PR #3705, opened 2016-08-04
Fixes #12488 - Add admin flag to usergroup param group description

mbacovsky:
To verify run 
`hammer -r user-group create --help`
and check the --admin option is in the output.
You may need to regenerate apipie cache on the server `rake apipie:cache:index`.

### DIFF
--- a/app/controllers/api/v2/usergroups_controller.rb
+++ b/app/controllers/api/v2/usergroups_controller.rb
@@ -22,6 +22,7 @@ module Api
       def_param_group :usergroup do
         param :usergroup, Hash, :required => true, :action_aware => true do
           param :name, String, :required => true
+          param :admin, :bool, :required => false, :desc => N_("is an admin user group")
           param :user_ids, Array, :require => false
           param :usergroup_ids, Array, :require => false
           param :role_ids, Array, :require => false


mbacovsky:
Dear reviewer, could you please advise me how to clear the codeclimate issues? Are there any related docs/guidelines? It seems these are false positives.

app/controllers/api/v2/usergroups_controller.rb
```diff
@@ -22,6 +22,7 @@ def show
       def_param_group :usergroup do
         param :usergroup, Hash, :required => true, :action_aware => true do
           param :name, String, :required => true
+          param :admin, :bool, :required => false, :desc => N_("is an admin group-account")
```
domcleal: The "group-account" in the description is unclear to me, I'd suggest "is an admin user group", "has admin access" or similar.

domcleal:
I don't think the Code Climate results are meaningful here, please ignore them. The comment by @theforeman-bot needs to be resolved though.

mbacovsky:
- description updated
- commit message was shortened

domcleal:
Merged as 7f8a818, thanks @mbacovsky.

I swapped it to the corresponding Foreman core ticket, the assigned Redmine ticket was in the Hammer CLI project.

PR #3659, opened 2016-07-19
fixes #3917 - replace protected_attrs with strong parameters

domcleal:
Filtering of attributes has moved from the protected_attributes gem to
strong parameters in controller concerns, to be in line with current
Rails recommendations.

Concerns are shared between UI and both API controllers and list the
attributes using Foreman::ParameterFilter, which provides additional
features:
1. Registration of additional attributes from plugins, through the
   plugin API or `attr_accessible` for short term compatibility.
2. Re-use of permitted attribute lists for nested models (e.g. host
   and interface).
3. Combining of lists of attributes from all sources in a single
   permit call.
4. A small DSL for changing accepted parameters based on controller,
   action and UI/API type.

Plugins should either temporarily depend on protected_attributes to
continue to protect their models or call permit/use ParameterFilter
similarly: http://projects.theforeman.org/projects/foreman/wiki/Strong_parameters

Some UI changes were required to make nested model hash keys all
integers instead of "new_123456" etc, else strong parameters would
filter the entries out.

The Role model's builtin default has been moved from initialize to the
database as the removal of protected_attrs changed the initialisation
order.

PR #3706, opened 2016-08-04
Fixes #15108 - allow editing cloned roles

ares:


dLobatog:
@ares Thanks, it works well :smile: I'm wondering why not using a migration to ensure `builtin` has a valid value?

ares:
Thanks for looking, that was my original thought but it seems that schema changes are not accepted for stable branches (and this is meant to be added to stable 1.12). In develop it's already fixed by https://github.com/theforeman/foreman/pull/3659/files#diff-dbdb4ceb4a2d790f9001a3bc2be67da1R3

dLobatog:
I see, I don't recall ever merging patches specific for a minor version outside CVE fixes - @domcleal OK to merge this ?

ares:
I think we do it regularly for regressions introduced in current or previous branch. Since the impact is quite big for users who actively use roles I think this should get to 1.12 branch.

domcleal:
@dLobatog fine by me, so long as it was addressed in develop. Having a different patch is OK to minimise changes in the stable branch.

dLobatog:
Merged as 919cde52ef017819767dabd0424a6c6bb6c9e788 for 1.12-stable, thanks !

PR #3707, opened 2016-08-04
fixes #15989 - add clone host button to host view

timogoebel:


dLobatog:
Merged as 1e9d49a7eb748598ee4cf99cf31a655b9b6d523c, thanks @timogoebel!

dLobatog:
By the way, we should think of a better layout for the host view, based on patternfly. It's starting to look cramped with several plugins adding buttons there

timogoebel:
@dLobatog : Yes, I agree. I do like the general design, though. We should keep it "Foreman".

PR #3708, opened 2016-08-05
Fixes #15996 - Multiple nav bars

anandkumaragrawal:
Multiple navigation bars (top bar starting with "Default Organization") on screen

In Firebug Console: Uncaught TypeError: Cannot read property 'hide' of undefined

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -457,9 +457,9 @@ function filter_permissions(item){
 }
 
 function setPowerState(item, status){
+  var power_actions = $('#power_actions');
   if(status=='success') {
     var place_holder = $('#loading_power_state').parent('.btn-group');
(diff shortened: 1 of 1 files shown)


timogoebel:
[test]

dLobatog:
Can you really reproduce this? The instructions just say go to 'all hosts' and click on a host, which clearly isn't enough to reproduce it.

mmoll:
That's probably the same as http://projects.theforeman.org/issues/15956

anandkumaragrawal:
@dLobatog, Yes i can reproduce that  when Foreman cannot find VM, it is similar to the one as mentioned by @mmoll

dLobatog:
Merged as 42c1cf29ff5702f9d0d483afa084194e7d2a7982, thanks @anandkumaragrawal!

PR #3709, opened 2016-08-05
Fixes #15232 - Showing inherited roles from User groups to user's tab

dhlavac:
Added showing inherited roles from User groups to user's tab

### DIFF
--- a/app/views/api/v2/users/show.json.rabl
+++ b/app/views/api/v2/users/show.json.rabl
@@ -2,6 +2,13 @@ object @user
 
 extends "api/v2/users/main"
 
+child @user.cached_usergroups.uniq => :cached_usergroups do
+  extends "api/v2/usergroups/base"
+  child :roles do
+    extends "api/v2/roles/base"
+  end
+end
+
 child :auth_source do
   extends "api/v2/auth_source_ldaps/base"
 end
@@ -16,6 +23,9 @@ end
 
 child :usergroups do
   extends "api/v2/usergroups/base"
+  child :roles do
+      extends "api/v2/roles/base"
+  end
 end
 
 node do |user|
--- a/app/views/users/_form.html.erb
+++ b/app/views/users/_form.html.erb
@@ -73,12 +73,49 @@
     <div class='tab-pane' id='roles'>
       <%= checkbox_f f, :admin if User.current.can_change_admin_flag? %>
       <%= multiple_checkboxes f, :roles, @user, Role.givable.for_current_user,
-                               { :label => _('Roles')}, {:disabled => @editing_self ? Role.givable.for_current_user.pluck(:id) : false } %>
+                              { :label => _('Roles')}, {:disabled => @editing_self ? Role.givable.for_current_user.pluck(:id) : false } %>
+      <% usergroups = @user.cached_usergroups.includes(:roles).uniq %>
+      <% if usergroups.any? %>
+        <div class="form-group">
+          <label class="col-md-2 control-label" for="roles"><%= _("Roles from user groups") %></label>
+          <div class="col-md-5">
+            <div class="dropdown">
+              <button class="btn btn-default dropdown-toggle" type="button" id="usergroupsDropdownMenuBtn" data-toggle="dropdown">
+                <%= usergroups.first %>
+                <span class="caret"></span>
+              </button>
+              <ul class="dropdown-menu" role="menu" aria-labelledby="usergroupsDropdownMenuBtn">
+                <% usergroups.each do |usergroup| %>
+                    <li role="presentation"><a role="menuitem" tabindex="-1" onclick=tfm.tools.userInheritedRolesUpdater(<%=usergroup.id %>)><%= usergroup %></a></li>
+                <% end %>
+              </ul>
+            </div>
+            <ul class="list-group" id='roles_tab'>
+              <% usergroups.each do |usergroup|  %>
+                  <% unless usergroup.roles.map(&:name).any? %>
+                      <li data-id="<%= usergroup.id %>" class="list-group-item"><%= _('This group has no roles') %></li>
+                  <%end %>
+                <% usergroup.roles.map(&:name).each do |role_name| %>
+                  <li data-id="<%= usergroup.id %>" class="list-group-item hidden"><%= role_name %></li>
+                <% end %>
+              <% end %>
+            </ul>
+          </div>
+        </div>
+      <% end  %>
     </div>
     <%= render 'taxonomies/loc_org_tabs', :f => f, :obj => @user,
                :html_options => user_taxonomies_html_options(@user)
-      %>
+    %>
   </div>
 
   <%= submit_or_cancel f %>
 <% end %>
+
+<% if @user.cached_usergroups.any? %>
+  <script>
+      $(function() {
+        tfm.tools.userInheritedRolesUpdater(<%= @user.cached_usergroups.uniq.first.id %>)
+      });
+  </script>
+<% end  %>
--- a/webpack/assets/javascripts/foreman_tools.js
+++ b/webpack/assets/javascripts/foreman_tools.js
@@ -35,3 +35,11 @@ export function activateDatatables() {
     });
   });
 }
+
+export function userInheritedRolesUpdater(id) {
+  $('#roles_tab li').hide();
+  $("#roles_tab li[data-id = '" + id + "']").show();
+  $('.dropdown-menu li a').click(function () {
+    $(this).parents('.dropdown').find('.btn').html($(this).text() + ' <span class="caret"></span>');
+  });
+}
--- a/app/assets/stylesheets/users.scss
+++ b/app/assets/stylesheets/users.scss
@@ -22,3 +22,8 @@
     font-size: 10px;
   }
 }
+
+.list-group{
+  max-height: 200px;
+  overflow-y:scroll;
+}


ares:
[test]

ares:
Another part of the issue was adding the list to the API/hammer output.

app/controllers/users_controller.rb
```diff
@@ -9,7 +9,7 @@ class UsersController < ApplicationController
   after_action       :update_activity_time, :only => :login
 
   def index
-    @users = User.authorized(:view_users).except_hidden.search_for(params[:search], :order => params[:order]).includes(:auth_source, :cached_usergroups).paginate(:page => params[:page])
+    @users = User.authorized(:view_users).except_hidden.search_for(params[:search], :order => params[:order]).includes(:auth_source, :cached_usergroups).paginate(:page => params[:page]).includes(:cached_usergroups)
```
dLobatog: This is already including `:cached_usergroups` - see the  method call before `paginate`

app/views/users/_form.html.erb
```diff
@@ -69,12 +69,28 @@
     <div class='tab-pane' id='roles'>
       <%= checkbox_f f, :admin if User.current.can_change_admin_flag? %>
       <%= multiple_checkboxes f, :roles, @user, Role.givable.for_current_user,
-                               { :label => _('Roles')}, {:disabled => @editing_self ? Role.givable.for_current_user.pluck(:id) : false } %>
+                              { :label => _('Roles')}, {:disabled => @editing_self ? Role.givable.for_current_user.pluck(:id) : false } %>
+      <div class="form-group">
+        <% if @user.cached_usergroups.uniq.any? %>
```
dLobatog: Please make this `if` condition show/hide the entire div, not just the title. Also use `.usergroups`, not `.cached_usergroups` - and no need for `.uniq` as far as I know

app/views/users/_form.html.erb
```diff
@@ -69,12 +69,28 @@
...
+            <label class="col-md-2 control-label" for="roles">Roles inherited from user groups</label>
+        <% end  %>
+        <div class="col-md-5">
+          <% @user.cached_usergroups.uniq.each do |usergroup| %>
```
dLobatog: Ditto, `.usergroups`

app/views/users/_form.html.erb
```diff
@@ -69,12 +69,28 @@
...
+                <h5><%= link_to_if_authorized trunc_with_tooltip(usergroup.name), hash_for_edit_usergroup_path(:id => usergroup.id).merge(:auth_object => usergroup, :authorizer => authorizer) %></h5>
+              <div class="row">
+                <div class="col-lg-offset-1">
+                  <%=usergroup.roles.map(&:name).join(', ')%>
```
dLobatog: Nitpick, missing spaces between the tags and the code inside

app/views/users/_form.html.erb
```diff
@@ -69,12 +69,28 @@
     <div class='tab-pane' id='roles'>
       <%= checkbox_f f, :admin if User.current.can_change_admin_flag? %>
       <%= multiple_checkboxes f, :roles, @user, Role.givable.for_current_user,
-                               { :label => _('Roles')}, {:disabled => @editing_self ? Role.givable.for_current_user.pluck(:id) : false } %>
+                              { :label => _('Roles')}, {:disabled => @editing_self ? Role.givable.for_current_user.pluck(:id) : false } %>
+      <div class="form-group">
+        <% if @user.cached_usergroups.uniq.any? %>
+            <label class="col-md-2 control-label" for="roles">Roles inherited from user groups</label>
```
dLobatog: Nitpick, if you can shorten it to `Roles from user groups` that'd be good, it looks kind of odd now
screenshot from 2016-08-08 08-06-46
screenshot from 2016-08-08 08-07-24

dLobatog:
@dhlavac Thanks, there are a few comments I left inline but overall it's interesting info to have in this tab plus the implementation is easy. 

We try to follow Patternfly as to provide users the same kind of look&feel in all enterprise applications, I looked at the patterns trying to find something that would fit this, as the plan list of words is a little dry. Can you look into displaying groups as [an accordion](http://www.patternfly.org/pattern-library/widgets/#accordion)? Every group would display the roles upon click, it seems a bit more user friendly and pleasing to the eye, not sure what you think.

Thanks again!

ares:
I'll let @dhlavac to respond to other comments, just one reaction from my side.

> Can you look into displaying groups as an accordion?

We tried that and it's not a good fit, user typically don't have many user groups and user group also does not have many roles so clicking to see what roles are assigned is extra burden. I'd be fine with restyling the list but I'd prefer to see the list of roles right away.

app/views/users/_form.html.erb
dhlavac: It doesn't work when using only `.usergroups`. Because it doesn't show usergroups and roles from the other usergroups when one usergroup contains another one.

app/views/users/_form.html.erb
dhlavac: Ditto, as previous comment.

app/views/users/_form.html.erb
ares: Note that the `cached_usergroups` relation contains all user groups including those that are part of directly associated user groups. That's why we have the cache association there. They might be duplicated since multiple user groups can contain the same user group which is why we need `uniq`.

ares:
[test]

app/views/api/v2/users/show.json.rabl
```diff
@@ -16,6 +23,9 @@ end
 
 child :usergroups do
   extends "api/v2/usergroups/base"
+  child :roles do
+      extends "api/v2/roles/base"
+    end
```
ares: indentation is off

ares:
thanks, you need to rebase it to get rid of hound-ci bot

dhlavac:
@ares Thank you

ares:
[test foreman]

dLobatog:
@rohoover Mind to take a look at this?

We're looking for the best way to display the roles that an user has, including those coming from the user group. The form needs to allow users to have their own roles that are set no matter if it's on an user group or not, so the duplicity you see in the "roles" list above is fine.

screenshot from 2016-09-06 11-30-29

Thanks!

Rohoover:
If the main issue is displaying the "Roles from user groups", it would be nice to use a list view with a drop down. The drop down being the group the user would wish to see roles for. This might save "space" for users who are in multiple groups such as the example above. The list view (or perhaps a table) would give a clearer view of all the roles assigned vs in paragraph format where it's difficult to breeze through the information. Let me know if you guys would like a wireframe or high fidelity mockup for this.

dhlavac:
@Rohoover can you send me a wireframe ?

Rohoover:
@dhlavac Sure. Give me a little bit. In an all day meeting, but I'll hack away at it during the meetings.

Rohoover:
Here is a mockup. Changing the drop down changes the roles within the list group. Add the scroll if necessary. The fixed height of the role area will reduce any shifts in the content below from user groups that have lots of roles. I did not include the outline of the box because there are already two outlined boxes at the top, adding a third may not be ideal for visual hierarchy. Let me know what you think or any feedback.
userroledisplay

ares:
Ok, works for me. I don't think user will have more than few user groups and user group would have more than few roles. If the select box will load first (and I think usually the only) user group automatically it will show most important info right away and will work with more complicated setups too. @dhlavac could you please incorporate patternfly [dropdowns](http://www.patternfly.org/pattern-library/widgets/#dropdowns) and [lists](http://www.patternfly.org/pattern-library/widgets/#list-group)?

dhlavac:
@ares sure and @Rohoover thank you for wireframe

dhlavac:
@ares @dLobatog @Rohoover can you look at that ?

webpack/assets/javascripts/foreman_tools.js
```diff
@@ -35,3 +35,11 @@ export function activateDatatables() {
     });
   });
 }
+
+export function UserInheritedRolesUpdater(id) {
```
ares: function name should start with small `u` I think

webpack/assets/javascripts/foreman_tools.js
```diff
@@ -35,3 +35,11 @@ export function activateDatatables() {
...
+  $("#roles_tab li[data-id = '" + id + "']").show();
+  $("#roles_tab li[data-id != '" + id + "']").hide();
```
ares: would it be better to first hide all `li`s and then show the right one? that way you wouldn't have to evaluate `!= id` for each `li` I think

something like

```
$("#roles_tab li]").hide();
$("#roles_tab li[data-id = '" + id + "']").show();
```

app/views/users/_form.html.erb
```diff
@@ -69,12 +69,46 @@
...
+    <%= submit_or_cancel f %>
 <% end %>
+
+<script>
+  <% if @user.cached_usergroups.any? %>
+    $(function() {
+      tfm.tools.UserInheritedRolesUpdater(<%= @user.cached_usergroups.uniq.first.id %>)
+    });
+  <% end  %>
+</script>
```
ares: nit: missing eol

app/views/users/_form.html.erb
```diff
@@ -69,12 +69,46 @@
...
-      %>
+    %>
   </div>
 
-  <%= submit_or_cancel f %>
+    <%= submit_or_cancel f %>
 <% end %>
+
+<script>
+  <% if @user.cached_usergroups.any? %>
```
ares: this should be wrapping the whole `<script>` IMHO

app/views/users/_form.html.erb
```diff
@@ -69,12 +69,46 @@
...
+      <% end  %>
     </div>
     <%= render 'taxonomies/loc_org_tabs', :f => f, :obj => @user,
                :html_options => user_taxonomies_html_options(@user)
-      %>
+    %>
   </div>
 
-  <%= submit_or_cancel f %>
+    <%= submit_or_cancel f %>
```
ares: this is probably not intentional

app/views/users/_form.html.erb
```diff
@@ -69,12 +69,46 @@
     <div class='tab-pane' id='roles'>
       <%= checkbox_f f, :admin if User.current.can_change_admin_flag? %>
       <%= multiple_checkboxes f, :roles, @user, Role.givable.for_current_user,
-                               { :label => _('Roles')}, {:disabled => @editing_self ? Role.givable.for_current_user.pluck(:id) : false } %>
+                              { :label => _('Roles')}, {:disabled => @editing_self ? Role.givable.for_current_user.pluck(:id) : false } %>
+      <% usergroups = @user.cached_usergroups.uniq %>
```
ares: I think `includes(:roles)` would help with (n+1) sql query issue on line 91 (haven't tested though)

app/views/users/_form.html.erb
```diff
@@ -69,12 +69,46 @@
...
+              </button>
+              <ul class="dropdown-menu" role="menu" aria-labelledby="dropdownMenu1">
+                <% usergroups.each do |usergroup| %>
+                    <li role="presentation"><a role="menuitem" tabindex="-1" onclick=tfm.tools.UserInheritedRolesUpdater(<%=usergroup.id %>)><%= usergroup %></a></li>
+                <% end %>
+              </ul>
+            </div>
+            <ul class="list-group" id='roles_tab'>
+              <% usergroups.each do |usergroup|  %>
+                <% usergroup.roles.map(&:name).each do |every_role| %>
```
ares: I think `role` or `role_name` as a variable name would be better

ares (changes requested):
I added few code related comments, otherwise it works and looks good.

ares (approved):

dhlavac:
@ares related changes - done

app/controllers/api/v2/users_controller.rb
```diff
@@ -23,6 +23,7 @@ class UsersController < V2::BaseController
       param :auth_source_ldap_id, String, :desc => N_("ID of LDAP authentication source")
       param :usergroup_id, String, :desc => N_("ID of user group")
       param :role_id, String, :desc => N_("ID of role")
+      param :cached_usergroups, Hash, :desc => N_("List all user groups included inherited user groups")
```
tstrachota: @dhlavac what is this parameter for?

app/views/users/_form.html.erb
```diff
@@ -73,12 +73,46 @@
...
       <%= multiple_checkboxes f, :roles, @user, Role.givable.for_current_user,
-                               { :label => _('Roles')}, {:disabled => @editing_self ? Role.givable.for_current_user.pluck(:id) : false } %>
+                              { :label => _('Roles')}, {:disabled => @editing_self ? Role.givable.for_current_user.pluck(:id) : false } %>
+      <% usergroups = @user.cached_usergroups.includes(:roles).uniq %>
+      <% if usergroups.any? %>
+        <div class="form-group">
+          <label class="col-md-2 control-label" for="roles">Roles from user groups</label>
+          <div class="col-md-5">
+            <div class="dropdown">
+              <button class="btn btn-default dropdown-toggle" type="button" id="dropdownMenu1" data-toggle="dropdown">
```
tstrachota: Could you please give it some more meaningful id? Eg. `usergroupsDropdownBtn` or some other less generic name.

app/views/users/_form.html.erb
```diff
@@ -73,12 +73,46 @@
...
+                <%= usergroups.first %>
```
tstrachota, line 84: If you wrap this name with a span and give it a class or id, you can avoid inserting `<span class="caret"></span>` on each click.

tstrachota:
Works fine, @dhlavac I added couple of nitpick notes to the code.

One more thing that might by worth improving is how groups with no roles are displayed:
screenshot from 2016-11-15 13 46 02

I understand that it's technically correct behaviour - it displays empty list, but I'm afraid that it might appear broken to users. What if we added text "This group has no roles" or something similar instead?
@ares @Rohoover what do you think?

Rohoover:
@tstrachota 

Yes, I think having that type of messaging would be good.

app/views/users/_form.html.erb
dhlavac: I already tried that, but it doesn't work properly

app/controllers/api/v2/users_controller.rb
dhlavac: Actually I don't remember, but I thing it is from last commit and I forgot delete it

dhlavac:
@tstrachota added text to list when usergroup have no roles and little changes that you mentioned

app/views/users/_form.html.erb
```diff
@@ -73,12 +73,49 @@
...
       <%= multiple_checkboxes f, :roles, @user, Role.givable.for_current_user,
-                               { :label => _('Roles')}, {:disabled => @editing_self ? Role.givable.for_current_user.pluck(:id) : false } %>
+                              { :label => _('Roles')}, {:disabled => @editing_self ? Role.givable.for_current_user.pluck(:id) : false } %>
+      <% usergroups = @user.cached_usergroups.includes(:roles).uniq %>
+      <% if usergroups.any? %>
+        <div class="form-group">
+          <label class="col-md-2 control-label" for="roles"><%= _("Roles from user groups") %></label>
+          <div class="col-md-5">
+            <div class="dropdown">
+              <button class="btn btn-default dropdown-toggle" type="button" id="dropdownMenu1" data-toggle="dropdown">
```
tstrachota: @dhlavac I guess you forgot to change `dropdownMenu1` to `usergroupsDropdownMenuBtn` here.

dhlavac: @tstrachota thank you

ares:
Foreman failure unrelated, dashboard integration on one combination. Thanks to all participants and of course extra thanks to @dhlavac. Merged as 2a3ede3c19f058a65088cb4a5a9df6fdda6f50f9

PR #3710, opened 2016-08-08
Fixes #13813 - Removed "\ " so ruby 1.8 could read the ENC

ShimShtein:


app/controllers/hosts_controller.rb
```diff
@@ -197,7 +197,7 @@ def externalNodes
     begin
       respond_to do |format|
         format.html { render :text => "<pre>#{ERB::Util.html_escape(@host.info.to_yaml)}</pre>" }
-        format.yml { render :text => @host.info.to_yaml }
+        format.yml { render :text => @host.info.to_yaml(:line_width => -1) }
```
iNecas: we should do this both fore html and yml version so have identical results

ShimShtein:
Fixed commit message and added fix for html output.

app/controllers/hosts_controller.rb
```diff
@@ -196,8 +196,8 @@ def externalNodes
 
     begin
       respond_to do |format|
-        format.html { render :text => "<pre>#{ERB::Util.html_escape(@host.info.to_yaml)}</pre>" }
-        format.yml { render :text => @host.info.to_yaml }
+        format.html { render :text => "<pre>#{ERB::Util.html_escape(@host.info.to_yaml(:line_width => -1))}</pre>" }
+        format.yml { render :text => @host.info.to_yaml(:line_width => -1) }
```
iNecas: i would suggest extracting the yaml into variable to reduce code duplicities. Good to merge after that.

ShimShtein:
Removed duplication.

iNecas:
Tested and it works well, Thanks @ShimShtein, merged as 0069c98dcb50a8960f92397921cfb8a98dc8401e

PR #3712, opened 2016-08-09
Fixes #16020 - Prevent reflective XSS on form validation

tbrisker:
Error messages for various form fields were not properly escaped to
prevent HTML from being insert into them. This caused a possible
reflective XSS in smart class parameter/varaible default value
validations.

### DIFF
--- a/app/helpers/form_helper.rb
+++ b/app/helpers/form_helper.rb
@@ -265,7 +265,7 @@ module FormHelper
   end
 
   def help_inline(inline, error)
-    help_inline = error.empty? ? inline : content_tag(:span, error.to_sentence.html_safe, :class => 'error-message')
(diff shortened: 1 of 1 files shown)


domcleal:
To the reviewer: please note that this was reported to foreman-security, though it appears to have no noticeable security impact, as the XSS is not stored. For this reason, no CVE will be requested, it's just a preventative measure.

dLobatog:
I looked for instances of `help_inline` and validations where we would call it with an error with HTML in it, didn't find any. It seems like the `.html_safe` call has been there for ages (0fe08ca9fe158e2cde30a15d2141a76c1de83e79) 

We call `.html_safe` on the result of `help_inline` on `add_help_to_label` too but I don't think it's a problem now. Merging

dLobatog:
Merged as cf0ce6d763014a0d637e402dc2266554b97beaa7, thanks @tbrisker!

PR #3713, opened 2016-08-09
refs #13768 - add random_name call to create_hosts permission

domcleal:


timogoebel:
Merged as 4a811e630eb93e4b38e9aaca5b0fb8dbe124557a, thanks @domcleal!

PR #3714, opened 2016-08-09
Fixes #16022 - Prevent stored XSS in host interface form

tbrisker:
The host interface form may contain a stored XSS in the identifier field
allowing a user allowed to edit a host's interfaces to cause code
execution by another user viewing that host's edit form.


dLobatog:
Merged as 53081ea14b30d66f0d67b62fe950a2c1463225f5, thanks @tbrisker!

PR #3716, opened 2016-08-09
Fixes #16025 - Fix ERB & params to import subnets from proxy

dLobatog:
Before this commit, the ERB would just throw a 500 because it tried to
show `@subnet.name` in a template for a `@subnet` that was still undefined.

Aside from that, the create_multiple method in the controller didn't
send the right call to filter the params, hence making it impossible to
create the imported subnets even with the ERB fix.

### DIFF
--- a/app/views/subnets/import.html.erb
+++ b/app/views/subnets/import.html.erb
@@ -1,19 +1,20 @@
 <% title _("Import IPv4 subnets") %>
 
-<h2><%= _("The following IPv4 subnets have been found") %></h2>
-
+<h4><%= _("The following IPv4 subnets have been found. Please review the details before creating them.") %></h4>
 <%= form_for "subnets[]", :url => create_multiple_subnets_path do |f| %>
   <% display_all = !minimal?(@subnets) %>
-  <div class="accordion" id='accordion1'>
+  <div class="panel-group" id='accordion1'>
   <% @subnets.each do |subnet| %>
-    <% id = 'subnet_fields_'+subnet.to_s.gsub('/','_').gsub('.','_') %>
-    <div class="accordion-group">
-      <div class="accordion-heading" >
-        <a class="accordion-toggle" data-toggle="collapse" data-parent='#accordion1' href="<%= '#'+id %>" data-original-title='review' rel='twipsy' > <%= subnet %> </a>
-        <%= link_to_function "x", "ignore_subnet(this)", :'data-original-title'=> 'ignore subnet', :rel=>'twipsy', :class => "label ignore-subnet" %>
+    <% id = 'subnet_fields_' + subnet.network_address.gsub('/','_').gsub('.','_') %>
+    <div class="panel panel-default">
+      <div class="panel-heading" >
+        <a class="panel-title" data-toggle="collapse" data-parent='#accordion1' href="<%= '#'+id %>" rel='twipsy' >
+          <%= subnet.network_address %>
+        </a>
+        <%= link_to_function "x", "ignore_subnet(this)", :'data-original-title'=> 'Ignore subnet', :rel=>'twipsy', :class => "label ignore-subnet" %>
       </div>
-      <div id="<%= id %>" class="accordion-body collapse <%= 'in' if display_all %>" >
-        <div class="accordion-inner">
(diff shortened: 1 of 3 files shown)


app/views/subnets/import.html.erb
```diff
@@ -1,19 +1,19 @@
...
+  <div class="panel-group" id='accordion1'>
   <% @subnets.each do |subnet| %>
-    <% id = 'subnet_fields_'+subnet.to_s.gsub('/','_').gsub('.','_') %>
-    <div class="accordion-group">
-      <div class="accordion-heading" >
-        <a class="accordion-toggle" data-toggle="collapse" data-parent='#accordion1' href="<%= '#'+id %>" data-original-title='review' rel='twipsy' > <%= subnet %> </a>
+    <% id = 'subnet_fields_' + subnet.network_address.gsub('/','_').gsub('.','_') %>
+    <div class="panel panel-default">
+      <div class="panel-heading" >
+        <a class="panel-title" data-toggle="collapse" data-parent='#accordion1' href="<%= '#'+id %>" data-original-title='review' rel='twipsy' >
```
domcleal: data-original-title should probably be removed here too, it just shows an untranslated tooltip saying "review" and isn't very helpful!

app/views/subnets/import.html.erb
```diff
@@ -1,19 +1,19 @@
 <% title _("Import IPv4 subnets") %>
 
-<h2><%= _("The following IPv4 subnets have been found") %></h2>
```
domcleal, line 3: This was probably still useful, and could even be expanded to say to review the details in the accordion below.

domcleal:
Could you also add a functional test for the import action itself, to check the page renders without errors based on some mocked proxy response?

test/functional/subnets_controller_test.rb
```diff
@@ -55,77 +55,117 @@ def test_destroy
     assert !Subnet.exists?(subnet.id)
   end
 
-  def test_freeip_fails_no_subnet
```
dLobatog, line 58: This is just makeup, nothing changed

test/functional/subnets_controller_test.rb
```diff
@@ -55,77 +55,117 @@ def test_destroy
...
+
+    test 'without view_params user should not see parameters in a subnet' do
+      setup_user "edit", "subnets"
+      subnet = FactoryGirl.create(:subnet_ipv4, :with_parameter)
+      get :edit, {:id => subnet.id}, set_session_user.merge(:user => users(:one).id)
+      assert_nil response.body['Parameter']
+    end
+  end
+
+  context 'import IPv4 subnets' do
```
dLobatog, line 124: Here start the changes

dLobatog:
Fixed - basically I added two tests for the import action in the controller & added back the h2 warning. Thanks for reviewing @domcleal !

app/views/subnets/import.html.erb
```diff
@@ -1,19 +1,20 @@
...
-
+<h2><%= _("The following IPv4 subnets have been found. Please review the details before creating them.") %></h2>
```
domcleal: Could you drop this to regular paragraph text, or perhaps try h3 etc? The text is very nearly the same size as the main title.

dLobatog:
Updated to decrease the font-size of the subtitle to h3 , also added col-md-8 so that the h3 tag is aligned with the title and the rest of the form

app/views/subnets/import.html.erb
```diff
@@ -1,19 +1,20 @@
 <% title _("Import IPv4 subnets") %>
 
-<h2><%= _("The following IPv4 subnets have been found") %></h2>
-
+<h4 class='col-md-8'><%= _("The following IPv4 subnets have been found. Please review the details before creating them.") %></h4>
```
domcleal: The `col-md-8` adds a `float: left` to this element which causes it to overlap:

selection_003

dLobatog:
Updated to remove the col-md-8 - sorry, I was trying it with the web inspector and I guess I disabled something and didn't realize, so the col-md looked fine.

domcleal:
[test] due to failed slave.

domcleal:
[test]

domcleal:
Merged as 4622318, thanks @dLobatog.

PR #3717, opened 2016-08-09
Fixes #16030 - Add foreman_cache_data to foreman-debug

chris1984:


dLobatog:
@chris1984 Thanks! This won't work with Puppet AIO though (it's , can you retrieve the lib directory via `puppet config print libdir`, or detect whether we have AIO vs regular puppet?

domcleal:
This directory will likely contain passwords and tokens which foreman-debug deliberately attempts _not_ to collect. The list of files should probably be filtered to exclude all credentials in this directory.

domcleal:
Did you mean to close this?

chris1984:
@domcleal sorry closed the wrong one

chris1984:
@domcleal This should filter out any passwords etc
https://github.com/theforeman/foreman/blob/develop/script/foreman-debug#L42-L55

domcleal:
> This should filter out any passwords etc
> https://github.com/theforeman/foreman/blob/develop/script/foreman-debug#L42-L55

The contents of the files are YAML strings without any of those keywords - the filter matches key/value type structures, filtering out any keys that match the known words.  Almost all files in this directory are likely to be sensitive, are they not?

chris1984:
Closing this out since its not needed

PR #3719, opened 2016-08-10
fixes #16046 - run webpack:compile as pre-req to test rake tasks

domcleal:
The webpack:compile task is now run before all rake test tasks,
including jenkins:*. Except for jenkins:integration, a failure in
webpack:compile is issued as a warning as unit tests don't require
webpack assets to run.

Using webpack:compile from the task definitions avoids the need to load
tasks from railties in test helpers and keeps rake task dependencies
defined together.


dLobatog:
Merged as 55de1d1cab2a70f5f94f4218f5f845c6b5463277, thanks @domcleal!

PR #3720, opened 2016-08-10
Fixes #16055 - Have webpack-dev-server bind to 0.0.0.0

daviddavis:
I run foreman inside a VM and am unable to connect to the webpack-dev-server unless it's bound to 0.0.0.0 or the specific external ip.


daviddavis:
Nevermind. This doesn't work unless I also edit config/initializers/secure_headers.rb.

PR #3721, opened 2016-08-11
Refs #12364 - Allow running webpack dev server on any hostname

tbrisker:
Currently, SecureHeaders only allows running the webpack dev server from
localhost or 0.0.0.0, and only using http. This leads to issues when
attempting to run foreman inside a virtual machine or using https. This
patch will generate the correct headers during runtime, to allow the dev
server to be run on the same host and protocol as the request uses.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -438,8 +438,10 @@ module ApplicationHelper
 
   def webpack_dev_server
     return unless Rails.env.development?
-    host = ::Rails.configuration.webpack.dev_server.host.call
     port = ::Rails.configuration.webpack.dev_server.port
-    javascript_include_tag "http://#{host}:#{port}/webpack-dev-server.js"
+    dev_server = "#{request.protocol}#{request.host}:#{port}"
+    response.headers['Content-Security-Policy']['script-src'] = "script-src #{dev_server}"
+    response.headers['Content-Security-Policy']['wss:'] = "wss: #{dev_server}"
+    javascript_include_tag "#{dev_server}/webpack-dev-server.js"
   end
 end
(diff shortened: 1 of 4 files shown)


tbrisker:
@ehelms @thomasmckay I know you hit issues with this, could you please give this a spin?

app/helpers/application_helper.rb
```diff
@@ -438,8 +438,10 @@ def hosts_count(resource_name = controller.resource_name)
 
   def webpack_dev_server
     return unless Rails.env.development?
-    host = ::Rails.configuration.webpack.dev_server.host.call
     port = ::Rails.configuration.webpack.dev_server.port
-    javascript_include_tag "http://#{host}:#{port}/webpack-dev-server.js"
+    dev_server = "#{request.protocol}#{request.host}:#{port}"
+    response.headers["Content-Security-Policy"].gsub!("script-src","script-src #{dev_server}")
```
dLobatog: micro-optimization , use `response.headers['Content-Security-Policy']['script-src'] = "script-src #{dev_server}"` - https://github.com/JuanitoFatas/fast-ruby#stringsub-vs-stringgsub-vs-string-code

ehelms:
:+1: This fixed my setup

app/helpers/application_helper.rb
tbrisker: Updated, also makes it more readable.

dLobatog:
:+1: if @thomasmckay can give it a spin and confirm it that'd be the best - it seems to work fine on centos7-devel (forklift)

thomasmckay:
per @daviddavis add host: '0.0.0.0' at https://github.com/theforeman/foreman/blob/develop/config/webpack.config.js#L80

tbrisker:
Moved the host 0.0.0.0 setting from the Procfile to the config file.

thomasmckay:
ack, thanks!

dLobatog:
Merged as 42ba771cc8bfc09257c15c5d388542d4d1b69358, thanks @tbrisker!

PR #3722, opened 2016-08-11
Refs #12364 - Only load the dev server js if it's enabled

tbrisker:


app/helpers/application_helper.rb
```diff
@@ -437,8 +437,8 @@ def hosts_count(resource_name = controller.resource_name)
   end
 
   def webpack_dev_server
-    return unless Rails.env.development?
-    port = ::Rails.configuration.webpack.dev_server.port
+    return unless Rails.env.development? && Rails.configuration.webpack.dev_server.enabled
```
domcleal: Should the `env.development?` check be removed here? If it simply returns based on `dev_server.enabled` then the environment config is authoritative.

tbrisker:
@domcleal changed to only check the configuration, and added a setting to allow easy disabling of the dev server for those that don't need it.

config/settings.yaml.example
```diff
@@ -17,6 +17,10 @@
 # Mark translated strings with X characters (for developers)
 :mark_translated: false
 
+# Disable the webpack development server (for developers)
+# Make sure to run `rake webpack:compile` if disabled.
+#:webpack_dev_server: false
```
domcleal: Nitpick, this should show the default/current value rather than being commented out as `false`, it'll be clearer. The comment above may need changing too.

tbrisker:
@domcleal updated the settings.yaml.example to reflect the default.

domcleal:
Merged as 31bbcd8, thanks @tbrisker.

PR #3723, opened 2016-08-11
Fixes #16073 - match up nic identifiers for unmanaged hosts

jlsherrill:


app/models/host/base.rb
```diff
@@ -349,11 +349,13 @@ def get_interface_scope(name, attributes, base = self.interfaces)
           rescue Net::Validations::Error
             logger.debug "invalid mac during parsing: #{attributes[:macaddress]}"
           end
-          base = base.where(:mac => macaddress)
+
+          nics = base.where(:mac => macaddress)
```
ares: I find this a bit more readable (not combining if in else branch)

``` ruby
mac_based = base.where(:mac => macaddress)
if attributes[:virtual]
  mac_based.virtual.where(:identifier => name)
elsif !self.managed?
  base.where(:identifier => name)
else
  mac_based.physical
end
```

one difference is that I don't even try `nics` and fallback to base which I think is also good, feel free to add it to elsif if you disagree

also please add a comment on line 343 why we handle unmanaged hosts differently

test/unit/host_test.rb
```diff
@@ -1276,6 +1276,21 @@ def teardown
...
+        host.set_interfaces(parser)
+      end
+
+      assert_equal '10.0.0.200', host.interfaces.where(:mac => '00:00:00:22:33:44').first.ip
+      assert_equal '2001:db8::2', host.interfaces.where(:mac => '00:00:00:22:33:44').first.ip6
+      assert_empty host.interfaces.where(:mac => '00:00:00:11:22:33')
+      refute host.interfaces.where(:mac => '00:00:00:22:33:44').first.link
+    end
+
+
```
ares: rubo:cop: is unhappy because of this line

app/models/host/base.rb
jlsherrill: I think we do need to fallback to 'base.physical' if we can't find it by name, but i do like your structure better.  Updating!

jlsherrill:
[test]

jlsherrill:
[test]

ares:
[test]

ares:
works nice for me, APJ

ares:
Merged as 7b75a6a018f560fba44916ded12d9fe8da92e820, thanks @jlsherrill!

PR #3724, opened 2016-08-12
Fixes #16082 - FactValue helper shouldn't assume valid .host

dLobatog:
As described on #16081 - the fact_values helper is expecting
that "fact_value.host" to return a meaningful value all the time.

However for certain hosts (DiscoveredHost now, but it could be others in
the future) /fact_values shows a 500 because "fact_value.host" doesn't
exist.

I think it'd be reasonable to expect fact_value.host to exist for Hosts
in core, but we should show N/A if that cannot be found. This way the
page will be more resilient to manual db manipulation, data corruption,
or plain bugs like the one mentioned above.

### DIFF
--- a/app/helpers/fact_values_helper.rb
+++ b/app/helpers/fact_values_helper.rb
@@ -13,9 +13,10 @@ module FactValuesHelper
       if parent.present? && h(parent.name) == memo
         current_name
       else
-        if value_name != memo || value.compose
+        if host_id(value) && (value_name != memo || value.compose )
           parameters = { :parent_fact => memo }
-          url = host_parent_fact_facts_path(parameters.merge({ :host_id => params[:host_id] || value.host.name }))
+          url = host_parent_fact_facts_path(
+            parameters.merge(:host_id => host_id(value)))
           link_to(current_name, url,
                   :title => _("Show all %s children fact values") % memo)
         else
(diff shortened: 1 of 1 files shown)


domcleal:
Seems to me that the issue is in the association or usage of the model (in associating a non-Host::Managed to FactValue), rather than the view.  Should the association be changed to Host::Base / all hosts, rather than just Host::Managed? I'd expect `fact_value.host` to resolve to a Host::Base then rather than returning `nil`.

Edit: submitted too early. Alternatively, we explicitly only support Host::Managed and Discovery is misusing this model/views.

dLobatog:
I workaround the issue (or fix it, depends how you look at it) on https://github.com/theforeman/foreman_discovery/pull/294 however I think of this PR as a preventative measure to not crash if you have an older version of Discovery or other plugins. Fine with me to close it - but I think it's a small change for potentially avoiding a 500

domcleal:
Would changing the association to Host::Base from Host::Managed ensure `fact_value.host` returns a Host::Base (derivative) without even needing a change in foreman_discovery?  I've not tested the issue, but that's my reading of cause from the ticket.

dLobatog:
Closing as it's not that relevant now that the other ticket was closed

dLobatog:
@domcleal Reopening this - I've experienced this crash again. To reproduce:
1. Create a discovery host and make the discovery image provide custom facts 
2. Provision said discovery host
3. These custom facts are now orphaned FactValues

domcleal:
> These custom facts are now orphaned FactValues

Shouldn't the host_id be unchanged when Host::Discovered _becomes_ Host::Managed? I'm not sure I understand what the patch is fixing.

I still think that the belongs_to on FactValue should probably be Host::Base rather than Host::Managed if Host::Discovered is using facts, as that would prevent `fact_value.host` finding the associated host.

PR #3725, opened 2016-08-12
fixes #16083 - pin gettext_i18n_rails for Ruby 2.0.0

domcleal:

PR #3726, opened 2016-08-12
fixes #16085 - update Rails to 4.2.7.1

domcleal:


domcleal:
The Ruby 2.0.0 and upgrade test failures are due to #3725. The Katello test failures are filed as  http://projects.theforeman.org/issues/16088 - @Katello/katello-core, please address.

dLobatog:
Merged as 3a59bbb72066d9d7a965fe49e40724fd55dee2c6, thanks @domcleal!

PR #3603, opened 2016-06-20
Fixes #13424 - c3 patternfly react implementation

gailsteiger:
The change can be seen on the statistics page.
This PR depends on the [Webpack PR](https://github.com/theforeman/foreman/pull/3433) which has been merged.

PR #3727, opened 2016-08-12
fixes #16089 - replace flot-rails with npm jquery-flot

domcleal:


tbrisker:
@domcleal Thanks, though I wonder if we should bother with this since flot.js is unmaintained and there is a PR in progress to replace flot with c3? (https://github.com/theforeman/foreman/pull/3603)

domcleal:
That PR doesn't remove flot-rails as a dependency, which is what I'm trying to fix in this issue. It's also only WIP, not in a reviewable state.

tbrisker:
@domcleal Very well. This works well, merging.

tbrisker:
Merged as 551407c293a9b60cc39e319d108f97e47dcb06b1, thanks @domcleal!

PR #3728, opened 2016-08-12
fixes #13759 - 500 when attempting to update a host

anandkumaragrawal:


app/controllers/concerns/api/compatibility_checker.rb
```diff
@@ -9,5 +9,11 @@ def check_create_host_nested
         Foreman::Deprecation.api_deprecation_warning("Field host_parameters_attributes.nested ignored") unless attribute.delete(:nested).nil?
       end
     end
+
+    def prevent_attributes
```
dLobatog, line 13: Bad naming IMO, how about `remove_capabilities`

anandkumaragrawal, line 13: @dLobatog I have made this method generic, so that in future, if we have more attributes like 'capabilities' that needs to be taken care. We can add that attribute as a part of the arrary. Hence named it as prevent_attributes.

Any thoughts?

anandkumaragrawal:
@domcleal : @dLobatog I got your point to name this method as 'remove_capabilities'. But as already mentioned the reason behind using 'prevent_attributes',  it would be good to use this generic name to remove keys from any given params in future. 

Let me know, if it really need to be changed to remove_capabilities?

ShimShtein:
How about `filter_params`?
@domcleal,it looks very similar to strong params, maybe we should use the same mechanism here?

xprazak2:
@anandkumaragrawal, could you rebase please? I like `filter_params` more than `prevent_attributes`...

amitkarsale:
[test]

amitkarsale:
can we have a status on closing or keeping open on this ticket?

dLobatog (changes requested):
@anandkumaragrawal Are you able to rebase so we can see if tests pass? I am not sure why this never got merged. Sorry for the extremely long delay

mmoll:
@amitkarsale what's the status here?

mmoll:
@anandkumaragrawal what's the status here?

PR #3730, opened 2016-08-12
Fixes #16075 - add authenticity token in setting params

amirfefer:
I can't reproduce the issue, but  it may caused by the lack of authenticity token in the settings params.


tbrisker:
Merged as b22a953a4feee3f23a47a8be6b9b79ed1daf8a67, thanks @amirfefer!

PR #3731, opened 2016-08-14
Fixes #16104 - Use parameterizable for HW models

tbrisker:
Correctly escape HW model names that may contain url-unsafe characters.
Also removed unnecessary truncation of the model name.

Resource finder now works correctly for models that may contain either a
friendly name or a parameterized id-name.

### DIFF
--- a/app/controllers/concerns/find_common.rb
+++ b/app/controllers/concerns/find_common.rb
@@ -11,7 +11,10 @@ module FindCommon
   def resource_finder(scope, id)
     raise ActiveRecord::RecordNotFound if scope.empty?
     result = scope.from_param(id) if scope.respond_to?(:from_param)
-    result ||= scope.friendly.find(id) if scope.respond_to?(:friendly)
+    begin
+      result ||= scope.friendly.find(id) if scope.respond_to?(:friendly)
+    rescue ActiveRecord::RecordNotFound
+    end
     result || scope.find(id)
   end
 
--- a/app/models/concerns/parameterizable.rb
+++ b/app/models/concerns/parameterizable.rb
@@ -27,7 +27,7 @@ module Parameterizable
       end
 
       def self.from_param(id_name)
-        self.find_by_id(id_name.to_i)
+        self.find_by_id(id_name.to_i) if id_name =~ /\A\d+-/
       end
     end
   end
--- a/app/views/models/index.html.erb
+++ b/app/views/models/index.html.erb
@@ -16,7 +16,7 @@
     <% @models.each do |model| %>
       <tr>
         <td class="display-two-pane ellipsis">
-          <%= link_to_if_authorized trunc_with_tooltip(model.name),
+          <%= link_to_if_authorized model.name,
             hash_for_edit_model_path(:id => model).merge(:auth_object => model, :authorizer => authorizer) %></td>
         <td class="ellipsis"><%= model.vendor_class %></td>
         <td class="ellipsis"><%= model.hardware_model %></td>
--- a/app/controllers/api/v2/models_controller.rb
+++ b/app/controllers/api/v2/models_controller.rb
@@ -49,12 +49,6 @@ module Api
       def destroy
         process_response @model.destroy
       end
-
-      private
-
-      def find_resource
-        @model = Model.friendly.find(params[:id]) || super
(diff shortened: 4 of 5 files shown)


app/controllers/concerns/find_common.rb
```diff
@@ -11,7 +11,7 @@ def find_resource
   def resource_finder(scope, id)
     raise ActiveRecord::RecordNotFound if scope.empty?
     result = scope.from_param(id) if scope.respond_to?(:from_param)
-    result ||= scope.friendly.find(id) if scope.respond_to?(:friendly)
+    result ||= scope.friendly.find(id) rescue nil if scope.respond_to?(:friendly)
```
tbrisker: This raises ResourceNotFound if the friendly find doesn't find anything, causing the `scope.find(id)` attempt to never run for models with friendly id.

app/models/concerns/parameterizable.rb
```diff
@@ -27,7 +27,7 @@ def to_param
       end
 
       def self.from_param(id_name)
-        self.find_by_id(id_name.to_i)
+        self.find_by_id(id_name.to_i) if id_name =~ /\A\d+-/
```
tbrisker, line 30: Only find by id if it is in the correct format - if the parameter is for example "1model", it will find the record with ID 1 even though it isn't the one we are looking for.

tbrisker:
[test]

app/controllers/concerns/find_common.rb
dLobatog: Can you write a functional test for a controller that has this problem? (Model?) 

Also if that's the exception that's causing the issue, rescue only that one, we still want to know if something different happened:

```
begin
 result ||= scope.friendly.find(id) if scope.respond_to?(:friendly)
rescue ActiveRecord::ResourceNotFound
end
result || scope.find(id)
```

or similar

app/views/models/index.html.erb
```diff
@@ -16,7 +16,7 @@
     <% @models.each do |model| %>
       <tr>
         <td class="display-two-pane ellipsis">
-          <%= link_to_if_authorized trunc_with_tooltip(model.name),
+          <%= link_to_if_authorized model.name,
```
dLobatog, line 19: Why?

app/models/concerns/parameterizable.rb
dLobatog, line 30: good catch!

dLobatog:
@tbrisker Thanks, please see the comment inline about the test & rescue, other than that :+1:

app/views/models/index.html.erb
tbrisker, line 19: This was actually a regression introduced in https://github.com/theforeman/foreman/commit/aa50ab15ff091f132be81861f03c6a9919f2410d#diff-d0a1a867ec446b231c4e11653e019a02R19, just happened to notice it when working on this issue.

app/controllers/concerns/find_common.rb
tbrisker: I actually found this issue because of this test: https://github.com/theforeman/foreman/blob/develop/test/functional/api/v2/models_controller_test.rb#L46, do you think we need another one?
I'm not sure if another exception might be raised, but I'm also not sure we care why friendly find failed-  we just fallback to find by id, and if the result is still empty after everything then a 404 will be returned anyways.

dLobatog: The point is that the inline rescue is also hiding errors that, 'scope', `scope.friendly` and `scope.friendly.find` could have, but the only one we're certain is legitimate is ResourceNotFound. Say 'scope.friendly' throws some other error because an update on friendly_id breaks it, we wouldn't know so easily as it falls back no matter what.

https://www.new-bamboo.co.uk/blog/2015/09/23/dont-inline-rescue-in-ruby/

dLobatog: Also don't worry about the test then, I was not aware of it, thanks

tbrisker: Fixed, thank you.

dLobatog:
Merged as a763e9764dfc1536936736abd6ea41d15155bdde, thanks @tbrisker!

PR #3704, opened 2016-08-04
Fixes #7806 - allow taxonomy association on role level

ares:

PR #3732, opened 2016-08-15
Fixes #12864 - verify host owner taxonomies

ares:


test/unit/host_test.rb
```diff
@@ -683,7 +683,21 @@ def teardown
...
+      loc1 = FactoryGirl.create(:location)
+      user = FactoryGirl.create(:user, :organizations => [], :locations => [])
+      host = FactoryGirl.build(:host, :organization => org1, :location => loc1, :owner => user)
+
+      SETTINGS[:locations_enabled] = true
+      SETTINGS[:organizations_enabled] = true
+      refute host.valid?
+      assert_operator 2, :<=, host.errors[:owner_id].size
+      SETTINGS[:locations_enabled] = false
+      SETTINGS[:organizations_enabled] = false
```
iNecas: I'm a bit puzzled here, in other tests we keep the taxonomies enabled in tests at the end, such as here https://github.com/theforeman/foreman/blob/develop/test/unit/host_test.rb#L607

iNecas:
Nice patch (given the subject is taxonomies:), one question inline

test/unit/host_test.rb
ares: in that link you can see it was disabled in before which is the pattern I copied, anyway I'll fix it by keeping original value which is better anyway (will fix the linked test too)

ares:
@iNecas fixed and calmed down :cop: please take another look

iNecas:
:+1: from me, leaving 24hours for others to speak up if they have something to improve here before merging

ares:
sorry, somehow I linked wrong redmine issue, I changed commit message

app/models/taxonomies/location.rb
```diff
@@ -15,8 +15,9 @@ class Location < Taxonomy
 
   scope :completer_scope, ->(opts) { my_locations }
 
-  scope :my_locations, lambda {
-    conditions = User.current.admin? ? {} : sanitize_sql_for_conditions([" (taxonomies.id in (?))", User.current.location_and_child_ids])
+  scope :my_locations, lambda { |user = nil|
```
dLobatog: Nitpick, move the `user ||= User.current` to the lambda args `|user =  User.current|`, if you want to provide a default I think that's the usual way

app/models/taxonomies/organization.rb
```diff
@@ -15,8 +15,9 @@ class Organization < Taxonomy
 
   scope :completer_scope, ->(opts) { my_organizations }
 
-  scope :my_organizations, lambda {
-    conditions = User.current.admin? ? {} : sanitize_sql_for_conditions([" (taxonomies.id in (?))", User.current.organization_and_child_ids])
+  scope :my_organizations, lambda { |user = nil|
```
dLobatog: https://github.com/theforeman/foreman/pull/3732/files#r74886627

app/models/host/managed.rb
```diff
@@ -926,6 +927,17 @@ def apply_compute_profile(modification)
 
   private
 
+  def owner_taxonomies_match
+    if self.owner.is_a?(User)
```
dLobatog: Move this `if` to the `validate` above, no point on going in the method if this condition is not met

app/models/user.rb
```diff
@@ -364,6 +364,14 @@ def set_current_taxonomies
     TopbarSweeper.expire_cache(self)
   end
 
+  def my_organizations
```
dLobatog, line 368: Aren't these methods nearly equivalent to `location_and_child_ids` and `organization_and_child_ids`? As far as I can see they return the object and the others the ids, but the id is enough in the query the validator does

app/models/host/managed.rb
```diff
@@ -926,6 +927,17 @@ def apply_compute_profile(modification)
...
+      if Organization.organizations_enabled && self.organization_id && !self.owner.my_organizations.where(:id => self.organization_id).exists?
```
dLobatog: `self.owner.organization_and_child_ids.include?(self.organization_id)` so you can avoid adding new methods in user.rb, location.rb and organization.rb

Why not implementing these through `Taxonomy.enabled_taxonomies` by the way? It's basically duplicated code as is now

test/unit/host_test.rb
```diff
@@ -598,13 +598,13 @@ def teardown
 
   context "location or organizations are not enabled" do
     before do
-      SETTINGS[:locations_enabled] = false
-      SETTINGS[:organizations_enabled] = false
+      @original_loc, SETTINGS[:locations_enabled] = SETTINGS[:locations_enabled], false
```
dLobatog: Why keeping track of the original value? We assume locs/orgs have to be enabled for test to pass, except for contexts like this where we explicitly disable it for a bit - and enable it right after

test/unit/host_test.rb
```diff
@@ -2058,6 +2058,20 @@ def to_managed!
     end
   end # end of context "location or organizations are not enabled"
 
+  test "should not save if owner taxonomies do not match host configuration" do
+    org1 = FactoryGirl.create(:organization)
+    loc1 = FactoryGirl.create(:location)
```
dLobatog: Is `create` necessary vs `.build`, or even nothing? Looks like `self.owner.organization_and_child_ids.include?(self.organization_id)` and `organization_id.present?` would resolve w/o creating the object

test/unit/host_test.rb
```diff
@@ -2058,6 +2058,20 @@ def to_managed!
...
+    user = FactoryGirl.create(:user, :organizations => [], :locations => [])
+    host = FactoryGirl.build(:host, :organization => org1, :location => loc1, :owner => user)
+
+    original_loc, SETTINGS[:locations_enabled] = SETTINGS[:locations_enabled], true
```
dLobatog: Same as before - assume they're true, it's fine

dLobatog:
Thanks @ares , please see the comments inline

app/models/host/managed.rb
ares: > self.owner.organization_and_child_ids.include?(self.organization_id) so you can avoid adding new methods in user.rb, location.rb and organization.rb

`User#my_organizations` uses `Organization.my_organizations` scope which reflects admin flag. If admin was not assigned to any org/loc (usually the case) this would always return false. `organization_and_child_ids` should only be used "internally" e.g. in `Organization.my_organizations` scope. Otherwise we'd had to add admin condition everywhere.

> Why not implementing these through Taxonomy.enabled_taxonomies by the way? It's basically duplicated code as is now

Since you asked [second time](https://github.com/theforeman/foreman/pull/3704#discussion_r74396509), I'll try to explain a bit more why I want to avoid it. I don't think this is duplication because
- Organization and Location are two different things. The implementation makes it easy to use them in same way leading to problems like when you later realize that Location should not act as Organization you'd have to split the logic everywhere. For example the taxonomy inheritance does not always make sense for both, should Subnet be shared with all sub-orgs? I'd say no. Should Subnet be shared with sub-locations? That makes more sense to me. I think it was mistake to use STI here. Similar code does not automatically mean duplication to me, if the logic os the same then it is. Since #my_organizations and #my_locations can potentially do different things I don't think it's duplication, only the name pattern is the same.
- metaprogramming is cool when you write it, pretty bad when you read it or try to grep source-code, I don't think the similarity here is justification to generate the code dynamically. The taxonomy code is already pretty hard to understand and big portion of this is because we use [this](https://github.com/theforeman/foreman/blob/develop/app/models/concerns/taxonomix.rb#L66)

The more of this we add the more we'll have to remove when we get to it. If someone later decides it's better to keep it this way, it's not too bad duplication since it's in one file near each other. I think worse duplication is that you can't spot easily (multiple methods, files etc).

app/models/user.rb
ares, line 368: they are not, https://github.com/theforeman/foreman/pull/3732#discussion_r74919424 (first part)

test/unit/host_test.rb
ares: I don't know how taxonomies settings change tomorrow for tests, this keeps tests less fragile, is there a reason to not? without keeping original value we could enable them even if they were disabled before this context

test/unit/host_test.rb
ares: same as before - this makes test less fragile

test/unit/host_test.rb
ares: we use `my_organizations` scope that does the query so we need it to exist in DB

Also one generic a bit off-topic comment for reviews, I don't think this is a way to systematically speed-up tests. It already happened that changing `create` to `build` made the test useless. If we want to have unit tests run in seconds we should avoid using ActiveRecord and do either full stubbing (yuck) or have separate layer for storage and business logic (yes please) which is unrealistic IMHO.

ares:
@dLobatog thanks, I fixed few things and replied to other comments

test/unit/host_test.rb
iNecas: +1 for keeping the original values: it makes the tests survive whatever the default is

app/models/taxonomies/organization.rb
iNecas: One observation: this stile of links to discussion is not good, as it gets invalid as soon as the diff was outdated. The correct way how to link the things is https://github.com/theforeman/foreman/pull/3732#discussion_r74886627 - you can get it while getting the link to the comment from `Conversation` tab. This helps a lot co-reviewers following up the discussion. - it even makes nicer links in the text as you see. Thanks

app/models/user.rb
iNecas, line 368: Ad linking to other comments in the PR https://github.com/theforeman/foreman/pull/3732#discussion_r74947625

app/models/host/managed.rb
iNecas: +1 for keeping the public methods/scopes outside of meta-programming. It makes it much easier to navigate through the code later.

I see the problem is we have `my_organizations` and `taxonomy_and_child_ids` that seem to do similar things. We should settle on using one of them publicly and stick on it. Since the `my_organizations`/`my_locations` takes into account the `admin` flag and it's more powerful, I would prefer using them as public methods, and hiding the `taxonomy_and_child_ids` into private method of `Taxonomy` class. This way it would be clear in the future which method/scope to use.

ares: I agree, I created http://projects.theforeman.org/issues/16128 to track this (and volunteer to do it). I hope that does not block this PR since it's unrelated change. These methods were added long time ago.

ares:
I'm not sure why katello tests fail, I see there are some host [fixtures](https://github.com/Katello/katello/blob/master/test/fixtures/models/hosts.yml) but I don't see where it would conflict with host owner which is by default the admin. @jlsherrill any ideas how this PR could affect it? TL;DR of what it does - adds validation to host that it's owner must be associated to at least same taxonomies as host's organization and location is set.

app/models/host/managed.rb
```diff
@@ -985,7 +995,7 @@ def ensure_associations
   end
 
   def set_default_user
-    return if self.owner_type.present? && !OWNER_TYPES.include?(self.owner_type)
+    return if self.owner_type.present? && !OWNER_TYPES.include?(self.owner_type) # to propagate invalid value to validations
     self.owner_type ||= 'User'
```
iNecas: In relation to https://github.com/theforeman/foreman/pull/3740, I guess we should set the default user only when the `owner_type` is not specified explicitly

ares: not sure I follow, we set it only if owner type is present which means it was specified

test/fixtures/taxable_taxonomies.yml
```diff
@@ -133,3 +133,13 @@ twentysix:
...
+
+restricted_user_organization1:
+  taxonomy: organization1
+  taxable: restricted
+  taxable_type: User
+
+restricted_user_location1:
+  taxonomy: location1
+  taxable: restricted
+  taxable_type: User
```
jlsherrill, line 145: I dug into one of the failures and adding these is actually the cause of the katello error.  In a lot of our tests, we are using the restricted user and creating filters without any organization or location.  This worked previously due to this line:

https://github.com/theforeman/foreman/blob/develop/app/models/concerns/taxonomix.rb#L186

but now that the user is restricted to a particular org/loc, each filter must have at least one of the user's orgs and locations.  In our tests we're using Role#add_add_permissions! to setup the role to test, and that doesn't seem to be able to handle orgs/locs at all.

ares, line 145: ok, I think I can avoid this and add these taxonomies to the restricted used only in specific tests or use different user, thanks @jlsherrill!

app/models/host/managed.rb
ares: After #3740 was merged I modified this so the set_default_user only sets default user if both are missing, ignoring OWNER_TYPES as it's handled by `Host::Base` validation. Now user can't create host without specifying either both or none, hammer example:

``` sh
# current user will be used
$ hammer host create --managed false --name a --location Brno --organization Ares --build false
Host created

$ hammer host create --managed false --name a --location Brno --organization Ares --build false --owner-type Usergroup
Could not create the host:
  Owner If owner type is specified, owner must be specified too.
$ hammer host create --managed false --name a --location Brno --organization Ares --build false --owner-type User     
Could not create the host:
  Owner If owner type is specified, owner must be specified too.

$ hammer host create --managed false --name a --location Brno --organization Ares --build false --owner-id 2     
Could not create the host:
  Owner type Owner type needs to be one of the following: User, Usergroup

$ hammer host create --managed false --name a --location Brno --organization Ares --build false --owner-id 2 --owner-type User
Host created
```

test/models/host_test.rb
```diff
@@ -698,6 +698,14 @@ def teardown
         assert_match(/Usergroup/, host.errors[:owner].first)
       end
     end
+
```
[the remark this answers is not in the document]
ares: fixed, not sure why this does not disappear

ares:
[test foreman]

ares:
@iNecas I fixed your comments and made an improvement in owner handling, assigning was moved to before_validate, rest is handled by proper validations. Please take another look when you have a chance. I also updated it so Katello tests won't be affected by the change.

dLobatog:
Thanks @ares  for the updates, I'll merge. 

I'm not very convinced that this is a legitimate bug when you can still assign an Usergroup as the host_owner, and Usergroup.users need not to have the Host taxonomies in order for that to happen. 

I don't think we should protect that use case (it can be rather limiting for users), this one I guess is OK since it's straightforward to check whether the assigned owner will be able to see this or not.

dLobatog:
Merged as a4d7a037ba2cd143713e6a92f49ad6dc8c314628, thanks @ares!

PR #3733, opened 2016-08-15
Fixes #16114 - Expose host parametesr to GET /api/hosts

treydock:


domcleal:
Thanks for the pull request, but I'm unsure this is desirable. The index should be built for speed, but retrieving parameters is a relatively expensive process - it requires looking up directly associated objects (subnets, domains, OSes) and loading parameters joined to all of those too.

The separation should be similar to the UI where only limited information is shown in the index and detailed information is per-host, to reduce the load time. I'm not sure parameters are really suitable for an index. Does that make sense?

treydock:
@domcleal The exclusion of parameters from index makes sense.  Would it be acceptable if some configuration option enabled the parameter behavior?  In our environment we use parameters across hundreds of systems to modify system behavior and right now querying the parameter states via API requires 800 GET requests which is time consuming.

domcleal:
An option in the request might be better than a server-side configuration option, it'd be more like other APIs - I'd suggest `include=parameters,all_parameters`.

(e.g. JSON API: http://jsonapi.org/format/#fetching-includes)

app/views/api/v2/hosts/main.json.rabl
```diff
@@ -40,6 +40,18 @@ node :hostgroup_title do |host|
   host.hostgroup.title if host.hostgroup.present?
 end
 
+if params[:action] == 'show' || (params[:include].present? && params[:include].include?('all_parameters'))
```
dLobatog: I think you meant 'parameters' here instead of 'all_parameters', otherwise  you're showing both keys with `all_parameters`

dLobatog:
@treydock There was a small typo in the RABL definition - should be ready to merge after that. Thanks !

treydock:
@dLobatog I've updated the PR

app/views/api/v2/hosts/main.json.rabl
```diff
@@ -40,6 +40,18 @@ node :hostgroup_title do |host|
   host.hostgroup.title if host.hostgroup.present?
 end
 
+if params[:action] == 'show' || (params[:include].present? && params[:include].include?('parameters'))
```
domcleal: Can you extract `params` out of the view? It ties the view to this controller quite tightly and may prevent foreman_hooks from using the view via `Rabl.render`. It might be better to use an instance variable set in the controller(s) as these can be passed into `Rabl.render`.

domcleal (changes requested):

treydock:
@domcleal I've updated PR to move params logic into controller.

domcleal:
Thanks, that's much better. Lastly, would you be able to add a functional test that passes parameters/all_parameters to the index API and checks the response works?

There are a couple tests of :index in test/functional/api/v2/hosts_controller_test.rb that can be extended trivially, e.g. by copying the test that checks for `ip` in the response to check for one/both of the fields you're adding, and passing the extra `include` parameter on the `get` line.

test/functional/api/v2/hosts_controller_test.rb
```diff
@@ -89,6 +89,38 @@ def expect_attribute_modifier(modifier_class, args)
...
+    refute parameters.empty?
+    assert_equal parameter['value'], 'bar'
+  end
+
+  test "should get all_parameters from index" do
+    hostgroup = FactoryGirl.create(:hostgroup, :with_parent, :with_domain, :with_os)
+    hostgroup.group_parameters = [GroupParameter.new(name: 'foobar', value: 'baz')]
+    hostgroup.save!
+    last_record.parameters = [HostParameter.new(name: 'foo', value: 'bar')]
+    last_record.save!
```
domcleal: This isn't doing anything, it can be removed (it probably should have been after the hostgroup set, but it won't work as last_record retrieves a new copy every time it's called).

test/functional/api/v2/hosts_controller_test.rb
```diff
@@ -89,6 +89,38 @@ def expect_attribute_modifier(modifier_class, args)
...
+    last_record.hostgroup = hostgroup
```
domcleal: This isn't being saved so the inherited_parameter test is failing. Use `last_record.update_attribute(:hostgroup_id, hostgroup.id)` to set and save it in one operation.

domcleal (changes requested):

isratrade:
@treydock, this use case of optionally returning nested data (such as host parameters) in a collection is one of issues that is solved by the [JSONAPI](http://jsonapi.org/) spec. FYI, I started a repo for a Foreman plugin that follows this spec called [foreman_jsonapi](https://github.com/isratrade/foreman_jsonapi). There's also an [RFC](https://github.com/isratrade/rfcs/blob/master/0000-foreman-api-v3.md). 

Btw, notice that the JSONAPI spec uses sideloading for nested data rather than embedding full json objects.

domcleal:
@treydock thanks for the update, but it's conflicting with develop. Please rebase it and hopefully tests will run without conflicts.

domcleal (approved):

domcleal:
Merged, thanks @treydock.

PR #3734, opened 2016-08-16
Refs #12364 - webpack server can serve assets on https

ares:
https://github.com/theforeman/foreman/commit/42ba771cc8bfc09257c15c5d388542d4d1b69358 does not fix the issue completely, the policy is required but is not sufficient if the webpack server runs on the same domain as the Foreman. Browser refuses content from both `https://$domain` and `http://$domain` if the domain is same in both urls. With this patch, webpack server can be started in https mode and we can configure the client to use https too in settings.yaml.

There's one gotcha though, webpack_rails only supports verify_none ssl mode, I plan to send them patch and after that we can expose this option too.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,4 +1,5 @@
 # Run Rails & Webpack concurrently
 # If you wish to use a different server then the default, use e.g. `export RAILS_STARTUP='puma -w 3 -p 3000 --preload'`
 rails: [ -n "$RAILS_STARTUP" ] && $RAILS_STARTUP || bin/rails server -b 0.0.0.0
-webpack: ./node_modules/.bin/webpack-dev-server --config config/webpack.config.js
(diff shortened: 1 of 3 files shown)


ares:
https://github.com/mipearson/webpack-rails/pull/49 - PR opened

tstrachota:
:+1: works nice for me!

dLobatog:
Just triggered http://ci.theforeman.org/job/test_develop/2319/ to see if test_develop comes back, ready to merge if tests pass :+1:

dLobatog:
Lots of 'java.io.IOException: remote file operation failed:'.. [test]?

dLobatog:
Merged as ab5c7c28f3c8e6a4fd4731cba590b62d5497ac97, thanks @ares!

PR #3735, opened 2016-08-16
fixes #16061 - automatical ip assignment for non-primary interfaces

Klaas-:


Klaas-:
I'm not sure if this should go into the if clause https://github.com/Klaas-/foreman/blob/develop/app/models/host/managed.rb#L631
or if it should not be part of that and check each interface if its managed; looking for some feedback :)

app/models/host/managed.rb
```diff
@@ -631,6 +631,11 @@ def set_ip_address
     if SETTINGS[:unattended] && (new_record? || managed?)
       self.ip  ||= subnet.unused_ip.suggest_ip if subnet.present?
       self.ip6 ||= subnet6.unused_ip.suggest_ip if subnet6.present?
+      self.interfaces.each do |nic|
+        next if nic.primary
```
timogoebel: self.ip is delegated to the primary interface. I think, a better approach is to just loop over all interfaces and leave out the "next" line.

``` ruby
self.interfaces.each do |nic|
  nic.ip  ||= nic.subnet.unused_ip.suggest_ip if nic.subnet.present?
  nic.ip6 ||= nic.subnet6.unused_ip.suggest_ip if nic.subnet6.present?
end
```

timogoebel:
@Klaas- : You definitely need the if `if SETTINGS[:unattended] && (new_record? || managed?)`. This just checks that the host is managed.
You probably want to limit the suggestion depending on whether the interface itself is managed or not.

Could you please add some testing for this?

app/models/host/managed.rb
```diff
@@ -629,8 +629,10 @@ def set_compute_attributes
 
   def set_ip_address
     if SETTINGS[:unattended] && (new_record? || managed?)
-      self.ip  ||= subnet.unused_ip.suggest_ip if subnet.present?
-      self.ip6 ||= subnet6.unused_ip.suggest_ip if subnet6.present?
+      self.interfaces.each do |nic|
+        nic.ip  ||= nic.subnet.unused_ip.suggest_ip if nic.subnet.present? and nic.managed
+        nic.ip6 ||= nic.subnet6.unused_ip.suggest_ip if nic.subnet6.present? and nic.managed
```
timogoebel: please use `&&` over `and`. the latter is for flow control.

Klaas-:
I'm not sure how to test this; I can't find the previous tests for the set_ip_address function

timogoebel:
If you want to test host api creation, you could add a test to `test/functional/api/v2/hosts_controller_test.rb` that creates the host. There are some examples you can use. You can then stub the calls to `suggest_ip`. There should be examples, too. Let me know, if you need more help. :-)

timogoebel:
[test]

Klaas-:
okay I refactored it to be less complex; seems like that worked, could you please rerun the tests again @timogoebel ?

timogoebel:
sure, [test]

timogoebel:
[test]

lzap:
LGTM, works as expected with discovery. Thanks!

lzap:
Merged as a86dcf44b236e59c1b4370c1a51749360420e69e, thank you!

lzap:
Fixed spelling in the commit message.

PR #3736, opened 2016-08-16
Fixes #16120 - Allow including css in webpack assets

tbrisker:
This commit demonstrates using the css from the select2 module using the
webpack css loaders. Images under 32k will be compiled into the css to
reduce request counts.


tbrisker:
[test]

gailsteiger:
Does webpack produce css source maps?

Looks good. I think this can be merged.

config/webpack.config.js
```diff
@@ -81,7 +95,6 @@ if (production) {
     port: devServerPort,
     headers: { 'Access-Control-Allow-Origin': '*' }
   };
-  config.output.publicPath = '//0.0.0.0:' + devServerPort + '/webpack/';
```
dLobatog, line 84: Before merging this, why can this line be removed now?

tbrisker, line 84: TBH I'm not sure why it was needed in the first place. I am removing it because it makes urls from css turn into absolute urls instead of relative ones, but that isn't needed because relative urls will work for both when the webpack dev server is enabled and disabled. (it would be relative to the relevant server - either rails or webpack)

tbrisker:
@gailsteiger I believe source maps can be added via another loader, but I think it can be done in a separate PR.

dLobatog:
@tbrisker I've been testing the PR when reviewing UEFI & others, should be good to merge. I'll do it in a bit. Do you want to try and do the packaging stuff too (remember, only 'dependencies', and 'devdependencies' used in `webpack:compile` are strictly needed) or I should do it?

Thanks @gailsteiger for taking a look!

dLobatog:
Merged as ff51a9f3cce9626481cecc4634e80221884b93cb, thanks @tbrisker!

package.json
```diff
@@ -5,13 +5,19 @@
...
-    "expose-loader": "~0.6.0",
     "babel-core": "~6.7.2",
     "babel-loader": "~6.2.4",
-    "babel-preset-es2015": "~6.6.0"
+    "babel-preset-es2015": "~6.6.0",
+    "css-loader": "^0.23.1",
+    "es6-promise": "^3.2.1",
+    "expose-loader": "~0.6.0",
+    "extract-text-webpack-plugin": "^1.0.1",
+    "file-loader": "^0.9.0",
```
dLobatog, line 15: Was this really needed?

dLobatog, line 15: yup, peerdep of url-loader

PR #3737, opened 2016-08-16
Fixes #15814 - Reset override params when override is off

daviddavis:


test/unit/lookup_key_test.rb
```diff
@@ -376,6 +376,20 @@ def setup
     end
   end
 
+  test "override params are reset after override changes back to false" do
+      @key = FactoryGirl.create(:puppetclass_lookup_key, :as_smart_class_param,
```
orrabin: rubocop doesn't like the indentation here, it should be 2 spaces instead of 4

app/models/lookup_keys/puppetclass_lookup_key.rb
```diff
@@ -41,6 +41,10 @@ def puppet?
   def check_override_selected
     return if (changed - ['description', 'override']).empty?
     return if override?
+    return if !override? && override_changed? && (changed - ['merge_overrides',
```
orrabin: Can `reset_override_params` be called only when override changed from true to false to avoid this addition?
Edit: looks like the condition will still need to be added because right now you can't change anything when turning override to false but `reset_override_params` can still be called only on the change.

daviddavis: Yea, my concern was systems which already have override = false and merge_overrides = true. They're going to still hit the validation error if they try to update their puppet classes. I suppose I could write a migration though?

orrabin: A migration would be good, to keep this from continuing to run each time you save the lookup key.

daviddavis:
@orrabin I added a migration and updated `reset_override_params`.

daviddavis:
[test]

app/models/lookup_keys/lookup_key.rb
```diff
@@ -203,6 +204,16 @@ def cast_default_value
     true
   end
 
+  def reset_override_params
+    if override_changed? && !override?
```
orrabin: nitpick: this can be moved to the `before_validation` and not get here at all unless override has changed.

test/unit/lookup_key_test.rb
```diff
@@ -376,6 +376,20 @@ def setup
     end
   end
 
+  test "override params are reset after override changes back to false" do
+    @key = FactoryGirl.create(:puppetclass_lookup_key, :as_smart_class_param,
+                              :override => true, :key_type => 'array',
+                              :default_value => '[]', :puppetclass => puppetclasses(:one),
+                              :use_puppet_default => true)
+
+    @key.merge_overrides = true
```
orrabin: maybe also add `merge_default` and `avoid_duplicates`?

test/unit/lookup_key_test.rb
```diff
@@ -376,6 +376,20 @@ def setup
...
+    @key.save
+    @key.override = false
+
+    assert @key.save
+    refute @key.errors.any?
```
orrabin, line 393: you should also check `merge_overrides` has changed to false

orrabin:
Tested both with api and ui - when I change override to false, all three turn to false. 
I added a few more comments (mostly nitpicks) and then this is ready to merge.

app/models/lookup_keys/lookup_key.rb
daviddavis: I don't see a benefit to doing so but if you think it's better, then I can.

**EDIT**: I don't see the harm in doing so either though so I'll go ahead and update.

daviddavis:
@orrabin updated

orrabin:
@daviddavis thanks, waiting for the tests to be green

app/models/lookup_keys/puppetclass_lookup_key.rb
tbrisker: The first clause will always be true (since we check the opposite in the line before). Also, If the description also changed, this check will fail although it should not - please add a test for that case. 
I'm not too fond of very complex conditions as they are very error prone, I would be happy if you could find a way to simplify this check.

daviddavis: > The first clause will always be true (since we check the opposite in the line before). 

I'm a bit torn. By explicitly checking for `!override`, line 44 makes sense in and of itself. Plus if line 43 is changed/removed, it'll still work. That said, I agree that simplifying the condition would be good so I can make this change.

> Also, If the description also changed, this check will fail although it should not

This is the current behavior on develop. I'd probably argue that this is a separate bug/issue.

daviddavis: > This is the current behavior on develop. I'd probably argue that this is a separate bug/issue.

Actually it does look like since we're setting merge_overrides, etc now when override changes, it bypasses line 42 so I'll fix that in this PR.

daviddavis:
@orrabin @tbrisker I updated the code to reset the params after validation which means not having to touch the check_override_selected method. Please review and test it out. Thanks.

By the way, I am working on the katello failure here: https://github.com/Katello/katello/pull/6260

config/environments/development.rb
```diff
@@ -31,7 +31,7 @@
   # Debug mode disables concatenation and preprocessing of assets.
   # This option may cause significant delays in view rendering with a large
   # number of complex assets.
-  config.assets.debug = true
+  config.assets.debug = false
```
orrabin: accidentally added to the commit?

daviddavis: Ugh yea. Thanks.

daviddavis:
[test]

orrabin:
merged as e21a3a47573306f58e43657afeed18e4b5807311, thanks @daviddavis!

PR #3740, opened 2016-08-17
Fixes #16125 - add owner type validation

ShimShtein:


ares:
Maybe this new validation or existing validation in `Host::Base` can be moved so we don't have it on 2 different places.

app/models/host/managed.rb
```diff
@@ -209,6 +209,7 @@ def compute_provides?(attr)
   before_validation :set_certname, :if => Proc.new {|h| h.managed? && Setting[:use_uuid_for_certificates] } if SETTINGS[:unattended]
   after_validation :trigger_nic_orchestration, :if => Proc.new { |h| h.managed? && h.changed? }, :on => :update
   before_validation :validate_dns_name_uniqueness
+  validate :validate_owner
```
iNecas: would be viable `validates :owner_id, :presence = true, :if => Proc.new { |h| h.owner_type.present? }`

ShimShtein: Not exactly, you can have `owner_id.present?`, but type mismatch. Then the `owner` property will remain empty (and will not be set).

ShimShtein:
@ares:

> Maybe this new validation or existing validation in Host::Base can be moved so we don't have it on 2 different places.

I would move the whole owner handling to a concern on the `Host::Base`, but I see that part of it is in `Host::Managed`, so I am not sure where to draw the line. I have used `Host::Managed` since the default owner is [set there](https://github.com/theforeman/foreman/blob/develop/app/models/host/managed.rb#L987).

iNecas:
Since owner_type is validated in Host::Base, I'm for having this validation with it. The fact that for managed host we set the default owner, is not that important to the validation.

ShimShtein:
Took the liberty to move the `:owner` relation to the `Host::Base` - otherwise I couldn't move the validator. It looks a bit odd that `Host::Base` contained validation on the `owner_type` field, but didn't have it defined.

test/unit/host_test.rb
```diff
@@ -686,6 +686,32 @@ def teardown
...
+        :architecture => architectures(:x86_64), :environment => environments(:production), :managed => true
+      assert host.valid?
+    end
+
+    test "should not save if owner_type is set without owner" do
+      host = Host.new :name => "myfullhost", :mac => "aabbecddeeff", :ip => "3.3.4.03", :medium => media(:one),
+        :domain => domains(:mydomain), :operatingsystem => operatingsystems(:redhat), :subnet => subnets(:two), :puppet_proxy => smart_proxies(:puppetmaster),
+        :architecture => architectures(:x86_64), :environment => environments(:production), :managed => true,
+        :owner_type => "Usergroup"
+      assert !host.valid?
```
orrabin: why not use `refute_valid` and `assert_valid` in these tests? or at least `refute host.valid?`

ShimShtein:
Changed syntax to `*_valid`, thanks @orrabin!

domcleal:
Sorry, I caused an error with the tests for this PR when making a change to Jenkins. Please amend/rebase and re-push the commit to re-run them for your reviewer.

ShimShtein:
Rebased

test/unit/host_test.rb
```diff
@@ -674,16 +674,42 @@ def teardown
...
     test "should not save if owner_type is not User or Usergroup" do
       host = Host.new :name => "myfullhost", :mac => "aabbecddeeff", :ip => "3.3.4.03", :medium => media(:one),
         :domain => domains(:mydomain), :operatingsystem => operatingsystems(:redhat), :subnet => subnets(:two), :puppet_proxy => smart_proxies(:puppetmaster),
         :architecture => architectures(:x86_64), :environment => environments(:production), :managed => true,
         :owner_type => "UserGr(up" # should be Usergroup
-      assert !host.valid?
+      refute_valid host
+    end
+
+    test 'should succeed validation if owner not set' do
```
dLobatog: nitpick, these tests are good, but they could use factories & be put in a context to be much easier to read - https://gist.github.com/ad4e32e5f751f956918362f79c6851ea

dLobatog:
@ShimShtein :+1: ready to merge, please check out the comment inline to make tests a bit easier to read, it's a small change. I'll merge after pushing those changes. Thanks!

ShimShtein:
Rebased and added test scope.

test/unit/host_test.rb
```diff
@@ -669,21 +669,38 @@ def teardown
...
-        :architecture => architectures(:x86_64), :environment => environments(:production), :managed => true,
-        :owner_type => "UserGr(up" # should be Usergroup
-      assert !host.valid?
+      test "should not save if owner_type is not User or Usergroup" do
+        host = FactoryGirl.build(:host, :owner_type => "UserGr(up") # should be Usergroup
+        refute_valid host
+      end
+
+      test 'should succeed validation if owner not set' do
+        # don't want to use FactoryGirl here, since I am not sure that the owner
```
dLobatog: It's not set, check `/test/factories/host_related` or inspect it with pry, please change this

ShimShtein:
Moved the last test to `FactoryGirl` with a trait that will ensure host's state.

dLobatog:
Merged as 698e916ce208b5040b83a908a058c83c94d158ee, thanks @ShimShtein!

PR #3741, opened 2016-08-18
Fixes #16159 - Rename mail alerts to not be Puppet specific

dLobatog:
Right now, under 'my account', users get emails with alerts about the
status of their hosts regarding configuration management (e.g: if there
was an error, get an alert, or get a weekly summary of all hosts
changes)

These emails gather info from reports that could come from Puppet,
Ansible, Salt or Chef, but there are references to Puppet all over
them.

I think these Puppet references should be removed as these reports
may not come from Puppet.
## 

Another step further in this direction would be to start storing a
'origin' or similar field in Reports and Facts that indicates where
they come from.

### DIFF
--- a/app/views/host_mailer/error_state.html.erb
+++ b/app/views/host_mailer/error_state.html.erb
@@ -1,4 +1,4 @@
-<h2 class="headline"><%= _("<b>Foreman</b> Puppet error report").html_safe %></h2>
+<h2 class="headline"><%= _("<b>Foreman</b> Configuration Management Error Report").html_safe %></h2>
 <div class="dashboard">
   <%= render :partial => 'error_states', :locals => {:logs => @report.logs} %>
   <br>
--- a/db/migrate/20160818062936_rename_puppet_mail_notifications.rb
+++ b/db/migrate/20160818062936_rename_puppet_mail_notifications.rb
@@ -1,0 +1,20 @@
+class RenamePuppetMailNotifications < ActiveRecord::Migration
+  # We have to load this class in the migration so that existing mail
+  # notifications of 'type=PuppetError' can load
+  class ::PuppetError < ConfigManagementError
+  end
+
+  def up
+    MailNotification.where(:name => 'puppet_summary').
+      update_all(:name => 'config_summary')
+    MailNotification.where(:name => 'puppet_error_state').
+      update_all(:name => 'config_error_state', :type => ConfigManagementError)
+  end
+
+  def down
+    MailNotification.where(:name => 'config_summary').
+      update_all(:name => 'puppet_summary')
+    MailNotification.where(:name => 'config_error_state').
+      update_all(:name => 'puppet_error_state', :type => PuppetError)
+  end
+end
--- a/test/fixtures/mail_notifications.yml
+++ b/test/fixtures/mail_notifications.yml
@@ -1,18 +1,18 @@
 # Read about fixtures at http://ar.rubyonrails.org/classes/Fixtures.html
-puppet_summary:
-  name: puppet_summary
-  description: Puppet summary
+config_summary:
+  name: config_summary
+  description: config summary
   mailer: HostMailer
   method: summary
   subscription_type: report
 
-puppet_error_state:
-  name: puppet_error_state
-  description: A notification when a host reports a puppet error
+config_error_state:
+  name: config_error_state
+  description: A notification when a host reports a config error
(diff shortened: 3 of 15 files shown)


app/models/mail_notifications/config_error.rb
```diff
@@ -0,0 +1,10 @@
+class ConfigError < MailNotification
```
lzap: Since this is missing namespace, I'd prefer something more specific. This is too generic term that can be defined in our dependencies. How about `ConfigManagementError` or similar?

app/views/host_mailer/error_state.html.erb
```diff
@@ -1,4 +1,4 @@
-<h2 class="headline"><%= _("<b>Foreman</b> Puppet error report").html_safe %></h2>
+<h2 class="headline"><%= _("<b>Foreman</b> configuration management error report").html_safe %></h2>
```
lzap: Mail subject has capitals while this subject not. Just an observation, same below.

test/fixtures/mail_notifications.yml
```diff
@@ -1,18 +1,18 @@
...
+  name: config_error_state
+  description: A notification when a host reports a config error
   mailer: HostMailer
   method: error_state
   subscription_type: alert
-  type: PuppetError
+  type: ConfigError
 
 welcome:
   name: welcome
```
lzap: Time to get rid of this? You are essentially changing 50 % of all lines of the file.

dLobatog:
@lzap Thanks for taking a look - I've updated this according to your suggestions, except for removing the fixture which involves changing most of the mail notifications test.

dLobatog:
@lzap ping?

db/migrate/20160818062936_rename_puppet_mail_notifications.rb
```diff
@@ -0,0 +1,30 @@
...
+  # We have to load this class in the migration so that existing mail
+  # notifications of 'type=PuppetError' can load
+  class ::PuppetError < ConfigManagementError
+  end
+
+  def up
+    puppet_summary = MailNotification.find_by_name('puppet_summary')
+    if puppet_summary.present?
+      puppet_summary.update_attributes(:name => 'config_summary')
+    end
```
tbrisker: `MailNotification.where(:name => 'puppet_summary').update_all(:name => 'config_summary')`?

db/migrate/20160818062936_rename_puppet_mail_notifications.rb
```diff
@@ -0,0 +1,30 @@
...
+    puppet_error_state = MailNotification.find_by_name('puppet_error_state')
+    if puppet_error_state.present?
+      puppet_error_state.update_attributes(:name => 'config_error_state',
+                                           :type => ConfigManagementError)
+    end
```
tbrisker: `MailNotification.where(:name => 'puppet_error_state').update_all(:name => 'config_error_state', :type => ConfigManagementError)`?

tbrisker (changes requested):
:+1: other then a couple of minor inline comments.

dLobatog:
@tbrisker Updated the migrations as you suggested, thanks 😄

dLobatog:
@tbrisker Rebased  - let me know if there is anything else to fix, thanks!

db/migrate/20160818062936_rename_puppet_mail_notifications.rb
```diff
@@ -0,0 +1,20 @@
...
+      update_all(:name => 'config_summary')
+    MailNotification.where(:name => 'puppet_error_state').
+      update_all(:name => 'config_error_state', :type => ConfigManagementError)
+  end
+
+  def down
+    MailNotification.where(:name => 'config_summary').
+      update_all(:name => 'puppet_summary')
+    MailNotification.where(:name => 'config_error_state').
+      update_all(:name => 'puppet_error_state', :type => ConfigManagementError)
```
tbrisker: shouldn't this be `type => PuppetError`?

tbrisker (changes requested):
one comment inline, otherwise code looks good. I don't have a working mail setup to test though.

dLobatog:
@tbrisker Updated to fix the problem in the `down` part of the migration. Thanks for reviewing!

tbrisker (approved):

tbrisker:
Thanks @dLobatog !

PR #3742, opened 2016-08-18
Fixes #16160 - added script/start-foreman helper

lzap:
I don't want to use "foreman" gem for my development environment, I need pry
without hacks. A simple shell script does the very same. When you hit Ctrl+C
enter, both processes are terminated. And pry does work just fine.

Second part of this improvement is ability to set logging settings per
environment. I want my test environment to have slightly different
configuration than what's defined in `config/settings.yml` but the file is in
git and I can't change it.

### DIFF
--- a/config/initializers/pry_warning.rb
+++ b/config/initializers/pry_warning.rb
@@ -1,0 +1,3 @@
+if ENV['PRY_WARNING'] && defined? pry
+  Rails.logger.warn "WARNING: Pry will not work with foreman gem, use script/foreman-start-dev"
+end
--- a/Procfile
+++ b/Procfile
@@ -1,5 +1,5 @@
 # Run Rails & Webpack concurrently
 # If you wish to use a different server then the default, use e.g. `export RAILS_STARTUP='puma -w 3 -p 3000 --preload'`
-rails: [ -n "$RAILS_STARTUP" ] && $RAILS_STARTUP || [ -n "$BIND" ] && bin/rails server -b $BIND || bin/rails server
+rails: [ -n "$RAILS_STARTUP" ] && env PRY_WARNING=1 $RAILS_STARTUP || [ -n "$BIND" ] && bin/rails server -b $BIND || env PRY_WARNING=1 bin/rails server
 # you can use WEBPACK_OPTS to customize webpack server, e.g. 'WEBPACK_OPTS='--https --key /path/to/key --cert /path/to/cert.pem --cacert /path/to/cacert.pem' foreman start '
 webpack: ./node_modules/.bin/webpack-dev-server --config config/webpack.config.js $WEBPACK_OPTS
--- a/script/foreman-start-dev
+++ b/script/foreman-start-dev
@@ -1,0 +1,3 @@
+#!/bin/sh
+./node_modules/.bin/webpack-dev-server --config config/webpack.config.js $WEBPACK_OPTS &
+./bin/rails server -b 0.0.0.0 "$@"


iNecas:
As irritating as it might be, it should be handled in two PRs, or at least two commits, to show the changes are not related. +1 for the bash script

iNecas:
looking at other files we have in the `scripts` directory, it would look nicer calling it 'foreman-start`, or`foreman-dev-start`

script/start-foreman
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+./node_modules/.bin/webpack-dev-server --config config/webpack.config.js $WEBPACK_OPTS &
+./bin/rails server -b 0.0.0.0 $*
```
iNecas: `"$@"` should be preferred to pass additional args in bash: it handles "" more nicely

lzap:
Will split it up, thanks for comments.

lzap:
Updated.

dLobatog:
Does it make sense having this one and also `foreman` to initialize the dev environment? I tend to think it doesn't but I haven't thought about it in detail. cc-ing @tbrisker since he added it

lzap:
> Does it make sense having this one and also `foreman` to initialize the dev environment? I tend to think it doesn't but I haven't thought about it in detail. cc-ing @tbrisker since he added it 

Not sure I follow, what you mean by "initializing" here.

## 

Later,
 Lukas #lzap Zapletal

ohadlevy:
one major downside of the foreman gem, is that we can't extend it from
plugins (e.g. if a plugin wants to add a backend service - dynflow, redis
etc) you can't make sure its all loaded at the same time

iNecas:
I guess @dLobatog has https://github.com/theforeman/foreman/blob/develop/Procfile in mind. If the functionality can be replaced with a 3 lines of bash, we don't use Foreman gem for production, and it makes it more complicated to use pry, I don't see any reason why we should have Foreman gen here.

tbrisker:
I have no preference here, I added foreman+Procfile as a simple way to start and stop both servers at once or one at a time, with the added bonus of the "foreman on foreman" pun. I don't mind changing it to a script that does the same, as long as we get the same capabilities.

dLobatog:
Same opinion basically, so anyone feel free to merge. I think it'd make sense to remove the foreman gem from our development.rb & the Procfile in this same PR too

If this is merged, @lzap please send an email to foreman-dev and update theforeman.org handbook/contribute instructions which reference 'foreman start'.

lzap:
I am in favor in keeping both scripts, that does not look like huge issue of maintaining both as those are just few lines. Breaking everyone dev setups is also something we do want to do every week. Both make sense:
- foreman for newcomers who will just issue "rails server" (and it should work out-of-box)
- the shell script for hackers who prefer pry compatibility

Instead removing "foreman" I suggest (and amend) a change that will issue a warning when pry is detected and running in "foreman" mode:

```
12:05:10 rails.1   | 2016-08-24T12:05:10  [app] [W] WARNING: Pry will not work with foreman gem, use script/foreman-start-dev
```

iNecas:
Personally, I would say keeping the dev entry point on 1 place is better then on 2: I don't have a feeling that Foreman would be that much of standard that it would make any difference: in fact it was the first time in my life I've seen Procfile really.

config/initializers/pry_warning.rb
```diff
@@ -0,0 +1,3 @@
+if ENV['PRY_WARNING'] && defined? pry
```
lzap, line 1: > Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

Rebased onto develop and you better not report this one, or else... ;-)

## 

Later,
 Lukas #lzap Zapletal

tbrisker:
:+1: to having only one entry point. Might also make sense to check if the dev server is enabled and run it only if it is, or `rake webpack:compile` if not (and the public/webpack folder is older then the webpack/assets folder).

lzap:
I rebased, this is ready for merge, I am not going to implement any other checks or try to integrate with "foreman" gem. My sole idea is to offer an alternative and simple enough entrypoint which is pry-friendly.

lzap:
The hound comment is irrelevant, this is ready for review.

ohadlevy:
@lzap whats the status of this pr? thanks.

lzap:
I have my local startup script which I use, I still think this would be beneficial to have and document.

dLobatog (approved):
Thanks @lzap - feel free to remove Foreman & Procfile on another PR...

ohadlevy:
how do i use this script?

lzap:
You *execute* it? ;-) This is what I have:

    alias bes="rm log/*log 2>/dev/null; [ -x script/foreman-start-dev ] && script/foreman-start-dev || [ -x bin/smart-proxy ] && bundle exec bin/smart-proxy || bundle exec foreman start"

There is a warning if you start via foreman gem, but it still works and it a default option for now.

PR #3744, opened 2016-08-18
Fixes #16167 - added list of safemode methods

lzap:
Visit About page to see the list. Could not figure out better place for this.
Maybe template editor (as hidden box). It is worth to cache this table perhaps
too.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -576,3 +576,9 @@ table {
     float: right;
   }
 }
+
+code.transparent {
+  padding: 0;
+  color: black;
+  background-color: transparent;
+}
--- a/app/controllers/provisioning_templates_controller.rb
+++ b/app/controllers/provisioning_templates_controller.rb
@@ -1,9 +1,14 @@
 class ProvisioningTemplatesController < TemplatesController
   include Foreman::Controller::Parameters::ProvisioningTemplate
+  helper_method :documentation_anchor
 
   def build_pxe_default
     status, msg = ProvisioningTemplate.authorized(:deploy_provisioning_templates).build_pxe_default(self)
     status == 200 ? notice(msg) : error(msg)
     redirect_to :back
+  end
+
+  def documentation_anchor
+    '4.4.3ProvisioningTemplates'
   end
 end
--- a/app/helpers/templates_helper.rb
+++ b/app/helpers/templates_helper.rb
@@ -20,12 +20,22 @@ module TemplatesHelper
     rights.all? && !rights.blank?
   end
 
-  def infobox_functions_and_macros(docs_section)
-    alert(:class => 'alert-info', :header => '',
-          :text => (_('Check out some %s') %
-                    link_to(_('useful template functions and macros'),
-                            documentation_url(docs_section))),
-          :rel => 'external')
+  def safemode_methods
+    @@safemode_methods ||= begin
+      objects = ObjectSpace.each_object(Class).select{|x| x < Safemode::Jail }
+      objects_with_methods = objects.map do |obj|
+        [obj.name.gsub(/::Jail$/, ''), obj.allowed_methods.sort.join(' ')]
+      end
+      objects_with_methods.uniq.sort_by(&:first)
+    end
+  end
+
+  def safemode_helpers
+    @@safemode_helpers ||= Foreman::Renderer::ALLOWED_HELPERS.sort.join(' ')
+  end
+
+  def safemode_variables
+    @@safemode_variables ||= Foreman::Renderer::ALLOWED_VARIABLES.sort.map{|x| "@#{x}"}.join(' ')
   end
 
   def locked_warning(template)
--- a/app/views/provisioning_templates/_alerts.html.erb
+++ b/app/views/provisioning_templates/_alerts.html.erb
@@ -1,5 +1,3 @@
 <% if @template.locked? -%>
   <%= locked_warning(::Template.find(params[:id])) %>
-<% else -%>
-  <%= infobox_functions_and_macros("4.4.3ProvisioningTemplates") %>
 <% end -%>
--- a/app/views/templates/_form.html.erb
+++ b/app/views/templates/_form.html.erb
@@ -9,7 +9,6 @@
   <ul class="nav nav-tabs" data-tabs="tabs">
     <li class="active"><a id="primary_tab" href="#primary" data-toggle="tab"><%= _("Template") %></a></li>
     <%= render "#{@type_name_plural}/custom_tab_headers" unless type == 'ptable'%>
-
     <li><a id='history_tab' href="#history" data-toggle="tab"><%= _("History") %></a></li>
     <% if show_location_tab? %>
       <li><a href="#locations" data-toggle="tab"><%= _("Locations") %></a></li>
@@ -17,6 +16,7 @@
     <% if show_organization_tab? %>
       <li><a href="#organizations" data-toggle="tab"><%= _("Organizations") %></a></li>
     <% end %>
+    <li><a id='help_tab' href="#help" data-toggle="tab"><%= _("Help") %></a></li>
   </ul>
 
   <div class="tab-content">
@@ -30,12 +30,9 @@
       <%= render "custom_tabs", :f => f if type == 'ptable' %>
       <%= render_if_exists "#{@type_name_plural}/_alerts" %>
 
-
-
       <% if pxe_with_building_hosts?(@template) -%>
         <% warning_text = (_("The template is associated to at least one host in build mode. To apply the change, disable and enable build mode on hosts to update the live templates or choose to %s their configuration from 'Select Action' menu") % link_to(_('recreate'), building_hosts_path(@template))).html_safe %>
-        <%= alert(:class => 'alert-warning', :header => 'Warning',
-                  :text => warning_text) %>
+        <%= alert(:class => 'alert-warning', :header => 'Warning', :text => warning_text) %>
       <% end -%>
 
       <div class='form-group'>
@@ -84,6 +81,74 @@
       <% end %>
     </div>
 
+    <div class="tab-pane" id="help">
+      <div>
+        <h1><%= _('Template syntax') %></h1>
+        <%= _('Provisioning templates are generated using the Embedded Ruby (ERB) language.') %>
+        <%= _('The following table and example snippet shows global methods, variables and class methods and how to use them.') %>
+        <%= _('These are guaranteed to work via the safe mode rendering, to ensure a template can do nothing harmful.') %>
+        <% unless Setting[:safemode_render] -%>
+          <%= _('To enable safe mode, navigate to Settings page and enable Safemode rendering option.') %>
+        <% end -%>
+        <%= link_to(_('For more info visit our documentation.'), documentation_url(controller.documentation_anchor)) %>
+        <h2><%= _('Examples') %></h2>
+        <div class="panel-group" id="accordion-markup">
+          <%= render "help_accordion", :title => _('Comments'), :contents => (capture { %>
+            <%%# this is a comment %>
+          <% }) %>
+          <%= render "help_accordion", :title => _('In-line code syntax'), :contents => (capture { %>
+            <%% if true -%><br/>
+            &nbsp;&nbsp;some content<br/>
+            <%% end -%>
+          <% }) %>
+          <%= render "help_accordion", :title => _('Printing data'), :contents => (capture { %>
+            <%%= "some content from variable" %>
+          <% }) %>
+          <%= render "help_accordion", :title => _('Variables'), :contents => (capture { %>
+            <%%= @host %>
+          <% }) %>
+          <%= render "help_accordion", :title => _('Calling methods on objects'), :contents => (capture { %>
+            <%%= @host.operatingsystem.major %>
+          <% }) %>
+          <%= render "help_accordion", :title => _('Safe mode methods'), :contents => (capture { %>
+            <%%# @host.operatingsystem.set_title("this method is not listed for Operatingsystem") %>
+          <% }) %>
+          <%= render "help_accordion", :title => _('Global functions'), :contents => (capture { %>
+            <%%= rand(9999) + 1 %>
+          <% }) %>
+          <%= render "help_accordion", :title => _('Iterating'), :contents => (capture { %>
+            <%% @host.interfaces.each do |interface| -%><br/>
+            &nbsp;&nbsp;NIC name: <%%= interface.name %><br/>
+            <%% end -%><br/>
+          <% }) %>
+        </div>
+      </div>
+      <h2><%= _('Safe mode methods and variables') %></h2>
+      <table class="<%= table_css_classes %>">
+        <thead>
+          <tr>
+            <th class='col-md-3'><%= _("Class") %></th>
+            <th class='col-md-9'><%= _("Allowed methods or members") %></th>
+          </tr>
+        </thead>
+        <tbody>
+          <tr>
+            <td class="ellipsis"><em><%= _("Global methods (functions)") %></em></td>
+            <td><code class="transparent"><%= safemode_helpers %></code></td>
+          </tr>
+          <tr>
+            <td class="ellipsis"><em><%= _("Global variables") %></em></td>
+            <td><code class="transparent"><%= safemode_variables %></code></td>
+          </tr>
+        <% safemode_methods.each do |record| %>
+          <tr>
+            <td class="ellipsis"><%= record.first %></td>
+            <td><code class="transparent"><%= record.second %></code></td>
+          </tr>
+        <% end %>
+        </tbody>
+      </table>
+    </div>
 
     <%= render "custom_tabs", :f => f unless type == 'ptable'%>
     <%= render 'taxonomies/loc_org_tabs', :f => f, :obj => @template %>
--- a/app/views/templates/_help_accordion.html.erb
+++ b/app/views/templates/_help_accordion.html.erb
@@ -1,0 +1,16 @@
+<div class="panel panel-default">
+  <div class="panel-heading">
+    <h4 class="panel-title">
+      <a data-toggle="collapse" data-parent="#accordion-markup" href="#<%= title.parameterize %>" class="collapsed">
+        <%= title %>
+      </a>
+    </h4>
+  </div>
+  <div id="<%= title.parameterize %>" class="panel-collapse collapse">
+    <div class="panel-body">
+      <code>
+        <%= contents %>
+      </code>
+    </div>
+  </div>
+</div>
--- a/app/controllers/ptables_controller.rb
+++ b/app/controllers/ptables_controller.rb
@@ -1,3 +1,8 @@
 class PtablesController < TemplatesController
   include Foreman::Controller::Parameters::Ptable
+  helper_method :documentation_anchor
+
+  def documentation_anchor
+    '4.4.4PartitionTables'
+  end
 end
--- a/app/views/ptables/_alerts.html.erb
+++ b/app/views/ptables/_alerts.html.erb
@@ -1,5 +1,3 @@
 <% if @template.locked? -%>
   <%= locked_warning(::Template.find(params[:id])) %>
-<% else -%>
-  <%= infobox_functions_and_macros("4.4.4PartitionTables") %>
 <% end -%>
--- a/test/functional/provisioning_templates_controller_test.rb
+++ b/test/functional/provisioning_templates_controller_test.rb
@@ -28,6 +28,14 @@ class ProvisioningTemplatesControllerTest < ActionController::TestCase
     assert_template 'edit'
   end
 
+  test "edit page contains help information" do
+    Setting[:safemode_render] = true
+    get :edit, {:id => templates(:pxekickstart).to_param}, set_session_user
+    assert_includes @response.body, Foreman::Renderer::ALLOWED_HELPERS.first.to_s
+    assert_includes @response.body, Foreman::Renderer::ALLOWED_VARIABLES.first.to_s
+    assert_includes @response.body, 'foreman_server_fqdn'
+  end
+
   test "lock" do
     @request.env['HTTP_REFERER'] = provisioning_templates_path
     get :lock, {:id => templates(:pxekickstart).to_param }, set_session_user
--- a/test/unit/helpers/templates_helper_test.rb
+++ b/test/unit/helpers/templates_helper_test.rb
@@ -1,0 +1,18 @@
+require 'test_helper'
+
+class TemplatesHelperTest < ActionView::TestCase
+  include TemplatesHelper
+
+  test "all safemode helpers should be returned" do
+    assert_includes safemode_helpers, Foreman::Renderer::ALLOWED_HELPERS.first.to_s
+  end
+
+  test "all safemode variables should be returned" do
+    assert_includes safemode_variables, Foreman::Renderer::ALLOWED_VARIABLES.first.to_s
+  end
+
+  test "all safemode jail methods should be returned" do
+    assert safemode_methods.any?{ |x| x.first == "Host::Managed" }
+    assert safemode_methods.any?{ |x| x.second.include?("subnet") }
+  end
+end


app/controllers/about_controller.rb
```diff
@@ -5,6 +5,7 @@ def index
     @smart_proxies = SmartProxy.authorized(:view_smart_proxies).includes(:features)
     @compute_resources = ComputeResource.authorized(:view_compute_resources)
     @plugins = Foreman::Plugin.all
+    @safemode = ObjectSpace.each_object(Class).select{|x| x < Safemode::Jail }.collect{|x| [x.name.gsub(/::Jail$/, ''), x.allowed_methods.sort.join(' ')]}.sort{|a,b| a.first <=> b.first}
```
domcleal: Please address the Code Climate result, this is very complex. It should be moved into a helper at the very least, and split into multiple lines with meaningful variable names.

domcleal:
> Maybe template editor (as hidden box).

This seems eminently more sensible than the About page, can you do that?

It's also missing tests.

lzap:
Done moving, I also moved the text field (the editor) under the name, it required scrolling on smaller screens:

image

If you click on the info box link, it will switch tabs to:

image

Note this is only available when safemode is turned on.

app/helpers/templates_helper.rb
```diff
@@ -28,6 +28,20 @@ def infobox_functions_and_macros(docs_section)
           :rel => 'external')
   end
 
+  def safemode_methods
+    Rails.cache.fetch("safemode_methods", :expires_in => 2.hours) do
+      objects = ObjectSpace.each_object(Class).select{|x| x < Safemode::Jail }
+      objects_with_methods = objects.collect do |obj|
```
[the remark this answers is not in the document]
lzap: Changed.

app/helpers/templates_helper.rb
```diff
@@ -28,6 +28,20 @@ def infobox_functions_and_macros(docs_section)
           :rel => 'external')
   end
 
+  def safemode_methods
+    Rails.cache.fetch("safemode_methods", :expires_in => 2.hours) do
```
domcleal: The version number (`SETTING[:version]`) should probably be added to the cache key, then it'll automatically invalidate on update.

app/helpers/templates_helper.rb
```diff
@@ -28,6 +28,20 @@ def infobox_functions_and_macros(docs_section)
...
+      objects = ObjectSpace.each_object(Class).select{|x| x < Safemode::Jail }
+      objects_with_methods = objects.map do |obj|
+        [obj.name.gsub(/::Jail$/, ''), obj.allowed_methods.sort.join(' ')]
+      end
+      objects_with_methods.uniq.sort{|a,b| a.first <=> b.first}
```
domcleal: `sort_by(&:first)`

app/helpers/templates_helper.rb
```diff
@@ -28,6 +28,20 @@ def infobox_functions_and_macros(docs_section)
...
+    end
+  end
+
+  def infobox_safemode
+    alert(:class => 'alert-info', :header => '', :text => (_('In safemode only %s are allowed') % link_to_function(_('selected methods or variables'), '$("#safemode_tab").click()')))
```
domcleal: a) I'm unsure if "safemode" is the best phrase to use.  I'd suggest keeping it really simple and use "In templates", because you already know safemode is enabled.  Else I'd consider calling it "safe mode", as this is the description used in settings - the name of the gem isn't the user's business.
b) Avoid splitting the sentence into three parts over two strings, they won't be translated together and translators won't know they're from the same text. I'd suggest having one sentence, a colon, then a link that can be translated more independently (e.g. "list of available template helpers").

app/views/about/index.html.erb
```diff
@@ -13,6 +13,7 @@
           <li><a href="#compute_resources" data-toggle="tab"><%= _('Compute Resources') %></a></li>
         <% end %>
         <li><a href="#plugins" data-toggle="tab"><%= _('Plugins') %></a></li>
+        <li><a href="#safemode" data-toggle="tab"><%= _('Safemode') %></a></li>
```
domcleal: ditto about "Safemode", consider making this more generic - "Template help", "Help" or similar?

app/views/provisioning_templates/_alerts.html.erb
```diff
@@ -1,3 +1,5 @@
+<%= infobox_safemode if Setting[:safemode_render] %>
```
domcleal: Should this be in the `else` branch with the other help box?

app/views/templates/_form.html.erb
```diff
@@ -9,35 +9,22 @@
...
-
     <li><a id='history_tab' href="#history" data-toggle="tab"><%= _("History") %></a></li>
     <% if show_location_tab? %>
       <li><a href="#locations" data-toggle="tab"><%= _("Locations") %></a></li>
     <% end %>
     <% if show_organization_tab? %>
       <li><a href="#organizations" data-toggle="tab"><%= _("Organizations") %></a></li>
     <% end %>
+    <% if Setting[:safemode_render] %>
+      <li><a id='safemode_tab' href="#safemode" data-toggle="tab"><%= _("Safemode") %></a></li>
```
domcleal: ditto re naming

app/views/templates/_form.html.erb
```diff
@@ -9,35 +9,22 @@
...
 
   <div class="tab-content">
     <div class="tab-pane active" id="primary">
       <%= text_f f, :name, :disabled => @template.locked? %>
 
-      <% if show_default? %>
-        <%= checkbox_f f, :default, :label=>_('Default'), :help_block => default_template_description %>
-      <% end -%>
-
-      <%= render "custom_tabs", :f => f if type == 'ptable' %>
```
domcleal, line 30: Why move the custom tabs and default checkbox? This is unrelated to the alerts, and spreads the ptable UI about somewhat.

app/views/templates/_form.html.erb
```diff
@@ -84,6 +85,32 @@
...
+        <thead>
+          <tr>
+            <th><%= _("Class") %></th>
+            <th><%= _("Allowed methods or members") %></th>
+          </tr>
+        </thead>
+        <tbody>
+          <tr>
+            <td class="ellipsis"><em><%= _("Global methods (functions)") %></em></td>
+            <td><%= Foreman::Renderer::ALLOWED_HELPERS.join(' ') %></td>
```
domcleal: This column should all have code style formatting, not plain text.

app/views/templates/_form.html.erb
```diff
@@ -84,6 +85,32 @@
...
+          </tr>
+          <tr>
+            <td class="ellipsis"><em><%= _("Global variables") %></em></td>
+            <td><%= Foreman::Renderer::ALLOWED_VARIABLES.map{|x| "@#{x}"}.join(' ') %></td>
```
domcleal: The list of variables (and helpers, above) should be sorted to match the jail methods.

app/helpers/templates_helper.rb
```diff
@@ -28,6 +28,20 @@ def infobox_functions_and_macros(docs_section)
           :rel => 'external')
   end
 
+  def safemode_methods
```
domcleal, line 23: Please add a unit test for this method under test/unit/helpers/ to check it returns some known jail members.

app/helpers/templates_helper.rb
lzap: > The version number (`SETTING[:version]`) should probably be added to the cache key, then it'll automatically invalidate on update.

Good idea, do we also have unique UUID of a process? I could simply use
PID for that (so it invalidates after restart). Cos only restart can
add/remove from the list. Then I could set expire to never.

## 

Later,
 Lukas #lzap Zapletal

domcleal: No point in using the Rails cache if you intend to create a per-process cache, use a class variable instead.

lzap: Indeed, missed that option completely. Will use that instead.

app/helpers/templates_helper.rb
lzap: Sure, going for: _In templates only selected methods and variables are allowed, list of available template helpers_

app/views/about/index.html.erb
lzap: Good idea, I like "Help"

app/views/provisioning_templates/_alerts.html.erb
lzap: Absolutely.

app/views/templates/_form.html.erb
lzap: Because I think it's better UX, but I admit it does not belong in here.

app/views/templates/_form.html.erb
lzap: Tried `<code>` but it looked ugly, created a CSS class for that.

app/views/templates/_form.html.erb
lzap: Yep, extracted it into helpers as well.

app/helpers/templates_helper.rb
lzap: I think that's overtesting, covered with functional test. But I will add one.

lzap:
https://gist.github.com/e843217de35a543fb417b48c66451f5c

app/views/about/index.html.erb
domcleal: This tab is unused?

app/helpers/templates_helper.rb
```diff
@@ -28,6 +28,28 @@ def infobox_functions_and_macros(docs_section)
...
+  def safemode_helpers
+    @@safemode_helpers ||= Foreman::Renderer::ALLOWED_HELPERS.sort.join(' ')
+  end
+
+  def safemode_variables
+    @@safemode_variables ||= Foreman::Renderer::ALLOWED_VARIABLES.sort.map{|x| "@#{x}"}.join(' ')
+  end
+
+  def infobox_safemode
+    alert(:class => 'alert-info', :header => '', :text => (_('In templates only selected methods and variables are allowed') + ', ' + link_to_function(_('list of available template helpers'), '$("#help_tab").click()')))
```
domcleal: I don't think this alert is really necessary. The "Help" tab is quite clear and visible, so this only uses up screen estate, I'd just remove it.

app/views/provisioning_templates/_alerts.html.erb
```diff
@@ -1,5 +1,6 @@
 <% if @template.locked? -%>
   <%= locked_warning(::Template.find(params[:id])) %>
 <% else -%>
+  <%= infobox_safemode if Setting[:safemode_render] %>
   <%= infobox_functions_and_macros("4.4.3ProvisioningTemplates") %>
```
domcleal: This can move onto the Help tab, either as a block of text or alert at the top of the tab, it'd be a much better place for it.

app/views/templates/_form.html.erb
```diff
@@ -9,33 +9,26 @@
...
     <%= render "#{@type_name_plural}/custom_tab_headers" unless type == 'ptable'%>
-
     <li><a id='history_tab' href="#history" data-toggle="tab"><%= _("History") %></a></li>
     <% if show_location_tab? %>
       <li><a href="#locations" data-toggle="tab"><%= _("Locations") %></a></li>
     <% end %>
     <% if show_organization_tab? %>
       <li><a href="#organizations" data-toggle="tab"><%= _("Organizations") %></a></li>
     <% end %>
+    <% if Setting[:safemode_render] %>
```
domcleal: This will need to be always displayed when moving the link to docs.

app/assets/stylesheets/application.scss
```diff
@@ -573,3 +573,7 @@ table {
 .masked-input {
   font-family: 'AllBullets';
 }
+
+.code-in-table {
```
domcleal: There should already be styling of `<code>` tags from Bootstrap, is this needed? (and should it be `table code`?)

app/assets/stylesheets/application.scss
```diff
@@ -573,3 +573,7 @@ table {
...
+  font-family: DejaVu Sans Mono, Courier New, monospace;
```
domcleal: Not the standard Bootstrap font family for code blocks, why change it?

app/assets/stylesheets/application.scss
domcleal: If you're intending to change the colour then I'd suggest either not doing it, using `<code>` tags around individual items in the list to better separate them, or extend the `code` CSS within the table to change the color/background properties.

app/helpers/templates_helper.rb
lzap: Sure.

app/views/provisioning_templates/_alerts.html.erb
lzap: Ok I will move it and keep it as alert. My idea is to be bold enough so users get the idea quickly.

app/views/templates/_form.html.erb
lzap: Not sure I understand but I will show it always, the warning box within the help tab will be only visible when safemode is on tho.

app/assets/stylesheets/application.scss
lzap: Yeah I will try to come up with something better.

app/assets/stylesheets/application.scss
lzap: Because I was hoping to retire without decent CSS knowledge. It does not work... :-D

app/views/templates/_form.html.erb
lzap: Shall I keep "In templates only selected methods and variables are allowed" or do you think something better?

lzap:
Pushed https://gist.github.com/e6ee45a7fcdf2d0aa342a0bd5c09effb

app/views/provisioning_templates/_alerts.html.erb
domcleal: This is still on the first tab, shouldn't it be on the Help tab?

app/views/templates/_form.html.erb
domcleal: > This will need to be always displayed when moving the link to docs.

In the previous comment, I'd suggested moving the link to the useful macros etc to the Help tab, which you said you'd move. If it moved then the help tab would always need to be visible - which it now is, so sorted.

app/helpers/templates_helper.rb
```diff
@@ -28,6 +28,28 @@ def infobox_functions_and_macros(docs_section)
...
+  def safemode_helpers
+    @@safemode_helpers ||= Foreman::Renderer::ALLOWED_HELPERS.sort.join(' ')
+  end
+
+  def safemode_variables
+    @@safemode_variables ||= Foreman::Renderer::ALLOWED_VARIABLES.sort.map{|x| "@#{x}"}.join(' ')
+  end
+
+  def infobox_safemode
+    alert(:class => 'alert-info', :header => '', :text => (_('In templates only selected methods and variables are allowed')))
```
domcleal: > Shall I keep "In templates only selected methods and variables are allowed" or do you think something better?

I don't think it's a good alert as it doesn't really provide much information. It's probably better as text introducing the table, perhaps briefly stating that the list shows which methods may be called depending on the data type.

domcleal (changes requested):
Per https://github.com/theforeman/foreman/pull/3744#discussion-diff-77809378, the link to useful macros etc should be on the Help tab now.

app/helpers/templates_helper.rb
lzap: Ok done, added examples block.

app/views/about/index.html.erb
domcleal: Still unused, should be removed.

app/views/templates/_form.html.erb
```diff
@@ -84,6 +83,70 @@
       <% end %>
     </div>
 
+    <div class="tab-pane" id="help">
+      <div>
+        <%= _('Provisioning templates use ERB syntax') %>
```
domcleal: Is this meant to be a heading? If it's meant to be part of the following sentence then it's missing a full stop, though doesn't make much sense as a full sentence.

app/views/templates/_form.html.erb
```diff
@@ -84,6 +83,70 @@
...
+        <%= _('The following table and example snippet shows global methods, variables and class methods and how to use them.') %>
+        <code class="transparent"><pre>
+<%%# this is a comment %>
+
+<%%# use Ruby for code flow %>
+<%% if true %>
```
domcleal: use `-%>` for conditionals, most do

app/views/templates/_form.html.erb
```diff
@@ -84,6 +83,70 @@
...
+<%%# host variable is available in most templates (Host::Managed table entry) %>
+<%%= @host %>
+
+<%%# associated operating system major version (Operatingsystem table entry) %>
+<%%= @host.operatingsystem.major %>
+
+<%%# associated hostgroup name (Hostgroup table entry) %>
+<%%= @host.hostgroup.name %>
+
+<%%# root password from associated hostgrup (without newline - note the trailing dash) %>
```
domcleal: "host group", missing u and space

domcleal (changes requested):
The "Check out some useful template functions and macros" link/alert is still on the first tab, but in an earlier update you said you'd move it to the Help tab. Instead, the new docs you've written are on a different tab to this link for more help.

app/views/templates/_form.html.erb
lzap: It ought to be a header, but I struggle to figure out something reasonable :-) Will do try then.

Edit: It was header not sentence. Whatever.

lzap: How about:

```
        <%= _('Provisioning templates are generated using Embedded Ruby (ERB) language.') %>
        <%= _('The following table and example snippet shows global methods, variables and class methods and how to use them.') %>
        <%= _('These are guaranteed to work via the safemode rendering, to ensure a template can do nothing harmful.') %>
        <%= _('With safemode disabled other macros may work, but are not supported at this time.') %>
        <%= _('To enable safemode, set "safemode_render" to "true" in Settings.') %>
```

Mostly coming from: http://projects.theforeman.org/projects/foreman/wiki/templatewriting

app/views/templates/_form.html.erb
lzap: Sure.

app/views/templates/_form.html.erb
lzap: Thanks.

app/views/templates/_form.html.erb
domcleal: It's OK, though the later sentences are quite short so it doesn't read too well to me. I'd also only suggest showing information about how to enable safemode if it's actually disabled - which is not the default, or leave it out.

Also I think the text in settings uses "safe mode" as two words.

lzap:
Reworded, pushed.

app/views/templates/_form.html.erb
```diff
@@ -84,6 +83,73 @@
       <% end %>
     </div>
 
+    <div class="tab-pane" id="help">
+      <div>
+        <%= _('Provisioning templates are generated using Embedded Ruby (ERB) language.') %>
+        <%= _('The following table and example snippet shows global methods, variables and class methods and how to use them.') %>
+        <%= _('These are guaranteed to work via the safemode rendering, to ensure a template can do nothing harmful.') %>
+        <%= _('With safemode disabled other macros may work, but are not supported at this time.') %>
+        <%= _('To enable safemode, set "safemode_render" to "true" in Settings.') %>
```
domcleal: Please see earlier comments at https://github.com/theforeman/foreman/pull/3744#discussion_r79589977

domcleal (changes requested):

app/views/templates/_form.html.erb
lzap: Sure, I will drop `With safemode disabled other macros may work, but are not supported at this time.` as it is not that relevant and only show the last sentence when appropriate. FYI the setting is named "Safemode rendering" but the description says "safe mode" - I will follow that. Pushing in a minute.

app/views/templates/_form.html.erb
domcleal: I don't think it's an improvement, the new position puts the snippet/family information directly between the template text and the template text upload field, which are related. I think the snippet/family information was more related to the name (and other general attributes), or even on another tab like provisioning templates.

app/views/templates/_form.html.erb
```diff
@@ -54,6 +45,14 @@
...
+      <%= render "custom_tabs", :f => f if type == 'ptable' %>
+
+      <% if pxe_with_building_hosts?(@template) -%>
+        <% warning_text = (_("The template is associated to at least one host in build mode. To apply the change, disable and enable build mode on hosts to update the live templates or choose to %s their configuration from 'Select Action' menu") % link_to(_('recreate'), building_hosts_path(@template))).html_safe %>
+        <%= alert(:class => 'alert-warning', :header => 'Warning',
+                  :text => warning_text) %>
+      <% end -%>
+
       <% unless @template.locked? -%>
         <%= file_field_f f, :template, :class => "editor_file_source",:size => "col-md-10", :id => 'template_file',
```
domcleal, line 55: To continue https://github.com/theforeman/foreman/pull/3744/files#r76795949, this field is related to the template edit box above - adding the other ptable fields here breaks the link between them.

app/views/templates/_form.html.erb
```diff
@@ -84,6 +83,74 @@
       <% end %>
     </div>
 
+    <div class="tab-pane" id="help">
+      <div>
+        <%= _('Provisioning templates are generated using Embedded Ruby (ERB) language.') %>
+        <%= _('The following table and example snippet shows global methods, variables and class methods and how to use them.') %>
+        <%= _('These are guaranteed to work via the safe mode rendering, to ensure a template can do nothing harmful.') %>
+        <% unless Setting[:safemode_render] -%>
+          <%= _('To enable safe mode, navigate to Settings page and flip Safemode rendering option.') %>
```
domcleal: "flip" is rather informal, "enable" or "set the "Safemode rendering" option to "Yes"." may be better.

app/views/templates/_form.html.erb
```diff
@@ -84,6 +83,74 @@
...
+<%%# global method random is listed in the table %>
+<%%= rand(9999) + 1 %>
+
+<%%# iterating through all network interaces of a host %>
+<%% @host.interfaces.each do |interface| -%>
+  NIC name: <%%= interface.name %>
+<%% end -%>
+        </pre></code>
+      </div>
+      <table class="table table-striped table-fixed">
```
domcleal: missing `table_css_classes` (which obsoletes table, table-striped)

app/views/templates/_form.html.erb
```diff
@@ -84,6 +83,74 @@
...
+<%%# print data from Ruby %>
+<%%= "some content from variable" %>
+
+<%%# host variable is available in most templates (Host::Managed table entry) %>
+<%%= @host %>
+
+<%%# associated operating system major version (Operatingsystem table entry) %>
+<%%= @host.operatingsystem.major %>
+
+<%%# associated hostgroup name (Hostgroup table entry) %>
```
domcleal: "hostgroup" is inconsistent with the next description, so use "host group" in the first bit of the text, but leave "Hostgroup" in parentheses as that's the class name in the table

app/views/templates/_form.html.erb
```diff
@@ -84,6 +83,74 @@
...
+        <% unless Setting[:safemode_render] -%>
+          <%= _('To enable safe mode, navigate to Settings page and flip Safemode rendering option.') %>
+        <% end -%>
+        <code class="transparent"><pre>
+<%%# this is a comment %>
+
+<%%# use Ruby for code flow %>
+<%% if true -%>
+  some content
+<%% end %>
```
domcleal: also `-%>` here

app/views/templates/_form.html.erb
```diff
@@ -84,6 +83,74 @@
...
 
+    <div class="tab-pane" id="help">
+      <div>
+        <%= _('Provisioning templates are generated using Embedded Ruby (ERB) language.') %>
+        <%= _('The following table and example snippet shows global methods, variables and class methods and how to use them.') %>
+        <%= _('These are guaranteed to work via the safe mode rendering, to ensure a template can do nothing harmful.') %>
+        <% unless Setting[:safemode_render] -%>
+          <%= _('To enable safe mode, navigate to Settings page and flip Safemode rendering option.') %>
+        <% end -%>
+        <code class="transparent"><pre>
```
domcleal: Is `<code>` superfluous here? Pre should keep formatting, and also doesn't appear to need the CSS override.

app/views/templates/_form.html.erb
```diff
@@ -84,6 +83,74 @@
       <% end %>
     </div>
 
+    <div class="tab-pane" id="help">
+      <div>
+        <%= _('Provisioning templates are generated using Embedded Ruby (ERB) language.') %>
```
domcleal: "using _the_ Embedded [..]"?

domcleal (changes requested):
Some inline comments there, but more generally, I'm concerned that the layout of the new code examples block and text you've added is poor - there's little space left between the elements (the text, to the pre block, to the table) and it looks rather thrown together.

Adding a paragraph break between the intro text and table, then a sub-heading above the table to describe what it is may make it look tidier.

> The "Check out some useful template functions and macros" link/alert is still on the first tab, but in an earlier update you said you'd move it to the Help tab. Instead, the new docs you've written are on a different tab to this link for more help.

This review comment from the last updates needs addressing still.

app/views/templates/_form.html.erb
lzap: I've reverted the whole form as it was before.

app/views/templates/_form.html.erb
lzap: I dropped `<code>` completely and redesigned the whole examples with PatternFly element.

app/views/templates/_form.html.erb
lzap: I removed this example completely.

lzap:
Pushed the all new UI. Looks much better now.

app/views/templates/_form.html.erb
```diff
@@ -84,6 +81,74 @@
...
+        <%= _('Provisioning templates are generated using the Embedded Ruby (ERB) language.') %>
+        <%= _('The following table and example snippet shows global methods, variables and class methods and how to use them.') %>
+        <%= _('These are guaranteed to work via the safe mode rendering, to ensure a template can do nothing harmful.') %>
+        <% unless Setting[:safemode_render] -%>
+          <%= _('To enable safe mode, navigate to Settings page and enable Safemode rendering option.') %>
+        <% end -%>
+        <%= _('For more info %s.').html_safe % link_to(_('visit our documentation'), documentation_url(controller.documentation_archor)) %>
+        <h2><%= _('Examples') %></h2>
+        <div class="panel-group" id="accordion-markup">
+          <%= render "help_accordion", :title => 'Comments', :contents => (capture { %>
```
domcleal: Please extract the `:title` strings for i18n in all of these accordions.

app/views/templates/_form.html.erb
```diff
@@ -84,6 +81,74 @@
...
+    <div class="tab-pane" id="help">
+      <div>
+        <h1><%= _('Template syntax') %></h1>
+        <%= _('Provisioning templates are generated using the Embedded Ruby (ERB) language.') %>
+        <%= _('The following table and example snippet shows global methods, variables and class methods and how to use them.') %>
+        <%= _('These are guaranteed to work via the safe mode rendering, to ensure a template can do nothing harmful.') %>
+        <% unless Setting[:safemode_render] -%>
+          <%= _('To enable safe mode, navigate to Settings page and enable Safemode rendering option.') %>
+        <% end -%>
+        <%= _('For more info %s.').html_safe % link_to(_('visit our documentation'), documentation_url(controller.documentation_archor)) %>
```
domcleal: I'd suggest just linking the entire sentence, it'll be far easier to translate - translators won't know that these two strings are related in any way from Transifex. `_("Visit our documentation for more information.")` or similar and link it all.

app/controllers/provisioning_templates_controller.rb
```diff
@@ -1,9 +1,14 @@
 class ProvisioningTemplatesController < TemplatesController
   include Foreman::Controller::Parameters::ProvisioningTemplate
+  helper_method :documentation_archor
```
domcleal: Typo: "anchor"

app/views/templates/_help_accordion.html.erb
```diff
@@ -0,0 +1,16 @@
+          <div class="panel panel-default">
```
domcleal: Nitpick, lose the indentation, I don't think we normally indent fragments to match where they're called from, it's easier to edit (even if it renders without indentation).

domcleal (changes requested):
Thanks, that's a huge improvement.

app/views/templates/_form.html.erb
lzap: Sure.

app/views/templates/_form.html.erb
lzap: Yeah I was thinking - when the rest (inline comments in the examples) are not translated - not to do it. Whatever.

app/views/templates/_help_accordion.html.erb
lzap: Sure, all comments done in https://gist.github.com/2ae30894d7a8157507650aa5e696e1ea

domcleal:
[test foreman]

domcleal (approved):

domcleal:
Merged as 7907984, thanks @lzap.

PR #3745, opened 2016-08-18
fixes #16172 - remove base_class override in template subclasses

domcleal:
On Rails 5.0, the base_class is used during model instantiation to
validate STI relationships as it now casts similarly to Foreman::STI,
causing failures when base_class returns the subclass rather than parent
class.

Removing the overriding of `base_class` in Template subclasses
necessitates changing polymorphic associations to store the base class
in the table's type field rather than the subclass name because the
has_many associations will search by `base_class`. This is recommended
by ActiveRecord when using STI with polymorphic associations:
https://github.com/rails/rails/blob/v5.0.0/activerecord/lib/active_record/associations.rb#L780

### DIFF
--- a/app/models/taxonomy.rb
+++ b/app/models/taxonomy.rb
@@ -14,8 +14,8 @@ class Taxonomy < ActiveRecord::Base
   has_many :smart_proxies, :through => :taxable_taxonomies, :source => :taxable, :source_type => 'SmartProxy'
   has_many :compute_resources, :through => :taxable_taxonomies, :source => :taxable, :source_type => 'ComputeResource'
   has_many :media, :through => :taxable_taxonomies, :source => :taxable, :source_type => 'Medium'
-  has_many :provisioning_templates, :through => :taxable_taxonomies, :source => :taxable, :source_type => 'ProvisioningTemplate'
-  has_many :ptables, :through => :taxable_taxonomies, :source => :taxable, :source_type => 'Ptable'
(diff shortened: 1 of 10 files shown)


app/models/taxonomy.rb
```diff
@@ -145,15 +145,15 @@ def dup
     #  if ignore?("Domain")
     #   Domain.pluck(:id)
     # else
-    #   self.taxable_taxonomies.where(:taxable_type => "Domain").pluck(:taxable_id)
+    #   super  # self.domain_ids
```
ares: `super()` which is different to `super`

ares:
One downside is that in audit page I only see that "template a was removed", previously I saw "partition table a was removed" but I suppose there's not much we can do about it. Otherwise it works well.

app/models/taxonomy.rb
domcleal: Thanks, comment fixed.

domcleal:
> One downside is that in audit page I only see that "template a was removed", previously I saw "partition table a was removed" but I suppose there's not much we can do about it.

Indeed, I can't see a reasonable solution to this without extending Audit further, or manipulating another field such as comment or auditable_name. Neither seem like great ideas to me, as they'll affect searching etc.

ares:
Same thoughts. Ready for merge pending jenkins.

dLobatog:
@ares Haven't checked this PR but Jenkins says it's ok

ares:
Thanks @domcleal, merged as 5908bf418d28c57dcdc4d2e01ccf3253c0e8a81a

PR #3746, opened 2016-08-19
fixes #16193 - change templates.type default to known STI class

domcleal:


dLobatog:
Merged as 9a1dfb99d676a177410e6bb54373f4ec2f49e06c, thanks @domcleal!

PR #3747, opened 2016-08-19
fixes #16194 - correct table name in remove_foreign_key call

domcleal:


dLobatog:
Merged as aeb74d31ad752779b2d3e65410e53297600f376d, thanks @domcleal!

PR #3748, opened 2016-08-19
fixes #16198 - remove therubyracer Bundler group

domcleal:
NodeJS must be installed instead, which ExecJS will use automatically.


dLobatog:
Merged as 5525e303ecb07c183249a4d38e45e33b2d265384, thanks @domcleal!

PR #3750, opened 2016-08-19
fixes #16199 - always run build related callbacks

timogoebel:
`logger.debug` might return false and causes the callback to be skipped.


lzap:
Nasty one, this always shows up here and there. :-(

Unrelated test failure, thanks.

lzap:
Merged as 31ddd916c745db7a6eb07f989706e8ed7dd3f98f, thank you!

PR #3752, opened 2016-08-19
Fixes #16197 - Host#smart_proxies returns only the host proxies

dLobatog:
Doing the following:
- Create 2 smart proxies, one for dhcp/dns/tft/puppetca/puppet, another
  for realms. Let's keep the realms proxy off, so it's unreachable.
- Create a hostgroup A that sets domain, subnet, realm, puppet_proxy,
  puppet_ca_proxy
- Create a host that uses hostgroup A, but uncheck 'inherit' and
  remove the 'realm' attribute. The host is created just fine with no
  realm
- When I try to rebuild the host, `HostBuildStatus#check_all_statuses`
  is called. This checks the host_status (OK), templates_status(OK), and
  smart_proxies_status(FAIL). The host cannot be rebuilt.

The reason `smart_proxies_status` fails is that `smart_proxy_ids`
tries to find whether Realm is set.
- If it's set, it adds the proxy to 'Host#smart_proxies`
- If it's not set, it looks in the hostgroup to see if it's set there.
  However, I have overridden the option, because I don't want to set a
  Realm since I don't want to boot the Realm proxy I have.

Host#smart_proxies should not look in the host.hostgroup object to
retrieve values for the proxy, but instead retrieve those from itself

### DIFF
--- a/app/models/concerns/hostext/smart_proxy.rb
+++ b/app/models/concerns/hostext/smart_proxy.rb
@@ -1,0 +1,22 @@
+module Hostext
+  module SmartProxy
+    extend ActiveSupport::Concern
+
+    def smart_proxies
+      ::SmartProxy.where(:id => smart_proxy_ids)
+    end
+
+    def smart_proxy_ids
+      ids = []
+      [subnet, subnet6].compact.each do |s|
+        ids << s.dhcp_id
+        ids << s.tftp_id
+        ids << s.dns_id
+      end
+      ids << domain.dns_id if domain.present?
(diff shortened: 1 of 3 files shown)


tbrisker:
Looks good, will merge if [test] is green

dLobatog:
@tbrisker They are greener than this apple :green_apple:

tbrisker:
Merged as 15f2f2918f60ec5af84196c9c5cd026ae5654731, thanks @dLobatog!

PR #3755, opened 2016-08-22
Fixes #16218 - override settings per environment

lzap:
This is a small change that will add possibility to setup loggers for test
environment. I don't like the hardcoded defaults and there is no way to change
that only for test environment. I want separate setups for development and
test.

### DIFF
--- a/config/settings.yaml.example
+++ b/config/settings.yaml.example
@@ -33,9 +33,12 @@
 :fqdn: 'localhost.localdomain.net'
 
 # Log settings for the current environment can be adjusted by adding them
-# here. For example, if you want to increase the log level.
+# here. It is also possible to override settings per environment. See
+# logging.yaml for more details and examples.
 #:logging:
 #  :level: debug
+#  :production:
+#    :type: syslog
 
 # Individual logging types can be toggled on/off here
 #:loggers:
--- a/lib/foreman/logging.rb
+++ b/lib/foreman/logging.rb
@@ -80,6 +80,7 @@ module Foreman
     def load_config(environment, overrides = {})
       fail "Logging configuration 'config/logging.yaml' not present" unless File.exist?('config/logging.yaml')
       overrides ||= {}
+      overrides = overrides[environment.to_sym] if overrides.has_key?(environment.to_sym)
       @config = YAML.load_file('config/logging.yaml')
       @config = @config[:default].deep_merge(@config[environment.to_sym]).deep_merge(overrides)
     end


config/settings.yaml.example
```diff
@@ -37,6 +37,11 @@
 #:logging:
 #  :level: debug
 
+# It is also possible to override settings per environment:
+#:logging:
+#  :test:
+#    :pattern: "[%c|%.1l] %m\n"
```
tbrisker: This is quite un-understandable in my POV... would it make sense to instead add an option to the loggers to only be environment specific? something line 

``` yml
:loggers:
  :app:
    :enabled: true #this would always log
  :ldap:
    :enabled: [production, test] #a list of more then one environment
  :sql:
    :enabled: development #a single environment
```

This would be slightly similar to what we have for the smart proxy where enabled option can take `true`,`false`,`http`,`https`

lzap: What's exactly isn't clear? What you suggest does not change a thing, I want to configure SYSLOG output and output format for particular environment. We already have the ability to override globally, this adds an option to add an environment. The old behavior still works fine.

lzap:
I rebased, there were unrelated test failures, this is ready for review.

ehelms:
:+1:  Makes sense to me to be able to override at a more global environment level instead of having to go change each and every logger.

config/settings.yaml.example
tbrisker: I guess what I don't understand is what the pattern does?

lzap: Do you mean the logging pattern? That's explained in `logging.yaml` but you are right that is not the best example. I changed it to more obvious one.

tbrisker: Could you please add a pointer in the comment to `logging.yaml` or somewhere in the docs for more info on the syntax so that users can easily find out what they can do here?

lzap: Sure, done.

dLobatog:
Thanks @lzap , just for the record I don't think this is worth talking about in the manual (as it's intended for production installations) and devs can already read the comments in this file.

lzap:
Thanks.

Here is a tip, if you do not like growing `log/development.log` or `log/test.log` (this can be huge sometimes), you can reconfigure it into syslog since override patch [1] was merged. It is possible also to change the pattern completely to something shorter to save terminal columns.

I like to redirect everything into syslog, because I use systemd-journald in memory only mode, which is convenient. I can still use logs during development or testing if I need to. Saves a lot of SSD writes, it's faster.

``` yaml
:logging:
  :development:
    :type: syslog
    :pattern: "[%c|%.1l] %m\n"
  :test:
    :type: syslog
    :pattern: "[%c|%.1l] %m\n"
    :level: debug
```

Here is the systemd-journald configuration for the record:

```
$ grep -v ^# /etc/systemd/journald.conf
[Journal]
Storage=volatile
Seal=no
RuntimeMaxUse=20M
ForwardToSyslog=no
ForwardToKMsg=no
ForwardToConsole=no
```

[1] https://github.com/theforeman/foreman/pull/3755

PR #3757, opened 2016-08-22
fixes #16220 - replace skip_callback class changes with stubs

domcleal:


tbrisker:
Merged as 8563088e6846876ea70ffd0a55c768feab01a589, thanks @domcleal!

PR #3758, opened 2016-08-22
Fixes #16057 - added dns_lookup template helper

lzap:
Also improved tests. Kindly ask reviewer to update our wiki page. Thanks!

### DIFF
--- a/lib/foreman/renderer.rb
+++ b/lib/foreman/renderer.rb
@@ -3,7 +3,7 @@ require 'tempfile'
 module Foreman
   module Renderer
     ALLOWED_GENERIC_HELPERS ||= [ :foreman_url, :snippet, :snippets, :snippet_if_exists, :indent, :foreman_server_fqdn,
-                                  :foreman_server_url, :log_debug, :log_info, :log_warn, :log_error, :log_fatal, :template_name ]
+                                  :foreman_server_url, :log_debug, :log_info, :log_warn, :log_error, :log_fatal, :template_name, :dns_lookup ]
     ALLOWED_HOST_HELPERS ||= [ :grub_pass, :ks_console, :root_pass,
                                :media_path, :param_true?, :param_false?, :match ]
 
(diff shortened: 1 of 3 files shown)


lib/foreman/renderer.rb
```diff
@@ -105,6 +105,18 @@ def indent(count)
       prefix + text.gsub(/\n/, "\n#{prefix}")
     end
 
+    def dns_lookup(name_or_ip)
+      resolver = Resolv::DNS.new
+      Timeout.timeout(Setting[:dns_conflict_timeout]) do
+        ipquery = IPAddr.new(name_or_ip) rescue false
```
dLobatog: @lzap Can we expect anything but a [IPAddr::Error](http://ruby-doc.org/stdlib-2.0.0/libdoc/ipaddr/rdoc/IPAddr/Error.html) here ?

test/lib/foreman/renderer_test.rb
```diff
@@ -49,6 +41,38 @@ def setup_safemode_renderer
       assert_equal 'bar', tmpl
     end
 
+    test "foreman_server_fqdn helper method" do
```
dLobatog, line 44: Mind to move all the `send('setup_#{renderer_name}')` calls to a setup block in the .each loop?

lib/foreman/renderer.rb
dLobatog: I tried with a variety of options, all of them fell under that class (most likely a subclass of it) of exception

dLobatog:
@lzap Thanks, this would be useful. Two minor comments inline, also can you add tests for `dns_lookup` with invalid parameters? For instance a wrong IP and an invalid argument

lib/foreman/renderer.rb
lzap: Sure.

test/lib/foreman/renderer_test.rb
lzap: Yup.

lzap: That's can't be created actually, the setup block creates a closure around renderer_name and it's not set appopriately.

lzap:
Rebased https://gist.github.com/8fd45cd74ef21d713c9c1ec78c6b02d7

dLobatog:
[test] was aborted

dLobatog:
Unrelated failure, merging

dLobatog:
Merged as 1ffe8b2b472730be60d4126eaeb1e56253df2cb2, thanks @lzap!

PR #3759, opened 2016-08-22
fixes #16221 - replace private AS timezone listing method

domcleal:


tbrisker:
Merged as abebb99425cb9d6c14f8067a12aea3013daa0dc0, thanks @domcleal!

PR #3760, opened 2016-08-22
Fixes #16163 - Hide spinner for vmware cluster

xprazak2:


ares:
[test]

app/assets/javascripts/compute_resource.js
```diff
@@ -248,7 +248,7 @@ function vsphereGetResourcePools(item) {
   var data = {cluster_id: $(item).val()};
   var url = $(item).data('url');
   tfm.tools.showSpinner();
-  selectbox = $('*[id*=resource_pool]');
+  selectbox = $('*[id=host_compute_attributes_resource_pool]');
```
tbrisker: I'm not sure why using a more specific selector would cause the spinner to hide? 
Any select box selected by this will have also been selected by the previous one, and I suspect there was a reason for using a more generalized selector before - perhaps this code is also used on other pages where the select box has a different ID? Even if not, matching on a precise id should be done with `$(#id)`, not `$(*[id=id])`.

timogoebel: it is also used on the compute profile page and has a different id there

xprazak2:
@tbrisker it was selecting a div as well:


@timogoebel than you for warning about compute profiles page

tbrisker:
Code looks good but I don't have a vmware set up to verify the fix. Will merge if someone can test this on vmware.

timogoebel:
I'll test this in a week and merge after that if no one has done it before.

@xprazak2: Do you have a vmware setup? I do have some prs open that make vmware better and need someone to take a look. :-)

xprazak2:
@timogoebel: I have access to vmware, I will try to find some time for your PRs.

ares:
Works nicely, thanks @xprazak2, merged as 105917c2f0cb2ef1200f2c5f12a304321b4f326d

PR #3762, opened 2016-08-22
Fixes #16044 - Load dashboard widgets via ajax

tbrisker:
This speeds up dasboard loading as widgets are loaded in the background
and do not block rendering, as well as allow the dashboard to load
correctly even if some of the widgets are broken.

### DIFF
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -2,8 +2,8 @@ class DashboardController < ApplicationController
   include Foreman::Controller::AutoCompleteSearch
   include Foreman::Controller::Parameters::Widget
 
-  before_action :prefetch_data, :only => :index
-  before_action :find_resource, :only => [:destroy]
+  before_action :prefetch_data, :only => :show
+  before_action :find_resource, :only => [:show, :destroy]
   skip_before_action :welcome
 
   def index
@@ -12,6 +12,16 @@ class DashboardController < ApplicationController
       format.yaml { render :text => @report.to_yaml }
       format.json
     end
+  end
+
+  def show
+    if @widget.present? && @widget.user == User.current
+      render(:partial => @widget.template, :locals => @widget.data)
+    else
+      render_403 "User #{User.current} attempted to access another user's widget"
+    end
+  rescue ActionView::MissingTemplate, ActionView::Template::Error => exception
+    process_ajax_error exception, "load widget"
   end
 
   def create
--- a/app/views/dashboard/index.html.erb
+++ b/app/views/dashboard/index.html.erb
@@ -17,7 +17,9 @@
               <a class='minimize'>&minus;</a>
             </div>
             <div class="widget <%= widget.name.parameterize %>">
-              <%= render_widget(widget) %>
+              <div data-ajax-url="<%= widget_path(widget) %>" data-on-complete="refreshCharts">
+                <%= spinner(_("%s widget loading...") % widget.name) %>
+              </div>
             </div>
           <% end %>
         <% end %>
--- a/app/helpers/dashboard_helper.rb
+++ b/app/helpers/dashboard_helper.rb
@@ -35,12 +35,6 @@ module DashboardHelper
     end
   end
 
-  def render_widget(widget)
-    render(:partial => widget.template, :locals => widget.data)
-  rescue ActionView::MissingTemplate
-    ::Foreman::Exception.new(N_("Missing template '%{template}' for widget '%{widget}'."), :widget => _(widget.name), :template => widget.template)
-  end
-
   def widget_data(widget)
     { :data => { :id    => widget.id,    :name  => _(widget.name), :row  => widget.row, :col => widget.col,
                  :sizex => widget.sizex, :sizey =>  widget.sizey,  :hide => widget.hide } }
--- a/app/services/foreman/access_permissions.rb
+++ b/app/services/foreman/access_permissions.rb
@@ -732,7 +732,7 @@ Foreman::AccessControl.map do |permission_set|
   end
 
   permission_set.security_block :dashboard do |map|
-    map.permission :access_dashboard, {:dashboard => [:index, :save_positions, :reset_default, :create, :destroy],
+    map.permission :access_dashboard, {:dashboard => [:index, :show, :save_positions, :reset_default, :create, :destroy],
                                       :"api/v1/dashboard" => [:index],
                                       :"api/v2/dashboard" => [:index]
     }
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -412,7 +412,7 @@ Foreman::Application.routes.draw do
 
   end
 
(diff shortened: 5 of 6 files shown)


app/controllers/dashboard_controller.rb
```diff
@@ -2,8 +2,8 @@ class DashboardController < ApplicationController
   include Foreman::Controller::AutoCompleteSearch
   include Foreman::Controller::Parameters::Widget
 
-  before_action :prefetch_data, :only => :index
-  before_action :find_resource, :only => [:destroy]
+  before_action :prefetch_data, :only => :show
```
tbrisker, line 5: This isn't ideal, as the same data will be loaded for each widget. In the next step this should be refactored so each widget only loads the data it actually needs, possibly using cache to prevent the same data from being loaded twice.

ohadlevy, line 5: Long term we can move the common widgets to be triggered via JavaScript with the same data. If this was done in react, you would have two components using the same store data ( where flux would fetch it only once)

app/controllers/dashboard_controller.rb
```diff
@@ -14,6 +14,16 @@ def index
     end
   end
 
+  def show
```
ohadlevy, line 17: missing permissions?

tbrisker: fixed

app/views/dashboard/index.html.erb
```diff
@@ -17,7 +17,7 @@
               <a class='minimize'>&minus;</a>
             </div>
             <div class="widget <%= widget.name.parameterize %>">
-              <%= render_widget(widget) %>
+              <div data-ajax-url="<%= widget_path(widget) %>" data-on-complete="refreshCharts"></div>
```
ohadlevy: can you please add a spinner?

app/controllers/dashboard_controller.rb
```diff
@@ -14,6 +14,16 @@ def index
...
+    if @widget.present? && @widget.user == User.current
+      render(:partial => @widget.template, :locals => @widget.data)
```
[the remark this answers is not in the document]
ohadlevy: it would be nice to log which widget we are rendering (not just the id) so we can troubleshoot more easily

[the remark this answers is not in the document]
ohadlevy: also, not sure if its related, but i got the following when trying to render the discovery widgets:

```
W] Failed to load widget
10:30:50 rails.1   |  | ActionView::MissingTemplate: Missing partial dashboard/_discovery_widget, application/_discovery_widget with {:locale=>[:en], :formats=>[:html, :html, :text, :js, :css, :ics, :csv, :vcf, :png, :jpeg, :gif, :bmp, :tiff, :mpeg, :xml, :rss, :atom, :yaml, :multipart_form, :url_encoded_form, :json, :pdf, :zip], :variants=>[], :handlers=>[:erb, :builder, :raw, :ruby, :rabl]}. Searched in:
10:30:50 rails.1   |  |   * "/home/ohad/git/foreman/app/views"
10:30:50 rails.1   |  |   * "/home/ohad/.rvm/gems/ruby-2.3.1@foreman/gems/apipie-rails-0.3.6/app/views"
10:30:50 rails.1   |  | 
```

note that the plugin path was not included?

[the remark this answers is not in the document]
tbrisker: Did you restart the rails server? it seems to work fine for me (tested with tasks plugin).
Also, the widgets are logged, e.g. 

```
2016-08-23T10:51:59 a9570e1f [app] [I] Started GET "/widgets/16" for 127.0.0.1 at 2016-08-23 10:51:59 +0300
2016-08-23T10:51:59 a9570e1f [app] [I] Processing by DashboardController#show as HTML
2016-08-23T10:51:59 a9570e1f [app] [I]   Parameters: {"id"=>"16"}
2016-08-23T10:52:00 a9570e1f [app] [I]   Rendered dashboard/_distribution_widget.html.erb (19.5ms)
2016-08-23T10:52:00 a9570e1f [app] [I] Completed 200 OK in 514ms (Views: 16.5ms | ActiveRecord: 235.1ms)
```

app/views/dashboard/index.html.erb
tbrisker: added

app/controllers/dashboard_controller.rb
ohadlevy, line 19: its logged only when it finished rendering,  i had of a problem to figure
out which widget was rendering, but i guess i could go both ways.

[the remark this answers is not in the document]
ohadlevy: oh my bad - the plugin was not in my bundle, so it actually worked :)

ohadlevy:
APJ

dLobatog:
@tbrisker Thanks! Notice integration tests for the dashboard didn't pass, I guess they have to wait for the widgets to load or something like that. Good proof that it's async now though :smile:

tbrisker:
@dLobatog updated the test to wait for ajax, let's hope this will please Jenkins :wink:

app/views/dashboard/index.html.erb
```diff
@@ -17,7 +17,9 @@
               <a class='minimize'>&minus;</a>
             </div>
             <div class="widget <%= widget.name.parameterize %>">
-              <%= render_widget(widget) %>
+              <div data-ajax-url="<%= widget_path(widget) %>" data-on-complete="refreshCharts">
+                <%= spinner _("Loading widget...") %>
```
dLobatog: How about displaying the name of what's being loaded, next to a spinner? e.g: `spinner _(widget.name)`

I looked up the names of all widgets and they seem to make sense, so it'd at least give some info about what's loading in case it takes more time than usual or whatever:

```
["Status table", "Status chart", "Report summary", "Distribution chart", "Discovery widget", "Status table", "Status chart", "Report summary", "Distribution chart", "Discovery widget", "Status table", "Status chart", "Report summary", "Distribution chart", "Discovery widget", "Status table", "Status chart", "Report summary", "Distribution chart", "Discovery widget", "Status table", "Status chart", "Report summary", "Distribution chart", "Tasks Status table", "Tasks in Error/Warning", "Discovery widget", "Status table", "Status chart", "Report summary", "Distribution chart", "Tasks Status table", "Tasks in Error/Warning", "Discovery widget", "Tasks Status table", "Tasks in Error/Warning"]
```

ohadlevy: :+1: that would be in line with how statistics ajax is done.

tbrisker: added

dLobatog:
Ready to merge if Jenkins tests pass

dLobatog:
Merged as 8685d6060088563e42456929531b7c2cac103d44, thanks @tbrisker!

PR #3763, opened 2016-08-22
Fixes #15806 - add babel-eslint dependency to package.json

gailsteiger:
Added missing dependency.


ohadlevy:
merged as 9d64124 thanks @gailsteiger

PR #3764, opened 2016-08-23
Fixes #16238 - Speed up dashboard loading time.

ohadlevy:
I've change report host to inner query and used a join, in
my testing, this speed up dramaticilly dashboard loading
time, on my setup from 60 seconds to under 10 seconds.

### DIFF
--- a/app/services/dashboard/data.rb
+++ b/app/services/dashboard/data.rb
@@ -17,7 +17,10 @@ class Dashboard::Data
 
   def latest_events
     # 9 reports + header fits the events box nicely...
-    @latest_events ||= ConfigReport.authorized(:view_config_reports).my_reports.interesting.where(:host_id => hosts.pluck(:id)).search_for('reported > "7 days ago"').limit(9).includes(:host)
+    @latest_events ||= ConfigReport.authorized(:view_config_reports).my_reports.interesting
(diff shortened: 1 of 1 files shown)


tbrisker:
Nice catch! Another small change I noticed that might also help would be to change the `eager_load` to `joins` here: https://github.com/theforeman/foreman/blob/develop/app/models/host/managed.rb#L82 as I don't think it really makes any sense to use a left join here

ohadlevy:
@tbrisker not sure, maybe @ares knows why was it initially required?

ohadlevy:
@tbrisker regardless, I think this one worth merging without host_status optimization since host_status has a wider effect, and might need to refactor the host_status usage from the dashboard vs from host list (for example).

tbrisker:
@ohadlevy Very well, opened http://projects.theforeman.org/issues/16245 to take care of too-eager eager-loading.

tbrisker:
Merged as 95fd95eaf0df8756b049dcaf993b4d9709d86401, thanks @ohadlevy!

ares:
I think @dLobatog might know - https://github.com/theforeman/foreman/commit/7fdb705b9698f773ddd15cd6d32b9a41b7e022d1 there also a good description in the commit.

PR #3768, opened 2016-08-23
fixes #16243 - make Host::Managed.new's options arg optional

domcleal:
The options argument no longer exists on Rails 5 as it was part of the
support for protected_attrs. The argument list to .new should be handled
more defensively, like Host::Base.


tbrisker:
Merged as 8e780e97b1651e778722a82a9a8d2d8e5cdb351b, thanks @domcleal!

PR #3770, opened 2016-08-23
Fixes #12544 - added errors from host to cancelBuild

ShimShtein:
Now the flash will contain detailed error messages.

### DIFF
--- a/test/functional/hosts_controller_test.rb
+++ b/test/functional/hosts_controller_test.rb
@@ -1278,6 +1278,19 @@ class HostsControllerTest < ActionController::TestCase
     assert_template :partial => '_interfaces'
   end
 
+  test 'failed cancelBuild shows errors' do
+    @request.env['HTTP_REFERER'] = hosts_path
+    HostsController.any_instance.stubs(:resource_finder).returns(@host)
+    @host.errors[:test] << 'my error'
+    @host.interfaces = [] # force save failure
+    get :cancelBuild, { id: @host.name }, set_session_user
+
+    assert_response :redirect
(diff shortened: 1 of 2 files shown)


ShimShtein:
Rebased, fixed test style.

test/functional/hosts_controller_test.rb
```diff
@@ -1278,6 +1278,21 @@ def test_submit_multiple_rebuild_config_pessimistic
...
+    @host.errors[:test] << 'my error'
+    @host.interfaces = [] # force save failure
+    get :cancelBuild, { id: @host.name }, set_session_user
+
+    assert_response :redirect
+    assert_match(/Failed to cancel/, flash[:error])
+    assert_match(/following errors/, flash[:error])
+    assert_match(/host must have/, flash[:error])
+
+    HostsController.any_instance.unstub(:find_resource)
```
dLobatog: No need to unstub it after the test, that happens automatically

dLobatog:
@ShimShtein Good addition, thanks! I always dug in the log for this things and didn't really give it much thought, unprivileged users can't do that. 

Minor comment on the test, it doesn't need to call 'unstub' at the end. I realize the context for 'when host is saved after setBuild' is calling it, but it's just a mistake. Please remove it from your test, and from the teardown too, repush and I'll be glad to merge. I tried it locally and tests passed.

ShimShtein:
Removed unnecessary `#unstub`

dLobatog:
Merged as 9b1708f79d0e180b88999c1c7db0d7dc2e39bd23, thanks @ShimShtein!

PR #3771, opened 2016-08-24
Fixes #16260 - Wait for dashboard to fully load in test

tbrisker:
This will hopefully prevent intermittent db deadlocks.


domcleal:
Merged as 1706113, thanks @tbrisker.

PR #3773, opened 2016-08-24
refs #12544 - replace ordered interpolation with named variables

domcleal:


tbrisker:
Merged as bfa131a8fc3d38c351f1b4945dc5638f27716f22, thanks @domcleal!

PR #3774, opened 2016-08-24
Fixes #2314 - fine grain for RAM selectors

ares:


ohadlevy:
@ares - generic question - does it make sense to write this as part of webpack new coding standards etc? /cc @tbrisker and @gailsteiger

ares:
Probably, I started this before webpack was there and I'm happy to modify it if someone provides guidance, I'm not sure what is the final set of tools though. E.g. do we enforce es2015 now?

tbrisker:
@ares es2015 is preferable, and linting is also in place during development for all the webpack/ directory to help you along. I didn't look into the code too much yet, but I could try and help you if you run into any issues.
This would also be good opportunity to move jquery.ui to npm, which should reduce its footprint (as only the used parts will be compiled into the webpack bundle).

dLobatog:
:+1: to moving this to our new assets in webpack (tests appreciated). We do enforce ES2015 under the webpack directory.

app/assets/javascripts/jquery.ui.custom_spinners.js
```diff
@@ -0,0 +1,105 @@
...
+  });
+
+  $.widget("ui.byte_spinner", $.ui.limited_spinner, {
+    options: {
+      step: 1,
+      min: 1,
+      incremental: false,
+      value_target: null
+    },
+    _megabyte: function() {
```
tbrisker: Why function instead of value?

app/helpers/form_helper.rb
```diff
@@ -203,6 +203,20 @@ def autocomplete_f(f, attr, options = {})
...
+  def byte_size_f(f, attr, options = {})
+    base_id = ActionView::Helpers::Tags::TextField.new(f.object_name, attr, nil, options).send :add_default_name_and_id, {}
+    options[:class] = options[:class].to_s + ' byte_spinner'
+    options[:help_inline] ||= popover('', _("When specifying custom value, add 'MB' or 'GB' at the end. Field is not case sensitive and MB is default if unspecified."))
+    options[:help_block] ||= content_tag(:span, :class => 'maximum-limit hidden') do
+      content_tag(:i, '', :class => 'pficon-warning-triangle-o') +
+      content_tag(:span, ' ' + _('Specified value is higher than recommended maximum'), :class => 'error-message')
+    end
+
+    javascript_include_tag("jquery.ui.custom_spinners.js") +
```
tbrisker: This will add the js multiple times on some pages and in the middle of the page, please only include once in pages that need it.

app/models/compute_resources/foreman/model/libvirt.rb
```diff
@@ -58,12 +58,13 @@ def max_cpu_count
       hypervisor.cpus
     end
 
-    # libvirt reports in KB
+    # returns available memory for VM in bytes
     def max_memory
-      hypervisor.memory * Foreman::SIZE[:kilo]
+      # libvirt reports in KB
+      hypervisor.memory * 1.kilobyte
```
tbrisker: `hypervisor.memory.kilobyte`?

app/helpers/form_helper.rb
```diff
@@ -203,6 +203,20 @@ def autocomplete_f(f, attr, options = {})
     end
   end
 
+  def byte_size_f(f, attr, options = {})
+    base_id = ActionView::Helpers::Tags::TextField.new(f.object_name, attr, nil, options).send :add_default_name_and_id, {}
+    options[:class] = options[:class].to_s + ' byte_spinner'
+    options[:help_inline] ||= popover('', _("When specifying custom value, add 'MB' or 'GB' at the end. Field is not case sensitive and MB is default if unspecified."))
+    options[:help_block] ||= content_tag(:span, :class => 'maximum-limit hidden') do
+      content_tag(:i, '', :class => 'pficon-warning-triangle-o') +
+      content_tag(:span, ' ' + _('Specified value is higher than recommended maximum'), :class => 'error-message')
```
tbrisker, line 200: Should this be validated by the server?

app/assets/javascripts/jquery.ui.custom_spinners.js
```diff
@@ -0,0 +1,105 @@
...
+      if (value % this._gigabyte() == 0) {
+        return (value / this._gigabyte()) + ' GB';
+      } else {
+        return (value / this._megabyte()) + ' MB';
+      }
+    }
+  });
+
+  $("input.byte_spinner").each(function () {
+    var target = $(this).parents('div').children('span.help-block').children('span.maximum-limit');
```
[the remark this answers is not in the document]
tbrisker: `$(this)` should be cached in a variable to save DOM lookups.
also, if you just want to go one level up use `.parent()`, or in this case, possibly `.siblings()`.

config/initializers/assets.rb
```diff
@@ -18,6 +18,8 @@
                   host_edit
                   host_edit_interfaces
                   hosts
+                  jquery.cookie
```
tbrisker: Why?

tbrisker:
@ares I opened https://github.com/theforeman/foreman/pull/3811 to move jquery.ui to npm which should help migrate this to webpack as well

app/assets/javascripts/application.js
```diff
@@ -2,6 +2,7 @@
 //= require turbolinks
 //= require i18n
 //= require jquery.ui.autocomplete
+//= require jquery.ui.spinner
```
gailsteiger, line 5: why use jquery.ui.spinner instead of patternfly?

app/assets/javascripts/jquery.ui.custom_spinners.js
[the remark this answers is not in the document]
ares: > $(this) should be cached in a variable to save DOM lookups.

:+1: 

> also, if you just want to go one level up use .parent(), or in this case, possibly .siblings().

I need to go more levels up, IIRC 2 levels up to the div but if it makes more sense I think I can change that to `.parent().siblings('span.help-block').children('span.maximum-limit')`, but it seems like hardcoding the path to exact DOM structure which is more fragile I think especially without tests

app/assets/javascripts/jquery.ui.custom_spinners.js
ares: no reason anymore, fixing

app/helpers/form_helper.rb
ares: I don't think so, the thing is that that many CR hardcode [some maximum](https://github.com/theforeman/foreman/blob/develop/app/models/compute_resources/foreman/model/ovirt.rb#L102) which should not limit the users if they know what they're doing. Later I'd like to remove artificial limits but that needs more CR code refactoring first. So now we can just display a warning.

config/initializers/assets.rb
ares: not needed anymore, removing

app/assets/javascripts/application.js
ares: patternfly spinner didn't allow me to customize the output and parsing - note that gb/mb is part of the input so user can type "5gb" and it will get parsed correctly

webpack/assets/javascripts/jquery.ui.custom_spinners.js
```diff
@@ -0,0 +1,111 @@
...
+  });
+
+  $.widget("ui.byte_spinner", $.ui.limited_spinner, {
+    options: {
+      step: 1,
+      min: 1,
+      incremental: false,
+      value_target: null
+    },
+    _megabyte: 1024 * 1024,
```
tbrisker: these could be set in the module context as `const megabyte = 1024 **2`. Since this is a module that won't pollute the global namespace and save all of the annoying `this._megabyte` etc

webpack/assets/javascripts/jquery.ui.custom_spinners.js
```diff
@@ -0,0 +1,115 @@
...
+
+$(function () {
+  $.widget('ui.limited_spinner', $.ui.spinner, {
+    options: {
+      softMaximum: 0,
+      errorTarget: null
+    },
+    _validate: function () {
+      if (this.options.softMaximum !== 0) {
+        if (this.value() > this.options.softMaximum) {
```
tbrisker: I think this could be done in one line - 
`this.options.errorTarget.toggle(this.value() > this.options.softMaximum);`

webpack/assets/javascripts/jquery.ui.custom_spinners.js
```diff
@@ -0,0 +1,114 @@
...
+      if (this.options['softMaximum'] != 0) {
+        if (this.value() > this.options['softMaximum']) {
+          this.options["errorTarget"].show();
+        } else {
+          this.options["errorTarget"].hide();
+        }
+      }
+    },
+    _spin: function (step, event) {
+      let result = this._super(step, event);
```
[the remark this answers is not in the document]
tbrisker: Any reason to save this in a variable instead of just `return this._super(step, event);` after running the validation?

webpack/assets/javascripts/jquery.ui.custom_spinners.js
```diff
@@ -0,0 +1,115 @@
...
+        if (this.value() % gigabyte === 0) {
+          step = -1 * gigabyte;
+        } else {
+          step = -1 * (this.value() % gigabyte);
+        }
+      }
+      return step;
+    },
+    _megabyteSpin: function (step) {
+      let megabyteStep = step * 256 * megabyte;
```
tbrisker: should this be `const`?

webpack/assets/javascripts/jquery.ui.custom_spinners.js
```diff
@@ -0,0 +1,115 @@
...
+      let result = null;
+
+      if ((this.value() > gigabyte && step < 0) ||
+        (this.value() >= gigabyte && step > 0)) {
+        step = this._gigabyteSpin(step);
+      } else {
+        step = this._megabyteSpin(step);
+      }
+
+      result = this._super(step, event);
```
tbrisker, line 76: again, any reason to save this rather then directly returning it?

webpack/assets/javascripts/jquery.ui.custom_spinners.js
```diff
@@ -0,0 +1,115 @@
+const megabyte = 1024 * 1024;
+const gigabyte = 1024 * megabyte;
+
+$(function () {
+  $.widget('ui.limited_spinner', $.ui.spinner, {
+    options: {
+      softMaximum: 0,
+      errorTarget: null
+    },
+    _validate: function () {
```
tbrisker, line 13: This doesn't seem to trigger on manual changes of the field (ie not using the spinner)

webpack/assets/javascripts/jquery.ui.custom_spinners.js
ares: there's one more call after this but I want to keep the same interface for this function so returning whatever `_super` returns

webpack/assets/javascripts/jquery.ui.custom_spinners.js
```diff
@@ -0,0 +1,115 @@
...
+      }
+
+      return (value / megabyte) + ' MB';
+    }
+  });
+});
+
+export function initByteSpinner() {
+  $('input.byte_spinner').each(function () {
+    let jqThis = $(this);
```
tbrisker: nitpick: `$this = $(this);` is the common idiom, unless a more descriptive name is used (field maybe?)

webpack/assets/javascripts/jquery.ui.custom_spinners.js
```diff
@@ -0,0 +1,115 @@
...
+    let target = jqThis.parents('div').children('span.help-block').children('span.maximum-limit');
+
+    jqThis.byte_spinner({
+      valueTarget: jqThis.attr('id') + '_hidden',
```
tbrisker: `this.id === jqThis.attr('id')`

webpack/assets/javascripts/jquery.ui.custom_spinners.js
```diff
@@ -0,0 +1,115 @@
...
+    },
+    _spin: function (step, event) {
+      let result = this._super(step, event);
+
+      this._validate();
+      return result;
+    }
+  });
+
+  $.widget('ui.byte_spinner', $.ui.limited_spinner, {
```
tbrisker: Please use camel case for functions even if they are defined using a factory which makes linting miss that

webpack/assets/javascripts/jquery.ui.custom_spinners.js
tbrisker: Correct me if i'm wrong, but the other call doesn't modify this?

webpack/assets/javascripts/jquery.ui.custom_spinners.js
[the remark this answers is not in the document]
ares: to keep the interface of method I'm overriding

webpack/assets/javascripts/jquery.ui.custom_spinners.js
ares: It does not but `_spin` must be called before we update the hidden which holds the real value, so we need to call super and then the update. But I want to keep method interface therefore returning the result of super call.

webpack/assets/javascripts/jquery.ui.custom_spinners.js
```diff
@@ -0,0 +1,114 @@
...
+        if (value.match(/gb$/i)) {
+          return parseFloat(value) * gigabyte;
+        } else if (value.match(/mb$/i) || parseInt(value, 10) < megabyte) {
+          return parseInt(value, 10) * megabyte;
+        }
+      }
+      return value;
+    },
+    // prints value with unit, if it's multiple of gigabytes use GB, otherwise format in MB
+    _format: function (value) {
```
tbrisker: This could be also done in one line using an arrow function:

``` javascript
_format: value => (value % gigabyte === 0) ? (value / gigabyte) + ' GB' : (value / megabyte) + ' MB';
```

app/helpers/form_helper.rb
```diff
@@ -192,6 +192,19 @@ def autocomplete_f(f, attr, options = {})
...
+    end
+    options[:onchange] = "$(this).byte_spinner('validate'); #{options[:onchange]}"
```
tbrisker: you have `onchange` here, you also add `onchange` listener below in the initialization, would it make sense to have one `changed` function that handles everything instead of multiple ones being called?

webpack/assets/javascripts/jquery.ui.custom_spinners.js
```diff
@@ -0,0 +1,114 @@
...
+
+      return (value / megabyte) + ' MB';
+    }
+  });
+});
+
+export function initByteSpinner() {
+  $('input.byte_spinner').each(function () {
+    let jqThis = $(this);
+    let target = jqThis.parents('div').children('span.help-block').children('span.maximum-limit');
```
tbrisker: the comment about this line got hidden, but you can use 
`.closest('.form-group').find('.maximum-limit')` to make sure you find the correct one. 
also, perhaps a better name for this variable?

config/initializers/assets.rb
```diff
@@ -18,6 +18,7 @@
                   host_edit
                   host_edit_interfaces
                   hosts
+                  jquery.ui.custom_spinners
```
tbrisker: This is no longer needed

ares: are you sure? I thought we have to wait for https://github.com/theforeman/foreman/pull/3811

tbrisker: yes, this is compiled by webpack not sprockets

app/helpers/form_helper.rb
```diff
@@ -192,6 +192,18 @@ def autocomplete_f(f, attr, options = {})
     end
   end
 
+  def byte_size_f(f, attr, options = {})
+    base_id = ActionView::Helpers::Tags::TextField.new(f.object_name, attr, nil, options).send :add_default_name_and_id, {}
```
tbrisker: This is a private function that may break in the future, why not just push the hidden field into one of the help blocks with no id and find it like we do for the warning message?

config/initializers/assets.rb
ares: I misread the line, I thought it was the generic ui spinner lib, ok makes sense, thanks

app/helpers/form_helper.rb
ares: right, makes more sense

app/helpers/form_helper.rb
```diff
@@ -192,6 +192,18 @@ def autocomplete_f(f, attr, options = {})
     end
   end
 
+  def byte_size_f(f, attr, options = {})
+    options[:class] = options[:class].to_s + ' byte_spinner'
+    options[:help_inline] ||= popover('', _("When specifying custom value, add 'MB' or 'GB' at the end. Field is not case sensitive and MB is default if unspecified."))
+    options[:help_block] ||= content_tag(:span, :class => 'maximum-limit hidden') do
+      content_tag(:i, '', :class => 'pficon-warning-triangle-o') +
+      content_tag(:span, ' ' + _('Specified value is higher than recommended maximum'), :class => 'error-message') +
+      f.hidden_field(attr, :class => "real-hidden-value")
```
tbrisker: unless you specify `:id=> nil` this field will get the same id as the fake one.
Also, I would add this after the block so it is added even if there is a different `options[:help_block]` - i.e. `options[:help_block] += f.hidden_field....`.
You can also do without adding a class - `.find('input[type=hidden]')` below will also work.

ares:
Thanks @tbrisker, updated. I kept the class to make it less fragile. Please take another look. I'll rebase other PRs once this is considered final.

Rohoover:
@tbrisker 
Would this be changeable to fit the requirement:
http://www.patternfly.org/pattern-library/forms-and-controls/time-picker/#_

Instead of hours or minutes, an arrow could control the numeric value and the button toggle between GB/MB?

Alternatively, if this does not meet your needs, I would box in the arrows. The majority of input fields do not let the attached actions float within the field. Example: http://www.patternfly.org/pattern-library/widgets/#bootstrap-combobox
The down arrow is boxed in. 

I can provide a mock-up with a few options if you'd like.

tbrisker:
@Rohoover this should look like a number input field (i.e. a number that can be manually entered or incremented/decremented by buttons/arrow keys), I could not find anything similar in patternfly. The time picker looks like overkill here.

ares:
I'd be happy to work on this further but I hope that design concerns do not block the actual fix which allows user to bypass hardcoded limits. Please don't get me wrong, I appreciate all suggestions I just think we can address it in separate PR. I can open redmine issue for that if you prefer.

Rohoover:
@tbrisker 
Makes sense. I've documented the fact that PatternFly does not have a standard for this. Instead of up/down arrows that are very small and close together, what about +/-? Here are a couple of quick mockups.
quanitityfield

app/helpers/form_helper.rb
```diff
@@ -192,6 +192,18 @@ def autocomplete_f(f, attr, options = {})
...
   end
 
+  def byte_size_f(f, attr, options = {})
+    options[:class] = options[:class].to_s + ' byte_spinner'
+    options[:help_inline] ||= popover('', _("When specifying custom value, add 'MB' or 'GB' at the end. Field is not case sensitive and MB is default if unspecified."))
+    options[:help_block] ||= content_tag(:span, :class => 'maximum-limit hidden') do
+      content_tag(:i, '', :class => 'pficon-warning-triangle-o') +
+      content_tag(:span, ' ' + _('Specified value is higher than recommended maximum'), :class => 'error-message')
+    end
+    options[:help_block] += f.hidden_field(attr, :class => "real-hidden-value", :id => "#{f.object.object_id}_hidden_real_value")
```
tbrisker: I'd prefer not having an id (`:id => nil`) here unless there is any reason for it. `f.object` may have multiple byte size fields.

vendor/assets/stylesheets/jquery.ui.spinner_custom.scss
```diff
@@ -0,0 +1,12 @@
+span.ui-spinner {
```
tbrisker: this file shouldn't be under vendor, but under app (similar to other scss files we have for overriding defaults from libraries)

tbrisker:
@Rohoover Thanks for the input. I think we'll go with just wrapping the arrows in buttons for now, and reiterate if and when patternfly suggests a better styling.

Rohoover:
@tbrisker Sounds good. I also contribute to PatternFly, there are opportunities to create functionality and float it over.

ares:
[test katello]

ares:
[test foreman] [test upgrade] seems like some networking issue

tbrisker:
Merged as eabbbeb4730abb51de2de8f9b302d26b14d6d33a, thanks @ares!

ares:
\o/

PR #3775, opened 2016-08-24
Fixes #16072 - Fixes compute profile two pane

tstrachota:
Fixes compute profile two-pane behaviour in compute resource detail page and
adds support for using id '#two_pane_content' for element that should be
replaced after closing the right pane. '#content' is still used by default.

### DIFF
--- a/app/views/compute_resources/show.html.erb
+++ b/app/views/compute_resources/show.html.erb
@@ -51,7 +51,7 @@
     </div>
   <% end %>
   <div id="compute_profiles" class="tab-pane">
-    <table class="<%= table_css_classes 'table-two-pane table-fixed' %>">
+    <table class="<%= table_css_classes 'table-two-pane table-fixed' %>" id="two_pane_content">
       <thead>
         <tr>
           <th class="col-md-8"><%= _("Compute profile") %></th>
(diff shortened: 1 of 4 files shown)


tstrachota:
Do we really need `Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.` check? It doesn't make much sense especially in the requires.

tbrisker:
@tstrachota disregard those comments, it seems the hound isn't picking up our rubocop settings correctly

daviddavis:
It looks like it's using [hound's config](https://git.io/v67hc) instead of [ours](https://git.io/v67hz). It does look like Hound should support inherit_from: https://github.com/houndci/hound/pull/971.

Also, why is hound passing this PR since we have it set to fail PRs with violations?

https://github.com/theforeman/foreman/blob/develop/.hound.yml#L15

tstrachota:
[test]

xprazak2:
Tested, takes care of the issue for me.

app/views/compute_resources/show.html.erb
```diff
@@ -51,7 +51,7 @@
     </div>
   <% end %>
   <div id="compute_profiles" class="tab-pane">
-    <table class="<%= table_css_classes 'table-two-pane table-fixed' %>">
+    <table class="<%= table_css_classes 'table-two-pane table-fixed' %>" id="two_pane_content">
```
tbrisker, line 54: what if we wanted to use two pane in another tab?

tstrachota: ...then it won't work. But it didn't work before either :smile: (not that I would test it before but the code already searches for the two-pane classes globally)

ares:
Keeping open until Monday so @tbrisker or anyone else can comment further but I think it's ready for merge. If there are no objections, I'll merge then. Btw thanks @tstrachota for integration test :+1:

dLobatog:
@ares @tbrisker This has been on RTM for nearly a week, anything missing?

ares:
[test foreman]

ares:
[test katello]

ares:
Thanks @tstrachota, merged as d5ff4a673446f65d188bcaeb71764d739213db7e

PR #3779, opened 2016-08-25
Fixes #14101 - Exception triggered when editing/adding a role

dhlavac:


ares:
[test]

ares:
[test]

ares:
Could you please add some reproducing steps? I failed to reproduce. I created a user which had all unlimited location permissions and permission to edit roles. I tried to let users without any locations as well as associate all. Roles editing and creation work just fine.

app/services/association_authorizer.rb
```diff
@@ -3,7 +3,11 @@ def self.authorized_associations(associations, klass_name = nil, should_raise_ex
     if associations.included_modules.include?(Authorizable)
       associations_klass = associations
       if associations.respond_to?(:klass)
-        associations_klass = associations.klass
+        if associations.klass == Taxonomy
```
ares: I was finally able to reproduce. After more digging I don't think this is the right place to fix it. When you check the backtrace you can see that this is called from https://github.com/theforeman/foreman/blob/105917c2f0cb2ef1200f2c5f12a304321b4f326d/app/helpers/form_helper.rb#L74 and that this method has optional argument called `klass_name`. The only thing that we need to do is passing right class when multiple_selects is used to generate taxonomies field. Something like this should work

``` ruby
def multiple_selects(f, attr, associations, selected_ids, options = {}, html_options = {})
    options.merge!(:size => "col-md-10")
    klass = case attr
              when :organizations
                Organization
              when :locations
                Location
              else
                nil
            end
    authorized = AssociationAuthorizer.authorized_associations(associations, klass).all
    ....
```

dhlavac:
@ares something else to do ?

ares:
I guess the test is not unrelated and can be removed, looks good otherwise.

dhlavac:
Test removed

dLobatog:
Merged as 8fde28895ec88297e88413e86ea191828bfabc7c, thanks @dhlavac!

PR #3782, opened 2016-08-26
Fixes #14468 - fix db:migrate when unattended false on empty db

orrabin:


dLobatog:
Merged as 4a3cb08b2847e27dd2256c24521834bd44e50296, thanks @orrabin!

PR #3787, opened 2016-08-28
Refs #16222 - remove no-global-assign rule from eslint

tbrisker:
Hound uses an older version of eslint that does not support this rule
leading to confusing comments on PRs. The rule is not crucial and can be
removed.


gailsteiger:
This is fine. I don't think it's a rule that many will try to break.

PR #3791, opened 2016-08-29
Fixes #16372 - Convert the header logo to svg instead of png

bkearney:


lzap:
That's increasing size of each request by 100k!

```
[lzap@lzapx foreman]$ ll app/assets/images/header_logo* -h
-rw-rw-r-- 1 lzap lzap 4.3K Mar 23 09:06 app/assets/images/header_logo.png
-rw-rw-r-- 1 lzap lzap  90K Aug 30 09:36 app/assets/images/header_logo.svg
```

Vectorization in Inkscape is not the way to go. We need to put the original instead. But I still think it would be better to provide two or three responsive PNG images:

``` css
.responsive-img {
     width: 100%;
     height: auto;
}
```

lzap:
I mean different images for different sizes, which gives the best possible quality:

``` css
    @media only screen and (max-width: 600px) {
        #image1 {
            background-image: url("image-small.png");
        }
    }
    @media only screen and (min-width: 601px) and (max-width: 900px) {
        #image1 {
            background-image: url("image-medium.png");
        }
    }
    @media only screen and (min-width: 901px)  {
        #image1 {
            background-image: url("image-large.png");
        }
    }
```

lzap:
By the way the largest version of the PNG I was able to find so far is in our git history (was removed): https://github.com/theforeman/foreman/blob/fe5010723b86f26018fc7fb4ee054c734173b5a2/app/assets/images/foreman_large.png

lzap:
For the record, the helmet was made by Mairín, we should be able to get the original (maybe SVG, definitely higher res).

Edit: I realized this is the original (https://github.com/theforeman/foreman-graphics/tree/master/logo) but it's still too big. I am playing with SVG optimizer to bring it down a bit, looks good so far. Hold on.

lzap:
I realized this is the original (https://github.com/theforeman/foreman-graphics/tree/master/logo) but it's still too big. I am plying with SVG optimizer to bring it down a bit, looks good so far. Hold on.

lzap:
So I was able to optimize original SVG logo to the size of the PNG (4.3K). The result is 4.6K which is not bad at all, but this is SVGZ version, I am not sure if browsers supports it. If not, we can still go with SVG version (16K) and configure web server to compress it. We already have asset pre-compressing for RPM installations.

https://github.com/theforeman/foreman-graphics/pull/1

bkearney:
@lzap next steps? Should I resubmit the pr with the smaller icon?

lzap:
Sorry for the confusion, we just merged the optimized SVG, please resubmit with it:

https://github.com/theforeman/foreman-graphics/blob/master/logo/foreman-web.svg

lzap:
Ok looks like our Rails precompiler is configured for SVG already, so I made a small change in the httpd.conf to serve gzipped file instead:

https://github.com/theforeman/foreman/blob/develop/config/initializers/assets.rb#L64

https://github.com/theforeman/puppet-foreman/pull/475

bkearney:
@lzap I force pushed a new version with the svg you suggested. I updated the commit message to reference where the image came from.

lzap:
It works great, before I merge, I want to point out that SVG is _not_ supported in IE8 and older. All other browsers supports it for years now: http://caniuse.com/#feat=svg

@bkearney are we fine with that?

tbrisker:
@lzap I think it's about time we dropped IE 8 support, however note that there seem to be some issues with scaling svgs on all IE versions - I don't have IE to test it, but we should try at least with a recent version to make sure we don't break the entire layout.

bkearney:
@lzap I am fine with IE 10+

bkearney:
@lzap FYI, that answer is based off of downstream support statements, and looking at https://upload.wikimedia.org/wikipedia/commons/7/74/Timeline_of_web_browsers.svg which tells me IE10 came out 4 years ago which I think is the typical life of desktops.

lzap:
I can only test in MSIE11, it does not stretch correctly:

vystrizek

lzap:
Ok I tested this with MSIE11 and it looks like specifying the size works fine, it renders fine in both. Please update the line with:

```
    <%= image_tag("header_logo.svg", size: "44x34") %>
```

That will work in all cases.

app/views/home/_topbar.html.erb
```diff
@@ -3,7 +3,7 @@
     <div class="container">
       <div class="navbar-brand">
         <!--[if IE]><div class='logo-small'> <![endif]-->
-        <%= image_tag("header_logo.png") %>
```
tbrisker, line 5: Please remove the old PNG if we don't use it anymore.

lzap:
Good point, remove it please. Also I furthermore optimized the web SVG, if this gets merged, please use it: https://github.com/theforeman/foreman-graphics/pull/2

bkearney:
@lzap I updated the style, but did it in the css instead of the html.
@tbrisker the old file is removed.

lzap:
I am waiting until https://github.com/theforeman/foreman-graphics/pull/2 is merged and then I will ask you to pull the more optimized version. I think it's important to do everything we can do to save bandwidth.

lzap:
I am in Brno tomorrow, might not check out github for the rest of the week, so feel free to merge. If the above patch gets merged, please put the more optimized version.

lzap:
More optimized version was merged, please rebase once again @bkearney with the smallest version and we are good to go. Thanks everyone.

bkearney:
@lzap rebased and pushed with the latest image.

app/assets/stylesheets/patternfly_and_overrides.scss
```diff
@@ -121,6 +121,10 @@ a {
   }
 }
 
+.header-logo{
+  size: "44x34";
```
tbrisker: This isn't valid css.

lzap: Bryan, can you change that to

```
width: 44px;
height: 34px;
```

bkearney:
@lzap fixed and rebased.

PR #3792, opened 2016-08-30
Fixes #16379 - Improve layout of hostname randomize button

tbrisker:
This moves the button from being a line of its own to an icon inside the
input. Also added a form_feedback option to the field helper to allow
easy addition of more icons in inputs in the future.

### DIFF
--- a/app/assets/stylesheets/patternfly_and_overrides.scss
+++ b/app/assets/stylesheets/patternfly_and_overrides.scss
@@ -216,7 +216,6 @@ body.modal-open {
 .input-addon {
   color: black;
 }
-// eof form-control-feedback
 
 legend {
   margin-bottom: 10px;
--- a/app/helpers/form_helper.rb
+++ b/app/helpers/form_helper.rb
@@ -15,17 +15,6 @@ module FormHelper
     end
   end
 
-  def button_input_group(content, options = {}, glyph = nil)
-    options[:type] ||= 'button'
-    options[:herf] ||= '#'
-    options[:class] ||= 'btn btn-default'
-    content_tag :span, class: 'input-group-btn' do
-      content_tag :button, content, options  do
-        content_tag :span,content, :class => glyph
-      end
-    end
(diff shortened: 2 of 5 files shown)


tbrisker:
Before:
screenshot from 2016-08-30 11-02-11

After:
screenshot from 2016-08-30 11-01-29

@rohoover - I'd be happy for your input

ares:
I think it deserves its own redmine issue.

tbrisker:
@ares thanks, updated

ares:
Works nicely, ACK pending jenkins and feedback from @rohoover

ohadlevy:
@tbrisker how about having it as a btn inside the input, similar to the date picker at https://www.patternfly.org/pattern-library/forms-and-controls/date-picker ?

tbrisker:
@ohadlevy I personally think it looks nicer not as a button, but I may be wrong. I'll do whatever @rohoover says will work better.

Rohoover:
Visually I am not bothered by either approach so I have to default to the closest standard for consistency which would be what @ohadlevy suggested.

Leaving it "naked" without the button would be a new pattern or exception.

tbrisker:
Changed to a button, updated screenshot:
screenshot from 2016-08-31 15-46-20

ohadlevy:
@tbrisker what do you see if the feature is disabled?

tbrisker:
@ohadlevy just a regular text input with no button

ohadlevy:
ack assuming this is the right icon to use (didnt find a better one).

Rohoover:
This is the correct icon. It's labeled as such in font awesome and is the generally accepted icon for "random" in a multitude of applications. +1

lzap:
Works as expected, thanks.

ehelms:
[test]

lzap:
Merged as 3b6f9323cd186c355a8519d887cc13c30c5e2276, thank you!

PR #3793, opened 2016-08-30
fixes #16376 - use correct params filter for nested Puppet LKs

domcleal:


orrabin:
:+1:  works for me

tbrisker:
Merged as ddf80e132ed9087e527c108f5fc0f2f5e0ac7639, thanks @domcleal!

PR #3794, opened 2016-08-30
Fixes #16092 - Test connection buttons don't disapper on error

tstrachota:
Adds new type of inline help that doesn't disappear when there is a validation error for the field.
It also moves the spinner onto the "Test connection" button. Previously it was placed right next to the field and validation errors were displayed between the spinner and the button.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -573,3 +573,12 @@ table {
 .masked-input {
   font-family: 'AllBullets';
 }
+
+.btn-spinner {
+  .spinner {
+    margin: 0.3em;
+  }
+  .caption {
+    float: right;
+  }
+}
--- a/app/helpers/compute_resources_helper.rb
+++ b/app/helpers/compute_resources_helper.rb
@@ -65,4 +65,25 @@ module ComputeResourcesHelper
   def unset_password?
     action_name == "edit" || action_name == "test_connection"
   end
+
+  def test_connection_button_f(f, success, caption = nil)
+    caption ||= _("Test Connection")
+    btn_class = success ? 'btn-success' : 'btn-default'
+    spinner_class = success ? 'spinner-inverse' : nil
+    spinner_button_f(f, caption, "testConnection(this)",
+                     :id => 'test_connection_button',
+                     :spinner_id => 'test_connection_indicator',
+                     :class => btn_class,
+                     :spinner_class => spinner_class,
+                     :'data-url' => test_connection_compute_resources_path)
+  end
+
+  def load_button_f(f, success, failure_caption)
+    caption = success ? _("Test Connection") : failure_caption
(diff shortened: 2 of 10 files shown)


tstrachota:
[test]

tbrisker:
I think it might make more sense to move this button into the input-group instead of introducing yet another option for that already complex method, this should be trivial using the changes introduced in https://github.com/theforeman/foreman/pull/3792. @rohoover - your input would also be appreciated here.

Rohoover:
@tbrisker 
Anyway I can get a screenshot of this?

tstrachota:
**Load datacenters button** 
using inline help position:
load_datacenters_inline_help
using input group position:
load_datacenters_input_grp

**Test connection button** 
using inline help position:
test_connection_inline_help
using input group position:
test_connection_input_grp

tstrachota:
While I agree it would be good to use existing position instead if adding more complexity, I don't think it's good to add visual connection between the dropdown and the button. Especially in the case of test connection button it doesn't seem correct to me, because testing connection has nothing to do with loading the datacenters from users perspective (even though we know it's the same api call in the background).

tbrisker:
@tstrachota I think that by placing the button on the same line we already have that association, maybe i would be best to place this button in a different position all together?

Rohoover:
I don't think it should be attached to the field for the reasons @tstrachota listed above. There are also already fields with icon buttons attached to them. This could give the permission to string multiple buttons to a field, it's a slippery slope to bad UI. Additionally, if you look at the size of the drop down button with relation to positioning another attached button, user miss clicks could increase.

If the "Test Connection" or "Load Data Center" isn't related to the fields at all perhaps its possible to move it down one line.

In a related note but not necessarily actionable for this issue - In line help (field level help) would be better placed next to the label. Here it is outlined in PatternFly. http://www.patternfly.org/pattern-library/forms-and-controls/field-level-help/#_

tstrachota:
I created screenshots of the full form with buttons below. I still like the buttons next to the field more, but having them below works for me too

**Load button**
load_datacenters_below
**Test connection button**
test_connection_below

tstrachota:
Since I'm leaving on longish PTO and I won't be able to adjust the PR according to outcome of the button placement discussion, I added one more commit that moves the buttons below the datacenters dropdown. Feel free to either squash it or leave it off based on what position you like more.

ares:
@tbrisker @Rohoover what is the conclusion here? based on last comments I think the last version that Tomas posted is preferred?

Rohoover:
@ares Yes, that is my impression.

tbrisker:
I'm willing to go with what @Rohoover says, she's the UX expert.

ares:
Thanks @tstrachota, merged as 02e8b5e7fb93b332c3f0f7a7f35c81d7e50d64ca

PR #3796, opened 2016-08-30
Fixes #13047, #16291 - Upgrade datatables and move it to npm

tbrisker:


ehelms:
[test]

app/assets/javascripts/application.js
```diff
@@ -100,7 +100,7 @@ function onContentLoad(){
         $(this).html(response);
       }
```
gailsteiger, line 101: It's not efficient to repeat jquery operations. For example, in this function, $(this) should be cached at the top of the function, like this: 

``` js
var $this = $(this);
```

and then $(this) replaced with $this throughout the function like this:

``` js
$(this).removeAttr('data-ajax-url');
```

becomes

``` js
$this.removeAttr('data-ajax-url');
```

dLobatog:
All good IMO, @gailsteiger can you do the $this caching in another PR? I think I'll merge this and update the packages.

tbrisker:
@dLobatog I've updated the PR to cache the jQuery object, at least on the lines I touched. Please verify this works on oVirt, only lines changed are those used for displaying the VMs using ajax calls (which is currently only used there).

dLobatog:
Merged as 77e12c8572085010eed3d2a42bb2cfa28de49d94, thanks @tbrisker!

dLobatog:
I'll create the new packages for lodash and datatables via [npm2rpm](https://github.com/dlobatog/npm2rpm)

PR #3800, opened 2016-08-31
Refs #15814 - Reset override params only for PuppetclassLookupKey

orrabin:


app/models/lookup_keys/lookup_key.rb
```diff
@@ -21,14 +21,13 @@ class LookupKey < ActiveRecord::Base
...
   before_validation :cast_default_value
-  after_validation :reset_override_params, :if => ->(key) { key.override_changed? && !key.override? }
 
   validates :key, :presence => true
   validates :validator_type, :inclusion => { :in => VALIDATOR_TYPES, :message => N_("invalid")}, :allow_blank => true, :allow_nil => true
   validates :key_type, :inclusion => {:in => KEY_TYPES, :message => N_("invalid")}, :allow_blank => true, :allow_nil => true
   validate :validate_default_value
   validates_associated :lookup_values
-  validate :disable_merge_overrides, :disable_avoid_duplicates, :disable_merge_default, :if => :override?
+  validate :disable_merge_overrides, :disable_avoid_duplicates, :disable_merge_default, :if => :override? || type != 'PuppetclassLookupKey'
```
ares: you need to use lambda here I think

orrabin: I could also remove it, since all 3 are turning to false the errors won't be added anyway.

daviddavis: Don't the parameters get set to false after validation though?

orrabin: Yeah but they couldn't be true without overriding anyway.
This prevents from checking without a reason but if the `if` is more complicated it might be worth just getting to the validation and checking it

ehelms:
[test]

db/migrate/20160817125655_reset_override_params.rb
```diff
@@ -3,7 +3,7 @@ class LookupKey < ::ActiveRecord::Base
   end
 
   def up
-    LookupKey.where(override: false).update_all(
+    PuppetclassLookupKey.where(override: false).update_all(
```
tbrisker, line 3: I'm not sure this would use the fake class correctly, might be better to use `LookupKey.where(override: false, type: 'PuppetclassLookupKey')...`

app/models/lookup_keys/lookup_key.rb
```diff
@@ -21,14 +21,13 @@ class LookupKey < ActiveRecord::Base
...
   before_validation :cast_default_value
-  after_validation :reset_override_params, :if => ->(key) { key.override_changed? && !key.override? }
 
   validates :key, :presence => true
   validates :validator_type, :inclusion => { :in => VALIDATOR_TYPES, :message => N_("invalid")}, :allow_blank => true, :allow_nil => true
   validates :key_type, :inclusion => {:in => KEY_TYPES, :message => N_("invalid")}, :allow_blank => true, :allow_nil => true
   validate :validate_default_value
   validates_associated :lookup_values
-  validate :disable_merge_overrides, :disable_avoid_duplicates, :disable_merge_default, :if => :override?
+  validate :disable_merge_overrides, :disable_avoid_duplicates, :disable_merge_default, :if =>  Proc.new { |lk| lk.override || !lk.puppet? }
```
tbrisker, line 30: Doesn't have to be in this pr, but I would consider combining these to one custom validator and including it with a different condition in the children, which would get rid of complex condition and separate concerns better.

tbrisker:
:+1: other then the inline comments.

app/models/lookup_keys/lookup_key.rb
orrabin: I'll open a refactoring issue after this pr.

db/migrate/20160817125655_reset_override_params.rb
orrabin: I'll remove the fake class

tbrisker:
Merged as d7f8cc79691f69cd4bddc818565cf70925162949, thanks @orrabin!

PR #3801, opened 2016-08-31
Fixes #16389 - enable taxable object creation in specific context

ares:


app/models/concerns/taxonomix.rb
```diff
@@ -36,7 +36,8 @@ def with_taxonomy_scope(loc = Location.current, org = Organization.current, inne
       self.which_location        = Location.expand(loc) if SETTINGS[:locations_enabled]
       self.which_organization    = Organization.expand(org) if SETTINGS[:organizations_enabled]
       scope = block_given? ? yield : where('1=1')
-      scope = scope.where(:id => taxable_ids) if taxable_ids
+      tax_ids = taxable_ids
+      scope = scope.where(:id => tax_ids) if tax_ids
```
ares: unrelated performance boost, taxable_ids is not cached

ehelms:
[test]

iNecas:
APJ, if there was emoji for :respect: I would use one :)

ares:
@iNecas one more change was needed, we have to be explicit about the table name and we can't escape the values inside `IN ()`, these injections should be always safe though. Please take another look, thanks.

dLobatog:
Merged as 35ed04f95f14c934ffebc874e90a3a4a6684fe71, thanks @ares!

PR #3802, opened 2016-08-31
Fixes #16397 - set nested objects to optional for template_combinations

ShimShtein:
Fixes a bug introduced in ee66b3af2f9497404734ef39f3d60a64a14b3928, nested objects are not required in this case: you should still be able to use `/template_combinations/:id` url.


ehelms:
[test]

domcleal:
Please add a functional test for the unnested case (both show and destroy), there are only tests for the nested case.

ShimShtein:
Tests added

domcleal:
Merged as d33fe9c, thanks @ShimShtein.

(Commit message changed - adding one new line into the summary doesn't follow correct message format.)

PR #3803, opened 2016-09-01
Fixes #13936 - Drop support for legacy browsers

tbrisker:
This patch removes various workarounds needed to get legacy browsers to
work. Mostly stuff needed for IE<10, but also some other vendor prefixes
that are no longer needed.


ohadlevy:
I didn't look, but with stats like: `+22 −99,102` lines changed you got the MVP commit of the day :100:

tbrisker:
@ohadlevy I somehow need to get rid of the +98k I got when merging the AllBullets.svg so I can go back to having a net negative impact :laughing:

lzap:
I was wondering if we could implement a warning banner when any of the unsupported browser is in use as part of this patch.

tbrisker:
We already have one for IE<=9, do you think we also need one for other?

app/assets/stylesheets/typography.scss
```diff
@@ -1,10 +1,7 @@
 @font-face {
   font-family: 'Overpass';
   src: local("Overpass"),
-  font-url('overpass_regular-web.eot?#iefix') format("eot"),
-  font-url('overpass_regular-web.woff') format("woff"),
-  font-url('overpass_regular-web.ttf') format("truetype"),
-  font-url('overpass_regular-web.svg#webfontLTZe4IYH') format("svg");
+       font-url('overpass_regular-web.woff') format("woff");
```
tbrisker, line 4: All modern browsers support woff, no need to ship other (heavier) formats.

dLobatog:
Merged as bbefa32cd3908d08712a78588a5cf2c04130791c, thanks @tbrisker!

PR #3804, opened 2016-09-01
Fixes #16378 - Move default_value casting to lookup_key children

orrabin:
For puppetclass_lookup_key casting/validation should only be done when override is true so each child of lookup_key needs to decide when to call them.


dLobatog:
Merged as 1ebc15ea40ebc4eaddc7afb8e83753841a724b1d, thanks @orrabin!

PR #3807, opened 2016-09-01
Fixes #16219 - Interfaces API works with scoped view_hosts

dLobatog:
Before this commit, parent_scope called 'merge' on a scope that may
contain conditions that do not make sense on 'resource_class'.

In this case, when a user with a filter :view_hosts and search
'hostgroup_fullname = foo' tried to view
api/v2/hosts/somehost/interfaces, that would not work.

The SQL generated by `scope_for` contains
`AND ((hostgroups.title = 'base'))`, which clearly cannot be
merged with the Nic::Base scope, as 'hostgroups' isn't a field in
Nic::Base.

To avoid this problem, we resolve the scope and call 'where' on the
scope ids, instead of merging the parent and the child scope.

Notice this problem exists for any child/parent relationship where the
user has a filter :view_parent , search 'parent_attribute = foo' if
parent_attribute is not defined in the child.

---

The old parent scope was wrong, by the way. It was a simple '"SELECT \"hosts\".\* FROM \"hosts\" WHERE \"hosts\".\"type\" IN ('Host::Managed')"' which does NOT apply any authorization filters, the fix for the CVE fixed that.

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -56,7 +56,22 @@ module Api
       association = resource_class.reflect_on_all_associations.find {|assoc| assoc.plural_name == parent_name.pluralize}
       #if couldn't find an association by name, try to find one by class
       association ||= resource_class.reflect_on_all_associations.find {|assoc| assoc.class_name == parent_name.camelize}
-      resource_class.joins(association.name).merge(scope)
+      result_scope = resource_class.joins(association.name).merge(scope)
+      # Check that the scope resolves before return
+      result_scope if result_scope.to_a
+    rescue ActiveRecord::ConfigurationError
+      # Chaining SQL with a parent scope does not always work, as the
+      # parent scope might have attributes the resource_class does not have.
+      #
+      # For example, chaining 'interfaces' with a parent scope (hosts) that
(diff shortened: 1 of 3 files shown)


ares:
Tests are failing, seems related to the change.

dLobatog:
Updated - I decided to only resolve the scope if we found that error where the parent_scope contains attributes the resource_name does not have.

dLobatog:
Changed the base_controller spec to expect `.to_a` to be called before returning the scope.

dLobatog:
@ares Ping? I think this is a problem for @nbarrientos 's production deployment

ares:
[test foreman]

ares:
APJ

dLobatog:
@ares The test that's failing is the intermittent problem we have with the dashboard in integration tests - http://ci.theforeman.org/job/test_develop_pr_core/10244/

ares:
I think tests are about to finish and so far seems green, the "default" seems to be pre jenkins plugin change

ares:
Thanks @dLobatog, merged as 3357cdaf190445fb38bd29de6f217c005dbc2e9e, nasty one.

PR #3808, opened 2016-09-02
Fixes #16424 - extracted create_new_host_when_facts_are_uploaded

lzap:
Discovery plugin needs to override this to true otherwise it can't import any
facts.

### DIFF
--- a/app/models/host/base.rb
+++ b/app/models/host/base.rb
@@ -102,9 +102,13 @@ module Host
       super - [ inheritance_column ]
     end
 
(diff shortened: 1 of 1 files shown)


ares:
I think you should use Foreman own redmine issue for this. Other than that it makes sense.

lzap:
Renamed.

dLobatog:
Merged as f0e9c7769eeaca238536f42e991508f372729a83, thanks @lzap!

PR #3811, opened 2016-09-04
Fixes #16285 - Move jQuery.ui from bundle to npm

tbrisker:


tbrisker:
Looks like removing the jquery-ui gem from core breaks katello. This is because they also require it, but don't limit its version to <5 causing version 5.x to be installed which apparently isn't compatible with their code..

ohadlevy:
@tbrisker it seems like this was merged on katello side, what is the status on this pr?

ohadlevy:
@tbrisker bump?

tbrisker:
Closing this PR for now, not worth the effort IMHO - we should replace the 2 places in core where we use jquery-ui (autocomplete and memory size fields) with react components and drop it completely.

PR #3812, opened 2016-09-06
Fixes #16439 - Don't fail recreate if DNS is ok

agx:
If rebuilding a host and DNS is not "feasible" we don't want to fail
that host but rather recreate what's currently feasible.

If a DNS record is still valid we don't want to fail that host but
rather take the still valid entry.

### DIFF
--- a/app/models/concerns/dns_interface.rb
+++ b/app/models/concerns/dns_interface.rb
@@ -35,7 +35,8 @@ module DnsInterface
   end
 
   def recreate_dns_record(type)
-    set_dns_record(type) unless dns_record(type).nil? || dns_record(type).valid?
+    return true if dns_record(type).nil? || dns_record(type).valid?
+    set_dns_record(type)
   end
 
   def set_dns_record(type)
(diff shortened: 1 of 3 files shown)


app/models/concerns/dns_interface.rb
```diff
@@ -35,7 +35,7 @@ def dns_feasible?(type)
   end
 
   def recreate_dns_record(type)
-    set_dns_record(type) unless dns_record(type).nil? || dns_record(type).valid?
+    dns_record(type).nil? || dns_record(type).valid? ? true : set_dns_record(type)
```
timogoebel: I think, a guard clause would make this more readable.

``` ruby
def recreate_dns_record(type)
  return true if dns_record(type).nil? || dns_record(type).valid?
  set_dns_record(type)
end
```

timogoebel:
@agx: Thanks for the patch. It looks good to me. Could you just add a tiny unit test for this?

agx:
@timogoebel test added and code changed as suggested

timogoebel:
[test foreman]

timogoebel:
@agx: Looks good, thank you. Test results are unrelated.

timogoebel:
Merged as 088c8f3d6fdbe9da1041a3812dae103f5c5a0cee, thanks @agx!

PR #3813, opened 2016-09-06
Fixes #10395 - fine grain for CPU selectors

ares:
based on https://github.com/theforeman/foreman/pull/3774

@tbrisker would you mind to take a look?

### DIFF
--- a/app/assets/javascripts/host_edit.js
+++ b/app/assets/javascripts/host_edit.js
@@ -75,7 +75,7 @@ function computeResourceSelected(item){
         activate_select2('#compute_resource');
         if ($('#compute_resource').find('.alert-danger').length > 0) $('#compute_resource_tab a').addClass('tab-error');
         update_capabilities($('#capabilities').val());
-        tfm.initByteSpinner();
+        tfm.numFields.initAll();
       }
     })
   }
--- a/app/helpers/form_helper.rb
+++ b/app/helpers/form_helper.rb
@@ -195,13 +195,25 @@ module FormHelper
   def byte_size_f(f, attr, options = {})
     options[:class] = options[:class].to_s + ' byte_spinner'
     options[:help_inline] ||= popover('', _("When specifying custom value, add 'MB' or 'GB' at the end. Field is not case sensitive and MB is default if unspecified."))
-    options[:help_block] ||= content_tag(:span, :class => 'maximum-limit hidden') do
-      content_tag(:i, '', :class => 'pficon-warning-triangle-o') +
-      content_tag(:span, ' ' + _('Specified value is higher than recommended maximum'), :class => 'error-message')
-    end
+    options[:help_block] ||= soft_limit_warning_block
     options[:help_block] += f.hidden_field(attr, :class => "real-hidden-value", :id => nil)
 
     text_f(f, attr, options)
+  end
+
+  def counter_f(f, attr, options = {})
+    options[:class] = options[:class].to_s + ' counter_spinner'
+    options[:help_block] ||= soft_limit_warning_block
+
+    text_f(f, attr, options)
+  end
+
+  def soft_limit_warning_block
+    content_tag(:span, :class => 'maximum-limit hidden') do
+      icon_text('warning-triangle-o',
+                content_tag(:span, ' ' + _('Specified value is higher than recommended maximum'), :class => 'error-message'),
+                :kind => 'pficon')
+    end
   end
 
   def form_to_submit_id(f)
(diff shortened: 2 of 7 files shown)


app/assets/javascripts/application.js
```diff
@@ -2,6 +2,7 @@
 //= require turbolinks
 //= require i18n
 //= require jquery.ui.autocomplete
+//= require jquery.ui.spinner
```
ohadlevy, line 5: can you provide more info on why patternfly spinner is not enough? I was hoping to get rid of jquery.ui when we rewrite the search autocompleter.... /cc @Rohoover

ohadlevy, line 5: also, a small image / animated gif would be great to see the ui changes :) thanks!

ares: Copying answer from first PR https://github.com/theforeman/foreman/pull/3774#discussion_r77506994

> patternfly spinner didn't allow me to customize the output and parsing - note that gb/mb is part of the input so user can type "5gb" and it will get parsed correctly

This cpu spinner just reuses some code that was added for the more advanced spinner. If there's a way to use patternfly one for this I'm happy to refactor this in next Foreman version. That should not block this PR I hope.

Here's the gif for simple CPU fields, changing the values using up/down keys

vmcpu

Here's the RAM being added in the first PR 
vmram

ares:
[test foreman]

app/assets/javascripts/host_edit.js
```diff
@@ -76,6 +76,7 @@ function computeResourceSelected(item){
         if ($('#compute_resource').find('.alert-danger').length > 0) $('#compute_resource_tab a').addClass('tab-error');
         update_capabilities($('#capabilities').val());
         tfm.initByteSpinner();
+        tfm.initCounterSpinner();
```
tbrisker: It seems like these two always come together, maybe have an `initFieldSpinners` function that runs both (and any more if needed in the future) instead of repeating?

app/helpers/form_helper.rb
```diff
@@ -204,6 +204,16 @@ def byte_size_f(f, attr, options = {})
     text_f(f, attr, options)
   end
 
+  def counter_f(f, attr, options = {})
+    options[:class] = options[:class].to_s + ' counter_spinner'
+    options[:help_block] ||= content_tag(:span, :class => 'maximum-limit hidden') do
+      content_tag(:i, '', :class => 'pficon-warning-triangle-o') +
+        content_tag(:span, ' ' + _('Specified value is higher than recommended maximum'), :class => 'error-message')
+    end
```
tbrisker: these 4 lines are common with the byte_size_f above, consider extracting to a function

app/assets/javascripts/host_edit.js
ares: sounds good, but we should keep both also available separately right?

tbrisker: yes, maybe import everything under `tfm.numFields` (so we can use e.g. `tfm.numField.initAll`, `tfm.numField.initByte` etc) - also hide away the confusing "spinner" name on the way.

ares:
thanks @tbrisker, I updated the code, could you please take another look?

webpack/assets/javascripts/bundle.js
```diff
@@ -9,5 +9,9 @@ require('./bundle_select2');
 require('./bundle_datatables');
 window.tfm = {
   tools: require('./foreman_tools'),
-  initByteSpinner: require('./jquery.ui.custom_spinners').initByteSpinner
+  numFields: {
```
tbrisker: no need to import each function separately, you can just use `numFields: require('./jquery.ui.custom_spinners')` and all exports will be imported (like we do for tools)

ares: ok, changed, please take another look

ares:
just for the record, failures are unrelated

ares:
[test foreman]

app/helpers/form_helper.rb
```diff
@@ -195,15 +195,26 @@ def autocomplete_f(f, attr, options = {})
...
+  def counter_f(f, attr, options = {})
+    options[:class] = options[:class].to_s + ' counter_spinner'
+    options[:help_block] ||= soft_limit_warning_block
+
+    text_f(f, attr, options)
+  end
+
+  def soft_limit_warning_block
+    content_tag(:span, :class => 'maximum-limit hidden') do
+      content_tag(:i, '', :class => 'pficon-warning-triangle-o') +
```
tbrisker: `icon_text(...)`?

ares: ok, sadly icon_text does not support block but I don't to introduce it in this commit...

tbrisker:
Merged as 80963357e57ddfb6969b9de7e5ca4636e36b4182, thanks @ares!

PR #3814, opened 2016-09-06
Fixes #5901 - fine grain for vmware CPU selectors

ares:
based on #3813 

@tbrisker would you mind to take a look?

### DIFF
--- a/app/views/compute_resources_vms/form/vmware/_base.html.erb
+++ b/app/views/compute_resources_vms/form/vmware/_base.html.erb
@@ -1,7 +1,9 @@
 <%= javascript 'compute_resource' %>
+<%= javascript_tag("$(document).on('ContentLoad', tfm.numFields.initCounter)"); %>
 <%= text_f f, :name, :label => _('Name'), :disabled => !new_host if show_vm_name? %>
-<%= selectable_f f, :cpus, 1..compute_resource.max_cpu_count, { }, :class => "col-md-2", :disabled => !new_host, :label => _('CPUs') %>
-<%= selectable_f f, :corespersocket, 1..compute_resource.max_cpu_count, { }, :class => "col-md-2", :disabled => !new_host, :label => _('Cores per socket') %>
+<%= counter_f f, :cpus, :disabled => !new_host, :label => _('CPUs'), :label_size => 'col-md-2', :'data-soft-max' => compute_resource.max_cpu_count %>
+
+<%= counter_f f, :corespersocket, :disabled => !new_host, :label => _('Cores per socket'), :label_size => 'col-md-2', :'data-soft-max' => compute_resource.max_cpu_count, :value => f.object.corespersocket || 1 %>
 <%= text_f f, :memory_mb, :class => "col-md-2", :disabled => !new_host, :label => _("Memory (MB)") %>
 <%= selectable_f f, :cluster, compute_resource.clusters, { :include_blank => _('Please select a cluster') },
                     :class => "col-md-2", :disabled => !new_host,


timogoebel:
Just my 2 cents, not a full review:

"Cores per socket" is initially empty for me. I guess, it shouldn't be.
The small "triangles" look odd. They should probably look similar so a select2 selectbox.

ares:
> "Cores per socket" is initially empty for me. I guess, it shouldn't be.

that was because fog object does not set this value in contrary to cpus, I restored default to be 1

> They should probably look similar so a select2 selectbox.

you mean two buttons? I don't think it would look very nice. Or you mean grey color?

timogoebel:
> you mean two buttons? I don't think it would look very nice. Or you mean grey color?

i meant the latter, grey color ;-)

ares:
Hm, I tried to enforce the color but it seems the jquery uses image sprite so I'm not sure how to change the color. Maybe some css3 transformation but that does not sound right.

ares:
[test upgrade]

ares:
the new version of base functionality added buttons for arrows so now it looks better IMHO

timogoebel:
Yes, it's better. Still looks a bit off, sorry. But that might just be a matter of taste.

image

Gray arrow vs black arrow...

timogoebel:
👍 from me though, when the base functionality is in.

ares:
[test foreman]

ares:
Rebased after #3813 was merged, @tbrisker would you mind looking at this one too? You already know other spinner PRs :-)

tbrisker:
@ares https://github.com/theforeman/foreman/pull/3829 finally got us unit test for JS, care to write a few?

app/views/compute_resources_vms/form/vmware/_base.html.erb
```diff
@@ -1,7 +1,9 @@
 <%= javascript 'compute_resource' %>
+<%= javascript_tag("$(document).on('ContentLoad', tfm.numFields.initCounter)"); %>
 <%= text_f f, :name, :label => _('Name'), :disabled => !new_host if show_vm_name? %>
-<%= selectable_f f, :cpus, 1..compute_resource.max_cpu_count, { }, :class => "col-md-2", :disabled => !new_host, :label => _('CPUs') %>
-<%= selectable_f f, :corespersocket, 1..compute_resource.max_cpu_count, { }, :class => "col-md-2", :disabled => !new_host, :label => _('Cores per socket') %>
+<%= counter_f f, :cpus, :disabled => !new_host, :label => _('CPUs'), :label_size => 'col-md-2', :'data-soft-max' => compute_resource.max_cpu_count %>
+
+<%= counter_f f, :corespersocket, :disabled => !new_host, :label => _('Cores per socket'), :label_size => 'col-md-2', :'data-soft-max' => compute_resource.max_cpu_count, :value => f.object.corespersocket || 1 %>
 <%= text_f f, :memory_mb, :class => "col-md-2", :disabled => !new_host, :label => _("Memory (MB)") %>
```
tbrisker, line 7: Why not change this also to the new ByteSpinner?

ares: The MB field returns the value in bytes, vmware requires megabytes, we need some unification layer for compute resources first or I could enhance the byte field so the unit for real value would be configurable. I'd prefer the first approach, since then the form or at least this field for now, would be the same.

tbrisker: I think the server should take care of converting the units to whatever the CR provider needs, the user shouldn't care about it nor the widget

ares:
> #3829 finally got us unit test for JS, care to write a few?

I'll be happy to try since I wanted to learn it anyway, but if possible in separate PR for 2 reasons, 1) I don't want to block this PR, 2) I hope to get this in 1.13.1 as the previous spinner PR since it's very related. I opened http://projects.theforeman.org/issues/16601 to track it.

app/views/compute_resources_vms/form/vmware/_base.html.erb
ares: I agree, but that's nontrivial big change. To make it work properly we need to migrate all compute profiles and hosts, the migration will be a bit tricky since it's serialized in yaml also currently the code passes the value to fog directly. Ideally we'll have one layer between fog and UI which will be generic so UI can be same for all compute resources where it make sense.

timogoebel:
Let me know if you need anyone to test this. I can do it easily.

tbrisker:
@timogoebel I'd be happy if you could verify on vmware, thanks. Code is OK to go on my part, feel free to merge once you verify.
@ares please do open a pr with tests for both spinners to make sure we don't break these, thanks.

timogoebel:
Merged as 0cd2ddf0cff1b07aeca2e2dee835937fde613b8b, thanks @ares!

PR #3815, opened 2016-09-06
Fixes #16380 - Eager load hosts attributes

shlomizadok:
Since Rails 4, in order to load delegated attributes,
we need to use 'eager_load' instead of 'includes'

### DIFF
--- a/app/controllers/api/v2/hosts_controller.rb
+++ b/app/controllers/api/v2/hosts_controller.rb
@@ -29,7 +29,8 @@ module Api
       param_group :search_and_pagination, ::Api::V2::BaseController
 
       def index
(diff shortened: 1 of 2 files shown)


dLobatog:
Thanks @shlomizadok . This is the old missing .references problem we had when updating to Rails 4.

```
Host.all.includes([ :host_statuses, :compute_resource, :hostgroup, :operatingsystem, :interfaces, :token ]).references
([ :host_statuses, :compute_resource, :hostgroup, :operatingsystem, :interfaces, :token ]).to_sql == Host.all.eager_load([ :host_status
es, :compute_resource, :hostgroup, :operatingsystem, :interfaces, :token ]).to_sql
```

dLobatog:
Merged as 7c5b8bd63f0fb7695b1b6e09a20e917004eca839, thanks @shlomizadok!

PR #3816, opened 2016-09-06
Fixes #16449 - reordered taxonomy detection during import

lzap:
Discovery need to have taxonomy set before the import.


dLobatog:
Merged as c49eb07f93f6821d7a7cad518ef537cd9ca41478, thanks @lzap!

PR #3819, opened 2016-09-07
Fixes #15992 - Truncate compute resources name in about page

shlomizadok:


app/views/about/index.html.erb
```diff
@@ -80,7 +80,7 @@
                 <tbody>
                 <% @compute_resources.each do |compute| %>
                   <tr>
-                    <td><%= link_to compute.name, compute %></td>
+                    <td><%= link_to(trunc_with_tooltip(compute.name), compute) %></td>
```
tbrisker: please don't use `trunc_with_tooltip` for this, give the table class `table-fixed`, set the column widths in the head and add class `ellipsis` to any cells that should be truncated.

shlomizadok: Changed. Thanks.

tbrisker:
Merged as 00042c21ca75bec98d246905733d6bedf5e9a06c, thanks @shlomizadok!

PR #3820, opened 2016-09-07
Fixes #16414 - reword alphanumeric validation error message

shlomizadok:


dLobatog:
It's not clear to me the new error says anything new, "underscored" is also kind of hard to translate, at least I don't know any translations for it. Deferring the review to others

ares:
maybe "must be alphanumeric with underscores and cannot contain spaces or hyphens"

shlomizadok:
@domcleal - I would appreciate your help with the right rewording :) - thanks

domcleal:
@ares' suggestion is most definitely better. Perhaps also add separate checks for spaces and hyphens if you want to mention them in the error, then add an error stating `cannot contain spaces` only if they're found. It keeps the default message much simpler ("must be alphanumeric with underscores"), and Rails will concatenate errors automatically.

Or simpler still, "must only contain alphanumeric or underscore characters". Should be clearer than a list of things that it can't contain.

shlomizadok:
Thanks @domcleal , this reads better.

dLobatog:
Thanks @ares @shlomizadok @domcleal !

PR #3824, opened 2016-09-07
Fixes #16463 - Fix with_taxonomy_scope when taxable_ids = []

dLobatog:
The old implementation of taxonomy_scope was relying on calling
`scope.where(:id => [])`, it did it 'if taxable_ids'. A recent change
introduced more checking into that condition (taxable_ids.present?)
which made the case when taxable_ids == [] not run any scope.

That results in objects not being scoped at all when taxable_ids
is an empty array.

To reproduce, simply go to any page with scoped content, like
smart proxies or compute resources, and see how the scoped content
still shows up when choosing the wrong taxonomy.

### DIFF
--- a/app/models/concerns/taxonomix.rb
+++ b/app/models/concerns/taxonomix.rb
@@ -36,12 +36,7 @@ module Taxonomix
       self.which_location        = Location.expand(loc) if SETTINGS[:locations_enabled]
       self.which_organization    = Organization.expand(org) if SETTINGS[:organizations_enabled]
       scope = block_given? ? yield : where('1=1')
-      tax_ids = taxable_ids
-      # we need to generate part of SQL query as a string, otherwise the default scope would set id on each
-      # new instance and the same taxable_id on taxable_taxonomy objects
-      if tax_ids.present?
-        scope = scope.where("#{self.table_name}.id IN (#{tax_ids.join(',')})")
-      end
+      scope = scope_by_taxable_ids(scope)
       scope.readonly(false)
     end
 
@@ -115,6 +110,21 @@ module Taxonomix
 
     def admin_ids
       User.unscoped.where(:admin => true).pluck(:id) if self == User
+    end
+
+    def scope_by_taxable_ids(scope)
+      case (cached_ids = taxable_ids)
+      when nil
+        scope
+      when []
+        # If *no* taxable ids were found, then don't show any resources
+        scope.where(:id => [])
+      else
+        # We need to generate the WHERE part of the SQL query as a string,
+        # otherwise the default scope would set id on each new instance
+        # and the same taxable_id on taxable_taxonomy objects
+        scope.where("#{self.table_name}.id IN (#{cached_ids.join(',')})")
+      end
     end
   end
 
--- a/test/unit/taxonomix_test.rb
+++ b/test/unit/taxonomix_test.rb
@@ -61,21 +61,51 @@ class TaxonomixTest < ActiveSupport::TestCase
     assert_empty @dummy.locations
   end
 
-  test ".with_taxonomy_scope expands organizations and locations to actual values" do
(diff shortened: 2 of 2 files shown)


dLobatog:
@ares Pinging you because few people know this area as well as you :grin:

app/models/concerns/taxonomix.rb
```diff
@@ -116,13 +111,27 @@ def inner_select(taxonomy, inner_method = which_ancestry_method)
...
+      when nil
+        scope
+      when []
+        # If *no* taxable ids were found, then don't show any resources
+        scope.where(:id => [])
+      else
+        # We need to generate the WHERE part of the SQL query as a string,
+        # otherwise the default scope would set id on each new instance
+        # and the same taxable_id on taxable_taxonomy objects
+        scope.where("#{self.table_name}.id IN (#{taxable_ids.join(',')})")
```
tbrisker: `taxable_ids` should remain cached to prevent calling this expensive function twice

app/models/concerns/taxonomix.rb
```diff
@@ -116,13 +111,27 @@ def inner_select(taxonomy, inner_method = which_ancestry_method)
...
+      end
+    end
   end
 
   def set_current_taxonomy
-    if self.new_record? && self.errors.empty?
-      self.locations     << Location.current     if add_current_location?
-      self.organizations << Organization.current if add_current_organization?
-    end
+    return unless self.new_record? && self.errors.empty?
```
ares: please don't use unless if there is `&&` or `||` in condition

test/unit/taxonomix_test.rb
```diff
@@ -61,21 +61,42 @@ def setup
...
+      assert_includes @dummy.class.which_organization, org2
+      refute_includes @dummy.class.which_organization, org3
+    end
+
+    test 'returns only objects in the given taxonomies' do
+      @dummy.name = 'foo'
+      @dummy.save
+      TaxonomixDummy.expects(:taxable_ids).returns([@dummy.id]).at_least_once
+      assert_equal 1, @dummy.class.with_taxonomy_scope(@loc, @org).count
+      assert_equal [@dummy], @dummy.class.with_taxonomy_scope(@loc, @org).to_a
```
ares: Maybe store the result of `@dummy.class.with_taxonomy_scope(@loc, @org)` so it's not called twice and it's more DRY.

test/unit/taxonomix_test.rb
```diff
@@ -61,21 +61,42 @@ def setup
...
     end
 
-    assert_empty dummy_class.which_location
-    assert_includes dummy_class.which_organization, org1
-    assert_includes dummy_class.which_organization, org2
-    refute_includes dummy_class.which_organization, org3
+    test 'does not return objects outside the specified taxonomies' do
+      @dummy.name = 'foo'
+      @dummy.save
+      TaxonomixDummy.expects(:taxable_ids).returns([@dummy.id + 20]).at_least_once
```
ares: What does `+ 20` mean here? if it's non-existing resource id, I think it would be better to verify it really does not exist to avoid intermittent tests failures. Also I think it would be good to have a test where taxable_ids `.returns([])`. I think the first test does it but it might change by future tests changes so I'd prefer having explicit one.

ares:
just few comments inline, looks good in general

dLobatog:
@ares @tbrisker Updated so it doesn't change set_current_taxonomy, it explicitly tests the `[]` case, and improves one test to ensure  `@dummy.id` is  not found in taxable_ids. Thanks for reviewing!

ares:
APJ

dLobatog:
Green! @ares

ares:
Thanks @dLobatog, merged as d08290be7b0bca2d4517e297b46daa41949f8791

PR #3827, opened 2016-09-08
Fixes #16491 - Upgrade rubocop to version 0.42

tbrisker:
Some new cops were introduced and HoundCI has been upgraded to 0.42 so that it comments on anything not explicitly disabled.

@theforeman/core - For now I just added all the new cops into .rubocop_todo.yml, but please take a look and see if there are cops we wish to fix and enable.


.rubocop_todo.yml
```diff
@@ -1,34 +1,45 @@
...
-# Offense count: 5
+# Offense count: 7
 Lint/HandleExceptions:
   Enabled: false
 
+# Offense count: 3
+# Cop supports --auto-correct.
+# Configuration parameters: EnforcedStyle, SupportedStyles.
+# SupportedStyles: runtime_error, standard_error
+Lint/InheritException:
```
iNecas: This cop makes sense, let's fix it.

.rubocop_todo.yml
```diff
@@ -1,34 +1,45 @@
...
 
-# Offense count: 10
+# Offense count: 12
 Lint/ParenthesesAsGroupedExpression:
   Enabled: false
 
-# Offense count: 56
+# Offense count: 1
+Lint/ShadowedException:
+  Enabled: false
```
iNecas: It it's just one, lets disable it in the code

.rubocop_todo.yml
```diff
@@ -38,53 +49,63 @@ Lint/UnusedBlockArgument:
...
-# Offense count: 5
+# Offense count: 6
 # Cop supports --auto-correct.
 Performance/Casecmp:
   Enabled: false
 
+# Offense count: 17
+# Cop supports --auto-correct.
+Performance/PushSplat:
+  Enabled: false
```
iNecas, line 91: A prove, that enforce documentation actually doesn't your documentation better. http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Performance/PushSplat is kind of useless to understand, why.

ares:
Thanks, from the new cops I'd like to see enabled these two 
- Lint/InheritException
- Style/MethodMissing

.rubocop_todo.yml
```diff
@@ -38,53 +49,63 @@ Lint/UnusedBlockArgument:
 Lint/UnusedMethodArgument:
   Enabled: false
 
-# Offense count: 508
+# Offense count: 3
+# Configuration parameters: ContextCreatingMethods.
+Lint/UselessAccessModifier:
+  Enabled: false
```
tbrisker: I think we should also enable this one

.rubocop_todo.yml
```diff
@@ -326,32 +347,57 @@ Style/LeadingCommentSpace:
...
 # Cop supports --auto-correct.
 # Configuration parameters: EnforcedStyle, SupportedStyles.
 # SupportedStyles: require_parentheses, require_no_parentheses, require_no_parentheses_except_multiline
 Style/MethodDefParentheses:
   Enabled: false
 
-# Offense count: 45
+# Offense count: 5
+Style/MethodMissing:
+  Enabled: false
```
ares, line 348: I like this one, it surprises me it's under Style

.rubocop_todo.yml
dLobatog: Please do, it's just 3 offences so it'd be easy to review here

.rubocop_todo.yml
```diff
@@ -148,6 +169,10 @@ Rails/HasAndBelongsToMany:
 Rails/Output:
   Enabled: false
 
+# Offense count: 96
+Rails/OutputSafety:
+  Enabled: false
```
dLobatog, line 158: We should enable this at some point

.rubocop_todo.yml
tbrisker: I understand why this rule makes sense - if you're combining two arrays, why splat one and push it to the other when you can just use the concat method - but I agree the documentation makes no sense

.rubocop_todo.yml
```diff
@@ -483,57 +550,72 @@ Style/SignalException:
...
 
-# Offense count: 511
+# Offense count: 493
 # Cop supports --auto-correct.
 Style/SpaceAfterComma:
   Enabled: false
 
+# Offense count: 1
+# Cop supports --auto-correct.
+Style/SpaceAfterNot:
```
dLobatog: Fix? It's 1 offence and it makes sense (prevents '! variable`)

.rubocop_todo.yml
tbrisker: checked those 3, looks like there is a bug in rubocop that mistakenly marks them as useless. I'll leave it.

daviddavis:
@tbrisker it might be nice to notify plugin owners of the version change to rubocop. Katello for one uses foreman's gem requirement. Speaking of which, I'll go ahead and open a PR to make sure katello will work with 0.42.

.rubocop_todo.yml
daviddavis: This cop is already enabled on develop. The checking for useless modifiers was simply improved in 0.42 so it caught three new useless modifiers.

I'm fairly certain the first two are useless:

https://git.io/viBsq
https://git.io/viBsZ

The third seems like it's just pointing to the wrong line. It's supposed to point to this `private` which is useless:

https://git.io/viBsE

If you remove it, it should pass. I've opened a rubocop issue for this as well: https://github.com/bbatsov/rubocop/issues/3483

.rubocop_todo.yml
tbrisker: I'm going to have to leave it disabled due to https://github.com/bbatsov/rubocop/issues/3358 but I fixed everything so this should be enabled on the next rubocop upgrade.

tbrisker:
Fixed some of the new cops and a couple of trivial old ones along the way. I'll send a mail to -dev list notifying of the upgrade.

dLobatog:
Merged as e7e5ef3ee20f8e3c6b8c996b7321f4d481037612, thanks @tbrisker!

domcleal:
Please note that this has broken some functional tests on develop as the tests were also failing in this PR. I filed http://projects.theforeman.org/issues/16512 for the issue.

It also looks like rubocop is failing on develop:

<pre>
config/routes.rb:1:1: C: Missing frozen string literal comment.
require 'api_constraints'
^
</pre>

(etc, many similar violations.)

app/services/power_manager/base.rb
```diff
@@ -5,10 +5,16 @@ def initialize(opts = {})
...
 
     def self.method_missing(method, *args)
+      super
+    rescue NoMethodError
       logger.warn "invalid power state request #{action} for host: #{host}"
       raise ::Foreman::Exception.new(N_("Invalid power state request: %{action}, supported actions are %{supported}"), { :action => action, :supported => SUPPORTED_ACTIONS })
     end
 
+    def self.respond_to_missing?(method_name, include_private = false)
+      super
```
ares, line 15: just out of curiosity, what is this good for?
edit: I think we should rather locally disable the cop in such cases

PR #3184, opened 2016-02-15
Fixes #13703 - added db info into foreman-debug

lzap:
This patch adds more information from our database into debug tarball. So far
we have been only shipping settings and auth_sources.

This is selection of our tables which I think might be useful in giving
support. All the data that can be sensitive are hashed via plain md5 function
(no salt). The idea is to make them unreadable. This is not top-security, but
debug tarballs are not usually disclosed publicly. It also allows to check
things in database when there are hostnames or logins mentioned in
ticket/bugzilla.

Here is an example output: https://gist.github.com/lzap/50a3760148413496e4ba

I would like also to export all templates in separate files once I find the correct PgSQL command: http://stackoverflow.com/questions/35409314/dump-individual-text-fields-into-separate-files-in-psql

PR #3828, opened 2016-09-08
#bz 1371745 - [RFE] Export list of users and their locale setting

tchellomello:
This patch added the capability for foreman-debug to grab the user's preferences for timezone and language locale

``` bash
COMMAND> echo 'select login, locale, timezone from users order by login' | su postgres -c 'psql foreman'

++ echo 'select login, locale, timezone from users order by login'
++ su postgres -c 'psql foreman'
could not change directory to "/root"
       login       | locale |          timezone          
-------------------+--------+----------------------------
 admin             | pt_BR  | Eastern Time (US & Canada)
 foreman_admin     |        | 
 foreman_api_admin |        | 
(3 rows)
```

### DIFF
--- a/script/foreman-debug
+++ b/script/foreman-debug
@@ -22,7 +22,7 @@ OPTIONS:
   -d DIR  Directory to place the tarball in (default /tmp/foreman-XYZ)
   -g      Skip generic info (CPU, memory, firewall etc.)
   -a      Do not generate a tarball from the resulting directory
-  -l      Include usernames to the locale report
+  -l      Include login names (usernames) on the locale settings
   -m NUM  Maximum lines to keep for each file (default 5000), zero means no limit
   -j PRG  Filter with provided program when creating a tarball
   -p      Additionally print password patterns being filtered out
--- a/script/foreman-debug
+++ b/script/foreman-debug
@@ -22,6 +22,7 @@ OPTIONS:
   -d DIR  Directory to place the tarball in (default /tmp/foreman-XYZ)
   -g      Skip generic info (CPU, memory, firewall etc.)
   -a      Do not generate a tarball from the resulting directory
+  -l      Include usernames to the locale report
   -m NUM  Maximum lines to keep for each file (default 5000), zero means no limit
   -j PRG  Filter with provided program when creating a tarball
   -p      Additionally print password patterns being filtered out
@@ -130,6 +131,7 @@ VERBOSE=0
 DEBUG=0
 UPLOAD=0
 UPLOAD_DISABLED=0
+EXPORT_USERNAMES=0
 
 if type -p xz >/dev/null; then
   COMPRESS="xz -9"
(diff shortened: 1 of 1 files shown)


tchellomello:
Example:

``` bash
USAGE: ./foreman-debug [options]

Collects configuration and log data for Foreman, Smart Proxies, backend
services and system information while removing security information like
passwords, tokens and keys.

This program can be used on Foreman instances, Smart Proxy instances or
backend services separately.

OPTIONS:
  -d DIR  Directory to place the tarball in (default /tmp/foreman-XYZ)
  -g      Skip generic info (CPU, memory, firewall etc.)
  -a      Do not generate a tarball from the resulting directory
  -l      Include usernames to the locale report
  -m NUM  Maximum lines to keep for each file (default 5000), zero means no limit
  -j PRG  Filter with provided program when creating a tarball
  -p      Additionally print password patterns being filtered out
  -q      Quiet mode
  -v      Verbose mode
  -u      Upload tarball
  -h      Shows this message

You may want to upload the tarball (with -u) to our public server via rsync.
This is a write-only directory (readable only by Foreman core developers)
Please note that the rsync transmission is UNENCRYPTED.

--without usernames


[root@sat6mmello ~]# ./foreman-debug 
Processing... (takes a while)Exporting tasks, this may take a few minutes.

[...snip...]

A debug file has been created: /tmp/foreman-debug-0ROJ6.tar.xz (1162664 bytes)

To upload a tarball to our secure site, please use the -u option.

[root@sat6mmello tmp]# cat /tmp/foreman-debug-0ROJ6/foreman_users_locale 
COMMAND> echo 'select locale, timezone from users order by login' | su postgres -c 'psql foreman'

could not change directory to "/root"
 locale |          timezone          
--------+----------------------------
 en     | Eastern Time (US & Canada)
        | 
        | 
(3 rows)

==== with usernames

[root@sat6mmello ~]# ./foreman-debug -l
Processing... (takes a while)

[..SNIP...]

A debug file has been created: /tmp/foreman-debug-eCL7m.tar.xz (1164288 bytes)

[root@sat6mmello tmp]# cat /tmp/foreman-debug-eCL7m/foreman_users_locale 
COMMAND> echo 'select login, locale, timezone from users order by login' | su postgres -c 'psql foreman'

could not change directory to "/root"
       login       | locale |          timezone          
-------------------+--------+----------------------------
 admin             | en     | Eastern Time (US & Canada)
 foreman_admin     |        | 
 foreman_api_admin |        | 
(3 rows)
```

dLobatog:
@tchellomello Thanks for the contribution! As @lzap mentioned in the ticket, I doubt this would be very useful as browser_locale is really what most people will have, but we'll see. You might want to query that via JS or something and store it in an column of 'users' that's not used for anything.

Please take a look at the comments by the Foreman bot, the commits must be squashed into one and have the proper format. I've cloned the BZ to a redmine ticket http://projects.theforeman.org/issues/16511 so you can use that.

lzap:
Hello @tchellomello,

please do not call psql directly from debug and use our exporter framework via rake:

https://github.com/theforeman/foreman/blob/develop/lib/tasks/exports.rake

This will create CSV export file for you. Note foreman-debug does not yet call exporter rake, but it will (I have plans to modify this patch to do this: https://github.com/theforeman/foreman/pull/3184

lzap:
Also this is very important, login names must not be exposed as it is security leak. You want to export only initial and last letter, something like `a***n`. Make sure the record ID is included as well.

PR #3829, opened 2016-09-09
Fixes #13748 - Add testing tools for code under webpack

dLobatog:
This commit adds jest as a tool to be used to test some of the code
under webpack, supporting ES2015 via babel-polyfill.

The purpose of this is just to give the tools for other developers
to test their javascript code.


dLobatog:
@gailsteiger @tbrisker mind to take a look?

lib/tasks/jenkins.rake
```diff
@@ -23,6 +23,11 @@ begin
         --no-color --out rubocop.xml")
       exit($?.exitstatus)
     end
+
+    task :jest do
```
dLobatog: Probably useless, I suppose jenkins can run `npm run jest`

package.json
```diff
@@ -35,6 +38,7 @@
     "select2": "~3.5.2-browserify"
   },
   "scripts": {
-    "lint": "eslint -c ./.eslintrc webpack/assets/**/*.js  || exit 0"
+    "lint": "./node_modules/.bin/eslint -c .eslintrc webpack/assets/**/*.js",
```
dLobatog: $? is 0 after running eslint w/o errors . I also added the node_modules portion because otherwise it might run a globally installed one

.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  "presets": ["es2015"]
+}
```
gailsteiger, line 3: since you added the preset via .babelrc, the following code is no longer necessary in webpack.config:

```
    query: {
      presets: ['es2015']
    }
```

package.json
```diff
@@ -35,6 +38,7 @@
...
+    "jest": "jest"
```
gailsteiger: would suggest naming the script "test" instead of jest. Then you can run "npm test

jest can also just be run directly from the command line

webpack/assets/javascripts/__tests__/foreman_tools-test.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
```
gailsteiger: I prefer the 'filename.test.js' test file naming convention.
Jest will look for files following this convention outside of the **tests** directory. This would enable us to make changes to test file locations if we decide that we prefer something other than the **tests** convention.

webpack/assets/javascripts/__tests__/foreman_tools-test.js
```diff
@@ -0,0 +1,37 @@
...
+
+  it('adds a bold text next to the label', () => {
+    expect(tools.iconText('plus', 'foo', 'patternfly'))
+    .toBe('<span class="patternfly patternfly-plus"/><strong>foo</strong>');
+  });
+});
+
+describe('activateDatatables', () => {
+  it('calls this.DataTable when it finds a data-table=server', () => {
+    const $ = require('jquery');
```
gailsteiger: it calls '$.fn.DataTable', so the description in the describe function is confusing.

gailsteiger:
I think that automatic mocking by default is one of jest's strengths and that was an important factor in my recommending it. Recently released version 15.0 sets automock false by default.

I suggest setting automock true. This will force explicitly unmocking all dependencies in the test which will give the dependencies more visibility.

In any case, automocking can be enabled/disabled at the testfile level. Setting the automock property to true (or false for that matter) gives an indication of how we would like tests written.

webpack/assets/javascripts/foreman_tools.js
```diff
@@ -1,3 +1,7 @@
+'use strict';
+
+const $ = require('jquery');
```
tbrisker: This shouldn't be needed as the webpack ProvidePlugin should make $ available to all our modules

gailsteiger: From what I can tell, $ is not available to the module when the tests are run

.babelrc
dLobatog: fixed

dLobatog:
@gailsteiger I think I've fixed the comments you've mentioned above. I've set automocking as true, however I'm having trouble with making the tests pass again. It seems like jQuery selectors are not working in the activateDatatables test, even if when jQuery is unmocked. Any idea? 

If we make our tests small and make them require only the files they need, what is automocking useful for? I think that can be our context, since we're starting to write tests for JS just now.

Thanks for reviewing!

gailsteiger:
@dLobatog try defining the jest section of package.json as follows:

``` json
  "jest": {
    "automock": true,
    "unmockedModulePathPatterns": [
      "jquery"
    ]
  }
```

Thanks for implementing. No testing has been a pain point for me.

dLobatog:
@gailsteiger  That worked :smile:  - tests do pass now. I removed the manual unmocking, any idea why unomcking it manually with `jest.unmock('jquery');` did not work?

gailsteiger:
@dLobatog Did you unmock before you required 'jquery'?

dLobatog:
@gailsteiger Yeah, I did it exactly as with foreman_tools. My _guess_ is that the manual require in 'foreman_tools.js' was mocked, and I did not find a way to unmock that manually other than without the options you said in https://github.com/theforeman/foreman/pull/3829#issuecomment-246278511

dLobatog:
@domcleal BTW I think we can safely test this in Travis as we wouldn't need to test this with other plugins or anything.

webpack/assets/javascripts/foreman_tools.js
gailsteiger: ES6 modules are always in strict mode so no need to declare it.

@dLobatog You can also use the import statement instead of require

``` js
import $ from 'jquery';
```

gailsteiger: @tbrisker I think we should consider declaring dependencies explicitly instead of relying on the webpack ProvidePlugin for our modules. Relying on the plugin creates a strange dependency on the build system. We wouldn't be able to export and package the code for reuse without explicitly declaring the dependencies.

I think leaving the plugin in place to as a safety net for legacy code that might be moved to webpack assets without being refactored makes sense.

webpack/assets/javascripts/__tests__/foreman_tools.test.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
```
gailsteiger: we don't use 'use strict' in es6

gailsteiger:
Looks good to me. I have 2 minor comments outstanding but there is no reason not to merge and get js testing started.

dLobatog:
@gailsteiger @tbrisker Updated to use `import` in foreman tools & remove use strict from the test

ohadlevy:
how should we run the tests? jenkins / travis?

dLobatog:
@ohadlevy I don't see any reason for using Jenkins here - tests don't affect plugins so Travis ought to work fine (+ it's less load on Jenkins, you have to npm install to run tests...)

ohadlevy:
sounds good, what do you need to enable ?

webpack/assets/javascripts/foreman_tools.js
```diff
@@ -1,3 +1,7 @@
+'use strict';
```
tbrisker: I think this is also not needed.

dLobatog:
@tbrisker Removed use strict on foreman_tools too since its ES6 
@ohadlevy I suppose we should enable Travis after it's merged?

package.json
tbrisker: The `|| exit 0` is there because if eslint finds errors it exits with a non-zero status, causing npm to think there was a bug in eslint and leading to confusing error message being displayed.

tbrisker:
Other then one minor comment this is good to merge.

dLobatog:
@tbrisker Updated!

tbrisker:
Merged as a467444a7c1bf2123a5d259869331c6a9a74bec5, thanks @dLobatog!

PR #3830, opened 2016-09-11
Fixes #16074 - Adds BIND variable to startup process

shlomizadok:
Instead of binding to 0.0.0.0 we default to localhost
and use ENV variable BIND to override.
~~This also introduces '.env' files which helps to set
other ENV variables~~

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,5 +1,5 @@
 # Run Rails & Webpack concurrently
 # If you wish to use a different server then the default, use e.g. `export RAILS_STARTUP='puma -w 3 -p 3000 --preload'`
-rails: [ -n "$RAILS_STARTUP" ] && $RAILS_STARTUP || bin/rails server -b 0.0.0.0
+rails: [ -n "$RAILS_STARTUP" ] && $RAILS_STARTUP || [ -n "$BIND" ] && bin/rails server -b $BIND || bin/rails server
 # you can use WEBPACK_OPTS to customize webpack server, e.g. 'WEBPACK_OPTS='--https --key /path/to/key --cert /path/to/cert.pem --cacert /path/to/cacert.pem' foreman start '
 webpack: ./node_modules/.bin/webpack-dev-server --config config/webpack.config.js $WEBPACK_OPTS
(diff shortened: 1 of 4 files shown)


shlomizadok:
/me wonders if the `.env` should be added - or explain how to use in docs

shlomizadok:
One note: At the beginning I have added a `.env`, but on second thought it should not be added (though can be introduced in documentation)

dLobatog:
@shlomizadok Does this work with [forklift](https://github.com/Katello/forklift)?

shlomizadok:
Would work on forklift if we'd add a BIND variable. I'd suggest creating a `.env` via ansible (can create a pull request once this is merged)

Procfile
```diff
@@ -1,5 +1,5 @@
 # Run Rails & Webpack concurrently
 # If you wish to use a different server then the default, use e.g. `export RAILS_STARTUP='puma -w 3 -p 3000 --preload'`
-rails: [ -n "$RAILS_STARTUP" ] && $RAILS_STARTUP || bin/rails server -b 0.0.0.0
+rails: [ -n "$RAILS_STARTUP" ] && $RAILS_STARTUP || [ -n "$BIND" ] && $BIND bin/rails server -b $BIND || bin/rails server -b 127.0.0.1
```
domcleal: The `$BIND` before bin/rails appears superfluous? It causes a warning:

<pre>
13:11 $ BIND=0.0.0.0 foreman start                                                                                                                                                                                                                                              
13:11:08 rails.1   | started with pid 24145
13:11:08 webpack.1 | started with pid 24146
13:11:08 rails.1   | sh: 0.0.0.0: command not found
</pre>

domcleal:
Please also add .env or .env\* to the gitignore file.

Procfile
shlomizadok: It was left there by accident. removed

shlomizadok:
Added .env\* to .gitignore and removed superfluous $BIND

domcleal:
Merged as 3b10c2a, thanks @shlomizadok.

Please update the notes at the bottom of https://theforeman.org/handbook.html#Technicalnotes on `foreman start` usage, and consider dropping an e-mail to foreman-dev in case people want to change the secure default.

PR #3831, opened 2016-09-12
fixes #16514 - redirect to login page if user is deleted

kgaikwad:
Previously, when user is deleted and they still have an active session in browser, Dashboard Not Found error is shown.

Added change which will redirect the user to login page with warning message "your session has expired".


daviddavis:
@kgaikwad please rebase your branch with develop. You're picking up the default houndci config because your branch is missing .hound.yml.

kgaikwad:
@daviddavis done with the rebase. Please check.

app/controllers/concerns/foreman/controller/authentication.rb
```diff
@@ -7,7 +7,7 @@ def available_sso
 
   def authenticate
     return true if (User.current && Rails.env.test? && api_request?) ||
-                   session[:user] && (User.current = User.unscoped.find(session[:user]))
+                   session[:user] && (User.current = User.unscoped.find_by(id: session[:user]))
```
domcleal, line 10: Why change this?

domcleal:
Please add functional tests to demonstrate the change.

app/controllers/concerns/foreman/controller/authentication.rb
kgaikwad: When user is deleted while they still have an active session in browser and they access any URL again then message 

   {Resource} Not Found
   Please try to update your request
   [ Back ]

Clicking back or reload does not help. Even even visiting /users/logout does not help.
What helps is visiting /users/login and logging in as different user.

Here, "find" method throws an exception  "ActiveRecord::RecordNotFound" which is handled in application controller and on web UI, it is showing warning "{Resource} Not Found" instead of redirection to login page with some warning message. For example - "Your session has expired".

Instead of exception, "find_by" method will return nil and assign to User.current then rest of the things already handled by authenticate method.

PR #3832, opened 2016-09-12
fixes #16512 - remove Setting#respond_to_missing?

domcleal:
respond_to_missing? changed how AR handles group_by calls on
collections, causing group_by to be interpreted as a scope on the model.
Rather than adding a database call to determine known setting names,
it's been removed again.

The method_missing implementation has been deprecated in favour of the
hash [] style method calls which are more idiomatic and avoids making
the respond_to? behaviour any more complex.

### DIFF
--- a/app/controllers/api/v2/base_controller.rb
+++ b/app/controllers/api/v2/base_controller.rb
@@ -78,7 +78,7 @@ module Api
       end
 
       def metadata_per_page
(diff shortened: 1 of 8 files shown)


ares:
Thanks @domcleal, fixes the issue for me and apparently also jenkins. Merged as 2e5f252ae12e20074f9f73815416637a8a6732ce

PR #3836, opened 2016-09-13
Fixes #14545 - Correctly parse y.z minor OS version from facts

tbrisker:
Some OSes use y.z minor version, for example CentOS uses versions such
as 7.2.1511. Currently, the puppet fact parser will only take the 'y'
part of the version, 2 in this case, which can lead to conflicts if the
OS has been defined manually with the 2.1511 minor version (as the
description will still contain the entire version).

### DIFF
--- a/app/services/fact_parser.rb
+++ b/app/services/fact_parser.rb
@@ -179,10 +179,4 @@ class FactParser
   def not_implemented_error(method)
     "#{method} fact parsing not implemented in #{self.class}"
   end
-
-  def is_numeric?(string)
-    !!Integer(string)
-  rescue ArgumentError, TypeError
-    false
-  end
 end
--- a/app/services/puppet_fact_parser.rb
+++ b/app/services/puppet_fact_parser.rb
@@ -7,7 +7,7 @@ class PuppetFactParser < FactParser
     if os_name == "Archlinux"
       # Archlinux is rolling release, so it has no release. We use 1.0 always
       args = {:name => os_name, :major => "1", :minor => "0"}
-      os = Operatingsystem.where(args).first || Operatingsystem.new(args)
+      os = Operatingsystem.find_or_initialize_by(args)
     elsif orel.present?
       if os_name == "Debian" && orel[/testing|unstable/i]
         case facts[:lsbdistcodename]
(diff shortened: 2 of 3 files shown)


app/services/puppet_fact_parser.rb
```diff
@@ -27,14 +27,14 @@ def operatingsystem
       elsif os_name[/FreeBSD/i]
         orel.gsub!(/\-RELEASE\-p[0-9]+/, '')
       end
-      major, minor = orel.split(".")
-      major.to_s.gsub!(/\D/, '') unless is_numeric? major
-      minor.to_s.gsub!(/\D/, '') unless is_numeric? minor
-      args = {:name => os_name, :major => major.to_s, :minor => minor.to_s}
-      os = Operatingsystem.where(args).first || Operatingsystem.create!(args)
```
tbrisker, line 34: No reason to save the model twice for new OSes, if there is a failing validation it will be thrown in the `.save!` below.

timogoebel:
LGTM.

@tbrisker : Have you come up with a solution how not to break media paths for, let's say, Centos? Could we just add a new variable that gets replaces in the url that does not contain the full minor version?

tbrisker:
@timogoebel It looks like actually this will un-break the media path for CentOS, take a look at: http://mirror.centos.org/centos/

timogoebel:
@tbrisker : You're correct. 👍

tbrisker:
Thanks to @dlevene1 for providing the reproducing details, please wait for him to verify this solves the issue for him before merging.

tbrisker:
Reporter has confirmed via mail that this PR fixed the issue for them.

test/unit/puppet_fact_parser_test.rb
```diff
@@ -99,6 +99,17 @@ def setup
...
   end
 
+  test 'should accept y.z minor version' do
+    FactoryGirl.create(:operatingsystem, name: "CentOS",
+                                         major: "7",
+                                         minor: "2.1511",
+                                         description: "CentOS Linux 7.2.1511")
+    assert_valid PuppetFactParser.new({ "operatingsystem" => "CentOS",
+                                        "lsbdistdescription" => "CentOS Linux release 7.2.1511 (Core) ",
+                                        "operatingsystemrelease" => "7.2.1511"
```
dLobatog, line 109: Not that I care, but here's an example where enabling the cop to enforce one type of hash syntax would've been good instead of allowing 2 different hashing syntaxes to coexist.

app/services/fact_parser.rb
```diff
@@ -179,10 +179,4 @@ def normalize_interfaces(interfaces)
   def not_implemented_error(method)
     "#{method} fact parsing not implemented in #{self.class}"
   end
-
-  def is_numeric?(string)
```
dLobatog, line 183: Didn't see any instance of this in the salt/chef/ansible parsers so I guess it's OK to remove w/o deprecation

dLobatog:
Merged as af6d54a3a184846f77f5fd7ca07bcdc7de70a264, thanks @tbrisker!

app/services/fact_parser.rb
tbrisker: I also scanned for it across all our repos (including katello) and found none which is why I just removed it. Just as a side note, the logic here doesn't actually match the function name - non-integer numeric values would return false. In addition, there are much better ways to check this w/o having to throw and catch exceptions.

test/unit/puppet_fact_parser_test.rb
tbrisker: For me it kind of makes sense to use the newer syntax when passing a hash as function arguments. The second hash I think expects string keys rather then symbols, so I kept it with the old style.

PR #3837, opened 2016-09-13
Fixes #15543 - External method to search fact's hostgroup

shlomizadok:


app/models/fact_value.rb
```diff
@@ -45,6 +45,15 @@ def self.search_by_host(key, operator, value)
     { :conditions => "fact_values.id IN(#{search.join(',')})" }
   end
 
+  def self.search_by_hostgroup(key, operator, value)
+    search_term = value =~ /\A\d+\Z/ ? 'id' : 'name'
+    conditions = sanitize_sql_for_conditions(["hostgroups.#{search_term} #{operator} ?", value_to_sql(operator, value)])
+    search     = FactValue.joins(:hostgroup).where(conditions).uniq.pluck('fact_values.id')
+
+    return { :conditions => "1=0" } if search.empty?
+    { :conditions => "fact_values.id IN(#{search.join(',')})" }
```
ohadlevy: why not use an inner query? if search is a scope, it will be done automatically.

tbrisker: Even better, I think this should work just as well and get rid of the extra query completely:

``` ruby
search_term = value =~ /\A\d+\Z/ ? 'id' : 'name'
conditions = sanitize_sql_for_conditions(["hostgroups.#{search_term} #{operator} ?", value_to_sql(operator, value)])
{ :joins => :hostgroup, :conditions => conditions }
```

shlomizadok: @tbrisker - this works perfect. Thanks !

tbrisker:
@shlomizadok Could you apply the same refactoring to the host search above while at it?

shlomizadok:
@tbrisker - yeah, tried to DRY them as well.

app/models/fact_value.rb
```diff
@@ -37,12 +37,17 @@ class FactValue < ActiveRecord::Base
...
+    search_by('host', key, operator, value)
+  end
 
-    return { :conditions => "1=0" } if search.empty?
-    { :conditions => "fact_values.id IN(#{search.join(',')})" }
+  def self.search_by_hostgroup(key, operator, value)
+    search_by('hostgroup', key, operator, value)
+  end
+
+  def self.search_by(host_or_hg, key, operator, value)
```
tbrisker: you can even DRY this more- key is either `host` or `host.hostgroup`, so you can use the same method for both and just check the key to decide which table to use.

tbrisker (changes requested):

shlomizadok:
@tbrisker - got it as DRY as I can...

app/models/fact_value.rb
```diff
@@ -35,14 +35,12 @@ class FactValue < ActiveRecord::Base
   scope :root_only, -> { with_roots.where(:fact_names => {:ancestry => nil}) }
 
   validates :fact_name_id, :uniqueness => { :scope => :host_id }
-
-  def self.search_by_host(key, operator, value)
+
```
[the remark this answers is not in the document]
shlomizadok: @houndci-bot woof!

tbrisker (approved):

app/models/fact_value.rb
shlomizadok: Excellent idea. Changed :+1:

shlomizadok:
Foreman [test] failure seem unrelated

shlomizadok:
@theforeman/core - Any clue why this is failing?

tbrisker:
@shlomizadok this is the cause of the failures: http://projects.theforeman.org/issues/16599

dLobatog:
[test foreman]

shlomizadok:
[test foreman]

tbrisker:
Merged as 5ce4c4c51917fdcf377fb168157a099108e5557f, thanks @shlomizadok!

PR #3838, opened 2016-09-13
fixes #16527 - replace where_values calls on relations

domcleal:


tbrisker:
Merged as 1e7ae52191070be4fd9d88eae68538c109b5a36f, thanks @domcleal!

PR #3839, opened 2016-09-13
fixes #16529 - require rake/testtask explicitly

domcleal:


dLobatog:
Merged as b79ef2a8c600d3be3a2354270c6d60d8ce43edb0, thanks @domcleal!

PR #3840, opened 2016-09-14
Fixes #16532 - fixed hostgroup architecture params

lzap:
We changed the `assign_parameter` implementation to accept the parameters as
regular hash with root element, but the JavaScript part was hardcoded to always
send `host`.

This patch changes the behavior to send `host` or `hostgroup` accordingly the
same way as in `medium_selected` which fixes the issue in hostgroup form (it
was not possible to set architecture).

Regression in 1.13 RC1 introduced by UEFI PXE loader implementation.

### DIFF
--- a/app/assets/javascripts/host_edit.js
+++ b/app/assets/javascripts/host_edit.js
@@ -346,11 +346,13 @@ function subnet_contains(network, cidr, ip) {
 }
 
 function architecture_selected(element){
-  var attrs   = attribute_hash(['architecture_id', 'organization_id', 'location_id']);
   var url = $(element).attr('data-url');
+  var type = $(element).attr('data-type');
+  var attrs = {};
+  attrs[type] = attribute_hash(['architecture_id', 'organization_id', 'location_id']);
   tfm.tools.showSpinner();
   $.ajax({
-    data: {host: attrs},
+    data: attrs,
     type:'post',
     url: url,
     complete: function(){
@@ -363,11 +365,13 @@ function architecture_selected(element){
 }
 
 function os_selected(element){
-  var attrs = attribute_hash(['operatingsystem_id', 'organization_id', 'location_id']);
   var url = $(element).attr('data-url');
+  var type = $(element).attr('data-type');
+  var attrs = {};
+  attrs[type] = attribute_hash(['operatingsystem_id', 'organization_id', 'location_id']);
   tfm.tools.showSpinner();
   $.ajax({
-    data: {host: attrs},
+    data: attrs,
     type:'post',
     url: url,
     complete: function(){
@@ -416,10 +420,9 @@ function update_provisioning_image(){
 function medium_selected(element){
   var url = $(element).attr('data-url');
   var type = $(element).attr('data-type');
-  var obj = (type == "hosts" ? "host" : "hostgroup");
   var attrs = {};
-  attrs[obj] = attribute_hash(['medium_id', 'operatingsystem_id', 'architecture_id']);
-  attrs[obj]["use_image"] = $('*[id*=use_image]').attr('checked') == "checked";
+  attrs[type] = attribute_hash(['medium_id', 'operatingsystem_id', 'architecture_id']);
+  attrs[type]["use_image"] = $('*[id*=use_image]').attr('checked') == "checked";
   $.ajax({
     data: attrs,
     type:'post',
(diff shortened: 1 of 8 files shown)


app/models/host/base.rb
```diff
@@ -59,8 +59,6 @@ def self.taxonomy_conditions
                                     :domain, :domain_id, :domain_name,
                                     :lookup_values_attributes].freeze
 
-    after_initialize :suggest_default_pxe_loader
-
```
lzap, line 63: I noticed this is not needed as it's defined in `PxeLoaderSuggestion` concern. Removing to match other three instances of this (managed_host and hostgroup files).

thomasmckay:
Do similar changes need to be made in https://github.com/Katello/katello/blob/master/app/assets/javascripts/katello/hosts/host_and_hostgroup_edit.js ?

app/assets/javascripts/host_edit.js
```diff
@@ -363,11 +366,14 @@ function architecture_selected(element){
 }
 
 function os_selected(element){
-  var attrs = attribute_hash(['operatingsystem_id', 'organization_id', 'location_id']);
   var url = $(element).attr('data-url');
+  var type = $(element).attr('data-type');
+  var obj = (type == "hosts" ? "host" : "hostgroup");
```
domcleal: The type string can be singularised in the view, no need to hardcode this here?

domcleal:
Thanks, please add test coverage for the change.

lzap:
Threre are couple of failures on Ruby 2.0, not sure what this is: `ActionView::Template::Error: Could not load compiled manifest from ...`

I will do an integration test for this case.

Not sure about Katello :-)

app/assets/javascripts/host_edit.js
```diff
@@ -363,11 +366,14 @@ function architecture_selected(element){
...
+  var attrs = {};
+  attrs[obj] = attribute_hash(['operatingsystem_id', 'organization_id', 'location_id']);
```
parthaa: May be out of scope but would it be possible for you  to have a global var or some mechanism by which you can make attribute_hash accept additional set attributes to send when os gets selected.  

Basically Katello needs these also to get sent when os gets selected

```
  var attrs = attribute_hash(['operatingsystem_id', 'organization_id', 'location_id', 'content_view_id',
                              'lifecycle_environment_id', 'content_source_id', 'architecture_id', 'hostgroup_id',
                              'medium_id', 'kickstart_repository_id']);
```

In fact we have to hack and over ride the complete method just for that additional set of attributes. Look at -> https://github.com/Katello/katello/blob/master/app/assets/javascripts/katello/hosts/host_and_hostgroup_edit.js#L252-L274

If there was ether a "os_attributes" or some global param we could very easily add the 3 params needed and not duplicate this code it would be great . As a bonus it will also fix -> http://projects.theforeman.org/issues/14699

domcleal: If you want to refactor this for extension points, please file a PR for it separately.

lzap: I tend to agree with Dominic in this case, the patch only changes the behavior and we will likely be backporting this into 1.13 RC2. On the other hand, if Katello is broken right now (RC1) and this patch changes it once again for RC2, it might be good idea to create such an extension.

Since this is important fix for 1.13, another PR could be good idea.

parthaa: yeah. lets do it in a diff PR

parthaa: @lzap how about adding something like a 

```
function get_attributes_for_os_selected() {
  return ['operatingsystem_id', 'organization_id', 'location_id'];
}
```

And using that . Is that too bad ?

lzap: And monkey patch this in each plugin? I don't like that. Global array seems a better idea than this :-)

app/assets/javascripts/host_edit.js
lzap: Singularized all of them, will update shortly.

lzap:
Added test as well. Looks stable on my laptop but that does not say anything...

https://gist.github.com/66c9346f8310a449eb6131d33d05ea14

test/integration/hostgroup_js_test.rb
```diff
@@ -5,6 +5,31 @@ class HostgroupJSTest < IntegrationTestWithJavascript
...
+    wait_for_ajax
+    select2 os.architectures.first.name, :from => 'hostgroup_architecture_id'
+    wait_for_ajax
+    select2 os.title, :from => 'hostgroup_operatingsystem_id'
+    wait_for_ajax
+    select2 os.media.first.name, :from => 'hostgroup_medium_id'
+    wait_for_ajax
+    select2 os.ptables.first.name, :from => 'hostgroup_ptable_id'
+    fill_in 'hostgroup_root_pass', :with => '12345678'
+    click_button('Submit')
```
domcleal: Use `assert_submit_button` or add an `assert_current_path` call (for the redirect path) after submitting, I don't think there's anything here that will wait until the form submission is complete - it isn't AJAX, so wait_for_ajax won't do it.

domcleal (changes requested):

test/integration/hostgroup_js_test.rb
lzap: When I remove the `wait_for_ajax` after submit line, the next assert fails (`host = Hostgroup.where(:name => "myhostgroup1").first`). I also don't see any JavaScript there, could be some kind of race condition (the hidden browser in a different process).

lzap: Anyway I refactored the test to use `assert_submit_button`

domcleal (approved):

domcleal:
Review complete, subject to integration tests passing.

domcleal:
[test foreman]

domcleal:
Merged as 1e180a5, thanks @lzap.

PR #3842, opened 2016-09-14
fixes #16545 - correctly search for classes from YAML import

domcleal:
Prevents a full class hash of name => parameters being passed to the AR
finder when using a parameterized (hash-style) YAML format.

### DIFF
--- a/test/unit/host_test.rb
+++ b/test/unit/host_test.rb
@@ -784,6 +784,15 @@ class HostTest < ActiveSupport::TestCase
       assert_equal '3.3.4.12', parameters['foreman_interfaces'].first['ip']
     end
 
+    test "should import from non-parameterized external nodes output" do
+      host = FactoryGirl.create(:host, :environment => environments(:production))
+      host.importNode("environment" => "production", "classes" => ["apache", "base"], "parameters" => {})
+
+      Setting[:Parametrized_Classes_in_ENC] = true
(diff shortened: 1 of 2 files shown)


test/unit/host_test.rb
```diff
@@ -784,6 +784,15 @@ def teardown
       assert_equal '3.3.4.12', parameters['foreman_interfaces'].first['ip']
     end
 
+    test "should import from non-parameterized external nodes output" do
+      host = FactoryGirl.create(:host, :environment => environments(:production))
+      host.importNode("environment" => "production", "classes" => ["apache", "base"], "parameters" => {})
+
+      Setting[:Parametrized_Classes_in_ENC] = true
+      Setting[:Enable_Smart_Variables_in_ENC] = true
+      assert_equal ['apache', 'base'], host.info['classes'].keys
```
dLobatog, line 793: This seems to pass even without the change?

domcleal: Yes, it probably will because the old method simply called `.each` on both arrays and hashes, but by luck it worked with hashes. The change uses `.is_a?` to correct behaviour for hashes, so this test ensures the new code still works for both (the method was only tested with hashes before).

If I'd just changed the code to `nodeinfo["classes"].keys.each` then it would have broken for arrays.

dLobatog:
Merged as 37c0b37ff2654b769f95812f27471baedc0ecaae, thanks @domcleal!

PR #3843, opened 2016-09-14
fixes #16557 - move tests into modern Rails layout

domcleal:
Models, controllers, mailers and helpers are now in test/ subdirs
matching app/ subdirs. Service, miscellaneous and old test/lib/ tests
are now in test/unit/ separated from models, which obsoletes the custom
test:lib rake task.

Removing the rake task also removed an entire second Rails environment
load from `rake test` which was writing to the Rails cache to initialise
settings. The reset_setting_cache test helper is now more robust against
the cache being missing on clean checkouts.

JSON fixtures are moved to the same test/static_fixtures/ directory so
they can be accessed uniformly from all tests.


lzap:
Holy moly this is a bummer, it's _not right time now_ having 1.13 RC1 out. Assuming we are not in rush for Rails 5, I suggest to wait until 1.13 comes out, perhaps wait more time until 1.13 stabilizes (few minor releases) to ease downstream process.

Until then, please implement a workaround so tests run under Rails 4 and Rails 5 without moving all of them.

domcleal:
I don't see why the timing of stable releases has anything to do with this, please reconsider this for review. 1.13 was branched to 1.13-stable, this only affects develop.

lzap:
> I don't see why the timing of stable releases has anything to do with
> this, please reconsider this for review. 1.13 was branched to
> 1.13-stable, this only affects develop.

I expect we will be backporting more than usual for the 1.13 release due
to downstream alignment and this makes it harder as we do the work in
develop first and then cherry pick (twice).

Can you make Rails 5 tests running first without the big rename? Because
I really think this is worth the effort. I am not saying no, I would
like us to wait a bit.

And what are the benefits of this move other than aligning with Rails 5
recommended paths? Keeping it as-is is also an option for an existing
project.

## 

Later,
 Lukas #lzap Zapletal

domcleal:
I doubt it makes it difficult to cherry pick commits, it's a trivial matter. In any case, maintenance of a stable release is not a matter that should prevent refactoring in develop.

> Can you make Rails 5 tests running first without the big rename?

Yes, but this is the correct layout on Rails 4 or higher.

> And what are the benefits of this move other than aligning with Rails 5 recommended paths?

It removes the custom lib test task, reducing the test initialisation time.

dLobatog:
@lzap :+1:  to getting this merged, it's nearly innocuous and it would save us a few cycles.. plus the new convenience of the get fixture method. 

I don't see how merging this would make anything harder to cherrypick, (OK, maybe you have to correct the path of the test)

domcleal:
Patch rebased, reset flags to match previous state.

dLobatog:
@domcleal Could you rebase this? I'd be happy to merge it, cons are pretty minor, but I'll give it until the end of the day in case anyone has more serious concerns with this

domcleal:
@dLobatog rebased, thanks.

domcleal:
[test foreman] probably an intermittent failure in AboutIntegrationTest.

dLobatog:
Thanks @domcleal !

PR #3844, opened 2016-09-14
Fixes #16562 - Enable JS testing on CI

dLobatog:
This enables JS testing on every PR. 
I've enabled it on my local repo to test it: https://travis-ci.org/dLobatog/foreman/builds/159996940
https://github.com/dLobatog/foreman/pull/28

cc @gailsteiger @ohadlevy @tbrisker

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,5 @@
+language: node_js
+node_js:
+  - "node"
+  - "4" # Fedora
+script: ./script/travis_run_js_tests.sh
--- a/script/travis_run_js_tests.sh
+++ b/script/travis_run_js_tests.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+set -ev
+if [[ $( git diff --name-only HEAD~1..HEAD webpack/ | wc -l ) -ne 0 ]]; then
+  npm run test;
+fi


timogoebel:
Do we really want to test this on travis? I am under the impression, that they are incredibly slow to queue tests.

ohadlevy:
Also.. do we want to run the tests conditionally? I would assume that are
way more non JS commits (for now), so the tests can be skipped? (see
https://docs.travis-ci.com/user/customizing-the-build/#Implementing-Complex-Build-Steps
)

dLobatog:
@timogoebel Any other suggestions? I don't want to add more load and 'npm installs' to our Jenkins when there's no reason to. 

@ohadlevy I guess I could do something like `if $( git diff --name-only HEAD~1..HEAD webpack/ ) then; npm run test;` I'll try it in my repo first

dLobatog:
Updated so that it only runs when the last commit contains changes in 'webpack/'

timogoebel:
> Any other suggestions? I don't want to add more load and 'npm installs' to our Jenkins when there's no reason to.

@dLobatog : I think, there is a good reason to use the jenkins farm. Travis being slow. But to be honest, I don't have strong feelings for this matter. We can try travis and if it doesn't work, try something else.

dLobatog:
@timogoebel We can always change from one to the other. I personally prefer Jenkins too as it's more configurable, but our infra is quite loaded already. I doubt it'll be much faster, it took 2m 50s to run it on 3 different versions of node (https://travis-ci.org/dLobatog/foreman/jobs/159996945) - I'll remove 0.10 as it isn't really relevant.

ohadlevy:
merged as 0d8ad89 thanks @dLobatog !

PR #3846, opened 2016-09-15
Fixes #16564 - Disable FrozenStringLiteralComment cop

tbrisker:
Also fixes a failng cop in config/initializers/secure_headers.rb


lzap (approved):
I think the main `.ruboccop.yml` is the right place.

Let's not play with the pragmas until our dependencies gets fixed, which will take quite some time I guess. This is useless until then.

PR #3847, opened 2016-09-15
Fixes #16565 - Correct host search by parameters

tbrisker:
Overrides for parameters values caused incorrect results for when
searching by parameter value. This commit fixes those results, and
improves the handling of parameter priorities in general. A new rake
task, `parameters:reset_priorities` is introduced that should be invoked
every time priorities are changed to update to the correct values.

### DIFF
--- a/db/migrate/20110327123639_add_priority_to_parameter.rb
+++ b/db/migrate/20110327123639_add_priority_to_parameter.rb
@@ -2,7 +2,7 @@ class AddPriorityToParameter < ActiveRecord::Migration
   def up
     add_column :parameters, :priority, :integer
     Parameter.reset_column_information
-    Parameter.reassign_priorities
+    Rake::Task['parameters:reset_priorities'].invoke
   end
 
   def down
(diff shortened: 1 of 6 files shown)


db/migrate/20110327123639_add_priority_to_parameter.rb
```diff
@@ -1,8 +1,7 @@
 class AddPriorityToParameter < ActiveRecord::Migration
   def up
     add_column :parameters, :priority, :integer
-    Parameter.reset_column_information
```
domcleal, line 4: Might this still be required? The column information may be cached, and the rake task still relies on it. (As may later migrations.)

domcleal (changes requested):

db/migrate/20110327123639_add_priority_to_parameter.rb
tbrisker: I'm not sure but I will leave that line there just in case.

domcleal (approved):

domcleal:
Merged as a29f3e3, thanks @tbrisker.

PR #3848, opened 2016-09-15
Fixes #16530 - For active session - Redirect to :back or /hosts

swapab:


app/controllers/users_controller.rb
```diff
@@ -155,4 +156,12 @@ def login_user(user)
   def parameter_filter_context
     Foreman::Controller::Parameters::User::Context.new(:ui, controller_name, params[:action], editing_self?)
   end
+
+  def verify_active_session
+    if params[:status].blank? && session[:user].present? && User.exists?(session[:user])
+      warning _("You have already Logged-In.")
```
domcleal: "logged in"

app/controllers/users_controller.rb
```diff
@@ -155,4 +156,12 @@ def login_user(user)
...
+      redirect_to root_path
```
domcleal: This should be the previous URL preferably, else `hosts_path` to match login behaviour, and to work for users without dashboard permissions.

test/functional/users_controller_test.rb
```diff
@@ -322,6 +322,7 @@ class UsersControllerTest < ActionController::TestCase
   end
 
   test "#login renders login page" do
+    session.clear
```
domcleal, line 325: Instead of undoing the setup method, it might be better to try and move tests relying on this into a context that redefines `setup_user` to an empty method.

test/functional/users_controller_test.rb
```diff
@@ -364,6 +365,39 @@ class UsersControllerTest < ActionController::TestCase
...
 
+  context "an active session is present" do
+    test "#login redirects to root(dashboard)" do
+      get :login, set_session_user
+
+      get :login
+      assert_redirected_to root_path
+    end
+
+    test "for user :one -> #login redirects to root(dashboard)" do
```
domcleal: What does this test that the block above doesn't?

test/functional/users_controller_test.rb
```diff
@@ -364,6 +365,39 @@ class UsersControllerTest < ActionController::TestCase
...
+      get :login, set_session_user.merge(user: users(:one).id)
+
+      get :login
+      assert_redirected_to root_path
+    end
+  end
+
+  context "no active session present" do
+    test "#login renders login page" do
```
domcleal: This looks identical to the "#login renders login page" test on line 324.

test/functional/users_controller_test.rb
```diff
@@ -364,6 +365,39 @@ class UsersControllerTest < ActionController::TestCase
...
+      session.clear
+      get :login
+      assert_template :login
+    end
+
+    test "logout a logged-in user and visit /users/login should render login page" do
```
domcleal: I don't think this tests anything that you haven't tested already.

app/controllers/users_controller.rb
```diff
@@ -7,6 +7,7 @@ class UsersController < ApplicationController
   skip_before_action :require_login, :authorize, :session_expiry, :update_activity_time, :set_taxonomy, :set_gettext_locale_db, :only => [:login, :logout, :extlogout]
   skip_before_action :authorize, :only => :extlogin
   after_action       :update_activity_time, :only => :login
+  before_action      :verify_active_session, only: :login, unless: -> { request.post? }
```
domcleal: I'd suggest sticking with the same hash style as the previous line for readability.

domcleal (changes requested):
Please address the comments from @theforeman-bot about the length and style of your commit messages, see the links provided for more details.  Please also squash the commits to one.

test/functional/users_controller_test.rb
swapab: line 324 tests the response of :login
In above case the test is on rendered template. Maybe I can extend 324 case itself. Wont't that violate 1 assert per test standard ?

``` ruby
  test "#login renders login page" do                                                                                                                                                         
    session.clear                                                                                                                                                                             
    get :login                                                                                                                                                                                
    assert_response :success                                                                                                                                                                  
  end
```

test/functional/users_controller_test.rb
swapab: I thought to also test it with different user(:one) along with default(:admin). Just to make sure  users with different role behaves the same.

test/functional/users_controller_test.rb
swapab: Sure @domcleal I am sorry I didn't quite get you. Could you please give an example?

swapab:
@domcleal Thank you for reviewing my PR.

I have made the requested changes. Could you please have a look ?

I have also rebased the commits to make it look cleaner.

test/functional/users_controller_test.rb
swapab: This should make sense now, with the new change of redirecting to `:back`.

app/controllers/users_controller.rb
swapab: Yes, that makes sense.
Previous url OR hosts_path in case no `HTTP_REFERER`

app/controllers/users_controller.rb
swapab: Changed.

app/controllers/users_controller.rb
swapab: Updated hash style.

test/functional/users_controller_test.rb
domcleal: Please disregard, I don't think it's worthwhile any more - there were more instances of it in the previous version of the PR.

test/functional/users_controller_test.rb
```diff
@@ -364,6 +365,42 @@ class UsersControllerTest < ActionController::TestCase
     assert_response :unprocessable_entity
   end
 
+  context "when user is logged in" do
+    setup do
+      @previous_url = "/bookmarks"
+      get :login, set_session_user
```
domcleal, line 371: I think this can be removed, it doesn't appear to be testing anything. The main requests are in the body of the test.

test/functional/users_controller_test.rb
```diff
@@ -364,6 +365,42 @@ class UsersControllerTest < ActionController::TestCase
...
+    test "#login if referer absent redirect to hosts_path" do
+      request.env['HTTP_REFERER'] = nil
+
+      get :login
+
+      assert_redirected_to hosts_path
+    end
+  end
+
+  context "when user has logged out" do
```
domcleal: I think you can remove this block, it tests the same thing as "#login renders login page" (rendering with an empty session).

app/controllers/users_controller.rb
```diff
@@ -155,4 +156,16 @@ def login_user(user)
   def parameter_filter_context
     Foreman::Controller::Parameters::User::Context.new(:ui, controller_name, params[:action], editing_self?)
   end
+
+  def verify_active_session
+    if !request.post? && params[:status].blank? && User.exists?(session[:user].presence)
+      warning _("You have already logged in")
+      redirect_to back_or_hosts
```
domcleal: There is already a `redirect_back_or_to` helper in the application controller, you should be able to call `redirect_back_or_to hosts_path`.

domcleal (changes requested):

domcleal:
[test foreman]

swapab:
@domcleal http://ci.theforeman.org/job/test_develop_pr_core/10484/#showFailuresLink
640 failures ? Are these usual ?
I am sure it is getting addressed somewhere. I would be happy to fix them.

test/functional/users_controller_test.rb
swapab: Test refactored.

test/functional/users_controller_test.rb
swapab: Removed.

app/controllers/users_controller.rb
swapab: Nice. I used the helper method.
Thanks!

domcleal:
Thanks for the update. They will be fixed under ticket http://projects.theforeman.org/issues/16599, I don't know if anybody is working on it right now.

I'll retest once the develop branch has stabilised again.

test/functional/users_controller_test.rb
domcleal: Sorry, I missed this point. It's worth leaving out unless there's a difference in the tested code for the two types of user, just to minimise the test time.

test/functional/users_controller_test.rb
domcleal: Sorry, I missed this point. One assertion per test isn't something we try to implement, it's OK if it's asserting multiple things as the test environment/request is identical.

Consider adding messages or comments to the assertions if it's unclear why they're testing what they do, but generally it's clearer to have one test block testing different things about the same request rather than spreading it out.

(So if you'd like to add the `assert_template :login` back to the existing test, that'd be fine.)

domcleal (approved):

domcleal:
ok to test

domcleal:
Merged as be36ce9, thanks @swapnilabnave.

PR #3850, opened 2016-09-15
Fixes #16566 - Allow css inline images

tbrisker:
Secure headers currently prevents inlining images in CSS using the
'data:' method. This add that to the allowed image sources.


lzap:
There are relevant test failures.

domcleal:
Merged as e2b73df, thanks @tbrisker.

PR #3851, opened 2016-09-15
Fixes #13704 - Defensive association name setter

shlomizadok:
When setting an association by name (e.g., hostgroup_name, organization_name)
and the name does not exist, we raise a not found error

### DIFF
--- a/test/lib/has_many_common_test.rb
+++ b/test/lib/has_many_common_test.rb
@@ -79,6 +79,13 @@ class HasManyCommonTest < ActiveSupport::TestCase
     refute_equal orig_id, new_id
   end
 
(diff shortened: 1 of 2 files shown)


test/lib/has_many_common_test.rb
```diff
@@ -79,6 +79,13 @@ class HasManyCommonTest < ActiveSupport::TestCase
     refute_equal orig_id, new_id
   end
 
+  test "should rails error if hostgroup name does not exist" do
```
shlomizadok: lol... should raise - not should _rails_

tbrisker:
Merged as 65af0932ee5ed8cf9a1ad1ac3c55958fcea77ab7, thanks @shlomizadok!

PR #3853, opened 2016-09-16
fixes #16578 - make keep_param idempotent, remove duplicate call

domcleal:
Using host_params (and keep_param) twice was causing compute_attributes
to be deleted permanently out of `params`. keep_param now restores
elements it deletes, and the controller now only calls host_params once
for efficiency.

### DIFF
--- a/app/controllers/concerns/foreman/controller/parameters/keep_param.rb
+++ b/app/controllers/concerns/foreman/controller/parameters/keep_param.rb
@@ -6,9 +6,15 @@
 # working around https://github.com/rails/rails/issues/9454.
 module Foreman::Controller::Parameters::KeepParam
   def keep_param(params, top_level_hash, *keys)
+    # Delete keys being kept from the `params` hash, so the block yielded to filters the others
     old_params = keys.inject({}) do |op,(key,val)|
       params[top_level_hash].has_key?(key) ? op.update(key => params[top_level_hash].delete(key)) : op
     end
-    yield.update old_params
+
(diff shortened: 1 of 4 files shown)


app/controllers/concerns/foreman/controller/parameters/keep_param.rb
```diff
@@ -9,6 +9,8 @@ def keep_param(params, top_level_hash, *keys)
     old_params = keys.inject({}) do |op,(key,val)|
       params[top_level_hash].has_key?(key) ? op.update(key => params[top_level_hash].delete(key)) : op
     end
-    yield.update old_params
+    filtered = yield.update(old_params)
+    params[top_level_hash].update old_params
```
dLobatog, line 17: Please leave a comment here explaining this restores deleted `params` , I don't think it's very clear at first glance

dLobatog (changes requested):
Just a small comment to explain what keep_params is doing and why - otherwise :+1: , thanks for fixing this @domcleal

domcleal:
@dLobatog comments added for both restoration lines, hopefully it's clearer now.

dLobatog (approved):

dLobatog:
Ready to merge if Jenkins tests pass

dLobatog:
[test foreman]

domcleal:
[test foreman] as develop changed.

dLobatog:
Thanks @domcleal !

PR #3854, opened 2016-09-16
Fixes #16577 - enhance unit detection

ares:


lib/core_extensions.rb
```diff
@@ -140,17 +140,20 @@ def to_translation
   end
 
   def to_gb
-    value, _, unit=self.match(/(\d+(\.\d+)?) ?(([KMGT]i?B?|B))$/i)[1..3]
```
ares, line 143: this used to fail, #match return nil if there's no match and we called `[1..3]` on it

brandonweeks (approved):

lib/core_extensions.rb
```diff
@@ -140,17 +140,20 @@ def to_translation
...
+    match_data = self.match(/(\d+(\.\d+)?) ?(([KMGT]i?B?|B|Bytes))$/i)
```
tbrisker: Shouldn't the match start at the start of the string?

lib/core_extensions.rb
```diff
@@ -140,17 +140,20 @@ def to_translation
...
+    if match_data.present?
+      value, _, unit = match_data[1..3]
+    else
+      raise "Unknown string: #{self.inspect}!"
+    end
     case unit.to_sym
```
tbrisker: we use case insensitive matching, so for example "10gb" will be caught, but there is no `:gb` symbol - perhaps `.lower.to_sym`?

tbrisker (changes requested):

lib/core_extensions.rb
ares: I didn't want to mix refactorings with fix since I thought this should be merged into 1.13 but since it was rejected I can fix existing code.

lib/core_extensions.rb
ares: alright, even though gB is not gb

ares:
thanks @tbrisker, I made changes you requested, could you take another look?

ares:
[test foreman]

tbrisker (approved):

dLobatog:
[test foreman]

lib/core_extensions.rb
tbrisker: technically GiB isn't the same as GB either, but we ignore that as well :)

tbrisker:
Merged as 0771cec2351c75514abaf1c7040cdd5b74e8ef1d, thanks @ares!

lib/core_extensions.rb
ares: right :-)

PR #3855, opened 2016-09-16
fixes #16580 - redact BMC password in ENC/templates with setting

domcleal:
Passwords stored against BMC NICs are accessible by default via
nic.password in a template, even in safe mode (861a03b), and are also
exported in the ENC output. A new bmc_credentials_accessible setting
allows the admin to redact the values in both locations, only allowing
the password to be used internally for BMC smart proxies.

Disabling BMC credential access requires safemode_render for complete
protection.

### DIFF
--- a/app/models/nic/bmc.rb
+++ b/app/models/nic/bmc.rb
@@ -12,7 +12,7 @@ module Nic
     end
     alias_method :virtual?, :virtual
 
-    attr_exportable :provider, :username, :password => ->(nic) { nic.decrypt_field(nic.password) }
+    attr_exportable :provider, :username, :password
 
     def proxy
       proxy = bmc_proxy
@@ -20,7 +20,7 @@ module Nic
       ProxyAPI::BMC.new({ :host_ip  => ip,
                           :url      => proxy.url,
                           :user     => username,
-                          :password => password })
+                          :password => password_unredacted })
     end
 
     def self.humanized_name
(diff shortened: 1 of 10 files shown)


app/models/nic/bmc.rb
```diff
@@ -12,15 +12,15 @@ def virtual
     end
     alias_method :virtual?, :virtual
 
-    attr_exportable :provider, :username, :password => ->(nic) { nic.decrypt_field(nic.password) }
+    attr_exportable :provider, :username, :password
```
lzap, line 15: The code looks good, but I struggle to understand where the decrypt routine ended up now. The behaviour on this line was to immediately decrypt the field while in your patch `password` is an alias to `password_unredacted` under some conditions (or `nil` which is not relevant).

At which stage the password gets decrypted? 

Do we have a test coverage to assure the password is encrypted in the database and decrypted correctly when using this API?

lzap (commented):
Other than that, it`s a merge.

app/models/nic/bmc.rb
domcleal: The Encryptable module automatically defines the `password` getter with a decryption routine, so this was redundant (https://github.com/theforeman/foreman/blob/develop/app/models/concerns/encryptable.rb#L32), overwriting the default Rails getter which would return the DB contents. The alias_method I added then renames the `password` getter to `password_unredacted`.

There was test coverage already at https://github.com/theforeman/foreman/blob/develop/test/unit/nics/bmc_test.rb#L29

lzap (approved):

app/models/nic/bmc.rb
lzap: Thanks, I will test this shortly.

PR #3856, opened 2016-09-16
fixes #16547 - handle nil for interfaces from facts

thomasmckay:


app/models/host/base.rb
```diff
@@ -175,7 +173,8 @@ def set_interfaces(parser)
       end
 
       parser.interfaces.each do |name, attributes|
-        iface = get_interface_scope(name, attributes).first || interface_class(name).new(:managed => false)
+        iface_scope = get_interface_scope(name, attributes)
```
domcleal, line 179: Which code path leads this to return nil?

app/models/host/base.rb
```diff
@@ -175,7 +173,8 @@ def set_interfaces(parser)
...
+        iface = iface_scope.nil? ? interface_class(name).new(:managed => false) : iface_scope.first
```
domcleal: This removes important handling of `iface_scope.first` being nil, hence `iface` remains nil.

domcleal (changes requested):
Please update the title and commit message to use the existing Redmine ticket, #16547.

Tests are failing and also missing for the issue that this fixes, please address.

app/models/host/base.rb
thomasmckay: https://github.com/theforeman/foreman/pull/3856/files#diff-07cbb99bfdb4bed86ad02486bb5adde2R384

app/models/host/base.rb
```diff
@@ -175,7 +175,10 @@ def set_interfaces(parser)
       end
 
       parser.interfaces.each do |name, attributes|
-        iface = get_interface_scope(name, attributes).first || interface_class(name).new(:managed => false)
+        iface = nil
```
ares, line 178: how about

``` ruby
iface = get_interface_scope(name, attributes).try(:first) || interface_class(name).new(:managed => false)
```

that should be equivalent to 4 lines patch

ares:
it would be great if you could add a test for the case that cause the scope to be nil

app/models/host/base.rb
domcleal: The link doesn't show anything for me, and I don't have that commit SHA. All code paths from `get_interface_scope` should return a scope, so please expand on why that returns nil? This appears to be working around a bug elsewhere.

As already noted, the unit test would assist.

ares: I think the part with `elsif`s does not have `else` which can result in nil, so if `!attributes[:virtual] && !mac_based.physical.any? && self.managed` then it can be nil. The nil should be valid return value, it indicates we didn't find matching interface and we want to create new one.

domcleal: Thanks, that clears it up. The `nil` return value was introduced in 7b75a6a, and it looks to me like the fallback for creating a new interface will be correct.

domcleal:
I've opened #3879 as an alternative with tests and review comments addressed.

dLobatog:
Closing as #3879 was merged

PR #3858, opened 2016-09-18
Fixes #14837 - Reword import buttons

shlomizadok:
Import buttons now say where they import from.

In /environments it looks like:
image

In /puppetclasses, it would state "Import environments from..." as it is what it is doing
and would look like:

image

### DIFF
--- a/app/helpers/puppetclasses_and_environments_helper.rb
+++ b/app/helpers/puppetclasses_and_environments_helper.rb
@@ -12,12 +12,13 @@ module PuppetclassesAndEnvironmentsHelper
   end
 
   def import_proxy_select(hash)
-    select_action_button(_('Import'), {}, import_proxy_links(hash))
+    select_action_button(_('Import'), {}, import_proxy_links(hash, true))
   end
 
-  def import_proxy_links(hash, classes = nil)
+  def import_proxy_links(hash, import_env_text = false, classes = nil)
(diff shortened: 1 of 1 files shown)


app/helpers/puppetclasses_and_environments_helper.rb
```diff
@@ -12,12 +12,13 @@ def class_update_text(pcs, env)
   end
 
   def import_proxy_select(hash)
-    select_action_button(_('Import'), {}, import_proxy_links(hash))
+    select_action_button(_('Import'), {}, import_proxy_links(hash, true))
   end
 
-  def import_proxy_links(hash, classes = nil)
+  def import_proxy_links(hash, import_env_text = false, classes = nil)
+    import_from_text = import_env_text ? "Import environments from %s" : "Import classes from %s"
```
domcleal: These strings must be extracted, move the `_()` from line 21.

domcleal (changes requested):

shlomizadok:
@domcleal - changed. Thanks

domcleal (approved):

domcleal:
Thanks, that's all fine subject to develop being fixed in #16599 and tests passing.

domcleal:
[test foreman]

domcleal:
Merged as 2272591, thanks @shlomizadok.

PR #3859, opened 2016-09-18
Fixes #14935 - Redirect self-editing user to /users

shlomizadok:


domcleal:
I don't think I understand what you're trying to fix here. Should this not store the previous page, like other edit actions do when accessed from indexes (e.g. they preserve searches etc)?

shlomizadok:
If you are in `/environments` and then go to `My account`, edit and save, you will be redirected back to `/environments` which seems (I think) out of context. If I was editing a user [myself]) shouldn't I return to /users?

domcleal:
I would say no, that redirecting to `/users` would be out of context and that I'd expect to return to the previous page after submitting the form.

shlomizadok:
OK, so shall I close this and also reject the issue?

domcleal:
Up to you, but I'd be inclined to, unless another committer wishes to merge this.

ares:
The original issue asked for staying at "my account" page which would mean staying on page with form. I think redirecting to users list after "my account" edit feels more natural to me but I don't have strong preference. Redmine, which also has my account link, remains on the user edit page, just prints notice that it has been saved. So opinions from others are more than welcome.

mbacovsky:
Returning to place where I was at  before hitting 'My Account' feels natural. So I'd vote for keeping the existing strategy. 
However while testing this I found out my instances do this just for the first time and then they keep returning to the same place: e.g. I'm on 'Dashbord' -> My account -> Submit -> I'm redirected to Dashboard -> go to Architectures -> My Account -> Submit -> I'm redirected to Dashboard (always from now on). If I log out and login I can  click My Account from e.g. Settings and keep returning there for the whole session. That seems like a bug. Even more interesting is that from pages added by Katello plugin submitting My Account redirects me  to the My Account form.

dhlavac:
I agree with @ares , redirecting to users list after editing "My Account" looks better

shlomizadok:
First let's see that [test foreman] passes :)
Second, I think that it is quite common that when saving an object, you'd return to the object's index.

ShimShtein:
I agree with @mbacovsky: When Katello is added to the mix - the redirect start behave unexpectedly. IMHO, a simpler rule like always return to users list is better. The system will be more predictable this way.
If our main concern is that the users list page is less useful - we can open it to configuration, so each user could decide what is the returning page for him.

shlomizadok:
@theforeman/core - rebased. Would love to have a verdict :see_no_evil: on this pull request :)

timogoebel:
I just took a quick vote here in the office among fellow Foreman users. It's 3:1 in favor of keeping the current behaviour. Me being the only one that likes the redirect to `/users`.

adamruzicka:
+1 for redirecting to `/users`

mbacovsky:
From users perspective I'd say return `/users` makes not much sense as it is not obvious that going to `My account` means editing one record from `/users` list. 

So for me it is :+1:  for returning to something fixed if we can't fix the current behaviour but it makes not much sense to return to `/users` then to return to `/settings` or `/hosts` or whatever.

dLobatog:
:-1:  to redirecting to '/users'. It's an avoidable 403. Regular users should not have access to that path even, only managers or admin users. I agree with @domcleal if I'm editing environments and I change something in my account, I want the controller to redirect to back, not to some random place

ShimShtein:
> if I'm editing environments and I change something in my account, I want the controller to redirect to back

You are right, the problem begins when you are editing Katello's content views for example. In this case you will be redirected to whatever you did before editing content views.

If `/users` is not good enough, let's make it `/hosts` or we can make it even configurable. IMHO anything would be better than non-consistent behavior.

shlomizadok:
The problem today is that you are not redirected back to where you where. This PR's attempt is to normalize the path of where to return. would `root_path` be sufficient?

dLobatog:
@ShimShtein Then the problem appears to be in Katello?

ShimShtein:
> Then the problem appears to be in Katello?

@dLobatog or any other plugin that does AJAX calls (Katello is a good example, because of its SPA design)

shlomizadok:
@theforeman/core - Shall I close this?

ares:
Doing the math
  
@shlomizadok /users
@domcleal  keep
@ares /users
@mbacovsky keep
@ShimShtein /user
@dhlavac /users
@timogoebel 3x keep, 1 /users
@adamruzicka /users
@dLobatog keep

results
* /users **6**
* keep **6** (Timo had 4 votes though :-))

Hm... we need one more vote I guess? Or would anyone's vote change if we redirected to `root_path`? I'd like to avoid making this configurable, we already have too many options. @rohoover what would be better UX?

TL;DR description: user is on e.g. hosts page, in top right corner clicks on my account link, it takes him/her to form where he/she can edit his/her account, where should we redirect to after submitting? Options are - always to specific page, e.g. user list or always to page from which he/she came to my account, e.g. hosts page, next time dashboard etc.

Rohoover:
I vote for /users. 

Going back to environments or hosts page, etc, assumes they wanted to go back there. I would opine that they intentionally wanted to go to their account settings and may have further activities there. Is there a reason you'd need to leave environment (for example) and edit your account in order for something to happen in environment?

Tangentially: I find it interesting that when you click on my account, you are given a full page view of your account but no other users. When you go to the Users section, you get two pane view. When you click on your row entry, it remains in two pane view.

ohadlevy:
how about a third option - open it within a modal?

I worry that a user with no permissions will get a 404 or a very limited list of users, or even might not have access to the /users in the first place?

Rohoover:
@ohadlevy 

+1

mbacovsky:
@ohadlevy, that would make best sense for me.

ares:
@shlomizadok would you mind trying to do it via modal window? I think it might work best even though there would be 5 tabs in this window.

shlomizadok:
Moved to editing "My Account" in modal window.
It currently looks like:

image

To do:

- [x] Handle Errors
- [x] Add / fix tests
- [x] Cleanup / Hound

shlomizadok:
Closing for the time being.

xprazak2:
Any progress on this?

ares:
I think this would be good task for UI team to take a look? IIRC it was mostly ready.

PR #3860, opened 2016-09-19
Fixes #16603 - Add translations to the new org process

bkearney:


dLobatog:
@bkearney We have this in most models and it works (try subnet or architectures) - why does it fail to show the translation here?

https://gist.github.com/dLobatog/9a27cc32335dd74c1c5e62515bcc9c50

bkearney:
@dLobatog I am not sure. I do not see the label or the description field int the model attributes for taxonomies. So, perhaps something to do with model inheritance?

bkearney:
@dLobatog I can not reomve the waiting on contributor label :(

app/views/taxonomies/_form.html.erb
```diff
@@ -59,8 +59,8 @@
                        :'data-url' => (controller_name == 'organizations' ? parent_taxonomy_selected_organization_path(taxonomy.id) :
                                parent_taxonomy_selected_location_path(taxonomy.id))
                      }) if show_parent?(taxonomy) %>
-        <%= text_f f, :name %>
-        <%= textarea_f f, :description, :rows => 5 %>
+        <%= text_f f, :name, :label => _('Name') %>
```
domcleal, line 62: Why extract the name? `Taxonomy|Name` is extracted via model_attributes.rb.

app/views/taxonomies/_form.html.erb
```diff
@@ -59,8 +59,8 @@
...
+        <%= textarea_f f, :description, :rows => 5, :label => _('Description') %>
```
domcleal, line 63: Use model_attributes.rb.

domcleal (changes requested):
> @dLobatog I am not sure. I do not see the label or the description field int the model attributes for taxonomies.

Label isn't an attribute of Taxonomy, it may not come from Foreman. Description can be added to the model_attributes.rb file rather than being extracted in the UI.

domcleal:
Your bug report says "Organization" isn't translated, but where exactly is that? I don't see it changed in the PR.

bkearney:
@domcleal do I edit the model file directly to add the new column, or is it extracted?

domcleal:
It's generated, see locale:find_model.

PR #3861, opened 2016-09-20
Fixes #16599 - Shrinkwrap npm modules

tbrisker:
This is a temporary workaround to pin the entire npm module tree until
we can upgrade node and npm on jenkins to newer versions.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,0 +1,16 @@
+{
+  "name": "TheForemanDevDeps",
+  "version": "1.14.0",
+  "dependencies": {
+    "postcss-svgo": {
+      "version": "2.1.4",
+      "from": "postcss-svgo@>=2.1.1 <3.0.0",
+      "resolved": "https://registry.npmjs.org/postcss-svgo/-/postcss-svgo-2.1.4.tgz"
+    },
+    "svgo": {
+      "version": "0.6.6",
+      "from": "svgo@>=0.6.0 <0.7.0",
+      "resolved": "https://registry.npmjs.org/svgo/-/svgo-0.6.6.tgz"
+    }
+  }
+}


dLobatog:
@tbrisker We can merge this and get tests to pass, however the issue runs deeper, el7 still ships 0.10, and I believe we should test in that environment (most of our userbase is in el7). 

Doesn't pinning `svgo` fix the issue, or do we need any new features from svgo that are not available in 0.10 compatible versions?

tbrisker:
@dLobatog looks like tests are still failing, i'll keep working on this. We can't pin just svgo since it is a sub dependency (actually about 4 levels deep) of css-loader. If we were using npm3 we could create a shrinkwrap for just that version (or possibly only its tree), but currently we have to pin the entire tree, at least for that package. I don't think we need to stay with node 0.10 since this is just used for compiling the assets, our users don't need to do that.

dLobatog:
@tbrisker Our users don't do that, but we do compile the assets prior to making the RPMs on el7, using 0.10

domcleal:
We compile assets on a number of platforms, not just EL7 - three of our five supported OSes use NodeJS 0.10.x.

johnpmitsch:
FWIW, this fixed the issue in my dev environment

swapab:
Upgrading nodejs to 6.6.0 worked for my dev environment.

tbrisker:
Updated this to use npm3's better ability to only pin a certain module, this will fail tests until https://github.com/theforeman/foreman-infra/pull/239 gets merged.

domcleal:
Will this require a minimum of npm 3.x to install, even when using NodeJS 0.10.0?  (That isn't a dependency we have on all OSes either, the two are somewhat linked.)

tbrisker:
@domcleal npm 3.x supports NodeJS>=0.10 [1](https://github.com/npm/npm). In fact, the official way of upgrading npm is by running "npm install npm -g" [2](https://docs.npmjs.com/getting-started/installing-node)

domcleal:
[test foreman]

domcleal (approved):

domcleal:
Merged as f7ccfcc, thanks @tbrisker.

PR #3863, opened 2016-09-20
Fixes #16622 - Disambiguate resource_name column name in #hosts_count

brandonweeks:


domcleal:
Could you add a test to test/unit/helpers/application_helper_test.rb please? You can create a user with a limited filter joining another table (like hostgroups), then with `as_user` check that the function succeeds.

lzap (approved):
LGTM

lzap:
Whoops, I tested tho, works fine.

domcleal:
Here's a suggestion for a unit test:

<pre>
test '#hosts_count works for non-admin user with joining filter' do
  host = FactoryGirl.create(:host, :with_hostgroup)
  user_role = FactoryGirl.create(:user_user_role)
  FactoryGirl.create(:filter, :role => user_role.role, :permissions => Permission.where(:name => 'view_hosts'), :search => "hostgroup = #{host.hostgroup.name}")
  as_user user_role.owner do
    assert_equal({host.operatingsystem_id => 1}, hosts_count('operatingsystem'))
  end
end
</pre>

The issue was fixed in Rails 4.2.7 via rails/rails@2c894cb3adb8f1c1f2653af05879a3e96d0add21, so this only affects RPM installations which use an older version (I'll simultaneously patch them). You can reproduce it with the test above by changing the Rails version to 4.2.6 in the Gemfile.

domcleal:
Thanks for the PR, I'm going to close it now as I think the issue's been fixed sufficiently in Rails (gems and packages) without needing this patch.

PR #3864, opened 2016-09-20
Fixes #16624 - Make AuthSourceLDAP taxable

dLobatog:
This allows users to set organizations/locations on AuthSourceLDAP
objects. That in itself might not be that useful, but it allows us to
follow on and assign the AuthSourceLDAP taxonomies to the users
autocreated through it.

### DIFF
--- a/app/models/auth_sources/auth_source_ldap.rb
+++ b/app/models/auth_sources/auth_source_ldap.rb
@@ -27,6 +27,7 @@ class AuthSourceLdap < AuthSource
   include Parameterizable::ByIdName
   include Encryptable
   encrypts :account_password
+  include Taxonomix
 
   validates :host, :presence => true, :length => {:maximum => 60}
   validates :attr_login, :attr_firstname, :attr_lastname, :attr_mail, :presence => true, :if => Proc.new { |auth| auth.onthefly_register? }
@@ -38,6 +39,14 @@ class AuthSourceLdap < AuthSource
 
   before_validation :strip_ldap_attributes
   after_initialize :set_defaults
+
+  scoped_search :on => :name, :complete_value => :true
+
+  default_scope lambda {
+    with_taxonomy_scope do
+      order("#{AuthSourceLdap.table_name}.name")
+    end
+  }
 
   DEFAULT_PORTS = {:ldap => 389, :ldaps => 636 }
   # Loads the LDAP info for a user and authenticates the user with their password
--- a/app/models/concerns/taxonomix.rb
+++ b/app/models/concerns/taxonomix.rb
@@ -219,7 +219,7 @@ module Taxonomix
   end
 
   def used_taxonomy_ids(type)
-    return [] if new_record?
+    return [] if new_record? || !self.respond_to?(:hosts)
     Host::Base.where(taxonomy_foreign_key_conditions).uniq.pluck(type).compact
   end
 
--- a/app/controllers/api/v2/auth_source_ldaps_controller.rb
+++ b/app/controllers/api/v2/auth_source_ldaps_controller.rb
@@ -1,11 +1,21 @@
 module Api
   module V2
     class AuthSourceLdapsController < V2::BaseController
+      include Api::TaxonomyScope
       include Foreman::Controller::Parameters::AuthSourceLdap
+
+      wrap_parameters AuthSourceLdap,
+        :include => auth_source_ldap_params_filter.
+                      accessible_attributes(parameter_filter_context)
 
       before_action :find_resource, :only => %w{show update destroy test}
 
       api :GET, "/auth_source_ldaps/", N_("List all LDAP authentication sources")
+      api :GET, '/locations/:location_id/auth_source_ldaps/',
+        N_('List LDAP authentication sources per location')
+      api :GET, '/organizations/:organization_id/auth_source_ldaps/',
+        N_('List LDAP authentication sources per organization')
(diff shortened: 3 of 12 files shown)


app/models/concerns/taxonomix.rb
```diff
@@ -219,7 +219,7 @@ def taxonomy_foreign_key_conditions
   end
 
   def used_taxonomy_ids(type)
-    return [] if new_record?
+    return [] if new_record? || !self.respond_to?(:hosts)
```
dLobatog, line 222: This allows the loc_orgs tab partial to work for models that are not directly associated with hosts (such as this one, or filter, or roles which I plan to move to use the partial after this is merged)

test/functional/api/v2/auth_source_ldaps_controller_test.rb
```diff
@@ -53,4 +53,17 @@ class Api::V2::AuthSourceLdapsControllerTest < ActionController::TestCase
...
   end
+
+  test 'taxonomies can be set' do
+    AuthSourceLdap.any_instance.stubs(:test_connection).raises(Foreman::Exception)
+    put :update, { :id => auth_sources(:one).to_param,
+                   :organization_names => [taxonomies(:organization1).name],
+                   :location_ids => [taxonomies(:location1).id] }
+    show_response = ActiveSupport::JSON.decode(@response.body)
+    assert_response :success
+    assert_equal taxonomies(:location1).id,
```
dLobatog: Notice this test tests both the PUT action accepts the right params, and the RABL in the response

dLobatog:
@ares Can you take a look?

app/controllers/auth_source_ldaps_controller.rb
```diff
@@ -47,4 +47,8 @@ def test_connection
     Foreman::Logging.exception("Failed to connect to LDAP server", exception)
     render :json => {:message => exception.message}, :status => :unprocessable_entity
   end
+
+  def controller_permission
+    'authenticators'
```
ares, line 54: is this separate issue that non-admin user could not work with auth source ldaps? and if so, isn't there the same problem for API controllers?

app/models/auth_sources/auth_source_ldap.rb
```diff
@@ -27,6 +27,7 @@ class AuthSourceLdap < AuthSource
   include Parameterizable::ByIdName
   include Encryptable
   encrypts :account_password
+  include Taxonomix
```
ares, line 30: I think it would be also desirable for AuthSourceExternal, when user is authenticated externally (apache module) it would be good to assign taxonomies too

ares (changes requested):
not tested yet but seems solid

app/controllers/auth_source_ldaps_controller.rb
dLobatog: Yeah it's kind of a separate issue, I noticed while testing it that permissions to non-admin users did not apply properly

dLobatog: And yeah it applies to the API controller too... will put this in a separate PR

app/models/auth_sources/auth_source_ldap.rb
dLobatog: How would you assign taxonomies to an AuthSourceExternal though? There's not even a controller to modify it

ares: for some reason I was under impression it was listed among other LDAP auth sources, but looking at the code now proves me wrong => please ignore this

dLobatog:
Updated, the permissions issue now is a separate PR here - https://github.com/theforeman/foreman/pull/3872 cc @ares

ares (approved):

ares:
[test foreman]

ares:
The [test failure](http://ci.theforeman.org/job/test_develop_pr_core/10577/database=mysql,ruby=2.2,slave=fast/testReport/junit/%28root%29/AuthSourceTest/test_0002_should_return_search_results_if_search_free_text_is_auth_source_name/) looks relevant, @dLobatog could you take a look? it might be something with STI and Taxonomix only in child, maybe all auth sources should be taxable even though other than LDAP one would be still global.

dLobatog:
@ares Thanks for pointing it out. The error happens because scoped_search, https://github.com/wvanbergen/scoped_search/blob/master/lib/scoped_search/definition.rb#L62 is trying to include 'locations' in the SQL query for an `AuthSource`, as the table has taxonomies.

It tries to find that relation in the AuthSource class, it can't find it and it errors out. Since scoped_search can never be used in the application (there isn't even a controller for non LDAP auth sources), I will move those tests and the scoped_search definition to the LDAP class.

If you look at the auth_sources table - it has all the fields for LDAP, and these are empty for AuthSourceExternal, AuthSourceHidden and AuthSourceInternal. AuthSource is nearly an abstract class, of which we make no use of scope_search of, the only child that does it is AuthSourceLDAP, the others do not need this functionality.

dLobatog:
@ares Should I rebase with develop and push? The integration tests are failing because this patch is based on 'develop' from a few days ago when it was broken

ares:
looks good, could you please rebase @dLobatog?

dLobatog:
Repushed - thanks for looking @ares

ares:
Thanks @dLobatog, merged as 05bd6ed86392a5bd57aad48d79fa916bfa25503b

PR #3865, opened 2016-09-20
Fixes #16625 - Remove show_taxonomy_tabs?

dLobatog:
This helper isn't really needed as the partial already takes care of it.

### DIFF
--- a/app/helpers/taxonomy_helper.rb
+++ b/app/helpers/taxonomy_helper.rb
@@ -11,6 +11,9 @@ module TaxonomyHelper
   end
 
   def show_taxonomy_tabs?
+    Foreman::Deprecation.deprecation_warning(
+      '1.16', 'The partial "taxonomy/loc_org_tabs" checks already if '\
+      'taxonomies are enabled or not.')
     SETTINGS[:locations_enabled] || SETTINGS[:organizations_enabled]
   end
 
--- a/app/views/environments/_form.html.erb
+++ b/app/views/environments/_form.html.erb
@@ -1,28 +1,20 @@
 <%= form_for @environment, :url => (@environment.new_record? ? environments_path : environment_path(:id => @environment)) do |f| %>
   <%= base_errors_for @environment %>
-  <% if show_taxonomy_tabs? %>
-    <ul class="nav nav-tabs" data-tabs="tabs">
-      <li class="active"><a href="#primary" data-toggle="tab"><%= _('Primary') %></a></li>
-      <% if show_location_tab? %>
-        <li><a href="#locations" data-toggle="tab"><%= _('Locations') %></a></li>
-      <% end %>
-      <% if show_organization_tab? %>
-        <li><a href="#organizations" data-toggle="tab"><%= _('Organizations') %></a></li>
-      <% end %>
-    </ul>
+  <ul class="nav nav-tabs" data-tabs="tabs">
+    <li class="active"><a href="#primary" data-toggle="tab"><%= _('Environment') %></a></li>
+    <% if show_location_tab? %>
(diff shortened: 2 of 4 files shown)


app/helpers/taxonomy_helper.rb
```diff
@@ -10,10 +10,6 @@ def show_organization_tab?
     SETTINGS[:organizations_enabled] && User.current.allowed_to?(:view_organizations)
   end
 
-  def show_taxonomy_tabs?
```
domcleal, line 13: I think this should issue a deprecation warning and not be removed immediately, it's used by at least the Foreman discovery plugin: https://github.com/theforeman/foreman_discovery/blob/fe346a413d3a69fc9e878e768caea1119d055d10/app/views/discovery_rules/_form.html.erb#L31.

app/views/environments/_form.html.erb
```diff
@@ -1,28 +1,20 @@
...
-      <li class="active"><a href="#primary" data-toggle="tab"><%= _('Primary') %></a></li>
-      <% if show_location_tab? %>
-        <li><a href="#locations" data-toggle="tab"><%= _('Locations') %></a></li>
-      <% end %>
-      <% if show_organization_tab? %>
-        <li><a href="#organizations" data-toggle="tab"><%= _('Organizations') %></a></li>
-      <% end %>
-    </ul>
+  <ul class="nav nav-tabs" data-tabs="tabs">
+    <li class="active"><a href="#primary" data-toggle="tab"><%= _('Primary') %></a></li>
```
domcleal: I'd suggest changing the tab text to "Environment" while you're here, since it will now be visible by default and "Primary" isn't descriptive. The other two forms have better text already.

domcleal (changes requested):

lzap (approved):
Code looks good, tested. Please re-test foreman when develop is fixed.

app/helpers/taxonomy_helper.rb
dLobatog: Fair enough, will do.

dLobatog:
Updated so that the method is just deprecated until current stable + 3 (1.12 + 3, 1.15) and the environments shows 'Environment' instead of 'Primary' as the tab title.

app/helpers/taxonomy_helper.rb
```diff
@@ -11,6 +11,9 @@ def show_organization_tab?
   end
 
   def show_taxonomy_tabs?
+    Foreman::Deprecation.deprecation_warning(
+      '1.15', 'The partial "taxonomy/loc_org_tabs" checks already if '\
```
domcleal: 1.13's the current (well, latest) stable, so this should be 1.16 (deprecated in 1.14, 1.15 and then is removed in 1.16).

domcleal (changes requested):

dLobatog:
Updated to deprecate this on 1.16

domcleal:
[test foreman]

domcleal (approved):

domcleal:
@dLobatog could you rebase the branch please? The integration tests are failing as it's prior to the fix in #16599. (Sorry, it's a behaviour change I made to fix other failures: https://groups.google.com/d/msg/foreman-dev/VioIxHBFjDs/BWJKXyS5AwAJ)

dLobatog:
No problem, rebased & repushed :smile:

domcleal:
Merged as 1e3f9ac, with the commit message amended to say deprecated, thanks @dLobatog.

Could you drop a very short e-mail to foreman-dev please? (e.g. https://groups.google.com/forum/#!topic/foreman-dev/Q_Czr1bHydw)

PR #3866, opened 2016-09-20
Introduce storybook to represent react components.

ohadlevy:
to use:

npm install storybook

npm run storybook

### DIFF
--- a/.storybook/head.html
+++ b/.storybook/head.html
@@ -1,0 +1,15 @@
+<link href='https://cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/3.3.7/css/bootstrap.min.css' rel='stylesheet' media="screen, print">
+<link href='https://cdnjs.cloudflare.com/ajax/libs/patternfly/3.10.0/css/patternfly.min.css' rel='stylesheet' media="screen, print">
+<link href='https://cdnjs.cloudflare.com/ajax/libs/patternfly/3.10.0/css/patternfly-additions.min.css' rel='stylesheet' media="screen, print">
+<script src='https://cdnjs.cloudflare.com/ajax/libs/jquery/1.12.4/jquery.min.js'></script>
+<script src='https://cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/3.3.7/js/bootstrap.min.js'></script>
+<script src='https://cdnjs.cloudflare.com/ajax/libs/d3/4.3.0/d3.min.js'></script>
+<script src='https://cdnjs.cloudflare.com/ajax/libs/c3/0.4.11/c3.min.js'></script>
+<script src='https://cdnjs.cloudflare.com/ajax/libs/patternfly/3.10.0/js/patternfly.min.js'></script>
+<!-- <script src='https://cdnjs.cloudflare.com/ajax/libs/jed/1.1.1/jed.min.js'></script> -->
+<!-- workaround until Jed is moved to npn - see http://projects.theforeman.org/issues/17263 -->
+<script>
+function __(str) {
+  return str;
+}
+</script>
--- a/webpack/stories/index.js
+++ b/webpack/stories/index.js
@@ -1,0 +1,93 @@
+import React from 'react';
+
+/* eslint-disable no-unused-vars */
+import { storiesOf, action, linkTo, addDecorator } from '@kadira/storybook';
+require('../assets/javascripts/bundle');
+require('../../app/assets/javascripts/application');
+import StatisticsChartsList from
+'../assets/javascripts/react_app/components/charts/StatisticsChartsList';
+import StatisticsChartBox from
+'../assets/javascripts/react_app/components/charts/StatisticsChartBox';
+import PowerStatus from
+'../assets/javascripts/react_app/components/hosts/PowerStatus';
+
+addDecorator((story) => (
+  <div className="ca" style={{textAlign: 'center'}}>
+    {story()}
+    <div id="targetChart"></div>
+  </div>
+));
+
+storiesOf('Statistics', module)
+  .add('Loading', () => (
+    <StatisticsChartBox
+      config={{data: {columns: [] }}}
+      noDataMsg={'No data here'}
+      title="Title"
+      status="PENDING" />
+  ))
+    .add('Without Data', () => (
+    <StatisticsChartBox
+      config={{data: {columns: [] }}}
+      noDataMsg={'No data here'}
+      title="Title"
+      status="RESOLVED" />
+  ))
+    .add('With Error', () => (
+    <StatisticsChartBox
+      config={{data: {columns: [] }}}
+      title="Title"
+      noDataMsg={'No data here'}
+      errorText="Ooops"
+      status="ERROR" />
+  ))
+  .add('With data', () => (
+    <StatisticsChartBox
+      config={{data: {columns: [1, 2]}}}
+      modalConfig={{}}
+      noDataMsg={'No data here'}
+      id="target"
+      status="RESOLVED"
+    />
+  )
+);
+
+storiesOf('Power Status', module)
+  .add('Loading', () => (
+    <PowerStatus
+      loadingStatus="PENDING"
+    />
+  ))
+  .add('ON', () => (
+    <PowerStatus
+      state="on"
+      title="on"
+      loadingStatus="RESOLVED"
+      statusText="On"
+    />
+  ))
+  .add('OFF', () => (
+    <PowerStatus
+      state="off"
+      title="off"
+      loadingStatus="RESOLVED"
+      statusText="Off"
+    />
+  ))
+  .add('N/A', () => (
+    <PowerStatus
+      state="na"
+      statusText="No power support"
+      loadingStatus="RESOLVED"
+      title="N/A"
+    />
+  ))
+    .add('Error', () => (
+    <PowerStatus
+      state="na"
+      statusText="Exception error some where"
+      loadingStatus="ERROR"
+      title="N/A"
+    />
+  )
+);
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -1,0 +1,7 @@
+import { configure } from '@kadira/storybook';
+
+function loadStories() {
+  require('../webpack/stories');
+}
+
+configure(loadStories, module);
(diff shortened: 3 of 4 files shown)


config/webpack.common.config.js
```diff
@@ -0,0 +1,60 @@
+'use strict';
+
+require('es6-promise').polyfill(); // needed for compatibility with older node versions
+
+var path = require('path');
```
gailsteiger: The webpack config files are not processed by babel. Add 

``` js
/* eslint-disable no-var */
```

at the top of the file to disable the no-var rule. 'var' declarations should still be at the top of the function scope.

webpack/assets/javascripts/react_app/components/notifications/Toast.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import {Button} from 'react-bootstrap';
+
+const Toast = (props) => {
+  let type = (props.type || 'success');
```
gailsteiger: use const here instead of let

ohadlevy:
@gailsteiger updated all of hound concerns and your comments.

config/webpack.common.config.js
ohadlevy: added the whole file to eslint ignore.

gailsteiger: I think that's extreme. I think that it would be better to lint the file and disable specific rules that aren't relevant.

ohadlevy:
I was just following the same as its done to webpack configuration it self
currently. /cc @tbrisker as in that case I would assume that we should lint
the current webpack configuration?

gailsteiger:
@ohadlevy I was surprised to find webpack config  in the ignore list. I think we should lint them both.

ohadlevy:
rebased, @tbrisker or @dLobatog will you have a minute to review? thanks!

ohadlevy:
rebased.

.storybook/head.html
```diff
@@ -0,0 +1,14 @@
+<link href='https://cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/3.3.7/css/bootstrap.min.css' rel='stylesheet' media="screen, print">
+<link href='https://cdnjs.cloudflare.com/ajax/libs/patternfly/3.10.0/css/patternfly.min.css' rel='stylesheet' media="screen, print">
+<link href='https://cdnjs.cloudflare.com/ajax/libs/patternfly/3.10.0/css/patternfly-additions.min.css' rel='stylesheet' media="screen, print">
+<script src='https://cdnjs.cloudflare.com/ajax/libs/jquery/1.11.3/jquery.min.js'></script>
+<script src='https://cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/3.3.7/js/bootstrap.min.js'></script>
+<script src='https://cdnjs.cloudflare.com/ajax/libs/d3/4.3.0/d3.min.js'></script>
+<script src='https://cdnjs.cloudflare.com/ajax/libs/c3/0.4.11/c3.min.js'></script>
+<!-- <script src='https://cdnjs.cloudflare.com/ajax/libs/jed/1.1.1/jed.min.js'></script> -->
+<!-- <script src='https://cdnjs.cloudflare.com/ajax/libs/patternfly/3.10.0/js/patternfly.min.js'></script> -->
```
tbrisker: any reason we are using CDN here rather then the versions already available in our assets? might lead to unexpected differences in layout if we have different versions.

.eslintignore
```diff
@@ -2,3 +2,4 @@
 app/**
 vendor/**
 config/webpack.config.js
+config/webpack.common.config.js
```
tbrisker: I think we should just ignore `config/**`, no point linting any of it (since we lint with es6 rules etc and config has to be legacy)

config/webpack.common.config.js
```diff
@@ -0,0 +1,60 @@
+/* eslint-disable no-var */
```
tbrisker: we don't lint this, no point in adding comments for the linter

config/webpack.config.js
```diff
@@ -1,17 +1,15 @@
+/* eslint-disable no-var */
```
tbrisker: ditto

config/webpack.config.js
```diff
@@ -1,17 +1,15 @@
...
-var webpack = require('webpack');
-var StatsPlugin = require('stats-webpack-plugin');
-var ExtractTextPlugin = require("extract-text-webpack-plugin");
+var GlobalConfig = require('../config/webpack.common.config');
 
 // must match config.webpack.dev_server.port
 var devServerPort = 3808;
 
 // set TARGETNODE_ENV=production on the environment to add asset fingerprints
 var production = process.env.RAILS_ENV === 'production' || process.env.NODE_ENV === 'production';
```
tbrisker, line 14: this is duplicated in webpack.common.config.js. Should we also split develop and production into different config files?

webpack/assets/javascripts/react_app/components/charts/Chart.js
```diff
@@ -30,7 +30,9 @@ class Chart extends React.Component {
   }
 
   componentWillUnmount() {
-    this.chart = this.chart.destroy();
+    if (this.hasData()) {
```
tbrisker: is this change related?

webpack/assets/javascripts/react_app/components/common/Search.js
```diff
@@ -0,0 +1,64 @@
+import React, { Component } from 'react';
+import helpers from '../../common/helpers';
+import {Form, FormGroup, ControlLabel, FormControl, Button} from 'react-bootstrap';
+
+class Search extends Component {
```
tbrisker: I would prefer not introducing new component in this PR as the subject is introducing storybook.

webpack/assets/javascripts/react_app/components/notifications/Toast.js
```diff
@@ -0,0 +1,31 @@
...
+import {Button} from 'react-bootstrap';
+
+const ToastIcons = {
+  success: 'ok',
+  danger: 'error-circle-o',
+  warning: 'warning-triangle-o',
+  'default': 'info'
+};
+
+const Toast = (props) => {
```
tbrisker: ditto

config/webpack.config.js
tbrisker: also, if we have multiple config files, maybe place them all under `config/webpack/`?

tbrisker (changes requested):

.storybook/head.html
ohadlevy: once they are moved into npm packages, we can use it, this is kind of quick and dirty approach to get it going

.eslintignore
ohadlevy: I'm in favor of linting it, but I don't want to do it in this pr, regardless, i can change it to ```*``` but I think we should not ignore by default.

config/webpack.common.config.js
ohadlevy: its my editor which complains :)

config/webpack.config.js
ohadlevy: or maybe webpack/config

webpack/assets/javascripts/react_app/components/charts/Chart.js
ohadlevy: semi related, most likely should be done in refactoring to charts components to be able to be used in storybook... /cc @gailsteiger

webpack/assets/javascripts/react_app/components/common/Search.js
ohadlevy: agree, I'll remove it.

config/webpack.config.js
tbrisker: whichever, as long as we paint the bikeshed green :smile:

webpack/stories/index.js
```diff
@@ -0,0 +1,93 @@
...
+      modalConfig={{}}
+      noDataMsg={'No data here'}
+      id="target"
+      status="RESOLVED"
+    />
+  )
+);
+
+storiesOf('Power Status', module)
+  .add('Loading', () => (
```
ohadlevy, line 56: @tbrisker I removed the other components and replace them with power status instead.

@gailsteiger from some reason, the imported css of the PowerStatus.js is not visible in storybook, i suspect something in webpack, but I was not able to identify it, could you please have a look?

gailsteiger: @ohadlevy Seems to work if you remove the '.storybook/webpack.config.js' file.

ohadlevy:
@tbrisker removed all webpack changes (seems like storybook defaults were working fine), added power status stories.

.storybook/head.html
ohadlevy: do you have an alternative idea? (I'm already importing the asset pipieline application.js - however webpack ignores the //= require statements)

.storybook/head.html
```diff
@@ -0,0 +1,15 @@
+<link href='https://cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/3.3.7/css/bootstrap.min.css' rel='stylesheet' media="screen, print">
+<link href='https://cdnjs.cloudflare.com/ajax/libs/patternfly/3.10.0/css/patternfly.min.css' rel='stylesheet' media="screen, print">
+<link href='https://cdnjs.cloudflare.com/ajax/libs/patternfly/3.10.0/css/patternfly-additions.min.css' rel='stylesheet' media="screen, print">
+<script src='https://cdnjs.cloudflare.com/ajax/libs/jquery/1.12.4/jquery.min.js'></script>
```
ohadlevy, line 4: updated to a current jquery version.

I did try to use npm jquery package, however, since bootstrap is not in npm i was unable, as jquery needs to load prior to bootstrap.
the way that bootstrap works is by adding the bundle inside the body vs the head section.

ohadlevy:
@tbrisker @dLobatog anything else needed to get this moving? thanks!

tbrisker (approved):

tbrisker:
Ready to merge on my side, going to give this a few hours before merging in case @dLobatog or anyone else have further comments.

ohadlevy:
@tbrisker :+1:

tbrisker:
Merged, Thanks @ohadlevy !

PR #3868, opened 2016-09-21
Refs #16074 - Load .env when not running a Procfile

xprazak2:
Loads .env file when Webpack is started from node_modules/.bin


xprazak2:
[test foreman]

domcleal:
Can you move the `bindOn` and `require('dotenv')` into the dev server setup block, near the bottom? I think this will reduce the dependencies needed when using webpack in production mode, as the BIND environment var isn't required there.

domcleal:
Oh, please also rebase the branch when you update it to ensure tests pass with the fix from #16599.

domcleal:
Merged as b831082, thanks @xprazak2.

Please do add a comment when changing a PR to say what's changed.

PR #3869, opened 2016-09-21
Fixes #6372 - User inherits taxonomies from LDAP on creation

dLobatog:
In an installation that uses taxonomies, when an user logs in for
the first time and gets its account created via LDAP, it has no
taxonomies.

This is a problem, as even if the user has all the roles it needs from
external user groups, an admin needs to manually set the taxonomies for
every new user.

This commit fixes that by assigning the taxonomies from the LDAP auth
source to the user when it logs in for the first time. It does _not_
update them on every login, as after the user has logged in, user
taxonomies are left to the Foreman administrator to manage.

---

This relies on https://github.com/theforeman/foreman/pull/3864 - so that has to be merged first.

I could see the case in which the Foreman administrator wants the AuthSourceLDAP taxonomies to be an authoritative source for User taxonomies, and hence every time they're changed, the User taxonomies are changed too. 

I did not implement that as it strays very far from the current implementation and might not be necessary in the end. If we were to implement it, it would be simple to just update user taxonomies every time the AuthSourceLDAP is saved and has the 'Propagate orgs/locs to users' property - it would habe to be optional.

cc @ares

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -395,41 +395,45 @@ class User < ActiveRecord::Base
   end
 
   def self.try_to_auto_create_user(login, password)
-    return nil if login.blank? || password.blank?
-
-    # user is not yet registered, try to authenticate with available sources
+    return if login.blank? || password.blank?
+    # User is not yet registered, try to authenticate with available sources
     logger.debug "Attempting to log into an auth source as #{login} for account auto-creation"
-    if (attrs = AuthSource.authenticate(login, password))
-      attrs.delete(:dn)
-      user = new(attrs)
+    return unless (attrs = AuthSource.authenticate(login, password))
 
-      # If an invalid data returned from an authentication source for any attribute(s) then set its value as nil
-      saved_attrs = {}
-      unless user.valid?
-        user.errors.each do |attr|
-          saved_attrs[attr] = user[attr]
-          user[attr] = nil
-        end
-      end
-
-      # The default user can't auto create users, we need to change to Admin for this to work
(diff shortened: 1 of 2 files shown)


timogoebel:
@dLobatog : This looks interesting. The code looks fine, though the diff is a little hard to read. Have you noticed the test failures?

dLobatog:
@timogoebel Yeah, it relies on https://github.com/theforeman/foreman/pull/3864 being merged so it'll always fail. I'll comment on the diff, the change is quite minimal

app/models/user.rb
```diff
@@ -395,41 +395,45 @@ def external_usergroups
...
-      unless user.valid?
-        user.errors.each do |attr|
-          saved_attrs[attr] = user[attr]
-          user[attr] = nil
-        end
+    return unless (attrs = AuthSource.authenticate(login, password))
+
+    attrs.delete(:dn)
+    user = new(attrs)
+    # Inherit taxonomies from authentication source on creation
```
dLobatog, line 405: These four lines are the only thing that changed. The rest of the diff is shown as 'diff' because I changed the huge 'if' statement for an early return.

dLobatog:
Rebased and pushed so that tests run now that the other PR has been merged

timogoebel:
Code looks good, tests are green, manual testing confirmed, that it worked. 👍

timogoebel:
Merged as 4db325b086b765b8a0345e27f02ab941eeb10a42, thanks @dLobatog!

PR #3871, opened 2016-09-21
fixes #16635 - welcome page works with scoped models

timogoebel:


timogoebel:
[test foreman]

lzap:
What was the error message? I am not getting it. Can you add a test to verify this action works fine?

timogoebel:
@lzap: Thanks for taking a look.
Let me explain the motiviation for this change:
If you have a model, say "ForemanOmaha:: OmahaReport" and a controller "OmahaReportsController", before the change the Model (named klass here) was resolved to "OmahaReport" so the could would call `OmahaReport.first.nil?`, but it should have called `ForemanOmaha:: OmahaReport.first.nil`? There is already the method `model_of_controller` that handles the class resolving this better.

I can add a test for this, sure.

timogoebel:
@lzap: Finally added the tests. Sorry for the delay.

lzap:
[test foreman]

timogoebel:
@lzap : Thanks. Tests are ok now.

lzap:
Rebased and merged via the new button! Crossing fingers. Thanks @timogoebel !

PR #3874, opened 2016-09-21
Fixes #16643 - Display origin for facts

dLobatog:
This commit sets the interfaces for the fact importers to start setting
a Fact origin, and display it in the UI. If this is accepted, I'll send PRs to foreman_chef, foreman_ansible and foreman_salt to adapt to this new interface. 

It looks like this:

screenshot from 2016-09-21 18-52-02


ares:
I think the ansible icon should be added by ansible plugin.

app/helpers/fact_values_helper.rb
```diff
@@ -43,4 +43,9 @@ def show_full_fact_value(fact_value)
       end
     end.html_safe
   end
+
+  def fact_origin_icon(origin)
+    return "N/A" if origin == 'N/A'
```
ares: `return origin if origin == 'N/A'` perhaps

db/migrate/20160921160155_add_origin_to_fact_values.rb
```diff
@@ -0,0 +1,8 @@
+class AddOriginToFactValues < ActiveRecord::Migration
+  def change
+    add_column :fact_values, :origin, :string, :limit => 255, :default => 'N/A'
```
ares: why is it necessary to introduce another column, FactName has STI which determines the fact origin

app/helpers/fact_values_helper.rb
dLobatog: :smile: .. I guess I was tired

db/migrate/20160921160155_add_origin_to_fact_values.rb
dLobatog: Good call, I created this column because I assumed we'd have problems distinguishing fact names with the same name for distinct config management systems. I'll make `FactName::Puppet` in this PR too and remove the migrations

ares: I didn't dare to ask, but `FactName::Puppet` would be awesome!

dLobatog:
I've updated this so that:
- it uses the FactName to figure out the value origin
- it does not require any migrations 
- the 'origin' column size matches the size of the icons
- the icons match the recommended Patternfly size for icons (same as the font-awesome icons)

screenshot from 2016-09-22 10-30-38

If you're wondering how it looks like, I would recommend changing the default of 'N/A' to 'Puppet'. Updating the plugins to support this is quite trivial as you can see in https://github.com/dLobatog/foreman_ansible/commit/87b9fbc523a5ae061e9d24888a4f503fed27df3f

cc @ares thanks for reviewing this

domcleal:
Please ensure app/assets/images/Puppet.png is under GPLv3+ (unlikely), or add its licensing terms to LICENSE if it can be freely distributed and used here, else I'd suggest using text or a replacement icon.

dLobatog:
I took it from their press kit - it doesn't mention anything about fair use unfortunately so I'll just shoot an email to their PR.

dLobatog:
https://puppet.app.box.com/s/eihp5xv6jp9j38f9x3wivycgtdu7o1o9 - 

"Puppet Logo: Written permission is always required. Contact us by email at trademarks@puppet.com."

Aight, I'll substitute it with one of @lzap 's 'letter logos' until we get permission :smile:

app/services/puppet_fact_importer.rb
```diff
@@ -1,9 +1,9 @@
 class PuppetFactImporter < StructuredFactImporter
   def self.authorized_smart_proxy_features
-    'Puppet'
+    'Puppet'.freeze
   end
 
   def fact_name_class
-    ::FactName
+    PuppetFactName
```
ares, line 7: should we have a migration for existing facts?

dLobatog: How would you figure out where they come from?

ares: is there a fact in any plugin that does not use it's own class? I think all existing fact name with type `::FactName` should become `PuppetFactName`

ares:
Tested, everything works nicely. Except the migration I think it's good to go.

app/services/puppet_fact_importer.rb
dLobatog: Facts coming from Discovery are FactNames, but I suppose I can say they come from Puppet... after all it's facter who provides them in the image

dLobatog:
@ares Updated, all FactNames that are plain FactNames (I check the type attribute for that) are now PuppetFactName. This includes discovery fact names that the user may have provided but I guess that's fine.

dLobatog:
https://github.com/theforeman/foreman_ansible/pull/50

db/migrate/20160922144222_change_fact_name_to_puppet_fact_name.rb
```diff
@@ -0,0 +1,13 @@
+class ChangeFactNameToPuppetFactName < ActiveRecord::Migration
+  def up
+    FactName.select { |f| f.type == 'FactName' }.each do |fact|
```
ares: This loads all fact names in memory (18328 in my case) and then only select among instantiated objects, why not using `FactName.where(:type => 'FactName')`? or even better using `FactName.update_all` to make just one query skipping validations?

db/migrate/20160922144222_change_fact_name_to_puppet_fact_name.rb
```diff
@@ -0,0 +1,13 @@
...
+      fact.update_attribute(:type, 'PuppetFactName')
+    end
+  end
+
+  def down
+    PuppetFactName.each do |puppetfact|
```
ares: same as above

ares:
Could you please rebase too?

db/migrate/20160922144222_change_fact_name_to_puppet_fact_name.rb
dLobatog: `FactName.update_all` retrieves `ForemanAnsible::FactName` too, so that can't  be done, but I can chain where and update_all

dLobatog:
Rebased and now the migrations chain `where` and `update_all` instead of iterating over all objects (thanks for the catch @ares )

db/migrate/20160922144222_change_fact_name_to_puppet_fact_name.rb
ares: sure, that's what I meant :+1:

ares:
Thanks @dLobatog, merged as f20020ce457d768c5418141ec367e2ee1dd3b9b9

dLobatog:
2 months later, no word from Puppetlabs yet :/

PR #3878, opened 2016-09-22
Fixes #16652 - UEFI Grub2 support for non-intel archs

lzap:
This small patch should enable GRUB2 UEFI booting of:
- ARM 64
- IBM POWER
- IBM POWER Little Endian

### DIFF
--- a/test/unit/architecture_test.rb
+++ b/test/unit/architecture_test.rb
@@ -25,28 +25,53 @@ class ArchitectureTest < ActiveSupport::TestCase
     assert_not architecture.destroy
   end
 
-  test "should return intel precision for i386" do
+  test "should return EFI filename for i386" do
     architecture = Architecture.new :name => "i386"
-    assert_equal "ia32", architecture.intel_precision
+    assert_equal "ia32", architecture.bootfilename_efi
   end
 
-  test "should return intel precision for i686" do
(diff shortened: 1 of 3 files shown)


test/unit/architecture_test.rb
```diff
@@ -25,28 +25,53 @@ class ArchitectureTest < ActiveSupport::TestCase
...
     architecture = Architecture.new :name => "x86-64"
-    assert_equal "x64", architecture.intel_precision
+    assert_equal "x64", architecture.bootfilename_efi
   end
 
-  test "should return intel precision for x86_64" do
+  test "should return EFI filename for x86_64" do
     architecture = Architecture.new :name => "x86_64"
-    assert_equal "x64", architecture.intel_precision
+    assert_equal "x64", architecture.bootfilename_efi
```
dLobatog, line 45: took me a split second to not raise an eyebrow and think this and the previous test were the same :)

dLobatog:
Merged as 4f5adf6525dfca9f80623bdd1c925cd336dd1639, thanks @lzap!

test/unit/architecture_test.rb
lzap: Yeah it was a bad name for sure.

PR #3881, opened 2016-09-24
Fixes #17044 - foreman-debug counts size instead lines

pmoravec:
…ad of tailing all (log)files

Signed-off-by: Pavel Moravec pmoravec@redhat.com

### DIFF
--- a/script/foreman-debug
+++ b/script/foreman-debug
@@ -22,10 +22,11 @@ OPTIONS:
   -d DIR  Directory to place the tarball in (default /tmp/foreman-XYZ)
   -g      Skip generic info (CPU, memory, firewall etc.)
   -a      Do not generate a tarball from the resulting directory
-  -m NUM  Maximum lines to keep for each file (default 5000), zero means no limit
+  -m NUM  Maximum lines to keep for each file, deprecated, use -s
   -j PRG  Filter with provided program when creating a tarball
   -p      Additionally print password patterns being filtered out
   -q      Quiet mode
+  -s SIZE Maximum log set size in MB (current+rotated files, defaults to 10 MB)
   -v      Verbose mode
   -u      Upload tarball$UPLOAD_INFO
   -h      Shows this message
@@ -79,41 +80,51 @@ add_cmd() {
   eval $CMD >> "$DIR/$OUT" 2>&1
 }
 
-# add and filter if it is a non zero, readable, regular file or symlink (skip otherwise)
-add_files() {
-  if [ $MAXLINES -eq 0 ]; then
-    TAIL_OR_CAT="cat"
-  else
-    TAIL_OR_CAT="tail -n ${MAXLINES}"
-  fi
-  for FILE in $*; do
-    if [ \( -f "$FILE" -o -h "$FILE" \) -a \( -r "$FILE" -a -s "$FILE" \) ]; then
-      printv " - $FILE"
-      SUBDIR=$(dirname $FILE)
-      [ ! -d "$DIR$SUBDIR" ] && mkdir -p "$DIR$SUBDIR"
-      MIME=$(file -bi "$FILE" | cut -d\; -f1)
-      case $MIME in
+# add and filter file of known MIME type, calculate size
+add_file() {
+  FILE=$1
+  SIZE=$(stat -c "%s" $FILE)
+  MIME=$(file -bi "$FILE" | cut -d\; -f1)
+  case $MIME in
         application/x-gzip)
-          zcat "$FILE" | $TAIL_OR_CAT | sed -r "$FILTER" > "$DIR$FILE.txt"
+          zcat "$FILE" | sed -r "$FILTER" > "$DIR$FILE.txt"
           ;;
         application/x-bzip2)
-          bzcat "$FILE" | $TAIL_OR_CAT | sed -r "$FILTER" > "$DIR$FILE.txt"
+          bzcat "$FILE" | sed -r "$FILTER" > "$DIR$FILE.txt"
           ;;
         application/x-xz)
-          xzcat "$FILE" | $TAIL_OR_CAT | sed -r "$FILTER" > "$DIR$FILE.txt"
+          xzcat "$FILE" | sed -r "$FILTER" > "$DIR$FILE.txt"
           ;;
         text/plain | application/xml)
-          $TAIL_OR_CAT "$FILE" | sed -r "$FILTER" > "$DIR$FILE"
+          sed -r "$FILTER" "$FILE" > "$DIR$FILE"
           [ $PRINTPASS -eq 1 ] && grep -H "\*\*\*\*\*" "$DIR$FILE"
-          ;;
-        inode/symlink)
-          cp -a "$FILE" "$DIR$FILE"
-          add_files $(readlink -f $FILE)
           ;;
         *)
           echo "Skipping file $FILE: unknown MIME type $MIME" >> "$DIR/skipped_files"
+          SIZE=0 # don't count the size to collected files
           ;;
       esac
+}
+
+# add files (from newest to oldest) that are non zero, readable, regular file or symlink, until $MAXSIZE limit is reached
+add_files() {
+  SUMSIZE=0
+  # sort regular+symlink readable nonempty files per modification time, newest first - assuming no space in a filename
+  for FILE in $(find -L $* -type f -readable -size +0b -printf "%T@ %p\n" 2> /dev/null | sort -nr | cut -d' ' -f2 2> /dev/null); do
+    # if we are over size limit, skip rest older files
+    if [ \( $SUMSIZE -gt $MAXSIZE \) -a \( $MAXSIZE -gt 0 \) ]; then
+      printv " - $FILE (skipped due to size)"
+    else
+      printv " - $FILE"
+      SUBDIR=$(dirname $FILE)
+      [ ! -d "$DIR$SUBDIR" ] && mkdir -p "$DIR$SUBDIR"
+      # if the file is symlink, copy source and add target
+      if [ -h $FILE ]; then
+        cp -a "$FILE" "$DIR$FILE"
+        FILE=$(readlink -f $FILE)
+      fi
+      add_file $FILE
+      SUMSIZE=$(($((SUMSIZE))+$((SIZE))))
     fi
   done
 }
@@ -122,7 +133,7 @@ add_files() {
 DIR=""
 NOGENERIC=0
 NOTAR=0
-MAXLINES=5000
+MAXSIZE=10485760   # 10 MB in bytes
 COMPRESS=""
 PRINTPASS=0
 QUIET=0
@@ -149,7 +160,7 @@ fi
 CONF_FILE=/usr/share/foreman/config/foreman-debug.conf
 test -f $CONF_FILE && source $CONF_FILE
 
-while getopts "d:gam:j:uqpvhx" opt; do
+while getopts "d:gam:s:j:uqpvhx" opt; do
   case $opt in
     d)
       DIR="$OPTARG"
@@ -170,11 +181,16 @@ while getopts "d:gam:j:uqpvhx" opt; do
       VERBOSE=1
       ;;
     m)
-      MAXLINES="$OPTARG"
+      error "Warning: -m option is deprecated, use -s option"
       ;;
     j)
       COMPRESS="$OPTARG"
       EXTENSION=".$(echo "$OPTARG" | awk '{ print $1 }')"
+      ;;
+    s)
+      #read the value and convert from MB to bytes
+      MAXSIZE="$OPTARG"
+      MAXSIZE=$((MAXSIZE*1024*1024))
       ;;
     u)
       UPLOAD=1
@@ -259,7 +275,9 @@ if [ $NOGENERIC -eq 0 ]; then
     add_cmd "scl -l" "software_collections"
 
   add_cmd "ps auxwwwZ" "process_list"
-  add_files /var/log/messages /var/log/audit/audit.log /var/log/syslog
+  add_files /var/log/messages*
+  add_files /var/log/audit/audit.log*
+  add_files /var/log/syslog*
   add_cmd "ausearch -m AVC -m USER_AVC -m SELINUX_ERR | head -n 100" "selinux_first_denials.log"
   add_cmd "ausearch -m AVC -m USER_AVC -m SELINUX_ERR || grep AVC /var/log/audit/audit.log" "selinux_denials.log"
   if [ -f /usr/sbin/selinuxenabled ] && /usr/sbin/selinuxenabled; then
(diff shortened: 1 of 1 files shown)


domcleal:
Administrative note: this ticket is not from the Foreman project, it's in Katello. Please file a ticket describing the changes in Foreman.

script/foreman-debug
```diff
@@ -79,50 +79,79 @@ add_cmd() {
...
+        add_file $FILE
+        SUMSIZE=$(($((SUMSIZE))+$((SIZE))))
+        # if we are over the size limit, skip rest older files
+        if [ \( $SUMSIZE -gt $MAXSIZE \) -a \( $MAXSIZE -gt 0 \) ]; then
+          SKIP=" (skipped due to size)"
+        fi
+      fi
     fi
   done
 }
```
lzap, line 130: So this function adds all files up to maximum size of 5000 bytes by default, that's 5k. Looks like a bad default?!

Also why are we counting the maximum for each individual call? Isn't making a "global" maximum size a better option?

script/foreman-debug
```diff
@@ -79,50 +79,79 @@ add_cmd() {
...
           ;;
       esac
+}
+
+# add files (from newest to oldest) that are non zero, readable, regular file or symlink, until $MAXSIZE limit is reached
+add_files() {
+  SUMSIZE=0
+  SKIP="" # will be nonempty to stop adding files
+  # sort files per modification time, newest first
+  for FILE in $(ls -t $* 2> /dev/null); do
```
lzap: This won't work if you enter two different directories, e.g. add_files "/var/log/one" "/var/log/two".

Consider using GNU find, you can print last modification time and filename as two columns, then sort it and process line by line. With GNU find you can also limit results (to regular files, symlinks etc). And you can also add additional column with file size so you have all the information needed at once.

script/foreman-debug
```diff
@@ -79,50 +79,79 @@ add_cmd() {
...
+        # if the file is symlink, copy source and add target
+        if [ -h $FILE ]; then
+          cp -a "$FILE" "$DIR$FILE"
+          FILE=$(readlink -f $FILE)
+        fi
+        add_file $FILE
+        SUMSIZE=$(($((SUMSIZE))+$((SIZE))))
+        # if we are over the size limit, skip rest older files
+        if [ \( $SUMSIZE -gt $MAXSIZE \) -a \( $MAXSIZE -gt 0 \) ]; then
+          SKIP=" (skipped due to size)"
```
lzap: This is little bit too much, I'd simply call printv twice, once with successful branch and once in this code flow branch.

script/foreman-debug
```diff
@@ -79,50 +79,79 @@ add_cmd() {
...
+# etc.
+# be aware, 1st argument is regexp, so use "[.]" instead of "." for dot char
+add_files_types() {
+  SEPARATOR="$1"
+  shift
+  FILES="$*"
+  for FILETYPE in $(ls $FILES | awk -F "$SEPARATOR" -v sep="$SEPARATOR" '{ print $1 sep }' | sort -u); do
+    add_files "${FILETYPE}*"
+  done
+}
```
lzap: Again, I'd strongly prefer GNU find. More importantly, I don't like using regexp patterns, globs are very expected in this case. Find can do the processing for you (it can do globs). Do you have a use case where you need to have regexp instead of a glob?

script/foreman-debug
```diff
@@ -79,50 +79,79 @@ add_cmd() {
...
+
 # default values
 DIR=""
 NOGENERIC=0
 NOTAR=0
-MAXLINES=5000
+MAXSIZE=5000
```
lzap: Assuming these are bytes that's too little.

script/foreman-debug
```diff
@@ -22,7 +22,7 @@ OPTIONS:
   -d DIR  Directory to place the tarball in (default /tmp/foreman-XYZ)
   -g      Skip generic info (CPU, memory, firewall etc.)
   -a      Do not generate a tarball from the resulting directory
-  -m NUM  Maximum lines to keep for each file (default 5000), zero means no limit
+  -m NUM  Maximum size in kB for each file type (default 5000 kB), zero means no limit
```
lzap: Oh I noticed this is kB, therefore 5 MB per log file or directory provided? I think we can go higher here. Ten to twenty would be reasonable I think.

lzap (changes requested):
Overall, I like the change, thank you. I have some notes in regard to the implementation.

lzap:
Also rename the commit message to something shorter please.

script/foreman-debug
pmoravec: It stops adding files just after reaching 5MB. This function is assumed to be called for one type of logs (i.e. /var/log/messages\* or /var/log/foreman/production\* or so).

If you need "to collect logs of multiple types (i.e. from /var/log/foreman, collect 5MB of production.log_, 5MB of apipe_cache.log_, 5MB of dynflow_executor.log\* etc.), let call:

```
add_files_types . "/var/log/foreman/*.log*"
```

This will 1) find all the files from 2nd argument, 2) call `add_files` for groups of files with common prefix (where 1st argument, the dot here, is the delimiter). Let add a debug just before

```
add_files "${FILETYPE}*"
```

line in the `add_files_types` method (or read the method description in comment before its body).

script/foreman-debug
pmoravec: OK, but find -type / -xtype assumes paths, not files, so the test

```
if [ \( -f "$FILE" -o -h "$FILE" \) -a \( -r "$FILE" -a -s "$FILE" \) ]; then
```

will remain..

script/foreman-debug
pmoravec: Yes, units are kB, default value 5000. I don't insist on either. Per my comparison of original and new foreman-debug, usually _more_ data is collected after the patch applied. Since originally we collected 5k lines (say 500kB) of each file, now around 5MB per each file type - worth comparing tarball of previous and new foreman-debug.

In my opinion, the parameter shall be lower for some log files (i.e. http error log or foreman cron, usually not big) and higher for others (production.log, usually few MBs per day, worth having few days of history). But I dont see an easy and understandable way how to achieve this.

pmoravec: Btw. the current method has two maybe unusual aspects:
- it stops adding files after reaching the limit. So if two newest files are 1MB and 10MB, both complete files are added. This makes sense, the files are usually daily rotated and it is worth having logs from whole day when some extra log events happened - not only tail of that events
- it calculates sizes of unpacked _.gz files. While gunzip (to apply password obfuscation) causes the foreman-debug-XXXX directory will have *bigger_ sizes. I see this as acceptable since xzip will shrink the sizes back.

pmoravec:
I updated the PR by incorporating the changes. About the default size limit of logs/configs to collect: First recall the limit is calculated per log type - i.e. 5MB per production.log, 5MB per apipe_cache.log, 5MB per httpd error log, 5MB per httpd foreman access SSL log, etc.

Some comparison of foreman-debug time execution and collected data (from my Satellite6.2, just one particular instance): columns stand for:

-m value, time of execution, tar.xz size in kB, unpacked size in kB
100 5m54.961s 6552 375452
500 5m38.980s 6724 383112
1000 4m35.432s 5624 388236
2000 6m53.514s 8596 491156
3000 6m33.501s 7396 500956
4000 6m48.183s 8604 511460
5000 4m19.430s 7464 518776
6000 7m6.514s 8708 532760
7000 6m42.826s 8836 550428
8000 7m52.490s 9776 989696
9000 6m58.363s 9896 998692
10000 8m46.581s 11208 1007076

(the times are quite non-deterministic..)

And original pre-patch value:
5000lines 4m43.564s 3228 72516

So the size is 2-3times bigger for 5MB default size.

lzap (approved):

script/foreman-debug
```diff
@@ -79,50 +79,77 @@ add_cmd() {
...
+        if [ -h $FILE ]; then
+          cp -a "$FILE" "$DIR$FILE"
+          FILE=$(readlink -f $FILE)
+        fi
+        add_file $FILE
+        SUMSIZE=$(($((SUMSIZE))+$((SIZE))))
+      fi
     fi
   done
 }
```
lzap, line 130: This still needs to be rewritten using GNU find and sort using `-type` options.

script/foreman-debug
```diff
@@ -79,50 +79,77 @@ add_cmd() {
...
+# etc.
+add_files_types() {
+  SEPARATOR="$1"
+  shift
+  FILES="$*"
+  for FILETYPE in $(find -L $FILES 2> /dev/null | awk -F "$SEPARATOR" '{ print $1 }' | sort -u); do
+    add_files "${FILETYPE}${SEPARATOR}*"
+  done
+}
+
```
lzap: Please remove this function, this is utterly complex. Just use add_files with shell blog for each individual 5MB bundle you want to add. Maybe add a comment below and into katello script but it will be quite obvious and cleaner.

lzap (changes requested):

script/foreman-debug
```diff
@@ -22,10 +22,11 @@ OPTIONS:
   -d DIR  Directory to place the tarball in (default /tmp/foreman-XYZ)
   -g      Skip generic info (CPU, memory, firewall etc.)
   -a      Do not generate a tarball from the resulting directory
-  -m NUM  Maximum lines to keep for each file (default 5000), zero means no limit
+  -m NUM  Maximum lines to keep for each file, deprecated, use -s
   -j PRG  Filter with provided program when creating a tarball
   -p      Additionally print password patterns being filtered out
   -q      Quiet mode
+  -s SIZE Stop collecting given type of logs after reaching SIZE kB, default 10000
```
lzap: I'd make this more clear (and slightly less accurate for the sake of readability):

```
Maximum log set size in kB (current and rotated files, defaults to 10 MB)
```

script/foreman-debug
```diff
@@ -79,41 +80,52 @@ add_cmd() {
...
           echo "Skipping file $FILE: unknown MIME type $MIME" >> "$DIR/skipped_files"
+          SIZE=0 # don't count the size to collected files
           ;;
       esac
+}
+
+# add files (from newest to oldest) that are non zero, readable, regular file or symlink, until $MAXSIZE limit is reached
+add_files() {
+  SUMSIZE=0
+  SKIP="" # will be nonempty to stop adding files
```
lzap: This one is unused, I believe.

script/foreman-debug
```diff
@@ -79,41 +80,52 @@ add_cmd() {
...
+  # sort regular+symlink readable nonempty files per modification time, newest first - assuming no space in a filename
+  for FILE in $(find -L $* -type f -readable -size +0b -printf "%T@ %p\n" 2> /dev/null | sort -nr | cut -d' ' -f2 2> /dev/null); do
+    # if we are over size limit, skip rest older files
+    if [ \( $SUMSIZE -gt $MAXSIZE \) -a \( $MAXSIZE -gt 0 \) ]; then
```
lzap, line 115: I think the `\( $MAXSIZE -gt 0 \)` test should be done outside of the loop, perhaps just make sure the user entered a positive regular number and that's it.

script/foreman-debug
```diff
@@ -170,12 +182,15 @@ while getopts "d:gam:j:uqpvhx" opt; do
       VERBOSE=1
       ;;
     m)
-      MAXLINES="$OPTARG"
+      error "Warning: -m option is deprecated, see -s option"
```
lzap: `use -s option`

script/foreman-debug
```diff
@@ -287,8 +304,19 @@ add_cmd "gem list" "gem_list"
...
 add_cmd "bundle --local --gemfile=/usr/share/foreman/Gemfile" "bundle_list"
 add_cmd "facter" "facts"
-add_files /etc/foreman/* /var/log/foreman/*.log*
-add_files /etc/foreman-installer/scenarios.d/{*,*/*,*/.*} /var/log/foreman-installer/*.log*
+
+add_files /var/log/foreman/apipie_cache*.log*
+add_files /var/log/foreman/cron*.log*
+add_files /var/log/foreman/db_migrate*.log*
+add_files /var/log/foreman/db_seed*.log*
+add_files /var/log/foreman/dynflow_executor*.log*
```
lzap: Please move dynflow into katello foreman debug addon-script.

lzap (changes requested):
Overall it looks good, couple of minor comments. Almost there!

pmoravec:
Incorporated all changes except one: `\( $MAXSIZE -gt 0 \)` on that place allows -s 0 to collect all logs, so I didn't remove it.

script/foreman-debug
```diff
@@ -79,41 +80,51 @@ add_cmd() {
...
-    if [ \( -f "$FILE" -o -h "$FILE" \) -a \( -r "$FILE" -a -s "$FILE" \) ]; then
-      printv " - $FILE"
-      SUBDIR=$(dirname $FILE)
-      [ ! -d "$DIR$SUBDIR" ] && mkdir -p "$DIR$SUBDIR"
-      MIME=$(file -bi "$FILE" | cut -d\; -f1)
-      case $MIME in
+# add and filter file of known MIME type, calculate size
+add_file() {
+  FILE=$1
+  SIZE=$(du -k $FILE | cut -f1)
```
lzap: Nitpick: `stat` can return just the number, but this works too.

script/foreman-debug
```diff
@@ -22,10 +22,11 @@ OPTIONS:
   -d DIR  Directory to place the tarball in (default /tmp/foreman-XYZ)
   -g      Skip generic info (CPU, memory, firewall etc.)
   -a      Do not generate a tarball from the resulting directory
-  -m NUM  Maximum lines to keep for each file (default 5000), zero means no limit
+  -m NUM  Maximum lines to keep for each file, deprecated, use -s
   -j PRG  Filter with provided program when creating a tarball
   -p      Additionally print password patterns being filtered out
   -q      Quiet mode
+  -s SIZE Maximum log set size (current and rotated files, defaults to 10000 kB)
```
lzap: Oh please reword this to be sure user will get the unit is actually kB:

```
-s SIZE Maximum log set size in kB (current and rotated files, defaults to 10 MB)
```

script/foreman-debug
```diff
@@ -259,7 +273,9 @@ if [ $NOGENERIC -eq 0 ]; then
     add_cmd "scl -l" "software_collections"
 
   add_cmd "ps auxwwwZ" "process_list"
-  add_files /var/log/messages /var/log/audit/audit.log /var/log/syslog
+  add_files /var/log/messages
+  add_files /var/log/audit/audit.log
+  add_files /var/log/syslog
```
lzap: Can we include rotated files of these as well?

lzap (changes requested):
Ok small changes and I am good to start testing this! Well done.

pmoravec:
Updated, also `du` -> `stat`. That caused I count sizes in bytes and not in kB so I changed `MAXSIZE` default value and its assignment in `getopts`

lzap:
Works fine thanks. Please ping me on in the other PRs, at least foreman-proxy must be modified in order to get the new feature.

PR #3882, opened 2016-09-24
Fixes #16681 - error creating config_group ActiveModel::ForbiddenAttr

ranjan:


domcleal:
Please address the feedback from @theforeman-bot. Rather than state an error message, the commit message should state what's changed, e.g. `use strong params in config groups API`.

domcleal:
If you change a pull request in future, please add a comment so the reviewer knows that it's been updated and what's changed.

domcleal:
I think this will also require `wrap_parameters` to be defined in the controller using the accessible_attributes method from the parameters helper.

With this PR, creating a config group with an _unwrapped_ hash containing puppetclass_ids/names fails to associate any classes. This would have worked prior to strong parameters as both associations were listed as accessible attributes.

ranjan:
@domcleal Creating config group with puppet_classes is working as expected. Added test which includes  puppet_classes.

domcleal:
This doesn't appear to fix the issue I described with _unwrapped_ requests - it will need a wrap_parameters statement to fix I think. Your test only tests a wrapped request.

ranjan:
Thanks @domcleal,  got the difference between wrap and unwrapped parameter :). Updated PR

domcleal:
Merged, thanks @ranjan.

PR #3884, opened 2016-09-25
Fixes #12111 - Set width for host list action column

tbrisker:
This makes no siginificant change to core but allows unbreaking
foreman_column_view plugin.

### DIFF
--- a/app/views/hosts/_list.html.erb
+++ b/app/views/hosts/_list.html.erb
@@ -10,7 +10,7 @@
       <th class="hidden-tablet hidden-xs" width="10%"><%= sort :model, :as => _('Model') %></th>
       <th class="hidden-tablet hidden-xs" width="15%"><%= sort :hostgroup, :as => _("Host group") %></th>
       <th class="hidden-tablet hidden-xs" width="10%"><%= sort :last_report, :as => _('Last report'), :default => 'DESC' %></th>
-      <th ><%= _('Actions') %></th>
+      <th width="100px"><%= _('Actions') %></th>
     </tr>
   </thead>
   <tbody>


dLobatog:
@tbrisker I'm getting `ActionView::Template::Error (couldn't find file 'bootstrap' with type 'application/javascript'` when trying the column_view PR , any idea what's wrong with it? I don't think it's a local setup issue because when I remove the plugin from my Gemfile I don't see the issue anymore.

app/views/hosts/_list.html.erb
```diff
@@ -10,7 +10,7 @@
       <th class="hidden-tablet hidden-xs" width="10%"><%= sort :model, :as => _('Model') %></th>
       <th class="hidden-tablet hidden-xs" width="15%"><%= sort :hostgroup, :as => _("Host group") %></th>
       <th class="hidden-tablet hidden-xs" width="10%"><%= sort :last_report, :as => _('Last report'), :default => 'DESC' %></th>
-      <th ><%= _('Actions') %></th>
+      <th width="100px"><%= _('Actions') %></th>
```
dLobatog, line 13: What you want to accomplish by using fixed-size width is that other buttons never cause this one to overflow, right?

tbrisker: That is correct.

tbrisker:
@dLobatog i'm not quite sure where you would be getting that, there is no mention of bootstrap anywhere in the column view plugin. Is it possible you may have something in your `settings.yaml` that causes it? or maybe an older version of the plugin? it seems to work just fine for me with a git checkout.

dLobatog:
Just saw what caused it, it's https://github.com/GregSutcliffe/foreman_column_view/pull/23 . This fixes the issue in the plugin and does not cause problems here so :+1:

dLobatog:
Thanks  @tbrisker !

PR #3885, opened 2016-09-25
Refs #16599 - Specify npm 3.x requirement in package.json

tbrisker:
npm >= 3 is required for npm-shrinkwrap to function correctly.


tbrisker:
Looks like engines is only checked when installing as a package, not when running npm install, closing as no one installs foreman as a package.

PR #3886, opened 2016-09-25
Fixes #15285 - save browser timezone in User

shlomizadok:
Not sure about this pull request:
Currently when a user chooses "Browser timezone" it is set to nil in the User model. And because it is empty we set it to the [cookie's timezone](https://github.com/theforeman/foreman/blob/2aa15bf1f40fff77a50bb9907fa993e067dd6346/app/controllers/concerns/application_shared.rb#L10-L18).
The change in this pull request will enforce that the cookie's timezone will be saved for the user.

Would love to hear your comments

### DIFF
--- a/app/controllers/concerns/foreman/controller/users_mixin.rb
+++ b/app/controllers/concerns/foreman/controller/users_mixin.rb
@@ -31,6 +31,12 @@ module Foreman::Controller::UsersMixin
     sub_hg.each { |hg| hg.users << @user }
   end
 
+  def update_timezone
+    if @user.timezone.blank? && cookies[:timezone]
+      @user.update(timezone: cookies[:timezone])
+    end
+  end
+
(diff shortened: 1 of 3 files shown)


dLobatog:
@shlomizadok Thanks for the effort but `app/controllers/concerns/application_shared.rb` does the job already, why do you want to save it? It's set on every request and it allows us to not have to take care of yet another property on user. You wrote that code so I guess you have reasons to want to save it..?

shlomizadok:
I am conflicted to be honest.
On one hand, as you have said: We don't really need it `application_shared.rb` is handling it correctly.
On the other hand, if a user picks the "Browser timezone", it is never really saved anywhere.

I'll let you decide :smile: - I'm good either way

dLobatog:
I prefer to close it - it means zero extra work & risk for a feature that already works pretty well (which you yourself did :smile: )

PR #3887, opened 2016-09-25
Fixes #16646 - Add ability to plugins to modify index scope

ShimShtein:
Added the ability for plugins to modify the scope used to get hosts for `#index` action. This is needed if a plugin wants to add more `includes` or `eager_load` statements, so Active Record won't try to lazy load additional relations and create N+1 queries.

### DIFF
--- a/app/controllers/concerns/scopes_per_action.rb
+++ b/app/controllers/concerns/scopes_per_action.rb
@@ -1,0 +1,48 @@
+# This module will hold a registry of scope directives that should be
+# part of the query per specific action in a controller. This is required by
+# plugins that add *_to_many relations to foreman core models like
+# `Host::Managed` and want to add some includes or eager_load directives to
+# the query.
+
+module ScopesPerAction
+  extend ActiveSupport::Concern
+
+  # returns a scope that includes all directives registered via add_scope_for
+  def action_scope_for(action, base_scope)
+    scope = base_scope
+    self.class.scopes_for(action).each do |scope_func|
+      scope = scope_func.call(scope) || scope
+    end
+    scope
+  end
+
+  module ClassMethods
+    # Add a new scope directive to action specified by action parameter using
+    # block statement. It will receive the base scope as a first parameter.
+    # example: add_scope_for(:my_action) { |base_scope| base_scope.includes(:my_new_table) }
+    def add_scope_for(action, &block)
+      local_scopes_for(action) << block
+    end
+
+    def scopes_for(action)
+      all_scopes = scopes_per_action.merge(scopes_per_action_from_plugins) { |k, left_v, right_v| left_v + right_v }
+      all_scopes[action] || []
+    end
+
+    def local_scopes_for(action)
+      scopes_per_action[action] ||= []
+    end
+
+    private
+
+    def scopes_per_action
+      @scopes_per_action ||= {}
+    end
+
+    def scopes_per_action_from_plugins
+      Foreman::Plugin.all.map {|plugin| plugin.action_scopes_hash_for(self) }.inject({}) do |memo, actions_hash|
+        memo.merge(actions_hash) { |k, left_v, right_v| left_v + right_v }
+      end
+    end
+  end
+end
--- a/app/services/foreman/plugin.rb
+++ b/app/services/foreman/plugin.rb
@@ -105,6 +105,7 @@ module Foreman #:nodoc:
       @parameter_filters = {}
       @smart_proxies = {}
       @permissions = {}
+      @controller_action_scopes = {}
     end
 
     def after_initialize
@@ -368,6 +369,18 @@ module Foreman #:nodoc:
 
     def smart_proxies(klass)
       @smart_proxies.fetch(klass.name, {})
+    end
+
+    def add_controller_action_scope(controller_class, action, &block)
+      controller_actions = @controller_action_scopes[controller_class.name] || {}
+      actions_list = controller_actions[action] || []
+      actions_list << block
+      controller_actions[action] = actions_list
+      @controller_action_scopes[controller_class.name] = controller_actions
+    end
+
+    def action_scopes_hash_for(controller_class)
+      @controller_action_scopes[controller_class.name] || {}
     end
 
     private
--- a/test/unit/plugin_test.rb
+++ b/test/unit/plugin_test.rb
@@ -383,6 +383,27 @@ class PluginTest < ActiveSupport::TestCase
     assert_equal({:foo => {:feature => 'Foo'}}, Foreman::Plugin.find(:test_smart_proxy).smart_proxies(Awesome))
   end
 
+  def test_hosts_controller_action_scope
+    mock_scope = ->(scope) { scope }
+    Foreman::Plugin.register :test_hosts_controller_action_scope do
+      add_controller_action_scope HostsController, :test_action, &mock_scope
+    end
+    scopes = HostsController.scopes_for(:test_action)
+    assert_equal mock_scope, scopes.last
+  end
+
+  def test_hosts_controller_action_scope_added_to_local
+    mock_scope = ->(scope) { scope }
+    HostsController.add_scope_for(:test_action) do |scope|
+      scope
+    end
+    Foreman::Plugin.register :test_hosts_controller_action_scope_added_to_local do
+      add_controller_action_scope HostsController, :test_action, &mock_scope
+    end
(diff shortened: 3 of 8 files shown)


app/controllers/concerns/scopes_per_action.rb
```diff
@@ -0,0 +1,37 @@
+# This module will hold a regstry of scope directives that should be
```
timogoebel: nitpick: `regstry´should be`registry´

ShimShtein: Done

app/controllers/concerns/scopes_per_action.rb
```diff
@@ -0,0 +1,37 @@
...
+# the query.
+
+module ScopesPerAction
+  extend ActiveSupport::Concern
+
+  # returns a scope that includes all directives registered via add_scope_for
+  def action_scope_for(action, base_scope)
+    scope = base_scope
+    self.class.scopes_for(action).each do |scope_func|
+      scope = scope_func.call(scope) || scope
```
ohadlevy, line 14: instead of looping them all, you can return the first match? or do you rather want to match the last one? the code is a bit hard to read.

app/services/foreman/plugin.rb
```diff
@@ -351,6 +351,12 @@ def parameter_filters(klass)
       @parameter_filters.fetch(klass.is_a?(Class) ? klass.name : klass, [])
     end
 
+    def hosts_controller_action_scope(action, &block)
+      # Add scope for both API and UI controller.
+      HostsController.add_scope_for(action, block)
```
ohadlevy: does it make sense to always assume you want to extend them both?

app/controllers/concerns/scopes_per_action.rb
ShimShtein: I am chaining the scopes, here is an example:
In puppet plugin:

``` ruby
HostsController.add_scope_for(:index) { |base_scope| base_scope.eager_load(:environment) }
```

In provisioning plugin:

``` ruby
HostsController.add_scope_for(:index) { |base_scope| base_scope.eager_load(:operatingsystem) }
```

I want to create a scope that will contain both operations:

``` ruby
base_scope = Hosts.authorized
actual_scope = base_scope.eager_load(environment).eager_load(operatingsystem)
```

I achieve this by calling each block, and passing it the result of the previous block.

app/services/foreman/plugin.rb
ShimShtein: It sounds reasonable to me that the info that is returned is similar between the API and UI. In the original bug in Katello I saw that the problem exists in both UI and API.
I won't fight over it though. If you think it's unlikely, I can separate it to different methods.

app/services/foreman/plugin.rb
```diff
@@ -351,6 +351,12 @@ def parameter_filters(klass)
       @parameter_filters.fetch(klass.is_a?(Class) ? klass.name : klass, [])
     end
 
+    def hosts_controller_action_scope(action, &block)
```
timogoebel: Could you please add a test for this? Extra points for inline usage example.

app/services/foreman/plugin.rb
timogoebel: I think, you need to pass the block like so (note the extra ampersand):

``` ruby
HostsController.add_scope_for(action, &block)
```

Same below.

timogoebel (changes requested):
Looks good to me all in all. I like the DSL and this feature is long past due. The Plugin method needs test coverage (and fixing), though.

ShimShtein:
Rebased, fixed plugin registration and added tests.

@timogoebel did I get the extra points?
Is it clear how to use it?

ShimShtein:
Rebased again

timogoebel (changes requested):
The code looks good, thanks for the tests. It is also clear how to use this.
My only concern is, that this won't work in develop mode when classes can be reloaded. My suggestion is to store the scope codeblocks in the plugin model and when the concern is included have it get the scopes from the plugin model. You can check c6760930cf08a4b584b75df8a621092dd787da01 for an example.

ShimShtein:
> My only concern is, that this won't work in develop mode when classes can be reloaded. 

In the past we had `config.to_prepare` block that was run after reload. I can see that in Rails 5, this block [has moved](https://github.com/activeadmin/activeadmin/issues/4438) to `ActiveSupport::Reloader.to_prepare` just put those definitions in the to_prepare block and the callback will be registered again after reload.

timogoebel:
@ShimShtein : That would be possible, however it's not the Straight-A student's solution. The code in "to_prepare" runs on every development request. I think, it's better to implement a solution that just runs when a class is actually reloaded.

ShimShtein:
Now the list is global - no reloads should affect it.

lzap:
FYI there are conflicting files.

ShimShtein:
Rebased

test/unit/plugin_test.rb
```diff
@@ -383,6 +383,19 @@ def test_add_smart_proxy_for
     assert_equal({:foo => {:feature => 'Foo'}}, Foreman::Plugin.find(:test_smart_proxy).smart_proxies(Awesome))
   end
 
+  def test_hosts_controller_action_scope
+    mock_scope = ->(scope) { scope }
+    Foreman::Plugin.register :test_parameter_filter do
```
timogoebel: should be `test_controller_action_scope`

timogoebel (changes requested):
@ShimShtein : Thanks for the latest updates. Class reloading still beaks the scopes from Plugins.

Open hosts index page -> everyting is fine.

To reproduce:

``` bash
touch app/controllers/concerns/scopes_per_action.rb
```

Open host index page again -> it detects a n+1 query.

ShimShtein:
> touch app/controllers/concerns/scopes_per_action.rb

Indeed, reloading of this class will break the scopes. But that class'es sole responsibility is to maintain the registry, so if it changes - the old registry is probably not valid.
If I store this registry in the plugin class and it will get reloaded, it will be lost too.

timogoebel:
@ShimShtein : True. But if you store the data in
1. the host controller class
2. the plugin class

When these classes get reloaded, the code is parsed as well and the information is restored.

When you access the data, you can get it from both classes (and merge it) and you shouldn't have a problem. What do you think?

timogoebel:
@domcleal: May @ShimShtein and I ask for your opinion regarding where to store the data so that it survives a class reload in development? Is the current implementation fine, where the data from plugins gets lost when the `ScopesPerAction` concern is reloaded?

domcleal:
> @domcleal: May @ShimShtein and I ask for your opinion regarding where to store the data so that it survives a class reload in development?

It really should be stored in Foreman::Plugin (which is immune from reloads) when the API is called. Either access it from there when called, or something akin to c676093 where it's retrieved when the concern is included.

I'd also suggest not adding an API specifically for hosts, else it'd spawn a new API for every single model/controller collection, it ought to be more generic.

> Is the current implementation fine, where the data from plugins gets lost when the ScopesPerAction concern is reloaded?

Not really, IMHO. This is normally treated as a bug, and it ought to be fixed.

ShimShtein:
@domcleal First, thanks for the review. My only question is what to do with scopes that are not part of a particular plugin. For example `API::V2::HostsController` adds a default scope for index [here](https://github.com/theforeman/foreman/pull/3887/files#diff-6e62e9d287766aaa9c3d35e6dc3470a1R21). Of course I can add this scope manually in the method, but it sounds like I should really use the same framework here too.

ShimShtein:
@domcleal, Sorry, ignore my previous comment, I did not read the code thoroughly enough.

I have change the implementation to store scopes in both the plugin and in the controller (locally). The local storage is there for cases like the `API::V2::HostsController`.
I didn't cache the result of querying the plugins intentionally. It is done to avoid issues with construction of the controller class before all the plugins are loaded.

ShimShtein:
test failure is not related.

timogoebel (approved):

timogoebel:
Merged, thanks @ShimShtein. Please add a usage example to http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin.

ShimShtein:
@timogoebel Done: http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Modifying-controllers-query

PR #3888, opened 2016-09-26
refs #16557 - move templates helper test to correct dir

domcleal:
Missed this new file during a rebase.


dLobatog:
Merged - thanks @domcleal

PR #3890, opened 2016-09-26
Fixes #16614 - Humanize Audit class name

shlomizadok:


shlomizadok:
[test hound]

shlomizadok:
[test houndci] or [test hound] ?

shlomizadok:
[test] then ?

dLobatog:
@shlomizadok I think they're having internal problems, it's been failing a little bit today on other PRs.

dLobatog:
Thanks @shlomizadok ! 

PS I realized this needs translating right after clicking the button, will follow up this myself, don't worry

shlomizadok:
@dLobatog - Thanks. I do not think that a class name needs translation. It is translated at the [Permission model](https://github.com/theforeman/foreman/blob/develop/app/models/permission.rb#L15-L17)

PR #3891, opened 2016-09-26
Refs #16614 - Translate Audit string for filters edit page

dLobatog:
The humanized string should be a translatable string


shlomizadok:
@dLobatog - I do not think this is necessary [Permission model](https://github.com/theforeman/foreman/blob/develop/app/models/permission.rb#L15-L17) handles the translation

PR #3893, opened 2016-09-28
Fixes #16689 - DRY index action with search

shlomizadok:


timogoebel:
👍 from me on successful tests

PR #3894, opened 2016-09-28
Refs #16689 - Missing DRY index action controllers

dLobatog:
Some of the controllers that could have used the refactor in #16689 were
not added in that commit.

Additionally, I've fixed the '.includes' leftover in those controllers
from Rails 3 to be '.eager_load' now. '.includes' does not change the
query in _any_ way unless you call '.references' afterwards.


shlomizadok:
@dLobatog - Perfect. Thanks !

timogoebel:
@dLobatog : Thanks, even better 👍 

"Always leave the campground cleaner than you found it."

PR #3895, opened 2016-09-28
Fixes #13968 - Add error when no known proxy features found

shlomizadok:


app/models/smart_proxy.rb
```diff
@@ -115,16 +115,20 @@ def associate_features
...
       end
     rescue => e
       errors.add(:base, _('Unable to communicate with the proxy: %s') % e)
       errors.add(:base, _('Please check the proxy is configured and running on the host.'))
     end
     features.any?
   end
+
+  def validate_features(features)
+    features.map{|f| Feature.name_map[f]} == nil
```
shlomizadok: I feel like a Tuna fish for sending this code :sob:

shlomizadok:
Maybe commit message should be something like:
"Skip feature if not in Feature model"

app/models/smart_proxy.rb
```diff
@@ -115,11 +115,12 @@ def associate_features
 
     begin
       reply = ProxyAPI::Features.new(:url => url).features
-      if reply.is_a?(Array) && reply.any?
```
domcleal, line 118: I think the Array check here on `reply` is important - it's moved below to `valid_features.is_a?(Array)`, but you know that's an array as you're constructing it with .map. Check the `reply` is an array before trying to use map on it.

domcleal:
> Maybe commit message should be something like: "Skip feature if not in Feature model"

I think it already did this. This now correctly logs an error when no _known_ features are found. Perhaps "add error when no known proxy features found", if you wanted to change it.

Could you add a test for this bug and validation? It contains a skip when Rails.env is test, but I think that could probably be removed and ProxyAPI::Features be stubbed instead when necessary (e.g. in the factory).

domcleal (changes requested):

app/models/smart_proxy.rb
shlomizadok: You are right. I'll change that.

shlomizadok:
Thanks for the review and comments @domcleal. Will add tests to validate this issue.

shlomizadok:
@domcleal  - Added tests, reworded commit message.
Let me know what you think.

app/models/smart_proxy.rb
```diff
@@ -111,15 +111,15 @@ def sanitize_url
   end
 
   def associate_features
-    return true if Rails.env == 'test'
```
domcleal, line 114: It looks like removing this is affecting a very large number of tests - sorry, the check might have to remain.

I do have one more suggestion if you'd like to continue with it - separate the feature association from the validation. So this method with the ProxyAPI::Features call becomes a `before_validation` and is skipped when Rails.env.test?, but the validation that the smart proxy has at least one feature is a normal validator which can be tested in isolation from the feature assignment.

app/models/smart_proxy.rb
```diff
@@ -111,15 +111,15 @@ def sanitize_url
...
-
     begin
       reply = ProxyAPI::Features.new(:url => url).features
-      if reply.is_a?(Array) && reply.any?
-        self.features = Feature.where(:name => reply.map{|f| Feature.name_map[f]})
+      return errors.add(:base, _('Invalid features format')) unless reply.is_a?(Array)
```
domcleal: Please expand this message a little, it'll be unclear to users - perhaps "An invalid response was received while requesting available features from this proxy" (similar to the message below)?

domcleal (changes requested):

app/models/smart_proxy.rb
shlomizadok: :+1:  Again your English is better than mine :smile: 
Will change.

app/models/smart_proxy.rb
shlomizadok: I have been pulling my hair with how to fix the tests... Your idea is better. Will refactor

shlomizadok: Refactored. Hope this is what you had in mind

shlomizadok:
@domcleal - first approach didn't work well for me (I may have done something wrong).
I have moved to stubbing the Proxy API calls: 
`ProxyAPI::Features.any_instance.stubs(:features => Feature.name_map.keys)`

Let me know if that's acceptable

test/models/nic_test.rb
```diff
@@ -233,7 +233,7 @@ def teardown
 
     test '.proxy uses any BMC SmartProxy if none is found in subnet' do
       assert @subnet.proxies.select { |proxy| proxy.features.map(&:name).include?('BMC') }
-      assert_equal @interface.proxy.url, SmartProxy.with_features('BMC').first.url + '/bmc'
+      assert_includes SmartProxy.with_features('BMC').map { |sp| sp.url + '/bmc' }, @interface.proxy.url
```
shlomizadok, line 236: We are not sure there is always only one proxy

test/controllers/smart_proxies_controller_test.rb
```diff
@@ -18,6 +18,7 @@ def test_create_invalid
   end
 
   def test_create_valid
+    ProxyAPI::Features.any_instance.stubs(:features =>["dns", "dhcp"])
```
shlomizadok: Not needed anymore. Will remove

shlomizadok:
OK, I think that besides https://github.com/theforeman/foreman/pull/3895#discussion_r81953386 I'm good :bowtie:

test/active_support_test_case_helper.rb
```diff
@@ -39,6 +40,10 @@ def skip_if_plugin_asked_to
     end
   end
 
+  def fetch_proxy_features
+    ProxyAPI::Features.any_instance.stubs(:features => Feature.name_map.keys)
```
domcleal: This adds a new SQL query to every single test, as it's called regardless of whether the stubbing's needed or not. What's it used for? Can the stubbing be added to the tests that need it?

shlomizadok: Yeah, it was a precaution for Katello tests. I will open a PR on the failing tests there (to use factories)

shlomizadok:
https://github.com/Katello/katello/pull/6377 should solve the failures in Katello

app/models/smart_proxy.rb
```diff
@@ -111,15 +111,15 @@ def sanitize_url
   end
 
   def associate_features
-    return true if Rails.env == 'test'
-
     begin
       reply = ProxyAPI::Features.new(:url => url).features
-      if reply.is_a?(Array) && reply.any?
-        self.features = Feature.where(:name => reply.map{|f| Feature.name_map[f]})
+      return errors.add(:base, _('An invalid response was received while requesting available features from this proxy')) unless reply.is_a?(Array)
```
dLobatog: How about printing the response too? That can be helpful while debugging

app/models/smart_proxy.rb
```diff
@@ -111,15 +111,15 @@ def sanitize_url
...
+      valid_features = reply.map{|f| Feature.name_map[f]}.compact
+      if valid_features.any?
+        self.features = Feature.where(:name => valid_features)
       else
         self.features.clear
-        errors.add :base, _('No features found on this proxy, please make sure you enable at least one feature')
+        errors.add :base, _('No matching features found on this proxy, please make sure you enable at least one feature')
```
dLobatog: Not sure what 'matching' means in this context for a user that doesn't know about the bug. Also it tells the user to 'enable at least one feature' but control flow allows Foreman to get here when there was certainly one feature enabled (albeit not known by Foreman)

How about 'Features #{invalid_features} in this proxy are not recognized by Foreman. If these features come from a Smart Proxy plugin, make sure Foreman has the plugin installed too.' when there are invalid features, and leaving the old message "No features found..." for the case where there are no features found?

dLobatog:
@shlomizadok I left a few comments concerning the 'not features found' case vs 'only unknown features found' - thanks for fixing this

app/models/smart_proxy.rb
shlomizadok: @dLobatog - Shall I print it in a logger message or just add it to the errors ?

shlomizadok:
[test katello]

app/models/smart_proxy.rb
shlomizadok: Changed. Let me know if this is what you had in mind.

app/models/smart_proxy.rb
shlomizadok: Added a logger message.

test/test_helper.rb
```diff
@@ -62,6 +62,10 @@ def set_basic_auth(user, password)
     @request.env['HTTP_ACCEPT'] = 'application/json'
   end
 
+  def fetch_proxy_features
+    ProxyAPI::Features.any_instance.stubs(:features => Feature.name_map.keys)
```
domcleal: Wasn't this going to be removed now? (https://github.com/theforeman/foreman/pull/3895#discussion_r81988688)

domcleal (changes requested):

test/test_helper.rb
shlomizadok: The previous one was for unit (/models) tests. This one is for functional (/controllers) tests. I guess I can remove it too and add per each test.

shlomizadok: @domcleal - removed ^, added per test

domcleal (approved):

domcleal:
Merged, thanks @shlomizadok.

PR #3897, opened 2016-09-28
fixes #16717 - input_group_btn option no longer added to field

domcleal:
When input_group_btn was passed to text_f (as in the host name field),
it was added as an attribute to the textbox as well as being output.


dLobatog:
Thanks @domcleal !

PR #3898, opened 2016-09-28
fixes #16718 - remove relation.to_a.delete_if in multiple actions

domcleal:
To obtain a list of hosts where updating their build status etc. didn't
work, the controller relied on calling .to_a.delete_if on a relation and
then checking the result later. This doesn't work under Rails 5, it
simply returns the cached result of the relation again as it hasn't
actually changed.

The results are now saved in a different array instead of affecting the
relation.

### DIFF
--- a/app/controllers/hosts_controller.rb
+++ b/app/controllers/hosts_controller.rb
@@ -532,7 +532,7 @@ class HostsController < ApplicationController
   def submit_multiple_build
     reboot = params[:host][:build] == '1' || false
 
-    @hosts.to_a.delete_if do |host|
+    missed_hosts = @hosts.select do |host|
       success = true
       forward_url_options(host)
       begin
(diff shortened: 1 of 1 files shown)


dLobatog:
Merged as ca59bac0da3e09601c6974f1ad4bcbddf4b13d04, thanks @domcleal!

app/controllers/hosts_controller.rb
```diff
@@ -543,31 +543,28 @@ def submit_multiple_build
         Foreman::Logging.exception(message, error)
         success = false
       end
-      success
+      !success
     end
```
lzap, line 547: The select block can be cleaned a bit now, the last line looks weird. I'd rename it to pick or similar, or ideally remove whole variable completely and use just true and false in both possible code flows as the begin block returns the last evaluation.

lzap:
Whoops, did't refresh the tab.

PR #3899, opened 2016-09-28
fixes #16700 - change CatchJsonParseErrors test to cleanup driver

domcleal:
On completion, the Capybara driver was left unset causing the next test
to execute with the JS driver and introducing AJAX concurrency issues.
Using the integration test helpers ensures it's reset correctly.

The AboutIntegrationTest now uses the JS driver explicitly to test its
AJAX features, and stubs the network-based ping calls.


dLobatog:
Thanks @domcleal !

PR #3900, opened 2016-09-28
Fixes #16704 - Allow filtering on dashboard

tbrisker:
Some filters on the dashboard caused the reports widget to break,
specifically searching using free text or on attributes that are not
part of the hosts table (such as environment name).
This is caused by an issue with Rails that digregards outer joins when a
query is used as an inner query for a where clause using hash syntax.

### DIFF
--- a/test/unit/dashboard_test.rb
+++ b/test/unit/dashboard_test.rb
@@ -1,0 +1,86 @@
+require 'test_helper'
+
+class DashboardTest < ActiveSupport::TestCase
+  setup do
+    @env = FactoryGirl.create(:environment)
+    @host = FactoryGirl.create(:host, :with_reports, :environment => @env)
+  end
+
+  test 'hosts returns correct host' do
+    data = Dashboard::Data.new
+
+    as_admin do
+      assert_equal 1, data.hosts.length
+    end
+  end
+
+  test 'hosts works with environment filter' do
+    data = Dashboard::Data.new("environment = #{@env.name}")
+
+    as_admin do
+      assert_equal 1, data.hosts.length
+    end
+  end
+
+  test 'hosts works with free text filter' do
+    data = Dashboard::Data.new(@env.name)
+
+    as_admin do
+      assert_equal 1, data.hosts.length
+    end
+  end
+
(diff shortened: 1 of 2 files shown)


domcleal:
Could you add tests for this please? I'd suggest at a couple of layers, e.g. a unit test for Dashboard::Data and a general integration test for the dashboard with filtering. Both should catch this class of errors, and the unit test will be easy and cheap to extend for more precise tests.

tbrisker:
@domcleal I've added several unit tests.
I did not manage to get a properly working integration test due to issues with the ajax loading of the widgets.
I'm leaving on an extended leave soon, if you or any other maintainer wishes to add integration tests while I'm away, my branch allows edits from maintainers.

domcleal:
There are some test failures - the dashboard tests are failing on SQLite3, and PuppetClassImporter tests are failing on multiple DBs.

It looks like the latter is expecting certain ordering of environments in queries - possibly the creation of new environments in the tests (pushing the count over 100) is affecting the sorting ("99" vs "100") so the comparisons in the test may need `.sort` adding.

tbrisker:
@domcleal Thanks, updated the tests to use `@env` instead of `env` for the created environment which seems to clear it correctly from the db, and moved it to the setup block since nearly all the tests use it. Also changed the column quoting so it should hopefully now pass on all dbs.

test/unit/dashboard_test.rb
```diff
@@ -0,0 +1,88 @@
...
+
+  test 'hosts returns correct host' do
+    data = Dashboard::Data.new
+    as_admin do
+      assert_equal 1, data.hosts.length
+    end
+  end
+
+  test 'hosts works with environment filter' do
+    @host.update_attribute(:environment_id, @env.id)
```
domcleal: Pass `:environment => @env` to the host creation in `setup` to avoid these additional updates in many of the tests.

domcleal (changes requested):

tbrisker:
@domcleal fixed, also fixed some minor spacing inconsistencies.

domcleal (approved):

domcleal:
Merged, thanks @tbrisker.

PR #3903, opened 2016-09-29
fixes #16514 - redirect to login page if user is deleted

kgaikwad:
#3831

I have deleted the previous branch due to some bad commits. I have created this branch without functional test. 
I will add those changes and update the branch.

### DIFF
--- a/app/controllers/concerns/foreman/controller/authentication.rb
+++ b/app/controllers/concerns/foreman/controller/authentication.rb
@@ -7,7 +7,7 @@ module Foreman::Controller::Authentication
 
   def authenticate
     return true if (User.current && Rails.env.test? && api_request?) ||
-                   session[:user] && (User.current = User.unscoped.find(session[:user]))
+                   session[:user] && (User.current = User.unscoped.find_by(id: session[:user]))
 
     if SETTINGS[:login]
       # authentication is enabled
@@ -17,15 +17,16 @@ module Foreman::Controller::Authentication
         logger.info("Authorized user #{user.login}(#{user.to_label})")
         set_current_user user
       else
-        if api_request?
-          false
-        else
-          # Keep the old request uri that we can redirect later on
-          session[:original_uri] = request.fullpath
-          @available_sso ||= SSO::Base.new(self)
-
-          redirect_to @available_sso.login_url unless @available_sso.has_rendered
+        return false if api_request?
+        # Keep the old request uri that we can redirect later on
+        session[:original_uri] = request.fullpath
+        @available_sso ||= SSO::Base.new(self)
+        if session[:user] && !User.current
+          backup_session_content { reset_session }
+          flash[:warning] = _('Your session has expired, please login again')
         end
+        return if @available_sso.has_rendered
+        redirect_to @available_sso.login_url
       end
     else
       # We assume we always have a user logged in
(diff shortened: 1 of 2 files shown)


kgaikwad:
@domcleal, please review a functional test.

test/controllers/application_controller_subclass_test.rb
```diff
@@ -255,6 +255,20 @@ class TestableResourcesControllerTest < ActionController::TestCase
       assert_template :partial => false
     end
   end
+
+  context 'logged in user is deleted' do
+    setup do
+      @user = users(:admin)
```
domcleal: This setup block's rather superfluous, `@user` can be a plain inline variable in the test itself (if needed) instead of an instance variable.

app/controllers/concerns/foreman/controller/authentication.rb
```diff
@@ -17,15 +17,13 @@ def authenticate
...
-          @available_sso ||= SSO::Base.new(self)
-
-          redirect_to @available_sso.login_url unless @available_sso.has_rendered
-        end
+        return false if api_request?
+        # Keep the old request uri that we can redirect later on
+        session[:original_uri] = request.fullpath
+        @available_sso ||= SSO::Base.new(self)
+        return if @available_sso.has_rendered
+        flash[:warning] = _('Your session has expired.') if session[:user] && !User.current
```
domcleal: 1. Please also add an assertion to the test for the flash message (`assert_equal "...", flash[:warning]`)
2. Should `session[:user]` be deleted at this point? It's now known that the user ID is incorrect, it might be safest to wipe the entire contents - even better, using `backup_session_content { reset_session }` in the same way as `expire_session` (app/controllers/concerns/foreman/controller/session.rb).
3. The message should probably match `expire_session`'s message by adding ", please login again".

domcleal (changes requested):

kgaikwad:
@domcleal , thanks for reviewing the pull-request.
I have added suggested changes in the same commit.

domcleal:
@theforeman-bot ok to test

app/controllers/concerns/foreman/controller/authentication.rb
```diff
@@ -17,15 +17,16 @@ def authenticate
...
-          # Keep the old request uri that we can redirect later on
-          session[:original_uri] = request.fullpath
-          @available_sso ||= SSO::Base.new(self)
-
-          redirect_to @available_sso.login_url unless @available_sso.has_rendered
+        return false if api_request?
+        # Keep the old request uri that we can redirect later on
+        session[:original_uri] = request.fullpath
+        @available_sso ||= SSO::Base.new(self)
+        return if @available_sso.has_rendered
```
domcleal: I think this needs to move _below_ the new block - if an SSO method is in use or enabled (e.g. enable `authorize_login_delegation` for SSO::Apache to activate) then the session reset and message don't appear to the user.

domcleal: I created a test that demonstrates the issue if you'd like to include it: https://github.com/domcleal/foreman/commit/9cf396bb7867570485bd63e6f2b6cb5cd21a5629

domcleal (changes requested):

app/controllers/concerns/foreman/controller/authentication.rb
kgaikwad: @domcleal , as per mentioned I have made changes as well as included your test in same pull-request.

domcleal (approved):

domcleal:
Merged, thank you @kgaikwad.

PR #3904, opened 2016-09-29
fixes #16725 - add plugin extension point for smart proxies

timogoebel:
This commit adds the ability to easily add smart proxies to models from plugins. This also includes tiny dsl to specify the proxies in a model directly.
The proxies automatically show up in the model's form and api.

If this gets merged, I'll send a pr to foreman_discovery to make use of the feature.


app/services/foreman/plugin.rb
```diff
@@ -351,6 +352,16 @@ def parameter_filters(klass)
       @parameter_filters.fetch(klass.is_a?(Class) ? klass.name : klass, [])
     end
 
+    def smart_proxy_for(klass, name, options)
+      @smart_proxies[klass.name] ||= {}
+      @smart_proxies[klass.name][name] = options
+      klass.register_smart_proxy(name, options)
```
domcleal, line 358: This probably won't work well with class autoloading, as if `klass` gets reloaded at any point the belongs_to association (etc) will be lost. You could either:
1. Add an `included` block to the concern so after it's loaded, it queries Foreman::Plugin to reload any smart proxies on the model.
2. Remove the `@smart_proxies` stuff (though it's my preferred method) and use `to_prepare` instead, which would be called on every development request (in the dev environment where reloading occurs).

app/views/api/v2/subnets/main.json.rabl
```diff
@@ -6,14 +6,8 @@ attributes :network, :network_type, :cidr, :mask, :priority, :vlanid, :gateway,
...
-  extends "api/v2/smart_proxies/base"
-end
-
-child :tftp => :tftp do
-  extends "api/v2/smart_proxies/base"
-end
-
-child :dns => :dns do
-  extends "api/v2/smart_proxies/base"
+Subnet.smart_proxies.keys.each do |proxy|
```
domcleal: `@object`?

app/helpers/subnets_helper.rb
```diff
@@ -1,4 +1,6 @@
 module SubnetsHelper
+  include SharedSmartProxiesHelper
```
domcleal: Shouldn't be needed, all helpers get loaded usually. Ditto in others.

app/helpers/realms_helper.rb
```diff
@@ -0,0 +1,3 @@
+module RealmsHelper
```
domcleal: Not required? All helpers should be loaded.

app/validators/proxy_features_validator.rb
```diff
@@ -6,7 +6,7 @@ def initialize(args)
 
   def validate_each(record, attribute, value)
     if value && !value.has_feature?(@options[:feature])
-      record.errors["#{attribute}_id"] << _(@options[:message])
+      record.errors["#{attribute}_id"] << _('does not have the %s feature') % @options[:feature]
```
domcleal: Best to prefer `@options[:message]` here if it's set, else (`||`) use the built in one - it's more like Rails ones then, and it's possible a plugin is using this validator.

app/views/api/v2/domains/main.json.rabl
```diff
@@ -3,3 +3,9 @@ object @domain
 extends "api/v2/domains/base"
 
 attributes :fullname, :dns_id, :created_at, :updated_at
+
+@object.smart_proxies.keys.each do |proxy|
```
domcleal: I'd suggest putting this into a partial like api/v2/taxonomies/children_nodes, it's identical in every location.

app/views/hosts/_form.html.erb
```diff
@@ -73,7 +73,10 @@
 
         <%= host_puppet_environment_field f %>
 
-        <%= puppet_master_fields f, true, @host.hostgroup && @host.hostgroup_id_was.nil? %>
+        <%= smart_proxy_fields f,
+          :can_override => true,
```
domcleal: Nitpick: indentation should be more than the `smart_proxy_fields` word itself.

domcleal (changes requested):

app/views/api/v2/subnets/main.json.rabl
timogoebel: It has to be `Subnet`, because `@object` would be `Subnet::Ipv4` for example. The smart proxies are registered only on `Subnet´.

domcleal: `class_attribute` may help here, as its value is inherited by subclasses if a more specific subclass value isn't set. See https://github.com/domcleal/foreman/commit/4d05ce2493ed5db55e6b167d43ab816906481e3d for example.

Note that's specifically fetching, merging, then setting the value of smart_proxies_from_model rather than modifying the same hash, which should allow for subclasses to vary their registered smart proxies from the parent (they'll receive a copy of the parent + their own alterations).

timogoebel:
@domcleal : Updated. Thanks for your comments and  domcleal@4d05ce2! I have also added api doc handling. It has some redundancy but it am unsure how to write an extension point for this without having to monkey-patch apipie-rails. So I haven't really looked into that.

app/views/api/v2/realms/main.json.rabl
```diff
@@ -1,3 +1,5 @@
 object @realm
 extends "api/v2/realms/base"
+extends "api/v2/smart_proxies/children_nodes"
+
 attributes :name, :realm_proxy_id, :realm_type, :created_at, :updated_at
```
domcleal: `realm_proxy_id` and `_name` should be replaced by the smart_proxy/children_nodes too I think, it can be generated easily.

Ditto in other views, e.g. host has puppet_proxy_id/puppet_ca_proxy_name which can be replaced.

domcleal (changes requested):

app/views/api/v2/realms/main.json.rabl
timogoebel: @domcleal : I saw the params, however we break compatibility if we just remove these. What is the normal procedure for this?

domcleal: Oh sorry - I hadn't meant to remove them. I meant to replace the explicit listings of `attributes :realm_proxy_id` by code that automatically calls `attributes` with `_id` and `_name` for each registered smart proxy association. IDs and names for new associations from plugins will automatically appear in the output too then, not just through full child nodes.

timogoebel:
@domcleal : Added the requested change and parameter_filter handling. Hope, we have all covered now. The smart proxy params for the host model were defined in `host_base`, I have moved them to `host_common`.  Was there any reason they were in `host_base`?

app/validators/proxy_features_validator.rb
```diff
@@ -6,7 +6,12 @@ def initialize(args)
 
   def validate_each(record, attribute, value)
     if value && !value.has_feature?(@options[:feature])
-      record.errors["#{attribute}_id"] << _(@options[:message])
+      if @options[:message].nil?
+        message = _('does not have the %s feature') % @options[:feature]
+      else
+        message = _(@options[:message])
+      end
+      record.errors["#{attribute}_id"] << message
```
domcleal: Use `record.errors.add("#{attribute}_id", message)`, it's the preferred API. (Deprecated later.)

timogoebel:
I just tested this a little more. Somehow the inheritance thing still does not work. I need to take a closer look.

timogoebel:
@domcleal : I reworked the registration from plugins a bit. They are also inherited now.

domcleal (changes requested):
Thanks, other than one small inline comment the registration system does seem to be more reliable now and it works well.

It would be reassuring if you could add some tests (either to the concern tests or subnet model tests) to check that the inheritance is indeed working between Subnet, Subnet::Ipv4/6 as I can't see any coverage of this right now.

domcleal:
> The smart proxy params for the host model were defined in host_base, I have moved them to host_common. Was there any reason they were in host_base?

No, probably a mistake from when protected attributes were initially added.

> @domcleal : I reworked the registration from plugins a bit. They are also inherited now.

Per the review comment, it'd be great if there was some test coverage of this, I don't see it.

app/validators/proxy_features_validator.rb
timogoebel: Done.

timogoebel:
> It would be reassuring if you could add some tests (either to the concern tests or subnet model tests) to check that the inheritance is indeed working between Subnet, Subnet::Ipv4/6 as I can't see any coverage of this right now.

Added the tests (in subnet/ipv4_test.rb, ipv6_test.rb and tests for the concern). Should have done that without an explicit request, sorry.

domcleal:
No worries, thanks. Functional tests seem to be failing now, but I can't immediately see why, or any change that might have introduced it. Could you check them please?

timogoebel:
The tests run fine on my laptop. I have rebased the code. Let's see if they run through now.

domcleal:
It's an interference between some of the unit tests that permanently change the behaviour of models via the plugin interface, then the controller tests fail later. You can reproduce it with:

```
be ruby -Itest:. -e "require 'test/unit/plugin_test.rb'; require 'test/controllers/api/v1/subnets_controller_test.rb'"
```

You may want to stub some of the class modifications (e.g. adding validators) in the plugin unit tests rather than letting them modify the class (and attempting to revert changes).

timogoebel:
Thanks a lot, @domcleal . You are awesome. Saved me the time of looking into this. That is why the test now runs against the "Awesome" module defined in the test class. Well, and it should fix the tests. I also renamed `smart_proxies` to ´registered_smart_proxies´ as apparently the former term is used quite a lot throughout the code and I didn't want to shadow it.

timogoebel:
There is still one failing test. I'll investigate tomorrow.

timogoebel:
@domcleal : That was it. Tests are finally green.

domcleal (approved):

domcleal:
Merged, thanks @timogoebel!

Please could you add a section to http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin, noting that it requires Foreman 1.14 or higher.

timogoebel:
I added the section to the docs.

lzap:
Nice patch!

PR #3905, opened 2016-09-30
Fixes #16752 - Extract template code from Host::Managed

dLobatog:
Host::Managed contains a lot of code related with determining the
template that a host should have. Other Host objects such as
Host::Discovered might need that logic, without becoming a
Host::Managed.

This commit extracts that into Hostext::OperatingSystem so it can
be used on plugins.

### DIFF
--- a/app/controllers/hosts_controller.rb
+++ b/app/controllers/hosts_controller.rb
@@ -638,8 +638,8 @@ class HostsController < ApplicationController
   # we don't need any has_many relation to determine what proxies are used and the view
   # renders only resulting templates set so the rest of form is unaffected
   def template_used
-    host = params[:id] ? Host.readonly.find(params[:id]) : Host.new
-    host.attributes = host_params.select { |k,v| !k.end_with?('_ids') }.except(:host_parameters_attributes)
+    host = params[:id] ? Host::Base.readonly.find(params[:id]) : Host.new
+    host.attributes = host_attributes_for_templates(host)
     templates = host.available_template_kinds(params[:provisioning])
     return not_found if templates.empty?
     render :partial => 'provisioning', :locals => { :templates => templates }
@@ -906,5 +906,16 @@ class HostsController < ApplicationController
 
   def host_power_state(key)
     HOST_POWER[key].merge(:title => _(HOST_POWER[key][:title]))
+  end
+
+  def host_attributes_for_templates(host)
+    # This method wants to only get the attributes applicable to the current
+    # kind of host. For example 'is_owned_by', even if it's in host_params,
+    # should be ignored by Host::Discovered or any other Host class that does
+    # not have that attribute
+    host_attributes = host.class.attribute_names
+    host_params.select do |k,v|
+       host_attributes.include?(k) && !k.end_with?('_ids')
+    end.except(:host_parameters_attributes)
   end
 end
--- a/app/models/concerns/hostext/operating_system.rb
+++ b/app/models/concerns/hostext/operating_system.rb
@@ -1,0 +1,60 @@
+module Hostext
+  module OperatingSystem
+    extend ActiveSupport::Concern
+
+    included do
+      scope :with_os, -> { where('hosts.operatingsystem_id IS NOT NULL') }
+      alias_attribute :os, :operatingsystem
+      validates :operatingsystem_id, :presence => true,
+        :if => ->(host) { host.managed }
+    end
+
+    # returns a configuration template (such as kickstart) to a given host
+    def provisioning_template(opts = {})
+      opts[:kind]               ||= "provision"
+      opts[:operatingsystem_id] ||= operatingsystem_id
+      opts[:hostgroup_id]       ||= hostgroup_id
+      opts[:environment_id]     ||= environment_id
+
+      Taxonomy.as_taxonomy(self.organization, self.location) do
+        ProvisioningTemplate.find_template opts
+      end
+    end
+
+    def available_template_kinds(provisioning = nil)
+      kinds = template_kinds(provisioning)
+      kinds.map do |kind|
+        ProvisioningTemplate.find_template({ :kind               => kind.name,
+                                             :operatingsystem_id => operatingsystem_id,
+                                             :hostgroup_id       => hostgroup_id,
+                                             :environment_id     => environment_id
+        })
+      end.compact
+    end
+
+    def template_kinds(provisioning = nil)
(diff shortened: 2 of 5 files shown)


app/models/concerns/hostext/operating_system.rb
```diff
@@ -0,0 +1,55 @@
...
+    def provisioning_template(opts = {})
+      opts[:kind]               ||= "provision"
+      opts[:operatingsystem_id] ||= operatingsystem_id
+      opts[:hostgroup_id]       ||= hostgroup_id
+      opts[:environment_id]     ||= environment_id
+
+      ProvisioningTemplate.find_template opts
+    end
+
+    def available_template_kinds(provisioning = nil)
```
timogoebel, line 24: Do you think, we can clean that one a bit up? Maybe extracting this to two methods? The if looks very scary.

app/controllers/hosts_controller.rb
```diff
@@ -681,8 +681,11 @@ def process_taxonomy
   # we don't need any has_many relation to determine what proxies are used and the view
   # renders only resulting templates set so the rest of form is unaffected
   def template_used
-    host = params[:id] ? Host.readonly.find(params[:id]) : Host.new
-    host.attributes = host_params.select { |k,v| !k.end_with?('_ids') }.except(:host_parameters_attributes)
+    host = params[:id] ? Host::Base.readonly.find(params[:id]) : Host.new
+    host.attributes = host_params.select do |k,v|
```
timogoebel: This looks scary as well. Let's extract this to a method, shall we? Maybe a method that prepares "host".

timogoebel (commented):
@dLobatog : Looks good to me, we should just use this opportunity to clean a bit. I left inline comments.

app/models/concerns/hostext/operating_system.rb
```diff
@@ -0,0 +1,55 @@
...
+      kinds.map do |kind|
+        ProvisioningTemplate.find_template({ :kind               => kind.name,
+                                             :operatingsystem_id => operatingsystem_id,
+                                             :hostgroup_id       => hostgroup_id,
+                                             :environment_id     => environment_id
+        })
+      end.compact
+    end
+
+    def render_template(template)
```
ares: I'd prefer not mixing template selecting logic with rendering. Note that some plugins tend to include own extensions to `Host::Managed` but this would allow rendering in more classes. If it's out of scope for this PR, please keep rendering in host object.

app/models/concerns/hostext/operating_system.rb
```diff
@@ -0,0 +1,55 @@
+module Hostext
+  module OperatingSystem
+    extend ActiveSupport::Concern
```
ares, line 3: Could you please rather extract it to a separate class, e.g. HostTemplatePicker that would require a host and only return templates? Hiding it in concern that is included does not make the object any less fat and make testing harder. With this pattern, one could inject testing host object in unit tests for picker class.

dLobatog: I agree, but it was a cheap and simple way to fix my problem with https://github.com/theforeman/foreman_discovery/pull/305 without making me write any (possibly wrong logic). I'll investigate how to make that class.

lzap:
What's the status boys?

app/controllers/hosts_controller.rb
```diff
@@ -907,4 +907,11 @@ def host_power_ping(result)
   def host_power_state(key)
     HOST_POWER[key].merge(:title => _(HOST_POWER[key][:title]))
   end
+
+  def host_attributes_for_templates(host)
+    host_params.select do |k,v|
+      host.class.new.attributes.include?(k) &&
```
timogoebel: What is this doing? Why wo we need to create a new instance of the host class for every param?

timogoebel (commented):
Looks good to me in general. I have just one question inline.

app/controllers/hosts_controller.rb
dLobatog: No need to do it on every param, that's for sure, thanks for the catch 😄 I can get the list of attributes beforehand.

The reason is that when I have a `Host::Discovered`, there may be parameters on `host_params` that I don't want to use. E.g: `is_owned_by`: is not defined by Host::Discovered, so those parameters I want to skip.

dLobatog:
@timogoebel Replied to your question and updated the method to explain what it's doing in a comment. It's necessary to get the attributes of the Host class at least once to verify we're not setting attributes not available to the current Host object (Host::Discovered may have different methods than Host::Managed)

app/controllers/hosts_controller.rb
```diff
@@ -907,4 +907,15 @@ def host_power_ping(result)
   def host_power_state(key)
     HOST_POWER[key].merge(:title => _(HOST_POWER[key][:title]))
   end
+
+  def host_attributes_for_templates(host)
+    # This method wants to only get the attributes applicable to the current
+    # kind of host. For example 'is_owned_by', even if it's in host_params,
+    # should be ignored by Host::Discovered or any other Host class that does
+    # not have that attribute
+    host_attributes = host.class.new.attributes
```
timogoebel: How about `host.class.attribute_names`?

app/models/concerns/hostext/operating_system.rb
```diff
@@ -0,0 +1,60 @@
...
+      opts[:hostgroup_id]       ||= hostgroup_id
+      opts[:environment_id]     ||= environment_id
+
+      Taxonomy.as_taxonomy(self.organization, self.location) do
+        ProvisioningTemplate.find_template opts
+      end
+    end
+
+    def available_template_kinds(provisioning = nil)
+      kinds = if provisioning == 'image'
```
timogoebel: This could easily be extracted to a method and might make CodeClimate happy.

timogoebel:
@dLobatog : Thanks. I don't want to waste any unnecessary time on this but left two more comments inline.

dLobatog:
@timogoebel Thanks for the tips - fixed now 😄

timogoebel (approved):
@dLobatog: Looks good to me.
@ares: Anything to add or am I free to merge?

ares:
Well, it's not ideal since in fact it does not extract it from the Host object, it's just making it mixable to more objects, but that's not a reason to block discovery feature, we can always refactor later. Take it as :+1: from me :-)

timogoebel:
Merged, thanks @dLobatog.

PR #3906, opened 2016-09-30
Fixes #16753 - Expose lodash as lodash, not _

xprazak2:


dLobatog:
@xprazak2 Most methods are the same in both, shouldn't katello use `lodash` too? It's supposed to be faster and more maintained. The migration guide points out few differences (https://github.com/lodash/lodash/wiki/Migrating), why did the tests not catch this / where is this a problem? Thanks for raising these questions :smile: !

xprazak2:
@dLobatog The problem is on smart_proxies/show_page when you have a pulp node on the proxy:



because at that point, Bastion js files get loaded and underscore is among them 



lodash has `_.get`but underscore does not, and underscore hijacks the `_` sometime during loading:





Renaming is a simple fix, swapping lodash for underscore has a benefit of removing one dependency from ecosystem, but could be a little more involved and I do not know how Katello feels about lodash. Any thoughts on this, @ehelms?

ehelms:
I'm fine with moving to lodash, provided we just do some basic poking around at the UI. I am confident they are compatible but worth just making sure nothing blows up. Should be easy enough to remove from Bastion and use it directly from the core JS set.

gailsteiger:
In my opinion, it is a bad practice to use both underscore and lodash. Things happen, but at this point the solution is to settle on one of the libraries.

tbrisker:
:+1: to remove underscore from bastion/katello, this issue should be solved in any plugins that hit it not in core.
Plugins in general should not include libraries that are provided by core exactly because of possibly conflicting versions (lodash is a fork of underscore), we have several similar cases with other libraries as well. 
I recommend closing this PR in favor of https://github.com/Katello/katello/pull/6375 and https://github.com/Katello/bastion/pull/133

app/assets/javascripts/application.js
```diff
@@ -432,7 +432,7 @@ function setPowerState(item, status){
       $('#loading_power_state').remove();
     }
   }else{
-    $('#loading_power_state').text(_('Unknown power state'))
+    $('#loading_power_state').text(__('Unknown power state'))
```
tbrisker, line 435: regardless of this PR's fate, this is a good catch and should be fixed separately.

gailsteiger: $('#loading_power_state') should be cached when the fix is made. No need to repeat DOM operation.

xprazak2:
The PRs replacing underscore with lodash in Bastion & Katello were merged and the translation typo is addressed separately, closing.

PR #3695, opened 2016-08-01
Fixes #15926 - [API] Add method to get ENC values of a host

soukron:
Fixes #15926 - [API] Add method to get ENC values of a host

PR #3907, opened 2016-10-03
Fixes #15926 - [API] Add method to get ENC values of a host

xprazak2:
As discussed with @ares earlier today, this is a rebase of #3695 since the original author is currently unavailable and we really want to get it in.


dLobatog:
Thanks @xprazak2 & @soukron  - notice the commit authorship is preserved

soukron:
Thanks for keeping the authorship and excuse me for not attending properly the comments.

PR #3908, opened 2016-10-03
Fixes #16765 - DB IPAM works with excluded IPs

ares:


tstrachota:
:+1: tested, it fixes the problem for me

app/services/ipam/db.rb
```diff
@@ -8,7 +8,7 @@ def suggest_ip
       to = subnet.to.present? ? IPAddr.new(subnet.to) : IPAddr.new(subnet_range.last.to_i - 2, subnet.family)
       (from..to).each do |address|
         ip = address.to_s
-        if !subnet.known_ips.include?(ip) && !excluded_ips.include?(ip)
+        if !subnet.known_ips.include?(ip) && !(excluded_ips || []).include?(ip)
```
timogoebel: You might want to fix this globally, i.e. for all providers here: https://github.com/theforeman/foreman/blob/79c0664a008afe0fea54a0b5256426f3867a02bc/app/services/ipam/base.rb#L13

timogoebel: I mean something like this:

``` ruby
{:excluded_ips => nil}.fetch(:excluded_ips, []) || []
```

ares: only internal db provider supports excluded_ips, but I could maybe define the attr_reader so it can never return nil which would be more robust in base class

ares: fetch already does that... anyway I redefined the reader, please take a look

timogoebel: Well, what you are trying to prevent is the following:

``` ruby
> {:abc => nil}.fetch(:abc, [])
 => nil
```

Fetch does not help you here. It just takes the provided default it the key cannot be found in the hash.

I do think, defining a custom getter is overengineering the whole thing. But that might just be a matter of taste.

ares: You're right, I'm starting to wonder if it would be better to keep it failing if someone explicitly send invalid value.

ares: I removed the default value, I think we shouldn't try to recover from bad wrong usage.

timogoebel:
@ares : Looks good to me. :-)

app/services/ipam/db.rb
timogoebel: ok, that's fair. On second thought we could've gone with:

``` ruby
{:abc => nil}[:abc] || []
```

PR #3909, opened 2016-10-03
fixes #16749 - improve NIC update performance during fact imports

domcleal:
- In get_facts_for_interface, replace regexes in hash key filtering
  with simple String methods and simplify hash construction
- Change logger.debug to lazy evaluation for production performance,
  especially when calling #inspect

Before:

```
  PuppetFactParser 1      6.114k (±16.1%) i/s -     57.836k in   9.951523s
 PuppetFactParser 50     15.127  (±26.4%) i/s -    135.000  in  10.036620s
PuppetFactParser 500      0.143  (± 0.0%) i/s -      2.000  in  14.027935s
Memory stats
Total objects allocated: 61005590
Total heap pages allocated: 4218
```

After:

```
  PuppetFactParser 1      9.308k (±14.8%) i/s -     87.936k in   9.934396s
 PuppetFactParser 50    213.821  (±16.4%) i/s -      2.018k in  10.007988s
PuppetFactParser 500      2.610  (± 0.0%) i/s -     27.000  in  10.357182s
Total objects allocated: 149652784
Total heap pages allocated: 4221
```

(approximately 15x faster, and sub-second for the ~400 interfaces in the
bug report)

In Host#set_interface, unchanged interfaces are no longer saved to avoid
creating an unnecessary validation and DB transactions in most imports.

### DIFF
--- a/app/models/host/base.rb
+++ b/app/models/host/base.rb
@@ -398,15 +398,17 @@ module Host
       update_virtuals(iface.identifier_was, name) if iface.identifier_changed? && !iface.virtual? && iface.persisted? && iface.identifier_was.present?
       iface.attrs = attributes
 
-      logger.debug "Saving #{name} NIC for host #{self.name}"
-      result = iface.save
+      if iface.new_record? || iface.changed?
+        logger.debug "Saving #{name} NIC for host #{self.name}"
+        result = iface.save
 
-      unless result
-        logger.warn "Saving #{name} NIC for host #{self.name} failed, skipping because:"
-        iface.errors.full_messages.each { |e| logger.warn " #{e}" }
+        unless result
+          logger.warn "Saving #{name} NIC for host #{self.name} failed, skipping because:"
+          iface.errors.full_messages.each { |e| logger.warn " #{e}" }
+        end
+
+        result
       end
-
-      result
     end
 
     def update_virtuals(old, new)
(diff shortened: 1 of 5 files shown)


app/models/host/base.rb
```diff
@@ -399,7 +399,7 @@ def set_interface(attributes, name, iface)
       iface.attrs = attributes
 
       logger.debug "Saving #{name} NIC for host #{self.name}"
-      result = iface.save
+      result = iface.save if iface.new_record? || iface.changed?
```
lzap: Could you pull the logger message in the statement as well to avoid confusion?

app/services/puppet_fact_parser.rb
```diff
@@ -119,10 +119,12 @@ def get_interfaces
   end
 
   def get_facts_for_interface(interface)
-    iface_facts = @facts.select { |name, value| name =~ /.*_#{interface}\Z/ }
-    iface_facts = iface_facts.map { |name, value| [name.gsub("_#{interface}", ''), value] }
+    iface_facts = @facts.inject([]) do |facts, (name, value)|
+      facts << [name.chomp("_#{interface}"), value] if name.end_with?("_#{interface}")
+      facts
+    end
```
lzap, line 125: Nicely done, from O(n^2) to O(n).

lzap (approved):
Nice, the comment is minor. I have a power outage so need to test tomorrow :-)

app/models/host/base.rb
domcleal: Thanks, done. I noticed that the whole block should have been within the conditional as it was also checking for errors after save. It's all been moved.

lzap:
Thanks @domcleal! Merged.

PR #3910, opened 2016-10-03
refs #16702 - fix deprecation passing 'controller' etc.

domcleal:


dLobatog:
Merged as 590b8293c49975d21da9cdfa0cd19c4798b2c584, thanks @domcleal!

PR #3911, opened 2016-10-04
Fixes #16785 - improve resource names for lookup keys

ares:


dLobatog:
Thanks @ares , this also makes the manual more understandable :+1:

PR #3912, opened 2016-10-04
fixes #16786 - don't call to_sentence on associations

domcleal:


daviddavis:
I'm not familiar with rails 5 but am curious why you're calling `map(&:name)` over `pluck(:name)`?

domcleal:
> I'm not familiar with rails 5 but am curious why you're calling map(&:name) over pluck(:name)?

The associations are all eager loaded in the index query, so the data (on say `SmartProxy#features`) is already preloaded and available. Using `pluck(:name)` would generate an additional query (actually, N queries).

daviddavis:
LGTM

dLobatog:
Thanks @domcleal @daviddavis !

PR #3896, opened 2016-09-28
refs #9016 - remove unnecessary condition, fix comment location

domcleal:
Corrects rubocop error raised for one-line conditional.

PR #3914, opened 2016-10-04
Fixes #16787 - Fix rubocop offense

daviddavis:
Apparently HoundCI only checks lines that are edited and therefore, it's possible that rubocop errors might occur outside those lines. Perhaps this error happened when Hound upgraded to rubocop 0.42? Not totally sure.


daviddavis:
Dupe of https://github.com/theforeman/foreman/pull/3896

PR #3915, opened 2016-10-04
fixes #16792 - don't modify AR record attributes in-place

domcleal:
Rails 5 calls #freeze on any string passed in as an attribute during
validation, so attributes must not be edited in-place with gsub! and
similar methods.

### DIFF
--- a/app/models/compute_resource.rb
+++ b/app/models/compute_resource.rb
@@ -333,7 +333,7 @@ class ComputeResource < ActiveRecord::Base
   end
 
   def sanitize_url
-    self.url.chomp!("/") unless url.empty?
+    self.url = url.chomp("/") unless url.empty?
   end
 
   def random_password
--- a/app/models/concerns/strip_whitespace.rb
+++ b/app/models/concerns/strip_whitespace.rb
@@ -7,8 +7,7 @@ module StripWhitespace
 
   def strip_spaces
     self.changes.each do |column, values|
-      # return string if RuntimeError: can't modify frozen String
-      self.send(column).strip! if (values.last.is_a?(String) && !skip_strip_attrs.include?(column)) rescue send(column)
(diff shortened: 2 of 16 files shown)


dLobatog:
Thanks @domcleal . Are these the only ones causing test failures? I also found these instances of in-place modifications after validation:

`sanitize_url` on ComputeResource
`cleanup_ip` on Subnet::IPv4
`remove_trailing_chars` on Template
`sanitize_url` on SmartProxy

domcleal:
Thanks for checking for more instances, they were the only ones found through tests. It prompted me to go back over models and found a few more. I've fixed as many instances as I could find and have added tests for any such methods that appeared to missing coverage.

I think that freezing only occurs in certain circumstances, currently when a model is built _through_ an association on another model - e.g. `host.fact_values.build(value: fact_value)` would freeze fact_value, hence the changes to OS models as fact values get frozen during import. Despite this, I think it's useful to change these to be more robust - it's possible that more strings get frozen automatically in future, either by Rails (e.g. `params`?) or by enabling the option in Ruby.

domcleal:
Seems that "fixing" StripWhitespace to work with frozen strings has caused tests to fail in a few different areas of the app, which I'll need to take a closer look at. Setting back to WoC.

domcleal:
> Seems that "fixing" StripWhitespace to work with frozen strings has caused tests to fail in a few different areas of the app

Fixed by changing StripWhitespace to use read/write_attribute instead of getters and setters, which caused inconsistent results on some models that overrode them (e.g. Host overwrites the setter, but not getter).

Tests pass, except for an intermittent integration test fixed in #3931. (Edit: PR rebased on develop with that fix.)

dLobatog:
Sorry for taking so long on coming back to this, thanks @domcleal

PR #3917, opened 2016-10-05
fixes #16798 - move scoped_search definitions to STI subclasses

domcleal:
scoped_search doesn't support class inheritance with STI, so registering
definitions on the subclass fixes various issues. This fixes an issue
where scoped_search on CommonParameter calls Parameter.all and is
returned a list of CommonParameters under Rails 4.2, as it relies on a
bug (#18806) for applying the type='CommonParameter' condition when
calling Parameter.all.

On Rails 5.0, this bug was fixed and calling Parameter.all within a
scope on CommonParameter now returns all types of parameters.
Registering all scoped_search definitions on the subclasses ensures that
scoped_search calls CommonParameter.all instead.

The taxonomies API changed as it called Taxonomix objects with scopes
such as `.locations.search_for`, which no longer existed when the
scoped_search definitions were removed from Taxonomy (.locations is a
Taxonomy association with a where clause). This now explicitly searches
via the appropriate Taxonomy subclass.

### DIFF
--- a/app/models/parameter.rb
+++ b/app/models/parameter.rb
@@ -9,7 +9,13 @@ class Parameter < ActiveRecord::Base
   include Authorizable
   validates :name, :presence => true, :no_whitespace => true
 
-  scoped_search :on => :name, :complete_value => true
+  def self.inherited(child)
+    child.instance_eval do
+      scoped_search :on => :name, :complete_value => true
+      scoped_search :on => :value, :complete_value => :true
+    end
+    super
+  end
 
   default_scope -> { order("parameters.name") }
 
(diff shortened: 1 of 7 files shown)


app/models/parameter.rb
```diff
@@ -9,7 +9,12 @@ class Parameter < ActiveRecord::Base
   include Authorizable
   validates :name, :presence => true, :no_whitespace => true
 
-  scoped_search :on => :name, :complete_value => true
+  def self.inherited(child)
+    child.instance_eval do
+      scoped_search :on => :name, :complete_value => true
```
ares, line 14: [this](https://github.com/theforeman/foreman/blob/develop/app/models/parameters/common_parameter.rb#L5) won't be necessary anymore I think

domcleal: Thanks, I've removed the search definitions from CommonParameter and added the :value definition here.

dLobatog:
Thanks @domcleal @ares !

PR #3918, opened 2016-10-05
Fixes #16803 - tolerate templates without documentation

ares:


ares:
easy to reproduce with remote execution plugin, try editing job template

domcleal:
Merged, thanks @ares.

PR #3919, opened 2016-10-05
Fixes #16806 - include renderer url dependencies

ares:
I took the liberty to improve the the existing test so it does not include the render module in class method and avoids using instance variable if it's not needed.

EDIT: I realized we had two separate test files for the same so I also joined them and applied the same fix.


domcleal:
Merged, thanks @ares!

PR #3920, opened 2016-10-05
fixes #16805 - reload host parameters on host group change

domcleal:
On Rails 4.2, `Host#type_changed?` may return nil due to bug #24220
causing data-type-changed to be missing from the host form and the JS
in handleHostgroupChangeEdit to assume host#edit is a hostgroup form.
Now that data-type-changed is always present, the full reload of
parameters occurs when the host group is changed on a host.

### DIFF
--- a/app/views/hosts/_form.html.erb
+++ b/app/views/hosts/_form.html.erb
@@ -5,7 +5,7 @@
 
 <% Taxonomy.as_taxonomy @organization , @location do %>
 
-  <%= form_for @host,  :html => {:data => {:id => @host.try(:id), :type_changed => @host.type_changed?, :submit => 'progress_bar'}} do |f| %>
+  <%= form_for @host,  :html => {:data => {:id => @host.try(:id), :type_changed => !!@host.type_changed?, :submit => 'progress_bar'}} do |f| %>
     <%= base_errors_for @host %>
 
     <ul class="nav nav-tabs" data-tabs="tabs">
--- a/test/integration/host_js_test.rb
+++ b/test/integration/host_js_test.rb
@@ -223,6 +223,17 @@ class HostJSTest < IntegrationTestWithJavascript
       host = Host::Managed.search_for('name ~ "myhost1"').first
       assert_equal env2.name, host.environment.name
     end
+
+    test 'setting host group updates parameters tab' do
+      hostgroup = FactoryGirl.create(:hostgroup, :with_parameter)
+      visit new_host_path
+      select2(hostgroup.name, :from => 'host_hostgroup_id')
+      wait_for_ajax
+
+      assert page.has_link?('Parameters', :href => '#params')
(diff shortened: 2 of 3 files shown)


test/integration/host_js_test.rb
```diff
@@ -315,6 +332,23 @@ class HostJSTest < IntegrationTestWithJavascript
       click_link 'Parameters'
       assert_equal find("#s2id_host_lookup_values_attributes_#{lookup_key.id}_value .select2-chosen").text, "true"
     end
+
+    test 'changing host group updates parameters tab' do
```
dLobatog, line 336: Nitpick, same name as the first test

test/integration/host_js_test.rb
```diff
@@ -315,6 +332,23 @@ class HostJSTest < IntegrationTestWithJavascript
...
+      assert page.has_link?('Parameters', :href => '#params')
+      click_link 'Parameters'
+      assert page.has_selector?("#inherited_parameters #name_#{hostgroup1.group_parameters.first.name}")
+
+      click_link 'Host'
+      select2(hostgroup2.name, :from => 'host_hostgroup_id')
+      wait_for_ajax
+
+      click_link 'Parameters'
+      assert page.has_selector?("#inherited_parameters #name_#{hostgroup2.group_parameters.first.name}")
```
dLobatog, line 351: Maybe it's worth asserting that the parameter from hostgroup 1 is not available at this point?

app/views/hosts/_form.html.erb
```diff
@@ -5,7 +5,7 @@
 
 <% Taxonomy.as_taxonomy @organization , @location do %>
 
-  <%= form_for @host,  :html => {:data => {:id => @host.try(:id), :type_changed => @host.type_changed?, :submit => 'progress_bar'}} do |f| %>
+  <%= form_for @host,  :html => {:data => {:id => @host.try(:id), :type_changed => !!@host.type_changed?, :submit => 'progress_bar'}} do |f| %>
```
dLobatog, line 8: :+1: , will be harmless in Rails 5

dLobatog (changes requested):
Minor comments on the integration test, thanks @domcleal !

test/integration/host_js_test.rb
domcleal: They're in different contexts, but I've changed the other one to "setting host group [..]" as it's not changing from one group to another.

test/integration/host_js_test.rb
domcleal: Added, thanks.

domcleal:
PR updated with the fixed test name and extra assertion, thanks for the review.

dLobatog (approved):

dLobatog:
Thanks @domcleal !

PR #3921, opened 2016-10-05
 Fixes #16548 - Changing user own password require current password 

dhlavac:
Changing user own password require current password and adding changing password activity to Audit log

### DIFF
--- a/app/controllers/concerns/foreman/controller/parameters/user.rb
+++ b/app/controllers/concerns/foreman/controller/parameters/user.rb
@@ -6,13 +6,14 @@ module Foreman::Controller::Parameters::User
   class_methods do
     def user_params_filter
       Foreman::ParameterFilter.new(::User).tap do |filter|
-        filter.permit :default_location_id,
+        filter.permit :current_password,
+          :default_location_id,
           :default_organization_id,
+          :description,
           :firstname,
           :lastname,
           :locale,
           :mail,
-          :description,
           :mail_enabled,
           :password,
           :password_confirmation,
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -15,7 +15,7 @@ class User < ActiveRecord::Base
   ANONYMOUS_API_ADMIN = 'foreman_api_admin'
 
   validates_lengths_from_database :except => [:firstname, :lastname, :format, :mail, :login]
-  attr_accessor :password, :password_confirmation
+  attr_accessor :password, :password_confirmation, :current_password
   after_save :ensure_default_role
   before_destroy EnsureNotUsedBy.new([:direct_hosts, :hosts]), :ensure_hidden_users_are_not_deleted, :ensure_last_admin_is_not_deleted
 
@@ -85,6 +85,8 @@ class User < ActiveRecord::Base
   validate :name_used_in_a_usergroup, :ensure_hidden_users_are_not_renamed, :ensure_hidden_users_remain_admin,
            :ensure_privileges_not_escalated, :default_organization_inclusion, :default_location_inclusion,
            :ensure_last_admin_remains_admin, :hidden_authsource_restricted, :ensure_admin_password_changed_by_admin
+  before_validation :verify_current_password, :if => Proc.new {|user| user == User.current},
+                    :unless => Proc.new {|user| user.password.empty?}
   before_validation :prepare_password, :normalize_mail
   before_save       :set_lower_login
 
@@ -480,6 +482,12 @@ class User < ActiveRecord::Base
 
   def set_default_widgets
     Dashboard::Manager.reset_user_to_default(self)
+  end
+
+  def verify_current_password
+    unless self.matching_password?(current_password)
+      errors.add :current_password, _("Incorrect password")
+    end
   end
 
   protected
--- a/app/views/users/_form.html.erb
+++ b/app/views/users/_form.html.erb
@@ -35,6 +35,9 @@
                         :label => _('Authorized by') } ) unless @editing_self %>
 
       <div id='password' <%= display? !@user.manage_password? %>>
+        <% if User.current == @user  %>
+          <%= password_f f, :current_password, :label => _('Current password'), :placeholder => '' %>
+        <% end %>
         <%= password_f f, :password, :label => _('Password'), :error => @user.errors[:password_hash] %>
         <%= password_f f, :password_confirmation, :label => _('Verify') %>
       </div>
--- a/test/controllers/users_controller_test.rb
+++ b/test/controllers/users_controller_test.rb
@@ -117,9 +117,9 @@ class UsersControllerTest < ActionController::TestCase
                     :login => "johnsmith", :password => "dummy", :password_confirmation => "DUMMY"
                   }
                  }, set_session_user
-    mod_user = User.find_by_id(user.id)
 
-    assert mod_user.matching_password?("changeme")
+    user.reload
+    assert user.matching_password?("changeme")
     assert_template :edit
   end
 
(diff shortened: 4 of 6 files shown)


app/controllers/concerns/foreman/controller/parameters/user.rb
```diff
@@ -14,6 +14,7 @@ def user_params_filter
           :mail,
           :description,
           :mail_enabled,
+          :current_password,
```
domcleal: Please keep this list alphabetised (it was prior to 6cc6c9e).

app/models/user.rb
```diff
@@ -9,13 +9,13 @@ class User < ActiveRecord::Base
   include Taxonomix
   include DirtyAssociations
   include UserTime
-  audited :except => [:last_login_on, :password, :password_hash, :password_salt, :password_confirmation]
+  audited :except => [:last_login_on, :password, :password_salt, :password_confirmation]
```
domcleal: Removing password_hash will cause the password itself to be stored in the audit details. Changing the UI to hide this is not sufficient because the hash is still stored and will be available via the API etc.

Either exclude the field from being audited and add a new Audit entry when the password is changed containing whatever details are required, or filter out the value of the password hash in AuditExtensions.

This also seems unrelated to the issue/PR reason about requiring the current password. I'd appreciate it if you'd split the auditing changes away from the password requirement, into another ticket and PR.

app/models/user.rb
```diff
@@ -85,6 +85,8 @@ def self.name_format
   validate :name_used_in_a_usergroup, :ensure_hidden_users_are_not_renamed, :ensure_hidden_users_remain_admin,
            :ensure_privileges_not_escalated, :default_organization_inclusion, :default_location_inclusion,
            :ensure_last_admin_remains_admin, :hidden_authsource_restricted, :ensure_admin_password_changed_by_admin
+  before_validation :verify_current_password, :if => Proc.new {|user| user == User.current},
+                    :unless => Proc.new {|user| user.password.empty?}
```
domcleal, line 89: When is this condition required?

app/models/user.rb
```diff
@@ -474,6 +476,12 @@ def set_default_widgets
     Dashboard::Manager.reset_user_to_default(self)
   end
 
+  def verify_current_password
+    if !self.matching_password?(current_password)
```
domcleal: Prefer `unless` to `if !`

app/models/user.rb
```diff
@@ -474,6 +476,12 @@ def set_default_widgets
...
+      errors.add :current_password, _("Wrong password")
```
domcleal: "Incorrect password"?

domcleal:
This is missing test coverage, it should at least have functional coverage, please add to test/controllers/users_controller_test.rb.

domcleal (changes requested):

domcleal:
@theforeman-bot ok to test

app/models/user.rb
dhlavac: @domcleal When user want to change some attribute without changing password. (e.g to change his own description or location without entering current password)

app/controllers/concerns/foreman/controller/parameters/user.rb
dhlavac: ok thanks

app/models/user.rb
dhlavac: Ok I'll try to do otherwise but I think that, a part of this issue was "We need to ensure that the password change activity is logged" and that means log in Audits with changed password

app/models/user.rb
dhlavac: It’s my fault

test/controllers/users_controller_test.rb
```diff
@@ -135,6 +135,23 @@ class UsersControllerTest < ActionController::TestCase
     assert_redirected_to users_url
   end
 
+  test "current user have to enter current password to change password" do
+    user = User.new :login => "John", :mail => "john@doe.com", :firstname => "john", :lastname => "doe", :auth_source => auth_sources(:internal)
+    User.current = user
+    user.password = "password"
+    assert user.save
```
domcleal: Prefer `FactoryGirl.create(:user, :password => 'password')` over manual construction. (Actually, 'password' is even the default for the factory.)

test/controllers/users_controller_test.rb
```diff
@@ -135,6 +135,23 @@ class UsersControllerTest < ActionController::TestCase
...
+
+    put :update, {:id => user.id,
+                  :user => {
+                    :current_password => "password", :login => "johnsmith", :password => "newpassword", :password_confirmation => "newpassword"
+                  }
+    }, set_session_user
+
+    modified_user = User.find_by_id(user.id)
```
domcleal: `user.reload`

app/models/user.rb
domcleal: I think that's a separate change, whether or not that's written in the issue description.

domcleal:
Could you expand the tests some more please? I'd suggest using unit tests for these as they don't need to exercise the controller. The missing coverage is:
1. If the _wrong_ current password is entered, the user fails to save/validate.
2. If user is being edited by another user/admin, the current password isn't required.

domcleal (changes requested):
Needs unit test coverage to check the _negative_ condition (that the current password is actually needed) and that admin/non-current user is unaffected.

app/models/user.rb
dhlavac: Ok I am going to open a new ticket for that issue

dhlavac:
@domcleal added tests , could you look at that ?

test/models/user_test.rb
```diff
@@ -874,4 +876,24 @@ def editing_self_helper
...
+    assert user.valid?
+  end
+
+  test "changing user's own password with incorrect current password" do
+    user = FactoryGirl.create(:user)
+    as_user user do
+      user.password = "newpassword"
+      user.current_password = "hatatitla"
+    end
+    assert user.save
```
domcleal: This test doesn't seem to check what's described - it says it checks the user changing their own password with an incorrect current password, but this line and lines 896/7 actually suggest it _passes_. It should be _failing_ validation.

I think it's because the save/validation are not in the `as_user user` block.

app/views/users/_form.html.erb
```diff
@@ -35,6 +35,9 @@
                         :label => _('Authorized by') } ) unless @editing_self %>
 
       <div id='password' <%= display? !@user.manage_password? %>>
+        <% if User.current == @user  %>
+          <%= password_f f, :current_password, :label => _('Current password') %>
```
domcleal: Should the placeholder be disabled here? The stored value is only really on the `:password` field, not current password.

test/models/user_test.rb
```diff
@@ -874,4 +876,24 @@ def editing_self_helper
...
+  test "changing user password as admin without setting current password" do
+    user = FactoryGirl.create(:user)
+    as_admin do
+      user.password = "newpassword"
+      assert user.save
+    end
+    assert user.valid?
+  end
+
+  test "changing user's own password with incorrect current password" do
```
domcleal, line 888: Please also add a test for the current user changing the password with the correct current password.

domcleal (changes requested):

test/models/user_test.rb
dhlavac: This test is already in controller.

domcleal: It should probably be a model test though, the logic is in the model, not the controller.

domcleal:
If you're changing a pull request, especially in response to comments, please state the changes you've made and ask for a re-review. It's unclear otherwise if and why a PR has been updated.

test/models/user_test.rb
dhlavac: OK then delete this tests in controller ?

domcleal: Up to you!

dhlavac: Ok thank you

dhlavac:
@domcleal added test for correct current password and removed placeholder from current password field ,could you look at that ?

domcleal (approved):

domcleal:
One last question: in the ticket, you wrote "This means one can change the password of other users in the same group".

Did you mean to introduce a restriction that other users must also have the current password of the user to change it, or only for the user logged in? The current behaviour in the PR seems most correct to me, but I wondered what this comment meant.

domcleal:
Your commit message also contains the text:

```
Please enter the commit message for your changes. Lines starting
```

Please remove this.

dhlavac:
@domcleal I copied that issue from Bugzilla and I think that only makes sense when a current user want to change his own password. If someone else wants to change the password of another user, he should have a permission for that. (comment removed)

domcleal:
Merged, thanks @dhlavac.

PR #3922, opened 2016-10-05
Fixes #5673 - Allow deletion of role with users

shlomizadok:


app/helpers/roles_helper.rb
```diff
@@ -8,4 +8,12 @@ def role_link(role)
       end
     end
   end
+
+  def delete_role_confirmation(role)
+    if role.users.any?
+      _("Warning! This will remove %{name} from %{number} users. are you sure?") % {name: role.name, number: role.users.size}
```
domcleal: Use `n_()` here for the pluralised form, i.e. "from %{number} user" (1) and "from %{number} users".

test/controllers/roles_controller_test.rb
```diff
@@ -51,12 +51,12 @@ class RolesControllerTest < ActionController::TestCase
     assert_nil Role.find_by_id(role.id)
   end
 
-  test 'roles in use cannot be destroyed' do
+  test 'roles in use can be destroyed' do
```
domcleal: Might as well delete this I think, or perhaps change it to a test to check you can't delete built in roles. Your unit test coverage of deletion of roles with users is better I think, and cheaper.

domcleal (changes requested):

test/controllers/roles_controller_test.rb
shlomizadok: Changed to testing built-in role

app/helpers/roles_helper.rb
shlomizadok: Changed to pluralized form

test/models/role_test.rb
```diff
@@ -36,6 +36,15 @@ class RoleTest < ActiveSupport::TestCase
     role.must_be :valid?
   end
 
+  it "Should delete role who has users" do
```
dLobatog, line 39: nitpick - lowercase should. can be fixed on merge. Also `FactoryGirl.build` is preferable for creating the User here, it'll prove the point just the same (and the relationship need not to be in db)

domcleal (approved):

test/models/role_test.rb
domcleal: Fixed the case on merge, but I'd prefer in this case to keep the objects persisted in the DB to ensure we test the whole destroy mechanism (there are a few dependent destroys with cached user roles etc).

domcleal:
Merged as a2c1e8d, thanks @shlomizadok.

PR #3923, opened 2016-10-06
Fixes #16813 - improve error messages for associations

ares:


domcleal:
Can you check the test failures please? Updating that test should suffice for unit tests, but a new test in test/controllers/api/v2/base_controller_subclass_test.rb would be good to check the API responds with the correct message when the error occurs.

ares:
Thanks @domcleal, I updated the unit test and added the function test to provisioning_controller. I'm not sure if stubbing in base controller would be worth the effort. I'd need two models with has_many relationship. Let me know if you have better idea how to test it in base controller.

dLobatog:
:+1: from my side - leaving it to @domcleal since I don't know if the test in base controller (vs the one now in provisioning templates) is that critical

domcleal:
Merged, thanks @ares.

PR #3924, opened 2016-10-06
#16528 - Subnet's Discovery Capsule can point to a deleted proxy

amitkarsale:
In foreman db, there is a missing key constraint from subnet.discovery_id to smart_proxies.id. That means, if a subnet refers a Discovery Capsule and I delete the capsule/proxy, no warning is raised.

Add a key constraint similar to e.g. "subnet.tftp_id -> smart_proxies.id".


domcleal:
I don't think this is a Foreman issue as it relates to the Discovery plugin. It should be submitted to https://github.com/theforeman/foreman_discovery instead.

amitkarsale:
okay will make a new pull request on foreman discovery. Thanks for the pointer.

lzap:
Thanks, for the record the relevant PR is: https://github.com/theforeman/foreman_discovery/pull/307

PR #3630, opened 2016-07-08
Fixes #15621 - WebUI: Cloned roles saved with nil builtin

vision2910:
The above fix will prevent nil value for 'builtin' attribute for Role while cloning it. The changed files are-

/app/controllers/roles_controller.rb#L84

/test/functional/roles_controller_test.rb#L83

PR #3925, opened 2016-10-07
Fixes #16828 - cloned-roles - role.builtin must be 0

swapab:
Though `roles.builtin` has [`:default => 0`](https://github.com/theforeman/foreman/blob/12612809c5deb885b13ceaf36b6a147c21688eb5/db/migrate/20160715131352_set_role_builtin_default.rb). This is to ensure cloned roles always have `builtin == 0`

### DIFF
--- a/test/controllers/roles_controller_test.rb
+++ b/test/controllers/roles_controller_test.rb
@@ -122,6 +122,7 @@ class RolesControllerTest < ActionController::TestCase
       cloned_role = Role.find_by_name('clonedrole')
       assert_not_nil cloned_role
       assert_equal @role.permissions, cloned_role.permissions
+      assert_equal cloned_role.builtin, 0
     end
   end
 end
--- a/app/controllers/roles_controller.rb
+++ b/app/controllers/roles_controller.rb
@@ -94,7 +94,7 @@ class RolesController < ApplicationController
       new_role.name = params[:role][:name]
       new_role.organization_ids = params[:role][:organization_ids]
       new_role.location_ids = params[:role][:location_ids]
-      new_role.builtin = false
+      new_role.builtin = 0
     else
       new_role = Role.new(role_params)
     end


test/controllers/roles_controller_test.rb
```diff
@@ -123,5 +123,15 @@ class RolesControllerTest < ActionController::TestCase
       assert_not_nil cloned_role
       assert_equal @role.permissions, cloned_role.permissions
     end
+
+    test "cloned roles should have builtin value 0 instead of `false`" do
```
domcleal: The comment about `false` isn't helpful after this patch is merged, it should simply state that it should have the value 0.

test/controllers/roles_controller_test.rb
```diff
@@ -123,5 +123,15 @@ class RolesControllerTest < ActionController::TestCase
...
+      params = { :role => {:name => 'clonedrole'},
+                 :original_role_id => @role.id,
+                 :cloned_role => true }
+      post :create, params, set_session_user
+
+      cloned_role = Role.find_by_name('clonedrole')
+      assert_equal cloned_role.builtin, 0
```
domcleal, line 125: Might as well put this assertion into the existing cloning test above? It'll save execution time.

domcleal (changes requested):
The test passes on current develop, so it's of limited value. I'd suggest just adding the assertion into the existing clone test, keeps everything simple.

domcleal:
@theforeman-bot ok to test

test/controllers/roles_controller_test.rb
swapab: @domcleal 
Believe me it was like that at some-point in time. But I was in 2 minds whether to add more asserts to one test or not. I totally agree on _saving execution time_.
Thanks for reviewing.

swapab:
@domcleal 
I wanted to discuss this with someone.
In `app/models/role.rb` - [`BUILTIN_DEFAULT_ROLE = 2`](https://github.com/theforeman/foreman/blob/develop/app/models/role.rb#L26) is for **default role**. Likewise shouldn't there be constants for 0 and 1 (w.r.t Roles) ?

ranjan:
@swapnilabnave  There is a ongoing [PR#3630](https://github.com/theforeman/foreman/pull/3630) for the same issue. I think that discussion thread will be helpful for you.

test/controllers/roles_controller_test.rb
swapab: Removed the redundant test.

swapab:
@ranjan Thanks for the direction. I think #3630 would be make it more meaningful.

domcleal (approved):

domcleal:
Merged, thanks @swapnilabnave.

lzap:
I think this patch is missing migration that will fix incorrect records:

```
Role.where(:builtin => nil).update_all(:builtin => 0)
```

Shell we create one? Edit: Creating one: http://projects.theforeman.org/issues/18321

PR #3926, opened 2016-10-07
Fixes #6492 - ipmi_boot permission renamed to ipmi_boot_hosts

dLobatog:
Authorizer expects permission names to follow a convention
'action'_'controller'. However this permission was not following it, and
this prevented the permission from being applied properly.

Before this fix, only admins could call ipmi_boot. I've also added a
small fix to the controller to check whether the BMC interface is
available before making the IPMI call - otherwise the error that Foreman
threw did not make much sense for the end user (NoMethodError on
bmc_proxy).

### DIFF
--- a/app/controllers/api/v2/hosts_controller.rb
+++ b/app/controllers/api/v2/hosts_controller.rb
@@ -242,6 +242,8 @@ Return the host's compute attributes that can be used to create a clone of this 
         else
           render :json => { :error => _("Unknown device: available devices are %s") % valid_devices.join(', ') }, :status => :unprocessable_entity
         end
+      rescue ::Foreman::Exception => e
+        render_message(e.to_s, :status => :unprocessable_entity)
       end
 
       api :POST, "/hosts/facts", N_("Upload facts for a host, creating the host if required")
@@ -351,7 +353,7 @@ Return the host's compute attributes that can be used to create a clone of this 
       end
 
       def permissions_check
-        permission = "#{params[:action]}_hosts".to_sym
+        permission = "#{action_permission}_hosts".to_sym
         deny_access unless Host.authorized(permission, Host).find(@host.id)
       end
 
--- a/db/migrate/20161007115719_rename_ipmi_boot_permission.rb
+++ b/db/migrate/20161007115719_rename_ipmi_boot_permission.rb
@@ -1,0 +1,11 @@
+class RenameIpmiBootPermission < ActiveRecord::Migration
+  def up
+    Permission.where(:name => 'ipmi_boot').
+      update_all(:name => 'ipmi_boot_hosts')
+  end
+
+  def down
+    Permission.where(:name => 'ipmi_boot_hosts').
+      update_all(:name => 'ipmi_boot')
+  end
+end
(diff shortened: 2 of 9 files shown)


test/controllers/hosts_controller_test.rb
```diff
@@ -1301,6 +1301,29 @@ def test_submit_multiple_rebuild_config_pessimistic
...
+    end
+
+    test 'responds gracefully if BMC is not available' do
+      put :ipmi_boot, { :id => @host.id, :ipmi_device => 'bios'},
+        set_session_user.merge(:user => @one.id)
+      assert_match(/No BMC NIC available for host/, flash[:error])
+      assert_redirected_to host_path(@host.id)
+    end
+
+    test 'responds gracefully if BMC is not available' do
```
domcleal: This test name should be changed, perhaps to say it responds correctly for a non-admin user or similar.

test/fixtures/permissions.yml
```diff
@@ -295,7 +295,7 @@
     created_at: "2013-12-04 08:41:04.931564"
     updated_at: "2013-12-04 08:41:04.931564"
   ipmi_boot:
```
domcleal: Nitpick, update the name here too, please.

db/migrate/20161007115719_rename_ipmi_boot_permission.rb
```diff
@@ -0,0 +1,6 @@
+class RenameIpmiBootPermission < ActiveRecord::Migration
+  def change
+    Permission.where(:name => 'ipmi_boot').
+      update_all(:name => 'ipmi_boot_hosts')
```
domcleal, line 4: This change isn't reversible, it'll need separate up/down methods.

app/controllers/hosts_controller.rb
```diff
@@ -305,6 +305,8 @@ def nics
   end
 
   def ipmi_boot
+    raise Foreman::Exception.new(
```
domcleal: This might need to be in the API v2 hosts controller (or the model) too? it calls ipmi_boot without any checking.

domcleal (changes requested):

dLobatog:
Thanks for taking a look @domcleal - I've updated it, the changes include:
- bmc_available check and exception moved to Host::Managed
- tests for controller and v2 controller handling that exception 
- fix for permissions check in API v2 
- migration is reversible

app/controllers/api/v2/hosts_controller.rb
```diff
@@ -351,7 +353,7 @@ def detect_host_type
       end
 
       def permissions_check
-        permission = "#{params[:action]}_hosts".to_sym
+        permission = "#{action_permission}_hosts".to_sym
```
dLobatog, line 356: Without this, we call for permissions that might not even exist in the next line. E.g: 'rebuild_config_hosts', 'boot_hosts', etc... I think this fixes permission problems with a bunch of other methods, not just 'ipmi_boot_hosts'

domcleal (approved):

domcleal:
Merged, thanks @dLobatog.

PR #3929, opened 2016-10-09
fixes #16819 - rebuild tftp with template kinds

timogoebel:


app/models/concerns/orchestration/tftp.rb
```diff
@@ -36,7 +36,9 @@ def tftp6
   def rebuild_tftp
     if tftp? || tftp6?
       begin
-        setTFTP
+        host.operatingsystem.template_kinds.each do |kind|
+          setTFTP(kind)
+        end
       rescue => e
```
domcleal: Can you move this _inside_ the template_kinds loop? Have a look at Orchestration::DNS for a similar example, where the rebuild method always tries to rebuild as much as possible, even if one fails, but then returns `false` on failure.

domcleal (changes requested):

lzap:
Ah, sorry about this one. This is when tests won't help much, compiler would do it.

timogoebel:
@lzap: No worries. ;-)

@domcleal : Pushed the latest changes that should address your comment.

domcleal (approved):

domcleal:
Merged, thanks @timogoebel.

PR #3931, opened 2016-10-10
refs #13424 - add wait_for_ajax to statistics page load test

domcleal:
Prevents errors occurring when the AJAX calls continue beyond the end of the test.

### DIFF
--- a/test/integration/statistic_test.rb
+++ b/test/integration/statistic_test.rb
@@ -3,6 +3,7 @@ require 'integration_test_helper'
 class StatisticIntegrationTest < IntegrationTestWithJavascript
   test "statistics page" do
     visit statistics_path
+    wait_for_ajax
     assert page.has_selector?('h3', :text => "OS Distribution")
     assert page.has_selector?('h3', :text => "Architecture Distribution")
     assert page.has_selector?('h3', :text => "Environment Distribution")


ohadlevy (commented):
I'm not sure if this will actually solve the problem as jquery is not fully used( react components has their own hooks/life cycle). We searched for a wait for react mount but it's considered a real bad practice and I'm not sure if it adds much value all together.

domcleal:
> I'm not sure if this will actually solve the problem as jquery is not fully used( react components has their own hooks/life cycle)

It appeared to wait for all AJAX requests to complete when I tested, did you check?

ohadlevy:
@domcleal I did, since we use JQuery.getJSON at https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/API.js#L8 it might actually work most of the time, the reason I'm hesitate is because:
1. there is more logic involved then just the ajax request, so in theory, it can finish the jquery part, and still not draw it on the screen (as the wait_for_ajax test simply check if jquery is active).
2. in theory finder methods of capibara should be smart enough to reload them self, i guess that's why we didn't include it in the first place, yet i assume the build was failing to start with?
3. i've read multiple articles (e.g. https://bibwild.wordpress.com/2016/02/18/struggling-towards-reliable-capybara-javascript-testing) - it didn't sound encouraging, as a global wait_for_react_mounting helper will need to be created, and I've yet seen a good / recommended way to do that.

I was happy as I thought this actually works, but if it turns out to be unstable, then I would prefer removing it as I think the jest coverage is high.

domcleal:
> 1. there is more logic involved then just the ajax request, so in theory, it can finish the jquery part, and still not draw it on the screen (as the wait_for_ajax test simply check if jquery is active).

Why does that matter? The purpose of the wait is to prevent requests being processed beyond the end of the test. If the requests have all completed, it's worked correctly.

> 1. in theory finder methods of capibara should be smart enough to reload them self, i guess that's why we didn't include it in the first place, yet i assume the build was failing to start with?

The test wasn't executed with JS enabled prior to https://github.com/theforeman/foreman/commit/659d49a1542e524c373b7af4a0a43aeda214074a#diff-05d411ead2ff2b4917be281afc963057 enabling it, so no AJAX requests were made. This isn't related to Capybara finders, it's about AJAX requests.

domcleal:
> I was under the assumption the main reason for waiting is to ensure the
> tests can be performed, e.g. to avoid
> https://github.com/jnicklas/capybara#asynchronous-javascript-ajax-and-friends

No, the `has_selector?` finders in the test already wait. They will work and may all match the widget headings before the AJAX requests for the individual charts have completed.

> Are you suggesting this was done to reduce load / impact on other tests?

Partly, yes. `wait_for_ajax` prevents an impact on further tests and stops errors during database cleanup within a test, when the server is still processing requests.

domcleal:
Sorry, I missed this point.

> Right, based on the link above, Capybara should wait by it self without
> additional logic..?

It will, but the matchers in the test only check for headings (that are part of the page) and don't check for anything in the AJAX responses.

ohadlevy (approved):

ohadlevy:
merged as 3f9c407 thanks @domcleal

PR #3932, opened 2016-10-12
refs #16725 - initialize smart_proxies for models

timogoebel:
This is needed for discovery where we have to add the concern "BelongsToProxies" to a model without any smart_proxies.


lzap:
Thanks :+1:

PR #3933, opened 2016-10-12
Fixes #16821 - Maintain test schema during initialization

adamruzicka:
Fixes the failing plugin tests

### DIFF
--- a/config/initializers/test_schema.rb
+++ b/config/initializers/test_schema.rb
@@ -1,0 +1,3 @@
+if Rails.env.test? && !Foreman.in_rake?('db:migrate') && !Foreman.in_rake?('db:test:prepare') && !defined?(Spring)
+  ActiveRecord::Migration.maintain_test_schema!
+end


config/initializers/test_schema.rb
```diff
@@ -0,0 +1,5 @@
+if Rails.env.test?
```
iNecas: I guess we can put this into one condition

iNecas (changes requested):
Small nitpick + I'm sure you have a link to a jenkins build that proves this fixes the issue.

config/initializers/test_schema.rb
```diff
@@ -0,0 +1,5 @@
...
+  unless Foreman.in_rake?('db:test:prepare')
```
dLobatog: Is this necessary? This task has been deprecated in Rails for a long time

dLobatog:
Here's a small script to verify it works :smile: 

```
#!/bin/sh
rm db/development.sqlite3 db/foreman_test.sqlite3
bundle exec rake db:drop db:create db:migrate
ruby -I'test' test/unit/foreman/access_permissions_test.rb
```

For some reason it gets quite stuck on my laptop on `2016-10-12T18:11:32  [app] [W] Recreating the whole DB cache` (more than 5m waiting already...). 

I'm OK with merging the fix, but keep in mind this is still a hack, the tests shouldn't expect the db to contain anything other than `fixtures`. Thanks @adamruzicka for stepping up

iNecas:
I know it works: I've seen it on adam's laptop:) but I want the jenkins to have the final word on that actually fixing the issue

adamruzicka:
Actually it looks there's some issue, the katello tests got stuck. I'll have to look into it tomorrow

config/initializers/test_schema.rb
adamruzicka: The `ActiveRecord::Migration.maintain_test_schema!` calls that rake task from [here](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/migration.rb#L583). Without that condition it would recurse infinitely

domcleal:
I've removed this hack once in f20c2a1, and I don't think it's a good idea to reintroduce it - as noted above, it's too easy for this to trigger an infinite recursion, among other problems.

I proposed a simple change to the plugin registration instead in #3939 to populate the permissions data all of the time in the test environment.

domcleal:
Closing as #3939 was merged.

PR #3934, opened 2016-10-12
Fixes #16887 - Display errors from compute resource in VM tab

dLobatog:
This commit allows these errors to show up in the VM tab. Without it
just a standard, unstyled 'internal server error' message shows up while
the error is only logged on Foreman's log.

I thought about either moving this to `webpack/` or write an integration test for this but it might be excessive for such a small change. Feel free to ask for it if you disagree.

screenshot from 2016-10-12 17-13-44 is how it looks now


domcleal:
Merged, thanks @dLobatog!

PR #3935, opened 2016-10-12
Fixes #16881 - Pass information from proxy error response

dLobatog:
Certain information that could be important during creation of hosts
should be displayed - currently the ProxyAPI abstracts the entire
exception message away.

This commit appends the message that comes from the proxy if there is
one to show it whenever a ProxyAPI::ProxyException is thrown

screenshot from 2016-10-12 17-58-13 an example of an error that  normally get's hidden under just '400 bad request'

### DIFF
--- a/lib/proxy_api/proxy_exception.rb
+++ b/lib/proxy_api/proxy_exception.rb
@@ -8,7 +8,15 @@ module ProxyAPI
     end
 
     def message
-      super + ' ' + _('for proxy') + ' ' + url
+      super + proxy_exception_response(exception) +
+        ' ' + _('for proxy') + ' ' + url
+    end
+
+    private
+
+    def proxy_exception_response(exception)
+      return '' unless exception.respond_to?(:response)
+      ": #{exception.response}"
     end
   end
 end
--- a/test/unit/proxy_api/proxy_exception_test.rb
+++ b/test/unit/proxy_api/proxy_exception_test.rb
@@ -1,0 +1,26 @@
+require 'test_helper'
+
+class ProxyApiExceptionTest < ActiveSupport::TestCase
+  test 'exception response body is added if available' do
+    fake_exception = OpenStruct.new(:response => 'useful message from proxy',
+                                    :message => 'useful message from proxy',
+                                    :code => 400)
+    proxy_exception = ProxyAPI::ProxyException.new(
+      'fakeurl',
+      fake_exception,
+      'unable to do something')
+    assert_match(
+      /ERF.*ProxyException.*unable to do something.*useful message from proxy.*/,
+      proxy_exception.message
+    )
+  end
+
+  test 'does not fail if response body is not available' do
+    proxy_exception = ProxyAPI::ProxyException.new('fakeurl',
+                                                   NoMethodError.new,
+                                                   'unable to do something')
+    assert_equal 'fakeurl', proxy_exception.url
+    assert_match(/unable to do something.*NoMethodError.*proxy fakeurl/,
+      proxy_exception.message)
+  end
+end


lib/proxy_api/proxy_exception.rb
```diff
@@ -3,12 +3,20 @@ class ProxyException < ::Foreman::WrappedException
     attr_reader :url
 
     def initialize(url, exception, message, *params)
+      message += append_exception_message(exception)
```
domcleal: Why append the message in this location and not in the `message` method? I think this will affect the error codes generated, which will now depend on the passed message and _any_ text supplied in the smart proxy response.

domcleal (changes requested):

lib/proxy_api/proxy_exception.rb
dLobatog: Shouldn't the error codes depend on this? Otherwise you're getting one error code for what could be very different proxy errors - here you could have errors coming from things as varied as MS DNS, libvirt, DHCP, BMC...  which are different errors that require different solutions.

domcleal: Each proxy feature should already have a unique set of error codes (e.g. ProxyExceptions are unique within each ProxyAPI class), but I agree there are different errors that the error code wouldn't reflect.

I'm concerned though that the text from the proxy may not be standardised - if it's used to generate the error code then the codes may not be reproducible at all, which is probably worse than them being too broad.

e.g. if the proxy returns an error stating it can't contact 192.168.0.100, then an error code generated based on that IP isn't useful to anybody except that one user, or ditto with file paths (cannot open file /srv/tftpboot/ or /var/lib/tftpboot).

dLobatog: Hmmm fair enough. I just realized after your comment that the error I did this for had to do with MAC addresses so it wouldn't be useful to have a different code per MAC :smile: I'll fix it, thanks @domcleal

dLobatog:
Updated so that the message is added at runtime

test/unit/proxy_api/proxy_exception_test.rb
```diff
@@ -0,0 +1,21 @@
+require 'test_helper'
+
+class ProxyApiExceptionTest < ActiveSupport::TestCase
+  test 'exception response body is added if available' do
+    fake_exception = OpenStruct.new(:response => 'useful message from proxy',
+                                    :code => 400)
+    proxy_exception = ProxyAPI::ProxyException.new(
+      'fakeurl',
+      OpenStruct.new(:message => 'some error'),
```
domcleal: Shouldn't this second parameter be the `fake_exception`, rather than using `expects(:exception)` below?

lib/proxy_api/proxy_exception.rb
```diff
@@ -8,7 +8,15 @@ def initialize(url, exception, message, *params)
     end
 
     def message
-      super + ' ' + _('for proxy') + ' ' + url
+      super + ' ' + proxy_exception_response(exception) +
```
domcleal: Nitpick, remove the extra space before `proxy_exception_response` as it'll cause a space before the colon and double-space otherwise.

test/unit/proxy_api/proxy_exception_test.rb
```diff
@@ -0,0 +1,21 @@
...
+      'unable to do something')
+    proxy_exception.expects(:exception).returns(fake_exception)
+    assert_match(/useful message from proxy/, proxy_exception.message)
```
domcleal: It'd be useful if this could test the entire message - or more of the concatenation, not only the proxy_exception_response bit.

test/unit/proxy_api/proxy_exception_test.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  test 'does not fail if response body is not available' do
```
domcleal, line 18: I think this needs to test `#message`, not just `#url`.

domcleal (changes requested):

dLobatog:
@domcleal I've updated this according to your comments, thanks for the review!

lib/proxy_api/proxy_exception.rb
```diff
@@ -8,7 +8,15 @@ def initialize(url, exception, message, *params)
...
     def message
-      super + ' ' + _('for proxy') + ' ' + url
+      super + proxy_exception_response(exception) +
+        ' ' + _('for proxy') + ' ' + url
+    end
+
+    private
+
+    def proxy_exception_response(exception)
+      return '' unless exception.respond_to?(:response)
```
domcleal, line 18: I think `exception` here is either the same or a new instance of _this_ exception. I think this needs to be `wrapped_exception`.

test/unit/proxy_api/proxy_exception_test.rb
```diff
@@ -0,0 +1,26 @@
+require 'test_helper'
+
+class ProxyApiExceptionTest < ActiveSupport::TestCase
+  test 'exception response body is added if available' do
+    fake_exception = OpenStruct.new(:response => 'useful message from proxy',
```
domcleal, line 5: If you change this message here and in the expectation, e.g. s/message/response/ then the test will fail:

    Expected /ERF.*ProxyException.*unable to do something.*useful response from proxy.*/ to match "ERF12-2902 [ProxyAPI::ProxyException]: unable to do something ([OpenStruct]: useful message from proxy) for proxy fakeurl".

The response isn't in the message.

domcleal (changes requested):
Doesn't seem to work!

PR #3937, opened 2016-10-13
Fixes #16899 - rename helper with conflicting path name

ares:
help_path is defined by [this](https://github.com/theforeman/foreman/blob/develop/config/routes.rb#L19), since the route helpers were included after application helper, so it re-replaced the method by which we had replaced the original method :-)

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -204,7 +204,7 @@ module ApplicationHelper
     text_field_tag(name, val, options)
   end
 
-  def help_path
+  def help_button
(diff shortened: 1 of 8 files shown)


dLobatog:
https://github.com/search?l=&q=help_path+user%3Atheforeman+user%3Akatello&ref=advsearch&type=Code&utf8=%E2%9C%93 notice discovery and tasks use this too. I think it'd be OK to merge this along changes to the plugins and not deprecate it.

ares:
yeah, I'll send patches to both, I agree with not deprecating it

ares:
https://github.com/theforeman/foreman_discovery/pull/309 and https://github.com/theforeman/foreman-tasks/pull/207 opened

dLobatog:
Thank you @ares!

PR #3938, opened 2016-10-13
fixes #16900 - ensure interface mac address is unicast

timogoebel:
This should prevent manually entered addresses to be either multicast or broadcast.

@dLobatog : This should have saved you a lot of debugging, I guess.

### DIFF
--- a/lib/net/validations.rb
+++ b/lib/net/validations.rb
@@ -63,6 +63,21 @@ module Net
       false
     end
 
+    def self.multicast_mac?(mac)
+      return false unless validate_mac(mac)
+
+      # Get the first byte
+      msb = mac.tr('.:-', '').slice(0..1).to_i(16)
+
+      # Is least significant bit set?
+      msb & 0b1 == 1
+    end
+
+    def self.broadcast_mac?(mac)
+      return false unless validate_mac(mac)
+      mac.downcase == 'ff:ff:ff:ff:ff:ff'
+    end
+
     # validates the mac and raises an error
     def self.validate_mac!(mac)
       raise Error, "Invalid MAC #{mac}" unless validate_mac(mac)
--- a/test/factories/host_related.rb
+++ b/test/factories/host_related.rb
@@ -48,7 +48,7 @@ FactoryGirl.define do
   factory :nic_base, :class => Nic::Base do
     type 'Nic::Base'
     sequence(:identifier) { |n| "eth#{n}" }
-    sequence(:mac) { |n| "00:00:00:00:" + n.to_s(16).rjust(4, '0').insert(2, ':') }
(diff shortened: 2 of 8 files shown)


lib/net/validations.rb
```diff
@@ -63,6 +63,21 @@ def self.validate_mac(mac)
       false
     end
 
+    def self.multicast_mac?(mac)
+      return false unless validate_mac(mac)
+
+      # Get most signitifcant byte
+      msb = mac.gsub(/[\.\:\-]/, '').slice(0..1).to_i(16)
```
domcleal: Prefer tr over gsub to remove/change characters, so `mac.tr('.:-', '')` will do the same, but quicker.

lib/net/validations.rb
```diff
@@ -63,6 +63,21 @@ def self.validate_mac(mac)
...
+
+      # Is most significant bit set?
+      msb & 0b1 == 1
+    end
+
+    def self.broadcast_mac?(mac)
+      return false unless validate_mac(mac)
+      mac.gsub(/[\.\:\-]/, '').downcase == 'ffffffffffff'
```
domcleal: ditto, prefer `tr` over gsub - or even remove entirely, as `mac.downcase` can be compared to `ff:ff:ff:ff:ff:ff`

lib/net/validations.rb
```diff
@@ -63,6 +63,21 @@ def self.validate_mac(mac)
       false
     end
 
+    def self.multicast_mac?(mac)
+      return false unless validate_mac(mac)
+
+      # Get most signitifcant byte
+      msb = mac.gsub(/[\.\:\-]/, '').slice(0..1).to_i(16)
+
+      # Is most significant bit set?
```
domcleal: least significant?

test/factories/host_related.rb
```diff
@@ -65,7 +65,7 @@ def set_nic_attributes(host, attributes, evaluator)
 
   factory :nic_managed, :class => Nic::Managed, :parent => :nic_interface do
     type 'Nic::Managed'
-    sequence(:mac) { |n| "01:23:45:ab:" + n.to_s(16).rjust(4, '0').insert(2, ':') }
+    sequence(:mac) { |n| "00:23:45:ab:" + n.to_s(16).rjust(4, '0').insert(2, ':') }
```
domcleal: Can you change one of the other bytes to ensure this sequence doesn't overlap with Nic::Base's above? I think that's why the first part was changed to 01. Note that the other factory sequences use 23:56, 23:67 etc to avoid duplicating.

domcleal (changes requested):

timogoebel:
@domcleal : Thanks for your review. I have made the suggested changes.

domcleal (approved):

domcleal:
Merged, thanks @timogoebel.

PR #3939, opened 2016-10-13
fixes #16821 - store plugin permissions in AccessControl in tests

domcleal:
When initialising a new test database, the "permission" directive in a
plugin registration would skip the Foreman::AccessControl mapping so
later tests using the access control lists would fail (e.g.
AccessPermissionsTest, if a route added by a plugin to the main app
engine used a plugin permission.)

The AccessControl entry is now always created as it's safe to do so even
without a database, but the Permission resource is not created from the
initialiser in the test environment. Instead, permissions, roles and
filters are added to the database via fixtures so plugins can rely on
them for unit or functional tests.

---

Alternative to #3933, which I feel is unreliable (I spent a long time removing this hack once in f20c2a1). The issue is simply missing AccessControl mappings, which are added trivially by changing the `#permission` method for plugin registrations.

I've gone further by adding permissions and roles automatically to fixtures so plugin authors can rely on them inside the test environment too.

### DIFF
--- a/test/fixtures/filterings.yml
+++ b/test/fixtures/filterings.yml
@@ -373,3 +373,11 @@ crud_hosts_1_2:
 crud_hosts_1_3:
   filter: crud_hosts_1
   permission: view_hosts
+<% Foreman::Plugin.all.map(&:default_roles).inject({}, :merge).each do |role,permissions| %>
+<% role_id = role.tr(' ', '_').scan(/[a-z_]/i).join %>
+<% permissions.each do |permission| %>
+<%= role_id %>_<%= permission %>:
+  filter: <%= role_id %>
+  permission: <%= permission %>
+<% end %>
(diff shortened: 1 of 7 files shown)


dLobatog:
Thanks @domcleal , this fixes it without us having to care about what's in the database.

test/fixtures/filterings.yml
```diff
@@ -373,3 +373,11 @@ crud_hosts_1_2:
 crud_hosts_1_3:
   filter: crud_hosts_1
   permission: view_hosts
+<% Foreman::Plugin.all.map(&:default_roles).inject({}, :merge).each do |role,permissions| %>
```
dLobatog, line 376: I had no idea this could be done in the yml files - great job. I thought about adding these fields through a script in the helper or something but this is definitely easier

PR #3941, opened 2016-10-14
Fixes #16835 - ptables search fails PG error

amitkarsale:
API get for search on ptables with organization ID fails with PGSQL error

### DIFF
--- a/app/models/concerns/taxonomix.rb
+++ b/app/models/concerns/taxonomix.rb
@@ -14,10 +14,10 @@ module Taxonomix
              :validate => false
     after_initialize :set_current_taxonomy
 
-    scoped_search :in => :locations, :on => :name, :rename => :location, :complete_value => true
-    scoped_search :in => :locations, :on => :id, :rename => :location_id, :complete_enabled => false, :only_explicit => true
     scoped_search :in => :organizations, :on => :name, :rename => :organization, :complete_value => true
     scoped_search :in => :organizations, :on => :id, :rename => :organization_id, :complete_enabled => false, :only_explicit => true
+    scoped_search :in => :locations, :on => :name, :rename => :location, :complete_value => true
+    scoped_search :in => :locations, :on => :id, :rename => :location_id, :complete_enabled => false, :only_explicit => true
 
     dirty_has_many_associations :organizations, :locations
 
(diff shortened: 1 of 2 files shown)


domcleal:
I can't reproduce the error reported in the ticket on Foreman, can you confirm that this bug exists?

It's unclear from the commit and the commit message (which should state what changed) why this reordering of the scoped search definitions fixes anything. Please supply a bit more context about the change and the cause.

amitkarsale:
I and one of my colleague was able to reproduce the issue on foreman. 
Not sure why just re-ordering solves the issue,  we use `scope_search` gem for searching, the query formed while not reordering the scopes and ordering the scopes were different and as they are applied on the polymorphic table might have lead to form inconsistent query on scope_search gem.
I was going to ask someone from the community who has knowledge on scope_search gem regarding the behavior.

domcleal:
I see the issue now when running the unit test, but it simply moves the bug from one taxonomy type to the other by reordering those conditions - it previously passes when using `location_id`, but now fails for `location_id`. I don't think this sufficiently fixes the issue.

shlomizadok:
This fix does not solve the issue - I was able to reproduce with organization.
calling `/api/v2/ptables?search=FreeBSD&organization_id=2` would raise:
`error": {"message":"PG::UndefinedTable: ERROR:  missing FROM-clause entry for table \"organizations_templates`

* Please note that calling `/api/v2/ptables?search=FreeBSD&organization=MyOrg` works.

@amitkarsale - let me know if you plan to continue or close this.

amitkarsale:
@shlomizadok I am looking into this issue, it should also works when `id` is provided. will submit a new PR on the same

ohadlevy:
@amitkarsale what is the status of this pr? thanks.

ares:
@amitkarsale was there any progress?

amitkarsale:
will update on this shortly. Thanks

tbrisker:
I can't seem to reproduce with current develop, does this only occur when certain plugins are enabled or was this resolved by some other fix?

tbrisker:
never mind, managed to reproduce.

tbrisker:
Closing after discussion with @amitkarsale.

PR #3943, opened 2016-10-16
fixes #16951 - ipv6 compute orchestration

timogoebel:


timogoebel:
This can be tested with https://github.com/theforeman/foreman-digitalocean/pull/16

app/models/concerns/orchestration/compute.rb
```diff
@@ -124,7 +128,7 @@ def setComputeDetails
       attrs.each do |foreman_attr, fog_attr|
         if foreman_attr == :mac
           return false unless match_macs_to_nics(fog_attr)
-        elsif foreman_attr == :ip
+        elsif [:ip, :ip6].include?(foreman_attr)
           value = vm.send(fog_attr) || find_address
```
domcleal: `find_address` isn't IPv6 aware here, it'll call `vm.ip_addresses` which can return v4 addresses if the compute claims to provide v6 but `vm.send(fog_attr)` returns nil. You can reproduce this by _not_ ticking the IPv6 box when using fog-digitalocean.

It probably needs to be changed to filter addresses from `vm.ip_addresses` depending on their type?

domcleal (changes requested):
find_address causes problems when IPv6 is not active on a v6-capable provider.

timogoebel:
@domcleal : The code is more reliable now. `find_address` filters the ip addresses by type and returns early if any other ip address is already set. I also removed the `delCompute` calls from `validate_forman_attr` as they would eventually try to delete the vm twice.

[test foreman]
[test katello]
[test upgrade]

domcleal:
[test foreman]

app/models/concerns/orchestration/compute.rb
```diff
@@ -124,16 +128,21 @@ def setComputeDetails
...
         else
           value = vm.send(fog_attr)
           self.send("#{foreman_attr}=", value)
-          return false unless validate_foreman_attr(value, Host, foreman_attr)
+          return false unless validate_required_foreman_attr(value, Host, foreman_attr)
         end
       end
+
+      if ip.blank? && ip6.blank? && (compute_provides?(:ip) || compute_provides?(:ip6))
+        return failure(_("failed to acquire ip addresses from compute resource for %s") % name)
```
domcleal: Capitalise "Failed" and "IP", important in translated (and therefore user-visible) strings.

app/models/concerns/orchestration/compute.rb
```diff
@@ -246,12 +263,17 @@ def find_address
...
-      logger.info "acquisition of ip address timed out, using #{ip}"
+      ip = filter_ip_addresses(vm.ip_addresses, type).first if ip.blank?
+      logger.info "acquisition of #{type} address timed out, using #{ip}"
     end
     ip
   end
 
+  def filter_ip_addresses(addresses, type)
+    check_method = type == :ip6 ? :ipv6? : :ipv4?
+    addresses.map {|ip| IPAddr.new(ip) rescue nil}.compact.select { |ip| ip.send(check_method) }.map(&:to_s)
```
domcleal: `select { |ip| ip.send(check_method) }` can be written as `select(&check_method)`

app/models/concerns/orchestration/compute.rb
```diff
@@ -124,16 +128,21 @@ def setComputeDetails
       attrs.each do |foreman_attr, fog_attr|
         if foreman_attr == :mac
           return false unless match_macs_to_nics(fog_attr)
-        elsif foreman_attr == :ip
-          value = vm.send(fog_attr) || find_address
+        elsif [:ip, :ip6].include?(foreman_attr)
+          value = vm.send(fog_attr) || find_address(foreman_attr)
           self.send("#{foreman_attr}=", value)
           return false unless validate_foreman_attr(value, ::Nic::Base, foreman_attr)
```
domcleal: If the provider claims to provide :ip6 but it doesn't (e.g. foreman_digitalocean, don't tick IPv6), then this still throws an error if `value.blank?` and there's another blank value in the database. Don't call it if the value's blank?

```
2016-11-01T14:36:30 c50b235a [sql] [W] ip6  is already used by 
```

test/models/orchestration/compute_test.rb
```diff
@@ -23,6 +23,15 @@ class ComputeOrchestrationTest < ActiveSupport::TestCase
       assert(@host.send :setComputeDetails)
     end
 
+    test "are set for CR providing IPv6" do
+      an_ip6 = '2001:db8::1'
+      @cr.stubs(:provided_attributes).returns({:ip6 => :ip6})
+      @host.vm.expects(:ip6).returns(an_ip6)
```
domcleal, line 39: A test that returns a blank value for :ip6 while claiming to provide ip4 + ip6 would be a good idea.

domcleal (changes requested):

app/models/concerns/orchestration/compute.rb
timogoebel: Oh, sorry. I have tested this with a fresh install and an empty database and it worked fine. I will make the necessary adjustments and add a test case for this.

timogoebel:
@domcleal : Thanks for giving this another spin. I have fixed the issue and tested with more hosts. It should work fine now. Also improved the test coverage of this.

domcleal (approved):

domcleal:
Merged, thanks @timogoebel.

PR #3948, opened 2016-10-17
fixes #16952 - pass type to proxy when destroying dns records

timogoebel:
Theoretically we could also pass the record type for reverse records. However smart proxy has code to detect this automatically and I prefer to keep this change minimal.


domcleal:
Merged, thanks @timogoebel!

PR #3949, opened 2016-10-17
Fixes #16970 - added list of Rails config files

lzap:
Side effect is providing list of foreman hooks installed, which was the main
motivation for the patch.


domcleal:
Under what circumstances is a list of packaged config files important? If you want to include a list of foreman_hooks scripts, please add it to the plugin instead, not Foreman.

lzap:
Do you prefer package only change?

domcleal:
I don't think I understand the question?

lzap:
If you prefer the debug script in the main foreman_hooks git repo, or just in RPM/DEB repos.

domcleal:
In the plugin repo.

lzap:
Sure https://github.com/theforeman/foreman_hooks/pull/40

PR #3950, opened 2016-10-18
Fixes #16990 - make access_perm test reusable

jlsherrill:


jlsherrill:
Using this I can easily add the same check to a plugin such as katello:  https://github.com/Katello/katello/pull/6403/files#diff-0997358e45358776de1e730a5e19cf4bR1

test/unit/foreman/access_permissions_test.rb
```diff
@@ -10,59 +11,8 @@
 #
 # In particular, it's important that actions for AJAX requests are added to
 # an appropriate permission so views using those requests function.
-class AccessPermissionsTest < ActiveSupport::TestCase
-  MAY_SKIP_REQUIRE_LOGIN = [
-    "users/login", "users/logout", "users/extlogin", "users/extlogout", "home/status", "notices/destroy",
+class AccessPermissionTest < ActiveSupport::TestCase
```
domcleal: Why rename this? It should still be plural I think to match the class and filename.

test/unit/foreman/access_permissions_test.rb
```diff
@@ -10,59 +11,8 @@
 #
 # In particular, it's important that actions for AJAX requests are added to
 # an appropriate permission so views using those requests function.
-class AccessPermissionsTest < ActiveSupport::TestCase
-  MAY_SKIP_REQUIRE_LOGIN = [
```
domcleal, line 14: The constants should probably remain in this test class, they're part of the test (linked to this app) rather than the reusable bit.

test/unit/foreman/access_permissions_test_base.rb
```diff
@@ -0,0 +1,69 @@
+module AccessPermissionsTestBase
```
domcleal: The path should probably be test/unit/shared/access_permissions_test_base.rb to match other common bits of test code.

domcleal (changes requested):

test/unit/foreman/access_permissions_test.rb
jlsherrill: Whoops, that was an accident. Fixed!

test/unit/foreman/access_permissions_test_base.rb
jlsherrill: Done

test/unit/foreman/access_permissions_test.rb
jlsherrill: Moving the constants were actually the whole reason I opened the PR. 

Without any of these changes, I could still test /my_plugin/ routes easily, however if I add any routes to the existing foreman routes I can't easily choose to test all the foreman routes because I need access to these constants.  'requiring'  the main test file (as it is in master) made those tests to get executed due to the way they were rewritten, and would fail since I haven't added my own exceptions yet.  The only options I saw were:

1) Move the constants to some other file that could be required and re-used (which is what is here)
2) Rewrite the tests so that they didn't declare one test per route.
3) Just redefine them in my plugin

1) seemed like the best route to me, but I'm happy to look at other suggestions

domcleal: http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Skipping-tests is provided if you want to exclude certain routes being tested, exclude the test.

Moving the constants will still cause the tests to be run from this file against the existing constants, it won't allow a plugin author to add additional constants. (The tests defined by this file will continue to fail.)

jlsherrill: Correct, but it does allow me to write my own test and call these constants and add my own routes, such as in the commit i mentioned:  https://github.com/Katello/katello/pull/6403/files#diff-0997358e45358776de1e730a5e19cf4bR45

If i skip the test completely (and don't run my own with my added routes), i'm not checking that my added routes within the foreman routes are fully 'permissioned'.  Do you have another suggestion to accomplish this?

domcleal: > If i skip the test completely (and don't run my own with my added routes), i'm not checking that my added routes within the foreman routes are fully 'permissioned'.

Rather than skipping the _complete_ test, only skip the individual tests for routes that you want to exclude (for whatever legitimate reason). This way the AccessPermissionTest will always pass. Needing to redefine the test to add skips/exclusions means the _original_ test has to be disabled to prevent failures, which shouldn't be necessary.

jlsherrill:
@domcleal thanks for the feedback, updated

test/unit/foreman/access_permissions_test.rb
```diff
@@ -1,6 +1,7 @@
 require 'test_helper'
 require 'foreman/access_control'
```
domcleal: These foreman/access_\* requires should be in the shared file, making it more reusable (as that's where they're referenced).

domcleal (changes requested):

domcleal (approved):

test/unit/foreman/access_permissions_test.rb
domcleal: Thanks for fixing, but please reply/leave a comment so the reviewer knows it's changed.

domcleal:
Merged, thanks @jlsherrill.

PR #3951, opened 2016-10-18
fixes #16996 - reduce eager_load of tables

schurzi:
Not eager_loading all data speeds up response time and reduces ressource consumption in larger environments. (actually makes it work with our environment)
In smaller environments there may be a negative performance impact, but this should be negligible.

additional tests seem to be not needed, because no new functionality is introduced

### DIFF
--- a/app/controllers/puppetclasses_controller.rb
+++ b/app/controllers/puppetclasses_controller.rb
@@ -9,9 +9,7 @@ class PuppetclassesController < ApplicationController
   before_action :setup_search_options, :only => :index
 
   def index
-    eager_load_tables = [:config_group_classes, :class_params, :environments,
-                         :hostgroups]
-    @puppetclasses = resource_base_search_and_page(eager_load_tables)
+    @puppetclasses = resource_base_search_and_page
     @hostgroups_authorizer = Authorizer.new(User.current, :collection => HostgroupClass.where(:puppetclass_id => @puppetclasses.map(&:id)).uniq.pluck(:hostgroup_id))
   end
 
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -48,6 +48,8 @@ Foreman::Application.configure do
     Bullet.rails_logger = true
     Bullet.add_footer = true
     Bullet.counter_cache_enable = false
+    Bullet.add_whitelist :type => :n_plus_one_query, :class_name => "Puppetclass", :association => :environments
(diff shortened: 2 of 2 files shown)


mmoll:
ok to test

app/controllers/puppetclasses_controller.rb
```diff
@@ -9,9 +9,7 @@ class PuppetclassesController < ApplicationController
   before_action :setup_search_options, :only => :index
 
   def index
-    eager_load_tables = [:config_group_classes, :class_params, :environments,
-                         :hostgroups]
```
domcleal, line 13: I think `:hostgroups` is worth keeping in here. Can you check there's no performance impact for you with it?

The others definitely seem to have a negative effect on performance as the eagerly loaded versions aren't used (e.g. the info about class_params isn't used, only a count), so this is otherwise correct.

domcleal (changes requested):
Removing `class_params` causes an alert from [Bullet](https://github.com/flyerhzm/bullet) that there's an N+1 query. Can you please add a whitelist entry to config/environments/development.rb per https://github.com/flyerhzm/bullet#whitelist to avoid it?

```
 | N+1 Query detected
 |   Puppetclass => [:class_params]
```

Sorry for the slight delay reviewing your PR.

app/controllers/puppetclasses_controller.rb
schurzi: After some quick tests I can see no performance difference when including `:hostgroups`. For `:environmens` there is also no impact. 

I have given this some additional thought. The view displays 20 records by default (can this even be changed?), so 90% of (eager) loaded data is thrown away in any case, since the generated query does not include a `LIMIT` clause. I really think any eager loading here is not necessary. I'm always thinking of the old proverb "premature optimization is the root of all evil". So I would argue that we need to show a case where eager loading noticeably increases performance.

But if you say so, i will re-add the loading for `:hostgroups`. :)

domcleal: I thought it reduced the SQL query count, but I've re-checked the logs and they're identical, so it can be left out. Apologies. The comment regarding Bullet stands, but now for both detected missing eager loads.

domcleal (approved):

domcleal:
Thanks for the update, please just reply or add a comment when a PR changes so the reviewer knows to look at it again.

domcleal:
Merged, thanks @schurzi.

PR #3954, opened 2016-10-19
[WIP] Fixes #16982 - "No taxonomies" fix for Host default scope

dLobatog:
The default scope for hosts did not restrict properly by taxonomies.
An user without organizations or locations, could do anything it's
permissions allow to. The list of hosts was unrestricted and showed
hosts in any location or organization.

This is fixed to work so that:
- Users without taxonomies, when set to 'any context' cannot see
  anything
- Users with taxonomies, when set to 'any context' can see everything
  within all of their taxonomies context.
- Admins set to 'any context' can see everything - regardless of
  whether it has a taxonomy or not.
- Users or admins set to some organization/location scope can only
  see stuff within scope.

### DIFF
--- a/app/controllers/unattended_controller.rb
+++ b/app/controllers/unattended_controller.rb
@@ -122,7 +122,7 @@ class UnattendedController < ApplicationController
     if (result = token.match(/^([a-z0-9-]+)(.slax)$/i))
       token, _suffix = result.captures
     end
-    Host.for_token(token).first
+    Host.unscoped.for_token(token).first
   end
 
   def find_host_by_ip_or_mac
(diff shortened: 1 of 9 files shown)


dLobatog:
Hmm - I think the way it should work should rather be:
- If current taxonomy type is set -> restrict by that
- If current taxonomy type is NOT set:
  - Restrict by `user.taxonomy_type.map(&:subtree_ids)`
  - Add hosts w/o taxonomies for admin users.

This is still a bit fuzzy and might require changes on UnattendedController and certain tests who expect that 'no taxonomies' == 'can see everything'

PR #3955, opened 2016-10-19
Fixes #16971 - CVE-2016-7077 remove unauthorized checkboxes

ares:
This is one way to fix it - add authorization for checkboxes including hidden field to keep existing associations that user can't modify. The other way would be removing the checkboxes completely and always use multiple_selects, which I find better and more safe.

I haven't find any recommendation for this in patternfly, @rohoover your input would be appreciated. When we have n:m associations we use checkboxes if there are less than 6 objects, and more sophisticated multiple selector otherwise. Would it be better/worse if we used only one form element, the multiple select one?

Illustration of current form elements for less than 6 options
checkboxes1

for 6 and more
multiple1

possible change to have the same thing for less items too
multiple2


Rohoover:
@ares One other option that may work for both use case scenarios found in PatternFly, is the multiple select option. 
http://www.patternfly.org/pattern-library/widgets/#bootstrap-select

screen shot 2016-10-19 at 8 01 21 am

Thoughts?

ares:
Thanks for prompt response. I like the patternfly multiple select bootstrap-select also seems to support live search and select all features. Since this is security fix, we'd like to keep the patch minimal, which means I'd like to avoid changing all forms in Foreman in this PR. That does not mean we can't open a separate issue to migrate to patternfly multiple select if others agree. If I read it correctly, you also suggest using one widget for any number of items, correct?

Rohoover:
@ares In the name of consistency, I would prefer to use one widget for the same type of interaction, but also understand that incremental steps may be necessary for full implementation. Fixing it in this one instance, and opening another PR for all others sounds reasonable. Thanks!

ares:
Thanks Roxanne! I'll change this PR so we use only one of two widgets we have which gets us closer to migrate to patternfly widget too.

dLobatog:
@ares Tests are failing, other than that :+1: at first glance, all this does is to remove the paths for `< 6` elements. Thanks !

ares:
I fixed the integration tests.

PR #3956, opened 2016-10-20
Fixes #17035 - removes webpack provider plugin

gailsteiger:
Webpack is currently configured to automatically load 'jquery' when the '$' or 'jquery' is used as free variable in a module. The identifier is filled with the exports of the loaded module.

This practice is not sufficient for our testing infrastructure.

### DIFF
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -63,10 +63,6 @@ var config = {
       modules: false,
       assets: true
     }),
-    new webpack.ProvidePlugin({
-      $: 'jquery',
-      jQuery: 'jquery',
-    }),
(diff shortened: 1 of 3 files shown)


dLobatog:
Thanks @gailsteiger ! I noticed `react_app/common/MountingService.js` uses `$` in the else condition however there is no explicit import of jquery there - I think there should be one there, let me know if I'm missing something. Would be glad to merge after that.

gailsteiger:
@dLobatog 
`$` here `${componentName}` is an ES6 template string interpolation, not a jquery call.

No import statement is needed.

Nice pickup though.

gailsteiger:
@dLobatog 
this is the contributor
what is it waiting on me for?

dLobatog:
Thanks @gailsteiger !

PR #3958, opened 2016-10-20
fixes #17041 - skips orchestration when importing facts

timogoebel:
This is a regression in 1.13. The orchestration callbacks only run if `unless :skip_orchestration?`, however `skip_orchestration?` is not a method for `Nic::Managed` causing the callbacks to always run, even if importing facts.

This leads to a chain reaction: Smart Proxy gets unresponsive, all Foreman threads start to wait for a response from Smart Proxy and the UI gets completely unresponsive.

### DIFF
--- a/app/models/concerns/orchestration.rb
+++ b/app/models/concerns/orchestration.rb
@@ -99,11 +99,15 @@ module Orchestration
   end
 
   def skip_orchestration?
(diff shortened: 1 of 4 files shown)


domcleal:
Merged as a390cfb, thanks @timogoebel.

PR #3959, opened 2016-10-20
Fixes #17042 - improved orchestration logging

lzap:
This improves orchestration logging, failed tasks were silently ignored (we
only logged those which failed due exception but not these that return false).

The patch adds this logging pattern:

```
[I] Processed 6/6 orchestration tasks
[D] Task Remove IPv4 DNS record for srv-192-168-122-148.local.lan *completed*
[D] Task Remove Reverse IPv4 DNS record for srv-192-168-122-148.local.lan *completed*
[D] Task Remove DHCP Settings for srv-192-168-122-148.local.lan *completed*
[D] Task Delete TFTP PXEGrub2 config for srv-192-168-122-148.local.lan *completed*
[D] Task Delete TFTP PXELinux config for srv-192-168-122-148.local.lan *completed*
[D] Task Delete TFTP PXEGrub config for srv-192-168-122-148.local.lan *completed*
```

Number of successful tasks (6) out of (6) is logged as INFO level, all tasks
are logged by their name along with the status (DEBUG only level).

### DIFF
--- a/app/models/concerns/orchestration.rb
+++ b/app/models/concerns/orchestration.rb
@@ -151,6 +151,18 @@ module Orchestration
     fail_queue(q)
 
     rollback
+  ensure
+    unless q.nil?
+      logger.info("Processed queue '#{queue_name}', #{q.completed.count}/#{q.all.count} completed tasks") unless q.empty?
+      q.all.each do |task|
+        msg = "Task #{task.name} *#{task.status}*"
+        if task.status?(:completed) || task.status?(:pending)
+          logger.debug msg
+        else
+          logger.error msg
+        end
+      end
+    end
   end
 
   def fail_queue(q)
--- a/app/services/orchestration/task.rb
+++ b/app/services/orchestration/task.rb
@@ -19,6 +19,10 @@ module Orchestration
       end
     end
 
+    def status?(symbol)
+      @status == symbol.to_s
+    end
+
     def to_s
       "#{name}\t #{priority}\t #{status}\t #{action}"
     end


domcleal:
@lzap there appear to be a lot of test failures, so I'm leaving this on WoC.

lzap:
Should be fixed now, also removed the `result` helper, `ensure` block should not modify the return value as I tested on Ruby 2.2.

app/models/concerns/orchestration.rb
```diff
@@ -151,6 +151,18 @@ def process(queue_name)
     fail_queue(q)
 
     rollback
+  ensure
+    unless q.nil?
+      logger.info("Processed #{q.completed.count}/#{q.all.count} orchestration tasks") if q.all.count > 0
+      q.all.each do |task|
+        msg = "Task #{task.name} *#{task.status}*"
+        if task.status != "completed"
```
domcleal: Should this also exclude pending tasks? I don't think they shouldn't be considered as errors.

app/models/concerns/orchestration.rb
```diff
@@ -151,6 +151,18 @@ def process(queue_name)
     fail_queue(q)
 
     rollback
+  ensure
+    unless q.nil?
+      logger.info("Processed #{q.completed.count}/#{q.all.count} orchestration tasks") if q.all.count > 0
```
domcleal: 1. Add the queue_name here too, there are two.
2. `if q.all.count > 0` => `unless q.empty?`

domcleal (changes requested):

lzap:
Done, example output:

```
14:25:38 rails.1   | [sql|D] Using host queue
14:25:38 rails.1   | [sql|I] Processed queue 'queue', 8/8 completed tasks
14:25:38 rails.1   | [sql|D] Task DHCP conflicts removal for fake7716c1.local.lan *completed*
14:25:38 rails.1   | [sql|D] Task Create DHCP Settings for fake7716c1.local.lan *completed*
14:25:38 rails.1   | [sql|D] Task Create IPv4 DNS record for fake7716c1.local.lan *completed*
14:25:38 rails.1   | [sql|D] Task Create Reverse IPv4 DNS record for fake7716c1.local.lan *completed*
14:25:38 rails.1   | [sql|D] Task Deploy TFTP PXEGrub2 config for fake7716c1.local.lan *completed*
14:25:38 rails.1   | [sql|D] Task Deploy TFTP PXELinux config for fake7716c1.local.lan *completed*
14:25:38 rails.1   | [sql|D] Task Deploy TFTP PXEGrub config for fake7716c1.local.lan *completed*
14:25:38 rails.1   | [sql|D] Task Fetch TFTP boot files for fake7716c1.local.lan *completed*
```

app/models/concerns/orchestration.rb
```diff
@@ -151,6 +151,18 @@ def process(queue_name)
     fail_queue(q)
 
     rollback
+  ensure
+    unless q.nil?
+      logger.info("Processed queue '#{queue_name}', #{q.completed.count}/#{q.all.count} completed tasks") unless q.empty?
+      q.all.each do |task|
+        msg = "Task #{task.name} *#{task.status}*"
+        if task.status !~ /^(completed|pending)$/
```
domcleal: No need to add a regexp in here, use a simple boolean please, or `if ['completed', 'pending'].include?(task.status)` (and flip the order of the conditional is the matching block, then the false block instead of negating it).

domcleal (changes requested):

app/models/concerns/orchestration.rb
lzap: Sure, I created helper so it's more readable.

domcleal (approved):

domcleal:
Merged, thanks @lzap.

PR #3960, opened 2016-10-21
Fixes #17060 - Add hammer ping and katello-service status commands to foreman-debug

ahumbe:


domcleal:
These are not commands present in Foreman or Hammer itself, please submit this change to the Katello project.

PR #3961, opened 2016-10-21
Fixes #16982 - CVE 2016-7078 - Scope properly when no taxonomies are set

dLobatog:
The default scope for hosts and other objects did not restrict
properly by taxonomies. An user without organizations or
locations, could do anything it's permissions allow to.
The list of hosts was unrestricted and showed hosts in
any location or organization.

This is fixed to work so that:

Users without taxonomies, when set to 'any context' cannot see
anything, only non-taxable objects.

Users with taxonomies, when set to 'any context' can see
everything within all of their taxonomies context (including
children taxonomies).

Admins set to 'any context' can see everything, regardless
of whether it has a taxonomy or not.

Users or admins set to some organization/location scope
can only see stuff within scope.

---

Problems with this change: a LOT of our tests depend on 'no taxonomies' behavior being broken as it used to be. There are still some tests that need changing. If you feel like the changes are not so many, feel free to edit this PR (it's set to allow edits from maintainers). I'm leaving for a week since now so I doubt I'll have time to look at this PR.

If you think the approach is wrong feel free to solve in any other way. I think the good thing about this is that it'll force us to be more explicit about taxonomies in tests.

@jlsherrill @ehelms @thomasmckay (can't mention @katello team in here :/) this is likely to affect Katello tests a lot as well.


dLobatog:
Done with the Foreman part - I'll try to fix Katello tests as well, help appreciated though.

app/models/concerns/taxonomix.rb
```diff
@@ -32,10 +32,11 @@ module ClassMethods
 
     # default inner_method includes children (subtree_ids)
     def with_taxonomy_scope(loc = Location.current, org = Organization.current, inner_method = :subtree_ids)
+      scope = block_given? ? yield : where('1=1')
```
tbrisker: Any reason to use `where('1=1')` rather then `where(nil)`?

dLobatog: None, just that I thought it'd be easier to read

dLobatog:
You can find the Katello part of the story in https://github.com/Katello/katello/pull/6447 - just a few tests were relying on non-admin users able to see 'unscoped' objects.

app/models/concerns/taxonomix.rb
```diff
@@ -87,11 +88,19 @@ def taxable_ids(loc = which_location, org = which_organization, inner_method = w
                           inner_select(org, inner_method)
                         end
       end
-      inner_ids   = inner_ids_loc & inner_ids_org if (inner_ids_loc && inner_ids_org)
+      inner_ids = inner_ids_loc & inner_ids_org if (inner_ids_loc && inner_ids_org)
       inner_ids ||= inner_ids_loc if inner_ids_loc
       inner_ids ||= inner_ids_org if inner_ids_org
-      # In the case of users we want the taxonomy scope to get both the users of the taxonomy and admins.
-      inner_ids.concat(admin_ids) if inner_ids && self == User
+      return nil if inner_ids.nil? && User.current.try(:admin?)
```
tbrisker: Would it make sense to create the shortcut for admin in any context earlier? 
i.e. first line in the function, something like `return nil if loc.nil? && org.nil? && User.current.try(:admin?)` so it is clear `nil` is only returned in one specific case?  (which might also make sense to comment that return nil == everything)
Then we could also simplify the logic, just have an array `inner_ids` that we push into and return. Let me know if i'm not clear.

test/controllers/api/v2/config_reports_controller_test.rb
```diff
@@ -48,9 +48,10 @@ def test_create_duplicate
       Setting[:require_ssl_smart_proxies] = false
 
       proxy = smart_proxies(:puppetmaster)
+      proxy.update_attribute(:url, 'http://configreports.foreman')
```
tbrisker, line 51: is this (and similar changes in other tests) related?

test/controllers/subnets_controller_test.rb
```diff
@@ -26,33 +26,34 @@ def test_create_valid_with_type
   end
 
   def test_update_invalid
+    # Find a way to raise the point where User.current changes
```
tbrisker: well, did you find?

tbrisker (changes requested):
In general looks good, added a few inline comments. Also, how about adding some more tests on the changed taxonomix functions to make sure we don't break this again?

test/controllers/subnets_controller_test.rb
dLobatog: Ah forget about that, it was a reminder to myself

test/controllers/api/v2/config_reports_controller_test.rb
dLobatog: Yeah, without it `auth_smart_proxies` in the controller will find other proxies with the feature that have 'else.where' in the name, and those might not necessarily have the Puppet feature

dLobatog:
Latest update only puts the 'return nil if admin' condition at the beginning of the `taxable_ids` method and changes `where 1=1` to `where nil`.

tbrisker:
I created https://github.com/dLobatog/foreman/pull/29 to explain what I meant for simplifying `taxable_ids`. Also, I'm still missing some new tests to make sure this method doesn't break again.

dLobatog:
@tbrisker Thanks for the PR, I've tested it and I don't think it's equivalent in functionality when you have both organizations and locations enabled: https://github.com/dLobatog/foreman/pull/29/files#r87873373

As for testing - as you can see a long list of tests will break if this breaks again, but I'll include a few more that specifically test returning nil when no orgs/locs are selected (and taxable_ids is called from admin) and for returning User.current along the list of users.

dLobatog:
Updated with extra tests!

test/models/taxonomix_test.rb
```diff
@@ -260,4 +261,41 @@ def setup
...
+    test "returns only visible objects when org/loc are selected" do
+      scoped_environments = Environment.
+        with_taxonomy_scope([taxonomies(:organization1)])
+      assert scoped_environments.include?(*taxonomies(:organization1).environments)
+      assert_not_equal Environment.unscoped.all, scoped_environments
+      assert_equal taxonomies(:organization1).environments, scoped_environments
+    end
+
+    test "returns nil (all objects) when there are no org/loc" do
+      assert_equal User.with_taxonomy_scope([],[]).sort, User.unscoped.all.sort
```
tbrisker, line 320: Shouldn't we also test that for a non-admin user empty taxonomy scope returns nothing?

dLobatog: Will do in a moment

dLobatog:
@tbrisker Updated - the test ('can work with array of taxonomies') covers the case for checking the visibility of stuff that's in the current organization but not in the current location. I wrote 'returns no IDs for non-admin user' to check that it doesn't return all objects for a regular user without taxonomies.

tbrisker (approved):

tbrisker:
LGTM, but I would be happy if another maintainer who is familiar with taxonomix/permissions could take a quick look just to be sure. Also, do we need the katello changes to get merged before or after this?

dLobatog:
Katello stuff should be merged after

dLobatog:
Repushed because a test from e18d69e89b544288d90926011d7bb8b1d46a182b was relying on the CVE to work - it's now fixed on `test/controllers/puppetclasses_controller_test.rb`

app/models/concerns/taxonomix.rb
```diff
@@ -87,11 +92,18 @@ def taxable_ids(loc = which_location, org = which_organization, inner_method = w
...
       inner_ids ||= inner_ids_org if inner_ids_org
-      # In the case of users we want the taxonomy scope to get both the users of the taxonomy and admins.
-      inner_ids.concat(admin_ids) if inner_ids && self == User
+
+      if self == User
+        # In the case of users we want the taxonomy scope to get both the users
+        # of the taxonomy, admins, and the current user.
+        inner_ids ||= []
+        inner_ids.concat(admin_ids)
+        inner_ids << User.current.id if User.current.present?
```
ares: calling `uniq` might remove some dupes, oth SQL can deal with it too

ares:
visually looks good, I'd like to give it a try tomorrow, this should be covered in released notes since it will have impact on users that only enabled orgs/locs but don't use them, regular users won't see anything after upgrade

ares:
> Users without taxonomies, when set to 'any context' cannot see
anything (at all)

I don't think this is entirely correct. User in any context can see resources that are not assigned to any orgnaization/location or are not even taxable. E.g. such user can see all unassigned provisioning templates if he or she has permission. That I could understand. With this patch applied it seems that user can see even more, e.g. hostgroup that is assigned to org A (user has no org/loc assigned).

Also it seems weird that with this patch, user in any context doesn't see resources from a specific taxonomies which he's assigned to. I have host group in org A, user is in org A, C (there's also B). In any context he does not see the host group, after choosing org A, the user can see it.

dLobatog:
> Also it seems weird that with this patch, user in any context doesn't see resources from a specific taxonomies which he's assigned to. I have host group in org A, user is in org A, C (there's also B). In any context he does not see the host group, after choosing org A, the user can see it.

@ares Have you tried that with the latest version of this PR? I tried the following to reproduce that:

* User "robotest" in Location 'Spain', Organization 'Lobacorp" , Role 'viewer' - can see all Hostgroups in those taxonomies when visiting /hostgroups ("Any context" selected) - ✔️ 
* Create a Hostgroup in location 'Czech Republic', organization 'E-Corp' - it can't be seen from user "robotest" - ✔️ 
* Add Location "Czech Republic" and Organization "E-Corp" to User "robotest" . Refresh the page ("Any context" selected) - hostgroups in both Czech/Spain Lobacorp/E-Corp are displayed ✔️ 

All of these sound reasonable to me, is anything wrong there?

About the former part "User without taxonomies in any context can see resources that are not assigned to any orgnaization/location or are not even taxable". I partly agree:

1. Things that are not taxable are definitely visible to users without taxonomies. I do agree with that.
2. Things that are taxable, but have *no taxonomies assigned*, should not be visible to any users other than Admin. I don't agree with what you said that "no taxonomies assigned" means "visible to users without taxonomies" AND "not visible to users with taxonomies". 

IIRC the convention is that when taxonomies are enabled, "no taxonomies assigned" means "admins only". I think there was an old CVE that had to do with this precisely, I'll look it up.

PS: I found yet another security issue when testing all of the above 😞 I'll ping you to check foreman-security in a few hours

dLobatog:
Hmmm found a bug regarding users with Orgs but not Locs or viceversa. Feel free to not review until that's fixed.

ares:
> @ares Have you tried that with the latest version of this PR?

yes

> Things that are taxable, but have no taxonomies assigned, should not be visible to any users other than Admin. I don't agree with what you said that "no taxonomies assigned" means "visible to users without taxonomies" AND "not visible to users with taxonomies".

I also think such users should not see such resources. The patch does not seem to change the fact that they do though. I think it was never meant to work like this and no one thought too much about non-admin users with no taxonomies. I wonder why someone would create a user without taxonomies, only to manage non-taxable resources? I think it would be better to forbid such users and require some orgs/locs if they are enabled.

dLobatog:
https://github.com/theforeman/foreman/pull/4070 - completes this basically, although some tests need to be fixed from a recent update

ares:
Sorry, it needs rebasing, also a lot of tests seem to fail with the latest version. I forgot to add results of our conversation here but I think we agreed also on adding a migration to create "Global Organization" for resources and users that are unassigned and adding a validation so that non-user can't be saved without any taxonomies. Or is that meant for a separate PR?

dLobatog:
@ares Seems like a separate issue, so separate PR (It'd be easier to review, this one is huge for a 'minor' change) - I already submitted this fix we talked about yesterday separately https://github.com/theforeman/foreman/pull/4070

I'll fix the new broken tests which I think stem from this line I added yesterday to fix the bug you mentioned in your previous comment https://github.com/theforeman/foreman/pull/3961/files#diff-501156756cdcbc510254e30f9e2a29daR102

ares:
Ok, I'm fine with separate PR but I'd like to get it in the same Foreman release and now when 1.14 is close I'm not sure whether it's good idea to merge it without? Even though I'd love to get this into 1.14.

tbrisker:
@dLobatog - this needs rebasing
@ares  - is there any reason not to merge after rebase?

ares:
Nope, the only concern I had was that both PRs should get in to the same release. The second PR is doing something else that we discussed here but that does not block this one...

dLobatog:
https://github.com/theforeman/foreman_remote_execution/pull/217 fixes related tests in REX.

dLobatog:
WoC until I fix the recent test breakages, this PR is quite tedious to maintain as new behavior/tests can easily break it.

dLobatog:
WoC until I fix the recent test breakages, this PR is quite tedious to maintain as new behavior/tests can easily break it.

dLobatog:
Seems like tests are 'randomly' failing - will check out why.

dLobatog:
Hound went crazy asking for double quotes everywhere, I removed the comments and fixed that on another PR

app/models/concerns/taxonomix.rb
```diff
@@ -69,28 +70,52 @@ def enforce_default
...
-                        end
+      # Return everything (represented by nil), including objects without
+      # taxonomies. This value should only be returned for admin users.
+      return nil if any_context?(loc) && any_context?(org) &&
+        User.current.try(:admin?)
+
+      inner_ids_loc = inner_ids(loc, Location, inner_method) if SETTINGS[:locations_enabled]
+      inner_ids_org = inner_ids(org, Organization, inner_method) if SETTINGS[:organizations_enabled]
+
+      inner_ids = inner_ids_loc & inner_ids_org
```
tbrisker: This will fail if only one taxonomy type is enabled as it will try to do `[...] & nil`. 
Also, using inner_ids for both a variable and a method is confusing, how about just calling it `ids`?

How about going at this the other way - start with all the ids and remove those that don't match:
```ruby
ids = unscoped.pluck(:id)
ids &= inner_ids(loc, Location, inner_method) if SETTINGS[:locations_enabled]
ids &= inner_ids(org, Organization, inner_method) if SETTINGS[:organizations_enabled]
```

app/models/concerns/taxonomix.rb
```diff
@@ -69,28 +70,52 @@ def enforce_default
...
     end
 
+    # Returns the IDs available for the passed context.
+    # Passing a nil or [] value as taxonomy equates to "Any context".
+    # Any other value will be understood as 'IDs available in this taxonomy'.
+    def inner_ids(taxonomy, taxonomy_class, inner_method)
+      return unscoped.pluck("#{table_name}.id") if taxonomy_class.ignore?(to_s)
+      return inner_select(taxonomy, inner_method) if taxonomy.present?
+
+      if any_context?(taxonomy) && User.current.present?
```
tbrisker: `any_context?(taxonomy)==true` here always, because if `taxonomy.present?` we already returned. Therefore, we could also short-circuit and return `[]` if no user found and get rid of this whole if block

dLobatog:
Latest update just fixes a couple of tests in api/v2/hosts_controller and seeds

app/models/concerns/taxonomix.rb
```diff
@@ -69,28 +70,52 @@ def enforce_default
...
+        # Any available taxonomy to the current user
+        return unscoped.pluck("#{table_name}.id") if User.current.admin?
+
+        taxonomy_relation = taxonomy_class.to_s.underscore.pluralize
+        any_context_taxonomies = User.current.public_send(:"#{taxonomy_relation}")
+        return any_context_taxonomies.map(&:"#{table_name}").flatten.uniq.map(&:id)
```
tbrisker: This should also include the child taxonomies, and could be improved:
```ruby
taxonomy_relation = taxonomy_class.to_s.underscore.pluralize
taxonomy_ids = User.current.taxonomy_and_child_ids(:"#{taxonomy_relation}")
return self.unscoped.joins(taxonomy_join_table).where('#{taxonomy_join_table}.taxonomy_id' => taxonomy_ids).pluck(:id)
```

tbrisker (changes requested):
I think I found a couple of bugs and some improvement suggestions inline.

dLobatog:
Seems like tests might randomly fail - http://ci.theforeman.org/job/test_develop_pr_core/11900/#showFailuresLink - will investigate.

Just saw those failures on https://github.com/theforeman/foreman_ansible/pull/74 without this PR being merged.. so I guess this happens on the current develop.

dLobatog:
[test] ok to test

dLobatog:
@tbrisker I implemented your suggestions here

dLobatog:
@tbrisker Updated to solve the conflicts - hope no tests break again (only one broke in auth_source_ldap), this PR becomes just bigger the more tests rely on buggy behavior.

test/controllers/hosts_controller_test.rb
```diff
@@ -449,8 +456,8 @@ def setup_multiple_environments
     post :update_multiple_environment, params,
       set_session_user.merge(:user => users(:admin).id)
 
-    assert Host.find(@host1.id).environment == hostgroup.environment
-    assert Host.find(@host2.id).environment == hostgroup.environment
+    assert hostgroup.environment, Host.unscoped.find(@host1.id).environment
```
tbrisker: shouldn't this be `assert_equal`?

app/models/concerns/taxonomix.rb
```diff
@@ -69,26 +70,46 @@ def enforce_default
...
-                        end
+      # Return everything (represented by nil), including objects without
+      # taxonomies. This value should only be returned for admin users.
+      return nil if any_context?(loc) && any_context?(org) &&
+        User.current.try(:admin?)
+
+      ids = unscoped.pluck(:id)
+      ids &= inner_ids(loc, Location, inner_method) if SETTINGS[:locations_enabled]
+      ids &= inner_ids(org, Organization, inner_method) if SETTINGS[:organizations_enabled]
+      ids ||= []
```
tbrisker: not necessary, `ids` will be an array for certain. (even if one of the previous lines makes it empty)

test/controllers/subnets_controller_test.rb
tbrisker: looks like this is still here

test/unit/tasks/seeds_test.rb
```diff
@@ -46,18 +47,18 @@ def seed
 
   context 'populating an initial admin user' do
     test 'with defaults' do
-      assert_difference 'User.where(:login => "admin").count', 1 do
+      assert_difference 'User.unscoped.unscoped.where(:login => "admin").count', 1 do
```
tbrisker: double unscoped

tbrisker (changes requested):
Just a few minor comments inline. Should the katello PR be merged before or after this one?

dLobatog:
@tbrisker I think this one should be merged before katello, otherwise it will not pass some of the tests there. I've updated this one with the changes you've commented on - thanks for reviewing!

tbrisker:
@dLobatog I'm sorry, but looks like another test broke...

test/controllers/hosts_controller_test.rb
```diff
@@ -449,8 +456,8 @@ def setup_multiple_environments
     post :update_multiple_environment, params,
       set_session_user.merge(:user => users(:admin).id)
 
-    assert Host.find(@host1.id).environment == hostgroup.environment
-    assert Host.find(@host2.id).environment == hostgroup.environment
+    assert_equal hostgroup.environment_id, Host.unscoped.find(@host1.id).environment_id
```
dLobatog, line 459: This is the test that broke in the last update, `.unscoped` was missing and also I compare now with `_id` instead of calling `.reload`.

dLobatog:
@tbrisker Updated - left a comment inline about the test that broke

tbrisker:
@dLobatog  - I'm going to merge this now so we have time to get any issues this may cause fixed in time for 1.15. Please make sure that the katello pr also gets merged asap.

db/seeds.d/07-provisioning_templates.rb
```diff
@@ -81,7 +81,6 @@
     { :name => 'puppet.conf', :source => 'snippets/_puppet.conf.erb', :snippet => true },
     { :name => 'puppet_setup', :source => 'snippets/_puppet_setup.erb', :snippet => true },
     { :name => 'puppetlabs_repo', :source => 'snippets/_puppetlabs_repo.erb', :snippet => true },
-    { :name => 'pxelinux_discovery', :source => 'snippets/_pxelinux_discovery.erb', :snippet => true },
```
lzap, line 84: Was this removed on purpose? This snippet is in use. @dLobatog

lzap:
@dLobatog please ping Discovery maintainers as well with such a breaking changes :-) Allright let's get hands dirty.

PR #3963, opened 2016-10-24
fixes #11656 - disable cr interface fields on host edit

timogoebel:


lzap:
Thanks Timo! So the ticket is categorized as VMWare only, but the patch applies to all CRs. Can you confirm?

timogoebel:
@lzap : Yes. It applies to all CRs that don't support updating. In Redmine there is #12057 which may be a duplicate but is in the "Host creation" category. I only found that after editing the PR. Redmine search, you know... I can use that, if you want.

app/assets/javascripts/host_edit.js
```diff
@@ -762,6 +762,11 @@ function disable_vm_form_fields() {
   $("[id^=host_compute_attributes]").each(function () {
     $(this).attr("disabled", "disabled");
   });
+  $("[id^=host_interfaces_attributes]").filter(function() {
+    return this.id.match(/^host_interfaces_attributes_[0-9]+_compute_attributes_.*/);
+  }).each(function () {
+    $(this).attr("disabled", "disabled");
```
tbrisker, line 768: should be `.prop("disabled", true)`. Or better, be disabled by the view instead of disabling with js after loading it.

timogoebel: I agree, however we should prioritize. Imho fixing the issue is more important. We can then refactor the whole code to disable _all_ fields via views.

lzap:
Huge apologize, I was on holidays. Let me quickly re-test.

app/assets/javascripts/host_edit.js
lzap: I agree, let's just pull this in, important fix. Thanks!

PR #3559, opened 2016-05-30
fixes #2113 - vmware: filter datastores and networks based on cluster

timogoebel:
This PR filters datastores and networks based on the selected cluster for vmware.
It needs https://github.com/fog/fog-vsphere/pull/35 

@angystardust : Would you mind testing this? Please pay special attention to compute profile handling as well. :-)

PR #3964, opened 2016-10-24
fixes #15519 - speed up host edit page rendering on vmware

timogoebel:
This commit speeds up rendering of the host edit page for hosts on
VMWare. The API calls to list networks and datastores are expensive.
When editing a host, these information is never showed to a user and
furthermore CR editing on VMWare is not supported.
So instead of displaying a select box, this commit shows a textbox.

On my local environment, rendering before this patch was ~ 20seconds. After the patch, it got down to approx. 4 seconds. This makes a huge difference to a user.

I talked to @iNecas and @tstrachota about this. There are more ways to improve this
- caching
- filtering the network list based on a the VMWware cluster (#3559)
- not displaying a form but some nice patternfly cards with VM information, ...

but I think, this is a good quick win.

### DIFF
--- a/app/views/compute_resources_vms/form/vmware/_network.html.erb
+++ b/app/views/compute_resources_vms/form/vmware/_network.html.erb
@@ -1,9 +1,12 @@
-<% networks = compute_resource.networks %>
 <%= select_f f, :type, compute_resource.nictypes, :first, :last, { },
-                 :class => "col-md-3 vmware_type",
-                 :label => _('NIC type'), :label_size => "col-md-3"
+  :class => "col-md-3 vmware_type",
+  :label => _('NIC type'), :label_size => "col-md-3"
 %>
-<%= select_f f, :network, networks, :name, :name, { },
-                 :class => "col-md-3 vmware_network",
-                 :label => _('Network'), :label_size => "col-md-3"
-%>
+<% if new_host %>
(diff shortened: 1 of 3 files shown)


app/views/compute_resources_vms/form/vmware/_network.html.erb
```diff
@@ -1,9 +1,12 @@
...
+  :class => "col-md-3 vmware_type",
+  :label => _('NIC type'), :label_size => "col-md-3"
 %>
+<% if new_host %>
+  <%= select_f f, :network, compute_resource.networks, :name, :name, { },
+    :class => "col-md-3 vmware_network",
+    :label => _('Network'), :label_size => "col-md-3"
+  %>
+<% else %>
+  <%= text_f f, :network, :class => 'col-md-3 vmware_network', :label => _("Network"), :label_size => "col-md-3" %>
```
tbrisker, line 11: If this can't be changed, shouldn't it be set to `readonly`? same applies for the other form.

timogoebel: We do this via JS (for whatever reason), please see https://github.com/theforeman/foreman/pull/3963

dLobatog:
Thanks @timogoebel !

PR #3965, opened 2016-10-24
fixes #17070 - unpin rabl to permit 0.13.1

domcleal:


timogoebel:
Merged, thanks @domcleal .

PR #3969, opened 2016-10-25
Fixes #17083 - log skipped orchestration steps

lzap:
If there is an ActiveRecord error during unattended/built orchestration
(queue_tftp), the method silently does not enqueue TFTP orchestration. We
should print the error in logs, otherwise it is impossible to investigate the
root cause.

Symptoms are even more difficult to spot because due to bug RM/17082 no error
is reported when unattended "built" REST API call fails, so the TFTP
orchestration is silently ignored and hosts stays in provisioning loop.

The patch aims only to add logging, note this is currently sent into SQL logger
due to limitation of our logging stack.

### DIFF
--- a/app/models/concerns/orchestration/common.rb
+++ b/app/models/concerns/orchestration/common.rb
@@ -5,4 +5,12 @@ module Orchestration::Common
     logger.debug "Error occured during validations of #{self.class.name}: #{e.message}"
     nil
   end
+
+  def log_orchestration_errors
+    logged_errors = []
+    logged_errors << errors.full_messages if self.respond_to?(:errors) && errors.any?
+    logged_errors << host.errors.full_messages if self.respond_to?(:host) && host.respond_to?(:errors) && host.errors.any?
+    logger.warn("Not queueing #{self.class.name}: #{logged_errors.to_sentence}") if logged_errors.any?
+    false
+  end
 end
--- a/app/models/concerns/orchestration/compute.rb
+++ b/app/models/concerns/orchestration/compute.rb
@@ -3,6 +3,7 @@ require 'timeout'
 
 module Orchestration::Compute
   extend ActiveSupport::Concern
+  include Orchestration::Common
 
   included do
     attr_accessor :compute_attributes, :vm
@@ -36,7 +37,7 @@ module Orchestration::Compute
   protected
 
   def queue_compute
-    return unless compute? && errors.empty?
+    return log_orchestration_errors unless compute? && errors.empty?
     # Create a new VM if it doesn't already exist or update an existing vm
     vm_exists? ? queue_compute_create : queue_compute_update
   end
--- a/app/models/concerns/orchestration/dhcp.rb
+++ b/app/models/concerns/orchestration/dhcp.rb
@@ -135,7 +135,7 @@ module Orchestration::DHCP
   end
 
   def queue_dhcp
-    return unless (dhcp? || (old && old.dhcp?)) && orchestration_errors?
+    return log_orchestration_errors unless (dhcp? || (old && old.dhcp?)) && orchestration_errors?
     queue_remove_dhcp_conflicts
     new_record? ? queue_dhcp_create : queue_dhcp_update
   end
--- a/app/models/concerns/orchestration/dns.rb
+++ b/app/models/concerns/orchestration/dns.rb
@@ -55,7 +55,7 @@ module Orchestration::DNS
   end
 
   def queue_dns
-    return unless (dns? || dns6? || reverse_dns? || reverse_dns6?) && errors.empty?
+    return log_orchestration_errors unless (dns? || dns6? || reverse_dns? || reverse_dns6?) && errors.empty?
     queue_remove_dns_conflicts if overwrite?
     new_record? ? queue_dns_create : queue_dns_update
   end
--- a/app/models/concerns/orchestration/puppetca.rb
+++ b/app/models/concerns/orchestration/puppetca.rb
@@ -1,5 +1,6 @@
 module Orchestration::Puppetca
   extend ActiveSupport::Concern
+  include Orchestration::Common
 
   included do
     attr_reader :puppetca
@@ -43,7 +44,7 @@ module Orchestration::Puppetca
   private
 
   def queue_puppetca
-    return unless puppetca? && errors.empty?
+    return log_orchestration_errors unless puppetca? && errors.empty?
     return unless Setting[:manage_puppetca]
     new_record? ? queue_puppetca_create : queue_puppetca_update
   end
--- a/app/models/concerns/orchestration/realm.rb
+++ b/app/models/concerns/orchestration/realm.rb
@@ -1,5 +1,6 @@
 module Orchestration::Realm
   extend ActiveSupport::Concern
+  include Orchestration::Common
 
   included do
     after_validation  :queue_realm
@@ -45,7 +46,7 @@ module Orchestration::Realm
   private
 
   def queue_realm
-    return unless realm? && errors.empty?
+    return log_orchestration_errors unless realm? && errors.empty?
     new_record? ? queue_realm_create : queue_realm_update
   end
 
--- a/app/models/concerns/orchestration/tftp.rb
+++ b/app/models/concerns/orchestration/tftp.rb
@@ -1,5 +1,6 @@
 module Orchestration::TFTP
   extend ActiveSupport::Concern
+  include Orchestration::Common
 
   included do
     after_validation :validate_tftp, :unless => :skip_orchestration?
@@ -150,7 +151,7 @@ module Orchestration::TFTP
   end
 
   def queue_tftp
-    return unless (tftp? || tftp6?) && no_errors
+    return log_orchestration_errors unless (tftp? || tftp6?) && no_errors
     # Jumpstart builds require only minimal tftp services. They do require a tftp object to query for the boot_server.
     return true if host.jumpstart?
     new_record? ? queue_tftp_create : queue_tftp_update


app/models/concerns/orchestration.rb
```diff
@@ -199,8 +199,26 @@ def execute(opts = {})
     end
   end
 
-  def orchestration_errors?
-    overwrite? ? errors.are_all_conflicts? : errors.empty?
+  def orchestration_errors?(check_overwrite = true)
+    if check_overwrite && overwrite?
```
timogoebel: I think, this gets more readable by doing this:

```
return errors.are_all_conflicts? if check_overwrite && overwrite?
return true if errors.empty?
logger.error "Orchestration errors: #{errors.full_messages.join(', ')}"
false
```

app/models/concerns/orchestration.rb
```diff
@@ -199,8 +199,26 @@ def execute(opts = {})
...
+        true
+      else
+        logger.error "Orchestration errors: #{errors.full_messages.join(', ')}"
+        false
+      end
+    end
+  end
+
+  def host_orchestration_errors?
+    if host.errors.empty?
```
timogoebel:
``` ruby
return true of host.errors.empty?
logger.error "Orchestration host errors: #{host.errors.full_messages.join(', ')}"
false
```

app/models/concerns/orchestration.rb
```diff
@@ -199,8 +199,26 @@ def execute(opts = {})
     end
   end
 
-  def orchestration_errors?
-    overwrite? ? errors.are_all_conflicts? : errors.empty?
+  def orchestration_errors?(check_overwrite = true)
```
timogoebel: Where is `check_overwrite` used?

timogoebel (changes requested):
Thanks, @lzap . I left some comments inline.

lzap:
Done, thanks.

timogoebel (approved):

timogoebel:
@lzap : Thanks, I am unsure if this is the correct place for logging the errors as the messages are logged several times.
The method is called here: https://github.com/theforeman/foreman/blob/b7cc39f867cc4092cebd173ee3dec05202a96321/app/models/concerns/orchestration.rb#L75

How about moving the logging to `around_save_orchestration`? Then the logging is done only when the model is actually saved?

lzap:
I am trying to print orchestration queue errors, these basically stops orchestration from happening, nothing is saved at all and errors are swallowed silently. Putting it into `around_save_orchestration` does nothing if I understand you correctly.

lzap:
@dLobatog Timo is travelling, what is your opinion on that?

timogoebel:
> I am trying to print orchestration queue errors, these basically stops orchestration from happening, nothing is saved at all and errors are swallowed silently.

That's a great feature and very annoying when debugging errors. I've been there myself. Very, very annoying.

> Putting it into around_save_orchestration does nothing if I understand you correctly.

The methods, that queue orchestration (like `queue_dhcp`, `queue_tftp`, ...) just put the  orchestration tasks in a queue. The queue is then processed when the model is saved. With your current implementation, logging occurs every time you validate the host model. It's usually ok, that a specific task is not queued. Is this, what you want to log?
The logging is done a lot of times, e.g. just by submitting an invalid new hosts form.
Nothing is more annoying than errors logged, when there aren't any errors. And an invalid host form is already logged (but as info).
How about putting the logging to "Host.built"?

lzap:
I don't know if this is a good place, I would like to have this generic. How about changing these messages to debug level if your concern is amount of messages? It was a hard time to find the root cause for me, but debug would really help.

timogoebel:
@lzap: Ok, I have another suggestion. Please bear with me here and let me know, what you think.

You could add a method to `app/models/concerns/orchestration.rb`, let's call it `log_orchestration_errors`. Let's add this method as a `after_validation` callback.

It could look like this:
```ruby

included do
  after_validation :log_orchestration_errors
end

def log_orchestration_errors
  logger.info "Orchestration errors: #{errors.full_messages.to_sentence}" if orchestration_errors?
  logger.info "Orchestration host errors: #{host.errors.full_messages.to_sentence}" if host_orchestration_errors?
  true
end
```

The advantages are:
- The logs are just logged once per validation.
- It uses `to_sentence` instead of `join(',')`.

What's still missing:
`host_orchestration_errors?` should not be defined in `app/models/concerns/orchestration.rb` as this is also included in `Host::Managed` and `host` isn't defined there.

Any thoughts?

dLobatog:
(setting to WoC waiting for @lzap 's response)

lzap:
@timogoebel I like your idea, redone the patch completely, I added an extra after validator for base nic, but it looks it does not get triggered and all errors I've been able to produce so far were all reported via orchestration concern method. Maybe we don't need that to be included in `NIC::Base` at all, that makes things little more simple.

Also I think the best default log level could be WARN rather than ERROR (I think INFO is not ideal - it's overused in general). It's because you can get these overwrite "errors" as well:

    Orchestration errors: Conflict DHCP records somehost-813430.local.lan-52:54:00:25:fe:d9/192.168.122.14 already exists and Conflict DHCP records somehost-813430.local.lan-52:54:00:25:fe:d9/192.168.122.14 already exists

lzap:
Rebased, can you take a look again please?

timogoebel:
@lzap : Could you please check the Katello test failure?

timogoebel:
> Also I think the best default log level could be WARN rather than ERROR

@lzap: If you prefer warn, feel free to change the log level to warn.

lzap:
Added missing `obj`.

app/models/nic/base.rb
```diff
@@ -271,6 +274,11 @@ def interface_attribute_uniqueness(attr, base = Nic::Base.where(nil))
       db_candidates = db_candidates.select { |c| c.id != self.id && in_memory_candidates.map(&:id).include?(c.id) }
       errors.add(attr, :taken) if db_candidates.present?
     end
+
+    def log_host_orchestration_errors
+      logger.error "Host orchestration errors: #{host.errors.full_messages.to_sentence}" unless orchestration_errors?(host)
```
timogoebel: Katello test still fails:
NoMethodError: undefined method `orchestration_errors?' for #<Nic::Base:0x0000001baa0558>

lzap: I think I will pull NIC::Base off this PR, I don't remember why it was added in the first place.

lzap:
Enough with this one! I am taking completely different approach - I do not want generic solution, I just want clean and working solution to the problem - skipped TFTP orchestration due to AR errors. This will do it.

app/models/concerns/orchestration/compute.rb
```diff
@@ -36,7 +36,11 @@ def vm_name
   protected
 
   def queue_compute
-    return unless compute? && errors.empty?
+    return unless compute?
+    unless errors.empty?
+      logger.warning("Not queueing Compute: #{errors.full_messages.to_sentence}")
```
timogoebel: @lzap: I'd prefer `logger.warn` over `logger.warning`?

app/models/concerns/orchestration/dhcp.rb
```diff
@@ -135,7 +135,11 @@ def dhcp_record_name(record_mac)
   end
 
   def queue_dhcp
-    return unless (dhcp? || (old && old.dhcp?)) && orchestration_errors?
+    return unless dhcp? || (old && old.dhcp?)
+    unless orchestration_errors?
+      logger.warning("Not queueing DHCP: #{(errors + host.errors).full_messages.to_sentence}")
```
dLobatog: It seems that ActiveModel errors cannot be added together (`undefined method `+' for #<ActiveModel::Errors:0x00000015624d40>.`), how about logging DHCP and Host errors in 2 different lines?

dLobatog (changes requested):
@lzap Tests are failing because of the way errors are added before logging them. I think the solution is simple to understand 👍

lzap:
Using `warn` instead, split into two lines.

app/models/concerns/orchestration/dns.rb
```diff
@@ -55,7 +55,12 @@ def rebuild_dns
   end
 
   def queue_dns
-    return unless (dns? || dns6? || reverse_dns? || reverse_dns6?) && errors.empty?
+    return unless dns? || dns6? || reverse_dns? || reverse_dns6?
+    unless errors.empty?
+      logger.warn("Not queueing DNS: #{errors.full_messages.to_sentence}")
```
timogoebel: I think, it's unclear if the errors are on the host object or nic just by reading the logs:
How about something like this (untested):
```ruby
logger.warn("Not queueing DNS because of Nic (#{identifier}/#{mac}) errors: #{errors.full_messages.to_sentence}") if errors.any?
logger.warn("Not queueing DNS because of Host errors: #{host.errors.full_messages.to_sentence}") if host.errors.any?
```

Same for DHCP and TFTP.

timogoebel (changes requested):
@lzap: Sorry for the delay. I still think, this is an important patch. I just have one small suggestion regarding the wording of the logs to make it easier for users to understand what's going on.

lzap:
OK how about this?

timogoebel (approved):
@lzap: Yep, works nicely. 👍 from me.

timogoebel:
@lzap: Could you please check out the test failures? I'm not sure if they're related.

lzap:
Yes that was my error, now it's fixed and more simple. It's not relevant if the error is in NIC or Host, we just need to actually *see* the error.

app/models/concerns/orchestration/common.rb
```diff
@@ -5,4 +5,11 @@ def handle_validation_errors
     logger.debug "Error occured during validations of #{self.class.name}: #{e.message}"
     nil
   end
+
+  def log_orchestration_errors
+    errors = []
+    errors << errors.full_messages if self.respond_to?(:errors) && errors.any?
+    errors << errors.full_messages if self.respond_to?(:host) && host.respond_to?(:errors) && host.errors.any?
```
timogoebel: I believe this must be `host.errors.full_messages`.

app/models/concerns/orchestration/dhcp.rb
```diff
@@ -135,7 +135,11 @@ def dhcp_record_name(record_mac)
   end
 
   def queue_dhcp
-    return unless (dhcp? || (old && old.dhcp?)) && orchestration_errors?
+    return unless dhcp? || (old && old.dhcp?)
+    unless orchestration_errors?
```
timogoebel: `if orchestration_errors?`?

app/models/concerns/orchestration/dns.rb
```diff
@@ -55,7 +55,11 @@ def rebuild_dns
   end
 
   def queue_dns
-    return unless (dns? || dns6? || reverse_dns? || reverse_dns6?) && errors.empty?
+    return unless dns? || dns6? || reverse_dns? || reverse_dns6?
+    unless orchestration_errors?
```
timogoebel: `if orchestration_errors?` ?
If there are errors, we want to log and not queue that.

app/models/concerns/orchestration/common.rb
```diff
@@ -5,4 +5,11 @@ def handle_validation_errors
     logger.debug "Error occured during validations of #{self.class.name}: #{e.message}"
     nil
   end
+
+  def log_orchestration_errors
+    errors = []
```
timogoebel: This cannot be named `errors` as this shadows the ActiveRecord `errors` collection and this breaks the wohle thing. Please rename this to `my_errors` or `orchestration_errors` or whatever...

timogoebel (dismissed):
@lzap: I left more comments.

lzap:
Good eye, holy moly what am I doing here. Frankly, I was this close closing this PR two times already, I thought like it is not worth the effort. Let's see.

timogoebel:
@lzap: This still fails some tests.

dLobatog (commented):
This is a pretty simple fix that saves a lot of trouble IMO - is it possible to rebase & test again?

dLobatog (approved):

lzap:
Rebased! Thanks.

dLobatog (changes requested):
@lzap Seems that a bunch of tests fail when you return earlier than expected, could you check that? Thanks!

lzap:
Weird, you do see why it fails? It should remain unchanged.

lzap:
Ok how about this? This is kind of chaining call stack in order to achieve return on one line, but I think it's not bad code at all.

dLobatog (approved):
Definitely makes sense. 👍 Thanks @lzap !

dLobatog:
Failing test is one of the usual DNS failures.

PR #3970, opened 2016-10-25
Fixes #16850 - Added record of password change to audit log

dhlavac:


domcleal:
@theforeman-bot ok to test

app/models/user.rb
```diff
@@ -445,6 +446,24 @@ def self.try_to_auto_create_user(login, password)
...
+    attribute_will_change!('password_changed') if password_changed != value
+    @password_changed = value
+  end
+
+  def password_changed_changed?
+    changed.include?('password_changed')
+  end
+
+  def set_password_changed
+    self.password_changed = true
```
domcleal: This seems to be missing a conditional, will it not mark password_changed on _every_ model validation?

app/models/user.rb
```diff
@@ -445,6 +446,24 @@ def self.try_to_auto_create_user(login, password)
     user
   end
 
+  def password_changed
+    @password_changed ||= nil
```
domcleal: Unsure what `||= nil` is achieving here, should this just be an `attr_reader :password_changed` ?

app/models/user.rb
```diff
@@ -445,6 +446,24 @@ def self.try_to_auto_create_user(login, password)
...
+  end
+
+  def password_changed=(value)
+    attribute_will_change!('password_changed') if password_changed != value
+    @password_changed = value
+  end
+
+  def password_changed_changed?
```
domcleal, line 453: Is this necessary? Doesn't AM::Dirty provide it?

app/models/user.rb
```diff
@@ -445,6 +446,24 @@ def self.try_to_auto_create_user(login, password)
...
+    changed.include?('password_changed')
+  end
+
+  def set_password_changed
+    self.password_changed = true
+    self.audit_comment = 'Password has been changed'
```
domcleal: Is this necessary if the attribute's recorded?

domcleal (changes requested):
Some inline comments, but the main thing this is missing is tests for the model changes. Please add some to ensure the auditing happens on password change, and also not on a regular update.

dhlavac:
@domcleal I changed pull request and to yours previous comments: 1.attr_reader is fine, thank you, 2. password_changed_changed? is neccessary because it working with virtual attribute and other things I rosolve please could yo look at that ?

app/models/user.rb
```diff
@@ -88,6 +90,7 @@ def self.name_format
   before_validation :verify_current_password, :if => Proc.new {|user| user == User.current},
                     :unless => Proc.new {|user| user.password.empty?}
   before_validation :prepare_password, :normalize_mail
+  before_validation :set_password_changed, :if => Proc.new { |user| user.manage_password? && !user.password.blank? }
```
domcleal: !..blank? should be `present?`

test/models/user_test.rb
```diff
@@ -320,6 +320,32 @@ def setup_user(operation)
     assert_includes record.errors.keys, :admin
   end
 
+  test "log of password change should be saved" do
+    user = FactoryGirl.create(:user)
+    user.password = nil
```
domcleal: This is rather hacky, reload the object from the database instead (`User.find_by_id(user.id)`) and add a comment to state why (i.e. to clear the value of user.password).

test/models/user_test.rb
```diff
@@ -320,6 +320,32 @@ def setup_user(operation)
...
+    user.save
+    as_admin do
+      user.password = "newpassword"
+      user.valid?
```
domcleal: `assert_valid user`

test/models/user_test.rb
```diff
@@ -320,6 +320,32 @@ def setup_user(operation)
     assert_includes record.errors.keys, :admin
   end
 
+  test "log of password change should be saved" do
```
domcleal: audit, not log

test/models/user_test.rb
```diff
@@ -320,6 +320,32 @@ def setup_user(operation)
...
+    as_admin do
+      user.password = "newpassword"
+      user.valid?
+      assert user.password_changed_changed?
+      assert user.password_changed
+      assert_includes user.changed, "password_changed"
+    end
+  end
+
+  test "log of password change shouldnt be saved" do
```
domcleal: 1. shouldn't, missing an apostrophe
1. There is no password change here - it isn't that the log of a password change shouldn't be saved, it's that there is no audit log entry created when there is no password change.

test/models/user_test.rb
```diff
@@ -320,6 +320,32 @@ def setup_user(operation)
...
+    user = FactoryGirl.create(:user)
+    user.password = nil
```
domcleal: ditto all comments above, they apply to this test too.

app/models/user.rb
```diff
@@ -16,7 +16,9 @@ class User < ActiveRecord::Base
 
   validates_lengths_from_database :except => [:firstname, :lastname, :format, :mail, :login]
   attr_accessor :password, :password_confirmation, :current_password
+  attr_reader :password_changed
   after_save :ensure_default_role
+  after_save :unset_password_changed
```
domcleal, line 21: Missing test coverage? A test should run two saves in a row and check for no second audit entry.

domcleal (changes requested):

test/models/user_test.rb
dhlavac: In this test I want to test that , it should not mark password_changed on every model validation

dhlavac:
thanks @domcleal, I fixed all your comments, could you please take another look?

test/models/user_test.rb
```diff
@@ -320,6 +320,50 @@ def setup_user(operation)
...
+    user = User.find_by_id(user.id) #to clear the value of user.password
+    as_admin do
+      user.password = "newpassword"
+      assert_valid user
+      assert user.password_changed_changed?
+      assert user.password_changed
+      assert_includes user.changed, "password_changed"
+      assert user.save
+    end
+    user = User.find_by_id(user.id) #to clear the value of user.password
```
domcleal: This is this meant to test the `after_save`, isn't it? Replacing the entire user object will also reset the value, so it doesn't test the callback.

domcleal (changes requested):

dhlavac:
@domcleal I have edited test which testing after_save

domcleal:
Thanks, it seems OK but I think this is still missing the first tests I requested to check that the auditing happens on password change. The new model methods seem well-tested, but since it's rather hacky (knowing which methods the audit gem will call), I think checking the audit record exists is quite important.

test/models/user_test.rb
```diff
@@ -320,6 +320,47 @@ def setup_user(operation)
...
+    as_admin do
+      user.password = "newpassword"
+      assert_valid user
+      assert user.password_changed_changed?
+      assert user.password_changed
+      assert_includes user.changed, "password_changed"
+    end
+  end
+
+  test "audit of password change should not be saved - due to no password change" do
```
dhlavac: @domcleal I think this test tested that the auditing happens on password change.

domcleal: 1. This test checks that there is _no audit_.
1. The test above ("audit of password change should be saved") doesn't assert that an audit entry is created. It only checks the return values of the new methods you added. It should verify that the audit is created.

dhlavac:
@domcleal added  test assert that an audit entry is created and password_changed is not saved in audit log ... is it this what you mean ?

domcleal:
> @domcleal added test assert that an audit entry is created and password_changed is not saved in audit log ... is it this what you mean ?

I guess, but I don't see it in the PR (d34e0b188c7aaa17a7e5680a9b4e5a749b2d2773).

test/models/user_test.rb
```diff
@@ -320,6 +320,60 @@ def setup_user(operation)
...
+    as_admin do
+      user.firstname = "Johnny"
+      assert_valid user
+      refute user.password_changed_changed?
+      refute user.password_changed
+      refute_includes user.changed, "password_changed"
+    end
+  end
+
+  test "audit of name change sholud contain only firstname and not password_changed" do
```
dhlavac: @domcleal right there

domcleal: This only seems to call methods on `user`, I don't see any checks on the audit object(s).

dhlavac: @domcleal  Ok added assertion in Audit object .. sorry for misunderstanding

domcleal: Where is it? Do you mean https://github.com/theforeman/foreman/pull/3970/files#diff-b6d7d06f7e7b01588ae44fbc220c0675R375 ?

    assert_includes Audit.last.audited_changes, "firstname"

If so, this asserts that the audit logs changes to firstname. The point of the PR is auditing password_changed, shouldn't there be assertions for _that_ field instead? (In the other tests)

dhlavac: Yes that is it. I think is not necessary because if user.changed contains attribute, than Audit log contains it too. But I can change that if you want.

domcleal: I think I'd rather see an assertion for it, instead of relying on that to remain true.

dhlavac:
@domcleal  I hope Final version added assertion for Audit object to every test

test/models/user_test.rb
```diff
@@ -320,6 +320,68 @@ def setup_user(operation)
     assert_includes record.errors.keys, :admin
   end
 
+  test "audit of password change should be saved only once, second time audited changes should not contain password_changed" do
+    user = FactoryGirl.create(:user)
```
dLobatog: Please move the user creation to a `setup` block, it's the same in all 3 hosts. You can group them in a context `audits for password change` or similar.

dLobatog:
@dhlavac I've put a minor comment on tests - otherwise I haven't tested it much but it seemed you addressed @domcleal comments.

app/models/user.rb
dLobatog: @dhlavac When I see your tests - it looks like you're calling at most `.valid?` and `.save`, so I'm not 100% positive this is ggcovered as per the comment above.

dLobatog (changes requested):

dhlavac:
@dLobatog Audit log was created before validation and after save I set the variable password_changed to false then I think this is correct, if I tested it is not still true. Thanks for comment tests were added to context and setup block was created.

dLobatog (approved):

dLobatog:
Thanks @dhlavac, waiting on @domcleal to approve.

domcleal (approved):

domcleal:
Merged, thanks @dhlavac.

PR #3971, opened 2016-10-26
fixes #17095 - ensure subnet belongs to ip from facts

timogoebel:


ares:
Maybe this should be configurable through "Ignore Puppet facts for provisioning" setting can be reused. This could disassociate a lot of interfaces if DHCP server was down temporarily. Also if managed interface gets disassociated and I remove the host afterwards, it wouldn't delete DHCP/TFTP/DNS PTR records I'm afraid.

timogoebel:
> Maybe this should be configurable through "Ignore Puppet facts for provisioning" setting can be reused.

I just looked through the code. This code does not run when ignore_puppet_facts_for_provisioning is set.

https://github.com/theforeman/foreman/blob/4d0870ad2503d9e1489aa18ddc4e837675a0b8f6/app/services/fact_parser.rb#L89-L91
https://github.com/theforeman/foreman/blob/a390cfb73dd55ed6735e9f8916699dc15415ce03/app/models/host/base.rb#L156

> This could disassociate a lot of interfaces if DHCP server was down temporarily.

Yes, I think therefore we have this setting.

Generally speaking, I agree. This can cause issues. Let me explain, why I think this change is necessary.
Let's imagine a Foreman user uses Foreman to provision laptops. The laptops get installed in a deployment network. Afterwards, they are handed out to the laptop users and they use the laptop in other networks. The ip address is never updated in Foreman and the DHCP range in the deployment network is eventually full.

> Also if managed interface gets disassociated and I remove the host afterwards, it wouldn't delete DHCP/TFTP/DNS PTR records I'm afraid.

Maybe we should actually update the network and not just set it to nil. We could also remove the "managed" flag. But orchestration is actually disabled during fact imports anyways.

I saw, that requiring that the IP belongs to the subnet for all managed interfaces (was all dhcp enabled interfaces before) causes a lot of tests to fail. I'll happily look into that, when this change is approved.

ares:
> I just looked through the code. This code does not run when ignore_puppet_facts_for_provisioning is set.

good, thanks for checking, I didn't look where exactly you call it...

> Let's imagine a Foreman user uses Foreman to provision laptops. The laptops get installed in a deployment network. Afterwards, they are handed out to the laptop users and they use the laptop in other networks. The ip address is never updated in Foreman and the DHCP range in the deployment network is eventually full.

Thanks for explanation, I have few more questions though. In this scenario, shouldn't you just delete the host from Foreman after provisioning to do the cleanup? Othwerwise it would also keep DNS and TFTP "reservations". I can see this is not such a big issue but still, what is the point of keeping this host in Foreman, later reprovisioning? Chances are that the host will boot from different DHCP changing it's domain hence FQDN, so it will register as a new host if it uses the same puppet master.

timogoebel:
> In this scenario, shouldn't you just delete the host from Foreman after provisioning to do the cleanup?

Then you cannot use Foreman as a puppet enc/dashboard, can you?
Of course you can just remove the "managed" flag vom the nic and everything gets cleaned up. The Host IP address does change now, however the subnet still does not match the ip address. (I know, the code should also check if foreman knows the network and change it accordingly. I'll add that later.).

> Chances are that the host will boot from different DHCP changing it's domain hence FQDN, so it will register as a new host if it uses the same puppet master.

In our network, hostnames for laptops are fixed. They register in dns through "dynamic dns". So this would work fine.

Let's imagine another scenario:
You remove the IPv6 address from a host. From now on, the nics aren't updated from facts because the subnet6 is still set and therefore a ipv6 address is required for the nic to be valid.

Do you have a suggestion how we could mitigate the issues?

timogoebel:
The code tries to match the interface subnet to the ip from that comes from facts.

timogoebel:
Fixed all tests.

ares:
> > In this scenario, shouldn't you just delete the host from Foreman after provisioning to do the cleanup?
> 
> Then you cannot use Foreman as a puppet enc/dashboard, can you?

Not sure I follow, if the host wakes up on another subnet it would register itself from there in Foreman as unmanaged host and you can still see its reports/facts and modify ENC for it.

> You remove the IPv6 address from a host. From now on, the nics aren't updated from facts because the subnet6 is still set and therefore a ipv6 address is required for the nic to be valid.

Right, I don't have good answer for that. I don't want to block this PR but I'd also like to hear others opinions on this. Few names who might have good ideas on the topic @lzap, @domcleal, @GregSutcliffe, @ohadlevy.

TL;DR allow changing subnet of managed interface based on IP coming from puppet facts? It could trigger the NIC orchestration.

GregSutcliffe:
I'm going to put an immediate reaction and then think about it more tomorrow...

This proposed behaviour feels like the behaviour people would expect - that when the IP changes, the orchestration is triggered to match (unless it's an unmanaged NIC etc). Obviously that's not what happens today, so from a new user perspective :+1: 

Thinking about the laptop scenario - my laptops exist in Foreman with both a wired and wireless NIC (both managed). That's because they'll only ever be on one subnet anyway, but for provisioning/roaming would it be a case of documenting the expectation? If orchestration is triggered on fact import, then I could see an argument for unmanaging the wireless NIC, but keep the wired one for provisioning?

As a final thought which I've not explored fully in my own head - should this be a setting per NIC or even per subnet (perhaps defaulting to the value of ignore_puppet_facts_for_provisioning)? I can definitely see situations where the current behaviour is desirable, and others (as shown by Timo's ipv6 example) where it isn't.

timogoebel:
@ares: How are we going to continue here? How about I add a setting, so that a user can choose which behavior he desires?

lzap:
I'd say yes, when we are changing such an important behavior, we usually decide to provide an opt-out way to keep the old one. I don't need this myself, so I would be fine with merging this right away, but users have weird setups I usually can't even think about.

timogoebel:
Added the setting `update_subnets_from_facts`.

app/models/nic/base.rb
```diff
@@ -178,6 +178,12 @@ def ip6=(addr)
       super(Net::Validations.normalize_ip6(addr))
     end
 
+    def matches_subnet?(ip_field, subnet_field)
+      return unless send(subnet_field).present?
+      value = send(ip_field)
```
dLobatog: Naming nitpick, 'value' -> 'ip_value'

test/models/host_test.rb
```diff
@@ -1604,6 +1604,77 @@ def teardown
...
+        assert_no_difference 'Nic::Base.count' do
+          host.set_interfaces(parser)
+        end
+
+        assert_equal '10.10.20.2', host.primary_interface.ip
+        assert_equal subnet2, host.primary_interface.subnet
+      end
+
+      test "#set_interfaces updates existing physical interface when subnet6 is set but facts report no ipv6 addr" do
+        subnet = FactoryGirl.create(:subnet_ipv4, :network => '10.10.10.0')
```
dLobatog: Looks like the `subnet` field is the same in all tests, please move it to the setup method

app/models/nic/base.rb
dLobatog: `public_send` preferred too, since those are public methods - it reduces the chances of accidentally calling `matches_subnet` with private methods

test/models/orchestration/dns_test.rb
```diff
@@ -138,7 +138,7 @@ def teardown
     test 'should queue dns update' do
       @host.save!
       @host.queue.clear
-      @host.ip = '1.2.3.4'
+      @host.ip = IPAddr.new(IPAddr.new(@host.ip).to_i + 1, Socket::AF_INET).to_s
```
dLobatog, line 141: Why these changes here?

dLobatog (changes requested):
@timogoebel Thanks - a few minor changes remaining and I'll merge. If you don't have the time let me know and I'll make them so we can merge this for 1.14.

test/models/orchestration/dns_test.rb
timogoebel: I need to change the host ip to trigger the dns update. The host is managed and this PR requires all managed hosts to have the ip to mach the subnet.  This wasn't the case before, so the test passed. To fix the test, i just calculate the succeeding ip. That seemed to be the easiest way to ensure the ip changes and stays within the subnet.

timogoebel:
@dLobatog : Thanks for your comments. I have addressed them all. Let me know if there is anything more that needs changing.

app/models/nic/base.rb
```diff
@@ -178,6 +178,12 @@ def ip6=(addr)
       super(Net::Validations.normalize_ip6(addr))
     end
 
+    def matches_subnet?(ip_field, subnet_field)
+      return unless send(subnet_field).present?
+      ip_value = send(ip_field)
+      value.present? && public_send(subnet_field).contains?(ip_value)
```
dLobatog: `value` here has to be changed to `ip_value`, this is causing a bunch of tests to fail

dLobatog (changes requested):
Nearly there - there's a typo ('value' instead of 'ip_value' on base.rb). Other than that 👍, will merge after that. Thanks @timogoebel

app/models/nic/base.rb
timogoebel: I'm always glad we have the tests. Fixed, thanks for your feedback.

dLobatog:
@timogoebel 1 test is failing 😄 http://ci.theforeman.org/job/test_develop_pr_core/11308/database=postgresql,ruby=2.3,slave=fast/testReport/junit/(root)/HostTest__location%20or%20organizations%20are%20not%20enabled__update_subnets_from_facts%20is%20enabled/test_0003__set_interfaces_updates_existing_physical_interface_when_subnet6_is_set_but_facts_report_no_ipv6_addr/ this one's a tough nut to crack!

timogoebel:
> this one's a tough nut to crack!

... at least when you're not in the office ;-)
Fixed and ran the tests locally. Should be 💚 now. Thanks, Daniel @dLobatog . :-)

dLobatog (approved):

dLobatog:
Thanks @timogoebel @ares @GregSutcliffe @lzap !

PR #3975, opened 2016-10-27
Fixes #17084 - Add webpack on welcome and unauthorized pages

bastilian:
On pages where the welcome action in the app controller was triggered
the webpack server was not added due to welcome rendering before

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -19,7 +19,7 @@ class ApplicationController < ActionController::Base
   before_action :set_taxonomy, :require_mail, :check_empty_taxonomy
   before_action :authorize
   before_action :welcome, :only => :index, :unless => :api_request?
-  before_action :allow_webpack, :if => -> {Rails.configuration.webpack.dev_server.enabled}
+  prepend_before_action :allow_webpack, if: -> { Rails.configuration.webpack.dev_server.enabled }
   around_action :set_timezone
   layout :display_layout?
 
@@ -407,10 +407,15 @@ class ApplicationController < ActionController::Base
   end
 
   def allow_webpack
+    webpack_csp = { script_src: [webpack_server], connect_src: [webpack_server],
+                    style_src: [webpack_server], img_src: [webpack_server] }
+
+    append_content_security_policy_directives(webpack_csp)
+  end
+
+  def webpack_server
     port = Rails.configuration.webpack.dev_server.port
-    @dev_server = "#{request.protocol}#{request.host}:#{port}"
-    append_content_security_policy_directives(script_src: [@dev_server], connect_src: [@dev_server],
-                                              style_src: [@dev_server], img_src: [@dev_server])
+    @dev_server ||= "#{request.protocol}#{request.host}:#{port}"
   end
 
   class << self
--- a/test/integration/middleware_test.rb
+++ b/test/integration/middleware_test.rb
@@ -12,14 +12,36 @@ class MiddlewareIntegrationTest < ActionDispatch::IntegrationTest
       "'self'; style-src 'unsafe-inline' 'self'"
   end
 
-  test "webpack dev server adds the dev server to Content-Security-Policy" do
-    begin
+  context 'webpack dev server is enabled' do
+    setup do
       Rails.configuration.webpack.dev_server.enabled = true
+      @webpack_url = "#{host}:#{Rails.configuration.webpack.dev_server.port}"
       Webpack::Rails::Manifest.stubs(:asset_paths).returns([])
-      visit '/'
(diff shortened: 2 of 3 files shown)


ares:
@theforeman-bot ok to test

ares:
@bastilian please change the commit message so it follows the pattern, e.g. "Fixes #17084 - add webpack on welcome packages" and `git push -f`. I tested the patch and it fixes the issue for me. If tests passes it's good to merge after the commit message change.

bastilian:
Commit message fixed and pushed.

app/controllers/application_controller.rb
```diff
@@ -18,8 +18,8 @@ class ApplicationController < ActionController::Base
   before_action :session_expiry, :update_activity_time, :unless => proc {|c| !SETTINGS[:login] || c.remote_user_provided? || c.api_request? }
   before_action :set_taxonomy, :require_mail, :check_empty_taxonomy
   before_action :authorize
```
domcleal, line 20: The same applies to `authorize`, the CSP headers must be in place for the 403 page.

domcleal (changes requested):
Also the commit message states "welcome packages" - should it say "welcome pages"?

bastilian:
@domcleal packages is now corrected to pages. and it now uses `prepend_before_action` and includes a test to verify it is also set on unauthorised page requests.

test/integration/middleware_test.rb
```diff
@@ -22,4 +22,28 @@ class MiddlewareIntegrationTest < ActionDispatch::IntegrationTest
       Rails.configuration.webpack.dev_server.enabled = false
     end
   end
+
+  test "webpack dev server adds the dev server to Content-Security-Policy on pages triggering a welcome page" do
+    begin
+      Rails.configuration.webpack.dev_server.enabled = true
+      Webpack::Rails::Manifest.stubs(:asset_paths).returns([])
+      Environment.stubs(:first).returns(nil)
```
domcleal, line 32: Both new tests are missing any verification that these stubs have triggered the filters you expect. Probably at a minimum they should be expectations, but some kind of verification of the page content may also help.

test/integration/middleware_test.rb
```diff
@@ -22,4 +22,28 @@ class MiddlewareIntegrationTest < ActionDispatch::IntegrationTest
...
+      visit '/environments'
+      assert page.response_headers['Content-Security-Policy'].include?(Rails.configuration.webpack.dev_server.port.to_s)
+    ensure
+      Rails.configuration.webpack.dev_server.enabled = false
+    end
+  end
+
+  test "webpack dev server adds the dev server to Content-Security-Policy on unauthirized page requests" do
```
domcleal: spelling, "unauthorized"

test/integration/middleware_test.rb
```diff
@@ -22,4 +22,28 @@ class MiddlewareIntegrationTest < ActionDispatch::IntegrationTest
...
+    begin
+      Rails.configuration.webpack.dev_server.enabled = true
+      Webpack::Rails::Manifest.stubs(:asset_paths).returns([])
+      User.stubs(:current).returns(nil)
```
domcleal: Prefer overriding `login_admin` (see user_test.rb), same comment as above applies - no verification that this tests the `authorize` filter behaviour.

domcleal (changes requested):

bastilian:
@domcleal thanks for taking a closer look at it!

I've added assertions to verify the behaviour of what the stubs should trigger as well, override the `login_admin` for the unauthorised test and wrap the three tests in a context to clean it up.

domcleal (approved):

test/integration/middleware_test.rb
```diff
@@ -12,14 +12,36 @@ class MiddlewareIntegrationTest < ActionDispatch::IntegrationTest
...
+      assert page.has_content? 'Learn more about this in the documentation.'
+      assert page.response_headers['Content-Security-Policy'].include?(Rails.configuration.webpack.dev_server.port.to_s)
+    end
+
+    context "on unauthorized page requests" do
+      def login_admin; end
+
+      test "webpack dev server adds the dev server to Content-Security-Policy as well" do
+        visit '/environments'
+        assert page.status_code, '403'
```
domcleal: `assert_equal '403', page.status_code` - this currently asserts that page.status_code is trueish

domcleal (changes requested):

bastilian:
@domcleal I've cleaned it up now and added `logout_admin`, which simple removes the cookie, since overriding `login_admin` did not trigger an unauthorised request in this case.

domcleal (approved):

domcleal:
Merged, thanks @bastilian.

bastilian:
Thanks for merging @domcleal! 🎉  :)

PR #3976, opened 2016-10-27
Fixes #17120 - Allow custom name for setting category

xprazak2:


app/helpers/settings_helper.rb
```diff
@@ -25,6 +25,10 @@ def short_cat(category)
     category.gsub(/Setting::/,'')
   end
 
+  def cat_label(category)
+    category.constantize.humanized_category || short_cat(category)
```
domcleal, line 29: `humanized_category` calls the deprecated Setting method_missing getter by default. Is there meant to be a default implementation in Setting?

domcleal (changes requested):

app/helpers/settings_helper.rb
xprazak2: implementing it in Setting will avoid hitting `method_missing`...

domcleal:
Could you also add the same humanised string to the API response for settings (probably main.json.raml)? Then it'll be usable from Hammer etc, which'll be an improvement over the class name in the `category`. Perhaps call it `category_name` or similar.

Please also add it to the example at http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Settings.

xprazak2:
I added the `:category_name` into main.json.rabl for settings.

app/views/api/v2/settings/main.json.rabl
```diff
@@ -3,3 +3,7 @@ object @setting
 extends "api/v2/settings/base"
 
 attributes :value, :description, :category, :settings_type, :default, :created_at, :updated_at
+
+node do |s|
```
domcleal: Nitpick, prefer `node :category_name` syntax instead of a hash if adding just one node.

app/views/api/v2/settings/main.json.rabl
```diff
@@ -3,3 +3,7 @@ object @setting
...
+  { :category_name => s.class.humanized_category }
```
domcleal: 1. This will return `nil` for most categories, it should follow the same logic as the UI helper (`s.class.name.demodulize`) when humanized_category is nil.
2. Missing `_( .. )` of the result.

domcleal (changes requested):

xprazak2:
I changed the syntax of `node`, added translation and `category_name` now returns the same thing as helper.

app/views/api/v2/settings/main.json.rabl
```diff
@@ -3,3 +3,7 @@ object @setting
 extends "api/v2/settings/base"
 
 attributes :value, :description, :category, :settings_type, :default, :created_at, :updated_at
+
+node :category_name do
+  _(@setting.category.constantize.humanized_category || @setting.category.gsub(/Setting::/,''))
```
domcleal: 1. The block has the setting passed in (`node :category_name do |s|`), use that instead of referring to `@setting` here. I thought this was correct in the previous version.
2. `@setting.class` should be more direct than constantising via `@setting.category.constantize`

domcleal (changes requested):

xprazak2:
I changed to `s.class`, referring to a setting through a block variable.

domcleal (approved):

domcleal:
Merged, thanks @xprazak2.

PR #3977, opened 2016-10-28
fixes #17071 - set ip after acquiring compute details

timogoebel:
This makes EUI-64 IPAM usable (that needs the mac address to calculate an ip) with compute resources that provide the mac address.

### DIFF
--- a/app/models/concerns/orchestration/compute.rb
+++ b/app/models/concerns/orchestration/compute.rb
@@ -29,18 +29,6 @@ module Orchestration::Compute
     compute? && compute_resource.provided_attributes.keys.include?(attr)
   end
 
-  def ip_available?
-    ip.present? || compute_provides?(:ip)
-  end
-
-  def ip6_available?
-    ip6.present? || compute_provides?(:ip6)
-  end
-
-  def mac_available?
-    mac.present? || compute_provides?(:mac)
-  end
-
   def vm_name
     Setting[:use_shortname_for_vms] ? shortname : name
   end
@@ -61,6 +49,8 @@ module Orchestration::Compute
                  :action => [self, :setComputeIP]) if compute_provides?(:ip) || compute_provides?(:ip6)
     queue.create(:name   => _("Query instance details for %s") % self, :priority => 4,
                  :action => [self, :setComputeDetails])
+    queue.create(:name   => _("Set IP addresses for %s") % self, :priority => 5,
+                 :action => [self, :setComputeIPAM]) if compute_provides?(:mac) && (mac_based_ipam?(:subnet) || mac_based_ipam?(:subnet6))
     queue.create(:name   => _("Power up compute instance %s") % self, :priority => 1000,
                  :action => [self, :setComputePowerUp]) if compute_attributes[:start] == '1'
   end
@@ -168,6 +158,20 @@ module Orchestration::Compute
   end
 
   def delComputeIP; end
+
+  def setComputeIPAM
+    set_ip_address
+
+    unless required_ip_addresses_set?(false)
+      failure _('Failed to set IPs via IPAM for %{name}: %{e}') % {:name => name, :e => primary_interface.errors.full_messages.to_sentence }
+      return false
+    end
+    true
+  rescue => e
+    failure _("Failed to set IP for %{name}: %{e}") % { :name => name, :e => e }, e
+  end
+
+  def delComputeIPAM; end
 
   def delCompute
     logger.info "Removing Compute instance for #{name}"
@@ -302,9 +306,7 @@ module Orchestration::Compute
   end
 
   def validate_required_foreman_attr(value, object, attr)
-    if value.blank?
-      return failure("#{attr} value is blank!")
-    end
+    return failure("#{attr} value is blank!") if value.blank?
     validate_foreman_attr(value, object, attr)
   end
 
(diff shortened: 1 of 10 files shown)


domcleal:
I see a couple of issues when testing this:

1) Creating a host (through the UI) with only an IPv6 subnet and a domain set (for forward DNS) causes a `can't be blank` validation failure on `:ip` on the NIC, as `require_ip4_validation` returns true (because `!other_ip_field.present?`).

It looks like this validation needs to change to understand that the v6 address may be filled in later - possibly replacing `send(other_ip_field).present?` with a check that the other_ip_field is either: `present`, `host.compute_provides?(other_ip_field)`, or `mac_based_ipam?(other_ip_field)`.

2) Adding a new NIC with only an IPv6 subnet causes a validation error on the MAC address field on the form, even with a suitable CR set.

Edit: I'd be interested to know which combination of settings you have when testing this (e.g. v4/v6 subnet presence, DHCP/DNS orchestrations, which CR).

timogoebel:
@domcleal : Thanks for testing this and apologies, it didn't work. I must have misclicked or something. It does work now, I have double checked. I refactored the `required_ip_address` method into a class. I think, this looks a lot cleaner now. There is still room for improvement, though. Any suggestions?
We might want to change the tests accordingly, however I don't like refactoring tests and code that is tested in one commit.
I also got rid of the redundant `ip_available?` and `mac_available?` methods.

I do have a VMWare CR and a smart proxy with isc dhcp and bind to test this. I have tested with IPv6 only and a dual-stacked host for now.

app/services/nic_ip_required/base.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    def forward_dns?
+      domain.present? && domain.dns_id.present? && !other_ip_protocol_provides_ip?
+    end
+
+    def require_ip_for_dhcp?
+      subnet.present? && subnet.dhcp_id.present?
+    end
+
+    def require_ip_for_token?
```
domcleal: I think the original code named this incorrectly - it's checking if the IP is required _instead_ of a token - more comments inline.

app/services/nic_ip_required/base.rb
```diff
@@ -0,0 +1,65 @@
...
+      tokens_enabled? && token_required? && !other_ip_protocol_provides_ip?
+    end
+
+    def tokens_enabled?
```
domcleal: disabled (0 is off, >0 is on)

app/services/nic_ip_required/base.rb
```diff
@@ -0,0 +1,65 @@
...
+      Setting[:token_duration] == 0
+    end
+
+    def token_required?
```
domcleal: This might be called "token or IP required", or "unattended controller used?", as it defines whether the host will need some identifier to access UnattendedController - either use a token, or require an IP address.

app/services/nic_ip_required.rb
```diff
@@ -0,0 +1,8 @@
+module NicIpRequired
+  FIELDS = [:ip, :ip6].freeze
+
+  def self.new(field, *args)
+    raise ::Foreman::Exception.new(N_("Unknown address field - can't continue")) unless FIELDS.include?(field)
+    "NicIpRequired::#{field.to_s.camelize}".safe_constantize.new(*args)
```
domcleal: I think this magic is unnecessary - in both cases in Nic::Base you can just call `NicIpRequired::Ip6.new` or Ip4 as desired.

app/models/nic/base.rb
```diff
@@ -142,30 +142,28 @@ def host_managed?
...
-      raise ::Foreman::Exception.new("invalid ip field: %s", ip_field) unless ip_field == :ip || ip_field == :ip6
-      raise ::Foreman::Exception.new("invalid other ip field: %s", other_ip_field) unless other_ip_field == :ip || other_ip_field == :ip6
-
-      # if it's not managed there's nowhere to specify an IP anyway
-      return false unless self.host.managed? && self.managed? && self.provision?
-
-      # if the CR will provide an IP, then don't validate yet
-      return false if host.compute_provides?(ip_field)
+    def require_ip4_validation?(from_compute = true)
+      (@require_ip4_validation ||= NicIpRequired.new(:ip, :nic => self, :from_compute => from_compute)).required?
```
domcleal: `NicIpRequired::Ip4.new` would be much simpler (or `Ipv4` rather than `Ip4` for consistency with Subnet)

app/models/concerns/orchestration/compute.rb
```diff
@@ -158,6 +152,20 @@ def setComputeIP
 
   def delComputeIP; end
 
+  def setComputeIPAM
+    set_ip_address
+
+    unless required_ip_addresses_set?(false)
+      failure _('Failed to set IPs via IPAM for %s') % name
```
domcleal: It'd be useful to include more debug here to indicate which IP(s) is expected + missing, but I'm not certain on the best method. Perhaps by adding errors to the model in the `required_ip_addresses_set?` method and ensuring they're logged?

domcleal (changes requested):
Tests worked much better, thanks for the update.

timogoebel:
@domcleal : Thanks for your comments. I have rebased, improved the logging, added a test for this, removed the playful magic and followed your naming suggestions for the methods.

domcleal (approved):

domcleal:
Merged, thanks @timogoebel!

PR #3979, opened 2016-10-31
[WIP] Fixes #4509 - [vmware] Adds per volumes scsi controller

shlomizadok:


shlomizadok:
@mention-bot how about @timogoebel ? he could be a candidate too.

timogoebel:
Don't worry, I'll take a look anyway. ;-)

timogoebel:
@shlomizadok : Thanks for this PR. I think, this is much appreciated by users.

Some feedback from a brief look:
As a user I'd like to do something like this:

SCSI-Controller 1:
- Harddisk 1
- Harddisk 2
- Harddisk 3

SCSI-Controller 2:
- Harddisk 4
- Harddisk 5

So I think, this lacks the ability to add n harddisks to a scsi controller. Not just a 1:1 relationship.

shlomizadok:
I see your point. I see your point?

Will change :)
hmmm

timogoebel:
> Will change :)

Thanks. Let me know, if you need any help. Just a small hint: `key` is `key` here. The first scsi controller has id "1000", the second "1001", ...
To add a harddisk to a controller, just set `controller_key` to the the key of the scsci controller.

``` ruby
attrs = {
  :volumes => [
    {
      :storage_pod=>"Storage",
      :name=>"Hard disk 1",
      :size_gb=>"10",
      :thin=>"true",
      :eager_zero=>"false",
      :controller_key => 1000
    },
    {
      :storage_pod=>"Storage",
      :name=>"Hard disk 2",
      :size_gb=>"10",
      :thin=>"true",
      :eager_zero=>"false",
      :controller_key => 1001
    }
  ],
  :scsi_controllers => [
    {
      :type=>"VirtualLsiLogicController",
      :key => 1000
    },
    {
      :type=>"VirtualLsiLogicController",
      :key => 1001
    }
  ]
}
```

shlomizadok:
@timogoebel, I have no more hair because I have pulled all of my hair... :smile: 
In the UI, you can assign to which scsi controller you wish to add the new volume:
image

The problem: if you assign two or more volumes per controller, I get a "boot order" error. 
Any idea?

shlomizadok:
And btw, the hash sent to vmware is exactly as described in your comment.

timogoebel:
@shlomizadok : Oh no, does ist work if you comment this line? https://github.com/theforeman/foreman/blob/develop/app/models/compute_resources/foreman/model/vmware.rb#L506

shlomizadok:
@timogoebel - it does !
How is that related ? :thinking:

timogoebel:
@shlomizadok : Glad to hear that.
To set the boot order for vmware, you need the deviceKey for each harddisk. It's being calculated here:

https://github.com/fog/fog-vsphere/blob/master/lib/fog/vsphere/requests/compute/create_vm.rb#L152

There might be a bug there somewhere. I can't tell without your code.

shlomizadok:
@timogoebel - same code...

timogoebel:
@shlomizadok : Strangely, it works for me and I am unable to reproduce the error.

shlomizadok:
@timogoebel here is the current status:
When I mark out the `:boot_order` param, I can create volumes on separate scsi controllers:
image
And on vmware respectively:
image

However, it would fail when:
1. `:boot_order` parameter is set 
and
2. Different scsi controllers are chosen

Would love some help :)

timogoebel:
@shlomizadok : Could you please try if this helps?

https://github.com/fog/fog-vsphere/pull/58

> 1. Different scsi controllers are chosen

Would you mind to elaborate this? I can't follow. Do you mean different types?

shlomizadok:
I do mean  different types of scsi controller. 
When I use the same controller type for all volumes - it works fine.
When I use different type, it fails with: `InvalidArgument: A specified parameter was not correct. configSpec.bootOptions.bootOrder`

timogoebel:
@shlomizadok : Could you please check if https://github.com/timogoebel/fog-vsphere/commit/5e34396d331c6a6d30f71213ef7e2f95ecc94f8f solves the issue for you? I'll open a pr then.

shlomizadok:
@timogoebel - it works with https://github.com/timogoebel/fog-vsphere/commit/5e34396d331c6a6d30f71213ef7e2f95ecc94f8f

Gonna modify current tests and fix conflicts. Will you kindly open a PR with ^ ?

timogoebel:
@shlomizadok : The PR is already opened. Thanks for bearing with me, here. https://github.com/fog/fog-vsphere/pull/59

A couple of things I noticed:
The hash sent to fog still contains a `scsi_controller` (singular) key. It should not. I think, there is a default that needs to be removed.
I think, the UI needs to change a bit. Creating SCSI controllers should not directly be associated with creating volumes. You should be able to attach a volume to a controller after you created the controller. Similar to the vmware ui on your screenshots.

timogoebel:
@shlomizadok : I hacked together something, as a suggestion how I'd expect this to be. Don't expect everything to work, just a suggestion how this might look. Feel free to use whatever code you want.
What is missing is the whole ui implementation. It should be possible for a user to add a new controller and attach a volume to that controller. Therefore the list of controllers (read: the select box for each volume) has to be updated when a new controller is added.

https://github.com/timogoebel/foreman/tree/4509-vmware-scsi-controllers

This is how it looks:

image

Please let me know, what you think about this.

timogoebel:
@angystardust: I'd like to hear your opinion about this. Any comments from a users perspective? What would be the absolute necessities for you for this to be useful? :-)

shlomizadok:
@timogoebel - Thanks. Will you open a PR against my branch? I'll merge it, and we'll take it from there.
@angystardust, I'd love to hear you opinion too as we proceed :)

timogoebel:
@shlomizadok : Sure, done. https://github.com/shlomizadok/foreman/pull/8/files

angystardust:
Hi @timogoebel and @shlomizadok, thanks for paying attention to my humble opinion.
Unfortunately I can't test the patch right know (i've left my old company 2 months ago and i still haven't a foreman setup to play with) but I think that from a UI-prospective it's coming up pretty nice. 
Keep up the good work :+1: 

P.s.: just remember that we can't create more than 4 scsi controllers per vm ;)

timogoebel:
> P.s.: just remember that we can't create more than 4 scsi controllers per vm ;)

Yeah, thanks for the hint. I think, we should pay attention to the disk limits as well.

To summarize:

4 x SCSI Controllers
15 Disks per SCSI Controller

timogoebel:
Just for the record: @shlomizadok and I took a look at the current state. We agreed what needs to be done. Looking forward to an update. :-)
The code looks good so far, we just need to get the UI working. A user should be able to add up to four scsi controllers and choose the type. The controller keys have to start with 1000 and be increased for every controller. The user can then choose to which controller he wants to add a hard disk. This is defined by setting the controller key on the storage volume. When a new scsi controller is added or removed, a new entry has to be added or removed in all select boxes on the volumes.
It might be a good opportunity to do this with react.

timogoebel:
@shlomizadok : I just retested this. This looks promising, however it doesn't work. I'm setting this on WoC for now. Let me know, if I can assist in any way.

mmoll:
I guess http://projects.theforeman.org/issues/7803 would also be fixed with this?

timogoebel:
> I guess http://projects.theforeman.org/issues/7803 would also be fixed with this?

Yes.

ohadlevy:
I've created a POC for the UI using react, you can see it in action at https://ohadlevy.github.io/?selectedKind=VM%20Storage&selectedStory=multiple%20controllers&full=0&down=1&left=1&panelRight=0&downPanel=kadirahq%2Fstorybook-addon-actions%2Factions-panel codebase at https://github.com/ohadlevy/foreman/tree/vmware-scsi-controllers

shlomizadok:
I'll pick from there @ohadlevy

2017-01-10 13:08 GMT+02:00 Ohad Levy <notifications@github.com>:

gailsteiger:
It would make sense to make the /vms/storage Button.js, CheckboxForm.js, and SelectForm.js generic and move them to forms or formsControls foldeer in /components/common.

Since both CheckboxForm and SelectForm are form controls, the names are misleading. I suggest simply Checkbox and Select

webpack/assets/javascripts/react_app/components/hosts/vms/storage/CheckboxForm.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+const CheckboxForm = (props) => {
+  return (
+    <div className="clearfix">
+      <label className="col-md-2">{props.label}</label>
+      <input type="checkbox"
+             checked={props.size}
```
gailsteiger: prop.size? not intuitive prop name for checked

webpack/assets/javascripts/react_app/components/hosts/vms/storage/StorageContainer.js
```diff
@@ -0,0 +1,77 @@
...
+    VMStorageStore.removeChangeListener(this.onChange);
+    VMStorageStore.removeErrorListener(this.onError);
+  }
+  onChange(event) {
+    this.setState({ controllers: VMStorageStore.getControllers() });
+  }
+  onError(info) {
+    if (this.props.id === info.id) {
+      this.setState({
+        errorMessage: info.textStatus
```
gailsteiger: this.state.errorMessage doesn't seem to be used

webpack/assets/javascripts/react_app/constants.js
```diff
@@ -2,11 +2,38 @@ export const ACTIONS = {
...
+  DISK_UPDATED: 'DISK_UPDATED'
 };
 
 export const STATUS = {
   PENDING: 'PENDING',
   RESOLVED: 'RESOLVED',
   ERROR: 'ERROR'
 };
+
+export const VMStorageVMWare = {
```
gailsteiger: constant property format convention is all caps separated by underscores

webpack/assets/javascripts/react_app/stores/VMStorageStore.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+}
+
+const VMStorageStore = new VMStorageEventEmitter();
+
+AppDispatcher.register(action => {
+  switch (action.actionType) {
+    case ACTIONS.CONTROLLER_ADDED: {
+
+      if (_vmStorage.controllers.length < VMStorageVMWare.MaxControllers) {
```
gailsteiger: This check is unnecessary for CONTROLLER_ADDED or DISK_ADDED. This problem can be picked up before the action is dispatch:
option 1: disable button when maximum reached
option 2: check in VMStorageActions and don't dispatch is maximum reached

I prefer to reserve emitError for cases that can't be anticipated in advance

app/controllers/hosts_controller.rb
```diff
@@ -90,6 +90,8 @@ def clone
   end
 
   def create
+    require 'pry'
```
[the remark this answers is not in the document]
shlomizadok: @houndci-bot is that what bothers you on `require 'pry'` :rofl:

app/helpers/compute_resources_vms_helper.rb
```diff
@@ -199,4 +199,10 @@ def ovirt_vms_data
   def vm_delete_action(vm, authorizer = nil)
     display_delete_if_authorized(hash_for_compute_resource_vm_path(:compute_resource_id => @compute_resource, :id => vm.identity).merge(:auth_object => @compute_resource, :authorizer => authorizer), :class => 'btn btn-danger')
   end
+
+  def vsphere_scsi_controllers(compute)
+    scsi_controllers = {}
+    compute.scsi_controller_types.map { |type| scsi_controllers[type[:key]] = type[:title] }
```
ohadlevy: nitpick, substitute map with each

app/models/compute_resources/foreman/model/vmware.rb
```diff
@@ -335,6 +335,9 @@ def parse_args(args)
         args[collection] = nested_attributes_for(collection, nested_attrs) if nested_attrs
       end
 
+      # Backwards compatibility for e.g. API requests.
+      # User can set the scsi_controller_type attribute
+      # to define a single scsi controller by that type
       if args[:scsi_controller_type].present?
         args[:scsi_controller] = {:type => args.delete(:scsi_controller_type)}
```
ohadlevy: deprecation warning needed?

app/models/compute_resources/foreman/model/vmware.rb
```diff
@@ -528,7 +535,7 @@ def vm_instance_defaults
         :memory_mb  => 768,
         :interfaces => [new_interface],
         :volumes    => [new_volume],
-        :scsi_controller => { :type => scsi_controller_default_type },
+        :scsi_controllers => [{ :type => scsi_controller_default_type }],
```
ohadlevy, line 573: will this break any api?

app/views/compute_resources_vms/form/vmware/_scsi_controller.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= select_f f, :type, compute_resource.scsi_controller_types, :first, :last, {},
+  { :label => _("Controller type"),:class => "col-md-2", :disabled => !new_host } %>
+  
+<%= f.hidden_field :key, :value => f.object.key %>
+דדד
```
ohadlevy: unknown extra chars?

do we need this partial at all? I thought this is done via react now?

app/views/hosts/_compute_detail.html.erb
```diff
@@ -4,15 +4,5 @@
 <%= render :partial => provider_partial(compute_resource, 'base'),
             :locals => { :f => f, :compute_resource => compute_resource, :new_host => new_host }.merge(args_for_compute_resource_partial(@host)) %>
 
-<!--NICS-->
```
ohadlevy: why are you removing this code?

webpack/assets/javascripts/react_app/components/common/forms/Checkbox.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
+const Checkbox = (props) => {
+  return (
+    <div className="clearfix">
+      <div className="form-group">
+        <label className="col-md-2">{props.label}</label>
+        <div className="col-md-4">
+          <input type="checkbox"
+                 checked={props.size}
```
ohadlevy: whats props.size? should it be props.checked

webpack/assets/javascripts/react_app/components/common/forms/Select.js
```diff
@@ -0,0 +1,22 @@
...
+  return (
+    <div className="clearfix">
+      <div className="form-group">
+        <label className="col-md-2">{props.label}</label>
+        <div className="col-md-4">
+          <select className="form-control"
+                  value={props.value}
+                  onChange={props.onChange}
+          >
+            <option value="">Please select</option>
```
ohadlevy: string not extract, I would set a default value as an argument vs hard coded text.

webpack/assets/javascripts/react_app/components/common/forms/TextInput.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+
+const TextInput = (props) => {
+  return (
+    <div className="clearfix">
```
ohadlevy: since the div structure more or less repeat it self, does it make sense to have a common base component with all of that structure once vs in each component? for example:

```jsx
return (
<CommonForm>
  <input type...>
</CommonForm>
)
```

webpack/assets/javascripts/react_app/components/hosts/vms/storage/Controller.js
```diff
@@ -0,0 +1,85 @@
...
+  removeController(id, e) {
+    VMStorageActions.removeController(id);
+  }
+
+  render() {
+    return (
+      <div>
+      <div className="fields removable-item">
+        <Select
+          label={__('Create SCSI controller')}
```
ohadlevy, line 61: imho this will fail because of http://projects.theforeman.org/issues/17263

webpack/assets/javascripts/react_app/components/hosts/vms/storage/Controller.js
```diff
@@ -0,0 +1,85 @@
...
+          onChange={this.controllerUpdated.bind(this, 'type')}
+          options={this.selectableTypes()}
+        />
+        {this.disks()}
+      </div>
+        <div className="right">
+          <Button
+            disabled={this.props.disks.length >= VMStorageVMWare.MaxDisksPerController}
+            onClick={this.addDisk.bind(this, this.props.position)}>
+            {__('Add volume')}
```
ohadlevy, line 72: same comment about all extracted strings.

app/views/hosts/_compute_detail.html.erb
timogoebel: The VMWare UI looks weird if this is still in. And the form fields have been moved a long time ago to the interfaces tab. So I think, the note is not needed anymore.

webpack/assets/javascripts/react_app/components/hosts/vms/storage/Disk.js
```diff
@@ -0,0 +1,101 @@
...
+
+    attributes[attribute] = value;
+    VMStorageActions.updateDisk(this.props.controllerId, this.props.id, attributes);
+  }
+
+  updateDiskBooleanAttributes(attribute, event) {
+    let attributes = Object.assign({}, this.props);
+
+    attributes[attribute] = !this.props[attribute];
+    VMStorageActions.updateDisk(this.props.controllerId, this.props.id, attributes);
```
ohadlevy, line 47: do you really need all props? I would imagine you could be able to update one attribute at a time.

webpack/assets/javascripts/react_app/components/hosts/vms/storage/Disk.js
```diff
@@ -0,0 +1,101 @@
...
+      const key = Object.keys(attribute)[0];
+      const value = Object.values(attribute)[0];
+
+      return (<option key={key} value={key}>{value}</option>);
+    });
+  }
+
+  selectablePods() {
+    return this.props.storagePods.map((attribute) => {
+      const key = Object.keys(attribute)[0];
```
ohadlevy, line 28: DRY, extract to a common function?

webpack/assets/javascripts/react_app/components/hosts/vms/storage/StorageContainer.js
```diff
@@ -0,0 +1,110 @@
...
+              <Button
+                onClick={this.addController}
+                disabled={this.state.controllers.length >= VMStorageVMWare.MaxControllers} >
+                Add Controller
+              </Button>
+            </div>
+          </div>
+          <input
+            value={this.format()}
+            name="host[compute_attributes][scsi_controllers]"
```
ohadlevy, line 103: should be passed as a param from rails view imho.

webpack/assets/javascripts/react_app/stores/VMStorageStore.js
```diff
@@ -0,0 +1,94 @@
...
+    return _vmStorage.controllers;
+  }
+  getControllersCount() {
+    return this.getControllers().length;
+  }
+}
+
+const VMStorageStore = new VMStorageEventEmitter();
+
+AppDispatcher.register(action => {
```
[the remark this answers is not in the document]
ohadlevy: add /* eslint-enable max-statements */

.eslintrc
```diff
@@ -65,7 +65,7 @@
     }],
     "max-nested-callbacks": ["error", 5],
     "max-params": ["error", 3],
-    "max-statements": ["error", 10, { "ignoreTopLevelFunctions": true }],
+    "max-statements": ["error", 34, { "ignoreTopLevelFunctions": true }],
```
ohadlevy: not sure this is the right place for it, i would put it only on the required places if needed...? @gailsteiger ?

gailsteiger:
@shlomizadok , @ohadlevy max-statements is an important rule and 34 is too high. The solution is refactoring.

shlomizadok:
Hey @Rohoover,
In this PR I am adding a functionality to have SCSI controllers for vmware hosts.
Under each controller one can add attached volumes.

I'd love to have you help in arranging the form / buttons.
Especially in differentiating between controllers.

image


image


image

timogoebel:
@shlomizadok : I retested this PR. I do see an issue. When I save the host, it always resets the value of the storage pod (datastore cluster). This value is then wrong, and the vm won't save.

Rohoover:
@shlomizadok 

Here is a quick mockup I did that adjusts the layout. We can get even fancier if you want, but wanted to first offer a lower level of effort solution. 

Please feel free to provide feedback.

create controller volume_nomarkup
create controller volume

ares:
@shlomizadok could you please look at the rubocop failures?

shlomizadok:
Closing in favor of https://github.com/theforeman/foreman/pull/4366

PR #3980, opened 2016-11-01
Fixes #17152 - Puppet proxy dashboard displays correctly

tbrisker:


erezzarum:
Applied the patch, works.

lzap:
Yep, thanks. Works as expected.

PR #3981, opened 2016-11-01
Fixes #17153 - Consistent name for variable_type in the api

orrabin:
Both `parameter_type` and `variable_type` are aliases for `key_type`.
Leaving `parameter_type` for backward compatibility.


domcleal:
Merged, thanks @orrabin.

PR #3982, opened 2016-11-01
Fixes #17151 - sort PXE labels by hostgroup title

amirfefer:
The title includes the whole path, sorting by hostgroup's name may result with unsorted nested hostgroups.

### DIFF
--- a/test/controllers/provisioning_templates_controller_test.rb
+++ b/test/controllers/provisioning_templates_controller_test.rb
@@ -144,6 +144,16 @@ class ProvisioningTemplatesControllerTest < ActionController::TestCase
       get :build_pxe_default, {}, set_session_user
       assert_redirected_to provisioning_templates_path
     end
+
+    test "pxe menu's labels should be sorted by full hostgroup title" do
+      first = FactoryGirl.build(:hostgroup, :with_parent, :name => "def", :operatingsystem => operatingsystems(:centos5_3), :architecture => architectures(:x86_64), :medium => media(:one))
+      second = FactoryGirl.build(:hostgroup, :with_parent, :name => "abc", :operatingsystem => operatingsystems(:centos5_3), :architecture => architectures(:x86_64), :medium => media(:one))
+      FactoryGirl.create(:template_combination, :provisioning_template => templates(:mystring2), :hostgroup => second)
+      FactoryGirl.create(:template_combination, :provisioning_template => templates(:mystring2), :hostgroup => first)
+      ProxyAPI::TFTP.any_instance.expects(:create_default).with(regexp_matches(/^PXE.*/), has_entry(:menu, regexp_matches(/#{first.name}.*#{second.name}/m))).returns(true).times(3)
+      get :build_pxe_default, {}, set_session_user
+      assert_redirected_to provisioning_templates_path
+    end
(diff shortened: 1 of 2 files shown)


domcleal:
There are some test failures, so setting WoC. If this is caused by the new fixture, it would be preferable to use factories for tests that need the additional host group rather than modifying fixtures.

Please also check the commit message, I don't think you mean hostnames, I think this refers to host group names.

test/controllers/provisioning_templates_controller_test.rb
```diff
@@ -144,6 +144,18 @@ class ProvisioningTemplatesControllerTest < ActionController::TestCase
       get :build_pxe_default, {}, set_session_user
       assert_redirected_to provisioning_templates_path
     end
+
+    test "pxe menu's labels should be sorted by full hostname title" do
+      t1 = TemplateCombination.new :hostgroup => hostgroups(:inherited), :environment => environments(:production)
+      t1.provisioning_template = templates(:mystring2)
+      t2 = TemplateCombination.new :hostgroup => hostgroups(:nested), :environment => environments(:production)
+      t2.provisioning_template = templates(:mystring2)
+      t1.save
```
domcleal: Use `TemplateCombination.create` to avoid all of this editing and saving, or preferably just use the :template_combination factory.

test/controllers/provisioning_templates_controller_test.rb
```diff
@@ -144,6 +144,18 @@ class ProvisioningTemplatesControllerTest < ActionController::TestCase
       get :build_pxe_default, {}, set_session_user
       assert_redirected_to provisioning_templates_path
     end
+
+    test "pxe menu's labels should be sorted by full hostname title" do
```
domcleal: hostgroup

test/controllers/provisioning_templates_controller_test.rb
```diff
@@ -144,6 +144,18 @@ class ProvisioningTemplatesControllerTest < ActionController::TestCase
...
+      t1 = TemplateCombination.new :hostgroup => hostgroups(:inherited), :environment => environments(:production)
+      t1.provisioning_template = templates(:mystring2)
+      t2 = TemplateCombination.new :hostgroup => hostgroups(:nested), :environment => environments(:production)
+      t2.provisioning_template = templates(:mystring2)
+      t1.save
+      t2.save
+      ProxyAPI::TFTP.any_instance.expects(:create_default).with(regexp_matches(/^PXE.*/), has_entry(:menu, regexp_matches(/#{hostgroups(:nested).name}.*#{hostgroups(:inherited).name}/m))).returns(true).times(3)
```
domcleal: If you really intend to test the ordering, explicitly create host groups with names above that will exercise it - it's not clear from the fixtures that the names are meant to be sorted this way.

domcleal (changes requested):

amirfefer:
@domcleal  Thanks for your review, I've changed the test by using factory.

domcleal (approved):

domcleal:
Merged, thanks @amirfefer.

lzap:
Nice one, this was driving me crazy :-)

PR #3983, opened 2016-11-01
Fixes #16740 - Access host params through macro

xprazak2:
The host params can now be accessed using host_param macro in templates.
There is also host_info, that gives access to host.info.
It seems only parameters are needed most of the time, so I decided
to go for 2 methods. Both methods call host.info on first use and subsequently
use cached result of the first call.

Usage: ~~host_param 'puppetmaster'
is equivalent to: host_info 'parameters', 'puppetmaster'~~

`host_enc` instead of `@host.info`
`host_param_true?` instead of `@host.param_true?`
`host_param_false?` instead of `@host.param_false?`
`host_param('param_name')` instead of `@host.params['param_name']`

### DIFF
--- a/app/models/concerns/host_common.rb
+++ b/app/models/concerns/host_common.rb
@@ -146,10 +146,12 @@ module HostCommon
   end
 
   def param_true?(name)
+    Foreman::Deprecation.renderer_deprecation('1.17', __method__, 'host_param_true?')
     params.has_key?(name) && Foreman::Cast.to_bool(params[name])
   end
 
   def param_false?(name)
+    Foreman::Deprecation.renderer_deprecation('1.17', __method__, 'host_param_false?')
     params.has_key?(name) && Foreman::Cast.to_bool(params[name]) == false
   end
 
--- a/app/models/host/managed.rb
+++ b/app/models/host/managed.rb
@@ -383,6 +383,10 @@ class Host::Managed < Host::Base
   # rubocop:disable Metrics/PerceivedComplexity
   # rubocop:disable Metrics/CyclomaticComplexity
   def info
+    renderer_regex = /renderer\.rb.*host_enc/
+    unless caller.first.match(renderer_regex) || caller[1].match(renderer_regex)
+      Foreman::Deprecation.renderer_deprecation('1.17', __method__, 'host_enc')
+    end
     # Static parameters
     param = {}
     # maybe these should be moved to the common parameters, leaving them in for now
@@ -422,16 +426,8 @@ class Host::Managed < Host::Base
     # Parse ERB values contained in the parameters
     param = SafeRender.new(:variables => { :host => self }).parse(param)
 
-    classes = if self.environment.nil?
-                []
-              elsif Setting[:Parametrized_Classes_in_ENC] && Setting[:Enable_Smart_Variables_in_ENC]
-                lookup_keys_class_params
-              else
-                self.puppetclasses_names
-              end
-
     info_hash = {}
-    info_hash['classes'] = classes
+    info_hash['classes'] = self.enc_puppetclasses
     info_hash['parameters'] = param
     info_hash['environment'] = param["foreman_env"] if Setting["enc_environment"] && param["foreman_env"]
 
@@ -928,6 +924,16 @@ class Host::Managed < Host::Base
   def firmware_type
     return unless pxe_loader.present?
     Operatingsystem.firmware_type(pxe_loader)
+  end
+
+  def enc_puppetclasses
+    if self.environment.nil?
+      []
+    elsif Setting[:Parametrized_Classes_in_ENC] && Setting[:Enable_Smart_Variables_in_ENC]
+      lookup_keys_class_params
+    else
+      self.puppetclasses_names
+    end
   end
 
   private
--- a/app/services/foreman/deprecation.rb
+++ b/app/services/foreman/deprecation.rb
@@ -2,12 +2,28 @@ module Foreman
   class Deprecation
     #deadline_version - is the version the deprecation is going to be deleted, the format must be a major release e.g "1.8"
     def self.deprecation_warning(foreman_version_deadline, info)
-      raise Foreman::Exception.new(N_("Invalid version format, please enter in x.y (only major version).")) unless foreman_version_deadline.to_s.match(/\A\d[.]\d+\z/)
+      check_version_format foreman_version_deadline
       ActiveSupport::Deprecation.warn("You are using a deprecated behavior, it will be removed in version #{foreman_version_deadline}, #{info}", caller(2))
+    end
+
+    def self.check_version_format(foreman_version_deadline)
+      raise Foreman::Exception.new(N_("Invalid version format, please enter in x.y (only major version).")) unless foreman_version_deadline.to_s.match(/\A\d[.]\d+\z/)
     end
 
     def self.api_deprecation_warning(info)
       ActiveSupport::Deprecation.warn("Your API call uses deprecated behavior, #{info}", caller)
+    end
+
+    def self.renderer_deprecation(foreman_version_deadline, method, new_method)
+      check_version_format foreman_version_deadline
+      called_from_params = false
+      caller.each_with_index do |item, index|
+        called_from_params = true if item.match(/host_params\.rb.*params/)
+        return if called_from_params && item.match(/managed\.rb.*info/)
+        next unless item.match(/renderer\.rb.*render_safe/)
+        Rails.logger.warn "DEPRECATION WARNING: you are using deprecated @host.#{method} in a template, it will be removed in #{foreman_version_deadline}. Use #{new_method} instead."
+        return
+      end
     end
   end
 end
--- a/lib/foreman/renderer.rb
+++ b/lib/foreman/renderer.rb
@@ -14,7 +14,9 @@ module Foreman
                                   :foreman_server_url, :log_debug, :log_info, :log_warn, :log_error, :log_fatal, :template_name, :dns_lookup,
                                   :pxe_kernel_options ]
     ALLOWED_HOST_HELPERS ||= [ :grub_pass, :ks_console, :root_pass,
-                               :media_path, :param_true?, :param_false?, :match ]
+                               :media_path, :param_true?, :param_false?, :match,
+                               :host_param_true?, :host_param_false?,
+                               :host_param, :host_puppet_classes, :host_enc ]
 
     ALLOWED_HELPERS ||= ALLOWED_GENERIC_HELPERS + ALLOWED_HOST_HELPERS
 
@@ -24,6 +26,30 @@ module Foreman
 
     def template_logger
       @template_logger ||= Foreman::Logging.logger('templates')
+    end
+
+    def host_enc(*path)
+      @enc ||= @host.info.deep_dup
+      return @enc if path.compact.empty?
+      enc = @enc
+      path.each { |step| enc = enc.fetch step }
+      enc
+    end
+
+    def host_param(param_name)
+      @host.params[param_name]
+    end
+
+    def host_puppet_classes
+      @host.puppetclasses
+    end
+
+    def host_param_true?(name)
+      @host.params.has_key?(name) && Foreman::Cast.to_bool(@host.params[name])
+    end
+
+    def host_param_false?(name)
+      @host.params.has_key?(name) && Foreman::Cast.to_bool(@host.params[name]) == false
     end
 
     def render_safe(template, allowed_methods = [], allowed_vars = {})
--- a/test/unit/foreman/renderer_deprecation_test.rb
+++ b/test/unit/foreman/renderer_deprecation_test.rb
@@ -1,0 +1,39 @@
+require 'test_helper'
+
+class RendererTest < ActiveSupport::TestCase
+  class DummyRenderer
+    attr_accessor :host
+
+    include Foreman::Renderer
+  end
+
+  test 'should show the deprecation for @host.info' do
+    renderer = DummyRenderer.new
+    renderer.host = FactoryGirl.create(:host, :with_puppet)
+    template = FactoryGirl.create(:provisioning_template, :template => '<%= @host.info %>')
+    Rails.logger.expects(:warn).with("DEPRECATION WARNING: you are using deprecated @host.info in a template, it will be removed in 1.17. Use host_enc instead.").once
+    renderer.unattended_render template
+  end
+
+  test 'should show the deprecation for @host.params' do
+    renderer = DummyRenderer.new
+    renderer.host = FactoryGirl.create(:host, :with_puppet)
+    template = FactoryGirl.create(:provisioning_template, :template => '<%= @host.params %>')
+    Rails.logger.expects(:warn).with("DEPRECATION WARNING: you are using deprecated @host.params in a template, it will be removed in 1.17. Use host_param instead.").once
+    renderer.unattended_render template
+  end
+
+  test 'should render host param using "host_param" helper without deprecation' do
+    renderer = DummyRenderer.new
+    renderer.host = FactoryGirl.create(:host, :with_puppet)
+    Rails.logger.expects(:warn).never
+    assert renderer.render_safe("<%= host_param('test') %>", DummyRenderer::ALLOWED_HELPERS).present?
+  end
+
+  test 'should render host info using "host_enc" helper without deprecation' do
+    renderer = DummyRenderer.new
+    renderer.host = FactoryGirl.create(:host, :with_puppet)
+    Rails.logger.expects(:warn).never
+    assert renderer.render_safe("<%= host_enc %>", DummyRenderer::ALLOWED_HELPERS).present?
+  end
+end
--- a/test/unit/foreman/renderer_test.rb
+++ b/test/unit/foreman/renderer_test.rb
@@ -221,4 +221,37 @@ class RendererTest < ActiveSupport::TestCase
     @renderer.instance_variable_set '@whatever_random_name', 'has_value'
     assert_equal({ :whatever_random_name => 'has_value' }, @renderer.send(:allowed_variables_mapping, [ :whatever_random_name ]))
   end
+
+  test 'should render puppetclasses using host_puppetclasses helper' do
+    @renderer.host = FactoryGirl.create(:host, :with_puppetclass)
+    assert @renderer.host_puppet_classes
+  end
+
+  test 'should render host param using "host_param" helper' do
+    @renderer.host = FactoryGirl.create(:host, :with_puppet)
+    assert @renderer.host_param('test').present?
+  end
+
+  test 'should have host_param_true? helper' do
+    @renderer.host = FactoryGirl.create(:host, :with_puppet)
+    FactoryGirl.create(:parameter, :name => 'true_param', :value => "true")
+    assert @renderer.host_param_true?('true_param')
+  end
+
+  test 'should have host_param_false? helper' do
+    @renderer.host = FactoryGirl.create(:host, :with_puppet)
+    FactoryGirl.create(:parameter, :name => 'false_param', :value => "false")
+    assert @renderer.host_param_false?('false_param')
+  end
+
+  test 'should have host_enc helper' do
+    @renderer.host = FactoryGirl.create(:host, :with_puppet)
+    assert @renderer.host_enc
+  end
+
+  test "should find path in host_enc" do
+    host = FactoryGirl.create(:host, :with_puppet)
+    @renderer.host = host
+    assert_equal host.puppetmaster, @renderer.host_enc('parameters', 'puppetmaster')
+  end
 end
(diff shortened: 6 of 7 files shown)


lib/foreman/renderer.rb
```diff
@@ -26,6 +26,18 @@ def template_logger
...
     end
 
+    def host_info(*path)
+      @info ||= @host.info
+      info = @info
+      path.each { |step| info = info.fetch step }
+      info
+    end
+
+    def host_param(*path)
```
domcleal: Isn't this nearly the same as `@host.params` (plus param_true? etc), why's it needed and why use `@host.info` (the ENC method)?

domcleal (changes requested):
Seems to duplicate existing permitted method, plus a rubocop failure.

lib/foreman/renderer.rb
xprazak2: Yes, it is basically the same. The main requirement behind this was to access both params and ENC output in unified way. Also not to call methods on `@host` directly.

ares: and to add lazy cache of values, since the the enc takes time and users don't know that they can keep it in some local variable

domcleal: I'm not sure I see the advantage then, it seems like a duplication with a different interface style - it'll just be confusing, not helpful. Caching can be added to `@host.info` if you want.

ares: I think we agreed some time ago that we want to avoid using instance variables directly such as `@host`. IMHO calling their methods is also problematic. Things like `@host.info` or `@host.operatingsystem` are internal Foreman methods, if we encourage people to use them in templates, we can't modify them without templates modifications. Imagine we would like to rename `Host#operatingsystem` to `Host#operating_system`. Yes, macros can be seen as a duplication but they isolate template writers from internal changes and help to define templates API.

There are other benefits of macros over internal methods, e.g. it's easier to extend macros from plugins than modifying `@host` variable. Also we could better support previewing since the method can contain logic to handle nil objects so if `@host` is nil we could render just string '$host' or something like that.

domcleal: OK, I agree it's preferable over using `@host`, but the change needs to:
1. Also include replacements for `param_true?` and `param_false?`, since these are used on `@host`.
2. Preferably include a deprecation warning for the old methods when called from templates (perhaps set a thread variable during rendering, then it's easy to only deprecate when called that way) so we can indicate which method is preferred.

lib/foreman/renderer.rb
```diff
@@ -26,6 +26,18 @@ def template_logger
...
+      path.unshift 'parameters'
+      host_info(*path)
```
domcleal: I think this should be using `@host.params`, not calling .info, that's only for the Puppet ENC output.

test/unit/foreman/renderer_test.rb
```diff
@@ -221,4 +221,11 @@ def setup_safemode_renderer
     @renderer.instance_variable_set '@whatever_random_name', 'has_value'
     assert_equal({ :whatever_random_name => 'has_value' }, @renderer.send(:allowed_variables_mapping, [ :whatever_random_name ]))
   end
+
+  test 'should render host param' do
+    @renderer.host = FactoryGirl.create(:host, :with_puppet)
+    assert_nil @renderer.instance_variable_get('@info')
+    assert @renderer.host_param('puppetmaster').present?
```
domcleal: This should be testing `host_info` if you're checking `@info` - have a separate test for parameters.

lib/foreman/renderer.rb
```diff
@@ -26,6 +26,18 @@ def template_logger
       @template_logger ||= Foreman::Logging.logger('templates')
     end
 
+    def host_info(*path)
```
domcleal: Can you choose a better name for this, or even remove it if unnecessary? The model method isn't well-named, it's specifically the Puppet ENC output, not "info".

I don't really see much use for the Puppet ENC output, I'd prefer to see more specific information like `host_param`, so if you're trying to provide a list of classes to the template for example, a helper to return host classes would be better.

domcleal (changes requested):
1. Both method names need adding to the allowed **host** helpers to be permitted in template renders and show in help.
2. Should have a deprecation warning on the old methods when called from template renders, see inline comment.
3. Missing param_true/false helpers, if you're going to move `@host.params` itself.

xprazak2:
I added `host_param_true?`, `host_param_false?` and `host_puppet_classes` helpers. I renamed `host_info` to `host_enc` and kept it in, because some templates are using it. A deprecation message is logged in each time one of `:info, :params, :param_true?, :param_false?` is called in a template.

app/models/host/managed.rb
```diff
@@ -916,6 +908,16 @@ def apply_compute_profile(modification)
     modification.run(self, compute_resource.try(:compute_profile_for, compute_profile_id))
   end
 
+  def puppet_classes
```
domcleal: This is very similar to `#puppetclasses` from the association, but it returns very different structures. I'd suggest naming it `enc_puppetclasses` or similar.

lib/foreman/renderer.rb
```diff
@@ -26,14 +28,42 @@ def template_logger
       @template_logger ||= Foreman::Logging.logger('templates')
     end
 
+    def host_enc(*path)
+      @enc ||= @host.info
+      return @enc.deep_dup if path.compact.empty?
+      enc = @enc
+      path.each { |step| enc = enc.fetch step }
+      enc
```
domcleal, line 36: No duplication here, but it does above if there is no path? This should probably dup too.

lib/foreman/renderer.rb
```diff
@@ -269,5 +299,99 @@ def pxe_config
       @kernel = @host.operatingsystem.kernel @host.arch
       @initrd = @host.operatingsystem.initrd @host.arch
     end
+
+    # registeres host to be observed for call of deprecated methods
+    def register_method_deprecations_for(host)
+      return if !@host || !@host.respond_to?(:info) || @host.respond_to?(:monitored_for_deprecations)
```
domcleal: What's the `.respond_to?(:info)` checking specifically? For a specific type of host?

lib/foreman/renderer.rb
```diff
@@ -269,5 +299,99 @@ def pxe_config
...
+      host.class_eval { include MethodDeprecations }
```
domcleal: If using a mixin, it should be included in Host always rather than at runtime - if it introduces a behaviour change for these methods, tests won't be reliable or indicative if they're not always run either with it. If this is never included in tests, then this introduces a lot of untested code.

lib/foreman/renderer.rb
```diff
@@ -269,5 +299,99 @@ def pxe_config
...
+            define_method new_method_name do |host|
+              stash_to_thread m
+              self.send(old_method_name, host)
+            end
+          end
+        end
+
+        def define_methods_without_args(new_name, old_name)
+          METHODS.each do |m|
+            define_method "#{m}_#{new_name}" do
```
domcleal: Instead of redefining these methods here, could the deprecation warnings just be added inside the methods? e.g.

``` ruby
def info
  renderer_deprecation("Use host_enc instead of @host.info") do

  end
end
```

It might be easier to maintain that way and avoid redefining classes dynamically.

(I'll avoid reviewing this module in detail until your response to this point.)

domcleal (changes requested):

lib/foreman/renderer.rb
xprazak2: The mixin does not really change the behaviour of the method. Redefined methods just take note the original method is called and call it, they are also covered with tests in `UnattendedControllerTest` to the same extent as the original methods.

lib/foreman/renderer.rb
xprazak2: There may be a hostgroup in `@host`, which causes [test failures](https://github.com/theforeman/foreman/blob/develop/test/controllers/unattended_controller_test.rb#L238).

lib/foreman/renderer.rb
xprazak2: If we did this, code related to one set of deprecations would be in multiple places. In current version, it is contained inside `Renderer` module. Moreover, it would take note each time `info` method was called (like [here](https://github.com/theforeman/foreman/blob/develop/app/controllers/hosts_controller.rb#L71) or [here](https://github.com/theforeman/foreman/blob/develop/app/controllers/api/v2/hosts_controller.rb#L146)), but we do not want to deprecate method itself, only its usage in templates. 

I extend the `@host` object directly before it is used for rendering and there is nothing else going on with `@host` after the template is rendered, so the methods are effectively redefined only inside `render_safe`. I agree that dynamic method definition is not the easiest thing to read, but I think it makes sense here for the reasons mentioned above.

ares: I understand the DRY point as well as detection of calling this method only during rendering. But I agree with @domcleal, this is complex code. For readability and maintainability I suggest trying to just add deprecation logs to those methods and use `caller` to determine whether it was called in `render_safe` method. The sooner we remove this from Host jail and all templates the better. Let's try how it would that work on one method first and see if feels better.

domcleal: > Moreover, it would take note each time info method was called (like here or here), but we do not want to deprecate method itself, only its usage in templates.

My earlier suggestion was to use a thread variable for this - set a thread variable in the renderer, unset it in an `ensure` block, and then the hypothetical `renderer_deprecation` method in the above example can check if the variable is present.

ares: ok, that would also work and would be probably more reliable, I just don't like thread variables as they are similar to global variables and could cause problems in tests :-) I'd be fine with whatever @xprazak2 prefers

xprazak2: Thanks for your comments, I added a commit where I use `caller` to check if `:info` was called in `render_safe`. I also use `Foreman::Deprecation`. The main benefits I see here are better readability and less lines of code. We should also be able to find these deprecations by greping the files (I put '1.16' there for now just as an example). Let me know if I should proceed with this solution.

domcleal: It looks preferable to me, please do finish it.

ares: +1

xprazak2:
I added newer deprecation style to the remaining methods.

domcleal (changes requested):
Please check RendererTest test failures and also rebase to fix integration test failures.

app/services/foreman/deprecation.rb
```diff
@@ -2,12 +2,31 @@ module Foreman
...
+    end
+
     def self.api_deprecation_warning(info)
       ActiveSupport::Deprecation.warn("Your API call uses deprecated behavior, #{info}", caller)
     end
+
+    def self.renderer_deprecation(foreman_version_deadline, method)
+      check_version_format foreman_version_deadline
+      macros = {
+        :info => 'host_enc',
```
domcleal: Would it be easier to pass this in as a third argument from the caller?

lib/foreman/renderer.rb
```diff
@@ -26,6 +28,30 @@ def template_logger
       @template_logger ||= Foreman::Logging.logger('templates')
     end
 
+    def host_enc(*path)
+      @enc ||= @host.info.deep_dup
+      return @enc if path.compact.empty?
+      enc = @enc
+      path.each { |step| enc = enc.fetch step }
```
domcleal, line 35: Missing test coverage of this branch, it only covers `path.compact.empty?`.

app/services/foreman/deprecation.rb
```diff
@@ -2,12 +2,31 @@ module Foreman
...
 
+    def self.check_version_format(foreman_version_deadline)
+      raise Foreman::Exception.new(N_("Invalid version format, please enter in x.y (only major version).")) unless foreman_version_deadline.to_s.match(/\A\d[.]\d+\z/)
+    end
+
     def self.api_deprecation_warning(info)
       ActiveSupport::Deprecation.warn("Your API call uses deprecated behavior, #{info}", caller)
     end
+
+    def self.renderer_deprecation(foreman_version_deadline, method)
```
domcleal: Missing test coverage in deprecation_test.rb? The caller matching is the key bit, something to instantiate a renderer and then check the logger's called is key.

domcleal (changes requested):
Needs test coverage improving, otherwise looks right.

xprazak2:
The macro name is now passed into `render_deprecation` as a third argument and I added the tests . `params` method is called from `@host.info` as well, so there is a minor change in logic to avoid needless false positives.

test/unit/foreman/renderer_deprecation_test.rb
```diff
@@ -0,0 +1,37 @@
...
+    attr_accessor :host
+
+    include Foreman::Renderer
+  end
+
+  setup do
+    @renderer = DummyRenderer.new
+    @renderer.host = FactoryGirl.create(:host, :with_puppet)
+    @logger = Rails.logger
+    Rails.logger = MiniTest::Mock.new
```
domcleal: Please use `mock()` from mocha, we don't use MiniTest::Mock.

test/unit/foreman/renderer_deprecation_test.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  teardown do
+    Rails.logger = @logger
+  end
+
+  test 'should show the deprecation for @host.info' do
+    template = templates(:rendering_test_template)
+    Rails.logger.expect :warn, true, ["DEPRECATION WARNING: you are using deprecated @host.info in a template. Use host_enc instead."]
```
domcleal: ditto, use mocha (`expects`, not minitest's `expect`)

test/unit/foreman/renderer_deprecation_test.rb
```diff
@@ -0,0 +1,37 @@
...
+    @renderer.unattended_render template
+    Rails.logger.verify
```
domcleal: Not required with mocha.

test/unit/foreman/renderer_deprecation_test.rb
```diff
@@ -0,0 +1,37 @@
+require 'test_helper'
+
+class RendererTest < ActiveSupport::TestCase
+  def clear_current_user
```
domcleal: What's this override for? It doesn't seem to make any difference to the test.

test/fixtures/templates.yml
```diff
@@ -93,3 +93,19 @@ locked:
   operatingsystems: centos5_3
   locked: true
   type: ProvisioningTemplate
+
+rendering_test_template:
```
domcleal: Use a factory instead of adding a new fixture per test. Simply `template = FactoryGirl.build(:provisioning_template, template: '<%= @host.info %>')` in the test.

app/services/foreman/deprecation.rb
```diff
@@ -2,12 +2,28 @@ module Foreman
...
     end
+
+    def self.renderer_deprecation(foreman_version_deadline, method, new_method)
+      check_version_format foreman_version_deadline
+      called_from_params = false
+      caller.each_with_index do |item, index|
+        called_from_params = true if item.match(/host_params\.rb.*params/)
+        return if called_from_params && item.match(/managed\.rb.*info/)
+        next unless item.match(/renderer\.rb.*render_safe/)
+        Rails.logger.warn "DEPRECATION WARNING: you are using deprecated @host.#{method} in a template. Use #{new_method} instead."
```
domcleal: `foreman_version_deadline` appears to be unused, is it meant to be in this message?

domcleal (changes requested):

domcleal:
Please reply/leave a comment if a PR needs re-reviewing, this has been updated but is failing tests.

xprazak2:
I addressed the previous comments, trying to fix the tests.

xprazak2:
Tests are green, could I ask for a re-review pls?

xprazak2:
[test foreman]

test/unit/foreman/renderer_deprecation_test.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    include Foreman::Renderer
+  end
+
+  test 'should show the deprecation for @host.info' do
+    renderer = DummyRenderer.new
+    renderer.host = FactoryGirl.create(:host, :with_puppet)
+    template = FactoryGirl.create(:provisioning_template, :template => '<%= @host.info %>')
+    logger = Rails.logger
+    Rails.logger = mock()
```
domcleal: A mock object isn't really required here, you can set the expectation on the existing Rails.logger. This removes the need to replace it and any errors from not restoring the original correctly (e.g. it's currently missing an ensure block).

test/unit/foreman/renderer_deprecation_test.rb
```diff
@@ -0,0 +1,31 @@
...
+    renderer.unattended_render template
+    Rails.logger = logger
+  end
+
+  test 'should show the deprecation for @host.params' do
+    renderer = DummyRenderer.new
+    renderer.host = FactoryGirl.create(:host, :with_puppet)
+    template = FactoryGirl.create(:provisioning_template, :template => '<%= @host.params %>')
+    logger = Rails.logger
+    Rails.logger = mock()
```
domcleal: ditto, remove these two lines and the `Rails.logger =` line below, just keep .expects.

app/models/concerns/host_common.rb
```diff
@@ -146,10 +146,12 @@ def crypt_root_pass
   end
 
   def param_true?(name)
+    Foreman::Deprecation.renderer_deprecation('1.16', __method__, 'host_param_true?')
```
domcleal: 1.14 was branched last week, this should now be 1.17 as it's unlikely to be included in 1.14 now.

domcleal (changes requested):

xprazak2:
I removed the mock object and changed the message to include 1.17 instead of 1.16

test/unit/foreman/renderer_deprecation_test.rb
```diff
@@ -0,0 +1,25 @@
...
+    attr_accessor :host
+
+    include Foreman::Renderer
+  end
+
+  test 'should show the deprecation for @host.info' do
+    renderer = DummyRenderer.new
+    renderer.host = FactoryGirl.create(:host, :with_puppet)
+    template = FactoryGirl.create(:provisioning_template, :template => '<%= @host.info %>')
+    Rails.logger.expects(:warn).with("DEPRECATION WARNING: you are using deprecated @host.info in a template, it will be removed in 1.16. Use host_enc instead.").once
```
domcleal: References 1.16 here, probably causing the test failures.

domcleal (changes requested):
Tests are not passing.

xprazak2:
I removed the reference to 1.16 which fixed the tests. Thanks!

domcleal:
[test foreman] due to a broken Jenkins/slave connection.

xprazak2:
[test foreman]

xprazak2:
[test upgrade]

app/services/foreman/deprecation.rb
```diff
@@ -2,12 +2,28 @@ module Foreman
...
       ActiveSupport::Deprecation.warn("Your API call uses deprecated behavior, #{info}", caller)
     end
+
+    def self.renderer_deprecation(foreman_version_deadline, method, new_method)
+      check_version_format foreman_version_deadline
+      called_from_params = false
+      caller.each_with_index do |item, index|
+        called_from_params = true if item.match(/host_params\.rb.*params/)
+        return if called_from_params && item.match(/managed\.rb.*info/)
+        next unless item.match(/renderer\.rb.*render_safe/)
```
domcleal, line 23: When using `host_param` in a template, two deprecation warnings seem to be logged. Please check this works properly. A simple unit test that fails:

    test 'should render host param using "host_param" helper without deprecation' do
      @renderer.host = FactoryGirl.create(:host, :with_puppet)
      Rails.logger.expects(:warn).never
      assert @renderer.render_safe("<%= host_param('test') %>", DummyRenderer::ALLOWED_HELPERS).present?
    end

The logic on this line seems incorrect to me, it skips the deprecation warning _only_ from render_safe. Surely any other entry in the callstack would trigger this?

I'm unsure why template rendering triggers two deprecation warnings without further investigation, perhaps something else internally is calling `@host.params` and triggering a second. Please check it's working correctly in real usage.

domcleal (changes requested):
Deprecation warning being logged when using new `host_param` method.

app/services/foreman/deprecation.rb
xprazak2: I fixed the problem with `@host.params` being called internally, similar thing was happening to `host_enc` as well. I added tests to cover these cases.

domcleal:
Thanks, the issue with the warnings appears to be resolved.

> Usage: host_param 'puppetmaster'
> is equivalent to: host_info 'parameters', 'puppetmaster'

Just re-reading the description, is this accurate? Was it? I'm just trying to check that the described issue is resolved - and I'm unsure if whether it's referring to accessing ENC top-level parameters via host_param (and if this is even desirable).

xprazak2:
Not sure if it was accurate but it is not any more. I updated the first comment to reflect the current state.

domcleal:
Merged, thanks @xprazak2.

domcleal:
Please also consider updating community-templates, or at least filing an issue for the necessary updates.

PR #3984, opened 2016-11-01
Fixes #16739 - unify parameters permissions

ares:
This unifies `view_globals` used for global parameters with `view_params` used for the rest of parameters. And all other similar permissions of course. Shared and only resource type for such filters is `Parameter`. It also fixes the autocompletion for parameter filters so you can define a fllter like this

    "name ~ a* or domain_name ~ example*com"

other resources can be used too, e.g. host_name, hostgroup_name, subnet_name etc. One can also filter on type so to limit to global parameters only we can use `type = CommonParameter` condition. And finally, it adds integration test covering granular permissions for user.

The PR includes the migration for existing CommonParameter filters.


domcleal:
Setting WoC as there are Foreman, upgrade and rubocop test failures.

ares:
Thanks @domcleal, all is fixed now. Katello failure seems unrelated

```
:48:53   7) Error:
13:48:53 Katello::SyncManagementControllerTest#test_index:
13:48:53 Sprockets::FileNotFound: couldn't find file 'bastion/underscore/underscore' with type 'application/javascript'
13:48:53 Checked in these paths: 
```

[test katello]

Gemfile
```diff
@@ -13,7 +13,7 @@ gem 'ldap_fluff', '>= 0.3.5', '< 1.0'
 gem 'apipie-rails', '~> 0.3.4'
 gem 'rabl', '>= 0.11', "< 0.13"
 gem 'oauth', '~> 0.4'
-gem 'deep_cloneable', '~> 2.0'
+gem 'deep_cloneable', '~> 2.2.2'
```
domcleal: `'>= 2.2.2', '< 3.0'` to let it upgrade to 2.3 and beyond.

app/controllers/api/v2/common_parameters_controller.rb
```diff
@@ -9,7 +9,7 @@ class CommonParametersController < V2::BaseController
       param_group :search_and_pagination, ::Api::V2::BaseController
 
       def index
-        @common_parameters = resource_scope_for_index(:permission => :view_globals)
+        @common_parameters = resource_scope_for_index(:permission => :view_params).where(:type => 'CommonParameter')
```
domcleal: Should be unnecessary, as resource_scope_for_index uses resource_scope (redefined below).

app/models/parameter.rb
```diff
@@ -10,6 +10,29 @@ class Parameter < ActiveRecord::Base
   validates :name, :presence => true, :no_whitespace => true
 
   scoped_search :on => :name, :complete_value => true
+  scoped_search :on => :type, :complete_value => true
+  scoped_search :on => :value, :complete_value => true
+
+  # children associations must be defined here, otherwise scoped search definitions won't find them
+  belongs_to :domain, :foreign_key => :reference_id, :inverse_of => :domain_parameters
```
domcleal, line 17: I think it would be best to keep these (and scoped_search definitions) in the subclasses, but call them on the `superclass`?

Mostly so if the subclasses are changed, anything relating to each parameter type is all in one place - or if somebody adds a subclass, this isn't missed. These kind of things get out of sync otherwise.

app/services/authorizer.rb
```diff
@@ -11,10 +11,11 @@ def initialize(user, options = {})
   end
 
   def can?(permission, subject = nil, cache = true)
+    return true if user.admin? # is probably separate fix, if authorized is used for single permission it does not work for admin
```
domcleal: Either remove this comment or move it to its own PR and ticket. It should have a unit test added to authorizer_test.rb.

test/integration/parameters_permissions_test.rb
```diff
@@ -0,0 +1,139 @@
...
+class ParametersPermissionsIntegrationTest < ActionDispatch::IntegrationTest
+  setup do
+    role = FactoryGirl.create(:role)
+    @filter = FactoryGirl.create(:filter,
+                                 :permissions => Permission.where(:name => ['view_params']),
+                                 :search => 'name ~ a* or domain_name ~ example*com')
+    domain_filter = FactoryGirl.create(:filter, :permissions => Permission.where(:name => ['edit_domains', 'view_domains']))
+
+    role.filters = [ @filter, domain_filter ]
+    @user = FactoryGirl.create(:user, :mail => 'a@b.com')
```
domcleal: Nitpick, `:with_mail` as the value's unimportant. You could probably also set `:roles => [role]`.

test/integration/parameters_permissions_test.rb
```diff
@@ -0,0 +1,139 @@
+require 'integration_test_helper'
+
+class ParametersPermissionsIntegrationTest < ActionDispatch::IntegrationTest
+  setup do
+    role = FactoryGirl.create(:role)
+    @filter = FactoryGirl.create(:filter,
```
domcleal, line 6: These filter factories create a role already, so specify `:role => role` to reduce their work. (Or let this one create a role, then append the domain_filter to it.)

db/migrate/20161006070258_migrate_common_parameter_permissions.rb
```diff
@@ -0,0 +1,29 @@
...
+    affected = Filter.includes(:permissions).where(:permissions => {:name => ['view_globals', 'edit_globals', 'create_globals', 'destroy_globals']})
+    affected.each do |filter|
+      new_names = filter.permissions.map { |old| old.name.sub(/globals/, 'params') }
+      new_permissions = all_new_permissions.select do |new|
+        new_names.include?(new.name)
+      end
+
+      fake_filter = FakeFilter.find(filter.id)
+      new_search = "type = CommonParameter"
+      new_search = filter.search + ' and ' + new_search if filter.search.present?
```
domcleal: `filter.search` should probably be wrapped in parentheses (inside the search string) in case it's an `or` expression or something, to keep precedence clear + correct

domcleal (changes requested):
I need to do some more testing, but the direction looks good. Please also rebase if you can - the parent commit's rather old, I'd like to ensure it's tested against a current version develop.

app/models/parameter.rb
ares: I agree it would make more sense. Unfortunately it wouldn't work as mentioned in comment above (should I rephrase it?). Scoped search [does not work](https://github.com/wvanbergen/scoped_search/issues/112) well with STI.

Getting this when I move both definitions to domain parameter and then trying to do `Domain.search_for('domain_name ~ example')`.

```
NoMethodError: undefined method `klass' for nil:NilClass
        from /home/ares/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/scoped_search-3.3.0/lib/scoped_search/definition.rb:62:in `klass'
        from /home/ares/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/scoped_search-3.3.0/lib/scoped_search/query_builder.rb:302:in `to_sql'
        from /home/ares/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/scoped_search-3.3.0/lib/scoped_search/query_builder.rb:219:in `sql_test'
        from /home/ares/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/scoped_search-3.3.0/lib/scoped_search/query_builder.rb:514:in `sql_test'
        from /home/ares/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/scoped_search-3.3.0/lib/scoped_search/query_builder.rb:471:in `to_single_field_sql'
        from /home/ares/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/scoped_search-3.3.0/lib/scoped_search/query_builder.rb:483:in `to_sql'
        from /home/ares/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/scoped_search-3.3.0/lib/scoped_search/query_builder.rb:54:in `build_find_params'
        from /home/ares/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/scoped_search-3.3.0/lib/scoped_search/query_builder.rb:22:in `build_query'
        from /home/ares/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/scoped_search-3.3.0/lib/scoped_search/definition.rb:267:in `block in register_named_scope!'
```

Or did I misinterpret your suggestion?

app/services/authorizer.rb
ares: sorry, somehow I missed this before opening PR, extracted to https://github.com/theforeman/foreman/pull/4000

test/integration/parameters_permissions_test.rb
ares: Thanks, I've applied the :with_mail trait, if I use `:roles` it fails on `can_assign?` check unless I configure `User.current` which I want to avoid.

app/controllers/api/v2/common_parameters_controller.rb
ares: good catch, I copied that from v1 controller where we don't use resource_scope

ares:
Thanks for the review @domcleal, I believe I have addressed all your comments. The PR is now based on #4000 so the tests should be green. Also it's rebased on the latest develop.

ares:
Foreman test fails on AboutIntegrationTest.test_0001_about page on one combination, so I guess that's unrelated and can be ignored.

ares:
Rebased after the authorizer PR was merged.

app/models/parameter.rb
domcleal: I think you misinterpreted - I meant to call `superclass.belongs_to` and `superclass.scoped_search` inside the subclasses. I'm not certain it's a good idea, but it might be preferable to keeping the child class associations in the parent class file.

ares: Alright, got it. I'll try to move the code and see how it looks and feels, I like the idea.

ares: Unfortunately, it does not work, if we define associations later, children don't seem to inherit them correctly in AR. If you want to experiment further, here's the gist with patch and error I experience https://gist.github.com/ares/ab2ceb70d7873ba962039831afe7007b

test/integration/parameters_permissions_test.rb
ares: I think this is addressed on lines below.

ares:
@domcleal please take another look, the moving of definitions does not seem to work. If you find another solution I'll be happy to try it. Otherwise I think it's no longer WoC.

app/models/parameter.rb
domcleal: Yes, I see your point, thanks for trying. It seems possible to re-open the Parameter class before declaring DomainParameter, e.g.

``` ruby
require_dependency 'parameter'
class Parameter
  # associations must be defined on superclass, otherwise scoped search definitions does not find them
  belongs_to :domain, :foreign_key => :reference_id, :inverse_of => :domain_parameters
  # scoped search must be defined on superclass, so the same terms can be used for searching in parent class
  scoped_search :in => :domain, :on => :name, :complete_value => true, :rename => 'domain_name'
end

class DomainParameter < Parameter
[..]
end
```

but the complexity's high and I don't think it's worth it (as plugins may not subclass Parameter correctly in any case).

domcleal (changes requested):
1. The two API controllers are checking for a `view_common_parameters` permission for non-admin users on show/delete type actions (index is fine). I think they need to override `controller_permission` - this probably existed prior to this PR, but I'd suggest fixing it now to match the new name.
2. On migrate, users with `*_params` automatically get access to all CommonParameters while previously this only allowed them access to other Parameter types. Should this be left and documented on upgrade, or should filters with `*_params` have searches added (`type != CommonParameter`)?

ares:
@domcleal I've rebased the PR and replaced the the `child.instance_eval` introduced in https://github.com/theforeman/foreman/commit/08763fc2d16a7ea0d51c34406a0dd0db29306765#diff-d1834e361ee1da20b397eb0b3a77dfd2R12 since now it's defined in superclass.

> The two API controllers are checking for a view_common_parameters permission for non-admin users on show/delete type actions (index is fine). I think they need to override controller_permission - this probably existed prior to this PR, but I'd suggest fixing it now to match the new name.

Are you sure? I got the same error only if I forgot the restart my Foreman after checking out the branch. This is driven by `access_permissions.rb` which is not automatically reloaded. After restart I can see correct permissions being used in common parameters controller. In parameters controller it helped me to find another issue, parameters were not authorized at all. I kept the fix it in separate commit for easier reviewing.

> On migrate, users with *_params automatically get access to all CommonParameters while previously this only allowed them access to other Parameter types. Should this be left and documented on upgrade, or should filters with *_params have searches added (type != CommonParameter)?

If you agree, I'd rather document this in release notes since from my experience, users expected to grant all params permissions by *_params filters. It's trivial change if you or anyone else has different opinion.

ares:
[test foreman]

ares:
[test foreman]

ares:
rebased to avoid jquery issues

app/controllers/api/base_controller.rb
```diff
@@ -47,6 +47,10 @@ def controller_permission
       controller_name
     end
 
+    def current_permission
```
domcleal: Could you move the existing impl in application_controller.rb into FindCommon or maybe even ApplicationShared, since they're the same for now? It should avoid adding it again here.

domcleal (changes requested):

domcleal:
> > The two API controllers are checking for a view_common_parameters permission for non-admin users on show/delete type actions (index is fine). I think they need to override controller_permission - this probably existed prior to this PR, but I'd suggest fixing it now to match the new name.
> 
> Are you sure? I got the same error only if I forgot the restart my Foreman after checking out the branch. This is driven by access_permissions.rb which is not automatically reloaded.

Thanks for the explanation, this is probably what happened.

> If you agree, I'd rather document this in release notes since from my experience, users expected to grant all params permissions by *_params filters. It's trivial change if you or anyone else has different opinion.

Sure.

ares:
I've moved the `current_permission` method to `ApplicationShared`. I also squash commits into one and rebased the PR on latest develop. @domcleal please take another look.

domcleal:
> Are you sure? I got the same error only if I forgot the restart my Foreman after checking out the branch. This is driven by access_permissions.rb which is not automatically reloaded.

I've re-tested this and it is still failing. The issue isn't the authorisation check for access to the routes, but the resource lookup inside the controller:

<pre>
2016-11-23T10:42:41 cd00b3a0 [app] [I] Started GET "/api/v2/common_parameters/40" for 127.0.0.1 at 2016-11-23 10:42:41 +0000
2016-11-23T10:42:41 cd00b3a0 [app] [I] Processing by Api::V2::CommonParametersController#show as JSON
2016-11-23T10:42:41 cd00b3a0 [app] [I]   Parameters: {"apiv"=>"v2", "id"=>"40", "common_parameter"=>{}}
2016-11-23T10:42:41 cd00b3a0 [sql] [D]   ActiveRecord::SessionStore::Session Load (0.2ms)  SELECT  "sessions".* FROM "sessions" WHERE "sessions"."session_id" = ?  ORDER BY "sessions"."id" ASC LIMIT 1  [["session_id", "cd00b3a00032f1eb5294642ab120564f"]]
2016-11-23T10:42:41 cd00b3a0 [sql] [D]   User Load (0.2ms)  SELECT  "users".* FROM "users" WHERE "users"."lower_login" = ? LIMIT 1  [["lower_login", "16739test"]]
2016-11-23T10:42:41 cd00b3a0 [sql] [D]   AuthSource Load (0.1ms)  SELECT  "auth_sources".* FROM "auth_sources" WHERE "auth_sources"."id" = ? LIMIT 1  [["id", 1]]
2016-11-23T10:42:41 cd00b3a0 [sql] [D]   CACHE (0.0ms)  SELECT  "users".* FROM "users" WHERE "users"."lower_login" = ? LIMIT 1  [["lower_login", "16739test"]]
2016-11-23T10:42:41 cd00b3a0 [sql] [D] Authenticated user 16739test against INTERNAL authentication source
2016-11-23T10:42:41 cd00b3a0 [sql] [D]   User Load (0.1ms)  SELECT  "users".* FROM "users" WHERE "users"."lower_login" = ? LIMIT 1  [["lower_login", "foreman_admin"]]
2016-11-23T10:42:41 cd00b3a0 [app] [D] Setting current user thread-local variable to foreman_admin
2016-11-23T10:42:41 cd00b3a0 [app] [D] Setting current user thread-local variable to nil
2016-11-23T10:42:41 cd00b3a0 [sql] [D] Post-login processing for 16739test
2016-11-23T10:42:41 cd00b3a0 [sql] [D]   CACHE (0.0ms)  SELECT  "users".* FROM "users" WHERE "users"."lower_login" = ? LIMIT 1  [["lower_login", "foreman_admin"]]
2016-11-23T10:42:41 cd00b3a0 [app] [D] Setting current user thread-local variable to foreman_admin
2016-11-23T10:42:41 cd00b3a0 [sql] [D]   SQL (4.2ms)  UPDATE "users" SET "last_login_on" = '2016-11-23 10:42:41.499985' WHERE "users"."id" = ?  [["id", 66]]
2016-11-23T10:42:41 cd00b3a0 [sql] [D]   Role Load (0.1ms)  SELECT  "roles".* FROM "roles" WHERE "roles"."builtin" = ? LIMIT 1  [["builtin", 2]]
2016-11-23T10:42:41 cd00b3a0 [sql] [D]   Role Exists (0.2ms)  SELECT  1 AS one FROM "roles" INNER JOIN "user_roles" ON "roles"."id" = "user_roles"."role_id" WHERE "user_roles"."owner_id" = ? AND "user_roles"."owner_type" = ? AND "roles"."id" = ? LIMIT 1  [["owner_id", 66], ["owner_type", "User"], ["id", 8]]
2016-11-23T10:42:41 cd00b3a0 [app] [D] Setting current user thread-local variable to nil
2016-11-23T10:42:41 cd00b3a0 [app] [D] Setting current user thread-local variable to 16739test
2016-11-23T10:42:41 cd00b3a0 [sql] [D]    (0.2ms)  SELECT auth_sources.id FROM "auth_sources" WHERE "auth_sources"."type" IN ('AuthSourceHidden')
2016-11-23T10:42:41 cd00b3a0 [sql] [D]   User Load (0.1ms)  SELECT  "users".* FROM "users" WHERE ("users"."auth_source_id" NOT IN (7)) AND "users"."lower_login" = ?  ORDER BY firstname LIMIT 1  [["lower_login", "16739test"]]
2016-11-23T10:42:41 cd00b3a0 [app] [I] Authorized user 16739test(16739 test)
2016-11-23T10:42:41 cd00b3a0 [app] [D] Setting current user thread-local variable to 16739test
2016-11-23T10:42:41 cd00b3a0 [sql] [D]   Usergroup Load (0.1ms)  SELECT "usergroups".* FROM "usergroups" INNER JOIN "cached_usergroup_members" ON "usergroups"."id" = "cached_usergroup_members"."usergroup_id" WHERE "cached_usergroup_members"."user_id" = ?  ORDER BY usergroups.name  [["user_id", 66]]
2016-11-23T10:42:41 cd00b3a0 [sql] [D]   Role Load (0.2ms)  SELECT DISTINCT "roles".* FROM "roles" INNER JOIN "cached_user_roles" ON "roles"."id" = "cached_user_roles"."role_id" WHERE "cached_user_roles"."user_id" = ?  [["user_id", 66]]
2016-11-23T10:42:41 cd00b3a0 [sql] [D]    (0.1ms)  SELECT permissions.name FROM "permissions" INNER JOIN "filterings" ON "permissions"."id" = "filterings"."permission_id" INNER JOIN "filters" ON "filterings"."filter_id" = "filters"."id" WHERE "filters"."role_id" = ?  ORDER BY filters.role_id, filters.id  [["role_id", 8]]
2016-11-23T10:42:41 cd00b3a0 [sql] [D]    (0.1ms)  SELECT permissions.name FROM "permissions" INNER JOIN "filterings" ON "permissions"."id" = "filterings"."permission_id" INNER JOIN "filters" ON "filterings"."filter_id" = "filters"."id" WHERE "filters"."role_id" = ?  ORDER BY filters.role_id, filters.id  [["role_id", 46]]
2016-11-23T10:42:41 cd00b3a0 [permissions] [D] checking permission view_common_parameters
2016-11-23T10:42:41 cd00b3a0 [sql] [D]   Filter Load (0.3ms)  SELECT "filters".* FROM "filters" INNER JOIN "filterings" ON "filterings"."filter_id" = "filters"."id" INNER JOIN "permissions" ON "permissions"."id" = "filterings"."permission_id" INNER JOIN "roles" ON "filters"."role_id" = "roles"."id" INNER JOIN "cached_user_roles" ON "roles"."id" = "cached_user_roles"."role_id" WHERE "cached_user_roles"."user_id" = ? AND (permissions.resource_type = 'Parameter') AND (permissions.name = 'view_common_parameters')  [["user_id", 66]]
2016-11-23T10:42:41 cd00b3a0 [permissions] [D] 
2016-11-23T10:42:41 cd00b3a0 [permissions] [D] no filters found for given permission
2016-11-23T10:42:41 cd00b3a0 [sql] [D]    (0.2ms)  SELECT COUNT(*) FROM "parameters" WHERE (1=0) AND "parameters"."type" = ?  [["type", "CommonParameter"]]
2016-11-23T10:42:41 cd00b3a0 [app] [I] ActiveRecord::RecordNotFound (ActiveRecord::RecordNotFound)
2016-11-23T10:42:41 cd00b3a0 [app] [I]   Rendered api/v2/errors/not_found.json.rabl within api/v2/layouts/error_layout (3.6ms)
2016-11-23T10:42:41 cd00b3a0 [app] [I] Completed 404 Not Found in 49ms (Views: 9.8ms | ActiveRecord: 6.6ms)
2016-11-23T10:42:41 cd00b3a0 [sql] [D]    (0.1ms)  begin transaction
2016-11-23T10:42:41 cd00b3a0 [sql] [D]    (0.1ms)  commit transaction
</pre>

ares:
Aha, I reproduced and fixed it in both v1 and v2, please take one more look.

domcleal (approved):

domcleal:
Merged, thanks @ares.

PR #3987, opened 2016-11-02
Fixes #16987 - Added ability to apply facet attributes.

ShimShtein:
Now each facet will be requested to extract its information from the hostgroup before the host is created.

### DIFF
--- a/app/models/concerns/facets/base.rb
+++ b/app/models/concerns/facets/base.rb
@@ -38,13 +38,13 @@ module Facets
 
     module ClassMethods
       # Change attributes that will be sent to an facet based on inherited values from the hostgroup.
-      def self.inherited_attributes(hostgroup, facet_attributes)
+      def inherited_attributes(hostgroup, facet_attributes)
         facet_attributes
       end
 
       # Use this method to populate host's fields based on fact values exposed by the importer.
       # You can populate fields in the associated host's facets too.
-      def self.populate_fields_from_facts(host, importer, type, proxy_id)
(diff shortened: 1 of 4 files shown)


lzap:
Hey, thanks. Are you planning to do the Katello counterpart? If not, I will go ahead and do it, but I think we should test this together.

ShimShtein:
I see that Katello is [not using facets](https://github.com/Katello/katello/blob/master/app/models/katello/host/content_facet.rb#L3) framework at all...
@jlsherrill, Is there a reason for that?

ShimShtein:
Fixed a small bug in facet definition.

dLobatog:
Thanks @ShimShtein !

amitkarsale:
@ShimShtein the creation or updation of Host fails as there is no ` include Facets::Base` in `katello/app/models/katello/host/subscription_facet.rb`  and `katello/app/models/katello/host/content_facet.rb` files. 

The tests written for hosts are not w.r.t facets hence they pass if ran individually. 

If you found my point valid I can add the respective statement and commit.

cc: @dLobatog , @lzap

dLobatog:
Those 2 references are in https://github.com/Katello/katello/pull/6474

amitkarsale:
Alright. Thanks

lzap:
Is there anything I need to do in discovery to take advantage of this @ShimShtein ?

ShimShtein:
Just make sure that you are using `#apply_inherited_attributes` when you want the inheritance mechanism to kick in. We are already handling this in https://github.com/theforeman/foreman_discovery/pull/316

PR #3988, opened 2016-11-02
fixes #17187 - IPv6 address not required when CR provides IPv4

timogoebel:
This is a small fix for a reported bug, that we should 🚢 for 1.13.2. This basically breaks AWS.
#3977 contains refactorings to improve the `require_ip_validation` madness.


domcleal:
Merged! Thanks for extracting this fix @timogoebel.

PR #3989, opened 2016-11-03
Fixes #17200 - Add option datacenter to compute resource

ranjan:
```
hammer compute-resource create --provider="VMWare" --name "vCenter" --description "Test vCenter" --user "Administrator" --url "vmware.example.com" --password ComplexPassword
Could not create the compute resource:
Datacenter can't be blank
```

```
hammer compute-resource create --provider="VMWare" --name "vCenter" --description "Test vCenter" --user "Administrator" --url "vmware.example.com" --password ComplexPassword --datacenter Datacenter
Could not create the compute resource:
Error: Unrecognised option '--datacenter'
```

### DIFF
--- a/app/controllers/api/v2/compute_resources_controller.rb
+++ b/app/controllers/api/v2/compute_resources_controller.rb
@@ -34,7 +34,8 @@ module Api
           param :description, String
           param :user, String, :desc => N_("Username for oVirt, EC2, VMware, OpenStack. Access Key for EC2.")
           param :password, String, :desc => N_("Password for oVirt, EC2, VMware, OpenStack. Secret key for EC2")
-          param :uuid, String, :desc => N_("for oVirt, VMware Datacenter")
+          param :uuid, String, :desc => N_("Deprecated, please use datacenter") # FIXME: deprecated
+          param :datacenter, String, :desc => N_("for oVirt, VMware Datacenter")
           param :region, String, :desc => N_("for EC2 only")
           param :tenant, String, :desc => N_("for OpenStack only")
           param :server, String, :desc => N_("for VMware")


dLobatog:
@ranjan This looks fine, thank you! I don't understand why you removed the `--uuid` option though, maybe just a mistake? The API is supposed to be stable so adding new attributes is fine, just don't remove old ones so people's scripts still work. Would be glad to merge afterwards. 

Thank you! :clap:

ranjan:
Thanks for review @dLobatog  uuid is an alias for datacenter for [ovirt](https://github.com/theforeman/foreman/blob/develop/app/models/compute_resources/foreman/model/ovirt.rb#L11) and [vmware](https://github.com/theforeman/foreman/blob/develop/app/models/compute_resources/foreman/model/vmware.rb#L415) so all script should work without an issue. uuid is already removed from most of the compute resource api but it looks like some we missed out to remove this from post api. Shall i add attribute uuid back?

mbacovsky:
@ranjan, I'd prefer to mark the option deprecated first as removing it may break automation for users. While we have not :deprecated => true in apipie yet We are using a comment temporarily as in  https://github.com/theforeman/foreman/blob/698e916ce208b5040b83a908a058c83c94d158ee/app/controllers/api/v2/operatingsystems_controller.rb#L52

I think it would be worth to mention  also in the option description that it is deprecated as it is passed over to a Hammer command help.

ranjan:
@mbacovsky, PR updated

dLobatog:
[test katello] not sure what happened

app/controllers/api/v2/compute_resources_controller.rb
```diff
@@ -34,7 +34,8 @@ def show
           param :description, String
           param :user, String, :desc => N_("Username for oVirt, EC2, VMware, OpenStack. Access Key for EC2.")
           param :password, String, :desc => N_("Password for oVirt, EC2, VMware, OpenStack. Secret key for EC2")
-          param :uuid, String, :desc => N_("for oVirt, VMware Datacenter")
+          param :uuid, String, :desc => N_("Deprecated, , please use datacenter") # FIXME: deprecated
```
mbacovsky: Thanks for the change. Is there a reason for the two commas BTW?

ranjan: My mistake. updated again Thanks @mbacovsky

dLobatog:
@ranjan I've manually moved the ticket to the Foreman project - not hammer CLI where it was. Merging...

dLobatog:
Thanks @ranjan & @mbacovsky !

PR #3990, opened 2016-11-03
fixes #17197 - don't update DNS on ip6 change to empty string

domcleal:


timogoebel:
@domcleal : Thanks. Looks great. Have you investigated where the "nil" vs. "empty" values come from?

domcleal:
`nil` could be set if the host's created through the API or fact imports etc, it's the default value of most fields. The web UI form will always send an empty string for an empty field, so it changes value.

timogoebel:
Thanks, l'll merge when the tests have run.

timogoebel:
Merged, thanks for this quick fix @domcleal .

PR #3994, opened 2016-11-03
fixes #17170 - use strong params in InterfacesController

domcleal:


lzap:
Thanks, that works as expected.

supernovae:
If i edit an existing host, i see the options for bond show up after applying this patch, however i can't save the interfaces associated with the bond. I don't get any errors but they simply don't save or show up when i re-edit the screen after saving. I also noticed that during provisioning of a discovered host, clicking bond doesn't drop down the bonding options - is there a duplicated interface i need to edit / apply this fix to in order to fix bonds for discovered hosts?

lzap:
Let's move this to foreman-devel list or IRC, but in short - check production.log there should be an error if the object did not pass validation.

PR #3996, opened 2016-11-06
Fixes #17195 - CVE-2016-8634 escape html in alert text

tbrisker:
The alert helper used to mark the alert text as html_safe by default.
However, in some cases it may be possible for a user to enter custom
text into the alert message leading to a possible XSS vulnerability.
This patch changes this so that text is escaped unless expilicitly
marked as html_safe in the code.


domcleal:
Merged, thanks @tbrisker.

Could you please inform plugin developers on foreman-dev that if they pass HTML into `alert` that they will need to a) call .html_safe, and most importantly, b) HTML escape any user input in such messages.

PR #3997, opened 2016-11-06
Fixes #17238 - Validates lookup value match exists in key path

orrabin:


app/models/lookup_value.rb
```diff
@@ -98,6 +98,18 @@ def ensure_matcher(match_type)
     matcher
   end
 
+  def ensure_matcher_exists
+    return false if match.blank?
+    match.split(LookupKey::KEY_DELM).each do |m|
+      key_element = m.split(LookupKey::EQ_DELM).first
+      unless lookup_key.path.include?(key_element)
+        errors.add(:match, _("%{key} does not exist in order field") % { :key => key_element })
+        return false
```
domcleal: No need to return here or at the end, adding the error's enough and lets the user see _all_ errors.

test/integration/hostgroup_test.rb
```diff
@@ -38,7 +38,7 @@ class HostgroupIntegrationTest < ActionDispatch::IntegrationTest
 
   test 'clones lookup values' do
     group = FactoryGirl.create(:hostgroup, :with_puppetclass)
-    lookup_key = FactoryGirl.create(:puppetclass_lookup_key, :as_smart_class_param, :with_override,
+    lookup_key = FactoryGirl.create(:puppetclass_lookup_key, :as_smart_class_param, :with_override, :path => "hostgroup/ncomment",
```
domcleal: `\n`

test/models/hostgroup_test.rb
```diff
@@ -408,7 +408,7 @@ class HostgroupTest < ActiveSupport::TestCase
 
     test 'without save makes no changes' do
       group = FactoryGirl.create(:hostgroup, :with_config_group, :with_puppetclass)
-      FactoryGirl.create(:puppetclass_lookup_key, :as_smart_class_param, :with_override, :puppetclass => group.puppetclasses.first, :overrides => {group.lookup_value_matcher => 'test'})
+      FactoryGirl.create(:puppetclass_lookup_key, :as_smart_class_param, :with_override, :path => "hostgroup/ncomment", :puppetclass => group.puppetclasses.first, :overrides => {group.lookup_value_matcher => 'test'})
```
domcleal: `\n`

test/models/lookup_value_test.rb
```diff
@@ -121,6 +121,9 @@ def valid_attrs2
     env = FactoryGirl.create(:environment)
     pc = FactoryGirl.create(:puppetclass, :with_parameters, :environments => [env])
     key = pc.class_params.first
+    key.path += '\nfoo'
```
domcleal: 1. `\nfoo` with single quotes wouldn't actually create a new line, is this really doing anything?
2. Would it be better to change the `foo` in the lookup_value below to `comment` or whatever's in the key path?

test/models/lookup_value_test.rb
```diff
@@ -227,7 +237,7 @@ def setup
   context "when key type is puppetclass lookup and value is empty" do
     def setup
       @key = FactoryGirl.create(:puppetclass_lookup_key, :as_smart_class_param,
-                                :with_override, :with_omit,
+                                :with_override, :with_omit, :path => 'hostgroup\ncomment',
```
domcleal: ditto, \n in a single quoted string?

test/models/lookup_value_test.rb
```diff
@@ -248,7 +258,7 @@ def setup
 
   test "should allow white space in value" do
     key = FactoryGirl.create(:puppetclass_lookup_key, :as_smart_class_param,
-                             :with_override,
+                             :with_override, :path => 'hostgroup\ncomment',
```
domcleal: ditto

app/models/lookup_value.rb
```diff
@@ -98,6 +98,18 @@ def ensure_matcher(match_type)
     matcher
   end
 
+  def ensure_matcher_exists
+    return false if match.blank?
+    match.split(LookupKey::KEY_DELM).each do |m|
+      key_element = m.split(LookupKey::EQ_DELM).first
+      unless lookup_key.path.include?(key_element)
```
domcleal: I think `path` needs to be split here to more precisely check whether the attribute is in the _array_ of path items, not just the path _string_ as a whole. The path array should be accessible through `path_elements`.

I think this is why some tests that used single quotes for `path` are passing when they shouldn't (see comments further down).

domcleal (changes requested):
I think the tests need re-checking after fixing `path.include?` to check the array of path elements instead of the path string, there will probably be more failures.

app/models/lookup_value.rb
orrabin: You're right. that's better. The fact that I missed a few or the single quoted strings proves this is still permitting unknown matchers, I'll change it.

test/models/lookup_value_test.rb
orrabin: 1. I'll change it, it doesn't really matter for this test.

orrabin:
@domcleal thanks for the review. `ensure_matcher_exists` now handles multiple matchers correctly and the tests are also fixed.

domcleal (approved):

domcleal:
Merged, thanks @orrabin.

PR #3351, opened 2016-03-20
Fixes #5176 - Adding vSphere as an OS type

LiamLeane:
Adds simple vCenter OS family.

Templates heading to theforeman/community-templates#249

PR #3998, opened 2016-11-06
Fixes #5176 - Adding vSphere as an OS type

LiamLeane:
This was originally theforeman/foreman#3351, github isn't opening the pull back up after clearing commits and submitting a new one so new PR.

Adds vSphere as an OS type for VCSA, ESXi etc.


timogoebel:
@theforeman-bot ok to test

timogoebel:
@LiamLeane : First of all, thanks for this PR. I think, every new os is a great addition to Foreman.

I didn't have time to take a closer look, but you definitely need to fix the test in `test/models/operatingsystem_test.rb:162`.

dLobatog:
Seems OK, thanks @LiamLeane ! What templates are you using to deploy this? The ones in the Foreman wiki? http://projects.theforeman.org/projects/foreman/wiki/VMware_ESXi ? 

If so - would you mind to share the ones you're using? I'd like to deploy ESXi using this and merge. Thanks!

timogoebel:
@dLobatog : There is a PR that adds the templates: https://github.com/theforeman/community-templates/pull/249

app/models/operatingsystems/vsphere.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  class << self
+    delegate :model_name, :to => :superclass
+  end
+
+  def mediumpath(host)
+    medium_uri(host).to_s
+  end
+
+  def pxe_type
```
dLobatog, line 12: This needs also 

``` ruby
def pxedir
  "EFI/BOOT"
end
```

Not sure 100% about whether that path suits all versions but for 6.0 that seems OK

app/models/operatingsystems/vsphere.rb
```diff
@@ -0,0 +1,23 @@
...
+    "kickstart"
+  end
+
```
dLobatog, line 15: This needs an `available_loaders` method too to list all of the possible PXE loaders

dLobatog (changes requested):

PR #3999, opened 2016-11-06
fixes #17076 - Duplicate "Smart Class Parameter" entries.

kgaikwad:
Previously, there is no association between puppet class with smart class parameters i.e puppet class lookup keys directly, it is through environment class.
When someone tries to delete puppet class then dependent environment class objects also get deleted, but the smart class parameters remains in database.

Likewise association between puppet class and smart variables i.e variable lookup keys, there should be the association required between the puppet class and smart class parameters.


kgaikwad:
After the discussion with @orrabin, found that it is not a bug. It is the expected behavior as system keep record of the history of each smart parameter.
So closing the pull request.

PR #4001, opened 2016-11-07
Fixes #16546 - update facts on existing hosts

jlsherrill:
when create_new_host_when_facts_are_uploaded is set to false

### DIFF
--- a/app/models/host/base.rb
+++ b/app/models/host/base.rb
@@ -112,7 +112,7 @@ module Host
 
     # expect a facts hash
     def import_facts(facts)
-      return false unless create_new_host_when_facts_are_uploaded?
+      return false if !create_new_host_when_facts_are_uploaded? && new_record?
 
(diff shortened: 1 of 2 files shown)


lzap:
Thank you this makes sense.

app/models/host/base.rb
```diff
@@ -112,7 +112,7 @@ def create_new_host_when_facts_are_uploaded?
 
     # expect a facts hash
     def import_facts(facts)
-      return false unless create_new_host_when_facts_are_uploaded?
+      return false if !create_new_host_when_facts_are_uploaded? && new_record?
```
ohadlevy, line 115: nitpick: new_record? is probably better to check first, as it doesn't need to query another object (such as settings).

lzap: Yeah, settings are cached tho.

PR #4002, opened 2016-11-07
Fixes #17256 - ignore order in Authorizer

ares:


ares:
I believe this method is already well tested, I don't think it makes sense to add another as this is specific to what is passed as resource. In this case it's `Role.givable.for_current_user`

dLobatog:
Thanks @ares !

PR #4003, opened 2016-11-07
Fixes #17257 - Allow specific configs to be recreated

treydock:


lzap:
Hello, thanks for the patch. This is indeed very useful feature to have.

app/controllers/api/v2/hosts_controller.rb
```diff
@@ -265,8 +265,9 @@ def facts
 
       api :PUT, "/hosts/:id/rebuild_config", N_("Rebuild orchestration config")
       param :id, :identifier_dottable, :required => true
+      param :only, Array, :desc => N_("Limit rebuild steps")
```
lzap: Can you give examples of values which are expected here? Users won't know if you expect IDs, names or something else. I think you should be able to call `rebuild_methods` directly and it will generate fine (this needs to be tested).

app/models/host/managed.rb
```diff
@@ -880,16 +880,23 @@ def build_status_label(options = {})
...
   # returns  : Hash with 'true' if rebuild was a success for a given key (Example: {"TFTP" => true, "DNS" => false})
-  def recreate_config
+  def recreate_config(only = nil)
     result = {}
-    Nic::Managed.rebuild_methods.map do |method, pretty_name|
+    nic_rebuild_methods = Nic::Managed.rebuild_methods
+    host_rebuild_methods = self.class.rebuild_methods
+    if only.present?
+      nic_rebuild_methods = nic_rebuild_methods.select { |k,v| only.include?(k) || only.include?(v) }
+      host_rebuild_methods = host_rebuild_methods.select { |k,v| only.include?(k) || only.include?(v) }
```
lzap: Can you move the `select` statement into class method of Orchestration object named `rebuild_methods_for` or similar? That's the place it should be defined, will save one extra line as well.

app/models/host/managed.rb
```diff
@@ -880,16 +880,23 @@ def build_status_label(options = {})
   # rebuilds orchestration configuration for a host
   # takes all the methods from Orchestration modules that are registered for configuration rebuild
   # returns  : Hash with 'true' if rebuild was a success for a given key (Example: {"TFTP" => true, "DNS" => false})
-  def recreate_config
+  def recreate_config(only = nil)
     result = {}
-    Nic::Managed.rebuild_methods.map do |method, pretty_name|
+    nic_rebuild_methods = Nic::Managed.rebuild_methods
+    host_rebuild_methods = self.class.rebuild_methods
```
lzap: Irrelevant inputs should be reported as error, add a test for that as well please.

app/models/host/managed.rb
```diff
@@ -880,16 +880,23 @@ def build_status_label(options = {})
   # rebuilds orchestration configuration for a host
   # takes all the methods from Orchestration modules that are registered for configuration rebuild
   # returns  : Hash with 'true' if rebuild was a success for a given key (Example: {"TFTP" => true, "DNS" => false})
-  def recreate_config
+  def recreate_config(only = nil)
```
lzap, line 898: Please update the documentation comment as well.

test/models/host_test.rb
```diff
@@ -3122,6 +3122,27 @@ def to_managed!
...
+      Host::Managed.expects(:rebuild_methods).returns(:rebuild_test => "TEST")
+      host = FactoryGirl.build(:host, :interfaces => [@nic])
+      host.expects(:rebuild_test).never
+      result = host.recreate_config(['TFTP'])
+      refute result["DHCP"]
+      refute result["DNS"]
+      assert result["TFTP"]
+      refute result["TEST"]
+    end
+  end
```
lzap, line 3305: Can you also add edge case tests? When only is empty array and when it's filled with all possible values (you can dynamically generate them).

lzap (changes requested):
Overall, great patch, there is still some room for improvement. Thanks!

app/models/host/managed.rb
treydock: @lzap Should the error checking be done in Host::Managed method or should this be done in API where the inputs are supplied?

dLobatog:
@lzap Does this look good now? It doesn't dynamically test all combinations for the array of configurations to test but there's definitely good coverage in my opinion. I'd merge but I prefer to wait since you opened the review.

lzap:
I apologize for the delay, when there is no comment after push, I tend to miss the update. Need to start checking for labels. Let me re-review.

app/models/concerns/orchestration.rb
```diff
@@ -22,6 +22,14 @@ def rebuild_methods=(methods)
       @rebuild_methods = methods || {}
     end
 
+    def rebuild_methods_for(only = nil)
+      if only.present?
+        (@rebuild_methods || {}).select { |k,v| only.include?(k) || only.include?(v) }
+      else
+        @rebuild_methods || {}
+      end
+    end
```
lzap, line 31: Why are you comparing both keys and values? To me only keys look better, they are more readable. Here is how accepted values look like at the moment:

```
[6] pry(main)> (Nic::Managed.rebuild_methods.values + Host::Managed.rebuild_methods.values + Nic::Managed.rebuild_methods.keys + Host::Managed.rebuild_methods.keys).join(', ')
=> "DHCP, DNS, TFTP, rebuild_dhcp, rebuild_dns, rebuild_tftp"
```

How about only accept "DHCP" "DNS" and "TFTP"?

app/controllers/api/v2/hosts_controller.rb
```diff
@@ -265,8 +265,12 @@ def facts
 
       api :PUT, "/hosts/:id/rebuild_config", N_("Rebuild orchestration config")
       param :id, :identifier_dottable, :required => true
+      param :only, Array, :desc => N_(<<-eos
+Limit rebuild steps, valid steps are #{(Nic::Managed.rebuild_methods.values + Host::Managed.rebuild_methods.values + Nic::Managed.rebuild_methods.keys + Host::Managed.rebuild_methods.keys).join(', ')}
+eos
+)
```
lzap: Please extract the list to a constant, freeze it and use it in the underscore function according to http://projects.theforeman.org/projects/foreman/wiki/Translating#Extracting-strings (using per-cent operator).

dLobatog:
@treydock I've this to Waiting on contributor after @lzap 's comments

app/controllers/api/v2/hosts_controller.rb
treydock: @lzap I tried a constant in Host::Managed but kept getting undefined method errors on rebuild_methods.  Using a class method seems to work

treydock:
@dLobatog @lzap The PR has been updated

app/models/concerns/orchestration.rb
```diff
@@ -22,6 +22,14 @@ def rebuild_methods=(methods)
       @rebuild_methods = methods || {}
     end
 
+    def rebuild_methods_for(only = nil)
+      if only.present?
+        (@rebuild_methods || {}).select { |k,v| only.include?(v) }
+      else
+        @rebuild_methods || {}
+      end
+    end
```
lzap, line 31: Hey sorry for the delay, tests are red please rebase and this can be rewritten as:

    def rebuild_methods_for(only = valid_rebuild_only_values)
      (@rebuild_methods || {}).select { |k,v| only.include?(v) }
    end

Or similar (not sure if the cop will like that).

lzap:
Other than that looks good, thanks!

lzap (approved):

lzap:
But this is not mandatory, your code is fine. Let me retrigger tests: [test foreman]

app/models/concerns/orchestration.rb
treydock: @lzap Setting only as defaulting to valid_rebuild_only_values breaks the tests that verified passing empty array would return all rebuild methods.  The use of `only.present?` was used to only filter rebuild methods when there was actual methods listed, so `nil` or `[]` would return all methods.  Should I modify this behavior so that an empty array returns no methods?

treydock:
@lzap I rebased and modified orchestration test changes to use Host instead of Nic::Managed, so that the tests would more accurately reflect which class is using the rebuild_methods_for method.

lzap:
Thanks.

PR #4005, opened 2016-11-08
Fixes #17248 - Run js tests with node versions we use

tbrisker:
Our package builders and test infrastracture use different versions of
nodejs. We should check that all js tests pass for all different
versions used, to prevent breakage in any of our supported OSes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,9 @@
 language: node_js
 node_js:
-  - "node"
-  - "4" # Fedora
+  #- "0.10" # EL6, Trusty, Jessie - sadly, jest does not work with 0.10
+  - "4.2" # Xenial
+  - "4.6" # Fedora 24
+  - "6.9" # EL 7
 before_install:
   npm install -g npm@'>=3.0.0'
 script: ./script/travis_run_js_tests.sh
--- a/script/travis_run_js_tests.sh
+++ b/script/travis_run_js_tests.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
 set -ev
-if [[ $( git diff --name-only HEAD~1..HEAD webpack/ | wc -l ) -ne 0 ]]; then
+if [[ $( git diff --name-only HEAD~1..HEAD webpack/ .travis.yml package.json | wc -l ) -ne 0 ]]; then
   npm run test;
 fi


script/travis_run_js_tests.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
 set -ev
-if [[ $( git diff --name-only HEAD~1..HEAD webpack/ | wc -l ) -ne 0 ]]; then
+if [[ $( git diff --name-only HEAD~1..HEAD webpack/ .travis.yml | wc -l ) -ne 0 ]]; then
```
tbrisker: we should also re-run tests if we change the travis config

.travis.yml
```diff
@@ -1,7 +1,9 @@
 language: node_js
 node_js:
-  - "node"
-  - "4" # Fedora
+  - "0.10" # EL6, Trusty, Jessie
```
ohadlevy: sadly tests are failing on 0.10 version.

script/travis_run_js_tests.sh
ohadlevy: :+1:

.travis.yml
gailsteiger: What tests are failing? What's the error message?

ohadlevy: @gailsteiger see https://travis-ci.org/theforeman/foreman/builds/174192667 (also linked bellow) but it does open the question if its the tests who fail because of old node versions or the actual code (e.g. if webpack:compile passes).

further, I don't know if its an issue to upgrade node/npm just when building the packages for other oss (e.g. non rhel7), /cc @mmoll

ohadlevy:
@tbrisker how about just removing latest node version? this would unblock all open JS pr's that are failing tests?

tbrisker:
@ohadlevy commented out 0.10, seems that jest can't run on it. not sure what we can do about that. Also added changes to package.json as something that triggers js tests.

ohadlevy:
test failure seems unrelated?

ohadlevy:
merged as 5cb5f8, thanks @tbrisker

.travis.yml
mmoll: For Debian it would be possible to use nodejs 4 from the nodesource repo for Debian/jessie and Ubuntu/trusty, but I'd like to avoid it.

PR #4006, opened 2016-11-08
Fixes #15498 - removes scopes from includes in taxonomix

ares:


dLobatog:
Thanks @ares , easy to reproduce and I used to find this bug too. Merging...

PR #4007, opened 2016-11-08
fixes #14178 - introduce patternfly toast notifications

ohadlevy:
this PR replaces the jnotify implementation.

TODO:
- add missing tests.


webpack/assets/javascripts/react_app/components/toasts/Toast.js
```diff
@@ -0,0 +1,43 @@
...
+  }
+  icon() {
+    return 'pficon pficon-' + (ToastIcons[this.type].icon || ToastIcons.default.icon);
+  }
+  closeBtn() {
+    return (
+      <Button className="close" data-dismiss="alert" aria-hidden="true">
+        <span className="pficon pficon-close"></span>
+      </Button>
+    );
```
gailsteiger: Why not use html button tag here as in patternfly reference markup. What do we gain from using the react-bootstrap component?

ohadlevy: most likely in this usage case its not that important.. i can go either way.

ohadlevy: but i dont think it implies we don't need react-bootstrap?

gailsteiger: It doesn't imply that we don't need react-bootstrap, just that we don't need it here.

ohadlevy:
now using plain buttons instead of a react-bootstrap component.

ohadlevy:
I've just noticed https://github.com/bentatum/react-fade, maybe we can remove the Expire component and replace it with <Fade>

ohadlevy:
@gailsteiger @tbrisker I've rebased this and updated, can you have a look please and comment on the approach? thanks.

webpack/assets/javascripts/foreman_notifications.js
```diff
@@ -0,0 +1,31 @@
+import ToastActions from './react_app/actions/ToastNotificationActions';
+import $ from 'jquery';
+
+export function notify(notification) {
+  ToastActions.addNotification(notification);
+}
+
+export function importFlashMessagesFromRails() {
```
tbrisker: Does this actually need exporting? if it is only used internally in this module, don't export it.

webpack/assets/javascripts/foreman_notifications.js
```diff
@@ -0,0 +1,31 @@
...
+
+  notifications.forEach(notification => {
+    let [type, message] = notification;
+
+    // normalize rails flash names
+    if (type === 'danger') {
+      type = 'error';
+    }
+
+    notify({type: type, message: message, sticky: (type === 'error')});
```
tbrisker: I think `notify({type, message, sticky: (type === 'error')});` will also work

app/assets/javascripts/application.js
```diff
@@ -378,15 +358,9 @@ function spinner_placeholder(text){
...
-  var icon = typeToIcon(type);
-  var options = { classMessage: "foreman-alert",
-                  classBackground: "",
-                  sticky: (type != 'success'),
-                  type: type };
-  $.jnotify("</div>" + icon + $(item).text(), options);
-  // jnotify does not support multiple classes passed via classMessage so added via jquery.
-  $('.foreman-alert').addClass("alert alert-" + type);
-  $(item).remove();
+  console.log("DEPRECATION WARNING: you are using deprecated notify() function, \
```
tbrisker: `console.warn` maybe?

tbrisker: Also, how about making this into `tfm.tools.deprecate('notify','1.17')`?

app/assets/javascripts/dashboard.js
```diff
@@ -53,12 +53,12 @@ function remove_widget(item){
             type: 'DELETE',
             url: $(item).data('url'),
             success: function(){
-                $.jnotify(__("Widget removed from dashboard."), 'success', false);
+                notify(__("Widget removed from dashboard."), 'success');
```
tbrisker: why not use  `tfm.notifications.notify` directly?

webpack/assets/javascripts/bundle.js
```diff
@@ -17,6 +17,7 @@ window.tfm = Object.assign(
     users: require('./foreman_users'),
     trends: require('./foreman_trends'),
     hostgroups: require('./foreman_hostgroups'),
+    notifications: require('./foreman_notifications'),
```
tbrisker: this might cause confusion with the notification drawer, maybe name it toast instead?

webpack/assets/javascripts/foreman_notifications.js
```diff
@@ -0,0 +1,31 @@
...
+import $ from 'jquery';
+
+export function notify(notification) {
+  ToastActions.addNotification(notification);
+}
+
+export function importFlashMessagesFromRails() {
+  const notifications = $('#notifications').data().flash;
+
+  notifications.forEach(notification => {
```
tbrisker: you can use destructuring directly in the function definition, like so:
`notifications.forEach(([type, message]) => {...`

webpack/assets/javascripts/foreman_notifications.js
tbrisker: also, previously warning was also sticky (the condition checked for `type!='success'`)

webpack/assets/javascripts/react_app/components/notifications/toast/Toast.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import Icon from '../../common/Icon';
+import Alert from '../../common/Alert';
+
+const Toast = (props) => {
```
tbrisker, line 5: Not crucial but how about destructuring the props here so that it's clear which props we expect?

webpack/assets/javascripts/react_app/common/MountingService.js
```diff
@@ -23,7 +25,11 @@ export function mount(component, selector, data) {
     NotificationDrawer: {
       type: NotificationDrawer,
       markup: <NotificationDrawer data={data} />
-    }
+    },
+    FlashNotifications: {
```
tbrisker: Shouldn't this also be called ToastNotifications?

webpack/assets/javascripts/react_app/components/common/Fade.js
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import helpers from '../../common/helpers';
+
+export default class Fade extends React.Component {
```
tbrisker, line 4: Correct me if i'm wrong but this doesn't actually fade, just removes the content when time is out?

webpack/assets/javascripts/react_app/stores/ToastNotificationsStore.js
```diff
@@ -0,0 +1,55 @@
...
+      } else {
+        action.ids.forEach(id => {
+          _notifications.data.slice(id);
+        });
+      }
+
+      ToastNotificationsStore.emitChange(action.actionType);
+      break;
+    }
+    case ACTIONS.FADED_NOTIFICATION: {
```
tbrisker: If this action isn't used remove it?

webpack/stories/index.js
```diff
@@ -11,6 +11,11 @@ import chartService from '../assets/javascripts/services/statisticsChartService'
 import mockData from './data/charts/donutChartMockData';
 import PowerStatus from
   '../assets/javascripts/react_app/components/hosts/PowerStatus';
+// import FlashNotifications
+// from '../assets/javascripts/react_app/components/notifications/toast/FlashNotifications';
```
tbrisker: leftovers?

tbrisker (changes requested):
:1st_place_medal: for getting rid of one of one of our unmaintained dependencies, see inline comments.

webpack/assets/javascripts/react_app/components/notifications/toast/Toast.js
ohadlevy: I tried, but it was more than 4, that was both lint error and pretty long, there is probably a react way to describe the props - @gailsteiger ?

webpack/assets/javascripts/react_app/components/common/Fade.js
ohadlevy: Yeah, would Expire be a better name?

ohadlevy:
@tbrisker I've addressed your comments, I did not handle the deprecation as I wasn't sure whats the best way to call a function from string, I think it deserve its own PR too.

LICENSE
```diff
@@ -28,7 +28,6 @@ The following files and directories are exceptions:
 * vendor/assets/fonts/ is (c) 2011-2016, Red Hat Inc. under SIL Open Font License v1.1 or LGPL v2.1 licenses.
 * vendor/assets/javascripts/noVNC/ is (c) 1999 AT&T Laboratories Cambridge, (c) 1996 Widget Workshop, Inc., (c) 1996 Jef Poskanzer <jef@acme.com>, (c) 2011 Joel Martin, (c) Hiroshi Ichikawa <http://gimite.net/en/>, (c) Martijn Pieters <mj@digicool.com>, Samuel Sieb <samuel@sieb.net> under (MPL 1.1 or GPL v2+ or LGPL v2.1+) and LGPL v3 and LGPL v2+ and MIT and Revised BSD licenses.
 * vendor/assets/javascripts/jquery.flot.axislabels.js is (c) 2010 Xuan Luo under MIT license.
-* vendor/assets/javascripts/jquery.jnotify.js is (c) 2010 Giva, Inc. under ASL 2.0 license.
```
tbrisker, line 31: This was the only one using ASL, we can remove it below.

app/assets/javascripts/application.js
```diff
@@ -378,15 +358,9 @@ function spinner_placeholder(text){
 }
 
 function notify(item, type) {
```
tbrisker: you can change the parameter name to `message`, then below call
```js
tfm.toastNotifications.notify({message, type});
```

vendor/assets/stylesheets/vendor.scss
```diff
@@ -1 +0,0 @@
-@import "/**/*";
```
tbrisker, line 1: Why not remove the folder entirely?

webpack/assets/javascripts/react_app/components/common/Fade.js
tbrisker: Maybe, or maybe use the fade component you found that actually does fade?

webpack/assets/javascripts/react_app/components/notifications/toast/Toast.js
```diff
@@ -0,0 +1,28 @@
...
+  return (
+    <Alert
+      type={props.type}
+      dismissable={props.close}
+      css="toast-pf"
+      >
+      <div className="pull-right toast-pf-action">
+        <a href="#">
+          {props.link}
```
tbrisker, line 14: What happens if there is no link? Also, shouldn't the link also have a target that isn't `#`?

tbrisker (changes requested):
Some more comments inline, opened https://github.com/theforeman/foreman/pull/4246 for the deprecation message.

vendor/assets/stylesheets/vendor.scss
ohadlevy: We no longer use it, and I hope we would not need to as we moved assets to npm/yarn packages

webpack/assets/javascripts/react_app/components/common/Fade.js
ohadlevy: I did that initially, but it ended up pulling a lot of unneeded dependencies.

webpack/assets/javascripts/react_app/components/common/Alert.js
```diff
@@ -0,0 +1,31 @@
...
+    classNames += ' ' + props.css;
+  }
+
+  if (props.dismissable) {
+    classNames += ' ' + ALERT_CSS.dismissable;
+  }
+
+  const closeBtn = (
+    <button className="close" data-dismiss="alert" aria-hidden="true">
+      <Icon type="close" />
```
gailsteiger, line 18: We shouldn't use the bootstrap js. It manipulates the DOM. Need to be careful about inconsistencies between the DOM, the virtual DOM and the data.

It's possible to unmount with `ReactDOM.unmountComponentAtNode(ReactDOM.findDOMNode(this).parentNode);
`. Can't use stateless component in you use that though. Notice the `this`.

webpack/assets/javascripts/react_app/components/notifications/toast/Toast.js
gailsteiger: You can always break the statement into 2 or more lines. The question is whether or not destructuring the props in this case adds clarity or causes confusion.

webpack/assets/javascripts/react_app/components/common/Fade.js
gailsteiger: Is the content being removing via React or through DOM manipulation?

webpack/assets/javascripts/react_app/components/notifications/toast/Toast.js
gailsteiger: building the JSX such that the <a> is only include when there actually is a link. pointer cursor not desirable when there is nothing to click.

gailsteiger:
I think the terminology could be confusing between toast and notification. Worth considering

webpack/assets/javascripts/foreman_toast_notifications.js
```diff
@@ -0,0 +1,30 @@
...
+    if (type === 'danger') {
+      type = 'error';
+    }
+
+    notify({type, message, sticky: (type !== 'success')});
+  });
+}
+
+// clear all notifications when leaving the page
+$(window).bind('beforeunload', function () {
```
matanwerbner, line 23: Shouldn't this action be fired on the "componentWillUnmount" lifecycle event of the notification component?

gailsteiger: The implementation is WIP. It is not pure react and contains carryover aspects which directly manipulate the DOM. This will be cleaned up.

ohadlevy:
closed in favor of #4300

PR #4011, opened 2016-11-09
Fixes #17296 - Add setting access_unattended_without_build

treydock:


dLobatog:
@treydock Not sure if I'm missing something, but as far as I know those URLs (without token) are already possible to visit. e.g: `https://myforeman.com/unattended/finish?hostname=alvin-eader.lobatolan.home`. 

On the Host page, you can click on Review (click on the arrow next to Edit) and then you'd be able to see those templates.
screenshot from 2016-11-11 09-49-13 

If that's the feature you were looking for - I guess no setting is needed. If not, sorry for the misunderstanding but can you explain how it differs from this one?

Thanks!

timogoebel:
I could imagine some use-cases where you'd want to access the rendered user-data template from the host even when the host is not in build mode anymore.

dLobatog:
@timogoebel Mind to elaborate a bit? Remember you can also set the token duration to 0 so that they last forever.

timogoebel:
Sure, it's just an idea and this PR might not be the ideal place to discuss this: But I was thinking about something like this:
AWS has a metadata service. You can just curl `http://169.254.169.254/latest/user-data` to get the user-data of an instance. OpenStack has something similar. When a host boots, cloud-init goes to that URL to retrieve the cloud-init template and configures the VM accordingly.
Other compute resources don't have that possibility, but you could bake a custom url into the image (or set it via [guestinfo.coreos.config.url for CoreOS and VMWare](https://coreos.com/os/docs/latest/booting-on-vmware.html#vmware-guestinfo-interface)). That custom url could be a Foreman instance.
I see some problems here: Foreman doesn't ship the user-data template when the host isn't in build mode (this pr possibly fix this, that's why I commented the PR) and the host doesn't know it's token (might not be a problem).

Disclaimer: I haven't tried this at home(, yet 😈 ).

treydock:
@dLobatog The goal here is to allow iPXE to use Foreman PXE urls from Foreman without a system being in build mode.  Right now a system's iPXE isn't accessible unless a host is in build mode, so something like `curl http://foreman/unattended/iPXE` will fail without build being set.  We use NFS root systems that never need to be in build mode but their PXE comes from Foreman and we frequently have to update the templates with host parameters to change kernel version or root filesystem paths.  Currently we have to use custom rake tasks to sync the TFTP files to foreman-proxy.  We would like to be able to use iPXE to pull rendered templates without using build mode.

Does this clarify the purpose of this pull request?

timogoebel:
@dLobatog : Sounds reasonable to me. Shall we merge?

GregSutcliffe:
I've seen this usecase come up a few times with situations where you want to run some first-time setup after reboot, but can't because you have to mark built before rebooting to avoid loops - on a variety of platforms.

However, it *is* a security risk, as the templates can contain passwords or other useful data. Does it make sense to allow this new setting to *only* work with a provisioning token and not MAC/IP matching? That would prevent a malicious user from retrieving the template weeks after build using a known IP/MAC.

timogoebel:
I think, having a special setting for this is enough. If a user turns the setting on, he should know what he is doing. In the described uses cases, you don't know the security token at the time you request the information.

GregSutcliffe:
Mmm, good point. In that case, documentation is clearly critical - I'm +1 for this so long as a good explanation of the setting is added to the manual :)

dLobatog:
@treydock Thanks - as @GregSutcliffe mentioned, docs would be greatly appreciated, specially with use cases like the one you described (I was kind of lost as to why this was needed myself). I'll merge in 3..2..1

PR #4012, opened 2016-11-10
Fixes #6502 - Add a list of ignored classes in the Puppet classes view

bastilian:
This PR addresses the issue reported in [#6502](http://projects.theforeman.org/issues/6502) by making ignored environments and classes visible when importing like so: 
screen shot 2016-11-14 at 13 50 43

In case all available environments are ignored as full and no other changes are detected it will also mention these in the notice:
screen shot 2016-11-14 at 14 02 51

### DIFF
--- a/app/controllers/concerns/foreman/controller/environments.rb
+++ b/app/controllers/concerns/foreman/controller/environments.rb
@@ -9,6 +9,7 @@ module Foreman::Controller::Environments
       opts[:env] = params[:env] unless params[:env].blank?
       @importer = PuppetClassImporter.new(opts)
       @changed  = @importer.changes
+
     rescue => e
       if e.message =~ /puppet feature/i
         error _("No smart proxy was found to import environments from, ensure that at least one smart proxy is registered with the 'puppet' feature.")
@@ -18,10 +19,20 @@ module Foreman::Controller::Environments
       end
     end
 
+    if @importer.ignored_boolean_environment_names?
+      warning(_("Ignored environment names resulting in booleans found. Please quote strings like true/false and yes/no in config/ignored_environments.yml."))
+    end
+
     if @changed["new"].size > 0 || @changed["obsolete"].size > 0 || @changed["updated"].size > 0
       render "common/_puppetclasses_or_envs_changed"
     else
-      notice _("No changes to your environments detected")
+      notice_message = _("No changes to your environments detected")
+
+      if @changed['ignored'].present?
+        list_ignored(notice_message, @changed['ignored'])
+      end
+
+      notice(notice_message)
       redirect_to :controller => controller_path
     end
   end
@@ -43,5 +54,18 @@ module Foreman::Controller::Environments
     error _("Failed to add task to queue: %s") % e.to_s
   ensure
     redirect_to :controller => controller_path
+  end
+
+  private
+
+  def list_ignored(notice, ignored)
+    environments = ignored.select { |_, values| values.first == '_ignored_' }
+    if environments.any?
+      ignore_notice = _("Ignored environments: %s") % environments.keys.to_sentence
+    else
+      ignore_notice = _("Ignored classes in the environments: %s") % ignored.keys.to_sentence
+    end
+
+    notice << "\n" + ignore_notice
   end
 end
--- a/app/services/puppet_class_importer.rb
+++ b/app/services/puppet_class_importer.rb
@@ -17,32 +17,44 @@ class PuppetClassImporter
 
   # return changes hash, currently exists to keep compatibility with importer html
   def changes
-    changes = { 'new' => { }, 'obsolete' => { }, 'updated' => { } }
+    changes = { 'new' => { }, 'obsolete' => { }, 'updated' => { }, 'ignored' => { } }
 
     if @environment.nil?
       actual_environments.each do |env|
-        new     = new_classes_for(env)
-        old     = removed_classes_for(env)
-        updated = updated_classes_for(env)
-        changes['new'][env] = new if new.any?
-        changes['obsolete'][env] = old if old.any?
-        changes['updated'][env] = updated if updated.any?
+        changes_for_environment(env, changes)
       end
 
       old_environments.each do |env|
         changes['obsolete'][env] ||= []
         changes['obsolete'][env] << "_destroy_" unless actual_environments.include?(env)
       end
+
+      ignored_environments.each do |env|
+        changes['ignored'][env] ||= []
+        changes['ignored'][env] << '_ignored_'
+      end
     else
-      env = @environment
-      new     = new_classes_for(env)
-      old     = removed_classes_for(env)
-      updated = updated_classes_for(env)
-      changes['new'][env] = new if new.any?
-      changes['obsolete'][env] = old if old.any?
-      changes['updated'][env] = updated if updated.any?
+      changes_for_environment(@environment, changes)
     end
+
     changes
+  end
+
+  # Adds class changes of an environment to a changes hash in new, obsolete and updated
+  #
+  # Params:
+  #  * +environment+: {String} of environments name
+  #  * +changes+: {Hash} to add changes to
+  #
+  def changes_for_environment(environment, changes)
+    new_classes     = new_classes_for(environment)
+    old_classes     = removed_classes_for(environment)
+    updated_classes = updated_classes_for(environment)
+    ignored_classes = ignored_classes_for(environment)
+    changes['new'][environment] = new_classes if new_classes.any?
+    changes['obsolete'][environment] = old_classes if old_classes.any?
+    changes['updated'][environment] = updated_classes if updated_classes.any?
+    changes['ignored'][environment] = ignored_classes if ignored_classes.any?
   end
 
   # Update the environments and puppetclasses based upon the user's selection
@@ -71,6 +83,15 @@ class PuppetClassImporter
     #  [e.to_s]
   end
 
+  # Returns all classes for a given environment
+  #
+  # Params:
+  #  * +environment+: {String} containing the name of the environment
+  #
+  def proxy_classes_for(environment)
+    @proxy_classes[environment] ||= proxy.classes(environment)
+  end
+
   def new_classes_for(environment)
     old_classes = db_classes_name(environment)
     HashWithIndifferentAccess[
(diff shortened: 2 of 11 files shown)


test/unit/puppet_class_importer_test.rb
```diff
@@ -18,11 +18,67 @@ def setup
...
-    proxy = smart_proxies(:puppetmaster)
-    importer = PuppetClassImporter.new(:url => proxy.url, :env => 'foreman-testing')
-    assert importer.changes['new'].include?('foreman-testing')
-    assert !importer.changes['new'].include?('foreman-testing-1')
+  describe '#changes' do
+    setup do
+      @proxy = smart_proxies(:puppetmaster)
+    end
+
+    context 'a spcefific environment is set' do
```
domcleal: specific

test/unit/puppet_class_importer_test.rb
```diff
@@ -18,11 +18,67 @@ def setup
...
+      test "should contain only the specified environment in changes" do
+        importer = PuppetClassImporter.new(url: @proxy.url, env: 'foreman-testing')
+
+        assert importer.changes['new'].include?('foreman-testing')
```
domcleal: Prefer assert_includes, refute_includes (better debug output on assertion failure).

app/views/api/v1/import_puppetclasses/show.json.rabl
```diff
@@ -25,3 +28,7 @@ end
 node(:removed_environment, :if => ->(environment) { @changed['obsolete'][environment.name].present? && @changed['obsolete'][environment.name].match(/_destroy_/) }) do |environment|
   environment.name
 end
+
+node(:ignored_environment, :if => ->(environment) { @changed['ignored'][environment.name].present? && @changed['ignored'][environment.name].match(/_ignored_/) }) do |environment|
```
domcleal: These changes should be made to API v2 - either only v2, or both v2 and v1 if you really wish to continue extending it.

app/services/puppet_class_importer.rb
```diff
@@ -94,6 +107,29 @@ def updated_classes_for(environment)
...
+    end
+  end
+
+  # Returns true when the class name matches any pattern in ignored_classes
+  #
+  # Params:
+  #  * ++class_name++:: {String} containing the class to be checked
+  #
+  def ignored_class?(class_name)
+    !ignored_classes.select do |filter|
```
domcleal: Use `.any?` instead of `! .. select .. empty?`, it's faster and easier to read.

app/controllers/concerns/foreman/controller/environments.rb
```diff
@@ -21,7 +21,12 @@ def import_environments
     if @changed["new"].size > 0 || @changed["obsolete"].size > 0 || @changed["updated"].size > 0
       render "common/_puppetclasses_or_envs_changed"
     else
-      notice _("No changes to your environments detected")
+      notice_message = _("No changes to your environments detected")
+      if !@changed['ignored'].empty?
+        notice_message << "\n"
+        notice_message << _("Ignored Environments: %s") % @changed['ignored'].keys.join(', ')
```
domcleal: 1. Lower case "environments"
2. Use `.to_sentence` instead of `join(', ')`, it's locale-aware

app/controllers/concerns/foreman/controller/environments.rb
```diff
@@ -21,7 +21,12 @@ def import_environments
     if @changed["new"].size > 0 || @changed["obsolete"].size > 0 || @changed["updated"].size > 0
       render "common/_puppetclasses_or_envs_changed"
     else
-      notice _("No changes to your environments detected")
+      notice_message = _("No changes to your environments detected")
+      if !@changed['ignored'].empty?
```
domcleal: .present? or unless

test/controllers/api/v1/smart_proxies_controller_test.rb
```diff
@@ -223,4 +223,18 @@ def setup_import_classes
       assert_equal env_name, response['results']['name']
     end
   end
+
+  test 'should contain ignored environments' do
```
domcleal: ditto re API versions, should be in v2

app/services/puppet_class_importer.rb
```diff
@@ -94,6 +107,29 @@ def updated_classes_for(environment)
     ]
   end
 
+  # Gives back the classes ignored for a given environment
+  #
+  # Params:
+  #  * ++environment++:: {String} name of the environment
```
domcleal: Is this meant to be RDoc? (normally single `+` isn't it for tt?)

app/services/puppet_class_importer.rb
```diff
@@ -94,6 +107,29 @@ def updated_classes_for(environment)
...
+  #
+  def ignored_classes_for(environment)
+    classes = proxy.classes(environment)
```
domcleal: Can you use the existing query instead of making a second request to the smart proxy? The performance of this API in some configurations may be poor so it shouldn't be called unnecessarily.

It looks like the information is broadly available via `actual_classes`, but it will need a little refactoring to keep the ignored classes.

app/services/puppet_class_importer.rb
```diff
@@ -94,6 +107,29 @@ def updated_classes_for(environment)
...
+  end
+
+  # Returns true when the class name matches any pattern in ignored_classes
+  #
+  # Params:
+  #  * ++class_name++:: {String} containing the class to be checked
+  #
+  def ignored_class?(class_name)
+    !ignored_classes.select do |filter|
+      filter =~ class_name
```
domcleal: This duplicates the check in `actual_classes` but without the safety check for Regexp - it should be shared code.

domcleal (changes requested):

domcleal:
@theforeman-bot ok to test

bastilian:
@domcleal thanks for reviewing! I've incorporated the changes you suggested.

There is an issue with moving the `ignored_environments` to `api/v2`.
`import_puppetclasses_common_controller.rb` only uses v1 templates see: [import_puppetclasses_common_controller.rb:37](https://github.com/theforeman/foreman/blob/acf4b6870c6d39fc2ea42811a713e4fe704be340/app/controllers/concerns/api/import_puppetclasses_common_controller.rb#L37)

Should I make the import_puppetclasses_common_controller.rb render the template matching the requested api version?

domcleal:
> Should I make the import_puppetclasses_common_controller.rb render the template matching the requested api version?

Yes, I think so.

bastilian:
I made it use `api_version` to use in the path for the templates. This should do the trick.

dLobatog:
Thanks @bastilian - overall :+1: on the implementation and tests, it worked pretty well. I found a small bug though:
- Set a 'truthy' or 'falsy' value as one of the ignored_environments, e.g: 

``` yml
:ignored:
  - no 
  - yes
```
- Try to import environments: you get a 500

The problem is that Foreman is reading those keys as actual Ruby yes or no values, instead of as strings, so that `@changed['ignored']` becomes `{"test"=>["_ignored_"], false=>["_ignored_"], true=>["_ignored_"]}` , notice false and true are booleans there - and it fails. 

A way to fix this is just to stringify those keys when you make the hash, that should be enough. A test for that would be nice too. :smile:

test/controllers/api/v1/smart_proxies_controller_test.rb
```diff
@@ -223,4 +223,12 @@ def setup_import_classes
       assert_equal env_name, response['results']['name']
     end
   end
+
+  test 'should render templates according to api version 2' do
```
domcleal: version 1

app/views/api/v2/import_puppetclasses/show.json.rabl
```diff
@@ -25,3 +26,7 @@ end
 node(:removed_environment, :if => ->(environment) { @changed['obsolete'][environment.name].present? && @changed['obsolete'][environment.name].match(/_destroy_/) }) do |environment|
   environment.name
 end
+
+node(:ignored_environment, :if => ->(environment) { @changed['ignored'][environment.name].present? && @changed['ignored'][environment.name].match(/_ignored_/) }) do |environment|
```
domcleal, line 34: Also needs a corresponding `ignored_puppetclasses` node.

app/controllers/concerns/foreman/controller/environments.rb
```diff
@@ -21,7 +21,13 @@ def import_environments
     if @changed["new"].size > 0 || @changed["obsolete"].size > 0 || @changed["updated"].size > 0
       render "common/_puppetclasses_or_envs_changed"
     else
-      notice _("No changes to your environments detected")
+      notice_message = _("No changes to your environments detected")
+
+      if @changed['ignored'].present?
+        notice_message << "\n" + _("Ignored environments: %s") % @changed['ignored'].keys.to_sentence
```
domcleal: The message here is including names of environments that contain _some_ ignored classes (via the filter), not only environments that are entirely ignored, which is misleading.

domcleal (changes requested):

app/views/api/v2/import_puppetclasses/show.json.rabl
dLobatog: https://github.com/theforeman/foreman/pull/4012/files#diff-fda05e73ec11be4e3f4c60d62f9bd9e7R26 fixed here, GitHub isn't smart enough to notice yet

dLobatog:
Seems to still struggle with boolean-like environments:
`
] [I] Completed 500 Internal Server Error in 2500ms (ActiveRecord: 170.2ms)
09:11:39 rails.1   | 2016-11-18T09:11:39 a75456b0 [app] [F]
09:11:39 rails.1   |  | ArgumentError - comparison of FalseClass with true failed:
09:11:39 rails.1   |  |   app/views/common/_puppetclasses_or_envs_changed.html.erb:32:in`sort'
`

(the ones I have are just called 'yes' and 'no')

bastilian:
I was looking into the issue yesterday as well. Stringifying keys would prevent a 500, but The problem would be that they already come as `true` and `false`, so they end up as `"true"`, and would make it impossible to ignore `yes` or 'no' envs.

I'm not sure if it would be a good idea to "escape" 'yes' and 'no' (and other truthy/falsy), before handing it to YAML (just at this instance). Doing it after will not work, since all that were any of the truthy values will then result in just "true".

What do you think @dLobatog?

dLobatog:
@bastilian I'd say in that case, detect when the key is a boolean and show that as an error to the user, either in the list of ignored environments or in a 'notice' popup. 

Another solution could be to purposefully make Foreman look for those kind of keys in `ignored_environments.yml` and warn you about the specific env, but it's probably not worth the trouble (I don't expect that many people to have these envs).

bastilian:
Showing a notice describing the circumstances is I think a better approach.

I'll add a notice when booleans are found, pointing out that keys were stringified and results in them probably not being found. Plus advice them to quote these names in the yml.

Does that sound like a plan, @dLobatog?

dLobatog:
That sounds good to me - no need to get too much into detail,  it's fine to say that boolean environments are not imported properly and point to the UI. Thanks @bastilian!

bastilian:
All change requests should now be addressed. The warning in case Boolean values as environment names are found is now shown as well when importing:

screen shot 2017-01-03 at 09 29 42

The warning might be a bit long and for some users it might be very vague. Would it maybe be better to add a proper note to the "[Classes](https://theforeman.org/manuals/1.13/index.html#4.2.2Classes)" section in the manual, shorten the warning and link to the note?

dLobatog:
@domcleal Anything missing here? It seems to work well

dLobatog:
Thanks @bastilian @domcleal  - let me know if you were missing anything, I checked your last comments and they seemed to be addressed, didn't want this to PR get stuck

PR #4013, opened 2016-11-10
fixes #17300 - accept Nic::Bond#attached_devices string inputs

domcleal:


timogoebel:
Looks good to me and fixes the issue. Thanks, @domcleal.

PR #4015, opened 2016-11-11
Fixes #17297 - added PXE loader flags for Debian family

lzap:
Enables users to use Grub/Grub2 with Debian family systems.


dLobatog:
Thanks @lzap !

PR #4016, opened 2016-11-13
Fixes #16619 - Set EC2 VPC instance private IP...

alejandrocfg:
...from host primary nic.

### DIFF
--- a/app/models/compute_resource.rb
+++ b/app/models/compute_resource.rb
@@ -132,7 +132,7 @@ class ComputeResource < ActiveRecord::Base
 
   def host_interfaces_attrs(host)
     host.interfaces.select(&:physical?).each.with_index.reduce({}) do |hash, (nic, index)|
-      hash.merge(index.to_s => nic)
+      hash.merge(index.to_s => nic.compute_attributes.merge(ip: nic.ip, ip6: nic.ip6))
     end
   end
 
--- a/app/models/compute_resource.rb
+++ b/app/models/compute_resource.rb
@@ -132,7 +132,7 @@ class ComputeResource < ActiveRecord::Base
 
   def host_interfaces_attrs(host)
     host.interfaces.select(&:physical?).each.with_index.reduce({}) do |hash, (nic, index)|
-      hash.merge(index.to_s => nic.compute_attributes)
+      hash.merge(index.to_s => nic)
     end
   end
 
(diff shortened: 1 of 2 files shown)


mmoll:
ok to test

app/models/compute_resource.rb
```diff
@@ -132,7 +132,7 @@ def host_compute_attrs(host)
 
   def host_interfaces_attrs(host)
     host.interfaces.select(&:physical?).each.with_index.reduce({}) do |hash, (nic, index)|
-      hash.merge(index.to_s => nic.compute_attributes)
+      hash.merge(index.to_s => nic)
```
domcleal: I don't think changing this is a good idea, it'll completely change the per-NIC attributes passed to every CR's interface creation, breaking them all. It might be best to _merge_ the general NIC model attributes (`ip`, `ip6`) with the compute_attributes.

domcleal (changes requested):
Please also address the commit message feedback from @theforeman-bot, there should only be one "#" and the message should be shorter. Move the bulk of it into the description if that helps.

app/models/compute_resource.rb
alejandrocfg: @domcleal thanks, does the following looks better to you?

`hash.merge(index.to_s => nic.compute_attributes)`
`hash.merge(index.to_s => { :ip => nic[:ip], **(nic[:ip6] ? {:ip6 => nic[:ip6]} : {}) })`

If not, where would be the best place to merge the IP attributes?

alejandrocfg: Made ip6 conditional so it works in versions older than 1.13.

domcleal: > hash.merge(index.to_s => { :ip => nic[:ip], **(nic[:ip6] ? {:ip6 => nic[:ip6]} : {}) })

This will overwrite the first `index.to_s` entry that had `nic.compute_attributes`, so use:

```
hash.merge(index.to_s => nic.compute_attributes.merge(ip: nic.ip, ip6: nic.ip6))
```

Don't worry about making ip6 optional, the patch should only be written for develop, not for backwards compatibility.

app/models/compute_resources/foreman/model/ec2.rb
```diff
@@ -51,6 +51,7 @@ def create_vm(args = { })
       args[:groups].reject!(&:empty?) if args.has_key?(:groups)
       args[:security_group_ids].reject!(&:empty?) if args.has_key?(:security_group_ids)
       args[:associate_public_ip] = subnet_implies_is_vpc?(args) && args[:managed_ip] == 'public'
+      args[:private_ip_address] = args[:interfaces_attributes][:"0"][:ip]
```
timogoebel, line 54: How does this handle the case when the VM has no ip set by Foreman, i.e. when EC2 provides the ip? It might not be a problem, just asking.

alejandrocfg: I tested that as well. If you don't set and IP address in the host interface amazon will assign one free from the Subnet selected  for that VPC instance because private_ip_address will have no value.

domcleal:
Merged, thanks @alejandrocfg.

PR #4017, opened 2016-11-13
Fixes #17309 - Made facets resilient to class reloading.

ShimShtein:
Now the configuration is stored in plugin declaration.
If you are using `Facets.register` directly, please make sure
you are re-registering it after reload occurs.


dLobatog:
Thanks @ShimShtein !

PR #4018, opened 2016-11-13
Fixes #17303 - correct position of validation msg for parameters

kgaikwad:
Instead of showing all validation error messages on top, this commit will show each validation error message on the top of respective parameter.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -154,7 +154,9 @@ function check_caps_lock(key, e) {
 
 function remove_fields(link) {
   $(link).prev("input[type=hidden]").val("1");
-  $(link).closest(".fields").hide();
+  var $field_row = $(link).closest(".fields");
+  $field_row.next("tr.error-msg-block").hide();
+  $field_row.hide();
   mark_params_override();
 }
 
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -590,3 +590,21 @@ code.transparent {
   color: black;
   background-color: transparent;
 }
+
+#parameters {
+  .table {
+    tr {
+      td {
+        border-top: none;
+      }
+      &.error-msg-block td {
+        .help-block {
+          margin-top: 0px;
+        }
+      }
+      &.fields td {
+        padding-top: 10px;
+      }
+    }
+  }
+}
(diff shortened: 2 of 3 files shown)


tbrisker:
Ok to test

orrabin:
The error is now in red and above the parameter that is actually causing it as requested:
image

Maybe stretch the error on the whole row so it doesn't push down each row so much?

Also, please change the commit message to the current status and not the bug, for example it can be: `Fixes #17303 - correct position of validation msg for parameters`

kgaikwad:
@orrabin, I have made suggested code changes. Added "colspan" attribute to table row cell.
Please have a look.

app/assets/javascripts/application.js
```diff
@@ -154,7 +154,10 @@ function check_caps_lock(key, e) {
 
 function remove_fields(link) {
   $(link).prev("input[type=hidden]").val("1");
-  $(link).closest(".fields").hide();
+  var $field_row = $(link).closest(".fields"),
+  $error_block = $field_row.prev("tr.error-msg-block");
+  if ($error_block) { $error_block.hide(); }
```
orrabin: Nitpick: no need for the if, calling hide on an empty error_block won't cause an error.

orrabin:
@kgaikwad I tested again and it looks much better when stretched, thanks for the change.
Comparing this to the other errors in tables I noticed 2 things:
1) We don't remove the error when clicking remove :+1:  for adding that, maybe a separate pr can remove it in other places.
2) The rest of the tables put the error message under the row that causes it and not above.
Can you try moving the error under the row?

kgaikwad:
@orrabin , done with the changes.
As I have removed top border of all rows for better UI with error indication, added little padding on top of each field row and removed the top margin of error row.

kgaikwad:
[Test]

orrabin:
[test foreman]

orrabin:
@Rohoover this pr fixes the location of validation errors in the parameters table, they used to be above the table:
image

After this pr they are located correctly in the table:
image

but the table doesn't have borders to avoid borders between the parameter and the error (avoid the table below):
image

is it better to keep the table without any borders? with the borders always (even when there is an error)? should it look different?

Rohoover:
@orrabin

I don't think the table borders (lines) are necessary as the fields themselves clearly define the rows. If borders are required, I would not put the error in its own row with borders - that visually separates the error from the field and could be confusing. 

Thanks!

orrabin:
Thanks @kgaikwad!

PR #4019, opened 2016-11-14
Fixes #17332 - Test email error renders propertly in user email preferences

amirfefer:
When test email didn't succeed in User -> Email Preferences, an error message was rendered not properly:
 
error_mail


domcleal:
Merged, thanks @amirfefer.

PR #4020, opened 2016-11-15
Fixes #17330 - skip admin email if it's unknown

ares:


domcleal:
I receive a different error when testing this from lib/tasks/config.rake - it enhances db:seed to set the db_pending_migration/seed settings to false, but they're not yet initialised:

<pre>
$ rm db/development.sqlite3; spring stop; bin/rake db:migrate; bin/rake db:seed
[..]
2016-11-15T12:02:57  [sql] [D]   ProvisioningTemplate Load (0.3ms)  SELECT "templates".* FROM "templates" WHERE "templates"."type" IN ('ProvisioningTemplate') AND "templates"."default" = ? AND "templates"."name" IN ('Boot disk iPXE - host', 'Boot disk iPXE - generic host')  ORDER BY templates.name  [["default", "t"]]
All seed files executed
2016-11-15T12:02:57  [sql] [D]   Setting Load (0.1ms)  SELECT  "settings".* FROM "settings" WHERE "settings"."name" = ?  ORDER BY "settings"."name" ASC LIMIT 1  [["name", "db_pending_migration"]]
rake aborted!
NoMethodError: undefined method `[]' for nil:NilClass
/home/dcleal/code/foreman/foreman/app/models/setting.rb:185:in `create'
/home/dcleal/.rvm/gems/ruby-2.3.0@foreman/gems/activerecord-4.2.7.1/lib/active_record/relation.rb:143:in `block in create'
/home/dcleal/.rvm/gems/ruby-2.3.0@foreman/gems/activerecord-4.2.7.1/lib/active_record/relation.rb:302:in `scoping'
/home/dcleal/.rvm/gems/ruby-2.3.0@foreman/gems/activerecord-4.2.7.1/lib/active_record/relation.rb:143:in `create'
/home/dcleal/.rvm/gems/ruby-2.3.0@foreman/gems/activerecord-4.2.7.1/lib/active_record/relation.rb:155:in `first_or_create'
/home/dcleal/code/foreman/foreman/app/models/setting.rb:91:in `[]='
/home/dcleal/code/foreman/foreman/lib/tasks/config.rake:175:in `block in <top (required)>'
/home/dcleal/.rvm/gems/ruby-2.3.0@foreman/gems/activesupport-4.2.7.1/lib/active_support/dependencies.rb:268:in `load'
/home/dcleal/.rvm/gems/ruby-2.3.0@foreman/gems/activesupport-4.2.7.1/lib/active_support/dependencies.rb:268:in `block in load'
/home/dcleal/.rvm/gems/ruby-2.3.0@foreman/gems/activesupport-4.2.7.1/lib/active_support/dependencies.rb:240:in `load_dependency'
/home/dcleal/.rvm/gems/ruby-2.3.0@foreman/gems/activesupport-4.2.7.1/lib/active_support/dependencies.rb:268:in `load'
-e:1:in `&lt;main>'
Tasks: TOP => db:seed
(See full trace by running task with --trace)
</pre>

The cause seems to be the preloading of the environment via spring, because when not using spring then the settings are initialised during the startup of `bin/rake db:seed`. Using spring, the environment's already running from `bin/rake db:migrate` but load_defaults is never called.

You could add further workarounds to config.rake, I suppose, but the settings won't then be set correctly.

IMHO settings need to move into seeds themselves instead of relying on initialisers, but that's a significant change.

ares:
I wonder if there was a reason why not having settings in seed in the first place. Maybe to keep settings table clean until the value is changed and ability to have default value dynamic e.g. email using `Facter.value(:domain)`?

domcleal:
Seeds were added after settings, and I think only moved data that was already in DB migrations. Settings always existed with their separate mechanism.

You're right though, changing default values is an additional feature/complexity of the current method (though probably not one that matters).

ares:
Sorry, I don't have capacity to migrate settings to seed. I leave up to you the decision whether we take this "partial" fix or not, at least, there's a workaround. Also spring issue has lower priority for me, since it's only in dev environment. Feel free to close the PR.

domcleal:
I don't think this should be merged while it's still throwing an error, that needs to be fixed before merge. Feel free to close if you wish, or improve the fix.

ares:
I added the workaround in a separate commit even though I think it's completely unrelated issue. One problem is that `Setting.where(:name => 'a').first_or_create` calls our redefined `#create` with `nil` as `opts` causing the exception. Second is that even after correctly loading name in such case, when the `config.rake` task tries to set new settings value, the setting has not been seeded yet. Calls like `Setting['whatever'] = value` always fail when `'whatever'` is not seeded yet because it's missing value, default and description attributes and validations fail. After fixing first one, second fails silently.

domcleal (changes requested):
Please check the test failures and rebase to fix the integration test failures.

ares:
fixed, tests should pass now

domcleal:
> Calls like Setting['whatever'] = value always fail when 'whatever' is not seeded yet because it's missing value, default and description attributes and validations fail. After fixing first one, second fails silently.

It doesn't seem to fail silently for me, how should this happen?

<pre>
2016-11-24T10:09:07  [sql] [D]   Setting Load (0.1ms)  SELECT  "settings".* FROM "settings" WHERE "settings"."name" = ?  ORDER BY "settings"."name" ASC LIMIT 1  [["name", "db_pending_migration"]]
2016-11-24T10:09:07  [sql] [D]   Setting Load (0.2ms)  SELECT  "settings".* FROM "settings" WHERE "settings"."name" = ? AND "settings"."name" = ?  ORDER BY "settings"."name" ASC LIMIT 1  [["name", "db_pending_migration"], ["name", "db_pending_migration"]]
2016-11-24T10:09:07  [sql] [D]    (0.0ms)  begin transaction
2016-11-24T10:09:07  [sql] [D]   Setting Exists (0.0ms)  SELECT  1 AS one FROM "settings" WHERE "settings"."name" = 'db_pending_migration' LIMIT 1
2016-11-24T10:09:08  [sql] [D]    (0.1ms)  rollback transaction
2016-11-24T10:09:08  [sql] [D]    (0.0ms)  begin transaction
2016-11-24T10:09:08  [sql] [D] String does not start with the prefix 'encrypted-', so Setting db_pending_migration was not decrypted
2016-11-24T10:09:08  [sql] [D]   Setting Exists (0.1ms)  SELECT  1 AS one FROM "settings" WHERE "settings"."name" = 'db_pending_migration' LIMIT 1
2016-11-24T10:09:08  [sql] [D]    (0.0ms)  rollback transaction
rake aborted!
ActiveRecord::RecordInvalid: Validation failed: »Description« can't be blank, »Default« can't be blank
/home/dcleal/.rvm/gems/ruby-2.3.0@foreman/gems/activerecord-4.2.7.1/lib/active_record/validations.rb:79:in `raise_record_invalid'
/home/dcleal/.rvm/gems/ruby-2.3.0@foreman/gems/activerecord-4.2.7.1/lib/active_record/validations.rb:43:in `save!'
/home/dcleal/.rvm/gems/ruby-2.3.0@foreman/gems/activerecord-4.2.7.1/lib/active_record/attribute_methods/dirty.rb:29:in `save!'
/home/dcleal/.rvm/gems/ruby-2.3.0@foreman/gems/activerecord-4.2.7.1/lib/active_record/transactions.rb:291:in `block in save!'
/home/dcleal/.rvm/gems/ruby-2.3.0@foreman/gems/activerecord-4.2.7.1/lib/active_record/transactions.rb:351:in `block in with_transaction_returning_status'
/home/dcleal/.rvm/gems/ruby-2.3.0@foreman/gems/activerecord-4.2.7.1/lib/active_record/connection_adapters/abstract/database_statements.rb:213:in `block in transaction'
/home/dcleal/.rvm/gems/ruby-2.3.0@foreman/gems/activerecord-4.2.7.1/lib/active_record/connection_adapters/abstract/transaction.rb:184:in `within_new_transaction'
/home/dcleal/.rvm/gems/ruby-2.3.0@foreman/gems/activerecord-4.2.7.1/lib/active_record/connection_adapters/abstract/database_statements.rb:213:in `transaction'
/home/dcleal/.rvm/gems/ruby-2.3.0@foreman/gems/activerecord-4.2.7.1/lib/active_record/transactions.rb:220:in `transaction'
/home/dcleal/.rvm/gems/ruby-2.3.0@foreman/gems/activerecord-4.2.7.1/lib/active_record/transactions.rb:348:in `with_transaction_returning_status'
/home/dcleal/.rvm/gems/ruby-2.3.0@foreman/gems/activerecord-4.2.7.1/lib/active_record/transactions.rb:291:in `save!'
/home/dcleal/code/foreman/foreman/app/models/setting.rb:93:in `[]='
/home/dcleal/code/foreman/foreman/lib/tasks/config.rake:175:in `block in <top (required)>'
/home/dcleal/.rvm/gems/ruby-2.3.0@foreman/gems/activesupport-4.2.7.1/lib/active_support/dependencies.rb:268:in `load'
/home/dcleal/.rvm/gems/ruby-2.3.0@foreman/gems/activesupport-4.2.7.1/lib/active_support/dependencies.rb:268:in `block in load'
/home/dcleal/.rvm/gems/ruby-2.3.0@foreman/gems/activesupport-4.2.7.1/lib/active_support/dependencies.rb:240:in `load_dependency'
/home/dcleal/.rvm/gems/ruby-2.3.0@foreman/gems/activesupport-4.2.7.1/lib/active_support/dependencies.rb:268:in `load'
-e:1:in `<main>'
Tasks: TOP => db:seed
(See full trace by running task with --trace)
</pre>

ares:
Hm, seems like I missed to kill spring or it has started somewhere between my tests, I was able to reproduce. I modified the config.rake so it loads the data from the `Setting::General`. I'm eager to split the `load_data` in other setting subclasses if you agree that's the way to go. Long term we need to clean this mess. Thanks for another look @domcleal

ares:
And I rebased the whole branch.

domcleal (approved):

domcleal:
> I'm eager to split the load_data in other setting subclasses if you agree that's the way to go.

Sure, it's best if they follow the same pattern else it's unclear why it's different. Please also then squash the commits with some explanation of why these changes were made.

ares:
Both done, @domcleal please take another look.

domcleal:
Please fix the code style comments and rebase the PR on develop for test fixes.

ares:
@domcleal ready again

domcleal:
Merged, thanks @ares.

PR #4021, opened 2016-11-15
Fixes #17334 - ignore username in password

ares:


domcleal:
I think the issue here is simply the wording of the message. It shouldn't say "cannot", but "should not".

The messages are advisory to help users increase the strength of their password, not to inform them of rules being enforced - this one is worded badly. There are many other checks that are enabled which will show similar messages but are also not enforced on the server, so I don't think disabling this one check is correct.

ares:
I think the whole rule does not make much sense, how does username in 'admin_sflkjsdflksjflksfjlsafjsaflsjdfsfd' make the password less secure? Anyway, if you prefer, I can change the wording instead.

domcleal:
It's quite a common password complexity requirement I think, and it's there to prevent the password being predictable for an attacker, i.e. the password "virtwho00" in the ticket is very predictable if you know the username "virtwho".

If the password's sufficient complex then the warnings are shown in orange/green so could be ignored by the user. I think the wording is the issue here - it's suggesting that it won't be accepted, not that it's advice.

ares:
Thanks @domcleal for quick response. I've modified the wording for this and other messages, please let me know if it's better or where the language could be improved.

domcleal:
Thanks @ares, merged with an updated commit message.

PR #4023, opened 2016-11-15
Fixes #14951 - Allow matcher for host turning into managed host

orrabin:
I've only been able to test this with discovery, I can add a corresponding discovery pr with tests unless someone has an idea how to test this in core.

### DIFF
--- a/app/models/lookup_value.rb
+++ b/app/models/lookup_value.rb
@@ -75,7 +75,8 @@ class LookupValue < ActiveRecord::Base
     md = ensure_matcher(/fqdn=(.*)/)
     return md if md == true || md == false
     fqdn = md[1].split(LookupKey::KEY_DELM)[0]
-    return true if Host.unscoped.find_by_name(fqdn) || host_or_hostgroup.try(:new_record?)
+    return true if Host.unscoped.find_by_name(fqdn) || host_or_hostgroup.try(:new_record?) ||
+        (host_or_hostgroup.present? && host_or_hostgroup.type_changed? && host_or_hostgroup.type == "Host::Managed")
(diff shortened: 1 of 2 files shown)


lzap:
As per our IRC discussion, can you please add a test for this in discovery plugin? I will test it together.

orrabin:
@lzap I was able to reproduce in a test in core so there is no need for a test in discovery

app/models/lookup_value.rb
```diff
@@ -75,7 +75,8 @@ def ensure_fqdn_exists
     md = ensure_matcher(/fqdn=(.*)/)
     return md if md == true || md == false
     fqdn = md[1].split(LookupKey::KEY_DELM)[0]
-    return true if Host.unscoped.find_by_name(fqdn) || host_or_hostgroup.try(:new_record?)
+    return true if Host.unscoped.find_by_name(fqdn) || host_or_hostgroup.try(:new_record?) ||
+        (host_or_hostgroup.present? && host_or_hostgroup.type_changed? && host_or_hostgroup.type == "Host::Managed")
```
lzap, line 79: What is the `host_or_hostgroup.present?` for? I mean, you can create discovered hosts without hostgroups just fine.

unorthodoxgeek:
@mention-bot - can you please not tag me? 🚯

app/models/lookup_value.rb
orrabin: `host_or_hostgroup` is the discovered host itself, you can either create a lookup_value matcher yourself or override a parameter from the host or hostgroup in which can host_or_hostgroup will be the host/hostgroup that created an override which triggers a creation of the matcher.

lzap:
@unorthodoxgeek https://github.com/mention-bot/how-to-unsubscribe

lzap:
Thanks, merged.

PR #4024, opened 2016-11-15
Fixes #17015 - Adds Key pairs controller

shlomizadok:


ohadlevy:
@shlomizadok note failing tests and hound errors?

shlomizadok:
Yeah, also adding KeyPair tests :)

2016-11-17 17:25 GMT+02:00 Ohad Levy notifications@github.com:

app/controllers/key_pairs_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+  before_action :find_compute_resource, :except => [:download]
+  before_action :find_resource, :only => [:download]
+
+  def index
+    @key_pairs = @compute_resource.get_compute_key_pairs
+  end
+
+  def download
+    @key_pair.update_attributes(audit_comment: _("%{user} Downloaded %{key} as pem file") % {user: User.current.name, key: @key_pair.name})
+    send_data @key_pair.secret, :filename => "#{@key_pair.name}.pem" if @key_pair
```
ohadlevy: i think `@key_pair` will always be defined because of the before_action?

app/controllers/key_pairs_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def recreate
```
ohadlevy: does it make sense to keep REST semantics? e.g. in this case this will be PUT  (/edit)

app/controllers/key_pairs_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+    if @compute_resource.try(:recreate)
+      process_success :success_msg => _('Successfully created'),
+                      :success_redirect => compute_resource_key_pairs_path(@compute_resource)
+    else
+      process_error :error_msg => _('Could not recreate a new key pair')
+    end
+  end
+
+  def remove
```
ohadlevy: ditto for destroy? (DELETE)

app/models/concerns/key_pair_compute_resource.rb
```diff
@@ -0,0 +1,55 @@
...
+    delete_key_from_resource(kay_pair_name)
+  end
+
+  private
+
+  def setup_key_pair
+    key = client.key_pairs.create :name => "foreman-#{id}#{Foreman.uuid}"
+    KeyPair.create! :name => key.name, :compute_resource_id => self.id, :secret => key.private_key
+  rescue => e
+    Foreman::Logging.exception("Failed to generate key pair", e)
```
ohadlevy, line 36: does it make sense to include the compute resource id/name?

app/models/concerns/key_pair_compute_resource.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def destroy_key_pair
+    return unless key_pair
+    delete_key_from_resource(key_pair.name)
+    key_pair.destroy!
+  end
+
+  def delete_key_from_resource(key_pair_name)
+    logger.info "removing #{provider_friendly_name} key #{key_pair_name}"
```
ohadlevy: consider adding the compute resource name? e.g. friendly name will only show amazon, but not the region?

app/models/key_pair.rb
```diff
@@ -1,8 +1,9 @@
 class KeyPair < ActiveRecord::Base
-  belongs_to :compute_resource
+  belongs_to :compute_resource, :dependent => :destroy
```
ohadlevy: wont this delete the CR if the key is removed?

app/services/compute_resource_key_pairs/base.rb
```diff
@@ -0,0 +1,23 @@
...
+    end
+
+    private
+
+    def is_active?(active_key_name)
+      @name == active_key_name
+    end
+
+    def used_elsewhere?
+      KeyPair.where(name: @name).size > 0 && !@active
```
ohadlevy: i would swap the condition, so there is no extra query in the active case.

app/views/audits/show.html.erb
```diff
@@ -51,6 +51,13 @@
         <% end %>
         </tbody>
       </table>
+      <% if @audit.comment %>
+        <div class="alert alert-info">
```
ohadlevy: use the alert helper?

ohadlevy:
@shlomizadok thanks! looks good, would you mind adding a couple of screenshots?

ohadlevy (changes requested):

app/models/key_pair.rb
shlomizadok: Someone, who is definitely not @shlomizadok, has added that. :no_mouth:

app/controllers/key_pairs_controller.rb
shlomizadok: Right. Hence, I am not looking for it here :)

app/controllers/key_pairs_controller.rb
shlomizadok: It is in `PUT`

shlomizadok:
Key pairs controller shows all of the available key pairs for a compute resource

image

With the option to delete unused key pairs, or download || recreate active key pair

image

app/controllers/key_pairs_controller.rb
ohadlevy: you still have the unnecessary if at the end of the line.

config/routes.rb
```diff
@@ -401,6 +401,15 @@
           put 'test_connection'
         end
         resources :images, :except => [:show]
+        resources :key_pairs, :only => [:index] do
+          member do
+            get 'download'
+          end
+          collection do
+            put 'recreate'
+            delete 'remove'
```
ohadlevy: this assumes you can only remove the compute resource key? I think we should support removing other keys as well. in that case you would still need to pass the key id.

config/routes.rb
```diff
@@ -401,6 +401,15 @@
           put 'test_connection'
         end
         resources :images, :except => [:show]
+        resources :key_pairs, :only => [:index] do
+          member do
+            get 'download'
+          end
+          collection do
+            put 'recreate'
```
ohadlevy: this is not mapped to REST semantics, you are doing a PUT on a collection, imho it should be a PUT on the key itself.
it makes sense to do a POST on a collection (creates a new instance of a key) but not a PUT IMHO.

app/services/compute_resource_key_pairs/base.rb
```diff
@@ -0,0 +1,23 @@
...
+    end
+
+    private
+
+    def is_active?(active_key_name)
+      @name == active_key_name
+    end
+
+    def used_elsewhere?
+      !@active && KeyPair.where(name: @name).size > 0
```
ohadlevy: shouldnt this be OR ?

ohadlevy: nitpick: can write where().any? instead

ohadlevy:
@shlomizadok I'm not sure I like that there is icon only for the active case, there are basicilly 3 different cases here:
1. key is used by the compute resource (the current active icon you have).
2. key is used by  another computer resource in foreman (so should be kind of in use elsewhere indication)
3. key is not used and can be deleted.

would it make sense to have difference color of icons to indicate in use / not used?

/cc @Rohoover

app/controllers/key_pairs_controller.rb
ohadlevy: its in put on the collection.

app/controllers/key_pairs_controller.rb
ohadlevy: where is the actual error message? (e.g. why it failed)

app/controllers/key_pairs_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+    key_to_delete = params[:key_name]
+    return not_found unless key_to_delete
+    if @compute_resource.delete_key_pair(key_to_delete)
+      process_success :success_msg => _('Successfully delete %s') % key_to_delete,
+                      :success_redirect => compute_resource_path(@compute_resource)
+    else
+      process_error :error_msg => _('Could not delete %s') % key_to_delete
```
ohadlevy, line 34: ditto about the actual error.

app/models/compute_resources/foreman/model/ec2.rb
```diff
@@ -29,7 +25,7 @@ def self.model_name
     end
 
     def capabilities
-      [:image]
+      [:image, :key_pair]
```
ohadlevy: nitpick: I wonder if this should be done as part of the mixin

app/models/concerns/key_pair_compute_resource.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  included do
+    has_one :key_pair, :foreign_key => :compute_resource_id, :dependent => :destroy
+    after_create :setup_key_pair
+    after_destroy :destroy_key_pair
+    delegate :key_pairs, :to => :client
+  end
+
+  def get_compute_key_pairs
+    return unless capabilities.include?(:key_pair)
```
ohadlevy: shouldnt this return an empty array?

app/models/concerns/key_pair_compute_resource.rb
```diff
@@ -0,0 +1,55 @@
...
+    key_pairs = self.try(:key_pairs)
```
ohadlevy: i think this line is a noop. is it required?

app/models/concerns/key_pair_compute_resource.rb
```diff
@@ -0,0 +1,55 @@
...
+      compute_resource_key_pair_class.new(key.name, key.fingerprint, active_key.name, active_key.id)
+    end
+  end
+
+  def recreate
+    destroy_key_pair
+    setup_key_pair
+  end
+
+  def delete_key_pair(kay_pair_name)
```
ohadlevy, line 26: i wonder if all of the deleation of keys (that are not used by the compute resource) should be in the ComputeResourceKeyPairs::Base class.

app/models/concerns/key_pair_compute_resource.rb
```diff
@@ -0,0 +1,55 @@
...
+    delete_key_from_resource(kay_pair_name)
+  end
+
+  private
+
+  def setup_key_pair
+    key = client.key_pairs.create :name => "foreman-#{id}#{Foreman.uuid}"
+    KeyPair.create! :name => key.name, :compute_resource_id => self.id, :secret => key.private_key
```
ohadlevy: nitpick, can drop the self

app/services/compute_resource_key_pairs/ec2.rb
```diff
@@ -0,0 +1,4 @@
+module ComputeResourceKeyPairs
+  class EC2 < Base
```
ohadlevy: do we actually need any of those child classes?

test/models/compute_resources/ec2_test.rb
```diff
@@ -17,9 +17,43 @@ class EC2Test < ActiveSupport::TestCase
...
+
+  context "key pairs" do
+    setup do
+      @aws_key_pairs = []
+      3.times do |i|
+        @aws_key_pairs << AWSKeyPair.new("foreman-#{1}", "13:01:73:0#{i}")
+      end
+    end
+
+    test "#get_compute_key_pairs" do
```
ohadlevy, line 34: nitpick why # in method name?

test/models/key_pair_test.rb
```diff
@@ -0,0 +1,12 @@
...
+class KeyPairTest < ActiveSupport::TestCase
+  should belong_to(:compute_resource)
+
+  test "ensure validations" do
+    key = KeyPair.new
+    refute key.valid?
+    assert_equal([:name, :secret, :compute_resource_id], key.errors.messages.keys)
+    assert_equal(["can't be blank"], key.errors.messages.values.first.compact)
+  end
+end
```
ohadlevy, line 44: could you please add tests to the ComputeResourceKeyPairs::Base class? especially around active and used elsewhere?

app/services/compute_resource_key_pairs/base.rb
ares: or even better `KeyPair.exists?(name: @name)`

app/services/compute_resource_key_pairs/ec2.rb
shlomizadok: Not sure. Thought of creating them so we could extend the base class in the future

test/models/compute_resources/ec2_test.rb
shlomizadok: Standard for "this test tests a method with the same name..." :)

app/models/concerns/key_pair_compute_resource.rb
shlomizadok: I'd like to separate. The compute resource is in charge of manipulating the resource (e.g., create a vm, create a key pair, destroy a key pair, etc.) The single responsibility of `ComputeResourceKeyPairs::Base` is to represent the key pair, and to decide if it is safe to delete / recreate it.

app/services/compute_resource_key_pairs/base.rb
shlomizadok: Should be `AND`. e.g., it is not the active key_pair **and** it exists?

app/models/compute_resources/foreman/model/ec2.rb
shlomizadok: Thought about that, yet the `:image` is not a part of the mixin - so left it here.

config/routes.rb
shlomizadok: On the contrary, it removes key with name as a parameter. Not the KeyPair

app/models/compute_resources/foreman/model/ec2.rb
ohadlevy: you can do alias_method_chain and add to the respond the key_pair symbol.

config/routes.rb
```diff
@@ -401,6 +401,15 @@
           put 'test_connection'
         end
         resources :images, :except => [:show]
+        resources :key_pairs, :only => [:index] do
+          member do
+            get 'download'
+            put 'recreate'
+          end
+          collection do
+            delete 'remove'
```
ohadlevy: why is this still on the collection? you are deleting a specific ID, so the route should be ../key_pairs/ID and the verb is DELETE).

shlomizadok: because I am not deleting an AR object with id. I am deleting a Fog::Compute::AWS::KeyPair object which has a name. The name is passed as a parameter. 
hmm... I guess I can pass the name param as id

ohadlevy:
correct.. hence my desire to split the key pairs from the compute resource
object a bit...

Rohoover:
@ohadlevy 

Maybe green checkmark, yellow check mark, and  perhaps fa-circle-o in grey for when not in use and can be deleted.

app/services/compute_resource_key_pairs/ec2.rb
ares: I wish we had generic KeyPair object which would use adapter/mapper for each CR type. Adding these children objects leads to inconsistent API among CR types and many `ifs` in view layer. If we don't need it now, I'd suggest keeping it in a single, shared class. If there will be some changes required later, I'd like to adopt the adapter pattern.

app/models/compute_resources/foreman/model/ec2.rb
shlomizadok: @ohadlevy - is it worth it?

shlomizadok:
@Rohoover @ohadlevy - So it looks like:
image

Where green ok is active, yellow ok is used elsewhere and circle is safe to delete.

Rohoover:
@shlomizadok bitmoji

app/assets/stylesheets/patternfly_and_overrides.scss
```diff
@@ -315,3 +315,7 @@ a.btn-info span{
   max-height: 80vh;
   overflow-y: auto;
 }
+
+.pficon-ok.warn:before {
+  color: #ec7a08;
```
ohadlevy, line 320: nitpick: I would add a comment this is pf-orange-400 color, just in case we ever search for it.

shlomizadok:
@Rohoover 
image

@ohadlevy, will add a comment :)

app/controllers/key_pairs_controller.rb
shlomizadok: It is logged

app/models/concerns/key_pair_compute_resource.rb
shlomizadok: Added

app/controllers/key_pairs_controller.rb
ohadlevy: also visible in the UI? or rater oops, check log files if you want to know how to handle it?

app/helpers/application_helper.rb
```diff
@@ -128,8 +128,9 @@ def link_to_function_if_authorized(name, function, options = {}, html_options =
 
   def display_delete_if_authorized(options = {}, html_options = {})
     text = options.delete(:text) || _("Delete")
+    method = options.delete(:method) || :delete
     options = {:auth_action => :destroy}.merge(options)
-    html_options = { :data => { :confirm => _('Are you sure?') }, :method => :delete, :class => 'delete' }.merge(html_options)
+    html_options = { :data => { :confirm => _('Are you sure?') }, :method => method, :class => 'delete' }.merge(html_options)
```
ohadlevy, line 133: is this still required after the route changes?

app/models/compute_resources/foreman/model/ec2.rb
ohadlevy: imho yes, as its unexpected that mixin does everything besides this.

app/views/key_pairs/index.html.erb
```diff
@@ -0,0 +1,39 @@
...
+  <thead>
+  <tr>
+    <th width="6%"><%= _('Active key') %></th>
+    <th width="40%"><%= _('Name') %></th>
+    <th width="40%"><%= _('Fingerprint') %></th>
+    <th width="100px"><%= _('Actions') %></th>
+  </tr>
+  </thead>
+  <tbody>
+  <% @key_pairs.each do |key_pair| %>
```
ohadlevy: what if there are no key pairs(deleted by mistake - e.g. in ec2 console)? I think it make sense to allow to recreate the key? if you go that path, maybe consider adding patternfly empty state ?

app/helpers/application_helper.rb
shlomizadok: Yes, for the `recreate` action, which I also ask "Are you sure?"
Regardless, the whole "Are you sure?" scenario should be rewritten, imo :smile:

app/views/key_pairs/index.html.erb
shlomizadok: Added

shlomizadok:
@ohadlevy - Added an empty state to key pairs index.

app/helpers/application_helper.rb
ohadlevy: mind opening a new issue for that?

app/controllers/key_pairs_controller.rb
ohadlevy: bump on this two errors on eating the exceptions?

app/models/compute_resources/foreman/model/ec2.rb
ohadlevy: can you simply add alias_method_chain in your mixin?

ohadlevy: and a test of course :)

app/views/key_pairs/index.html.erb
```diff
@@ -0,0 +1,53 @@
+<% title _("Key pairs for: %s") % @compute_resource.name %>
+
+<table class="<%= table_css_classes 'table-fixed' %>">
+  <% if @key_pairs.any? %>
+  <thead>
+  <tr>
+    <th width="6%"><%= _('Active key') %></th>
```
ohadlevy: should we drop the title? or just call it something else?

app/views/key_pairs/index.html.erb
```diff
@@ -0,0 +1,53 @@
...
+    <th width="40%"><%= _('Name') %></th>
```
ohadlevy: or maybe just fold it in next to the name? (without its on TH)

app/views/key_pairs/index.html.erb
```diff
@@ -0,0 +1,53 @@
...
+  </tbody>
+  <% else %>
+    <tbody>
+      <tr>
+        <td class="blank-slate-pf">
+            <div class="blank-slate-pf-icon">
+              <%= icon_text("key", "", :kind => "fa") %>
+            </div>
+            <h1><%= _('Key pairs') %></h1>
+            <p><%= _("Key pair (private and public keys) help connecting to VMs on a compute resource.") %></p>
```
ohadlevy: I would write:
```
When using SSH based provisioning, a special SSH key is created in order to automate the provisioning process.
Normally when setting up a Compute Resource, a key pair (Private and Public) is created for you automatically, however, if it was deleted externally this page you to recreate it
```

app/views/key_pairs/index.html.erb
shlomizadok: Like that?
image

app/helpers/application_helper.rb
shlomizadok: Will do

app/views/key_pairs/index.html.erb
ohadlevy: I think so.. what do you think?

shlomizadok: When looking closely, the "Active" icon seems a little bigger from the other icons.
When it was on a separate column, I did not notice that, now I do.

shlomizadok: image

It may be only me, but this itches me :)
Previous solution works better, maybe only change the title (or remove the title completely)?

ohadlevy: will defer to @Rohoover

ohadlevy:
also note failing tests.

shlomizadok:
@Rohoover @ohadlevy - so we've got three options to display the icons:
1. little column, no title
image

2. (current) little column title: Active Key
image

3. remove column, attach to name (which I dislike because of the icons sizes
image
image

:+1: for option # 1, :heart: for option # 2 or :laughing: for option # 3

shlomizadok:
@ohadlevy - changed UI to look like option # 1, added a prepend to override capabilties (+test)

app/helpers/application_helper.rb
ohadlevy: can't we simply re-use create as recreate? I don't think we use it at all?

shlomizadok:
@ohadlevy - removed `recreate` action, using `create` action only. Thanks!

app/assets/stylesheets/patternfly_and_overrides.scss
```diff
@@ -315,3 +315,7 @@ a.btn-info span{
   max-height: 80vh;
   overflow-y: auto;
 }
+// Color origin: pf-orange-400
+.pficon-ok.warn:before {
```
ohadlevy, line 319: is this really the right place to put this? wont this change all references to this icon?

app/helpers/application_helper.rb
ohadlevy: this change is no longer required afaiu?

app/controllers/key_pairs_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def download
+    @key_pair.update_attributes(audit_comment: _("%{user} Downloaded %{key} as pem file") % {user: User.current.name, key: @key_pair.name})
+    send_data @key_pair.secret, :filename => "#{@key_pair.name}.pem"
+  end
+
+  def create
+    if @compute_resource.recreate
+      process_success :success_msg => _('Successfully created'),
```
ohadlevy: should the message be re-created (to match the UI)?

app/helpers/key_pairs_helper.rb
```diff
@@ -0,0 +1,8 @@
+module KeyPairsHelper
+  def key_pair_status_icon(key_pair)
+    return icon_text('ok', '', {kind: 'pficon', class: 'center', title: _('Active key')}) if key_pair.active
+    return icon_text('ok', '', {kind: 'pficon', class: 'center warn',
+                                title: _('Key used with other compute resource')}) if key_pair.used_elsewhere
+    icon_text('circle-o', '', {kind: 'fa', class: 'center', title: _('Key not connected to compute resource')})
```
ohadlevy: this is any compute resource right? in that case might change the wording to any compute resource?

app/models/compute_resources/foreman/model/ec2.rb
```diff
@@ -1,14 +1,11 @@
 module Foreman::Model
   class EC2 < ComputeResource
-    has_one :key_pair, :foreign_key => :compute_resource_id, :dependent => :destroy
-
+    include KeyPairComputeResource
+    prepend KeyPairCapabilities
```
ohadlevy: does it work if the prepend is within KeyPairComputeResource mixin too? (if it does, maybe move it inside the mixin)

app/assets/stylesheets/patternfly_and_overrides.scss
shlomizadok: This is the right place to put is. I will change only to `pficon-ok.warn` and not to whole of `pficon-ok`

app/controllers/key_pairs_controller.rb
shlomizadok: Sometimes it is created (when no key pair), sometimes it is re-created (actually more common use)

app/models/key_pair.rb
```diff
@@ -3,6 +3,7 @@ class KeyPair < ActiveRecord::Base
   validates_lengths_from_database
   validates :name, :secret, :presence => true
   validates :compute_resource_id, :presence => true, :uniqueness => true
+  audited
```
ohadlevy: will this not show the private key in the audit log when its created??

app/views/audits/show.html.erb
```diff
@@ -51,6 +51,10 @@
         <% end %>
         </tbody>
       </table>
+      <% if @audit.comment %>
+        <%= alert(:class => 'alert-info', :header => '',
```
ohadlevy: should this always be an alert? mind adding a screenshot?

app/views/compute_resources/show.html.erb
```diff
@@ -12,6 +12,9 @@
   <% if @compute_resource.capabilities.include?(:image) %>
     <li><a href="#images" data-toggle="tab"><%= _("Images") %></a></li>
   <% end %>
+  <% if @compute_resource.capabilities.include?(:key_pair) %>
+    <li><a href="#key_pairs" data-toggle="tab"><%= _("Key pairs") %></a></li>
```
ohadlevy: I wonder if we should just call it with its real name.. SSH Keys (key pair is an internal implementation name?)

app/views/key_pairs/index.html.erb
```diff
@@ -0,0 +1,56 @@
...
+                                                 .merge(:auth_object => @compute_resource, :authorizer => authorizer, :method => 'post', :text => _("Recreate")))) %>
+          <% elsif !key_pair.used_elsewhere %>
+            <%= action_buttons(display_delete_if_authorized(hash_for_remove_compute_resource_key_pair_path(@compute_resource, key_pair.name)
+                                                            .merge(:auth_object => @compute_resource, :authorizer => authorizer, :method => 'delete'))) %>
+          <% end %>
+        </td>
+      </tr>
+    <% end %>
+    </tbody>
+  <% else %>
```
ohadlevy, line 37: nitpick: I would break it into two partials, one for list and another for blank state.

app/views/key_pairs/index.html.erb
```diff
@@ -0,0 +1,56 @@
...
+    <tbody>
+      <tr>
+        <td class="blank-slate-pf">
+            <div class="blank-slate-pf-icon">
+              <%= icon_text("key", "", :kind => "fa") %>
+            </div>
+            <h1><%= _('Key pairs') %></h1>
+            <p>
+              <%= _("When using SSH based provisioning, a special SSH key is created in order to automate the provisioning process.") %><br />
+              <%= _("Normally when setting up a Compute Resource, a key pair (Private and Public) is created for you automatically, however, if it was deleted externally this page you to recreate it") %>
```
ohadlevy: text is not correct English :) lowercase private and public and at the end of the line helps/allows you to...

app/views/audits/show.html.erb
shlomizadok: image

app/views/key_pairs/index.html.erb
shlomizadok: lol... I copied from your previous comment :smiling_imp:

test/controllers/key_pairs_controller_test.rb
```diff
@@ -0,0 +1,59 @@
...
+    get :index, {:compute_resource_id => compute_resource.to_param}, set_session_user
+    assert_response :not_found
+  end
+
+  test "should download pem file" do
+    key = FactoryGirl.create(:key_pair)
+    get :download, {:compute_resource_id => @compute_resource.to_param, :id => key.id}, set_session_user
+    assert_response :success
+    assert_equal(key.secret, @response.body)
+    downloaded = Tempfile.new('tmp_pem')
```
ohadlevy: why do you write it to a file? cant you check the response.body.size.zero?

app/models/key_pair.rb
shlomizadok: image

And

image
 
Seems ok, right?

test/controllers/key_pairs_controller_test.rb
shlomizadok: Trying to simulate the download of a file, and check the file's size

app/views/audits/show.html.erb
shlomizadok: I think as long it is `alert-info` it looks good

app/assets/stylesheets/patternfly_and_overrides.scss
ohadlevy: my main concern this effects all pficon-ok.warn across the app, not sure if its intended or not.

app/controllers/key_pairs_controller.rb
ohadlevy: i would go with the common case.... create implies a bug?

app/models/key_pair.rb
ohadlevy: no, make sure you don't audit the field, e.g.
```ruby
audited :except => [:secret]
```

shlomizadok: lol... silly me. changing `audited` on KeyPair.

app/views/audits/show.html.erb
ohadlevy: two concerns:
1. its inside the table, yet its not related to any of the table actions, probably needs a re-review of the entire show diff part, /cc @Rohoover 
2. the alert will be shown to any other audit that has a comment, not just key pairs. e.g. templates.

app/views/key_pairs/index.html.erb
ohadlevy: ask me again tomorrow, maybe it will change again ;)

test/controllers/key_pairs_controller_test.rb
ohadlevy: i read that, i just wonder why you need to actually save it to the slave disk..?

shlomizadok: It gets deleted afaik ([tempfile docs](http://ruby-doc.org/stdlib-1.9.3/libdoc/tempfile/rdoc/Tempfile.html#class-Tempfile-label-Explicit+close))

app/assets/stylesheets/patternfly_and_overrides.scss
shlomizadok: There are none at the moment. This override will help others who wish to use it in the future.

app/models/compute_resources/foreman/model/ec2.rb
shlomizadok: Yeah, good advise.

app/views/audits/show.html.erb
shlomizadok: I think it is cool that we add the comment (see the example for template):
image

As for the table layout: It is right below the table and it makes sense to me.

app/views/compute_resources/show.html.erb
```diff
@@ -50,6 +53,13 @@
       </div>
     </div>
   <% end %>
+  <% if @compute_resource.capabilities.include?(:key_pair) %>
+    <div id="key_pairs" class="tab-pane">
+      <div id="key_pairs_list" data-url=<%= compute_resource_key_pairs_path(@compute_resource) %>>
+        <%= spinner(_('Loading key pairs information ...')) %>
```
ohadlevy: loading ssh key information?

ohadlevy: or just spinner.

app/views/key_pairs/index.html.erb
```diff
@@ -0,0 +1,57 @@
...
+              <%= icon_text("key", "", :kind => "fa") %>
+            </div>
+            <h1><%= _('SSH keys') %></h1>
+            <p>
+              <%= _("When using SSH based provisioning, a special SSH key is created in order to automate the provisioning process.") %><br />
+              <%= _("Normally when setting up a Compute Resource, a key pair (private and public) is created for you automatically.") %><br />
+              <%= _("However, if it was deleted externally this page helps you to recreate it") %>
+             </p>
+            <div class="blank-slate-pf-main-action">
+              <%= link_to(_('New key pair'), compute_resource_key_pairs_path(@compute_resource), :method=> :post, :class => "btn-lg" ) %>
```
ohadlevy: ssh key, here and across other messages...

shlomizadok: Changed.

app/controllers/key_pairs_controller.rb
```diff
@@ -0,0 +1,56 @@
+class KeyPairsController < ApplicationController
+  before_action :find_compute_resource, :except => [:download]
+  before_action :find_resource, :only => [:download]
+
+  def index
+    @key_pairs = @compute_resource.get_compute_key_pairs
```
ohadlevy, line 8: missing exception handling for unable to fetch key pairs. can easily be simulated by using the wrong user/pass on your compute resource.

ohadlevy: spinner will just go on for ever..

ohadlevy:
@shlomizadok one thing I would like to see if you can add is a bit of improvement in the audit log, two things that i found anonying:
1. download is considered an update to the key, is there a way to simply write the user downloaded the key vs the user updated the key.
2. when looking at a key history, it reflect only the key itself, e.g. if i recreate it, its no longer listed. while thats understandable, i do expect to see that kind of break down in the compute resource history vs the key it self. this is probably a wider area for compute resource events (such as vm operations etc).

if there is no simple way to make audit as a "event log" I'll leave it as is(read another issue), but i would be happy if you could look into it briefly.

otherwise :+1: feature works well (minor issue with exception handling on failure - mentioned in another comment).

test/controllers/key_pairs_controller_test.rb
```diff
@@ -0,0 +1,56 @@
+require 'test_helper'
+
+class KeyPairsControllerTest < ActionController::TestCase
+  setup do
+    @compute_resource = FactoryGirl.create(:ec2_cr)
+  end
+
+  test "cr with key_pair should get index" do
+    get :index, {:compute_resource_id => @compute_resource.to_param}, set_session_user
```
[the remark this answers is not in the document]
shlomizadok: wat?

ares:
I guess we should disable cops that we can't fullfil on older ruby versions 
e.g. on ubuntu trusty

shlomizadok:
@ohadlevy @ares - fixed failing cops

.rubocop.yml
```diff
@@ -36,3 +36,7 @@ Metrics/ClassLength:
 Performance/FixedSize:
   Exclude:
     - 'test/**/*'
+
+# Don't enforce params attr on tests (won't work with set_session_user)
+Rails/HttpPositionalArguments:
+  Enabled: false
```
ohadlevy: maybe this should be added as a pr of it self, as i guess most current prs are hitting this? /cc @dLobatog / @tbrisker

app/controllers/application_controller.rb
```diff
@@ -414,6 +414,11 @@ def webpack_server
     @dev_server ||= "#{request.protocol}#{request.host}:#{port}"
   end
 
+  def compute_resource_error(listing, error)
```
ohadlevy: i dont think it make sense to create a compute resource specific error handling in the application controller....

shlomizadok: Where would you put this code? If I want to DRY this, application_controller seems right (I can move this to a concern though)

ohadlevy:
On Tue, Dec 20, 2016 at 10:23 AM, Shlomi Zadok <notifications@github.com>
wrote:

> *@shlomizadok* commented on this pull request.
> ------------------------------
>
> In app/controllers/application_controller.rb
> <https://github.com/theforeman/foreman/pull/4024>:
>
> > @@ -414,6 +414,11 @@ def webpack_server
>      @dev_server ||= "#{request.protocol}#{request.host}:#{port}"
>    end
>
> +  def compute_resource_error(listing, error)
>
> Where would you put this code? If I want to DRY this,
> application_controller seems right (I can move this to a concern though)
>

please do, its related to compute resource errors, not the entire
application

app/controllers/application_controller.rb
shlomizadok: Moved to a concern

.rubocop.yml
shlomizadok: Added as a separate PR @ https://github.com/theforeman/foreman/pull/4121

ohadlevy: @shlomizadok can you rebase please?

shlomizadok:
@ohadlevy - rebased, removed cop, moved to a concern :)

app/models/key_pair.rb
```diff
@@ -3,6 +3,7 @@ class KeyPair < ActiveRecord::Base
   validates_lengths_from_database
   validates :name, :secret, :presence => true
   validates :compute_resource_id, :presence => true, :uniqueness => true
+  audited :except => :secret
```
ohadlevy: I think you need to add ```:associated_with => :compute_resource```

shlomizadok: Why?

ohadlevy:
it links to the compute resource, so later on we could show the
associated(e.g. ssh key change) data in the compute resource audit log,
have a look at audited docs https://github.com/collectiveidea/audited

shlomizadok:
test failures unrelated :sweat_smile:

shlomizadok:
[test] just to be sure :)

app/views/audits/show.html.erb
ohadlevy: this breaks other comment listing, such as an audit of templates without a message:
template

app/views/audits/show.html.erb
```diff
@@ -51,6 +51,10 @@
         <% end %>
         </tbody>
       </table>
+      <% if @audit.comment %>
```
ohadlevy: most likely need to add .blank?

app/views/audits/show.html.erb
ohadlevy: please include a timestamp in this view, and maybe user email address or something additionally to identify the user? I wan it to look more like a log.

(audit has a reference user object if it still exists).

ohadlevy: also, one should not be allowed to close the warning.

shlomizadok: Done

app/views/audits/show.html.erb
shlomizadok: Done

app/views/audits/show.html.erb
```diff
@@ -51,6 +51,10 @@
         <% end %>
         </tbody>
       </table>
+      <% if @audit.comment && !@audit.comment.blank? %>
```
ohadlevy: you can simply do ```<%= unless @audit.comment.blank? %>```
nil.blank? => true

shlomizadok:
Foreman error unrelated

config/routes.rb
```diff
@@ -416,6 +416,12 @@
           put 'test_connection'
         end
         resources :images, :except => [:show]
+        resources :key_pairs, :only => [:index, :create] do
+          member do
+            get 'download'
+            delete 'remove'
```
ohadlevy: sorry for catching this pretty late, but this should simply default to rest destroy action

shlomizadok: Restified it

ohadlevy:
@shlomizadok note tests are failing, plus can you please add scope search definition to search based on the keypair and compute_resources type?

thanks

shlomizadok:
@ohadlevy - fixed failing tests. Failed test unrelated.

ohadlevy:
thanks @shlomizadok, finally merged.

Could you please open two follow up issues:
1. expose key_pair info via API responses (I assume both show/recreate etc) and possibility added info on the compute resource response (e.g. just the key name).
2. add audit search for compute resource type (similar to what you added to ssh_keys).

PR #4025, opened 2016-11-16
fixes #17335 - parse Solaris update as minor version

domcleal:
Also fixes idempotency of OSes with only a major version (as Solaris
was prior to this change), where `minor` was nil instead of "", causing
the OS to be recreated with a validation error.


dLobatog:
Thanks @domcleal !

PR #4026, opened 2016-11-16
fixes #17354 - test and improve idempotency of OS fact parser

domcleal:
Contains the commit from #4025 which fixes an issue when parsing OSes with only a major version.

### DIFF
--- a/app/models/operatingsystems/aix.rb
+++ b/app/models/operatingsystems/aix.rb
@@ -1,11 +1,6 @@
 class AIX < Operatingsystem
   PXEFILES = {:kernel => "powerpc", :initrd => "initrd"}
 
-  # Override the class representation, as this breaks many rails helpers
-  def class
-    Operatingsystem
-  end
-
   def pxe_type
     "nim"
   end
--- a/app/services/puppet_fact_parser.rb
+++ b/app/services/puppet_fact_parser.rb
@@ -2,7 +2,7 @@ class PuppetFactParser < FactParser
   attr_reader :facts
 
   def operatingsystem
-    orel = os_release
+    orel = os_release.dup
 
     if os_name == "Archlinux"
       # Archlinux is rolling release, so it has no release. We use 1.0 always
@@ -40,14 +40,20 @@ class PuppetFactParser < FactParser
     end
     if os.description.blank?
       if os_name == 'SLES'
-        os.description = os_name + ' ' + orel.gsub!('.', ' SP')
+        os.description = os_name + ' ' + orel.gsub('.', ' SP')
       elsif facts[:lsbdistdescription]
         family = os.deduce_family || 'Operatingsystem'
         os.description = family.constantize.shorten_description facts[:lsbdistdescription]
       end
     end
-    os.save!
-    os
+
+    if os.new_record?
+      os.save!
+      Operatingsystem.find_by_id(os.id) # complete reload to be an instance of the STI subclass
+    else
+      os.save!
+      os
+    end
   end
 
   def environment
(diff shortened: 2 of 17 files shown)


app/models/operatingsystems/aix.rb
```diff
@@ -1,9 +1,8 @@
 class AIX < Operatingsystem
   PXEFILES = {:kernel => "powerpc", :initrd => "initrd"}
 
-  # Override the class representation, as this breaks many rails helpers
-  def class
-    Operatingsystem
+  class << self
```
timogoebel: You could add this to the superclass to not repeat yourself:

``` ruby
def self.inherited(child)
  child.instance_eval do
    # rubocop:disable Rails/Delegate
    def model_name
      superclass.model_name
      end
    # rubocop:enable Rails/Delegate
  end
  super
end
```

domcleal: Thanks, I've used this and removed existing model_name impls.

app/services/puppet_fact_parser.rb
```diff
@@ -40,14 +40,20 @@ def operatingsystem
...
       elsif facts[:lsbdistdescription]
         family = os.deduce_family || 'Operatingsystem'
         os.description = family.constantize.shorten_description facts[:lsbdistdescription]
       end
     end
-    os.save!
-    os
+
+    if os.new_record?
+      os.save!
```
dLobatog, line 51: nitpick, why checking whether it's a new record? I suppose already existing Operatingsystem instances can be reloaded just the same by calling `.find` and you can avoid the entire if condition just with:

``` ruby
os.save!
OperatingSystem.find(os.id)
```

:question:

app/services/puppet_fact_parser.rb
```diff
@@ -40,14 +40,20 @@ def operatingsystem
     end
     if os.description.blank?
       if os_name == 'SLES'
-        os.description = os_name + ' ' + orel.gsub!('.', ' SP')
+        os.description = os_name + ' ' + orel.gsub('.', ' SP')
```
dLobatog, line 43: I'm not sure I understand why this change. It doesn't seem to change the behavior as we're not using `orel` anywhere after this anyway. gsub! (in place) should be slightly faster. If we're just going to make one substitution `orel['.'] = ' SP'` [should be even faster](https://github.com/JuanitoFatas/fast-ruby#stringsub-vs-stringgsub-vs-string-code) (petty, I know)

dLobatog (changes requested):
Just 2 minor comments on the parser, I tried making those changes and tests passed just fine. Thanks @domcleal

app/services/puppet_fact_parser.rb
domcleal: I was trying to avoid doing the extra .find call if the OS already existed. If it's a new record then its class will be Operatingsystem with `family` set to say, `Debian` so has to be reloaded to be the right class.

If the OS already existed and we're just updating it, the family will already be known, it'll be a `Debian` class and so saving it is enough - it doesn't need to reload from the DB, saving a call.

app/services/puppet_fact_parser.rb
domcleal: Yeah, I found this in the SLES parser test before adding `.dup` to `os_release`. They fix the same issue, but I left this here because it's more correct.

`#gsub!` returns `nil` if there are no changes made, so shouldn't be used in concatenation:

``` ruby
2.3.0 :001 > "1.2".gsub!('.', ' SP')
 => "1 SP2" 
2.3.0 :002 > "1.2".gsub!('.', ' SP').gsub!('.', ' SP')
 => nil 
```

What was happening in the tests was the same facts hash was parsed twice and using `gsub!` modified the original fact. When the test called the parser again, the gsub! returned nil and the test failed trying to concatenate String and NilClass.

dLobatog (approved):

dLobatog:
Thanks @domcleal !

dLobatog:
Thanks @domcleal !

PR #3670, opened 2016-07-22
Fixes #15779 - make background processing available

ares:
blocked by https://github.com/theforeman/foreman-tasks/pull/194 but I no longer want to wait with publishing the PR so discussion/reviewing can start

PR #4027, opened 2016-11-16
Fixes #17361 - ensure plugins migration paths are set

iNecas:


iNecas:
We've noticed this while working on https://github.com/theforeman/foreman/pull/3670

ares:
nasty issue :-) I think that's reasonable solution, I'll keep it open until Friday to see whether someone comes different better idea but you can consider this ACKed from me

ares:
Thanks @iNecas, merged as afd5e5122282a65e96e58e3532d3d683bdd6305b

PR #4029, opened 2016-11-17
Fixes #17377 - Update fog-core to 1.43.0

dLobatog:
This is a requirement for certain compute resources that use fog. In
particular
https://github.com/fog/fog-azure-rm/blob/master/fog-azure-rm.gemspec
needs 1.43 to work.


domcleal:
This might cause a problem with foreman-digitalocean, since fog-core 1.43.0 assumes the DO provider's being loaded from the fog-digitalocean gem (at `fog/digitalocean/compute`). In develop and in foreman-digitalocean master, they still use the main `fog` gem which contains the DO provider (at `fog/digitalocean/compute_v2`).

You might be able to fix this in foreman-digitalocean by adding a dependency on fog-digitalocean, but I'm unsure if there will be any issues arising from having the DO provider in _both_ the `fog` and `fog-digitalocean` gems present (e.g. classes modifying each other).

We can't yet remove the `fog` gem, it's still used by a few providers. If you're OK with potentially having both `fog` and `fog-digitalocean` loaded (after adding the dep to foreman-do) then this can be merged.

dLobatog:
Thanks for the warning, I didn't know about that. I'll check the problem with fog-digitalocean at some other point. My intent was to do this in order to enable https://github.com/fog/fog-azure-rm in fog-azure, but I was able to work around it, so there's no real need for this _right now_. When there is, I'll test DO too. Closing!

PR #4030, opened 2016-11-18
Fixes #16884 - enforce granular filters even during save

ares:
Seems to work surprisingly well, if this gets :+1: to get in, we should probably cover it in release notes. Note that I'm unable to exactly tell which attribute caused the authorization to fail so I add error message to base.

### DIFF
--- a/app/models/concerns/authorizable.rb
+++ b/app/models/concerns/authorizable.rb
@@ -1,11 +1,50 @@
 module Authorizable
   extend ActiveSupport::Concern
 
-  included do
-    def authorized?(permission)
-      return false if User.current.nil?
-      User.current.can?(permission, self)
+  def check_permissions_after_save
+    authorizer = Authorizer.new(User.current)
+    creation = self.id_changed?
+    name = permission_name(creation ? :create : :edit)
+
+    Foreman::Logging.logger('permissions').debug { "verifying the transaction by permission #{name} for class #{self.class}" }
+    unless authorizer.can?(name, self)
+      org_loc_string = Taxonomy.enabled_taxonomies.map { |tax| _(tax) }.join(' ' + _('or') + ' ')
+      errors.add :base, _("You don't have permission %{name} with attributes that you have specified or you don't have access to specified %{tax_string}") % { :name => name, :tax_string => org_loc_string }
+
+      # This is required in case the rollback happend, the instance must look like new record so that all url helpers work correctly. Rails don't rollback these attributes.
+      if creation
+        self.id = nil
+        @new_record = true
+      end
+
+      # we need to rollback orchestration tasks if this object orchestrates something
+      if self.class.included_modules.include?(Orchestration)
+        self.send :fail_queue, self.queue
+      end
+
+      raise ActiveRecord::Rollback
     end
+  end
+
+  def authorized?(permission)
+    return false if User.current.nil?
+    User.current.can?(permission, self)
+  end
+
+  def permission_name(action)
+    type = Permission.resource_name(self.class)
+    permissions = Permission.where(:resource_type => type).where(["#{Permission.table_name}.name LIKE ?", "#{action}_%"])
+
+    # some permissions are grouped for same resource, e.g. edit_comupute_resources and edit_compute_resources_vms, in such case we need to detect the right permission
+    if permissions.size > 1
+      permissions.detect { |p| p.name.end_with?(type.underscore.pluralize) }.try(:name)
+    else
+      permissions.first.try(:name)
+    end
+  end
+
+  included do
+    after_save :check_permissions_after_save
   end
 
   module ClassMethods
--- a/app/models/permission.rb
+++ b/app/models/permission.rb
@@ -25,4 +25,21 @@ class Permission < ApplicationRecord
   def self.reset_resources
     @all_resources = nil
   end
+
+  # converts klass to name used as resource_type in permissions table
+  def self.resource_name(klass)
+    return 'Operatingsystem' if klass <= Operatingsystem
+    return 'ComputeResource' if klass <= ComputeResource
+    return 'Subnet' if klass <= Subnet
+    return 'Parameter' if klass <= Parameter
+
+    case (name = klass.to_s)
+    when 'Audited::Audit'
+      'Audit'
+    when /\AHost::.*\Z/
+      'Host'
+    else
+      name
+    end
+  end
 end
--- a/app/services/authorizer.rb
+++ b/app/services/authorizer.rb
@@ -11,6 +11,7 @@ class Authorizer
   end
 
   def can?(permission, subject = nil, cache = true)
+    return false if user.nil?
     return true if user.admin?
 
     if subject.nil?
@@ -161,21 +162,8 @@ class Authorizer
     [organizations, locations, values]
   end
 
-  # sometimes we need exceptions however we don't want to just split namespaces
   def resource_name(klass)
-    return 'Operatingsystem' if klass <= Operatingsystem
-    return 'ComputeResource' if klass <= ComputeResource
-    return 'Subnet' if klass <= Subnet
-    return 'Parameter' if klass <= Parameter
-
-    case (name = klass.to_s)
-    when 'Audited::Audit'
-      'Audit'
-    when /\AHost::.*\Z/
-      'Host'
-    else
-      name
-    end
+    Permission.resource_name(klass)
   end
 
   def base_ids
--- a/test/controllers/notification_recipients_controller_test.rb
+++ b/test/controllers/notification_recipients_controller_test.rb
@@ -93,11 +93,20 @@ class NotificationRecipientsControllerTest < ActionController::TestCase
 
     test "notification when host has no owner" do
       host = FactoryGirl.create(:host, :managed)
-      User.current = nil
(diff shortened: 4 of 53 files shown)


app/models/permission.rb
```diff
@@ -23,4 +23,21 @@ def self.with_translations
   def self.reset_resources
     @all_resources = nil
   end
+
+  # converts klass to name used as resource_type in permissions table
+  def self.resource_name(klass)
+    return 'Operatingsystem' if klass <= Operatingsystem
+    return 'ComputeResource' if klass <= ComputeResource
+    return 'Parameter' if klass <= Parameter
```
ares: note for the reviewer: I already applied https://github.com/theforeman/foreman/pull/3984/files#diff-2a8c1ced7ef4dfe08f35154c5a3ef0f1R169 since this is moving to different file and wouldn't cause conflict

app/models/concerns/authorizable.rb
```diff
@@ -6,6 +6,33 @@ def authorized?(permission)
...
+    end
+
+    def check_permissions_after_save
+      authorizer = Authorizer.new(User.current)
+      creation = self.id_changed?
+      name = permission_name(creation ? :create : :update)
+
+      unless authorizer.can?(name, self)
+        errors.add :base, "You don't have permission #{name} with attributes that you have specified"
+        if creation
```
ares: note for the reviewer: This is required so the instance looks like new record and all url helpers works correctly. Rails wouldn't rollback these.

ares:
Tests fail  since there's no user set for permissions verification. Maybe we can skip this check in test environment.

ares:
before I start fixing tests somehow, I'd like to know if this is the right direction, any comments from @theforeman/core would be appreciated

ares:
Pinging @theforeman/core didn't work so I'm sorry but I'm pinging specific people. The question is, do you think this PR is a viable solution for limiting create and update actions by granular filters? It does the save in transaction, it verifies permissions using scoped search and rollbacks the transaction if user is not authrorized. @dLobatog @domcleal @iNecas @lzap @tbrisker @timogoebel @orrabin @ohadlevy

timogoebel:
I can't think of a better solution, so 👍 
However, I must admit, that I'm not too familiar with the permissions code.

tbrisker:
I think `after_save` is the correct place for this, although I'm surprised this actually works - since our filters are based on scoped search, it searches the database to check if the object matches the filter. Only concern on my part is that permission checking is quite slow and this might cause a significant performance problem when creating multiple objects at once, and possibly deadlocking the database in some cases (like we saw with the cached counters).

orrabin:
Edit: because it's update and create the permission can't be checked until we get to after_save because the object to find might not even exist yet.

I can't think of any other solution that will easily work for both api and ui so :+1: from me.

app/models/concerns/authorizable.rb
iNecas: I think it's good to put the comment not only to the PR, but actually into the code for future generations

iNecas:
I think it's a viable solution (don't see any concerns right now). About performance: since we provide permissions system, we can't sacrifice it for performance, at least not by default, and at places we opt it out, we should have a reason why. I guess we can figure out the best way to turn the check off once we hit the case where it's not a good idea doing this way

iNecas:
I don't see a reason (other than failling jenkins at the moment) to not pull this PR in. If there are some hidden issues, I guess there is not other way than merge it and address the issues when they appear.

ares:
Thanks Ivan, I started to fix tests last week. So far, more than 150 tests have been fixed but there are still more to be fixed. Setting WoC until I fix the rest.

ares:
Can't believe my eyes, Foreman tests green. I'll look at Katello tomorrow.

ares:
Katello tests should be green with https://github.com/theforeman/foreman-tasks/pull/261 and https://github.com/Katello/katello/pull/6842

It would be great to have this in 1.16, the change itself is quite simple, the biggest part are test fixes.

ares:
[test katello] [test upgrade]

ares:
[test katello]

ares:
it seems katello tests are unrelated, same issue in other PRs

app/services/authorizer.rb
```diff
@@ -11,6 +11,7 @@ def initialize(user, options = {})
   end
 
   def can?(permission, subject = nil, cache = true)
+    return false if user.nil?
```
dLobatog, line 14: Does this work when `SETTINGS[:login]` is set to false?

app/models/concerns/authorizable.rb
```diff
@@ -1,13 +1,52 @@
...
+    # some permissions are grouped for same resource, e.g. edit_comupute_resources and edit_compute_resources_vms, in such case we need to detect the right permission
+    if permissions.size > 1
+      permissions.detect { |p| p.name.end_with?(type.underscore.pluralize) }.try(:name)
+    else
+      permissions.first.try(:name)
     end
   end
 
+  included do
+    after_save :check_permissions_after_save
```
dLobatog, line 47: Does this work when `SETTINGS[:login]` is set to false? Maybe just an `:if => SETTINGS[:login]` is all we need to ensure it works in both scenarios

app/models/concerns/authorizable.rb
```diff
@@ -1,13 +1,52 @@
...
-      return false if User.current.nil?
-      User.current.can?(permission, self)
+  def check_permissions_after_save
+    authorizer = Authorizer.new(User.current)
+    creation = self.id_changed?
+    name = permission_name(creation ? :create : :edit)
+
+    Foreman::Logging.logger('permissions').debug { "verifying the transaction by permission #{name} for class #{self.class}" }
+    unless authorizer.can?(name, self)
+      org_loc_string = Taxonomy.enabled_taxonomies.map { |tax| _(tax) }.join(_(" or "))
```
dLobatog: nitpick, best to add the spaces in the string below rather than in the translated string

dLobatog (commented):
@ares Just had a few comments, overall 👍 if tests pass - if they pass but something in Foreman doesn't work, it means we should cover that. We still have a lot of time to harden 1.16 paths.

app/models/concerns/authorizable.rb
ares: I think it should since we [set User.current]((https://github.com/theforeman/foreman/blob/develop/app/controllers/concerns/foreman/controller/authentication.rb#L34)) to admin in such case (same elsewhere, when login is disabled, all should work in a same way)

app/models/concerns/authorizable.rb
ares: I hope that's what you meant

app/services/authorizer.rb
ares: user should be admin in this case, so not a `nil`

ares:
@iNecas @dLobatog thanks for taking a look, tests are now green including katello

iNecas (approved):

iNecas:
:+1: from me

dLobatog:
Thanks @ares !

test/controllers/notification_recipients_controller_test.rb
```diff
@@ -93,11 +93,20 @@ class NotificationRecipientsControllerTest < ActionController::TestCase
...
+      assert host.update_attribute(:owner_id, nil)
+      Host::Managed.where(:owner_id => nil).update_all(:owner_type => nil) # owner type must be set by hack because of sti.rb
+      host.reload
+      assert host.update_attribute(:build, false)
       assert_nil host.owner
-      assert host.built
-      get :index, { :format => 'json' }, set_session_user
+
+      setup_user 'edit', 'hosts'
+      host.stub :owner_suggestion, nil do
```
tbrisker, line 103: This line seems to break develop tests locally, not sure why it wasn't caught by jenkins. @ares can you take a look?

ares: do we have a trace or some link to failing jenkins job?

tbrisker: It seems to only fail locally:
```
1) Error:
NotificationRecipientsControllerTest::with seeded notification types#test_0003_notification when host has no owner:
NoMethodError: undefined method `stub' for #<Host::Managed:0x00000009106ec8>
    test/controllers/notification_recipients_controller_test.rb:103:in `block (2 levels) in <class:NotificationRecipientsControllerTest>'
```

timogoebel: Maybe it should be `host.stubs` instead of `host.stub`? I don't know, just a hunch.

ares: the stub method comes from minitest, perhaps something changed there and the stubbing framework was extracted in some newer version? I have minitest 5.10.3 here

tbrisker: I'm also on 5.10.3. if you run this test locally it passes?

ares: it worked when I was working on the test, I'll retest later to verify

ares: it failed only for single test runs, addressed by https://github.com/theforeman/foreman/pull/4705

PR #4031, opened 2016-11-19
Fixes #17403 - enable exporting of templates

ares:
this adds new action for all kinds of templates
export

the same can be achieved through V2 API.

### DIFF
--- a/app/controllers/templates_controller.rb
+++ b/app/controllers/templates_controller.rb
@@ -4,7 +4,7 @@ class TemplatesController < ApplicationController
   include Foreman::Controller::AutoCompleteSearch
 
   before_action :handle_template_upload, :only => [:create, :update]
-  before_action :find_resource, :only => [:edit, :update, :destroy, :clone_template, :lock, :unlock]
+  before_action :find_resource, :only => [:edit, :update, :destroy, :clone_template, :lock, :unlock, :export]
   before_action :load_history, :only => :edit
   before_action :type_name_plural, :type_name_singular, :resource_class
 
@@ -95,6 +95,10 @@ class TemplatesController < ApplicationController
     safe_render(@template)
   end
 
+  def export
+    send_data @template.to_erb, :type => 'text/plain', :disposition => 'attachment', :filename => @template.filename
+  end
+
   private
 
   def safe_render(template)
@@ -124,7 +128,7 @@ class TemplatesController < ApplicationController
     case params[:action]
       when 'lock', 'unlock'
         :lock
-      when 'clone_template', 'preview'
+      when 'clone_template', 'preview', 'export'
         :view
       else
         super
--- a/app/helpers/provisioning_templates_helper.rb
+++ b/app/helpers/provisioning_templates_helper.rb
@@ -13,7 +13,10 @@ module ProvisioningTemplatesHelper
   end
 
   def permitted_actions(template)
-    actions = [display_link_if_authorized(_('Clone'), template_hash_for_member(template, 'clone_template'))]
+    actions = [
+      display_link_if_authorized(_('Clone'), template_hash_for_member(template, 'clone_template')),
+      display_link_if_authorized(_('Export'), template_hash_for_member(template, 'export'), { :data => { :no_turbolink => true } })
+    ]
 
     if template.locked?
       confirm = [
--- a/app/models/ptable.rb
+++ b/app/models/ptable.rb
@@ -33,6 +33,8 @@ class Ptable < Template
 
   alias_attribute :layout, :template
 
+  attr_exportable :os_family
+
   # with proc support, default_scope can no longer be chained
   # include all default scoping here
   default_scope lambda {
--- a/app/models/template.rb
+++ b/app/models/template.rb
@@ -1,4 +1,6 @@
 class Template < ActiveRecord::Base
+  include Exportable
+
   validates_lengths_from_database
 
   validates :name, :presence => true
@@ -9,6 +11,8 @@ class Template < ActiveRecord::Base
   before_destroy :check_if_template_is_locked
 
   before_save :remove_trailing_chars
+
+  attr_exportable :name, :snippet
 
   class Jail < Safemode::Jail
     allow :name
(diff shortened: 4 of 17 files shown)


config/routes/api/v2.rb
```diff
@@ -40,6 +40,7 @@
         (resources :locations, :only => [:index, :show]) if SETTINGS[:locations_enabled]
         (resources :organizations, :only => [:index, :show]) if SETTINGS[:organizations_enabled]
         post :clone, :on => :member
+        get :export, :on => :member
```
shlomizadok: Maybe change to:
```ruby
member do
  post :clone
  get :export
end
```

test/models/template_test.rb
```diff
@@ -0,0 +1,115 @@
+require 'test_helper'
+
+class TemplateTest < ActiveSupport::TestCase
+  describe "generating metadata" do
+    before do
```
domcleal: Prefer `setup` for minitest's unit test style throughout.

test/models/template_test.rb
```diff
@@ -0,0 +1,115 @@
...
+  describe "#to_erb" do
+    before do
+      content = "<%#
+name: basic
+%>
+data"
+      @template = Template.new :name => 'basic', :template => content
+    end
+
+    it "generates fresh fresh metadata and replaces original ones" do
```
domcleal: Prefer `test` over `it` for consistency within this file and unit test style.

app/models/template.rb
```diff
@@ -31,6 +35,28 @@ def preview_host_collection
     Host.authorized(:view_hosts).order(:name).limit(100)
   end
 
+  def metadata
+    "<%#\n#{to_export(false).to_yaml.sub(/\A---$/, '').strip}\n%>\n\n"
```
domcleal: Is the second `\n` required here? When exporting a template with metadata, it causes an additional newline between the new metadata segment and whatever follows (which could be a problem in a template containing a shebang line.)

app/controllers/templates_controller.rb
```diff
@@ -95,6 +95,10 @@ def preview
     safe_render(@template)
   end
 
+  def export
+    send_data @template.to_erb, :type => 'text/plain', :disposition => 'attachment', :filename => @template.filename
```
domcleal, line 99: I'd suggest putting this in a concern to prevent duplication across all Template-subclass controllers.

test/controllers/api/v2/provisioning_templates_controller_test.rb
```diff
@@ -96,6 +96,11 @@ class Api::V2::ProvisioningTemplatesControllerTest < ActionController::TestCase
     assert_response :unprocessable_entity
   end
 
+  test 'export should export the erb of the template' do
+    get :export, { :id => templates(:pxekickstart).to_param }
+    assert_response :success
```
domcleal, line 101: Some form of the content and/or response type (that it's an attachment with a given filename) should be tested here too.

domcleal (changes requested):

test/models/template_test.rb
```diff
@@ -0,0 +1,115 @@
...
+      refute_includes without, 'name: basic'
+    end
+  end
+
+  describe "#filename" do
+    before do
+      @template = Template.new
+    end
+
+    test "filename ads erb suffix" do
```
domcleal: adds

app/controllers/templates_controller.rb
ares: sorry, I don't follow, this controller is base for all Template-subclass controllers

domcleal: Sorry, please disregard, I meant to put it against the API controllers. I'll log a different comment there though.

Edit: no, disregard completely, thanks.

app/models/template.rb
ares: Shebang should be on the first line then, same as we have `<?xml` and others, `to_erb` keeps the first line because of this. The reason was that I find it more readable if there an empty line to separate. But I removed the "\n", I agree that splitting first line and the rest is even worse and one day we could fix templates so metadata is always first.

ares:
thanks @shlomizadok and @domcleal, I fixed all your comments, could you please take another look?

app/models/ptable.rb
```diff
@@ -33,6 +33,8 @@ class Ptable < Template
 
   alias_attribute :layout, :template
 
+  attr_exportable :os_family
```
domcleal, line 36: `:snippet` should be added here as it's separate for Ptables, or perhaps in the base class.

For ProvisioningTemplates snippet replaces the kind - either the exported "kind" field should state "snippet" (like the metadata in current shipped templates) or simply add snippet: true by adding snippet to the base class.

test/controllers/provisioning_templates_controller_test.rb
```diff
@@ -55,6 +55,11 @@ class ProvisioningTemplatesControllerTest < ActionController::TestCase
     assert_template 'new'
   end
 
+  test "export" do
+    get :export, { :id => templates(:pxekickstart).to_param }, set_session_user
+    assert_response :success
```
domcleal, line 60: Please also add the response assertions in the API tests to both of these UI controller tests.

app/helpers/provisioning_templates_helper.rb
```diff
@@ -13,7 +13,10 @@ def template_kind(template)
   end
 
   def permitted_actions(template)
-    actions = [display_link_if_authorized(_('Clone'), template_hash_for_member(template, 'clone_template'))]
+    actions = [
+      display_link_if_authorized(_('Clone'), template_hash_for_member(template, 'clone_template')),
+      display_link_if_authorized(_('Export'), template_hash_for_member(template, 'export'))
```
domcleal: When clicking this export link, the loading spinner remains in the title bar of the application. (Tested in Chrome, the download completes in the background.)

domcleal (changes requested):

app/models/ptable.rb
ares: I decided to add :snippet to base class because of following reasons
* consistency with ptables
* job templates (rex plugin) consitency
* foreman_templates can simplify [the condition](https://github.com/theforeman/foreman_templates/blob/master/app/models/concerns/foreman_templates/provisioning_template_import.rb#L7)

one downside is that since it's boolean, it will be always present, mostly as false. We could probably move the scope search definition to base class too, but that's a different story.

ares:
@domcleal addressed again, please review changes

domcleal (approved):

domcleal:
Merged, thanks @ares.

PR #4032, opened 2016-11-20
Fixes #17279 - Changing os search in host to search on os_title

orrabin:


ohadlevy:
@orrabin how about adding a test?

orrabin:
@ohadlevy the search was not broken but it searched on a field that is not mandatory so for some operating systems would provide an empty search. I don't think there should be a test to force searching on title because if someone decides to change is back or switch to a different field I wouldn't want the test to break.

ohadlevy:
[test foreman]

PR #4033, opened 2016-11-20
Fixes #17394 - remove npm shrinkwrap file

tbrisker:
npm@4 broke the feature we were using to only pin some sub-dependencies
without pinning everything using shrinkwrap. Luckily, the package that
was breaking on node 0.10 is now fixed so we don't need this pin any
more. At least until the next sub-dependency decides to drop node 0.10
support.


dLobatog:
Thanks @tbrisker 😄

PR #4034, opened 2016-11-21
fixes #17407 - add a new hosts dashboard widget

timogoebel:
This commit adds a small dashboard that shows a list of the latest eight hosts that were created. Nothing too fancy.

### DIFF
--- a/app/views/dashboard/_new_hosts_widget.html.erb
+++ b/app/views/dashboard/_new_hosts_widget.html.erb
@@ -1,0 +1,11 @@
+<% hosts = @data.hosts.includes(:operatingsystem).preload(:owner).order(:created_at => :desc).limit(9) %>
+
+<h4 class="header">
+  <%= N_('New Hosts') %>
+</h4>
+
+<% if hosts.empty? %>
+  <p class="ca"><%= _("No hosts available.") %></p>
+<% else %>
+  <%= render :partial => 'new_hosts_widget_host_list', :locals => { :hosts => hosts } %>
+<% end %>
--- a/app/views/dashboard/_new_hosts_widget_host_list.html.erb
+++ b/app/views/dashboard/_new_hosts_widget_host_list.html.erb
@@ -1,0 +1,20 @@
+<table class="table table-striped table-fixed">
+  <thead>
+    <th width="30%"><%= _('Host') %></th>
+    <th><%= _('Operating system') %></th>
+    <th><%= _('Owner') %></th>
+    <th><%= _('Created') %></th>
+    <th><%= _('Installed') %></th>
+  </thead>
+  <tbody>
+    <% hosts.each do |host| %>
+      <tr>
+        <td class="ellipsis"><%= link_to host.name, host_path(host) %></td>
+        <td class="hidden-tablet hidden-xs ellipsis"><%= safe_join([icon(host.operatingsystem, :size => '16x16'), host.operatingsystem.to_label]) if host.operatingsystem.present? %></td>
+        <td class="hidden-tablet hidden-xs ellipsis"><%= host.owner %></td>
+        <td class="hidden-tablet hidden-xs ellipsis"><%= _('%s ago') % time_ago_in_words(host.created_at) %></td>
+        <td class="hidden-tablet hidden-xs ellipsis"><%= host.installed_at.present? ? _('%s ago') % time_ago_in_words(host.installed_at) : '-' %></td>
+      </tr>
+  <% end %>
+  </tbody>
+</table>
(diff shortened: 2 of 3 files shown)


timogoebel:
Tests failures look unrelated.

app/views/dashboard/_new_hosts_widget.html.erb
```diff
@@ -0,0 +1,11 @@
+<% hosts = Host.authorized(:view_hosts).includes(:owner, :operatingsystem).order(:created_at => :desc).limit(8) %>
```
ohadlevy: does it also make sense to do something based on installed_at attribute as well?

app/views/dashboard/_new_hosts_widget_host_list.html.erb
```diff
@@ -0,0 +1,18 @@
...
+  <thead>
+    <th><%= _('Host') %></th>
+    <th><%= _('Operating system') %></th>
+    <th><%= _('Owner') %></th>
+    <th><%= _('Created') %></th>
+  </thead>
+  <tbody>
+    <% hosts.each do |host| %>
+      <tr>
+        <td><%= link_to trunc_with_tooltip(h(host.name)), host_path(host) %></td>
```
ohadlevy: nitpick: h is not longer needed.

app/views/dashboard/_new_hosts_widget.html.erb
tbrisker: the dashboard data service already loads hosts respecting authorization and filters, you can use `@data.hosts.includes...`  instead and get the widget to respect the filter.

app/views/dashboard/_new_hosts_widget_host_list.html.erb
```diff
@@ -0,0 +1,18 @@
+<table class="table table-striped ellipsis">
```
tbrisker: `ellipsis` class should be on the `<td>`, not the table. the table should have `table-fixed` and the `th` elements should set their width.

app/views/dashboard/_new_hosts_widget_host_list.html.erb
tbrisker: `trunc_with_tooltip` is not needed if using `ellipsis` on the `td`.

timogoebel:
@ohadlevy, @tbrisker : Thanks for your comments. I added a "installed_at" column and fixed the ellipsis code.

dLobatog:
@timogoebel 👍 by me, however I'll wait until @tbrisker takes a final look since he's been the one reviewing this. Thank you 😄

app/views/dashboard/_new_hosts_widget.html.erb
```diff
@@ -0,0 +1,11 @@
+<% hosts = @data.hosts.includes(:operatingsystem).preload(:owner).order(:created_at => :desc).limit(8) %>
```
tbrisker: any reason not to use `.joins(:operatingsystem, :owner)` and getting the data in one call? also seems like there is room for 9 rows in the widget.

app/views/dashboard/_new_hosts_widget_host_list.html.erb
```diff
@@ -0,0 +1,20 @@
...
+    <th width="30%"><%= _('Host') %></th>
+    <th><%= _('Operating system') %></th>
+    <th><%= _('Owner') %></th>
+    <th><%= _('Created') %></th>
+    <th><%= _('Installed') %></th>
+  </thead>
+  <tbody>
+    <% hosts.each do |host| %>
+      <tr>
+        <td class="ellipsis"><%= link_to host.name, host_path(host) %></td>
```
tbrisker, line 12: Ellipsis on the dashboard seems broken but it is unrelated to this. Opened http://projects.theforeman.org/issues/17501 to handle that.

tbrisker (changes requested):
Mostly looks good, just a couple of inline comments.

app/views/dashboard/_new_hosts_widget.html.erb
timogoebel: Yes, rails joins seem to be broken for polymorphic associations. You can try that yourself, just use `includes(:operatingsystem, :owner)` and add a filter that doesn't match any host.

app/views/dashboard/_new_hosts_widget_host_list.html.erb
timogoebel: Thanks.

app/views/dashboard/_new_hosts_widget.html.erb
timogoebel: For the record, rails throws a `ActiveRecord::EagerLoadPolymorphicError: Cannot eagerly load the polymorphic association :owner` error.

timogoebel:
@tbrisker : Added one more row to the dashboard.

tbrisker (approved):

tbrisker:
good to merge once :green_apple:

PR #4037, opened 2016-11-22
Fixes #17602 - mention-bot won't ping inactive past contributers

tbrisker:


.mention-bot
```diff
@@ -0,0 +1,4 @@
+{
+  "requiredOrgs": ["theforeman"]
```
timogoebel: I think, this is missing a trailing comma to be valid json.

.mention-bot
```diff
@@ -0,0 +1,4 @@
...
+  "userBlackList": ["abenari", "unorthodoxgeek", "alongoldboim", "bgupta", "logicminds"]
```
timogoebel: Shouldn't we "remove" this guys from the organization if they're definitely inactive?

timogoebel (changes requested):
@tbrisker : Thanks. In addition to my inline comments, please file a redmine ticket for this.

.mention-bot
tbrisker: I think we should, at least from the committer group, but only @domcleal and @ohadlevy have the permissions to do that.

tbrisker:
@timogoebel added ticket and fixed comment.

dLobatog:
There's https://github.com/mention-bot/how-to-unsubscribe too..

tbrisker:
@dLobatog thanks, that is useful for users not seeing notifications they don't care about but it will still ping them instead of relevant contributes, plus this is not useful for users that may be using mentionbot in other projects.

timogoebel (approved):
@tbrisker : Sorry for the delay, I was off. Looks good to me, will merge in a sec.

PR #4038, opened 2016-11-22
refs #14688 - correct i18n extraction for all new power methods

domcleal:
- HOST_POWER being translated at the class level, not in action context
- messages with variables interpolated prior to i18n lookup
- timeout message being translated without prior extraction


ohadlevy:
thanks @domcleal

PR #4039, opened 2016-11-23
Fixes #17457 - multiple_checkboxes doesn't work with orgs/locs

dLobatog:
multiple_checkboxes calls ActiveModel::Naming on associations that could
be simply "Taxonomy". This results on a 500 error 'undefined method
`taxonomies' for #<Filter:0x005593624bdd18>' (or whatever other class
has the problem). Simply trying to edit a filter will throw this 500
error.

We've always had this problem , but since #16971 removed the check for
only using that method for '> 5' associations, you'd only see the error
if you had at least 6 organizations or locations (unlikely). Now it
became more apparent and it shows up every time.

The fix I've used is just to take advantage of the options hash to send
the association name directly, and if it's not set, then use
ActiveModel::Naming to get it (as usual).

### DIFF
--- a/app/helpers/form_helper.rb
+++ b/app/helpers/form_helper.rb
@@ -46,8 +46,9 @@ module FormHelper
   end
 
   def multiple_checkboxes(f, attr, klass, associations, options = {}, html_options = {})
-    associated_obj = klass.send(ActiveModel::Naming.plural(associations.new))
-    selected_ids = associated_obj.select("#{associations.new.class.table_name}.id").map(&:id)
+    association_name = attr || ActiveModel::Naming.plural(associations)
+    associated_obj = klass.send(association_name)
+    selected_ids = associated_obj.select("#{associations.table_name}.id").map(&:id)
     multiple_selects(f, attr, associations, selected_ids, options, html_options)
   end
 
--- a/app/views/filters/_form.html.erb
+++ b/app/views/filters/_form.html.erb
@@ -62,16 +62,13 @@
 
     <% if show_location_tab? %>
       <div class="tab-pane" id="locations">
-        <% collection = User.current.locations.empty? ? Location : User.current.locations %>
-        <%= multiple_checkboxes f, :locations, f.object, collection, {}, :label => _("Locations") %>
+        <%= multiple_checkboxes f, :locations, f.object, User.current.my_locations, :label => _("Locations") %>
       </div>
     <% end %>
 
     <% if show_organization_tab? %>
       <div class="tab-pane" id="organizations">
-        <% collection = User.current.organizations.empty? ? Organization : User.current.organizations %>
-        <%= multiple_checkboxes f, :organizations, f.object, collection, {},
-                                {:label => _("Organizations")} %>
+        <%= multiple_checkboxes f, :organizations, f.object, User.current.my_organizations, :label => _("Organizations") %>
       </div>
     <% end %>
   </div>
--- a/app/views/roles/_form.html.erb
+++ b/app/views/roles/_form.html.erb
@@ -23,13 +23,11 @@
       <% end %>
 
       <% if show_location_tab? %>
-        <% collection = User.current.locations.empty? ? Location : Location.my_locations %>
-        <%= multiple_checkboxes f, :locations, f.object, collection, {}, { :label => _("Locations") } %>
+        <%= multiple_checkboxes f, :locations, f.object, User.current.my_locations, { :label => _("Locations") } %>
       <% end %>
 
       <% if show_organization_tab? %>
-        <% collection = User.current.organizations.empty? ? Organization : Organization.my_organizations %>
-        <%= multiple_checkboxes f, :organizations, f.object, collection, {}, { :label => _("Organizations") } %>
+        <%= multiple_checkboxes f, :organizations, f.object, User.current.my_organizations, { :label => _("Organizations") } %>
       <% end %>
     </div>
 
--- a/app/views/subnets/_form.html.erb
+++ b/app/views/subnets/_form.html.erb
@@ -22,7 +22,7 @@
       <%= render 'fields', :f => f %>
     </div>
     <div class="tab-pane" id="domains">
-      <%= multiple_checkboxes f, :domain, f.object, Domain, {:help_inline => _("Domains in which this subnet is part"), :size => 'col-md-6', :prefix => f.object_name}, { :label => _("Domain")} %>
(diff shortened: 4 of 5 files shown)


tbrisker:
Does this mean the workaround in https://github.com/theforeman/foreman/pull/3779 is no longer needed?

dLobatog:
@tbrisker It's still needed, that's a different issue

app/helpers/form_helper.rb
```diff
@@ -46,7 +46,9 @@ def checkbox_f(f, attr, options = {}, checked_value = "1", unchecked_value = "0"
   end
 
   def multiple_checkboxes(f, attr, klass, associations, options = {}, html_options = {})
-    associated_obj = klass.send(ActiveModel::Naming.plural(associations.new))
+    association_name = options[:association_name] ||
```
tbrisker: We already sent this as `attr`, no reason to send again in the options.

app/helpers/form_helper.rb
```diff
@@ -46,7 +46,9 @@ def checkbox_f(f, attr, options = {}, checked_value = "1", unchecked_value = "0"
...
+                       ActiveModel::Naming.plural(associations.new)
```
tbrisker: `.new` here is not needed.

app/helpers/form_helper.rb
```diff
@@ -46,7 +46,9 @@ def checkbox_f(f, attr, options = {}, checked_value = "1", unchecked_value = "0"
...
+    associated_obj = klass.send(association_name)
     selected_ids = associated_obj.select("#{associations.new.class.table_name}.id").map(&:id)
```
tbrisker: `associations.new.class.table_name == associations.table_name` AFAICT

tbrisker (changes requested):
I could not reproduce the condition which causes this error, however if you manage to reproduce it there is much simpler way to fix this w/o requiring change to the funtion caller - see inline comments

tbrisker:
Also, if this is reproducible please add a test reproducing it so we don't break again.

ares:
I haven't done the review but it fixes the issue for me. I just went to role -> filters -> new filter and got the exception, after applying the patch, it works as expected.

app/helpers/form_helper.rb
```diff
@@ -46,7 +46,9 @@ def checkbox_f(f, attr, options = {}, checked_value = "1", unchecked_value = "0"
   end
 
   def multiple_checkboxes(f, attr, klass, associations, options = {}, html_options = {})
-    associated_obj = klass.send(ActiveModel::Naming.plural(associations.new))
```
dLobatog, line 49: The bug is basically `klass.send(ActiveModel::Naming.plural(associations.new))` when associations is `User.current.organizations` produces 'taxonomies'

dLobatog: Which doesn't work as `filter.taxonomies` is not a valid method

dLobatog:
Updated with tests @tbrisker , to reproduce I think just visiting the Filter page with an user with one Organization or Location will break it, as `klass.send(ActiveModel::Naming.plural(User.current.organizations))` becomes `klass.send(:taxonomies)`. Hopefully the test is clear enough. 

cc @ares thanks for taking a look too.

tbrisker:
This prevents the exception, however when going to edit a filter as a non-admin user the taxonomy selection fields are empty.

app/services/association_authorizer.rb
```diff
@@ -7,7 +7,7 @@ def self.authorized_associations(associations, klass_name = nil, should_raise_ex
       end
       klass_name ||= associations_klass
       permission = permission_name(klass_name, action, should_raise_exception)
-      associations.authorized(permission, associations_klass) if permission
+      associations.authorized(permission, klass_name) if permission
```
dLobatog: This is what was causing the issue where non admin users wouldn't see their locs/orgs. @tbrisker notice `klass_name` is passed from the patch #3779 but it wasn't being used (only for creating the permission name)

app/views/filters/_form.html.erb
```diff
@@ -62,16 +62,15 @@
 
     <% if show_location_tab? %>
       <div class="tab-pane" id="locations">
-        <% collection = User.current.locations.empty? ? Location : User.current.locations %>
-        <%= multiple_checkboxes f, :locations, f.object, collection, {}, :label => _("Locations") %>
+        <% collection = User.current.locations.empty? ? Location : User.current.my_locations %>
```
dLobatog: switched to `my_locations` which includes truly all user locations including children

app/views/filters/_form.html.erb
```diff
@@ -62,16 +62,15 @@
...
+        <%= multiple_checkboxes f, :locations, f.object, collection, :label => _("Locations") %>
       </div>
     <% end %>
 
     <% if show_organization_tab? %>
       <div class="tab-pane" id="organizations">
-        <% collection = User.current.organizations.empty? ? Organization : User.current.organizations %>
-        <%= multiple_checkboxes f, :organizations, f.object, collection, {},
-                                {:label => _("Organizations")} %>
+        <% collection = User.current.organizations.empty? ? Organization : User.current.my_organizations %>
```
dLobatog: switched to my_organizations too. this was another bug...

dLobatog:
Set to WoC as I broke something

dLobatog:
Tests back green again, and I am able to see my organizations as a non-admin user, can you check back @tbrisker ?

app/views/filters/_form.html.erb
tbrisker: This allows a non-admin user with no locations assigned to see all locations. This should just be `User.current.my_locations` always.

app/views/filters/_form.html.erb
tbrisker: ditto

app/views/roles/_form.html.erb
```diff
@@ -24,12 +24,12 @@
 
       <% if show_location_tab? %>
         <% collection = User.current.locations.empty? ? Location : Location.my_locations %>
```
tbrisker: same comment applied to this form.

tbrisker (changes requested):
Mostly works, just a minor comment about a certain use case.

dLobatog:
Moved everything to `User.current.my_xxxxx` , thanks for the catch @tbrisker

dLobatog:
@tbrisker anything else I should change here?

tbrisker (approved):

tbrisker:
Thanks @dLobatog !

PR #4042, opened 2016-11-23
Fixes #17343 - set deep munge config off

kgaikwad:
deep_munge was introduced as a solution to keep Rails secure by default which results in
'empty array becomes nil in params'. Thats why, set deep_munge config off in application.rb.
Also, added changes which will cast param argument to string while calling find_by_{string_type_attr} method on object.

### DIFF
--- a/app/controllers/concerns/api/v2/lookup_keys_common_controller.rb
+++ b/app/controllers/concerns/api/v2/lookup_keys_common_controller.rb
@@ -49,7 +49,7 @@ module Api::V2::LookupKeysCommonController
     id = params.keys.include?('smart_variable_id') ? params['smart_variable_id'] : params['id']
     @smart_variable   = VariableLookupKey.authorized(:view_external_variables).smart_variables.find_by_id(id.to_i) if id.to_i > 0
     @smart_variable ||= (puppet_cond = { :puppetclass_id => @puppetclass.id } if @puppetclass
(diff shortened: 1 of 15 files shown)


dLobatog:
[test foreman]
[test katello]
[test update]

domcleal:
@theforeman-bot ok to test

dLobatog:
@kgaikwad Mind to add a couple of tests for the API that send `[]` to clear organizations or locations? That way when we turn this on again in Rails 5 we know it's still working. Other than that 👍 good job! I'd merge after that if tests pass.

For the onlookers, the `.to_s` are basically guardrails in case someone happens to send numeral types via the API to string values. Check the issue in redmine or http://guides.rubyonrails.org/security.html#unsafe-query-generation to learn more about why `[]` parameters were broken.

kgaikwad:
@dLobatog , thank you for review. I will add couple of tests as you mentioned.

kgaikwad:
@dLobatog , added only two tests i.e for Organisation & Location. Please let me know if any other changes required.

dLobatog:
@kgaikwad Thanks - that's what we were looking for, merging... 

PS: the duplication we have on the location/organization tests is quite revolting in case you want to fix it on another PR

kgaikwad:
@dLobatog , thank you for reviewing and merging the pull request. 
I also felt the same while writing the test for location because both tests are quite similar and having the same purpose. 
If any one of them failed for empty array then second also definitely going to fail for same reason.

I will remove one of them in my next PR.

PR #4044, opened 2016-11-24
Fixes #17468 - build pxe returns error when no tftp in proxy

amirfefer:
`build_pxe_default` in `ProvisioningTemplate`  returns status code 200 when no TFTP
has found in proxy, should return  422, and therefore an error message should render in UI.

### DIFF
--- a/test/controllers/provisioning_templates_controller_test.rb
+++ b/test/controllers/provisioning_templates_controller_test.rb
@@ -132,6 +132,12 @@ class ProvisioningTemplatesControllerTest < ActionController::TestCase
       assert_redirected_to provisioning_templates_path
     end
 
+    test "build menu should return with error code if no TFTP defined" do
+      SmartProxy.stubs(:with_features).with('TFTP').returns([])
+      get :build_pxe_default, {}, set_session_user
+      assert flash[:error].present?
+    end
+
     test "pxe menu's labels should be sorted" do
       t1 = TemplateCombination.new :hostgroup => hostgroups(:db), :environment => environments(:production)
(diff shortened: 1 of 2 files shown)


test/controllers/provisioning_templates_controller_test.rb
```diff
@@ -132,6 +132,12 @@ class ProvisioningTemplatesControllerTest < ActionController::TestCase
       assert_redirected_to provisioning_templates_path
     end
 
+    test "build menu should return with error code if no TFTP defined" do
+      SmartProxy.stubs(:with_features).with('TFTP').returns('')
```
domcleal: `with_features` is a scope, wouldn't it return an array?

test/controllers/provisioning_templates_controller_test.rb
```diff
@@ -132,6 +132,12 @@ class ProvisioningTemplatesControllerTest < ActionController::TestCase
...
+      get :build_pxe_default, {}, set_session_user
+      assert_response 422
```
domcleal: This action always redirects, it doesn't return a 4xx status code to the browser.

domcleal (changes requested):
The issue and commit message are unclear, I think you're talking about changing the alert type rather than the response code - and only when there no smart proxies, not on general failure.

amirfefer:
@domcleal Thanks for your review, I've used `flash[:error]` instead status code.

domcleal (approved):

domcleal:
Merged, thanks @amirfefer.

PR #4045, opened 2016-11-24
Fixes #17487 - support sessions for api calls

tstrachota:
- authenticated api calls save user to session and set flag api_authenticated_session
- sessions with such flag allow posting requests without CSRF token
- api sessions exipre the same way as UI sessions
- api sessions don't store any additional data to keep the requests stateless

This way the standard UI requests as well as API requests authenticated with session created from UI remain protected against CSRF. At the same time applications using API (such as hammer) can benefit from using session authentication and avoid the need of storing two tokens (CSRF and _session_id).

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -371,7 +371,7 @@ module Api
     end
 
     def protect_api_from_forgery?
-      session[:user].present?
+      session[:user].present? && !session[:api_authenticated_session]
     end
 
     def parameter_filter_context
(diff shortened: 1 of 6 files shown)


domcleal (changes requested):
Thanks, that works nicely when retaining the session ID cookie. Could you please rebase to ensure the test suite is passing?

tstrachota:
Rebased

domcleal:
Sorry, please rebase again to ensure ee7d140 is included - the test suite isn't fully running currently.

tstrachota:
Rebased

domcleal (approved):

domcleal:
Merged, thank you for the nice addition @tstrachota.

PR #4046, opened 2016-11-24
fixes #17474 Postgres restore task now cleans the db prior to import.

ohadlevy:


dLobatog:
When I wrote the task, I did not run foreman installer, I was under the assumption that one would restore the db manually and not rely on foreman-installer to create it first. This will fix it - so 👍 

It's dangerous in case you have a existing db and you wipe it by accident with this, so I'll add a confirmation step on a separate PR.

Thanks @ohadlevy

lzap:
Today I spent some time with incorrect restore, I highly suggest to add `-e` (exit on error) there as well.

ohadlevy:
Send a new pr?

PR #4047, opened 2016-11-24
Fixes #17439 - Moved facet base to a concern

ShimShtein:
If the plugin that wants to create a new facet doesn't want to take the inheritance slot (for example if it already has some base class), it can `include` the new concern.


ShimShtein:
Rebased

dLobatog:
Thanks @ShimShtein !

PR #3551, opened 2016-05-25
fixes #15176 - match fog's pin of fog-google

domcleal:

PR #4048, opened 2016-11-24
Fixes #3551 - Show unavailable Puppet Classes in Host Group edit

bastilian:
Puppet Classes can be assigned to a Host Group even if they are
not in the set environment. They will persist through out changes
of the environment, but were previously hidden from the UI.

Unavailable Puppet Classes are now shown in the UI as unavailable
and a warning is triggered to inform the user about them.

**Host Group edit form / Puppet Classes with changed environment:**
screen shot 2016-11-24 at 16 08 21

**Warning:**
screen shot 2016-11-24 at 16 08 10

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -359,6 +359,7 @@ function update_puppetclasses(element) {
       $('#puppet_klasses').html(request);
       reload_puppetclass_params();
       tfm.tools.activateTooltips();
+      tfm.hostgroups.checkForUnavailablePuppetclasses();
     },
     complete: function() {
       reloadOnAjaxComplete(element);
--- a/app/assets/stylesheets/puppetclasses.scss
+++ b/app/assets/stylesheets/puppetclasses.scss
@@ -32,8 +32,12 @@
   .matcher .matcher-group {
     display: block;
   }
-  
+
   .matcher_value {
     width: 60%;
   }
+}
+
+.selected_puppetclass.unavailable a {
+  color: $color-pf-black-400;
 }
--- a/app/models/concerns/host_common.rb
+++ b/app/models/concerns/host_common.rb
@@ -203,16 +203,14 @@ module HostCommon
     end
   end
 
+  # Returns Puppetclasses of a Host or Hostgroup
+  #
+  # It does not include Puppetclasses of it's ConfigGroupClasses
+  #
   def individual_puppetclasses
     ids = host_class_ids - cg_class_ids
     return puppetclasses if ids.blank? && new_record?
-
-    conditions = {:id => ids}
-    if environment
-      environment.puppetclasses.where(conditions)
-    else
-      Puppetclass.where(conditions)
-    end
+    Puppetclass.includes(:environments).where(id: ids)
   end
 
   def available_puppetclasses
--- a/app/views/puppetclasses/_selectedClasses.html.erb
+++ b/app/views/puppetclasses/_selectedClasses.html.erb
@@ -1,4 +1,4 @@
-<%= content_tag_for :li, klass, :selected, :class => cycle('even', 'odd') do %>
+<%= content_tag_for :li, klass, :selected, :class => [cycle('even', 'odd'), ('unavailable' unless klass.environments.include?(@environment))] do %>
   <%= link_to_remove_puppetclass(klass, host) %>
   <%= hidden_field_tag "#{type}[puppetclass_ids][]", klass.id %>
 <% end %>
--- a/test/controllers/hostgroups_controller_test.rb
+++ b/test/controllers/hostgroups_controller_test.rb
@@ -141,16 +141,38 @@ class HostgroupsControllerTest < ActionController::TestCase
     assert_template :partial => "common/os_selection/_architecture"
   end
 
-  test "should return the selected puppet classes on environment change" do
-    env = FactoryGirl.create(:environment)
-    klass = FactoryGirl.create(:puppetclass)
-    hg = FactoryGirl.create(:hostgroup, :environment => env)
-    assert_equal 0, hg.puppetclasses.length
-    post :environment_selected, { :id => hg.id,
-                                  :hostgroup => { :name => hg.name,
-                                                  :puppetclass_ids => [klass.id],
-                                                  :environment_id => "" }}, set_session_user
-    assert_equal(1, (assigns(:hostgroup).puppetclasses.length))
+  describe '#environment_selected' do
+    setup do
+      @environment = FactoryGirl.create(:environment)
+      @puppetclass = FactoryGirl.create(:puppetclass)
+      @hostgroup = FactoryGirl.create(:hostgroup, :environment => @environment)
+      @params = {
+        id: @hostgroup.id,
+        hostgroup: {
+          name: @hostgroup.name,
+          environment_id: "",
+          puppetclass_ids: [@puppetclass.id]
+        }
+      }
+    end
+
+    test "should return the selected puppet classes on environment change" do
+      assert_equal 0, @hostgroup.puppetclasses.length
+
+      post :environment_selected, @params, set_session_user
+      assert_equal(1, (assigns(:hostgroup).puppetclasses.length))
+      assert_include assigns(:hostgroup).puppetclasses, @puppetclass
+    end
+
+    context 'no environment_id param is set' do
+      test 'it will take the hostgroup params environment_id' do
+        other_environment = FactoryGirl.create(:environment)
+        @params[:hostgroup][:environment_id] = other_environment.id
+
+        post :environment_selected, @params, set_session_user
+        assert_equal assigns(:environment), other_environment
+      end
+    end
   end
 
   test 'user with view_params rights should see parameters in a hostgroup' do
--- a/webpack/assets/javascripts/foreman_hostgroups.js
+++ b/webpack/assets/javascripts/foreman_hostgroups.js
@@ -1,0 +1,21 @@
+import $ from 'jquery';
+
+export function checkForUnavailablePuppetclasses() {
+  let unavailableClasses = $('#puppet_klasses #selected_classes .unavailable');
+  let puppetKlassesTab = $('#puppet_klasses');
+  let tab = puppetKlassesTab.closest('form').find('.nav-tabs a[href="#puppet_klasses"]');
+  let warningMessage = __('Some Puppet Classes are unavailable in the selected environment');
+  let warning = `<div class="alert alert-warning" id="puppetclasses_unavailable_warning">
+      <span class="pficon pficon-warning-triangle-o"></span>
+      ${warningMessage}
+    </span>`;
+
+  if (unavailableClasses.size() > 0) {
+    tab.prepend('<span class="pficon pficon-warning-triangle-o"></span> ');
+    puppetKlassesTab.prepend(warning);
+  } else {
+    puppetKlassesTab.find('#puppetclasses_unavailable_warning').remove();
+    tab.find('.pficon-warning-triangle-o').remove();
+  }
+}
+
--- a/app/assets/javascripts/host_edit.js
+++ b/app/assets/javascripts/host_edit.js
@@ -496,6 +496,8 @@ function load_with_placeholder(target, url, data){
 
 function onHostEditLoad(){
   update_interface_table();
+  tfm.hostgroups.checkForUnavailablePuppetclasses();
+
   $("#host-conflicts-modal").modal({show: "true", backdrop: "static"});
    $('#host-conflicts-modal').click(function(){
      $('#host-conflicts-modal').modal('hide');
--- a/app/controllers/hostgroups_controller.rb
+++ b/app/controllers/hostgroups_controller.rb
@@ -82,7 +82,8 @@ class HostgroupsController < ApplicationController
   end
 
   def environment_selected
-    return not_found unless (@environment = Environment.find(params[:environment_id])) if params[:environment_id].to_i > 0
+    env_id = params[:environment_id] || params[:hostgroup][:environment_id]
+    return not_found unless (@environment = Environment.find(env_id)) if env_id.to_i > 0
 
     @hostgroup ||= Hostgroup.new
     @hostgroup.environment = @environment if @environment
--- a/app/controllers/hosts_controller.rb
+++ b/app/controllers/hosts_controller.rb
@@ -165,8 +165,11 @@ class HostsController < ApplicationController
   end
 
   def hostgroup_or_environment_selected
+    @environment = Environment.find(params['host']['environment_id'])
+    @hostgroup = Hostgroup.find(params['host']['hostgroup_id'])
+
     Taxonomy.as_taxonomy @organization, @location do
-      if params['host']['environment_id'].present? || params['host']['hostgroup_id'].present?
+      if @environment || @hostgroup
         render :partial => 'puppetclasses/class_selection', :locals => {:obj => (refresh_host)}
       else
         logger.info "environment_id or hostgroup_id is required to render puppetclasses"
--- a/app/views/puppetclasses/_class_selection.html.erb
+++ b/app/views/puppetclasses/_class_selection.html.erb
@@ -14,7 +14,7 @@
     <ul id="selected_classes">
       <% if authorized_for(:controller => :host_editing, :action => :edit_classes) %>
         <%= render :partial => "puppetclasses/selectedClasses",
-          :collection => obj.individual_puppetclasses ,:as => :klass, :locals => { :type => obj_type(obj), :host => obj } %>
+          :collection => obj.individual_puppetclasses,:as => :klass, :locals => { :type => obj_type(obj), :host => obj } %>
         <% unless controller_name == 'config_groups' %>
             <%= render :partial => "puppetclasses/classes_in_groups",
           :collection => obj.classes_in_groups ,:as => :klass, :locals => { :type => obj_type(obj), :obj => obj } %>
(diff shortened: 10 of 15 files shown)


dLobatog:
@theforeman-bot ok to test

dLobatog:
@domcleal @ohadlevy Could you add @bastilian to @theforeman org please?

app/models/concerns/host_common.rb
```diff
@@ -201,16 +201,25 @@ def classes_in_groups
...
   end
 
-  def individual_puppetclasses
+  # Returns PuppetClasses of a Host or Hostgroup
+  #
+  # Params:
+  #  * +with_environments+: {Boolean} if possible environments should be included
+  #  * +in_environments+: {Boolean} to consider only classes in environments
+  #
+  def individual_puppetclasses(with_environments: false, in_environment: true)
```
dLobatog: I'm not sure I understand why this method has arguments if it's only called once (in the view), where 'with_environments' is true and 'in_environment' is false. I'd suggest to either:

* remove the arguments completely (adds complexity to the method that I don't think we're using anywhere)
* make the arguments default to what we're using

test/controllers/hostgroups_controller_test.rb
```diff
@@ -141,16 +141,37 @@ def setup_user(operation, type = 'hostgroups')
...
+          puppetclass_ids: [@puppetclass.id]
+        }
+      }
+    end
+
+    test "should return the selected puppet classes on environment change" do
+      assert_equal 0, @hostgroup.puppetclasses.length
+
+      post :environment_selected, @params, set_session_user
+      assert_equal(1, (assigns(:hostgroup).puppetclasses.length))
```
dLobatog, line 163: Nitpick, you probably want to check `assert_include @puppetclass, assigns(:hostgroup).puppetclasses` to ensure it's the right class

dLobatog (changes requested):
@bastilian Thank you, this works pretty well from what I've seen. 

Once you save the host group, when you edit it again, the Puppet classes are colored in red, but it doesn't tell you why. You only see the 'unavailable' warning message when you change the environment, after that it's hard to know.

Can you add an `alert(:class => 'alert-warning'` to display this information? (or any equivalent way to show the classes are indeed unavailable) The rest of the PR is pretty much ready!

app/assets/javascripts/application.js
```diff
@@ -351,6 +351,12 @@ $(function() {
   });
 });
 
+function check_for_unavailable_puppetclasses() {
+  if ($('#puppet_klasses #selected_classes .unavailable').size() > 0) {
+    notify('<span>' + __('Some Puppet Classes are unavailable in the selected environment') + '</span>', 'warning')
```
dLobatog: I think it's 'Puppet class' (lowercase class) - we're certainly not being very consistent about this around the app but a quick search revealed it's only 'Puppet Class' on a couple of titles, not texts

app/models/concerns/host_common.rb
bastilian: Ah, you are right! I overlooked that it is just used once in the whole app.

I'll rewrite it to not need arguments.

app/assets/javascripts/application.js
bastilian: I've taken the tab title, that capitalises both, considering that is what the user need to look for when the alert pops up.

bastilian: BUT. I do think that "Puppet class" would be the correct way to write it, as "Puppet" is a "brand" and "class" is just another word.

ohadlevy:
@bastilian why do you show the error as a toast notification? shouldn't it be within (or next) to the puppet class section on the form?
also, does red (in your image) means you can't remove it, or disabled? I was unsure what it means from looking at your screenshot. if its unavialavble, shouldn't it be disabled? /cc @rohoover

Rohoover:
@ohadlevy 
I interpreted the red text as being part of the form error. Perhaps I'm mistaken.

bastilian:
@ohadlevy  & @Rohoover yeah, it is not obvious why they are red, @dLobatog pointed that out as well.

I'm gonna add a note on the Puppet classes tab elaborating on why they are red and that they are going to be wiped if submitted.

ohadlevy:
can you explain it here as well? maybe @Rohoover could suggest an
alternative approach? thanks!

bastilian:
The case here is that the user has changed the environment of the host group. The environment that has been selected environment does not include the puppet classes, that have been previously configured for the host group with an environment that had these classes.

Marking these red may not be appropriate, now that I think about it, as the are actually not wiped, but are just "ineffective" in the set environment not including them. 

In case the environment changes to one that includes them they will be considered again.

Let me know if you have more questions to find a good way to communicate this in the UI, @Rohoover.

Rohoover:
@bastilian 

A few thoughts:

Placing an inline notification at the top (yellow) would help give the user more insight. (http://www.patternfly.org/pattern-library/communication/inline-notifications/#/api)
After you explanation - I agree, red may not be accurate. Perhaps grey them out?
Lastly, in your example no included classes were applicable. What remedies does the user have other than to switch environments?

bastilian:
@Rohoover Switching the environment is the only way to have them active again.

Another action the user could take is to remove the unavailable classes to avoid mistakenly activate them by updating the environment, later on. A link offering that in the message explaining the state would make sense I think.

test/controllers/hostgroups_controller_test.rb
dLobatog: fixed

app/assets/stylesheets/puppetclasses.scss
```diff
@@ -32,8 +32,12 @@
...
   }
-  
+
   .matcher_value {
     width: 60%;
   }
 }
+
+.selected_puppetclass.unavailable a {
+  color: #cc0000
```
dLobatog: http://www.patternfly.org/styles/color-palette/#_ I believe you can use one of these

dLobatog:
Set to WoC @bastilian if you can find an way to show the 'resolution' (maybe just link to the primary tab and highlight the Environment field) with js is good - @Rohoover is that what you're looking for?

dLobatog (approved):

Rohoover:
@bastilian @dLobatog 

I think offering the user the feedback in the notification will go a long way. If it's possible to link to the tab within that notification then that could also be helpful.

app/assets/javascripts/application.js
```diff
@@ -338,6 +338,26 @@ $(function() {
   });
 });
 
+function check_for_unavailable_puppetclasses() {
+  var unavailable_classes = $('#puppet_klasses #selected_classes .unavailable');
```
ohadlevy: tbh: I think this needs to be rewritten in webpack, this way you can add tests to the function...

app/assets/stylesheets/puppetclasses.scss
```diff
@@ -32,8 +32,12 @@
...
   }
-  
+
   .matcher_value {
     width: 60%;
   }
 }
+
+.selected_puppetclass.unavailable a {
+  color: #d1d1d1;
```
ohadlevy: is this a patternfly color? maybe add its name / reference to color pallet.

app/models/puppetclass.rb
```diff
@@ -41,7 +41,10 @@ class Puppetclass < ActiveRecord::Base
   scoped_search :in => :class_params, :on => :key, :complete_value => :true, :only_explicit => true
 
   scope :not_in_any_environment, -> { includes(:environment_classes).where(:environment_classes => {:environment_id => nil}) }
-
+  scope :with_possible_environments, lambda {
+    joins('LEFT OUTER JOIN environment_classes ON puppetclasses.id = environment_classes.puppetclass_id
+           LEFT OUTER JOIN environments ON environment_classes.environment_id = environments.id')
```
ohadlevy: why left outer join? wont you get duplicates this way? maybe I'm missing something

app/assets/stylesheets/puppetclasses.scss
dLobatog: https://github.com/theforeman/foreman/pull/4048#discussion_r91099151 - should be a patternfly color

Rohoover: pf-black-300 =  #d1d1d1

ohadlevy: can't we simply access it via js/css directly? would make it both readable and if pf colors change we will have less of an headache :)

app/assets/javascripts/application.js
bastilian: Yes, I am also in favour of rewriting it under webpack, but wouldn't it be better to to have a component for the edit form in general first and then move over components like in this case for editing Puppet classes?

To have this tested now we can use an integration test, which can be removed (or not) once it is rewritten with tests under webpack.

bastilian:
@Rohoover I've changed this to have the warning now next to the environment select and is updated when it changes:

screen shot 2016-12-22 at 10 53 52

It didn't feel right to have a notification every time this state is detected, I think warning this way is much more effective.

In the Puppet Classes tab the unavailable ones are now only grey out (using the patternfly `$grey-lighter` color):

screen shot 2016-12-22 at 10 20 34

If we have more cases like this, where things are unavailable/inactive, it might be a good idea to have this in `colors.scss` (or `status-colors.scss`).

@ohadlevy you were right, the outer join was not the right thing to use here. The `individual_puppetclasses` method now uses Rails' includes, as it should have in the first place and the scope with join is removed.

app/assets/javascripts/application.js
dLobatog: I have to agree with @ohadlevy here, the function is relatively complicated so if we can get unit tests for them that's great. It doesn't have to be a React component, just moving the function there would make sense IMO so you can use Jest to test it. Like this one: https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/foreman_users.js

dLobatog (changes requested):
@bastilian This looks basically ready except for the JS function missing unit tests. I haven't tested it much yet, but please move the function to `webpack/`, write a couple of tests (for unavailable_classes.size greater than 0 and equal to 0) and ping me so we can get this merged soon.

Thanks!

bastilian:
@dLobatog & @ohadlevy I was reluctant before, because I thought it would be beyond the scope of the PR, if moved to webpack and written as a component. I'm gonna move the function and from now on only add JS under `webpack`. :)

Rohoover:
@bastilian My only concern is that the two pieces of messaging below it for Puppet Master and Puppet CA are persistent. The warning for environment may be less identifiable because of the persistent messaging. A possible solution would be to place the persistent messaging in inline help:
http://www.patternfly.org/pattern-library/forms-and-controls/field-level-help/#_

As a note: I'm gong to be addressing with the patternfly team if that icon is correct... there are currently two info icons, I'd like to consolidate them.

bastilian:
@Rohoover To be consistent with other forms, where it is next to the label as an (i)-icon tooltip, it does make sense to use the field help for Puppet Master and Puppet CA.

Let me know once you heard back from the patternfly team.

app/assets/javascripts/application.js
```diff
@@ -359,6 +359,7 @@ function update_puppetclasses(element) {
       $('#puppet_klasses').html(request);
       reload_puppetclass_params();
       tfm.tools.activateTooltips();
+      tfm.hostgroups.check_for_unavailable_puppetclasses();
```
dLobatog: `checkForUnavailablePuppetclasses` typo

app/assets/stylesheets/puppetclasses.scss
```diff
@@ -32,8 +32,12 @@
...
   }
-  
+
   .matcher_value {
     width: 60%;
   }
 }
+
+.selected_puppetclass.unavailable a {
+  color: $gray-lighter;
```
dLobatog: Use something from http://www.patternfly.org/styles/color-palette/#_ , the current one is barely visible

`pf-black-400 #bbbbbb` looks nice IMO for disabled elements

dLobatog (changes requested):
@bastilian I've been looking at it and the functionality is good, it detects the envs just fine (after fixing the typo in application.js). My only comments are:

* The color is barely distinguishable from white, it might be my background, but I think using some color from the patternfly pallette would make it more visible, see the comment inline.
* The warning isn't really noticeable on the "Host" tab, plus as @Rohoover it's confusing to have both persistent and ephemeral information there. Personally I didn't even notice the information when I was testing it, and it wasn't until I came back to the PR that I saw it. What I meant by adding `pf-warning` is that you can put it in the "Puppet classes" tab, on top of everything as @ohadlevy mentioned. That way it's immediately obvious why these are grey, while the current way forces you to go back. 
* It doesn't seem to work in the Host form, which is the same.
* The warning doesn't look good when Katello is installed, see the screenshot.

Long comment, but the fixes needed are quite minimal, thanks @bastilian

dLobatog:
The warning with Katello installed:
screenshot from 2017-01-02 11-29-26

bastilian:
@dLobatog I've addressed the minor changes and move the warning to the Puppet Classes tab now:

screen shot 2017-01-02 at 13 11 05

One thing that would be good though is to somehow let the user know that "something happened" in the Puppet Classes tab when the environment was changed or do you think it is enough for them to see it only in the tab?

Rohoover:
@bastilian 

The correct icon should be pficon_info, not the blue info icon. (For the field level help). 

Also - I think currently the tab turns red when there is an error, not sure if the same holds true for a warning. That said, being that the content has changed in the tab, it should be made known to the user. Though this is not currently defined, it may be good to have something more or other than just a color change - as I have talked to a user who is color-blind, and color only indicators were useless for him.

bastilian:
@Rohoover How about adding a warning sign to the tab, instead of it turning red? Like this:

screen shot 2017-01-04 at 08 03 32

Rohoover:
@bastilian +1

bastilian:
@Rohoover It's now implemented and the screens look as follows when an environment is selected with classes unavailable:

screen shot 2017-01-04 at 19 48 36
screen shot 2017-01-04 at 19 48 43

dLobatog (approved):

dLobatog:
@bastilian Awesome - if tests pass, I'll merge (and rebase first, as I solved a small conflict to let tests run).

Great job at figuring out how to show warnings like this with @Rohoover , to be fair we basically relied on making the fields red any time anything wrong happens. As you said that's not going to work for color-blind users. Also, a warning when changing something in the form doesn't have to look the same as an error when you submit it. 

Is there any recommendation on the Patternfly guidelines for forms, or can we make this (warning icon on the tab) a Patternfly recommendation?

webpack/assets/javascripts/foreman_hostgroups.js
```diff
@@ -0,0 +1,21 @@
...
+  let warning = `<div class="alert alert-warning" id="puppetclasses_unavaliable_warning">
+      <span class="pficon pficon-warning-triangle-o"></span>
+      ${warningMessage}
+    </span>`;
+
+  if (unavailableClasses.size() > 0) {
+    tab.prepend('<span class="pficon pficon-warning-triangle-o"></span> ');
+    puppetKlassesTab.prepend(warning);
+  } else {
+    puppetKlassesTab.find('#puppetclasses_unavaliable_warning').remove();
```
dLobatog: typo (here and above), `unavailable`

webpack/assets/javascripts/foreman_hostgroups.js
```diff
@@ -0,0 +1,21 @@
...
+  let unavailableClasses = $('#puppet_klasses #selected_classes .unavailable');
+  let puppetKlassesTab = $('#puppet_klasses');
+  let tab = puppetKlassesTab.closest('form').find('.nav-tabs a[href="#puppet_klasses"]');
+  let warningMessage = __('Some Puppet Classes are unavailable in the selected environment');
+  let warning = `<div class="alert alert-warning" id="puppetclasses_unavaliable_warning">
+      <span class="pficon pficon-warning-triangle-o"></span>
+      ${warningMessage}
+    </span>`;
+
+  if (unavailableClasses.size() > 0) {
```
dLobatog, line 13: Seems like `unavailableClasses.size() > 0` is true even after you change back from the environment that has the classes, so every time I change environments, a new pficon warning is added

dLobatog (changes requested):
@bastilian Found a problem when switching back to the original environment (where the classes are available), it added another warning icon. Every time I change environments now it adds a new icon basically. I'll try to figure out what's going on

app/views/puppetclasses/_selectedClasses.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= content_tag_for :li, klass, :selected, :class => cycle('even', 'odd') do %>
+<%= content_tag_for :li, klass, :selected, :class => [cycle('even', 'odd'), ('unavailable' unless klass.environments.include?(@environment))] do %>
```
dLobatog, line 1: Found the issue. `@environment` is only available on the *first* load of the page. On subsequent loads, it doesn't exist as the host controller didn't set it.

I fixed it locally by adding two lines to `hostgroup_or_environment_selected`, so it's the following:

```ruby

  def hostgroup_or_environment_selected
    Taxonomy.as_taxonomy @organization, @location do
      if params['host']['environment_id'].present? || params['host']['hostgroup_id'].present?
        @environment = Environment.find(params['host']['environment_id'])
        @hostgroup = Hostgroup.find(params['host']['hostgroup_id'])
        render :partial => 'puppetclasses/class_selection', :locals => {:obj => (refresh_host)}
      else
        logger.info "environment_id or hostgroup_id is required to render puppetclasses"
      end
    end
  end
```

I think that's all that's required for it to work.

dLobatog (changes requested):
See https://github.com/theforeman/foreman/pull/4048/files#r95572403 for the change needed for this to work again

dLobatog (approved):

dLobatog:
Thanks @bastilian @ohadlevy  & @Rohoover !

PR #4049, opened 2016-11-24
fixes #11127 - show error when rebuild reboot fails

timogoebel:


dLobatog:
@timogoebel Thank you !

PR #4050, opened 2016-11-24
fixes #17476 - bmc power actions have correct status

timogoebel:


test/unit/proxy_api/bmc_test.rb
```diff
@@ -217,4 +217,40 @@ def setup
...
+      {
+        'action' => 'cycle',
+        'result' => true
+      }
+    ))
+    assert_equal true, @testbmc.power(data)
+  end
+
+  test "#power should return true when result was ok" do
+    action = 'cycle'
```
dLobatog: This would be more DRY (and easier to read IMO) if you could extract the first part of this to a method, something like `stub_bmc_response(result)` and call that everywhere. 

The fact all tests start with '#power' suggests you probably want to group them in a `context '#power'` block

dLobatog:
@timogoebel Thanks - one minor comment about tests and this should be ready. Keep in mind IPMI responses can vary depending on the vendor and freeipmi/ipmitool, so we should be basically accepting 'anything' that is returned by the proxy.

dLobatog (changes requested):

timogoebel:
@dLobatog : Thanks, updated.

dLobatog (approved):

dLobatog:
Thanks @timogoebel - waiting for tests now...

timogoebel:
@dLobatog : Katello tests failed, but look unrelated. If you want, you can restart them of course.

dLobatog:
Thanks @timogoebel !

PR #4052, opened 2016-11-25
Fixes #13775 - merge only on same class collections

amitkarsale:
`merging` is an intersection between two collection objects, it makes valid sense if the both collections are of same class otherwise it will always return an empty array set.

### DIFF
--- a/test/controllers/api/v2/hosts_controller_test.rb
+++ b/test/controllers/api/v2/hosts_controller_test.rb
@@ -45,6 +45,13 @@ class Api::V2::HostsControllerTest < ActionController::TestCase
     )
   end
 
+  def basic_attrs_with_hg
+    hostgroup_attr = {
+      :hostgroup_id => Hostgroup.first.id
+    }
+    basic_attrs.merge(hostgroup_attr)
+  end
+
   def nics_attrs
     [{
       :primary => true,
@@ -256,6 +263,12 @@ class Api::V2::HostsControllerTest < ActionController::TestCase
 
   test "should update host" do
     put :update, { :id => @host.to_param, :host => valid_attrs }
+    assert_response :success
+  end
+
+  test "should update hostgroup_id of host" do
+    @host = FactoryGirl.create(:host, basic_attrs_with_hg)
+    put :update, { :id => @host.to_param, :hostgroup_id => Hostgroup.last.id }
     assert_response :success
   end
 
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -61,7 +61,7 @@ module Api
       association = resource_class.reflect_on_all_associations.find {|assoc| assoc.plural_name == parent_name.pluralize}
       #if couldn't find an association by name, try to find one by class
       association ||= resource_class.reflect_on_all_associations.find {|assoc| assoc.class_name == parent_name.camelize}
-      result_scope = resource_class.joins(association.name).merge(scope)
+      result_scope = resource_class_join(association, scope)
       # Check that the scope resolves before return
       result_scope if result_scope.to_a
     rescue ActiveRecord::ConfigurationError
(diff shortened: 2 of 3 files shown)


amitkarsale:
[test]

dLobatog:
Notice tests are failing - http://ci.theforeman.org/job/test_develop_pr_core/11222/

@amitkarsale Can you write a test so we know for sure we will not have this regression again? Thanks for the patch!

amitkarsale:
yes @dLobatog  looking for the same. will make the necessary change and merge. Thanks. !!

test/controllers/api/v2/hosts_controller_test.rb
```diff
@@ -259,6 +266,12 @@ def expect_attribute_modifier(modifier_class, args)
     assert_response :success
   end
 
+  test "should update hostgroup_id of host" do
```
dLobatog, line 269: Hard to believe such a simple thing failed, I had to run this manually first on 'develop' to check it. Wow!

dLobatog:
@amitkarsale Thanks for the test! There are 3 failures http://ci.theforeman.org/job/test_develop_pr_core/11291/#showFailuresLink yet to be fixed, so I'm setting this to Waiting on Contributor until then. It'd be ready to merge after that.

test/controllers/api/v2/hosts_controller_test.rb
amitkarsale: @dLobatog sure, please point me if am messing something somewhere, the other 36 tests mentioned, are they failing because of this change? am unsure about it.

amitkarsale:
@dLobatog I've tried running test on develop branch and on my branch but there smells nothing fishy with the code I've written. Can you verify it on your end also please.

test/controllers/api/v2/hosts_controller_test.rb
dLobatog: It's really the same 3 tests failing in the 12 different ruby/database scenarios

dLobatog (changes requested):
The OsDefaultTemplates one is indicating something in the code is wrong. When I do 'Operatingsystem.find(params[:operatingsystem_id]).os_default_templates.count` I get 6 templates. However the way `resource_base` looks in this PR, I get 10 (some of them are not related to the OS defined by `params[:operatingsystem_id]`- which shows it's returning the wrong result, thankfully the test caught it)

I'm not sure why you say merging two scopes from different classes is wrong. `resource_class_join.merge(scope).to_sql` between `resource_class_join (OsDefaultTemplates)` and `scope (Operatingsystem)` is definitely valid and desirable. It allows us to get precisely the join on `operatingsystem.id` and `os_default_templates.operatingsystem_id`. 

```sql 
"SELECT \"os_default_templates\".* FROM \"os_default_templates\" INNER JOIN \"operatingsystems\" ON \"operatingsystems\".\"id\" = \"os_default_templates\".\"operatingsystem_id\" WHERE (\"operatingsystems\".\"id\" = 1073012828 OR \"operatingsystems\".\"title\" = '1073012828-RHEL 6-1')  ORDER BY \"operatingsystems\".\"title\" ASC, CASE WHEN \"operatingsystems\".\"title\" = '1073012828-RHEL 6-1' THEN 1 ELSE 0 END"
```

@amitkarsale Hope that's enough to keep you moving forward! I'd appreciate if you could look into why tests failed next time instead of saying 'my code doesn't smell fishy' 😄 . I mostly used [pry-byebug](https://github.com/deivid-rodriguez/pry-byebug) and just `.to_sql` to see what was going on, to be honest a good deal of debugging in Foreman comes down to those 2 tools! 😆  No offense intended, just want us to save us time from commenting back and forth and hopefully help you sharpen your analytic skills. Please keep the contributions coming!

dLobatog (approved):

dLobatog:
Thanks @amitkarsale !

PR #4053, opened 2016-11-25
fixes #17484 - remove reported_at sequence from report factory

domcleal:
Prevents reports getting ever more out of sync from Time.now.

### DIFF
--- a/test/factories/host_related.rb
+++ b/test/factories/host_related.rb
@@ -168,8 +168,8 @@ FactoryGirl.define do
         report_count 5
       end
       after(:create) do |host,evaluator|
-        evaluator.report_count.times do
-          report = FactoryGirl.create(:report, :host => host)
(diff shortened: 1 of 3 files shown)


domcleal:
[test foreman]
[test upgrade]

domcleal:
@theforeman-bot ok to test

shlomizadok:
++ LGTM

domcleal:
I made a further change to the _host_ factory that creates many reports to change the reported_at value to different values. The precision of the timestamp field is low by default, and there's a unique index on it so the value needs to be changed when adding multiple reports.

Most uses of the factory for single reports should be fine, but any user creating more than one report (like the host factory) will need to supply unique values for `reported_at`.

dLobatog:
Thanks @domcleal & @shlomizadok !

PR #4054, opened 2016-11-25
Fixes #17482 - make strings translatable

ares:
I can't combine the string into one as asked in the issue because the environments part is generated dynamically, the result can be following

```
Update environment one (5 classes), environment two (3 classes)
```

I added it as translation parameter at least in case order needs to be changed, hopefully that helps. I also updated button labels and plural form.

### DIFF
--- a/app/services/actions/foreman/puppet_class/import.rb
+++ b/app/services/actions/foreman/puppet_class/import.rb
@@ -15,9 +15,9 @@ module Actions
           return nil if input[:changed].nil?
 
           humanized_output = []
-          humanized_output << _('Add') + ' ' + format_env_and_classes_input(input[:changed][:new]) if input[:changed][:new].present?
-          humanized_output << _('Remove') + ' ' + format_env_and_classes_input(input[:changed][:obsolete]) if input[:changed][:obsolete].present?
-          humanized_output << _('Update') + ' ' + format_env_and_classes_input(input[:changed][:updated]) if input[:changed][:updated].present?
+          humanized_output << _('Add %s') % format_env_and_classes_input(input[:changed][:new]) if input[:changed][:new].present?
+          humanized_output << _('Remove %s') % format_env_and_classes_input(input[:changed][:obsolete]) if input[:changed][:obsolete].present?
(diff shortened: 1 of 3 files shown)


app/services/actions/foreman/puppet_class/import.rb
```diff
@@ -38,11 +38,12 @@ def self.cleanup_after
...
             classes = JSON.parse(classes)
-            no_classes_info = classes.include?('_destroy_')
-            result += " (#{classes.size} " + _('classes') + ")" unless no_classes_info
-            result
+            if classes.include?('_destroy_')
+              _('environment %s') % environment
+            else
+              n_('environment %{environment} (%{count} class)', 'environment %{environment} (%{count} classes)', classes.size) % { :environment => environment, :count => classes.size }
+            end
           end.join(', ')
```
domcleal, line 47: use `.to_sentence`

app/services/actions/foreman/puppet_class/import.rb
```diff
@@ -15,9 +15,9 @@ def humanized_output
           return nil if input[:changed].nil?
 
           humanized_output = []
-          humanized_output << _('Add') + ' ' + format_env_and_classes_input(input[:changed][:new]) if input[:changed][:new].present?
-          humanized_output << _('Remove') + ' ' + format_env_and_classes_input(input[:changed][:obsolete]) if input[:changed][:obsolete].present?
-          humanized_output << _('Update') + ' ' + format_env_and_classes_input(input[:changed][:updated]) if input[:changed][:updated].present?
+          humanized_output << _('Add %s') % format_env_and_classes_input(input[:changed][:new]) if input[:changed][:new].present?
```
domcleal, line 18: The verb should still be part of the string _inside_ the method as it's part of the same sentence - it isn't a separate sentence or fragment, i.e. inside `format_env_and_classes_input`:

    _("Add environment %s")

and append versions without the verb:

    _("environment %s")

unless you rewrite it entirely into separate fragments or sentences, which it isn't right now.

domcleal (changes requested):

ares:
Closing, the PR adding these files was reverted.

PR #4056, opened 2016-11-25
fixes #17485 - provision tftp for bond children

timogoebel:


lzap:
Crisp and clear, thanks. If you can, drop a line into our docs about this.

PR #1556, opened 2014-07-03
fixes #5414: LDAP + TLS is not LDAP + TLS but LDAP + SSL

orrabin:

PR #4058, opened 2016-11-27
Fixes #1556 #11615 - Remove puppet rdoc support

tbrisker:
Puppet rdoc support has been broken for a long time. Removing it from
Foreman.


dLobatog:
👍 , I'm OK with removing this as not only it's unmaintained but I've rarely seen anyone asking for this. I tried to make a few small changes to print puppetlabs/apache docs but I didn't succeed, no idea about what would we need to make it work again (+ to make the proxies send puppet docs to Foreman...)

I'll leave this open for a few hours today in case anyone else wants to comment to keep this open, otherwise I'll merge then. If anyone wants to reimplement this based on the old implementation, cherry-picking this commit shouldn't be hard.

ares:
:+1: for removing

dLobatog:
Thanks @tbrisker @ares !

PR #4060, opened 2016-11-27
Fixes #16422 - Foreman API throws ruby exception with ovirt network API

jkroepke:
See http://projects.theforeman.org/issues/16422#note-6

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -321,7 +321,7 @@ module Api
         parent_class
       end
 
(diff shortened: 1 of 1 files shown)


mmoll:
ok to test

timogoebel:
[test foreman]

dLobatog:
Passing a 'nil' resource_class to the `scope_for` method will always result in a 500 as you cannot call `.where` on nil - merging 😄 Thanks @jkroepke & @mmoll @timogoebel

PR #3679, opened 2016-07-27
Fixes #12634 - New HW Model flag pxe_loader

lzap:
Implementation of

https://github.com/theforeman/rfcs/blob/master/text/0001-PXE-Booting-UEFI.md

In short, new host/hostgroup flag PXE loader is added to specify TFTP
DHCP filename and Foreman now orchestrates _all_ PXE-capable templates
deployment (e.g. PXELinux, Grub1, Grub2 if associated to an OS).

Test this patch against develop smart proxy as it requires https://github.com/theforeman/smart-proxy/pull/443 (merged).

There are few test-related TODOs and the patch is missing:
- [x] Add the sane default to all hosts, pxe_loader should not be empty (migration)
- [x] Validate the pxe_loader field of a host
- [x] Add new templates - https://github.com/theforeman/community-templates/pull/291
- [x] Build PXE Default button deploys all templates on all proxies
- [x] Editing PXE loader on existing host should trigger DHCP orchestration (w/o conflicts)
- [x] PXE Loader field works fine with hostgroup
- [x] Ability to set blank PXE Loader field (for iPXE provisioning)
- [ ] Test host/hostgroup provisioning rendering
- [x] UI integration test
- [x] API

I will update the list as we start with review.

PR #4063, opened 2016-11-28
Fixes #17503 - Only trigger PXELoader suggestion when needed

tbrisker:
Previously every host instantiation triggered a PXELoader suggestion,
which led to up to 3 extra queries per host loaded. This changes so that
the suggestion is only applied when a host changes it's OS.

### DIFF
--- a/app/controllers/api/v2/hosts_controller.rb
+++ b/app/controllers/api/v2/hosts_controller.rb
@@ -112,6 +112,7 @@ module Api
         @host = Host.new(host_attributes(host_params))
         @host.managed = true if (params[:host] && params[:host][:managed].nil?)
         apply_compute_profile(@host)
+        @host.suggest_default_pxe_loader if params[:host] && params[:host][:pxe_loader].nil?
 
         forward_request_url
         process_response @host.save
--- a/app/controllers/api/v1/hostgroups_controller.rb
+++ b/app/controllers/api/v1/hostgroups_controller.rb
@@ -41,6 +41,8 @@ module Api
 
       def create
         @hostgroup = Hostgroup.new(hostgroup_params)
+        @hostgroup.suggest_default_pxe_loader if params[:hostgroup] && params[:hostgroup][:pxe_loader].nil?
+
         process_response @hostgroup.save
       end
 
--- a/app/controllers/api/v1/hosts_controller.rb
+++ b/app/controllers/api/v1/hosts_controller.rb
@@ -71,6 +71,7 @@ module Api
       def create
         @host = Host.new(host_params)
         @host.managed = true if (params[:host] && params[:host][:managed].nil?)
+        @host.suggest_default_pxe_loader if params[:host] && params[:host][:pxe_loader].nil?
         forward_request_url
         process_response @host.save
       end
--- a/app/controllers/api/v2/hostgroups_controller.rb
+++ b/app/controllers/api/v2/hostgroups_controller.rb
@@ -53,6 +53,8 @@ module Api
 
       def create
         @hostgroup = Hostgroup.new(hostgroup_params)
+        @hostgroup.suggest_default_pxe_loader if params[:hostgroup] && params[:hostgroup][:pxe_loader].nil?
+
         process_response @hostgroup.save
       end
 
--- a/app/controllers/concerns/foreman/controller/host_details.rb
+++ b/app/controllers/concerns/foreman/controller/host_details.rb
@@ -9,7 +9,9 @@ module Foreman::Controller::HostDetails
   end
 
   def os_selected
-    assign_parameter "operatingsystem", "common/os_selection/"
+    assign_parameter "operatingsystem", "common/os_selection/" do |item|
+      item.suggest_default_pxe_loader
+    end
   end
 
   def medium_selected
@@ -47,6 +49,7 @@ module Foreman::Controller::HostDetails
     Taxonomy.as_taxonomy @organization, @location do
       item = instance_variable_get("@#{controller_name.singularize}") || controller_name.classify.constantize.new(item_params)
       instance_variable_set("@#{name}", item.send(name.to_sym))
+      yield item if block_given?
       render :partial => root + name, :locals => { :item => item }
     end
   end
--- a/app/controllers/hosts_controller.rb
+++ b/app/controllers/hosts_controller.rb
@@ -363,6 +363,7 @@ class HostsController < ApplicationController
   def toggle_manage
     if @host.toggle! :managed
       if @host.managed
+        @host.suggest_default_pxe_loader
         msg = _("Foreman now manages the build cycle for %s") % (@host.name)
       else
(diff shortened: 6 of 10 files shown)


timogoebel (changes requested):
This breaks creating Hosts via API calls. I think, doing this in the model is the right approach. Why put logic on the controller level?

lzap:
@tbrisker sorry I misled you when I said this was only for WebUI. We used this actually to keep API V2 compatibility - when the flag is not provided, the code provides "sane default" which works great so users do not need to change their scripts.

On the other hand, I don't believe this belongs to the model. This should be a controller concern or helper rather than this.

tbrisker:
I've added the suggestion on host[group] create via api, I tend to agree this should be done explicitly in the controller when needed rather then implicitly by callback in the model, but I'm open to change it if there is a good reason.

timogoebel:
> I tend to agree this should be done explicitly in the controller when needed rather then implicitly by callback in the model

I think, storing it in the model has the advantage, that it just works. Everywhere. Basically it shouldn't make any difference, when the code is run. I would just expect model behavior (read: the intelligence) to be in the model rather the controller. It's not the controllers place to change model behavior. It's job is solely to pass data from the view to the model (and enforce filters). Moving this to the controller(s) just duplicates code. More code, more bugs.

@tbrisker : What do you want to prevent with this patch? The code only runs, when a model doesn't have a pxe loader. Imho it shouldn't matter, whether it's done in the controller or the model.

Does this work with APIv1?

tbrisker:
@timogoebel according to my talk with @lzap yesterday (and his comment above), this code was introduced to handle 2 cases: 

1. Automatically choosing the best default PXELoader during host creation when an OS is selected, so the drop-down picks a sane option by default.
2. Preventing API scripts from breaking when not passing the new `pxeloader` parameter by choosing a sane default.

The current implementation handles this by running the suggest_default_pxe_loader every time a `Host` object is instantiated - not just new hosts, but also every time a host is loaded from the db for example. When this happens, an additional 0-3 queries are executed for each host loaded - so for example in the host list view we could be seeing up to 60 extra queries that serve no purpose. 
The other option to fix this would be to have the callback execute conditionally, however I don't think these cases are common enough to justify running this by callback (which also adds complexity as this isn't done explicitly - it took me a while just to figure out where these extra queries are coming from).
As for APIv1 I'm not sure, perhaps @lzap will know?

BTW, the way I found this was when I noticed a bunch of extra queries when testing your new hosts dashboard widget.

timogoebel:
@tbrisker : Thanks for the explanation. I agree, we should get rid of unnecessary db calls. Why not add a `if new_record?` condition to the model? That would prevent, what you were seeing, right? What stops from working is, that existing hosts that don't have a PXELoader might not be able to build anymore. We don't have a migration that actually adds the sane default for existing hosts. So when a user rebuilds a host, that's broken. In addition I believe, that the Host view page might not show the PXELoader to be empty for such hosts. But I haven't tested this, yet.

tbrisker:
@timogoebel a default was added to all existing host[group]s in the [original patch](https://github.com/theforeman/foreman/pull/3679/files#diff-7cb410ccb7ad70aae1e31aa68880d762), so there is no benefit to re-running this on them as they will already have pxeloader set and this will be a noop.
We could add a `new_record?` condition (which is the second option I mentioned), however this will trigger multiple times during new host creation in UI as almost all changes in the host form actually instantiate a new host with the updated parameters.

timogoebel:
> @timogoebel a default was added to all existing host[group]s in the original patch, so there is no benefit to re-running this on them as they will already have pxeloader set and this will be a noop.

You are right. I didn't remember this correctly. So that's not actually not a problem. Great.

> We could add a new_record? condition (which is the second option I mentioned), however this will trigger multiple times during new host creation in UI as almost all changes in the host form actually instantiate a new host with the updated parameters.

Yes, I know. If we create the Hosts with the pxe_loader parameter already set, this shouldn't be a problem as the suggestion only runs if the parameter is empty.

timogoebel (approved):

timogoebel:
@tbrisker : As discussed on IRC, I'll test this on Friday. More voices are appreciated.

app/controllers/api/v2/hosts_controller.rb
```diff
@@ -112,6 +112,7 @@ def create
         @host = Host.new(host_attributes(host_params))
         @host.managed = true if (params[:host] && params[:host][:managed].nil?)
         apply_compute_profile(@host)
+        @host.suggest_default_pxe_loader if params[:host][:pxe_loader].nil?
```
lzap: The UI works fine but V2 API (both host and hostgroup) do not.

On this line, it looks like:

```
@host.try(:operatingsystem)
=> nil
@host.hostgroup.operatingsystem
=> #<CentOS7...>
```

This is not likely related to your change, but can you fix this please? We need to take hostgroup OS into account. Also a test for this would help to prevent this. We have a couple, but not for this hostgroup use case.

lzap:
Other than that, I am merging :-)

timogoebel:
> Other than that, I am merging :-)

I think, there is now rush. The question, why the controllers are better than the model is still unanswered.

timogoebel:
Ok, so I dug into this a little more. I think, moving this to the controller is probably the easiest choice. I still don't like it, but it's a quick fix that doesn't hurt to much and we should focus one more important issues/workflows.
What we should just be aware though is, that hosts created from facts don't get the pxe loader set. Should we trigger the automatic suggestion there, too? Or do we just don't care. I'm uncertain.

@tbrisker : I have one wish, though. Could you please add a test for APIv1 to check that the suggestion still works there.

tbrisker:
@timogoebel Thanks, I will add the tests you and @lzap asked for and fix the issue with hostgroup OS.  As for hosts that are created from facts, I think they are created as unmanaged by default so it shouldn't matter, we could trigger the suggestion when making a host managed as well.

timogoebel:
> I think they are created as unmanaged by default so it shouldn't matter

They are. Sounds good. 👍 

> we could trigger the suggestion when making a host managed as well.

Great idea.

app/controllers/api/v2/hosts_controller.rb
tbrisker: @lzap AFAICT, this actually should work for attributes inherited from the hostgroup, because the call to `host_attributes` above should inherit any attributes not set on the host from the hostgroup, therefor if you don't pass an os parameter but the hostgroup has one it will inherit and work correctly.

tbrisker:
@lzap @timogoebel Added a bunch of tests, APIv1 support and trigger on making a host managed.

lzap:
Works fine, any objections @timogoebel ?

timogoebel:
Feel free to merge this when you think its ready. I won't get to this for the next two weeks.

lzap:
Thanks @tbrisker for the investigation and fix!

PR #4064, opened 2016-11-29
Refs #1556 - Remove puppet docroot setting from db and tests

tbrisker:


dLobatog:
👍 if tests pass

dLobatog:
Thanks @tbrisker

PR #4065, opened 2016-11-29
Fixes #17516 - Update jquery to 2.2.4 to fix XSS

dLobatog:
Affected versions of the package (< 1.12) are vulnerable to Cross-site Scripting (XSS) attacks when a cross-domain ajax request is performed without the dataType option causing text/javascript responses to be executed.

https://github.com/jquery/jquery/issues/2432 for more information

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "events": "^1.1.1",
     "flux": "^2.1.1",
     "ipaddr.js": "~1.2.0",
-    "jquery": "~1.11.0",
+    "jquery": "~2.2.4",
     "jquery-flot": "~0.8.3",
     "jquery-ujs": "~1.2.0",
     "jquery.cookie": "~1.4.1",
--- a/webpack/assets/javascripts/foreman_users.test.js
+++ b/webpack/assets/javascripts/foreman_users.test.js
@@ -39,6 +39,6 @@ describe('initInheritedRoles', () => {
     $('.dropdown-menu li a').last().click();
     expect($('.btn').text()).toContain('First');
     expect($('.list-group li[data-id="1"]').is(':visible')).toBeTruthy();
-    expect($('.list-group li[data-id="2"]').is(':visible')).toBeFalsy();
+    expect($('.list-group li[data-id="2"]').css('display')).toEqual('none');
   });
 });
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -44,7 +44,7 @@
     "events": "^1.1.1",
     "flux": "^2.1.1",
     "ipaddr.js": "~1.2.0",
-    "jquery": "~1.11.0",
+    "jquery": "~1.12.0",
```
tbrisker: shouldn't this be 1.13?

dLobatog: There's no 1.13, 1.12 is the one that contains the fix - it was a typo

dLobatog:
Repushed with the right commit title  (it said 1.13 instead of 1.12)

tbrisker:
@dLobatog please rebase, tests are failing due to https://github.com/theforeman/foreman/pull/4064

dLobatog:
Rebased!

tbrisker:
tests are failing. Looks like this upgrade broke something.

tbrisker:
@dLobatog application.js line 66 should be changed to
```javascript
$('a[href="#'+this.id+'"]').addClass("tab-error");
```
to fix the failing test. Care to upgrade jQuery to 2.x while you are at it? It is compatible with 1.x except it dropped IE<9 support. I am looking into upgrading to 3.x which does introduce some breaking changes.

dLobatog:
Updated , including jquery 2.1 and the application.js fix, thanks @tbrisker

webpack/assets/javascripts/foreman_users.test.js
```diff
@@ -39,6 +39,6 @@ describe('initInheritedRoles', () => {
     $('.dropdown-menu li a').last().click();
     expect($('.btn').text()).toContain('First');
     expect($('.list-group li[data-id="1"]').is(':visible')).toBeTruthy();
-    expect($('.list-group li[data-id="2"]').is(':visible')).toBeFalsy();
+    expect($('.list-group li[data-id="2"]').css('display')).toEqual('none');
```
dLobatog, line 42: Seems like :visible meaning changed on 2.1, so I made it more specific

package.json
```diff
@@ -44,7 +44,7 @@
     "events": "^1.1.1",
     "flux": "^2.1.1",
     "ipaddr.js": "~1.2.0",
-    "jquery": "~1.11.0",
+    "jquery": "~2.1.0",
```
tbrisker: Latest version is 2.2.4, any reason not to use that?

dLobatog:
@tbrisker I think I saw some breakages on tests with 2.2.4 but it seems fine now. Updated to 2.2.4

tbrisker:
Works just fine, thanks @dLobatog!

PR #4066, opened 2016-11-29
Refs #17491 - Revert "polyfill Map for ARMv8 node"

mmoll:
This reverts commit f101368449b450b1ede5c822af0852166c45bc4c.

We do use a patched nodejs package on ARMv8 now, as using the polyfill didn't fix it reliably.


tbrisker:
test failure unrelated, merging. Thanks @mmoll !

PR #4067, opened 2016-11-29
Fixes #17501 - Display tooltips on dashboard widgets

tbrisker:


webpack/assets/javascripts/foreman_tools.js
```diff
@@ -43,3 +43,16 @@ export function userInheritedRolesUpdater(id) {
     $(this).parents('.dropdown').find('.btn').html($(this).text() + ' <span class="caret"></span>');
   });
 }
+
+export function activateTooltips(el = $('body')) {
```
ohadlevy, line 47: could you please add tests to this function?

tbrisker:
@ohadlevy and now with tests

webpack/assets/javascripts/foreman_tools.js
```diff
@@ -43,3 +43,16 @@ export function userInheritedRolesUpdater(id) {
...
+  el.find('[rel="twipsy"]').tooltip({ container: 'body' });
+  el.find('.ellipsis').tooltip({ container: 'body',
+                                 title: function () {
```
ohadlevy: nitpick: you can use write something like:
```js
const title = () => {
  return (this.scrollWidth > this.clientWidth ? this.textContent : null);
 };


```

tbrisker: right, still not used to all the nice es6 features :) pushing now fixed

ohadlevy:
thanks @tbrisker

PR #4068, opened 2016-11-29
Fixes #17521 - Use short name for hosts

parthaa:
This commit enforces the policy of using short name for hosts
while preserving the fqdn for interface values provided the user has
enabled 'use_shortname_for_vms' Setting.

### DIFF
--- a/app/services/name_synchronizer.rb
+++ b/app/services/name_synchronizer.rb
@@ -15,10 +15,16 @@ class NameSynchronizer
   # we have to use write_attribute, since host#name= is delegated to primary interface
   # which triggers the sync
   def sync_name
-    @host.send :write_attribute, :name, @interface.name
+    @host.send :write_attribute, :name, interface_name
   end
 
   def sync_required?
-    @interface.primary? && @host.present? && (@host.name != @interface.name)
+    @interface.primary? && @host.present? && (@host.name != interface_name)
+  end
+
+  private
+
+  def interface_name
+    Setting[:append_domain_name_for_hosts] ? @interface.name : @interface.shortname
   end
 end
--- a/app/models/setting/general.rb
+++ b/app/models/setting/general.rb
@@ -21,7 +21,8 @@ class Setting::General < Setting
       self.set('host_power_status', N_("Show power status on host index page. This feature calls to compute resource providers which may lead to decreased performance on host listing page."), true, N_('Show host power status')),
       self.set('http_proxy', N_('Sets a proxy for all outgoing HTTP connections.'), nil, N_('HTTP(S) proxy')),
       self.set('http_proxy_except_list', N_('Set hostnames to which requests are not to be proxied'), [], N_('HTTP(S) proxy except hosts')),
-      self.set('lab_features', N_("Whether or not to show a menu to access experimental lab features (requires reload of page)"), false, N_('Show Experimental Labs'))
+      self.set('lab_features', N_("Whether or not to show a menu to access experimental lab features (requires reload of page)"), false, N_('Show Experimental Labs')),
+      self.set("append_domain_name_for_hosts", N_("Foreman will append domain names when new hosts are provisioned"), true, N_("Append domain names to the host")),
     ]
   end
 
--- a/test/fixtures/settings.yml
+++ b/test/fixtures/settings.yml
@@ -364,3 +364,8 @@ attributes76:
   category: Setting::Provisioning
   default: "['lo', 'usb*', 'vnet*', 'macvtap*', '_vdsmdummy_', 'veth*', 'docker*', 'tap*', 'qbr*', 'qvb*', 'qvo*', 'qr-*', 'qg-*', 'vlinuxbr*', 'vovsbr*']"
   description: 'Ignore fact names that match these values during facts importing, you can use * wildcard to match names with indexes e.g. macvtap*'
+attribute77:
+  name: append_domain_name_for_hosts
+  category: Setting::General
+  default: "true"
+  description: "Should append domain names when new hosts are provisioned"
--- a/test/unit/name_synchronizer_test.rb
+++ b/test/unit/name_synchronizer_test.rb
@@ -3,7 +3,10 @@ require 'test_helper'
 class NameSynchronizerName < ActiveSupport::TestCase
   def setup
     @host = FactoryBot.build_stubbed(:host, :managed)
-    @nic  = FactoryBot.build_stubbed(:nic_managed, :host => @host, :name => 'myname', :primary => true)
+    @domain = FactoryBot.build_stubbed(:domain, :name => "a-domain.to-be-found.com")
+    @nic = FactoryBot.build_stubbed(:nic_managed, :host => @host,
+                              :domain => @domain,
+                              :name => "myname.#{@domain.name}", :primary => true)
     @hsync = NameSynchronizer.new(@host)
     @nsync = NameSynchronizer.new(@nic)
   end
@@ -22,6 +25,23 @@ class NameSynchronizerName < ActiveSupport::TestCase
     end
   end
 
+  context "synchronizer build from host on shortnames" do
+    before do
+      Setting[:append_domain_name_for_hosts] = false
+    end
+    test "#sync_required? detects difference between names" do
+      refute_equal @host.name, @host.primary_interface.shortname
+      assert @hsync.sync_required?
+    end
+
+    test "#sync_name synchronizes name based on interface" do
+      refute_equal @host.name, @host.primary_interface.shortname
+      @hsync.sync_name
+      assert_equal @host.name, @host.primary_interface.shortname
+      assert @host.name.blank?
+    end
+  end
+
   context 'synchronizer build from nic' do
     test '#sync_required? detects difference between names' do
       refute_equal @nic.name, @nic.host.name
@@ -32,7 +52,7 @@ class NameSynchronizerName < ActiveSupport::TestCase
       refute_equal @nic.name, @nic.host.name
       @nsync.sync_name
       assert_equal @nic.name, @nic.host.name
-      assert_equal 'myname', @nic.host.name
+      assert_equal "myname.a-domain.to-be-found.com", @nic.host.name
     end
   end
 


ohadlevy:
@parthaa mind rebasing your PR? test failure seems related to a previous fix.

ares:
should also `sync_required?` be modified to reflect this? also please add tests to https://github.com/theforeman/foreman/blob/develop/test/unit/name_synchronizer_test.rb

parthaa:
> should also sync_required? be modified to reflect this? also please add tests to

updated. added tests too..

parthaa:
Based on comments by @domcleal at http://projects.theforeman.org/issues/17521#note-2
I created a setting "append_domain_name_to_hosts"  to make this work instead of use_shortname_for_vms . I have updated the tests accordingly. 

@ohadlevy @ares @domcleal  @dLobatog really need this to be reviewed asap.. High priority on downstream.

app/models/lookup_value.rb
```diff
@@ -75,7 +75,8 @@ def ensure_fqdn_exists
     md = ensure_matcher(/fqdn=(.*)/)
     return md if md == true || md == false
     fqdn = md[1].split(LookupKey::KEY_DELM)[0]
-    return true if Host.unscoped.find_by_name(fqdn) || host_or_hostgroup.try(:new_record?) ||
+    fqdn_matched = Host.search_for("facts.fqdn=#{fqdn}").any? || Host.unscoped.where(:name => fqdn).any?
```
ares: If users prefer to use short names, shouldn't the matchers also use short names?

If we keep this change, I think the order should be the other way around, first we should search the host by fqdn and then fallback to facts. I see two reasons for this, first that the host name would have higher priority in case these two values get out of sync so it would prefer current behavior, second is that the search for is more resource consuming while the host name will be mostly used.

Also shouldn't the search_for be called on `Host.unscoped`?

ares: This change also needs to be covered by tests

app/services/name_synchronizer.rb
```diff
@@ -15,10 +15,12 @@ def initialize(object)
...
   # which triggers the sync
   def sync_name
-    @host.send :write_attribute, :name, @interface.name
+    name = Setting[:append_domain_name_for_hosts] ? @interface.name : @interface.shortname
+    @host.send :write_attribute, :name, name
   end
 
   def sync_required?
-    @interface.primary? && @host.present? && (@host.name != @interface.name)
+    name = Setting[:append_domain_name_for_hosts] ? @interface.name : @interface.shortname
```
ares: please extract this to some private method so it's not duplicated in `sync_name` and `sync_required?`

ares (changes requested):
Could you please share more information why this is needed? I don't like the fact that we'd store different value in host name based on settings. What if user later changes the setting value? There's no simple way to update all hosts right?

parthaa:
> Could you please share more information why this is needed? I don't like the fact that we'd store different value in host name based on settings. What if user later changes the setting value? There's no simple way to update all hosts right?

Entire point of this PR is because some customers do not want domain names appended to the host name *even* if they choose an interface. They *always* want the short name. Since thats the not the default setting for foreman we need this setting.

parthaa:
> If we keep this change, I think the order should be the other way around, first we should search the host by fqdn and then fallback to facts. I see two reasons for this, first that the host name would have higher priority in case these two values get out of sync so it would prefer current behavior, second is that the search for is more resource consuming while the host name will be mostly used.

Lets start with
https://github.com/theforeman/foreman/blob/develop/app/models/host/base.rb#L306

So when you do a Host.save with a overridden smart parameter, the host creates a "LookupValueMatcher" that has a match attribute  ```fqdn=<fqdn> || name ```  

When the newly created LookupValue tries to get saved it hits -> https://github.com/theforeman/foreman/blob/develop/app/models/lookup_value.rb#L73
" Host.unscoped.find_by_name(fqdn) " 

This causes the lookup fail and the host save to thus fail.. Basically we need a way to change this code to say the equivalent of
"Host.unscoped.find_by_fqdn(fqdn)" instead of " Host.unscoped.find_by_name(fqdn) " 
Changes in that file are meant to address that.

Order wise I think we have it better here. Because if the LookupValueMatcher has ```fqdn=<fqdn> || name ```  one would think fqdn gets the higher priority.  I can see the "more resources" argument though. I changed it anyway.

parthaa:
> Also shouldn't the search_for be called on Host.unscoped?

changed. Good call

app/services/name_synchronizer.rb
```diff
@@ -15,10 +15,18 @@ def initialize(object)
   # we have to use write_attribute, since host#name= is delegated to primary interface
   # which triggers the sync
   def sync_name
-    @host.send :write_attribute, :name, @interface.name
+    name = interface_name
```
ares: I'd avoid creating a variable here, how about this

```ruby
@host.send :write_attribute, :name, interface_name
```

app/services/name_synchronizer.rb
```diff
@@ -15,10 +15,18 @@ def initialize(object)
...
+    @host.send :write_attribute, :name, name
   end
 
   def sync_required?
-    @interface.primary? && @host.present? && (@host.name != @interface.name)
+    name = interface_name
+    @interface.primary? && @host.present? && (@host.name != name)
```
ares:
```ruby
 @interface.primary? && @host.present? && (@host.name != interface_name)
```

app/models/setting/provisioning.rb
```diff
@@ -20,6 +20,7 @@ def self.default_settings
       self.set('libvirt_default_console_address', N_("The IP address that should be used for the console listen address when provisioning new virtual machines via Libvirt"), "0.0.0.0", N_('Libvirt default console address')),
       self.set('update_ip_from_built_request', N_("Foreman will update the host IP with the IP that made the built request"), false, N_('Update IP from built request')),
       self.set('use_shortname_for_vms', N_("Foreman will use the short hostname instead of the FQDN for creating new virtual machines"), false, N_('Use short name for VMs')),
+      self.set("append_domain_name_for_hosts", N_("Foreman will append domain names when new hosts are provisioned"), true, N_("Append domain names to the host")),
```
ares: I don't think this is specific to provisioning. The same would apply when new hosts registers by facts upload or is created by API call. I think it should go to general settings and the help should be reworded to something like "Append domain names to new hosts".

ares (dismissed):
Sorry, it does not work. When I disable domain appending and I override some smart class parameter, I see the default value in host edit form in Parameters tab correctly. But when I define matcher in smart class parameter form with this fqdn equal to this short name without domain, the new value is not displayed in the Parameters tab in the host form. I still see only the default one. Also when I try to override it from host form, once I click submit I'm redirected back to host form without displaying any error message, in logs I can find following

```
2017-01-18T16:52:39 5b2b4524 [app] [D] Ignoring #<LookupValue id: nil, match: "fqdn=joy-ruffins.example.tst", value: "bbb", lookup_key_id: 157, created_at: nil, updated_at: nil, omit: false> match errors: ["fqdn=joy-ruffins.example.tst does not match an existing host"]
```

it tries to create the matcher with fqdn instead of hostname.

Also when I switch the setting to apply domains again, it leaves all matchers with short names making them invalid when I try to save the smart class parameter. We probably have the same problem with other renaming so maybe that is acceptable. Also the host correctly changes name after settings switch and new save.

app/models/setting/provisioning.rb
ares: Since it applies whenever the host is saved maybe the wording should not mention new hosts but something like "Append domain names to host names upon save"

dLobatog: 👍 "Append domain names to new hosts" - then explain what it means on the description sounds good

ares: it's not only for new hosts (my second comment), it applies whenever the host is saved and if the setting changed, save can append or remove the domain e.g. when I change the owner

parthaa:
> Sorry, it does not work. When I disable domain appending and I override some smart class parameter, I see the default value in host edit form in Parameters tab correctly. But when I define matcher in smart class parameter form with this fqdn equal to this short name without domain, the new value is not displayed in the Parameters tab in the host form. I still see only the default one. Also when I try to override it from host form, once I click submit I'm redirected back to host form without displaying any error message, in logs I can find following

@ares @dLobatog 

So I am guessing might have happened is you the override set to  match based on shortname instead of using fqdn even as the matcher said it was going to use fqdn. Would you mind confirming this again. When setting up the override matcher please set the full hostname instead of just the shortname.

ohadlevy:
@ares what is the status of this PR ? thanks?

app/models/setting/provisioning.rb
dLobatog: Correct, it should probably be moved to general.

app/models/lookup_value.rb
```diff
@@ -81,7 +81,8 @@ def ensure_fqdn_exists
     md = ensure_matcher(/fqdn=(.*)/)
     return md if md == true || md == false
     fqdn = md[1].split(LookupKey::KEY_DELM)[0]
-    return true if Host.unscoped.find_by_name(fqdn) || host_or_hostgroup.try(:new_record?) ||
+    fqdn_matched = Host.unscoped.where(:name => fqdn).any? || Host.unscoped.search_for("facts.fqdn=#{fqdn}").any?
```
dLobatog: This is not really user-friendly, why not adding a new matcher 'shortname'? Maybe it'd be easy to add this matcher in another PR then use it here, just to not mix stuff and keep PRs simple.

dLobatog (changes requested):
@parthaa 👍 to the setting, and the name synchronizer seems to work pretty well for created and imported hosts.  

The fqdn matcher though, it's counterintuitive IMO. I would suggest leaving fqdn matcher as it stands and making a new shortname one.. not sure about other people's thoughts on it here. 

The new `shortname` matcher would be useful for people that mix shortnames/fqdns too, and it could be documented on it's own without risking confusing people. Also it'd allow to merge part of this PR right away.

Thanks @parthaa !

iNecas:
@parthaa what is the status of this PR?

app/models/lookup_value.rb
parthaa: > This is not really user-friendly, why not adding a new matcher 'shortname'? Maybe it'd be easy to add this matcher in another PR then use it here, just to not mix stuff and keep PRs simple.


@dlobatog Ok let me see if I can state this in a clearer fashion. That line in lookup value is deceptively simple looking.

The bug is that Customer wants short hostnames when using the API But wants to have a fully formed fqdn in their puppet fact. They were able to get away with this setup before Host unification with Katello because Content Host and Host were 2 different things. However when they unified they wanted the host to have the short host name. But they also wanted the fqdn fact in puppet to reflect fully qualified name including the interface domain name.

Not just that. On top of this they want to be able Override certain puppet configuration (aka smart class parameters) based on the FQDN
Foreman code prior to this change made the assumption that Host.Name == FQDN when matching.
That is why we see code like
```return true if Host.unscoped.find_by_name(fqdn) || host_or_hostgroup.try(:new_record?) ||```

This works fine in most cases because foreman always appends interface name. However it would be better if in addition to matching host names for fqdn it also  tries to identify/find/match hosts with their actual fqdn value as specified in their puppet fact. 

All my code does -> When Matching..
- Try matching by the existing method first. i.e See if you can find a host that has name given in the FQDN matcher . Identical to the line above
-  If that does not exist See if there is a host that has the FQDN puppet fact matching the value given in the FQDN matcher
that 2nd case is where this customer's matcher would match.

I would argue that this change improves the existing 'matching' setup by actually matching fqdns for real instead of assuming name given the foreman host == FQDN

Adding a shortname matcher is not really going to solve the issue for this customer because they want the existing FQDN matcher to actually match FQDN. Not name.

parthaa:
> Correct, it should probably be moved to general.
Done

app/models/lookup_value.rb
ares: If customer needs to combina short name and FQDN in single matcher, I think a new matcher should be added. Something like `fqdn_or_shortname`. We're trying to improve project usability, this would go the opposite direction. I think the best solution is to add `shortname` and `fqdn_or_shortname` to already existing `fqdn` and leave `fqdn` untouched.

parthaa:
@dLobatog @ares created -> http://projects.theforeman.org/issues/21929 and updated this.
Thanks for reading my comments earlier.

dLobatog (approved):
@parthaa 👍 from my side to the current changes, I think it's clear that if you change the setting, that only applies to new hosts. 

@ares any further comments?

ares:
I did not test it, I still think this is a bit confusing especially if the setting is global it does not address the use case where some hosts have FQDN and some not. But I'm happy to dismiss my review so it's not blocked on me finding time to do another round of review.

dLobatog (approved):
Thanks @parthaa  - let's follow up this with http://projects.theforeman.org/issues/21929 when you or someone else gets a chance.

ares:
Thanks, setting the release to 1.18 in redmine, please don't forget to update redmine issue after merging.

PR #4069, opened 2016-11-30
fixes #17157 - inconsistent mapping of host owner

kgaikwad:
There is user_group which is assigned to one host as owner
with id=x. For organization assigned to user with same id=x,
on edit page showing a grey color with message
"this is used by a host".
With this commit, it should not show message to organization
assigned to user with same id as of user_groups.

Also, refactored the association code for
(User, Usergroup) -> UserRoles.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -26,7 +26,7 @@ class User < ActiveRecord::Base
   has_many :auditable_changes, :class_name => '::Audit', :as => :user
   has_many :direct_hosts,      :class_name => 'Host',    :as => :owner
   has_many :usergroup_member,  :dependent => :destroy,   :as => :member
-  has_many :user_roles,        -> { where(:owner_type => 'User') }, :dependent => :destroy, :foreign_key => 'owner_id'
+  has_many :user_roles,        :dependent => :destroy, :as => :owner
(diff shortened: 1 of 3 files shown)


ohadlevy:
@kgaikwad thanks for the PR. does it make sense to add a test to show the issue is no longer present?

ohadlevy:
ok to test

kgaikwad:
@ohadlevy , yes I will add test and update the PR.

kgaikwad:
@ohadlevy , thank you for reviewing the PR. Added one test in [taxonomix_test.rb](https://github.com/theforeman/foreman/pull/4069/commits/d52f2747ba4abde264e6b4898b63411c4d5c4061#diff-ce286c52a1c29f62fc12edf6efa6eff8). Let me know if any changes required.

dLobatog:
`Bundler::GemRequireError: There was an error while trying to load the gem 'autoprefixer-rails'.
Gem Load Error is: Could not find a JavaScript runtime. See https://github.com/rails/execjs for a list of available runtimes. ` is the error found on pg on `upgrade` , I don't think the change has anything to do with the JS runtime so I'll go ahead and merge. Thanks @kgaikwad @ohadlevy !

PR #4070, opened 2016-11-30
Fixes #17526 - Taxonomy.ignore? does not work with "any context"

dLobatog:
In this scenario:

1. User has one organization - "E-Corp" - this organization allows the
user to see every hostgroup (via selecting "all host groups" when
editing the organization).

2. User has a location "Czech republic" that also includes all host
groups through that checkbox.

3. User visits /hostgroups with "any context" selected, or "E-Corp/any
location", or "any organization/Czech republic". The result is that the
user does not see all host groups even though there's an
organization/location combination (e-corp/czech republic) that should
allow the user to see all host groups.

---

The reason is that when `Hostgroup.taxable_ids` is called
`Organization.ignore?` does not realize that there it should look in all
Organizations to see if any of them 'ignores' (has "all host groups"
    checked) the resource. The same thing happens with Locations.

The fix is to make `ignore?` aware the 'Organization.current == nil'
means 'Any organization', not 'No organization'.


test/models/shared/taxonomies_base_test.rb
```diff
@@ -0,0 +1,441 @@
...
+    end
+
+    test "taxonomy should not save when matcher is exactly 256 characters" do
+      parent = FactoryGirl.create(:"#{taxonomy_name}", :name => 'a' * (255 - taxonomy_name.length - 2))
+      taxonomy = taxonomy_class.new :parent => parent, :name => 'b'
+      refute_valid taxonomy
+      assert_equal _("is too long (maximum is 0 characters)"), taxonomy.errors[:name].first
+    end
+
+    test 'ignores the taxable_type if current taxonomy ignores it' do
```
dLobatog, line 402: These two are the only new tests, the rest of them are just removing the duplication between location_test and organization_test and moving them both to a module

dLobatog:
Tons of tests broke, setting to WoC

dLobatog:
@theforeman-bot ok to test

dLobatog:
Test failure in Katello is due to https://github.com/Katello/katello/pull/6480/files

tbrisker:
Thanks @dLobatog !

PR #4073, opened 2016-12-01
Fixes #17495 - align task and report expiration

ares:


app/services/actions/foreman/report/import.rb
```diff
@@ -31,6 +31,10 @@ def humanized_name
         def humanized_input
           input[:report_class].constantize.humanized_name
         end
+
+        def self.cleanup_after
+          "#{Hash[::Report::DEFAULT_EXPIRATION.parts][:days]}d"
```
ohadlevy, line 36: this will just set the default to 7 days, yet it will ignore any custom change, maybe change the commit message then?

app/models/report.rb
```diff
@@ -71,7 +72,7 @@ def <=>(other)
   # Expire reports based on time and status
   # Defaults to expire reports older than a week regardless of the status
   def self.expire(conditions = {})
-    timerange = conditions[:timerange] || 1.week
+    timerange = conditions[:timerange] || DEFALT_EXPIRATION
```
ohadlevy: nitpick: typo DEFAULT

app/services/actions/foreman/report/import.rb
ares: I updated the commit message to mention default

ares:
thanks @ohadlevy, I addressed all comments, would you mind giving it another look?

ohadlevy:
thanks @ares that's a better default.

I would love seeing task deletion based on task status (and in that case the default can be lower than a day, or even an hour ).

will merge once tests are passing.

ares:
it can be adjust by config/settings.plugins.d/foreman_tasks.yaml, see https://github.com/theforeman/foreman-tasks#tasks-cleanup for an example, the task status can be set there too, we could start shipping some default or at least example file

ohadlevy:
@ares please rebase your pr to avoid the test failures..

ares:
rebased

ohadlevy:
thanks @ares, katello failure unrelated.

PR #4074, opened 2016-12-01
Fixes #5483,#11106 - pass the virtualswitch value to fog

iNecas:
We can't pass the network id directly to the fog, but it accepts the
virtualswitch parameter, that we can use to point to the right value.

Also, it fixes another issue, where we used the vsphere network id to
match against the interface network value, in case of portgroup, but a
`key` value should be used instead, as the key and id might differ in
some situations.


iNecas:
mistakenly opened

PR #4075, opened 2016-12-01
Refs #15779 - move the foreman-tasks to own bundle group

iNecas:
This makes it easier to achieve a minimal setup, that might be needed for
things like plugins building etc.

See discussion in https://github.com/theforeman/foreman-packaging/pull/1436


dLobatog:
`NameError: uninitialized constant Dynflow::Testing::Algebrick`

iNecas:
I've updated the application.rb to try to require the group when available

dLobatog:
Good call, thanks @iNecas

PR #4077, opened 2016-12-02
Fixes #6006 - Add options to disable updates from facts

treydock:
Add option ignore_puppet_facts_for_operatingsystem - do not update system's operating system based on facts
Add option ignore_puppet_facts_for_domain - do not update system's domain based on facts

### DIFF
--- a/app/models/host/managed.rb
+++ b/app/models/host/managed.rb
@@ -463,7 +463,15 @@ class Host::Managed < Host::Base
   end
 
   def attributes_to_import_from_facts
-    super + [:domain, :architecture, :operatingsystem]
+    attrs = [:architecture]
+    if !Setting[:ignore_facts_for_operatingsystem] || (Setting[:ignore_facts_for_operatingsystem] && !operatingsystem.present?)
+      attrs << :operatingsystem
+    end
+    if !Setting[:ignore_facts_for_domain] || (Setting[:ignore_facts_for_domain] && !domain.present?)
+      attrs << :domain
+    end
+
+    super + attrs
   end
 
   def populate_fields_from_facts(facts = self.facts_hash, type = 'puppet')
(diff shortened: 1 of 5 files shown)


dLobatog:
@parthaa #4068 would play well with this

tbrisker:
How about just making one setting, `ignore_facts_update` or something, that is an array of all the attributes that should be ignored when updating hosts from facts? There are are already similar options for environments, subnets, ip and mac that could be combined into this. Also, facts are also provided by multiple config management plugins, so it is better not to name the option with a puppet-specific name.

dLobatog:
@tbrisker I do agree about the naming, however it's pretty useful to be able to disable only certain things. Lately we had a customer which did want subnets & OSs to be imported but not domains and we talked about possibly fixing it by doing what this PR exactly does. So IMO it'd be configurable as it is with this PR - I won't merge though if you have any concerns about it.

tbrisker:
@dLobatog I'm not against this per se, I'm just saying instead of ~5 different options that each disables one attribute let's have one options that takes a list and you can choose which attributes to disable, i.e. `disable_fact_update = [os, subnet]`. It will also make the importer part easier to implement.

treydock:
@tbrisker I like the idea of a single option to determine which facts are not imported.  I suppose this PR could focus on adding `ignore_facts_update` option and later commits could focus on migrating existing settings into the new setting.  Let me know what the final verdict is on how the option(s) should be named.

dLobatog:
That sounds fine to me - I'd rather have the feature than not , especially when the code it's right there. If this gets merged by someone else please ping me and I'll try to craft a PR quickly to migrate the settings to that.

ares:
I don't think we should merge these config options into one array. It does not lower the complexity of various settings combination. The inline description for all of these "ignore" settings would be also pretty big. I'd just remove puppet from the name, and hopefully from other settings too.

treydock:
@dLobatog @ares I've updated PR to remove the word puppet from settings.  Should I update PR further to change this into a single setting that takes an array of facts to ignore?

dLobatog:
Merged as ef9f2f9d14ac0a501c76e23b3d2aca39b267943a, thanks @treydock @ares @tbrisker 

Let's discuss how to make the importer better (although I think this handles it fine already, and it's quite easy to read) on another PR

ares:
:clap:

PR #4078, opened 2016-12-04
Fixes #17548 - Orphaned smart class parameters are hidden

orrabin:


ohadlevy:
:+1: ack on green tests.

ohadlevy:
@orrabin note the test failures...

orrabin:
@ohadlevy I added fixtures because the integration tests where using orphaned smart class parameters. Please notice the api would always filter them out which is why I also had to fix some api tests after connecting these parameters to environment classes.

ohadlevy:
thanks @orrabin

PR #4080, opened 2016-12-04
Fixes #17570 - Only execute `onContentLoad` once per page

tbrisker:
When an event is triggered, jQuery attempts to execute the on<name>
method if it exists for event <name>. We were also manually triggering
the `onContentLoad` every time the `ContentLoad` event was triggered,
leading to the function executing twice on every page.
This commit removes the explicit event listener so that the function
will only execute once.

### DIFF
--- a/app/views/hosts/show.html.erb
+++ b/app/views/hosts/show.html.erb
@@ -56,7 +56,7 @@
         <%= spinner(_('Loading VM information ...')) %>
         </div>
       <% end %>
-      <div id="nics" class="tab-pane" data-ajax-url='<%= nics_host_path(@host)%>' data-on-complete='onContentLoad'>
+      <div id="nics" class="tab-pane" data-ajax-url='<%= nics_host_path(@host)%>' data-on-complete='tfm.tools.activateTooltips'>
         <%= spinner(_('Loading NICs information ...')) %>
       </div>
       <% if @host.bmc_available? %>
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -21,8 +21,6 @@
 //= require editable/bootstrap-editable
 //= require editable/rails
 
-$(document).on('ContentLoad', onContentLoad);
-
 $(document).on("page:fetch", tfm.tools.showSpinner)
 
 $(document).on("page:change", tfm.tools.hideSpinner)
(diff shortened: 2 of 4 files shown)


app/views/hosts/show.html.erb
```diff
@@ -56,7 +56,7 @@
         <%= spinner(_('Loading VM information ...')) %>
         </div>
       <% end %>
-      <div id="nics" class="tab-pane" data-ajax-url='<%= nics_host_path(@host)%>' data-on-complete='onContentLoad'>
+      <div id="nics" class="tab-pane" data-ajax-url='<%= nics_host_path(@host)%>' data-on-complete='tfm.tools.activateTooltips'>
```
tbrisker, line 59: There is no need to execute the entire function here, only tooltip activation is needed.

webpack/assets/javascripts/foreman_tools.js
```diff
@@ -44,7 +44,8 @@ export function userInheritedRolesUpdater(id) {
   });
 }
 
-export function activateTooltips(el = $('body')) {
+export function activateTooltips(el = 'body') {
+  el = $(el);
```
tbrisker, line 39: make sure the element passed into the function is a jQuery object. This is needed when the function is called as a `data-on-complete` callback.

tbrisker:
On certain pages (the more complex forms such as host edit) this also saves ~100 ms per page load.

dLobatog:
@tbrisker Thanks, this is huge 😄 glad to see you acted on it. Mind to rebase? There are conflicts with the ES6 PR that have to be resolved.

tbrisker:
@dLobatog rebased, thanks

shlomizadok:
@tbrisker++ looks good

lzap:
Is this related to turbolinks sometimes loading some index pages twice? I tried to reproduce it today and I can't anymore...

lzap:
For example when I click on All hosts and then Discovered hosts. For the first time it reloads all assets, while when I click on the same link it's fast. Is this expected behavior?

tbrisker:
@lzap this is not related to turbolinks. This has to do with the onContentLoad() function being executed twice by mistake.
The double-loading behaviour of turbolinks is when it tries to only load changes but realizes the page changed too much and it needs to load it all (or something like that). TBH I'm not sure turbolinks actually improves our performance.

lzap:
Ok thanks, tests are green and it works, I guess that's a point for you!

PR #4081, opened 2016-12-04
Fixes #17571 - Refactor tfm.tools to use ES6 syntax

tbrisker:
This also extracts the inherited user roles function to a sepereate
module, adds tests and refactors it. These changes also fix two other
issues:
1. a possible stored XSS in usergroup names
2. tooltips on tables were broken (due to => syntax not changing `this`)

### DIFF
--- a/webpack/assets/javascripts/foreman_tools.js
+++ b/webpack/assets/javascripts/foreman_tools.js
@@ -9,10 +9,10 @@ export function hideSpinner() {
 }
 
 export function iconText(name, innerText, iconClass) {
-  let icon = '<span class="' + iconClass + ' ' + iconClass + '-' + name + '"/>';
+  let icon = `<span class="${iconClass} ${iconClass}-${name}"/>`;
 
   if (innerText !== '') {
-    icon += '<strong>' + innerText + '</strong>';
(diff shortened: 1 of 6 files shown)


webpack/assets/javascripts/foreman_tools.js
```diff
@@ -36,17 +35,10 @@ export function activateDatatables() {
...
-    $(this).parents('.dropdown').find('.btn').html($(this).text() + ' <span class="caret"></span>');
-  });
-}
-
-export function activateTooltips(el = $('body')) {
+export function activateTooltips(el = 'body') {
+  el = $(el);
   el.find('[rel="twipsy"]').tooltip({ container: 'body' });
-  el.find('.ellipsis').tooltip({ container: 'body', title: () => {
+  el.find('.ellipsis').tooltip({ container: 'body', title: function () {
```
tbrisker, line 41: We can't use `=>` here because we want `this` to point at the object the tooltip is added to - not the tools module.

gailsteiger:
ready to merge

dLobatog:
Thanks @tbrisker @gailsteiger !

PR #4082, opened 2016-12-05
fixes #17574, #17364 - update scoped_search to 4.x

domcleal:
scoped_search's sort() now accepts a replacement for params which is
filtered in an overridden Foreman sort helper, permitting additional
parameters used in Foreman's index pages.

Replaces :in and :alias in search definitions with preferred keywords.


app/helpers/application_helper.rb
```diff
@@ -192,6 +192,11 @@ def auto_complete_search(name, val, options = {})
     text_field_tag(name, val, options)
   end
 
+  def sort(field, permitted: [], **kwargs)
+    kwargs[:url_options] ||= params.permit(permitted + [:locale, :search])
+    super(field, kwargs)
```
ohadlevy, line 197: what is super in this context? shouldnt we use alias_method_chain in an helper/mixin?

domcleal: `super` is always the overridden method from scoped_search as dependencies will be loaded before Foreman's own helpers. I can change it to a chain or prepend, but it seems unnecessary at the moment.

ohadlevy: OK, I had a vague memory that we had similar problem with will paginate, but can't remember the root cause.. OK from my side.

PR #4083, opened 2016-12-05
fixes #17395 - use admin user for load_template_vars

domcleal:
When `load_template_vars` is called from before_action, the current user
must already be set to ensure caching of parameters inside Host#params
is done with consistent permissions. Otherwise when Foreman::Renderer's
kickstart_attributes (or similar) renders a ptable that uses params,
the method will cache an empty parameters hash.

### DIFF
--- a/app/controllers/unattended_controller.rb
+++ b/app/controllers/unattended_controller.rb
@@ -12,6 +12,7 @@ class UnattendedController < ApplicationController
     alias_method_chain f, :unattended
   end
 
+  before_action :set_admin_user, :unless => Proc.new { preview? }
   # We want to find out our requesting host
   before_action :get_host_details, :allowed_to_install?, :except => :hostgroup_template
   before_action :handle_ca, :if => Proc.new { params[:kind] == 'provision' }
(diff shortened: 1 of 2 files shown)


app/controllers/unattended_controller.rb
```diff
@@ -12,6 +12,7 @@ class UnattendedController < ApplicationController
     alias_method_chain f, :unattended
   end
 
+  before_action :set_admin_user, :unless => Proc.new { preview? }
```
tbrisker, line 15: I think `preview?` will return false on hostgroup template view, leading to any user being able to possibly see details they should not.

tbrisker (changes requested):

app/controllers/unattended_controller.rb
domcleal: Host group templates, almost by definition, are on well-known URLs and can be opened by anybody without the need for a preview mechanism. They should probably always be rendered with an admin account.

There's little point in restricting which parameters a logged in user can view in a host group template render if one can simply log out and view them all.

tbrisker (approved):

PR #4084, opened 2016-12-05
Fixes #17580 - React Chart Component has unsafe function

gailsteiger:
make hasData function more robust

### DIFF
--- a/webpack/assets/javascripts/react_app/components/charts/Chart.js
+++ b/webpack/assets/javascripts/react_app/components/charts/Chart.js
@@ -8,7 +8,10 @@ class Chart extends React.Component {
   }
 
   hasData() {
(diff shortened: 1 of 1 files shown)


domcleal:
I don't know why you have attributed this commit (a8c9f1e) to me, I think you're mistaken.

mmoll:
it seems 78e86121 got squahed into this?

gailsteiger:
@tbrisker @dLobatog could you please review

ohadlevy:
does it require any additional tests?

gailsteiger:
@ohadlevy I don't think it needs another test. It's a very edge case. but I made the change because I generally use the more robust coding convention.

It's not something I ever test, just programming defensively.

dLobatog:
Thanks @gailsteiger !

PR #4086, opened 2016-12-06
Fixes #17587 - remove autoprefixer gem

tbrisker:


tbrisker:
TBH I'm not certain that we really need this gem at all.

dLobatog:
Feel free to remove it - it was a dependency of bootstrap-sass that we needed to pin so that Foreman w/ Patternfly worked in Ruby 1.9.3. We don't care about that ruby now so...

tbrisker:
test failure unrelated

dLobatog:
Thanks @tbrisker !

PR #4089, opened 2016-12-07
Fixes #6817 - AWS VPC groups - show after saving

swapab:
The VPC group(security groups) are getting saved but it's not visible after save. Probably due to AWS-fog API updates, the formed data-structure got outdated.
Covered various cases where
* security groups not showing after saving compute-profile
* security groups weren't visible when subnet has been set but security group for that subnet isn't selected. Security groups listed only after subnet drop-down selection was changed


dLobatog:
Thanks @swapab

shawnlopresto:
Thanks @swapab.

PR #4091, opened 2016-12-07
Fixes #17599 - Smart vars allow you to select 'none' class

dLobatog:
When you edit a smart variable, it allows in the form to 'not' associate
it with any Puppet class. This doesn't show any error on the UI, however
it's not associated and you will see "Failed to save: Puppetclass can't
be blank" on the logs.

Additionally some strings were not translated at all so I took the
chance to add support for that.

### DIFF
--- a/app/helpers/lookup_keys_helper.rb
+++ b/app/helpers/lookup_keys_helper.rb
@@ -24,7 +24,7 @@ module LookupKeysHelper
       text_f(f, :puppetclass_id, :label => _('Puppet Class'), :value => f.object.param_class, :disabled => true)
     else # new smart-var with no particular context
          # Give a select for choosing the parent puppetclass
-      select_f(f, :puppetclass_id, Puppetclass.all, :id, :to_label, { :include_blank => _('None') }, {:label => _("Puppet class")})
+      select_f(f, :puppetclass_id, Puppetclass.all, :id, :to_label, {}, {:label => _("Puppet class")})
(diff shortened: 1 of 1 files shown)


app/views/lookup_keys/_fields.html.erb
```diff
@@ -4,9 +4,9 @@
   <%= remove_child_link(_("Remove %s?") % (f.object.new_record? ? _("Variable") : f.object), f, { :class => 'btn btn-danger hide' }) if controller_name ==  "puppetclasses" && !is_param %>
   <fieldset>
     <h2><%= _("Parameter details") %></h2>
-    <%= text_f f, :key, :disabled => is_param, :size => "col-md-8" %>
+    <%= text_f f, :key, :disabled => is_param, :size => "col-md-8", :label => _('Key')  %>
```
domcleal: Note sure that model attributes need extracting, they're already in model_attributes.rb (but probably under the wrong name as 6b4b489 seemed to rename them).

This will mean twice the number of translations, but only one gets used. Better fix this separately if it's really an issue, it's unrelated.

domcleal (changes requested):
Best to remove the unrelated i18n changes and file a ticket with specifics about the issue.

dLobatog:
Updated to remove translations - they're on https://github.com/theforeman/foreman/pull/4097 now

domcleal (approved):

domcleal:
Merged with the i18n references removed from the commit message. Thanks @dLobatog.

PR #4094, opened 2016-12-08
Fixes #17576 - check for mismatch revised

amitkarsale:
on selecting 'fail on mismatch' while assigning the org or location it always used to fail with mismatch error, appropriate method is invoked for the same now

### DIFF
--- a/app/controllers/concerns/foreman/controller/taxonomy_multiple.rb
+++ b/app/controllers/concerns/foreman/controller/taxonomy_multiple.rb
@@ -42,7 +42,7 @@ module Foreman::Controller::TaxonomyMultiple
       #hosts location needs to be updated before import missing ids
       taxonomy.import_missing_ids
     else
-      if taxonomy.need_to_be_selected_ids.count == 0
+      if taxonomy.mismatches.count == 0
         @hosts.update_all("#{type}_id".to_sym => taxonomy.id)
       else
         error "Cannot update #{taxonomy.type} to #{taxonomy.name} because of mismatch in settings"


amitkarsale:
[test]

dLobatog:
@amitkarsale Please take a look at broken tests 😄 Can you explain a bit the change? This area of the code is always a bit complicated to remember. Thanks !

dLobatog:
Seems like you squashed https://github.com/theforeman/foreman/pull/4052 here by mistake?

dLobatog:
Thanks @amitkarsale  - I was able to reproduce, but "Fix on mismatch" is supposed to do this. "Fail on mismatch" is really expected to fail if you're moving a host in Loc A with hostgroups/domains/subnets/compute_resources/whatever... in Loc A to some other Loc B. 

Fix on mismatch is supposed to move all objects to the new Loc B, while this one, Fail on mismatch is supposed to throw an error. The idea is that you run "fail on mismatch" if you don't want to change the loc/org of the associated objects.

From what I read on the BZ @ares agrees with that, and the issue is more about displaying what exactly failed. You'd have to go through the elements where it "Failed on mismatch". E.G: 

1. Host in location A, with subnet in location A. Let's say all other associations (hostgroup, compute resource, are global..)
2. Move host from location A to B. It will fail on mismatch with an error "Cannot update Location to B because of mismatch in settings". Users will not like the error as it doesn't tell them how to fix it or what went wrong.

The idea is that you fix what happens in step 2, by providing a good error message like: "Cannot update location to B because of mismatch: Host's subnet #{@host.subnet} only exists in location A"

Could you write a test that does what the issue says (move a host in location A with hostgroup in A to location B using 'fail on mismatch')? Additionally it'd be nice if we get tests for the following scenario:

1. hostgroup with locations A and B
2. host in location A
3. move host to location B using fail on mismatch. It shouldn't fail as the resources are in B too.

PR #4095, opened 2016-12-08
Fixes #10539 - Add default names for ifaces for oVirt CR VMs

bagasse:
Add default names for interface of new VM on oVirt compute resource.
The naming scheme is the same as used in oVirt WebUI. If no name is provided for the network interface, it will be named "nic{nic_number}" where nic_number is an integer > 1.
This is done on the oVirt compute resource backend and not on the form part of foreman WebUI so this can be used via hammer/API calls too. Any advises on how to add test on this feature are welcome.


bagasse:
[test]

bagasse:
[test foreman]

dLobatog:
Thanks @bagasse !

PR #4097, opened 2016-12-08
Fixes #17611 - Localize smart variables strings

dLobatog:
This commit rename removes the prefixes of LookupKey from model
attributes so that we can reuse the same strings across smart class
params and smart variables. In addition to that it uses the label
attribute to ensure those are used

### DIFF
--- a/locale/model_attributes.rb
+++ b/locale/model_attributes.rb
@@ -267,57 +267,33 @@ _('Lookup key')
 # TRANSLATORS: "Table name" or "Table name|Column name" for error messages
 _('Variable lookup key')
 # TRANSLATORS: "Table name" or "Table name|Column name" for error messages
-_('VariableLookupKey|Avoid duplicates')
+_('LookupKey|Avoid duplicates')
 # TRANSLATORS: "Table name" or "Table name|Column name" for error messages
-_('VariableLookupKey|Default value')
+_('LookupKey|Default value')
 # TRANSLATORS: "Table name" or "Table name|Column name" for error messages
-_('VariableLookupKey|Description')
+_('LookupKey|Description')
 # TRANSLATORS: "Table name" or "Table name|Column name" for error messages
-_('VariableLookupKey|Is param')
+_('LookupKey|Is param')
 # TRANSLATORS: "Table name" or "Table name|Column name" for error messages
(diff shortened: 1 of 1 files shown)


app/views/lookup_keys/_fields.html.erb
```diff
@@ -4,9 +4,9 @@
   <%= remove_child_link(_("Remove %s?") % (f.object.new_record? ? _("Variable") : f.object), f, { :class => 'btn btn-danger hide' }) if controller_name ==  "puppetclasses" && !is_param %>
   <fieldset>
     <h2><%= _("Parameter details") %></h2>
-    <%= text_f f, :key, :disabled => is_param, :size => "col-md-8" %>
+    <%= text_f f, :key, :disabled => is_param, :size => "col-md-8", :label => _('Key') %>
```
domcleal: No need to extract the strings again here, or even translate them, the form helpers do it automatically. They're likely to get out of sync.

locale/model_attributes.rb
```diff
@@ -267,58 +267,34 @@
 # TRANSLATORS: "Table name" or "Table name|Column name" for error messages
 _('Variable lookup key')
 # TRANSLATORS: "Table name" or "Table name|Column name" for error messages
-_('VariableLookupKey|Avoid duplicates')
+_('Avoid duplicates')
```
domcleal: These should be `Table name|Colum name` per the comment, i.e. `LookupKey|Avoid duplicates` or they won't match the automatic string lookups.

domcleal (changes requested):

dLobatog:
I've updated it according to your comments @domcleal  - but I'm probably missing something. If I do *not* use `:label` on the view, I don't see any translations. That's because PuppetclassLookupKey and VariableLookupKey are used still on the .po files. Should I change those too?

domcleal:
If you want to test it you could change them temporarily, but they shouldn't be changed directly - they would be updated when the POT is updated and synced from Transifex.

You appear to have squashed your changes into an unrelated commit.

dLobatog:
Fair enough - thanks @domcleal !

domcleal (approved):

domcleal:
Merged, thanks @dLobatog.

PR #4100, opened 2016-12-11
Fixes #17627 - remove tooltips after clicking back

tbrisker:
Turbolinks will keep the page cached as it is when a link with a tooltip
is clicked, so that clicking back will show the page with the tooltip
still active. This will make sure any tooltips are hidden when a page is
loaded from turbolinks' cache.


dLobatog:
Thanks @tbrisker !

(broken tests are a transient failure `cannot allocate memory - node..`)

PR #4101, opened 2016-12-12
fixes #17630 - show tooltips after deselecting selected item

kgaikwad:
For example, on Organization tab under edit user page
if user have already organization which is associated
with host then it will show tooltip - "This is used by host".
When you will select any other Organization and de-select it
again, on de-selection of item all existing tooltips
are disappeared.

### DIFF
--- a/app/assets/javascripts/jquery.multi-select.js
+++ b/app/assets/javascripts/jquery.multi-select.js
@@ -11,6 +11,7 @@ function multiSelectOnLoad(){
         var current_select = $(item).closest('.tab-pane').find('select[multiple]');
         current_select.data('descendants', null);
         $(current_select).multiSelect('refresh');
+        multiSelectToolTips();
       }
     })
   });


dLobatog:
@kgaikwad Maybe I'm doing something wrong but I cannot reproduce your setup. I have a host with orgs/locs, and users with those same orgs/locs yet I do not see the tooltip. Is your setup the latest develop?

kgaikwad:
@dLobatog ,
Thanks for reviewing the pull-request. 
My bad, I have skipped to write that when you edit an user who must be the **owner of that host**, is having an organization which is assigned to the same host then on edit page, you will see tooltip on that organization hover saying "This is used by host".

Now, select any other Organization and de-select it again, on de-selection of item all existing tooltips
are disappeared. 
In your case, please try to edit user "dlobatog" and let me know your observation.

Also, please do tell me if my understanding is wrong somewhere about it.

kgaikwad:
@dLobatog , I have modified and corrected description of the commit.

dLobatog:
Thanks @kgaikwad !

PR #4102, opened 2016-12-12
Fixes #17632 - Reports widget will not fail on ambiguous filter

tbrisker:
A user with a filter on a joined table that has a host_id column (such
as nics for example) will cause the report widget to break due to an
ambiguous column name. This fix makes sure the columns use the table
name and are correctly quoted.

### DIFF
--- a/test/unit/dashboard_test.rb
+++ b/test/unit/dashboard_test.rb
@@ -82,5 +82,14 @@ class DashboardTest < ActiveSupport::TestCase
         assert_equal 0, data.latest_events.length
       end
     end
+
+    test 'latest_events does not fail on ambiguous column name host_id' do
+      data = Dashboard::Data.new
+      #instead of going through the authorizer, force a join that will cause ambiguity.
+      ConfigReport.expects(:authorized).with(:view_config_reports).returns(ConfigReport.joins(:host => :interfaces))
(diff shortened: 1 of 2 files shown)


test/unit/dashboard_test.rb
```diff
@@ -82,5 +82,14 @@ class DashboardTest < ActiveSupport::TestCase
         assert_equal 0, data.latest_events.length
       end
     end
+
+    test 'latest_events does not fail on ambiguous column name host_id' do
+      data = Dashboard::Data.new
+      #instead of going through the authorizer, force a join that will cause ambiguity.
+      ConfigReport.stubs(:authorized).with(:view_config_reports).returns(ConfigReport.joins(:host => :interfaces))
```
domcleal: Nitpick, please use `expects` instead of `stubs` to prevent the test incorrectly succeeding if the implementation changes for any reason.

domcleal (changes requested):

test/unit/dashboard_test.rb
tbrisker: fixed.

domcleal (approved):

domcleal:
Merged, thanks @tbrisker.

PR #3404, opened 2016-04-07
Fixes #14479 - removed unwanted check for token param

lzap:
In function foreman_url we explicitly check for tokens and I can't find a
reason why we do that. This simply blocks provisioning when using Templating
Smart Proxy plugin.

PR #4103, opened 2016-12-12
Fixes #17636 - token no longer needed for proxied preview

lzap:
I removed this in renderer method, but we do have this twice in our codebase.

https://github.com/theforeman/foreman/pull/3404

I am aware this needs to be refactored, we are tracking this as a separate
ticket:

http://projects.theforeman.org/issues/8290


dLobatog:
Thanks @lzap !

PR #4104, opened 2016-12-12
Fixes #17641 - Use force when destroying VMs

agx:
There's no point in waiting several seconds for any guest utils to shut
down the VM when we want to get rid of it as quickly as possible.  This
also works around cases like

  2016-12-12T10:33:22 [app] [W] Failed to destroy a compute foo (VMware) instance bar.example.com: ToolsUnavailable: Der Vorgang kann nicht abgeschlossen werden, weil VMware Tools auf dieser virtuellen Maschine nicht ausgeführt wird.
   | RbVmomi::Fault: ToolsUnavailable: Der Vorgang kann nicht abgeschlossen werden, weil VMware Tools auf dieser virtuellen Maschine
  nicht ausgeführt wird.
   | /usr/share/foreman/vendor/ruby/2.1.0/gems/rbvmomi-1.8.2/lib/rbvmomi/connection.rb:61:in `parse_response'
   | /usr/share/foreman/vendor/ruby/2.1.0/gems/rbvmomi-1.8.2/lib/rbvmomi/connection.rb:90:in `call'
   | /usr/share/foreman/vendor/ruby/2.1.0/gems/rbvmomi-1.8.2/lib/rbvmomi/basic_types.rb:205:in `_call'
   | /usr/share/foreman/vendor/ruby/2.1.0/gems/rbvmomi-1.8.2/lib/rbvmomi/basic_types.rb:74:in `block (2 levels) in init'
   | /usr/share/foreman/vendor/ruby/2.1.0/gems/fog-1.34.0/lib/fog/vsphere/requests/compute/vm_power_off.rb:17:in `vm_power_off'
   | /usr/share/foreman/vendor/ruby/2.1.0/gems/fog-1.34.0/lib/fog/vsphere/models/compute/server.rb:94:in `stop'
   | /usr/share/foreman/vendor/ruby/2.1.0/gems/fog-1.34.0/lib/fog/vsphere/models/compute/server.rb:107:in `destroy'
   | /usr/share/foreman/app/models/compute_resource.rb:155:in `destroy_vm'
   | /usr/share/foreman/app/models/concerns/orchestration/compute.rb:164:in `delCompute'
   | /usr/share/foreman/app/models/concerns/orchestration.rb:168:in `execute'

where VMWare misdetecs the Openvm tools status.

### DIFF
--- a/app/models/compute_resources/foreman/model/vmware.rb
+++ b/app/models/compute_resources/foreman/model/vmware.rb
@@ -385,6 +385,13 @@ module Foreman::Model
       client.servers.new opts
     end
 
+    def destroy_vm(uuid)
+      find_vm_by_uuid(uuid).destroy :force => true
+    rescue ActiveRecord::RecordNotFound
+      # if the VM does not exists, we don't really care.
+      true
+    end
+
     # === Power on
     #
     # Foreman will try and start this vm after clone in a seperate request.


app/models/compute_resource.rb
```diff
@@ -184,7 +184,7 @@ def create_vm(args = {})
   end
 
   def destroy_vm(uuid)
-    find_vm_by_uuid(uuid).destroy
+    find_vm_by_uuid(uuid).destroy :force => true
```
dLobatog: @agx If this is just for VMWare - mind to move the call to `vmware.rb`? It'd be as simple as rewriting the method in that file, you can see other methods like `new_vm` etc... have been overridden there already.

The reason is just that we cannot ensure all fog compute resource providers accept this option (or other options at all). Azure for example does not accept any argument on .destroy calls

dLobatog (changes requested):
Please check out the comment inline, should be ready to merge after that. Thanks @agx !

agx:
On Wed, Dec 14, 2016 at 01:00:34AM -0800, Daniel Lobato García wrote:
> dLobatog requested changes on this pull request.
> 
> Please check out the comment inline, should be ready to merge after that. Thanks @agx !
> 
> > @@ -184,7 +184,7 @@ def create_vm(args = {})
>    end
>  
>    def destroy_vm(uuid)
> -    find_vm_by_uuid(uuid).destroy
> +    find_vm_by_uuid(uuid).destroy :force => true
> 
> @agx If this is just for VMWare - mind to move the call to `vmware.rb`? It'd be as simple as rewriting the method in that file, you can see other methods like `new_vm` etc... have been overridden there already.
> 
> The reason is just that we cannot ensure all fog compute resource providers accept this option (or other options at all). Azure for example does not accept any argument on .destroy calls

The ones I looked at in fog take a hash so I thought it might be good to
pass this for all resources (some like vmfusion even accept force). If
there are resources that don't accept an options hash at all we should
probably fix that?

agx:
@dLobatog  patch updated as you requested

dLobatog (approved):

dLobatog:
Thanks - @agx the problem with using it in general for all compute resources is that Foreman relies on fog (https://github.com/fog/) for these operations. We'd have to go over each of the ones that do not support `:force`, add it, then wait until a new release, then incorporate the release in Foreman. 

I'll wrap your commit to the 72nd column & merge, thanks again

timogoebel:
@agx : Could you please "fix" the commit message? I'll he happy to quickly test this after and we're ready to 🚢 .

agx:
On Wed, Dec 21, 2016 at 12:36:24AM -0800, Timo Goebel wrote:
> @agx : Could you please "fix" the commit message? I'll he happy to quickly test this after and we're ready to 🚢 .

Will do but how? Line breaking an error message that has no line breaks
originally?

timogoebel:
@agx : Daniel has already merged this, so nevermind. I was a few seconds too slow. :-)
Thanks for your contribution!

PR #4105, opened 2016-12-13
Fixes #17653 - Allow all parameters api to unhide values

orrabin:


app/models/lookup_keys/puppetclass_lookup_key.rb
```diff
@@ -20,6 +20,10 @@ class PuppetclassLookupKey < LookupKey
 
   scope :smart_class_parameters, -> { joins(:environment_classes).readonly(false) }
 
+  def can_edit?
```
tbrisker: This name is a bit confusing - the LK isn't doing the editing. Perhaps change it to `editable_by_user?` or something similar?

app/models/lookup_keys/puppetclass_lookup_key.rb
```diff
@@ -20,6 +20,10 @@ class PuppetclassLookupKey < LookupKey
...
+    User.current.allowed_to?(:edit_external_parameters)
```
tbrisker: This checks if the user has that permission, but what if it is only limited to certain parameters?

app/views/api/v2/common_parameters/base.json.rabl
```diff
@@ -1,3 +1,10 @@
 object @common_parameter
 
-attributes :id, :name, :value
+attributes :id, :name
+
+if params["show_hidden"] && User.current.allowed_to?(:edit_params)
```
tbrisker: this is quite repetitive, perhaps extract to a concern that is used by both params and lks?

test/unit/rabl_test.rb
```diff
@@ -3,14 +3,14 @@
 
 class RablTest < ActiveSupport::TestCase
   test 'render of single template' do
-    @domain = FactoryGirl.build(:domain)
```
tbrisker, line 6: Are these changes related or added by mistake to this commit?

tbrisker (changes requested):
Nice work on the test coverage! just a few fairly minor inline comments.

test/unit/rabl_test.rb
orrabin: Because it's only testing the rendering I didn't want to make changes to fit the new permissions so I changed it to `medium` which doesn't have parameters.

app/views/api/v2/common_parameters/base.json.rabl
orrabin: The concern would need to know when to use view and default_view and which permission it uses.

orrabin:
@tbrisker thanks for your comments, I addressed most of them, resolving the duplication would need more `if`s then it's worth in my opinion  but I don't mind changing it.

app/models/lookup_keys/puppetclass_lookup_key.rb
```diff
@@ -20,6 +20,10 @@ class PuppetclassLookupKey < LookupKey
 
   scope :smart_class_parameters, -> { joins(:environment_classes).readonly(false) }
 
+  def editable_by_user?
+    PuppetclassLookupKey.authorized(:edit_external_parameters).present? && PuppetclassLookupKey.authorized(:edit_external_parameters).find(self.id)
```
tbrisker: This will cause two queries on the DB, better to do something like:
```ruby
PuppetclassLookupKey.authorized(:edit_external_parameters).where(:id => id).exists?
```
which will generate a much simpler query.

app/models/lookup_value.rb
```diff
@@ -1,10 +1,11 @@
 class LookupValue < ActiveRecord::Base
   include Authorizable
   include PuppetLookupValueExtensions
+  include HiddenValue
 
   validates_lengths_from_database
   audited :associated_with => :lookup_key
-  delegate :hidden_value?, :hidden_value, :to => :lookup_key, :allow_nil => true
+  delegate :hidden_value?, :editable_by_user?, :to => :lookup_key, :allow_nil => true
```
tbrisker, line 8: no need for the `allow_nil` here, these are not attributes now

app/views/api/v2/common_parameters/base.json.rabl
```diff
@@ -1,3 +1,11 @@
 object @common_parameter
 
-attributes :id, :name, :value
+attributes :id, :name
+
+authorization = Parameter.authorized(:edit_params)
+if params["show_hidden"] && @common_parameter.present? && authorization.present? && authorization.find(@common_parameter.id)
```
tbrisker: how about making an `editable_by_user?` method on parameter model as well?

tbrisker (changes requested):
I still think that removing the duplication in the api is worth it, and if you create a `editable_by_user` method for parameters as well it should reduce the need for ifs (as you would just call it on whatever object)

app/models/lookup_value.rb
orrabin: if the lookup_key is nil and I remove the allow_nil there will be an error.

app/views/api/v2/common/show_hidden.json.rabl
```diff
@@ -0,0 +1,7 @@
+object @parameter
+
+if params["show_hidden"] == "true" && @object.present? && @object.editable_by_user?
+  attribute locals[:value].to_sym
```
tbrisker: why not pass the field name as a symbol instead of passing as a string and calling `to_sym`?

app/views/api/v2/override_values/base.json.rabl
```diff
@@ -3,3 +3,7 @@ object @override_value
 attributes :id, :match, :value, :omit
 # compatibility
 attribute :omit => :use_puppet_default
+
+node do
+  partial("api/v2/common/show_hidden", :locals => {:value => 'value'}, :object => @object)
+end
```
tbrisker: nitpick - missing newline at EOF

app/views/api/v2/common/show_hidden.json.rabl
```diff
@@ -0,0 +1,7 @@
+object @parameter
+
+if params["show_hidden"] == "true" && @object.present? && @object.editable_by_user?
```
tbrisker: is the `@object.present?` check still needed? wouldn't a missing object cause a failure anyways when attempting to access the attribute?

tbrisker (changes requested):
Almost done, just a few more minor comments inline

tbrisker:
Thanks @orrabin !

PR #4109, opened 2016-12-13
Fixes #17666 - prevent migration failures on counter cache

tbrisker:
Removal of counter cache causes migrations relying on the
`reset_counters` method to fail. The migrations are changes to not
update the counters (which are now removed anyways).

### DIFF
--- a/db/migrate/20140314004243_add_counter_caches.rb
+++ b/db/migrate/20140314004243_add_counter_caches.rb
@@ -3,59 +3,27 @@ class AddCounterCaches < ActiveRecord::Migration
     # Architectures
     add_column :architectures, :hosts_count, :integer, :default => 0
     add_column :architectures, :hostgroups_count, :integer, :default => 0
-    Architecture.all.each do |a|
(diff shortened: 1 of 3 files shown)


ohadlevy (commented):
Looks fine, but will this also address the linked bz? Afaiu counter cache still exists at that stage.

tbrisker:
@ohadlevy the reason this was hit in the BZ is that satellite 6.2.5 includes the counter cache removal. This will however hit foreman users as well if attempting to upgrade several releases at once, since the `reset_counters` method will fail to run with the current models (as the relations have no counter_cache defined on them).

mccun934:
@tbrisker so if a user goes from Foreman 1.7 -> 1.13 they most likely will hit this right?

tbrisker:
@mccun934 even 1.10->1.13, the last of these migrations was only included in 1.11

lzap:
Thanks, this indeed helps to mitigate the upgrade hop issue.

PR #4110, opened 2016-12-14
Fixes #17670 - fixed empty input for fact upload

lzap:
Also improved API doc.

### DIFF
--- a/app/controllers/api/v2/hosts_controller.rb
+++ b/app/controllers/api/v2/hosts_controller.rb
@@ -250,14 +250,14 @@ Return the host's compute attributes that can be used to create a clone of this 
         render_message(e.to_s, :status => :unprocessable_entity)
       end
 
-      api :POST, "/hosts/facts", N_("Upload facts for a host, creating the host if required")
+      api :POST, "/hosts/facts", N_("Upload facts for a host, creating the host if required (bypasses username/password authentication and validates trusted hosts setting or known smart proxies based on client cert CN or PTR record)")
       param :name, String, :required => true, :desc => N_("hostname of the host")
       param :facts, Hash,      :required => true, :desc => N_("hash containing the facts for the host")
       param :certname, String, :desc => N_("optional: certname of the host")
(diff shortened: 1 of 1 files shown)


app/controllers/api/v2/hosts_controller.rb
```diff
@@ -250,14 +250,14 @@ def boot
         render_message(e.to_s, :status => :unprocessable_entity)
       end
 
-      api :POST, "/hosts/facts", N_("Upload facts for a host, creating the host if required")
+      api :POST, "/hosts/facts", N_("Upload facts for a host, creating the host if required (bypasses username/password authentication and validates trusted hosts setting or known smart proxies based on client cert CN or PTR record)")
```
domcleal, line 253: 1. This seems unrelated to the ticket?
1. Use a description field for this, not the method title.
1. Also applies to other APIs such as report uploads, those should be updated at the same time.

app/controllers/api/v2/hosts_controller.rb
```diff
@@ -250,14 +250,14 @@ def boot
...
       param :name, String, :required => true, :desc => N_("hostname of the host")
       param :facts, Hash,      :required => true, :desc => N_("hash containing the facts for the host")
       param :certname, String, :desc => N_("optional: certname of the host")
       param :type, String,     :desc => N_("optional: the STI type of host to create")
 
       def facts
-        @host = detect_host_type.import_host params[:name], params[:facts][:_type] || 'puppet', params[:certname], detected_proxy.try(:id)
+        @host = detect_host_type.import_host params[:name], (params[:facts] || {})[:_type] || 'puppet', params[:certname], detected_proxy.try(:id)
```
domcleal, line 260: 1. Needs a functional test.
1. This should probably raise a 4xx error if the input is missing, since it's a required field - not silently accept it.

domcleal (changes requested):
Unsure that the proposed change is the correct behaviour - it implies a 2xx success response. While better than a 5xx response, a 4xx response seems more apt.

PR #4111, opened 2016-12-14
fixes #6150 - users need taxonomy added on "all Users"

kgaikwad:
With this commit, if "All users" is ticked then
locations and organizations are added to users. On edit user page,
you can not edit this taxonomy as they are disabled with tooltip
"Select all option enabled for this taxonomy".

Likewise, it will fix 'select all' option functionality for other ignore_types as well.

### DIFF
--- a/app/models/concerns/taxonomix.rb
+++ b/app/models/concerns/taxonomix.rb
@@ -12,6 +12,7 @@ module Taxonomix
       :through => TAXONOMY_JOIN_TABLE, :source => :taxonomy,
       :validate => false
     after_initialize :set_current_taxonomy
+    before_save :add_taxonomies_with_ignore_type_enabled
 
     scoped_search :relation => :locations, :on => :name, :rename => :location, :complete_value => true, :only_explicit => true
     scoped_search :relation => :locations, :on => :id, :rename => :location_id, :complete_enabled => false, :only_explicit => true, :validator => ScopedSearch::Validators::INTEGER
@@ -83,7 +84,6 @@ module Taxonomix
     # Passing a nil or [] value as taxonomy equates to "Any context".
     # Any other value will be understood as 'IDs available in this taxonomy'.
     def inner_ids(taxonomy, taxonomy_class, inner_method)
-      return unscoped.pluck("#{table_name}.id") if taxonomy_class.ignore?(to_s)
       return inner_select(taxonomy, inner_method) if taxonomy.present?
       return [] unless User.current.present?
       # Any available taxonomy to the current user
@@ -209,6 +209,22 @@ module Taxonomix
         errors.add key, _("Invalid %{assoc} selection, you must select at least one of yours and have '%{perm}' permission.") % { :assoc => _(assoc), :perm => "assign_#{assoc}" }
       end
     end
+  end
+
+  def add_taxonomies_with_ignore_type_enabled
+    Taxonomy.enabled_taxonomies.each do |taxonomy_pluralize_str|
+      taxonomy_class_name = taxonomy_pluralize_str.classify
+      obj_ids = taxonomy_class_name.constantize.taxonomy_ids_by_ignore_type(self.class.to_s)
+      assign_taxonomy_ids(taxonomy_class_name, obj_ids) if obj_ids.present?
+    end
+  end
+
+  def assign_taxonomy_ids(taxonomy_class_name, taxonomy_ids_by_ignore_type)
+    taxonomy_ids_meth = taxonomy_class_name.underscore + '_ids'
+    existing_obj_ids = self.send(taxonomy_ids_meth)
+    return if (taxonomy_ids_by_ignore_type - existing_obj_ids).blank?
+    taxonomy_ids_by_ignore_type.concat(existing_obj_ids).uniq!
+    self.send("#{taxonomy_ids_meth}=", taxonomy_ids_by_ignore_type)
   end
 
   protected
--- a/app/models/concerns/taxonomix.rb
+++ b/app/models/concerns/taxonomix.rb
@@ -222,11 +222,11 @@ module Taxonomix
 
   def assign_taxonomy_ids(taxonomy_class_name, taxonomy_ids_by_ignore_type)
     taxonomy_ids_meth = taxonomy_class_name.underscore + '_ids'
-    existing_obj_ids = self.send(taxonomy_ids_meth)
+    existing_obj_ids = send(taxonomy_ids_meth)
     return if (taxonomy_ids_by_ignore_type - existing_obj_ids).blank?
 
     taxonomy_ids_by_ignore_type.concat(existing_obj_ids).uniq!
-    self.send("#{taxonomy_ids_meth}=", taxonomy_ids_by_ignore_type)
+    send("#{taxonomy_ids_meth}=", taxonomy_ids_by_ignore_type)
   end
 
   protected
--- a/app/models/concerns/taxonomix.rb
+++ b/app/models/concerns/taxonomix.rb
@@ -212,10 +212,11 @@ module Taxonomix
   end
 
   def add_taxonomies_with_ignore_type_enabled
-    Taxonomy.enabled_taxonomies.each do |taxonomy_pluralize_str|
-      taxonomy_class_name = taxonomy_pluralize_str.classify
+    ['location', 'organization'].map(&:classify).each do |taxonomy_class_name|
       obj_ids = taxonomy_class_name.constantize.taxonomy_ids_by_ignore_type(self.class.to_s)
-      assign_taxonomy_ids(taxonomy_class_name, obj_ids) if obj_ids.present?
+      if obj_ids.present?
+        assign_taxonomy_ids(taxonomy_class_name, obj_ids)
+      end
     end
   end
 
@@ -223,6 +224,7 @@ module Taxonomix
     taxonomy_ids_meth = taxonomy_class_name.underscore + '_ids'
     existing_obj_ids = self.send(taxonomy_ids_meth)
     return if (taxonomy_ids_by_ignore_type - existing_obj_ids).blank?
+
     taxonomy_ids_by_ignore_type.concat(existing_obj_ids).uniq!
     self.send("#{taxonomy_ids_meth}=", taxonomy_ids_by_ignore_type)
   end
--- a/app/views/taxonomies/_loc_org_tabs.html.erb
+++ b/app/views/taxonomies/_loc_org_tabs.html.erb
@@ -5,12 +5,14 @@
 
 <% if show_location_tab? %>
   <div class="tab-pane" id="locations">
+    <% location_ids_with_select_all_for_ignore_type = Location.taxonomy_ids_by_ignore_type(obj.class.to_s) %>
     <%= location_selects f, obj.used_or_selected_location_ids,
-            {:disabled => obj.used_location_ids }.merge!(select_options),
+            {:disabled => obj.used_location_ids.concat(location_ids_with_select_all_for_ignore_type)}.merge!(select_options),
             {'data-descendants' => obj.children_of_selected_location_ids.to_json,
-             'data-useds' => obj.used_location_ids.to_json }.merge!(html_options[:location]) %>
+             'data-useds' => obj.used_location_ids.to_json,
+             'data-used-all' => location_ids_with_select_all_for_ignore_type.to_json }.merge!(html_options[:location]) %>
     <% if obj.kind_of? User %>
-      <%= select_f f, :default_location_id, (obj.admin? ? Location.all : obj.my_locations), :id, :title,
+       <%= select_f f, :default_location_id, (obj.admin? ? Location.all : obj.my_locations), :id, :title,
                     { :include_blank => true }, { :label => _('Default on login') } %>
     <% end %>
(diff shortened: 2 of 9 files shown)


kgaikwad:
I have updated the pull-request. 
codeclimate is showing complexity warnings for taxonomix.rb file.
But I am not getting how to resolve it.
Please provide me some pointers so that I will do further modifications.

kgaikwad:
I have updated the pull-request with modifications to solve complexity warning for file taxonomix.rb

ares:
Thanks @kgaikwad, could you please rebase the PR?

kgaikwad:
@ares ,
Yes, I will update PR soon.

kgaikwad:
@ares , 
Done. I have updated the PR.
Please let me know if any further changes required.

app/models/user.rb
```diff
@@ -560,6 +561,19 @@ def verify_current_password
     end
   end
 
+  # assign all taxonomies for which ignore_type includes 'User' resource
+  def add_select_all_taxonomies
+    if self.new_record?
```
ares: the condition could be moved to `after_initialize :add_select_all_taxonomies, :if => :new_record?` so it's more obvious it does not happen for existing users

app/views/taxonomies/_loc_org_tabs.html.erb
```diff
@@ -18,10 +20,12 @@
...
 <% if show_organization_tab? %>
   <div class="tab-pane" id="organizations">
+    <% enabled_select_all_organization_ids = obj.enabled_select_all_ids_thr_type(Organization) %>
     <%= organization_selects f, obj.used_or_selected_organization_ids,
-            {:disabled => obj.used_organization_ids }.merge!(select_options),
+            {:disabled => obj.used_organization_ids.concat(enabled_select_all_organization_ids)}.merge!(select_options),
             {'data-descendants' => obj.children_of_selected_organization_ids.to_json,
-             'data-useds' => obj.used_organization_ids.to_json }.merge!(html_options[:organization]) %>
+             'data-useds' => obj.used_organization_ids.to_json,
+	     'data-usedall' => enabled_select_all_organization_ids.to_json }.merge!(html_options[:organization]) %>
```
ares: indentation is a bit off, 
also, `usedall` should be split to `data-used-all`

lib/tasks/taxonomy.rake
```diff
@@ -0,0 +1,20 @@
+namespace :taxonomy do
+  desc <<-END_DESC
+	This task will update organization & location records with all users for which ignore_types includes 'User' resource.
+  END_DESC
+  task :update_taxonomy => :environment do
```
ares, line 5: the rake task should probably run in `User.anonymous_admin do ... end` (required since recent changes, probably was not required back then)

app/models/user.rb
```diff
@@ -100,6 +100,7 @@ def self.name_format
 
   after_create :welcome_mail
   after_create :set_default_widgets
+  after_initialize :add_select_all_taxonomies
```
ares: This seems to work for new users but I think we should do similar check for existing ones, e.g. when I update some user through API and select subset of organizations. I can't do this through WebUI because of disabled fields but API/hammer could still do it.

ares (changes requested):
Thanks Kavita, patches improving this area are highly appreciated. In general I think this improves the current situation but does not solve it completely.

I think that it could get out of sync again in case a new organization or location ignoring users is added after some user already exists. Perhaps some callback on Taxonomy model should be added that will make sure all resources are linked if they're ignored.

Last but not least, this fixes the issue for User model only. Could it be generalized in taxonomix.rb so it applies to all taxable resources? If it's implemented this way and we can rely on having proper relations in DB between resource and taxonomy, I think it could simplify the [existing code](https://github.com/theforeman/foreman/blob/develop/app/models/concerns/taxonomix.rb#L96)

I think more tests would be helpful since this is area is prone to errors. Covering all new methods and a test verifying that after taxonomy starts to ignore a resource class, every such `$resource.taxonomies` includes this taxonomy

kgaikwad:
@ares ,

Thank you for explaining in brief. As per your suggestions, I have modified the code and added before_save hook for all ignore_types as well as on taxonomy.
I have queries regarding few points:
- As you mentioned that code modifications will simplify the [existing code](https://github.com/theforeman/foreman/blob/develop/app/models/concerns/taxonomix.rb#L96).
Instead of querying taxable_class with new changes we can query on taxonomy itself.
i.e. `taxonomy.send("#{base_class.name.underscore}_ids")`
Please correct me if I misunderstood something here.

-  In [taxonomy.rb](https://github.com/theforeman/foreman/blob/develop/app/models/taxonomy.rb#L155),
`_ids` methods are overwritten where ignore_types are returned by querying on type class instead of DB relations. For users, all with `unscoped.except_admin.except_hidden` and for other types, it simply returns all records.
It may cause conflicts in above mentioned case and other areas.
Please guide me on this.

Pending Work - I will add test cases once code changes done.

app/models/concerns/taxonomix.rb
```diff
@@ -12,6 +12,7 @@ module Taxonomix
              :through => TAXONOMY_JOIN_TABLE, :source => :taxonomy,
              :validate => false
     after_initialize :set_current_taxonomy
+    before_save :add_taxonomies_having_select_all_enabled
```
orrabin: Why do this every save and not only on create?
Isn't it already saved by the taxonomy when it's saved for all existing taxable taxonomies?

app/models/concerns/taxonomix.rb
```diff
@@ -180,12 +181,17 @@ def used_organization_ids
...
   end
 
+  def enabled_select_all_ids_thr_type(which_type)
+    return [] unless [Location, Organization].include?(which_type)
+    which_type.taxonomies_by_ignore_type(self.class.to_s).map(&:id)
+  end
+
   def used_or_selected_location_ids
-    (location_ids + used_location_ids).uniq
+    (location_ids + used_location_ids + enabled_select_all_ids_thr_type(Location)).uniq
```
orrabin: After the association is saved the ids should already exist in `location_ids`.

app/models/concerns/taxonomix.rb
kgaikwad: @orrabin,

Previously I have added it for new_objects only. 
Later in discussion with @ares found that from web-ui user can't modify taxonomies as field will be disabled but using API/hammer, there might be possibility user can change taxonomy.

orrabin:
[test foreman]
[test katello]

app/models/concerns/taxonomix.rb
ares: I suggested this in https://github.com/theforeman/foreman/pull/4111#discussion_r128496408, the other option is to disable this from hammer/API. We could add a validation check and disable update if some of linked taxonomies "all" has been checked. But we'd need to load all taxonomies in this case too. If there's no change we don't fire any extra SQL call. Maybe the performance could be slightly improved by using `organization_ids=` instead of `organizations=` but we need to load them anyway.

app/models/concerns/taxonomix.rb
```diff
@@ -224,6 +230,16 @@ def ensure_taxonomies_not_escalated
     end
   end
 
+  def add_taxonomies_having_select_all_enabled
+    [Organization, Location].each do |taxonomy_class|
```
ares: Please use `Taxonomy.enabled_taxonomies` if possible, some users might have enabled only a subset of taxonomies.

kgaikwad:
@ares and @orrabin ,

I have modified code as per your comments.
I have added some additional changes and test cases to `taxonomy_test.rb`.
Please review and feel free to add comments if any additional changes required.

app/models/concerns/taxonomix.rb
kgaikwad: Currently I am using `organizations=` only, I will try to modify it to `organization_ids=`  and will check if I can minimize any SQL call.

kgaikwad: @ares and @orrabin,

I have now used `organization_ids=` instead of `organizations=`.
Previously I thought that disabling update from hammer/API will be good to have to make data consistent. To add this change, we need to load taxonomies again like you mentioned. 

So I will prefer to go with update on every save. I have added check that if taxonomies are not already assigned to an object in that case only update will happen for taxonomies.
And I think current code changes will take care of data consistency.

Please let me know if I misunderstood anything.

kgaikwad:
@ares ,
While working on one issue that current code is not working properly for default organization and location. 
As for default organization and location, `taxable_taxonomies` gives all records but value for ignore_types is empty. In this case, it fails to save these taxonomies as in callbacks I have assigned records based on ignore_types.
Please let me know what would be the best way to handle it. 
Do I need to handle this default condition in callbacks?

xprazak2:
@ares, any suggestions how to continue?

ares:
I'd be happy to rereview, sorry I missed the last version completely. @kgaikwad would you be able to rebase? I think this was really good change.

jhutar:
Hello. I have attempted to apply this patch to Sat6.4.0 snap10 and besides I do not know how to apply *.js changes (do we have some docs here please?), migration failed:

```
[root@sat640snap10 foreman]# foreman-rake db:migrate
/usr/share/foreman/lib/foreman.rb:8: warning: already initialized constant Foreman::UUID_REGEXP
/usr/share/foreman/lib/foreman.rb:8: warning: previous definition of UUID_REGEXP was here
/usr/share/foreman/lib/core_extensions.rb:182: warning: already initialized constant ActiveSupport::MessageEncryptor::DEFAULT_CIPHER
/opt/theforeman/tfm-ror51/root/usr/share/gems/gems/activesupport-5.1.6/lib/active_support/message_encryptor.rb:22: warning: previous definition of DEFAULT_CIPHER was here
== 20170922082617 UpdateTaxonomiesWithSelectedIgnoreTypesObjects: migrating ===
-- This Migration will take time for updating organization & location records with all objects of selected ignore_types.
Failed to save Organization 1- ["Environments expecting environments used by hosts or inherited (check mismatches report).", "Domains expecting domains used by hosts or inherited (check mismatches report)."]
== 20170922082617 UpdateTaxonomiesWithSelectedIgnoreTypesObjects: migrated (1.3069s)
```

Any idea on what might be wrong?

jhutar:
Reported possibly related https://bugzilla.redhat.com/show_bug.cgi?id=1599284

kgaikwad:
@jhutar,
Thank you for looking into this PR and addressing Bugzilla issue. 
I will check it after rebase.
As this change is related to taxanomies, there are many combinations. 
After all changes when I tested it, migration ran well on my instance but ones I got random error that I am not able to reproduce it later.
It might be same or different but I would suggest this patch needs some thorough testing.

bkearney:
@kgaikwad is this still in active development?

kgaikwad:
@bkearney,
There are many scenarios with taxonomies to test. I have faced data loss issues with my changes for few scenarios.
Currently, looking into other tasks so yes, it is not active.
But I would like to fix those issues.

ares:
@kgaikwad hello, is this something we could revisit after 1.24 branching? Do you think you'll have capacity to finish it? Improvements in this area are more than welcome.

kgaikwad:
> @kgaikwad hello, is this something we could revisit after 1.24 branching? Do you think you'll have capacity to finish it? Improvements in this area are more than welcome.

Thank you @ares. 
Yeah, after 1.24 branching we can revisit this as I am not sure about how many changes will be required to fix this "all <resources>" options and what will be the impact on current behaviour.

lib/tasks/taxonomy.rake
kgaikwad: Tried the scenario with newly created DB, this migration fails because user `ANONYMOUS_ADMIN` not being present at that time. Added condition to handle this in rake task but is there any better way to run the rake task in case this user is not present?

kgaikwad:
@ares, 
I have done with rebase.
Tests are green. As I mentioned [above](https://github.com/theforeman/foreman/pull/4111/files#r360877161), I would like to know if any additional migration or rake task changes are required so that it will run smoothly on any environment.

EDIT - I have tested with few general scenarios:
- Tested with organization/location where select all option is enabled for few ignore_types.
- Tested behavior of child organization/location when respective parent is having this "select all" option enabled against ignore_types.
- DB changes as per UI modification while creating/updating records having organizations/locations with such options.
If any other complex scenarios that I should test then please let me know.

lib/tasks/taxonomy.rake
ares: The admin user should exist when this task is needed, since admins are seeded before taxonomies. If the database is not seeded at all, this task shouldn't be needed and can be skipped. The condition could check `if Organization.any?` or `if Location.any?` but it's similar to what you have already. So I guess it can remain as is.

kgaikwad: Okay then I am keeping the condition as it is.
Thank you @ares

bkearney:
@Ares are we still waiting on the updates from @kgaikwad

kgaikwad:
@bkearney, @ares,
Had conflicts as few days back fixed "Style/RedundantSelf cop" to master. Resolved and done with rebase.

tbrisker:
closing for now, thanks @kgaikwad !

PR #4112, opened 2016-12-15
Fixes #6952 - Set hostgroup after provisioning from pxe

amirfefer:
After hostgroup provisioning, when the machine has registered
with foreman, it should be in the same hostgroup.
to achieve this behavier, the [referenced](https://github.com/theforeman/community-templates/pull/327) snippet should be added
to hostgroup provisioning template.

### DIFF
--- a/app/models/host/base.rb
+++ b/app/models/host/base.rb
@@ -120,6 +120,7 @@ module Host
 
       facts[:domain] = facts[:domain].downcase if facts[:domain].present?
 
+      hostgroup_title = facts.delete(:foreman_hostgroup)
       time = facts[:_timestamp]
       time = time.to_time if time.is_a?(String)
       self.last_compile = time if time
@@ -128,6 +129,7 @@ module Host
       importer = FactImporter.importer_for(type).new(self, facts)
       importer.import!
 
+      set_hostgroup(hostgroup_title) if hostgroup_title.present? && new_record?
       save(:validate => false)
       set_taxonomies(facts)
       populate_fields_from_facts(facts, type)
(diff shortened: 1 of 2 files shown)


app/models/host/base.rb
```diff
@@ -120,6 +120,7 @@ def import_facts(facts)
 
       facts[:domain].try(:downcase!)
 
+      hostgroup_title = facts.delete(:foreman_hostgroup)
```
dLobatog, line 123: This fact isn't standard, from what I read on the issue description it's a custom fact. I think the functionality is neat, but please add docs to this after it's merged

app/models/host/base.rb
```diff
@@ -209,6 +211,10 @@ def ==(comparison_object)
         comparison_object.id == id
     end
 
+    def set_hostgroup(title)
+      self.hostgroup = Hostgroup.find_by_title(title)
```
dLobatog, line 215: Call `set_hostgroup_defaults` after, so that the attributes from the host group are applied

dLobatog (changes requested):
It's nearly done. Just call `set_hostgroup_defaults` and please add something to theforeman.org documenting what this does, as `foreman_hostgroup` isn't provided by Ansible, Puppet etc... 

Thanks @amirfefer

amirfefer:
Thanks for your review @dLobatog, I've added `set_hostgroup_defaults` call, and a related test.

dLobatog (approved):

dLobatog:
@amirfefer I'll fix the HoundCI qualms on merge, remember to write some docs for this on https://github.com/theforeman/theforeman.org/ please. Thanks!

dLobatog:
Merged as 9883149acbb88bd59695b92207933c074a72c427 , thanks @amirfefer !

domcleal:
9883149 looks significantly different to the reviewed commit 604a651, adding `hostgroup` to `attributes_to_import_from_facts` and removing the `set_hostgroup` method + calls. Why isn't it the same?

PR #4113, opened 2016-12-15
Fixes #17678 - APIdoc - URL desc add missing CR types

swapab:


app/controllers/api/v2/compute_resources_controller.rb
```diff
@@ -30,7 +30,7 @@ def show
         param :compute_resource, Hash, :required => true, :action_aware => true do
           param :name, String, :required => true
           param :provider, String, :desc => N_("Providers include %{providers}") # values are defined in apipie initializer
-          param :url, String, :desc => N_("URL for Libvirt, oVirt, and OpenStack")
+          param :url, String, :desc => N_("URL for Docker, Libvirt, oVirt, OpenStack and Rackspace")
```
dLobatog: Sorry - Docker is not a compute resource provided by Foreman, but by https://github.com/theforeman/foreman-docker/ , it should be added there

dLobatog (changes requested):
Please take a look at the comment inline - Rackspace could probably be left here but not Docker. Ideally the URL can be overridden through plugins. Thanks @swapab

swapab:
@dLobatog 
> Ideally the URL can be overridden through plugins.

Does this mean extending `Api::V2::ComputeResourcesController` in foreman-docker plugin ?

Overriding controller just for document correction doesn't sound good to me.
I think defining a class method `ComputeResource::providers_requiring_url` makes more sense.

```ruby
  def self.providers_requiring_url
    "Libvirt, oVirt, OpenStack and Rackspace"
  end
```
And then override `ComputeResource` class in foreman-bootdisk plugin.

ares:
currently, it's impossible to extend apipie doc from plugins, feel free to upvote http://projects.theforeman.org/issues/16630, I added #17678 on its list of blocked items

katello addresses this by overriding the whole definition but that does not work if two or more plugins try to do the same

app/models/compute_resource.rb
```diff
@@ -73,6 +73,10 @@ def self.providers
     end
   end
 
+  def self.providers_requiring_url
+    "Libvirt, oVirt, OpenStack and Rackspace"
```
dLobatog: I believe this ought to be a translation friendly string , `_('Libvirt, oVirt, OpenStack and Rackspace')`

dLobatog (changes requested):

dLobatog:
@ares Any issues with providing a method that contains the string, which plugins can override, like this? Thanks @swapab for coming up quickly with a possible solution.

ares:
looks good, just note that apipie cache needs to be regenerated once we install docker plugin, I don't think this happens as part of the installation process so the manual rake task invocation is probably required

app/models/compute_resource.rb
swapab: > I believe this ought to be a translation friendly string , _('Libvirt, oVirt, OpenStack and Rackspace')

I thought it's been taken care by controller: https://github.com/theforeman/foreman/pull/4113/commits/361da397071cb23f15cb67b0a633958de1ee2fdc#diff-cf2965b610f2fa0c776c7129f985f26dR33

No worries, integral string translation is always better.

swapab:
@dLobatog Thanks for reviewing my PR.

@ares 
> just note that apipie cache needs to be regenerated once we install docker plugin

are you referring to some production/test setup? Should this instruction be part of documentation?

> I don't think this happens as part of the installation process so the manual rake task invocation is probably required

Again, I understand its `rake apipie:cache` is the rake you are referring to. Is there a post docker installation setup that needs to be updated?

_Do I need to run the rake and push apipie html files?_
But I don't see them in the current version. Maybe you can explain it a bit more.

ares:
> are you referring to some production/test setup? Should this instruction be part of documentation?

yes

> Again, I understand its rake apipie:cache is the rake you are referring to. Is there a post docker installation setup that needs to be updated?

correct, I think it wouldn't hurt to be mentioned for every plugin that extends API, so far I think it's in core installation manual https://theforeman.org/manuals/1.13/index.html#3.2.3InstallationScenarios, but that does not block this PR from my point of view...

dLobatog (approved):

dLobatog:
Thanks @swapab & @ares !

swapab:
Yes. Thanks for clearing things @ares.

On Dec 20, 2016 5:12 PM, "Marek Hulán" <notifications@github.com> wrote:

are you referring to some production/test setup? Should this instruction be
part of documentation?

yes

Again, I understand its rake apipie:cache is the rake you are referring to.
Is there a post docker installation setup that needs to be updated?

correct, I think it wouldn't hurt to be mentioned for every plugin that
extends API, so far I think it's in core installation manual
https://theforeman.org/manuals/1.13/index.html#3.2.3InstallationScenarios,
but that does not block this PR from my point of view...

—
You are receiving this because you were mentioned.
<https://github.com/theforeman/foreman/pull/4113#issuecomment-268222970>,
or mute the thread

PR #4114, opened 2016-12-15
Fixes #17679 - Fixing the style issues in Login Page

junezhang:
1, Based on [Terminology and Wording Style Guide](http://www.patternfly.org/styles/terminology-and-wording/#terms-and-words), Suggested to use a verb, we should say "Log In" as the button label for the login page.
2, The username should align the text box when the screen is smaller than 768px.
Please check the attachment picture.
screenshot from 2016-11-05 03-06-49

### DIFF
--- a/app/assets/stylesheets/login.scss
+++ b/app/assets/stylesheets/login.scss
@@ -58,6 +58,7 @@
 }
 .login-page .container .form-horizontal .control-label {
   font-weight: 400;
+  padding-left: 20px;
   text-align: left;
 }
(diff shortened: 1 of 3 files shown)


ohadlevy:
ok to test

ohadlevy:
@junezhang thanks! please note you would need to update the test as well, have a look at test/integration/user_test.rb:23

junezhang:
@ohadlevy Thanks very much for pointing out where needs to update. It's helpful for a newbie contributor. :-) I've updated it.

PR #4115, opened 2016-12-15
Fixes #17680 - templates_used host helper

dLobatog:
templates_used contains the names of all of the provisioning templates
kickstart used during a host build. This allows users to be able to show the
name of a provisioning template during PXEBoot, for example.

Submitted in behalf of Peter Vreman.

### DIFF
--- a/app/models/host/managed.rb
+++ b/app/models/host/managed.rb
@@ -72,7 +72,7 @@ class Host::Managed < Host::Base
       :image_build?, :pxe_build?, :otp, :realm, :param_true?, :param_false?, :nil?, :indent, :primary_interface,
       :provision_interface, :interfaces, :bond_interfaces, :bridge_interfaces, :interfaces_with_identifier,
       :managed_interfaces, :facts, :facts_hash, :root_pass, :sp_name, :sp_ip, :sp_mac, :sp_subnet, :use_image,
-      :multiboot, :jumpstart_path, :install_path, :miniroot, :medium, :bmc_nic
+      :multiboot, :jumpstart_path, :install_path, :miniroot, :medium, :bmc_nic, :templates_used
   end
 
   scope :recent,      ->(*args) { where(["last_report > ?", (args.first || (Setting[:puppet_interval] + Setting[:outofsync_interval]).minutes.ago)]) }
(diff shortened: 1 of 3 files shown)


app/models/host/managed.rb
```diff
@@ -828,6 +828,14 @@ def available_template_kinds(provisioning = nil)
     end.compact
   end
 
+  def templates_used
```
ares, line 826: this whole method can be written as one-liner
```ruby
Hash[available_template_kinds.map { |t| [t.template_kind_name, t.name] }]
```

but if you find it less readable, I don't insist

ares (changes requested):
could you please also add not only the test for host method but for renderer to make sure it can be rendered in template? the best place is probably here test/unit/foreman/renderer_test.rb

dLobatog:
@ares I've added a test for renderer.rb that tests the method can indeed be called on Host. The test doesn't check for the contents -that's the purpose of the unit test in Host itself- but it checks that one is allowed to run templates_used. I don't think testing for the content there would make much sense as it's writing the same test twice.

Thanks for taking a look!

test/unit/foreman/renderer_test.rb
```diff
@@ -221,4 +221,8 @@ def setup_safemode_renderer
     @renderer.instance_variable_set '@whatever_random_name', 'has_value'
     assert_equal({ :whatever_random_name => 'has_value' }, @renderer.send(:allowed_variables_mapping, [ :whatever_random_name ]))
   end
+
+  test 'templates_used is allowed to render for host' do
+    assert Safemode.find_jail_class(Host::Managed).allowed? :templates_used
```
ares, line 274: I think this test only verifies that safemode jail can list allowed methods. I agree it's not required to test the output but I'd like to have a test that verifies it can render without errors at least. Renderer is sort of a user API so I'd like to have it covered.

you could use something like this in block near line 151
```ruby
    test "#{renderer_name} should render templates used" do
      send "setup_#{renderer_name}"
      @renderer.host = FactoryGirl.build(:host)
      template = mock('template')
      template.stubs(:template).returns('<%= @host.templates_used %>')
      assert_nothing_raised do
        @renderer.unattended_render(template)
      end
    end
```

dLobatog:
@ares Updated with a test to check it renders correctly and displays the names

ares (approved):
Thanks @dLobatog, works well. I'll merge shortly, would you mind describing it at [TemplateWriting wiki](http://projects.theforeman.org/projects/foreman/wiki/templatewriting)?

ares:
Merged as 22439f753cbf76ad93cacdc5e48bbe1ce24f65c9

PR #4116, opened 2016-12-15
Fixes #17683 - Wrong resource type in hostgroup AJAX permission check

sean797:
Specifying `Host` causes this not to work on `Host::Discovered` and I guess others.

### DIFF
--- a/app/controllers/hosts_controller.rb
+++ b/app/controllers/hosts_controller.rb
@@ -671,7 +671,7 @@ class HostsController < ApplicationController
     return head(:not_found) unless @hostgroup
 
     @host = if params[:host][:id]
-              host = Host::Base.authorized(:view_hosts, Host).find(params[:host][:id])
+              host = Host::Base.authorized(:view_hosts).find(params[:host][:id])
               host = host.becomes Host::Managed
               host.attributes = host.apply_inherited_attributes(host_params)
               host
--- a/test/controllers/hosts_controller_test.rb
+++ b/test/controllers/hosts_controller_test.rb
@@ -1310,6 +1310,28 @@ class HostsControllerTest < ActionController::TestCase
     end
   end
 
+  test '#process_hostgroup works on Host subclasses' do
+    class Host::Test < Host::Base; end
+    user = FactoryGirl.create(:user, :with_mail, :admin => false)
+    FactoryGirl.create(:filter, :role => user.roles.first, :permissions => Permission.where(:name => [ 'edit_hosts', 'view_hosts' ]))
+    hostgroup = FactoryGirl.create(:hostgroup)
+    host = FactoryGirl.create(:host, :type => "Host::Test", :hostgroup => hostgroup)
+    host.stubs(:set_hostgroup_defaults)
+    host.stubs(:set_compute_attributes)
+    host.stubs(:architecture)
+    host.stubs(:operatingsystem)
+    host.stubs(:environment)
+    host.stubs(:domain)
+    host.stubs(:subnet)
+    host.stubs(:compute_profile)
+    host.stubs(:realm)
+    attrs = host_attributes(host)
+    attrs[:id] = host.id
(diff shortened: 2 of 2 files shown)


lzap:
This looks good to me, but I'd ask @ares for confirmation. Can you please fix the typo in the commit message?

app/controllers/hosts_controller.rb
```diff
@@ -671,7 +671,7 @@ def process_hostgroup
     return head(:not_found) unless @hostgroup
 
     @host = if params[:host][:id]
-              host = Host::Base.authorized(:view_hosts, Host).find(params[:host][:id])
+              host = Host::Base.authorized(:view_hosts).find(params[:host][:id])
```
ares, line 674: This will work since new resource type will now be `Host::Base` and when we search for collection of filters, we translate it back to `Host` at https://github.com/sean797/foreman/blob/bc929bc30b277e907ceda208773932e6a2a92a37/app/services/authorizer.rb#L174

Since it's not clear at first look, it would be great to add a test in which we'd create a new Host subclass and tested it with it. This way it would make sure it won't break discovery in future.

lzap:
So for the record, asking for one test here. Also renamed the PR name to more appropriate subject.

lzap:
@sean797 here is the test, sorry was a bit different than I guided you:

```ruby
  test '#process_hostgroup works on Host subclasses' do
    class Host::Test < Host::Base; end
    hostgroup = FactoryGirl.create(:hostgroup)
    host = FactoryGirl.create(:host, :type => "Host::Test", :hostgroup => hostgroup)
    host.stubs(:set_hostgroup_defaults)
    host.stubs(:set_compute_attributes)
    host.stubs(:architecture)
    host.stubs(:operatingsystem)
    host.stubs(:environment)
    host.stubs(:domain)
    host.stubs(:subnet)
    host.stubs(:compute_profile)
    host.stubs(:realm)
    attrs = host_attributes(host)
    attrs[:id] = host.id
    attrs[:hostgroup_id] = hostgroup.id
    xhr :put, :process_hostgroup, { :host => attrs }, set_session_user
    assert_response :success
  end
```

You need to put it *outside* of the contexts, a bit above (note the indentation) otherwise you will get expectation errors of Libvirt.

lzap:
Tests fail because: You need to put it outside of the contexts, a bit above (note the indentation) otherwise you will get expectation errors of Libvirt. Put the test outside of this context (must be not in any other context):

    context 'Fog.mock!' do

sean797:
Massive thanks @lzap! Tests passed 👍

test/controllers/hosts_controller_test.rb
```diff
@@ -1310,6 +1310,26 @@ def test_submit_multiple_rebuild_config_pessimistic
...
+    host.stubs(:operatingsystem)
+    host.stubs(:environment)
+    host.stubs(:domain)
+    host.stubs(:subnet)
+    host.stubs(:compute_profile)
+    host.stubs(:realm)
+    attrs = host_attributes(host)
+    attrs[:id] = host.id
+    attrs[:hostgroup_id] = hostgroup.id
+    xhr :put, :process_hostgroup, { :host => attrs }, set_session_user
```
ares: `set_session_user` uses admin user so the authorization code is skipped, should it be tested under user to verify `.authorized(:view_hosts)` works correctly?

lzap: Yeah, please do following:

```ruby
    user = FactoryGirl.create(:user, :with_mail, :admin => false)
    FactoryGirl.create(:filter, :role => user.roles.first, :permissions => Permission.where(:name => 'view_hosts'))
    .... set_session_user(user)
```

sean797: Thanks @lzap! Tests passed @ares  :)

lzap (approved):

ares (approved):

lzap:
@sean797 thanks a lot!

PR #4117, opened 2016-12-15
Fixes #17686 - wider focus area of action buttons

iNecas:
The buttons in tables were not capturing the click, if it was outside of the
text. Let's avoid the frustration this might cause.

table-button-click

### DIFF
--- a/app/assets/stylesheets/patternfly_and_overrides.scss
+++ b/app/assets/stylesheets/patternfly_and_overrides.scss
@@ -77,6 +77,9 @@ span.input-group-btn .hidden-xs {display:inline !important;}
 span.btn a {
   text-decoration: none;
   color: #333333;
+  // make sure the link inside span covers the whole area of the button
+  padding: 8px;
+  margin: -8px
 }
 
 a {


tbrisker:
This works generally, however there is some overflow of the `a` element outside the button, and the padding size depends on the `btn-x` class. Also, I am worried even if we use the correct padding sizes to match the span, this may break in the future if bootstrap/patternfly change the button design.
I think it may be better to fix the underlying issue - we wrap the `a` in a `span` to give it the various button classes instead of placing them on the `a` itself. I started to play with it at https://github.com/tbrisker/foreman/commit/ddf4f4a5696084f3d392c16b38f8ae78e704e627 which mostly works, only issue i noticed so far is that delete links aren't red anymore, which should be simple to fix.

iNecas:
While I agree fixing the underlying issue would be better, TBH doing the `sub!` on the string and expecting a link being always there doesn't look like much better solution :)

ares:
I agree with @iNecas, parsing html text and injecting class there does not seem any better. I'd say we should either take short-term hack presented in this PR and open refactor issue or reject and do the proper fix which means fix all buttons to be created correctly.

tbrisker:
Looking back on this I agree changing the string is not much better then this, although fixing it correctly will require a major refactoring of the code - we only need to add the classes for the first link, but these functions receive the links as an array of strings, meaning we can't easily add classes to them... 
At the very least, if we are going to accept this hack, please make sure the extra margin/padding matches those on the buttons and add a comment to the scss explaining what this is for.

app/assets/stylesheets/patternfly_and_overrides.scss
```diff
@@ -77,6 +77,9 @@ span.input-group-btn .hidden-xs {display:inline !important;}
 span.btn a {
   text-decoration: none;
   color: #333333;
+  // make sure the link inside span covers the whole area of the button
+  padding: 8px;
+  margin: -8px
 }
```
gailsteiger, line 84: I always solve this by giving the `<a/>` element `display: inline-block;` styling. The enclosing element should also have `block` or `inline-block display`

Generally, use `padding` instead of `margin` between an element and its enclosing element

ares:
What's the status here? I'd love to see this in.

sean797:
I too would love to see this in 1.15. I would suggest we take this short term hack for now

tbrisker:
I'm willing to let the hack in, if the sizes are fixed to match the correct sizes of the buttons, unless the solution suggested by @gailsteiger (changing the `display` option) proves to be a better solution.

iNecas:
I've tried and tempering with `display` values both for `a` and `span` without any positive effect on the results. I'd be very grateful for people specializing in frontend  to fix this better: this hack is the top I can do with my skills :)

ohadlevy:
@iNecas can you explain @matanwerbner what exactly you ask for? :) thanks.

matanwerbner:
Can you add a 'btn-container' class to the btn div if you know it's going to contain clickable elements?
then 
```
.btn-container { padding: 0 }
.btn-container * { padding: 2px 6px }
```

iNecas:
@matanwerbner what I want to achieve is recorded in the original gif (it's the reproducer), the goal is for the button to respond on all places, not just on the caption.

I think the most effective way right now would be a UI expert (@matanwerbner ?) opening a new PR with better solution (if it exists without rewritting half of our helpers ?:) and I will be more than happy to close this PR for the new one.

matanwerbner:
@iNecas i would be happy to help, but can you see my above comment first? This should be done on the ruby part - adding .btn-container to buttons that contain children.

iNecas:
It doesn't work for me and it's not fun for me anymore: non of the hints I got so far seemed to have any effect on the actual behavior. I would appreciate for folks understanding css better than I do to really reproduce the issue and find a suitable solution (other than this one): at this point, I need patches, not comments. Thanks

matanwerbner:
@iNecas i'm sorry if it's no fun, and i'm trying to help, but i don't do ruby.
the only way i can help is if you are able to achieve what i wrote in previous comments.
if it involves re-writing some helpers, then that needs to be done.
that is the only solution i can think of right now.

iNecas:
As I already mentioned, I've tried it and it didn't help. Is there anyone with a versions better than this and actually fixing the original problem? if not, please consider merging this and lets move on. Or just close this PR.

ares:
Closing this makes me sad, could new @theforeman/ui-ux perhaps take a look and consider whether this is acceptable?

terezanovotna:
The button functionality shows the user where to click, so it is a great idea to make the entire button clickable.

PR #4119, opened 2016-12-20
Fixes #17794 - Update user groups on EXTERNAL user creation

dLobatog:
Given an user who logs in via external delegation (REMOTE_USER). This user
is created automatically upon login. The user is not authenticated using
regular LDAP authentication sources, but through the 'authorized login
delegation' setting. Since REMOTE_USER_GROUP_* is not set, the user
groups are not set.

The user in reality comes from ActiveDirectory which is listed in
Foreman as an LDAP auth source - but using the krb5 ticket. Since the
login provided by REMOTE_USER and LDAP is the same, one would expect
that the LDAP user groups would be updated when you log in. However the
user groups are only updated through the cron or click on
'refresh user groups'.

This patch moves all logic of refreshing user groups to
post_successful_login, which is called from try_to_login, and
find_or_create_external_user (which is called from SSO::Apache).

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -237,7 +237,6 @@ class User < ApplicationRecord
               logger.warn "Failed to update #{user.login} attributes: #{user.errors.full_messages.join(', ')}"
             end
           end
-          user.auth_source.update_usergroups(user.login)
         end
 
         # clean up old avatar if it exists and the image isn't in use by anyone else
@@ -259,12 +258,30 @@ class User < ApplicationRecord
   end
 
   def post_successful_login
+    update_external_usergroups
     logger.debug "Post-login processing for #{login}"
     User.as_anonymous_admin do
       update_columns(:last_login_on => Time.now.utc)
       ensure_default_role
     end
     User.current = self
+  end
+
+  def update_external_usergroups
+    case auth_source.auth_method_name
+    # when 'INTERNAL'
+    # Internal users cannot have external user groups
+    when 'EXTERNAL'
+      # In the case of external authentication, we check on all LDAP auth
+      # sources to update user groups on login.
+      AuthSourceLdap.unscoped.all.map do |ldap_source|
+        ldap_source.update_usergroups(login)
+      end
+    when 'LDAP'
+      # For LDAP users, we just call update_usergroups on the auth_source
+      # we know the user comes from.
+      auth_source.update_usergroups(login)
+    end
   end
 
   def self.find_or_create_external_user(attrs, auth_source_name)
@@ -501,7 +518,6 @@ class User < ApplicationRecord
     # The default user can't auto create users, we need to change to Admin for this to work
     User.as_anonymous_admin do
       if user.save
-        auth_source.update_usergroups(user.login)
         logger.info "User '#{user.login}' auto-created from #{user.auth_source}"
       else
         logger.info "Failed to save User '#{user.login}' #{user.errors.full_messages}"
--- a/test/controllers/users_controller_test.rb
+++ b/test/controllers/users_controller_test.rb
@@ -247,6 +247,7 @@ class UsersControllerTest < ActionController::TestCase
     Setting['authorize_login_delegation_auth_source_user_autocreate'] = 'apache_mod'
     @request.session.clear
     @request.env['REMOTE_USER'] = 'ares'
+    AuthSourceLdap.any_instance.expects(:update_usergroups).with('ares')
     get :extlogin, {}, {}
     assert_redirected_to edit_user_path(User.unscoped.find_by_login('ares'))
   end
--- a/test/models/user_test.rb
+++ b/test/models/user_test.rb
@@ -7,6 +7,10 @@ class UserTest < ActiveSupport::TestCase
     @user = User.create :auth_source => auth_sources(:one), :login => "foo", :mail  => "foo@bar.com"
   end
 
+  def expect_ldap_connection(user = @user.login)
+    AuthSourceLdap.any_instance.expects(:update_usergroups).with(user)
+  end
+
   # Presence
   should validate_presence_of(:login)
   should validate_presence_of(:auth_source_id)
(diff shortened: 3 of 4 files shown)


tbrisker:
@dLobatog Code looks good, however tests are failing

dLobatog:
@tbrisker Thanks, updated to fix tests. More stubbing of LDAP connections was necessary 😄

ares:
@dLobatog would it make sense if we find some mapping to also set this source as users auth source? also why there are no user groups populated in REMOTE_USER_GROUP_*, is that a bug in apache module that should provide it?

tbrisker:
@dLobatog @ares What is the status here?

ohadlevy:
@ares @dLobatog bump?

ares:
I didn't want to set WoC as I didn't do the full review but my questions were not answered yet. Setting it now to get things moving.

bkearney:
@dLobatog bump on this.

dLobatog (commented):
@ares How would you make the mapping? Users who log in using REMOTE_USER could be in any of the auth sources. For example:

User 1 - Admin - logs in through krb5, REMOTE_USER='Admin'. This is connected to ActiveDirectory
User 2 - Admin - logs in through form. Admin is found on FreeIPA, so it's associated to this AuthSource. 

Linking based on user name (which is the only thing we know about users who log in with REMOTE_USER) isn't right IMO, let me know if that's what you meant or is it something else.

ares:
I'm afraid I don't understand the logic of this change.

EXTERNAL auth source is not ldap auth source. We should not look at any ldap auth source when user is linked to external one. We should not mix the two. If users setup ldap auth source, they don't have kerberos authentication. If they need it, they should go the more difficult path using the installer which setups the service ticket for the host and all the rest. Only IPA supports that, AD can be used by setting up a trust between IPA and AD.

Having said that, EXTERNAL auth source should work as any other ldap source in terms of user groups. The source in this case is not any remote LDAP server (we don't even know if the user is in any LDAP) but the env variables. We're missing the ability to define external user groups mapping for EXTERNAL auth source but that's WIP. See https://github.com/theforeman/foreman/pull/4865 and http://projects.theforeman.org/issues/19651, the UI will follow.

The redmine issue says that the problem is that such users don't get usergroups synced on first login not that they should get usergroups from all available ldap auth sources.

EDIT: also see http://projects.theforeman.org/issues/17794#note-4

dLobatog:
@ares I agree the change is strange - the problem in question is that the user (I had a remote session with them) does not have control of LDAP. They were able to have someone configure REMOTE_USER for them, but not REMOTE_USER_GROUP_*. At first they were confused because they logged in through REMOTE_USER and they didn't get the user linked to the external user groups. They thought external user groups worked for EXTERNAL auth sources - but they don't, as when Foreman tries to call `find_usergroups(external_user)`, there is no response..

However (and here's the trick, or the bug...) when they refresh usergroups manually or the cronjob runs, the call is the opposite. They added a LDAP auth source which they use to create external user groups. Then `find_users(external_usergroup)` returns the names of the users, which were created through REMOTE_USER. Then the link between user groups and users works.

It's definitely a hack, just a way to work around the fact this user could not set REMOTE_USER_GROUP_* variables, but it causes no harm in other situations (I think) and allows REMOTE_USERS to be linked to external user groups on login (instead of after a cronjob/manual refresh)

ares:
I think I'm struggling to understand the issue. If they setup the authentication using external auth source (I suppose using installer option), the REMOTE_USER_GROUP should be populated, see https://github.com/theforeman/puppet-foreman/blob/master/templates/lookup_identity.conf.erb. If that does not work, it most likely just silently fails during authentication (we don't want to block user login in case mod_lookup_identity can't fetch additional info). If they configured it manually,
 double check they have correctly configured [mod_identity_lookup](https://theforeman.org/manuals/1.15/index.html#5.7.5Populateusersandattributes) that would fetch user groups. The sync in this case happens only on user login, there's no URL we could use to ask, as the authentication happens externally. Users have a choice, they can get regular updates for LDAP auth sources when Foreman talks to LDAP directly but that does not provide SSO. Or they want to use SSO which is more complicated setup but means that user groups are synced only on login.

If for whatever reason mod_identity_lookup can't fetch the information, I'd like to avoid mixing auth sources. This hack would cause issues if you had both external and direct ldap authentication but these would be two different sources. If user would get some user group from env variable and second from some LDAP (two different accounts with same user name) that would lead to very unexpected behavior.

I suggest we reject this and ask user to verify why standard external authentication does not work for them. If we want to implement this RFE, then I think it should be explicit link between external auth source and ldap auth source that loads information for users of this external auth source. But that seems like something we should not be adding to our codebase in the first place.

Few extra arguments:
* external auth source does not have to be supported LDAP, it could be whatever LDAP sssd talks to
* external auth source does not have to be LDAP at all
* we don't want to duplicate the logic that already exists in mod_identity_lookup

dLobatog (commented):
@ares 

I think you got the gist of it. 

>  If they setup the authentication using external auth source (I suppose using installer option), the REMOTE_USER_GROUP should be populated,

It's not, the user who requested this only configured the external authentication so that REMOTE_USER is populated but REMOTE_USER_GROUP is not. In fact, they can't change this because they control the Foreman servers but not Apache or LDAP.

> Users have a choice, they can get regular updates for LDAP auth sources when Foreman talks to LDAP directly but that does not provide SSO. Or they want to use SSO which is more complicated setup but means that user groups are synced only on login.

Correct. The thing is, in the end it's Active Directory the source that is providing REMOTE_USER & SSO. So what these people did is to add Active Directory as an Auth Source LDAP & hope that since `REMOTE_USER == user from AD`,  groups would be synced when they log in through SSO. 

From Foreman's POV, obviously they won't because REMOTE_USER is an user in External Auth Source and "user from AD" is an user in Auth Source LDAP. 

I'm OK with rejecting this too as it's a hack, the real fix should be to ask their LDAP group to provide REMOTE_USER_GROUP in sssd. I opened the original BZ after having a remote session with a customer who wanted this. 

If we both agree it's a hack, feel free to close (close the Redmine/BZ too please)

ares:
Updated the BZ, I think this can be closed now. In such case, please also reset status in redmine, thank you.

PR #4118, opened 2016-12-16
Fixes #17681 - Switch to newly created taxonomy

xprazak2:
The context switches to newly created taxonomy as soon as user leaves the wizard.

PR #4121, opened 2016-12-20
Fixes #17796 - bump to rubocop 0.46

shlomizadok:


shlomizadok:
@mention-bot, you have found the perfect reviewers. Thank you!!1 :heart:

tbrisker:
@shlomizadok This cop was only introduced in rubocop 0.44, while we currently pin rubocop to 0.42. 
Also, I'm not certain we should disable it but rather add to the todo since it will be required for rails 5 upgrade IIUC.

shlomizadok:
@tbrisker - 
It is present (see: https://github.com/theforeman/foreman/pull/4024#discussion_r92941720, https://github.com/theforeman/foreman/pull/4118#discussion_r92996691 and other PRs)

I'll remove to .rubocop_todo

tbrisker:
@shlomizadok hound runs with its own version of rubocop, which is currently 0.46. We should upgrade ours to match so that we can also run the same cops locally.

shlomizadok:
OIC, so upgarde to 0.46 would be the right thing to do?

dLobatog:
Thanks @shlomizadok @tbrisker !

PR #4122, opened 2016-12-21
Fixes #17807 - Handy script to mirror rails runner

parthaa:
In a production environment I ve had to use "foreman-rake console" quite a few times to get information about objects while debugging customer instances. I know rails provides a "rails runner" script that can be used to run and fetch/execute data.

This is a  handy script to say something like
"foreman-rake runner EVAL='Hostgroup.find(5).inspect'"
or
"foreman-rake runner SCRIPT='script-to-fix-data-issues'
and have it executed as  admin.

### DIFF
--- a/lib/tasks/runner.rake
+++ b/lib/tasks/runner.rake
@@ -1,0 +1,19 @@
+task :runner=> :environment do
+  Rails.application.require_environment!
+  Rails.application.load_runner
+  file=ENV['SCRIPT']
+  code=ENV['EVAL']
+  if !code && !file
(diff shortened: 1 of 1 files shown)


bastilian:
@parthaa I think this could also be as simple as ['foreman-rake console'](https://github.com/theforeman/foreman/blob/develop/lib/tasks/console.rake) in terms of code, by calling Rails' runner's [`perform`](https://github.com/rails/rails/blob/master/railties/lib/rails/commands/runner/runner_command.rb#L17) method.

dLobatog:
@parthaa Can you verify if there's any difference on what your code does and what @bastilian proposes? I think it ought to be basically the same, if you could confirm and change the code to that I believe it should be ready to merge.

Thanks you both!

PR #4123, opened 2016-12-21
Fixes #17810 - Deleting smart class parameters with puppetclass

orrabin:


test/models/lookup_keys/puppetclass_lookup_key_test.rb
```diff
@@ -31,4 +31,41 @@ class PuppetclassLookupKeyTest < ActiveSupport::TestCase
     lookup_key.override = false
     assert lookup_key.valid?
   end
+
+  test "deleting puppetclass should delete smart class parameters" do
+    env1 = FactoryGirl.create(:environment)
```
dLobatog: Looks like you can put these in a context 'delete params with class', then move the creation of `env1, puppetclass, lookup_key`, and the `environment_class` that uses `env1` to a `setup` block

app/models/environment_class.rb
```diff
@@ -28,4 +29,10 @@ def self.is_in_any_environment(puppetclass, puppetclass_lookup_key)
   def self.key_in_environment(env, puppetclass, puppetclass_lookup_key)
     EnvironmentClass.where(:environment_id => env, :puppetclass_id => puppetclass, :puppetclass_lookup_key_id => puppetclass_lookup_key).first
   end
+
+  def delete_orphaned_lookup_keys
+    if puppetclass_lookup_key.environment_classes.empty?
```
dLobatog: What's the difference between this and `belongs_to :puppetclass_lookup_key, :dependent => :destroy` ?

dLobatog (changes requested):
@orrabin Thanks - I also think it's quite difficult to reimport a class & expect the params to still work (happened to me when I upgraded to puppet 4, where some modules would have vastly different params due to newer versions). Check the comments inline and let me know when it's ready again 😄

dLobatog:
Also please check tests 😄 !

app/models/environment_class.rb
orrabin: There can be multiple `environment_classes` connected to the same `puppetclass_lookup_key` and I don't want the lookup key destroyed when you remove one of those records. I want to make sure it deletes only if this is the last environment class for this key.

test/models/lookup_keys/puppetclass_lookup_key_test.rb
orrabin: Thanks, will do.

app/models/environment_class.rb
```diff
@@ -28,4 +29,10 @@ def self.is_in_any_environment(puppetclass, puppetclass_lookup_key)
   def self.key_in_environment(env, puppetclass, puppetclass_lookup_key)
     EnvironmentClass.where(:environment_id => env, :puppetclass_id => puppetclass, :puppetclass_lookup_key_id => puppetclass_lookup_key).first
   end
+
+  def delete_orphaned_lookup_keys
+    if puppetclass_lookup_key.present? && puppetclass_lookup_key.environment_classes.empty?
```
orrabin, line 34: This comment got deleted:
> What's the difference between this and belongs_to :puppetclass_lookup_key, :dependent => :destroy ?

There can be multiple environment_classes connected to the same puppetclass_lookup_key and I don't want the lookup key destroyed when you remove one of those records. I want to make sure it deletes only if this is the last environment class for this key.

orrabin:
@dLobatog I responded to one comment and fixed the rest, can you take another look please?

db/migrate/20161205142618_delete_orphaned_smart_class_parameters.rb
```diff
@@ -0,0 +1,8 @@
+class DeleteOrphanedSmartClassParameters < ActiveRecord::Migration
+  def up
+    PuppetclassLookupKey.where("NOT EXISTS (SELECT * FROM environment_classes WHERE environment_classes.puppetclass_lookup_key_id = lookup_keys.id)").destroy_all
```
dLobatog, line 3: Just a comment for any onlookers, I tried `PuppetclassLookupKey.where.not(:id => EnvironmentClass.pluck(:puppetclass_lookup_key_id))` but the query produced is quite terrible (because of the .pluck), so this is probably fine.

dLobatog:
Thanks @orrabin !

PR #4127, opened 2016-12-23
Fixes #17837 - Fix hosts:scan_out_of_sync task

dLobatog:
The task relies on Ping, which was removed since Ruby 1.9. I've moved it
to rely on net-ping instead. Notice the ping is not really on ICMP, as
those pings require root privileges, which 'foreman-rake' is most likely
not to have.

### DIFF
--- a/lib/tasks/hosts.rake
+++ b/lib/tasks/hosts.rake
@@ -1,67 +1,94 @@
 # TRANSLATORS: do not translate
 desc <<-END_DESC
-  Try to figure out the out of sync hosts real status
+  This task search hosts that have been out of sync for longer than
+  twice your out of sync interval by using DNS lookups and ping.
 
-  try to search them by DNS lookups and ping, if a host is not in DNS it will allow you to delete it.
+  Once it finishes scanning, it will prompt you to delete hosts that
+  don't have DNS entries nor respond to ping.
 
-  legend:
+  You can provide the number of seconds after which a host is considered
+  out of sync by using the environment variable 'OUTOFSYNC_INTERVAL'.
+
+  Legend:
   "." - pingable
   "x" - no ping response
 
-  Example:
+  Examples:
     rake hosts:scan_out_of_sync RAILS_ENV="production"
+    rake hosts:scan_out_of_sync RAILS_ENV="production" OUTOFSYNC_INTERVAL=3600
 
 END_DESC
 
 namespace :hosts do
   task :scan_out_of_sync => :environment do
-    require 'ping'
+    require 'net/ping/external'
     require 'resolv'
 
     def printhosts(list, description)
       unless list.empty?
         puts
-        puts "found #{list.size} #{description} hosts:"
+        puts "Found #{list.size} #{description}:"
         puts "Name".ljust(40)+"Environment".ljust(20)+"Last Report"
-        puts "#{"*"*80}"
+        puts '*' * 80
         list.each do |h|
           puts h.name.ljust(40) + h.environment.to_s.ljust(20) + h.last_report.to_s(:short)
         end
       end
     end
 
+    def out_of_sync_interval
+      return ENV['OUTOFSYNC_INTERVAL'] if ENV['OUTOFSYNC_INTERVAL'].present?
+      2 * (Setting[:puppet_interval] + Setting[:outofsync_interval])
+    end
+
     pingable = []
     missingdns = []
     offline = []
 
-    Host.out_of_sync(1.hour.ago).all(:order => 'environment_id asc').collect do |host|
+    Host::Managed.out_of_sync(out_of_sync_interval.to_i.seconds.ago).
+      order('environment_id ASC').collect do |host|
       $stdout.flush
       ip = Resolv::DNS.new.getaddress(host.name).to_s rescue nil
       if ip.empty?
         missingdns << host
       else
-        puts "conflict IP address for #{host.name}" unless ip == host.ip
-        if Ping.pingecho host.ip
+        if host.ip.blank?
+          puts "Host #{host.name} does not have an IP in Foreman. It resolved to IP: '#{ip}' through DNS. Skipping..."
+          next
+        end
+        all_host_ips = host.interfaces.pluck(:ip)
+        puts "Conflict IP address for #{host.name}" unless all_host_ips.include? ip
+        begin
+        if Net::Ping::External.new(host.ip).ping
           print "."
           pingable << host
         else
           print "x"
           offline << host
         end
+        rescue => e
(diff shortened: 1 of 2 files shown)


lib/tasks/hosts.rake
```diff
@@ -34,14 +34,15 @@ namespace :hosts do
     missingdns = []
     offline = []
 
-    Host.out_of_sync(1.hour.ago).all(:order => 'environment_id asc').collect do |host|
+    Host::Managed.out_of_sync.order('environment_id ASC').collect do |host|
```
ohadlevy: the idea behind bumping the interval was to avoid those machines that might be running long puppet runs, maybe we should default to puppet_interval*2 instead?

lib/tasks/hosts.rake
```diff
@@ -34,14 +34,15 @@ namespace :hosts do
...
       $stdout.flush
       ip = Resolv::DNS.new.getaddress(host.name).to_s rescue nil
       if ip.empty?
         missingdns << host
       else
         puts "conflict IP address for #{host.name}" unless ip == host.ip
-        if Ping.pingecho host.ip
+        host_icmp = Net::Ping::TCP.new(host.ip)
```
ohadlevy: if you are just using tcp ping, please change: 

1. require line to ```require 'net/ping/tcp'```
2. the variable name is wrong
3. ECONNREFUSED error is considered a failed ping, yet in our case it shows the machine is up (see https://github.com/chernesk/net-ping/blob/master/doc/ping.txt)

TBH: maybe using Net::Ping::External will yield more reliable results.

dLobatog:
@ohadlevy I think this should be ready again, it uses external ping (whatever the OS has), which doesn't require root. I extended a bit the task by allowing to provide your own interval, and made the deletion confirmation a bit more 'foolproof' (now users have to write 'y' instead of 'any button will delete these hosts').

lib/tasks/hosts.rake
```diff
@@ -33,15 +39,20 @@ namespace :hosts do
...
+      ENV['OUTOFSYNC_INTERVAL'].to_i.seconds.ago || long_interval
+    ).order('environment_id ASC').collect do |host|
       $stdout.flush
       ip = Resolv::DNS.new.getaddress(host.name).to_s rescue nil
       if ip.empty?
         missingdns << host
       else
         puts "conflict IP address for #{host.name}" unless ip == host.ip
-        if Ping.pingecho host.ip
+        if Net::Ping::External.new(host.ip).ping
```
ohadlevy, line 62: Net::Ping can create exceptions, could you wrap it in block?

lib/tasks/hosts.rake
```diff
@@ -33,15 +39,20 @@ namespace :hosts do
...
-    Host.out_of_sync(1.hour.ago).all(:order => 'environment_id asc').collect do |host|
+    Host::Managed.out_of_sync(
+      ENV['OUTOFSYNC_INTERVAL'].to_i.seconds.ago || long_interval
+    ).order('environment_id ASC').collect do |host|
       $stdout.flush
       ip = Resolv::DNS.new.getaddress(host.name).to_s rescue nil
       if ip.empty?
         missingdns << host
       else
         puts "conflict IP address for #{host.name}" unless ip == host.ip
```
ohadlevy: also, this now be updated to multi nic

dLobatog:
Updated to handle exceptions and hosts without IP.

lib/tasks/hosts.rake
dLobatog: The way it works right now it will check just the primary interface which is the one meant for DNS, which other interfaces would you check?

lib/tasks/hosts.rake
```diff
@@ -33,35 +39,54 @@ namespace :hosts do
...
-        if Ping.pingecho host.ip
+        if host.ip.nil? || host.ip.empty?
```
ohadlevy: nitpick, ```host.ip.blank?``` will work on both cases.

lib/tasks/hosts.rake
```diff
@@ -33,35 +39,54 @@ namespace :hosts do
...
+        begin
+        if Net::Ping::External.new(host.ip).ping
           print "."
           pingable << host
         else
           print "x"
           offline << host
         end
+        rescue => e
+          puts "Could not ping host #{host.name} due to an exception: #{e} - skipping"
```
ohadlevy: should probably STDERR (or warn method)

lib/tasks/hosts.rake
ohadlevy: we can just check that the ip is part of one the interfaces we have assigned to the host.
e.g.``` host.all_ips.include?(ip)```

dLobatog:
@ohadlevy Updated according to comments

lib/tasks/hosts.rake
```diff
@@ -33,35 +39,55 @@ namespace :hosts do
     pingable = []
     missingdns = []
     offline = []
+    long_interval = (
+      2 * (Setting[:puppet_interval] + Setting[:outofsync_interval])
+    ).seconds.ago
 
-    Host.out_of_sync(1.hour.ago).all(:order => 'environment_id asc').collect do |host|
+    Host::Managed.out_of_sync(
+      ENV['OUTOFSYNC_INTERVAL'].to_i.seconds.ago || long_interval
```
ohadlevy: ``` ENV['OUTOFSYNC_INTERVAL'].to_i``` will default to 0 if OUTOFSYNC_INTERVAL is not defined.

lib/tasks/hosts.rake
```diff
@@ -33,35 +39,55 @@ namespace :hosts do
...
+    ).order('environment_id ASC').collect do |host|
       $stdout.flush
       ip = Resolv::DNS.new.getaddress(host.name).to_s rescue nil
       if ip.empty?
         missingdns << host
       else
-        puts "conflict IP address for #{host.name}" unless ip == host.ip
-        if Ping.pingecho host.ip
+        if host.ip.blank?
+          puts "host #{host.name} does not have an IP in Foreman. It resolved to IP: '#{ip}' through DNS"
```
ohadlevy: i would add -  "skipiing..."

lib/tasks/hosts.rake
```diff
@@ -33,35 +39,55 @@ namespace :hosts do
...
+          next
+        end
+        all_host_ips = host.interfaces.map(&:ip)
```
ohadlevy: prefer pluck instead of map.

dLobatog:
@ohadlevy Updated again with changes:

* map -> pluck
* 'skipping...' added to message
* handling for `ENV["OUTOFSYNC_INTERVAL"].nil?`

ohadlevy:
works well, thanks @dLobatog

PR #4128, opened 2016-12-23
fixes #17838 - translate New Hosts title string

domcleal:


timogoebel (approved):
@domcleal : Thanks, will merge once the tests are green.

PR #4129, opened 2016-12-23
fixes #17533 - provision dhcp for bond child macs

timogoebel:
This commit adds the ability to provision a host which has a bond interface as its provisioning interface. The bond interface doesn't have a mac address set, but is linked to usually two physical interfaces that are connected to the same VLAN for additional redundancy or speed.
While the host is PXE-booting it just sees two independent interfaces, which it tries to boot from. We therefore need to set TFTP (bcb107cf8d5039486f07b136e4b41c337fda2921) and DHCP records for both physical interfaces. This commit adds the latter functionality.

The idea is, that two static leases would be written with the same IP-address but different mac. For a bond interface, the records are not named like the hostname, but the hostname suffixed with a consecutive numbering (e.g. "example.com-01", "example.com-02"). 

Cobbler already supports this functionality and handles the dhcp records in a similar matter.

Feedback is much appreciated.

This needs changes to Smart Proxy to work, see https://github.com/theforeman/smart-proxy/pull/485

### DIFF
--- a/app/models/concerns/orchestration/dhcp.rb
+++ b/app/models/concerns/orchestration/dhcp.rb
@@ -19,58 +19,58 @@ module Orchestration::DHCP
         !subnet.nil? && subnet.dhcp? && SETTINGS[:unattended] && (!provision? || operatingsystem.present?)
   end
 
-  def dhcp_record
-    return unless dhcp? || @dhcp_record
-    handle_validation_errors do
-      @dhcp_record ||= (provision? && jumpstart?) ? Net::DHCP::SparcRecord.new(dhcp_attrs) : Net::DHCP::Record.new(dhcp_attrs)
+  def dhcp_records
+    return [] unless dhcp?
+    @dhcp_records ||= mac_addresses_for_provisioning.map do |record_mac|
+      build_dhcp_record(record_mac)
     end
   end
 
   def reset_dhcp_record_cache
-    @dhcp_record = nil
+    @dhcp_records = nil
   end
 
   def rebuild_dhcp
-    if dhcp?
-      del_dhcp_safe
-      begin
-        set_dhcp
-      rescue => e
-        Foreman::Logging.exception "Failed to rebuild DHCP record for #{name}, #{ip}", e, :level => :error
-        false
-      end
-    else
+    unless dhcp?
       logger.info "DHCP not supported for #{name}, #{ip}, skipping orchestration rebuild"
-      true
+      return true
+    end
+
+    del_dhcp_safe
+    begin
+      set_dhcp
+    rescue => e
+      Foreman::Logging.exception "Failed to rebuild DHCP record for #{name}, #{ip}", e, :level => :error
+      false
     end
   end
 
   protected
 
   def del_dhcp_safe
-    if dhcp_record
-      begin
-        del_dhcp
-      rescue => e
-        Foreman::Logging.exception "Proxy failed to delete DHCP record for #{name}, #{ip}", e, :level => :error
-      end
-    end
+    del_dhcp
+  rescue => e
+    Foreman::Logging.exception "Proxy failed to delete DHCP record for #{name}, #{ip}", e, :level => :error
   end
 
   def set_dhcp
-    dhcp_record.create
+    dhcp_records.all? { |record| record.create }
   end
 
   def set_dhcp_conflicts
-    dhcp_record.conflicts.each{|conflict| conflict.create}
+    dhcp_records.all? do |record|
+      record.conflicts.each { |conflict| conflict.create }
+    end
   end
 
   def del_dhcp
-    dhcp_record.destroy
+    dhcp_records.all? { |record| record.destroy }
   end
 
   def del_dhcp_conflicts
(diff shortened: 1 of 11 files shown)


timogoebel:
The smart-proxy PR has been merged. This is ready for a review, now.

app/models/concerns/orchestration/dhcp.rb
```diff
@@ -93,16 +96,27 @@ def boot_server
 
   private
 
+  def build_dhcp_record(record_mac)
+    raise ::Foreman::Exception.new(N_("DHCP not supported for this NIC")) unless dhcp?
+    record_attrs = dhcp_attrs(record_mac)
+    record_type = (provision? && jumpstart?) ? Net::DHCP::SparcRecord : Net::DHCP::Record
+    handle_validation_errors do
```
dLobatog, line 100: Nice job extracting this method, it's much more readable and it's easy to see how the record is created now

app/models/concerns/orchestration/dhcp.rb
```diff
@@ -19,58 +19,61 @@ def dhcp?
...
 
   def del_dhcp
-    dhcp_record.destroy
+    dhcp_records.all? { |record| record.destroy }
   end
 
   def del_dhcp_conflicts
-    dhcp_record.conflicts.each{|conflict| conflict.destroy}
+    dhcp_records.all? do |record|
+      record.conflicts.each { |conflict| conflict.destroy }
```
dLobatog: This should be `.map`, because `.each` will not return the results of `conflict.destroy`.

app/models/concerns/orchestration/dhcp.rb
```diff
@@ -19,58 +19,61 @@ def dhcp?
...
   protected
 
   def del_dhcp_safe
-    if dhcp_record
-      begin
-        del_dhcp
-      rescue => e
-        Foreman::Logging.exception "Proxy failed to delete DHCP record for #{name}, #{ip}", e, :level => :error
-      end
+    return unless dhcp_records.any?
```
dLobatog: I believe you can move this call `dhcp_records.any?` to `del_dhcp` so that it looks like:

```
   def del_dhcp
    return true unless dhcp_records.any?
    dhcp_records.all? { |record| record.destroy }
  end
```

Then you can remove it from here

app/models/concerns/orchestration/dhcp.rb
dLobatog: These are the possibilities:

No conflicts - `record.conflicts.each` will return [], since it's not `false`, it's `true` for `all?`. This works fine.
Conflicts - `record.conflicts.each` will return [false, true]. That will be counted as `true` for `all?`. This does not work fine. For example:

```ruby
[1,2,3].all? { |e| [e % 2 == 0, true] }
=> true
[1,2,3].all? { |e| e % 2 == 0 }
=> false
```

You should do `record.conflicts.map(&:destroy).uniq == ([true] || [])` to ensure all conflicts were destroyed properly (or there were none to begin with).

test/unit/proxy_api/dhcp_test.rb
```diff
@@ -15,6 +15,41 @@ def setup
...
+                    :nextServer=>"1.2.3.4",
+                    :filename=>"pxelinux.0",
+                    :name=>"www.example.com",
+                    :related_macs=>[]}, result.attrs)
+    end
+  end
+
+  context '#records_by_ip' do
+    test 'retrieves an array with a single dhcp record' do
+      fake_response = [{'name'=>'www.example.com', 'mac'=>'00:11:22:33:44:55', 'ip'=>'192.168.0.10', 'filename'=>'pxelinux.0', 'nextServer'=>'1.2.3.4', 'hostname'=>'www.example.com', 'subnet'=>'192.168.0.0/255.255.255.0'}]
```
dLobatog: Looks like you can extract the first 2 lines of these tests to a method or put them in the same context. I believe these two tests `context` should actually be `describe` instead, it's most commonly used to describe a method, whereas `context` is for situations.

dLobatog (changes requested):
@timogoebel Left a few comments inline, I was able to do provisioning with the smart-proxy patch, didn't try with an old version of the proxy, what happens in such a case? Mostly everything worked well except the 'del_dhcp_conflicts' not noticing when a removing conflict returns false.

Thanks!

app/models/concerns/orchestration/dhcp.rb
timogoebel: Actually, i don't think the guard clause is needed at all.
`[].all? {...}` will not execute any code but just return true.

app/models/concerns/orchestration/dhcp.rb
timogoebel: How about using `#all?` twice?
```ruby
def del_dhcp_conflicts
  dhcp_records.all? do |record|
     record.conflicts.all? { |conflict| conflict.destroy }
  end
end
```

timogoebel:
@dLobatog : Addressed the comments. Thanks again for the review!

dLobatog (approved):

dLobatog:
Thanks @timogoebel !

PR #4131, opened 2016-12-25
Fixes #17842 - Adds 'type = compute_resource' to Audit search

shlomizadok:


tbrisker:
@shlomizadok How about adding all the missing types?

test/models/concerns/audit_extensions_test.rb
```diff
@@ -41,4 +41,9 @@ def setup
     FactoryGirl.create :lookup_value, :lookup_key_id => key.id, :value => false, :match => "hostgroup=Common"
     refute_empty Audit.search_for("type = lookupvalue")
   end
+
+  test "search for type=compute_resource in audit" do
+    FactoryGirl.create(:ec2_cr)
+    refute_empty Audit.search_for("type = compute_resource")
```
tbrisker, line 47: `assert_empty` before the creation please to make sure that it isn't empty because the creation added an entry.

shlomizadok: I do not care if it is empty... I just want to ensure I have at least one when I search. 
The purpose of this test is to ensure I can seach Audit by type=compute_resource

shlomizadok:
@tbrisker - each type at a time :smile:

test/models/concerns/audit_extensions_test.rb
ohadlevy: also would be nice to show that you search for different compute resource types ?(e.g. that STI works correctly).

PR #4132, opened 2016-12-26
Fixes #4238 - Prevent login brute forcing

tbrisker:
After 5 failed attempts from the same ip, login will be blocked for 5
minutes.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -2,6 +2,7 @@ class UsersController < ApplicationController
   include Foreman::Controller::AutoCompleteSearch
   include Foreman::Controller::UsersMixin
   include Foreman::Controller::Parameters::User
+  include Foreman::Controller::BruteforceProtection
 
   skip_before_action :require_mail, :only => [:edit, :update, :logout]
   skip_before_action :require_login, :authorize, :session_expiry, :update_activity_time, :set_taxonomy, :set_gettext_locale_db, :only => [:login, :logout, :extlogout]
@@ -64,6 +65,14 @@ class UsersController < ApplicationController
   # Stores the user id in the session and redirects required URL or default homepage
   def login
     User.current = nil
+
+    if bruteforce_attempt?
+      inline_error _("Too many tries, please try again in a few minutes.")
+      log_bruteforce
+      render :layout => 'login', :status => :unauthorized
+      return
+    end
+
     if request.post?
       backup_session_content { reset_session }
       intercept = SSO::FormIntercept.new(self)
@@ -75,6 +84,7 @@ class UsersController < ApplicationController
       if user.nil?
         #failed to authenticate, and/or to generate the account on the fly
         inline_error _("Incorrect username or password")
+        count_login_failure
         redirect_to login_users_path
       else
         #valid user
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -2,6 +2,7 @@ module Api
   #TODO: inherit from application controller after cleanup
   class BaseController < ActionController::Base
     include ApplicationShared
+    include Foreman::Controller::BruteforceProtection
 
     protect_from_forgery
     force_ssl :if => :require_ssl?
@@ -167,7 +168,14 @@ module Api
     end
 
     def authorize
+      if bruteforce_attempt?
+        log_bruteforce
+        render_error('unauthorized', :status => :unauthorized)
+        return false
+      end
+
       unless authenticate
+        count_login_failure
         render_error('unauthorized', :status => :unauthorized, :locals => { :user_login => @available_sso.try(:user) })
         return false
       end
--- a/app/controllers/concerns/foreman/controller/bruteforce_protection.rb
+++ b/app/controllers/concerns/foreman/controller/bruteforce_protection.rb
@@ -1,0 +1,19 @@
+module Foreman::Controller::BruteforceProtection
+  extend ActiveSupport::Concern
+
+  def count_login_failure
+    Rails.cache.write("failed_login_#{request.ip}", get_login_failures+1, :expires_in => 5.minutes)
+  end
+
+  def get_login_failures
+    Rails.cache.fetch("failed_login_#{request.ip}") {0} if request.ip.present?
+  end
+
+  def bruteforce_attempt?
+    get_login_failures >= 30
+  end
+
+  def log_bruteforce
+    logger.warn("Brute-force attempt blocked from IP: #{request.ip}")
+  end
+end
--- a/test/controllers/api/base_controller_subclass_test.rb
+++ b/test/controllers/api/base_controller_subclass_test.rb
@@ -130,6 +130,14 @@ class Api::TestableControllerTest < ActionController::TestCase
       assert_response :unauthorized
     end
 
+    it "prevents brute-force attempts" do
+      @controller.expects(:authenticate).times(30).returns(false)
+      @controller.expects(:log_bruteforce).once
+      31.times do
+        get :index, {:user => 'admin', :password => 'brute-force'}
+      end
+    end
+
     context "and SSO (plain) authenticates" do
       setup do
         @sso = mock('dummy_sso')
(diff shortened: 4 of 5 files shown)


dLobatog:
@tbrisker I'm not sure this is really feasible by blocking based on a per-IP basis, or even on the `:login` method. 

**Foreman supports 4 ways of authenticating:**

1. Username/password (internal auth source)
2. Username/password (LDAP auth source)
3. Delegation API, REMOTE_USER , often used to implement Single Sign On. (External auth source)
4. OAuth (External auth source)

For 1 and 2, this approach is fine, but it produces many false positives as checking for the IP isn't a reliable method if you have a front-end load balancer. All requests to Foreman will go through the balancer, so if there are 5 legitimate login attempts (for different users), they will come from the same IP, blocking future login attempts (for any user!).
For 3 and 4, this approach doesn't do anything as those 2 methods don't use the login method at all. 

**How are we protected already?**

1. uses SHA1 (...) so we're not protected from brute force logins, it should be Bcrypt or some other tunable password-hashing function, not a general purpose message digest meant to be fast.
2. we are relatively protected as `conn.authenticate?(login, password)` will make 2 network requests to LDAP, to authenticate the admin and the user, which is going to be relatively slow. Still some systems like AD 2008 will use fast hashing functions so it's not like we're well covered in that area.
3. No protection, we could implement something like `restrict_registered_smart_proxies` but for hosts that can send REMOTE_USER.
4. No idea how long it takes to verify the OAuth signature, but it doesn't look good at first glance.

**What can we do**

I would suggest implementing a similar counter to this one, but doing so on a per-user basis. So, on every attempt to login, increment the counter and check if we've reached the limit. For 1 and 2, this can be done in the user model. For 3 and 4, I think it will have to be done in app/services/sso/apache and app/services/sso/oauth

lzap:
I agree that user is better handle to use, I also vote for increasing the initial value from 5 attempts to at least 10 or 20. Some sites insist on three attempts, that makes zero sense, brute force attack involves millions of rounds and 3 or 20 barely makes any difference, but UX is higher (I often forget password alternation and need to test 4-8 password variants until I get the correct one and this would prevent me from doing this).

tbrisker:
Closing this for now as good implementation is more complex then thought at first and I don't have capacity for it at the moment.

lzap:
I was thinking about this @dLobatog and @tbrisker and I think per-user protection is utterly complex and the way how @tbrisker initially implemented it is good enough. Doing this per remote IP is fine, all I'd suggest would be to increase the amount of allowed trials to reasonable number like 30 or 50. I think a configurable value would do this in case this causes problems behind NAT, value of zero would turn this feature off.

Why I am commenting this? I think it's pity to not continue on this patch, it's good and I think it is acceptable as is, just with the setting.

lzap:
The only changes I would like to see is to have at least 30 tries in 10 minutes and a configurable value to set this (or disable this).

Unless @dLobatog objects, I would like to carry the patch into develop.

tbrisker:
Thought about this some more and I think I'll reopen this. 
I'll increase the limit to 30 failed tries in 5 minutes, which should be enough to allow real users making mistakes but make brute-forcing impractical. Note that only failed attempts are counted, so that multiple legitimate users will not cause the counter to increase.
Locking out by user account can lead to DoS attacks, which is why I think IP-based blocking is better in this case. If there is a NAT in place this may still cause DoS but will at least still allow local access to users. It will also prevent brute-forcing logins (by alternating the login names instead of the password).
I prefer not to add another setting for this, we already have too many.
External auth brute-force prevention should be handled by the external auth source, not by us.

app/controllers/users_controller.rb
```diff
@@ -83,6 +84,9 @@ def login
         login_user(user)
       end
     else
+      if @brute_force_attempt
```
ohadlevy: maybe extract into a method?
```ruby
def brute_force_attempt?
  @brute_force_attempt
end
```

ohadlevy (commented):
Overall, i think this is a good improvement compared to before but IMHO in order to have a full feature, I woud like to discuss things like:
* Account log out policy (e.g. lock the account after X tries)
* consider the usage of CAPTCHA ? (not sure if we can get an offline version)
* consider having a failure lock period that is relative to the amount of failures? e.g. if its 3 failures its a minute, 5 failures would be 3 minutes, 10 failures would be an hour etc.

thanks!

app/controllers/users_controller.rb
lzap: Yeah that's nitpick but let's do it.

app/controllers/users_controller.rb
```diff
@@ -165,4 +169,16 @@ def verify_active_session
...
     end
   end
+
+  def count_failure(ip)
+    Rails.cache.write("failed_login_#{ip}", get_failures(ip)+1, :expires_in => 5.minutes)
+  end
+
+  def get_failures(ip)
+    Rails.cache.fetch("failed_login_#{ip}") {0}
+  end
```
lzap: Please add a check that IP is not empty string or nil because then it would essentially test for all users. I can imagine that in some constrained environments, IP could become empty or something.

app/controllers/users_controller.rb
```diff
@@ -165,4 +169,16 @@ def verify_active_session
...
+
+  def prevent_bruteforce
+    @brute_force_attempt = get_failures(request.ip) > 30
```
lzap: Can we make this configurable? I know we have many options, but I am pretty sure users (downstream) will ask for this.

app/controllers/users_controller.rb
```diff
@@ -83,6 +84,9 @@ def login
         login_user(user)
       end
     else
+      if @brute_force_attempt
+        inline_error _("Too many tries, please try again in a few minutes.")
```
lzap: Can you be more specific saying "five minutes" or is this on purspose? :-)

app/views/users/login.html.erb
```diff
@@ -36,7 +36,7 @@
           </div>
           <div class="form-group">
             <div class="col-xs-offset-8 col-xs-4 submit">
-              <%= submit_tag(_("Log In").html_safe, :class => "btn btn-primary", :data => { :disable_with => _("Log In") }) %>
+              <%= submit_tag(_("Log In").html_safe, :class => "btn btn-primary", :data => { :disable_with => _("Log In") }, :disabled => (@brute_force_attempt)) %>
```
lzap: Is this really necessary? It forces you to reload, I think we don't need this.

lzap (changes requested):
Nice thanks, overall I agree and I don't want to overengineer this - I am fine merging after my nitpicks.

app/controllers/users_controller.rb
tbrisker: done

app/controllers/users_controller.rb
tbrisker: on purpose :) not that it would be hard to figure out, but no point saying when the attacker should resume.

app/controllers/users_controller.rb
tbrisker: I suggest we make this configurable if we get requests for it, 30/5 min is high enough that users should never hit it in practice.

app/views/users/login.html.erb
tbrisker: removed

tbrisker:
Uploaded newer version:
* extracted the checks to a concern
* cleaned up code not to rely on instance variables
* added protection to api as well.
* added logging when brute-force has been detected.

@ohadlevy - These are all good suggestions, but I think that they may be overkill (gradually increasing time) or harder to implement (captcha on disconnected servers). a 5 minute delay for every 30 failed attempts should be good enough to make most brute-force attacks impractical.

lzap:
Test unrelated, gonna merge this once tested.

lzap (approved):

lzap:
It works fine, I agree that 5 mins 30 attempts per IP is sane default, although I would like to see this a configurable option. Amount of settings is not an argument, indeed we are big project. I am merging tho, if you change your mind feel free to file another PR later. Thanks for this patch!

lzap:
Prepared documentation PR: https://github.com/theforeman/theforeman.org/pull/985

dLobatog:
@tbrisker @lzap Did you consider https://github.com/kickstarter/rack-attack for this?

PR #4134, opened 2016-12-27
Fixes #17850 - Prevent 414 on architecture/usergroup

dLobatog:
These two fields had been modified using change_column in other
migrations which removed the limit on 'name'. This causes a 414 if
you create a very long name and try to edit the field.

After the change, the limit is restored so very long names are
forbidden by validates_length_in_database and 414s are prevented

### DIFF
--- a/db/migrate/20161227082709_change_architecture_name_limit.rb
+++ b/db/migrate/20161227082709_change_architecture_name_limit.rb
@@ -1,0 +1,5 @@
+class ChangeArchitectureNameLimit < ActiveRecord::Migration
+  def change
+    change_column :architectures, :name, :string, :default => nil, :limit => 255
+  end
+end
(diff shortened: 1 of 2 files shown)


dLobatog:
[test foreman]
[test katello]

tbrisker:
These limits are already in place AFAICT, which migration removed them? Also, architecture name cannot be nil.

dLobatog:
@tbrisker Thanks for taking a look, here's what happened: 

* The limits are removed by `db/migrate/20150602153751_change_usergroup_name_to_be_required.rb` and `db/migrate/20140707113214_remove_architecture_default.rb`, where the column was changed without setting a limit. A way to reproduce this is to create a database from scratch, migrate up to that migration, and check `\d+ architectures` on `psql`, then continue the migration and run `\d+ architectures` again, you'll see how the limit is removed.
* I purposefully kept the default as the same of http://projects.theforeman.org/issues/6506

I rebased and pushed again to make sure tests pass (this PR was submitted prior to the concurrent-ruby-edge fix)

tbrisker:
@dLobatog Huh, I tried creating a new db and indeed the limits were removed. very strange it didn't appear on my older db, possibly rails changed something in how change_columns works? anyways, merging, thanks.

PR #4135, opened 2016-12-27
Fixes #17852 - Pin concurrent-ruby-edge to 0.2.3

dLobatog:
test_develop is red due to the foreman_tasks -> dynflow ->
concurrent_ruby_edge dependency. 0.3.0 Promises are incompatible with
0.2.3 Futures.

https://github.com/Dynflow/dynflow/pull/212 is meant to fix it there,
but since all committers are out we have to fix it in Foreman until a
new dynflow release comes out.

### DIFF
--- a/bundler.d/foreman_tasks.rb
+++ b/bundler.d/foreman_tasks.rb
@@ -1,3 +1,4 @@
 group :foreman_tasks do
   gem 'foreman-tasks', '>= 0.8.5'
+  gem 'concurrent-ruby-edge', '0.2.3'
 end


dLobatog:
[test upgrade]

dLobatog:
[test upgrade]

dLobatog:
The foreman tests job tries to install `nokogiri 1.6.8` (which works on ruby 2.0) 'http://ci.theforeman.org/job/test_develop_pr_core/11543/database=postgresql,ruby=2.0.0,slave=fast/consoleFull' because roadie-rails requires it.  

However on upgrades (from 1.11-stable!) there's no `roadie-rails` to require `nokogiri < 1.7.0`. It tries to install `nokogiri 1.7.0` on ruby 2.0.0 and it fails. I've fixed it by pinning nokogiri too. 

edit: nope, it has to be fixed on 1.11-stable, I don't think that makes sense

dLobatog:
https://github.com/theforeman/foreman-infra/pull/265 is supposed to fix the upgrade job by upgrading from a more recent version

dLobatog:
(in any case, this PR can be merged to fix all jobs but test_develop_upgrade)

ohadlevy:
thanks @dLobatog, I can confirm this solves the current issue.

PR #4136, opened 2016-12-27
Fixes #17851 - Set hostgroup taxonomies after saving host

dLobatog:
To reproduce:

1. Have a host with no hostgroup but some location set
2. Have a hostgroup with no location set
3. Assign the hostgroup to the host
4. Access the hostgroup's "Locations" tab

Now it seems the hostgroup already has a location active which is not
true. The hostgroup still has no location.
The location is assigned only after you click submit, even though it
looks like its already assigned.

### DIFF
--- a/app/models/concerns/hostext/taxonomies.rb
+++ b/app/models/concerns/hostext/taxonomies.rb
@@ -1,0 +1,28 @@
+module Hostext
+  module Taxonomies
+    extend ActiveSupport::Concern
+
+    included do
+      if SETTINGS[:organizations_enabled]
+        validates :organization_id, :presence => true, :if => ->(host) { host.managed? }
+      end
+      if SETTINGS[:locations_enabled]
+        validates :location_id, :presence => true, :if => ->(host) { host.managed? }
+      end
+
+      after_validation :fix_hostgroup_mismatches, :if => ->(host) { host.hostgroup_id.present? }
+    end
+
+    def fix_hostgroup_mismatches
+      Taxonomy.enabled_taxonomies.each do |taxonomy|
+        host_taxonomy = public_send(taxonomy.singularize.to_sym)
+        next if host_taxonomy.blank?
+        TaxableTaxonomy.where(
+          :taxonomy_id  => host_taxonomy.id,
+          :taxable_id   => hostgroup_id,
+          :taxable_type => 'Hostgroup'
+        ).first_or_create
+      end
+    end
+  end
+end
--- a/app/models/host/managed.rb
+++ b/app/models/host/managed.rb
@@ -3,6 +3,7 @@ class Host::Managed < Host::Base
   include Hostext::Search
   include Hostext::SmartProxy
   include Hostext::Token
+  include Hostext::Taxonomies
   include SelectiveClone
   include HostParams
   include Facets::ManagedHostExtensions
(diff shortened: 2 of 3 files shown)


app/models/concerns/hostext/taxonomies.rb
```diff
@@ -0,0 +1,20 @@
+module Hostext
+  module Taxonomies
+    extend ActiveSupport::Concern
+
+    included do
+      validates :organization_id, :presence => true, :if => ->(host) { host.managed? && SETTINGS[:organizations_enabled] }
+      validates :location_id,     :presence => true, :if => ->(host) { host.managed? && SETTINGS[:locations_enabled] }
```
ohadlevy: nitpick: I would extract the setting check condition to the entire line vs a if within the lambda.

app/models/concerns/hostext/taxonomies.rb
```diff
@@ -0,0 +1,20 @@
...
+      after_validation :fix_hostgroup_mismatches, :if => ->(host) { host.hostgroup.present? }
```
ohadlevy: nitpick: use host.hostgroup_id to avoid extra sql call.

app/models/concerns/hostext/taxonomies.rb
```diff
@@ -0,0 +1,20 @@
...
+    end
+
+    def fix_hostgroup_mismatches
+      Taxonomy.enabled_taxonomies.each do |taxonomy|
+        host_taxonomy = public_send(taxonomy.singularize.to_sym)
+        next if host_taxonomy.blank? ||
+          hostgroup.public_send(:"#{taxonomy}").include?(host_taxonomy)
```
ohadlevy: nitpick: you can work on the join table direct instead of looping though hostgroup, organizations and locations.

e.g. to save some extra un-required sql you can
```ruby
TaxableTaxonomy.first_or_create(:taxonomy => taxonomy, :taxable => hostgroup)
```
or even 
```ruby
TaxableTaxonomy.first_or_create(:taxonomy_id => host.taxonomy_id, :taxable_id => host.hostgroup_id, :taxable_type => 'Hostgroup')
```

dLobatog:
@ohadlevy Repushed to address your comments inline:

* Taxonomy checks for the validations are moved outside of the if so they only run on initialization
* Fixing mismatches is done through the TaxableTaxonomies table to speed it up.
* Called host.hostgroup_id where I can to save time

ohadlevy:
@dLobatog note the failing tests.

ohadlevy:
@dLobatog I'm trying to figure out what exactly is the user story here, it looks to me like this solution would surprise the user, please correct me if i'm wrong, but this is how i read it.

- a host can change the hostgroup

if thats true, I don't that makes any sense, and we should not allow the initial condition rather then allow this kind of workflow.

dLobatog:
This also affects 'users' and 'usergroups' as they can be set as Host.owner

dLobatog:
cc @lzap - this is the problem we had with Domains that are not in an Organization even though they appear like they are on it in the UI (Organizations tab). Therefore people get confused when the Domain does not show up in the UI.

This is a problem for Subnet, Domains, and any other thing that is "fixed on mismatch" when one assigns a Organization/Location to a Host

PR #4137, opened 2016-12-27
Fixes #17087 - default and overrides values converted to string

orrabin:


dLobatog:
[test foreman]
[test katello]

dLobatog:
@orrabin Why is this not handled on `Foreman::Parameters::Validator#validate!` (which is what validates `default_value`.

Shouldn't that validate that `default_value` is a string, or is the problem that `cast_default_value` goes wrong and does not give you info when you pass non-string values?

orrabin:
@dLobatog the problem is we cast the value and then save it as a non string, if I move it to the validation it will fail the next time the validation runs.

dLobatog:
@orrabin I see, I've been trying different things to move this to the model but nothing really convinces me, it's a problem at the controller level, the model is doing its job just fine. I see other problems, though , let me comment through a review

app/controllers/api/v2/override_values_controller.rb
```diff
@@ -104,6 +105,11 @@ def rename_use_puppet_default
         params[:override_value][:omit] = params[:override_value].delete(:use_puppet_default)
         Foreman::Deprecation.api_deprecation_warning('"use_puppet_default" was renamed to "omit"')
       end
+
+      def validate_value
+        return unless params[:override_value] && params[:override_value].key?(:value)
+        render_error(:custom_error, :locals => {:message => _('value must be a string')}, :status => :bad_request) unless params[:override_value][:value].is_a?(String)
```
dLobatog: This error can be misleading, it says `value must be a string`  when it can be boolean, integer, etc.. but contained in a string. I'm not sure how to word it, maybe `value must be contained in a string. e.g: if you want to send a boolean true, send 'true' as a string`.

app/controllers/concerns/api/v2/lookup_keys_common_controller.rb
```diff
@@ -125,6 +127,12 @@ def return_if_smart_mismatch
     end
   end
 
+  def validate_default_value
+    obj = smart_variable_id? ? "smart_variable" : "smart_class_parameter"
+    return unless params[obj] && params[obj].key?(:default_value)
+    render_error(:custom_error, :locals => {:message => _('default value must be a string')}, :status => :bad_request) unless params[obj][:default_value].is_a?(String)
```
dLobatog: Same problem as in the other controller

test/controllers/api/v2/override_values_controller_test.rb
```diff
@@ -128,4 +128,46 @@ class Api::V2::OverrideValuesControllerTest < ActionController::TestCase
     end
     assert_response :error
   end
+
+  test "should not accept boolean that is not sent as string" do
```
dLobatog: Is there a test for accepting booleans contained in strings, e.g: testing that I pass `"true"` and the overridden value is actually saved as a boolean? It works but I had to test it manually.

test/controllers/api/v2/override_values_controller_test.rb
```diff
@@ -128,4 +128,46 @@ class Api::V2::OverrideValuesControllerTest < ActionController::TestCase
...
+    lookup_key = FactoryGirl.create(:variable_lookup_key, :puppetclass => puppetclasses(:two), :key_type => 'boolean')
+    post :create, {:smart_variable_id => lookup_key.id, :override_value =>  { :match => 'os=string', :value => true}}
+    assert_response 400
```
dLobatog: nitpick `:bad_request`

test/controllers/api/v2/override_values_controller_test.rb
```diff
@@ -128,4 +128,46 @@ class Api::V2::OverrideValuesControllerTest < ActionController::TestCase
...
+    post :create, {:smart_variable_id => lookup_key.id, :override_value =>  { :match => 'os=string', :value => ['a', 'b']}}
+    assert_response 400
+  end
+
+  test "should accept array as string" do
+    lookup_key = FactoryGirl.create(:variable_lookup_key, :puppetclass => puppetclasses(:two), :key_type => 'array')
+    assert_difference('LookupValue.count', 1) do
+      post :create, {:smart_variable_id => lookup_key.id, :override_value =>  { :match => 'os=string', :value => "['a', 'b']"}}
+    end
+    assert_response :success
```
dLobatog: nitpick, `:created`

app/controllers/api/v2/override_values_controller.rb
```diff
@@ -104,6 +105,11 @@ def rename_use_puppet_default
         params[:override_value][:omit] = params[:override_value].delete(:use_puppet_default)
         Foreman::Deprecation.api_deprecation_warning('"use_puppet_default" was renamed to "omit"')
       end
+
+      def validate_value
```
dLobatog: A comment explaining why we're validating things in the API vs models would be helpful for other devs

test/controllers/api/v2/smart_class_parameters_controller_test.rb
```diff
@@ -192,4 +192,40 @@ class Api::V2::SmartClassParametersControllerTest < ActionController::TestCase
     get :show, {:id => lookup_keys(:five).id}, set_session_user.merge(:user => users(:one).id)
     assert_response :success
   end
+
+  test "should not accept boolean that is not sent as string" do
```
dLobatog: This code is duplicated because you're testing a concern, can you test it in `/test/controllers/shared` or something similar so that the code isn't basically identical in 2 places?

test/integration/host_js_test.rb
```diff
@@ -25,8 +25,8 @@ class HostJSTest < IntegrationTestWithJavascript
     test 'class parameters and overrides are displayed correctly for strings' do
       host = FactoryGirl.create(:host, :with_puppetclass)
       FactoryGirl.create(:puppetclass_lookup_key, :as_smart_class_param, :with_override,
-                                      :key_type => 'string', :default_value => true, :path => "fqdn",
-                                      :puppetclass => host.puppetclasses.first, :overrides => {host.lookup_value_matcher => false})
+                                      :key_type => 'string', :default_value => 'true', :path => "fqdn",
```
dLobatog: I don't understand these changes nor the changes to unit tests. The model still takes in booleans, integers, arrays, etc.. it's just the API controllers should only take strings. Why changing the unit tests to test only strings? At the model level we don't need to cast every time.

dLobatog (changes requested):
@orrabin Thanks - 3 issues commented inline, duplicate code, changes in unit tests and error messages when this happens in the controllers.

orrabin:
@dLobatog this solution looks better to me, what do you think?
tests are still duplicated but I can combine them.

app/controllers/concerns/api/v2/lookup_keys_common_controller.rb
```diff
@@ -125,6 +127,14 @@ def return_if_smart_mismatch
     end
   end
 
+  def cast_default_value
+    obj = smart_variable_id? ? "smart_variable" : "smart_class_parameter"
+
+    return unless params[obj] && params[obj].key?(:default_value) && !params[obj][:default_value].is_a?(String)
```
dLobatog: This code is basically a duplicate of `cast_value` in override_values_controller.rb, please refactor to remove the duplication

test/controllers/api/v2/smart_variables_controller_test.rb
```diff
@@ -102,4 +102,46 @@ class Api::V2::SmartVariablesControllerTest < ActionController::TestCase
       assert_equal parameter.hidden_value, show_response['default_value']
     end
   end
+
+  test "should accept boolean that is not sent as string" do
```
dLobatog: All three files here have 42 new lines of testing the same thing, the logic of `cast_value`. Once you extract that logic to a common place (app/services sounds fine but it's your choice), just test that logic somewhere *once*, and use `use_before_action` in the controller tests to ensure the controllers run it before the defined actions https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/action_controller/callback_matcher.rb#L54

dLobatog (changes requested):
@orrabin Thanks, the new way seems clear to me, I'm only concerned about the duplication in code and tests (so is codeclimate as it drops the score after this PR). Check out the comments inline.

orrabin:
@dLobatog I refactored the casting and tests.

app/controllers/concerns/api/v2/lookup_keys_common_controller.rb
```diff
@@ -125,6 +127,19 @@ def return_if_smart_mismatch
...
   end
 
+  def cast_default_value
+    obj = smart_variable_id? ? "smart_variable" : "smart_class_parameter"
+
+    cast_value(obj, :default_value)
+  end
+
+  def cast_value(obj = :override_value, value = :value)
+    return unless params[obj] && params[obj].key?(value) && !params[obj][value].is_a?(String)
```
tbrisker: This condition is quite complex, why do we care about doing another `.to_s` if it's a string?

app/controllers/concerns/api/v2/lookup_keys_common_controller.rb
```diff
@@ -125,6 +127,19 @@ def return_if_smart_mismatch
...
+    value = params[obj][value]
+    params[obj][value] = value if value.is_a?(Hash)
```
tbrisker: This doesn't make sense to assign here then do another assignment on the next line.

test/controllers/concerns/lookup_keys_common_controller_test.rb
```diff
@@ -0,0 +1,36 @@
...
+  def self.before_action(*args)
+    self.callbacks = args
+  end
+
+  include Api::V2::LookupKeysCommonController
+end
+
+class Api::V2::LookupKeysCommonControllerTest < ActiveSupport::TestCase
+  setup do
+    @dummy = DummyController.new
```
tbrisker, line 16: You can use `tests DummyController` and have `@controller` set to it correctly, and then you can use `ActiveController::TestCase`

tbrisker (changes requested):
see a couple of inline comments

test/controllers/concerns/lookup_keys_common_controller_test.rb
orrabin: I'm testing the functionality of a concern, I don't actually need this to have all controller functionality, that is already tested in the controllers that include `Api::V2::LookupKeysCommonController`  so I tested this like other concerns.

orrabin:
@tbrisker I addressed one comment inline and fixed the rest of the comments

tbrisker (approved):

tbrisker:
[test katello] just to make sure it's unrelated

orrabin:
rebased.
@tbrisker is there anything else missing besides the tests?

tbrisker:
Thanks @orrabin ! Sorry I forgot to merge in April...

ares:
Thanks @tbrisker for merging, please don't forget to set Release in redmine issue until we have prprocessor to do it automatically.

PR #4139, opened 2016-12-27
fixes #17853 - resolve ips in interface scope

timogoebel:
`to_ip_address` should be executed in nic's scope, not in host's scope. This refactors the resolving into a class and adds some tests.
I saw the issue using Discovery when changing the provision interface to a bond (e.g. provision a discovered host, create a new bond interface and set it as managed, primary and provisioning; save the host).

### DIFF
--- a/app/models/host/managed.rb
+++ b/app/models/host/managed.rb
@@ -912,20 +912,9 @@ class Host::Managed < Host::Base
     result
   end
 
-  # converts a name into ip address using DNS.
-  # if we are managing DNS, we can query the correct DNS server
-  # otherwise, use normal systems dns settings to resolv
   def to_ip_address(name_or_ip)
-    return name_or_ip if name_or_ip =~ Net::Validations::IP_REGEXP
-    if dns_record(:ptr4)
-      lookup = dns_record(:ptr4).dns_lookup(name_or_ip)
-      return lookup.ip unless lookup.nil?
-    end
-    # fall back to normal dns resolution
-    domain.resolver.getaddress(name_or_ip).to_s
-  rescue => e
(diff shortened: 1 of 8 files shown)


dLobatog:
[test foreman]
[test katello]

timogoebel:
Rebased to fix the test failures.

app/models/host/managed.rb
```diff
@@ -912,22 +912,6 @@ def recreate_config(only = nil)
     result
   end
 
-  # converts a name into ip address using DNS.
-  # if we are managing DNS, we can query the correct DNS server
-  # otherwise, use normal systems dns settings to resolv
-  def to_ip_address(name_or_ip)
```
tbrisker, line 918: Should we deprecate this first in case some plugin uses it? I didn't find any that does but there may one.

tbrisker:
@timogoebel Looks good in general, just one comment/question inline.

app/models/host/managed.rb
timogoebel: @tbrisker : Thanks for reviewing. I don't think, this is necessary. But it's easily done. Let me know and I'll get started.

dLobatog: Yeah - please do it with Foreman::Deprecation. It shouldn't take much time & we ensure nothing breaks for sure. Maybe some test in some plugin indirectly relies on this, we'd rather not break anything without warning.

timogoebel: Done, thanks @tbrisker and @dLobatog.

tbrisker:
Thanks @timogoebel!

PR #4141, opened 2016-12-27
Fixes #17858 - OpenStack scheduler doesn't show spinner

dLobatog:
Foreman does not show the spinner when the OpenStack scheduler hints are
loading, instead it just shows a javascript error.


tbrisker:
Thanks @dLobatog!

PR #4142, opened 2016-12-28
Fixes #17864 - Show smart proxy errors on unattended fail

dLobatog:
If your proxy is has some error (like misconfigured sudoers, puppet not
available, etc...) and you try to boot a host through unattended
mode [1], the exception will not show up when you try to fetch the
kickstart template. Instead, it will show something like No HTTP_REFERER
was set in the request to this action, so redirect_to :back could not be
called successfully.

The application controller should be aware of this situation and don't
try to redirect to back in it, instead just render the error.

[1] an example of such error - ProxyAPI::ProxyException: ERF12-0104
[ProxyAPI::ProxyException]: Unable to set PuppetCA autosign for
tonia-rippeon.lobatolan.home ([RestClient::NotAcceptable]: 406 Not
    Acceptable) for proxy
https://centos7-devel.lobatolan.home:9090/puppet/ca

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -114,8 +114,13 @@ class ApplicationController < ActionController::Base
   end
 
   def smart_proxy_exception(exception = nil)
-    process_error(:redirect => :back, :error_msg => exception.message)
     Foreman::Logging.exception("ProxyAPI operation FAILED", exception)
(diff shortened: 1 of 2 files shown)


dLobatog:
[test upgrade]

tbrisker:
Thanks @dLobatog !

PR #4144, opened 2016-12-29
Fixes #17870 - Empty boolean matcher should not turn into false

orrabin:


app/models/lookup_value.rb
```diff
@@ -44,6 +44,7 @@ def value_before_type_cast
   end
 
   def validate_value
+    validate_and_cast_value
```
tbrisker, line 47: since this no longer runs in a callback, it doesn't need to return true/false, which means it can be greatly simplified.

tbrisker (changes requested):

tbrisker (approved):

dLobatog:
Thanks @orrabin @tbrisker !

PR #4145, opened 2016-12-29
Fixes #17872 - Update button label from New to Create

junezhang:
Based [Terminology and Wording Style Guide](http://www.patternfly.org/styles/terminology-and-wording/#_), If what you are creating is not readily apparent from the context, consider adding a noun to the button label. (e.g. Create User).  "New" or "Add" are not recommended for this use case. 

So we suggest to make a change from New xxx to Create xxx.

### DIFF
--- a/app/helpers/environments_helper.rb
+++ b/app/helpers/environments_helper.rb
@@ -3,7 +3,7 @@ module EnvironmentsHelper
 
   def environments_title_actions
     title_actions import_proxy_select(hash_for_import_environments_environments_path),
-                  button_group(new_link(_('New Puppet Environment'))),
+                  button_group(new_link(_('Create Puppet Environment'))),
                   button_group(help_button)
   end
 end
--- a/app/helpers/hosts_helper.rb
+++ b/app/helpers/hosts_helper.rb
@@ -65,7 +65,7 @@ module HostsHelper
   end
 
   def new_host_title
-    t = _("New Host")
+    t = _("Create Host")
     title(t, (t + ' <span id="hostFQDN"></span>').html_safe)
   end
 
--- a/app/services/menu/loader.rb
+++ b/app/services/menu/loader.rb
@@ -55,7 +55,7 @@ module Menu
 
         menu.sub_menu :hosts_menu,      :caption => N_('Hosts') do
           menu.item :hosts,             :caption => N_('All hosts')
-          menu.item :newhost,           :caption => N_('New host'),
+          menu.item :newhost,           :caption => N_('Create host'),
                     :url_hash => {:controller => '/hosts', :action => 'new'}
           if SETTINGS[:unattended]
             menu.divider                :caption => N_('Provisioning Setup')
--- a/app/views/architectures/index.html.erb
+++ b/app/views/architectures/index.html.erb
@@ -1,5 +1,5 @@
 <% title _("Architectures") %>
-<% title_actions new_link(_("New Architecture")), help_button %>
+<% title_actions new_link(_("Create Architecture")), help_button %>
 
 <table class="<%= table_css_classes 'table-two-pane table-fixed' %>">
   <thead>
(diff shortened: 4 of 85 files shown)


tbrisker:
ok to test

orrabin:
For the record: corresponding issues should be opened for existing plugins.

dLobatog:
@junezhang Based on your comments on https://groups.google.com/forum/?fromgroups#!topic/foreman-dev/jy2Ytas4SRw , aren't these buttons missing `pficon-add-circle-o` or did I misunderstand it and that's for the 'Submit' buttons?

omaciel:
I apologize in advance for I'm looking at this PR from my phone but do these buttons/links have unique IDs? If not, could you add them? Unique IDs allow QE to fully automate the UI and not have to rely on text values.

junezhang:
@dLobatog Your understanding is correct. but in this time, we suggested if the button includes icon, then we updated it to PF icon, as if the button doesn't include icon at all, then we won't add the icon to them.
Because in Foreman different plugins, we can see some buttons have icon, some buttons doesn't have icon. It is not consistent, This is what we will do next to make the guideline for how to use icon alone, icon+label, and label alone. Please waiting for our conclusion. :-)

junezhang:
@omaciel Thanks for your comment. I've got your meaning. But I found in the finally page, we can see there is an ID for every label. For example: Create Modal. the code is following:
`<a class="btn btn-primary " data-id="aid_models_new" href="/models/new">Create Model</a>`
I don't know if we could use this data-id.
At this time, I've fixed all labels in the testing file. all the checks have passed. :-)

dLobatog:
@junezhang Oh I see now, thanks - I understood the email to foreman-dev as if you folks thought having the icon would be the better choice everywhere, so I even started doing it: https://github.com/Katello/katello/pull/6515 I guess I'll keep buttons without icons to ensure they're consistent until we decide to put the icon everywhere?

ares:
@junezhang I don't think data-id is what @omaciel is looking for as it's not necessarily unique on the page. We should provide unique and predictable id for all our clickable elements so in this case a patch to `new_link` helper would be required which I think is out of scope for this PR. @omaciel would you mind reporting this need as a separate issue? Ideally list element types you need.

ohadlevy:
On Mon, Jan 2, 2017 at 10:04 AM, Marek Hulán <notifications@github.com>
wrote:

> @junezhang <https://github.com/junezhang> I don't think data-id is what
> @omaciel <https://github.com/omaciel> is looking for as it's not
> necessarily unique on the page. We should provide unique and predictable id
> for all our clickable elements so in this case a patch to new_link helper
> would be required which I think is out of scope for this PR. @omaciel
> <https://github.com/omaciel> would you mind reporting this need as a
> separate issue? Ideally list element types you need.
>

@ares - isn't that already handled via
https://projects.theforeman.org/issues/3751 ?

omaciel:
@ares sure can, will wait until you reply to @ohadlevy question to see if it is necessary. And you're absolutely correct that I would like to have unique, predictable IDs for all clickable elements for our ui automation.

junezhang:
@dLobatog Yea, that's right. :-)

ares:
@ohadlevy most of it yes, e.g. ordering links in table headers, pagination buttons or tabs do not have it. Since it's based on URL of the link, there's a chance it's not unique if there are two different links leading to the same target. Which is I think why it's not commonly used id attribute but data attribute which I think is less testing friendlier still usable though. Currently data-id is hopefully not unique only for `aid_not_defined` value, I haven't found another dup example. So @omaciel I think you should open separate issues for things that do not have it if you need it.

dLobatog:
Is there anything missing here?

junezhang:
@dLobatog I've checked again, nothing is missed. Thank You.

dLobatog:
@junezhang Thanks - I'm merging so if any problems with automation such as what @omaciel suggested arise we're able to discover them as quickly as possible.

PR #4146, opened 2016-12-29
fixes #4166 - vmware: update cpus, memory for vms

timogoebel:
This commit adds support for editing cpus, corespersocket and memory of an existing vm. All other form fields are disabled when editing a vm.

image

 🔥 Video demonstrating the change: http://recordit.co/KwTP31J5vc 🔥 

Corresponding PR against fog-vsphere: https://github.com/fog/fog-vsphere/pull/63

### DIFF
--- a/app/helpers/compute_resources_vms_helper.rb
+++ b/app/helpers/compute_resources_vms_helper.rb
@@ -159,9 +159,9 @@ module ComputeResourcesVmsHelper
     host.try(:new_record?)
   end
 
-  def vsphere_resource_pools(form, compute_resource)
+  def vsphere_resource_pools(form, compute_resource, new_host = false)
     resource_pools = compute_resource.available_resource_pools(:cluster_id => form.object.cluster) rescue []
-    selectable_f form, :resource_pool, resource_pools, { }, :class => "col-md-2", :label => _('Resource pool')
+    selectable_f form, :resource_pool, resource_pools, { }, :class => "col-md-2", :label => _('Resource pool'), :disabled => !new_host
   end
 
   def vms_table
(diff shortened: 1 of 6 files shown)


dLobatog:
@timogoebel Notice you can add https://github.com/fog/fog-vsphere/pull/63 to a fog-extension for the time being, so that we can merge this already 😄 Thanks!

timogoebel:
@dLobatog : Yes, thanks. However I don't think, that's worth it. The maintainer of fog-vsphere merges that stuff usually pretty fast and releases a new version of the gem afterwards. I'd suggest to wait just a little bit more. But thanks for the hint. :-)

timogoebel:
fog-vsphere 1.6.0 is now available and contains the necessary prerequisites for this PR: https://rubygems.org/gems/fog-vsphere/versions/1.6.0

timogoebel:
[test foreman]

timogoebel:
[test upgrade]

mmoll:
@timogoebel could you bump the version requirement in bundler.d/vmware.rb? I guess that version specification is out of date because of earlier fixes anyway?

timogoebel:
@mmoll : Sure, done. The version requirement is totally out of date. Just curious, is that actually relevant for Debian packaging?

mmoll:
@timogoebel implicitly very little... we simply do "bundle install/package" in the Debian packaging process. For RPM it's more relevant, as things will fail and thus get noticed if the fog-vsphere RPM isn't updated in such a case.

timogoebel:
[test foreman]

dLobatog:
Thanks @mmoll @timogoebel !

PR #4148, opened 2016-12-30
fixes #8290 - remove foreman_url redundancy

timogoebel:


lib/foreman/renderer.rb
```diff
@@ -2,13 +2,8 @@
...
 module Foreman
   module Renderer
-    # foreman_url macro uses url_for, therefore we need url helpers and fake default_url_options
-    # if it's not defined in class the we mix into
-    include Rails.application.routes.url_helpers
-
-    def default_url_options
-      {}
-    end
+    extend ActiveSupport::Concern
```
dLobatog: What is this `ActiveSupport::Concern` for? I think it's causing the error in Katello and bootdisk (or any plugins that extend controllers who implement `Foreman::Renderer`)

with it:

```
UnattendedController
NameError: undefined method `kickstart_attributes' for class `UnattendedController'
        from /home/vagrant/.rvm/gems/ruby-2.2.4/gems/activesupport-4.2.7.1/lib/active_support/core_ext/module/aliasing.rb:32:in `alias_method'
        from /home/vagrant/.rvm/gems/ruby-2.2.4/gems/activesupport-4.2.7.1/lib/active_support/core_ext/module/aliasing.rb:32:in `alias_method_chain'
        from /home/vagrant/katello/app/lib/katello/concerns/renderer_extensions.rb:7:in `block in <module:RendererExtensions>'
```

without it - no error

dLobatog (changes requested):
@timogoebel The implementation is good in my opinion, thanks! There seems to be some issues with plugins that extend the controllers with the concern, can you take a look?

lib/foreman/renderer.rb
timogoebel: Yeah, that's not necessary. I just removed it.

dLobatog (approved):

dLobatog:
Thanks @timogoebel

PR #4149, opened 2017-01-01
Fixes #17886 - move config groups under parameters

ares:


dLobatog:
@ares Seems like the test in host_test:2653 broke, fix is the same as you did in your test here - I agree with the premise of the issue, but can you add a deprecation notice and leave it in both places in case anyone is relying on it? This seems like one of those things that break scripts and things like that.

Thanks!

ares:
@dLobatog I hope I've fixed the second test as well. I don't think we should deprecate it, it was not part of any release, it's targeted to 1.14 which is rc2 atm. Keeping it in ENC output would "break" https://docs.puppet.com/guides/external_nodes.html#enc-output-format (not that puppet master would care). Out of curiosity, how would you deprecate it? Once we put it into YAML, I don't see a way to find out user is reading this data from this key.

dLobatog:
@ares Ah my bad, I didn't look at when it was introduced - I'll tag it for 1.14.0 here. My thought was to deprecate it just server-side, you can show a warning on those lines if the config_groups attr is created on an ENC call. Anyway that's moot now, will merge if tests pass 😄

dLobatog:
Test failure is irrelevant - merging

dLobatog:
Thanks @ares!

PR #4150, opened 2017-01-02
Fixes #17894 - load needed dependent api param group

jlsherrill:
when rendering facet api param docs

### DIFF
--- a/app/services/facets/entry.rb
+++ b/app/services/facets/entry.rb
@@ -80,6 +80,10 @@ module Facets
       @compatibility_properties = property_symbols
     end
 
+    def load_api_controller
+      @api_controller.to_s.constantize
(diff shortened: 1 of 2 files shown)


app/services/facets/entry.rb
```diff
@@ -80,6 +80,10 @@ def template_compatibility_properties(*property_symbols)
       @compatibility_properties = property_symbols
     end
 
+    def load_api_controller
+      @api_controller.to_s.constantize
```
jlsherrill, line 84: Note, this was inspired by https://github.com/Apipie/apipie-rails/blob/master/lib/apipie/application.rb#L413

dLobatog:
Thanks @jlsherrill !

PR #4151, opened 2017-01-03
Fixes #17902 - Smart variable puppet class is not auto selected

orrabin:


app/helpers/lookup_keys_helper.rb
```diff
@@ -24,7 +24,8 @@ def show_puppet_class(f)
       text_f(f, :puppetclass_id, :label => _('Puppet Class'), :value => f.object.param_class, :disabled => true)
     else # new smart-var with no particular context
          # Give a select for choosing the parent puppetclass
-      select_f(f, :puppetclass_id, Puppetclass.all, :id, :to_label, {}, {:label => _("Puppet class")})
+      puppetclasses = accessible_resource(f.object, :puppetclass)
+      select_f(f, :puppetclass_id, puppetclasses, :id, :to_label, { :include_blank => true }, {:label => _("Puppet class")})
```
ohadlevy, line 28: does the error message shows up? @dLobatog complaint was that it was only visible in the logs?

orrabin: it was only in the logs because the validation was on puppetclass and not puppetclass_id and the field is on puppetclass_id

app/models/lookup_keys/variable_lookup_key.rb
```diff
@@ -16,8 +16,12 @@ def param_class
     puppetclass
   end
 
-  def self.humanize_class_name
-    "Smart variable"
+  def self.humanize_class_name(options = nil)
```
orrabin, line 23: After changing the validation the error added tries to call `VariableLookupKey.human_attribute_name(:puppetclass_id)` to use in the error message. This got to `humanize_class_name` and failed because it doesn't expect any attributes.

app/models/lookup_keys/variable_lookup_key.rb
```diff
@@ -16,8 +15,19 @@ def param_class
...
+    else
+      "Smart variable"
+    end
+  end
+
+  def validate_puppet_class_presence
+    # Variable lookup key can be created through puppet class.
+    # If this is a new puppet class the id is still nil so validating presence on puppetclass_id will fail.
+    # Adding the validation on the puppet class itself will not show up in the UI because the displayed field is puppetclass_id.
+    errors[:puppetclass_id] = _("can't be blank") unless self.puppetclass.present?
```
orrabin: I changed the validation because this test was failing: https://github.com/theforeman/foreman/blob/develop/test/models/puppetclass_test.rb#L46
I don't think we can actually get to this case through the api/ui but the test makes me think there may be some way to so this through the importer so I changed the validation.
If this really isn't possible I'd prefer to change the test and move the validation back to:
`validates :puppetclass_id, :presence => true`

tbrisker: There is an api for creating new puppet classes. However not sure if the test is a good reflection of what happens when creating using the api, maybe try moving that test to the api to see what it does when attempting to create a puppetclass with smart variables via the API.

orrabin:
@tbrisker I changed the validation to use `validates :puppetclass_id, :presence => true` with a condition on the puppetclass instead of a custom validation.

ohadlevy:
@tbrisker test failure seems unrelated?

tbrisker:
@ohadlevy correct, it was fixed already in another PR. merging.

tbrisker:
Thanks @orrabin !

PR #4152, opened 2017-01-03
Fixes #14714 - Make inherited Puppet Classes more apparent

bastilian:
To better recognise which Puppet Classes are inherited for Host
or Hostgroup they are under a sub-heading as a separated list and mentions the Hostgroup inherited from by name in the tooltip.

new host group - puppet classes

The issue suggests to have the Hostgroup the class is inherited from as part of the module name, should this be as following the pattern of `Parent-Hostgroup/module-name` or is the approach with with the header and the mention in the tooltip already a good approach to address #14714?

### DIFF
--- a/app/assets/stylesheets/patternfly_and_overrides.scss
+++ b/app/assets/stylesheets/patternfly_and_overrides.scss
@@ -319,3 +319,30 @@ a.btn-info span{
 .pficon-ok.warn:before {
   color: #ec7a08;
 }
+
+.paneless {
+  .panel {
+    box-shadow: none;
+  }
+
+  .panel-title {
+    font-size: inherit;
+  }
+
+  .panel-heading + .panel-collapse .panel-body {
+    border: none;
+    padding: 0;
+  }
+
+  .panel-heading {
+    background: none;
+    border: none;
+    padding-left: inherit;
+    padding-right: inherit;
+
+    a {
+      font-weight: inherit;
+      font-size: inherit;
+    }
+  }
+}
--- a/app/assets/stylesheets/puppetclasses.scss
+++ b/app/assets/stylesheets/puppetclasses.scss
@@ -41,3 +41,11 @@
 .selected_puppetclass.unavailable a {
   color: $color-pf-black-400;
 }
+
+#inherited_ids {
+  padding-left: 0;
+}
+
+#inherited-classes .panel-title {
+  font-size: 1.6rem;
+}
--- a/app/models/concerns/host_common.rb
+++ b/app/models/concerns/host_common.rb
@@ -62,6 +62,14 @@ module HostCommon
     end
   end
 
+  def parent_name
+    if is_a?(Host::Base) && hostgroup
+      hostgroup.name
+    elsif is_a?(Hostgroup) && parent
+      parent.name
+    end
+  end
+
   # Returns a url pointing to boot file
   def url_for_boot(file)
     "#{os.medium_uri(self)}/#{os.url_for_boot(file)}"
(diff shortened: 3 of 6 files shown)


orrabin:
Maybe the Included classes title should be bigger? because when all the classes are inherited it looks like this:
image

Do we still need the tooltip if the classes are now grouped under the title that explains where they came from?

bastilian:
@orrabin I've used a heading smaller than "Included Classes" on first instinct, since they are included, but inherited. Using a the same font size does make sense as well. In both cases though I think it would be nice to have a message that are no classes are added yet that are included, as it is when no environment is selected.

The tooltip makes sense, in my opinion, though I can't speak from experience (yet), when the list of modules grows and the heading is out of sight.

Rohoover:
Would it be worth making the Inherited Classes an accordion? Then also add field level help (http://www.patternfly.org/pattern-library/forms-and-controls/field-level-help/#_) to accommodate your tooltip? 

Also - I'm trying to get feedback now whether that blue icon is correct, or if we should use pficon_info.

bastilian:
@Rohoover We can try using an accordion for this, then it would make sense to do the include Classes as well, which is expanded by default. Would the field help then be on the title or with each puppet class?

Rohoover:
@bastilian If the tooltip content is the same for all puppet classes then I would put it on the title.

bastilian:
@Rohoover I will remove the tooltips. On the title it would be redundant, as it would say the same. The separation by the title should be enough, or?

Rohoover:
@bastilian I think the title should be enough

bastilian:
@Rohoover Did you mean the accordion to be used like this for the inherited classes:

screen shot 2017-01-09 at 09 25 32

Rohoover:
@bastilian 

No, just with the chevron/carrot.

dLobatog:
@bastilian Setting to WoC to change the accordion by the chevron/carrot + fixing the broken test

dLobatog:
@Rohoover Shouldn't it be h3 like the other headers as @orrabin suggested, even with the chevron? It looks a bit odd otherwise IMO

screenshot from 2017-01-11 13-00-18

Rohoover:
@dLobatog 
Yes, it looks like it should be the same as "Included Classes" and "Available Classes"

bastilian:
@Rohoover but with the chevron/carrot to expand and collapse?

Rohoover:
@bastilian 
Yes, unless you aren't concerned with how long the list would be.

bastilian:
@Rohoover I think @dLobatog can make better judgement of that.

In either case putting it into an accordion would be useful, then the question is only to collapse it or not by default.

Rohoover:
@bastilian 
I would collapse by default.

bastilian:
@Rohoover The inherited classes are now in an accordion, collapsed by default and styled to fit in better:
screen shot 2017-01-11 at 17 40 17


Even thought the title styles are now aligned it might still be a good idea to have a "placeholder" for included classes:

screen shot 2017-01-11 at 17 40 32

With classes it looks good:

screen shot 2017-01-11 at 17 40 55

The styles to remove borders and backgrounds can be applied to any accordion via the `.paneless` class and are separate from the title specific styling.

Rohoover:
@bastilian 

bitmoji

dLobatog:
@bastilian Please check out integration tests failures - http://ci.theforeman.org/job/test_develop_pr_core/11740/ - should be ready to merge afterwards. Thanks!

shlomizadok:
@Rohoover where do you get those cool images from?

Rohoover:
@shlomizadok 

It's a bitmoji, and I have the chrome plugin that adds a button to tool bars so I can insert them.
screen shot 2017-01-12 at 9 33 46 am

dLobatog:
Thanks @bastilian , @orrabin & @Rohoover !

PR #4154, opened 2017-01-04
Fixes #17910 - Quote strings for JQuery href attribute selector

bastilian:
When the attribute starts with an hash (#) the query can not be
parsed by JQquery, quoting them prevents that from happening.


tbrisker:
Thanks @bastilian !

shlomizadok:
@bastilian++

PR #4156, opened 2017-01-04
Fixes #17904 - Show tooltips for long proxy log messages

dLobatog:
To reproduce, get a proxy with long messages in the logs, and try to do this:

One can click on the "ERROR" part of the table but it's not obvious, the ellipsis fields are supposed to show a tooltip with the full content.

### DIFF
--- a/app/assets/javascripts/proxy_status/logs.js
+++ b/app/assets/javascripts/proxy_status/logs.js
@@ -52,6 +52,8 @@ function activateLogsDataTable() {
     if (link.data('message')) modal.find('#modal-bt-message').text(link.data('message'));
     if (link.data('backtrace')) modal.find('#modal-bt-backtrace').text(link.data('backtrace'));
   })
+  // Activate tooltips for fields with ellipsis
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/proxy_status/logs.js
```diff
@@ -52,6 +52,8 @@ function activateLogsDataTable() {
     if (link.data('message')) modal.find('#modal-bt-message').text(link.data('message'));
     if (link.data('backtrace')) modal.find('#modal-bt-backtrace').text(link.data('backtrace'));
   })
+  // Activate tooltips for fields with ellipsis
+  tfm.tools.activateTooltips();
```
tbrisker: you can limit this to just the table by passing its id as a parameter, i.e. `tfm.tools.activateTooltips('#table-proxy-status-logs')`

dLobatog:
@tbrisker Updated to pass the ID to activateTooltips

lzap:
Note you can still click on the message to see the detail pane, click on ERROR link to open it up.

PR #4157, opened 2017-01-05
Fixes #17923 - Add user default taxonomy by name

shlomizadok:


ares:
@shlomizadok please note that tests are failing

shlomizadok:
@ares - fixed

app/models/taxonomies/location.rb
```diff
@@ -9,7 +9,7 @@ class Location < Taxonomy
   before_destroy EnsureNotUsedBy.new(:hosts)
 
   has_many :location_parameters, :class_name => 'LocationParameter', :foreign_key => :reference_id, :dependent => :destroy, :inverse_of => :location
-  has_many :default_users,       :class_name => 'User',              :foreign_key => :default_location_id
+  has_many :default_users,       :class_name => 'User',              :foreign_key => :default_location_id, :inverse_of => :location
```
shlomizadok: Should be `default_location`, will change when tests are finished. Please don't merge yet.

tstrachota:
For the sake of consistency, shouldn't we fix the issue in hammer? I'm afraid this would be the only api endpoint where we would use `*_name` as an input parameter. We use only id's everywhere else and hammer/other api consumer is supposed to deal with translating names to ids itself.

shlomizadok:
It's a good point @tstrachota. I'd argue that other clients / scripts may also benefit from this change.
But if you / others feel it should be done on the hammer side - let me know.

ares:
I'd prefer hammer fox first for the consistency. I'm not against also improving API and allow searching by name but that should probably be added to all resources at once.

tstrachota:
I agree. There's certainly benefit in using names as identifiers when they're globally unique. But let's make it consistent first and if we decide to extend the api we should do it for all resources.

PR #4158, opened 2017-01-06
fixes #17951: Make email domain validation less strict and add test cases

vStone:


tbrisker:
ok to test

test/unit/validators/email_validator_test.rb
```diff
@@ -30,4 +20,27 @@ def setup
   test 'should allow blank' do
     assert @validatable.valid?
   end
+
+  [
```
shlomizadok: maybe use `%w(admin@example.com admin@example--domain.com)` ?

test/unit/validators/email_validator_test.rb
```diff
@@ -30,4 +20,27 @@ def setup
...
+    'admin@example.com',
+    'admin@example--domain.com',
+  ].each do |mail|
+    test "should pass email '#{mail} as valid" do
+      @validatable.mail = mail
+      assert @validatable.valid?
+    end
+  end
+
+  [
```
shlomizadok: Ditto

sean797:
@vStone Looks like tests are failing, can you take a look?

vStone:
Looks like i have an issue with the @localhost email address the tests are using. Let me figure out how to fix that. Probably  make the regex more relaxed :/

ohadlevy:
@vStone what is the status of the PR? thanks

ares:
I suppose the label here was added as 1.13 and later someone reused it by just changing its name. I don't think this is 1.15.1 blocker PR so removing it.

vStone:
I think this is no longer relevant.

PR #4159, opened 2017-01-08
Fixes #17844 - unify test mail js code

tbrisker:


app/controllers/users_controller.rb
```diff
@@ -125,7 +125,7 @@ def extlogout
   def test_mail
     begin
       user = find_resource(:edit_users)
-      if params[:user_email].blank? && user.mail.blank?
+      if (params.has_key?(:user_email) && params[:user_email].blank?) || user.mail.blank?
```
tbrisker, line 128: This check was actually incorrect before, since if the mail was changed to empty for a user that had a mail before this would return a success but try to send a mail with no recipient address.

app/views/settings/category/_email.html.erb
```diff
@@ -1,3 +1,3 @@
-<%= link_to_function(_("Test email"), "test_mail(this, #{User.current.id}, '#{test_mail_user_url(User.current)}')",
+<%= link_to_function(_("Test email"), "tfm.users.testMail(this, '#{test_mail_user_url(User.current)}')",
```
tbrisker, line 1: We don't actually need to send an ID in this case, the ID is already being sent from the url path.

webpack/assets/javascripts/foreman_users.js
```diff
@@ -9,3 +10,24 @@ export function initInheritedRoles() {
              .html(`${_.escape(target.text)} <span class="caret"></span>`);
   }).first().click();
 }
+
+/* eslint-disable no-undef */
```
tbrisker, line 28: need to disable no-undef here because `notify` has not been migrated to the new code yet.

tbrisker:
Rebased

dLobatog:
Seems to work on the settings page and the users page - merging. Thanks @tbrisker

PR #4162, opened 2017-01-08
Fixes #17972 - Simplify new trend form js and helpers

tbrisker:
Migrated the js to webpack as well.


dLobatog:
Thanks @tbrisker !

PR #4163, opened 2017-01-08
Fixes #17939 - Refactored classification to reduce complexity

ShimShtein:
1. Created a query object class to encapsulate the complex query for finding values that are relevant for a key
2. Made this query accessible via scope on `LookupKey` class
3. Encapsulated results rendering into a results class that behaves a bit like hash
4. Encapsulated matches generation logic into `MatchesGenerator` class.

For transition purposes, I have left the old classification classes but stripped down the logic and made them only adapters to the new system.

### DIFF
--- a/app/services/classification/base.rb
+++ b/app/services/classification/base.rb
@@ -14,7 +14,7 @@ module Classification
     end
 
     def inherited_values
-      values_hash :skip_fqdn => true
+      keys.inherited_values(@host).raw
     end
 
     protected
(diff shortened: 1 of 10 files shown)


app/services/classification/base.rb
```diff
@@ -23,212 +23,15 @@ def inherited_values
...
-          values |= lookup_value.value
-        else
-          values += lookup_value.value
-        end
-      end
-
-      return nil unless values.present?
-      {:value => values, :element => elements, :element_name => element_names}
+    def keys
+      raise NotImplementedError
```
orrabin, line 30: why return keys in `class_parameters` and have this and not just implement `class_parameters`?

app/services/classification/matches_generator.rb
```diff
@@ -0,0 +1,94 @@
...
+      return host.send(element) if is_method?(element)
+      # host parameter
+      return host.host_params[element] if is_parameter?(element)
+      # fact attribute
+      if (fn = fact_name(element))
+        return fact_value(fn)
+      end
+    end
+
+    def is_method?(element)
```
orrabin, line 58: Why extract this? it was easier to understand when it was part of the line. It's also not a method, it's associations to the host.

app/services/classification/class_param.rb
```diff
@@ -1,36 +1,40 @@
 module Classification
   class ClassParam < Base
     def enc
-      key_hash = hashed_class_parameters
-      values   = values_hash
+      key_hash = hashed_class_keys(keys)
```
orrabin, line 4: Why send the keys when you can call it from `hashed_class_keys`?

app/services/classification/matches_generator.rb
```diff
@@ -0,0 +1,94 @@
...
+      @keys = keys
+    end
+
+    def matches
+      matches = []
+      possible_value_orders.each do |rule|
+        match = generate_match(rule)
+        matches << match.join(LookupKey::KEY_DELM)
+
+        hostgroup_matches.each do |hostgroup_match|
```
orrabin, line 20: This whole part can move into `add_hostgroup_matches?` if you change the name to `add_hostgroup_matches`.

app/services/classification/matches_generator.rb
```diff
@@ -0,0 +1,94 @@
...
+      if (fn = fact_name(element))
+        return fact_value(fn)
+      end
+    end
+
+    def is_method?(element)
+      host.respond_to?(element)
+    end
+
+    def is_parameter?(element)
```
orrabin, line 62: I don't think it's worth extracting.

app/services/classification/matches_generator.rb
```diff
@@ -0,0 +1,94 @@
...
+    attr_reader :host, :keys
+    def initialize(host, keys)
+      @host = host
+      @keys = keys
+    end
+
+    def matches
+      matches = []
+      possible_value_orders.each do |rule|
+        match = generate_match(rule)
```
orrabin, line 17: :+1: for breaking this up

app/services/classification/matches_generator.rb
```diff
@@ -0,0 +1,94 @@
...
+
+    # translates an element such as domain to its real value per host
+    # tries to find the host attribute first, parameters and then fallback to a puppet fact.
+    def attr_to_value(element)
+      # direct host attribute
+      return host.send(element) if is_method?(element)
+      # host parameter
+      return host.host_params[element] if is_parameter?(element)
+      # fact attribute
+      if (fn = fact_name(element))
```
orrabin, line 53: Instead of extracting one liners into 2 methods why not extract this if into one method?

app/services/classification/class_param.rb
```diff
@@ -1,36 +1,40 @@
...
-        if key_hash[klass.id]
-          key_hash[klass.id].each do |key|
-            key_value = value_of_key(key, values)
-            klasses[klass.name][key.to_s] = key_value unless key_value.nil?
-          end
-          klasses[klass.name] = nil if klasses[klass.name] == {}
-        else
-          klasses[klass.name] = nil
-        end
+        klasses[klass.name] = smart_class_params_for(klass, key_hash, values)
```
orrabin, line 9: The name is confusing, this is the way the hash will look and not smart class parameters.

app/services/classification/classification_result.rb
```diff
@@ -0,0 +1,58 @@
...
+      if key_hash
+        {:value => key_hash[:value], :managed => key_hash[:managed] }
+      else
+        default_value_method = %w(yaml json).include?(key.key_type) ? :default_value_before_type_cast : :default_value
+        {:value => key.send(default_value_method), :managed => key.omit}
+      end
+    end
+
+    def key_hash(key)
+      raw[key.id][key.to_s] if raw[key.id]
```
orrabin, line 43: Why extract this?

app/services/classification/values_hash_query.rb
```diff
@@ -0,0 +1,167 @@
...
+        values_hash host, keys, :skip_fqdn => true
+      end
+
+      # Calculates values for a given set of keys and a host supplied in the constructor.
+      # Inputs:
+      # keys: an ActiveRecord::Scope with a set of LookupKey instances
+      # options:
+      #   :skip_fqdn => true, if there is no need to calculate values specified directly for the host.
+      # returns: ClassificationResult instance.
+      def values_hash(host, keys, options = {})
```
orrabin, line 17: :+1: for breaking this up, it was huge and unreadable.

app/services/classification/values_hash_query.rb
```diff
@@ -0,0 +1,167 @@
...
+      private
+
+      def lookup_values_cache(host, keys)
+        LookupValue.where(:match => Classification::MatchesGenerator.matches(host, keys)).where(:lookup_key_id => keys).includes(:lookup_key).to_a
+      end
+
+      def calculate_value(key, lookup_values_cache, options)
+        lookup_values_for_key = lookup_values_cache.select{|i| i.lookup_key_id == key.id}
+        sorted_lookup_values = sort_lookup_values(key, lookup_values_for_key)
+        value = nil
```
orrabin, line 35: Since this is now a method why not return instead of keeping the value and returning it at the end?

test/unit/classification_test.rb
```diff
@@ -83,18 +83,21 @@ def setup
     pc = FactoryGirl.create(:puppetclass, :environments => [env])
     lkey = FactoryGirl.create(:puppetclass_lookup_key, :as_smart_class_param, :with_override, :puppetclass => pc)
     classparam = get_classparam(env, pc)
-    classparam.expects(:attr_to_value).with('comment').returns('override')
+
+    host = classparam.send(:host)
+    host.expects(:comment).returns('override')
```
orrabin, line 88: Why is this on the host?

test/unit/classification_test.rb
```diff
@@ -83,18 +83,21 @@ def setup
...
     assert_equal({pc.name => {lkey.key => 'overridden value'}}, classparam.enc)
   end
 
   test "#values_hash should contain element's name" do
     env = FactoryGirl.create(:environment)
     pc = FactoryGirl.create(:puppetclass, :environments => [env])
     lkey = FactoryGirl.create(:puppetclass_lookup_key, :as_smart_class_param, :with_override, :puppetclass => pc)
-    classparam = Classification::ClassParam.new
+    host = mock('host')
+    classparam = Classification::ClassParam.new(:host => host)
```
orrabin, line 97: Isn't this what `get_classparam` does? (with the env and puppetclass)

test/unit/classification_test.rb
```diff
@@ -126,10 +130,11 @@ def setup
                                    :puppetclass => pc, :key_type => 'yaml', :default_value => 'a: b')
     json_lkey = FactoryGirl.create(:puppetclass_lookup_key, :as_smart_class_param, :override => true,
                                    :puppetclass => pc, :key_type => 'json', :default_value => '{"a": "b"}')
-    classparam = Classification::ClassParam.new
+    host = FactoryGirl.build(:host, :environment => env, :puppetclasses => [pc])
+    classparam = Classification::ClassificationResult.new(host, {})
```
orrabin, line 134: looks like the host should be the second parameter

orrabin (changes requested):

mmoll:
what's the status here? :)

ShimShtein:
It's good that this commit is closed. It's obsolete, everything from it is already in core's code base!

PR #4165, opened 2017-01-09
Fixes #17918 - Smart class parameters should appear in audits

orrabin:


db/migrate/20170109115157_fix_lookup_key_auditable_type.rb
```diff
@@ -0,0 +1,12 @@
+class FixLookupKeyAuditableType < ActiveRecord::Migration
+  def up
+    Audit.where(:auditable_type => 'LookupKey').each do |audit|
```
tbrisker: This would potentially load a very large amount of audits from the DB to check their types. would it be possible to perform this in the database without loading all of the audits and going over them one by one? (maybe change all the puppetLKs in one query and all the variableLKs in a second?) If it isn't possible, please use `find_each` instead of `each` so at least this is done in batches.

app/models/concerns/audit_extensions.rb
```diff
@@ -67,6 +67,8 @@ def fix_auditable_type
     self.associated_type = "Host"         if associated_type =~ /Host::/
     self.auditable_type = auditable.type  if auditable_type == "Taxonomy" && auditable
     self.associated_type = auditable.type if auditable_type == "Taxonomy" && auditable
+    self.auditable_type = auditable.type  if auditable_type == "LookupKey" && auditable
+    self.associated_type = auditable.type  if auditable_type == "LookupKey" && auditable
```
tbrisker: Perhaps refactor these 4 lines to something like this?
```ruby
self.auditable_type = self.associated_type = auditable.type if auditable && ['LookupKey', 'Taxonomy'].includes? auditable_type
```

tbrisker (changes requested):
Looks good in general, just a couple of comments inline

orrabin:
@tbrisker thanks, I fixed you comments.

db/migrate/20170109115157_fix_lookup_key_auditable_type.rb
```diff
@@ -0,0 +1,15 @@
+class FixLookupKeyAuditableType < ActiveRecord::Migration
+  def up
+    lookup_key_ids = Audit.where(:auditable_type => 'LookupKey').pluck(:auditable_id).uniq
+
+    puppetclass_lookup_key_ids =  PuppetclassLookupKey.where(:id => lookup_key_ids).pluck(:id)
+    Audit.where(:auditable_type => 'LookupKey', :auditable_id => puppetclass_lookup_key_ids).update_all(:auditable_type => 'PuppetclassLookupKey')
```
tbrisker: How about something like:
```ruby
Audit.joins('JOIN lookup_keys ON lookup_keys.id = audits.auditable_id').where(:auditable_type => 'LookupKey', :lookup_keys => {:type => 'VariableLookupKey'}).reorder(nil)
```
Also, make sure to `reorder(nil)` on all the queries to remove the default order scope.

tbrisker (changes requested):
One more sql optimization :)

db/migrate/20170109115157_fix_lookup_key_auditable_type.rb
```diff
@@ -0,0 +1,15 @@
+class FixLookupKeyAuditableType < ActiveRecord::Migration
+  def up
+    Audit.reorder(nil).joins('JOIN lookup_keys ON lookup_keys.id = audits.auditable_id')
+         .where(:auditable_type => 'LookupKey', :lookup_keys => {:type => 'VariableLookupKey'})
+         .update_all(:auditable_type => 'VariableLookupKey')
+
+    Audit.reorder(nil).joins('JOIN lookup_keys ON lookup_keys.id = audits.auditable_id')
+         .where(:auditable_type => 'LookupKey', :lookup_keys => {:type => 'PuppetclassLookupKey'})
+         .update_all(:auditable_type => 'PuppetclassLookupKey')
```
orrabin: tried to get the two queries into one with `update_all('auditable_type = lookup_keys.type')` but: https://github.com/rails/rails/issues/13496

app/helpers/audits_helper.rb
```diff
@@ -136,6 +136,10 @@ def audited_type(audit)
                     "#{audit.associated_type || 'Global'}-#{type_name}"
                   when 'LookupKey'
                     'Smart Variable'
```
orrabin: Keeping this because of old audits that don't have a connected lookup key and therefore I can't determine their type.

db/migrate/20170109115157_fix_lookup_key_auditable_type.rb
```diff
@@ -0,0 +1,15 @@
+class FixLookupKeyAuditableType < ActiveRecord::Migration
+  def up
+    Audit.reorder(nil).joins('JOIN lookup_keys ON lookup_keys.id = audits.auditable_id').
+         where(:auditable_type => 'LookupKey', :lookup_keys => {:type => 'VariableLookupKey'}).
+         update_all(:auditable_type => 'VariableLookupKey')
+
+    Audit.reorder(nil).joins('JOIN lookup_keys ON lookup_keys.id = audits.auditable_id').
+         where(:auditable_type => 'LookupKey', :lookup_keys => {:type => 'PuppetclassLookupKey'}).
+         update_all(:auditable_type => 'PuppetclassLookupKey')
```
orrabin: These can't be combined into one using `update_all('auditable_type = lookup_keys.type')` because of a bug in rails.

app/models/concerns/audit_extensions.rb
```diff
@@ -65,8 +65,7 @@ def fix_auditable_type
     # STI Host class should use the stub module instead of Host::Base
     self.auditable_type = "Host"          if auditable_type =~  /Host::/
     self.associated_type = "Host"         if associated_type =~ /Host::/
-    self.auditable_type = auditable.type  if auditable_type == "Taxonomy" && auditable
-    self.associated_type = auditable.type if auditable_type == "Taxonomy" && auditable
```
orrabin: `auditable` seems to be nil at this point

app/models/concerns/audit_extensions.rb
```diff
@@ -65,8 +65,8 @@ def fix_auditable_type
...
+    self.auditable_type = auditable.type if auditable && ["Taxonomy", "LookupKey"].include?(auditable_type)
+    self.associated_type = auditable.type if auditable && ["Taxonomy", "LookupKey"].include?(auditable_type)
```
tbrisker: after some debugging, looks like while this condition is always false, actually a side effect of the `auditable` call is what caused this to work correctly, possibly because it led to a reload of the auditable object. Otherwise the save fails. Some preliminary testing indicated that moving this callback from `after_validation` to `before_save` allows removing the second `auditable` call and saving the object correctly, but this needs to be checked to make sure this doesn't cause other things to break.

app/models/concerns/audit_extensions.rb
```diff
@@ -95,8 +95,8 @@ def fix_auditable_type
     # STI Host class should use the stub module instead of Host::Base
     self.auditable_type = "Host"          if auditable_type =~  /Host::/
     self.associated_type = "Host"         if associated_type =~ /Host::/
-    self.auditable_type = auditable.type  if auditable_type == "Taxonomy" && auditable
-    self.associated_type = auditable.type if auditable_type == "Taxonomy" && auditable
+    self.auditable_type = auditable.type if auditable && ["Taxonomy", "LookupKey"].include?(auditable_type)
```
ShimShtein: You can drop `auditable &&`, if you fix two issues:
1. Remove duplicate deletion from [puppetclass_importer](https://github.com/theforeman/foreman/blob/develop/app/services/puppet_class_importer.rb#L301), as you have already fixed this in https://github.com/theforeman/foreman/commit/b5f207decca34bf1ef0f3b97b1d793b1613be690
2. Add default value to `:type` column in `:lookup_key` [factory](https://github.com/theforeman/foreman/blob/develop/test/factories/puppet_related.rb#L11)

app/models/concerns/audit_extensions.rb
```diff
@@ -95,8 +95,8 @@ def fix_auditable_type
...
+    self.associated_type = auditable.type if auditable && ["Taxonomy", "LookupKey"].include?(auditable_type)
```
ShimShtein: This line looks a bit odd - you are setting `:associated` field to the type of `:audited` object.
It's unrelated to this PR, let's track it as a separate bug [here](http://projects.theforeman.org/issues/19354).

ShimShtein (commented):
Looks good, a few suggestions though.

app/models/concerns/audit_extensions.rb
orrabin: `&& auditable` was already in the code, it's a good suggestion to refactor it but could lead to a lot of changes in unrelated tests and might affect taxonomy as well so I'd rather have it in a separate pr.

I'll move it to the end to avoid calling auditable again for anything but taxonomy and lookup keys.

app/models/concerns/audit_extensions.rb
orrabin: This line should be changed anyway for: http://projects.theforeman.org/issues/19354

One thing I could think of that could break is the callbacks we have today on `before_save` that might have counted on `auditable` being accessed in this callback.
I moved the callback and did a few create/update/delete that showed up correctly in the audits and didn't see anything break right away.

ShimShtein:
:+1: ACK. Works as expected, no more comments.

orrabin:
@tbrisker all your comments are now addressed.

tbrisker (approved):

tbrisker:
Thanks @orrabin !

PR #4166, opened 2017-01-09
Fixes #17979 - Pin nokogiri to last version supporting ruby 2.0

tbrisker:


Gemfile
```diff
@@ -37,6 +37,7 @@ gem 'x-editable-rails', '~> 1.5.5'
 gem 'deacon', '~> 1.0'
 gem 'webpack-rails', '~> 0.9.8'
 gem 'mail', '~> 2.6'
+gem 'nokogiri', '< 1.7'
```
domcleal: This should be pinned conditionally, only on Ruby 2.0.

domcleal (changes requested):

Gemfile
tbrisker: Fixed, thanks.

domcleal (approved):

domcleal:
Merged, thanks @tbrisker.

PR #4167, opened 2017-01-10
Fixes #17998 - Only print LDAP bind status if available

dLobatog:
The ldap logger tries to call payload[:bind].status but it might not be
defined if the bind goes wrong in some LDAP sources.

This causes a 500 instead of logging the status of the bind. We should
only display payload[:bind].status if it's available.


dLobatog:
Rebased to get the Nokogiri pinning

tbrisker:
Thanks @dLobatog !

PR #3600, opened 2016-06-19
Fixes #15405 - Moved puppet UI to helpers

ShimShtein:
Created puppet-specific helpers as part of pulling puppet outside of core.

PR #2572, opened 2015-07-24
Fixes #11206 - plugin API now expose default role permissions

lzap:
If we were able to list default roles and it's permissions we can use this in
unit/functional tests to verify permissions and filters. More than that, this
gives us possibility to create a mechanism for comparing expected permissions
with existing ones. This might be useful after upgrades when users want to
check if some important permissions were not added in the new version.

This patch that adds this, it's simple. It also adds a simple rake task that
does the check. I would like to have this for 1.9 because we would like to
merge a big permission review for Discovery 4.0 together with bunch of new
tests covering this area which greatly improves the quality.

PR #4168, opened 2017-01-10
Fixes #18001 - Add plugin permissions to core's roles

xprazak2:


app/services/foreman/plugin.rb
```diff
@@ -396,5 +405,30 @@ def extend_template_helpers_by_module(mod)
       end
       allowed_template_helpers(*(mod.public_instance_methods - Module.public_instance_methods))
     end
+
+    def add_manager_roles(resources, opts)
+      manager = Role.find_by :name => 'Manager'
```
ares: what happens if user deleted the role? should we recreate it? also I need to test it but I think this can happen when you install Foreman, DB was not seeded yet but the rails env loads which means it loads all plugin initializers.

app/services/foreman/plugin.rb
```diff
@@ -396,5 +405,30 @@ def extend_template_helpers_by_module(mod)
...
+      add_permissions(manager, resources, opts) do |resource|
+        Permission.where(:resource_type => resource)
+      end
+    end
+
+    def add_viewer_roles(resources, opts)
+      viewer = Role.find_by :name => 'Viewer'
+      add_permissions(viewer, resources, opts) do |resource|
+        Permission.where(:resource_type => resource).where(Permission.arel_table[:name].matches("view_%"))
```
ares: just out of curiosity, why using `arel_table` here? do you prefer this over the string in condition?

app/services/foreman/plugin.rb
```diff
@@ -396,5 +405,30 @@ def extend_template_helpers_by_module(mod)
...
+      end
+    end
+
+    def add_permissions(role, resources, opts, &block)
+      to_add = []
+      resources.each do |resource|
+        all_permissions = yield resource
+        to_add += all_permissions.reject do |permission|
+          role.permissions.where(:name => permission.name).any? || opts[:except].include?(permission.name.to_sym)
```
ares: if you change the order, the SQL query won't be run if the permission is on blacklist

app/services/foreman/plugin.rb
```diff
@@ -396,5 +405,30 @@ def extend_template_helpers_by_module(mod)
...
+    def add_viewer_roles(resources, opts)
+      viewer = Role.find_by :name => 'Viewer'
+      add_permissions(viewer, resources, opts) do |resource|
+        Permission.where(:resource_type => resource).where(Permission.arel_table[:name].matches("view_%"))
+      end
+    end
+
+    def add_permissions(role, resources, opts, &block)
+      to_add = []
+      resources.each do |resource|
```
ares: would you mind splitting this method a bit, you have block in block and you call yield in one of them, I'd suggest using 

```ruby
def add_permissions(...)
   to_add = resources.map { |r| new_permissions_for(r) }.inject(&:+)
   role.add_permissions! to_add.map(&:name) if to_add.any?
end

def new_permissions_for(r)
  ...
end
```

I don't think the block is worth it, I'd probably prefer only to pass the condition which can be optional. That would also DRY up `Permission.where(:resource_type => resource)` that you have in add_manager_roles and add_viewer_roles

ares (changes requested):

ares:
In general I agree such plugin DSL would be very beneficial.

app/services/foreman/plugin.rb
```diff
@@ -243,6 +243,15 @@ def role(name, permissions)
       end
     end
 
+    # Add plugin permissions to core's Manager and Viewer roles
+    # Usage: add_permissions_to_core_roles ['MyPlugin::FirstResource', 'MyPlugin::SecondResource'], :except => [:skip_this_permission]
```
ares: I guess some permissions might not be tight to any specific resource

app/services/foreman/plugin.rb
xprazak2: Right. Skipping when role is not found.

xprazak2:
I refactored a bit to remove block and I use string condition in `.where`. I also added a method that can handle permissions not tied to any resource

app/services/foreman/plugin.rb
```diff
@@ -243,6 +243,27 @@ def role(name, permissions)
       end
     end
 
+    # Add plugin permissions to core's Manager and Viewer roles
+    # Usage:
+    # add_permissions_to_core_roles ['MyPlugin::FirstResource', 'MyPlugin::SecondResource'], :except => [:skip_this_permission]
+    def add_resource_permissions_to_core(resources, opts = {})
```
ehelms: If the roles are referred to as the "default roles" everywhere else, can we name this method similarly? e.g. add_resource_permissions_to_default_roles 

I think that would make things more readable/obvious.

app/services/foreman/plugin.rb
```diff
@@ -243,6 +243,27 @@ def role(name, permissions)
...
+      Role.transaction do
+        add_viewer_roles(resources, opts.dup)
+        add_manager_roles(resources, opts.dup)
```
ehelms: These method names are a little misleading as we aren't adding manager roles, but rather adding permissions to the single manager or viewer role.

ehelms:
Something to consider (since this is adding more functionality) is taking all of the permission/role related functionality and re-factoring it into a class similar to how the plugin logging functionality was done (https://github.com/theforeman/foreman/blob/develop/app/services/foreman/plugin/logging.rb).

app/services/foreman/plugin.rb
```diff
@@ -396,5 +417,47 @@ def extend_template_helpers_by_module(mod)
       end
       allowed_template_helpers(*(mod.public_instance_methods - Module.public_instance_methods))
     end
+
+    def add_manager_roles(resources, opts)
+      manager = Role.find_by :name => "Manager"
+      return if !manager || resources.empty?
+      add_permissions manager, resources, opts
+    end
```
ehelms: Is the idea that for this role, all "assumed" permissions are added for a given resource? Those being:

 * view_
 * edit_
 * create_
* delete_

test/unit/plugin_test.rb
```diff
@@ -424,6 +424,25 @@ def test_hosts_controller_action_scope_added_to_local
     assert_equal 2, scopes.count
   end
 
+  def test_add_resource_permissions_to_core
+    Foreman::Plugin.register :test_plugin do
+      add_resource_permissions_to_core ["Test::Resource"], :except => [:create_test]
+    end
+    manager = Role.find_by :name => "Manager"
+    viewer = Role.find_by :name => "Viewer"
+    assert_equal 2, manager.permissions.where(:resource_type => "Test::Resource").count
```
ehelms: Why does this count go up by 2 only? I would have assumed 3 would have been added (view, edit and delete).

app/services/foreman/plugin.rb
xprazak2: The idea is that all permissions for a given resource are added to 'Manager', regardless of how the permission is named, unless it is explicitly stated they should be skipped by using `:except => ['some_permission_name']` in opts

test/unit/plugin_test.rb
xprazak2: Perhaps related to comment above, 3 out of 4 fixtures I added are associated with the resource (view_test, create_test, edit_test). `:except => [:create_test]` says I do not want to add create_test permission.

xprazak2:
Moving this to a separate class is a good idea, thanks for suggestion @ehelms

lzap:
What is wrong with defining "Plugin Reader" and "Plugin Manager" in the fisrt place? I kinda miss the problem here, if each plugin provides Viewer/Reader and Manager, then users can very easily create view-only users or groups. This is a bit artificial issue trying to solve situation when user is created just with "Viewer" role and then something does not work.

Discovery is one of the plugins that is using Roles for a long time and we've been pretty successful with what we have today. We simply define two roles and add all required core permissions into it, so it works "out-of-box". Also the plugin API was extended for core and other plugins to get list of default roles. This allows you to create viewer that is only dedicated to discovery and not giving access to any other parts of the system. This cannot be achieved with what you propose.

https://github.com/theforeman/foreman_discovery/blob/develop/lib/foreman_discovery/engine.rb#L102-L144

I think the better approach than explicitly "injecting" roles into core is to provide default sets of permissions for both roles. And then pulling this information when necessary, but I still do not think it's good idea for plugins to change core roles.

We already have a plugin API that allows publishing default roles *today*. I believe only Discovery plugin takes advantage of this:

    Foreman::Plugin.find('foreman_discovery').default_roles['Discovery Reader']

This was implemented in http://projects.theforeman.org/issues/11206 (https://github.com/theforeman/foreman/pull/2572/files) and it can be used in two places. First, there is this `plugin:validate_roles` rake task that is able to compare expected permissions and current state. Users often delete permissions when trying to tighten security and when something does not work, it's hard to find the missing roles.

Second, this can be used in unit/functional testing where you can easily setup Manager and Reader roles for tests. We tend to test with admin accounts a lot which prevents from hitting permission issues at the test stage.

To make this short, I **do not agree** that plugins should be modifying core roles and I think this is **wrong approach**. This tightens our arms in making use of roles seed script (03-roles.rb) enabling users to reset permissions back to installed state (we don't have this feature yet but I think this would be very useful) and creates confusion and more complex testing scenarios (plugin is missing, permission is missing).

In the worst case, I vote against pushing this information and reusing what we already have in the API (default_roles hash) allowing the core to pull this information. This allows better integration with tests and rake task (possibly a WebUI feature) allowing comparison of existing roles with expected (default) state for both core and plugins. At least I'd build on top of that.

ehelms:
I think @lzap has a valid point and per plugin roles could provide better granularity. What I would then ask is can core make it easier for plugins to define their plugin roles through the plugin API (if for example, we say, all plugins should define a core set of roles "Viewer" and "Manager") to both help developers and ensure that users are presented with "default plugin roles" per plugin gauranteed.

xprazak2:
> What is wrong with defining "Plugin Reader" and "Plugin Manager" in the fisrt place?

@lzap, there is nothing wrong with it, what made you think I consider it wrong? I said on the mailing list plugins should do that anyway.

> I kinda miss the problem here, if each plugin provides Viewer/Reader and Manager, then users can very easily create view-only users or groups.

The initially reported problem is that 'Manager' role does not have permissions for things that come from for plugins. I believe users expect 'Manager' to be able to manage everything regardless of where the functionality comes from. They expect that when a new plugin is installed, users with 'Manager' role not need any changes. That is the main issue I was trying to address here and a reason why I initially chose the approach outlined in this PR. I admit, this solution is far from ideal.

> This is a bit artificial issue trying to solve situation when user is created just with "Viewer" role and then something does not work.

Not sure what you mean by that, could you elaborate?

ares:
As a user that installed foreman with some plugins (since they are enabled in the default scenario) I expect Manager role to contain all permissions including plugin ones. Otherwise it should be called "Core Manager" and again as a user, I don't have a clue what it means.

As @xprazak2 said, per plugin roles are also fine, user can create user group and include only what they need. But out-of-the box installation should provide Viewer and Manager roles that work for the whole application as their names suggest no matter what plugins I installed.

lzap:
> The initially reported problem is that 'Manager' role does not have permissions for things that come from for plugins. I believe users expect 'Manager' to be able to manage everything regardless of where the functionality comes from. They expect that when a new plugin is installed, users with 'Manager' role not need any changes.

I think we are at point we need to simply say no to whoever asked this. I suggest to rename "Manager" to something more explanatory like "Core Manager" and/or adding help text.

We can do some effort to make easier with implementing a way to automatically include all "* Manager" roles for a users via checkbox "[x] Always keep all *Manager roles" and the same for Reader/Viewer. This would be effective on every log in. This can be easily implemented if plugins exports their default manager and reader roles via API. But what I don't like is letting plugins to play around (modify) the core roles.

After reading Marek's comment, I see we share the opinion on renaming "Manager" role to "Core Manager", that's a good start.

lzap (dismissed):
I will continue to discuss the necessity of this change in the thread, but my overall concern is that the API only provides a way of adding permissions. How plugin authors will be removing or renaming them? We needed this several times already in the past with discovery (and did that via migrations).

ares:
I think the linked PR in foreman_templates can be understood as +1, if we don't provide a way, people will try to find workarounds.

lzap:
Let me elaborate my concern with the current proposal - our permissions API currently only provide ability to add permissions (`add_permissions`) and if we are building such an API for plugins where we want to be sure it's stable and won't change much, I wonder if we want to redesign this in a way plugins can fully manage permissions.

For this reason I think it could be better for plugins to provide list of "default" roles and permissions. These roles could be ideally read-only for users, so it's easier to work with them. From time to time we needed to remove a permission from a role, when role is user-managed it's a bit of issue because we break users (they no longer can do what they used to do).

I know what I am trying to discuss here is a change of permissions and roles, but I think it is worth it. Rather than building another (plugin) API on what we have I would like to improve core in this regard. I was doing fine in discovery managing my own Manager/Reader just fine without proposed plugin API, so for this reason I think we can rather live without this, or maybe take a chance and do it in a different way.

Is this blocking some other PRs that needs to go in? If so, I suggest to do step back, implement it without this proposed API and take our time to discuss this.

xprazak2:
> From time to time we needed to remove a permission from a role, when role is user-managed it's a bit of issue because we break users (they no longer can do what they used to do).

If we remove a permission from a role that is being actively used we break it for users and it does not matter if role is user-managed or our "default".

>  I was doing fine in discovery managing my own Manager/Reader just fine without proposed plugin API, so for this reason I think we can rather live without this, or maybe take a chance and do it in a different way.

I am certain discovery can live without it, but this is about Foreman as a whole. Each component can do well on its own, but the result can still be a broken app. I do not have to go far in history for an [example](https://github.com/theforeman/foreman/pull/3600#issuecomment-273588305). In this case, it is "only" annoying users. If it wasn't, they would not raise an issue.

> Is this blocking some other PRs that needs to go in?

This is not blocking anything AFAIK. @lzap, you raised an important questions that are related to plugins in general. If we found a reasonable solution how to remove plugin permissions/roles, this could take us one step closer to plugin uninstallation. I was thinking about the "reset to defaults" feature you mentioned above, maybe it could be a part of the solution.

dLobatog:
I'm confused about why the Puppet UI helpers PR was linked as an example of a similar problem, I don't think it was.  About the issue you're discussing here, I would 👎 renaming to "Core Manager" and "Core Viewer", those terms make no sense to users that are not devs too. 

Having this API does not take away from having plugin specific roles if you want to have them too.

Generally it's a 👍 from me, but I don't feel very strongly either way. Is there any user comment/feedback that triggered http://projects.theforeman.org/issues/17954 or http://projects.theforeman.org/issues/18001 ? To me it seems positive to provide a more coherent set of permissions & roles, but if people are OK with the current approach, I'd say YAGNI & carry on

xprazak2:
@dLobatog This was initially reported [downstream](https://bugzilla.redhat.com/show_bug.cgi?id=1304608)

> I'm confused about why the Puppet UI helpers PR was linked as an example of a similar problem

I never said it was a similar problem. I was referencing it to make my point that even if one component is fine, the application as a whole does not have to be - in response to a comment just [above](https://github.com/theforeman/foreman/pull/4168#issuecomment-274445308). In case of Puppet helpers, Foreman was fine, Foreman + Katello was not. In this case, Foreman + discovery is fine, and if Foreman + other plugins was fine as well, this issue would not have been raised.

ShimShtein:
Puppet UI PR is an excellent example of non-formal contracts that prevent us from making any change to the code (I have moved a call to helper to a different file, and it broke a plugin - not reasonable at all). Actually, @ares, you can use this example in the "Let's make our contracts with plugins more formal" discussion.

My two cents about the issue. Maybe we should look at the permissions in a more "faceted" way. I mean let's divide plugin models to two categories:
1. Models that extend core objects, like puppet/chef that add config management, content e.t.c. I call them facets. I am not sure that the user should be aware of different permissions for each specific facet of that object. For example, It's a bit too granular to say that a user has permission to edit some properties of the host, but he can't change properties that are related to content.
2. Completely new objects/resources introduced by the plugin. A good example for those would be puppet classes - It's a standalone resource that could be managed by itself - you can delete a class from the system, you can add a new one and so on. Here, I think, we as the developers could not say whether a user that has permissions to some objects should also have permissions to these new objects or not. I think in this case, we should leave the choice to the user, so he would be able to decide which roles will include these permissions and which won't.

Basically, I suggest to treat permissions for objects from the group#1 as suggested by @ares and @xprazak2. And resources from group#2 to treat as suggested by @lzap.
For me it makes a bit more sense.

ares:
@ShimShtein I mostly agree. One thing I disagree with is 2) for Manager and Viewer roles, these IMO clearly says what user expect from these roles. If plugin integrates with Foreman, it should integrate with its permission system which defines Manager role which contains all permissions available in the system. I agree with it for other core roles such as "View hosts", plugin should not probably extend these if it adds new type of host (discovery example) or at all if it's only similar (docker). TL;DR I think Viewer role should grant permission to view puppet classes if puppet was extracted to a plugin but installed.

ShimShtein:
@ares We do have an "Admin" checkbox that should override all permissions, don't we? This one should be respected by plugins of course. I am not sure about the "viewer" part though.

ares:
@ShimShtein it's not the same thing, e.g. Setting page can be change only by admin, not by user with Manager role. The purpose of Manager role is to contain all user permissions and to be a base that you can clone and assign to a Org/Loc to effectively create a "Manager of ACME"

ShimShtein:
@ares Hmmm... Another idea (not too good, so you can dismiss it). In that case maybe we should turn the `is_admin` to something with more states. Let's call it user "kind"? I mean there will be three kinds: Admin, Manager and Viewer. Every plugin will have to respect it just as they do with the admin right now.

ares:
@ShimShtein now we're getting pretty offtopic so moving to another media :-)

app/services/foreman/plugin.rb
```diff
@@ -396,5 +417,47 @@ def extend_template_helpers_by_module(mod)
...
+      opts[:except] ||= []
+      opts[:except].include?(permission.name.to_sym) || role.permissions.where(:name => permission.name).any?
+    end
+
+    def add_permissions_to_role(role, permission_names)
+      to_add = permission_names.reject do |name|
+        role.permissions.find_by :name => name
+      end
+      role.add_permissions! to_add if to_add.any?
+    end
```
lzap: I will simplify my concerns to just single item. Currently there is `default_roles` that provides a hash of roles defined by plugin and list of permissions. Can you deprecate this hash and incorporate it into the new API so every plugin exposes all its roles and permissions via some method? Then modify the `plugin:validate_roles` rake task, it is still useful when users play with roles and does not restart Rails too often. The default permissions exposed are also useful for easier permission testing, so this is useful API feature that I am missing here.

lzap (changes requested):
I left one comment - this is my only concern. This whole area needs improvements, there were some ideas in the thread and here, but this is really out of scope of this.

ares (approved):

xprazak2:
I moved this along a bit, still very much WIP. Changes:

* I extracted the roles we seed from seed files, moved them to services/foreman/constants
* We are able to detect if user changed permissions for 'Viewer' and 'Manager' and restore the original state in migration. If user modified these roles, they are renamed to 'Custom Viewer/Manager' and new roles are created with default set of permissions that we seed.
* Plugins can now provide information about roles and permissions they create. `default_roles` hash and rake task that compares default permissions for plugin roles works as before

I will start working on locking our seeded roles, which will prevent us from modifying roles that users have already customized.

xprazak2:
I continued where I [left off](https://github.com/theforeman/foreman/pull/4168#issuecomment-281677646). Seeded roles are now locked for user modifications, I added a couple of tests.

app/models/role.rb
```diff
@@ -125,17 +129,26 @@ def add_permissions(permissions, options = {})
...
+
+      filter ||= self.filters.build(:search => search)
       filter.role ||= self
 
       grouped_permissions.each do |permission|
         filtering = filter.filterings.build
         filtering.filter = filter
         filtering.permission = permission
       end
+      filter.save if filter.id
```
ares: Please use `filter.persisted?` so it's clear when and why we call save.

Also I wonder what happens if filter is invalid for whatever reason? I think it would fail without any notice. Wouldn't it be better to set [autosave](http://api.rubyonrails.org/v4.2.7//classes/ActiveRecord/AutosaveAssociation.html) on `has_many :fillters` relation to be consistent with the case when we build new filters? If you think that would be too big change, we could perhaps add our own before filter that would save changed filters. That way it should work nicely with the transaction in plugin [DSL](https://github.com/theforeman/foreman/blob/develop/app/services/foreman/plugin.rb#L240-L243)

app/models/role.rb
```diff
@@ -125,17 +129,26 @@ def add_permissions(permissions, options = {})
...
     end
   end
 
+  def permission_diff(permission_names)
+    current_names = permissions.map(&:name).map(&:to_sym)
+    current_names - permission_names | permission_names - current_names
```
ares: I think this line deserves parenthesis, do I read it correctly?

```ruby
(current_names - permission_names) | (permission_names - current_names)
```

app/services/foreman/plugin/rbac_support.rb
```diff
@@ -0,0 +1,76 @@
...
+        resources.map(&:constantize)
+        Role.without_auditing do
+          Role.transaction do
+            add_resource_permissions_to_viewer(resources, opts.dup)
+            add_resource_permissions_to_manager(resources, opts.dup)
+          end
+        end
+      end
+
+      def add_permissions_to_default_roles(args)
```
ares: the method would be more readable if variables would be named according to what they represent, I'd suggest

s/args/roles/
s/key/role_name/
s/array/permissions/

app/services/foreman/plugin/rbac_support.rb
```diff
@@ -0,0 +1,76 @@
...
+        args.each do |key, array|
+          check_role_name_before_extending key
+          role = Role.find_by :name => key
+          next unless role
+          role.modify_locked = true
```
ares: should this be set back to false after the `add_permissions_to_role` call? maybe the `modify_locked` (or other method) could accept block so one could write
```ruby
role.ignore_locking do
  add_permissions_to_role(role, array)
end
```

app/services/foreman/plugin/rbac_support.rb
```diff
@@ -0,0 +1,76 @@
...
+          add_permissions_to_role(role, array)
```
ares: if `add_permissions_to_role` won't save existing filters right away, I think explicit save is required here (might be needed for built filters anyway

app/services/foreman/plugin/rbac_support.rb
```diff
@@ -0,0 +1,76 @@
...
+        end
+      end
+
+      def check_role_name_before_extending(role_name)
+        msg = "Invalid role name, only 'Viewer' of 'Manager' is allowed to be extended from a plugin"
+        raise msg if role_name != "Viewer" && role_name != "Manager"
```
ares: Foreman::Exception should be used I think

app/services/foreman/plugin/rbac_support.rb
```diff
@@ -0,0 +1,76 @@
...
+        viewer = Role.find_by :name => "Viewer"
+        opts[:condition] = "name LIKE :name"
+        opts[:condition_hash] = { :name => "view_%" }
+        return if !viewer || resources.empty?
+        add_permissions viewer, resources, opts
+      end
+
+      def add_permissions(role, resources, opts)
+        role.modify_locked = true
+        to_add = resources.map { |resource| new_permissions_for role, resource, opts }.reduce(:+).map(&:name)
```
ares: `reduce(:+)` should be equivalent to `sum` since we use activesupprt

app/services/foreman/plugin/rbac_support.rb
```diff
@@ -0,0 +1,76 @@
...
+      end
+
+      def permission_already_included?(role, permission, opts)
+        opts[:except] ||= []
+        opts[:except].include?(permission.name.to_sym) || role.permissions.where(:name => permission.name).any?
+      end
+
+      def add_permissions_to_role(role, permission_names)
+        Role.without_auditing do
+          to_add = permission_names.reject do |name|
```
ares: I think it would be better to first load all role permissions so avoid separate SQL for each permission, e.g.

```ruby
existing_permissions = role.permissions.where(:name => permission_names).pluck(:name)
to_add = permission_names.select { |name| !existing_permissions.include?(name) }
```

app/models/role.rb
```diff
@@ -145,6 +158,11 @@ def disable_filters_overriding
     self.filters.where(:override => true).map { |filter| filter.disable_overriding! }
   end
 
+  def locked?
+    return false unless respond_to? :origin
```
ares: is this here because migration adding `origin` attribute might not ran yet? in such case, could you please add a comment?

db/migrate/20170221195674_tidy_current_roles.rb
```diff
@@ -0,0 +1,45 @@
...
+    if role.valid?
+      role.save!
+    else
+      rename_with_valid_name role, prefix, original_name
+    end
+  end
+
+  def rename_with_valid_name(role, prefix, original_name)
+    num = 0
+    until role.valid?
```
ares: What if role is invalid not because of the name? this would never exit the loop. I think we the migration should not even try to do the validations. The reason is that the migration can become broken later when Role model would introduce new validations. I'd recommend changing the code so that it find the first name that can be used (finding using active recrod is safe). Then do the role update skipping validations and callback, normally we use `update_all`, e.g. `Role.where(:id => role.id).update_all(:name => found_free_name)`

db/seeds.d/03-permissions.rb
```diff
@@ -1,169 +1,3 @@
-permissions = [
```
ares, line 1: I needed to do something similar for templates and I ended up moving it to separate file in [seeds](https://github.com/theforeman/foreman/pull/4320/files#diff-f63d6f59b7e054afbab817ac932cd46bR15), I suppose we should do it the same way. I'm happy to change my approach if you think it makes more sense to put it under app/services/foreman/constants but I was thinking seed related data should be in db/seeds.d. I'd appreciate others ideas.

lib/seed_helper.rb
```diff
@@ -0,0 +1,52 @@
+# methods which are used in seeds and migrations
+class SeedHelper
```
ares, line 2: I like extracting it to separate class. I suppose the motivation was for these helpers to be testable but I don't see any SeedHelper test. Or was the motivation different?

ares (changes requested):
Thanks @xprazak2, I left few comments in the code. During the review, github notified me that a new version was pushed so I hope most will be still relevant. I plan to do the testing after comments are resolved.

app/services/foreman/plugin/rbac_support.rb
xprazak2: I will use `flat_map` here.

db/seeds.d/03-permissions.rb
xprazak2: Moving to separate file in seeds.d does not work for me, I need the lists available in migrations as well. I am not sure if the current location is the most suitable though.

Edit: I missed you have to require them from seeds. I will make changes.

lib/seed_helper.rb
xprazak2: Main motivation was to reuse the methods in migrations, but adding tests is not a bad idea, I will add some.

xprazak2:
Main changes:

* added `ignore_locking` method for roles which takes a block where you can modify the locked role
* autosave for `has_many :filters`
* I renamed a couple of methods in `rbac_support`, I had `add_permissions` there and same method was defined for roles, which could cause confusion, since both methods were used in there. Btw. I realize the class name is not that great, so if anyone has suggestions, I will be grateful. 
* I skip validations in migrations and just look for a name that has not yet been taken.

Still facing failures in migrations on Jenkins, updates for that will follow.

- [x]  Jenkins migration failures
- [x]  `locked?` responding to `:origin`
- [x]  `SeedHelper` tests

xprazak2:
I pushed changes to fix tests + `:autosave` working nicely.

test/unit/seed_helper_test.rb
```diff
@@ -0,0 +1,21 @@
...
+
+class SeedHelperTest < ActiveSupport::TestCase
+  test "should create locked role" do
+    role_name = "Test role"
+    permissions_names = [:view_hosts, :destroy_hosts]
+    refute Role.find_by(:name => role_name)
+    SeedHelper.create_role role_name, permissions_names, 0
+    role = Role.find_by(:name => role_name)
+    assert role
+    assert_equal permissions_names, role.permissions.pluck(:name).sort.map(&:to_sym)
```
ares: this fails because of order, make sure you sort both sets

ares:
Couple of things found during reviewing and testing last version:

1. cloning locked role does not work, getting "Unable to save
This role is locked from being modified by users.", I think cloning needs to reset the role.
2. when applied on existing DB, I get customized roles cloned e.g. "Custom Manager", but newly created Manager is not locked
3. The "Default role" has a special meaning which is determined by it's name. People might need to change it's filters. I'm not sure what would be the best option here, the most consistent would be to have a setting to chose which role is considered default (as you recently did for global pxe template). But the change of this setting should ensure that all users have this role assigned and the previous one unassigned so it's not that simple.
4. We have now two locking attributes, one is `builtin` attribute, second is origin. While origin is nice to have for sure, could we rather use the `builtin` attribute for built-in roles?
5. Many tests are broken

I'm going to test it further but I wanted to provide early feedback ASAP.

xprazak2:
Changes:
1. cloning of roles should work now, I just needed to exclude `:origin`
2. `Role.reset_column_information` helped locally, lets see what Jenkins tells us
3. This is a tricky one. I am not sure if users should be allowed to modify filters for "Default role", I think it depends on what setup they have. "Global default role" could be a nice feature, but I am opened to leaving it unlocked for now, resetting and locking it in the future should not be that difficult if we decide to go that way. I will make changes.
4. I was thinking about the same thing, but then I rejected it. "Default role" is now both builtin and locked, which means I would have to use some sort of (bit)mask like we have in host status and I wanted to avoid that. Even if I unlock the "Default role" (as planned in 3.), I have to use some arbitrary "magic number" to determine the role is locked, and it really does not say much. And I would be effectively storing 2 unrelated pieces of information in one column, which could be confusing. The nice thing about origin is that it clearly states where the role comes from and we will be able to look up all roles that belong to a certain plugin quite easily.

ares (changes requested):

ares:
foreman tests seems to be green now, did you have a chance to look at katello tests?

EDIT: the PR needs rebasing now

xprazak2:
All foreman tests should turn green now. I am dealing with katello tests [separately](https://github.com/Katello/katello/pull/6663), but I expect some seed tests to fail because I changed `audit_modified?` to `SeedHelper.audit_modified?`. I will have a separate PR for that on katello side as well. 

Note to self: do not forget to unlock the default role
Edit: default role is no longer locked

xprazak2:
[test katello]

xprazak2:
[test katello]

xprazak2:
All tests are now passing except for a couple of katello's seed tests that require change from `audit_modified?` to `SeedHelper.audit_modified?` and too much complexity in role model flagged by code climate.

app/services/foreman/plugin.rb
```diff
@@ -234,15 +238,30 @@ def permission(name, hash, options = {})
...
         role = Role.where(:name => name).first_or_create
         role.add_permissions!(permissions) if role.permissions.empty?
+        rbac_registry.role_ids << role.id
       end
     end
 
+    # Add plugin permissions to core's Manager and Viewer roles
+    # Usage:
+    # add_permissions_to_default_roles ['MyPlugin::FirstResource', 'MyPlugin::SecondResource'], :except => [:skip_this_permission]
+    def add_resource_permissions_to_default_roles(resources, opts = {})
```
ares: If we have all permissions registered, why do we need to specify resources here? don't we know these already? it seems weird to me that I basically need to list all classes of my plugin?

EDIT: maybe we could leave this for better granularity but expose another method "add_all_permissions_to_default_roles" that would go through registry and add all (with or without resource)?

ares:
> All tests are now passing except for a couple of katello's seed tests that require change from audit_modified? to SeedHelper.audit_modified? and too much complexity in role model flagged by code climate.

What if we kept the method there and it would use `SeedHelper.audit_modified?` internally? Perhaps with deprecation warning?

app/models/role.rb
```diff
@@ -149,13 +159,25 @@ def disable_filters_overriding
   end
 
   def clone(role_params = {})
-    new_role = self.deep_clone(:except => [:name, :builtin],
+    new_role = self.deep_clone(:except  => [:name, :builtin, :origin],
```
ares: nit: extra space

ares:
could you please rebase?

app/models/role.rb
```diff
@@ -149,13 +159,25 @@ def disable_filters_overriding
...
   end
 
+  def locked?
+    return false unless respond_to? :origin
+    origin.present? && builtin != BUILTIN_DEFAULT_ROLE
+  end
+
+  def ignore_locking(&block)
+    self.modify_locked = true
+    yield self
```
ares: I think it'd be better either explicitly call the block by `block.call(self)` or don't specify it's name if you're using `yield`

xprazak2:
I rebased, modified `ignore_locking` method as suggested and returned `audit_modified?` into seeds

db/seeds.rb
```diff
@@ -12,52 +12,9 @@ def format_errors(model = nil)
...
-def create_role(role_name, permission_names, builtin)
-  return if Role.find_by_name(role_name)
-  return if audit_modified?(Role, role_name) && (builtin == 0)
-
-  role         = Role.new(:name => role_name)
-  role.builtin = builtin
-  role.save!
-  permissions = Permission.where(:name => permission_names)
-  create_filters(role, permissions)
+  puts "This method will be removed, use SeedHelper.audit_modified? instead."
```
ares: why not using  `Foreman::Deprecation.deprecation_warning('1.17', 'warning and reasoning message')`?

app/services/foreman/plugin.rb
```diff
@@ -234,15 +238,34 @@ def permission(name, hash, options = {})
...
+    end
+
+    # Add plugin permissions to Manager and Viewer roles. Use this for permissions without resource_type or to handle special cases
+    # Usage:
+    # add_permissions_to_default_roles 'Role Name' => [:first_permission, :second_permission]
+    def add_permissions_to_default_roles(args)
+      Plugin::RbacSupport.new.add_permissions_to_default_roles args
+    end
+
+    def add_all_permissions_to_default_roles
```
ares: Thanks, would you mind adding the similar docs as for add_permissions_to_default_roles (mentioning it adds all registered permissions to Manager and Viewer roles). Also a test for this method would be useful I think.

ares:
Thanks @xprazak2, I left two minor comments. I'll do another round of testing meanwhile. @lzap is there something missing from your point of view? Given I don't find any issues during testing and @xprazak2 will address those two comments I'd like to merge this.

db/migrate/20170221195674_tidy_current_roles.rb
```diff
@@ -0,0 +1,59 @@
...
+    new_name = generate_name prefix, original_name, num
+    while Role.find_by :name => new_name
+      new_name = generate_name prefix, original_name, num
+      num += 1
+    end
+    role.update_attribute :name => new_name
+  end
+
+  def generate_name(prefix, original, num)
+    "#{prefix} #{original_name} #{num}"
```
ares: I suppose the argument name should have been `original_name`
```
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

undefined local variable or method `original_name' for #<TidyCurrentRoles:0x0055b0ca671968>
/home/ares/Projekty/Zdrojaky/foreman/db/migrate/20170221195674_tidy_current_roles.rb:49:in `generate_name'
/home/ares/Projekty/Zdrojaky/foreman/db/migrate/20170221195674_tidy_current_roles.rb:40:in `rename_with_free_name'
/home/ares/Projekty/Zdrojaky/foreman/db/migrate/20170221195674_tidy_current_roles.rb:32:in `rename_existing'
/home/ares/Projekty/Zdrojaky/foreman/db/migrate/20170221195674_tidy_current_roles.rb:24:in `process_existing'
/home/ares/Projekty/Zdrojaky/foreman/db/migrate/20170221195674_tidy_current_roles.rb:12:in `block (2 levels) in up'
/home/ares/Projekty/Zdrojaky/foreman/db/migrate/20170221195674_tidy_current_roles.rb:9:in `each'
/home/ares/Projekty/Zdrojaky/foreman/db/migrate/20170221195674_tidy_current_roles.rb:9:in `block in up'
/home/ares/Projekty/Zdrojaky/foreman/db/migrate/20170221195674_tidy_current_roles.rb:8:in `up'
-e:1:in `<main>'
```

db/migrate/20170221195674_tidy_current_roles.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def rename_with_free_name(role, prefix, original_name)
+    num = 0
+    new_name = generate_name prefix, original_name, num
+    while Role.find_by :name => new_name
+      new_name = generate_name prefix, original_name, num
+      num += 1
+    end
+    role.update_attribute :name => new_name
```
ares: update_attribute requires two parameters

`role.update_attribute :name, new_name`

db/migrate/20170221195674_tidy_current_roles.rb
```diff
@@ -0,0 +1,59 @@
...
+    role_name = "#{prefix} #{original_name}"
+    if Role.find_by(:name => role_name)
+      rename_with_free_name role, prefix, original_name
+    else
+      role.update_attribute :name, role_name
+    end
+  end
+
+  def rename_with_free_name(role, prefix, original_name)
+    num = 0
```
ares: this generates "Custom Manager 0" as the first index, maybe `num = 1` would make more sense

db/migrate/20170221195674_tidy_current_roles.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def process_existing(original_name, role, permission_names)
+    diff = role.permission_diff permission_names
+    return if diff.empty?
+    rename_existing role, original_name
+    create_from_seeds original_name, permission_names
+  end
+
+  def rename_existing(role, original_name)
+    prefix = "Custom"
```
ares: would "Customized" prefix make more sense? maybe @GregSutcliffe or some other native speaker could let us know which one sounds better "Custom Manager" or "Customized Manager"

app/services/foreman/plugin.rb
```diff
@@ -234,15 +238,34 @@ def permission(name, hash, options = {})
 
     # Add a new role if it doesn't exist
     def role(name, permissions)
-      @default_roles[name] = permissions
+      default_roles[name] = permissions
       return false if pending_migrations || Rails.env.test?
 
       Role.transaction do
         role = Role.where(:name => name).first_or_create
```
ares, line 245: now when we lock shipped roles, should role created by plugin be also locked? and in case we find existing (from previous installation, we should probably make sure we lock it from now on)

app/services/foreman/plugin.rb
```diff
@@ -234,15 +238,34 @@ def permission(name, hash, options = {})
...
         role.add_permissions!(permissions) if role.permissions.empty?
```
ares, line 246: Now when roles are locked and you added support for in memory searching, we don't need to add new filters only if there are no assigned already. Without the if condition it works as expected, if the permission is missing, it's added. No duplicates seems to be created. The reason why we avoided this in past was that user could set some filter condition and we'd add the permission to such filter but this is no longer possible for shipped (locked) roles.

TL;DR pls remove the if condition on this line :-)

app/services/foreman/plugin.rb
ares: actually I was wrong, dupes are created, it also needs update in `add_permissions` method to skip existing permissions, this seems to do the trick

```rb
grouped_permissions.each do |permission|
  next if filter.permissions.include?(permission) # this line is required
  filtering = filter.filterings.build
  filtering.filter = filter
  filtering.permission = permission
end

```

xprazak2:
I changed the rename indexing from 0 to 1, used `Foreman::Deprecation` and I use the correct params for `update_attribute`

lzap (approved):

ares (approved):

ares:
Foreman test failure is unrelated. I'd kindly ask you for follow up PR that locks plugin roles with the same mechanism we have here and update of 1.15 manual. It does not block this PR though so I'll merge in a minute.

ares:
Merged as 72a28408aa23bf4f7f7a6fa201abee8305c253ba, thanks @xprazak2

lzap:
Good job folks, thanks.

PR #4169, opened 2017-01-10
fixes #13810 - increase value range for id colums

timogoebel:


timogoebel:
[test foreman]

timogoebel:
Changed this to only run on only mysql and postgres.
Bigint is the default for Rails 5.1.

Deleted the houndci comments. Single quoted strings are better than double quoted.

timogoebel:
[test upgrade]

dLobatog:
@timogoebel Please rebase to get the upgrade test passing (& no comments from Hound)

timogoebel:
@dLobatog : I already rebased, test upgrade fails because of

```
ERROR:  While executing gem ... (Gem::FilePermissionError)
    You don't have write permissions for the /usr/local/rvm/gems/ruby-2.0.0-p481@test_develop_pr_upgrade-1 
```

I think, there is nothing I can do about that.

dLobatog:
Thanks @timogoebel !

PR #2160, opened 2015-02-15
Fixes #9388 - Adding cname support in foreman.

RodolpheFouquet:
Hello, sorry I am redoing my pull request since I screwed up the previous one (#2158, you can delete it) 

I have added a cname support in foreman UI and in the V2 API.

This consists in a field in the host tab, this field splits a comma separated list of aliases.

Theses aliases are automatically formatted using javascript if the aliases are only separated by spaces of only by a comma or by a comma + spaces.

This new issue is related to the tracking issue: http://projects.theforeman.org/issues/5409

PR #2158, opened 2015-02-14
Fixes #9388 - Adding cname support in foreman.

RodolpheFouquet:
I have added a cname support in foreman UI and in the V2 API.

This consists in a field in the host tab, this field splits a comma separated list of aliases.

Theses aliases are automatically formatted using javascript if the aliases are only separated by spaces of only by a comma or by a comma + spaces.

This new issue is related to the tracking issue: http://projects.theforeman.org/issues/5409

PR #4171, opened 2017-01-10
Fixes #9388 - CNAME Support on UI

guimaluf:
WIP Include CNAMERecord for DNS

### DIFF
--- a/app/models/host_alias.rb
+++ b/app/models/host_alias.rb
@@ -1,0 +1,13 @@
+class HostAlias < ActiveRecord::Base
+  belongs_to :domain
+  belongs_to :nic, :foreign_key => :nic_id
+
+  validates :name, :nic_id, :presence => true
+
+  delegate :host, :to => :'nic.host'
+  delegate :hostname, :to => :'nic.host.hostname'
+
+  def to_s
+    name
+  end
+end
--- a/app/models/host_alias.rb
+++ b/app/models/host_alias.rb
@@ -1,13 +1,15 @@
 class HostAlias < ActiveRecord::Base
   belongs_to :domain
-  belongs_to :nic, :foreign_key => :nic_id
+  belongs_to :nic, :class_name => 'Nic::Base'
 
-  validates :name, :nic_id, :presence => true
-
-  delegate :host, :to => :'nic.host'
-  delegate :hostname, :to => :'nic.host.hostname'
+  validates :nic_id, :presence => true
+  validates :name, :presence => true, :uniqueness => {:scope => :domain_id}
 
   def to_s
     name
+  end
+
+  def cname
+    nic.host.fqdn
   end
 end
--- a/app/controllers/concerns/foreman/controller/parameters/host_alias.rb
+++ b/app/controllers/concerns/foreman/controller/parameters/host_alias.rb
@@ -1,0 +1,18 @@
+module Foreman::Controller::Parameters::HostAlias
+  extend ActiveSupport::Concern
+
+  class_methods do
+    def host_alias_params_filter
+      Foreman::ParameterFilter.new(::HostAlias).tap do |filter|
+        filter.permit_by_context :nic, :nic_id,
+          :name,
+          :domain, :domain_id,
+          :nested => true
+
+        filter.permit_by_context :id,
+          :_destroy,
+          :ui => false, :api => false, :nested => true
+      end
+    end
+  end
+end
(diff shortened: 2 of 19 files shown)


timogoebel:
@guimaluf, thanks for your PR.

First of all, have you seen this already? https://github.com/theforeman/foreman/pull/2158/commits/7ac9b965fffe4a2a4e2d6dee4df3603996d40f55

This might give you some more ideas.

Could you briefly describe, what exactly do you want to archive with this. How do you imagine CNAME-Records to work?

guimaluf:
@timogoebel 
Yes, I saw it! I'm using #2160  as a reference for what I'm doing.

I don't know exactly how foreman works behind the scenes, I'm still studying it.
In a high level view, I imagine two possibilities:
1. Include another input text in interface form allowing an array of alias.
2. Reuse the same input text for DNS name allowing an array of strings where the first string is the hostname and all the rest aliases.

What do you think? What detail level do you need?

I'll keep working on it and thanks for replying!

dLobatog:
@guimaluf The hound is just being stupid, we have to fix it. Single quotes are better unless there's string interpolation, sorry for the mess of comments.

guimaluf:
@dLobatog now I've changed all my single quotes? hahahahaha no problem, I'll rollback later

dLobatog:
@guimaluf No worries, I don't think we enforce either style 😄 It's just that hound for some reason is using the default style they have instead of picking up our hound.yaml for some reason

timogoebel:
> What do you think? What detail level do you need?

That's all I wanted to know. So basically you want the CNAME aliases to be deloyed when the host is created and deleted when the host is deleted, correct?

guimaluf:
@dLobatog now I've changed all my single quotes? hahahahaha no problem, I'll rollback later!

I would like to address everyone who commented on PR #2160 and here
@timogoebel @ohadlevy @ares @dLobatog 

**How should I include host aliases on interfaces?**
Should I create a new model/table belonging to `DnsInterface`(or `Nic::Managed`? or ?) containing the aliases or a string attribute which will contain the whole list of aliases?

do you guys suggests anything better then this?

timogoebel:
@guimaluf : Some thoughts:

- I'd create a new model (let's say `AliasRecord`). The `Nic::Base` can than have many AliasRecords.
- The `AliasRecord` should belong to a `Domain`.
- I'd suggest doing the orchestration (e.g. deploying the dns record), when the model and ui is completely done.
- You can get the smart proxy to deploy the Dns Changes from the domain.
- `DnsInterface` is just a concern. I wouldn't add the cname records here.
- I'd suggest that the UI looks similiar to Parameters. You could have a text field to specify the record name and a select field to specify the domain.

app/models/host_alias.rb
```diff
@@ -0,0 +1,4 @@
+class HostAlias < ActiveRecord::Base
+  belongs_to :domains
+  belongs_to :nics
```
ohadlevy: could you add a validation that the alias is unique across the domain

guimaluf: This is WIP! :)
I'll do more stuff, just stuck with other things!

guimaluf:
@timogoebel 
> I'd suggest that the UI looks similiar to Parameters. You could have a text field to specify the record  name and a select field to specify the domain.

Should I create a tab on Interface modal or include in the bottom the aliases fields?

guimaluf:
screenshot from 2017-02-07 12-33-05


what do you guys think about this type of interface?

app/models/host_alias.rb
```diff
@@ -0,0 +1,13 @@
+class HostAlias < ActiveRecord::Base
+  belongs_to :domain
```
ohadlevy, line 2: should it be unique in the context of a domain? e.g. you can't have more than one from a domain.

guimaluf: I don't think so
you could have `mywebserver.example.com` been an alias to `api.foobar.com` and to `www.foobar.com`

ohadlevy: i mean it the other way around, only one server can have an alias for api.foobar.com

guimaluf:
I'm struggling to write the interface/controller. I need help!

##### Can't align `remove` link on alias table with `host alias` and `domain` input
##### Can't set host_aliases parameters correctly in controller/view
host_alias input are like `host[interfaces_attributes][0][host_aliases][name]` while it should be
`host[interfaces_attributes][0][host_aliases][0][name]`. I really don't know where to set this

ohadlevy:
@guimaluf what is the status of the PR ? is the last comment still accurate ?

guimaluf:
@ohadlevy I needed to focus on other tasks and stop working on this.  

I'm stucked on interface process, as I mentioned above. I did not submitted my last changes, so I'm doing this right now.

I'd appreciate If anyone could help!

PR #4174, opened 2017-01-11
help block from libvirt updated

waldirio:
help block from libvirt updated from 
*e.g. qemu://host.example.com/system* 
  to 
*e.g. qemu+ssh://root@host.example.com/system*

According the official documentation


dLobatog:
@waldirio Thanks for the PR, but I'm afraid this isn't really necessary. `qemu://` works well with authentication via certificates or without authentication. The docs also tell you how to use `qemu+ssh` and `qemu+tcp`, true, but the current default option definitely works too. Let us know if you had any problems with it, I'll close this for the time being, please reopen if you beg to differ.

lzap:
To be fair, I've only used `qemu:///system` without any hostname or qemu+ssh and nothing else.

PR #4176, opened 2017-01-11
Fixes #12599 - Add config groups related apipie doc

xprazak2:


app/controllers/api/v2/config_groups_controller.rb
```diff
@@ -23,6 +23,8 @@ def show
       def_param_group :config_group do
         param :config_group, Hash, :required => true, :action_aware => true do
           param :name, String, :required => true
+          param :puppetclass_ids, Array, :desc => N_("IDs of associated Puppet classes")
+          param :puppetclass_names, Array, :desc => N_("Names of associated Puppet classes")
```
ares: I think we encountered this elsewhere recently too, we normally don't provide name keys in API and let hammer do it. While I agree this would be handy in API it would probably be better to do it for all resources at once. @tstrachota correct?

xprazak2: When I add a param for ids, hammer adds the param for names automatically, so I am removing this line.

tstrachota: +1 I agree @ares

app/controllers/api/v2/hostgroups_controller.rb
```diff
@@ -40,6 +40,8 @@ def show
           param :subnet_id, :number, :desc => N_('Subnet ID')
           param :domain_id, :number, :desc => N_('Domain ID')
           param :realm_id, :number, :desc => N_('Realm ID')
+          param :config_group_ids, Array, :desc => N_("IDs of associated config groups")
+          param :config_group_names, Array, :desc => N_("Names of associated config groups")
```
ares: should the [same thing](https://github.com/theforeman/foreman/pull/4176#discussion_r95616147) be applied here and in app/controllers/api/v2/hosts_controller.rb ?

xprazak2: Yes it should. I pushed poorly.

app/controllers/api/v2/config_groups_controller.rb
```diff
@@ -23,6 +23,7 @@ def show
       def_param_group :config_group do
         param :config_group, Hash, :required => true, :action_aware => true do
           param :name, String, :required => true
+          param :puppetclass_ids, Array
```
ares, line 26: if we had a proper test [here](https://github.com/theforeman/foreman/blob/develop/test/controllers/api/v2/config_groups_controller_test.rb#L13) and verify that puppet classes were added we'd probably notice the missing parameter sooner, would you mind extending it?

ares (changes requested):
Mostly ready I think.

xprazak2:
I expanded the test case a bit.

dLobatog:
@ares Test is ready here - should be ready to merge IMO

ares (approved):
Yup, sorry I didn't get to re-review sooner

dLobatog:
Thanks @xprazak2 @ares

PR #4177, opened 2017-01-11
fixes #17899 - Use fake objects in orphaned records migration

orrabin:


db/migrate/20161205142618_delete_orphaned_smart_class_parameters.rb
```diff
@@ -1,6 +1,19 @@
...
+  class FakePuppetclassLookupKey < ActiveRecord::Base
+    self.table_name = 'lookup_keys'
+
+    has_many :lookup_values, :dependent => :destroy, :class_name => 'FakePuppetclassLookupKey'
+  end
+
+  class FakeLookupValue < ActiveRecord::Base
+    self.table_name = 'lookup_values'
+
+    belongs_to :lookup_key, :class_name => 'FakePuppetclassLookupKey', :foreign_key => 'lookup_key_id'
```
ohadlevy: if we are not using any of the callbacks in the object, then why not mass delete the lookup_values and then all of the lookup keys in one transaction?

tbrisker:
Thanks @orrabin !

PR #4179, opened 2017-01-11
Fixes #4519, #9531 - Correctly handle invalid host name changes

tbrisker:
Removing a host name or otherwise changing it to an invalid name (such
as a name containing /) caused the form to fail loading. This makes sure
all paths are generated correctly and validation error is shown.


ShimShtein:
:+1: Works as expected. Let's wait for tests to finish...

dLobatog:
Thanks @tbrisker @ShimShtein !

PR #4185, opened 2017-01-12
Fixes #18040 - URL escape PuppetCA CN on proxy view

dLobatog:
If the CN contains characters that cannot be displayed in an URL, like
'mcollective/OL=mcollective', the puppetca list will not be able to
render.

The reason is that Rails cannot generate an URL for such CNs, so we need
to convert it into URL-friendly style.

### DIFF
--- a/app/views/puppetca/_list.html.erb
+++ b/app/views/puppetca/_list.html.erb
@@ -23,11 +23,11 @@
         <td>
           <%= action_buttons(
                     if cert.state == "pending"
-                      display_link_if_authorized(_("Sign"), hash_for_smart_proxy_puppetca_path(:smart_proxy_id => @proxy.to_param, :id => cert), :method => :put)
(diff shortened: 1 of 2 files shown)


domcleal (changes requested):
Could you add a controller test please, covering the index action?

dLobatog:
@domcleal Updated with a test now for the index action. If you remove the `url_encode`, you'll see the action will throw `ActionView::Template::Error: No route matches {:action=>"update", :controller=>"puppetca", :id=>"mcollective/OL=mcollective", :smart_proxy_id=>"182953976-Puppetmaster Proxy"} missing required keys: [:id]` which is the current behavior without the patch.

domcleal (approved):

domcleal:
Merged, thanks @dLobatog.

PR #4186, opened 2017-01-12
Fixes #18045 - Puppet classes show up choosing only env.

dLobatog:
After #3551 was merged, the hosts controller requires both environment
and hostgroup to be set in order to display puppetclasses.

It shouldn't be required to have both, so we should check what's
available and use it. If it's only the hostgroup or the environment, it
should work.

To reproduce, simply go to Hosts > New host, and select environment (without selecting hostgroup), notice how puppet classes do not load anymore.

cc @bastilian

### DIFF
--- a/app/controllers/concerns/foreman/controller/puppet/hosts_controller_extensions.rb
+++ b/app/controllers/concerns/foreman/controller/puppet/hosts_controller_extensions.rb
@@ -8,6 +8,7 @@ module Foreman::Controller::Puppet::HostsControllerExtensions
                              select_multiple_puppet_proxy update_multiple_puppet_proxy
                              select_multiple_puppet_ca_proxy update_multiple_puppet_ca_proxy)
   PUPPET_MULTIPLE_ACTIONS = %w(multiple_puppetrun update_multiple_puppetrun) + MULTIPLE_EDIT_ACTIONS
+
   included do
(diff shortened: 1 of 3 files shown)


dLobatog:
Hold off before merging, 007bf25498bf157fc7b95f46be5a457e51bda237 introduce this method twice

ShimShtein:
It would be better, if you added call to `set_class_variables` from `#hostgroup_or_environment_selected` in `puppet/host_controller_extensions`. Then just add to `#set_class_variables`
``` ruby
@hostgroup = host.hostgroup
```
and in `puppet/host_controller_extensions` already adds `@environment` to `#set_puppet_class_variables`

dLobatog:
Updated - to see difference, notice how choosing *only* environment or hostgroup before this patch doesn't load the Puppet Classes tab, but it does load it with this patch.

@ShimShtein  Fixed the PR to call set_class_variables and find the host right away, if it found `host.environment` or `host.hostgroup`, it'll load the classes tab, otherwise it won't.

dLobatog:
Test failure is  one of these unreliable integration tests

ShimShtein:
:+1: Works as expected!

ShimShtein:
[test foreman]

ohadlevy:
works as expected - thanks @dLobatog

PR #4187, opened 2017-01-12
Refs #16740 - migration templates to use param macros

ares:


stbenjam:
I believe users can also do `<%= @host.params['foo'] %>` in smart paramteter values, can't they?

ares:
@stbenjam not for smart variables or smart class parameters with safe mode on but I guess we can convert it the same way. ERB in Parameters seems to work even with safe mode so we can convert them too. I needed to add `require_dependency` on order ofr `descendats` to work, seems it was forgotten when STI was introduced in these models.

dLobatog:
I really think we should revert the commit and stop wasting time on writing macros for very stable interfaces like `@host.params`. See https://groups.google.com/forum/?fromgroups#!topic/foreman-dev/P364HozAYZA before continuing.

ares:
I don't think reverting makes sense. If you prefer to keep accepting the `@host.params` then only the deprecation could be reverted. Honestly, I don't think the original PR was wasting of time. Some devs don't see benefits, some I hope do.

ares:
I didn't want to add tests until I get the confirmation people see the value. Since the macros would now do something different than just `<%= @host.params['realm'] %>` I think we could have all variants and let user decide what's better for him. I'm happy to provide more, e.g. `host_param('name', 'default')` which would replace repeating pattern we have in nearly every template `@host.params['name'] || 'default'`

a
b

ares:
seems that hound died :-)

tbrisker:
:-1: I agree with Daniel and with others who have commented on the mailing list. Lets keep the `@host` interface stable instead of rewriting our users' templates. It can be done easily with a proxy object that allows all the checks that are missing in the actual host, additional functionality and ability to change internal host implementation.

ares:
Closing :disappointed:

ares:
Reopening since it's relevant again, now with added value suggested by @ekohl at https://github.com/theforeman/community-templates/pull/343#issuecomment-284984652

lib/foreman/renderer.rb
```diff
@@ -27,26 +29,36 @@ def template_logger
     end
 
     def host_enc(*path)
+      check_host
```
iNecas, line 33: Shouldn't the `host_enc` also have the `host_enc!`, `host_enc_true?` and `host_enc_false?`

ares: sorry it took me a while,

`host_enc` already raises and exception (`KeyError`) if it can't fetch a given parameter. Removing it would break existing API so I just improved it so it raises more appropriate exception that inherits from `RenderingError`

`host_enc_true?` and `host_enc_false?` should not be required since these values keep their types, if smart class parameter or smart variable is defined as Boolean, it will be always caster to true or false.

ares:
@iNecas I kept the update of exception in separate commit for easier reviewing. Could you please take another look? I also added 2 tests for `host_enc`.

db/migrate/20170112175131_migrate_template_to_parameters_macros.rb
```diff
@@ -0,0 +1,53 @@
+class MigrateTemplateToParametersMacros < ActiveRecord::Migration
+  def up
+    Template.descendants.each do |klass|
+      klass.all.each do |template|
```
iNecas: it needs to be `unscoped`, otherwise taxonomy (or user) filter leads to return an empty set and nothing actually gets migrated.

iNecas (changes requested):

iNecas:
Tested and the templates change works well. One last thing: the migration updated the template, but there is not update in the template's audit log. I wonder if that should not be the case? On the other hand, we don't do it even for db:seed when updating the templates, so I guess out of scope for this issue. Anyway, do you think it's worth a Redmine issue?

iNecas:
Pending adding the `unscoped` to the migrations

ares:
@iNecas updated, originally the PR was opened before the `unscoped` was necessary, good catch. Regarding audits, I don't think we should audit changes not caused by users. The same thing happens during Foreman upgrades since we update the template content. Hopefully we'll have better mechanism for updates using rubocop soon.

iNecas:
Well, I think it would be nice feature to be able to see what changed in the templates I'm using between the releases. Anyway, outside of scope for this PR.

iNecas:
There is no point in keeping this PR any longer in opened state. Thanks @ares, tested and works well.

lib/foreman/renderer.rb
```diff
@@ -27,26 +30,39 @@ def template_logger
...
       @enc ||= @host.info.deep_dup
       return @enc if path.compact.empty?
       enc = @enc
+      step = nil
       path.each { |step| enc = enc.fetch step }
       enc
+    rescue KeyError
+      raise(HostENCParamUndefined, _('Parameter %{name} is not set in host %{host} ENC output') % { :name => step, :host => @host })
```
iNecas: One last thing: I guess we should have `path` here, rather than `step`, as if the `step` is nested deeper in some path, it might not be clear which step exactly it is.

iNecas (changes requested):
One last thing: sorry about that :)

lib/foreman/renderer.rb
ares: I added both, from path only it would not be clear what step is missing which might be hard to find in case of big hashes

ares:
[test katello]

iNecas:
Katello test failures addressed https://github.com/theforeman/foreman/pull/4826, let's wait that getting resolved before we proceed

ares:
[test katello] let's see

iNecas (approved):

iNecas:
Could you try to rebase - I'm not sure our jenkins does the merge to develop in the katello tests: from the tests it seems the fix we did was not present there

ares:
Rebased, let's see again :-)

iNecas:
All green now. thanks @ares

iNecas:
@ares the original issue is assigned to 1.15, but I suppose we're not cherry-picking this one

ares:
No, I should have opened extra ticket for it :-( that's the problem of old PRs and refs...

PR #4189, opened 2017-01-13
Fixes #16846 - Org create failure with specific msg

amitkarsale:
While creating an Organization and passing non existing attributes as parameter values should give appropriate message.

[this commit has partial dependency on katello side as katello is making a call to foreman's base controller action, can be merged in any order, katello first or foreman first]

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -39,8 +39,8 @@ module Api
 
     rescue_from Foreman::MaintenanceException, :with => :service_unavailable
 
-    def get_resource
-      instance_variable_get(:"@#{resource_name}") || raise('no resource loaded')
+    def get_resource(message = "Couldn't find resource")
+      instance_variable_get(:"@#{resource_name}") || raise(message)
     end
 
     def controller_permission
@@ -121,7 +121,7 @@ module Api
     end
 
     def process_resource_error(options = { })
-      resource = options[:resource] || get_resource
+      resource = options[:resource] || get_resource(options[:message])
 
       raise 'resource have no errors' if resource.errors.empty?
 


dLobatog:
@amitkarsale Please rebase with the latest develop to clean the test failures. How do I reproduce what this PR fixes? Mind to write some test to demonstrate the behavior? Thanks!

app/controllers/api/base_controller.rb
```diff
@@ -39,8 +39,8 @@ class BaseController < ActionController::Base
 
     rescue_from Foreman::MaintenanceException, :with => :service_unavailable
 
-    def get_resource
-      instance_variable_get(:"@#{resource_name}") || raise('no resource loaded')
+    def get_resource(options = {})
```
ShimShtein: Maybe it's time to start using keyword arguments?
@dLobatog, what do you think?

iNecas: I guess this would be a question the whole @theforeman/core, probably on mailing list? Personally +1, but I would be interested into opinions of wider range before going for it

amitkarsale: +1 @ShimShtein , would be revising it.

ares: Debian Jessie has ruby 2.1, Ubuntu Trusty seems to have 2.0 only but maybe it's time to drop Trusty support in 1.15, @mmoll what do you think? I'd like to start using keyword arguments :-)

ShimShtein: @ares, According to [this](https://robots.thoughtbot.com/ruby-2-keyword-arguments) keyword arguments are there since 2.0, so it should be OK for Trusty too.

Unless you are looking for an excuse to drop Trusty support ;)

ares: I think 2.0 has limited support, all keyword arguments would have to be optional. Maybe that's enough for this case though.

mmoll: I would appreciate if Ruby 2.0 could get supported as long as possible, but if support for it gets dropped in the core code, there's of course nothing I can do except stopping trusty builds ;)

amitkarsale:
@dLobatog the test here to run and test for is in the following PR: https://github.com/Katello/katello/pull/6543/files

app/controllers/api/base_controller.rb
```diff
@@ -121,7 +121,7 @@ def service_unavailable(exception = nil)
     end
 
     def process_resource_error(options = { })
-      resource = options[:resource] || get_resource
+      resource = options[:resource] || get_resource(options)
```
dLobatog: Given what I see on https://github.com/Katello/katello/pull/6543/files, isn't `options[:message]` what you want to pass here? Otherwise you're just passing a hash and calling `raise()` on a hash instead of a string

dLobatog (changes requested):
@amitkarsale Checkout the inline comment, I was able to test it, but with a slight change on `process_resource_error`. Thanks!

app/controllers/api/base_controller.rb
amitkarsale: is it a bad practice to call raise() over hash?

dLobatog (approved):

dLobatog:
Thanks @amitkarsale !

amitkarsale:
@dLobatog Thanks.. :)

PR #4190, opened 2017-01-13
Fixes #18065 - make API v1 deprecation warning more apparent

tstrachota:
Since the text is wrapped in `N_()` and the whole string is translated later in apipie I didn't find any better solution than placing html tags into the string. Ideas are welcome.

### DIFF
--- a/app/controllers/api/v1/base_controller.rb
+++ b/app/controllers/api/v1/base_controller.rb
@@ -5,7 +5,7 @@ module Api
 
       resource_description do
         api_version "v1"
-        app_info N_("Foreman API v1 is deprecated. Please use v2. If you still need to use v1, you may do so by either passing 'version=1' in the Accept Header or using api/v1/ in the URL.")
+        app_info N_("<b>Foreman API v1 is deprecated. Please use v2.</b><br />If you still need to use v1, you may do so by either passing 'version=1' in the Accept Header or using api/v1/ in the URL.")
       end
     end
   end


dLobatog:
@domcleal Mind to take a look? I think at first that the tags in the string is a bad practice and the string should be split to translate it in chunks (and not translate the tags). I've messed up a few times translations so I think it's best if you also can check it out. Thanks!

domcleal:
Yes, translating it as two sentences might be best here, or use % placeholders to add the HTML tags instead of putting them in the translation.

tstrachota:
OK, I'll use the placeholders. I'm afraid we cannot split it into chunks here, because gettext wouldn't match the resulting string with existing translation later in apipie (`N_()` vs `_()`).

tstrachota:
Hmm, even placeholders won't solve the issue. So with current state of apipie there's imho no other solution than having the html tags in the string. If you're against please close the PR. I'm not going to change apipie now as the issue is really low prio.

dLobatog:
@tstrachota 

Would:

```ruby
'<b>' +
 N_("Foreman API v1 is deprecated. Please use v2.") +
'</b><br />' + 
N_("If you still need to use v1, you may do so by either passing 'version=1' in the Accept Header or using api/v1/ in the URL.")
```

work?

ohadlevy:
@dLobatog maybe add it to a notification warning? :)

dLobatog:
@ohadlevy Not sure if serious but this is about `apidoc/v1.html`, not worth adding notifications to that page in my opinion!

tstrachota:
@dLobatog IMHO it wouldn't, because:
```ruby
docs = '<b>' +
 N_("Foreman API v1 is deprecated. Please use v2.") +
'</b><br />' + 
N_("If you still need to use v1, you may do so by either passing 'version=1' in the Accept Header or using api/v1/ in the URL.")

#evaluates as 
'<b>Foreman API v1 is deprecated. Please use v2.</b><br />If you still need to use v1, you may do so by either passing 'version=1' in the Accept Header or using api/v1/ in the URL.'

#and then _(docs) won't match because you only have the first and the second part in translations separately.
```
Please see:
https://github.com/grosser/fast_gettext/blob/master/lib/fast_gettext/translation.rb#L52-L55

ohadlevy:
@dLobatog my thought process was to add a warning to notification if we detect someone using a deprecate methods, including API v1. (no need to duplicate it on every api call, but rather one warning with a list of calls/users who used it.)

tstrachota:
@ohadlevy this could be a another improvement. This PR is trying to fix the case when someone comes to the apidoc, looks at v1 and doesn't notice there's v2 that should be used preferably.

ares:
I'm sorry, what exactly is the issue with HTML tags in translated strings? I understand they could be mistakenly broken during translations but we already use them in popovers on multiple places

* https://github.com/theforeman/foreman/blob/develop/app/views/roles/_form.html.erb#L19
* https://github.com/theforeman/foreman/blob/develop/app/views/filters/_form.html.erb#L47
* https://github.com/theforeman/foreman/blob/develop/app/views/nic/_base_form.html.erb#L14

I suppose in this case it does not matter that much. Notifications are definitely a good idea but we can merge this too so users get more visible warning before they even do the API call.

So if breaking strings on translations is the only concern, :+1: from me and if no one NACKs within a week I think we can merge it.

ares:
Thanks @tstrachota, merged as 9d9af9e8b5ffa8498fd36722bb9a91a429735884

PR #4191, opened 2017-01-15
Fixes #12547 - Added validators to interger fields

kgaikwad:
With this commit, instead of postgresql exception it will
show proper validation message for interger fields.


dLobatog:
Thanks @kgaikwad!

PS: For the record, I'm not asking for tests to display the validation error here as I see scoped_search already covers that quite a lot, so it's somewhat like testing Rails-provided validations

PR #4194, opened 2017-01-16
Fixes #17545 - UI notification backend

dLobatog:
- Initial data model for notification support, based on manageiq design.
- adds notification JSON endpoints.

To be tested with https://gist.github.com/6326e104b73906b273174b5dd80cf153
and https://github.com/theforeman/foreman/pull/3902/files

screenshot from 2017-01-16 16-14-59
screenshot from 2017-01-16 16-15-12

### DIFF
--- a/app/models/notification_blueprint.rb
+++ b/app/models/notification_blueprint.rb
@@ -1,0 +1,25 @@
+# Model to store all information related to a notification.
+# The audience for this notification is calculated on real time by
+# Notification.
+#
+# Foreman and plugins should define NotificationBlueprints for various
+# actions that are notification-worthy. This keeps the responsibilities
+# separate, Notifications taking care of serving the content, and
+# NotificationBlueprint of storing it.
+class NotificationBlueprint < ActiveRecord::Base
+  has_many :notifications
+  belongs_to :subject, :polymorphic => true
+
+  validates :message, :presence => true
+  validates :group, :presence => true
+  validates :name, :presence => true
+  validates :level, :inclusion => { :in => %w(success error warning info) }, :presence => true
+  validates :expires_in, :numericality => {:greater_than => 0}
+  before_validation :set_default_expiry
+
+  private
+
+  def set_default_expiry
+    self.expires_in ||= 24.hours
+  end
+end
--- a/db/migrate/20160927071039_create_notification_blueprints.rb
+++ b/db/migrate/20160927071039_create_notification_blueprints.rb
@@ -1,0 +1,13 @@
+class CreateNotificationBlueprints < ActiveRecord::Migration
+  def change
+    create_table :notification_blueprints do |t|
+      t.string :group, index: true
+      t.string :level
+      t.string :message
+      t.text :name
+      t.integer :expires_in
+      t.timestamps null: false
+      t.references :subject, polymorphic: true
+    end
+  end
+end
--- a/app/controllers/concerns/foreman/controller/parameters/notification_recipient.rb
+++ b/app/controllers/concerns/foreman/controller/parameters/notification_recipient.rb
@@ -1,0 +1,15 @@
+module Foreman::Controller::Parameters::NotificationRecipient
+  extend ActiveSupport::Concern
+
+  class_methods do
+    def notification_recipient_params_filter
+      Foreman::ParameterFilter.new(::NotificationRecipient).tap do |filter|
+        filter.permit :seen
+      end
+    end
+  end
+
+  def notification_recipient_params
+    self.class.notification_recipient_params_filter.filter_params(params, parameter_filter_context)
+  end
+end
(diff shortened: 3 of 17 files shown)


db/migrate/20160927071039_create_notification_blueprints.rb
```diff
@@ -0,0 +1,12 @@
+class CreateNotificationBlueprints < ActiveRecord::Migration
+  def change
+    create_table :notification_blueprints do |t|
+      t.string :group, index: true
+      t.string :level
```
ohadlevy, line 5: i would keep the name too.

app/controllers/concerns/foreman/controller/parameters/user_notification.rb
```diff
@@ -0,0 +1,15 @@
+module Foreman::Controller::Parameters::UserNotification
+  extend ActiveSupport::Concern
```
ohadlevy: why do we need it as a mixin?

db/migrate/20160927071039_create_notification_blueprints.rb
ohadlevy: maybe a better name for it would be label.

app/controllers/concerns/foreman/controller/parameters/user_notification.rb
dLobatog: Just tidiness, *all* controllers have their parameters extended through concerns in this directory. Additionally the plugin interface needs this to exist to add more parameters to the list

db/migrate/20160927071039_create_notification_blueprints.rb
dLobatog: What's name used for in http://www.patternfly.org/pattern-library/communication/notification-drawer/#/api?

ohadlevy:
its not, its just a way for us to distinguish one notification from another
for example:
```new_notification group: provisioning, name: 'finished with an error'```
vs
```new_notification :group: provisionining, message: <full message ...>```

**Comment by dLobatog** [comment]
Added `:name` to the notification blueprint class

**Comment by dLobatog** [comment]
Fixed the controller to use `find_resource` and added a minor change to the filter parameters concern so that it would work with the UI.

**Comment by dLobatog** [comment]
@rohoover Mind to take a look at the screenshots at the beginning of this PR? I think https://github.com/theforeman/foreman/pull/4170 (the UI part) follows the Patternfly spec http://www.patternfly.org/pattern-library/communication/notification-drawer/#/api pretty well.

However since we're not using full-width, it looks kinda strange. I'm all for moving to full-width in future PRs and unify the style with Patternfly-related projects and Katello, just wondering what you thought about the drawer in this state.

File: app/models/notification_blueprint.rb [hunk 3]
```diff
@@ -0,0 +1,23 @@
...
+# actions that are notification-worthy. This keeps the responsibilities
+# separate, Notifications taking care of serving the content, and
+# NotificationBlueprint of storing it.
+class NotificationBlueprint < ActiveRecord::Base
+  has_many :notifications
+  belongs_to :subject, :polymorphic => true
+
+  validates :message, :presence => true
+  validates :level, :inclusion => { :in => %w(success error warning info) }, :presence => true
+  validates :expires_in, :numericality => {:greater_than => 0}
```
ohadlevy, line 17: what about group and name validation?

db/migrate/20160927071039_create_notification_blueprints.rb
ohadlevy: its not, its just a way for us to distinguish one notification from another for example:
```new_notification group: provisioning, name: 'finished with an error'```
vs
```new_notification :group: provisionining, message: <full message ...>```

File: app/models/notification_blueprint.rb (hunk 3 above)
Comment (dLobatog) [reply app/models/notification_blueprint.rb:17]: I forgot to validate `group`, it seems that the patternfly UI doesn't allow notifications without group. I'll add it.

For `name` I can do so too, but if it's a field that's not going to be displayed in the UI and it isn't really used for anything that dev. convenience, why should we enforce having it?

File: app/models/notification_blueprint.rb [hunk 4]
```diff
@@ -0,0 +1,24 @@
...
+# Foreman and plugins should define NotificationBlueprints for various
+# actions that are notification-worthy. This keeps the responsibilities
+# separate, Notifications taking care of serving the content, and
+# NotificationBlueprint of storing it.
+class NotificationBlueprint < ActiveRecord::Base
+  has_many :notifications
+  belongs_to :subject, :polymorphic => true
+
+  validates :message, :presence => true
+  validates :group, :presence => true
```
ohadlevy, line 14: i would validate name, as you dont want to blueprints with the same name... as a developer you would need to find the blueprint based on its name..

dLobatog: Sounds fair, I guess I just didn't understand `:name` use in the first place.

dLobatog:
Latest update just changes the name field from NotificationBlueprint from text to string

dLobatog:
Gah - forgot to set the right file name to the notifications controller test, updated

PR #4195, opened 2017-01-16
fixes #18064 - import host from compute resource

timogoebel:
This commit allows a user to import an existing VM into Foreman.
The code tries to pre-fill host details automatically.

This is still WIP. Any comments are appreciated.

image

### DIFF
--- a/app/controllers/hosts_controller.rb
+++ b/app/controllers/hosts_controller.rb
@@ -81,7 +81,10 @@ class HostsController < ApplicationController
   end
 
   def new
-    @host = Host.new :managed => true
+    @host = Host.new(
+      :managed => true,
+      :build => true
+    )
   end
 
   # Clone the host
--- a/app/models/compute_resources/foreman/model/vmware.rb
+++ b/app/models/compute_resources/foreman/model/vmware.rb
@@ -9,6 +9,9 @@ module Foreman::Model
     validates :user, :password, :server, :datacenter, :presence => true
     before_create :update_public_key
 
+    alias_attribute :server, :url
+    alias_attribute :datacenter, :uuid
+
     def self.available?
       Fog::Compute.providers.include?(:vsphere)
     end
@@ -472,22 +475,6 @@ module Foreman::Model
       client.servers.get(client.vm_clone(opts)['new_vm']['id'])
     end
 
-    def server
-      url
-    end
-
-    def server=(value)
-      self.url = value
-    end
-
-    def datacenter
-      uuid
-    end
-
-    def datacenter=(value)
-      self.uuid = value
-    end
-
     def console(uuid)
       vm = find_vm_by_uuid(uuid)
       raise "VM is not running!" unless vm.ready?
@@ -524,6 +511,24 @@ module Foreman::Model
 
     def self.provider_friendly_name
       "VMware"
+    end
+
+    def vm_compute_attributes(vm)
+      vm_attrs = super
+      dc_networks = networks
+      interfaces = vm.interfaces || []
+      vm_attrs[:interfaces_attributes] = interfaces.each_with_index.inject({}) do |hsh, (interface, index)|
+        network = dc_networks.detect { |n| [n.id, n.name].include?(interface.network) }
+        raise Foreman::Exception.new(N_("Could not find network %s on VMWare compute resource"), interface.network) unless network
+        interface_attrs = {}
+        interface_attrs[:compute_attributes] = {}
+        interface_attrs[:mac] = interface.mac
+        interface_attrs[:compute_attributes][:network] = network.name
+        interface_attrs[:compute_attributes][:type] = interface.type.to_s.split('::').last
+        hsh[index.to_s] = interface_attrs
+        hsh
+      end
+      vm_attrs
     end
 
     private
--- a/app/models/concerns/fog_extensions/aws/server.rb
+++ b/app/models/concerns/fog_extensions/aws/server.rb
@@ -6,7 +6,7 @@ module FogExtensions
       attr_accessor :managed_ip
 
       def to_s
-        tags["Name"] || identity
+        tags.try([], 'Name') || identity
       end
 
       def name
--- a/app/models/concerns/orchestration/compute.rb
+++ b/app/models/concerns/orchestration/compute.rb
@@ -37,7 +37,8 @@ module Orchestration::Compute
 
   def queue_compute
     return unless compute? && errors.empty?
-    new_record? ? queue_compute_create : queue_compute_update
+    # Create a new VM if it doesn't already exist or update an existing vm
+    vm_exists? ? queue_compute_create : queue_compute_update
   end
 
   def queue_compute_create
@@ -52,7 +53,7 @@ module Orchestration::Compute
     queue.create(:name   => _("Set IP addresses for %s") % self, :priority => 5,
                  :action => [self, :setComputeIPAM]) if compute_provides?(:mac) && (mac_based_ipam?(:subnet) || mac_based_ipam?(:subnet6))
     queue.create(:name   => _("Power up compute instance %s") % self, :priority => 1000,
-                 :action => [self, :setComputePowerUp]) if compute_attributes[:start] == '1'
+                 :action => [self, :setComputePowerUp]) if compute_attributes && compute_attributes[:start] == '1'
   end
 
   def queue_compute_update
@@ -215,8 +216,9 @@ module Orchestration::Compute
 
   def compute_update_required?
     return false unless compute_resource.supports_update? && !compute_attributes.nil?
-    old.compute_attributes = compute_resource.vm_compute_attributes_for(uuid)
-    compute_resource.update_required?(old.compute_attributes, compute_attributes)
+    attrs = compute_resource.vm_compute_attributes_for(uuid)
+    old.compute_attributes = attrs if old
+    compute_resource.update_required?(attrs, compute_attributes.symbolize_keys)
   end
 
   def find_image
@@ -344,5 +346,10 @@ module Orchestration::Compute
       return false unless validate_required_foreman_attr(mac,Nic::Base.physical,:mac)
     end
     true
+  end
+
+  def vm_exists?
+    return true unless compute_object
+    !compute_object.persisted?
   end
 end
--- a/app/services/compute_resource_host_importer.rb
+++ b/app/services/compute_resource_host_importer.rb
@@ -1,0 +1,69 @@
+class ComputeResourceHostImporter
+  attr_accessor :host, :compute_resource, :vm
+
+  delegate :logger, :to => :Rails
+
+  def initialize(opts = {})
+    self.compute_resource = opts.fetch(:compute_resource)
+    self.vm = compute_resource.find_vm_by_uuid(opts[:uuid]) if opts[:uuid]
+    self.vm ||= opts.fetch(:vm)
+    self.host = Host::Managed.new(
+      :managed => true,
+      :build => false,
+      :compute_resource_id => compute_resource.id,
+      :vm => vm
+    )
+    copy_vm_attributes
+    parse_vm_name
+    initialize_interfaces
+  end
+
+  private
+
+  def copy_vm_attributes
+    compute_resource.provided_attributes.each do |foreman_attr, fog_attr|
+      value = vm.public_send(fog_attr)
+      host.send("#{foreman_attr}=", value)
+    end
+  end
+
+  def parse_vm_name
+    name, domain = vm_name.split('.', 2)
+    host.name = name
+    return unless Net::Validations.validate_hostname(domain)
+    host.domain = Domain.where(:name => domain).first_or_create
+  end
+
+  def initialize_interfaces
+    vm_interfaces = vm_interface_attributes
+
+    host.interfaces.select(&:physical?).each do |interface|
+      vm_interface = vm_interfaces.shift
+      return if vm_interface.nil?
+      set_nic_compute_attributes(interface, vm_interface)
+    end
+
+    vm_interfaces.each do |vm_interface|
+      host.interfaces << set_nic_compute_attributes(Nic::Managed.new, vm_interface)
+    end
+  end
+
+  def set_nic_compute_attributes(interface, vm_interface)
+    interface.assign_attributes(vm_interface)
+  end
+
+  def vm_interface_attributes
+    vm_attrs = compute_resource.vm_compute_attributes_for(host.uuid)
+    attr_name = compute_resource.interfaces_attrs_name
+    attr_key = "#{attr_name}_attributes"
+    vm_attrs.with_indifferent_access[attr_key].try(:values) || []
+  end
+
+  def vm_name
+    [
+      vm.try(:hostname),
+      vm.try(:name),
+      vm.try(:identity)
+    ].detect(&:present?)
+  end
+end
--- a/app/views/compute_resources_vms/form/vmware/_base.html.erb
+++ b/app/views/compute_resources_vms/form/vmware/_base.html.erb
@@ -1,34 +1,34 @@
 <%= javascript 'compute_resource' %>
 <%= javascript_tag("$(document).on('ContentLoad', tfm.numFields.initCounter)"); %>
-<%= text_f f, :name, :label => _('Name'), :disabled => !new_host if show_vm_name? %>
+<%= text_f f, :name, :label => _('Name'), :disabled => !new_vm if show_vm_name? %>
 <%= counter_f f, :cpus, :label => _('CPUs'), :label_size => 'col-md-2', :'data-soft-max' => compute_resource.max_cpu_count %>
 
 <%= counter_f f, :corespersocket, :label => _('Cores per socket'), :label_size => 'col-md-2', :'data-soft-max' => compute_resource.max_cpu_count, :value => f.object.corespersocket || 1 %>
 <%= text_f f, :memory_mb, :class => "col-md-2", :label => _("Memory (MB)") %>
 <%= field(f, :firmware, :label => _('Firmware'), :label_size => "col-md-2") do
   compute_resource.firmware_types.collect do |type, name|
-    radio_button_f f, :firmware, {:disabled => !new_host, :value => type, :text => _(name)}
+    radio_button_f f, :firmware, {:disabled => !new_vm, :value => type, :text => _(name)}
   end.join(' ').html_safe
 end %>
 <%= selectable_f f, :cluster, compute_resource.clusters, { :include_blank => _('Please select a cluster') },
-                    :class => "col-md-2", :disabled => !new_host,
+                    :class => "col-md-2", :disabled => !new_vm,
                     :label => _('Cluster'), :onchange => 'vsphereGetResourcePools(this)',
                     :help_inline => :indicator,
                     :data => {:url => resource_pools_compute_resource_path(compute_resource)} %>
-<%= vsphere_resource_pools(f, compute_resource, new_host) %>
-<%= select_f f, :path, compute_resource.folders, :path, :to_label , {}, { :label => _("Folder"), :class => "col-md-2", :disabled => !new_host } %>
-<%= select_f f, :guest_id, compute_resource.guest_types, :first, :last, {}, { :label => _("Guest OS"), :class => "col-md-2", :disabled => !new_host } %>
-<%= select_f f, :scsi_controller_type, compute_resource.scsi_controller_types, :first, :last, {}, { :label => _("SCSI controller"), :class => "col-md-2", :disabled => !new_host } %>
-<%= select_f f, :hardware_version, compute_resource.vm_hw_versions, :first, :last, {}, { :label => _("Virtual H/W version"), :class => "col-md-2", :disabled => !new_host } %>
-<%= checkbox_f f, :memoryHotAddEnabled, { :help_inline => _("Memory hot add lets you add memory resources for a virtual machine while the machine is powered on."), :label => _('Memory hot add'), :label_size => "col-md-2", :disabled => !new_host } %>
-<%= checkbox_f f, :cpuHotAddEnabled, { :help_inline => _("CPU hot add lets you add CPU resources for a virtual machine while the machine is powered on."), :label => _('CPU hot add'), :label_size => "col-md-2", :disabled => !new_host } %>
-<%= checkbox_f f, :add_cdrom, { :help_inline => _("Add a CD-ROM drive to the virtual machine."), :label => _('CD-ROM drive'), :label_size => "col-md-2" } if new_host %>
+<%= vsphere_resource_pools(f, compute_resource, !new_vm) %>
+<%= select_f f, :path, compute_resource.folders, :path, :to_label , {}, { :label => _("Folder"), :class => "col-md-2", :disabled => !new_vm } %>
+<%= select_f f, :guest_id, compute_resource.guest_types, :first, :last, {}, { :label => _("Guest OS"), :class => "col-md-2", :disabled => !new_vm } %>
+<%= select_f f, :scsi_controller_type, compute_resource.scsi_controller_types, :first, :last, {}, { :label => _("SCSI controller"), :class => "col-md-2", :disabled => !new_vm } %>
+<%= select_f f, :hardware_version, compute_resource.vm_hw_versions, :first, :last, {}, { :label => _("Virtual H/W version"), :class => "col-md-2", :disabled => !new_vm } %>
+<%= checkbox_f f, :memoryHotAddEnabled, { :help_inline => _("Memory hot add lets you add memory resources for a virtual machine while the machine is powered on."), :label => _('Memory hot add'), :label_size => "col-md-2", :disabled => !new_vm } %>
+<%= checkbox_f f, :cpuHotAddEnabled, { :help_inline => _("CPU hot add lets you add CPU resources for a virtual machine while the machine is powered on."), :label => _('CPU hot add'), :label_size => "col-md-2", :disabled => !new_vm } %>
+<%= checkbox_f f, :add_cdrom, { :help_inline => _("Add a CD-ROM drive to the virtual machine."), :label => _('CD-ROM drive'), :label_size => "col-md-2" } if new_vm %>
 
 <!--TODO # Move to a helper-->
 <% checked = params[:host] && params[:host][:compute_attributes] && params[:host][:compute_attributes][:start] || '1' %>
-<%= checkbox_f f, :start, { :checked => (checked == '1'), :help_inline => _("Power ON this machine"), :label => _('Start'), :label_size => "col-md-2" } if new_host and controller_name != 'compute_attributes' %>
+<%= checkbox_f f, :start, { :checked => (checked == '1'), :help_inline => _("Power ON this machine"), :label => _('Start'), :label_size => "col-md-2" } if new_vm && controller_name != 'compute_attributes' %>
 
-<%= textarea_f f, :annotation, :class => "col-md-8", :rows=> '3', :disabled => !new_host, :label => _("Annotation Notes") %>
+<%= textarea_f f, :annotation, :class => "col-md-8", :rows=> '3', :disabled => !new_vm, :label => _("Annotation Notes") %>
 
 <%
   arch ||= nil ; os ||= nil
--- a/app/views/compute_resources_vms/form/vmware/_network.html.erb
+++ b/app/views/compute_resources_vms/form/vmware/_network.html.erb
@@ -1,12 +1,12 @@
 <%= select_f f, :type, compute_resource.nictypes, :first, :last, { },
   :class => "col-md-3 vmware_type",
-  :label => _('NIC type'), :label_size => "col-md-3", :disabled => !new_host
+  :label => _('NIC type'), :label_size => "col-md-3", :disabled => !new_vm
 %>
-<% if new_host %>
+<% if new_vm %>
   <%= select_f f, :network, vsphere_networks(compute_resource), :first, :last, { },
     :class => "col-md-3 vmware_network",
     :label => _('Network'), :label_size => "col-md-3"
   %>
 <% else %>
-  <%= text_f f, :network, :class => 'col-md-3 vmware_network', :label => _("Network"), :label_size => "col-md-3", :disabled => true %>
+  <%= text_f f, :network, :class => 'col-md-3 vmware_network', :label => _("Network"), :label_size => "col-md-3", :readonly => true %>
 <% end %>
--- a/app/views/hosts/_form.html.erb
+++ b/app/views/hosts/_form.html.erb
@@ -51,7 +51,7 @@
 
         <%= select_f f, :compute_resource_id, accessible_resource(@host, :compute_resource), :id, :to_label,
           { :include_blank => _('Bare Metal') },
-          {:label => _('Deploy on'), :disabled => !@host.new_record?, :'data-url' => compute_resource_selected_hosts_path,
+          {:label => _('Deploy on'), :disabled => (!@host.new_record? || @host.uuid.present?), :'data-url' => compute_resource_selected_hosts_path,
             :onchange => 'computeResourceSelected(this);',
             :help_inline => :indicator } if SETTINGS[:unattended] && @host.new_record? || @host.compute_resource_id %>
 
@@ -85,6 +85,10 @@
       <%= render_tab_content_for(:main_tabs, :subject => @host) %>
 
       <%= f.hidden_field :managed %>
+      <% # Required for import and edit actions %>
+      <%= f.hidden_field :uuid if @host.uuid.present? %>
+      <% # Required for import action %>
+      <%= f.hidden_field :compute_resource_id if @host.new_record? && @host.uuid.present? %>
       <%= f.hidden_field :progress_report_id, :data=>{:url=>task_path(@host.progress_report_id)} %>
       <%= f.hidden_field :type, :value => @host.type if @host.type_changed? %>
 
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -83,7 +83,7 @@ Metrics/MethodLength:
 # Offense count: 22
 # Configuration parameters: CountComments.
 Metrics/ModuleLength:
-  Max: 420
+  Max: 424
 
 # Offense count: 9
 # Configuration parameters: CountKeywordArgs.
--- a/app/assets/javascripts/host_edit.js
+++ b/app/assets/javascripts/host_edit.js
@@ -124,15 +124,13 @@ function update_capabilities(capabilities){
 var stop_pooling;
 
 function submit_with_all_params(){
-  var url = window.location.pathname.replace(/\/edit$|\/new|\/\d+.*\/nest$/,'');
+  var url = $('form').attr('action');
   if (url.match('hostgroups')) {
     resource = 'hostgroup'
   } else {
     resource = 'host'
   }
-  resources = resource + 's';
   capitalized_resource = resource[0].toUpperCase + resource.slice(1);
-  if(/\/clone$/.test(window.location.pathname)){ url = foreman_url('/' + resources); }
   $('form input[type="submit"]').attr('disabled', true);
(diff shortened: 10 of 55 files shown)


timogoebel:
Added import links and tests for all compute resources.

app/controllers/hosts_controller.rb
```diff
@@ -90,6 +91,15 @@ def clone
     @host.valid?
   end
 
+  # Import a host from a compute resource
+  def import
```
ohadlevy: does it make sense to separate it to a ImportHostController ? i could see us implementing the create / destroy and maybe update rest verbs ( to add, remove and change a host to compute resource mapping). not to mention make the score of hosts controller better :)

app/models/concerns/orchestration/compute.rb
```diff
@@ -212,8 +212,9 @@ def delComputeUpdate
 
   def compute_update_required?
     return false unless compute_resource.supports_update? && !compute_attributes.nil?
-    old.compute_attributes = compute_resource.find_vm_by_uuid(uuid).attributes
-    compute_resource.update_required?(old.compute_attributes, compute_attributes.symbolize_keys)
+    attrs = compute_resource.find_vm_by_uuid(uuid).attributes
+    old.compute_attributes = attrs if old
```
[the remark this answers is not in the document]
timogoebel: @houndci-bot: This is only possible with Ruby > Ruby 2.3.0. If I recall correctly the Foreman project supports older rubies.

app/models/compute_resources/foreman/model/vmware.rb
```diff
@@ -8,6 +8,9 @@ class Vmware < ComputeResource
     validates :user, :password, :server, :datacenter, :presence => true
     before_create :update_public_key
 
+    alias_attribute :server, :url
+    alias_attribute :datacenter, :uuid
```
ohadlevy, line 13: does this works for setters? I would assume you would need to alias 'datacenter=' to 'uuid=' ?

app/models/concerns/fog_extensions/aws/server.rb
```diff
@@ -6,7 +6,7 @@ module Server
       attr_accessor :managed_ip
 
       def to_s
-        tags["Name"] || identity
+        tags.try([], 'Name') || identity
```
ohadlevy, line 9: why change it? isnt tags always an array?

app/models/compute_resources/foreman/model/vmware.rb
timogoebel: Quoted from the manual: `alias_attribute` allows you to make aliases for attributes, which includes getter, setter, and query methods.

app/models/concerns/fog_extensions/aws/server.rb
timogoebel: Tags may be nil, at least when using the mocked data.

app/models/concerns/orchestration/compute.rb
```diff
@@ -37,7 +37,7 @@ def vm_name
 
   def queue_compute
     return unless compute? && errors.empty?
-    new_record? ? queue_compute_create : queue_compute_update
+    new_vm? ? queue_compute_create : queue_compute_update
```
ohadlevy: this will map a new host in foreman to an exists compute resource host, while I like this option I wonder if people would want that by default? (or at least not configurable)?

app/services/compute_resource_host_importer.rb
```diff
@@ -0,0 +1,69 @@
+class ComputeResourceHostImporter
+  attr_accessor :host, :compute_resource, :vm
+
+  delegate :logger, :to => :Rails
+
+  def initialize(opts = {})
+    self.compute_resource = opts.fetch(:compute_resource)
+    self.vm = compute_resource.find_vm_by_uuid(opts[:uuid]) if opts[:uuid]
+    self.vm ||= opts.fetch(:vm)
```
ohadlevy, line 9: what if uuid or vm dont exists? I think safer code would be:
```ruby
if opts[:uuid]
  self.vm = compute_resource.find_vm_by_uuid(opts[:uuid])
elsif opts[:vm]
  self.vm = opts[:vm]
else
  raise..
end
```

app/services/compute_resource_host_importer.rb
```diff
@@ -0,0 +1,69 @@
...
+    compute_resource.provided_attributes.each do |foreman_attr, fog_attr|
+      value = vm.public_send(fog_attr)
+      host.send("#{foreman_attr}=", value)
+    end
+  end
+
+  def parse_vm_name
+    name, domain = vm_name.split('.', 2)
+    return unless Net::Validations.validate_hostname(name)
+    domain = Domain.find_by_name(domain)
```
ohadlevy: since we are importing the vm, should we create the domain if it doesn't exists (just like other fact importers)?

app/services/compute_resource_host_importer.rb
timogoebel: `find_vm_by_uuid` raises `ActiveRecord::NotFound` so we should be covered. Your suggestion is fine as well, though.

app/services/compute_resource_host_importer.rb
```diff
@@ -0,0 +1,69 @@
...
+    attr_key = "#{attr_name}_attributes"
+    vm_attrs.with_indifferent_access[attr_key].try(:values) || []
+  end
+
+  def vm_name
+    [
+      vm.try(:hostname),
+      vm.try(:name),
+      vm.try(:identity)
+    ].select(&:present?).first
```
ohadlevy: nitpick: use detect to stop on first match

app/models/concerns/orchestration/compute.rb
timogoebel: The idea behind this was the following:
For an API request: When a user passes `compute_resource_id` and `uuid` when creating a host, the user wants to associate an existing VM to Foreman.  So no new VM should be created.
For an UI request, `uuid` is always empty (and `new_vm?` false) except when using the import option.

app/services/compute_resource_host_importer.rb
timogoebel: 👍  - we should check if the domain is valid, though. I have seen VMs named like "abc.example.com Test ABC". We shouldn't create a domain "example.com Test ABC" then.

app/views/hosts/_form.html.erb
```diff
@@ -85,6 +85,8 @@
       <%= render_tab_content_for(:main_tabs, :subject => @host) %>
 
       <%= f.hidden_field :managed %>
+      <%= f.hidden_field :uuid %>
+      <%= f.hidden_field :compute_resource_id if @host.new_record? && @host.uuid.present? %>
```
ohadlevy: can you explain a bit more about how its used? I'm afraid its unclear whats needed for import vs new vs edit. (maybe code comment would be best).

app/models/concerns/orchestration/compute.rb
ohadlevy: code comment could help, or even better variable names?

timogoebel:
@ohadlevy : I moved the code to a dedicated controller. Was that, what you had in mind?

timogoebel:
@ohadlevy : Do you have a suggestion how we should handle permissions with the controller pattern you suggested? Permissions are tightly coupled with the controller name and I want to allow a user, that can create a host and view compure_resource_vms to be able to import a host.

app/models/concerns/orchestration/compute.rb
```diff
@@ -212,8 +213,9 @@ def delComputeUpdate
 
   def compute_update_required?
     return false unless compute_resource.supports_update? && !compute_attributes.nil?
-    old.compute_attributes = compute_resource.vm_compute_attributes_for(uuid)
-    compute_resource.update_required?(old.compute_attributes, compute_attributes)
+    attrs = compute_resource.vm_compute_attributes_for(uuid)
+    old.compute_attributes = attrs if old
```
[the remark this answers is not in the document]
timogoebel: @houndci-bot: Again, this is only possible with Ruby >= Ruby 2.3.0. If I recall correctly the Foreman project supports older rubies.

timogoebel:
Rebased and added permissions for the new controller.

timogoebel:
Rebased and fixed vmware CR tests.

app/controllers/host/vms_controller.rb
```diff
@@ -0,0 +1,28 @@
+class Host::VmsController < ApplicationController
+  before_action :find_compute_resource, :only => :new
+  before_action :find_vm, :only => :new
+
+  helper :hosts
+
+  # Import a host from a compute resource
+  def new
```
ohadlevy: shouldn't this be POST/create pr REST? or am I understanding this correctly, this doesnt really import, it rather precreate the attributes into an existing host form?
that would be different to other importers who actually import the host into foreman and not force the user to edit (create) the record?

also, how would the API look like in this case?

app/views/compute_resources_vms/form/vmware/_base.html.erb
```diff
@@ -1,41 +1,41 @@
...
-<%= textarea_f f, :annotation, :class => "col-md-8", :rows=> '3', :disabled => !new_host, :label => _("Annotation Notes") %>
+<%= textarea_f f, :annotation, :class => "col-md-8", :rows=> '3', :disabled => !new_vm, :label => _("Annotation Notes") %>
 
 <%
   arch ||= nil ; os ||= nil
   images = possible_images(compute_resource, arch, os)
 %>
 
-<div id='image_selection'>
+<div id="image_selection">
```
ohadlevy: nitpick: why?

app/views/compute_resources_vms/form/vmware/_network.html.erb
```diff
@@ -1,12 +1,12 @@
...
 %>
-<% if new_host %>
+<% if new_vm %>
   <%= select_f f, :network, vsphere_networks(compute_resource), :first, :last, { },
     :class => "col-md-3 vmware_network",
     :label => _('Network'), :label_size => "col-md-3"
   %>
 <% else %>
-  <%= text_f f, :network, :class => 'col-md-3 vmware_network', :label => _("Network"), :label_size => "col-md-3", :disabled => true %>
+  <%= text_f f, :network, :class => 'col-md-3 vmware_network', :label => _("Network"), :label_size => "col-md-3", :readonly => true %>
```
ohadlevy, line 11: why?

app/controllers/host/vms_controller.rb
timogoebel: I think, GET is correct here as it - as you already said - it just renders a host edit form with predefined values. The actual creation is happening like for every other host. In discovery, we do something similar.

The API call is just a simple create. The only difference is, that when you pass `uuid` and `compute_resource_id` Foreman doesn't create a new VM but links the existing one.

app/views/compute_resources_vms/form/vmware/_base.html.erb
timogoebel: Just personal preference. HTML with single ticks looks weird imho. I can change that back, though.

app/views/compute_resources_vms/form/vmware/_network.html.erb
timogoebel: When the field is disabled, the value doesn't get serialized and when the host is saved it doesn't save the `interfaces_attributes`.

app/controllers/host/vms_controller.rb
```diff
@@ -0,0 +1,28 @@
+class Host::VmsController < ApplicationController
```
dLobatog: I think you & @ohadlevy  may have missed `ComputeResourcesVmsController`. This controller already implements these `find` methods, and could have a `import` method that implements the `new` method I see in this controller.

app/helpers/application_helper.rb
```diff
@@ -161,10 +161,18 @@ def display(attribute)
     "display:#{attribute ? 'none' : 'inline'};"
   end
 
+  def controller_namespace
```
dLobatog: These changes might not be needed if the import action is implemented on `ComputeResourcesVmsController` and we can rely on the Rails provided `controller_name` method

app/controllers/host/vms_controller.rb
timogoebel: This is actually very interesting. There already is a `new` action `ComputeResourcesVmsController` - but there is no route for that. Weird...

We could just use that? What do you think, @dLobatog & @ohadlevy?

app/models/concerns/orchestration/compute.rb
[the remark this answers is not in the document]
dLobatog: Fixed by https://github.com/theforeman/foreman/commit/d89db8e23356166ef2c9441e5170be314bb18120

app/controllers/host/vms_controller.rb
dLobatog: I'm fine with that - `import` sounds more meaningful to me though. I played with that `new` route a while ago to write a plugin for more OpenStack stuff, it might be possible that plugins define their own routes to it.

app/services/compute_resource_host_importer.rb
```diff
@@ -0,0 +1,69 @@
...
+  attr_accessor :host, :compute_resource, :vm
+
+  delegate :logger, :to => :Rails
+
+  def initialize(opts = {})
+    self.compute_resource = opts.fetch(:compute_resource)
+    self.vm = compute_resource.find_vm_by_uuid(opts[:uuid]) if opts[:uuid]
+    self.vm ||= opts.fetch(:vm)
+    self.host = Host::Managed.new(
+      :managed => true,
```
dLobatog, line 11: When I import a host from hypervisor, I would expect the VM to be already built. Therefore I would expect `build => false` here. Otherwise the Host will be put in build mode after importing it (which risks deleting all data in the machine).

dLobatog (changes requested):
@timogoebel Please notice the katello test is breaking due to the changes on orchestration (http://ci.theforeman.org/job/test_develop_pr_katello/1739/database=postgresql,ruby=2.2,slave=fast/testReport/junit/(root)/Katello__UnifyHostsTest/test_unify_two_compute_resource/)

Another issue I found is that 'build' should be false by default in my opinion. Have you found any way of importing the Operating System too? It's a bit annoying to have to select the new host OS for an already existing host.

Overall great job :smile: ! Let me know what you think about the issues above and I'll be happy to give it a 2nd review and merge if all works well.

timogoebel:
@dLobatog : Before I fix your comments, what are we going to do regarding the controller thing?

I asked @ohadlevy on IRC about that:
> i didn't look in great details, but import sounds like an action of it self, vms controller is mostly around crud for the vm it self, where import controller care about associated / de-associating a host and a vm, hence my initial reaction is to separate it. 

I think, he has a valid point. I don't mind rewriting the code to yet another controller, but I want this to be the last rewrite ;-)

timogoebel:
@dLobatog : Addressed your comments. The controller is still unchanged, though.

Sadly I didn't come up with a way to match a compute resource's OS with Foreman's OS model, that's why this isn't done automatically. If you have an idea how we could do this, I'm all ears.

app/services/compute_resource_host_importer.rb
```diff
@@ -0,0 +1,70 @@
...
+  def copy_vm_attributes
+    compute_resource.provided_attributes.each do |foreman_attr, fog_attr|
+      value = vm.public_send(fog_attr)
+      host.send("#{foreman_attr}=", value)
+    end
+  end
+
+  def parse_vm_name
+    name, domain = vm_name.split('.', 2)
+    return unless Net::Validations.validate_hostname(name)
```
dLobatog: Shouldn't this run on the Host object? I faked an invalid name and domain and this didn't show any error on the UI, just reloaded the page

app/services/compute_resource_host_importer.rb
dLobatog: Fixed

app/services/compute_resource_host_importer.rb
dLobatog: Raising an exception here might also work to display the issue

dLobatog (changes requested):
Found a minor issue on the hostname validation on the importer service, everything else worked like a charm. Thanks @timogoebel !

Regarding the controller, @ohadlevy have you seen the find methods in both controllers? They are identical, which IMO is a clear signal they operate for the same model. ComputeResourceVmsController always seemed to me like the responsible for all VM-related actions in a compute resource. In any case the code is mostly on the service (good call @timogoebel). It's not a major issue to have another controller but I think it complicates things a bit more than they should (duplicated code, more files to manage the same object - ComputeResource.vms).

app/services/compute_resource_host_importer.rb
timogoebel: You're totally right. I guess, I was being a little overcautious here. I'll remove the line and then the host object simply gets invalid when an invalid hostname is chosen.

timogoebel:
@dLobatog : Changed the controller to `ComputeResourcesVmsController`.

```diff
diff --git a/app/controllers/compute_resources_vms_controller.rb b/app/controllers/compute_resources_vms_controller.rb
index d74b1b9..b20febc 100644
--- a/app/controllers/compute_resources_vms_controller.rb
+++ b/app/controllers/compute_resources_vms_controller.rb
@@ -1,7 +1,18 @@
 class ComputeResourcesVmsController < ApplicationController
   include Foreman::Controller::ComputeResourcesCommon
+  include ::Foreman::Controller::ActionPermissionDsl
+
+  before_action :find_compute_resource
+  before_action :find_vm, :only => [:import, :associate, :show, :console, :pause, :power]
+
+  helper :hosts
+
+  def controller_permission
+    return :compute_resources if params[:action] == 'associate'
+    super
+  end
+
   def index
-    @compute_resource = find_compute_resource(:view_compute_resources_vms)
     load_vms
     @authorizer = Authorizer.new(User.current, :collection => [@compute_resource])
     respond_to do |format|
@@ -20,13 +31,10 @@ class ComputeResourcesVmsController < ApplicationController
   end

   def new
-    @compute_resource = find_compute_resource
     @vm = @compute_resource.new_vm
   end

   def create
-    @compute_resource = find_compute_resource
-
     if (vm = @compute_resource.create_vm params[:vm])
       @compute_resource.start_vm(vm.identity) if params[:vm][:start]=='1'
       process_success :success_redirect => compute_resource_vms_path(@compute_resource)
@@ -36,24 +44,20 @@ class ComputeResourcesVmsController < ApplicationController
   end

   def associate
-    @compute_resource = find_compute_resource(:edit_compute_resources)
-    @vm = find_vm
     if Host.for_vm(@compute_resource, @vm).any?
       process_error(:error_msg => _("VM already associated with a host"), :redirect => compute_resource_vm_path(:compute_resource_id => params[:compute_resource_id], :id => @vm.identity))
+      return
+    end
+    host = @compute_resource.associated_host(@vm) if @compute_resource.respond_to?(:associated_host)
+    if host.present?
+      host.associate!(@compute_resource, @vm)
+      process_success(:success_msg => _("VM associated to host %s") % host.name, :success_redirect => host_path(host))
     else
-      host = @compute_resource.associated_host(@vm) if @compute_resource.respond_to?(:associated_host)
-      if host.present?
-        host.associate!(@compute_resource, @vm)
-        process_success(:success_msg => _("VM associated to host %s") % host.name, :success_redirect => host_path(host))
-      else
-        process_error(:error_msg => _("No host found to associate this VM with"), :redirect => compute_resource_vm_path(:compute_resource_id => params[:compute_resource_id], :id => @vm.identity))
-      end
+      process_error(:error_msg => _("No host found to associate this VM with"), :redirect => compute_resource_vm_path(:compute_resource_id => params[:compute_resource_id], :id => @vm.identity))
     end
   end

   def show
-    @compute_resource = find_compute_resource(:view_compute_resources_vms)
-    @vm = find_vm
     respond_to do |format|
       format.html
       format.json { render :json => @vm }
@@ -61,19 +65,14 @@ class ComputeResourcesVmsController < ApplicationController
   end

   def power
-    @compute_resource = find_compute_resource(:power_compute_resources_vms)
-    @vm = find_vm
     run_vm_action(@vm.ready? ? :stop : :start)
   end

   def pause
-    @compute_resource = find_compute_resource(:power_compute_resources_vms)
-    @vm = find_vm
     run_vm_action(@vm.ready? ? :pause : :start)
   end

   def destroy
-    @compute_resource = find_compute_resource(:destroy_compute_resources_vms)
     if @compute_resource.destroy_vm params[:id]
       process_success({ :success_redirect => compute_resource_vms_path(@compute_resource), :success_msg => _('The virtual machine is being deleted.') })
     else
@@ -82,8 +81,6 @@ class ComputeResourcesVmsController < ApplicationController
   end

   def console
-    @compute_resource = find_compute_resource(:console_compute_resources_vms)
-    @vm = find_vm
     @console = @compute_resource.console @vm.identity
     @encrypt = Setting[:websockets_encrypt]
     render case @console[:type]
@@ -98,14 +95,26 @@ class ComputeResourcesVmsController < ApplicationController
     process_error :redirect => compute_resource_vm_path(@compute_resource, @vm.identity), :error_msg => (_("Failed to set console: %s") % e), :object => @vm
   end

+  def import
+    @host = ComputeResourceHostImporter.new(
+      :compute_resource => @compute_resource,
+      :vm => @vm
+    ).host
+  end
+
   private

-  def find_compute_resource(permission = :view_compute_resources)
-    ComputeResource.authorized(permission).find(params[:compute_resource_id])
+  define_action_permission 'console', :console
+  define_action_permission ['pause', 'power'], :power
+  define_action_permission 'import', :view
+  define_action_permission 'edit', :associate # edit_compute_resources
+
+  def find_compute_resource
+    @compute_resource = ComputeResource.authorized(current_permission).find(params[:compute_resource_id])
   end

   def find_vm
-    @compute_resource.find_vm_by_uuid params[:id]
+    @vm = @compute_resource.find_vm_by_uuid(params[:id])
   end

   def run_vm_action(action)
diff --git a/app/controllers/host/vms_controller.rb b/app/controllers/host/vms_controller.rb
deleted file mode 100644
index 2cbeb54..0000000
--- a/app/controllers/host/vms_controller.rb
+++ /dev/null
@@ -1,28 +0,0 @@
-class Host::VmsController < ApplicationController
-  before_action :find_compute_resource, :only => :new
-  before_action :find_vm, :only => :new
-
-  helper :hosts
-
-  # Import a host from a compute resource
-  def new
-    @host = ComputeResourceHostImporter.new(
-      :compute_resource => @compute_resource,
-      :vm => @vm
-    ).host
-  end
-
-  def controller_permission
-    'host'
-  end
-
-  private
-
-  def find_compute_resource
-    @compute_resource = ComputeResource.authorized(:edit_compute_resources).find(params[:compute_resource_id])
-  end
-
-  def find_vm
-    @vm = @compute_resource.find_vm_by_uuid(params[:uuid])
-  end
-end
diff --git a/app/helpers/application_helper.rb b/app/helpers/application_helper.rb
index 7b7c86b..3eedd39 100644
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -161,18 +161,10 @@ module ApplicationHelper
     "display:#{attribute ? 'none' : 'inline'};"
   end

-  def controller_namespace
-    controller_path.split('/').first
-  end
-
-  def controller_scope
-    controller_namespace.pluralize
-  end
-
   # return our current model instance type based on the current controller
   # i.e. HostsController would return "host"
   def type
-    controller_namespace.singularize
+    controller_name.singularize
   end

   def checked_icon(condition)
@@ -211,7 +203,7 @@ module ApplicationHelper
   end

   def method_path(method)
-    send("#{method}_#{controller_scope}_path")
+    send("#{method}_#{controller_name}_path")
   end

   def edit_textfield(object, property, options = {})
diff --git a/app/helpers/compute_resources_vms_helper.rb b/app/helpers/compute_resources_vms_helper.rb
index 840328b..235306d 100644
--- a/app/helpers/compute_resources_vms_helper.rb
+++ b/app/helpers/compute_resources_vms_helper.rb
@@ -153,7 +153,7 @@ module ComputeResourcesVmsHelper
   end

   def show_vm_name?
-    controller_scope != 'hosts' && controller_name != 'compute_attributes'
+    controller_name != 'hosts' && controller_name != 'compute_attributes'
   end

   def vsphere_resource_pools(form, compute_resource, disabled = false)
@@ -206,6 +206,6 @@ module ComputeResourcesVmsHelper

   def vm_import_action(vm)
     return unless Host.for_vm(@compute_resource, vm).empty?
-    display_link_if_authorized(_("Import"), hash_for_new_host_vms_path(:compute_resource_id => @compute_resource, :uuid => vm.identity))
+    display_link_if_authorized(_("Import"), hash_for_import_compute_resource_vm_path(:compute_resource_id => @compute_resource, :id => vm.identity))
   end
 end
diff --git a/app/helpers/hosts_helper.rb b/app/helpers/hosts_helper.rb
index 837872f..cecc04f 100644
--- a/app/helpers/hosts_helper.rb
+++ b/app/helpers/hosts_helper.rb
@@ -482,4 +482,9 @@ module HostsHelper
   def power_status_visible?
     SETTINGS[:unattended] && Setting[:host_power_status]
   end
+
+  def host_or_hostgroup_path(method)
+    controller = controller_name == 'hostgroups' ? 'hostgroups' : 'hosts'
+    send("#{method}_#{controller}_path")
+  end
 end
diff --git a/app/services/authorizer.rb b/app/services/authorizer.rb
index 1c3e844..eb1dde2 100644
--- a/app/services/authorizer.rb
+++ b/app/services/authorizer.rb
@@ -25,7 +25,7 @@ class Authorizer

   def find_collection(resource_class, options = {})
     permission = options.delete :permission
-    Foreman::Logging.logger('permissions').debug "checking permission #{permission}"
+    Foreman::Logging.logger('permissions').debug "checking permission #{permission} for class #{resource_class}"

     # retrieve all filters relevant to this permission for the user
     base = user.filters.joins(:permissions).where(["#{Permission.table_name}.resource_type = ?", resource_name(resource_class)])
diff --git a/app/services/foreman/access_permissions.rb b/app/services/foreman/access_permissions.rb
index 6b18b8d..7d53673 100644
--- a/app/services/foreman/access_permissions.rb
+++ b/app/services/foreman/access_permissions.rb
@@ -340,7 +340,7 @@ Foreman::AccessControl.map do |permission_set|
                                      :organizations =>  [:mismatches]
                                   }
     map.permission :create_hosts,  {:hosts => [:new, :create, :clone].push(*ajax_actions),
-                                    :"host/vms" => [:new],
+                                    :compute_resources_vms => [:import],
                                     :compute_resources => cr_ajax_actions,
                                     :puppetclasses => pc_ajax_actions,
                                     :subnets => subnets_ajax_actions,
diff --git a/app/views/common/os_selection/_architecture.html.erb b/app/views/common/os_selection/_architecture.html.erb
index aded7fe..e2091ca 100644
--- a/app/views/common/os_selection/_architecture.html.erb
+++ b/app/views/common/os_selection/_architecture.html.erb
@@ -4,6 +4,6 @@
     { :label => _("Operating system"),
       :disabled => arch_oss.empty? ? true : false,
       :help_inline => :indicator,
-      :onchange => 'os_selected(this);', :'data-url' => method_path('os_selected'), :'data-type' => controller_name.singularize, :required => true}
+      :onchange => 'os_selected(this);', :'data-url' => host_or_hostgroup_path('os_selected'), :'data-type' => controller_name.singularize, :required => true}
   %>
 <% end %>
diff --git a/app/views/common/os_selection/_operatingsystem.html.erb b/app/views/common/os_selection/_operatingsystem.html.erb
index ddef94f..18613a5 100644
--- a/app/views/common/os_selection/_operatingsystem.html.erb
+++ b/app/views/common/os_selection/_operatingsystem.html.erb
@@ -2,7 +2,7 @@
   <%= select_f f, :medium_id, os_media, :id, :to_label,
     {:include_blank => blank_or_inherit_f(f, :medium), :selected => item.medium_id},
     {:label => _("Media"), :disabled => os_media.empty?,
-     :help_inline => :indicator, :onchange => 'medium_selected(this);', :'data-url' =>  method_path('medium_selected'),
+     :help_inline => :indicator, :onchange => 'medium_selected(this);', :'data-url' => host_or_hostgroup_path('medium_selected'),
      :'data-type' => controller_name.singularize, :required => true }
   %>
   <%= select_f f, :ptable_id, os_ptable, :id, :to_label,
diff --git a/app/views/host/vms/new.html.erb b/app/views/compute_resources_vms/import.html.erb
similarity index 100%
rename from app/views/host/vms/new.html.erb
rename to app/views/compute_resources_vms/import.html.erb
diff --git a/app/views/hosts/_interfaces.html.erb b/app/views/hosts/_interfaces.html.erb
index a8eea25..5e15f31 100644
--- a/app/views/hosts/_interfaces.html.erb
+++ b/app/views/hosts/_interfaces.html.erb
@@ -47,7 +47,7 @@
   <!-- hidden interface forms -->
   <div id="interfaceForms">
     <%= f.fields_for :interfaces do |interface| %>
-      <%= render :partial => '/' + interface.object.to_partial_path, :locals => {:f => interface, :id => interface.object.object_id}, :layout => 'nic/hidden_layout' %>
+      <%= render :partial => interface.object, :locals => {:f => interface, :id => interface.object.object_id}, :layout => 'nic/hidden_layout' %>
     <% end %>
   </div>
   <!-- modal window -->
@@ -68,6 +68,6 @@
     </div>
   </div>

-  <%= new_child_fields_template(f, :interfaces, {:partial => '/' + Nic::Managed.new.to_partial_path, :layout => 'nic/hidden_layout'}) %>
+  <%= new_child_fields_template(f, :interfaces, {:partial => Nic::Managed.new, :layout => 'nic/hidden_layout'}) %>

 </fieldset>
diff --git a/app/views/hosts/_operating_system.html.erb b/app/views/hosts/_operating_system.html.erb
index fe140e6..3307bde 100644
--- a/app/views/hosts/_operating_system.html.erb
+++ b/app/views/hosts/_operating_system.html.erb
@@ -1,6 +1,6 @@

 <%= select_f f, :architecture_id, accessible_resource(f.object, :architecture), :id, :to_label, {:include_blank => true},
-    {:onchange => 'architecture_selected(this);', :'data-url' => method_path('architecture_selected'), :'data-type' => controller_name.singularize,
+    {:onchange => 'architecture_selected(this);', :'data-url' => host_or_hostgroup_path('architecture_selected'), :'data-type' => controller_name.singularize,
     :help_inline => :indicator, :required => true} %>

 <span id="os_select">
diff --git a/config/routes.rb b/config/routes.rb
index 974fd95..7bf90c0 100644
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -16,10 +16,6 @@ Foreman::Application.routes.draw do
     end
   end

-  namespace :host do
-    resource :vms, only: :new
-  end
-
   get '(:controller)/help', :action => 'welcome', :as => "help"
   constraints(:id => /[^\/]+/) do
     resources :hosts do
@@ -412,6 +408,7 @@ Foreman::Application.routes.draw do
               put 'pause'
               put 'associate'
               get 'console'
+              get 'import'
             end
           end
         end
diff --git a/test/controllers/compute_resources_vms_controller_test.rb b/test/controllers/compute_resources_vms_controller_test.rb
index 0607a2e..95f1044 100644
--- a/test/controllers/compute_resources_vms_controller_test.rb
+++ b/test/controllers/compute_resources_vms_controller_test.rb
@@ -7,8 +7,9 @@ class ComputeResourcesVmsControllerTest < ActionController::TestCase
     get_test_vm
   end

-  def setup_user(operation, type = 'compute_resources_vms')
-    super(operation, type, "id = #{@compute_resource.id}")
+  def setup_user(operation, type = 'compute_resources_vms', search = nil)
+    search ||= "id = #{@compute_resource.id}"
+    super(operation, type, search)
   end

   test "should not get index when not permitted" do
@@ -176,6 +177,35 @@ class ComputeResourcesVmsControllerTest < ActionController::TestCase
     assert_redirected_to @request.env['HTTP_REFERER']
   end

+  context '#import' do
+    setup { Fog.mock! }
+    teardown { Fog.unmock! }
+
+    let(:compute_resource) { FactoryGirl.create(:compute_resource, :vmware, :uuid => 'Solutions') }
+
+    test 'imports a host' do
+      setup_user('create', 'hosts', '')
+      setup_user('view', 'compute_resources_vms', "id = #{compute_resource.id}")
+      get :import, {:compute_resource_id => compute_resource.id, :id => '5032c8a5-9c5e-ba7a-3804-832a03e16381'}, set_session_user
+      assert assigns(:host)
+      assert_template 'compute_resources_vms/import'
+    end
+
+    test 'should not import when not permitted to view compute_resource' do
+      setup_user('none')
+      get :import, {:compute_resource_id => compute_resource.id, :id => '5032c8a5-9c5e-ba7a-3804-832a03e16381'}, set_session_user
+      assert_response :forbidden
+    end
+
+    test 'should not import when not permitted to create hosts' do
+      setup_user('view', 'compute_resources_vms', "id = #{compute_resource.id}")
+      get :import, {:compute_resource_id => compute_resource.id, :id => '5032c8a5-9c5e-ba7a-3804-832a03e16381'}, set_session_user
+      assert_response :forbidden
+    end
+  end
+
+  private
+
   def get_test_vm
     @test_vm = @compute_resource.vms.find { |vm| vm.name == 'test' }
   end
diff --git a/test/controllers/host/vms_controller_test.rb b/test/controllers/host/vms_controller_test.rb
deleted file mode 100644
index 8186880..0000000
--- a/test/controllers/host/vms_controller_test.rb
+++ /dev/null
@@ -1,16 +0,0 @@
-require 'test_helper'
-
-class Host::VmsControllerTest < ActionController::TestCase
-  context '#new' do
-    setup { Fog.mock! }
-    teardown { Fog.unmock! }
-
-    let(:compute_resource) { FactoryGirl.create(:compute_resource, :vmware, :uuid => 'Solutions') }
-
-    test 'imports a host' do
-      get :new, {:compute_resource_id => compute_resource.id, :uuid => '5032c8a5-9c5e-ba7a-3804-832a03e16381'}, set_session_user
-      assert assigns(:host)
-      assert_template 'host/vms/new'
-    end
-  end
-end
```

app/services/compute_resource_host_importer.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  def initialize(opts = {})
+    self.compute_resource = opts.fetch(:compute_resource)
+    self.vm = compute_resource.find_vm_by_uuid(opts[:uuid]) if opts[:uuid]
+    self.vm ||= opts.fetch(:vm)
+    self.host = Host::Managed.new(
+      :managed => true,
+      :build => false,
+      :compute_resource_id => compute_resource.id,
+      :vm => vm
```
dLobatog, line 14: Today I was trying to import a virtual machine, for a Host that already existed. I expected the import code would associate the host and the VM, however it attempted to create a new one (and failed because they had the same fqdn). Can you check that? I think the only required thing would be to change this `new` call to

`.where(....properties....).first_or_initialize`

dLobatog (changes requested):
@timogoebel I left a comment about importing VMs for existing hosts, let me know if you think it makes sense to do what I say there. I tried it locally and it worked.

dLobatog (changes requested):
@timogoebel I left a comment about importing VMs for existing hosts, let me know if you think it makes sense to do what I say there. I tried it locally and it worked.

timogoebel:
@dLobatog: I'm curious how you actually managed to import an already associated host. If a host exists for the compute resource and the uuid, the "import" button is not shown in the UI. I also added a test to demonstrate, that it doesn't work via API.
I don't think the import action should make any changes to existing hosts. We could maybe properly merge the attributes to not overwrite any existing values and just add information, but that would be even more code and I'd prefer to do this in a separate PR.

Rebased.

timogoebel:
Rebased.

Anything more to do to get this in line for 1.15?

timogoebel:
Rebased and fixed a test that needed org and loc set.

dLobatog:
@timogoebel Will merge if tests pass, I couldn't reproduce the import of an existing VM, maybe it was an old branch.

dLobatog (approved):

timogoebel:
@dLobatog : Thanks for the review. Tests are 🍏 .

dLobatog:
Thanks @timogoebel & @ohadlevy !

PR #4198, opened 2017-01-17
Fixes #18099 - Update the close button style in two-pane pages

junezhang:
In the two-pane pages, the color of close button in the right pane is a bit light, not easy to see.
So based my proposal on the close button in this link https://groups.google.com/forum/#!msg/foreman-dev/jy2Ytas4SRw/s8yos26AEAAJ.
I updated the close button to PatternFly style.

To check the screenshots:
close-button


dLobatog:
Thanks @junezhang

PR #4200, opened 2017-01-17
fixes #18102 - a host cannot be backed by the same vm

timogoebel:


dLobatog:
@timogoebel Seems to break tests, sorry. The change makes sense to me 😄 would be glad to merge after the tests are back up. Thanks!

app/models/concerns/orchestration/compute.rb
```diff
@@ -11,7 +11,7 @@ module Orchestration::Compute
   end
 
   def compute?
-    compute_resource_id.present? && (compute_attributes.present? || uuid.present?)
+    compute_resource_id.present? && compute_resource.present? && (compute_attributes.present? || uuid.present?)
```
ohadlevy: not sure i follow exactly the logic here?

app/models/host/managed.rb
```diff
@@ -196,6 +196,7 @@ class Jail < ::Safemode::Jail
     validate :short_name_periods
     before_validation :set_compute_attributes, :on => :create, :if => Proc.new { compute_attributes_empty? }
     validate :check_if_provision_method_changed, :on => :update, :if => Proc.new { |host| host.managed }
+    validates :uuid, uniqueness: { :scope => :compute_resource_id }
```
ohadlevy: its a good question if it should be unique across vs a specific compute resource (e.g. if someone moved a vm? or created a duplicate compute resource with different creds for example)?

db/migrate/20170117110442_add_unique_constraint_to_host_uuid.rb
```diff
@@ -0,0 +1,9 @@
+class AddUniqueConstraintToHostUuid < ActiveRecord::Migration
+  def up
+    add_index(:hosts, [:uuid, :compute_resource_id], :unique => true)
```
ohadlevy: if there are any failures, this would fail, should we check it / notify it / skip it ?

app/models/concerns/orchestration/compute.rb
timogoebel: The shoulda matchers test inserts an arbitrary id in the `compute_resource_id` column. This leads to the condition, that `compute_resource_id` is present but `compute_resource` is not. I guess, `compute_resource_id.present?` is there instead of `compute_resource.present?` because it's faster. I thought, we could keep it but add the more specific check to catch the described errors.

app/models/host/managed.rb
timogoebel: I'm undecided here. Both make sense. We actually do have several compute resources backed by the same VMWare infrastructure.

db/migrate/20170117110442_add_unique_constraint_to_host_uuid.rb
timogoebel: Do you mean show an explanation in the migration?

ohadlevy: TBH I'm not sure if we should add the index without validating / fixing the duplicated data first.

timogoebel:
Removed the db migration and the scope on a compute resource, fixed the tests.

ohadlevy:
thanks @timogoebel

PR #4203, opened 2017-01-18
fixes #18130 - add autoincrement for bigint id columns

timogoebel:
MySQL lost the auto_increment flag when the id column was changed to bigint. Postgres uses a sequence so there was no issue. Sqlite was not changed at all.

@ohadlevy : Thanks again for reporting this. Could you try if this fixes the issue for you, please?

### DIFF
--- a/db/migrate/20170118142654_add_auto_increment_to_bigint_ids.rb
+++ b/db/migrate/20170118142654_add_auto_increment_to_bigint_ids.rb
@@ -1,0 +1,16 @@
+class AddAutoIncrementToBigintIds < ActiveRecord::Migration
+  def self.up
+    if ['mysql', 'mysql2'].include? ActiveRecord::Base.connection.instance_values['config'][:adapter]
+      change_column :logs, :id, 'SERIAL'
+      change_column :reports, :id, 'SERIAL'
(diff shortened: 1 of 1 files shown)


ohadlevy:
@timogoebel this fixes the issue for me - thanks! ACK pending tests.

ohadlevy:
I assume the reason we didnt see this issue is because nightly is not published so E2E tests are not fully running? /cc @domcleal ?

domcleal:
This has nothing to do with packaging tests, it's an issue in Foreman that should be covered by Foreman's tests.

timogoebel:
@domcleal : Do you have a suggestion how we could test this properly? I was under the impression that at least one of the existing tests should have failed on mysql because the auto_increment flag got lost. Apparently I was wrong.

ohadlevy:
@timogoebel I'm going to merge in favor of fixing the current issue, but I would very much would like to figure out why our current testing did not find this issue.

@domcleal regarding the question, I was under the impression that our pipepline will only execute the End to End testing scenarios when packages are being created (prior to publishing), hence we didn't see any jenkins job that fail?

ohadlevy:
thanks @timogoebel !

PR #4204, opened 2017-01-18
fixes #18136 - add index for reports type column

timogoebel:
This increases performance when using plugins like openscap or omaha that do add an reports STI class.


dLobatog:
Thanks @timogoebel

PR #4205, opened 2017-01-18
Fixes #18143 - enable variables for snippets

ares:
snippet "ahoy" definition 
a
template definition
b
template output
c


dLobatog:
Thanks @ares !

ares:
oh no, I just pushed a newer version that also adds that to snippet_if_exists too, I'll open separate PR

PR #4206, opened 2017-01-19
Refs #18143 - extend snippet_if_exists too

ares:
Sorry, I realized `snippet_if_exists` needs it as well but too late. See #4205 for more details.

### DIFF
--- a/lib/foreman/renderer.rb
+++ b/lib/foreman/renderer.rb
@@ -86,11 +86,11 @@ module Foreman
     end
 
     # provide embedded snippets support as simple erb templates
-    def snippets(file)
+    def snippets(file, options = {})
       if Template.where(:name => file, :snippet => true).empty?
         render :partial => "unattended/snippets/#{file}"
       else
-        return snippet(file.gsub(/^_/, ""))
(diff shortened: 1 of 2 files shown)


ares:
Once it's merged, I'll update the wiki too.

ares:
And there's possibly one more - `snippets` so please don't merge yet.

ares:
I updated snippets too, I didn't find a way to pass variables the same way into file based rendering context. I haven't seen any use of it in community templates so I hope that's fine. Honestly I don't know why this macro is useful and whether we should have it in the first place. It looks like some left-over.

dLobatog:
Thanks @ares - about `if_exists`, I think it was useful when we didn't have locked templates so that if people removed a snippet, we'd still be able to render the template. It can still be useful for plugin snippets. E.g: a finish template from community_templates includes the REX and Salt snippets *only* if the plugin exists, and if I uninstall them, the finish template still renders fine

ares:
@dLobatog I understand the `if_exists` and agree it's useful. I don't understand `snippets` - note the **s** at the end. We have three -`snippet`, `snippet_if_exists` and `snippets`

PR #4207, opened 2017-01-19
Refs #17653 - add show_hidden to apidoc

orrabin:
To fix: http://projects.theforeman.org/issues/17652 there is no need for changes in hammer, the only thing that needs to change is the api doc and it was not added in the original commit for 17653: https://github.com/theforeman/foreman/commit/9643a2cb1e67f4f5fb75e47e83729fa9c168fdb8

### DIFF
--- a/app/controllers/api/v2/domains_controller.rb
+++ b/app/controllers/api/v2/domains_controller.rb
@@ -33,6 +33,7 @@ module Api
 
       api :GET, "/domains/:id/", N_("Show a domain")
       param :id, :identifier, :required => true, :desc => N_("Numerical ID or domain name")
+      param :show_hidden_parameter, :bool, :desc => N_("Should unhide hidden parameters values for display")
 
       def show
       end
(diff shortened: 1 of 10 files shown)


app/controllers/api/v2/domains_controller.rb
```diff
@@ -33,6 +33,7 @@ def index
 
       api :GET, "/domains/:id/", N_("Show a domain")
       param :id, :identifier, :required => true, :desc => N_("Numerical ID or domain name")
+      param :show_hidden_parameter, :bool, :desc => N_("Should unhide hidden parameters values for display")
```
dLobatog, line 36: I think the wording is wrong, both the "should" and the "unhide", I'd say "Display hidden parameters", "Reveal hidden parameters" or "Expose hidden parameters" convey the right meaning better.

Mind @domcleal @GregSutcliffe or some other native speaker to check it out? 

@orrabin Since tests pass already and the issue is just with the strings, I'll likely change it on merge as it's just a string.

dLobatog (changes requested):

app/controllers/api/v2/domains_controller.rb
domcleal: > I think the wording is wrong, both the "should" and the "unhide", I'd say "Display hidden parameters"

Indeed, or "Display hidden parameter values" to indicate it's the value and not the record being shown.

dLobatog:
Merged as 3b9d7933dc1f0599744080f0053858b8d7114d01, thanks @orrabin and @domcleal for the feedback!

PR #4208, opened 2017-01-19
fixes #18155 - add string to empty form-encoded arrays in test

domcleal:
This test attempts to send an empty `:smart_proxy_ids => []` parameter
emulating the UI to unset smart_proxy_ids on a taxonomy, but Rails 5
filters this empty parameter out when converting it internally to a form
encoded parameter (since it's not possible to specify an empty array in
a form) for the purposes of the test.

Comparing the test to the real UI shows that it really submits
`:smart_proxy_ids => [""]` to pass an empty array, so update the test
data to match.


dLobatog:
Thanks @domcleal !

PR #4209, opened 2017-01-19
fixes #18157 - use AS::Duration division to get days from expiry

domcleal:
ActiveSupport::Duration#parts was an unreliable API as it contains data
used internally when constructing durations.


dLobatog:
Thanks @domcleal

PR #4211, opened 2017-01-20
Fixes #18175 - Optionaly disable preview for templates

xprazak2:
If a template is not allowed for previewing, its clones will not be allowed as well.

### DIFF
--- a/app/controllers/concerns/foreman/controller/parameters/template.rb
+++ b/app/controllers/concerns/foreman/controller/parameters/template.rb
@@ -8,6 +8,7 @@ module Foreman::Controller::Parameters::Template
         :locked,
         :name,
         :snippet,
+        :preview_enabled,
         :template,
         :template_kind, :template_kind_id, :template_kind_name,
         :vendor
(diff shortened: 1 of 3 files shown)


xprazak2:
[test foreman]

ares:
Tests errors are unrelated but let's first try to test whether alternative fix in bootdisk would help. Maybe we could fix it by using `extend_template_helpers`

dLobatog:
@xprazak2 Thanks for the PR, I'm struggling to understand which templates shouldn't be previewed?

xprazak2:
@dLobatog preview for foreman_bootdisk tempalates causes errors and I was told preview for those templates is not relevant anyway. There may be an alternative approach that does not need this, I will close if it gets is.

dLobatog:
@xprazak2 Thanks! I just saw https://github.com/theforeman/foreman/pull/4211 , seems more sensible to me too. I'll close for the time being, feel free to reopen in case the bootdisk PR doesn't get anywhere, but it seems small enough and it works on my setup 👍

lzap:
What this was replaced with? You linked the same page, @dLobatog :-)

ares:
https://github.com/theforeman/foreman_bootdisk/pull/37 I guess

PR #4212, opened 2017-01-20
fixes #18176 - replace clear_association_cache with reload

domcleal:


dLobatog:
@domcleal Thanks! Please check out the failing tests 😄

domcleal:
The failing NIC BMC test has been fixed with an additional `reload` call. The old `clear_association_cache` was clearing unrelated subnet to smart proxy associations which the setup of the test was indirectly relying on as it modified proxy features.

dLobatog:
Thanks @domcleal

PR #4216, opened 2017-01-23
Fixes #18205 - Add more useful macros

ares:


tbrisker:
:-1: to adding more host macros on the renderer, let's create a proxy object that maintains the existing interface and allows extending it.

ares:
ok, so the first concern with host macros was that we'd force users to upgrade which is definitely not the case here, now the issue is that we **must** first add proxy objects before we can move somewhere? I might try to find some motivation after all of this to look at proxy objects during some weekend, but before I spend any more time on this, would you at least accept `save_to_file` macro which is not host specific?

tbrisker:
@ares yes, that makes sense. I will also try looking into making a proxy object this week.

ares:
Repushed the single macro and one fix test name conflict (unrelated to the macro). Once merge I'll update the template writing wiki. @tbrisker please update your thumb :-)

tbrisker:
Thumb updated :+1:

tbrisker:
Thanks @ares !

PR #4218, opened 2017-01-24
fixes #18208 - replace audited-activerecord with audited

domcleal:


shlomizadok:
:+1: ack

dLobatog:
Thanks @domcleal !

PR #4221, opened 2017-01-24
fixes #18214 - accept `params` as AC::Parameters or Hash

domcleal:
In Rails 5, ActionController::Parameters no longer inherits from Hash
and so type checks for params.is_a?(Hash) now return false and should
accept either. (Rails commit 14a3bd5.)

### DIFF
--- a/app/controllers/api/v2/hosts_controller.rb
+++ b/app/controllers/api/v2/hosts_controller.rb
@@ -304,7 +304,7 @@ Return the host's compute attributes that can be used to create a clone of this 
         params = params.deep_clone
         if params[:interfaces_attributes]
           # handle both hash and array styles of nested attributes
-          if params[:interfaces_attributes].is_a? Hash
+          if params[:interfaces_attributes].is_a?(Hash) || params[:interfaces_attributes].is_a?(ActionController::Parameters)
             params[:interfaces_attributes] = params[:interfaces_attributes].values
           end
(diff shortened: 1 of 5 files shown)


app/controllers/api/v2/hosts_controller.rb
```diff
@@ -304,7 +304,7 @@ def host_attributes(params, host = nil)
         params = params.deep_clone
         if params[:interfaces_attributes]
           # handle both hash and array styles of nested attributes
-          if params[:interfaces_attributes].is_a? Hash
+          if params[:interfaces_attributes].is_a?(Hash) || params[:interfaces_attributes].is_a?(ActionController::Parameters)
```
ohadlevy, line 307: does it make sense to extract to common mixin?

domcleal: Extract what exactly?

ohadlevy: sorry, the test condition - something like

```ruby
module Mixin
  def method(params)
    params.is_a?(Hash) || params.is_a?(ActionController::Parameters)
  end
end
```

domcleal: No, it doesn't make much sense to me, it's checked in very different places.

ohadlevy:
thanks @domcleal

PR #4223, opened 2017-01-24
fixes #18191 - notification should not keep initiating new pull requests when there is an active request

gailsteiger:


webpack/assets/javascripts/react_app/stores/NotificationsStore.js
```diff
@@ -5,6 +5,7 @@ import moment from 'moment';
 
 let _notifications = {};
 let _expandedTab = null;
+let _waitingForResponse = false;
```
ohadlevy: why  not use the PENDING constant we already have?

ohadlevy:
@gailsteiger initial review looks good, would you mind adding a test? I could think of multiple scenarios (e.g. server is down, timeout is reached, or just not to have multiple requests at the same time)

gailsteiger:
@ohadlevy after these changes I think this PR also fixes
Bug #18192 notification should not keep fetching events when timeout occured

webpack/assets/javascripts/react_app/actions/NotificationActions.js
```diff
@@ -1,18 +1,27 @@
...
-import {ACTIONS} from '../constants';
+import NotificationsStore from '../stores/NotificationsStore';
+import { ACTIONS, STATUS } from '../constants';
 
 const TIMER = 10000;
 
 export default {
   getNotifications(url) {
-    if (document.visibilityState === 'visible') {
+    if (document.visibilityState !== 'hidden' &&
```
ohadlevy: whats the difference between the two? (hidden vs visible)

ohadlevy:
> @ohadlevy after these changes I think this PR also fixes
> Bug #18192 notification should not keep fetching events when timeout occured

Nope:
notification-pending

ohadlevy:
@gailsteiger I might be missing something, but it looks like it just doesnt work at all..?

webpack/assets/javascripts/react_app/actions/NotificationActions.js
gailsteiger: The Document.visibilityState read-only property returns the visibility of the document, that is in which context this element is now visible. It is useful to know if the document is in the background or an invisible tab, or only loaded for pre-rendering. Possible values are:

- 'visible' : the page content may be at least partially visible. In practice this means that the page is the foreground tab of a non-minimized window.
- 'hidden' : the page content is not visible to the user. In practice this means that the document is either a background tab or part of a minimized window, or the OS screen lock is active.
- 'prerender' : the page content is being prerendered and is not visible to the user (considered hidden for purposes of document.hidden). The document may start in this state, but will never transition to it from another value. Note: browser support is optional.
- 'unloaded' : the page is being unloaded from memory. Note: browser support is optional.

I made the change only because of the test since visibilityState is read-only and had a value of 'prerender' in my test. Alternatively, It is possible to use:

```js
(document.visibilityState === 'visible'  || document.visibilityState === 'prerender')
```

gailsteiger:
@ohadlevy nicer catch! please check again

ohadlevy:
@gailsteiger what did you change?

gailsteiger:
During refactor I had changed the action name from
```js
ACTIONS.NOTIFICATIONS_SET_WAITING_FOR_RESPONSE
```
to
```js
ACTIONS.NOTIFICATIONS_SET_REQUEST_STATUS
```
but failed to pick up the change In NotificationsStore, where

```js
    case ACTIONS.NOTIFICATIONS_SET_WAITING_FOR_RESPONSE: {
      _requestStatus = action.status;
      break;
    }    
```
should have been 

```js
case ACTIONS.NOTIFICATIONS_SET_REQUEST_STATUS: {
      _requestStatus = action.status;
      break;
    }
```

ohadlevy:
sorry, I've closed it by mistake.

ohadlevy:
@gailsteiger it looks better now, however, when 401 is reached, I get a JS console error
```GET http://localhost:3000/notification_recipients 401 (Unauthorized)```
It looks like the API get method does not handle the .error callback? can you please have a look?

**Comment by gailsteiger** [comment]
@ohadlevy I set a breakpoint in the error handler which does handle the 401 error. The console message was produced by jQuery.

**Comment by ohadlevy** [comment]
@gailsteiger last comment / question, in http://www.patternfly.org/pattern-library/communication/notification-drawer/#/_design point 9. there is an example for infinite scroll and loading state, should we consider adding the loading state to the UI or completely ignore it UI wise?

File: webpack/assets/javascripts/react_app/components/notifications/NotificationDrawerTitle.js [hunk 3]
```diff
@@ -2,9 +2,13 @@ import React from 'react';
 
 const NotificationDrawerTitle = ({text}) => (
   <div className="drawer-pf-title">
-    <a className="drawer-pf-toggle-expand"></a>
```
ohadlevy, line 5: this probably best to do in a separate pr, as there is a redmine issue for it already.

ohadlevy: or change the commit message to include all relevant issues that are fixed.

webpack/assets/javascripts/react_app/stores/NotificationsStore.js
```diff
@@ -1,10 +1,11 @@
 import AppDispatcher from '../dispatcher';
-import { ACTIONS } from '../constants';
+import { ACTIONS, STATUS } from '../constants';
 import AppEventEmitter from './AppEventEmitter';
 import moment from 'moment';
 
 let _notifications = {};
 let _expandedTab = null;
```
ohadlevy, line 7: I assume this is not required if you are removing the expand functionality?

webpack/assets/javascripts/react_app/components/notifications/NotificationDrawerTitle.js
gailsteiger: OK, I'll revert

webpack/assets/javascripts/react_app/stores/NotificationsStore.js
gailsteiger: No. _expandedTab here refers to which of the accordion tabs (group) in the drawer is expanded

gailsteiger:
@ohadlevy I would ignore loading state in the UI.

webpack/assets/javascripts/react_app/actions/NotificationActions.js
ohadlevy: unloaded means we navigate to another page right? maybe we should also check if turbolinks is not active (e.g. I clicked on a really slow page to load that is going to refresh the entire page, no point in calling notifications as we'll do it again for the new page?)

gailsteiger: @ohadlevy 
I favor changing the 'condition' to 
```js
(document.visibilityState === 'visible'  || document.visibilityState === 'prerender')
```
and so prevent the polling request when page is being unloaded. Does this take care of the turbolinks issue you mentioned?

ohadlevy: @gailsteiger i dont think so - see https://github.com/turbolinks/turbolinks#full-list-of-events

ohadlevy:
[test foreman]

webpack/assets/javascripts/react_app/actions/NotificationActions.js
ohadlevy: @gailsteiger do you still want to change it regardless?

gailsteiger: I made the change and pushed

ohadlevy:
merged as 8ed57bb with minor whitespace changes, thanks @gailsteiger

PR #4224, opened 2017-01-25
Fixes #12294 - Update existing host params using API

xprazak2:


xprazak2:
Parameters for hostgroup, domain and taxonomies may be affected as well, investigating.

orrabin:
For the record: there needs to be a decision about whether the parameters will be updated or replaced.
Most places in the api seem to replace when getting a list of association (for example sending organizations to hostgroup will remove all of the hostgroup's organizations and connect only those sent in the list).
Also, if the parameters are updating should they include ids for existing parameters?

mbacovsky:
Update vs. replace - I believe that API v2 was originally designed to "replace" when updating multi-value attributes. Changing that behavior would violate v2 compatibility and bring in inconsistency in the concept.

I can see the "update" can be more natural from user perspective so can we add a new attribute e.g. `host_parameter_attributes_update` that would "update" instead of "replace"?

xprazak2:
I made changes to make this work for hostgroup, domain, subnet, os and taxonomies as well. It does update instead of replace - if a param with given name is found, the value gets updated, otherwise it is created. The outcome is the same as if the old ones were deleted and created with same name but new values. I will change into replace so we do not have inconsistencies.

xprazak2:
@orrabin you do not really need ids for parameters when you update because their names are unique in reference scope. I thought not using ids was intentional, to be more user-friendly.

test/controllers/api/v2/subnets_controller_test.rb
```diff
@@ -153,4 +153,13 @@ class Api::V2::SubnetsControllerTest < ActionController::TestCase
       assert_equal 'bar', show_response['parameters'].first['value']
     end
   end
+
+  test "should update existing subnet parameters" do
```
dLobatog, line 157: This test is failing

dLobatog (changes requested):
@xprazak2 Other than the failing test, the implementation looks quite neat, thanks! It looks ready to merge after fixing the test unless @orrabin @mbacovsky think otherwise

xprazak2:
Adding and removing parameters should be now handled, I added tests to cover both cases. All will hopefully turn green now.

xprazak2:
[test foreman]

xprazak2:
All tests are passing now, codeclimate raises similar code issue for apipie

tstrachota:
Fixes the issue for me. :+1: to merge.
```
>> hammer host update --id 2 --parameters 'param1=val1'
Host updated
>> hammer host update --id 2 --parameters 'param1=val2'
Host updated
>> hammer host info --id 2   
...
Parameters:               
    param1 => val2
...
```

dLobatog (approved):

dLobatog:
Thanks @xprazak2 !

app/controllers/concerns/parameter_attributes.rb
```diff
@@ -0,0 +1,73 @@
...
+  def process_parameter_attributes
+    model_params = params[parameter_method_mapping]
+    return unless model_params
+    parameter_params = model_params["#{parameter_class_mapping}_parameters_attributes"]
+    return unless parameter_params
+
+    param_names = parameter_params.map { |hash| hash[:name] }
+
+    process_update obj_names, param_names, parameter_params, param_klass
+    process_destroy obj_names, param_names, parameter_params, param_klass
```
junovitch, line 13: @xprazak2 I'm curious your thoughts on this in regards to the http://projects.theforeman.org/issues/20725 regression. Is there a good way to fix this so this line of code doesn't delete all parameters?

xprazak2: Maybe it could be removed altogether, so update would not remove any existing parameters and could be uses only for update. You would then need to use `hammer host delete-parameter` to remove the parameters.

PR #4226, opened 2017-01-25
fixes #18233, #18235 - isolate pagelets state between tests

domcleal:
Moves pagelets state into instances of Pagelets::Manager with clearer
dup semantics, and allows the instance to be backed up and restored
around tests using the PageletsIsolation test case mixin.


domcleal:
The integration test failure appears unrelated, filed as http://projects.theforeman.org/issues/18239.

timogoebel (approved):
@domcleal : Merging, thanks a lot for making sure our ci stays reliable.

PR #4228, opened 2017-01-25
Fixes #18234 - display correct icon when no notifications

lizagilman:


tbrisker:
ok to test

ohadlevy:
@gailsteiger can you please review? - thanks.

webpack/assets/javascripts/react_app/components/notifications/NotificationDrawerToggle.test.js
```diff
@@ -3,34 +3,56 @@ jest.unmock('../../stores/NotificationsStore');
...
 import React from 'react';
-import { shallow } from 'enzyme';
+import { mount, shallow } from 'enzyme';
 import NotificationDrawerToggle from './NotificationDrawerToggle';
 import NotificationsStore from '../../stores/NotificationsStore';
 import NotificationActions from '../../actions/NotificationActions';
 
 function setup() {
-  return shallow(<NotificationDrawerToggle />);
+    return shallow(<NotificationDrawerToggle />);
```
ohadlevy: nitpick: please keep the 2 spaces indentation.

webpack/assets/javascripts/react_app/stores/NotificationsStore.js
```diff
@@ -15,6 +15,10 @@ class NotificationsEventEmitter extends AppEventEmitter {
     return (_notifications.data || []);
```
ohadlevy: nitpick: should we return empty hash vs empty array on line 15?

webpack/assets/javascripts/react_app/stores/NotificationsStore.test.js
```diff
@@ -25,4 +25,10 @@ describe('NotificationsStore', () => {
     expect(result.info[0].id).toBe(18);
     expect(result.info[1].id).toBe(9);
   });
+
+  it('', () => {
+    const result = NotificationsStore.prepareNotifications([]);
+
```
ohadlevy: what does this test do?

gailsteiger:
This is lovely work and thanks for the tests.

However, the [Notification Drawer design](http://www.patternfly.org/pattern-library/communication/notification-drawer/#/_design) calls for the empty icon to be shown when there are no unread notifications, not when there are no notifications at all. (See Note 7)

ohadlevy:
@gailsteiger good catch!

tbrisker:
@gailsteiger I would argue that when there are no notifications at all there are also no unread notifications. I don't think it makes sense to have a full icon when there are no notifications.

ohadlevy:
@tbrisker i read it as we need to check that there are no unread notifications, having no notifications also qualify of course.

gailsteiger:
@tbrisker @ohadlevy There is no contradiction. Empty icon should appear when there are no unread notifications, including when there are no notifications but also when all notifications are read

That was the original intention of my comment.

tbrisker:
@gailsteiger sorry, I misunderstood your previous comment.
@lizagilman - could you please update this to show the correct icon when there are read notifications as well?

webpack/assets/javascripts/react_app/stores/NotificationsStore.js
```diff
@@ -12,7 +12,11 @@ class NotificationsEventEmitter extends AppEventEmitter {
   }
 
   getNotifications() {
-    return (_notifications.data || []);
+    return (_notifications.data || {});
+  }
+
+  getNotificationsCount() {
```
ohadlevy: probably not needed?

webpack/assets/javascripts/react_app/stores/ToastNotificationsStore.js
```diff
@@ -0,0 +1,107 @@
+import AppDispatcher from '../dispatcher';
```
ohadlevy: I guess this file is not part of this PR?

webpack/assets/javascripts/react_app/components/notifications/NotificationDrawerToggle.js
```diff
@@ -21,7 +21,7 @@ class NotificationDrawerToggle extends Component {
     switch (actionType) {
       case ACTIONS.RECEIVED_NOTIFICATIONS: {
         this.setState({
-          count: NotificationsStore.getNotifications().length,
+          unreadNotificationsCount: NotificationsStore.hasUnreadNotifications(),
```
gailsteiger: calling NotificationsStore.hasUnreadNotifications without arguments
method signature is `hasUnreadNotifications(notifications)`
note also that this only checks unread notifications when notifications are received. Consider the case when there is one unread notification and it is manually marked as read. Then there are no unread notifications but state won't reflect that until next polling request

webpack/assets/javascripts/react_app/stores/NotificationsStore.test.js
```diff
@@ -18,11 +18,29 @@ describe('NotificationsStore', () => {
...
+    expect(result).toBeFalsy();
+  });
+
+  it('returns true when there are unread notifications', () => {
+    const result = NotificationsStore.hasUnreadNotifications(NotificationsDataSeen);
+
+    expect(result).toBeTruthy();
+  });
 });
```
gailsteiger: Both test data sets contain unread notifications, notification where property seen is false
method is not returning correct result

webpack/assets/javascripts/react_app/stores/NotificationsStore.js
```diff
@@ -12,9 +12,13 @@ class NotificationsEventEmitter extends AppEventEmitter {
   }
 
   getNotifications() {
-    return (_notifications.data || []);
+    return (_notifications.data || {});
   }
 
+  hasUnreadNotifications(notifications) {
```
ohadlevy: shouldnt the function be:
```js
hasUnreadNotifications() {
  return !!(_notifications.find(notification => notification.seen));
}
```

webpack/assets/javascripts/react_app/components/notifications/NotificationDrawerToggle.js
ohadlevy: I'm a bit puzzled how the tests results did not reflect the missing argument on this function?

ohadlevy: @gailsteiger any suggestions/examples on how its best to test updates to component based on store events?

webpack/assets/javascripts/react_app/stores/NotificationsStore.js
```diff
@@ -12,9 +12,13 @@ class NotificationsEventEmitter extends AppEventEmitter {
...
+    return !!(notifications.find(notification => notification.seen));
+
+  }
```
gailsteiger: function does not return correct result
(a notification is unread if it's seen property is false)

I suggest checking `_notifications` (which is an object and does not support `find`) instead of calling the method with an argument

webpack/assets/javascripts/react_app/components/notifications/NotificationDrawerToggle.js
ohadlevy: @gailsteiger wont mark as read trigger an update event which will cause the unread counter to be updated anyway?

gailsteiger: for the polling use case perhaps it would be best to keep a running count of unread notifications, reset when notifications are received and adjusted when notification is marked as read. The store data for the read notification should then also be updated (my bad)

webpack/assets/javascripts/react_app/stores/NotificationsStore.js
gailsteiger: that function I think returns true if any of the notifications has its seen property set to true

webpack/assets/javascripts/react_app/components/notifications/NotificationDrawerToggle.js
lizagilman: @ohadlevy @gailsteiger 
will go over everything in this PR and fix according to your comments!

gailsteiger: @lizagilman great. Have a look at NotificationPanel.js Consider extracting the calculation of the unread notifications to a helper or services file and reuse for your purposes.

```js
    let unread;

    if (notifications) {
      unread = notifications.reduce((total, curr) => {
        if (!curr.seen) {
          total = total + 1;
        }
        return total;
      }, 0);
    } else {
      unread = 0;
    }
```

webpack/assets/javascripts/react_app/components/notifications/NotificationDrawerToggle.test.js
```diff
@@ -18,6 +18,14 @@ describe('NotificationDrawerToggle', () => {
...
   });
 
+  it('should show bell icon', () => {
+    const wrapper = setup();
+    const emptyBellIcon = wrapper.find('.fa-bell-o');
+    const fullBellIcon = wrapper.find('.fa-bell');
+
+    expect(emptyBellIcon.length || fullBellIcon.length).toBe(1);
+  });
+
```
lizagilman, line 28: overall tests will be followed when finished moving from flux to redux

ohadlevy:
@matanwerbner @gailsteiger can you please review? thanks

gailsteiger:
@lizagilman I added some unread messages to the database and css class is still `fa-bell-o`

webpack/assets/javascripts/react_app/components/notifications/NotificationDrawerToggle.js
```diff
@@ -25,6 +26,8 @@ class NotificationDrawerToggle extends Component {
       case ACTIONS.RECEIVED_NOTIFICATIONS: {
         this.setState({
           count: NotificationsStore.getNotifications().length,
+          unreadCount: _.filter(NotificationsStore.getNotifications(),
+            notification => notification.seen).length,
           isLoaded: true
         });
         break;
```
gailsteiger: I've found that it is best to make the `setState` call as simple as possible. I would remove the function calls from the `setState` call, do all the calculations before and pass a simple object literal.

Also, don't call `NotificationsStore.getNotifications()` twice. Cache the result of the first call.

`NotificationsStore.getNotifications()` returns an object of arrays of notifications. You have to do your calculations on the underlying arrays.

matanwerbner: I think this should be !seen...we both missed it lol

gailsteiger: All the more reason not to do the calculations in `setState`. Kinda hard to set a break point.

ohadlevy: anyway we can add a test for this ? :)

matanwerbner: i dont think there is a way to test flux connected components.. @gailsteiger  do you know a way to mock the store?

lizagilman:
@gailsteiger can you please review again? thanks.

webpack/assets/javascripts/react_app/components/notifications/NotificationDrawerToggle.js
```diff
@@ -23,10 +24,18 @@ class NotificationDrawerToggle extends Component {
   onChange(actionType) {
     switch (actionType) {
       case ACTIONS.RECEIVED_NOTIFICATIONS: {
-        this.setState({
-          count: NotificationsStore.getNotifications().length,
-          isLoaded: true
-        });
+        const notifications = NotificationsStore.getNotifications();
+
+        const newState = _.reduce(notifications, (curr, next) => {
```
ohadlevy: maybe i'm missing something, but why do we care about the current count ? just set the new state regardless ?

lizagilman: we might need it for later use @matanwerbner ?

matanwerbner: I think that If we should need it in the future, we will implement. Currently, if its not needed, it should be removed

matanwerbner: All this component needs is a bool called "hasUnreadMessages" imho 😃

gailsteiger: I think it is important to write some tests that will check for various scenarios, including a scenario where there is more than 1 array of notifications.

matanwerbner: @gailsteiger i agree, we're a bit lost on testing UI components with flux, do you have any idea how to do this?
in redux it would much simpler as the store is injected into the component, so different data can be mocked in the store.
in flux currently nothing is injected so there is much less control.

lizagilman:
@ohadlevy @gailsteiger can you please review again?

ohadlevy:
thanks @lizagilman

PR #4229, opened 2017-01-25
fixes #18239 - wait for AJAX spinners removal on about page test

domcleal:
`wait_for_ajax` alone may return after page load before AJAX requests
have even begun, so also wait for the spinners to be replaced.

---

Recreate (hopefully) the same assertion failure locally by disabling the proxy_status/about JS assets on the page, which shows the new assertion will wait.


dLobatog:
I was able to recreate the failure only on sqlite for some reason - but it makes sense and could not reproduce with this.

dLobatog:
Thanks @domcleal !

PR #4233, opened 2017-01-26
fixes #18252 - added docker* to ignore

thomasmckay:
Ignore docker* by default in ignored_interface_identifiers


dLobatog:
Thanks, good addition @thomasmckay 👍

PR #4234, opened 2017-01-26
Fixes #18260 - Display proper organizations count on table

dLobatog:
If the user is scoped to taxonomy "A", "Host.count" will only return
hosts in that taxonomy. That is correct. However there is a usecase
where we want to display the hosts count regardless of the current
taxonomy.

The table for locations and organizations was showing the "Host.count"
scoped to the current taxonomy, so it would show 0 everywhere but in the
current taxonomy.

The fix is to make the helper aware of this issue, and show the unscoped
Host.count when displaying the organizations/locations table

"Spot the difference"

screenshot from 2017-01-26 16-23-36
screenshot from 2017-01-26 16-24-04

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -433,7 +433,13 @@ module ApplicationHelper
   end
 
   def hosts_count(resource_name = controller.resource_name)
-    @hosts_count ||= Host::Managed.reorder('').authorized.group("#{resource_name}_id").count
+    # If we are on /organizations or /locations, this allows to display the
+    # count for hosts not in the current organization & location.
+    hosts_scope = Host::Managed.reorder('')
+    if ['organization', 'location'].include? resource_name
+      hosts_scope = hosts_scope.unscoped
+    end
+    @hosts_count ||= hosts_scope.authorized.group("#{resource_name}_id").count
   end
 
   def webpack_dev_server


app/helpers/application_helper.rb
```diff
@@ -433,7 +433,13 @@ def hidden_spinner(text = '', options = {})
   end
 
   def hosts_count(resource_name = controller.resource_name)
-    @hosts_count ||= Host::Managed.reorder('').authorized.group("#{resource_name}_id").count
+    # If we are on /organizations or /locations, this allows to display the
+    # count for hosts not in the current organization & location.
+    if ['organization', 'location'].include? resource_name
+      hosts_scope = Host::Managed.reorder('').unscoped
```
tbrisker: This would mean the user may know about hosts he is not authorized to see.

tbrisker (changes requested):
See inline

app/helpers/application_helper.rb
dLobatog: No? This would only display the count, once you click on the number to be directed to the search, it will not display you anything you cannot see.

dLobatog:
@tbrisker Replied inline, I don't think this opens up anything for users - just the host count. I'm OK with hiding the number if that's preferred, but showing the count doesn't mean the user will see any of these hosts.

app/helpers/application_helper.rb
tbrisker: Oh, actually I was wrong, there is another `.authorized` below. However there is some repetition here, how about 
```ruby
hosts_scope = Host::Managed.reorder('')
hosts_scope = hosts_scope.unscoped if ['organization', 'location'].include? resource_name
@hosts_count ||= hosts_scope.authorized.group("#{resource_name}_id").count
```

tbrisker (changes requested):

dLobatog:
@tbrisker Updated according to your changes 😄

tbrisker (approved):

PR #4236, opened 2017-01-27
refs #17526 - reset taxonomy.current after each test

domcleal:
The shared "ignores the taxable_type if current taxonomy ignores it"
test failed to reset the taxonomy in teardown, instead using the setup
routine. This causes failures in different test cases when the test was
last to run.


dLobatog:
Thanks @domcleal  - I've seen these failures on several PRs but didn't find a good way to reproduce, not even using the same 'seed'.

PR #4238, opened 2017-01-27
fixes #13618 - cache expensive vmware api calls

timogoebel:
In my dev environment the `compute_resource_selected` call got approx. 10 times faster. I'm really thrilled!

The caching can be enabled for every compute resource. The code is reusable to plugins or other compute resources can reuse the caching mechanism.
Cache invalidation was not straight forward to implement because I wanted to support memcached so I had to use a "scope" when creating the cache keys. The scope is changed when the cache is invalidated.

Some "benchmarks":

Before:
Completed 200 OK in 16121ms (Views: 15963.0ms | ActiveRecord: 6.4ms)

After:
Completed 200 OK in 1710ms (Views: 1583.8ms | ActiveRecord: 5.1ms)

### DIFF
--- a/app/controllers/api/v2/compute_resources_controller.rb
+++ b/app/controllers/api/v2/compute_resources_controller.rb
@@ -10,7 +10,7 @@ module Api
       before_action :find_resource, :only => [:show, :update, :destroy, :available_images, :associate,
                                               :available_clusters, :available_flavors, :available_folders,
                                               :available_networks, :available_resource_pools, :available_security_groups, :available_storage_domains,
-                                              :available_zones, :available_storage_pods]
+                                              :available_zones, :available_storage_pods, :refresh_cache]
 
       api :GET, "/compute_resources/", N_("List all compute resources")
       param_group :taxonomy_scope, ::Api::V2::BaseController
@@ -41,6 +41,7 @@ module Api
           param :server, String, :desc => N_("for VMware")
           param :set_console_password, :bool, :desc => N_("for Libvirt and VMware only")
           param :display_type, %w(VNC SPICE), :desc => N_('for Libvirt only')
+          param :caching_enabled, :bool, :desc => N_('enable caching, for VMware only')
           param_group :taxonomies, ::Api::V2::BaseController
         end
       end
@@ -162,11 +163,21 @@ module Api
         render 'api/v2/hosts/index', :layout => 'api/v2/layouts/index_layout'
       end
 
+      api :PUT, "/compute_resources/:id/refresh_cache/", N_("Refresh Compute Resource Cache")
+      param :id, :identifier, :required => true
+      def refresh_cache
+        if @compute_resource.respond_to?(:refresh_cache) && @compute_resource.refresh_cache
+          render_message(_('Successfully refreshed the cache.'))
+        else
+          render_message(_('Failed to refresh the cache.'), :status => :unprocessable_entity)
+        end
+      end
+
       private
 
       def action_permission
         case params[:action]
-          when 'available_images', 'available_clusters', 'available_flavors', 'available_folders', 'available_networks', 'available_resource_pools', 'available_security_groups', 'available_storage_domains', 'available_zones', 'associate', 'available_storage_pods'
+          when 'available_images', 'available_clusters', 'available_flavors', 'available_folders', 'available_networks', 'available_resource_pools', 'available_security_groups', 'available_storage_domains', 'available_zones', 'associate', 'available_storage_pods', 'refresh_cache'
             :view
           else
             super
--- a/app/controllers/compute_resources_controller.rb
+++ b/app/controllers/compute_resources_controller.rb
@@ -4,7 +4,7 @@ class ComputeResourcesController < ApplicationController
 
   AJAX_REQUESTS = [:template_selected, :cluster_selected, :resource_pools]
   before_action :ajax_request, :only => AJAX_REQUESTS
-  before_action :find_resource, :only => [:show, :edit, :associate, :update, :destroy, :ping] + AJAX_REQUESTS
+  before_action :find_resource, :only => [:show, :edit, :associate, :update, :destroy, :ping, :refresh_cache] + AJAX_REQUESTS
 
   #This can happen in development when removing a plugin
   rescue_from ActiveRecord::SubclassNotFound do |e|
@@ -73,6 +73,20 @@ class ComputeResourcesController < ApplicationController
     end
   end
 
+  def refresh_cache
+    if @compute_resource.respond_to?(:refresh_cache) && @compute_resource.refresh_cache
+      process_success(
+        :success_msg => _('Successfully refreshed the cache.'),
+        :success_redirect => @compute_resource
+      )
+    else
+      process_error(
+        :error_msg => _('Failed to refresh the cache.'),
+        :redirect => @compute_resource
+      )
+    end
+  end
+
   #ajax methods
   def provider_selected
     @compute_resource = ComputeResource.new_provider :provider => params[:provider]
@@ -125,7 +139,7 @@ class ComputeResourcesController < ApplicationController
     case params[:action]
       when 'associate'
         'edit'
-      when 'ping', 'template_selected', 'cluster_selected', 'resource_pools'
+      when 'ping', 'template_selected', 'cluster_selected', 'resource_pools', 'refresh_cache'
         'view'
       else
         super
--- a/app/models/compute_resources/foreman/model/vmware.rb
+++ b/app/models/compute_resources/foreman/model/vmware.rb
@@ -4,6 +4,7 @@ require 'foreman/exception'
 module Foreman::Model
   class Vmware < ComputeResource
     include ComputeResourceConsoleCommon
+    include ComputeResourceCaching
 
     validates :user, :password, :server, :datacenter, :presence => true
     before_create :update_public_key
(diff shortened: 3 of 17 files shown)


app/services/compute_resource_cache.rb
```diff
@@ -0,0 +1,82 @@
...
+# rails cache to speed up slow or expensive API calls
+class ComputeResourceCache
+  attr_accessor :compute_resource, :cache_duration, :cache_enabled
+  alias_method :cache_enabled?, :cache_enabled
+
+  delegate :logger, :to => ::Rails
+
+  def initialize(compute_resource, opts = {})
+    self.compute_resource = compute_resource
+    self.cache_duration = opts.fetch(:cache_duration, 180.minutes)
```
tbrisker: You can use a kwarg here

app/controllers/api/v2/compute_resources_controller.rb
```diff
@@ -162,12 +162,24 @@ def associate
...
+      end
+
       private
 
       def action_permission
         case params[:action]
           when 'available_images', 'available_clusters', 'available_flavors', 'available_folders', 'available_networks', 'available_resource_pools', 'available_security_groups', 'available_storage_domains', 'available_zones', 'associate', 'available_storage_pods'
             :view
+          when 'invalidate_cache'
+            :edit
```
tbrisker: I think it makes sense to allow viewers to also invalidate the cache (for example if they notice the data seems outdated), this doesn't actually do anything to the compute resource other then reload the data.

app/controllers/compute_resources_controller.rb
```diff
@@ -123,7 +137,7 @@ def resource_pools
 
   def action_permission
     case params[:action]
-      when 'associate'
+      when 'associate', 'invalidate_cache'
```
tbrisker: ditto

app/models/concerns/compute_resource_caching.rb
```diff
@@ -0,0 +1,13 @@
...
+  extend ActiveSupport::Concern
+
+  def invalidate_cache
+    cache.invalidate
+  end
+
+  private
+
+  def cache
+    @cache || ComputeResourceCache.new(self)
```
tbrisker: I think this needs to be `||=` for the new cache to be saved

app/views/compute_resources/form/_vmware.html.erb
```diff
@@ -7,4 +7,6 @@
 <%= checkbox_f f, :set_console_password, :checked => f.object.set_console_password?,
                   :label => _("Console passwords"),
                   :help_inline => _("Set a randomly generated password on the display connection") %>
+<%= checkbox_f f, :caching_enabled, :label => _("Enable caching"),
```
tbrisker, line 10: This parameter should also be added to the api

db/migrate/20170127094357_add_caching_enabled_to_compute_resource.rb
```diff
@@ -0,0 +1,5 @@
+class AddCachingEnabledToComputeResource < ActiveRecord::Migration
+  def change
+    add_column :compute_resources, :caching_enabled, :boolean
```
tbrisker: This should have a default false I believe

tbrisker (changes requested):
Looks very good in general, a few comments inline about the code. Couldn't test as I don't have a vmware resource.

app/services/compute_resource_cache.rb
ohadlevy: would it make sense to include some sort of a hourly task that keeps refreshing the cache offline?

app/views/compute_resources/show.html.erb
```diff
@@ -4,6 +4,7 @@
 <% title_actions display_link_if_authorized(_("Associate VMs"),
                                             hash_for_associate_compute_resource_path(:compute_resource_id => @compute_resource).merge(:auth_object => @compute_resource, :permission => 'edit_compute_resources'),
                                             :title => _("Associate VMs to Foreman hosts"), :method => :put, :class=>"btn btn-default"),
+(display_link_if_authorized(_('Clear Cache'), hash_for_invalidate_cache_compute_resource_path(@compute_resource).merge(:auth_object => @compute_resource), :method => :put, :class => "btn btn-default") if @compute_resource.respond_to?(:invalidate_cache) && @compute_resource.caching_enabled),
```
ohadlevy: nitpick: I would call it reload vs clear cache.

app/views/compute_resources/show/_vmware.html.erb
```diff
@@ -6,3 +6,7 @@
   <td><%= _("Console passwords") %></td>
   <td><%= @compute_resource.set_console_password? ? _("Enabled") : _("Disabled") %></td>
 </tr>
+<tr>
+  <td><%= _("Caching") %></td>
+  <td><%= @compute_resource.caching_enabled ? _("Enabled") : _("Disabled") %></td>
```
ohadlevy, line 11: consider using an icon ? from http://www.patternfly.org/styles/icons/#_ maybe  pficon-ok and  pficon-warning-triangle-o?

config/routes.rb
```diff
@@ -399,6 +399,7 @@
           get 'resource_pools'
           post 'ping'
           put 'associate'
+          put 'invalidate_cache'
```
ohadlevy: nitpick: i would prefer reload vs invalidate (sounds scary as we all know cache invalidation is a hard problem ;))

db/migrate/20170127094357_add_caching_enabled_to_compute_resource.rb
ohadlevy: do you see a need to default to false? I would rather opt in by default, no one wants the slower option unless they have a problem?

ohadlevy:
I would like to see a refresh / reload button on the host edit page, i think this is where you would really need  up to date / latest info as you dont have the option (dc, cluster etc).

app/services/compute_resource_cache.rb
timogoebel: Yes, I do want to add a hourly rake task for this that can be called by cron. I'd prefer to do this in a separate PR, though.

app/views/compute_resources/show.html.erb
timogoebel: How about "refresh cache"?

app/views/compute_resources/show/_vmware.html.erb
timogoebel: I actually do prefer the text vs. the icons to be consistent in the UI. We can change this to icons in a separate PR. `pficon-warning-triangle-o` is too much warning. If a user disables the caching, it won't break anything.

timogoebel:
> I would like to see a refresh / reload button on the host edit page, i think this is where you would really need up to date / latest info as you dont have the option (dc, cluster etc).

@ohadlevy : Would it be ok for you to add this in a separate PR? I think, this will make coding and reviewing easier.

app/services/compute_resource_cache.rb
```diff
@@ -0,0 +1,89 @@
+# This class caches attributes for a compute resource in
+# rails cache to speed up slow or expensive API calls
+class ComputeResourceCache
+  attr_accessor :compute_resource, :cache_duration
+
+  delegate :logger, :to => ::Rails
+
+  def initialize(compute_resource:, cache_duration: nil)
```
tbrisker: I'm afraid required kwargs is supported only in ruby 2.1, so we can't use that yet. I'd suggest making the first parameter a regular argument and just the cache duration a kwarg. Then you can set the default to 180.minutes in the kwarg rather then below, like so:
```ruby
def initialize(compute_resource, cache_duration: 180.minutes)
```

db/migrate/20170127094357_add_caching_enabled_to_compute_resource.rb
```diff
@@ -0,0 +1,5 @@
+class AddCachingEnabledToComputeResource < ActiveRecord::Migration
+  def change
+    add_column :compute_resources, :caching_enabled, :boolean, default: true
```
tbrisker, line 3: Won't this make compute resources that don't support caching appear to have caching enabled?

tbrisker (commented):
A couple more comments inline. As for the naming of the action, "reload" seems to be a bit misleading, as it isn't actually reloaded, rather "cleared" or "flushed" IIUC.

db/migrate/20170127094357_add_caching_enabled_to_compute_resource.rb
timogoebel: @tbrisker : Yes, it will. But it's not shown in the UI or API. So it figured, it doesn't really matter.   Do you see that differently?

tbrisker: ok, makes sense.

app/services/compute_resource_cache.rb
```diff
@@ -0,0 +1,89 @@
+# This class caches attributes for a compute resource in
+# rails cache to speed up slow or expensive API calls
+class ComputeResourceCache
+  attr_accessor :compute_resource, :cache_duration
+
+  delegate :logger, :to => ::Rails
+
+  def initialize(compute_resource, cache_duration: nil)
+    self.compute_resource = compute_resource
+    self.cache_duration = cache_duration || 180.minutes
```
tbrisker: Instead of defaulting the kwarg to nil then defaulting here to 180.minutes, you can do that in the kwarg itself

timogoebel:
@tbrisker , @ohadlevy : Addressed all of your comments. Thanks a lot for these.

> As for the naming of the action, "reload" seems to be a bit misleading, as it isn't actually reloaded, rather "cleared" or "flushed" IIUC.

@tbrisker : Would `refresh cache` be fine for your opinion? I had `clear cache` before, but @ohadlevy suggested to use something different.

I currently see these follow up tasks:

- [ ] Add a rake task to refresh the cache.
- [ ] Add a cache refresh button to the select boxed in the host form that flushes the cache and reloads the data.

tbrisker:
@timogoebel I guess refresh is better, @ohadlevy what do you think? Still one minor comment on the kwargs, other than that LGTM but I would like someone with access to VMWare to test it as well before merging.

timogoebel:
@shlomizadok : Would you mind testing this for me? I think a VM show and create should suffice.

timogoebel:
Caching is now disabled if the compute resource has not been saved yet, fixed the kwarg default and removed caching from `dc` method as method is quite fast.

timogoebel:
Rebased and renamed `reload_cache` to `refresh_cache`.

timogoebel:
Fixed an issue with caching of `dc.clusters`.

@TheKangaroo: Would you mind testing this PR and let us know if you see any issues? Thanks.

app/models/compute_resources/foreman/model/vmware.rb
```diff
@@ -53,41 +54,54 @@ def max_memory
...
       end
-      dc.clusters.map(&:full_path).sort
+      dc_clusters.map(&:full_path).sort
     end
 
     def datastores(opts = {})
       if opts[:storage_domain]
-        name_sort(dc.datastores.get(opts[:storage_domain]))
+        cache.cache(:"datastores-#{opts[:storage_domain]}") do
+          name_sort(dc.datastores.get(opts[:storage_domain]))
```
tbrisker, line 80: not directly required, but this could benefit from a kwarg for storage domain

tbrisker (approved):

tbrisker:
Looks good to me pending testing, test failure is unrelated.

TheKangaroo:
Hi @tbrisker, I tested it in our environment.
I couldn't find any issues and the runtime improvements in creating hosts is huge :)
I really like your PR @timogoebel

shlomizadok:
@tbrisker - testing this now. give me 2 minutes to ack :)

shlomizadok:
ACK :+1:

tbrisker:
Thanks @timogoebel !

timogoebel:
Created two follow-up tickets:

Cache-Warming rake task:
http://projects.theforeman.org/issues/18314

Refresh from host page:
http://projects.theforeman.org/issues/18315

PR #4239, opened 2017-01-27
Fixes #18289 - fix IP based proxy authentication

ares:


dLobatog:
@ares Thanks - would it be OK to check `request.env['REMOTE_ADDR']` against the list of possible proxy addresses `(Regexp.new(Setting[:remote_addr])` like we do in app/controllers/unattended_controller.rb ?

ares:
I'm not sure I understand the question. Which part exactly you have in mind? If you mean the initializer (which would probably make sense) then I'm not sure how to reliably parse the regexp to specific IP addresses that would be trusted proxies.

ares:
ping

config/initializers/trusted_proxies.rb
```diff
@@ -0,0 +1,2 @@
+# Workaround Rails issue https://github.com/rails/rails/issues/5223
+Foreman::Application.config.action_dispatch.trusted_proxies = %w(::1 127.0.0.1). map { |proxy| IPAddr.new(proxy) }
```
dLobatog, line 2: Looks like the fix was merged already?

dLobatog (changes requested):
@ares Looks good, what I meant to check is just that in the concern, when we use a remote IP, like (`request.remote_ip`) as far as I understand, the regular IP (`request.ip`) should match `Setting[:remote_addr]`

xprazak2:
What is the status here?

ares:
closing as I know longer need the fix, though I think it's still valid issue

PR #4240, opened 2017-01-27
Fixes #18292 - Make notifications from an RSS feed

dLobatog:
This adds a task to connect to an RSS feed, by default the one from our
community blog. The idea is to make this run every night with 'cron' on
every Foreman installation to keep people updated about changes, and
inform them about events, etc...

Also, people can setup the task for their internal company blogs or give
it any other uses they may think about. The task can be easily removed
from cron, and I'd suggest enabling it by default.

The links don't look like actual HREF tags, which would require making React render the HTML without escaping. I don't think that's a good idea, so this PR is probably a good place to introduce an endpoint in the component to add links like the one provided here http://www.patternfly.org/pattern-library/communication/notification-drawer/#/example-overview-2.

Some milestones for this feature:

- [x] - Add endpoint to notifications component for links (https://github.com/theforeman/foreman/pull/4265)
- [x] - Add service that allows tasks that generate notifications to run on a schedule.
- [ ] - Add option to puppet-foreman to choose whether to enable this or not. Make a noticeable warning if it's enabled so the operator knows this app will contact external sources (theforeman.org) by default

screenshot from 2017-01-27 22-04-43
screenshot from 2017-01-27 22-04-58

cc @GregSutcliffe @shlomizadok I hope you make good use of this if it gets merged 
cc @bkearney I think this would be cool to use downstream to connect it to the blog or other sources

### DIFF
--- a/app/jobs/create_rss_notifications.rb
+++ b/app/jobs/create_rss_notifications.rb
@@ -1,0 +1,16 @@
+class CreateRssNotifications < ApplicationJob
+  after_perform do |job|
+    self.class.set(:wait => 12.hours).perform_later(job.arguments.first)
+  end
+
+  def perform(options = {})
+    # Defaults to theforeman.org blog RSS
+    UINotifications::RssNotificationsChecker.new(options).deliver!
+  end
+
+  rescue_from(StandardError) do |error|
+    Foreman::Logging.logger('background').error(
+      'RSS notification checker: '\
+      "Error while creating notifications #{error}: #{error.backtrace}")
+  end
+end
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -66,7 +66,7 @@ class Notification < ApplicationRecord
 
   def set_custom_attributes
     return unless notification_blueprint # let validation catch this.
-    self.actions = UINotifications::URLResolver.new(
+    self.actions ||= UINotifications::URLResolver.new(
       subject,
       notification_blueprint.actions
     ).actions if notification_blueprint.actions.any?
--- a/app/services/ui_notifications/rss_notifications_checker.rb
+++ b/app/services/ui_notifications/rss_notifications_checker.rb
@@ -1,0 +1,74 @@
+require 'rss'
+require 'date'
+
+module UINotifications
+  class RssNotificationsChecker
+    class Item
+      def initialize(feed_item)
+        @item = feed_item
+      end
+
+      def link
+        @link ||= @item.link.respond_to?(:href) ? @item.link.href : @item.link
+      end
+
+      def summary
+        @summary ||= begin
+          summary = @item.summary if @item.respond_to?(:summary)
+          summary = @item.description if @item.respond_to?(:description)
+          summary.respond_to?(:content) ? summary.content : summary
+        end
+      end
+
+      def title
+        @title ||= @item.title.respond_to?(:content) ? @item.title.content : @item.title
+      end
+    end
+
+    def initialize(options = [])
+      @url = options[:url] || Setting[:rss_url]
+      @latest_posts = options[:latest_posts] || 3
+      @force_repost = options[:force_repost] || false
+      @audience = options[:audience] || Notification::AUDIENCE_GLOBAL
+    end
+
+    def deliver!
+      # This is a noop every time rss_enable=false, the moment it
+      # gets enabled, notifications for RSS feeds are created again
+      return true unless Setting[:rss_enable]
+      feed = RSS::Parser.parse(@url, false)
+      feed.items[0, @latest_posts].each do |feed_item|
+        item = Item.new(feed_item)
+        blueprint = rss_notification_blueprint
+        if notification_already_exists?(item)
+          next unless @force_repost
+        end
+        Notification.create(
+          :initiator => User.anonymous_admin,
+          :audience => @audience,
+          :message => item.title,
+          :notification_blueprint => blueprint,
+          :actions => {
+            :links => [
+              {
+                :href => item.link,
+                :title => _('Open'),
+                :external => true
+              }
+            ]
+          }
+        )
+      end
+    end
+
+    private
+
+    def rss_notification_blueprint
+      NotificationBlueprint.unscoped.find_by_name('rss_post')
+    end
+
+    def notification_already_exists?(item)
+      !!Notification.unscoped.find_by_message(item.summary)
+    end
+  end
+end
--- a/config/application.rb
+++ b/config/application.rb
@@ -247,7 +247,7 @@ module Foreman
         if defined?(ForemanTasks)
           ForemanTasks.dynflow
         else
-          ::Dynflow::Rails.new(nil, Foreman::Dynflow::Configuration.new)
+          ::Dynflow::Rails.new(nil, ::Foreman::Dynflow::Configuration.new)
         end
       @dynflow.require!
       @dynflow
--- a/config/initializers/rss_notifications.rb
+++ b/config/initializers/rss_notifications.rb
@@ -1,0 +1,12 @@
+# First, we check if there's a job already enqueued for RSS notifications
+::Foreman::Application.dynflow.config.on_init do |world|
+  pending_jobs = world.persistence.find_execution_plans(filters: { :state => 'scheduled' })
+  scheduled_job = pending_jobs.select do |job|
+    delayed_plan = world.persistence.load_delayed_plan job.id
+    next unless delayed_plan.present?
+    delayed_plan.to_hash[:serialized_args].first["job_class"] == 'CreateRssNotifications'
+  end
+
+  # Only create notifications if there isn't a scheduled job
+  CreateRssNotifications.perform_later unless scheduled_job.present?
+end
--- a/lib/tasks/rss.rake
+++ b/lib/tasks/rss.rake
@@ -1,0 +1,42 @@
+namespace :rss do
+  desc <<-END_DESC
+Create notifications from an RSS feed.
+
+By default, the last 3 posts from the feed are sent to a global audience,
+and expire in one week since the announcement. Previous posts in the feed will not
+create a new notification unless "FOREMAN_RSS_FORCE_REPOST" is set to true.
+
+This will trigger a job that will check for notifications every 12 hours.
+
+It accepts the following environment variables:
+ - FOREMAN_RSS_URL
+   + An RSS URL and finds the latest posts, by default 'https://theforeman.org/feed.xml'
+ - FOREMAN_RSS_LATEST_POSTS
+   + The number of latest posts to retrieve, by default '3'
+ - FOREMAN_RSS_FORCE_REPOST
+   + If 'true', the task will send notifications again for posts in the RSS feed that already created
+     notifications.
+   + For example, on the first run of this task, it creates notifications for the latest 3 posts,
+     "Newsletter December", "Newsletter November", "Newsletter October". If you run it again without
+     FOREMAN_RSS_FORCE_REPOST, no new notifications will be created for these three posts. If you do,
+     the task will create notifications once again for "Newsletter December", November and October.
+ - FOREMAN_RSS_AUDIENCE
+   + The audience that will recieve the notification. By default, 'global'
+   + Possible values:
+     - 'global' - All users will receive the notification
+     - 'admin' - Only administrator users will receive the notification
+
+Examples:
+  # foreman-rake rss:create_notifications FOREMAN_RSS_AUDIENCE='admin'
+END_DESC
+
+  task :create_notifications => :environment do
+    rss_checker = UINotifications::RssNotificationsChecker.new(
+      :url => ENV['FOREMAN_RSS_URL'],
+      :latest_posts => ENV['FOREMAN_RSS_LATEST_POSTS'],
+      :audience => ENV['FOREMAN_RSS_AUDIENCE'],
+      :force_repost => ENV['FOREMAN_RSS_FORCE_REPOST'] == 'true',
+    )
+    rss_checker.deliver!
+  end
+end
--- a/app/jobs/application_job.rb
+++ b/app/jobs/application_job.rb
@@ -1,0 +1,2 @@
+class ApplicationJob < ActiveJob::Base
+end
--- a/app/models/setting/notification.rb
+++ b/app/models/setting/notification.rb
@@ -1,0 +1,23 @@
+class Setting::Notification < Setting
+  def self.default_settings
+    [
+      self.set('rss_enable', N_('Whether to get RSS notifications or not'), true, N_('RSS enable')),
+      self.set('rss_url', N_('URL to fetch RSS notifications from'), 'https://theforeman.org/feed.xml', N_('RSS URL')),
+    ]
+  end
+
+  def self.load_defaults
+    # Check the table exists
+    return unless super
+
+    self.transaction do
+      default_settings.each { |s| self.create! s.update(:category => "Setting::Notification")}
+    end
+
+    true
+  end
+
+  def self.humanized_category
+    N_('Notifications')
+  end
+end
--- a/db/seeds.d/170-notification_blueprints.rb
+++ b/db/seeds.d/170-notification_blueprints.rb
@@ -32,6 +32,20 @@ blueprints = [
         title: N_('Update host')
       ]
     }
+  },
+  {
+    group: _('Community'),
+    name: 'rss_post',
+    level: 'info',
+    message: _('RSS post message goes here'),
+    actions:
+    {
+      links:
+      [
+        title: _('URL'),
+        external: true,
+      ]
+    }
   }
 ]
 
--- a/test/fixtures/settings.yml
+++ b/test/fixtures/settings.yml
@@ -353,3 +353,13 @@ attribute73:
   category: Setting::Provisioning
   default: PXELinux default local boot
   description: 'Set up default local boot template for PXELinux'
+attribute74:
+  name: rss_enable
+  category: Setting::Notification
+  default: true
+  description: 'Whether to enable or not RSS feed notifications'
+attribute75:
+  name: rss_url
+  category: Setting::Notification
+  default: "http://theforeman.org/feed.xml"
+  description: 'Default URL for RSS feed notifications'
--- a/test/unit/ui_notifications/rss_notifications_checker_test.rb
+++ b/test/unit/ui_notifications/rss_notifications_checker_test.rb
@@ -1,0 +1,21 @@
+require 'test_helper'
+require 'ostruct'
+
+module UINotifications
+  class RssNotificationsCheckerTest < ActiveSupport::TestCase
+    context 'when force_repost is disabled' do
+      setup do
+        @notifications_service = RssNotificationsChecker.new(:force_repost => false)
+      end
+
+      test 'already existing notifications are not created' do
+        feed = OpenStruct.new(:items => [1,2,3])
+        RSS::Parser.expects(:parse).returns(feed)
+        Notification.expects(:create).never
+        @notifications_service.expects(:notification_already_exists?).
+          returns(true).at_least_once
+        @notifications_service.deliver!
+      end
+    end
+  end
+end


lib/tasks/rss.rake
```diff
@@ -0,0 +1,73 @@
...
+   + Possible values:
+     - 'global' - All users will receive the notification
+     - 'admin' - Only administrator users will receive the notification
+
+Examples:
+  # foreman-rake rss:create_notifications FOREMAN_RSS_AUDIENCE='admin'
+END_DESC
+
+  task :create_notifications => :environment do
+    url = ENV['FOREMAN_RSS_URL'] || 'https://theforeman.org/feed.xml'
```
ohadlevy: IMHO it should be a separate service class, this way we could trigger it either from a task, or maybe from the about page manually?

lib/tasks/rss.rake
```diff
@@ -0,0 +1,73 @@
...
+    feed = RSS::Parser.parse(url, false)
+    feed.items[0, ENV['FOREMAN_RSS_LATEST_POSTS'] || 3].each do |item|
```
ohadlevy: why limit to 3 vs from the last time you fetched the data?

lib/tasks/rss.rake
ohadlevy: I could see having one task that trigger all async stuff for notifications (e.g hourly or daily) and this will be one of them.
further, i could see users disabling this functionality in some sort of UI instead of stopping the cron job, so we could probably let the user select it in the UI.

ohadlevy: not to mention you can write better tests this way :)

lib/tasks/rss.rake
```diff
@@ -0,0 +1,73 @@
...
+      group = ENV['FOREMAN_RSS_NOTIFICATION_GROUP'] || 'Foreman Community'
```
ohadlevy: should be extracted?

ohadlevy: also, maybe drop the word Foreman, just community or Community Notifications/Blogs?

lib/tasks/rss.rake
```diff
@@ -0,0 +1,73 @@
...
+        # If none exist, create one.
+        blueprint = NotificationBlueprint.new(
+          :group => group,
+          :message => "#{summary} - #{link}",
+          :level => 'info',
+          :expires_in => expiration_date,
+          :name => name)
+      end
+      # Create a notification using the blueprint
+      audience = ENV['FOREMAN_RSS_NOTIFICATION_GROUP'] || 'global'
```
ohadlevy: should this really be global for every blog entry? maybe we should consider adding metadata to our blog? (or tags) for example, a security announcement would probably go only to admins?

ohadlevy: nitpick: I would use the constant vs create the 'global' string again.

lib/tasks/rss.rake
```diff
@@ -0,0 +1,73 @@
...
+      initiator = User.anonymous_admin
+      notification = Notification.new(
+        :initiator => initiator,
+        :audience => audience,
+        :notification_blueprint => blueprint
+      )
+      blueprint.save
```
ohadlevy: if it fails, should it report it somewhere?

lib/tasks/rss.rake
dLobatog: Where would you store the last time you fetched the data? A few lines below, I'm matching the already created notifications - if they have the same name as the latest blog posts, no new notifications will be created

ohadlevy: creation of the blueprint maybe?

lib/tasks/rss.rake
dLobatog: The UI - yeah I agree, the user should be able to opt-out, although having both things on the same PR makes it harder to review. 

I'll move to a service class when I get a chance, that's true too. 

I am not sure what you exactly mean by a task that triggers all async stuff for notifications, just one task that manually calls all services that create notifications?

ohadlevy: I think that over time, we would add more notifications that relay on checking something every interval, for example:

- is your certs about to expire?
- do we have new blogs on the community blog
- does your smart proxy active?
- ...

my wish is to have one cron job to trigger all kind of probes vs multiple cronjobs.

Last comment about it, we should ensure that the cron entry does not happen on the exact ms across multiple servers to reduce DoS on our blog server.

dLobatog:
I've repushed and moved the RSS parsing code to a service - not all of the comments have been addressed though, like the cronjob for 'all notification services', test for the service, or the action links in the React component. 

@ohadlevy How much should I cover in this one PR?

ohadlevy:
@dLobatog did you push your new version? I can't see it here.

I would create a separate PR for links, and cron (as its in the installer anyway?)

dLobatog:
@ohadlevy Sorry, I've just pushed it - it asked for my pw and I forgot to put it.

lib/tasks/rss.rake
```diff
@@ -0,0 +1,50 @@
+require 'rss'
```
ohadlevy: move to service class?

app/services/rss_notifications_manager.rb
```diff
@@ -0,0 +1,42 @@
...
+      summary = item.summary.content # display summary if available
+      link = item.link.href # link href is indispensable
+
+      # Find previous notification blueprints with this name.
+      if (blueprint = NotificationBlueprint.find_by_name(name))
+        # If some exist use them
+        next unless @force_repost
+      else
+        # If none exist, create one.
+        blueprint = NotificationBlueprint.new(
```
ohadlevy: does this really works? I would think blueprint is only defined inside the else block? or does it work as its part of if statement? regardless, I would break it down to multiple methods, so its more readable and testable.

app/services/rss_notifications_manager.rb
```diff
@@ -0,0 +1,42 @@
...
+          :group => @group,
+          :message => "#{summary} - #{link}",
+          :level => 'info',
+          :expires_in => @expiration_date,
+          :name => name)
+      end
+      # Create a notification using the blueprint
+      initiator = User.anonymous_admin
+      notification = Notification.new(
```
ohadlevy: maybe I'm missing something obvious, but wont this create duplicate notifications if you run this multiple times?

app/services/rss_notifications_manager.rb
dLobatog: Yes it works, try to run it - I can move it outside indeed

app/services/rss_notifications_manager.rb
dLobatog: It will not run multiple times unless you actively set FORCE_REPOST to true https://github.com/theforeman/foreman/pull/4240/files#diff-1bdbf04d1f1ed2e0cd02d313de48ad67R19

ohadlevy: ok, so we assume a blueprint for every blog post, rather then a generic blog post blueprint..? Ideally, you should have one notification blueprint for all blog posts, and a notification which is the actual blog entry.

dLobatog: I don't understand that - the notification blueprint stores the notification message, which is what is shown on the drawer. What would be a 'generic blog post blueprint'? 

`Notification` doesn't store anything, it's just a way of delivering notifications to recipients and a way to remove them easily.

ohadlevy: I believe the original idea was to combine the blueprint and subject and create the actual notification out of the two.

dLobatog:
Setting to WoC as it kind of depends on https://github.com/theforeman/foreman/pull/4265 - putting the links in the message is too ugly even for a proof of concept

app/jobs/rss_notifications_checker.rb
```diff
@@ -0,0 +1,59 @@
+require 'rss'
+require 'date'
+
+class RssNotificationsChecker < ApplicationJob
+  queue_as :notifications
+  attr_reader :url
+
+  after_perform do |job|
+    logger.info(_("Finished creating notifications for RSS feed: #{job.url}"))
+    self.class.set(:wait => 12.hours).perform_later
```
dLobatog: This is what makes the task 'recurring'. Any ActiveJob task can call itself again after it finishes to create a 'recurring job'

app/jobs/rss_notifications_checker.rb
```diff
@@ -0,0 +1,59 @@
+require 'rss'
+require 'date'
+
+class RssNotificationsChecker < ApplicationJob
+  queue_as :notifications
```
dLobatog: We can also assign executors for each queues if we care about it. E.g: maybe the notifications queue is less important than the pulp_sync queue

config/application.rb
```diff
@@ -168,6 +168,9 @@ class Application < Rails::Application
     # New config option to opt out of params "deep munging" that was used to address security vulnerability CVE-2013-0155.
     config.action_dispatch.perform_deep_munge = false
 
+    # Use delayed_job_active_record as the backend for ActiveJob notifications
+    config.active_job.queue_adapter = :delayed_job
```
dLobatog: Supports MySQL & Postgres. I assume we will want to change it to the https://github.com/theforeman/foreman-tasks adapter at some point.

dLobatog:
@ohadlevy I think you'd be interested in this. There is little to no reason to maintain a Notifications service that takes care of running the notification fetching tasks in an async way. With the ActiveJob implementation showcased here, it's as simple as:

```
class SubscriptionExpirationChecker < ApplicationJob
  after_perform do
    self.class.set(:wait => 12.hours).perform_later
  end

  def perform
    create_subs_expired_notification if subscriptions_about_to_expire.present?
  end
end
```

Of course, we would need to run `bin/delayed_job` to consume the tasks with this implementation (any executor will require to do this, be it dynflow, delayed_job or whatever). I've been looking at what would be required to do an interface for ActiveJob that makes it this simple to enqueue jobs (https://github.com/theforeman/foreman-tasks/pull/231) - enqueing tasks is not a problem. 

However with ForemanTasks ([an example of all the code needed for a simple task](https://github.com/theforeman/foreman_ansible/blob/master/app/lib/actions/foreman_ansible/play_host_roles.rb)) , there is really little point on using ActiveJob if I already have to write so much code for a task, it won't make it any simpler but it will just add another abstraction layer on top. 

I would suggest reassessing our assumptions as for why it's mandatory for us to use Dynflow (tasks UI could consume other executors with not so much changes, I think) when it's this simple to do it with other open source tools with a much larger community & prod installations ([but that's just like my opinion..](https://www.youtube.com/watch?v=pWdd6_ZxX8c))

cc @thomasmckay

ohadlevy:
> @ohadlevy I think you'd be interested in this. There is little to no reason to maintain a Notifications service that takes care of running the notification fetching tasks in an async way. With the ActiveJob implementation showcased here, it's as simple as:
> 

I do support this, in fact, I already expressed my opinion that not all async operations needs to be tasks, ActiveJob seems like a reasonable interface for this kind of jobs, that allows portability between different backends when needed.

> Of course, we would need to run bin/delayed_job to consume the tasks with this implementation (any executor will require to do this, be it dynflow, delayed_job or whatever). 

Since this is a packaging / installer effort, I would like to ask @domcleal, @mmoll and @ehelms for early feedback at this time, I would really not want to undo patches all over again - please ack / nack this approach.

> I've been looking at what would be required to do an interface for ActiveJob that makes it this simple to enqueue jobs (theforeman/foreman-tasks#231) - enqueing tasks is not a problem.
> 
> However with ForemanTasks (an example of all the code needed for a simple task) , there is really little point on using ActiveJob if I already have to write so much code for a task, it won't make it any simpler but it will just add another abstraction layer on top.

Sadly, since foreman-tasks in core is currently not an option, I'm not sure it makes sense to keep on blocking because of it, I'm very welcome counter feedback from @iNecas, @ares and others of course.
>
> I would suggest reassessing our assumptions as for why it's mandatory for us to use Dynflow (tasks UI could consume other executors with not so much changes, I think) when it's this simple to do it with other open source tools with a much larger community & prod installations (but that's just like my opinion..)

I don't think it was mandatory at any point of time, but we of course did not want to add yet another very similar backend service, please note that its not enough to add it to code, but also include installer, packaging and supporting tools (e.g. katello-service etc).

I've also expressed my believe that we need to split between orchestration / planning workflow, to simple execution of async operations and would prefer large scale tested implementations instead of fighting with all the bugs our self.

While it might be a future discussion, I do wonder if delayed job is the best candidate to start with, i personally think it might worth the effort to introduce sidekiq and redis instead.

I do realize we are getting close to code freeze, and would like to have the functionality earlier than later (e.g. read: I would like it in 1.15).

thanks @dLobatog for promoting this discussion!

mmoll:
> Since this is a packaging / installer effort, I would like to ask @domcleal, @mmoll and @ehelms for early feedback at this time, I would really not want to undo patches all over again - please ack / nack this approach.

I'm not getting the question entirely... this is about using ActiveJob and having a ~/foreman/bin/delayed_job and its startup script, correct?

ehelms:
I am also trying to understand. Let me see if I can recap what I read.

An async task runner is needed in order to routinely fetch the data. The idea is to introduce a new async runner to the Foreman stack to perform this action because the feeling is that Foreman Tasks is too heavy weight a solution. If this recap is correct, my immediate reaction is to avoid adding yet another process to our stack.

We seem to be butting up against this question more and more and I'd rather see a frank discussion with an eye toward the future about what we do with async operations given we are split. Having multiple ways to do things has it's downsides with respect to developer mindshare and debugging. This feels similar to the whole UI discussion that plagued us for years.

Baking the async discussion into this particular PR makes it harder to separate the two issues. I hate to drag out a PR but I think this warrants a separate discussion either over foreman-dev list or an RFC that outlines the needs for async operations, and the classes of async operations. I think there are a number of folks with good knowledge from having implemented or used our tools and those that exist elsewhere to have good feedback to arrive at a final solution.

dLobatog:
@ehelms I think you misread the situation:

1. This PR initially started with a rake-task that could be put into cron (https://github.com/dlobatog/foreman/tree/notifications-rss-task). Since it would be very unwieldy to add all tasks that create notifications as a new cron job, the idea was to draft a model solution for other notification-related tasks.
2. foreman-tasks in core seemed to not be an option now because of (https://github.com/theforeman/foreman-packaging/pull/1436)
3. I have a goal of offering a way for tasks that have to routinely check something (smart-proxy pulp size, subscriptions expiration, RSS notifications..) a good way of doing this. foreman-tasks actions do not allow to schedule anything right now, it doesn't allow multiple consumer executors for different tasks, plus it's significantly more complicated than (https://github.com/theforeman/foreman/pull/4240#issuecomment-280104637)
4. The current solution is to use ActiveJob from Rails. If we write an adapter for Dynflow, we should be able to just change it here (https://github.com/theforeman/foreman/pull/4240/files#diff-b1fe55db50c712fef0673345e5b9c0d9R172) and the application should keep on working. I started some drafting for this on https://github.com/theforeman/foreman-tasks/pull/231 
5. It seems to me that foreman-tasks should provide an adaptor for Dynflow based on ActiveJob API and deprecate the current way of running tasks.
6. In any case, I think we should reconsider if we want every single async, queued, planned, etc.. operation to run as a ForemanTasks::Task I think the ActiveJob approach is much simpler and powerful, and delayed_job has been much more battle-tested than Dynflow (plus it offers things like prioritization of jobs and different queues that Dynflow does not). 
7. I'm definitely willing to change the PR back to the rake task since using `delayed_job` will make it unmergeable as the installer/packaging/tooling etc.. does not handle it. Just wanted to showcase how it would look like.

iNecas:
I think the ActiveJob is the right approach, introducing Delayed Job as hard dependency here not so much. especially if the cron job can do that.

I've already expressed that I don't insist foreman-tasks to be the default async processor (it's more orchestrator, than async processor anyway), if the tools use the same UI/API infrastructure.

+1 for proper discussion about what async processor to introduce.

ohadlevy:
> I'm not getting the question entirely... this is about using ActiveJob and having a ~/foreman/bin/delayed_job and its startup script, correct?

@mmoll that is correct, its about packaging and having another service / daemon (delayed job or maybe sidekiq) and the relevant installer parts.

ohadlevy:
> I am also trying to understand. Let me see if I can recap what I read.
> 
> An async task runner is needed in order to routinely fetch the data. The idea is to introduce a new async runner to the Foreman stack to perform this action because the feeling is that Foreman Tasks is too heavy weight a solution. If this recap is correct, my immediate reaction is to avoid adding yet another process to our stack.

@ehelms As mentioned earlier by @dLobatog the problem is not with foreman tasks being too heavy rather than we can't get it any time soon into core, and that most of core usage cases do not require planning, rather a simple async executor. 
> 
> We seem to be butting up against this question more and more and I'd rather see a frank discussion with an eye toward the future about what we do with async operations given we are split. Having multiple ways to do things has it's downsides with respect to developer mindshare and debugging. This feels similar to the whole UI discussion that plagued us for years.

100% with you on that, I don't want to repeat it all over again.
I personally don't think that every action needs to be a foreman task action, and I think the really valuable part is that if you don't have foreman-tasks around, the developer is forced to create a UI for that workflow rather then to leave it to foreman tasks UI, in many ways, I think this was a mistake I don't want to repeat in core.
> 
> Baking the async discussion into this particular PR makes it harder to separate the two issues. I hate to drag out a PR but I think this warrants a separate discussion either over foreman-dev list or an RFC that outlines the needs for async operations, and the classes of async operations. I think there are a number of folks with good knowledge from having implemented or used our tools and those that exist elsewhere to have good feedback to arrive at a final solution.

ohadlevy:
@iNecas @dLobatog and others, what about creating a dynflow ActiveJob adapter instead? I can see two benefits:

1. we keep on using our current stack (dynflow)
2. we avoid the packaging problem as foreman now relay on dynflow which is a gem vs foreman-tasks which is a plugin?

I think this will also allow more flexibility going forward - meaning we will allow a developer to decide if he needs a foreman task or a simply a Job ?

mmoll:
Regading dynflow as gem dependency, I wouldn't be against that.

However, for Debian the consequence is (like for all core dependencies) that the current relevant dynflow version is included into the core package at package build time and it can't be updated without rebuilding the core package, so normally that will only happen with point releases.

That would still mean, there has to be a dynflow executor init script coming with the core package, correct?

ohadlevy:
> However, for Debian the consequence is (like for all core dependencies) that the current relevant dynflow version is included into the core package at package build time and it can't be updated without rebuilding the core package, so normally that will only happen with point releases.

I would believe this means we move away the pinning of dynflow from foreman-tasks gem into foreman core.
> 
> That would still mean, there has to be a dynflow executor init script coming with the core package, correct?

AFAIU we would need to package https://github.com/theforeman/foreman-tasks/blob/master/bin/dynflow-executor in core

dLobatog:
Setting to WoC as I'd like this to use https://github.com/theforeman/foreman/pull/4316

ohadlevy:
@dLobatog this needs a rebase and update to use the new ui notification structure.

dLobatog:
Rebased and fixed to work with the new notification structure. I will add tests in a few.

app/models/notification_blueprint.rb
```diff
@@ -10,7 +10,6 @@ class NotificationBlueprint < ActiveRecord::Base
   has_many :notifications, :dependent => :destroy
   store :actions, :accessors => [:links], :coder => JSON
 
-  validates :message, :presence => true
```
ohadlevy: why? How could a blueprint exists without a message?

app/models/notification.rb
```diff
@@ -68,7 +68,7 @@ def set_notification_recipients
 
   def set_custom_attributes
     return unless notification_blueprint # let validation catch this.
-    self.actions = UINotifications::URLResolver.new(
+    self.actions ||= UINotifications::URLResolver.new(
```
ohadlevy, line 69: not sure  i follow, related to the comment blue on message validation on the blueprint.

app/services/rss_notifications_checker.rb
```diff
@@ -0,0 +1,48 @@
+require 'rss'
+require 'date'
+
+class RssNotificationsChecker
```
ohadlevy: this file is in the wrong place, it should be in app/services/ui_notifications/.

Also, it should inherit from the notification base class imho.

lib/tasks/rss.rake
```diff
@@ -0,0 +1,42 @@
...
+END_DESC
+
+  task :create_notifications => :environment do
+    rss_checker = RssNotificationsChecker.new(
+      :url => ENV['FOREMAN_RSS_URL'],
+      :latest_posts => ENV['FOREMAN_RSS_LATEST_POSTS'],
+      :audience => ENV['FOREMAN_RSS_AUDIENCE'],
+      :force_repost => ENV['FOREMAN_RSS_FORCE_REPOST'] == 'true',
+    )
+    rss_checker.create_notifications
```
ohadlevy: please follow the UINotification conventions, it should respond to:

```ruby
UINotifications::RSS::CommunityBlog.deliver!(options = {})
```

db/seeds.d/17-notification_blueprints.rb
```diff
@@ -32,6 +32,18 @@
         title: _('Update host')
       ]
     }
+  },
+  {
+    group: _('Community'),
```
ohadlevy: based on a comment from @tbrisker, I think we need to extract with N_(..) ?

ohadlevy (dismissed):
thanks @dLobatog !

Overall, getting there, but I would like to see more consistency with existing UINotification constructs.

Also, I want to fully understand if the changes to blueprint and notification objects are required.

this is the first time where the subject is not really an AR object, so I believe there are some existing gaps to solve.

thanks!

app/models/notification_blueprint.rb
dLobatog: The message is created completely from scratch on every notification, e.g: the summary of the blog entry, that does not have to rely on any subject.

lib/tasks/rss.rake
dLobatog: `CommunityBlog`? I meant the task to be flexible enough to support any backend

app/models/notification_blueprint.rb
ohadlevy: since uinotifiations can override the blueprint when creating the message, I'm not in favor of removing the validation. lets create some dummy sample instead of reducing the validation for the "edge" case.

lib/tasks/rss.rake
ohadlevy: It can be generic, but I don't think we should support any generic rss feed into foreman, foreman is not an rss reader :)

app/models/notification.rb
dLobatog: I cannot use the URL resolver here, the issue is this notification has no subject. I just want to put an arbitrary link. If I leave this as `=`, my link will get overridden by this (and not work, as there is no subject).

ohadlevy: I wonder if in this case we should extend the actions json attributes with additional data instead of hacking around subject? I understand that subject must be an AR object, but TBH I dont like the fact that we can't provide an official interface for plain ruby classes.

app/services/ui_notifications/rss_notifications_checker.rb
```diff
@@ -0,0 +1,50 @@
...
+          :actions => {
+            :links => [
+              {
+                :href => item.link.href,
+                :title => _('Link to blog')
+              }
+            ]
+          }
+        )
+        blueprint.save
```
ohadlevy: why are you saving the blueprint? blueprint should be created once (e.g. RSS Blog blueprint via seeds.) and multiple notifications based on the rss input.

I would also prefer to use create! so have an indication that it failed.

db/seeds.d/17-notification_blueprints.rb
```diff
@@ -32,6 +32,19 @@
...
+    name: 'rss_post',
+    level: 'info',
+    message: _('RSS post message goes here'),
+    actions:
+    {
+      links:
+      [
+        title: _('URL')
```
ohadlevy: please add external: true, so it will work with #4401

app/services/ui_notifications/rss_notifications_checker.rb
dLobatog: This is a leftover from the changes I made to adapt to UINotifications, the `.save`call won't do anything as the blueprint attributes aren't modified, but it's indeed useless so I'll remove it

db/seeds.d/17-notification_blueprints.rb
dLobatog: 'community' should be translated shouldn't it?

dLobatog:
@ohadlevy Rebased and updated the PR, thanks for taking a look!

lib/tasks/rss.rake
```diff
@@ -0,0 +1,42 @@
+namespace :rss do
```
dLobatog, line 1: This can probably go away if we rely on the initializer, or maybe we can leave it so other people can put their own blog RSS and schedule it with cron...

dLobatog (commented):
Updated for ActiveJob

app/services/ui_notifications/rss_notifications_checker.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+
+    private
+
+    def rss_notification_blueprint
+      NotificationBlueprint.unscoped.find_by_name('rss_post')
+    end
+
+    def notification_already_exists?(item)
+      !!Notification.unscoped.find_by_message(item.summary.content)
```
ohadlevy: do we know if content is always smaller than message? does it make sense to hash it instead?

app/services/ui_notifications/rss_notifications_checker.rb
```diff
@@ -0,0 +1,50 @@
...
+        notification = Notification.new(
+          :initiator => initiator,
+          :audience => @audience,
+          :message => item.title.content,
+          :notification_blueprint => blueprint,
+          :actions => {
+            :links => [
+              {
+                :href => item.link.href,
+                :title => _('Link to blog'),
```
ohadlevy: maybe should be Open blog ?

app/models/notification.rb
ohadlevy: is this really needed (as there is an if at the end of the line)?

ohadlevy (commented):
minor comments below - thanks!

app/services/ui_notifications/rss_notifications_checker.rb
dLobatog: It's the title of the RSS entry, unlikely to be very big. "Foreman Community Newsletter - July 2017" is an example. I don't think it's critical to hash it.

If `item.summary.content` is hashed in the notification, how would you display it in the view?

app/models/notification.rb
dLobatog: Yes, it's needed. If it's `=` , the blueprint actions:

```ruby
 :actions => {"links"=>[{"title"=>"URL"}]}
```

will override the actual actions created by `UINotifications::RssNotificationsChecker`

```ruby
actions => {
            :links => [
              {
                :href => item.link.href,
                :title => _('Open blog'),
                :external => true
              }
            ]
          }
```

dLobatog (commented):
@ohadlevy Sorry, forgot to ping back, I updated this a week ago, I have to check what's going on with tests though

app/services/ui_notifications/rss_notifications_checker.rb
```diff
@@ -0,0 +1,50 @@
+require 'rss'
+require 'date'
+
+module UINotifications
+  class RssNotificationsChecker
+    def initialize(options = [])
+      @url = options[:url] || 'https://theforeman.org/feed.xml'
```
ohadlevy: just wondering, are we going to get ssl verification errors/warnings ?

app/services/ui_notifications/rss_notifications_checker.rb
```diff
@@ -0,0 +1,50 @@
...
+      @latest_posts = options[:latest_posts] || 3
+      @force_repost = options[:force_repost] || false
+      @audience = options[:audience] || Notification::AUDIENCE_GLOBAL
+    end
+
+    def deliver!
+      feed = RSS::Parser.parse(@url, false)
```
ohadlevy: if it fails to reach out (lets say a proxy, network transient error etc) is it being logged some where? should we?

app/services/ui_notifications/rss_notifications_checker.rb
```diff
@@ -0,0 +1,50 @@
...
+      feed.items[0, @latest_posts].each do |item|
+        blueprint = rss_notification_blueprint
+        if notification_already_exists?(item)
+          next unless @force_repost
+        end
+        initiator = User.anonymous_admin
+        notification = Notification.new(
```
ohadlevy: nitpick, just use Notification.create!()..

config/initializers/rss_notifications.rb
```diff
@@ -0,0 +1,10 @@
+# First, we check if there's a job already enqueued for RSS notifications
+pending_jobs = ::Foreman::Application.dynflow.world.persistence.find_execution_plans(filters: { :state => 'scheduled' })
+scheduled_job = pending_jobs.select do |job|
+  delayed_plan = ::Foreman::Application.dynflow.world.persistence.load_delayed_plan job.id
+  next unless delayed_plan.present?
+  delayed_plan.to_hash[:serialized_args].first["job_class"] == 'CreateRssNotifications'
+end
+
+# Only create notifications if there isn't a scheduled job
+CreateRssNotifications.perform_later unless scheduled_job.present?
```
ohadlevy: is this the best way to load future / scheduled tasks? I was hoping this can be kept in somewhere and load all of them as one process.

Also, shouldnt this done on dynflow load vs Foreman load? /cc @ares and @iNecas

db/seeds.d/17-notification_blueprints.rb
ohadlevy: @tbrisker ?

ares:
@dLobatog  what's the status here?

config/initializers/rss_notifications.rb
dLobatog: Not sure about 'best way', but it's a way. As far as I know this instruction is still the same process as Foreman (it's just enqueing the job in Dynflow),

I don't think anything should be done on Dynflow load, I prefer to treat it as a background processing engine completely detached from Foreman, like Resque, Sidekiq, etc.. -

app/services/ui_notifications/rss_notifications_checker.rb
dLobatog: I don't see any, you can try to run `CreateRssNotifications.perform_now` on the console to verify

app/services/ui_notifications/rss_notifications_checker.rb
dLobatog: Good catch - I have included a rescue in the job itself for this situation, it will be logged to the background logger

app/services/ui_notifications/rss_notifications_checker.rb
dLobatog: Fixed

dLobatog (commented):
Thanks for the review @ohadlevy , I've updated the PR addressing your comments

ares (changes requested):
Tests failure seems related

UINotifications::RssNotificationsCheckerTest::when force_repost is enabled.test_0001_already existing notifications are not created

dLobatog (commented):
@ares Updated, the test was using a method that I renamed 😄

dLobatog:
[test foreman]
[test upgrade]
[test katello] 

Looks like there was a connection problem last time

dLobatog (commented):
Will trigger tests again, I don't think there's anything particular to this PR that would cause "db disconnected"...

dLobatog:
[test foreman]
[test katello]

ares:
it seems tests still fail on the same, could you try to rebase? are tests green in your setup?

app/jobs/create_rss_notifications.rb
```diff
@@ -0,0 +1,17 @@
+class CreateRssNotifications < ApplicationJob
+  after_perform do |job|
+    self.class.set(:wait => 12.hours).perform_later(job.arguments.first)
+  end
+
+  def perform(options = {})
+    # Defaults for theforeman.org blog RSS
+    UINotifications::RssNotificationsChecker.new(options).deliver!
+    raise StandardError.new('OHNO')
```
ares: is this some leftover from testing?

app/services/ui_notifications/rss_notifications_checker.rb
```diff
@@ -0,0 +1,48 @@
+require 'rss'
+require 'date'
+
+module UINotifications
+  class RssNotificationsChecker
+    def initialize(options = [])
+      @url = options[:url] || 'https://theforeman.org/feed.xml'
```
ares: would it make more sense to extract the default url to settings?

config/initializers/rss_notifications.rb
ares: it would be imho a good approach to have a separate service object - scheduler, that would plan other jobs and keep an eye on them, it would be regularly called using dynflow job but I think we can add it later when we have more repetitive jobs. This is good for now and and easy to replace later.

lib/tasks/rss.rake
```diff
@@ -0,0 +1,42 @@
...
+     FOREMAN_RSS_FORCE_REPOST, no new notifications will be created for these three posts. If you do,
+     the task will create notifications once again for "Newsletter December", November and October.
+ - FOREMAN_RSS_AUDIENCE
+   + The audience that will recieve the notification. By default, 'global'
+   + Possible values:
+     - 'global' - All users will receive the notification
+     - 'admin' - Only administrator users will receive the notification
+
+Examples:
+  # foreman-rake rss:create_notifications FOREMAN_RSS_AUDIENCE='admin'
```
ares, line 30: How is this supposed to be used? When I run this and restart Foreman, is it somehow persisted? Or do you rely on dynflow to keep it in it's DB? Is there a way to disable/remove it later without going to dynflow console?

ares (changes requested):
The initial commit mentions "Add option to puppet-foreman to choose whether to enable this or not. Make a noticeable warning if it's enabled so the operator knows this app will contact external sources (theforeman.org) by default". Is there this PR already? Also how is the configuration planned? Maybe I missed that but I don't see any configuration option to disable this job.

Is there a chance all above could be addressed before 1.17 branching? I'd make sure it's reviewed in time in such case.

app/jobs/create_rss_notifications.rb
dLobatog: Definitely, it was meant to test what happens when notifications fail, my bad.

lib/tasks/rss.rake
dLobatog: This can be removed - it's a way to force running the task, normally https://github.com/theforeman/foreman/pull/4240/files#diff-9d19612ef63c6c68d133845b585a29dcR1 takes care of running and scheduling the task, but if the user wants to run it more often, they can run this command.

dLobatog: Also they can put this in a cronjob to create notifications for other RSS feeds, not just Foreman's..

app/services/ui_notifications/rss_notifications_checker.rb
dLobatog: It could be possible to do that but it hasn't ever changed

dLobatog (commented):
@ares Good point, there is no way to disable the RSS notifications for the blog at the moment. 

I think this could be done by adding a setting, which could be used by puppet-foreman too.

Updated, I rebased and removed the `raise StandardError`

lib/tasks/rss.rake
ares: yeah, that's fine, for some reason I thought this was scheduling the repeated task but I see it's just one shot

ares:
After a long battle I made this work :-) the problem is with the `global` audience which is broken in Foreman. That's not related to this PR, however that made me wonder, is global the right audience? The user with limited permission might not be interested in fact that there's new version of Foreman available. Should we only generate notifications for admins?

ares (changes requested):
test failures at least on mysql looks related, could you please have a look? also see the comment about audience, last but not least, as we discussed it would be good to have settings for default url and disabling/enabling this feature

ares:
I opened a PR with the fix for global audience at  - https://github.com/theforeman/foreman/pull/5032

dLobatog:
@ares Blog posts are not only about new versions published. Other useful things for all kinds of users are published, e.g: newsletter, new features, case studies..

ares:
> @ares Blog posts are not only about new versions published. Other useful things for all kinds of users are published, e.g: newsletter, new features, case studies..

Ok, I still have doubts whether all users will be interested, but IIRC we don't have setups with many users. And in the worst case, it should be possible to disable it entirely. We can always change the audience later based on feedback. So to sum up what's I think pending

- [] Setting for disable/enable
- [] Setting for default url
- [] make tests green

and then it's good to go

dLobatog (commented):
Updated @ares , now there's a setting for URL and another for enabling RSS feed notifications. The task gets enqueued in Dynflow regardless of this setting. 

This is on purpose, so that creating RSS notifications is a `noop` until the moment `Setting[:rss_enable]` becomes true. It's much easier than killing the job in Dynflow and scheduling a new one every time the setting is flipped - that's more complex and I don't think there's much to gain vs a noop job.

ares:
There are still some tests failures, see http://ci.theforeman.org/job/test_develop_pr_core/15229/database=sqlite3,ruby=2.2,slave=fast/testReport/junit/(root)/UINotifications__RssNotificationsCheckerTest__when%20force_repost%20is%20disabled/test_0001_already_existing_notifications_are_not_created/

dLobatog (commented):
@ohadlevy  @ares Fixed the test failures, they were caused by accessing `Foreman::Application.dynflow.world` more than once, I started using `::Foreman::Application.dynflow.config.on_init` as per @iNecas recommendation to avoid this problem, looks like everything's 💚  now

app/jobs/create_rss_notifications.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  def perform(options = {})
+    # Defaults to theforeman.org blog RSS
+    UINotifications::RssNotificationsChecker.new(options).deliver!
+  end
+
+  rescue_from(StandardError) do |exception|
+    Foreman::Logging.logger('background').error(
+      'RSS notification checker: '\
+      "Error while creating notifications #{exception}")
```
ares: either use `Foreman::Logging.exception` or log the backtrace please

ares (changes requested):
One improvement for logging and I think we're good to go.

dLobatog (commented):
Thanks for the comments @ares, updated to log the backtrace & with the `Item` class.

ares (approved):

ares:
good to be merged when foreman tests run confirms

dLobatog:
@ares  🍏

ares:
Thanks @dLobatog, merged.

app/jobs/create_rss_notifications.rb
```diff
@@ -0,0 +1,16 @@
...
+  end
+
+  def perform(options = {})
+    # Defaults to theforeman.org blog RSS
+    UINotifications::RssNotificationsChecker.new(options).deliver!
+  end
+
+  rescue_from(StandardError) do |error|
+    Foreman::Logging.logger('background').error(
+      'RSS notification checker: '\
```
ohadlevy, line 13: whats the reason for the multiline? I assume its not multiline in the log?

app/models/notification.rb
ohadlevy: TBH I'm still not sure i follow this change - I'm still concerned that this is a "hack".

app/services/ui_notifications/rss_notifications_checker.rb
```diff
@@ -0,0 +1,74 @@
...
+      # gets enabled, notifications for RSS feeds are created again
+      return true unless Setting[:rss_enable]
+      feed = RSS::Parser.parse(@url, false)
+      feed.items[0, @latest_posts].each do |feed_item|
+        item = Item.new(feed_item)
+        blueprint = rss_notification_blueprint
+        if notification_already_exists?(item)
+          next unless @force_repost
+        end
+        Notification.create(
```
ohadlevy, line 46: I would create! so that exception is handled / logged somewhere.

app/services/ui_notifications/rss_notifications_checker.rb
```diff
@@ -0,0 +1,74 @@
...
+            ]
+          }
+        )
+      end
+    end
+
+    private
+
+    def rss_notification_blueprint
+      NotificationBlueprint.unscoped.find_by_name('rss_post')
```
ohadlevy, line 67: AFAIR find_by_ methods are deprecated?

app/models/notification.rb
ares: the condition is `if notification_blueprint.actions.any?` which means it would be always overridden if blueprint defines some default actions

app/services/ui_notifications/rss_notifications_checker.rb
ares: I don't think they are deprecated as long as the dynamic part is name of the column. `find_or_create_by_name` was deprecated some time ago. See http://guides.rubyonrails.org/active_record_querying.html#dynamic-finders and http://api.rubyonrails.org/classes/ActiveRecord/Base.html scroll down to "Dynamic attribute-based finders"

app/jobs/create_rss_notifications.rb
dLobatog: Being able to stick to 80 chars & read stuff nicely in vim/tmux 😄

app/services/ui_notifications/rss_notifications_checker.rb
dLobatog: https://github.com/theforeman/foreman/pull/4240/files#diff-bbe58f068bd900b15ec6debded3fbcf8R8 - the `rescue_from` in the job logs any possible error from here

app/models/notification.rb
ohadlevy: @ares I understand the condition works, my concern is with how the actions is be used, I'm just wondering if this is inline with the original design of the notification actions (being part of a blueprint). I'm more than ok to review the notification implementation and to see if that makes sense or not, but was concerned with adding this without a good understanding on the impact.

app/services/ui_notifications/rss_notifications_checker.rb
ares: @dLobatog the thing is that create fails silently returning false, not raising an exception, with `create!` we'd be able to find in logs that it failed and why

app/models/notification.rb
ares: Ok, so what was the idea behind blueprint and actions? from the code I understood that blueprint is a notification template but you can override specific attributes. Attribute message already had `||=` therefore it felt like a bug that actions behave differently.

ares: @ohadlevy I know you're busy, but is there a chance you can take a look at this? I don't want it to slip off from my radar but the older it gets the bigger chance it will.

PR #4241, opened 2017-01-29
Fixes #14716 - Added option to transform thin template to thick disk

ShimShtein:
Unfortunately, it's a single setting for all disks - you can't choose which disks (inherited from template) will be thin and which not.

For new disks, the logic remains the same - "thin" flag would be used.

### DIFF
--- a/app/models/compute_resources/foreman/model/vmware.rb
+++ b/app/models/compute_resources/foreman/model/vmware.rb
@@ -423,6 +423,8 @@ module Foreman::Model
         "resource_pool" => [args[:cluster], args[:resource_pool]]
       }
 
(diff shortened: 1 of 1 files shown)


ShimShtein:
Message length fixed

dLobatog:
Merging - although @ShimShtein you probably want to look at @timogoebel & @shlomizadok 's PR on the reimplementation of the storage UI, it might be possible to add support for thin/thick on individual disks there. https://github.com/theforeman/foreman/pull/3979

dLobatog:
@ShimShtein thanks!

ShimShtein:
@dLobatog the problem is that the restriction is on the API level - from reading VMWare's API (the one that is wrapped by fog), there is no place where we can define this kind of transformation. VMWare ignores the "thin" flag for disks coming from templates (which is sent correctly BTW), and the only way to influence this behavior was to use this global `transform` flag.

PR #4242, opened 2017-01-30
Fixes #17463 - Detect user's orgs when importing from puppet

xprazak2:


app/services/puppet_class_importer.rb
```diff
@@ -182,6 +182,10 @@ def old_environments
     db_environments - actual_environments
   end
 
+  def current_user_environments
+    Environment.where(:id => Organization.where(:id => User.current.organization_ids).map { |org| org.environment_ids }.reduce(:+)).map(&:to_s)
```
ares: Please don't use `User.current.organization_ids` directly, use [Organization.my_organizations](https://github.com/theforeman/foreman/blob/develop/app/models/taxonomies/organization.rb#L18) instead, it will respect admin exception and nesting of Organizations.

ares (changes requested):
Don't we have the same problem for Locations?

xprazak2:
Made changes to use  `Organization.my_organizations` and handle the same case for locations as well.

app/services/puppet_class_importer.rb
```diff
@@ -182,6 +182,13 @@ def old_environments
     db_environments - actual_environments
   end
 
+  def current_user_environments
+    ids = [Organization.my_organizations, Location.my_locations].flatten.map do |taxonomy|
+      taxonomy.environment_ids
+    end
+    Environment.where(:id => ids.reduce(:+).uniq).map(&:to_s)
```
tbrisker: How about 
```ruby
Environment.unscoped  #to get rid of the default taxonomy scope
           .authorized #to make sure we aren't leaking data
           .joins(:taxable_taxonomies)
           .where('taxable_taxonomies.taxonomy_id' => Organization.my_organizations.pluck(:id) + Location.my_locations.pluck(:id))
           .distinct
           .pluck(:name)
```

ares: If you use authorized scope, please always specify permission which you want user to be authorized for.

xprazak2:
I improved the query as suggested

dLobatog:
@xprazak2 I'm able to reproduce and fix the problem with your patch - can you check the tests though? It seems that some tests were affected after this, can you fix them and write a couple that check that the problem is fixed? Thanks!

(the katello failures are unrelated)

app/services/puppet_class_importer.rb
```diff
@@ -171,7 +171,28 @@ def db_environments
...
+
+  def proxy_environments
+    proxy.environments.map(&:to_s)
+  end
+
+  def to_be_created_environments
+    proxy_environments - Environment.where(:name => proxy_environments).pluck(:name)
+  end
+
+  def current_user_taxonomy_ids
```
dLobatog: I would say this and the method below really belong in the user model, this logic only related to fetching information from User:

```
class User
  def taxonomy_ids
    { :organizations => my_organizations.pluck(:id), :locations => my_locations.pluck(:id) }
  end
 
  def visible_environments
        Environment.unscoped
               .authorized(:view_environemnts)
               .joins(:taxable_taxonomies)
               .where('taxable_taxonomies.taxonomy_id' => taxonomy_ids[:organizations] + taxonomy_ids[:locations])
               .distinct
               .pluck(:name)
  end
end
```

test/controllers/api/v2/smart_proxies_controller_test.rb
```diff
@@ -102,12 +102,14 @@ def setup_import_classes
       Puppetclass.destroy_all
       Environment.destroy_all
     end
+    orgs = [taxonomies(:organization1)]
```
dLobatog: This code is duplicated on both API controllers, can you extract it? (see test/controllers/shared/ for examples)

dLobatog:
@ares The problem is now fixed on locations too. I left a couple of minor comments, I wouldn't block the PR from being merged if you think it's ready as it is now. 

@xprazak2 Thank you!

app/services/puppet_class_importer.rb
ares: I agree, that would also fix part of http://projects.theforeman.org/issues/16128

ares (approved):

ares (commented):
Thanks for fixing locations. Good from my side after comments from @dLobatog are resolved.

xprazak2:
I made changes as suggested.

dLobatog:
Thanks @xprazak2 !

PR #4244, opened 2017-01-31
Fixes #18321 - added migration for builtin roles

lzap:
We fixed this in code but existing incorrectly cloned roles stays undeletable.


tbrisker:
Thanks @lzap !

PR #4248, opened 2017-02-01
Fixes #18348 - foreman-debug compress faster

lzap:
When testing make sure you don't measure the difference on a clean install, on
1 MB tarball there is barely a difference. Test with 100+ MB tarball.


dLobatog:
Thanks @lzap !

PR #4249, opened 2017-02-02
fixes #18366 - vmware: delete vm only if created with errors

timogoebel:
If there were errors when deploying the vm, the `vm` object exists but `vm.id` is `nil`. Trying to delete a VM with name `nil` causes an exception.


dLobatog:
Thanks @timogoebel  - I've seen this problem in other compute resources too 😥

PR #4250, opened 2017-02-02
Fixes #18367 - Add patternfly welcome to user groups index

shlomizadok:


dLobatog:
Merged as 878ecc3a79659aabfb57cf78eb9015167135d2b8 , notice I changed the icon and removed the html_safe calls that I believe were not needed. Thanks @shlomizadok !

PR #4251, opened 2017-02-02
fixes #18373 - vmware: do not pass automatic firmware to vm

timogoebel:
`vm_instance_defaults` sets `firmware` to `automatic` as this should be the default value shown in the UI. The default value may be overwritten by a user. It it's unset, the default (and invalid) value of `automatic` may be passed to fog for vm creation. This commit prohibits this by defaulting to `bios`.

### DIFF
--- a/app/models/compute_resources/foreman/model/vmware.rb
+++ b/app/models/compute_resources/foreman/model/vmware.rb
@@ -398,6 +398,7 @@ module Foreman::Model
         clone_vm(args)
       else
         vm = new_vm(args)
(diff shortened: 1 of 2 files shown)


dLobatog:
@timogoebel Check out the broken tests - I triaged the bug and patch and can confirm it works. Thanks!

timogoebel:
@dLobatog : Fixed the tests and added a new one for this issue.

dLobatog:
Thanks @timogoebel

PR #4253, opened 2017-02-04
Fixes #12054 - Openstack v3 support

achevalet:
I am new in ruby so it's probably not perfect but it could be a good start, at least it works with my environments both v2 and v3


achevalet:
I was not able to rebase.. closed and reopened at https://github.com/theforeman/foreman/pull/4254

PR #4255, opened 2017-02-05
Fixes #18364 - disable hide of dropdown

matanwerbner:


orrabin:
ok to test

gailsteiger:
Not related to this PR but why is this an 'erb' file. Is this common in the project?

Otherwise, looks good.

ohadlevy:
@gailsteiger it is not common.. git log would be your friend :)

app/assets/javascripts/host_checkbox.js.erb
```diff
@@ -42,10 +42,10 @@ function readFromCookie() {
 function toggle_actions() {
   var dropdown = $("#submit_multiple a");
   if ($.foremanSelectedHosts.length == 0) {
-    dropdown.addClass("disabled hide");
+    dropdown.addClass("disabled");
```
ohadlevy: does it make sense to have a pop over / hover on the btn when its disabled? - telling the user that he needs to select at least one host?

app/assets/javascripts/host_checkbox.js.erb
```diff
@@ -40,13 +40,17 @@ function readFromCookie() {
 }
 
 function toggle_actions() {
-  var dropdown = $("#submit_multiple a");
+  var dropDownContainer = $("#submit_multiple");
+  var dropdown = dropDownContainer.find("a");
+  var disabledMessage = __("Please select hosts to perform action on.");
```
ohadlevy, line 45: would it be easier to add the title when creating the action in the first place? e.g. at https://github.com/theforeman/foreman/blob/develop/app/helpers/hosts_helper.rb#L175 ?

matanwerbner: we will still need to keep the code at host_checkbox.js, since it handles removing and adding the title when hosts are selected/unselected (we dont want the title for an active dropdown)

ohadlevy:
[test katello]

ohadlevy (approved):

ohadlevy:
katello test failure seems unrelated, re-running and will merge once :green_apple:

ohadlevy:
Katello test failure seems unrelated - thanks @matanwerbner

PR #4256, opened 2017-02-06
fixes #18248 - dashboard auto refresh mounts duplicate components

gailsteiger:
This is an unfortunate side-effect of turbolinks.

The alternatives are 

1. not to refresh the page
2. refresh the page without using Turbolinks.visit

This PR implements the second option.

### DIFF
--- a/app/views/home/_topbar.html.erb
+++ b/app/views/home/_topbar.html.erb
@@ -14,7 +14,6 @@
           </ul>
           <% if SETTINGS[:login] %>
             <div id='notification_drawer'> </div>
-            <%= mount_react_component('NotificationDrawer', '#notification_drawer', {}) %>
           <% end %>
         </div>
       <% end %>
@@ -48,3 +47,5 @@
     <% end %>
   </div>
 </div>
+
+<%= mount_react_component('NotificationDrawer', '#notification_drawer', {}) %>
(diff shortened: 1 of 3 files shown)


app/views/home/_topbar.html.erb
```diff
@@ -48,3 +47,5 @@
     <% end %>
   </div>
 </div>
+
+<%= mount_react_component('NotificationDrawer', '#notification_drawer', {}) %>
```
gailsteiger, line 51: code moved to bottom to separate the script code from the html

app/views/home/_user_dropdown.html.erb
```diff
@@ -13,3 +12,5 @@
     <% end %>
   </ul>
 </li>
+
+<%= mount_react_component('NotificationDrawerToggle', '#notification_icon', {:url => notification_recipients_path}.to_json ) %>
```
gailsteiger, line 16: code moved to bottom to separate the script code from the html

dLobatog:
Thanks @gailsteiger - it took me actually a long time to review this as I wanted to see whether it was necessary to have it on Turbolinks 5 / or Turbo-React (https://github.com/ssorallen/turbo-react). Seems that in both cases it's necessary - so let's get this in

gailsteiger:
Thanks @dLobatog

PR #4257, opened 2017-02-07
Fixes #18410 - log permission denials

iNecas:
I've added logging of reason for permissions denials + changed the level
in UI logging from debug to info, as I believe this information is on
different level than other debug info from permissions.

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -273,8 +273,10 @@ module Api
     def set_error_details(error, options)
       case error
       when 'access_denied'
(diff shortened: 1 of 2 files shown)


dLobatog:
@iNecas Thanks - will fix the issue number on merge.

iNecas:
Thanks @dLobatog

lzap:
Thank you, we all own you :beer: ! I'd perhaps go for warning instead of info (because it really is either misconfiguration or half-incident), but info is fine for sure :-)

PR #4260, opened 2017-02-08
Fixes #18381 - PXE loader attr load in host/hostgroup form

swapab:
Override pxe_loader attribute to fail safe find nested pxe_loader from parent hostgroup. Host#pxe_loader needed override as similar behavior was observed in new host form.

### DIFF
--- a/app/controllers/hostgroups_controller.rb
+++ b/app/controllers/hostgroups_controller.rb
@@ -98,6 +98,7 @@ class HostgroupsController < ApplicationController
     define_hostgroup
     inherit_parent_attributes
     load_vars_for_ajax
+    reset_explicit_attributes
 
     render :partial => "form"
   end
@@ -154,5 +155,9 @@ class HostgroupsController < ApplicationController
     @hostgroup.subnet             ||= @parent.subnet
     @hostgroup.realm              ||= @parent.realm
     @hostgroup.environment        ||= @parent.environment
+  end
+
+  def reset_explicit_attributes
+    @hostgroup.pxe_loader = nil if @parent.present?
   end
 end
--- a/app/helpers/form_helper.rb
+++ b/app/helpers/form_helper.rb
@@ -332,7 +332,8 @@ module FormHelper
 
   def blank_or_inherit_f(f, attr)
     return true unless f.object.respond_to?(:parent_id) && f.object.parent_id
-    inherited_value   = f.object.send(attr).try(:name_method)
+    inherited_value   = f.object.send(attr)
+    inherited_value   = inherited_value.name_method if inherited_value.present? && inherited_value.respond_to?(:name_method)
     inherited_value ||= _("no value")
     _("Inherit parent (%s)") % inherited_value
   end
--- a/app/models/host/managed.rb
+++ b/app/models/host/managed.rb
@@ -750,6 +750,14 @@ class Host::Managed < Host::Base
     read_attribute(:provision_method) || capabilities.first.to_s
   end
 
+  def explicit_pxe_loader
+    read_attribute(:pxe_loader).presence
+  end
+
+  def pxe_loader
+    explicit_pxe_loader || hostgroup.try(:pxe_loader)
+  end
+
   def image_build?
     self.provision_method == 'image'
   end
--- a/app/models/hostgroup.rb
+++ b/app/models/hostgroup.rb
@@ -181,6 +181,14 @@ class Hostgroup < ApplicationRecord
     Setting[:root_pass]
   end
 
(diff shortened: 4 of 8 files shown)


lzap:
Thanks, it fixes the issue but when I create hg A with some loader setting and then I click on Nest button, it does explicitly set it instead of Inherited value (The value), is this expected behavior?

swapab:
@lzap Yes, it is showing the PXE Loader value explicitly now. Is it mandatory to show it like `Inherited value(the value)`?
If yes, I'll revisit this and try to bring back the previous behavior(along with the fix).

lzap:
I don't mind this behavior, it's just different. Is there a reason for that? @ares ?

ares:
I'd prefer to have the inheritance as for other fields unless there's some technical limitation.

swapab:
@lzap @ares PR has been updated to the original behavior of showing `Inherit parent (<pxe-loader>)`

lzap:
Looks good but this does not work:

Create hostgroup A, select Grub2 UEFI for example, save.
Create nested hg B, select Grub1, save.
Now edit B again, select Inherit parent (Grub1) - but this should be Inherit parent (Grub2 UEFI).

Perhaps add a test for that? The one there is there did not catch this.

swapab:
@lzap Thanks for catching this :+1: 

Q: What should be the behavior in the following case:
1. Create Hostgroup-A, select Grub2 UEFI save
2. Create _independent_ Hostgroup-B, select Grub1 save
3. Edit Hostgroup-B, set parent Hostgroup-A, goto Operating systems -> PXE Loader shows 'Grub1'.
4. Dropdown has value 'Inherited parent (Grub2 UEFI)'

step#3 PXE Loader should be auto selected to step#4?
On edit, this appears to be a bug but looks like `FormHelper#blank_or_inherit_f` is not supposed to handle this, dropdown had already got its value. I think we need to reset the dropdown so that, include_blank plays it's role.

swapab:
A point to mention, after saving on step#3 the functionality might be working as expected, as the parent hostgroup relation is achieved. This is the UI part that should show what actually is.

One fix could be, let's not reset the drop-down instead reset `pxe_loader` on the server side while updating(since parent hostgroup is there).
**Edit:** This would correct the UI on next load.

lzap:
There are test failures.

swapab:
@lzap Tests are green.

Q: https://github.com/theforeman/foreman/pull/4260#issuecomment-288764905 do you think this is a valid use-case/scenario?

To summarize, shall we reset hostgroup.pxe_loader value if parent_id is present? Does an explicitly set pxe_loader has any effect even if hostgroup has a parent(with a valid parent_hg.pxe_loader)?

lzap (approved):
This is expected behavior, thanks.

lzap:
I'd like to have extra look by @ares as I would like to propose this into 1.15, looks good, works as expected. Any concerns?

ares:
It does not seem to work for me when I have multiple levels of inheritance. E.g. host groups A -> B -> C, B correctly inherits from A but C displays `Inherit parent ()`. Other values, such as architecture, seems to work fine.

lzap:
Can you add a test for three levels of inheritance please? Good catch thanks.

swapab:
@ares Thanks for catching this. I have used attribute `selected` to fix multi-level inheritance issue.

lzap (changes requested):
Unfortunately, on Hostgroup edit, I am not able to save nested hostgroup with explicitly set flag. It always falls back to Inherited parent no matter what I set. If possible, make a test for this.

swapab:
@lzap Fixed the explicit use case https://github.com/theforeman/foreman/pull/4260#issuecomment-292926741 and added a js test https://github.com/theforeman/foreman/pull/4260/files#diff-5a2960f2e24206bf7cca4c8b3b029e40

lzap (approved):

lzap:
Thanks it works fine, althought there is one wish I have - when you create new host without any hostgroup set, currently PXE loader is blank, I wish it was set to PXELinux (BIOS) which was the default behavior before I introduced PXE loader. In the original patch this was never implemented, but I wish we had this. If you happen to find some time to work on this as well, that would be great. Anyway good patch, thanks a lot!

swapab:
@lzap Thanks for merging the patch :smiley: 
I'll take a note of the wish. It does make a lot of sense to have default `pxe_loader`. I am yet to understand 'behind the scene' part of `pxe_loader` but I have created a new ticket to address it : http://projects.theforeman.org/issues/19327

PR #3541, opened 2016-05-23
Refs #13741 - Hidden values use masked input for default value

tbrisker:
This was missed in the pervious commit. Opening a hidden lookup key led
to the default value being displayed using a password input instead of
as a masked text area.

PR #3269, opened 2016-03-02
Fixes #3541,#13769 - make sure the ssh orchestration methods return true on success

iNecas:
Strange things start happening otherwise.

PR #4262, opened 2017-02-08
Fixes #3541,#13769 - make sure the ssh orchestration methods return true on success

iNecas:
Strange things start happening otherwise.

Less controversal version after https://github.com/theforeman/foreman/pull/3269
not making it in.

### DIFF
--- a/app/models/concerns/orchestration/ssh_provision.rb
+++ b/app/models/concerns/orchestration/ssh_provision.rb
@@ -64,12 +64,15 @@ module Orchestration::SSHProvision
     self.handle_ca
     return false if errors.any?
     logger.info "Revoked old certificates and enabled autosign"
+    true
   end
 
   def delSSHCert
     # since we enable certificates/autosign via here, we also need to make sure we clean it up in case of an error
     if puppetca?
       respond_to?(:initialize_puppetca,true) && initialize_puppetca && delCertificate && delAutosign
+    else
+      true
     end
   rescue => e
     failure _("Failed to remove certificates for %{name}: %{e}") % { :name => name, :e => e }, e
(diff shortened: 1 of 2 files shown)


ares (approved):
Fixes the problem and refactors the condition in delSSHCert, :+1:

ares (changes requested):
actually a I'm afraid I found another combination where some step fails if log level is set to error, investigating

app/models/concerns/orchestration/ssh_provision.rb
```diff
@@ -80,7 +81,9 @@ def setSSHProvision
     if client.deploy!
       # since we are in a after_commit callback, we need to fetch our host again, and clean up puppet ca on our own
       Host.find(id).built
-      respond_to?(:initialize_puppetca,true) && initialize_puppetca && delAutosign if puppetca?
+      if puppetca? && respond_to?(:initialize_puppetca, true)
```
ares, line 86: when I deploy host without puppet ca, this fails since this is the last expression and it's missing

```ruby
else
  true
end
```

ares (changes requested):

app/models/concerns/orchestration/ssh_provision.rb
iNecas: Ah, I assumed we ignore when returning nil (and only false is significant), but you're right, updated

ares (approved):

ares:
Foreman tests failure unrelated, thanks @iNecas, merged as a4e1e08ceb388e9749fceee8ae4487ab3c76c105

lzap:
I was wondering, how about to insist on a return value of some type in the orchestration code. We'd need to refactor all orchestration return values, but that is easy change. I see this kind of bug for second time now, it had to be difficult to track. Any comments? Shall I make a refactor ticket for that?

iNecas:
@lzap I think it would make sense: the biggest obstacle I see is you need a buy-in from multiple people here, and especially as it's compatibility breaking change, it might be a challenged. But you initial :+1 from me at least.

PR #4263, opened 2017-02-08
Fixes #18433 - Remove the Data not available message.

bkearney:
This is causing confusion, as it appears to infer an error state


dLobatog:
Thanks @bkearney  - I encountered this with the Ansible and Dynflow smart proxy plugins and it was kind of confusing indeed

PR #4265, opened 2017-02-09
Fixes #18426 - Add actions to notification entries

dLobatog:
This adds support for the notifications to display actions. In case no
actions are provided, the dropdown kebab isn't shown. The actions
possible right now only include GET links.

screenshot from 2017-02-09 12-16-25
screenshot from 2017-02-09 12-16-44

### DIFF
--- a/app/models/notification_recipient.rb
+++ b/app/models/notification_recipient.rb
@@ -17,6 +17,7 @@ class NotificationRecipient < ActiveRecord::Base
       :subject    => notification_blueprint.subject,
       :created_at => notification.created_at,
       :group      => notification_blueprint.group,
+      :actions    => notification_blueprint.actions
     }
   end
 
--- a/webpack/assets/javascripts/react_app/components/notifications/Notification.js
+++ b/webpack/assets/javascripts/react_app/components/notifications/Notification.js
@@ -1,11 +1,12 @@
 import React from 'react';
 import Icon from '../common/Icon';
 import moment from 'moment';
+import NotificationDropdown from './NotificationDropdown';
 import NotificationActions from '../../actions/NotificationActions';
 import '../../common/commonStyles.css';
 
 /* eslint-disable camelcase */
-const Notification = ({created_at, seen, text, level, id}) => {
+const Notification = ({created_at, seen, text, level, id, actions}) => {
   const created = moment(created_at);
   const title = __('Click to mark as read').toString();
   const tooltip = {
@@ -27,6 +28,7 @@ const Notification = ({created_at, seen, text, level, id}) => {
   return (
     <div className="drawer-pf-notification">
       <Icon type={level} css="pull-left"></Icon>
+      {actions.links && <NotificationDropdown links={actions.links} id={id} />}
       {markup}
       <div className="drawer-pf-notification-info">
         <span className="date">{created.format('M/D/YY')}</span>
--- a/webpack/assets/javascripts/react_app/components/notifications/NotificationDropdown.js
+++ b/webpack/assets/javascripts/react_app/components/notifications/NotificationDropdown.js
@@ -1,0 +1,28 @@
+import React from 'react';
+import { Dropdown, Glyphicon, MenuItem } from 'react-bootstrap';
+
+const NotificationDropdown = ({ links, id }) => {
+  const listLinks = links.map((link, i) => {
+    const liKey = `notification-link-${i}-${id}`;
+
+    return (
+      <MenuItem key={liKey} id={liKey} target="_blank" href={link.href}>
+        {link.title}
+      </MenuItem>
+    );
+  });
+
+  return (
+    <Dropdown className="pull-right dropdown-kebab-pf" pullRight key={id}
+              id={`notifications-dropdown-${id}`}>
+      <Dropdown.Toggle noCaret bsStyle="link">
+        <Glyphicon bsClass="fa" glyph="ellipsis-v" />
+      </Dropdown.Toggle>
+      <Dropdown.Menu>
+        {listLinks}
+      </Dropdown.Menu>
+    </Dropdown>
+  );
+};
+
+export default NotificationDropdown;
--- a/app/models/notification_blueprint.rb
+++ b/app/models/notification_blueprint.rb
@@ -9,6 +9,7 @@
 class NotificationBlueprint < ActiveRecord::Base
   has_many :notifications
   belongs_to :subject, :polymorphic => true
+  store :actions, :accessors => [:links], :coder => JSON
 
   validates :message, :presence => true
   validates :group, :presence => true
--- a/db/migrate/20170209084517_add_actions_to_notification_blueprint.rb
+++ b/db/migrate/20170209084517_add_actions_to_notification_blueprint.rb
@@ -1,0 +1,5 @@
+class AddActionsToNotificationBlueprint < ActiveRecord::Migration
+  def change
+    add_column :notification_blueprints, :actions, :text
+  end
+end
--- a/webpack/assets/javascripts/react_app/components/notifications/Notification.test.js
+++ b/webpack/assets/javascripts/react_app/components/notifications/Notification.test.js
@@ -9,7 +9,15 @@ const notification = {
   id: 1,
   text: 'Job well done',
   level: 'success',
-  created_at: '2016-12-13 16:52:47Z'
+  created_at: '2016-12-13 16:52:47Z',
+  actions: {
+    'links': [
+      {
+        'href': 'https://theforeman.org',
+        'title': 'Foreman blog'
+      }
+    ]
+  }
 };
 /* eslint-enable camelcase */
 
@@ -18,6 +26,8 @@ function setup(notification) {
 }
 
 describe('Notification', () => {
+  let wrapper;
+
   beforeEach(() => {
     global.__ = (text) => text;
     global.tfm = {
@@ -25,29 +35,40 @@ describe('Notification', () => {
         activateTooltips: () => {}
       }
     };
+    wrapper = setup(notification);
   });
 
   it('displays text', () => {
-    const wrapper = setup(notification);
     const messageElement = wrapper.find('.drawer-pf-notification-message');
 
     expect(messageElement.text()).toBe('Job well done');
   });
   it('displays icon', () => {
-    const wrapper = setup(notification);
     const iconElement = wrapper.find('.pficon.pficon-ok.pull-left');
 
     expect(iconElement.length).toBe(1);
 
   });
   it('displays created date', () => {
-    const wrapper = setup(notification);
     const dateElement = wrapper.find('.date');
 
     expect(dateElement.text()).toBe('12/13/16');
   });
+  it('display actions dropdown if links are provided', () => {
+    const dropdownMenu = wrapper.find('a');
+
+    expect(dropdownMenu.props().href).toBe('https://theforeman.org');
+    expect(dropdownMenu.text()).toBe('Foreman blog');
+  });
+  it('does not display actions dropdown if links are NOT provided', () => {
+    let notificationWithoutAction = notification;
+
+    delete notificationWithoutAction.actions.links;
+    const dropdownMenu = setup(notificationWithoutAction).find('a');
+
+    expect(dropdownMenu.node).not.toBeDefined();
+  });
   xit('displays created time', () => {
-    const wrapper = setup(notification);
     const timeElement = wrapper.find('.time');
 
     expect(timeElement.text()).toBe('06:52:47 PM');


ohadlevy:
@gailsteiger or @matanwerbner can you have a look please?

webpack/assets/javascripts/react_app/components/notifications/Notification.js
```diff
@@ -18,6 +18,33 @@ const Notification = ({created_at, seen, text, level, id}) => {
...
+
+      links.push((<li key={li_key} id={li_key}>
+                    <a target="_blank" href={actions.links[i].href}>
+                    {actions.links[i].title}</a>
+                  </li>
+                 ));
+    }
+
+    actions_dropdown = (
+      <div className="dropdown pull-right dropdown-kebab-pf">
```
ohadlevy: does it make sense to make kebab a component by it self? i can see us using it in many different places

webpack/assets/javascripts/react_app/components/notifications/Notification.js
```diff
@@ -18,6 +18,33 @@ const Notification = ({created_at, seen, text, level, id}) => {
...
+        <button className="btn btn-link dropdown-toggle" type="button" id="dropdownKebabRight{id}"
```
ohadlevy: maybe it make sense to use https://react-bootstrap.github.io/components.html#btn-dropdowns ?

webpack/assets/javascripts/react_app/components/notifications/Notification.js
```diff
@@ -18,6 +18,33 @@ const Notification = ({created_at, seen, text, level, id}) => {
     (<strong {...tooltip} className="drawer-pf-notification-message pointer"
              onClick={markAsRead}>{text}</strong>);
 
+  const links = [];
+
+  let actions_dropdown = '';
+
```
matanwerbner: This component may best be put on it's own as a seperate component, accepting a list of links, and then on the parent's render:
{ actions.links && `<ActionsDropdown  links={actions.links } />`   }
that way we dont inflate the notification component too much, and dont have to deal with different types of notifications

webpack/assets/javascripts/react_app/components/notifications/Notification.js
```diff
@@ -18,6 +18,33 @@ const Notification = ({created_at, seen, text, level, id}) => {
...
+  if (actions.links != null) {
+    for (let i = 0; i < actions.links.length; i++) {
```
matanwerbner: `const links = actions.links.map((link, index) => {
  const li_ley = `notifcation-link${index}-${id}`;
  return <li key={li_key} id={ li_ley }>
                 <a target="_blank" href={link.href}>
                      {link.title}
                 </a></li>)`

matanwerbner: it might be nicer to use array.map 
`const links = actions.links.map((link, index) => {

})`

webpack/assets/javascripts/react_app/components/notifications/Notification.js
matanwerbner: agree. also, shouldn't there be a onClick handler here?

webpack/assets/javascripts/react_app/components/notifications/Notification.js
matanwerbner: i think we can easily implement this component ourself using a stateful component, more reason for it to be on it's own

webpack/assets/javascripts/react_app/components/notifications/Notification.js
matanwerbner: i think we can implement this component ourself using a stateful React class, more reason for it to be on it's own

webpack/assets/javascripts/react_app/components/notifications/Notification.js
gailsteiger: +1 for using `map`

webpack/assets/javascripts/react_app/components/notifications/Notification.js
gailsteiger: We should implement the [kebab](http://www.patternfly.org/pattern-library/widgets/#kebabs) as a stand alone component using the patternfly markup as a reference. We probably should remove the bootstrap js functionality and replace with a pure react implementation.

webpack/assets/javascripts/react_app/components/notifications/Notification.js
gailsteiger: yes

webpack/assets/javascripts/react_app/components/notifications/Notification.js
gailsteiger: yes

webpack/assets/javascripts/react_app/components/notifications/Notification.js
matanwerbner: thinking of it, the kebab is just an icon, so perhaps we can have it as a css class icon, with onClick to open various panels as component requires

gailsteiger: In patternfly, the kebab is a restyled version of a dropdown supporting the same features as the bootstrap dropdown component. I'll check with Roxanne as to whether there are use cases for the dropdown menu or whether the kebab should always be used.

dLobatog:
@ohadlevy @gailsteiger @matanwerbner 

Updated - the Dropdown is now in its own component. It accepts the ID in order to create unique ids for the elements (which QE deems useful). I could see making it more general but I think it's best to leave it 'YAGNI' and add it when we actually want it.

Aside from that, I believe I addressed the other comments (using map instead of the for iterator, etc..)

webpack/assets/javascripts/react_app/components/common/Dropdown.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+
+const Dropdown = ({ links, id }) => {
+  const listLinks = [];
+
```
matanwerbner: functional approach to writing this would be:

```

const listLinks = links.map((link, i) => { 
  const liKey = `notification-link-${i}-${id}`
 return <li key={liKey} id={liKey}>
                    <a target="_blank" href={link.href}>
                    {link.title}</a>
                  </li>
}
```

dLobatog: That's true - haste is the enemy of doing a good job 😄  Will update in a moment

app/models/notification_recipient.rb
```diff
@@ -17,6 +17,7 @@ def payload
       :subject    => notification_blueprint.subject,
       :created_at => notification.created_at,
       :group      => notification_blueprint.group,
+      :actions    => notification_blueprint.actions
```
ohadlevy, line 20: how does the actions translate to actual urls given the subject is part of the notification vs notification_blueprint object?

dLobatog: Actions have nothing to do with subject - it's just a dumb set of links for the moment, it can be any hash https://github.com/theforeman/foreman/pull/4265/files#diff-61072566998937ff274a277f94d284e8R12

It can be made more complicated but I'm not sure how good is it to start coding logic nothing is using, this approach works well with the RSS PR (if passing an array of links isn't good enough ,e.g: you want to pass the HTTP action - POST, etc.., that's easily changeable if we need to do it)

webpack/assets/javascripts/react_app/components/common/Dropdown.js
```diff
@@ -0,0 +1,27 @@
...
+      </li>
+    );
+  });
+
+  return (
+    <div className="dropdown pull-right dropdown-kebab-pf">
+      <button className="btn btn-link dropdown-toggle" type="button" id="dropdownKebabRight{id}"
+               data-toggle="dropdown" aria-haspopup="true" aria-expanded="true">
+        <span className="fa fa-ellipsis-v"></span>
+      </button>
```
gailsteiger: The component is still using bootstrap's dropdown.js plugin to handle the toggle. This is not desirable

dLobatog: Why not? I see it's handled like that on Patternfly and the Angular implementation of Patternfly too.

gailsteiger: Patternfly itself doesn't implement a javascript framework so it makes sense that it uses the bootstrap jquery plugins.

With Angular and React the paradigm moves away from direct manipulation markup and adopts the approach of rendering to reflect state. With React the concept of one way data flow is also introduced.

If you have a look at [AngularUI Bootstrap](https://angular-ui.github.io/bootstrap/) (the port of Bootstrap to Angular) you can see that it contains native AngularJS directives and has no dependency on either jQuery or Bootstrap's JS. I don't know why Angular Patternfly chose to not to implement in the same manner.

I think that our components will be easier to reason about if we adopt the pure React approach instead of mixing Reach with the Bootstrap plugins. Specifically, the bootstrap approach was not workable with the Navigation Drawer and Alert/Toasts (which you can see when I update the PR).

I think would be better to encapsulate a click handler in the component rather than rely on the boostrap plugins. In that way it will be easier for us to manage state, render in the React way and initiate actions.

@matanwerbner what do you think?

matanwerbner: Agree. We should try to not depend on bootstrap at all, especially on simple or trivial components. A bit of css wil get us where we need.

dLobatog:
@matanwerbner @gailsteiger Pushed an updated version that uses the `Dropdown` component from react-bootstrap, it's quite useful. 

I'll submit this implementation to https://idimaster.github.io/patternfly-react/ see if we can hopefully rely on this npm package for every Patternfly component 😄

webpack/assets/javascripts/react_app/components/notifications/NotificationDropdown.js
```diff
@@ -0,0 +1,28 @@
...
+
+    return (
+      <MenuItem key={liKey} id={liKey} target="_blank" href={link.href}>
+        {link.title}
+      </MenuItem>
+    );
+  });
+
+  return (
+    <Dropdown className="pull-right dropdown-kebab-pf" pullRight key={id}
```
matanwerbner, line 16: small comment on style, styling with flex box is easier then pull-right/pull-left... if the containing div has display:flex and justify-content:space-between, the elements drop into position nicely :)

gailsteiger:
@dLobatog I don't like react-bootstrap. I think it's overly complicated and tedious to use. I'm not going to object to the PR but I think a simple component would be better.

In this case, you had to use many element tags with the dropdown:
```js
    <Dropdown className="pull-right dropdown-kebab-pf" pullRight key={id}
              id={`notifications-dropdown-${id}`}>
      <Dropdown.Toggle noCaret bsStyle="link">
        <Glyphicon bsClass="fa" glyph="ellipsis-v" />
      </Dropdown.Toggle>
      <Dropdown.Menu>
        {listLinks}
      </Dropdown.Menu>
    </Dropdown>
```
I would have to consult the documentation every time I used it.

Instead, I think we should be able to achieve something along the lines of 
```js
    <Dropdown type="kebab" key={id}  id={`notifications-dropdown-${id}`}>
        {listLinks}
    </Dropdown>
```

PS: I'm not opposed to using component libraries but some are better than others

webpack/assets/javascripts/react_app/components/notifications/NotificationDropdown.js
dLobatog: I seem to not be able to do this properly as flex-box centers the text 😞

dLobatog:
@gailsteiger thanks for taking a look! About react-bootstrap, I thought it would be a good idea to use it since it is already a dependency of Foreman and the component is fully implemented there. This means different types of toggle, dividers, etc.. we don't have to write them. Additionally the docs are pretty good and with a large community I would expect it to not have many bugs or at least have them fixed at a faster rate than if I wrote our own foreman notifications dropdown component (we can always contribute back 😄 ). 

@matanwerbner Is the flex-box CSS thing a deal-breaker? I'm no expert or anything but it's what I see the Patternfly implementors use in their examples (http://www.patternfly.org/pattern-library/communication/notification-drawer/#/_code) so I suppose they code Patternfly with the expectations pull-direction is used instead of flex-box.

@gailsteiger @matanwerbner ++ appreciate the feedback!

edit - react-bootstrap, not react-patternfly

webpack/assets/javascripts/react_app/components/notifications/NotificationDropdown.js
matanwerbner: It's ok, it was just something i thought is the responsibility of the composer (wrapper) component to organize it's children - and not for the child to be opinionated about it's location.

i suggest reading up on flexbox anyway, you may find it easier to deal with then floats

gailsteiger:
@dLobatog I presume your above comments are about react-bootstrap and not about react-patternfly.
I am guilty of having introduced the react-bootstrap dependency into foreman. At present I think we use only one component - modal. I tried a few of the other components and removed them for various reasons.
For the purpose of this PR, leave the react-bootstrap Dropdown in.

I still would like to see the feature running. Could you refer me to a live site or help me get the appropriate data into my database.

webpack/assets/javascripts/react_app/components/notifications/NotificationDropdown.js
gailsteiger: @dLobatog you might find this useful [A Complete Guide to Flexbox](https://css-tricks.com/snippets/css/a-guide-to-flexbox/)

dLobatog:
@gailsteiger If you want to test this, you can create notifications this way:

1st - Run `rake db:migrate` to run the migration from this pull request that adds 'actions' to the Notifications table.
2nd - Open the Rails console with `rake console` from the Foreman directory
3rd - Issue this commands:
```ruby
blueprint = NotificationBlueprint.new(:group => 'React devs',
                                      :message => 'Hi Gail! This is a notification message',
                                      :level => 'info',
                                      :expires_in => 10.months,
                                      :name => 'gailnotification',
                                      :actions => {"links"=>[{"href"=>"https://theforeman.org/blog", "title"=>"Link to blog"}]})

blueprint.save
notification = Notification.new(:initiator => User.first, :audience => Notification::AUDIENCE_GLOBAL, :notification_blueprint => blueprint)
notification.save
```

It should automatically show up on the UI after that.

matanwerbner:
@dLobatog seems good from my end, 
I feel like the structure of the notification drawer can be greatly improved, but that is for another PR

gailsteiger:
@dLobatog thanks for the rails console instructions. Looks fine by me.

webpack/assets/javascripts/react_app/components/notifications/NotificationDropdown.js
```diff
@@ -0,0 +1,28 @@
...
+              id={`notifications-dropdown-${id}`}>
+      <Dropdown.Toggle noCaret bsStyle="link">
+        <Glyphicon bsClass="fa" glyph="ellipsis-v" />
```
ohadlevy, line 19: why not use our Icon component?

ohadlevy:
@matanwerbner can to explain regarding notification structure comment?

ohadlevy:
@dLobatog note that a test has failed, it does look like its un-related, but if you push again it would be nice to confirm.

webpack/assets/javascripts/react_app/components/notifications/NotificationDropdown.js
dLobatog: I did not see any advantage on using that component vs using what's provided by react-bootstrap, is there any?

dLobatog:
Repushed to confirm the error some node tests in Travis is not this PR's fault

ohadlevy:
thanks @dLobatog, katello tests failures are unrelated to this pr.

PR #4266, opened 2017-02-09
fixes #18445 - always define eager_load_paths, use require_dep

domcleal:
Allows `eager_load!` to be called or enabled in any environment without
loading files under lib/ twice. Switches many `require` calls to Rails'
`require_dependency` to always use its dependency loader, so when eager
loading it will track whether the file is loaded or not.

### DIFF
--- a/app/models/concerns/orchestration.rb
+++ b/app/models/concerns/orchestration.rb
@@ -1,5 +1,5 @@
-require "proxy_api"
-require 'orchestration/queue'
+require_dependency "proxy_api"
+require_dependency 'orchestration/queue'
 
 module Orchestration
   extend ActiveSupport::Concern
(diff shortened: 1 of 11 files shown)


domcleal:
Setting to WoC as I'm unsure this fixed the issue, it's occurring still.

domcleal:
Fixed the remaining error enabling eager loading by changing `require` calls to `require_dependency` so as to use Rails' dependency loader.

ares (commented):
It looks and works fine. The only remaining `require $our_code` I found was in rake tasks. I suppose it does not matter that much since rake task is only run once so no reloading should be required. If you agree and you left it there intentionally I'm happy to merge.

domcleal:
Thanks, I hadn't intentionally left them. I've updated them to `require_dependency` for consistency.

ares:
Thanks @domcleal, merged as 7923c95d829898062a593d0be8cac9e4a91acef4.

iNecas:
This broke the plugins API in development, to reproduce:

1. enable a plugin (see it in about page)
2. edit some model (such as app/models/setting.rb)
3. see the about page again: no plugin present (the same applies to menus)

PR #4267, opened 2017-02-09
Fixes #18317 - Clone user role via API

xprazak2:
I do not clone description since it may be removed in the future based on output we got from component demos. Let me know if it should be cloned as well.

### DIFF
--- a/app/controllers/api/v2/roles_controller.rb
+++ b/app/controllers/api/v2/roles_controller.rb
@@ -4,7 +4,7 @@ module Api
       include Foreman::Controller::Parameters::Role
 
       before_action :find_optional_nested_object
-      before_action :find_resource, :only => %w{show update destroy}
+      before_action :find_resource, :only => %w{show update destroy clone}
 
       api :GET, "/roles/", N_("List all roles")
       param_group :search_and_pagination, ::Api::V2::BaseController
@@ -24,6 +24,7 @@ module Api
       def_param_group :role do
         param :role, Hash, :required => true, :action_aware => true do
           param :name, String, :required => true
+          param :description, String, :desc => N_('Role description')
           param_group :taxonomies, ::Api::V2::BaseController
         end
       end
@@ -51,10 +52,27 @@ module Api
         process_response @role.destroy
       end
 
+      api :POST, "/roles/:id/clone", N_("Clone a role")
+      param :id, String, :required => true
+      param_group :role
+      def clone
+        @role = @role.clone(role_params)
+        process_response @role.save
+      end
+
       private
 
       def allowed_nested_id
         %w(user_id)
+      end
+
+      def action_permission
+        case params[:action]
+        when 'clone'
+          :create
+        else
+          super
+        end
       end
     end
   end
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -21,7 +21,6 @@ class Role < ActiveRecord::Base
   friendly_id :name
 
   include Parameterizable::ByIdName
-
   # Built-in roles
   BUILTIN_DEFAULT_ROLE = 2
   audited
@@ -143,6 +142,13 @@ class Role < ActiveRecord::Base
 
   def disable_filters_overriding
     self.filters.where(:override => true).map { |filter| filter.disable_overriding! }
+  end
+
+  def clone(role_params = {})
+    new_role = self.deep_clone(:except  => [:name, :builtin],
+                               :include => [:locations, :organizations, { :filters => :permissions }])
+    new_role.attributes = role_params
+    new_role
   end
 
   private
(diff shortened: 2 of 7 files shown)


ares:
I think the atrribute is not planned to be removed. It was more about the inline help for description field I believe.

ares:
Tests needs fixing `ArgumentError: Factory not registered: locations`

dLobatog:
@xprazak2 API needs a permission - didn't check the rest of the PR yet but works fine for admin users from what I see 😄

ares:
@xprazak2 please leave a message when you update the PR, I see the permission is there now

ares:
[test foreman]

xprazak2:
[test foreman]

app/models/role.rb
```diff
@@ -145,6 +144,11 @@ def disable_filters_overriding
     self.filters.where(:override => true).map { |filter| filter.disable_overriding! }
   end
 
+  def clone
+    self.deep_clone(:except  => [:name, :builtin],
+                    :include => [:locations, :organizations, { :filters => :permissions }])
```
ares: would you mind updating https://github.com/theforeman/foreman/blob/develop/app/controllers/roles_controller.rb#L92 to use this `#clone` as well?

ares (changes requested):
one nit, otherwise looks good to me, once resolved, I'll test it

xprazak2:
I updated controller to use `clone` method as well.

app/controllers/api/v2/roles_controller.rb
```diff
@@ -51,11 +51,30 @@ def destroy
         process_response @role.destroy
       end
 
+      api :POST, "/roles/:id/clone", N_("Clone a role")
+      param :id, String, :required => true
+      param :new_name, String, :required => true, :desc => N_('Name of the cloned role')
```
ares: I see param `name` is used few lines below, which one is correct? also could it be enhanced so it accepts the whole role param_group? I'd like to set the organization while cloning it as well as description.

ares (changes requested):
Works well except the fact I can't specify additional attribute values. Please see the inline comment for more details.

app/controllers/api/v2/roles_controller.rb
xprazak2: I changed to `:new_name`, which is correct one. Interestingly enough hammer sends both `:name` and `:new_name` with the same value, that is why I haven't discovered it:



I can see the value of adding different description to a newly cloned role, but how setting organizations should work is not obvious. Should new clone have the organizations from original when you do not supply any orgs, or should it have none? I believe it would be better to use update for any modifications.

ares: I think all attributes should be consistent. Description is cloned and would not be appended if I specify new one correct/ So I think orgs and locs should behave the same. If you don't specify any, it should have the same as origin. If I specify some (including empty array), it should set this new collection.

ares (changes requested):
Please note that tests are failing.

xprazak2:
We can now specify taxonomies and description that will be set for the newly created role.

ares:
I'll retest but it seems it breaks katello tests, maybe the role cloning is used there somewhere? There are couple of messages like

`ActiveRecord::AssociationTypeMismatch: Taxonomy(#84531320) expected, got NilClass(#8884800)...`

ares (changes requested):
I'm sorry I still don't understand where `:new_name` is used. It was not allowed in strong params filters so I think it's ignored. 2 of 3 tests in this PR use `:name` only, third one uses `new_name` but also set role hash which contains name so I don't think they test this attribute. My testing result:

clone

Using name it works just fine. Please let me know what I'm doing wrong.

ares:
Another finding is that cloning role that has some taxonomies set does not keep it for new clone (even if I don't specify organization_ids at all)

I'd expect this sets org 1 for the clone and leaves locations untouched
```json
{
  "role": {
    "name": "Manager of Ares23",
    "organization_ids": [1]
  }
}
```

If I'd like to reset locations to empty list, following should be used
```json
{
  "role": {
    "name": "Manager of Ares23",
    "organization_ids": [1],
    "location_ids": []
  }
}
```

Since it does not work this way right now, I think a test for this would be useful.

xprazak2:
I removed the `:new_name` completely, it is a hammer thing and is not needed here. I do not see how I could have broken functional test for subscriptions/index by adding a new method to role. I will revisit it if katello tests remain red. I added tests to cover taxonomy modification cases .

ares (approved):
Works nicely, :+1:

ares:
Thanks @xprazak2, merged as 8813aee5130dc46b8a713864a272054bb7b73f7d

PR #4268, opened 2017-02-12
Fixes #18461 - reword 'External groups'

shlomizadok:


app/views/usergroups/_form.html.erb
```diff
@@ -4,7 +4,7 @@
     <li class="active"><a href="#primary" data-toggle="tab"><%= _("User group") %></a></li>
```
tbrisker: How about fixing the case here as well while you're at it?

shlomizadok: Done :+1:

tbrisker:
APJ

tbrisker:
thanks @shlomizadok !

PR #4270, opened 2017-02-12
Fixes #18760 - Allow export to CSV

tbrisker:
This introduces a way of exporting tables from the UI to CSV.
There are 3 steps to adding a CSV export to a table:

1. Add the CsvResponder concern to the relevant controller.
2. Add a `format.csv` block to the index controller action. This block
should include a call to `csv_response` with the wanted resources.
3. Add `csv_link` to the title actions. (Or wherever you want the export
button to appear)

By default, the csv will include all columns that exist for that model
in the database, except for the `created_at` and `updated_at` ones. To
override the default, add a method `csv_columns` to the controllers that
returns an array of columns. Columns can be actual column names in the
database, reflections, or any method that can be called on an instance
of the model.

This commit also includes an example of the feature implemented on the
Hosts index page.

### DIFF
--- a/app/controllers/hosts_controller.rb
+++ b/app/controllers/hosts_controller.rb
@@ -9,6 +9,7 @@ class HostsController < ApplicationController
   include Foreman::Controller::SmartProxyAuth
   include Foreman::Controller::Parameters::Host
   include Foreman::Controller::Puppet::HostsControllerExtensions
+  include Foreman::Controller::CsvResponder
 
   SEARCHABLE_ACTIONS= %w[index active errors out_of_sync pending disabled ]
   AJAX_REQUESTS=%w{compute_resource_selected current_parameters process_hostgroup process_taxonomy review_before_build scheduler_hint_selected}
@@ -58,8 +59,10 @@ class HostsController < ApplicationController
         @hostgroup_authorizer = Authorizer.new(User.current, :collection => @hosts.map(&:hostgroup_id).compact.uniq)
         render :index if title && (@title = title)
       end
-      format.yaml { render :text => search.all(:select => "hosts.name").map(&:name).to_yaml }
-      format.json
+      format.csv do
+        @hosts = search.includes(included_associations - [:host_statuses, :token, :compute_resource])
+        csv_response(@hosts)
+      end
     end
   end
 
@@ -917,5 +920,9 @@ class HostsController < ApplicationController
     host_params.select do |k,v|
        host_attributes.include?(k) && !k.end_with?('_ids')
     end.except(:host_parameters_attributes)
+  end
+
+  def csv_columns
+    [:name, :operatingsystem, :environment, :model, :hostgroup, :last_report]
   end
 end
--- a/app/controllers/templates_controller.rb
+++ b/app/controllers/templates_controller.rb
@@ -99,6 +99,10 @@ class TemplatesController < ApplicationController
     send_data @template.to_erb, :type => 'text/plain', :disposition => 'attachment', :filename => @template.filename
   end
 
+  def resource_class
+    @resource_class ||= controller_name.singularize.classify.constantize
+  end
+
   private
 
   def safe_render(template)
@@ -142,10 +146,6 @@ class TemplatesController < ApplicationController
 
   def resource_name
     'template'
-  end
-
-  def resource_class
-    @resource_class ||= controller_name.singularize.classify.constantize
   end
 
   def type_name_plural
--- a/app/services/csv_exporter.rb
+++ b/app/services/csv_exporter.rb
@@ -1,0 +1,17 @@
+require 'csv'
+
+module CsvExporter
+  def self.export(resources, columns)
+    Enumerator.new do |csv|
+      csv << csv_header(columns)
+
+      resources.reorder(nil).limit(nil).find_each do |obj|
+        csv << CSV.generate_line(columns.map{|c| obj.send(c)})
+      end
+    end
+  end
+
+  def self.csv_header(columns)
+    CSV.generate_line(columns.map{|c| c.to_s.titleize})
+  end
+end
--- a/app/controllers/concerns/foreman/controller/csv_responder.rb
+++ b/app/controllers/concerns/foreman/controller/csv_responder.rb
@@ -1,0 +1,15 @@
+module Foreman::Controller::CsvResponder
+  extend ActiveSupport::Concern
+
+  def csv_response(resources, columns = csv_columns)
+    headers["Cache-Control"] = "no-cache"
+    headers["Content-Type"] = "text/csv; charset=utf-8"
+    headers["Content-Disposition"] = %(attachment; filename="#{controller_name}-#{Date.today}.csv")
+    self.response_body = CsvExporter.export(resources, columns)
+  end
+
+  private
+  def csv_columns
+    resource_class.column_names - ['created_at', 'updated_at']
+  end
+end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -152,6 +152,11 @@ module ApplicationHelper
     display_link_if_authorized(name, options, html_options)
   end
 
+  def csv_link
+    link_to(_('Export'), params.merge(:format => :csv),
+      {:title => _('Export to CSV'), :class => 'btn btn-default', 'data-no-turbolink' => true})
+  end
+
   # renders a style=display based on an attribute properties
   def display?(attribute = true)
     "style=#{display(attribute)}"
--- a/app/views/hosts/index.html.erb
+++ b/app/views/hosts/index.html.erb
@@ -1,2 +1,2 @@
-<% title_actions multiple_actions_select, button_group(new_link(_("Create Host"))) %>
+<% title_actions multiple_actions_select, csv_link, button_group(new_link(_("Create Host"))) %>
 <%= render 'list', :hosts => @hosts, :header => @title || _("Hosts")  %>
--- a/test/controllers/concerns/csv_responder_test.rb
+++ b/test/controllers/concerns/csv_responder_test.rb
@@ -1,0 +1,50 @@
+require 'test_helper'
+
+class FakeController < ApplicationController
+  include Foreman::Controller::CsvResponder
+
+  def index
+    csv_response(Domain.unscoped, [:name])
+  end
+end
+
+class Api::V2::FakeController < Api::V2::BaseController
+  include Foreman::Controller::CsvResponder
+
+  def index
+    csv_response(Domain.unscoped, [:name])
+  end
+end
+
+#add the fake controllers to the routes table
+Rails.application.routes.disable_clear_and_finalize = true
+Rails.application.routes.draw do
+  get '/fake' => 'fake#index'
+  get '/fake_api' => 'api/v2/fake#index'
+end
+
+class CsvResponderTest < ActionController::TestCase
+  tests FakeController
+
+  test "response is streamed correctly with right headers" do
+    get :index, {}, set_session_user
+    assert_equal "text/csv; charset=utf-8", response.headers["Content-Type"]
+    assert_equal "no-cache", response.headers["Cache-Control"]
+    assert_equal "attachment; filename=\"fake-#{Date.today}.csv\"", response.headers["Content-Disposition"]
+    assert response.stream.instance_variable_get(:@buf).is_a? Enumerator
+    assert_equal "Name\n", response.stream.instance_variable_get(:@buf).next
+  end
+end
+
+class CsvApiResponderTest < ActionController::TestCase
+  tests Api::V2::FakeController
+
+  test "response is streamed correctly with right headers" do
+    get :index
+    assert_equal "text/csv; charset=utf-8", response.headers["Content-Type"]
+    assert_equal "no-cache", response.headers["Cache-Control"]
+    assert_equal "attachment; filename=\"fake-#{Date.today}.csv\"", response.headers["Content-Disposition"]
+    assert response.stream.instance_variable_get(:@buf).is_a? Enumerator
+    assert_equal "Name\n", response.stream.instance_variable_get(:@buf).next
+  end
+end
--- a/test/unit/csv_exporter_test.rb
+++ b/test/unit/csv_exporter_test.rb
@@ -1,0 +1,29 @@
+require 'test_helper'
+
+class CsvExporterTest < ActiveSupport::TestCase
+  test 'return correct amount of lines' do
+    result = CsvExporter.export(Host::Managed, [:id])
+    assert_equal "Id\n", result.next
+    assert_equal result.count, Host::Managed.count+1
+  end
+
+  test 'handles empty results correctly' do
+    result = CsvExporter.export(Host::Managed.where(:name => 'no-such-host'), [:id, :name])
+    assert_equal "Id,Name\n", result.next
+    assert_equal 1, result.count
+    assert_raises StopIteration do
+      result.next
+    end
+  end
+
+  test 'calls methods on records' do
+    id = Domain.first.id
+    Domain.any_instance.expects(:test_method).once.returns('success!')
+    result = CsvExporter.export(Domain.where(:id => id), [:id, :test_method])
+    assert_equal "Id,Test Method\n", result.next
+    assert_equal "#{id},success!\n", result.next
+    assert_raises StopIteration do
+      result.next
+    end
+  end
+end


app/controllers/hosts_controller.rb
```diff
@@ -60,6 +60,9 @@ def index(title = nil)
       end
       format.yaml { render :text => search.all(:select => "hosts.name").map(&:name).to_yaml }
       format.json
+      format.csv do
+        send_data search.to_csv(columns: [:id, :name, :os, :environment, :model, :hostgroup, :last_report])
```
ohadlevy: can we add all attributes by default? I think most users won't mind to remove data vs adding would be more complex?

ohadlevy: or maybe a way to reuse default json attributes?

ohadlevy: also, should it export to a file by default? e.g.
```ruby
format.csv { send_data @users.to_csv, filename: "users-#{Date.today}.csv" }
```

app/models/concerns/csv_exporter.rb
```diff
@@ -0,0 +1,17 @@
+require 'csv'
+
+module CsvExporter
+  extend ActiveSupport::Concern
+
+  module ClassMethods
+    def to_csv(columns: column_names, csv_options: {})
+      CSV.generate(csv_options) do |csv|
+        csv << columns.map(&:to_s)
+        columns.map!(&:to_sym)
```
ohadlevy: why are you looping columns twice?

app/models/concerns/csv_exporter.rb
```diff
@@ -0,0 +1,17 @@
...
+        all.limit(nil).find_each do |obj|
```
ohadlevy: should consider using batches instead of loading all objects into memory.

ohadlevy: really missing search filter here.

app/models/concerns/csv_exporter.rb
```diff
@@ -0,0 +1,17 @@
+require 'csv'
+
+module CsvExporter
+  extend ActiveSupport::Concern
+
+  module ClassMethods
+    def to_csv(columns: column_names, csv_options: {})
+      CSV.generate(csv_options) do |csv|
+        csv << columns.map(&:to_s)
```
ohadlevy: column_names is unused?

app/controllers/hosts_controller.rb
tbrisker: The default if no columns are passed is to use `column_names` which just gets all the columns from the database. Here I tried to match the UI table, but we don't have to. However `column_names` doesn't provide stuff like `hostgroup`, only `hostgroup_id` - this way we can call methods besides the data that is stored in the db.
I can add the filename export, sure.

app/models/concerns/csv_exporter.rb
tbrisker: `column_names` returns a list of strings, I want to be able to pass a list of symbols as well, and I need to make sure they are in the correct format in both places - headers are strings, sent parameters are symbols.

app/models/concerns/csv_exporter.rb
tbrisker: `find_each` loads them in batches of 1000 by default.
Also, this method is called on a scope that includes the search filter.

app/models/concerns/csv_exporter.rb
tbrisker: `column_names` is an activerecord method that gives a list of all the column names, and it is used to provide the default value for the `columns` kwarg if no other value is passed.

app/controllers/hosts_controller.rb
```diff
@@ -60,6 +60,10 @@ def index(title = nil)
       end
       format.yaml { render :text => search.all(:select => "hosts.name").map(&:name).to_yaml }
       format.json
+      format.csv do
+        send_data search.to_csv(columns: [:id, :name, :os, :environment, :model, :hostgroup, :last_report]),
+                  filename: "hosts-#{Date.today}.csv"
```
ohadlevy: this generates N+1 Queries, see:
```
16:04:27 rails.1   | 2017-02-12T16:04:27 d703ba15 [app] [I] Sent data hosts-2017-02-12.csv (5.7ms)
16:04:27 rails.1   | 2017-02-12T16:04:27 d703ba15 [app] [I] Completed 200 OK in 408666ms (Views: 5.5ms | ActiveRecord: 4177.0ms)
16:04:27 rails.1   | 2017-02-12T16:04:27 d703ba15 [app] [W] user: ohad
16:04:27 rails.1   |  | GET /hosts.csv
16:04:27 rails.1   |  | USE eager loading detected
16:04:27 rails.1   |  |   Host::Managed => [:operatingsystem]
16:04:27 rails.1   |  |   Add to your finder: :includes => [:operatingsystem]
16:04:27 rails.1   |  | Call stack
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/models/concerns/csv_exporter.rb:12:in `block (3 levels) in to_csv'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/models/concerns/csv_exporter.rb:12:in `map'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/models/concerns/csv_exporter.rb:12:in `block (2 levels) in to_csv'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/models/concerns/csv_exporter.rb:11:in `block in to_csv'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/models/concerns/csv_exporter.rb:8:in `to_csv'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/controllers/hosts_controller.rb:64:in `block (2 levels) in index'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/controllers/hosts_controller.rb:51:in `index'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/controllers/concerns/application_shared.rb:14:in `set_timezone'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/models/concerns/foreman/thread_session.rb:32:in `clear_thread'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/lib/middleware/catch_json_parse_errors.rb:8:in `call'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/lib/middleware/tagged_logging.rb:18:in `call'
16:04:27 rails.1   |  | 
16:04:27 rails.1   |  | 
16:04:27 rails.1   | 2017-02-12T16:04:27 d703ba15 [app] [W] user: ohad
16:04:27 rails.1   |  | GET /hosts.csv
16:04:27 rails.1   |  | USE eager loading detected
16:04:27 rails.1   |  |   Host::Managed => [:environment]
16:04:27 rails.1   |  |   Add to your finder: :includes => [:environment]
16:04:27 rails.1   |  | Call stack
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/models/concerns/csv_exporter.rb:12:in `block (3 levels) in to_csv'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/models/concerns/csv_exporter.rb:12:in `map'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/models/concerns/csv_exporter.rb:12:in `block (2 levels) in to_csv'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/models/concerns/csv_exporter.rb:11:in `block in to_csv'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/models/concerns/csv_exporter.rb:8:in `to_csv'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/controllers/hosts_controller.rb:64:in `block (2 levels) in index'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/controllers/hosts_controller.rb:51:in `index'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/controllers/concerns/application_shared.rb:14:in `set_timezone'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/models/concerns/foreman/thread_session.rb:32:in `clear_thread'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/lib/middleware/catch_json_parse_errors.rb:8:in `call'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/lib/middleware/tagged_logging.rb:18:in `call'
16:04:27 rails.1   |  | 
16:04:27 rails.1   |  | 
16:04:27 rails.1   | 2017-02-12T16:04:27 d703ba15 [app] [W] user: ohad
16:04:27 rails.1   |  | GET /hosts.csv
16:04:27 rails.1   |  | USE eager loading detected
16:04:27 rails.1   |  |   Host::Managed => [:model]
16:04:27 rails.1   |  |   Add to your finder: :includes => [:model]
16:04:27 rails.1   |  | Call stack
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/models/concerns/csv_exporter.rb:12:in `block (3 levels) in to_csv'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/models/concerns/csv_exporter.rb:12:in `map'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/models/concerns/csv_exporter.rb:12:in `block (2 levels) in to_csv'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/models/concerns/csv_exporter.rb:11:in `block in to_csv'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/models/concerns/csv_exporter.rb:8:in `to_csv'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/controllers/hosts_controller.rb:64:in `block (2 levels) in index'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/controllers/hosts_controller.rb:51:in `index'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/controllers/concerns/application_shared.rb:14:in `set_timezone'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/models/concerns/foreman/thread_session.rb:32:in `clear_thread'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/lib/middleware/catch_json_parse_errors.rb:8:in `call'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/lib/middleware/tagged_logging.rb:18:in `call'
16:04:27 rails.1   |  | 
16:04:27 rails.1   |  | 
16:04:27 rails.1   | 2017-02-12T16:04:27 d703ba15 [app] [W] user: ohad
16:04:27 rails.1   |  | GET /hosts.csv
16:04:27 rails.1   |  | USE eager loading detected
16:04:27 rails.1   |  |   Host::Managed => [:hostgroup]
16:04:27 rails.1   |  |   Add to your finder: :includes => [:hostgroup]
16:04:27 rails.1   |  | Call stack
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/models/concerns/csv_exporter.rb:12:in `block (3 levels) in to_csv'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/models/concerns/csv_exporter.rb:12:in `map'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/models/concerns/csv_exporter.rb:12:in `block (2 levels) in to_csv'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/models/concerns/csv_exporter.rb:11:in `block in to_csv'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/models/concerns/csv_exporter.rb:8:in `to_csv'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/controllers/hosts_controller.rb:64:in `block (2 levels) in index'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/controllers/hosts_controller.rb:51:in `index'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/controllers/concerns/application_shared.rb:14:in `set_timezone'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/app/models/concerns/foreman/thread_session.rb:32:in `clear_thread'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/lib/middleware/catch_json_parse_errors.rb:8:in `call'
16:04:27 rails.1   |  |   /home/ohad/git/foreman/lib/middleware/tagged_logging.rb:18:in `call'
16:04:27 rails.1   |  | 
```

app/controllers/hosts_controller.rb
```diff
@@ -48,18 +48,20 @@ def index(title = nil)
...
-        @hosts = search.includes(included_associations).paginate(:page => params[:page])
         # SQL optimizations queries
         @last_report_ids = ConfigReport.where(:host_id => @hosts.map(&:id)).group(:host_id).maximum(:id)
         @last_reports = ConfigReport.where(:id => @last_report_ids.values)
         # rendering index page for non index page requests (out of sync hosts etc)
         @hostgroup_authorizer = Authorizer.new(User.current, :collection => @hosts.map(&:hostgroup_id).compact.uniq)
         render :index if title && (@title = title)
       end
-      format.yaml { render :text => search.all(:select => "hosts.name").map(&:name).to_yaml }
-      format.json
```
tbrisker, line 62: These formats seem to be broken and not used anywhere anyways afaict.

app/controllers/templates_controller.rb
```diff
@@ -99,6 +99,10 @@ def export
     send_data @template.to_erb, :type => 'text/plain', :disposition => 'attachment', :filename => @template.filename
   end
 
+  def resource_class
```
tbrisker, line 102: need resource_class to not be private so the index view can check if the to_csv method is supported.

ohadlevy:
@tbrisker set WoC to handle N+1 problem. thanks.

tbrisker:
@ohadlevy I think I fixed it in my last push, could you please check again?

app/controllers/hosts_controller.rb
```diff
@@ -48,18 +48,20 @@ def index(title = nil)
       error e.to_s
       search = resource_base.search_for ''
     end
+    @hosts = search.includes(included_associations).paginate(:page => params[:page])
```
ohadlevy: testing with newer version is a lot faster:
```
17:12:27 rails.1   | 2017-02-12T17:12:27 c58ea131 [app] [I] Sent data hosts-2017-02-12.csv (0.5ms)
17:12:27 rails.1   | 2017-02-12T17:12:27 c58ea131 [app] [I] Completed 200 OK in 61069ms (Views: 0.4ms | ActiveRecord: 1690.6ms)
```

However, most of the AR time was spent in Token and Host Status loading, both are not reflected in the CSV, in that regard, it does open the question if we should enable power status and host-status visibility or not in the csv.

tbrisker: I think power status would require much too long for it to be useful as it will have to query external apis. Not sure about status, might also be complicated to include as it is computed currently in a complex way and not as a method on host that we can easily call. I'll remove for now and see how that goes.

tbrisker:
@ohadlevy removed power and status related includes, please try again.

thomasmckay:
This is a fairly significant feature that will impact all the tables. Is it WYSIWYG? Do the other export tools (hammer csv, rake tasks, etc.) need to be updated to leverage this new functionality? Are plugin data extensions taken into account?

ohadlevy (changes requested):
@tbrisker I would still want to see two more items in this PR:

1. the ability to pass the attributes in which a user can ask for data, for example:
/hosts.csv?search=<>&fields=<fieldA,fieldB..>

2. extract the default fields into a method, so plugins can override it easily.

ohadlevy:
@rohoover what do you think about the icon and placement of the action(note its a img while hovering the btn)? thanks!

image

dLobatog:
@ohadlevy To me that image looks like I'm exporting the results of the search, instead of whatever is actually shown in the table. I would say next to the pagination numbers it would convey the meaning of 'export this table' better - but I defer to @rohoover - might be just a first impression

Rohoover:
@ohadlevy @dLobatog 

If it is exporting the table results, then I would not put it on the search bar. I would place it with the table actions. That is where many of the export buttons in the application are located. I'd also compare the other export buttons for verbiage and iconography. I do not think all the export buttons have an icon.

ohadlevy:
@tbrisker latest test shows another big improvement!
```
17:58:18 rails.1   | 2017-02-13T17:58:18 d617a551 [app] [I] Sent data hosts-2017-02-13.csv (13.8ms)
17:58:18 rails.1   | 2017-02-13T17:58:18 d617a551 [app] [I] Completed 200 OK in 42063ms (Views: 13.6ms | ActiveRecord: 645.7ms)
```

sadly you can still see that out of the 42 seconds, only 6.5 seconds were in the db, the rest is mostly in the csv generation.

app/controllers/hosts_controller.rb
```diff
@@ -58,8 +58,11 @@ def index(title = nil)
         @hostgroup_authorizer = Authorizer.new(User.current, :collection => @hosts.map(&:hostgroup_id).compact.uniq)
         render :index if title && (@title = title)
       end
-      format.yaml { render :text => search.all(:select => "hosts.name").map(&:name).to_yaml }
-      format.json
+      format.csv do
+        @hosts = search.includes(included_associations - [:host_statuses, :token, :compute_resource])
+        send_data @hosts.to_csv(columns: [:id, :name, :os, :environment, :model, :hostgroup, :last_report]),
+                  filename: "hosts-#{Date.today}.csv"
```
lzap: One note, this will easily eat all memory of a server process for big tables like reports or audits. While `send_data` streams (which is exactly what we want to do), method `to_csv` creates in-memory string. We already treat memory badly in Foreman and we do not want to make it worse, so my point is, can you rewrite this into a helper method that will stream the data row by row. Ideally you can read the rows in batches using `find_in_batches` so it's also "streaming" from DB directly.

lzap (changes requested):
Requesting small change to avoid allocating in-memory string completely.

ohadlevy:
@lzap something like http://smsohan.com/blog/2013/05/09/genereating-and-streaming-potentially-large-csv-files-using-ruby-on-rails/ ?

lzap:
> @lzap something like http://smsohan.com/blog/2013/05/09/genereating-and-streaming-potentially-large-csv-files-using-ruby-on-rails/ ?

Yes, something like that. I think enterprise users will totally love this, we will create API for that as well, the code will be reused, they will run it every hour and bang - memory usage goes wild.

ohadlevy:
@Rohoover do you have a suggestion of where and what text/icon to place?

Rohoover:
@ohadlevy 

Probably to the right of the table actions... in the screen I see a Select All. In that row, most likely not the first action. If the CSV file format is common, then I would just have a button that says "Export". If CSV is uncommon, the "Export CSV"

dLobatog:
@Rohoover The button you're seeing is not 'Select all' but 'Select Action', which looks like this already:

screenshot from 2017-02-16 09-59-05

I thing the button is overused, and it represents "actions to be applied to the hosts". The option I was thinking of, is to put the icon next to the "Displaying all entries (bottom left corner)" or next to the pagination numbers like in this table:

screenshot from 2017-02-16 09-58-43

ShimShtein:
@ohadlevy, about:

> the ability to pass the attributes in which a user can ask for data, for example:
/hosts.csv?search=<>&fields=<fieldA,fieldB..>

Do we want to consider a some standard way of specifying queries? I would suggest [JSON API](http://jsonapi.org/format/#fetching-sparse-fieldsets), since we could benefit from using this standard in other areas too (I am talking about nested fields update semantics as opposed to full overwrite that we have today).

Rohoover:
@dLobatog That is the longest list of actions I've ever seen. Yikes!

I am still of the opinion that an action should not be placed with the pagination or total display entry field. I've been hunting around and that would be a highly unusual interaction pattern. For an action affecting the table or multiple rows, it would be placed in the same line with "Select Action", "Select Host". I would create a new button in that row for Export.

tbrisker:
Updated the PR:

* Stream data instead of building whole csv in memory
* Move button to title actions and change from icon to "Export".
* Changed CSV Exporter from an activerecord concern to a service.
* Added CSV Builder controller concern to set headers and initiate the data streaming. 
* Extract default fields to a method in the concern that can be overridden in the controller. 

Not in this PR: Allowing users to specify columns, as this requires checking user input for permitted values (otherwise a user can ask for a column `delete` which will delete the entire table for example). I would rather separate that ability to a future PR so this one doesn't grow too big to review (which is also the reason I added the button only on one table).

tbrisker:
setting to WoC since I need to fix the tests as well.

app/controllers/templates_controller.rb
tbrisker: actually this is no longer necessary, but i think it makes sense to leave it here anyways for consistency with the default.

tbrisker:
@ohadlevy @lzap Care to have another look? tests have been fixed.

app/services/csv_exporter.rb
```diff
@@ -0,0 +1,21 @@
+require 'csv'
+
+module CsvExporter
+  def self.to_csv(resources, columns)
+    Enumerator.new do |csv|
+      csv << csv_header(columns)
+
+      resources.find_each do |obj|
```
dLobatog: I've ran into limitations with `find_each` on trying to use this with objects in Katello API controllers. The issue is, find_each is OK when you want to call it on ActiveRecord relations. However some controllers directly return Arrays/Hashes etc.. 

I will check what to do with those so that those controllers can work with this.

app/services/csv_exporter.rb
```diff
@@ -0,0 +1,21 @@
...
+        csv << csv_row(obj, columns)
```
dLobatog: I think `obj` here has to be flattened. Otherwise relations (e.g: column_names for Product has `sync_plan`, but `sync_plan` by itself will just return an object) will not show up.

If it's not flattened, we will have to write a `to_s` for everything that doesn't display well as a string

dLobatog: That example actually isn't good. Feel free to disregard until I provide better feedback.

tbrisker:
@ohadlevy @lzap Fixed streaming, should work now. It was missing a header before (`no-cache`) which strangely didn't work for me yesterday. Also changed a couple of names to be more expressive, and removed a function call in the main csv loop (didn't see a significant difference in timing for 40K records but might be significant in a larger dataset or in memory consumption).

lzap (approved):

lzap:
Looks good to me, thanks for digging into streaming so memory does not goes high. Kudos!

thomasmckay:
See https://github.com/Katello/katello/pull/6646#issuecomment-284859529

Would it be possible to change the UI button to be a dropdown of choices in this PR? Follow up PRs could then add choices to match the existing CSV exports (hammer --csv host list and hammer csv hosts --export).

ohadlevy:
test foreman

app/controllers/hosts_controller.rb
```diff
@@ -58,8 +59,10 @@ def index(title = nil)
         @hostgroup_authorizer = Authorizer.new(User.current, :collection => @hosts.map(&:hostgroup_id).compact.uniq)
         render :index if title && (@title = title)
       end
-      format.yaml { render :text => search.all(:select => "hosts.name").map(&:name).to_yaml }
-      format.json
+      format.csv do
+        @hosts = search.includes(included_associations - [:host_statuses, :token, :compute_resource])
```
ohadlevy, line 63: I think its harmless, but in the logs i see:
```
  [sql] [W] Scoped order and limit are ignored, it's forced to be batch order and batch size
```

ohadlevy (approved):
Streaming is awesome, good job @tbrisker

app/controllers/hosts_controller.rb
```diff
@@ -918,4 +921,8 @@ def host_attributes_for_templates(host)
        host_attributes.include?(k) && !k.end_with?('_ids')
     end.except(:host_parameters_attributes)
   end
+
+  def csv_columns
+    [:id, :name, :operatingsystem, :environment, :model, :hostgroup, :last_report]
```
ohadlevy: What do you think about removing the ID column? I'm not sure if its required (nor its visible in the current table).

ohadlevy:
@tbrisker please note the test failures.

app/controllers/hosts_controller.rb
tbrisker: I could clear the order and limit explicitly to get rid of the warning.

app/controllers/hosts_controller.rb
tbrisker: I'm not sure, might be useful to have it just in case but I don't mind removing either.

tbrisker:
@ohadlevy Pushed again with cleared limit and order. previous test failures seem unrelated, let's see what happens now.

tbrisker:
@thomasmckay You are welcome to add or ask for more features to this, but please in a separate PR. 
As I mentioned already, this is not meant to replace or be identical in output to other CSV tools, this feature only tries to address the "I want this table I see on the screen in front of me in csv format" use case. 
Future additions can include allowing users to select which columns they want (http://projects.theforeman.org/issues/18761). 
If there are other use cases or features you want to add, feel free to add them to the tracker (http://projects.theforeman.org/issues/18463).

app/controllers/hosts_controller.rb
ohadlevy: I would drop the id.

ohadlevy (changes requested):
@tbrisker :+1: for the code, I would like to add a bit more testing, if you can:
1. add tests for controller headers
2. add tests that you actually get enum

also following on nested / ref data, maybe we should have a test that fails csv response if it has _id in an attribute name?

tbrisker:
@ohadlevy Added controller test, dropped ID.

ohadlevy (approved):

ohadlevy:
pending Mr. Jenkins.

ohadlevy:
thanks @tbrisker

PR #4275, opened 2017-02-13
Fixes #12163 - add a rake task to clean up orphaned facts

ares:
On my system with 11k fact names it took 44 seconds. It could be more optimized but I don't think it's necessary at this point.

### DIFF
--- a/app/services/fact_cleaner.rb
+++ b/app/services/fact_cleaner.rb
@@ -1,0 +1,49 @@
+class FactCleaner
+  delegate :logger, :to => :Rails
+
+  def clean!
+    to_delete = find_leaf_orphaned_facts
+    # we also delete all composes that either don't have any descendant or all of them are to be deleted
+    to_delete += find_compose_orphaned_facts(to_delete)
+    delete!(to_delete)
+    @deleted_count = to_delete.count
+    self
+  end
+
+  def deleted_count
+    raise 'cleaner has not cleaned anything yet, run #clean! first' if @deleted_count.nil?
+    @deleted_count
+  end
+
+  private
+
+  def find_leaf_orphaned_facts
+    result = []
+    logger.info "Searching for leaf orphaned facts..."
+    FactName.leaves.includes(:fact_values).find_in_batches do |batch|
+      result += batch.select do |fact|
+        fact.fact_values.empty?
+      end
+    end
+    result
+  end
+
+  def find_compose_orphaned_facts(found)
+    logger.info "Searching for compose orphaned facts..."
+    FactName.composes.find_in_batches do |batch|
+      found += batch.select do |compose|
+        compose.descendants.all? { |fact| found.include?(fact) }
+      end
+    end
+    found
+  end
+
+  def delete!(to_delete)
+    logger.info "#{to_delete.size} facts will be removed"
+    to_delete.each do |fact|
+      logger.debug { "Removing fact #{fact}" }
+      fact.destroy
+    end
+    logger.info "Cleanup of facts finished!"
+  end
+end
--- a/lib/tasks/facts.rake
+++ b/lib/tasks/facts.rake
@@ -1,0 +1,17 @@
+namespace :facts do
+  desc <<-END_DESC
+Removes facts without any values
+
+This is useful when you used custom facts of different facts sourcers that you don't use anymore.
+After cleaning of such facts they should no longer appear in auto-completion suggestions.
+
+Examples:
+  # foreman-rake facts:clean
+END_DESC
+
+  task :clean => :environment do
+    puts 'Starting orphaned facts clean up...'
+    count = FactCleaner.new.clean!.deleted_count
+    puts "Finished, cleaned #{count} facts"
+  end
+end
--- a/test/unit/fact_cleaner_test.rb
+++ b/test/unit/fact_cleaner_test.rb
@@ -1,0 +1,56 @@
+require 'test_helper'
+
+class FactCleanerTest < ActiveSupport::TestCase
+  let(:cleaner) do
+    FactCleaner.new
+  end
+
+  test "it cleans orhpaned root leaves" do
+    root_1_fact = FactoryGirl.create(:fact_name)
+    cleaner.clean!
+    assert_not_include remaining_of(root_1_fact), root_1_fact
+  end
+
+  test "it cleans root composes without any leaves" do
+    root_2_fact = FactoryGirl.create(:fact_name, :compose => true)
+    cleaner.clean!
+    assert_not_include remaining_of(root_2_fact), root_2_fact
+  end
+
+  test "it cleans composes and it's leaves composes if there are no values for them" do
+    root_3_fact = FactoryGirl.create(:fact_name, :compose => true)
+    root_3_child_fact = FactoryGirl.create(:fact_name, :parent_id => root_3_fact.id)
+    cleaner.clean!
+    remaining = remaining_of(root_3_fact, root_3_child_fact)
+    assert_not_include remaining, root_3_fact
+    assert_not_include remaining, root_3_child_fact
+  end
+
+  test "it keeps composes and their only children that have values" do
+    root_4_fact = FactoryGirl.create(:fact_name, :compose => true)
+    root_4_child_1_fact = FactoryGirl.create(:fact_name, :parent_id => root_4_fact.id)
+    # root_4_child_1_fact_value
+    FactoryGirl.create(:fact_value, :fact_name => root_4_child_1_fact)
+    root_4_child_2_fact = FactoryGirl.create(:fact_name, :parent_id => root_4_fact.id)
+
+    cleaner.clean!
+    remaining = remaining_of(root_4_fact, root_4_child_1_fact, root_4_child_2_fact)
+    assert_include remaining, root_4_fact
+    assert_include remaining, root_4_child_1_fact
+    assert_not_include remaining, root_4_child_2_fact
+  end
+
+  test "it keeps root leaves if the have value" do
+    root_6_fact = FactoryGirl.create(:fact_name)
+    # root_6_fact_value
+    FactoryGirl.create(:fact_value, :fact_name => root_6_fact)
+    cleaner.clean!
+    assert_include remaining_of(root_6_fact), root_6_fact
+  end
+
+  private
+
+  def remaining_of(*names)
+    FactName.where(:name => names.map(&:name))
+  end
+end
--- a/app/models/fact_name.rb
+++ b/app/models/fact_name.rb
@@ -9,6 +9,9 @@ class FactName < ActiveRecord::Base
 
   scope :no_timestamp_fact, -> { where("fact_names.name <> ?",:_timestamp) }
   scope :timestamp_facts, -> { where(:name => :_timestamp) }
+  scope :composes, -> { where(:compose => true) }
+  scope :leaves, -> { where(:compose => false) }
+
   scope :with_parent_id, lambda { |find_ids|
     conds, binds = [], []
     [find_ids].flatten.each do |find_id|


bastilian:
If orphan facts happen wouldn't it make sense to have a FactCleaner service?

Having it as a service rather than running code in the rake task also makes it more testable and helps with future (performance) improvements.

ohadlevy:
this is another example of something we want to run on a reoccurring basis, other candidates i saw recently are:
1. @dLobatog pr to pull rss feeds to notifications
2. @timogoebel keep vmware cache warm
3. all the existings cron tabs (expire reports etc)

maybe its time to introduce a proper service that tasks can hook into it? rfc?

ares:
I agree with the need of having proper service, but I hope that's orthogonal to this PR. There are users with the problem and this rake task would help them. Bonus points if we can automate the cleanup later. @bastilian you're right, I think I can move the code to the service class and change the interface of it later when we have something we can hook into. It started as a one-liner... :-)

ares:
@bastilian I extracted the code into service object and properly split the logic into methods. I also covered it with tests. Would you mind to take another look?

app/services/fact_cleaner.rb
```diff
@@ -0,0 +1,44 @@
+class FactCleaner
+  delegate :logger, :to => :Rails
+
+  def clean!
+    @to_delete = []
+    logger.info "Searching for orphaned facts..."
```
bastilian: I'd rather put the logging into the method.

app/services/fact_cleaner.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    delete!
+    logger.info "Cleanup of facts finished!"
+    @to_delete.count
+  end
+
+  private
+
+  def find_leaf_orphaned_facts
+    FactName.where(:compose => false).includes(:fact_values).find_in_batches do |batch|
```
bastilian: Having this defined as a scope in `FactName` would make it reusable and composable. We can test this better in the model tests as well.

app/services/fact_cleaner.rb
```diff
@@ -0,0 +1,44 @@
...
+      @to_delete += batch.select do |fact|
+        fact.fact_values.empty?
```
bastilian: This as well might be good to have as a scope, which then can be combined with others.

app/services/fact_cleaner.rb
```diff
@@ -0,0 +1,44 @@
+class FactCleaner
+  delegate :logger, :to => :Rails
+
+  def clean!
+    @to_delete = []
```
bastilian: This might turn into an issue if the the cleaner lives longer, but we should be able to just have a ActiveRecord result-set for delete! to act on.

lib/tasks/facts.rake
```diff
@@ -0,0 +1,17 @@
...
+After cleaning of such facts they should no longer appear in auto-completion suggestions.
+
+Examples:
+  # foreman-rake facts:clean
+END_DESC
+
+  task :clean => :environment do
+    puts 'Starting orphaned facts clean up...'
+    count = FactCleaner.new.clean!
+    puts "Finished, cleaned #{count} facts"
```
bastilian: If the log statement would be in `FactCleaner` itself `clean!` would not need to return the count.

app/services/fact_cleaner.rb
```diff
@@ -0,0 +1,44 @@
...
+  def find_compose_orphaned_facts
+    FactName.where(:compose => true).find_in_batches do |batch|
+      @to_delete += batch.select do |compose|
+        compose.descendants.all? { |fact| @to_delete.include?(fact) }
+      end
+    end
+  end
+
+  def delete!
+    @to_delete.each do |fact|
```
bastilian: Using `destroy_all` on results might be better. This does mean we can't yield what fact it currently attempts to remove when running the task, but this can be seen in the logs of queries.

app/services/fact_cleaner.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def find_leaf_orphaned_facts
+    FactName.where(:compose => false).includes(:fact_values).find_in_batches do |batch|
+      @to_delete += batch.select do |fact|
+        fact.fact_values.empty?
+      end
+    end
+  end
+
+  def find_compose_orphaned_facts
```
bastilian: This too might be in better context under the `FactName` model. Turning the `select` condition into a arel query might be a bit brain-wrenching, but doable.

test/unit/fact_cleaner_test.rb
```diff
@@ -0,0 +1,38 @@
...
+    @root_6_fact = FactoryGirl.create(:fact_name)
+    @root_6_fact_value = FactoryGirl.create(:fact_value, :fact_name => @root_6_fact)
+  end
+
+  test "searching orphaned root facts" do
+    cleaner.clean!
+    remaining = FactName.where(:name => [@root_1_fact, @root_2_fact, @root_3_fact, @root_4_fact,
+                                         @root_5_fact, @root_3_child_fact, @root_4_child_1_fact,
+                                         @root_4_child_2_fact, @root_6_fact].map(&:name))
+    assert_include remaining, @root_4_fact
```
bastilian: These assertion could be split up into separate tests in `context`s, to make them more descriptive of what they are testing. They can then also have their own `setup` and `let`s specific to their cases.

test/unit/fact_cleaner_test.rb
```diff
@@ -0,0 +1,38 @@
+require 'test_helper'
+
+class FactCleanerTest < ActiveSupport::TestCase
+  let(:cleaner) do
```
bastilian, line 4: I'd prefer `subject {}`

bastilian (changes requested):
Thanks @ares! I've taken another look. Here are my notes in addition to the line comments.

At the moment I feel `FactCleaner` knows more than it should about `FactName` and `FactValue`.
This is reflected in the complexity of the `setup` for the test.
A nicer implementation would allow the tests to merely verify expectations of calls to `FactName`

The finder methods would be better when implemented as scopes. This also puts them then into the right place to describe what an orphan fact is.
As private methods in the `FactCleaner` they also might get overlooked in case that definition changes.

As mentioned above, the setup for the test feels too long and hard to understand, 
but moving the finders to the model and testing them there, should remove complexity from that.

The test for `clean!` itself asserts too much, each test should only assert one thing, two at most.
As a result of that rule of thumb, the tests will become much more readable and contribute to documenting how `FactCleaner` works and what it does.

Instance variables in tests can become a problem, preferring `let() {}` and `subject {}` makes their scope and lifetime predictable.
Using `describe {}` and `context {}` blocks help in that regard even more and improve readability of the tests as well.

These would be my suggestions for improving this PR.

app/services/fact_cleaner.rb
ares: the only public method is `clean!`, I might be missing something but every time someone calls it, it allocates new empty array so I don't see the potential issue.

app/services/fact_cleaner.rb
ares: done everywhere

app/services/fact_cleaner.rb
ares: I don't think `FactName.where(:compose => false).includes(:fact_values)` should be put in a scope? That fact we include fact_values is specific to our local usage. The condition itself could be useful - added `.leaves` and `.composes`.

app/services/fact_cleaner.rb
ares: How would you put that into scope especially until we have [rails 5 left_outer_jons](http://guides.rubyonrails.org/active_record_querying.html#left-outer-joins)? I think in this case it's not worth it, especially if we only need it for one maintenance task. I'm happy to revisit when we are on Rails 5 or there's at least one use case elsewhere to find orphaned facts.

app/services/fact_cleaner.rb
ares: the fact we search for it in batches is specific to this task that is expect to deal with big number of records, if we only extracted the condition in form of scope, then argument from above applies (not worth at least until left outer joins)

app/services/fact_cleaner.rb
ares: When users run Foreman they usually don't have sql logger enabled, for me it's easier to tell them how to enable debug log and rerun the task to see which fact failed than how to add sql logger, also it's more prone to errors since whitespaces in yaml matter. `destroy_all` does not bring much value, it instantiates every record and run all callback anyway.

lib/tasks/facts.rake
ares: logger is not user interface for running the script, puts print it to stdout

test/unit/fact_cleaner_test.rb
ares: Believe me I'd prefer all the goodness of spec-like syntax too, but I was pushed back several times :-) I'm happy we now use `let` and `describe` but rest usually does not slip in.

test/unit/fact_cleaner_test.rb
ares: Before I split it into contexts specifying befores etc I converted it into separate tests which makes it more readable. Coverage remains the same. As noted earlier, most of tests in Foreman do not use spec syntax capabilities, I made the mistake once and then had to rewrite all role tests from spec to testunit syntax :-)

ares:
@bastilian I incorporated what made sense to me, I tried to explain my view on the rest. Please let me know if you want me to do further modifications.

app/services/fact_cleaner.rb
bastilian: It's not an issue now, when `clean!` is only called once in the lifetime of the cleaner, but when using it elsewhere, possibly cached and `clean!` gets called multiple times. A situation could occur, where `@to_delete` is reset to an empty array while other methods are still working with it and assuming a different state.

Using a local variable to pass to `delete!` could avoid these situations and issues as a result of that.

lib/tasks/facts.rake
bastilian: By "log statement" I meant the `puts` :).

Ideally both puts/log statements would be within the `FactCleaner`, so the statements are also present when it is run not in a rake task. To not mix puts and logger statements FactCleaner could have a `stdout` logger and a log method to log to both. Something like this:

```ruby
  def initialize(options = {})
     self.stdout = stdout_logger if options[:stdout]
  end 
 
  def stdout_logger
    user_log = Logger.new(STDOUT)
    user_log.level = Logger::INFO
    user_log
  end

  def log(message, level = :info)
    stdout.send(level, message) if stdout
    logger.send(level, message)
  end
```

`clean!` could then just simply return true or nothing on a successful run and we have both logging and output for the user.

The code for the rake task would then be like this:

```ruby
cleaner = FactCleaner.new(stdout: true)
cleaner.clean!
```

Most methods with an exclamation mark return the modified object or for example with `create!` and `save!` it is more a way of saying "please, let me  know if there is something wrong", in this case though, since `FactCleaner` is a service I'd expect the `!` to be a sign to tell wether or not it is done.

bastilian (changes requested):
You are right about the scope this maybe to much for now and the use of `subject` is just a question of taste.

However it would be nice to have the changes to `clean!` and the use of `@to_delete`, as I think it would make a nice round up of the implementation.

lib/tasks/facts.rake
ares: After the discussion on IRC, the motivation was rather the count returned by clean! than unifying output.

The convention I know from Programming Ruby book is "Methods that are "dangerous," or modify the receiver, might be named with a trailing "!". but I it's vague and especially in Rails it's used in other contexts. If you didn't expect that, others might be surprised as well so I'll add explicit reader.

ares:
@bastilian I updated the PR, please take another look.

bastilian (approved):
👍  i'll leave it to someone with commit rights to merge.

timogoebel:
Merged, thanks @ares. And thanks @bastilian for the review.

PR #4276, opened 2017-02-13
fixes #18481 - host index page loads faster

timogoebel:


dLobatog:
Thanks @timogoebel

lzap:
Nice.

ohadlevy:
:+1: indeed, but in all honestly, i would prefer we fix the root cause instead of the "short term" solution.

timogoebel:
> i would prefer we fix the root cause instead of the "short term" solution.

I fully agree. The index might prove valuable in any case.

PR #4277, opened 2017-02-14
Fixes #18468 - Include JS in libvirt compute profile form

bastilian:
When the form is loaded via XHR the needed JavaScript is not loaded and selecting of certain attributes fail.

The oVirt templates include their JS via the `javascript` helper to included it as `content_for` in the head, but that won't be loaded there and here when requesting via XHR.

I will look into the oVirt form as well and open an issue in case, but #18468 would be fixed by this PR.

Including the JavaScript other than in the head is not the most elegant solution, having the `javascript` helper either include directly or set as content_for might be a better approach, but might cause issues in other places.

### DIFF
--- a/app/views/compute_resources_vms/form/libvirt/_base.html.erb
+++ b/app/views/compute_resources_vms/form/libvirt/_base.html.erb
@@ -1,3 +1,5 @@
+<%= javascript_include_tag 'compute_resource' %>
+
 <%= javascript_tag("$(document).on('ContentLoad', tfm.numFields.initAll)"); %>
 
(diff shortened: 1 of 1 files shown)


dLobatog:
@bastilian Seems to work well - will merge if tests pass. Thanks!

dLobatog:
Errors in jenkins seem to have to do with Jenkins itself, not the tests - http://ci.theforeman.org/job/test_develop_pr_core/12052/#showFailuresLink merging.

dLobatog:
Thanks @bastilian

PR #4278, opened 2017-02-14
Fixes #18488 - API remove GET config_templates/build_pxe_default

swapab:


domcleal:
This is a supported but deprecated API, why remove it? I can't see any justification.

(It also seems unrelated to the error in the issue title.)

swapab:
@domcleal 
The first thought was to not allowing GET request on an update action altogether.
I removed this after consulting with @tstrachota to remove the deprecated GET request since it's more than a year old, but I think the commit came too soon.
I missed @tstrachota comment on IRC saying "compatibility is kept till two version back"

The whole idea was to force use of POST instead of GET for api/v2/config_templates/build_pxe_default.
But since the compatibility needs to be maintained, this PR can be closed.

---
The issue title is an error occurred due to incorrect method(GET) in conjunction with URL.
```
GET /api/v2/provisioning_templates/build_pxe_default calls
Api::V2::ConfigTemplatesController#show
```

domcleal:
> The whole idea was to force use of POST instead of GET for api/v2/config_templates/build_pxe_default.
> But since the compatibility needs to be maintained, this PR can be closed.

Both are supported in API v2, it was only deprecated in this version. If you want to close the PR, do.

> GET /api/v2/provisioning_templates/build_pxe_default calls

Yes, not a documented API.

swapab:
Closing based on above discussion.

PR #4279, opened 2017-02-14
fixes #18489 - test AuthSourceLdap missing and blank passwords

domcleal:
After 2348fc8e, providing an empty password during update would erase
the existing password. Other resource type tests were updated, but not
AuthSourceLdap, so these now reflect the current behaviour.

These started failing on Rails 5 as the test params passed converted a
value of nil (not really possible) to an empty string, causing the
password to be blanked. On develop, the original password was nil
and the update to another nil value was a no-op.


shlomizadok:
LGTM, APJ :+1:

dLobatog:
Thanks @domcleal

PR #4280, opened 2017-02-14
Fix #18464 - Add Redux to statistics page

matanwerbner:
This is a first step to replacing Flux with Redux


tbrisker:
ok to test

ohadlevy:
@gailsteiger please review thanks!

webpack/assets/javascripts/react_app/common/MountingService.js
```diff
@@ -4,25 +4,25 @@ import PowerStatusContainer from '../components/hosts/PowerStatusContainer';
...
   const components = {
     StatisticsChartsList: {
       type: StatisticsChartsList,
-      markup: <StatisticsChartsList data={data}/>
+      markup: <StatisticsChartsList store={store} data={data}/>
     },
     PowerStatusContainer: {
       type: PowerStatusContainer,
-      markup: <PowerStatusContainer url={data.url} id={data.id}/>
+      markup: <PowerStatusContainer store={store} url={data.url} id={data.id}/>
```
ohadlevy: i assume the other components (besides) statistics should not use the store?

webpack/assets/javascripts/react_app/components/charts/StatisticsChartsList.js
```diff
@@ -1,111 +1,43 @@
...
-import { STATUS } from '../../constants';
-import StatisticsStore from '../../stores/StatisticsStore';
-import StatisticsChartActions from '../../actions/StatisticsChartActions';
 import './StatisticsChartsListStyles.css';
+import store from '../../redux';
+import { connect } from 'react-redux';
+import * as StatisticsChartActions from '../../redux/actions/statistics';
 
 class StatisticsChartsList extends React.Component {
   constructor(props) {
```
ohadlevy: if the constructor is now noop, can't we remove it?

webpack/assets/javascripts/react_app/redux/actions/statistics.js
```diff
@@ -0,0 +1,19 @@
+import API from '../../API';
```
ohadlevy: are you still using the "old" api here?

webpack/assets/javascripts/react_app/redux/actions/statistics.js
```diff
@@ -0,0 +1,19 @@
...
+
+export const getStatisticsData = charts => dispatch => {
+  dispatch({ type: STATISTICS_DATA_REQUEST, payload: charts });
+  charts.map(c => c.url).forEach(url => {
+    API.get(url).then(
+      result => dispatch({ type: STATISTICS_DATA_SUCCESS, payload: result }),
+      (jqXHR, textStatus, errorThrown) => dispatch({
+        type: STATISTICS_DATA_FAILURE,
+        payload: errorThrown
+      })
```
ohadlevy: what about error?

webpack/assets/javascripts/react_app/redux/actions/statistics/index.js
```diff
@@ -0,0 +1,18 @@
+import API from '../../../API';
+import {
+  STATISTICS_DATA_REQUEST,
+  STATISTICS_DATA_SUCCESS,
+  STATISTICS_DATA_FAILURE
+} from '../../consts';
+
```
ohadlevy, line 7: is this a duplicate file? whats the diff between actions/statistics and actions/statistics/index.js?

webpack/assets/javascripts/react_app/redux/consts.js
```diff
@@ -0,0 +1,3 @@
+export const STATISTICS_DATA_REQUEST = 'STATISTICS_DATA_REQUEST';
+export const STATISTICS_DATA_SUCCESS = 'STATISTICS_DATA_SUCCESS';
```
ohadlevy, line 2: q: does it make sense to group const into one object? e.g.
```js
export const STATISTICS = {
  data: {
    request: '...',
   success: '..'
   ...
}

File: webpack/assets/javascripts/react_app/redux/reducers/index.js [hunk 7]
```diff
@@ -0,0 +1,6 @@
+import { combineReducers } from 'redux';
+import statistics from './statistics';
+
+export default combineReducers({
```
ohadlevy, line 4: is this a global file for all reduces? wont it grow quickly for every component?

webpack/assets/javascripts/services/dal/statistics.js
```diff
@@ -0,0 +1,8 @@
+import axios from 'axios';
```
ohadlevy: not used?

webpack/assets/javascripts/react_app/redux/reducers/statistics/index.js
```diff
@@ -0,0 +1,35 @@
...
+  [STATISTICS_DATA_SUCCESS](state, { payload: { data, id } }) {
+    const newChartsArray = flatMap(
+      state.charts, (chart) => chart.id === id ?
+        { ...chart, data, status: STATUS.RESOLVED } :
+        chart
+    );
+
+    return state.set('charts', newChartsArray);
+  },
+  [STATISTICS_DATA_FAILURE](state, { payload: { error, id } }) {
```
ohadlevy: are these reducers being tested?

webpack/assets/javascripts/react_app/common/MountingService.js
matanwerbner: yup...will fix

webpack/assets/javascripts/react_app/components/charts/StatisticsChartsList.js
matanwerbner: yes

webpack/assets/javascripts/react_app/redux/consts.js
matanwerbner: this is more common approach for action names...
that way you can explicitly import the action names you need, which makes the code more readable as the reader sees which actions the code will use

webpack/assets/javascripts/react_app/redux/reducers/index.js
matanwerbner: this is only the root of the state tree.. it may grow, but not much, as the leafs may also export "combineReducers"

webpack/assets/javascripts/react_app/redux/actions/statistics.js
matanwerbner: still using jquery for ajax calls, i figure maybe this should be a seperate issue and discussion.

webpack/assets/javascripts/react_app/redux/actions/statistics.js
matanwerbner: error is the line starting with (jqXHR ... 
in then() the first parameter is resolve, the second is reject

webpack/assets/javascripts/react_app/redux/reducers/statistics/index.js
matanwerbner: yes, left out in the commit, see below

webpack/assets/javascripts/react_app/redux/actions/statistics/index.js
matanwerbner: right, this is the correct one. actions/statistics.js has been removed.

webpack/assets/javascripts/react_app/redux/actions/statistics.js
```diff
@@ -0,0 +1,16 @@
...
+
+export const getStatisticsData = charts => dispatch => {
+  dispatch({ type: STATISTICS_DATA_REQUEST, payload: charts });
+  charts.map(c => c.url).forEach(url => {
+    API.get(url).then(
+      result => dispatch({ type: STATISTICS_DATA_SUCCESS, payload: result }),
+      (jqXHR, textStatus, errorThrown) => dispatch({ type: STATISTICS_DATA_FAILURE, payload: errorThrown })
+    );
+  });
+};
```
gailsteiger: Am I missing something or isn't this file redundant with 'redux/actions/statistics/index.js' below?

matanwerbner: yes it is , it's removed by now

webpack/assets/javascripts/react_app/redux/actions/statistics.js
gailsteiger: +1 for reviewing use of jquery for ajax calls

webpack/assets/javascripts/react_app/redux/consts.js
gailsteiger: I have a bias towards using only one constants file. Also, prefer a name that doesn't have a spelling error.

gailsteiger:
Kudos Matan. It works, and to the extent that I understand, looks good.

There are a couple of instances where I have needed to use local/session storage to 'maintain state' following page navigation. You can see that with Notifications and it will also be necessary with Toasts. Is this going to be a problem?

webpack/assets/javascripts/react_app/redux/consts.js
matanwerbner: not sure i'm following you :)

matanwerbner:
Thanks, 
Initial state is built in to redux, as you can see in the store - each reducer accepts his initial state, which often comes from local storage.

gailsteiger:
We should settle on a consistent way to handle test file naming and directory location. I'll have another look  at jest conventions. In most cases I don't have strong opinions but we should use the same convention.

gailsteiger:
Where would we update local storage when state changes?  For instance, the navigation drawer (if open) should look the same after navigation as before. The user can open/close the drawer or open a different accordion panel before the navigation. This will need to be reflected in local storage before the navigation.

matanwerbner:
side effects such as API calls, session storage or similar, should be handled in the action creators, please see how the API is called on redux/actions/statistics/index.js
it would be almost the same handling sessionStorage: make the side-effect, then dispatch the actual action.

gailsteiger:
Is webpack/assets/javascripts/react_app/actions/StatisticsChartActions.js being used. Likewise for API.getStatisticsData

matanwerbner:
good point, i'll remove those

gailsteiger:
Looks like webpack/assets/javascripts/react_app/stores/StatisticsStore.js isn't being used either

webpack/assets/javascripts/react_app/redux/actions/statistics/index.js
```diff
@@ -0,0 +1,18 @@
...
+  dispatch({ type: STATISTICS_DATA_REQUEST, payload: charts });
+  charts.forEach(c => {
+    API.get(c.url).then(
+      result => dispatch({ type: STATISTICS_DATA_SUCCESS, payload: result }),
+      (jqXHR, textStatus, error) => dispatch(
+        { type: STATISTICS_DATA_FAILURE, payload: { error, id: c.id } }
+      )
+    );
+  });
+};
```
gailsteiger: Will the above replace 'webpack/assets/javascripts/react_app/actions/ServerActions.js' in the event that we completely replace flux with redux.

matanwerbner: absolutley

webpack/assets/javascripts/react_app/components/charts/spec/fixtures.js
```diff
@@ -0,0 +1,14 @@
+export const statisticsData = [
+  {
+    id: 'operatingsystem',
+    title: 'OS Distribution',
+    url: 'statistics/operatingsystem',
```
ohadlevy: this spec does not include actually any data, is that on purpose?

matanwerbner: Yes, and It was a dilema, i thought that the way components are currently built, data-specific tests may belong in the chartbox component, since the list has a job of rendering x panels for x items in the state.
tests of this component can be improved for sure, including a mount test to see if actions are fired properly and empty store gets proper values when given data.

matanwerbner:
made the reducer syntax more simple with swtich / case, removed "spec" libraries to make file structue more uniform, and removed "status" property from the state ( it is implicit ).

i'll be happy to answer any other suggestions, if not this PR is complete by me.

ohadlevy:
@matanwerbner can you please remove the statistics actions from webpack/assets/javascripts/react_app/actions/ServerActions.js ?
there are references also to the constant file

webpack/assets/javascripts/react_app/redux/index.js
```diff
@@ -0,0 +1,12 @@
+import { applyMiddleware, createStore } from 'redux';
+import thunk from 'redux-thunk';
+import createLogger from 'redux-logger';
+import reducer from './reducers';
+
+const logger = createLogger();
+
+export default createStore(
+  reducer,
+  window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__(),
```
ohadlevy, line 10: is this part of the PR?

webpack/assets/javascripts/react_app/redux/actions/statistics/index.js
```diff
@@ -0,0 +1,18 @@
+import API from '../../../API';
+import {
+  STATISTICS_DATA_REQUEST,
+  STATISTICS_DATA_SUCCESS,
+  STATISTICS_DATA_FAILURE
+} from '../../consts';
+
+export const getStatisticsData = charts => dispatch => {
+  dispatch({ type: STATISTICS_DATA_REQUEST, payload: charts });
+  charts.forEach(c => {
```
ohadlevy: nitpick: can you use chart vs c for more meaningful var names?

webpack/assets/javascripts/react_app/redux/reducers/statistics/index.js
```diff
@@ -0,0 +1,29 @@
...
+  charts: []
+});
+
+export default (state = initialState, action) => {
+  const { payload } = action;
+
+  switch (action.type) {
+    case STATISTICS_DATA_REQUEST: return state.set('charts', payload);
+    case STATISTICS_DATA_SUCCESS:
+      return state.set('charts', state.charts.map(chart => chart.id === action.payload.id ?
```
ohadlevy: nitpick: action.payload.id can simply be payload.id ?

webpack/assets/javascripts/react_app/components/charts/StatisticsChartsList.js
```diff
@@ -1,111 +1,49 @@
...
-  constructor(props) {
-    super(props);
-
-    this.state = {charts: this.stateSetup(this.props.data)};
-
-    helpers.bindMethods(this, [
-      'onChange',
-      'onError']
-    );
+const getStatusFromChart = (chart) => {
```
ohadlevy, line 9: @gailsteiger any concerns about this change?

ohadlevy, line 9: @matanwerbner also, why isnt this a function within the react class?

webpack/assets/javascripts/react_app/redux/index.js
matanwerbner: yes, this enabled redux devtools on the browser
http://extension.remotedev.io/

ohadlevy:
@matanwerbner note webpack/assets/javascripts/react_app/constants.js still reference statistics.

webpack/assets/javascripts/react_app/components/charts/StatisticsChartsList.js
matanwerbner: this is a utility function. i dont think it belongs in the component, since it is not react-related.
anyway we should refactor this to avoid rendering the "smart" chartbox component.
after refactoring this function will go away.
i think it can also be in a utilities file, along with the translation strings.

matanwerbner:
@ohadlevy fixed.

webpack/assets/javascripts/react_app/components/charts/StatisticsChartsList.spec.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import StatisticsChartsList from './StatisticsChartsList';
+import { statisticsData } from './StatisticsChartsList.fixtures';
+
```
ohadlevy: nitpick: extra line

webpack/assets/javascripts/react_app/redux/reducers/statistics/statistics.fixtures.js
```diff
@@ -0,0 +1,38 @@
...
+  {
+    id: 'operatingsystem',
+    title: 'OS Distribution',
+    url: 'statistics/operatingsystem',
+    search: '/hosts?search=os_title=~VAL~'
+  }
+];
+export const response = {
+  id: 'operatingsystem',
+  data: [['redHat 3', 2]]
```
ohadlevy: nitpick: redHat is not a valid brand, either write RHEL or RedHat :)

webpack/assets/javascripts/react_app/redux/reducers/statistics/statistics.fixtures.js
```diff
@@ -0,0 +1,38 @@
...
+};
+
+export const error = 'some error happened';
+
+export const stateBeforeResponse = Immutable({
+  charts: request
+});
+
+export const stateAfterSuccess = Immutable({
+  charts: request.map(c => ({
```
ohadlevy: same nitpick about one chart variable names :)

ohadlevy:
@matanwerbner besides minor nitpicks big :+1:  for this change.

I hope @gailsteiger is around to double check me, but from my side :+1: to merge.

webpack/assets/javascripts/react_app/redux/actions/statistics/index.js
gailsteiger: +1

webpack/assets/javascripts/react_app/components/charts/StatisticsChartsList.js
gailsteiger: OK

gailsteiger:
OK to merge. Super!

webpack/assets/javascripts/react_app/components/charts/StatisticsChartsList.js
gailsteiger: @matanwerbner could you elaborate on refactoring to avoid rendering the "smart" chartbox component?

matanwerbner: i meant that i believe the list component should contain logic as to what chart it renders.
it should choose between rendering a chart, a loading panel,  or an error panel,  based on the item it is handling.
so the chart box might be not neccesary

ohadlevy:
thanks @matanwerbner !

webpack/assets/javascripts/react_app/components/charts/StatisticsChartsList.js
gailsteiger: Based on what you're saying, the `Loader` would not be necessary but I'm not sure about the `ChartBox`. 

Generally, what thoughts about the pros and cons of using components like the [`Loader`](https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/components/common/Loader.js) or [`Fade`](https://github.com/gailsteiger/foreman/blob/pf-react-toast/webpack/assets/javascripts/react_app/components/common/Fade.js) which wrap other components and manage part of the wrapped components' rendering logic as opposed to including the relevant logic in the wrapped components themselves?

matanwerbner: i think wrapper components are great, it's a very common pattern in react and functional programming, in common usecases.
i feel that there are cases where it's not required, and adds a level of complexity to the code.
if we have common functionality that spans several components, it's worth considering making a wrapper (if normal solutions such as css dont work out)...

matanwerbner: after looking at loader, i think i would prefer manually adding a loading component for components that require it. 
makes it easier to directly describe and understand the logic for showing/hiding the loader, rather then requiring status parameter

gailsteiger: good point

PR #4281, opened 2017-02-14
Fixes #18490 - run eslint on all files in webpack

tbrisker:


gailsteiger:
what other files other than .js do you have in mind?

tbrisker:
@gailsteiger for some reason not all files are matched by the existing pattern, even though they should.

gailsteiger:
Wow. Which files is it missing? Nice catch.

Does ` webpack/assets/**/*/*.js` work?

tbrisker:
@gailsteiger I didn't manage to figure out the exact condition, but when testing #4280 a lot of the files weren't being linted by `npm run lint`, which led to a lot of barking from the hound.

tbrisker:
@gailsteiger Actually I know why - bash was interpolating the `*`'s so only the top level was being linted, nothing under the other folders.

package.json
```diff
@@ -61,7 +61,7 @@
     "select2": "~3.5.2-browserify"
   },
   "scripts": {
-    "lint": "./node_modules/.bin/eslint -c .eslintrc webpack/assets/**/*.js || exit 0",
+    "lint": "./node_modules/.bin/eslint -c .eslintrc webpack/assets/**/* || exit 0",
```
ohadlevy: this will catch also css files and fail.

i believe ```webpack/**/*/*js``` works for me.

tbrisker: no, eslint only checks .js files unless --ext is used.
Also that does not work, you should see a warning on API.js.

tbrisker:
Amended to remove the glob, eslint can be passed an entire directory as a target.

PR #4282, opened 2017-02-14
refs #18468 - move compute_resources JS to compute profile index

domcleal:
The CR JS should always be ready and loaded on the compute profile
index, as the new attributes form is rendered into the existing page
after the profile is created.

Because the new attributes form is loaded via the two-pane #content
replacement, the additional JavaScript tag specified in
compute_profiles/show is not automatically loaded into the page.

This fixes the issue for all compute resources, not only Libvirt. It
also uses the `javascript` header to properly track the asset with
Turbolinks so it's unloaded later.

### DIFF
--- a/app/views/compute_profiles/index.html.erb
+++ b/app/views/compute_profiles/index.html.erb
@@ -1,3 +1,5 @@
+<%= javascript 'compute_resource' %>
+
 <% title _('Compute profiles') %>
 
 <% title_actions new_link(_('Create Compute Profile')),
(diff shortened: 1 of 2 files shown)


ohadlevy (approved):

bastilian:
This makes the call to `javascript` in [`ovirt/_base.html` ](https://github.com/theforeman/foreman/blob/develop/app/views/compute_resources_vms/form/ovirt/_base.html.erb#L3) redundant as well.

dLobatog:
Thanks @domcleal

PR #4283, opened 2017-02-14
Fixes #17228 - Select templates as global defaults

xprazak2:
This should allow user to select templates that will be used
for building PXE Default on TFTP proxy. These templates will
not be updated from seeds during updates.

### DIFF
--- a/app/models/provisioning_template.rb
+++ b/app/models/provisioning_template.rb
@@ -50,6 +50,8 @@ class ProvisioningTemplate < Template
     end
   }
 
+  scope :of_kind, ->(kind) { joins(:template_kind).where("template_kinds.name" => kind) }
+
   def self.template_ids_for(hosts)
     hosts = hosts.with_os.uniq
     oses = hosts.pluck(:operatingsystem_id)
@@ -105,24 +107,39 @@ class ProvisioningTemplate < Template
     template.is_a?(ProvisioningTemplate) ? template : nil
   end
 
+  def self.find_global_default_template(name, kind)
+    ProvisioningTemplate.unscoped.joins(:template_kind).where(:name => name, "template_kinds.name" => kind).first
+  end
+
+  def self.global_default_name(kind)
+    "#{kind} global default"
+  end
+
+  def self.global_template_name_for(kind, renderer)
+    global_setting = Setting.find_by(:name => "global_#{kind}")
+    return global_setting.value if global_setting && global_setting.value.present?
+    global_template_name = global_default_name(kind)
+    renderer.logger.info "Could not find user defined global template from Settings for #{kind}, falling back to #{global_template_name}"
+    global_template_name
+  end
+
   def self.build_pxe_default(renderer)
     return [422, _("No TFTP proxies defined, can't continue")] if (proxies = SmartProxy.with_features("TFTP")).empty?
-
     error_msgs = []
+    used_templates = []
     TemplateKind::PXE.each do |kind|
-      default_template_name = "#{kind} global default"
-      if (default_template = ProvisioningTemplate.find_by_name(default_template_name)).nil?
-        error_msg = _("Could not find a Configuration Template with the name \"%s global default\", please create one.") % kind
+      global_template_name = global_template_name_for(kind, renderer)
+      if (default_template = find_global_default_template global_template_name, kind).nil?
+        error_msg = _("Could not find a Configuration Template with the name \"%s\", please create one.") % global_template_name
       else
         begin
           @profiles = pxe_default_combos
           allowed_helpers = Foreman::Renderer::ALLOWED_GENERIC_HELPERS + [ :default_template_url ]
           menu = renderer.render_safe(default_template.template, allowed_helpers, :profiles => @profiles)
         rescue => exception
-          Foreman::Logging.exception("Cannot render '#{default_template_name}'", exception)
+          Foreman::Logging.exception("Cannot render '#{global_template_name}'", exception)
           error_msgs << "#{exception.message} (#{kind})"
         end
-
         return [422, error_msg] unless error_msg.empty?
         proxies.each do |proxy|
           begin
@@ -130,17 +147,18 @@ class ProvisioningTemplate < Template
             tftp.create_default(kind, {:menu => menu})
             fetch_boot_files_combo(tftp)
           rescue => exception
-            Foreman::Logging.exception("Cannot deploy rendered template '#{default_template_name}' to '#{proxy}'", exception)
+            Foreman::Logging.exception("Cannot deploy rendered template '#{global_template_name}' to '#{proxy}'", exception)
             error_msgs << "#{proxy}: #{exception.message} (#{kind})"
           end
         end
       end
+      used_templates << global_template_name
     end
 
     if error_msgs.empty?
-      [200, _("PXE Default file has been deployed to all Smart Proxies")]
+      [200, _("PXE files for templates %s have been deployed to all Smart Proxies") % used_templates.to_sentence]
     else
-      [500, _("There was an error creating the PXE Default file: %s") % error_msgs.join(", ")]
+      [500, _("There was an error creating the PXE file: %s") % error_msgs.join(", ")]
     end
   end
 
@@ -177,6 +195,13 @@ class ProvisioningTemplate < Template
       end
     end
     combos.sort_by! { |profile| [profile[:hostgroup].title, profile[:template]] }
+  end
+
+  def global_default?
+    return unless template_kind
+    setting = Setting.find_by :name => "global_#{template_kind.name}"
+    return unless setting
+    setting.value == name
   end
 
   private
--- a/app/models/setting/provisioning.rb
+++ b/app/models/setting/provisioning.rb
@@ -1,5 +1,14 @@
 require 'facter'
 class Setting::Provisioning < Setting
+  def self.default_global_labels
+    TemplateKind::PXE.map do |pxe_kind|
+      "global_#{pxe_kind}"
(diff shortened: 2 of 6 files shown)


xprazak2:
Tries to find template from settings and if fails, it tries to fall back on the "`#{kind} global default"`

app/controllers/provisioning_templates_controller.rb
```diff
@@ -3,8 +3,9 @@ class ProvisioningTemplatesController < TemplatesController
   helper_method :documentation_anchor
 
   def build_pxe_default
-    status, msg = ProvisioningTemplate.authorized(:deploy_provisioning_templates).build_pxe_default(self)
+    status, msg, info_msgs = ProvisioningTemplate.authorized(:deploy_provisioning_templates).build_pxe_default(self)
     status == 200 ? notice(msg) : error(msg)
+    info_msgs.each { |m| logger.info m }
```
timogoebel: Why not log these messages directly?

db/seeds.d/07-provisioning_templates.rb
```diff
@@ -12,6 +12,14 @@
   raise "Unable to create template kind: #{format_errors kinds[type]}" if kinds[type].nil? || kinds[type].errors.any?
 end
 
+def is_global_default? input
+  return unless input[:template_kind]
+  s = Setting.find_by :name => "global_#{input[:template_kind].name}"
```
timogoebel: nitpick, prefer `setting` over `s`. Makes it easier to read imho.

app/models/provisioning_template.rb
```diff
@@ -105,42 +105,51 @@ def self.find_template(opts = {})
...
 
+  def self.build_pxe_default(renderer)
+    return [422, _("No TFTP proxies defined, can't continue"), []] if (proxies = SmartProxy.with_features("TFTP")).empty?
     error_msgs = []
+    info_msgs = []
     TemplateKind::PXE.each do |kind|
-      default_template_name = "#{kind} global default"
-      if (default_template = ProvisioningTemplate.find_by_name(default_template_name)).nil?
-        error_msg = _("Could not find a Configuration Template with the name \"%s global default\", please create one.") % kind
+      setting_name = "global_#{kind}"
```
timogoebel: I'd suggest extracting this to a dedicated method to reduce complexity of `build_pxe_default`.

timogoebel (changes requested):
@xprazak2: Thanks. This will be greatly appreciated by users. Left some comments inline. :-)

xprazak2:
Thanks for comments, logging is now done directly and I also made minor style changes as suggested.

db/seeds.d/07-provisioning_templates.rb
```diff
@@ -12,6 +12,14 @@
   raise "Unable to create template kind: #{format_errors kinds[type]}" if kinds[type].nil? || kinds[type].errors.any?
 end
 
+def is_global_default? input
```
timogoebel: Why not put this on the template model itself?
You could then do something like this:

```ruby
if (t = ProvisioningTemplate.find_by_name(input[:name])) && !audit_modified?(ProvisioningTemplate, input[:name])
  next if t.is_global_default?
```

app/models/provisioning_template.rb
```diff
@@ -105,42 +105,53 @@ def self.find_template(opts = {})
...
+
   def self.build_pxe_default(renderer)
     return [422, _("No TFTP proxies defined, can't continue")] if (proxies = SmartProxy.with_features("TFTP")).empty?
-
     error_msgs = []
     TemplateKind::PXE.each do |kind|
-      default_template_name = "#{kind} global default"
-      if (default_template = ProvisioningTemplate.find_by_name(default_template_name)).nil?
-        error_msg = _("Could not find a Configuration Template with the name \"%s global default\", please create one.") % kind
+      global_setting = Setting.find_by(:name => "global_#{kind}")
```
timogoebel: I'd suggest extracting this to a dedicated method, something like

```ruby
def global_template_name_for(kind, renderer)
  global_setting = Setting.find_by(:name => "global_#{kind}")
  return global_setting.value if global_setting && global_setting.value.present?

  global_template_name = global_default_name(kind)
  renderer.logger.info "Could not find user defined global template from Settings for #{kind}, falling back to #{global_template_name}"
  global_template_name
end
```

timogoebel (changes requested):
@xprazak2 : Thanks, Ondřej. I left some more comments inline.

xprazak2:
I moved the `global_default?` method from seeds to provisioning templates and extracted `global_template_name_for` method.

app/models/provisioning_template.rb
```diff
@@ -105,42 +105,56 @@ def self.find_template(opts = {})
...
+            Foreman::Logging.exception("Cannot deploy rendered template '#{global_template_name}' to '#{proxy}'", exception)
             error_msgs << "#{proxy}: #{exception.message} (#{kind})"
           end
         end
       end
     end
 
     if error_msgs.empty?
-      [200, _("PXE Default file has been deployed to all Smart Proxies")]
+      [200, _("PXE file has been deployed to all Smart Proxies")]
```
timogoebel: Does it make sense to log the template name that was deployed? Same below.

app/models/setting/provisioning.rb
```diff
@@ -46,4 +55,10 @@ def validate_safemode_render(record)
       record.errors[:base] << _("Unable to disable safemode_render when bmc_credentials_accessible is disabled")
     end
   end
+
+  def self.default_global_templates
+    TemplateKind::PXE.map do |pxe_kind|
+      self.set("global_#{pxe_kind}", N_("Global default %s template that will not be affected by upgrades") % pxe_kind, nil, N_("Global default %s template") % pxe_kind)
```
timogoebel: Can we show a select box instead of an input field? That might be easier for users to select an appropriate template.

timogoebel (commented):
@xprazak2: Thanks for extracting the code. It's much easier to read, now. While testing this, I have two more questions.

xprazak2:
I added the names of deployed templates to the message and changed text field to select box in settings. Not that select boxes for settings do not have filter, so finding/selecting the right item when users have large number of templates may be problematic, that is why I did not go for select box from the beginning

app/models/provisioning_template.rb
```diff
@@ -105,42 +107,58 @@ def self.find_template(opts = {})
...
             error_msgs << "#{proxy}: #{exception.message} (#{kind})"
           end
         end
       end
+      used_templates << global_template_name
     end
 
     if error_msgs.empty?
-      [200, _("PXE Default file has been deployed to all Smart Proxies")]
+      [200, _("PXE file for templates %s has been deployed to all Smart Proxies") % used_templates.to_sentence]
```
timogoebel: PXE files?

app/models/setting/provisioning.rb
```diff
@@ -46,4 +55,11 @@ def validate_safemode_render(record)
       record.errors[:base] << _("Unable to disable safemode_render when bmc_credentials_accessible is disabled")
     end
   end
+
+  def self.default_global_templates
+    TemplateKind::PXE.map do |pxe_kind|
+      templates = Proc.new { { "" => "" }.merge Hash[ProvisioningTemplate.unscoped.of_kind(pxe_kind).map { |tmpl| [tmpl.name, tmpl.name] }] }
+      self.set("global_#{pxe_kind}", N_("Global default %s template that will not be affected by upgrades") % pxe_kind, nil, N_("Global default %s template") % pxe_kind, nil, { :collection =>  templates })
```
timogoebel: I'd suggest changing the text a little so that it's more clear, that this is the template that gets deployed to the tftp proxies and not just the template that is skipped when upgrading.

How about?
"Global default %s template. This template gets deployed to all configured TFTP servers. It will not be affected by upgrades."

timogoebel (changes requested):
@xprazak2 : Thanks again. I tested this and it works well. Two minor nitpicks, then this is ready to 🚢

xprazak2:
I rebased and changed the messages to be more user-friendly.

timogoebel:
[test katello]

travis is fine although github doesn't know this.

timogoebel (approved):
@xprazak2 : Thanks. Looks good to me now! I'll merge once the Katello tests reran.

timogoebel:
@xprazak2 : The Katello tests are still failing. Any idea, why?

xprazak2:
No idea, I will investigate.

dLobatog:
@xprazak2 @timogoebel I'll merge. The Katello tests are failing due to https://github.com/Katello/katello/pull/6447 not being merged yet.

db/seeds.d/07-provisioning_templates.rb
```diff
@@ -95,6 +95,8 @@
     contents = File.read(File.join("#{Rails.root}/app/views/unattended", input.delete(:source)))
 
     if (t = ProvisioningTemplate.unscoped.find_by_name(input[:name])) && !audit_modified?(ProvisioningTemplate, input[:name])
+      next if t.global_default?
```
ares, line 98: @xprazak2 why was this skipping added? if the recommended workflow now is to clone the template and change the link it via settings why do we skip this? if user is fine with the default template (which would be locked), it would not be updated it with newer version I think

timogoebel: I think, we should remove this when we actually implement the new workflow (i.e. lock the shipped templates).

ares: alright, makes sense, I'll add it to #4320, thanks for quick response

PR #3701, opened 2016-08-03
Fixes #15711 - Redesigned host info to providers.

ShimShtein:
Now you can add your own info by registering a host method that will return info hash.
#### Main files

`app/models/concerns/host_info.rb`: Core concern, responsible for registration and querying of different info providers.
`app/models/concerns/host_info_providers/static_info.rb`: Provider of the basic (and static) information, like hostgroup, interfaces e.t.c.
`app/models/concerns/host_info_providers/puppet_info.rb`: Provider for the puppet specific information, like smart class parameters and smart variables.
`app/models/concerns/host_params.rb`: Became a provider of (global) host parameters.
#### Classification

Removed the hierarchy from `Classification` and left the base class as a service class that knows how to calculate values for a given set of keys (the type of keys does not matter).
The logic that was stored in the child classes moved to PuppetInfo concern.

PR #4288, opened 2017-02-15
fixes #18512 - support sprockets-rails 3.x

domcleal:
The app.assets environment is no longer always configured, and instead
is only configured when the regular Rails environment is loaded. Loading
only the 'assets' group initializers isn't sufficient.

Because the asset group initializers are no longer run, the plugin API
has been extended with `precompile_assets` and `automatic_assets` to
store and optionally generate a list of assets for that plugin. It
deprecates using an assets initializer, modifying SETTINGS and
`config.assets.precompile` to configure the plugin asset list.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -30,7 +30,7 @@ gem 'net-ldap', '>= 0.8.0'
 gem 'net-ping', :require => false
 gem 'activerecord-session_store', '>= 0.1.1', '< 2'
 gem 'sprockets', '~> 3'
-gem 'sprockets-rails', '>= 2.3.3', '< 3'
+gem 'sprockets-rails', '>= 2.3.3', '< 4'
 gem 'responders', '~> 2.0'
 gem 'roadie-rails', '>= 1.1', (RUBY_VERSION < '2.2' ? '< 1.2' : '< 2')
 gem 'x-editable-rails', '~> 1.5.5'
--- a/app/registries/foreman/plugin.rb
+++ b/app/registries/foreman/plugin.rb
@@ -93,6 +93,7 @@ module Foreman #:nodoc:
       end
     end
 
+    prepend Foreman::Plugin::Assets
     prepend Foreman::Plugin::SearchOverrides
 
     def_field :name, :description, :url, :author, :author_url, :version, :path
--- a/app/registries/foreman/plugin/assets.rb
+++ b/app/registries/foreman/plugin/assets.rb
@@ -1,0 +1,69 @@
+module Foreman
+  class Plugin
+    module Assets
+      attr_reader :assets
+
+      def initialize(id)
+        super
+        @assets = []
+        @automatic_assets = true
+      end
+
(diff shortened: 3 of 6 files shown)


ohadlevy:
[test katello]

dLobatog (changes requested):
@domcleal I think the deprecation message showed up fine but it didn't fall back to the previous behavior. One of the failed tests in Katello precisely has to do with including assets this way (https://github.com/Katello/katello/blob/master/test/controllers/foreman/containers/steps_controller_test.rb#L25)  

http://ci.theforeman.org/job/test_develop_pr_katello/2816/database=postgresql,ruby=2.2,slave=fast/consoleFull contains the messages. I'll investigate what happened when loading foreman_docker + Katello.

domcleal:
> I think the deprecation message showed up fine but it didn't fall back to the previous behavior. One of the failed tests in Katello precisely has to do with including assets this way

The deprecation warning is when SETTINGS is populated, which is only used in the `plugin:assets:precompile` rake task. The development or test mode asset pipeline should continue to work fine.

The assertion in that test is probably too tight, it assumes that digests are disabled, which is not the case by default in 3.x.

ohadlevy:
@dLobatog ping?

dLobatog:
Sorry for the long turnaround, @domcleal you're right on https://github.com/theforeman/foreman/pull/4288#issuecomment-304845600. I changed the test on https://github.com/Katello/katello/pull/6841, as soon as it's merged I will trigger the test katello job again and hopefully merge.

dLobatog (approved):

mmoll:
[test katello]

jlsherrill:
[test katello]

dLobatog:
[test katello]

mmoll:
[test katello]

mmoll:
I have not the slightest idea, what's happening here, maybe one of the deprecations warnings is treated as an error at some point, leading to an incomplete DB...

dLobatog:
@mmoll I think I've seen the katello job failing like random on other PRs here in a similar way [test katello]

mmoll:
[test katello] again

mmoll:
I think this one is failing consistently...

mmoll:
https://github.com/Katello/bastion/pull/198 is related, BTW, but no idea if that fixes anything

mmoll:
[test katello]

jlsherrill:
[test katello]

jlsherrill:
[test katello]

jlsherrill:
[test katello]

dLobatog (approved):
This branch does not have host_info from https://github.com/theforeman/foreman/pull/3701/files which is needed to pass Katello tests. However the commit is included in https://github.com/theforeman/foreman/pull/4646 including authorship so I think it's safe to close this if tests pass on that one.

dLobatog:
Closed as it was merged with #4646

PR #4290, opened 2017-02-15
fixes #18476 - users have ssh keys

timogoebel:
This PR adds the ability for a user to have ssh keys. These keys can be used in templates and are available in ENC data.
A user without explicit permission can create/delete keys for himself.

image


db/migrate/20170214132230_create_ssh_keys.rb
```diff
@@ -0,0 +1,12 @@
+class CreateSshKeys < ActiveRecord::Migration
+  def change
+    create_table :ssh_keys do |t|
+      t.string :name, :limit => 255
+      t.string :key, :text, :limit => 5000
```
sean797: What is this text column user for?

timogoebel: it stores the actual ssh key, e.g. `ssh-rsa abcdef foreman@example.com`

sean797: Isn't that stored in `key` ?

```
katello=# \d ssh_keys
                                      Table "public.ssh_keys"
   Column    |            Type             |                       Modifiers                       
-------------+-----------------------------+-------------------------------------------------------
 id          | integer                     | not null default nextval('ssh_keys_id_seq'::regclass)
 name        | character varying(255)      | 
 key         | character varying           | 
 text        | character varying           | 
 fingerprint | character varying           | 
 user_id     | integer                     | 
 created_at  | timestamp without time zone | not null
 updated_at  | timestamp without time zone | not null
Indexes:
    "ssh_keys_pkey" PRIMARY KEY, btree (id)

Comment (timogoebel) [reply db/migrate/20170214132230_create_ssh_keys.rb]: Ah, good catch! Thanks.
This line should obviously read `t.text :key, :limit => 5000`

**Comment by timogoebel** [comment]
I just realised, we already use the `SSHKey` gem as a dependency. Changing the homemade code in favor of this gem.

File: app/models/user.rb [hunk 2]
```diff
@@ -38,6 +38,7 @@ class User < ActiveRecord::Base
   has_many :permissions,       :through => :filters
   has_many :cached_usergroup_members
   has_many :widgets, :dependent => :destroy
+  has_many :ssh_keys, :dependent => :destroy
```
ares, line 43: It would be great if you defined a Jail class and allow this for safe mode so so we could do something like `@host.owner.ssh_keys` in provisioning and job templates. Also it would be probably handy to have a Jail definition in SshKey class as well so we could print the public key and potentially fingerprint. Another bonus points for macro that would print public keys for a given user. No push, just an idea :-)

timogoebel: @ares : Good point. `owner` is currently now allowed in host's jail. I do remember, that there were some concerns about this. Can you think of some?

ares: Hm, I can't think of any. The tricky thing is that sometimes the owner is user, sometimes it's user group.

timogoebel:
Code now uses the `SSHKey` gem. Added Jail `SshKey`, `User` and `Usergroup`. `ssh_authorized_keys` now follows the law of demeter. `SshKey` now has a `length` attribute.

app/models/user.rb
timogoebel: Added the Jail classes.

app/models/ssh_key.rb
```diff
@@ -0,0 +1,66 @@
+class SshKey < ActiveRecord::Base
+  include Authorizable
```
ohadlevy, line 2: is there a way to reuse KeyPair model?

app/views/ssh_keys/_ssh_keys_tab.html.erb
```diff
@@ -0,0 +1,45 @@
...
+    </thead>
+    <tbody>
+      <% user.ssh_keys.each do |key| %>
+        <%= render 'ssh_keys/ssh_key_row', :f => f, :key => key %>
+      <% end %>
+    </tbody>
+  <% else %>
+    <tbody>
+      <tr>
+        <td class="blank-slate-pf">
```
ohadlevy, line 21: why not use the welcome partial instead?

webpack/assets/javascripts/foreman_ssh_keys.js
```diff
@@ -0,0 +1,12 @@
+import $ from 'jquery';
+
+export function autofillSshKeyName() {
```
ohadlevy, line 3: two comments:
1. missing tests for functions
2. maybe its just me, but I would love seeing it implemented via react instead ;)

app/views/ssh_keys/_ssh_keys_tab.html.erb
timogoebel: As discussed on IRC, this is unfortunately not possible as the welcome partial is only shown on the index controller. The ssh key list is only rendered as a tab on the users edit page.

webpack/assets/javascripts/foreman_ssh_keys.js
timogoebel: Added a test. As discussed on IRC, react is probably not worth it in this case.

app/models/ssh_key.rb
timogoebel: I'm reluctant on this one. It might be possible, but imho we try to model something completely different. There isn't any real code duplication now and the models behave differently. If we force them into a single model, it just gets harder to maintain and test this.

app/models/usergroup.rb
```diff
@@ -112,4 +116,12 @@ def ensure_last_admin_group_is_not_deleted
   def other_admins
     User.unscoped.only_admin.except_hidden - all_users
   end
+
+  def ssh_keys
+    users.map(&:ssh_keys).flatten
```
bastilian: You could use `users.flat_map(&:ssh_keys)` here.

app/controllers/api/v2/ssh_keys_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+  module V2
+    class SshKeysController < V2::BaseController
+      include Foreman::Controller::Parameters::SshKey
+      include Foreman::Controller::SshKeysCommon
+
+      wrap_parameters :ssh_key, :include => ssh_key_params_filter.accessible_attributes(parameter_filter_context)
+
+      before_action :find_resource, :only => %w{show destroy}
+
+      api :GET, "/users/:user_id/ssh_keys", N_("List all ssh keys for a user")
```
dLobatog: nitpick, 'SSH' uppercase in user-facing strings

app/controllers/concerns/foreman/controller/ssh_keys_common.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def find_user
+    if editing_self?
+      @user = User.current
+    else
+      @user = User.find(params[:user_id])
+    end
+  end
+
+  def editing_self?
```
dLobatog: This is duplicating logic already existing in the UsersMixin, extract to a shared concern or see if you can simply reuse the UsersMixin in both SSHKeys and User controllers

app/helpers/application_helper.rb
```diff
@@ -127,11 +127,19 @@ def link_to_function_if_authorized(name, function, options = {}, html_options =
   end
 
   def display_delete_if_authorized(options = {}, html_options = {})
+    if authorized_for(options)
```
dLobatog: nitpick, a guard clause is more clear   (https://github.com/bbatsov/ruby-style-guide#no-nested-conditionals)

```ruby
return "" unless authorized_for(options)
display_delete(options, html_options)
```

app/models/host/base.rb
```diff
@@ -57,6 +57,8 @@ def self.taxonomy_conditions
     scope :no_location,     -> { rewhere(:location_id => nil) }
     scope :no_organization, -> { rewhere(:organization_id => nil) }
 
+    delegate :ssh_authorized_keys, :to => :owner, :allow_nil => true
```
dLobatog, line 60: This one shouldn't delegate when the owner is an user group. Or alternatively, you can implement the method to fetch all ssh keys for all users in the user group.

```ruby
:if => ->(host) { host.owner_type == 'User' }
```

OR


```ruby
def ssh_authorized_keys
return owner.ssh_authorized_keys if owner_type == 'User'
owner.users.map(&:ssh_authorized_keys).flatten.uniq
end
```

Probably the 2nd option is better

app/models/ssh_key.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  private
+
+  def generate_fingerprint
+    self.fingerprint = nil
+    return unless self.key.present?
+    self.fingerprint = SSHKey.fingerprint(self.key)
+    true
+  rescue SSHKey::PublicKeyError
+    nil
```
dLobatog: Why?

dLobatog (changes requested):
@timogoebel Sorry didn't get to review the entire thing, but I left some comments already, feel free to fix them in the meantime, I'll finish it some time next week.

IMO - if we kept the API portion in another PR it'd get easier to review 😄

app/validators/ssh_key_validator.rb
```diff
@@ -0,0 +1,15 @@
...
+      record.errors[attribute] << _('is invalid')
+    end
+  end
+
+  private
+
+  def valid_ssh_public_key?(key)
+    SSHKey.valid_ssh_public_key?(key)
+  rescue SSHKey::PublicKeyError
+    false
```
dLobatog: I don't understand why this exception is rescued as `nil` in the model but `false` here.

app/models/host/base.rb
timogoebel: > Or alternatively, you can implement the method to fetch all ssh keys for all users in the user group.

I think, I did implement it for both `User` and `Usergroup`.

app/models/ssh_key.rb
timogoebel: If I return `false` here, Rails will stop processing all further validations. The docs suggest otherwise, but the tests completely blew.

app/validators/ssh_key_validator.rb
timogoebel: `false` is the intended and correct return value. See my other comment why I chose the other method to return `nil`.

timogoebel:
@bastilian, @dLobatog : Thanks for your comments. I pushed an updated version that should address them.

bastilian:
`.flat_map` FTW! 🎉

app/views/ssh_keys/_form.html.erb
```diff
@@ -0,0 +1,18 @@
...
+
+    <div class="alert alert-warning alert-block base">
+      <% [:fingerprint, :length].each do |field|%>
+        <% @ssh_key.errors.full_messages_for(field).each do |e| %>
+          <li><%= e %></li>
+        <% end %>
+      <% end %>
+    </div>
+  <% end %>
+  <%= textarea_f f, :key, :size => "col-md-8" , :rows=> '3' %>
```
ohadlevy: should we allow to upload ssh key as well?
also, i could see us using the ACE editor to display / edit the key.
for example, again from openshift:
image

timogoebel: @ohadlevy : I think, the upload idea is good. The ACE editor doesn't bring much value imho.
Anyways, both features aren't really necessary to get the core functionality in. @dLobatog already suggested to split the API part in a separate PR. I think, we should keep this out of this PR and create follow-up tickets.

app/controllers/concerns/foreman/controller/user_self_editing.rb
```diff
@@ -0,0 +1,9 @@
+module Foreman::Controller::UserSelfEditing
+  extend ActiveSupport::Concern
+
+  protected
+
+  def editing_self?
```
dLobatog, line 6: Seems like this isn't very useful, it forces you to change `editing_self?`, and completely duplicate the method every time you use it (changing the sliced params).

How about:

```
def editing_self?
  @editing_self ||= User.current.editing_self?(editing_self_params)
end

def editing_self_params
  params.slice(:controller, :action, :id)
end
```

Then on SshKeysCommon,

```
def editing_self_params
  params.slice(:controller, :action).merge { :id => params[:id] } 
end
```

At that point I think you can also safely revert the changes made here to the `editing_self?` method and leave it unchanged. Any further implementation just needs to provide its own params instead of overriding the entire action

app/controllers/concerns/foreman/controller/parameters/ssh_key.rb
```diff
@@ -0,0 +1,16 @@
+module Foreman::Controller::Parameters::SshKey
+  extend ActiveSupport::Concern
+
+  class_methods do
+    def ssh_key_params_filter
+      Foreman::ParameterFilter.new(::SshKey).tap do |filter|
+        filter.permit :key,
+          :name
```
dLobatog: nitpick, I guess same line is fine for 2 keys

app/helpers/application_helper.rb
```diff
@@ -127,11 +127,16 @@ def link_to_function_if_authorized(name, function, options = {}, html_options =
   end
 
   def display_delete_if_authorized(options = {}, html_options = {})
+    return "" unless authorized_for(options)
```
dLobatog: Not sure why `display_delete_if_authorized` doesn't work for you here? It seems a bit risky to modify this method to me..

app/models/usergroup.rb
```diff
@@ -112,4 +116,12 @@ def ensure_last_admin_group_is_not_deleted
   def other_admins
     User.unscoped.only_admin.except_hidden - all_users
   end
+
+  def ssh_keys
+    users.flat_map(&:ssh_keys)
```
dLobatog: `all_users` instead, otherwise it will not return the SSH keys for nested user groups

app/models/usergroup.rb
```diff
@@ -112,4 +116,12 @@ def ensure_last_admin_group_is_not_deleted
...
+  end
+
+  def ssh_authorized_keys
```
dLobatog: Duplicate code (the Jail and this method) between User and Usergroup, please extract to a concern

app/models/user.rb
```diff
@@ -351,7 +356,13 @@ def editing_self?(options = {})
       options[:id].to_i == self.id ||
     options[:controller].to_s =~ /\Aapi\/v\d+\/users\Z/ &&
       options[:action] =~ /show|update/ &&
-      (options[:id].to_i == self.id || options[:id] == self.login)
+      (options[:id].to_i == self.id || options[:id] == self.login) ||
```
dLobatog, line 363: I think these changes are avoidable (see my previous comment on the controller)

app/models/ssh_key.rb
dLobatog: Wouldn't it be better to display a general error if this happens as `errors.add(:base => "Could not generate fingerprint - #{e})`? 

As it stands right now if this exception happens, the user cannot know what happened, or how to fix it - not even the Administrator as it does not log anything.

app/models/ssh_key.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  def calculate_length
+    self.length = nil
+    return unless self.key.present?
+    self.length = SSHKey.ssh_public_key_bits(self.key)
+    true
+  rescue SSHKey::PublicKeyError
+    nil
```
dLobatog: Ditto as in https://github.com/theforeman/foreman/pull/4290/files#r102976270

app/validators/ssh_key_validator.rb
dLobatog: I see - please add a message though to the error logger when this happens - and ideally a message with the exception to display in the UI

app/views/ssh_keys/_form.html.erb
dLobatog: I do agree - no need to complicate it further

app/views/ssh_keys/_ssh_keys_tab.html.erb
```diff
@@ -0,0 +1,45 @@
...
+            <div class="blank-slate-pf-icon">
+              <%= icon_text("key", "", :kind => "fa") %>
+            </div>
+            <h1><%= _('SSH keys') %></h1>
+            <p>
+              <%= _("You can add SSH public keys to a user in Foreman.") %><br />
+              <%= _("The keys can then be used in templates and are available in config management tools managed by Foreman.") %><br />
```
dLobatog: nitpick, I think the 'then' here can be spared, and "config" -> "configuration", or "config.". Also "available for".

"The keys can be used in provisioning templates and are available for configuration management tools managed by Foreman."

cc @domcleal @GregSutcliffe if you can double check.. sorry

app/views/ssh_keys/_ssh_keys_tab.html.erb
```diff
@@ -0,0 +1,45 @@
+<div class="tab-pane" id="ssh_keys">
+  <table class="<%= table_css_classes 'table-fixed' %>">
+    <% if user.ssh_keys.any? %>
+    <thead>
+      <tr>
+        <th width="40%"><%= _('Name') %></th>
```
dLobatog: Please use `class="col-md-x col-sm-x"` instead of fixed percentages and pixels - it's way more flexible and looks good on smaller screens

db/migrate/20170214132230_create_ssh_keys.rb
```diff
@@ -0,0 +1,13 @@
+class CreateSshKeys < ActiveRecord::Migration
+  def change
+    create_table :ssh_keys do |t|
+      t.string :name, :limit => 255
+      t.text :key, :limit => 5000
```
dLobatog: Maximum is 16384, is there any reason to limit it here?

dLobatog (changes requested):
@timogoebel Overall pretty good - I was able to test it successfully, just found a few little problems in the implementation. Good job 😄 !

app/views/ssh_keys/_ssh_keys_tab.html.erb
johnpmitsch: @dLobatog If I'm understanding the sentence correctly, maybe add an "also" 

"The keys can be used in provisioning templates and are also available for configuration management tools managed by Foreman."

app/helpers/application_helper.rb
timogoebel: I need to display the link if either the user is authorized _or_ `@editing_self` is true.
I think, the change is quite harmless.
Before this commit, `display_delete_if_authorized` would just call `display_link_if_authorized`. The latter just calls `link_to` if `authorized_for` returns a truthy value or `""`.
Now this logic is also present in `display_delete_if_authorized`. Any better ideas are greatly appreciated, though?

app/models/ssh_key.rb
timogoebel: I'll add some logging. The messages provided by the gem aren't translated and are quite technical (e.g. "byte array too short"). That's why I don't want to show them to a user.

app/helpers/application_helper.rb
timogoebel: Ignore that, I found a better way to make this work. Please see `ApplicationHelper`'s `authorized_for`.

app/models/user.rb
timogoebel: @dLobatog : This method looks like a code smell to me. I don't see how we can get rid of the method, though. Any suggestions?

timogoebel:
@dLobatog : Thanks, I think I addressed all your comments. Let's refactor `User.editing_self?` in a separate PR to not overcomplicate things.

timogoebel:
- Fixed a bug, where the focus on the SSH keys form was set to the name but not to the textarea for the key.
- The key comment is now using the user's login and the foreman url setting.
- The ENC data now contains a user's hash as well that can e.g. used in a puppet module to create users on a VM based on the owner field.

Thanks @ekohl for testing this and the great ideas.

timogoebel:
Rebased, renamed the `users` ENC parameter to `foreman_users`

ekohl:
Untested, but https://github.com/ekohl/puppet-foreman_simple_user should create the users with their SSH keys.

ekohl:
So yesterday we tested the module, made some changes and it works. @timogoebel also has a snippet that can be used during provisioning that implements the same logic (it's based on https://gist.github.com/ekohl/ad6863798aed19c5fb45d3b91e8a4d87 but has a fix for when the owner is a group).

@ares you might have an opinion about the templating aspect.

@GregSutcliffe I believe this is a feature you really wanted.

app/models/usergroup.rb
```diff
@@ -112,4 +121,8 @@ def ensure_last_admin_group_is_not_deleted
   def other_admins
     User.unscoped.only_admin.except_hidden - all_users
   end
+
+  def ssh_keys
```
ares: should this method be protected if it's allowed by Jail?

ares:
Jail definitions seems to be fine except one nit. If you point me to the template snippet, I can review that. I think it would be beneficial if it's part of community-templates too.

timogoebel:
@ares: Added the template to this commit. @ekohl will create a PR in community-templates as he did most of the work and deserves the credit.

- Made the protected method public.
- Fixed a typo in a Jail definition.

ares:
> @ares: Added the template to this commit.

Sorry if I confused you, the template is normally synced from community-templates and should not be added directly in the PR, so please remove it.

timogoebel:
Removed the template.
Anything more to do or is this ready to 🚢 ?

ares:
My comments were addressed :+1: (I didn't do the full review)

dLobatog:
I'll spare some time today to review this - last time it was nearly ready but I was waiting for @ares 👍

dLobatog (approved):

dLobatog:
My comments were addressed nicely too, thanks @timogoebel @ares @ekohl @ohadlevy , let's get this on track for 1.15 !

PR #4291, opened 2017-02-16
Fixes #18530 - move redux devDependencies to dependencies

matanwerbner:
also, disable redux-logger in production.


dLobatog:
@matanwerbner Thanks - it seems to install the deps properly now. 

When this is merged, production dependencies for npm have to be packaged into rpm for distribution. We have an automated script that will get you 99% of the way there

Please install `npm2rpm` and `git-annex`, then submit the packaging PRs to foreman-packaging following this doc:

https://github.com/theforeman/foreman-packaging/tree/rpm/develop#howto-create-a-new-core-package-or-dependency-npm-modules

Summing up, please run `./add_npm_package.sh $NPMPACKAGENAME  $NPMPACKAGEVERSION bundle` for packages that have dependencies, using `single` for packages that don't have subdependecies and submit the result to foreman-packaging. (each package in its own separate commit please)

Without the RPMs for the npm packages your code will not run in nightly, so this is important to do every time new npm dependencies are added.

package.json
```diff
@@ -65,7 +58,13 @@
...
     "react-bootstrap": "^0.30.0",
     "react-dom": "^15.1.0",
-    "select2": "~3.5.2-browserify"
+    "select2": "~3.5.2-browserify",
+    "react-redux": "^5.0.2",
+    "redux": "^3.6.0",
+    "redux-logger": "^2.8.1",
+    "redux-mock-store": "^1.2.2",
+    "redux-thunk": "^2.2.0",
+    "seamless-immutable": "^7.0.1"
```
tbrisker: Could you please order these alphabetically? Some npm versions automatically do that, just to prevent the next one touching this file from changing it in an unrelated commit.

tbrisker (changes requested):

package.json
matanwerbner: fixed.

tbrisker (approved):

tbrisker:
Looks like @domcleal already took care of packaging for this (thanks Dominic!) so you can disregard @dLobatog 's comment about packaging. 
Thanks @matanwerbner !

matanwerbner:
cheers @tbrisker and @domcleal !

dLobatog:
Just remember next time :)

PR #4294, opened 2017-02-16
refs #18530 - move redux-mock-store to dev/test deps

domcleal:
mock-store doesn't appear to be a production dependency.


tbrisker:
Thanks @domcleal !

PR #4295, opened 2017-02-16
refs #18464 - rename *.spec.js to *.test.js

tbrisker:


tbrisker:
Honestly I couldn't care less if we name them `.spec.js` or `.test.js` as long as we're consistent. If you want to change the convention please do so in a separate PR.

dLobatog:
Thanks @tbrisker

PR #4296, opened 2017-02-16
fixes #18540 - update domain on primary in hosts_count tests

domcleal:
domain/domain_id are not attributes of the host itself, they're
delegated to the primary interface. When using update_attribute on the
host, the model doesn't appear `changed?` so Rails 5 skips the save.

The test "should update hosts_count" is left with delegation as the
direct interface modification is tested in the next test.


dLobatog:
Thanks @domcleal !

PR #4297, opened 2017-02-17
Fixes #18549 - Don't rebuild already correct DNS entries

agx:
If we do so and try to recreate them right away ecreate_dns_record might
still see a valid entry due to zone transfer slowness or other caching
issues. In this case we'd end up without a DNS entry after all. So
intead do nothing if the entry is already good.

This is a forward port from a fix I applied to a local 1.10.x instance.

Signed-off-by: Guido Günther <agx@sigxcpu.org>

### DIFF
--- a/app/models/concerns/orchestration/dns.rb
+++ b/app/models/concerns/orchestration/dns.rb
@@ -42,7 +42,7 @@ module Orchestration::DNS
     results = {}
 
     DnsInterface::RECORD_TYPES.each do |record_type|
-      del_dns_record_safe(record_type)
+      dns_record(record_type).nil? || dns_record(record_type).valid? || del_dns_record_safe(record_type)
 
       begin
         results[record_type] = dns_feasible?(record_type) ? recreate_dns_record(record_type) : true


app/models/concerns/orchestration/dns.rb
```diff
@@ -42,7 +42,7 @@ def rebuild_dns
     results = {}
 
     DnsInterface::RECORD_TYPES.each do |record_type|
-      del_dns_record_safe(record_type)
+      dns_record(record_type).nil? || dns_record(record_type).valid? || del_dns_record_safe(record_type)
```
dLobatog, line 45: The first check for `.nil?` is already done on `del_dns_record_safe` on `app/model/concerns/dns_interface.rb`.

Could you move the `.valid?` check to that method, and add a test on `test/models/orchestration/dns_test.rb` for rebuild_dns with an already correct DNS record?

dLobatog (changes requested):
@agx One minor comment inline, please check it out. Thanks for the contribution!

app/models/concerns/orchestration/dns.rb
agx: @dLobatog  wouldn't that obfuscate the meaning of del_dns_record_safe? It'd rather add a del_incorrect_dns_record

PR #4298, opened 2017-02-17
fixes #18551 - skip table_exists? error when the DB doesn't exist

domcleal:
MySQL and PostgreSQL DB adapters now throw a connection error under
Rails 5 if the the DB doesn't already exist, as the schema cache is no
longer used to answer the method call.

### DIFF
--- a/config/initializers/foreman.rb
+++ b/config/initializers/foreman.rb
@@ -34,4 +34,4 @@ Dashboard::Loader.load
 
 # clear our users topbar cache
 # The users table may not be exist during initial migration of the database
-TopbarSweeper.expire_cache_all_users if User.table_exists?
+TopbarSweeper.expire_cache_all_users if (User.table_exists? rescue false)


iNecas:
There are multiple places we have this code, some of them already fixed, some of them not https://github.com/theforeman/foreman/blob/develop/app/models/concerns/facets/managed_host_extensions.rb#L44,

Maybe the right time to encapsulate this into some helper method and use it from all the places?

domcleal:
The code you highlighted is inside a record accessor, so I'm not sure I see how that could fail during initialisation - it shouldn't need any modification.

iNecas:
I think it's better, if we can, not to expecting from the develper to know the nuances of different versions of Rails behavior, regardless of where exactly the code is. I see this as a code duplicity, and it would be better to have it on one place. Just my 2c: will not block this PR on this.

domcleal:
Code within models that already implicitly depend on a database shouldn't be using helpers that rescue database errors - so no, I don't think these are equivalent.

Developers and maintainers should be acutely aware of the risks and discouraged from adding database-dependent code into Foreman's initialisation code path.

dLobatog:
I see line 17 in this initializer and also fix_cache.rb contain similar checks - in any case, it's true it has little to do with these checks in the code (post initialization). Let's keep an eye on initializers and in case we see more of these db checks are needed (hopefully not) make some abstraction.

dLobatog:
Thanks @domcleal @iNecas !

PR #4299, opened 2017-02-17
Fixes #18554 - skip collection_cache_lookup on authorized_for

beav:
The call to `authorized_for` can take multiple seconds in some cases, leading
to performance issues.

This patch will bypass the cache lookup, resulting in similar times for admins
and non-admins when viewing pages with lots of hosts.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -97,7 +97,7 @@ module ApplicationHelper
       user.allowed_to?({ :controller => controller_name, :action => action, :id => id }) rescue false
     else
       authorizer = options.delete(:authorizer) || Authorizer.new(user)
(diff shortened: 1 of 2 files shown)


dLobatog:
Seems to have a significant error on REX jobs, merging..

Thanks @beav !

ares:
This patch does not make sense. The original problem was on one page that probably does not use authorizer object properly. After the cache is disabled, on every listing page (e.g. host index) we verify "view_hosts" permission on each table row for each subject separately. The cache normally loads all `@hosts` that are about to be displayed at once and do the verification in memory. I think this should be reverted and the fix should be in remote execution plugin where the authroized_for does not use properly setup authorizer. Thanks @iNecas for spotting this!

PR #4300, opened 2017-02-19
fixes #14178 - Introduce patternfly toast notifications

gailsteiger:
this PR replaces jnotify implementation.


webpack/assets/javascripts/foreman_toast_notifications.js
```diff
@@ -0,0 +1,47 @@
...
+      sticky: (type !== 'success'),
+      visible: true
+    });
+  });
+}
+
+// clear all notifications when leaving the page
+// questionable
+$(window).bind('beforeunload', function () {
+  ToastActions.closeNotifications();
```
ohadlevy: should move to component unmount?

gailsteiger: Question: Should toasts persist on navigation? I think so.

gailsteiger:
TODO:
- [x] replace Fade with Timer component
- [x] replace flux with redux
- [x] tests
- [ ] persist toasts after page navigation
- [ ] mark flash originated toasts so they will be deleted instead of hidden on dismiss

gailsteiger:
@ohadlevy 
In anticipation of your questions:
1. Do you think toasts should persist after navigation? I think they should,
2. Patternfly Toast Notifications overview indicates that it should include a message list that enables the user to view messages later. The design, however, provides no guidance on this feature. That's why I hide the toasts rather than delete them. Where would you like to go with this? Is this related to the whole issue  of which use cases are appropriate for notifications and which for toasts?

ohadlevy:
@rohoover  can you provide guidance ?

gailsteiger:
We could handle the message list in a separate PR.

ohadlevy (changes requested):
@gailsteiger thanks, I can see the code contains leftovers to the flux implementation.

@matanwerbner can you please review?

gailsteiger:
@ohadlevy unfluxed :)

webpack/assets/javascripts/foreman_toast_notifications.js
```diff
@@ -0,0 +1,39 @@
...
+      message,
+      sticky: (type !== 'success')
+    });
+  });
+}
+
+// load notifications from Rails
+$(document).on('ContentLoad', function () {
+  importFlashMessagesFromRails();
+  // should the flash toasts be persisted?
```
ohadlevy: please remove the comment, and the answer is no.

webpack/assets/javascripts/react_app/components/common/Alert.js
```diff
@@ -0,0 +1,35 @@
...
+
+  dismiss() {
+    this.setState({ visible: false });
+  }
+
+  render() {
+    const { type, dismissable, children } = this.props;
+
+    return (
+      this.state.visible ?
```
ohadlevy: nitpick: can write it as this.state.visible && (...) ?

webpack/assets/javascripts/react_app/components/common/CloseButton.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import Icon from './Icon';
+
+const CloseButton = ({onClick}) => {
+  return (
+    <button className="close" aria-hidden="true" onClick={onClick}>
+      <Icon type="close" />
```
ohadlevy, line 7: should icon have a hover / title ?

webpack/assets/javascripts/react_app/components/common/Fade.js
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import helpers from '../../common/helpers';
+
+export default class Fade extends React.Component {
+  constructor(props) {
```
ohadlevy: i believe you no longer use this component?

webpack/assets/javascripts/react_app/components/common/Timer.js
```diff
@@ -0,0 +1,20 @@
+class Timer {
```
ohadlevy: some docs on how to use would be nice :)

webpack/assets/javascripts/foreman_toast_notifications.js
```diff
@@ -0,0 +1,39 @@
+import * as ToastActions from './react_app/redux/actions/toasts';
```
matanwerbner, line 1: would this file be better off inside it's own dedicated folder?

also, is there a way that this can be a react component that's mounted on contentload and dispatches actions on mount?

webpack/assets/javascripts/react_app/components/notifications/toast/Toast.js
```diff
@@ -0,0 +1,85 @@
+import React from 'react';
+import Icon from '../../common/Icon';
+import helpers from '../../../common/helpers';
+import { ALERT_CSS } from '../../../constants';
+import CloseButton from '../../common/CloseButton';
+import * as ToastActions from '../../../redux/actions/toasts';
+// eslint-disable-next-line no-unused-vars
+import Timer from '../../common/Timer';
+
+const timerDelay = 8000;
```
ohadlevy: shouldnt this be in the default props?

webpack/assets/javascripts/react_app/components/notifications/toast/Toast.js
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  componentDidMount() {
+    if (!this.props.sticky) {
+      this.timer = new Timer(this.execute, timerDelay);
+      this.timer.startTimer();
+    }
+  }
+
+  componentWillUnmount() {
```
ohadlevy, line 24: does this also takes into account when I'm navigating to a new page? e.g. on the login page, if i submit incorrect password, i want the toast to be removed once i submitted a new pw (and before the server response)

webpack/assets/javascripts/react_app/components/common/Alert.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
```
matanwerbner: i feel like maybe this should be a functional component, without a state. parent should render it or not

webpack/assets/javascripts/react_app/components/notifications/toast/ToastsList.js
```diff
@@ -0,0 +1,33 @@
...
+class ToastsList extends Component {
+
+  render() {
+    const { messages } = this.props;
+
+    const visibleToasts = _.filter(messages, message => {
+      return message.visible;
+    });
+
+    const markup = _.map(visibleToasts, message => {
```
ohadlevy, line 17: nitpick:maybe its me, but i find those extra vars unneeded, i can see one loop over all visible messages that generate their markup inside the return block..

webpack/assets/javascripts/react_app/redux/actions/toasts/index.js
```diff
@@ -0,0 +1,32 @@
...
+  TOASTS_HIDE,
+  TOASTS_DELETE
+  // TOASTS_STORE_IN_LOCALSTORAGE,
+  // TOASTS_GET_FROM_LOCALSTORAGE
+} from '../../consts';
+import store from '../../../redux';
+
+// provide mechanism to persist toasts to localStorage
+export const addToast = (toast) => {
+  const payload = Object.assign({}, toast, {visible: true});
```
ohadlevy: visible:true is the default isnt it?

webpack/assets/javascripts/react_app/components/common/Timer.js
matanwerbner: i don't know if this class belongs here, maybe in a "utils" folder?

webpack/assets/javascripts/react_app/components/notifications/toast/Toast.js
```diff
@@ -0,0 +1,85 @@
...
+    if (this.timer) {
+      this.timer.clearTimer();
+      this.timer = null;
+    }
+  }
+
+  dismiss() {
+    ToastActions.hideToast(this.props.id);
```
matanwerbner: since toast is a dumb component, it can receive the callback to hideToast from the smart component parent

webpack/assets/javascripts/react_app/components/notifications/toast/Toast.js
```diff
@@ -0,0 +1,85 @@
...
+           onMouseEnter={this.onMouseEnter} onMouseLeave={this.onMouseLeave} >
+        {dismissable && <CloseButton onClick={this.dismiss}></CloseButton>}
+
+        <div className="pull-right toast-pf-action">
+          <a href="#">
+            {link}
+          </a>
+        </div>
+        <Icon type={type}/>
+        <strong>
```
matanwerbner: maybe `<strong>` is better replaced with `<div>` ?

webpack/assets/javascripts/react_app/components/notifications/toast/ToastsList.js
```diff
@@ -0,0 +1,33 @@
...
+      <div className="toast-notifications-list-pf">
+        {markup}
+      </div>
+    );
+  }
+}
+
+const mapStateToProps = state => ({
+  messages: state.toasts.messages,
+  counter: state.toasts.counter
```
matanwerbner: is counter used?

webpack/assets/javascripts/react_app/redux/actions/toasts/index.js
```diff
@@ -0,0 +1,32 @@
+import {
+  TOASTS_ADD,
+  TOASTS_HIDE,
+  TOASTS_DELETE
+  // TOASTS_STORE_IN_LOCALSTORAGE,
+  // TOASTS_GET_FROM_LOCALSTORAGE
+} from '../../consts';
+import store from '../../../redux';
```
matanwerbner: this is usually not required, action creators return plain objects for connected components to dispatch , or thunk functions (dispatch => {} )

webpack/assets/javascripts/react_app/redux/actions/toasts/index.js
```diff
@@ -0,0 +1,32 @@
...
+// provide mechanism to persist toasts to localStorage
+export const addToast = (toast) => {
+  const payload = Object.assign({}, toast, {visible: true});
+
+  // add to store
+  store.dispatch({ type: TOASTS_ADD, payload: payload });
+};
+
+export const hideToast = (id) => {
+  const payload = { id: id };
```
matanwerbner: `const payload = { id }` can be used when you have things like { id: id }

webpack/assets/javascripts/react_app/redux/reducers/toasts/index.js
```diff
@@ -0,0 +1,39 @@
...
+  counter: 0,
+  visibilityFilter: 'all'
+});
+
+export default (state = initialState, action) => {
+  const { payload } = action;
+
+  switch (action.type) {
+    case TOASTS_ADD: {
+      const id = (state.counter + 1).toString();
```
matanwerbner, line 19: not sure counter is required here... you can always use state.messages.length, no?

ohadlevy:
@gailsteiger also note the test failures.

Rohoover:
@ohadlevy @gailsteiger 

If I understand your first question correct. Yes, the toast timer is independent of a main navigation interaction. 

Toast notifications and the Notification Drawer, in my mind are not mutually exclusive. One could find cases where an event would trigger a toast notification and also an entry in the notification drawer because perhaps there is a follow up action to be taken or the content was deemed important enough to maintain.

I'm not sure I answered your question. Happy to have further discussion.

webpack/assets/javascripts/react_app/components/common/Alert.js
gailsteiger: No, that won't work. I need the expression to return null if `this.state.visible` is false. The expression you suggest would return false.
Alert is not used and has been removed.

webpack/assets/javascripts/react_app/components/common/CloseButton.js
gailsteiger: In this case no. It doesn't in the patternfly example. I think the X is pretty straight forward.

gailsteiger: Never mind, the docs say false will work too

webpack/assets/javascripts/react_app/components/common/Alert.js
gailsteiger: Never mind, the docs say false will work too

webpack/assets/javascripts/foreman_toast_notifications.js
gailsteiger: notify needs to be accessible from the legacy javascript so I don't think a react component is feasible

webpack/assets/javascripts/react_app/components/notifications/toast/Toast.js
gailsteiger: moved

webpack/assets/javascripts/react_app/components/common/Alert.js
gailsteiger: The component has been removed

webpack/assets/javascripts/react_app/components/notifications/toast/ToastsList.js
gailsteiger: I think this boils down to coder's preferences. I think it is cleaner to extract the mapping logic from the render return block

webpack/assets/javascripts/react_app/redux/reducers/toasts/index.js
gailsteiger: messages will be deleted so length won't work

matanwerbner: ok..maybe id is not needed? 
Im just trying to think of explicit code we can eliminate in favour of implicit computations

gailsteiger: There's nothing unique in the content that can be used. The only other alternative I can think of is datetime.

webpack/assets/javascripts/react_app/redux/actions/toasts/index.js
gailsteiger: That is my least favorite feature in ES6. I think it sacrifices clarity. I'll make the change though.

webpack/assets/javascripts/react_app/components/notifications/toast/ToastsList.js
gailsteiger: no.

webpack/assets/javascripts/react_app/redux/reducers/toasts/index.js
matanwerbner: Hmmm maybe the state should be an array... Seems like the natural choice, up to you.

webpack/assets/javascripts/react_app/redux/actions/toasts/index.js
gailsteiger: This seems more robust to me. I'd rather not make an assumption about the toast parameter but I'll remove it if you like. I'm sort of 55/45 on the issue.

webpack/assets/javascripts/react_app/components/notifications/toast/Toast.js
gailsteiger: It should be an inline element. You can see some examples in the patternfly reference markup

webpack/assets/javascripts/react_app/components/common/Timer.js
gailsteiger: It's a common widget. The location makes sense to me.

matanwerbner: Would it be possible to use native setTimeout instead of Timer?

webpack/assets/javascripts/react_app/components/notifications/toast/ToastsList.js
matanwerbner: agree, this looks fine to me

webpack/assets/javascripts/react_app/components/common/Fade.js
gailsteiger: gone

webpack/assets/javascripts/react_app/redux/reducers/toasts/index.js
gailsteiger: I think object is better since I need to access specific toasts.

webpack/assets/javascripts/react_app/redux/actions/toasts/index.js
gailsteiger: done

webpack/assets/javascripts/react_app/components/notifications/toast/Toast.js
gailsteiger: done

webpack/assets/javascripts/react_app/components/common/Timer.js
gailsteiger: Timer wraps the native setTimeout so yes it could be used. I think this is cleaner and reusable.

ohadlevy:
@gailsteiger what is the status of this pr? thanks

gailsteiger:
@ohadlevy I'm working on the test failures. Assuming I can fix that the PR should handle all toast types properly in the same manner they were handle by jnotify.

webpack/assets/javascripts/react_app/components/common/Timer.js
matanwerbner: Ok, i still think that it might belong more in reacrt_app/common, since it does no classify as a ui component in my mind.

webpack/assets/javascripts/react_app/components/notifications/toast/Toast.js
matanwerbner: `<span>` ? `<strong>` is hardly used anymore, and it would be more readable as a span with a dedicated css class in Toasts.scss

gailsteiger:
pushed code to see how changes to  `webpack/assets/javascripts/foreman_toast_notifications.js` help with test failures.

webpack/assets/javascripts/react_app/components/notifications/toast/Toast.js
gailsteiger: On further investigation I see that the `title` property is not yet used in the application.
I think we are getting ahead of ourselves by implementing a feature whose requirements aren't defined so I've removed the relevant markup.

ohadlevy:
@gailsteiger @waldenraines was looking into using this feature, we should make it accessible from Katello / pure JS  (not webpack) too.

ohadlevy:
@gailsteiger please note you must rebase the PR.

gailsteiger:
@ohadlevy I know that I must rebase. I'm focusing on the failing tests which is more challenging and I don't think that they are related to the indicated conflicting files.

waldenraines:
> @gailsteiger @waldenraines was looking into using this feature, we should make it accessible from Katello / pure JS (not webpack) too.

Please do.  I have an open PR on this (https://github.com/Katello/katello/pull/6575) that I need to rework to use the new toast notifications.

webpack/assets/javascripts/foreman_toast_notifications.js
```diff
@@ -0,0 +1,46 @@
...
+import store from '../javascripts/react_app/redux';
+import $ from 'jquery';
+
+export function notify(message, type, sticky) {
```
ohadlevy: I do wonder if we need to expose a better interface here for katello, or for "current" foreman ui to be able to provide more attributes (e.g. actions) into the toast.

waldenraines: Does this interface support html?  I know that we will have the need to include links in our toast notifications.

ohadlevy: @waldenraines @gailsteiger I would expect it to support the same interface we use for the notification drawer... ? ```(actions: {links: [{href:, title:...?}] } ```

gailsteiger: The Toast component accepts a link property which displays a link inline after the message.
I agree that the toast and notification drawer should support the same interface, including an action dropdown menu. Toast dropdown not yet implemented.

There is no support for html in the toast message. It is possible to implement via children (transclude for you Walden) but I would prefer not. For more specific formatting I prefer adding properties to the toast component and using a structured format instead of free form html.

gailsteiger: The Toast component is decoupled from redux. The ToastList ,the container component is not. This probably means a different container for implementations not using redux or requiring redux when using toasts.

waldenraines:
> The Toast component accepts a link property which displays a link inline after the message.
> I agree that the toast and notification drawer should support the same interface, including an action dropdown menu. Toast dropdown not yet implemented.
> 
> There is no support for html in the toast message. It is possible to implement via children (transclude for you Walden) but I would prefer not. For more specific formatting I prefer adding properties to the toast component and using a structured format instead of free form html.

That's fair, as allowing free form html puts us at risk of XSS.   We don't really need HTML other than the ability to add a link (usually to a task).  As long as the toast provides this functionality then we are good from our side.

ohadlevy:
@gailsteiger I've looked at the test failures..:

I've initially changed the test/integration/parameters_permissions_test.rb to run with javascript, the main point was that those tests had the ability to detect a successful toast notification as part of the editing process 
(https://github.com/theforeman/foreman/blob/develop/test/integration/parameters_permissions_test.rb#L115)

This change initiated a few failures(http://ci.theforeman.org/job/test_develop_pr_core/12519/database=postgresql,ruby=2.4,slave=fast/testReport/) -  base on my basic research to ACE editor, which exists on that page, which manipulate the form.

I assume its fine to remove the javascript tests in this form, if we can replace it with a specific integration test that checks the toast notifications?

gailsteiger:
@ohadlevy Thank you
Would the integration test be ruby or js?

ohadlevy:
@gailsteiger what is the status of the PR? thanks.

ohadlevy:
for reference, another implementation of pattenrfly toast notifications using react at: 
https://github.com/openstack/tripleo-ui/tree/master/src/js/components/notifications

tbrisker:
Closing in favor of https://github.com/theforeman/foreman/pull/4469.

tbrisker:
Thanks @gailsteiger !

PR #4302, opened 2017-02-19
fixes #18492 - move to vertical navigation

matanwerbner:
feb 22 2017 10-28 am


app/views/home/_navBar.html.erb
```diff
@@ -0,0 +1,15 @@
+<div id="vertical-navigation-container" class="nav-pf-vertical
```
ohadlevy, line 1: nitpick: filename should be _nar-bar.html.erb

ohadlevy:
@matanwerbner could you please:
1. add screenshots for this change?
2. fix the tests? some are broken now due to different in id/class names. 

thanks!

timogoebel:
Honestly speaking: I think, this looks horrible.

But if we really want to go down that road, some comments:

- The "Any Context" Link looks odd.

image

- The "Power" Column in the host view looks odd.
image

- One more glitch when the Taxonomy Menu is open:
image

- The Taxonomy menu can't fully expand:
image

- The spinner should be centered:
image

matanwerbner:
@timogoebel thanks for your comment
as for the org switcher - i do not have a functioning environment with this, when i do i will work on it (or perhaps you can assist)
as for the table - this happens when the screen width is small, and is not related to the PR.
screenshot from 2017-02-21 04-10-24
as for aligning the spinner - done, thanks for the heads up.
as for the way it looks - i actually prefer it to the previous design, but it's a matter of taste for sure.

ohadlevy:
comparing to other apps that uses pf vertical nav, openshift has a similar example of a context switcher (in this case a project) we can probably do something similar for the org/loc selector?

image

sean797:
It seems to be that a fair amount of people aren't on-board with this change, I would suggest they voice their concerns on https://groups.google.com/forum/#!msg/foreman-dev/lzLYcgD65TE/LHyDznBZBwAJ (just so its all in one place)

Personally I like it, I think it makes much better use small screens. A few comments:
- My scroll bar seems to have gone? I cant scroll on the Statistics or dashboard :'(
- Spinner still doesn't look centered to me I think its because the the bar along the top that holds the User options looks mostly transparent on my system.
screenshot from 2017-02-21 10-45-14
dashboard & statistics shows transparent top bar (maybe more), others do display correctly like My Account and Hosts pages.

Edit:
Also the sub menus colour needs to change - thanks @timogoebel stole your image 😆 
06731e38-f813-11e6-8913-815d6724db4d

matanwerbner:
@sean797 i am not seeing that locally, can you detail which kind of setup you're running?

sean797:
@matanwerbner I just done a `rake db:drop db:create db:migrate db:seed` and it doesn't happen on the dashboard anymore. Though the statistics pages is the same. (no scroll & transparent top bar) I'm running a devel install like https://theforeman.org/contribute.html#Setupdevelopmentenvironment

Is anyone else able to scroll on the Statistics page?

waldenraines:
Does it matter that we have this scrolling when the screen is small vertically?

screen shot 2017-02-22 at 9 18 53 am

app/assets/stylesheets/application.scss
```diff
@@ -19,6 +20,39 @@ html {
...
+                  &:hover {
+                    color: #fff;
+                  }
+                }
+              }
+        }
+      }
+      #main {
+        overflow: auto;
+        flex: 1;
```
waldenraines: This doesn't work on katello pages because we use `.container-fluid` for our container and because we don't have a `<div id="main">` so `flex: 1` is never set.  Can we move these styles to some element that automatically is inherited by plugins?

waldenraines: Another thing to note is that flexbox isn't fully supported on IE 10 and 11.  Not sure what this would look like in IE but it's something worth trying.

waldenraines (changes requested):

timogoebel:
I gave this another spin. Basically, good job @matanwerbner.

However, I still think this looks horrible. I reminds me of the days when there was Netscape Communicator 4.5 and everybody used framesets. But this might be a matter of taste. What really bugs me is that the vertical navigation costs precious space. On my 13" mac book pro, one fifth of the screen space is wasted by the nav menu. The horizontal menu was a lot more space saving and compact.
In addition I believe Foreman looses it's unique look. With the vertical navigation, it looks just like any other enterprise application.
If you compare the before and after, the "old" layout looks a lot nicer and brighter then the "new" layout.

All in all: :-1: from me. Sorry.

waldenraines:
> What really bugs me is that the vertical navigation costs precious space. On my 13" mac book pro, one fifth of the screen space is wasted by the nav menu. The horizontal menu was a lot more space saving and compact.

This pattern supports a collapsible menu with icons that would take up only about 20 - 30 px so that would solve this problem.  I don't think that has been implemented yet in this PR (at least not the last I checked).

Keep in mind also that the horizontal navigation starts to wrap to more than one line when your screen is narrow enough.

> In addition I believe Foreman looses it's unique look. With the vertical navigation, it looks just like any other enterprise application.

I think that's the point, to bring this application more in line style-wise with other offerings from a company named after a color and an article of clothing you wear on your head.  Also, how is foreman's look unique?  It looks like every other late 2000s web application (fixed width, horizontal nav with tree structure navigation, etc.).  Moreover, being unique isn't always good in UI, being user tested and thoroughly vetted is though.

timogoebel:
> This pattern supports a collapsible menu with icons that would take up only about 20 - 30 px so that would solve this problem.

I think, this is a ugly hack. With a horizontal navigation this problem just doesn't exist. And it saves you a click (expanding the menu).

> Also, how is foreman's look unique?

I mean that regarding the colors. The dark colors this PR adds feel very new (and kind of depressing). I prefer the light blueish and yellowish colors. Maybe the nav bar looks better on a white background.

> Moreover, being unique isn't always good in UI, being user tested and thoroughly vetted is though.

I fully agree. We can have that with a horizontal navigation as well.

waldenraines:
> I think, this is a ugly hack. With a horizontal navigation this problem just doesn't exist. And it saves you a click (expanding the menu).

I think it would be mouseover but point taken.

> I fully agree. We can have that with a horizontal navigation as well.

That's true.

Rohoover:
screen shot 2017-03-02 at 9 41 28 am

Horizontal navs do have their limitations (as shown above), that are not as easily addressed. Vertical nav is more flexible for growing menu items.

> I mean that regarding the colors. The dark colors this PR adds feel very new (and kind of depressing). I prefer the light blueish and yellowish colors. Maybe the nav bar looks better on a white background.

foreman_verticalnav

When using Foreman's colors the nav became overwhelming. 
screen shot 2017-03-02 at 10 51 57 am
screen shot 2017-03-02 at 10 52 36 am
screen shot 2017-03-02 at 10 52 54 am

It also failed a web accessibility test which the newer colors did not. While I know we aren't 100% compliant, making an effort to be as much as possible to be inclusive of users is ideal.

Rohoover:
@ohadlevy 
> comparing to other apps that uses pf vertical nav, openshift has a similar example of a context switcher (in this case a project) we can probably do something similar for the org/loc selector?

Yes - I would recommend this, but to add... Openshift is changing it a bit to be more inline with another initiative that directly affects them. I'll attach a wireframe and we can do something similar, but just a bit different. 

screen shot 2017-02-28 at 10 31 55 am

They have a secondary horizontal nav being added to house their project switcher.

I think we can utilize this for Orgs and Locations. However, I would recommend keeping the style of drop down that you show in your screenshot. I would also break out Location and Organization into two separate items so that way users can see upfront both what org they have selected and what location. Currently they can only see Organization unless they initiate the dropdown to view the location. 

The placement of this secondary horizontal nav also makes sense from a hierarchy standpoint, as it seems like to me (and I could be wrong). The context all of content and actions below are dependent on a users org/location.

matanwerbner:
i think we need to ask ourself: what is the best way to make the most out of the screen space we have available.
adding another horizontal nav bar whose sole purpose is to contain a small dropdown may not be the most efficient way to use our space.

also, how many users click on this dropdown? is it very commonly used? we should know these things in order to make an educated decision about where to place it - otherwise it might annoy most users who dont usually interact with it

Rohoover:
@matanwerbner 

I agree that if we had any metrics available they would be most appreciated.

timogoebel:
> Horizontal navs do have their limitations (as shown above), that are not as easily addressed. Vertical nav is more flexible for growing menu items.

@Rohoover : I'm not an UI expert, but I vaguely remember that in user tests horizontal navigation made a better impression than vertical ones. I think it was because users are used to read books. And books don't have a menu on the left side.
I agree, that using the blue foreman colors looks overwhelming. But I think, the dark background colors aren't much better. What do you think of a light background (maybe just plain white) for the navbar. That could also help that the content doesn't look so framed.

Rohoover:
@timogoebel Honestly there isn't a definitive guide on which is better. There are pros and cons to both and can be dependent on the circumstances of their use. In this case, because of the number of horizontal menu items, the fact that it is an application vs. marketing, and it's inability to scale well - Vertical Nav is a good fit.

ohadlevy:
@matanwerbner I've noticed all modal ui is broken, a few steps to reproduce:
1. goto host list, select one or more hosts, try any bulk action
2. goto host edit page, interface tab is broken. (and its modal)

matanwerbner:
@Rohoover in relation to the context switcher dropdown:
since it is the main thing missing currently, 
would it be possible to get a mock-up of it in context with the current look & feel of foreman?
we need to take into account the existing color scheme, and also the nested drop down functionality...
looking at openShift implementation, it does not answer these requirements, in my mind.

ohadlevy:
@matanwerbner @Rohoover I would also make sure it matchs / aligns with the other drop down on the top bar (e.g user selection, and if you have the access insight plugin - it has a drop down too).

Rohoover:
@matanwerbner Yes, I will work on that now. 

Also - I posted the links to the css provided to me on the Trello board.

ohadlevy:
@matanwerbner modal looks better, I still see another issue (see screenshot):
image

matanwerbner:
had a fix but didnt push, please update and look again it should be resolved

ohadlevy:
@Rohoover any update from your side? thanks!

Rohoover:
@ohadlevy @matanwerbner 

Here are a couple mock ups. Sorry for the delay I was on PTO yesterday. Let me know if you need any tweaks to these. It seems like to me the width can be adjusted on the drop downs as required, and there is still room for the spinner next to the Foreman Logo.

foreman_verticalnav_loc
foreman_verticalnav_org

timogoebel:
@Rohoover : Thanks. Just by looking at the mockups, I don't see where there is an improvement to the current design. I think, your suggestion just wastes even more space. Can we make a design that  either has a top bar or a left bar? Not both?
I'm all for improving the UI, but I honestly don't think we're doing users a favor with this.

I think, the new design looks like SUSE Manager. And that design shouldn't be our role model. It's also very overloaded.
Suse Manager

timogoebel:
@Rohoover : Why don't we go more into the direction that digitalocean has? They have a nice video on their front page with a clean, minimal design and excellent user experience imho. https://www.digitalocean.com/

Rohoover:
The SUSE manager example shows nav items on seemingly the same hierarchal level on both horizontal and vertical nav. They also have an overwhelming number of menu items in which secondary vertical items load below all the primary items-- I'm not sure from the screenshot. Though I understand your concern, I don't think the examples are similar.


Talking specifically about the Org and Location. In the current UI the Org is most visible with actions for location hidden in that nav. From my understanding they are of equal value in determining the content available to the user. In the most recent mockup provided, they are given equal nav footing and are more obvious. Their location in the mast-head reinforces their importance in determining the content below. Though saving space can be a good thing it shouldn't be the only contributor in determining navigation.

I'll check out the example you provided - thanks!

ohadlevy:
@Rohoover a few comments.
1. your first image is invalid, not sure if has anything different in it.
2. you've added the home menu item, is that on purpose?
3. I've noticed you also added a highlight on the current tab, that probably makes sense to add anyway - @matanwerbner.
4. you are OK that the drop downs for the user settings and the DD for org/loc are different style?

@timogoebel I've looked at the video in the landing page, and while it looks nice for host creation, i did not fully understand how it can handle many top level items  (I counted 7 or 6 if we remove home link in @Rohoover screenshot above)?

ohadlevy:
@timogoebel I assume this is the video?

timogoebel:
@ohadlevy : You are right. The video does not solve the nav problem. But they do have an admin panel I want to share as a inspiration.

image

My impression is, that moving to a vertical navigation has not been fully thought through. I can see that when using Foreman with all plugins enabled, the navigation can become overwhelming. I believe the design isn't the problematic part here but all the badly structured pages added by specific plugins. That is nothing we should try to solve in core or discuss here.
As said before: I'm all for improving the design and user experience. But a PR like this should only come as part of a whole concept and be one of the last steps in the process.
We discuss for ages whether it's reasonable for users to change their templates from `@host` to something else (they have to learn a new thing) but a design change like this has a way larger impact. Foreman feels like a completely different application - at least to me. One where the frame design doesn't match the content's.
I really like that ui experts define standards for the app, like how icons should looks for example. That helps a lot while developing because you don't have to reinvent the wheel. Using patternfly as a library is fine as well. Reusing existing components makes total sense. But just taking a vertical nav layout and trying to squeeze Foreman into it doesn't make sense.

Why do we want to change to a vertical nav?
What is the goal of this?
What are the alternatives?
Why do we need both a top and a sidebar? They take up significantly more space than the topbar did before.
Patternfly suggests a [grey background and only the "content boxes" have a white layout](http://www.patternfly.org/pattern-library/navigation/vertical-navigation/vertical-navigation-without-icons.html). Doing something similar might help connecting the "frame" to the "content".

matanwerbner:
@Rohoover can i get a psd file with at least the top bar as layers, so that i can see the exact sizes and dimensions you are aiming to?

ohadlevy:
@timogoebel 
> My impression is, that moving to a vertical navigation has not been fully thought through. I can see that when using Foreman with all plugins enabled, the navigation can become overwhelming. I believe the design isn't the problematic part here but all the badly structured pages added by specific plugins. That is nothing we should try to solve in core or discuss here.

If there are many items that do not fit in a single line, what else would you suggest to do?

> As said before: I'm all for improving the design and user experience. But a PR like this should only come as part of a whole concept and be one of the last steps in the process.

I believe @Rohoover tried to share over foreman-dev items in which she believe will make a positive impact, do you have any suggestion in terms of how to communicate so we can better align on UX changes?

> We discuss for ages whether it's reasonable for users to change their templates from @host to something else (they have to learn a new thing) but a design change like this has a way larger impact. Foreman feels like a completely different application - at least to me. One where the frame design doesn't match the content's.

Maybe this is where we tried to differentiate from patternfly, and not sure if it was the right change, we wanted to keep the original colors, and that created a few mismatches, for example, if it was all blue, the overall feedback was negative, if we used a different color for blue, it was not friendly for color blind.

> I really like that ui experts define standards for the app, like how icons should looks for example. That helps a lot while developing because you don't have to reinvent the wheel. Using patternfly as a library is fine as well. Reusing existing components makes total sense. But just taking a vertical nav layout and trying to squeeze Foreman into it doesn't make sense.
> 
> Why do we want to change to a vertical nav?
> What is the goal of this?
1. Allow more space for top level items
2. move to full width across the application.

> What are the alternatives?
BTH not sure, reduce the top level items overall?

> Why do we need both a top and a sidebar? They take up significantly more space than the topbar did before.
While true, I think its common for users to find basic operations at a top bar, having said that, we also gained about 30% screen size due to using full width.
also, we can (and should) minimize the nav bar on small display - I did ask @Rohoover to define on which size we should fold/minimize the navigation.

> Patternfly suggests a grey background and only the "content boxes" have a white layout. Doing something similar might help connecting the "frame" to the "content".

Already mentioned it before, should we drop the blue color then?

timogoebel:
@ohadlevy 

> > What is the goal of this?
>
> move to full width across the application.

On my MacPro I'm already at full width now. Can we do this without changing the nav?

> If there are many items that do not fit in a single line, what else would you suggest to do?

If there are more items than currently fit in the menu (with a regular screen size), there are simply too many items. On small screens I guess you want to hide the menu and just show it when a user clicks on it.

> I believe @Rohoover tried to share over foreman-dev items in which she believe will make a positive impact, do you have any suggestion in terms of how to communicate so we can better align on UX changes?

I do read and enjoy the notes from the weekly meetings. Sean has already addressed my concerns regarding them.

> should we drop the blue color then?

I personally would try to keep the blue color as it makes Foreman recognizable (on T-Shirts, stickers, ...). Maybe we can use the patternfly colors everywhere on the UI and just have a small blue bar on top of everything or pick up the color somewhere else?

Rohoover:
@ohadlevy 

Can the admin menu be styled to be similar? 

I wouldn't want to get rid of the blue completely... I agree with @timogoebel that is representative of the Foreman brand. If you're looking at investigating different implementation of the blue, I'd be happy to assist in that.

timogoebel:
I showed the PR in an internal meeting with some of our engineers who use Foreman on a daily basis to get their Feedback. Long story short: Five out of five did not like the change (me not included).

The points mentioned (unfiltered) were:
* This will confuse users who don't use Foreman on a regular basis even more. Foreman is complicated enough. They will have to relearn everything.
* "If they do that, there should be a switch to return to the old design."
* That looks like the new gitlab design. (Note: We're complaining about that all the time)
* The top menu felt like "tabs" and that was a good thing to have.
* The taxonomy switches look off and should be part of the left menu.
* It takes more space than before. The content looks squeezed together. 
* The colors / page looks too dark now.
* Why would you do that? It was good before.
* If the motivation is to make Foreman look better, the usability is worse now. And it doesn't even look better now.

When I explained, that the change was necessary to gain more space for menu items the reactions were that it would be a bad idea to add even more "main menu items" as it's complicated enough right now. The suggestion was to consider reducing the menu items as of now.
When you add Katello, in my experience users are even more overwhelmed by the menu items as with just Foreman core.

Another reaction was that a speech navigation "Foreman, deploy a new server" would be a better approach. ;-) 

Just wanted to relay the feedback. I hope it helps to improve this.

tbrisker:
I must say I don't care much for this proposed layout either. I tried playing around with it a bit and it seems to me that it doesn't much help the situation in smaller screen widths, as it causes some pages to look extremely cramped (hosts for example) because of the navigation taking up a non-proportional part of the view. Even with many plugins installed, the current navigation only starts to break at 1200px, which is less then most common monitors today. I understand that the menu can be collapsed (although not implemented yet), but that would increase the number of clicks needed to get from one page to another. I also agree with the feedback that the gray colors appear gloomy and a break with the existing graphic language of the application. I expect many users will not like this change as they will see no value in it, and it will force them to retrain their muscle memory of navigating around the app. This is a big change in UI for what may be solved in a simpler way - either by reducing the number of menus or the width at which the current menu collapses. 
This may just be my opinion, but I would like the rest of @theforeman/core team to weigh in on this please as this feels to me like a change that is too significant to be taken only with the feedback seen so far.

orrabin:
I don't really like the grey either.
I think it would look nice when it's collapsed and seeing the selected organization and location clearly is great but the claim of extra clicks might take precedence here.
Also, it does take over a lot more of the screen then the current menu does. I think that is expected on mobile but using a bigger screen it feels a little too wide for me.

sean797:
Some relevant user feedback about our current interface:
> Foreman has the least barrier to entry to start using it, from everything to the install and the user interface so my colleagues can look into using it and deploying different OSes. Gotta give credit to the interface guys. Once I used it a couple days, the layout of where to find stuff became 2nd nature.

Source: https://groups.google.com/forum/m/#!topic/foreman-users/YJ3_EnzTn8s

serenamarie125:
From a scalability and responsive perspective, vertical navigation seems to be the correct decision.  I agree @waldenraines that the collapsible menu with an icon is a great option - this allows users to collapse the menu as desired if they want more space.

I'm not quite sure that changing from horizontal to vertical navigation will confuse users.  Both of these navigation patterns are fairly standard and recognizable.

johnpmitsch:
While foreman menu may not be cluttered, the addition of plugins makes it very cluttered, and it breaks at small/medium browser size.
here
If Foreman is trying to support other plugins instead of being standalone app, shouldn't its menu support them as well?
A vertical nav will add more space for plugins to utilize. While there will definitely be minutia to resolve, I think this will be a positive change. A vertical nav shouldn't be anything new to a user. I would hate to see this idea passed over because its something new and we just go back to doing things the "old way", which seems broken to me.

timogoebel:
@johnpmitsch 

> While foreman menu may not be cluttered, the addition of plugins makes it very cluttered, and it breaks at small/medium browser size.

I agree, that we have an issue there and I'm all for supporting plugins.

> I would hate to see this idea passed over because its something new and we just go back to doing 
things the "old way"

I fully agree. I'm not saying that we can not improve this or change things. However we should not change things for the worse.

I think we have exchanged the opinions about this PR for now and got stuck there. To move forward, I think we need to see some alternatives to this approach. How about a horizontal nav that changes to a vertical navigation? How about a friendlier, less "framed" design?
In my opinion, we need a more general approach to this. How about doing a complete new design instead of just changing the way the navigation works? I'd love to see a proposal for a new design that works for everybody.

ohadlevy:
Going forward on this pr, I would suggest that we:
1. add the ability to completely hide the bar, on small displays it can be hidden by default and visible only on hover.
2.  introduce a small version of the bar (e.g. Icon only? or some other way? /cc @Rohoover )

any objections @timogoebel ?

timogoebel:
> any objections @timogoebel ?

Actually, yes. I'm sorry to say that, but I don't think, this solves any of my issues with this. It just makes them worse. My point is to not have a button like this image or a menu visible on hover.

Don't get me wrong, I'd really like to spend my time contributing stuff that brings the project forward. And I really like, that somebody is thinking about accessibility and design. This is important. But I have a very strong feeling, that this is not improving the project. My concerns are not about the code in this PR or even the people working on this. It's just the user experience.

Is there any other way than the proposed design? Let's think out of the box here for a second. Let's have at least one different proposal.
What about [this](https://maas.io/tour) design? It has a -vertical- horizontal navigation, lot's of space, bright colors and there is no unnecessary mobile-first stuff. And it's very minimalistic.

If we change the design, let's not just change the navigation, but the whole thing. But let's agree on something first, before we start implementing it and let's do it so there is a choice.

Rohoover:
@timogoebel 
I'm looking at your example but I'm not seeing the vertical navigation you're referring to. Is there a specific page you're looking at?

timogoebel:
@Rohoover : Apologies, I meant a horizontal navigation.

tbrisker:
I wonder if a better approach would be just to better handle the break points, and possibly programatically alter the menu text size and spacing on smaller monitors. 
Right now, with a large amount of plugins installed, the menu breaks for width <1200px or so (depending on the context). The "mobile" navigation only kicks in for resolution of <768px. 
What would be the minimum monitor width we want the full menu displayed on? 
if we push that mobile navigation to trigger for <1024px and some minor changes to the spacing and font sizes, we can mitigate the issue of width without going to vertical nav. Possibly we could even programatically trigger it if we can't fit everything in the width.

Rohoover:
Altering the text size and spacing on navigation isn't recommended. It generally hampers people with poor vision who will now manually blow up the screen size and break the nav all over again. Reducing the space between items increases chance of miss click as well as produces a visually cluttered appearance.

timogoebel:
@Rohoover , @serenamarie125: Thanks for explaining the motivation why you suggest a vertical menu and sorry I couldn't make it to the youtube broadcast. I would have liked to discuss this face-to-face. I do see, that a vertical nav might be the only way to support large menus. While I still believe a lot of menu items aren't a good think, I guess we have to support this. So, let's try to move forward with this and find a design with a vertical nav that works.

I believe, doing a user survey regarding the design would be a great idea.

I really enjoyed the part where you showed how other projects switched to the patternfly design.

What I liked was the top bar of the manageiq design because it is unique to manageiq and helps to resamble the brand. Can we do something similar for Foreman?
The navigation bar looks too similar to me across the projects. I think, we can keep the layout but should at least choose different colors. While a similar design makes sense if you have related products (like Word, Excel, PowerPoint) I don't see any value for the Foreman Project. Having a unique look makes marketing and recognition easier. While it's true the design should be intuitive to use and not exotic a litte "love" can't hurt, can it?
image


Regarding the menu:
Below you can see a screenshot of the menu from opsgenie.com. It expands on hover and shows the meaning of the icons. Is that what you suggest for Foreman?
I like icons in the menu, but I do have a hard time figuring out what the icons in the example mean and the expansion on hover doesn't work well on a mobile phone.
image

What do you think of a global search bar (for hosts) on every page as e.g. youtube and github have it? When using Foreman I see myself switching between pages and using the menu a lot. I don't do that on youtube e.g. (which has a hamburger menu and which is fine there, because a user usually uses the front page, related videos and the search results page). Foreman is all about hosts and most of the time I open Foreman to look up a host (because i need to know the name, reports, status or change something). Maybe we can add a shortcut for hosts and then the vertical navigation will not be so impractical to use. When I think about that, I don't use the "Monitor" menu at all but always access the info via a host. I rarely use the "Infrastructure" menu. Once the stuff in there is set up, you don't change them a lot. From the "Hosts" menu, I use "New Host" or "All Hosts" a lot. Like 90% of the time overall. And from the Configure Menu "Hostgroups", and "Environments". The rest: not so much.
Maybe we should write a small tool that examines Foreman's logfiles to find out what the most used pages are. We can then create shortcuts for these pages and "hide" the rest in a hamburger menu with a vertical navigation. The navigation would be completely hidden by default and there would still be a horizontal bar with a taxonomy switch, search box with typeahead to quickly access a particular host, a plus sign to create a new host, the Foreman logo and the 🍔 to open the vertical nav. My point is: Maybe we should differentiate between working with Foreman and configuring / setting up Foreman. If we find a design where users can work with Foreman fast and without a lot of clicks and and make it a little harder to set things up, that's actually a good change for users.

Have you thought about an alternative to the "two-pane" layout on pages? How does that work with the vertical nav?

Rohoover:
@timogoebel 

I'm going to use bullet points to respond so I can organize my thoughts better :)

1) I think we all agree some taxonomy work on the top could help improve not only the number of items but also their logical order. I don't think everything can be accounted for at once, but it's certainly worth the effort.

2) Yes we can play with the colors and branding of the top bar, lets just try to do it in a way that allows for good interactions for those with visual impairment.

3) The patternfly icon menu is user driven, meaning - they can switch between having only icons displayed or having icon + word... whatever their preference is. I believe in mobile it goes to icon + text, since there hover on mobile devices isn't a real capability yet (though I'm convinced it will be one day).

4) I have no opposition to a global search bar so long as the results are relevant and well organized. Any search that lacks those two qualities will be useless. 

5) To your point about Hosts... I've been toying with the idea that the main landing page for users perhaps is not the dashboard, maybe it's hosts or maybe it's a new landing page that has more relevant information, like top actions performed (which includes hosts), recent changes, etc. Aside from this point, it goes back around to the navigation taxonomy. I do not think hosts should be the 3 or 4th item in the menu, I believe (without any real data) that it is most possible the first menu item.

6) If we can see usage stats from users I will buy you a beer if we ever meet in person. Additionally, I'm not keen on having a main menu that changes based on user preferences, BUT a new landing page with top actions... I could totally see that. That would be a nice piece of predictive analytics :). A user logs in, sees a relevant landing page with the top actions that user regularly performs along with some super useful statuses to draw there attention if need be. Maybe some messaging about new features, bug fixes, help documents, etc. It starts to let the application interact with them vs. the user always interacting with it (if that makes sense).

7) Yes, I've done mock ups for two-pane view or rather it's removal. I think removal would lead to better utilization of the space, a bread-crumb would let the user know where they are and we've got a nifty interaction to still allow users to switch between items and search. I'm happy to talk more about that with you and show you what I've got.

I've been able to talk to some customers about this stuff and so far the feedback has been very positive.

timogoebel:
@Rohoover:

> 1. I think we all agree some taxonomy work on the top could help improve not only the number of items but also their logical order. I don't think everything can be accounted for at once, but it's certainly worth the effort.

Let's do it then. As written above, I'd suggest that the top bar gets more host centric. I imagine something like this (made with love and "paint"). Search bar for hosts. The plus-sign is a link for new hosts. The hamburger icon would show the actual nav.
image
Don't take this as a real proposal, just something to support my idea. The github top bar is also a good example of what I mean but doesn't have the hamburger menu icon.

> Yes we can play with the colors and branding of the top bar, lets just try to do it in a way that allows for good interactions for those with visual impairment.

Perfect. I fully agree. Let's make it beautiful as well. Do you have a suggestion for that?

> I've been toying with the idea that the main landing page for users perhaps is not the dashboard

Sounds like a good idea. Let's pursue this further. Same probably is for the hosts view page. Most space is used by two charts that don't have much use imho. If we can change the design to look good with the new layout, that will make users happy.

> I do not think hosts should be the 3 or 4th item in the menu, I believe (without any real data) that it is most possible the first menu item.

Maybe it can even be the only menu item in a top bar. All the other menu items can be hidden in the vertical nav.

> Additionally, I'm not keen on having a main menu that changes based on user preferences

I wasn't suggesting anything like that. I just wanted to handle hosts in a special way, outside of the nav menu.

> Yes, I've done mock ups for two-pane view or rather it's removal.

I think, this is an important piece of the whole strategy. I'd really like to see a "complete" mockup of such a page.

> If we can see usage stats from users I will buy you a beer if we ever meet in person.

I think @GregSutcliffe is your go-to person here. I believe, he wants to do some analytics plugin. It shouldn't be too much effort to write a tool to collect (anonymized) access logs from users and feed them into some tool that can draw nice graphs (Kibana, ...).


All in all, how are we going to proceed with this? I think we have exchanged ideas and concerns by now. To continue I think we need some mockups or wireframes so we can discuss the design proposal better. Even better, but not necessary would be a clickdummy. We could use them for a user survey, when we have agreed on a baseline. I believe the wireframes should show a complete page though and not just parts of the page.

serenamarie125:
@timogoebel glad to hear that the podcast helped with some of your concerns!

I guess my question is, do the colors need to be tweaked before this PR is approved?

I'm assuming that object details / 2 pane view changes can be handled outside of this PR, does that make sense to you?

timogoebel:
> I guess my question is, do the colors need to be tweaked before this PR is approved?
> 
> I'm assuming that object details / 2 pane view changes can be handled outside of this PR, does that make sense to you?

@serenamarie125: While we can address the 2 pane view outside of this PR, I do think we need a general plan before merging this. And yes, I think this is not ready to merge right now. The colors and the way both the top bar and navigation bar work need further tweaking.
I'll be happy to "beta test" this with experienced Foreman users and share the results. We have quite a few Foreman users here.

Rohoover:
I've done some work around styling of the navigation. This includes color options (which I've tested to pass accessibility requirements) as well as updating the style to which interest was expressed earlier. Lastly, I did do a general reorder of the items in the main menu based on feedback I've received from users as to more common items.

I think there are future items that need to be looked at but probably not for this PR:

- Additional ordering of the menu
---- Future redesign of monitoring to be more useful or introduction of some kind of alternate landing page
---- Merge hosts would reduce menu items within Hosts. There may be other opportunities as well to merge or consolidate UI's which in turn reduce menu items
- Introduction of a global search
[Foreman - Vertical Nav.pdf](https://github.com/theforeman/foreman/files/1153063/Foreman.-.Vertical.Nav.pdf)

timogoebel:
@Rohoover : Thanks. I like your proposal. It is definitely a Foreman design now. 👍 A co-worker actually pointed me to the fact, that this looks very similar to the [Microsoft Azure UI](https://msdnshared.blob.core.windows.net/media/2016/12/extension-ga.png). However they use colors for different services which he finds very confusing. I think, a lot of Foreman users will be familiar with the Azure UI. So we might be on the right track here.

I have two questions for your design:
What happens on click of the hamburger icon? Does it "shrink" the menu to only show icons?
What is the default menu state (just icons vs. icons + text)?
How does the second level of the menu look?
Is the hamburger a necessary requirement or could we think of something without?

I presented your proposal at our team meeting and feedback was a lot better then last time. So we're moving in the right direction :-)
The global search (definitely another PR) got really, really good feedback. I absolutely want to have that feature. So does everybody I talked to about this.

I'm curious how the content / content layout, e.g. the host list page, blends with the new navigation. Could you extend the proposals to show some pages with content?

GregSutcliffe:
Hey all, sorry from the radio silence on my end. Two things:

* @timogoebel glad the video helped - I'm sure another hangout (unrecorded this time) with @Rohoover is possible, if you both want to (of course, we'll want to report back here on the PR after).
* I'm setting up a public demo box on Scaleway so we can start showing this to the wider community. I doubt the current PR will work with either `develop` or `1.15-stable`, is there someone who can fix it up? @tbrisker?
* I also promised a quick survey, but I'll get the demo box up first.

Glad to see some resolution happening now :)

ohadlevy:
replaced by #4794 thanks @matanwerbner

PR #4303, opened 2017-02-19
Fixes #18560 - Remove searchbar from 403, change 403 style

shlomizadok:


app/views/common/403.html.erb
```diff
@@ -1,11 +1,15 @@
...
 
-<div class='col-md-offset-4 col-md-4'>
-  <%= alert :header => _("Permission denied"),
-    :text => (_("You are not authorized to perform this action.") + '</br>'.html_safe +
-                 _("Please request one of the required permissions listed below from a Foreman administrator:") +
-                 content_tag(:ul, permissions.join.html_safe)).html_safe,
-    :actions => link_to(_('Back'), main_app.root_path, :class => 'btn btn-default'),
-    :class => 'alert-danger',
-    :close => false %>
+<% content_for(:title, _("Permission denied")) %>
```
ohadlevy, line 3: you can use 
```erb
<% title(..) %>
```
instead which will also set the page title and header.

ohadlevy:
@shlomizadok can you provide a screenshot ? thanks!

shlomizadok:
@rohoover  - May I have your take on the suggested styling here (I couldn't find any ref on @patternfly, so I have created my own...)

Currently the 403 page looks like:

image

In this PR, I have removed the search bar and changed the appearance to:

image

Would love to hear your comments :smile:

app/views/common/403.html.erb
shlomizadok: Cowardly copied from `welcome` pages... :wink: 
Will change

dLobatog: I can change this on merge, it's fine

dLobatog:
Thanks @shlomizadok  - I'll test changing the `content_for` to `title` on merge and will merge it if it works fine. I am not 100% convinced the empty state pattern should be used here, but after looking in various places (ManageIQ, patternfly-react-starter), I didn't see anyone using anything in particular for pages that deny actions due to permissions)

dLobatog:
`content_for` is fine, title adds also the title at the beginning which we don't want:

screenshot from 2017-02-20 16-56-25

I'll merge for the moment as we will not get feedback from @rohoover until next week, this is an improvement already IMO, if more things have to be changed, that can be done later. 

@rohoover the merged version does not differ from the screenshot @shlomizadok posted

Rohoover:
@shlomizadok Nice!

Definitely no title @dLobatog 

bitmoji

PR #4304, opened 2017-02-20
refs #7495 - correct syntax in API error handler

domcleal:


dLobatog:
Thanks @domcleal , I thought this was possible to do even on 1.x rubies

PR #4305, opened 2017-02-20
fixes #18565 - move constantize inside access permission tests

domcleal:
Matches normal test loading so constants (and source files) are only
loaded when the test is executed. Syntax errors will then cause a
failure for tests loading the file instead of outside the test process.


dLobatog:
[test]
test foreman

domcleal:
Patch rebased. http://ci.theforeman.org/job/test_develop_pr_core/12147/ shows the original failure with a syntax error present.

dLobatog:
Thanks @domcleal !

PR #4307, opened 2017-02-20
fixes #18568 - replace deprecated AC::Parameters#update

domcleal:
The #update method in Rails 5 is returning a HWIA rather than the
ActionController::Parameters instance, causing keep_param to return the
wrong object type. It is also deprecated in 5.0, so replace it with a
simpler and supported assignment through #[]=.


dLobatog:
Thanks @domcleal !

PR #4309, opened 2017-02-20
Refs #18576 - update OpenSUSE mirror url

mmoll:


timogoebel (approved):
@mmoll: Looks good to me. I'll merge if there are no objections.

PR #4310, opened 2017-02-21
Fixes #18581 - conditionalize SUSE mediapath

mmoll:


ares (changes requested):
Looks good, would you mind adding a test to https://github.com/theforeman/foreman/blob/develop/test/unit/foreman/renderer_test.rb that verifies this method renders properly for host without installation medium? The easiest way would be using `send` for calling private method or you can setup the host and call `load_template_vars`

ares:
Also isn't there the same problem for all methods below? (coreos_attributes, aif_attributes, ...) Maybe it would make more sense to fix it at https://github.com/theforeman/foreman/blob/develop/app/models/operatingsystem.rb#L126 but this method is exposed in templates

mmoll:
@ares like this?
```diff
diff --git a/app/models/operatingsystem.rb b/app/models/operatingsystem.rb
index 948ec2e89..65a872d26 100644
--- a/app/models/operatingsystem.rb
+++ b/app/models/operatingsystem.rb
@@ -124,6 +124,7 @@ class Operatingsystem < ActiveRecord::Base
   end
 
   def medium_uri(host, url = nil)
+    return '' unless host.medium
     url ||= host.medium.path
     medium_vars_to_uri(url, host.architecture.name, host.operatingsystem)
   end
```

ares:
it should probably take url variable into account as well, host.medium is only used if passed url argument is nil

e.g.

```
def medium_uri(host, url = nil)
  if url.nil?
    return '' unless host.medium
    url = host.medium.path
  end
  medium_vars_to_uri(url, host.architecture.name, host.operatingsystem)
end
```

but I'm not sure whether this is the best approach. As a dev I want to get an exception when I try to access medium_uri without specifying url and host doesn't have a medium. The exception should be clearer, e.g. ArgumentError clearly stating that host does not have medium and no url was specified. I wish we had a macro or decorator for OS already...

So maybe it's best to go with your original approach, just covering it by tests and ideally fixing other similar methods below `yast_attributes`.

mmoll:
I guess this PR is good to go then, as the error only happens with image based provisioning and the other OSes don't support that anyway(?).

ares (approved):

ares:
Ok, fair enough. Thanks @mmoll, merged as 42e9040adcd44a489cd55353cbc4343fb8d1bdc3

PR #4313, opened 2017-02-21
Fixes #18587 and #18588 - fix 'multipart form data' IE bug for domains, subnets and lookup_keys

evgeni:
According to this bug (https://connect.microsoft.com/IE/Feedback/Details/868498),
multipart form data is malformed if there are fields without names.
It been solved within Edge browser, however no solution is expected for IE10/IE11

It cause a failure with Rack - EOFError: bad content body, therefore pages like
domains and subnet create/edit cannot be submitted (and all other forms which are
submitted via ajax with multipart form data, and an unnamed field within).

This is identical to https://github.com/theforeman/foreman/pull/4125
but on the domains and subnets pages.

### DIFF
--- a/app/helpers/form_helper.rb
+++ b/app/helpers/form_helper.rb
@@ -254,7 +254,7 @@ module FormHelper
         options.merge! :'data-id' => form_to_submit_id(f) unless options.has_key?(:'data-id')
         link_to(_("Cancel"), args[:cancel_path], :class => "btn btn-default") + " " + f.submit(text, options)
       end
-    end
+    end + ie_multipart_fix
   end
 
   def add_help_to_label(size_class, label, help_inline)
@@ -373,6 +373,14 @@ module FormHelper
           end
         end.html_safe
       end
+    end
+  end
+
+  def ie_multipart_fix
+    # This hidden input is a workaround to fix IE Multipart form data bug (https://connect.microsoft.com/IE/Feedback/Details/868498)
+    # Only add this fix for two-pane layouts, as these are the only ones that trigger the bug
+    if request.headers["X-Foreman-Layout"] == 'two-pane'
+      content_tag(:input, '', {:type => "hidden", :name => "_ie_support"})
     end
   end
 end
--- a/app/views/templates/_form.html.erb
+++ b/app/views/templates/_form.html.erb
@@ -153,13 +153,11 @@
     <%= render "custom_tabs", :f => f unless type == 'ptable'%>
     <%= render 'taxonomies/loc_org_tabs', :f => f, :obj => @template %>
 
+    <%# These hidden inputs are used for the diff Preview. They intentionally do
+        not have a 'name' attribute so browsers will not send them in the form POST. %>
+    <input type="hidden" id="old" value="<%= @template.template %>" />
+    <input type="hidden" id="new" value="<%= @template.template %>" />
+
     <%= submit_or_cancel f, false, :cancel_path => template_path_for(@template.class) %>
   </div>
-  <%# These hidden inputs are used for the diff Preview. They intentionally do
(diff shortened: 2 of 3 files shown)


sean797:
ok to test

tbrisker:
@evgeni - could you please extract that snippet to a helper and use that in both places as well as the previous one (in the templates)? I've also seen this occur on the smart variable form, care to add it there as well?

evgeni:
Created a snippet as @tbrisker asked and also fixed https://bugzilla.redhat.com/show_bug.cgi?id=1422188

app/helpers/form_helper.rb
```diff
@@ -375,4 +375,9 @@ def field(f, attr, options = {})
       end
     end
   end
+
+  def internetexplorer_multipart_form_data_fix(f)
```
tbrisker: No need to pass the form object (`f`) in here. Also name is a bit long, how about `ie_multipart_fix` or something like that?

tbrisker (changes requested):
Just a minor comment, looks good in general. 
I'll try to test if there are other forms that have the same issue so we can solve it once and for all.

tbrisker (approved):

tbrisker:
Code looks good, pending testing on IE and checking for other places hitting the same bug.

amirfefer:
@evgeni, @tbrisker domains and subnets pages work well on IE11, however I've found few more forms with the same issue: Global parameters, Architectures, Installation media and Operating systems.

evgeni:
@amirfefer cool, thanks for checking. I must admit I was too lazy to click though *all* the dialogs in my IE11 VM.

Will see that I fix the other dialogs too.

I wonder if there is a nicer way to inject that snipet into all Ajax dialogs.

dLobatog:
Setting to WoC based on @amirfefer 's  comment

tbrisker:
@evgeni - There is a way to automatically add this on all two pane forms. In the `submit_or_cancel` method in `app/helper/form_helper.rb`, check if `request.env['HTTP_X_FOREMAN_LAYOUT'] == 'two-pane'`. If it is true, add this fix after the buttons. 
We can even further only apply it on IE browsers by checking if `request.env['HTTP_USER_AGENT'].include?('MSIE')`.

evgeni:
@tbrisker there we go :)

* I used `request.headers` not `.env` as I think that's cleaner
* I did not add the `MSIE` filter, as I think this is ugly and only makes debugging harder

app/views/templates/_form.html.erb
```diff
@@ -159,7 +159,4 @@
       not have a 'name' attribute so browsers will not send them in the form POST. %>
   <input type="hidden" id="old" value="<%= @template.template %>" />
   <input type="hidden" id="new" value="<%= @template.template %>" />
```
tbrisker: I think these also need to be reordered with the submit_or_cancel so the fix is the last input

tbrisker (changes requested):
Thanks @evgeni , this looks much cleaner. one concern though - the fake field has to be the last field in the form, we need to make sure that is the case for all relevant forms. One case noted inline.

evgeni:
Right. That is what you get when you commit at 8000m :)
Fixed.

timogoebel:
I just tested `develop` without this patch with:
image

The menu didn't work at all but I had no issues submitting the subnet form. :-(

timogoebel:
@evgeni : I was finally able to verify this works. Good job.

@tbrisker +1 from me

timogoebel (approved):

tbrisker (approved):

tbrisker:
Thanks @evgeni  and @timogoebel !

PR #4316, opened 2017-02-22
Fixes #18618 - Use Dynflow as an ActiveJob backend

dLobatog:
This commit initializes the Dynflow world in a very similar manner to
foreman-tasks (if a Dynflow world is available from foreman-tasks, we
use that one). 

This is ready now that https://github.com/Dynflow/dynflow/pull/218 is merged and released.

To test, create a job on `app/jobs/`, for example (app/jobs/sample_job.rb)

```ruby
class SampleJob < ActiveJob::Base
  def perform(msg)
    5.times { logger.info("#{msg}") }
  end
end
```

If you use foreman-tasks, and apply https://github.com/theforeman/foreman-tasks/pull/235 , you can also define the usual `def humanize_name` etc.. methods so that the jobs are shown in the list of tasks. 

Then on the console run any of the methods from ActiveJob, such as `SampleJob.perform_now`, `SampleJob.perform_later`, `SampleJob.set(:wait => 10.seconds).perform_later`, etcetera. The output should be similar to https://gist.github.com/anonymous/429e7171e867c94182839a4b0d1ddddc 

cc @iNecas as you're familiar with this initialization (mostly drawn from foreman-tasks)

If you want to run it as in production, run `rake dynflow:executor` on a different terminal, and run the job I posted above.

screenshot from 2017-02-22 10-29-43

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,6 +50,9 @@ gem 'webpack-rails', '~> 0.9.8'
 gem 'mail', '~> 2.6'
 gem 'sshkey', '~> 1.9'
 gem 'ruby2ruby', '2.3.2'
+gem 'dynflow', '>= 0.8.25', '< 1.0.0'
+gem 'daemons'
+gem 'get_process_mem'
 
 Dir["#{File.dirname(FOREMAN_GEMFILE)}/bundler.d/*.rb"].each do |bundle|
   self.instance_eval(Bundler.read_file(bundle))
--- a/config/application.rb
+++ b/config/application.rb
@@ -6,6 +6,7 @@ require 'apipie/middleware/checksum_in_headers'
 require 'rails/all'
 
 require File.expand_path('../../config/settings', __FILE__)
+require File.expand_path('../../lib/foreman/dynflow', __FILE__)
 
 if File.exist?(File.expand_path('../../Gemfile.in', __FILE__))
   # If there is a Gemfile.in file, we will not use Bundler but BundlerExt
@@ -183,6 +184,9 @@ module Foreman
     # New config option to opt out of params "deep munging" that was used to address security vulnerability CVE-2013-0155.
     config.action_dispatch.perform_deep_munge = false
 
+    # Use Dynflow as the backend for ActiveJob
+    config.active_job.queue_adapter = :dynflow
+
     Foreman::Logging.configure(
       :log_directory => "#{Rails.root}/log",
       :environment => Rails.env,
@@ -197,7 +201,9 @@ module Foreman
       :permissions => {:enabled => false},
       :sql => {:enabled => false},
       :templates => {:enabled => true},
-      :notifications => {:enabled => true}
+      :notifications => {:enabled => true},
+      :background => {:enabled => true},
+      :dynflow => {:enabled => true}
     ))
 
     config.logger = Foreman::Logging.logger('app')
@@ -230,6 +236,19 @@ module Foreman
 
     # Use the database for sessions instead of the cookie-based default
     config.session_store :active_record_store, :secure => !!SETTINGS[:require_ssl]
+
+    def dynflow
+      return @dynflow if @dynflow.present?
+      @dynflow =
+        if defined?(ForemanTasks)
+          ForemanTasks.dynflow
+        else
+          Dynflow::Rails.new(Foreman::Dynflow::Configuration.new)
+        end
+      @dynflow.require!
+      @dynflow.initialize!
+      @dynflow
+    end
   end
 
   def self.setup_console
--- a/config/settings.yaml.example
+++ b/config/settings.yaml.example
@@ -59,4 +59,7 @@
 #  :templates:
 #    :enabled: true
 #  :notifications:
+#  :background:
+#    :enabled: true
+#  :dynflow:
 #    :enabled: true
--- a/lib/foreman/dynflow.rb
+++ b/lib/foreman/dynflow.rb
@@ -1,0 +1,7 @@
+require 'dynflow'
+
+module Foreman
+  module Dynflow
+    require_relative 'dynflow/configuration'
+  end
+end
--- a/lib/foreman/dynflow/configuration.rb
+++ b/lib/foreman/dynflow/configuration.rb
@@ -1,0 +1,15 @@
+module Foreman
+  module Dynflow
+    class Configuration < ::Dynflow::Rails::Configuration
+      # Action related info such as exceptions raised inside the actions' methods
+      def action_logger
+        Foreman::Logging.logger('background')
+      end
+
+      # Dynflow related info about the progress of the execution
+      def dynflow_logger
+        Foreman::Logging.logger('dynflow')
+      end
+    end
+  end
+end
--- a/lib/tasks/dynflow.rake
+++ b/lib/tasks/dynflow.rake
@@ -1,0 +1,13 @@
+namespace :dynflow do
+  desc <<-END_DESC
+In development mode, the Dynflow executor is part of the web server process. However, in production, it's more suitable to have the web server process separated from the async executor. Therefore, Dynflow is set to use external process in production mode by default (can be changed with Foreman.dynflow.config.remote = false). This task can be used to start the executor manually in development mode.
+
+The executor process needs to be executed before the web server. You can run it by:
+
+  foreman-rake foreman_tasks:dynflow:executor
+
+END_DESC
+  task :executor => :environment do
+    Dynflow::Rails::Daemon.new.run
+  end
+end
--- a/bin/dynflow-executor
+++ b/bin/dynflow-executor
@@ -1,0 +1,72 @@
+#!/usr/bin/env ruby
+
+require 'optparse'
+
+class ArgvParser
+  attr_reader :options, :command
+
+  def initialize(argv, file)
+    @options = { rails_root: Dir.pwd }
+
+    opts = OptionParser.new do |opts|
+      opts.banner = banner(file)
+
+      opts.on('-h', '--help', 'Show this message') do
+        puts opts
+        exit 1
+      end
+      opts.on('-f', '--foreman-root=PATH', "Path to Foreman Rails root path. By default '#{@options[:rails_root]}'") do |path|
+        @options[:rails_root] = path
+      end
+      opts.on('-c', '--executors-count=COUNT', 'Number of parallel executors to spawn. Overrides EXECUTORS_COUNT environment varaible.') do |count|
+        @options[:executors_count] = count.to_i
+      end
+      opts.on('-m', '--memory-limit=SIZE', 'Limits the amount of memory an executor can consume. Overrides EXECUTOR_MEMORY_LIMIT environment varaible. You can use kb, mb, gb') do |size|
+        @options[:memory_limit] = size
+      end
+      opts.on('--executor-memory-init-delay=SECONDS', 'Start memory polling after SECONDS. Overrides EXECUTOR_MEMORY_MONITOR_DELAY environment varaible.') do |seconds|
+        @options[:memory_init_delay] = seconds.to_i
+      end
+      opts.on('--executor-memory-polling-interval=SECONDS', 'Check for memory useage every SECONDS sec. Overrides EXECUTOR_MEMORY_MONITOR_INTERVAL environment varaible.') do |seconds|
+        @options[:memory_polling_interval] = seconds.to_i
+      end
+    end
+
+    args = opts.parse!(argv)
+    @command = args.first || 'run'
+  end
+
+  def banner(file)
+    banner = <<BANNER
+Run Dynflow executor for Foreman tasks.
+
+Usage: #{File.basename(file)} [options] ACTION"
+
+ACTION can be one of:
+
+* start   - start the executor on background. It creates these files
+          in tmp/pid directory:
+
+            * dynflow_executor_monitor.pid - pid of monitor ensuring
+                                             the executor keeps running
+            * dynflow_executor.pid         - pid of the executor itself
+            * dynflow_executor.output      - stdout of the executor
+* stop    - stops the running executor
+* restart - restarts the running executor
+* run     - run the executor in foreground
+
+BANNER
+    banner
+  end
+end
+
+# run the script if it's executed explicitly
+if $PROGRAM_NAME == __FILE__
+  parser = ArgvParser.new(ARGV, $PROGRAM_NAME)
+
+  app_file = File.expand_path('./config/application', parser.options[:rails_root])
+  require app_file
+
+  Dynflow::Rails::Daemon.new.run_background(parser.command, parser.options)
+end
+


Gemfile
```diff
@@ -38,6 +38,7 @@ gem 'deacon', '~> 1.0'
 gem 'webpack-rails', '~> 0.9.8'
 gem 'mail', '~> 2.6'
 gem 'nokogiri', '< 1.7' if RUBY_VERSION.start_with? '2.0.'
+gem 'dynflow', '~> 0.8', '>= 0.8.20'
```
ohadlevy: i assume we need a newer version of dynflow with AJ support?

config/application.rb
```diff
@@ -216,8 +220,22 @@ class Application < Rails::Application
...
-    puts "For some operations a user must be set, try User.current = User.first"
+  class << self
+    def setup_console
+      ENV['IRBRC'] = File.expand_path('../irbrc', __FILE__)
+      puts "For some operations a user must be set, try User.current = User.first"
+    end
+
+    def dynflow
+      return @dynflow if @dynflow.present?
+      if defined?(ForemanTasks)
```
ohadlevy: does foreman tasks load prior to this or do we need to update foreman tasks initialization to run first?

config/application.rb
```diff
@@ -216,8 +220,22 @@ class Application < Rails::Application
...
+        @dynflow ||= ForemanTasks.dynflow
```
ohadlevy: no need for ||= here.

config/application.rb
```diff
@@ -216,8 +220,22 @@ class Application < Rails::Application
...
+      else
+        @dynflow ||= Foreman::Dynflow.new
```
ohadlevy: ditto here, it will never be pre initialized because of the return statemet

config/application.rb
```diff
@@ -216,8 +220,22 @@ class Application < Rails::Application
...
-    ENV['IRBRC'] = File.expand_path('../irbrc', __FILE__)
-    puts "For some operations a user must be set, try User.current = User.first"
+  class << self
+    def setup_console
+      ENV['IRBRC'] = File.expand_path('../irbrc', __FILE__)
+      puts "For some operations a user must be set, try User.current = User.first"
+    end
+
+    def dynflow
+      return @dynflow if @dynflow.present?
```
ohadlevy: when will that be ever set? and if it is already set, must we require and init it? i'm worried on a secnario when its defined, but did not call require or initialize.

config/application.rb
```diff
@@ -216,8 +220,22 @@ class Application < Rails::Application
...
+      if defined?(ForemanTasks)
+        @dynflow ||= ForemanTasks.dynflow
+      else
+        @dynflow ||= Foreman::Dynflow.new
+        @dynflow.require!
+        @dynflow.initialize!
+        @dynflow
+      end
```
ohadlevy: nitpick: I would also consider writing it in a shorter format:
```ruby
def dynflow
  @dynflow ||= defined?(ForemanTasks) ? ForemanTasks.dynflow : Foreman::Dynflow.new
  @dynflow.require!
  @dynflow.initialize!
  @dynflow
end
```

lib/foreman/dynflow.rb
```diff
@@ -0,0 +1,98 @@
...
+
+    def initialized?
+      !@world.nil?
+    end
+
+    def initialize!
+      return unless @required
+      return @world if @world
+
+      if config.lazy_initialization && defined?(PhusionPassenger)
```
ohadlevy: does this make sense to extract to some other place? at least that if we ever move away from passenger its handled in one place?

lib/foreman/dynflow/configuration.rb
```diff
@@ -0,0 +1,152 @@
...
+      @on_init = []
+    end
+
+    # Loggers use the default logger for the time being
+
+    # for logging action related info (such as exceptions raised in side
+    # the actions' methods
+    def action_logger
+      #Foreman::Logging.logger('foreman/action')
+      Rails.logger
```
ohadlevy: why not a new logger just for tasks?

lib/foreman/dynflow/configuration.rb
```diff
@@ -0,0 +1,152 @@
...
+    # what rake tasks should run their own executor, not depending on the external one
+    attr_accessor :rake_tasks_with_executor
+
+    # if true, the ForemanTasks::Concerns::ActionTriggering will make
+    # no effect. Useful for testing, where we mignt not want to execute
+    # the orchestration tied to the models.
+    attr_accessor :disable_active_record_actions
+
+    def initialize
+      self.pool_size                = 5
```
ohadlevy: should this be configurable? what effects the size

lib/foreman/dynflow/daemon.rb
```diff
@@ -0,0 +1,71 @@
+require 'fileutils'
```
ohadlevy: should this file be in bin/ ?

ohadlevy (changes requested):
thanks @dLobatog.

I've added a few comments inline + I would like to see added testing for dynflow init, it looks pretty "complex" and potentially error prune.

Gemfile
dLobatog: That's precisely what this line is supposed to do 😄

config/application.rb
dLobatog: Any time you call `Foreman.dynflow` after the first time, it will be set (and it will be equal to the world defined in the lines below). 

E.g: RssNotifications calls `perform_later` (which calls Foreman.dynflow.world inditrectly). The next time a job runs, the variable will be fetched from here.

config/application.rb
dLobatog: Tasks loads first as this code just defines a method, until it's called it is not loaded

config/application.rb
dLobatog: I don't think that's a good idea, that code calls initialize! every time (which creates a new Dynflow world)

Gemfile
```diff
@@ -38,6 +38,7 @@ gem 'deacon', '~> 1.0'
 gem 'webpack-rails', '~> 0.9.8'
 gem 'mail', '~> 2.6'
 gem 'nokogiri', '< 1.7' if RUBY_VERSION.start_with? '2.0.'
+gem 'dynflow', '>= 0.8.20', '< 1.0.0'
```
dLobatog: Set to `:path => '../dynflow` or similar when testing, the gem is not out yet for https://github.com/Dynflow/dynflow/pull/216/files

lib/foreman/dynflow/daemon.rb
dLobatog: Actually I forgot to push the task last time, now it's pushed

ohadlevy: why is it a task? shouldn't it be a full blown daemon / service ?

Additionally, I'm not sure if we should run it using some tool to monitor its health, for example runit.

lib/foreman/dynflow/configuration.rb
dLobatog: Ran into some trouble with this - but yes, it should have it's own logger

lib/foreman/dynflow/daemon.rb
dLobatog: I was just trying to replicate the way foreman-tasks work. (https://github.com/theforeman/foreman-tasks/blob/master/lib/foreman_tasks/tasks/dynflow.rake)

@iNecas @ares Does the service in tasks do anything different?

ShimShtein:
In general, it looks nice.
I would suggest refactoring foreman-tasks to get the daemon and configuration parts out of it, and make foreman-tasks use the world defined in here.
In any case, we need to add the `dynflow-executor` script here too, since if not using foreman-tasks you won't have the ability to run dynflow as a daemon.

lib/foreman/dynflow.rb
dLobatog: I can move it if you want, it's just a warning that we have not changed in a while though

dLobatog:
Pending stuff then:

- [x] - Loggers for tasks/actions
- [x] - Bring in the [systemd service](https://github.com/theforeman/foreman-tasks/blob/master/deploy/foreman-tasks.service) and `bin/foreman-tasks`

After merge:
- [ ] - Make sure the package uses it (in foreman-packaging) - https://github.com/theforeman/foreman-packaging/pull/1549
- [x] - Get rid of the world initializing code (in foreman-tasks), except for the persistence layer to create the tasks as db objects

lib/foreman/dynflow/configuration.rb
dLobatog: Just took the default from foreman_tasks to be honest, I have not experimented with different sizes (lame excuse, I know)

iNecas: We were just planning to add this to be configured in short term

dLobatog:
@ohadlevy @iNecas Updated with loggers and the introduction onf `bin/foreman-tasks` (which the systemd foreman service should start)

bin/foreman-tasks
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env ruby
```
ohadlevy: should it be called foreman-tasks ? wont that be confusing? (or conflicting in packaging)?

bin/foreman-tasks
```diff
@@ -0,0 +1,5 @@
...
+
+foreman_root = '/usr/share/foreman'
```
ohadlevy: should this be set by a variable so installer can change it if needed?

config/application.rb
```diff
@@ -214,6 +219,18 @@ class Application < Rails::Application
 
     # Use the database for sessions instead of the cookie-based default
     config.session_store :active_record_store, :secure => !!SETTINGS[:require_ssl]
+
+    def dynflow
+      return @dynflow if @dynflow.present?
+      if defined?(ForemanTasks)
+        @dynflow = ForemanTasks.dynflow
+      else
+        @dynflow = Foreman::Dynflow.new
```
ohadlevy: nitpick - i still find this more readable:
```ruby
@dynflow = defined?(ForemanTasks) ? ForemanTasks.dynflow : Foreman::Dynflow.new
```

config/settings.yaml.example
```diff
@@ -54,3 +54,7 @@
 #    :enabled: false
 #  :templates:
 #    :enabled: true
+#  :action:
+#    :enabled: true
```
ohadlevy: what is action?

lib/foreman/dynflow.rb
```diff
@@ -0,0 +1,98 @@
+# -*- coding: utf-8 -*-
+require 'dynflow'
+
+module Foreman
```
ohadlevy: just wondering, shouldn't all of this logic be in dynflow vs foreman?

lib/foreman/dynflow.rb
ohadlevy: I do wonder if we should keep supporting this use case, now that we have foreman, we can always install dynflow service when foreman starts.

I don't know if we are actually gaining anything by running it in the same process vs using it more similar to how its used in production?

lib/foreman/dynflow/configuration.rb
```diff
@@ -0,0 +1,147 @@
+module Foreman
+  class Dynflow::Configuration
```
ohadlevy: again, why is this not part of dynflow rails integration ?

lib/foreman/dynflow/daemon.rb
```diff
@@ -0,0 +1,71 @@
...
+
+module Foreman
+  class Dynflow::Daemon
+    # load the Rails environment and initialize the executor
+    # in this thread.
+    def run(foreman_root = Dir.pwd)
+      STDERR.puts('Starting Rails environment')
```
ohadlevy: why ERR?

lib/foreman/dynflow/daemon.rb
```diff
@@ -0,0 +1,71 @@
...
+    ensure
+      STDERR.puts('Exiting')
+    end
+
+    # run the executor as a daemon
+    def run_background(command = 'start', options = {})
+      default_options = { foreman_root: Dir.pwd,
+                          process_name: 'dynflow_executor',
+                          pid_dir: "#{Rails.root}/tmp/pids",
+                          log_dir: File.join(Rails.root, 'log'),
```
ohadlevy: why file join here and not above?

bin/foreman-tasks
dLobatog: The idea is to remove it from foreman-tasks and move it here - so no - in fact it should have the same name in my opinion, to make sure it plays well with existing tooling that expects it

bin/foreman-tasks
dLobatog: I can do it but is it worth it? We haven't changed the target installation directory ever and we've relied on this path on foreman-tasks forever

config/application.rb
dLobatog: duly noted, I'll update the PR with that

config/settings.yaml.example
dLobatog: https://github.com/theforeman/foreman/pull/4316/files#diff-b227cab047536b0261bee94c657af2deR41

lib/foreman/dynflow.rb
dLobatog: Yes, in my opinion Foreman should just define the hook points that change (like `action_logger`, `dynflow_logger`). Everything else looks like it should be the same on all Rails applications, including the daemon.rb and most of configuration.rb

lib/foreman/dynflow/configuration.rb
dLobatog: https://github.com/theforeman/foreman/pull/4316/files#diff-b227cab047536b0261bee94c657af2deR41 explained here

lib/foreman/dynflow/daemon.rb
dLobatog: Carelessness, no good reason. Will change.

lib/foreman/dynflow/daemon.rb
dLobatog: No good reason, will change.

dLobatog:
Things to change:

- [x] - STDOUT in daemon
- [x] - File.join /tmp/pids
- [x] - Using a ternary in application.rb to check whether Dynflow is already started by foreman-tasks or not.
- [x] - Moving most of the code to Dynflow as it's 99% is related to it's Rails integration, not really something specific for Foreman except for the loggers. 

@ohadlevy I don't mind doing the latter, it seems easy but in case it's not - is it a deal breaker? I fully agree it's really not Foreman's business to do all this initialization, but given the timeframes we're moving on, if moving it to dynflow would delay this integration for 1.15, I would prefer to keep it here.

I'll give it a try - hopefully it's easy enough to do 😄

dLobatog:
Latest update moves the world initialization over to dynflow. Foreman just needs to override the loggers, the rest of the stuff is provided by dynflow (a much nicer interface IMO than before).

https://github.com/Dynflow/dynflow/pull/218 is necessary for this to work

cc @ohadlevy @iNecas

bin/foreman-tasks
ehelms: I would just double check that this path is the same as for Debian deployments. Perhaps @mmoll  can weigh in.

ehelms:
I think this is a good direction to head, but I do have some questions looking at this from a high level (and happy to help test some of the scenario questions below):

 1) Would you mind opening the packaging PR alongside this so we have it and can ensure we aren't missing something big that bit the foreman-tasks being included change?
 2) Has this been tested yet with foreman-tasks present as it is today? Will the two clash?
 3) Have you tested this with a fuller stack? e.g. Katello and REX
 4) One nice part of Foreman Tasks is that you get a UI and API for seeing and monitoring the tasks that are running including constantly running or periodic tasks. This won't provide that ability. For those type of operations, how do we keep track of or identify failures?
 5) Will this run in the same process as the web server for development and production or keep the split between processes depending on environment?

dLobatog:
@ehelms 

1. Yes - that's not a problem. I will open the PR in packaging that does the same thing as foreman-tasks, ship the service with the RPM and puppet-foreman should take care of starting it as it does now.  I would also have to make a PR to foreman-tasks removing the Dynflow dependency AND removing the initializer code except for the Persistence layer. (done https://github.com/theforeman/foreman-tasks/pull/235)
2. Yes - it uses the Dynflow world that ForemanTasks initializes (https://github.com/theforeman/foreman/pull/4316/files#diff-b1fe55db50c712fef0673345e5b9c0d9R225). The first ActiveJob that you run, will show a bunch of warnings though, because the consistency check that ForemanTasks provides (https://github.com/theforeman/foreman-tasks/blob/master/app/models/foreman_tasks/task/dynflow_task.rb#L96) will fail. That's expected for the moment though, as you're creating tasks in dynflow outside of the ForemanTasks framework - so consistency.. doesn't check.
3. Yes - I only saw the problem described in 2, after syncing products and sending simple jobs.
4. That's what https://github.com/theforeman/foreman-tasks/blob/master/lib/foreman_tasks/dynflow/persistence.rb is supposed to do, so as long as you have the plugin, tasks will show up as usual. If you don't have it, they won't. See a few lines below for the rationale behind this change.
5. Same as with foreman-tasks currently, you can start the executor independently (see rake dynflow:executor) or let it run in development in the same process.

The reason for the introduction of this change, is that doing things like a service that checks recurring things and creates notifications (say, RSS blog checking, Subscription expiration, Pulp disk space remaining in proxies...) is impossible w/o the -tasks dependency. Furthermore, tasks on it's own does not allow you to schedule things easily, so we would have to bring the scheduling code from RemoteExecution. https://github.com/theforeman/foreman/pull/4240 is an example of a simple RSS notification system that needs some backend executor and scheduler (it uses DelayedJob but making it use dynflow should be as simple as changing `config.active_job_adapter = :dynflow`)

In addition to that, ActiveJob provides a well-documented, known by the community, and radically simple way of creating and scheduling tasks. This is something ForemanTasks unfortunately did not have in my opinion (see the first post in this PR for an example of a simple task!). 

After this PR, I think ForemanTasks would still remain useful as a persistence/UI/API layer for Dynflow, but I would expect most tasks to be written with ActiveJob instead (http://edgeguides.rubyonrails.org/active_job_basics.html).

bin/foreman-tasks
dLobatog: As far as I can tell, the init script for debian does indeed use this script (with the hardcoded path) for tasks (https://github.com/theforeman/foreman-tasks/blob/master/deploy/foreman-tasks.init)

mmoll: I guess this can stay hardcoded for now, as core is only supported on Linux, where that path is always correct.

dLobatog:
[test foreman]
[test upgrade]
[test katello]

Gemfile
```diff
@@ -38,6 +38,8 @@ gem 'deacon', '~> 1.0'
...
+gem 'daemons'
```
ohadlevy: doesn't dynflow require daemons? should it?

bin/foreman-tasks
ohadlevy: I still find it confusing, and I think users would too.

config/settings.yaml.example
ohadlevy: I find the name action unfriendly, not sure why would users ever want to look in the "action" logs?

ohadlevy (changes requested):
minor comments, mostly about naming things.

Please make sure tests pass :)

as mentioned on IRC, I think a few more things need to be resolved:
1. packaging (what is moved from tasks to core etc)
2. installer updates (new service etc)
3. docs for both developers (how to use) and users.

thanks!

Gemfile
dLobatog: It doesn't, foreman-tasks does: https://github.com/theforeman/foreman-tasks/blob/master/foreman-tasks.gemspec#L35

Gemfile
dLobatog: It's out now actually

bin/foreman-tasks
dLobatog: For users - if they used foreman-tasks, they should not see any difference. For devs, I do agree it's confusing that this is in core but I prefer that to possibly break any tooling related with this.

config/settings.yaml.example
dLobatog: I can keep the old name 'foreman-tasks/action' if you want to, does that help? Otherwise I can call it 'job-processing' if it's any better

config/application.rb
```diff
@@ -181,7 +185,9 @@ class Application < Rails::Application
       :ldap => {:enabled => false},
       :permissions => {:enabled => false},
       :sql => {:enabled => false},
-      :templates => {:enabled => true}
+      :templates => {:enabled => true},
```
dLobatog: All test failures happened because I removed this logger (templates) by mistake

dLobatog:
Ruby 2.0.0 tests are broken on Dynflow (https://github.com/Dynflow/dynflow/pull/219) - after that they should pass here. If you check in Jenkins, tests on Ruby post 2.0 pass.

https://github.com/theforeman/puppet-foreman/pull/530 contains the puppet-foreman changes. foreman-installer doesn't need anything as far as I can see.

Other than this, I think the only change left required is to ship bin/foreman-tasks and the services in the `foreman` package instead of the `foreman-tasks` package, then we're ready (https://github.com/theforeman/foreman-packaging/pull/1549)

Gemfile
ohadlevy: should it?

config/settings.yaml.example
ohadlevy: or background etc, action doesnt really mean anything.

Gemfile
dLobatog: Probably not, it's only used if you use Dynflow with Rails. I don't mind either way (requiring daemons here or in Dynflow), @iNecas ?

ShimShtein: if you plan on running `bin/foreman-tasks`, you will need this gem. If it's not installed, the system will tell you that this gem is missing.
Originally, it was in a separate group, so the user could decide if it's wanted or no.

deploy/foreman-tasks.init
```diff
@@ -0,0 +1,231 @@
+#!/bin/bash
```
ohadlevy: why put it here and not in the packaging pr? it seems consistent with the rest?

deploy/foreman-tasks.service
```diff
@@ -0,0 +1,17 @@
+
+Unit]
```
ohadlevy: ditto

deploy/foreman-tasks.sysconfig
```diff
@@ -0,0 +1,14 @@
+FOREMAN_USER=foreman
```
ohadlevy: ditto

ohadlevy:
@dLobatog this needs a rebase now :)

ohadlevy (changes requested):
Please move the service related files (all the files under /deploy) to the packaging pr - thanks!

dLobatog:
Moved the deploy/ files to the packaging PR

config/application.rb
```diff
@@ -215,6 +221,15 @@ class Application < Rails::Application
 
     # Use the database for sessions instead of the cookie-based default
     config.session_store :active_record_store, :secure => !!SETTINGS[:require_ssl]
+
+    def dynflow
+      return @dynflow if @dynflow.present?
+      @dynflow = defined?(ForemanTasks) ? ForemanTasks.dynflow :
+        Dynflow::Rails.new(Foreman::Dynflow::Configuration.new)
```
ohadlevy: nitpicl: I know I originally asked for it,  but if its multi line, then it can probably go back to the older version or 
```ruby
@dynflow = 
if defined(ForemanTask?)
  ForemanTasks.dynflow
else
  Dynflow::Rails.new(Foreman::Dynflow::Configuration.new)
end
```

ohadlevy: or even extract to a simple method for clarity

ohadlevy:
overall :+1:

dLobatog:
Updated config/application.rb to the old version

bin/foreman-tasks
ohadlevy: @dLobatog what is the status of this?

lib/tasks/dynflow.rake
```diff
@@ -0,0 +1,5 @@
+namespace :dynflow do
```
ohadlevy, line 1: where / when is this being used?

dLobatog: In development, to have a separate executor from the main process (https://github.com/theforeman/foreman-tasks/#production-mode)

ohadlevy: so why not use foreman Procfile vs run in the same process?

ohadlevy:
@dLobatog I'm uncertain what is the status of the PR ?

dLobatog:
Rebased with the latest develop - this was ready to merge, however there have been some changes in Foreman-Tasks initialization (it was improved in ways that never got to Dynflow::Rails - https://github.com/theforeman/foreman-tasks/commit/d3051b91d63a1c809b78fbd48187ba364c391a89). 

I'm porting the memory monitoring support from foreman-tasks to Dynflow now, it should be easy, mostly cut and paste, once that's done, this should be ready to merge with:

https://github.com/theforeman/foreman-tasks/pull/235
https://github.com/theforeman/foreman-packaging/pull/1549
https://github.com/theforeman/puppet-foreman/pull/530 (I'll remove the 1.14 bits from here)
https://github.com/theforeman/foreman-installer/pull/222
https://github.com/theforeman/foreman-tasks/pull/238 (optional, allows jobs created with ActiveJob to show up in the foreman-tasks UI)

dLobatog (commented):
Updated to use Dynflow 0.8.25 which contains the latest fixes for memory profiling

bin/foreman-tasks
ohadlevy: I still find the bin name confusing.

lib/tasks/dynflow.rake
ohadlevy: still open question?

bin/foreman-tasks
ohadlevy: I would prefer this is relative to the file path, I just tried on my checkout 
```
➜  foreman git:(4316) ✗ ./bin/foreman-tasks -h                                                                                                                           git:(4316|… 
/home/ohad/.rvm/rubies/ruby-2.4.0/lib/ruby/2.4.0/rubygems/core_ext/kernel_require.rb:55:in `require': cannot load such file -- /usr/share/foreman/config/application (LoadError)
	from /home/ohad/.rvm/rubies/ruby-2.4.0/lib/ruby/2.4.0/rubygems/core_ext/kernel_require.rb:55:in `require'
	from ./bin/foreman-tasks:4:in `<main>'
```

ohadlevy: you can replace it with 
```ruby
foreman_root = File.join( __FILE__, "..", "..")
```

dLobatog: This shouldn't be run on your local checkout (just like you're not running bin/foreman-tasks from the plugin folder). This is moved to /usr/bin , then the systemd service calls it (https://github.com/theforeman/foreman-packaging/pull/1549/files#diff-a46acbb74d582e9d3a8ee262a72b237cR11). 

On local setups as yours, you should use https://github.com/theforeman/foreman-tasks/#production-mode if you want to have the executor separate from the Rails process.

lib/tasks/dynflow.rake
dLobatog: I don't really understand the question. This is the current situation (before merging this PR):

1. In development, by default foreman-tasks & Dynflow run as part of the Rails process. 
2. In development, you can decide to run Dynflow as a separate process by calling `rake dynflow:executor`. Most people don't do this unless it's a git checkout installation, I only knew about this as I wanted to test ActiveJob with the detached executor. It works the same as the one in the web process.
3. In production, this task isn't used, instead `/usr/bin/foreman-tasks start` starts the executor via [systemd](https://github.com/theforeman/foreman-packaging/pull/1549/files#diff-a46acbb74d582e9d3a8ee262a72b237cR11) (and https://github.com/theforeman/foreman-tasks/blob/master/deploy/foreman-tasks.service#L11).

None of 1 2 or 3 are modified after this PR - it works exactly the same. What does the Procfile have to do here? You want to modify the "default" (if you use the Procfile) to run the Dynflow executor as a separate process? Why?

ohadlevy: > In development, by default foreman-tasks & Dynflow run as part of the Rails process.
that's foreman-tasks, does it make sense to continue this way in core too?

TBH imho the answer is no, but I'm fine if we do it outside the scope of this pr.

> In development, you can decide to run Dynflow as a separate process by calling rake dynflow:executor. Most people don't do this unless it's a git checkout installation, I only knew about this as I wanted to test ActiveJob with the detached executor. It works the same as the one in the web process.

Same answer as before.

> In production, this task isn't used, instead /usr/bin/foreman-tasks start starts the executor via systemd (and https://github.com/theforeman/foreman-tasks/blob/master/deploy/foreman-tasks.service#L11).

I get that - my point is that this is foreman-tasks behavior - do we want to keep it? 
I personally find the name confusing, and would expect a package called foreman-tasks to deliver the foreman-task binary.

dLobatog (commented):
Updated to *remove* `bin/foreman-tasks` - what it does can be done just as well with `bin/dynflow-executor`, with some changes in Puppet + Packaging

ohadlevy (approved):

ohadlevy:
test failures are unrelated - thanks @dLobatog !!!

ares:
Thanks for merging @ohadlevy, please don't forget to set the release field in redmine after merge. I set it to 1.16.0.

ohadlevy:
Thanks, I was so excited that I forget to ;)

PR #4318, opened 2017-02-22
Fixes #17866 - Remove PXE Properties from Gentoo Operatingsystem

daugustin:
The Gentoo Operatingsystem cannot be installed via PXE and can only be provisioned via Image.

Implementing the pxe method causes the exception as described in #[17866](http://projects.theforeman.org/issues/17866).

Removing the methods fixes the template rendering of Gentoo Templates.


mmoll:
ok to test

dLobatog:
Thanks @daugustin  - the tests in katello are broken for other reason (https://github.com/Katello/katello/pull/6447)

PR #4320, opened 2017-02-23
Fixes #18639 - lock all templates we seed

ares:
The PR looks big but what it does is very simple
- extracts names into constants and into separate seed files so they can be reused safely in migration without triggering the seed
- adds a migration that sets locked flag for seeded templates to true
- makes sure all newly seeded templates are locked

For more information and motivations, please see the discussion on [foreman-dev list](https://groups.google.com/forum/#!searchin/foreman-dev/templates%7Csort:relevance/foreman-dev/qCWVoHPBhd0/N9LbfQxrDAAJ)

### DIFF
--- a/app/models/template.rb
+++ b/app/models/template.rb
@@ -1,5 +1,6 @@
 class Template < ActiveRecord::Base
   include Exportable
+  attr_accessor :modify_locked, :modify_default
 
   validates_lengths_from_database
 
@@ -58,6 +59,20 @@ class Template < ActiveRecord::Base
     name.downcase.delete('-').gsub(/\s+/, '_') + '.erb'
   end
 
+  def ignore_locking
+    self.modify_locked = true
+    yield
+    self.modify_locked = false
+    self
+  end
+
+  def ignore_default
+    self.modify_default = true
+    yield
+    self.modify_default = false
+    self
+  end
+
   private
 
   def allowed_changes
@@ -72,19 +87,19 @@ class Template < ActiveRecord::Base
     actual_changes = changes
 
     # Locked & Default are Special
-    if actual_changes.include? 'locked'
-      unless User.current.can?("lock_#{self.class.to_s.underscore.pluralize}", self)
+    if actual_changes.include?('locked') && !self.modify_locked
+      if User.current.nil? || !User.current.can?("lock_#{self.class.to_s.underscore.pluralize}", self)
         errors.add(:base, _("You are not authorized to lock templates."))
       end
     end
 
-    if actual_changes.include? 'default'
-      unless User.current.can?(:create_organizations) || User.current.can?(:create_locations)
+    if actual_changes.include?('default') && !self.modify_default
+      if User.current.nil? || !(User.current.can?(:create_organizations) || User.current.can?(:create_locations))
         errors.add(:base, _("You are not authorized to make a template default."))
       end
     end
 
-    unless actual_changes.delete_if { |k, v| allowed_changes.include? k }.empty?
+    if  !self.modify_locked && !actual_changes.delete_if { |k, v| allowed_changes.include? k }.empty?
       errors.add(:base, _("This template is locked. Please clone it to a new template to customize."))
     end
   end
--- a/db/migrate/20170223114114_lock_seeded_templates.rb
+++ b/db/migrate/20170223114114_lock_seeded_templates.rb
@@ -1,0 +1,14 @@
+require Rails.root + 'db/seeds.d/07-data.rb'
+require Rails.root + 'db/seeds.d/08-data.rb'
+
+class LockSeededTemplates < ActiveRecord::Migration
+  def up
+    names = (SEEDED_TEMPLATES + SEEDED_PARTITION_TABLES).map { |attrs| attrs[:name] }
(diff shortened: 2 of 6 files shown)


timogoebel:
@ares: How about taking a more radical approach and rename all seeded templates that have previously been modified so we can make sure all original templates are present and locked. I think, we should go all the way with this.

db/seeds.d/07-provisioning_templates.rb
```diff
@@ -105,7 +17,8 @@
 
       t = ProvisioningTemplate.create({
         :snippet  => false,
-        :template => contents
+        :template => contents,
+        :locked => true
       }.merge(input))
 
       if t.default?
```
timogoebel: I know, this is unrelated to this PR, but isn't `t.default?` always true here?

ares: yes, it's always set like that few lines above, I'll remove the if condition while I'm at it

ares:
> How about taking a more radical approach and rename all seeded templates that have previously been modified so we can make sure all original templates are present and locked. I think, we should go all the way with this.

@timogoebel I'm afraid this won't work reliably for snippets, their names are hardcoded in templates. I'd like to avoid updating their names in existing templates. Since we override default templates with every seed (not sure for how many versions) I hope there won't be custom changes in seeded templates. If renaming is preferred, maybe we could do it for all templates except snippets?

dLobatog:
Tested and 👍 to the change once Katello tests are green. I was a bit confused by one thing:

If I modified the "PXElinux default template" because I wanted to add custom stuff, then update Foreman to a version with this, my change remains in the locked template, but if I want to keep it i should clone it or rename it, correct?

If that's the case, still 👍 but let's warn about it in the release and upgrade notes

timogoebel:
> If I modified the "PXElinux default template" because I wanted to add custom stuff, then update Foreman to a version with this, my change remains in the locked template, but if I want to keep it i should clone it or rename it, correct?

Actually, I think this is still bad design. I think, we should only lock seeded templates and not continue with this mess.
Let's ship locked default templates managed by us and users can "fork" them if they choose to do so.

ares:
I think that's what the PR does. @dLobatog if you want to keep such changes to survive the update, you need to rename it anyway because we currently override any custom changes on `rake db:seed`.

dLobatog:
@timogoebel That's what this PR is for 😄  the locking functionality was there because Katello did this for a long time already.

ares:
Rebased to avoid compute resource test failures, should be green now.

timogoebel:
@dLobatog / @tbrisker  : Does anyone of you have time to check this one out? We should get this in. :-)

ares:
Foreman issues seems unrelated
* HostsControllerTest.test_0070_select multiple action with valid host_ids param should return a selection page (pg/ruby 2.4.0)
* ComputeResourceJSIntegrationTest.test_0003_compute resource password isn't deleted when testing connection (pg/ruby 2.0.0)

ares:
[test katello]

ares:
Rebased again

tbrisker:
code-wise looks good, waiting for feedback from @dLobatog since he reviewed this more deeply.

app/models/template.rb
```diff
@@ -1,5 +1,6 @@
 class Template < ActiveRecord::Base
   include Exportable
+  attr_accessor :modify_locked, :modify_default
```
dLobatog, line 3: How about making this an `attr_reader` in order to prevent modifications outside of this class (using the ignore_locking/ignore_default helpers)?

db/migrate/20170223114114_lock_seeded_templates.rb
```diff
@@ -0,0 +1,14 @@
+require Rails.root + 'db/seeds.d/07-data.rb'
+require Rails.root + 'db/seeds.d/08-data.rb'
+
+class LockSeededTemplates < ActiveRecord::Migration
+  def up
+    names = (SEEDED_TEMPLATES + SEEDED_PARTITION_TABLES).map { |attrs| attrs[:name] }
+    Template.where(:name => names).update_all(:locked => true)
```
dLobatog, line 7: Why is this needed if `:locked => true` is set in the seed https://github.com/theforeman/foreman/pull/4320/files#diff-151633105c29141fe3aac7603de185a5R25 ? Is this only relevant for partition tables?

dLobatog (approved):
@ares 👍 from me again, I just have 2 questions you can reply to inline. Thanks for this!

app/models/template.rb
ares: yeah I could do that but I though people might find it useful e.g. in tests when they just disable locking in setup and re-enable in teardown even though the `ignore_locking` is preferable elsewhere

db/migrate/20170223114114_lock_seeded_templates.rb
ares: On existing db we need to lock all templates, they were seeded unlocked in previous versions. So we find all templates by name that would be locked in new installation and lock them.

dLobatog:
Thanks @ares & @tbrisker @timogoebel !

PR #4321, opened 2017-02-23
Fixes #18538 - add sass support to webpack

matanwerbner:
This can really improve css maintainablity

### DIFF
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -48,6 +48,13 @@ var config = {
       {
         test: /(\.png|\.gif)$/,
         loader: "url-loader?limit=32767"
+      },
+      {
+        test: /\.scss$/,
+        loader: ExtractTextPlugin.extract(
+          'style', // The backup style loader
+          'css?sourceMap!sass?sourceMap'
+        )
       }
     ]
   },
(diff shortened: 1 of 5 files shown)


tbrisker:
@matanwerbner - looks like you got a typo in the issue number? Also one of the tests seems to have broken. other than that :+1:

dLobatog:
@matanwerbner Thanks - I agree with @tbrisker, make sure tests pass and please check out packaging for these 2 new packages.
https://github.com/theforeman/foreman/pull/4291#issuecomment-280316317

gailsteiger:
Excellent. OK to merge.

 +1 for the source maps.

@tbrisker  Can we remove the 'sass-rails' gem and rely on webpack exclusively?

tbrisker:
@gailsteiger tests are failing, so this isn't ok to merge yet. As for `sass-rails`, afaiu it is needed so the legacy assets can be written in sass, so we won't be able to remove it until all stylesheet assets are moved to webpack (which i'm not sure is feasible anyways).

ohadlevy:
@matanwerbner for future PR's, there is no need to close the PR, simply force push another branch into it.

PR #4322, opened 2017-02-23
Fixes #18640 - realign templates structure

ares:
Based on #4320 to avoid rebasing problems later. What this does
- move files in app/views/unattended according to structure in community templates
- updates `script/sync_templates.sh` to work with the new structure of community templates
- update all hardcoded paths and names accordingly
- syncs up with current community-templates version

We can't get rid of hardcoded paths just yet. Next step will be syncing metadata from Foreman to community-templates so we have os_family for partition tables and template kinds pro provisioning templates set correctly. Then we can just iterate over files and do the import based on what we find in metadata.

I verified that seed and file are in sync by comparing SEEDED_TEMPLATES and files I found using Dir.glob. Also I tried doing the seed on empty DB, all seems to fit. The same applies to partition tables.


app/views/unattended/provisioning_templates/PXEGrub/kickstart_default_pxegrub.erb
```diff
@@ -0,0 +1,68 @@
...
+    # static network configuration
+    ip = @host.provision_interface.ip
+    mask = subnet.mask
+    gw = subnet.gateway
+    dns = [subnet.dns_primary]
+    if subnet.dns_secondary != ''
+      dns.push(subnet.dns_secondary)
+    end
+    if (@host.operatingsystem.name.match(/Fedora/i) && major < 17) || major < 7
+      # old Anacoda found in Fedora 16 or RHEL 6 and older
```
dLobatog, line 39: nitpick 'Anaconda'

dLobatog:
I didn't find any problems in terms of the number of templates moved. I didn't check the content but I trust it'll be fine - or you can send PRs afterwards to fix it. Check out tests in foreman & katello though @ares 😄

app/views/unattended/provisioning_templates/PXEGrub/kickstart_default_pxegrub.erb
ares: Must be fixed in community-templates -> https://github.com/theforeman/community-templates/pull/376

ares:
Foreman tests failures are unrelated HostsControllerTest.test_0070_select multiple action with valid host_ids param should return a selection page on pg/ruby 2.0 and 2.3

dLobatog:
Thanks @ares !

PR #4324, opened 2017-02-23
Refs #17272 - fix sql condition

iNecas:
1==1 is not a valid SQL condition.


ares:
Thanks @inecas, merged as 4ada010b716a46bbfea045b0fcac2a5ec60a9609

PR #4325, opened 2017-02-24
Fixes #18657 - Allow ERB in settings files

ehelms:
This will allow, for example, the ability to have environment variables in a settings file.

### DIFF
--- a/config/settings.rb
+++ b/config/settings.rb
@@ -4,7 +4,7 @@ require "#{root}/app/services/foreman/version"
 
 settings_file = Rails.env.test? ? 'config/settings.yaml.test' : 'config/settings.yaml'
 
-SETTINGS = YAML.load_file("#{root}/#{settings_file}")
+SETTINGS = YAML.load(ERB.new(File.read("#{root}/#{settings_file}")).result)
 SETTINGS[:version] = Foreman::Version.new
 SETTINGS[:unattended] = SETTINGS[:unattended].nil? || SETTINGS[:unattended]
 SETTINGS[:login]    ||= SETTINGS[:ldap]
@@ -14,5 +14,5 @@ SETTINGS[:puppetssldir]  ||= "#{SETTINGS[:puppetvardir]}/ssl"
 
 # Load plugin config, if any
 Dir["#{root}/config/settings.plugins.d/*.yaml"].each do |f|
-  SETTINGS.merge! YAML.load_file f
+  SETTINGS.merge! YAML.load(ERB.new(File.read(f)).result)
 end


timogoebel:
I'm unsure if this is a good idea. What is the use case for this? What happens if there are parse failures in the ERB?

ehelms:
The use case for this is being able to provide settings files from core or plugins that can make use of things like environment variables. For example, to be able to define a secret in an environment variable, or the location of a backend service. You can do this today with the Rails `database.yml` file.

To be fair, what happens if the yaml parses incorrectly in the current state? I don't mind trying to add error recovery, but I don't feel that's a new issue.

dLobatog:
I don't think it would cause any more trouble than writing invalid YAML as @ehelms said. If you don't mind, let's see if other people comment on it and revisit it in a couple of days. I say 👍 - although I'd like to see some plan to use this rather than have a feature no one is using!

PS: I'm setting "Reached an impasse" just so that the issue is more visible and hopefully people will wonder why and voice their opinion here. It doesn't look very controversial to me but let's see if there is anyone else who could be against this.

ehelms:
I wouldn't say no one is using, since I opened this so I could use it instead of carrying around a patched fork. The environment in which this becomes highly useful is my side project to run our stack in OpenShift V3 where environment variables become the way to handle some secrets and identify services more dynamically.

ares (commented):
I think it's a nice feature.  If you add it for the main config file few lines above (consistency even if not used right away) and add a test for this file (hard to test it for plugins), I don't see a problem. :+1 from me.

timogoebel:
@ehelms : I wasn't going to block this, I was just curious what the actual aim is here. If you think, the ERB rendering is useful, I don't mind having it in the code. As said before, rails allows this for `database.yml` already. I'd suggest allowing this for the main settings file as well as @ares already said.

I do like the idea of being able to configure Foreman via environment variables. But wouldn't it be better to properly integrate this than relying on a config file with ERB rendering? E.g. we could make an environment variable called `FOREMAN_MARK_TRANSLATED=true` available as `SETTINGS[:mark_translated]`.

ehelms:
Thats a good thought @timogoebel and I think would be do-able for the mainline settings. The tough part would be how to handle nested settings such as plugins have. For example:

```
katello:
  pulp:
    url: "https://example.com"
```

I could essentially try to recurse through the hash collecting keys and tell users the assumption is that each nested piece is separated by underscores. Thus, in the example above you'd get KATELLO_PULP_URL. This might be too complex though.

lzap:
Instead of ERB I'd rather see something more explicit, like `${ENV_VARIABLE}`, then we could walk down the tree and replace those.

ares:
I'd rather avoid introducing new variable mechanism. Why not using the same thing that works in the Rails ecosystem? I mean fixtures yamls and database.yml

bastilian:
I would also not introduce a new way of defining variables. 
Using ERB for/in YAML is simple and effective and used for rails' database.yml for [some time](https://github.com/rails/rails/commit/b25f9af998ad26653083e004b7a6cb9ca0061940).

It might be better to have a `SettingsLoader` in lib that just does the ERB/Yaml loading or even better having a `Settings` object to assign to `SETTINGS`, which takes care of `settings.yml` loading, parsing, etc. essentially does what `config/settings.rb` does, but in a proper manner and testable.

We could also go for an existing solution like [railsconfgi/config](/railsconfig/config).

ares:
I agree, setting loader needs some improvements. But let's keep that separate from this PR, @bastilian I think that might be worth of RFC. Meanwhile I think this change would be useful as it is today.

bastilian:
Yes, agreed, if we want to make settings loading more robust, we should do so in separately and see what is the best approach with an RFC.

dLobatog:
Let's merge since there seems to be a consensus this addition does no harm & fixes @ehelms 's use case.

About the Travis failure - the logs do not show anything at all, and tests seem to pass just fine on 3 node versions locally so I'll go with merging. Maybe it was down when the PR was submitted due to the S3 fiasco or something.

dLobatog:
Thanks @ehelms !

PR #4327, opened 2017-02-24
refs #16982 - check _ids getters as admin user in taxonomy tests

domcleal:
After a request and User.current is back to nil, the _ids getters will
not return any results as the default taxonomix scopes don't permit it.
The clone test was instead testing that _no_ records were associated.


dLobatog:
Thanks @domcleal , the new assertions you added will prevent this from happening again

PR #4328, opened 2017-02-24
refs #16982 - pass ID, not models into model.find

domcleal:


dLobatog:
Thanks @domcleal - didn't know what was the issue, I now see it's deprecated behavior

domcleal:
Yes, it was originally fixed in 331fcdc.

PR #4329, opened 2017-02-24
refs #16982 - remove User.current deassignment (no such user)

domcleal:
Allows the scope change to be reverted, as User.current is no longer set
to `nil` (there is no 'admin' user).

This was relying on a bug in Ruby on Rails 4.2 where the `unscoped` call
filtered through thread variables into Subnet.subnet_for which calls
Subnet.all. This is fixed in 5.0, so the user must be set correctly.


tbrisker:
Thanks @domcleal !

PR #4330, opened 2017-02-24
fixes #18662 - stop default scope being overridden by association

domcleal:
Ensure the Taxonomix empty default scope isn't overridden by association
scopes which (effectively) calls `.where(:id => ...)` and overrides the
value of :id set in this default scope. This occurs on Rails 5.0 which
merges the scopes more correctly/effectively than 4.2, and so invisible
resources became visible through the association getter.

Like the case where there is at least one visible resource, a string SQL
fragment is used instead to prevent it being overridden by ActiveRecord.

The host test now uses an admin user as it was testing that the host's
environment was nil (since hostgroup#environment now returns nil due to
the default scope), rather than a present value. This failed on 5.0 when
hostgroup#environment returned the invisible record.

### DIFF
--- a/app/models/concerns/taxonomix.rb
+++ b/app/models/concerns/taxonomix.rb
@@ -134,7 +134,7 @@ module Taxonomix
         scope
       when []
         # If *no* taxable ids were found, then don't show any resources
-        scope.where(:id => [])
+        scope.where('1=0')
       else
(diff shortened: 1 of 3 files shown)


dLobatog:
Thanks @domcleal !

PR #4331, opened 2017-02-24
refs #17228 - modify, don't re-define existing constant

domcleal:
Prevents warnings during startup:

    app/models/setting/provisioning.rb:9: warning: already initialized constant Setting::BLANK_ATTRS
    app/models/setting.rb:13: warning: previous definition of BLANK_ATTRS was here


dLobatog:
Thanks @domcleal !

PR #4332, opened 2017-02-24
fixes #18664 - ignore missing callbacks in SmartProxyAuth concern

domcleal:
Filters that are only registered on the UI controllers cause errors
when using SmartProxyAuth on an API controller under Rails 5.0.


dLobatog:
👍 - I agree it's not worth separating into multiple concerns and just keep on ignoring if the filter doesn't exist as usual - it's actually how we've operated for some time so it's the better tested behavior.

PR #4333, opened 2017-02-24
fixes #18665 - call #to_h before comparing AC::Parameters to hash

domcleal:
Allows comparisons when ActionController::Parameters is separated from
Hash in Rails 5.0. #permit! is now called on inner hashes sent through
KeepParam (similar to rails/rails@e86524c in 5.1) so they are included
in the #to_h permitted output.


tbrisker:
@domcleal Looks good in general, one question though - is the change to `keep_param` only needed for 5.0 and will not be needed for 5.1? If so, could you add a comment or an issue in redmine to remove it when upgrading to 5.1? Otherwise ready to merge.

domcleal:
Sorry, I didn't mean that this change is only necessary for 5.0, it will need to be kept as long as KeepParam is present.

I meant instead that KeepParam as a whole can be deprecated and replaced with the feature in Rails 5.1 at some point, and this roughly matches its behaviour in permitting all of the sub-hash.

tbrisker:
@domcleal Alright, thanks for the clarification. Merging.

PR #4334, opened 2017-02-26
Fixes #18638 - add sass support to webpack

matanwerbner:
This will allow us to use sass in from webpack modules using require()

### DIFF
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -48,6 +48,13 @@ var config = {
       {
         test: /(\.png|\.gif)$/,
         loader: "url-loader?limit=32767"
+      },
+      {
+        test: /\.scss$/,
+        loader: ExtractTextPlugin.extract(
+          'style-loader', // The backup style loader
+          'css-loader?sourceMap!sass-loader?sourceMap'
+        )
       }
     ]
   },
--- a/webpack/assets/javascripts/react_app/components/hosts/PowerStatus.js
+++ b/webpack/assets/javascripts/react_app/components/hosts/PowerStatus.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import Loader from '../common/Loader';
-import './PowerStatus.css';
+import './PowerStatus.scss';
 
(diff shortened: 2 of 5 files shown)


webpack/assets/javascripts/react_app/components/hosts/PowerStatus.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import Loader from '../common/Loader';
-import './PowerStatus.css';
+import './PowerStatus.scss';
```
ohadlevy, line 3: not related to this pr?

matanwerbner: i put it in as a kind of test to see that scss import works.

config/webpack.config.js
```diff
@@ -48,6 +48,13 @@ var config = {
       {
         test: /(\.png|\.gif)$/,
         loader: "url-loader?limit=32767"
+      },
+      {
+        test: /\.scss$/,
+        loader: ExtractTextPlugin.extract(
+          'style', // The backup style loader
```
tbrisker: shouldn't this be 'style-loader'? Or is the loader above (for regular css) incorrect?

matanwerbner: from webpack documentation:

> By convention, though not required, loaders are usually named as XXX-loader, where XXX is the context name. For example, json-loader.
> You may reference loaders by its full (actual) name (e.g. json-loader), or by its shorthand name (e.g. json).

in webpack2 the "loader" appendix is mandatory, so i will change my code

tbrisker: Let's be consistent then, since others are `-loader` please use that for now and open another PR to remove them if you prefer.

tbrisker (changes requested):

matanwerbner:
done. anything else guys? @ohadlevy ?

tbrisker (approved):

ohadlevy (approved):

tbrisker:
Thanks @matanwerbner !

PR #4336, opened 2017-02-26
Fixes #18675 - added telemetry stack with PCP integration

lzap:
Rails provides many instrumentation points which can be very easily hooked and exported to some monitoring framework. This is completely rewrriten from scratch integration with PCP (Performance Co-Pilot) framework.

This patch aims to be as simple as possible to allow some integration for existing stable versions. It is the first version which we can extend afterwards. It requires just one dependency (pcptrace) I wrote which does not have a RPM package. It uses simple API called "trace" for application instrumentation. For production deployments do:

~~~
yum -y install rh-ruby22-ruby-devel
scl enable tfm -- gem install pcptrace
~~~

For the record, I am leaving my initial description about initial version which was using statsd protocol instead:

Protocol statsd is very simple, UDP based and provides decent integration options. This change will be unintrusive, telemetry will be opt-in and it will be always disabled for test environment. It will also work without any external service sending the telemetry data into Rails logger if needed.

This initial implementation provides Rails controller processing times for individual controllers and actions (view, db, total times in ms). In addition, it provides number of success/failed login attempts, ActiveRecord instantianed classes, Foreman templates rendering time and ActionMailer sent mail counts.

The only dependency introduced is pure-Ruby library "statsd-instrument" that provides client library for statsd daemon. The idea is to run this daemon on localhost aggregating the data and/or sending to RRD or other monitoring solutions.

My goal is to spend as little cpu cycles on telemetry as possible, for this reason Rails will export the data via the UDP standard protocol to statsd and the daemon itself will do aggregation and further integration. More details in the articles below.

I will provide more details next week about how to set statsd with Foreman, you can test with "logger" output setting that will put the telemetry data to Rails log meanwhile. My favourite server implementation is collectd (plugin collectd-statsd), because I already wrote a Foreman plugin that integrates with collectd. Further reading

https://codeascraft.com/2011/02/15/measure-anything-measure-everything/
https://signalvnoise.com/posts/3091-pssst-your-rails-application-has-a-secret-to-tell-you
https://anomaly.io/statsd-in-collectd/
https://github.com/Shopify/statsd-instrument

### DIFF
--- a/config/initializers/telemetry.rb
+++ b/config/initializers/telemetry.rb
@@ -1,0 +1,30 @@
+require 'pcptrace'
+
+ActiveSupport::Notifications.subscribe /process_action.action_controller/ do |*args|
+  event = ActiveSupport::Notifications::Event.new(*args)
+  controller = event.payload[:controller]
+  action = event.payload[:action]
+  format = event.payload[:format] || "all"
+  format = "all" if format == "*/*"
+  status = event.payload[:status]
+  key = "#{controller}.#{action}.#{format}"
+  r = PCPTrace::obs "#{key}.total_duration", event.duration
+  Rails.logger.warn("Telemetry error: " + PCPTrace::errstr(r)) if r != 0
+  PCPTrace::obs "#{key}.db_time", event.payload[:db_runtime]
+  PCPTrace::obs "#{key}.view_time", (event.payload[:view_runtime] || 0)
+  PCPTrace::counter "#{controller}.hit_#{status}", 1
+  PCPTrace::counter "total.hit_#{status}", 1
+end
+
+ActiveSupport::Notifications.subscribe /instantiation.active_record/ do |*args|
+  event = ActiveSupport::Notifications::Event.new(*args)
+  class_name = event.payload[:class_name]
+  record_count = event.payload[:record_count]
+  PCPTrace::counter "instantiation.#{class_name}", record_count
+end
+
+ActiveSupport::Notifications.subscribe /deliver.action_mailer/ do |*args|
+  event = ActiveSupport::Notifications::Event.new(*args)
+  to = event.payload[:to].gsub(/[^0-9a-z ]/i, '_')
+  PCPTrace::counter "mail_delivery.#{to}", 1
+end


config/initializers/telemetry.rb
```diff
@@ -0,0 +1,87 @@
...
+
+  def self.gauge(key, value, *metric_options)
+    if value.is_a?(Hash) && metric_options.empty?
+      metric_options = [value]
+      value = value.fetch(:value, nil)
+    end
+
+    ActiveSupport::Notifications.instrument 'telemetry.gauge', :key => key, :value => value, :metric_options => metric_options
+  end
+end
```
lzap: This should go into lib/ directory I assume.

config/initializers/telemetry.rb
```diff
@@ -0,0 +1,87 @@
...
+  Foreman::Telemetry.measure "#{key}.view_time", (event.payload[:view_runtime] || 0)
+  Foreman::Telemetry.increment "#{key}.status.#{status}"
+end
+
+ActiveSupport::Notifications.subscribe /instantiation.active_record/ do |*args|
+  event = ActiveSupport::Notifications::Event.new(*args)
+  class_name = event.payload[:class_name]
+  record_count = event.payload[:record_count]
+  Foreman::Telemetry.increment "instantiation.#{class_name}"
+end
```
lzap: This probe might be too much, I need to test on an instance with some load. Maybe we can do 10% sampling to reduce load, it will be very interesting to see if we don't make too many instances of DB records somewhere.

config/initializers/telemetry.rb
```diff
@@ -0,0 +1,87 @@
+require 'statsd-instrument'
+require 'socket'
+
+if Rails.env.test?
+  StatsD.backend = StatsD::Instrument::Backends::NullBackend.new
+else
+  if SETTINGS[:telemetry_destination] == 'logger'
```
ohadlevy: nitpick, use case statement?

config/initializers/telemetry.rb
```diff
@@ -0,0 +1,87 @@
...
+    StatsD.backend = StatsD::Instrument::Backends::LoggerBackend.new(::Foreman::Logging.logger('telemetry'))
+  elsif SETTINGS[:telemetry_destination] == 'statsd'
+    target = "#{SETTINGS[:telemetry_statsd_host]}:#{SETTINGS[:telemetry_statsd_port]}"
+    StatsD.backend = StatsD::Instrument::Backends::UDPBackend.new(target, SETTINGS[:telemetry_statsd_protocol].to_s)
+  else
+    StatsD.backend = StatsD::Instrument::Backends::NullBackend.new
```
ohadlevy: dry?

config/initializers/telemetry.rb
```diff
@@ -0,0 +1,87 @@
...
+  end
+end
+StatsD.prefix = SETTINGS[:telemetry_prefix] || "foreman.#{Socket.gethostname}"
+
+module Foreman::Telemetry
```
ohadlevy: why here vs a service class?

config/settings.yaml.example
```diff
@@ -54,3 +63,5 @@
 #    :enabled: false
 #  :templates:
 #    :enabled: true
+#  :telemetry:
+#    :enabled: true
```
ohadlevy: do you have log samples ? interesting to see if it doesnt clutter the logs.

ehelms:
A couple of high level questions:

 1) This requires a statsd daemon to be running correct? there is no way for me to simply hit an API and collect the metrics routinely from a service?
 2) Does statsd have to be running on the box? Can it be running remotely? If so, is that documented somewhere?
 3) Does this introduce another port that has to be open?

lzap:
> This requires a statsd daemon to be running correct? there is no way for me to simply hit an API and collect the metrics routinely from a service?

Currently there is no API, I was thinking about this but without aggregating the values the API would generate useless results. And I would like to keep this simple, statsd is here to do the math, not sure if we should be doing that ourselves.

The patch does not require it, it can work in "logger" only mode or even "null" mode, which is the default.

> Does statsd have to be running on the box? Can it be running remotely? If so, is that documented somewhere?

Technically possible, it's just UDP, but not the recommended setup. You want to give the numbers out quickly. Documentation will be part of the final patch indeed. My thinking is to provide a simple framework and documentation about how to set things up. My goal is to provide decent (local only) telemetry with as little configuration changes and running components as possible. I am thinking just with collectd with statsd, cpu, load, memory and perhaps rrd plugins as the documented way of doing things, not yet done. I already have a prototype of plugin (https://github.com/lzap/foreman_colly) that shows graphs on Foreman dashboard (live data), we could provide ad-hoc easy-to-use telemetry.

> Does this introduce another port that has to be open?

Only 8125 on localhost, I don't think we should be sending anything via network by default.

I should say that we have https://github.com/theforeman/foreman_graphite/blob/master/lib/foreman_graphite/engine.rb already but that's graphite - full blown web application and backend. I don't aim for that although these telemetry details can be (quite easily) integrated with graphite and similar apps.

config/settings.yaml.example
lzap: It does, it is definitely not pretty, the feature was just there (in the dependency), so why not to allow it for ad-hoc poor man telemetry.

lzap:
Math is not hard, if you think the API would be useful, I can rip off https://github.com/fetep/ruby-statsdserver/blob/master/lib/statsdserver/math.rb

lzap:
More I think about our own API more I like it. I can provide additional setting called "internal" where we can keep last 10 or 100 measurements and calculate avg/means/min/max on each telemetry API request. This could be just enough for ad-hoc measurements. And if we do this a runtime option, users could easily turn telemetry on, provide results and turn it off.

But one thing is that multiple passenger processes will not share the results, we'd need to store the results in database or cache perhaps. That could complicate holding history of N measurements, we could do just last value and min/max.

lzap:
Other idea is to integrate with Performance Co-Pilot (http://pcp.io/) which ships with RHEL and Debian distros as well. It's performance collecting tool, very easy to setup, zero configuration, just one daemon (and second one if data should be archived). The integration is not that straightforward as with statsd tho.

config/initializers/telemetry.rb
ShimShtein: I would leave this metric to more dedicated tools, like [stackprof](https://github.com/tmm1/stackprof) or other allocation profiling tools.

IMHO this info is relevant only for performance analysis, and in that case you will probably need more info, like the allocation stacks e.t.c.

lzap: Since Rails provides this already and it's a simple counter, I don't think we should give up on this one. The tool you mentioned needs special configuration and restart while telemetry will be shipped by default and users will be able to enable it just by clicking in Administer - Settings.

config/settings.yaml.example
akofink: Should this be? :

```rb
:loggers:
  :telemetry:
    :enabled: true
```

Also, here's some example output: https://gist.github.com/akofink/142c2dda7412b4fca8176fe5aa6d2c7e

lzap: Yes that's correct.

lzap:
Ok I got another idea of storing the telemetry data in archive files, my prototype works fine. I will incorporate this in this patch: https://github.com/Shopify/statsd-instrument/pull/101

lzap:
Closing in fav of https://github.com/theforeman/foreman/pull/5096

PR #4337, opened 2017-02-26
fixes #18676 - don't ignore notification expiry

ohadlevy:
This also adds the ability to mass refresh expiry dates.

### DIFF
--- a/app/controllers/notification_recipients_controller.rb
+++ b/app/controllers/notification_recipients_controller.rb
@@ -6,7 +6,7 @@ class NotificationRecipientsController < Api::V2::BaseController
 
   def index
     @notifications = NotificationRecipient.
-      where(:user_id => User.current.id).
+      where(:user_id => User.current.id, :notification_id => Notification.active).
       order(:created_at).
       eager_load(:notification, :notification_blueprint)
 
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -21,11 +21,11 @@ class Notification < ActiveRecord::Base
     :in => [AUDIENCE_USER, AUDIENCE_GROUP, AUDIENCE_TAXONOMY,
             AUDIENCE_GLOBAL, AUDIENCE_ADMIN]
   }, :presence => true
-  before_create :calculate_expiry, :set_notification_recipients,
+  before_create :set_expiry, :set_notification_recipients,
     :set_default_initiator
 
-  scope :active, -> { where('expired_at >= :now', {:now => Time.now.utc}) }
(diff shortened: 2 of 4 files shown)


app/models/notification.rb
```diff
@@ -48,8 +48,8 @@ def subscriber_ids
 
   private
 
-  def calculate_expiry
-    self.expired_at = Time.now.utc + notification_blueprint.expires_in
+  def set_expiry
```
dLobatog: A comment explaining that you want this value to be saved only once, when the Notification is created, would be useful. Otherwise I think it's going to lead people unfamiliar with the code to wonder why it's not `delegate :expired_at, :to => :notification_blueprint` instead.

Since it has the potential of override an existing attribute, I think it would be good to warn by naming it `set_expiry!`

app/controllers/notification_recipients_controller.rb
```diff
@@ -8,6 +8,7 @@ def index
     @notifications = NotificationRecipient.
       where(:user_id => User.current.id).
       order(:created_at).
+      joins(:notification).merge(Notification.active).
```
dLobatog: I think this can do without the `join`, by adding it to the `where`

`where(:user_id => User.current.id, :notification_id => Notification.active)`

app/models/notification_blueprint.rb
```diff
@@ -18,6 +18,14 @@ class NotificationBlueprint < ActiveRecord::Base
   validates :expires_in, :numericality => {:greater_than => 0}
   before_validation :set_default_expiry
 
+  def mass_update_expiry
```
dLobatog, line 21: I'd rather not add unused code if we can avoid it, is it used anywhere?

```
blueprint.notifications.map(&:set_expiry!) // is clear enough IMO to not need a method
```

dLobatog (changes requested):
@ohadlevy Left a few minor comments, particularly it took me a bit of time to understand why `expired_at` was not using delegation. Overall I agree it's a good change, thanks!

app/models/notification_blueprint.rb
ohadlevy: its used in the discovery notification pr - see https://github.com/theforeman/foreman_discovery/issues/328

ohadlevy: also, map will create a rails ar object, the approach is mass update on the db level.

ohadlevy:
@dLobatog:
1. removed the join to inner select
2. added a comment.
thanks for the review!

dLobatog (approved):

dLobatog:
Thanks @ohadlevy !

PR #4338, opened 2017-02-26
Fixes #16043 - add 'select all hosts' option into hosts page

amirfefer:
`Select all items` checkbox in hosts page selects only the hosts within the page.  when the user wish to select all hosts for a bulk action, he has to select all hosts page after page.

multiple_selection1

multiple_selection

ezgif com-resize

### DIFF
--- a/app/assets/javascripts/host_checkbox.js
+++ b/app/assets/javascripts/host_checkbox.js
@@ -4,11 +4,17 @@ $.foremanSelectedHosts = readFromCookie();
 
 // triggered by a host checkbox change
 function hostChecked(box) {
+  var multiple_alert = $("#multiple-alert");
   var cid = parseInt(box.id.replace("host_ids_", ""));
   if (box.checked)
     addHostId(cid);
-  else
+  else{
     rmHostId(cid);
+    if (multiple_alert.length){
+      multiple_alert.hide('slow');
+      multiple_alert.data('multiple', false)
+    }
+  }
   $.cookie($.cookieName, JSON.stringify($.foremanSelectedHosts), { secure: location.protocol === 'https:' });
   toggle_actions();
   update_counter();
@@ -89,14 +95,41 @@ function cleanHostsSelection() {
   return false;
 }
 
+function multiple_selection() {
+  var total = $("#pagination").data("count");
+  var alert_text = Jed.sprintf(__("All <b> %d </b> hosts are selected. "), total);
+  var undo_text = __("Undo selection");
+  var multiple_alert = $("#multiple-alert");
+  multiple_alert.find(".text").html(alert_text + '<a href="#" onclick="undo_multiple_selection();">' + undo_text + '</a>');
+  multiple_alert.data('multiple', true)
+}
+
+function undo_multiple_selection() {
+  var pagination = pagination_metadata();
+  var alert_text = Jed.sprintf(__("All %s hosts on this page are selected. "), pagination.per_page);
+  var select_text = Jed.sprintf(__("Select all<b> %s </b> hosts"), pagination.total);
+  var multiple_alert = $("#multiple-alert");
+  multiple_alert.find(".text").html( alert_text + '<a href="#" onclick="multiple_selection();">' + select_text + '</a>');
+  multiple_alert.data('multiple', false)
+}
+
 function toggleCheck() {
+  var pagination = pagination_metadata();
+  var multiple_alert = $("#multiple-alert");
   var checked = $("#check_all").is(':checked');
   $('.host_select_boxes').each(function(index, box) {
     box.checked = checked;
     hostChecked(box);
   });
-  if(!checked)
+  if(checked && (pagination.per_page - pagination.total < 0) ) {
+    multiple_alert.show('slow');
+    multiple_alert.data('multiple', true);
+  }
+  else if (!checked) {
+    multiple_alert.hide('slow');
+    multiple_alert.data('multiple', false);
     cleanHostsSelection();
+  }
   return false;
 }
 
@@ -118,20 +151,33 @@ $(function() {
 function submit_modal_form() {
   if (!$('#keep_selected').is(':checked'))
     removeForemanHostsCookie();
+  if ($("#multiple-alert").data('multiple')){
+    var query = $("<input>")
+            .attr("type", "hidden")
+            .attr("name", "search").val($("#search").val());
+    $("#confirmation-modal form").append(query);
+  }
   $("#confirmation-modal form").submit();
   $('#confirmation-modal').modal('hide');
 }
 
 function build_modal(element, url) {
+  var is_multiple = $("#multiple-alert");
+  if (is_multiple.data('multiple'))
+    var data = {search: $("#search").val()};
+  else
+    var data = {host_ids: $.foremanSelectedHosts};
   var title = $(element).attr('data-dialog-title');
   $('#confirmation-modal .modal-header h4').text(title);
   $('#confirmation-modal .modal-body').empty()
     .append("<div class='modal-spinner spinner spinner-lg'></div>");
   $('#confirmation-modal').modal();
-  $("#confirmation-modal .modal-body").load(url + " #content",{host_ids: $.foremanSelectedHosts},
+  $("#confirmation-modal .modal-body").load(url + " #content", data,
       function(response, status, xhr) {
         $("#loading").hide();
         $('#submit_multiple').val('');
+        if (is_multiple.data('multiple'))
+          $("#multiple-modal-alert").show();
         var b = $("#confirmation-modal .btn-primary");
         if ($(response).find('#content form select').length > 0)
           b.addClass("disabled").attr("disabled", true);
@@ -146,12 +192,17 @@ function build_redirect(url) {
   window.location.replace(url);
 }
 
+function pagination_metadata() {
+  var pagination = $("#pagination");
+  var total = pagination.data("count");
+  var per_page = pagination.data("per-page");
+  return { total: total, per_page: per_page }
+}
+
 function update_counter() {
   var item = $("#check_all");
-  if ($.foremanSelectedHosts) {
+  if ($.foremanSelectedHosts)
     $(".select_count").text($.foremanSelectedHosts.length);
-    item.attr("checked", $.foremanSelectedHosts.length > 0 );
-  }
   var title = "";
   if (item.attr("checked"))
     title = $.foremanSelectedHosts.length + " - " + item.attr("uncheck-title");
--- a/app/controllers/hosts_controller.rb
+++ b/app/controllers/hosts_controller.rb
@@ -41,6 +41,7 @@ class HostsController < ApplicationController
   before_action :validate_power_action, :only => :update_multiple_power_state
 
   helper :hosts, :reports, :interfaces
+  helper_method :multiple_with_filter?
 
   def index(title = nil)
     begin
@@ -751,6 +752,10 @@ class HostsController < ApplicationController
     @host
   end
 
+  def multiple_with_filter?
+    params.key?(:search)
+  end
+
   def load_vars_for_ajax
     return unless @host
 
@@ -764,10 +769,11 @@ class HostsController < ApplicationController
 
   def find_multiple
   # Lets search by name or id and make sure one of them exists first
-    if params[:host_names].present? || params[:host_ids].present?
-      @hosts = resource_base.where("hosts.id IN (?) or hosts.name IN (?)", params[:host_ids], params[:host_names])
+    if params.key?(:host_names) || params.key?(:host_ids) || multiple_with_filter?
+      @hosts = resource_base.search_for(params[:search]) if multiple_with_filter?
+      @hosts ||= resource_base.where("hosts.id IN (?) or hosts.name IN (?)", params[:host_ids], params[:host_names])
       if @hosts.empty?
-        error _('No hosts were found with that id or name')
+        error _('No hosts were found with that id, name or query filter')
         redirect_to(hosts_path)
         return false
       end
--- a/app/helpers/hosts_and_hostgroups_helper.rb
+++ b/app/helpers/hosts_and_hostgroups_helper.rb
@@ -42,4 +42,11 @@ module HostsAndHostgroupsHelper
                 { :help_inline   => :indicator }
             ).html_safe
   end
+
+  def multiple_filter(hosts)
+    return unless multiple_with_filter?
+    no_filter   = _("Reminder: <strong> All #{hosts.size} hosts are selected </strong>").html_safe
+    with_filter = _("Reminder: <strong> All #{hosts.size} hosts are selected </strong> for query filter #{h(params[:search])}").html_safe
+    params[:search].blank? ? no_filter : with_filter
+  end
 end
--- a/app/helpers/layout_helper.rb
+++ b/app/helpers/layout_helper.rb
@@ -81,7 +81,8 @@ module LayoutHelper
   end
 
   def will_paginate_with_info(collection = nil, options = {})
-    content_tag(:div, :id => "pagination", :class => "row") do
+    content_tag(:div, :id => "pagination", :class => "row",
+                :data => ({'count' => collection.total_entries, 'per-page' => per_page(collection)})) do
       page_entries_info(collection, options) +
         will_paginate(collection, options)
     end
@@ -186,6 +187,10 @@ module LayoutHelper
 
   def render_if_partial_exists(path, f)
     render path, :f => f if lookup_context.exists?(path, [], true)
+  end
+
+  def per_page(collection)
+    [Setting[:entries_per_page], collection.total_entries].min
   end
 
   private
--- a/app/views/hosts/_selected_hosts.html.erb
+++ b/app/views/hosts/_selected_hosts.html.erb
@@ -1,6 +1,9 @@
 <div class="row">
-  <table class="<%= table_css_classes %>">
-    <thead>
+  <%= alert(:class => 'alert-info hide', :id => 'multiple-modal-alert', :close => true, :header => '',
+            :text => multiple_filter(hosts)) %>
+  <% unless multiple_with_filter? %>
+    <table class="<%= table_css_classes %>">
+      <thead>
       <tr>
         <th>
           <%= _("Name") %>
@@ -14,8 +17,8 @@
         <%= raw("<th>" + _("Location") + "</th>") if SETTINGS[:locations_enabled]%>
         <%= raw("<th>" + _("Organization") + "</th>") if SETTINGS[:organizations_enabled]%>
       </tr>
-    </thead>
-    <tbody>
+      </thead>
+      <tbody>
       <% associations = [:hostgroup, :environment] %>
       <% associations << :organization if SETTINGS[:organizations_enabled]  %>
       <% associations << :location     if SETTINGS[:locations_enabled]  %>
@@ -33,8 +36,9 @@
           <% end %>
         </tr>
       <% end %>
-    </tbody>
-  </table>
+      </tbody>
+    </table>
   <%= check_box_tag "keep_selected", "", false, :title => _("Keep selected hosts for a future action") %>
   <%= _('Keep selected hosts for a future action') %><br/><br/>
+  <% end %>
 </div>
--- a/app/views/hosts/index.html.erb
+++ b/app/views/hosts/index.html.erb
@@ -1,2 +1,5 @@
 <% title_actions multiple_actions_select, csv_link, button_group(new_link(_("Create Host"))) %>
+<%= alert(:class => 'alert-info hide', :id => 'multiple-alert', :close => false, :header => '',
+          :text => _("All #{per_page(@hosts)} hosts on this page are selected. ").html_safe +
+                   link_to_function((_("Select all <b> #{@hosts.total_entries.to_s} </b> hosts")).html_safe, "multiple_selection()")) %>
 <%= render 'list', :hosts => @hosts, :header => @title || _("Hosts")  %>
--- a/test/controllers/hosts_controller_test.rb
+++ b/test/controllers/hosts_controller_test.rb
@@ -374,7 +374,7 @@ class HostsControllerTest < ActionController::TestCase
     post :update_multiple_hostgroup, {:host_ids => [-1], :host_names => ["no.such.host"]}, set_session_user
 
     assert_redirected_to hosts_url
-    assert_equal "No hosts were found with that id or name", flash[:error]
+    assert_equal "No hosts were found with that id, name or query filter", flash[:error]
   end
 
   test 'multiple hostgroup change by host ids' do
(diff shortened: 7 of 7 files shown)


app/views/hosts/_selected_hosts.html.erb
```diff
@@ -1,4 +1,6 @@
 <div class="row">
+  <%= alert(:class => 'alert-info hide', :close => true, :header => '',
```
ohadlevy: maybe this should be warning? @rohoover ?

Rohoover: @ohadlevy For which part? The selection of all items or that all items are changing owners?

ohadlevy: I'm looking for a way to say that you are seeing only the first 20 but you actually selected all of them, so if you take an action against them its the entire list not just he one you see (same workflow if you select all of your emails on gmail).

app/helpers/layout_helper.rb
```diff
@@ -81,7 +81,7 @@ def page_entries_info(collection, options = {})
   end
 
   def will_paginate_with_info(collection = nil, options = {})
-    content_tag(:div, :id => "pagination", :class => "row") do
+    content_tag(:div, :id => "pagination", :class => "row", 'data-meta-count' => collection.count) do
```
tbrisker: Use `collection.total_entries` which caches to total count

app/controllers/hosts_controller.rb
```diff
@@ -750,8 +750,9 @@ def load_vars_for_ajax
 
   def find_multiple
   # Lets search by name or id and make sure one of them exists first
-    if params[:host_names].present? || params[:host_ids].present?
-      @hosts = resource_base.where("hosts.id IN (?) or hosts.name IN (?)", params[:host_ids], params[:host_names])
+    if params[:host_names].present? || params[:host_ids].present? || params.key?(:search)
```
tbrisker: mind turning all of these to `.key?`?

app/assets/javascripts/host_checkbox.js.erb
```diff
@@ -115,12 +146,26 @@ $(function() {
...
+  }
+
   $("#confirmation-modal form").submit();
   $('#confirmation-modal').modal('hide');
 }
 
 function build_modal(element, url) {
-  var url = url + "?" + $.param({host_ids: $.foremanSelectedHosts});
+  if ($("#multiple-chosen").data('multiple')) {
+    var query = $("<input>")
```
tbrisker: why do you need to create an input here if the value is encoded directly to the query?

app/assets/javascripts/host_checkbox.js.erb
```diff
@@ -115,12 +146,26 @@ $(function() {
 function submit_modal_form() {
   if (!$('#keep_selected').is(':checked'))
     removeForemanHostsCookie();
+  if ($("#multiple-chosen").data('multiple')){
+    var query = $("<input>")
+            .attr("type", "hidden")
+            .attr("name", "search").val($("#search").val());
+    $("#confirmation-modal form").append($(query));
```
tbrisker: no need to wrap `query` here, it is already a jquery object

app/assets/javascripts/host_checkbox.js.erb
```diff
@@ -7,8 +7,13 @@ function hostChecked(box) {
   var cid = parseInt(box.id.replace("host_ids_", ""));
   if (box.checked)
     addHostId(cid);
-  else
+  else{
     rmHostId(cid);
+    if ($(".alert").length){
```
tbrisker: might this be a different alert?

tbrisker (changes requested):
Added a few comments inline. Seeing the amount of changes to host_checkbox.js.erb, I wonder if it makes sense to just rewrite it in ES6? afaict the only reason for it to be an .erb is for the path to the spinner, which I think can be done with just a class anyways.

app/views/hosts/_selected_hosts.html.erb
Rohoover: @ohadlevy 

I actually like the way it's address in the gif. I don't think it should be a warning as nothing "wrong" happened.

amirfefer:
thanks @tbrisker for your review, I've followed your comments.

ohadlevy:
@tbrisker I would favor to move this to ES6 in a separate commit, as that most likely will invlove the entire file vs just this particular feature?

also the js.erb file has been fixed in #4349

tbrisker:
In general this works well, however there is an issue when selecting a very large amount of hosts - loading the modal with the host list can take several minutes as it isn't paginated. 
The table in the modal should be paginated or only display the matching search when "select all" is chosen. Also it seems to join tables that aren't needed for the fields it displays and can be removed to speed it up.

amirfefer:
@Rohoover: What do you think is more suitable for the modal dialog alert?

**Info alert:**
info_alert
**Warning alert:**
warning_alert

@tbrisker: Thanks, I think it better to display just the alert with hosts quantity and filter query, as you can see at the screenshots above. I've fixed the join issue.

Rohoover:
Use the grey box if the action isn't particular serious.

If it has strong consequences use the warning (orange) box.

Remove the chevrons in both. It would be nice to have (but not required) if the messaging would read" Reminder: **All 10 hosts are selected** for query filter architecture=..." because the important part is not what query they searched but that they have them all selected.

amirfefer:
@Rohoover Thanks :+1: . I've kept the grey box, and changed the bold string. otherwise the user would think he made something wrong.   The chevron mark notes that the string is translatable.

app/views/hosts/_selected_hosts.html.erb
```diff
@@ -1,6 +1,9 @@
 <div class="row">
-  <table class="<%= table_css_classes %>">
-    <thead>
+  <%= alert(:class => 'alert-info hide', :id => 'multiple-modal-alert', :close => true, :header => '',
+            :text => _("Reminder: <strong> All #{hosts.count.to_s} hosts are selected </strong> for query filter #{multiple_filter}").html_safe) %>
```
tbrisker: I think it makes sense to only show the "for query filter..." part if there is a filter, otherwise it looks odd:
image

app/helpers/layout_helper.rb
```diff
@@ -81,7 +81,7 @@ def page_entries_info(collection, options = {})
   end
 
   def will_paginate_with_info(collection = nil, options = {})
-    content_tag(:div, :id => "pagination", :class => "row") do
+    content_tag(:div, :id => "pagination", :class => "row", 'data-meta-count' => collection.total_entries) do
```
tbrisker: We also have a meta tag on the host list for number per page, it makes sense to me to use the same here as well using a `<meta name="totalResults" content="#{collection.total_entries}"/>` tag.

app/views/hosts/index.html.erb
```diff
@@ -1,2 +1,4 @@
 <% title_actions multiple_actions_select, button_group(new_link(_("Create Host"))) %>
+<%= alert(:class => 'alert-info hide', :id => 'multiple-alert', :close => false, :header => '',
+          :text => _("All #{Setting[:entries_per_page]} hosts on this page are selected. ").html_safe + link_to_function((_("Select all <b> #{@hosts.count.to_s} </b> hosts")).html_safe, "multiple_selection()")) %>
```
tbrisker: This should also use the cached `@hosts.total_entries` instead of `@hosts.count`.

app/views/hosts/_selected_hosts.html.erb
tbrisker: The filter should also be escaped to prevent XSS :
image

app/views/hosts/_selected_hosts.html.erb
```diff
@@ -33,8 +36,9 @@
...
         </tr>
       <% end %>
-    </tbody>
-  </table>
-  <%= check_box_tag "keep_selected", "", false, :title => _("Keep selected hosts for a future action") %>
-  <%= _('Keep selected hosts for a future action') %><br/><br/>
+      </tbody>
+    </table>
+    <%= check_box_tag "keep_selected", "", false, :title => _("Keep selected hosts for a future action") %>
+    <%= _('Keep selected hosts for a future action') %><br/><br/>
```
tbrisker: shouldn't this be possible also when selecting all? (i.e. outside the if block)

app/controllers/hosts_controller.rb
```diff
@@ -750,8 +761,9 @@ def load_vars_for_ajax
 
   def find_multiple
   # Lets search by name or id and make sure one of them exists first
-    if params[:host_names].present? || params[:host_ids].present?
-      @hosts = resource_base.where("hosts.id IN (?) or hosts.name IN (?)", params[:host_ids], params[:host_names])
+    if params.key?(:host_names) || params.key?(:host_ids) || multiple_with_filter?
```
tbrisker, line 772: I wonder - do we actually ever call this method with a `host_names` parameter?

app/controllers/hosts_controller.rb
```diff
@@ -737,6 +738,16 @@ def find_resource
...
 
+  def multiple_with_filter?
+    params.key?(:search)
+  end
+
+  def multiple_filter
+    if multiple_with_filter?
+      params[:search].blank? ? _("Empty") : params[:search]
+    end
+  end
```
tbrisker: I think this should go in the helper instead of the controller.

test/controllers/hosts_controller_test.rb
```diff
@@ -500,6 +500,28 @@ def setup_multiple_compute_resource
...
+  test "find multiple hosts by filter query" do
+    setup_user_and_host "edit"
+    post :update_multiple_owner, { :search => "",
+      :owner => { :id => users(:one).id_and_type}},
+      set_session_user.merge(:user => users(:admin).id)
+    as_admin do
+      assert_equal users(:one).id_and_type, @host1.reload.is_owned_by
+      assert_equal users(:one).id_and_type, @host2.reload.is_owned_by
+    end
+  end
```
tbrisker, line 512: Please add a test that has an actual search query and make sure only the matching hosts are selected.

app/assets/javascripts/host_checkbox.js
```diff
@@ -86,14 +91,40 @@ function cleanHostsSelection() {
...
+  var undo_text = __("Undo selection");
+  $(".text").html('<span id="multiple-chosen">' + alert_text + '<a href="#" onclick="undo_multiple_selection();">' + undo_text + '</a></span>');
+  $("#multiple-chosen").data('multiple', true)
+}
+
+function undo_multiple_selection(){
+  var count = $("#pagination").data("meta-count");
+  var per_page = $("meta[name='itemsPerPage']").attr("content");
+  var alert_text = Jed.sprintf(__("All %s hosts on this page are selected. "), per_page);
+  var select_text = Jed.sprintf(__("Select all") + "<b> %s </b>" + __("hosts"), count);
```
tbrisker: This won't translate correctly in all languages as the strings `Select all` and `hosts` will be translated separately and not as one sentence. Also above on line 96.

tbrisker (changes requested):
Getting there, a few mostly minor inline comments.

test/controllers/hosts_controller_test.rb
amirfefer: I've added two additional tests

app/controllers/hosts_controller.rb
amirfefer: I haven't found anything using `host names` in params...

app/controllers/hosts_controller.rb
amirfefer: `host_helper` is too large, hence I've extract this logic into `hosts_and_hostgroup_helper`

app/helpers/layout_helper.rb
amirfefer: I've added this particularly  to pagination div, so the total collection size would be collectable from any pagination page.

As you can see at the following screenshot, the total size is visable though hardly collectable. 
total
 The meta name `per page` exists only in hosts page.

app/views/hosts/_selected_hosts.html.erb
amirfefer: I've thought at the beginning that it useless due to cookie implementation , but I've managed not to expand this cookie :smile:

amirfefer:
@tbrisker I've fixed most of your inline comments, I've left some inline comments.

app/helpers/hosts_and_hostgroups_helper.rb
```diff
@@ -42,4 +42,12 @@ def realm_field(f, can_override = false, override = false)
                 { :help_inline   => :indicator }
             ).html_safe
   end
+
+  def multiple_filter(hosts)
+    if multiple_with_filter?
+      no_filter   = _("Reminder: <strong> All #{hosts.size} hosts are selected </strong>").html_safe
+      with_filter = _("Reminder: <strong> All #{hosts.size} hosts are selected </strong> for query filter").html_safe + h(params[:search])
```
amirfefer: In the helper file, I've couldn't use `total_entries` (no method exception),  instead I've used `size`.

app/helpers/layout_helper.rb
ohadlevy: @tbrisker @amirfefer meta attribute was remove at https://github.com/theforeman/foreman/commit/ee7002f8f2bffb3a589c28ea1b35460b6bdbf3c0#diff-7eb571b855131f643240b7b93ddfd44dL4

amirfefer: I've removed the use of meta attribute, instead I've expanded  `will_paginate_with_info` with `include_data_tag` flag.

ohadlevy:
@amirfefer please rebase.
@tbrisker what is left here? thanks

app/helpers/hosts_and_hostgroups_helper.rb
```diff
@@ -42,4 +42,11 @@ def realm_field(f, can_override = false, override = false)
                 { :help_inline   => :indicator }
             ).html_safe
   end
+
+  def multiple_filter(hosts)
+    return unless multiple_with_filter?
+    no_filter   = _("Reminder: <strong> All #{hosts.size} hosts are selected </strong>").html_safe
+    with_filter = _("Reminder: <strong> All #{hosts.size} hosts are selected </strong> for query filter").html_safe + h(params[:search])
```
tbrisker: nitpick: missing either a `: ` in the end of the sentence or placing the search inside the translated string (since if it's part of the sentence it might not be the end of it in all languages).

app/helpers/layout_helper.rb
```diff
@@ -81,7 +81,8 @@ def page_entries_info(collection, options = {})
   end
 
   def will_paginate_with_info(collection = nil, options = {})
-    content_tag(:div, :id => "pagination", :class => "row") do
+    content_tag(:div, :id => "pagination", :class => "row",
+                :data => ({'count' => collection.total_entries, 'per-page' => Setting[:entries_per_page]} if options[:include_data_tag])) do
```
tbrisker: Is there any reason not to have these always available for anything that needs them rather then add yet another option?

app/assets/javascripts/host_checkbox.js
```diff
@@ -86,14 +94,40 @@ function cleanHostsSelection() {
   return false;
 }
 
+function multiple_selection(){
+  var count = $("#pagination").data("count");
+  var alert_text = Jed.sprintf(__("All <b> %d </b> hosts are selected."), count);
+  var undo_text = __("Undo selection");
+  $(".text").html('<span id="multiple-chosen">' + alert_text + '<a href="#" onclick="undo_multiple_selection();">' + undo_text + '</a></span>');
```
tbrisker: nitpick: missing a space here.

tbrisker (changes requested):
Works well in general, a few minor inline comments. 
One thing I also noticed when testing this - when selecting all hosts on a page, then searching for something else, the "select all" isn't cleared leading to a confusing header- e.g. "All 20 hosts on this page are selected. Select all 4 hosts" when non of the hosts on the page are actually selected. This also appears to be present on develop, but is less obvious because there is no big header telling you how many hosts are selected (it's only listed at the bottom of the page).

app/helpers/layout_helper.rb
amirfefer: No special reason..

amirfefer:
@tbrisker Thanks, I've rebased and fixed above issues, including the search bug.

ohadlevy:
test failure unrelated.

tbrisker:
I'm afraid the resetting of selection on filter change still does not work as expected:
image
This is after I selected all 89 hosts in a different search, then searched for something else. Clicking on a multiple action from the action select button leads to loading the modal with the first 20 from the previous search. If the new search return a subset of the previous this is even more confusing - you expect only some hosts to be affected but then the modal contains all.
@Rohoover - correct me if I'm wrong, but to me it makes sense that a new search should clear any previously selected hosts, right?

Rohoover:
@tbrisker 

Yes, that would be my assumption as well.

tbrisker:
@amirfefer Also, if I select all hosts, then in the modal choose "keep hosts selected", then it reverts to only the first 20 instead of all matching the filter.

amirfefer:
@tbrisker IMO the cookie is the problem here.
I think it better to keep the cookie behavior as it was, which means only for save specific hosts selection.
I don't see any benefit for saving the user's choice in a cookie when he select a whole collection by a search filter, (because it simple and quick to re-use a whole collection selection - just  2 clicks -> `all_hosts` checkbox >> `select all x hosts`).
Therefore I've removed the `keep hosts selected` checkbox when a user select all hosts by query filter.

app/helpers/hosts_and_hostgroups_helper.rb
```diff
@@ -42,4 +42,15 @@ def realm_field(f, can_override = false, override = false)
...
+
+  def multiple_filter(hosts)
+    return unless multiple_with_filter?
+    no_filter   = _("Reminder: <strong> All #{hosts.size} hosts are selected </strong>").html_safe
+    with_filter = _("Reminder: <strong> All #{hosts.size} hosts are selected </strong> for query filter #{h(params[:search])}").html_safe
+    params[:search].blank? ? no_filter : with_filter
+  end
+
+  def per_page(collection)
+     collection.total_entries > Setting[:entries_per_page] ? Setting[:entries_per_page] : collection.total_entries
```
tbrisker: Shouldn't this be in layout helper?
also, how about `[Setting[:entries_per_page], collection.total_entries].min`?

app/assets/javascripts/host_checkbox.js
```diff
@@ -86,14 +91,40 @@ function cleanHostsSelection() {
...
+  var count = $("#pagination").data("count");
+  var alert_text = Jed.sprintf(__("All <b> %d </b> hosts are selected. "), count);
+  var undo_text = __("Undo selection");
+  $(".text").html('<span id="multiple-chosen">' + alert_text + '<a href="#" onclick="undo_multiple_selection();">' + undo_text + '</a></span>');
+  $("#multiple-chosen").data('multiple', true)
+}
+
+function undo_multiple_selection(){
+  var count = $("#pagination").data("count");
+  var per_page = $("#pagination").data("per-page");
```
tbrisker: Please cache `$("#pagination")` to save dom lookups

app/assets/javascripts/host_checkbox.js
```diff
@@ -86,14 +91,40 @@ function cleanHostsSelection() {
...
+  var alert_text = Jed.sprintf(__("All %s hosts on this page are selected. "), per_page);
+  var select_text = Jed.sprintf(__("Select all<b> %s </b> hosts"), count);
+  $(".text").html('<span id="multiple-chosen">' + alert_text + '<a href="#" onclick="multiple_selection();">' + select_text + '</a></span>');
```
tbrisker: Isn't `.text` a bit of a generic selector?

app/assets/javascripts/host_checkbox.js
```diff
@@ -86,14 +91,40 @@ function cleanHostsSelection() {
...
+
 function toggleCheck() {
   var checked = $("#check_all").is(':checked');
   $('.host_select_boxes').each(function(index, box) {
     box.checked = checked;
     hostChecked(box);
   });
-  if(!checked)
+  var per_page = $("#pagination").data("per-page");
+  var total_host = $("#pagination").data("count");
```
tbrisker: ditto

tbrisker (changes requested):
Here is a screenshot explaining the problem with not clearing the cookie on a new search:
screenshot from 2017-03-27 11-45-07
Note that the top shown "20 selected", but the bottom says "40 selected" which is the correct number. (to get to this situation, search for something, mark some hosts, search a different search and select all)

amirfefer:
@tbrisker :  From your screenshot It looks like the top number is presented correctly  while the bottom does not. 
This issue occurs due to a counter bug which exists in develop, as you can see in the following screenshot.
selected_counter

I've fixed the rest.

app/assets/javascripts/host_checkbox.js
```diff
@@ -56,6 +61,8 @@ function toggle_actions() {
 
 // setups checkbox values upon document load
 $(function() {
+  $.count = $("#pagination").data("count");
+  $.per_page = $("#pagination").data("per-page");
```
tbrisker: Why are you adding these to the jQuery object?

app/assets/javascripts/host_checkbox.js
```diff
@@ -7,8 +7,13 @@ function hostChecked(box) {
   var cid = parseInt(box.id.replace("host_ids_", ""));
   if (box.checked)
     addHostId(cid);
-  else
+  else{
     rmHostId(cid);
+    if ($("#multiple-alert").length){
+      $("#multiple-alert").hide('slow');
+      $("#multiple-alert").data('multiple', false)
```
tbrisker: `$("#multiple-alert")` should be cached to reduce lookups

app/assets/javascripts/host_checkbox.js
tbrisker: also please cache `$("#pagination")`

app/assets/javascripts/host_checkbox.js
```diff
@@ -86,14 +93,35 @@ function cleanHostsSelection() {
...
     hostChecked(box);
   });
-  if(!checked)
+  if(checked && ($.per_page - $.count < 0) ) {
+    $("#multiple-alert").show('slow');
+    $("#multiple-alert").data('multiple', true);
+  }
+  else if (!checked) {
+    $("#multiple-alert").hide('slow');
+    $("#multiple-alert").data('multiple', false);
```
tbrisker: please cache these too

tbrisker (changes requested):
a few more comments, please also note related tests seem to be failing intermittently.

amirfefer:
@tbrisker , thanks, I've added caching to reduce lookups.

tbrisker:
@amirfefer Following our talk, looks like you didn't push an updated version with correct caching for jquery objects?

amirfefer:
@tbrisker , You're right, now you can see caching for jquery objects

tbrisker:
Thanks @amirfefer ! I'd be happy to see all of this code rewritten in es6 or react, could you please open a refactor issue for that?

PR #4339, opened 2017-02-26
fixes #18681 - moves polymorphic subject into notification object

ohadlevy:


dLobatog (changes requested):
@ohadlevy Thanks - the reason it got implemented like this is because.. I did not understand what you wanted to do with the subject until we talked about it in person.

Can you add a test that shows the behavior? For example, you may have a 'blueprint.message = "Hello from subject #{notification.subject.name}"'. Then test that by creating a Notification using that NotificationBlueprint with 2 subjects (say, 2 hosts), the message the user gets is:

"Hello from subject host1.example.com"
"Hello from subject host2.example.com"

I'm not sure this is possible to do with the current code in this PR, as the message in the blueprint is static. Perhaps it can be static as "Hello from subject %s", then on the NotificationRecipient you return the message as `_("Hello from subject %s") % notification.subject`.

If that's not the purpose of subject could you explain in a bit more detail what do you want to use it for?

ohadlevy:
> If that's not the purpose of subject could you explain in a bit more detail what do you want to use it for?

You are correct, the subject has two use cases.

1. providing a customized messages - e.g. ```  :message: Host %{subject} has been provisioned.```
2. it defines the recipients, for example, we would send the message above only to host owners.

I'll work on implementing 1 as well, 2. is already implemented.

ohadlevy:
@dLobatog this is an initial implementation to add a notification that uses subject, I've implemented host built notification.

early feedback is much appreciated.

app/services/ui_notifications.rb
```diff
@@ -0,0 +1,31 @@
+# Namespace for Foreman UI notifications event handling
+module UiNotifications
```
dLobatog: nitpick, why not UINotifications? Lowercase letters ('i' in this case) shouldn't be part of an acronym

app/services/ui_notifications.rb
```diff
@@ -0,0 +1,31 @@
...
+  class Base
+    attr_reader :state, :subject
+
+    def initialize(subject, state)
+      raise(Foreman::Exception, 'must provide notification subject') if subject.nil?
```
dLobatog: Please translate the string and use `::Foreman::Exception.new(N_("Some error"))` 

Nitpick but the string first letter should be uppercase

app/services/ui_notifications.rb
```diff
@@ -0,0 +1,31 @@
...
+      @subject = subject
+      raise(Foreman::Exception, 'invalid state') unless allowed_states.include?(state)
```
dLobatog: https://github.com/theforeman/foreman/pull/4339/files#r103540123 , also if you want to validate, how about putting all validations at the beginning? That way we don't set `@subject = subject` for nothing in the case of an invalid `state`

app/services/ui_notifications.rb
```diff
@@ -0,0 +1,31 @@
...
+      @state = state
+    end
+
+    def deliver!
+      raise(Foreman::Exception, 'not Implemented')
```
dLobatog: https://github.com/theforeman/foreman/pull/4339/files#r103540123 , `Not implemented`

app/services/ui_notifications.rb
```diff
@@ -0,0 +1,31 @@
...
+    end
+
+    protected
+
+    def user
+      User.anonymous_api_admin
+    end
+
+    def allowed_states
```
dLobatog: Please leave a comment explaining why `[]` is here, it's basically forcing the exception in the initializer

app/services/ui_notifications.rb
```diff
@@ -0,0 +1,31 @@
...
+      @state = state
+    end
+
+    def deliver!
+      raise(Foreman::Exception, 'not Implemented')
+    end
+
+    protected
+
+    def user
```
dLobatog: Just curious, why not `User.anonymous_admin`

app/services/ui_notifications.rb
```diff
@@ -0,0 +1,31 @@
...
+      User.anonymous_api_admin
+    end
+
+    def allowed_states
+      []
+    end
+
+    def logger
```
dLobatog: `delegate :logger, :to => :Rails`

app/services/ui_notifications/hosts/provision.rb
```diff
@@ -0,0 +1,51 @@
...
+        case state
+        when :build_completed
+          host_build
+        end
+      end
+
+      private
+
+      def host_build
+        logger.debug('Host finished provisioning (build is done) - checking for notficiations')
```
dLobatog: typo, `notifications`. Also "build is done" reads a bit weird, "build completed", or "Host built completed (provisioning has finished). Checking for notifications" 

IMO this should be `logger.info`, as it is already for importing facts or reports

Since we have the subject, the message is more traceable by using `Host #{subject}` in the message.

app/services/ui_notifications/hosts/provision.rb
```diff
@@ -0,0 +1,51 @@
...
+        blueprint = host_build_blueprint
+
+        if blueprint.notifications.where(:subject => subject).empty?
+          logger.debug('creating new UI notification host_build_completed')
```
dLobatog: Ditto, if we have the host name at this point, it'd be better to add it in the message for traceability

app/services/ui_notifications/hosts/provision.rb
```diff
@@ -0,0 +1,51 @@
...
+          ::Notification.create!(
+            initiator: User.find(subject.owner_id),
+            subject: subject,
+            # TODO fix usrgroup
+            audience: audience,
+            notification_blueprint: blueprint
+          )
+        else
+          logger.debug('UI notification already exists, updating timestamps.')
```
dLobatog: Ditto, if we have the host name at this point, it'd be better to add it in the message for traceability

app/services/ui_notifications/hosts/provision.rb
```diff
@@ -0,0 +1,51 @@
...
+      end
+
+      def allowed_states
+        [:build_completed]
+      end
+
+      def audience
+        if subject.owner_type == 'User'
+          ::Notification::AUDIENCE_USER
+        else
```
dLobatog: If the host has no owner (very common case), there is no notification? Taxonomy could fit in that case

app/services/ui_notifications/hosts/provision.rb
```diff
@@ -0,0 +1,51 @@
...
+      private
+
+      def host_build
+        logger.debug('Host finished provisioning (build is done) - checking for notficiations')
+        blueprint = host_build_blueprint
+
+        if blueprint.notifications.where(:subject => subject).empty?
+          logger.debug('creating new UI notification host_build_completed')
+          ::Notification.create!(
+            initiator: User.find(subject.owner_id),
```
dLobatog: `initiator` has to be present, but `subject.owner_id` doesn't... so I don't believe it's a good choice here

db/seeds.d/17-notification_blueprints.rb
```diff
@@ -0,0 +1,30 @@
...
+      NotificationBlueprint.create!(attributes)
+    end
+  end
+end
+
+blueprints = [
+  {
+    group: _('Provisioning'),
+    name: 'host_build_completed',
+    message: _('%{subject} has been provisioned successfully'),
```
dLobatog: This is missing `{ :actions => { :links => hosts_path(subject) } ` (probably you'll have to wrap `hosts_path` in a lambda or something though)

app/services/ui_notifications/hosts.rb
```diff
@@ -0,0 +1,6 @@
+module UiNotifications
+  module Hosts
+    class Base < UiNotifications::Base
```
dLobatog: Dead code?

app/services/ui_notifications.rb
dLobatog: Actually.. is this used anywhere? It can probably be removed

dLobatog (changes requested):
Thanks @ohadlevy - I tried it and found a few of the issues you can see in the comments, particularly if owner_id is not set. 

To be fair, I think this is a *lot* of YAGNI ceremony for a simple action: sending a notification. I would rather see a:

``` 
class HostBuiltNotificationSender < ApplicationJob
end
```

implementing the `host_build` in `ui_notifications/hosts/provision.rb`. Then in `app/models/host/managed.rb` you just call `HostBuiltNotificationSender.perform_later(self)`, and dynflow picks it up whenever it wants instead of making the action synchronous.

I'm not really opposed to the idea of the UINotifications service but to be honest, I want to keep notifications as simple as possible so people use them, and this (while not complicating notifications on its own) - to a Notifications newbie (most of the people these days) might make them think.. "oh it's not just define a method and call it, it's way more involved"

I don't mind merging if you think it makes sense, but I think in most cases I will resort to ActiveJob and asynchronous notification creation, and keeping stuff contained in its own Job class.

ohadlevy:
@dLobatog I believe I've addressed all of your concerns.

I still need to wrap up additional testing the url resolver service.

app/services/ui_notifications/url_resolver.rb
```diff
@@ -0,0 +1,52 @@
...
+  class URLResolver
+    # needed in order to resolve rails urls for notifications
+    include Rails.application.routes.url_helpers
+    def initialize subject, message, actions
+      @subject = subject
+      @raw_message = message
+      @raw_actions = actions
+    end
+
+    def message
```
dLobatog: Not sure what this or `parse_string` have to do with URLs?

app/services/ui_notifications/hosts/provision.rb
```diff
@@ -0,0 +1,43 @@
...
+        else
+          logger.debug('UI notification already exists, updating timestamps.')
+          blueprint.notifications.
+            where(subject: subject).
+            update_all(expired_at: blueprint.expired_at)
+        end
+      end
+
+      def destroyed
+        # IMHO no point in keeping notifications for deleted hosts, as it may
```
dLobatog: nitpick but 'imho' doesn't look like it belongs in a comment

app/models/host/managed.rb
```diff
@@ -1016,4 +1020,8 @@ def send_built_notification
   rescue SocketError, Net::SMTPError => e
     Foreman::Logging.exception("Host has been created. Failed to send email", e)
   end
+
+  def remove_ui_notifications
```
dLobatog: Is there any way we can have these changes in a concern `Hostext::Notifications` or similar? I'm just worried we keep on adding things to this object. Plus, `Host::Discovered` uses `Hostext` concerns when it needs behavior from `Host::Managed.`

app/services/ui_notifications.rb
```diff
@@ -0,0 +1,37 @@
...
+module UINotifications
+  class Base
+    attr_reader :subject, :event
+
+    def self.deliver!(subject, event)
+      new(subject, event).deliver!
+    end
+
+    def initialize(subject, event)
+      raise(Foreman::Exception, 'must provide notification subject') if subject.nil?
```
dLobatog: Exceptions text should be translated

app/services/ui_notifications.rb
```diff
@@ -0,0 +1,37 @@
...
+      raise(Foreman::Exception, 'invalid event') unless allowed_events.include?(event)
```
dLobatog: Missing translation, also I think it'd be useful to include the event that was passed in the text. `invalid_event %s` or something

app/services/ui_notifications.rb
```diff
@@ -0,0 +1,37 @@
...
+      @subject = subject
+      @event = event
+    end
+
+    def deliver!
+      logger.debug("Notification event: #{event} - checking for notifications")
+      send(event) if respond_to?(event, true)
```
dLobatog: Nitpick, `public_send`, or even better, the whole line could just be `try(event)`, correct?

app/services/ui_notifications/hosts.rb
```diff
@@ -0,0 +1,23 @@
+module UINotifications
+  module Hosts
+    class Base < UINotifications::Base
+      def deliver!
+        if audience.nil?
+          raise(Foreman::Exception, "Invalid owner for #{subject}, unable to send notifications")
```
dLobatog: Please extract the string for translations

app/services/ui_notifications/hosts/provision.rb
```diff
@@ -0,0 +1,43 @@
...
+        # contain links to non existing actions.
+        logger.debug("Removing all notifications for host: #{subject}")
+        Notification.where(subject: subject).delete_all
```
dLobatog: `.destroy_all`, if I'm not wrong `delete_all` is just going to delete stuff in the Notifications table (not NotificationRecipients)

app/services/ui_notifications/hosts/provision.rb
```diff
@@ -0,0 +1,43 @@
...
+      end
+
+      def initiator
+        return subject.owner if subject.owner.is_a?(User)
```
dLobatog: This is not acknowledging Hosts without owner

app/services/ui_notifications/hosts.rb
```diff
@@ -0,0 +1,23 @@
...
+
+      protected
+
+      def audience
+        case subject.owner_type
+        when 'User'
+          ::Notification::AUDIENCE_USER
+        when 'Usergroup'
+          ::Notification::AUDIENCE_GROUP
+        end
```
dLobatog, line 22:
```
else
  initiator 
end
```

or something like that - I guess you want the notification even if you don't set the owner field

dLobatog (changes requested):
@ohadlevy Thanks for the changes, please see the comments. 

@domcleal the exceptions here are not to be shown to the user.. we use to translate exceptions everywhere but maybe it doesn't make that much sense in this case?

app/services/ui_notifications/url_resolver.rb
```diff
@@ -0,0 +1,52 @@
...
+    private
+
+    attr_reader :subject, :raw_message, :raw_actions
+    def parse_string string
+      string % { :subject => subject }
+    end
+
+    def parse_link path_method, title
+      {
+        href: public_send(path_method, subject),
```
ShimShtein: It will fail for index paths like `hosts_path`, we can't pass the subject to such paths.

``` ruby
hosts_path(Host.first)
#=> "/hosts.tommy-havier.mydomain.com"
```

app/services/ui_notifications.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    protected
+
+    # Defaults to anonymous api admin, override in subclasses as needed.
+    def initiator
+      User.anonymous_api_admin
+    end
+
+    def allowed_events
```
ShimShtein: I would like to see a DSL for extending this list instead of overriding the method. Good example would be `Host` events that could be extended by Katello to add some content specific events.

app/services/ui_notifications/hosts.rb
ohadlevy: do you expect to see this anywhere? i can change it to a log entry instead of raising, I was assuming this will be used by ActiveJob so it can halt the job, you are correct that currently with active record call back it might disable the action.

app/services/ui_notifications/hosts.rb
ohadlevy: no, who would i notify?

ohadlevy: also, the handling is in the deliver! method.

ohadlevy:
@dLobatog would appericiate another review - thanks.

for exceptions, I believe i'll change it away from Foreman::Exception as they should be mostly useful for developers adding notifications vs users?

app/services/ui_notifications/hosts/provision.rb
```diff
@@ -0,0 +1,43 @@
+module UINotifications
+  module Hosts
+    class Provision < Base
+      private
+      def build_completed
+        blueprint = NotificationBlueprint.find_by(name: 'host_build_completed')
+
+        if blueprint.notifications.where(:subject => subject).empty?
```
ShimShtein: You could spare one query:

``` ruby
updated_count = blueprint.notifications.where(subject: subject).update_all(expired_at: blueprint.expired_at)

if updated_count == 0
  ::Notification.create!(...)
else
  log
end
```

app/models/host/managed.rb
ohadlevy: yes, but can we do it in a separate pr? this one is large enough? :)

ohadlevy:
@ShimShtein @dLobatog I think I've addressed all of your concerns?

app/services/ui_notifications/hosts/build_completed.rb
```diff
@@ -0,0 +1,36 @@
+module UINotifications
+  module Hosts
+    class BuildCompleted < Base
+      private
+
+      def create
+        if blueprint.notifications.where(:subject => subject).empty?
```
ShimShtein: Nitpick, but we can spare this query:

``` ruby
if update_notifications() == 0 #no old notifications
  add_notification
end
```

app/services/ui_notifications/hosts/build_completed.rb
```diff
@@ -0,0 +1,36 @@
...
+
+      def update_notifications
+        logger.debug('UI notification already exists, updating timestamps.')
+        blueprint.notifications.
+          where(subject: subject).
+          update_all(expired_at: blueprint.expired_at)
+      end
+
+      def blueprint
+        @blueprint ||= NotificationBlueprint.find_by(name: 'host_build_completed')
```
ShimShtein, line 26: We can cache this blueprint on class level - it will spare a query for each notification.

app/services/ui_notifications/hosts/destroy.rb
```diff
@@ -0,0 +1,29 @@
...
+        )
+      end
+
+      def delete_others
+        logger.debug("Removing all notifications for host: #{subject}")
+        Notification.where(subject: subject).destroy_all
+      end
+
+      def blueprint
+        @blueprint ||= NotificationBlueprint.find_by(name: 'host_destroyed')
```
ShimShtein, line 25: ditto

app/services/ui_notifications/url_resolver.rb
```diff
@@ -0,0 +1,61 @@
+module UINotifications
+  # interpolates blueprint strings and links for notifications
+  class URLResolver
+    # needed in order to resolve rails urls for notifications
+    include Rails.application.routes.url_helpers
+    def initialize subject, actions = nil
```
ShimShtein: parenthesis PLZ

app/services/ui_notifications/hosts.rb
ohadlevy: how about adding a new notification in that case to notify admins they have a host without an owner?

app/services/ui_notifications/hosts/build_completed.rb
ohadlevy: @ShimShtein but the blueprint is per class? each class will have its own?

app/services/ui_notifications/url_resolver.rb
ohadlevy: updated, even if rubocop does not enforce this.

ohadlevy:
@ShimShtein updated based on your comments.

ohadlevy:
adding a screenshot for the 3 new notifications:
1. host finished provisioning
2. host has no owner set
3. host is destroyed
notifications

app/models/host/managed.rb
```diff
@@ -50,7 +51,7 @@ def initialize(*args)
   end
 
   def build_hooks
-    return unless respond_to?(:old) && old && (build? != old.build?)
+    return if previous_changes['build'].nil?
```
dLobatog, line 54: nitpick, `blank`

app/models/notification.rb
```diff
@@ -61,4 +65,17 @@ def set_notification_recipients
...
   end
+
+  def set_custom_attributes
+    return unless notification_blueprint # let validation catch this.
+    self.actions = UINotifications::URLResolver.new(
+      subject,
+      notification_blueprint.actions
+    ).actions if notification_blueprint.actions.any?
+    # copy notification message in case we didn't create a custom one.
+    self.message ||= UINotifications::StringParser.new(
```
dLobatog, line 76: Nice separation of concerns

app/services/ui_notifications/hosts/destroy.rb
```diff
@@ -0,0 +1,29 @@
+module UINotifications
+  module Hosts
+    class Destroy < Base
+      private
+
+      def create
+        # I'm defaulting to deleting older notifications as it may
+        # contain links to non existing actions.
```
dLobatog, line 8: 👍

app/services/ui_notifications/hosts/missing_owner.rb
```diff
@@ -0,0 +1,29 @@
...
+      def update_notifications
+        blueprint.notifications.
+          where(subject: subject).
+          update_all(expired_at: blueprint.expired_at)
+      end
+
+      def add_notification
+        Notification.create!(
+          initiator: initiator,
+          audience: 'admin',
```
dLobatog: nitpick, `Notification::AUDIENCE_ADMIN` to be resilient to changes and use the frozen string

dLobatog (changes requested):
@ohadlevy Code-wise 👍 , I'm going to test this a bit more thoroughly and merge if it works. Thanks!

app/models/host/managed.rb
ohadlevy: this can be nil or an array, empty string / array is not possible for this method.

ohadlevy:
@dLobatog or others, do you have any idea why the tests are failing? I'm unsure why the hook is not firing / updating correctly.

dLobatog (approved):

ohadlevy:
@dLobatog I think I've addressed all failing tests. I've also added exception handling in the case notification action fails, I do not want to block the caller (e.g. think AR callbacks). once we introduce ActiveJob this might change, but regardless I think its good idea not to block the UI/Action triggering the notification regardless.

ohadlevy:
@dLobatog to make sure, the lines I've added are at https://github.com/theforeman/foreman/pull/4339/files#diff-59ce2b15fb9e1d3ecb10bf58cf982db8R8

dLobatog:
Thanks @ohadlevy ! Feel free to start using this in discovery, I'll do the same.

PR #4340, opened 2017-02-27
fixes #18691 - add as_deprecation_tracker to test environment

domcleal:
Causes test failures if any new deprecation warnings are triggered by
changes in Foreman. Plugin code is automatically whitelisted unless it
supplies its own ASDT whitelist configuration, so developers can opt-in.

Warnings are now hidden in production as they're only relevant in
development or test environments.

---

Try reverting 0804d85 to see the effect.

### DIFF
--- a/bundler.d/test.rb
+++ b/bundler.d/test.rb
@@ -15,4 +15,5 @@ group :test do
   gem 'test_after_commit', '>= 0.4', '< 2.0'
   gem 'shoulda-matchers', '~> 3.0'
   gem 'shoulda-context', '~> 1.2'
+  gem 'as_deprecation_tracker', '~> 1.4'
 end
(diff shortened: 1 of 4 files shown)


dLobatog:
Thanks @domcleal ! I'm surprised to only see 1 error in the whitelist. 

For plugin devs, you can run:

`AS_DEPRECATION_WHITELIST=~/plugin_path AS_DEPRECATION_RECORD=yes bin/rake test ` to create your whitelist.

domcleal:
More information added to http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Testing-for-deprecations for plugin developers.

ares:
This breaks tests for plugins using deface, see https://github.com/spree/spree/issues/7812, in my case it was fine to remove deface but I expect all other plugin maintainers will report this soon. The error they will see after migration is `can't modify frozen Array`

PR #4343, opened 2017-02-27
fixes #18712 - Spinner

gailsteiger:
create react Spinner component with patternfly styling

### DIFF
--- a/webpack/assets/javascripts/react_app/components/common/Spinner.js
+++ b/webpack/assets/javascripts/react_app/components/common/Spinner.js
@@ -1,0 +1,25 @@
+import React, { PropTypes } from 'react';
+
+const Spinner = ({size, inline, inverse}) => {
+  const css = `spinner spinner-${size} ` +
+    (inline ? ' spinner-inline ' : '') + (inverse ? ' spinner-inverse ' : '');
+
+  return (
+    <div className={css} >
+    </div>
+  );
+};
+
+Spinner.propTypes = {
+  inline: PropTypes.bool,
+  inverse: PropTypes.bool,
+  size: PropTypes.oneOf(['lg', 'sm', 'xs'])
+};
+
+Spinner.defaultProps = {
+  size: 'lg',
+  inline: false,
+  inverse: false
+};
+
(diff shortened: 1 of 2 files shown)


ohadlevy (changes requested):
thanks @gailsteiger.

Can you please update all other spinners to use this ? not much point in having multiple spinner implementations.

dLobatog:
Given my lack of success with https://github.com/idimaster/patternfly-react/pull/1 , and our growing set of Patternfly related components, how about we start extracting these to an npm package that we can use (and that we maintain?). 

I see a lot of benefits to doing it that way:

- The package can have it's own lifecycle, Foreman would just have to mount the components wherever needed and adapt it to use our API when needed.
- Maintainers don't have to be the same. We have a good number of people these days working exclusively on UI stuff who cannot merge things in the Foreman repo, but they should have access to merge stuff in the Patternfly React repo.
- Other projects can benefit from such a package and contribute to it. The few patternfly-react projects I've been seeing are unfortunately unmaintained or no actual project uses it. Since Foreman relies on React and Patternfly, we have a duty to keep that repository well maintained.

I don't mind to start this effort by moving some components to an npm package. My biggest concern with this, is that I'm not sure how possible is it to move our current components to an state where they don't need flux/redux. 

PS - If no one is starting this, I will after 1.15 is released, not before

gailsteiger:
@dLobatog @matanwerbner @ohadlevy @lizagilman and anyone else :)

You might want to have a look at  [Web Components/Native Framework Proposal](https://t.co/z9r6UZHxPe) proposed at the February 22 Patternfly Community Meeting. [slides](https://github.com/patternfly/patternfly-design/blob/master/resources/community-meetings/PatternFly%20Community%20Meeting%202017-02.pdf),  [recording](https://bluejeans.com/s/Cnu5T/)

I think participating in this could be fruitful.

tbrisker:
TBH I'm not sure I understand the value in creating a react component that just creates a spinner div - seems like a lot of code instead of just writing `<div class='spinner spinner-lg'>`. Am I missing something here?

PR #4345, opened 2017-02-28
Fixes #18718 - Empty boolean matcher should not turn into false

orrabin:


domcleal:
Please note that ticket #17870 has already been fixed and shipped in d836a838. A new ticket describing whatever bug this fixes should be filed.

tbrisker:
Thanks @orrabin !

PR #4346, opened 2017-02-28
Fixes #18717 - Allows use of proxies on multiple interfaces

sean797:
This moves Smart Proxies URL to its own model and during provision you select a URL instead of a proxy. All communication between foreman -> proxy is still done via the "Primary URL".


sean797:
Setting WIP, this needs a fair bit more work - only raised this really so I could run the whole test suite.

sean797:
This is ready for review now 👍 I'll have to send a PR to Katello to fix those test failures.

app/controllers/api/v1/hostgroups_controller.rb
```diff
@@ -33,10 +33,10 @@ def show
         param :architecture_id, :number
         param :medium_id, :number
         param :ptable_id, :number
-        param :puppet_ca_proxy_id, :number
```
stbenjam: These should get deprecated instead of removed entirely, you can pick the primary url if the user just gives the proxy ID

app/controllers/api/v1/smart_proxies_controller.rb
```diff
@@ -22,10 +22,18 @@ def index
...
+      def_param_group :url_attributes do
+        param :id, Fixnum, :desc => N_("Smart Proxy URL ID")
+        param :url, String, :required => true, :desc => N_("URL of Smart Proxy")
+        param :primary, :bool, :desc => N_("Primary URL. Defaults to false")
+      end
+
       api :POST, "/smart_proxies/", "Create a smart proxy."
       param :smart_proxy, Hash, :required => true do
         param :name, String, :required => true
-        param :url, String, :required => true
```
stbenjam: Ditto, deprecate this instead of removing it

stbenjam (commented):
Thanks @sean797!!!

I know a ton of people want something like this.  Not a full review from me, but just a few initial thoughts:

- I'm a little concerned we're removing some API's instead of deprecating them.

- Do we need to explicit support for multiple URL's, or wouldn't just alternate host names be enough? The proxy is still going to be using the same protocol and port isn't it? Maybe there is a use case for different ports?

- What about letting the user specify a name as well as a URL? If a proxy exists in a lot of networks, it'd be easier to pick something like "Atlanta - Dev Network" than try to select a URL.

- We also have another use case that's similar that might be able to fit in some kind of solution like this: this one seems to solve one proxy with multiple host names. The other use case that's common is when you have multiple proxies with 1 host name (load balanced). That's relevant for things like Katello where we want to make sure all the proxies have the same content assigned to them.  Also REX (@inecas may be interested) where you might want to assign an entire group of proxies to one network.  Just something to think about.

sean797:
Thanks for taking a look @stbenjam 

> - I'm a little concerned we're removing some API's instead of deprecating them.

Fair point, I've update it so if a proxy is given we use the primary URL. I've also on deprecated the API v2 params. I guess we should leave API v1 alone & stable ?

> - Do we need to explicit support for multiple URL's, or wouldn't just alternate host names be enough? The proxy is still going to be using the same protocol and port isn't it? Maybe there is a use case for different ports?

Host names are enough for our purposes. Though I'm not sure if a URL would be useful to any plugins?

> - What about letting the user specify a name as well as a URL? If a proxy exists in a lot of networks, it'd be easier to pick something like "Atlanta - Dev Network" than try to select a URL.

I'm not adverse to that, though I'm not sure its really that helpful. In my experience people tend to name things in foreman by another attribute, e.g Networks by CIDR, Proxies by hostname which a lot of the time creates unnecessary duplication.

sean797:
Just a couple of screenshots in its current state

screenshot from 2017-03-22 16-07-16
screenshot from 2017-03-22 16-02-04
screenshot from 2017-03-22 16-41-26

dmitri-d:
Tangential to this PR, but to support https on multiple interfaces, smart-proxy will need a wildcard certificate. ATM there's no way to associate a certificate with a specific interface.

sean797:
Thanks @witlessbird, https://github.com/theforeman/puppet-puppet/pull/490 fixes that using dns_alt_names 👍

PR #4347, opened 2017-02-28
Fixes #18514 - switch prefix/mask fields

chris1984:


dLobatog:
@chris1984 Thanks for the PR! However I don't understand the motivation. Both fields affect each other, e.g: changing the network mask changes the network prefix and viceversa. Why is it better to have the prefix first?

I see the original bug (https://bugzilla.redhat.com/show_bug.cgi?id=1377294) but that's no longer the case, it was fixed as part of the IPv6 work.

chris1984:
@dLobatog Thanks! Looking at the redmine issue I thought it was more of a user convenience, type in a 2 digit # and out pops the netmask for timesaving compared to 12 #'s, also thought it was still an issue with the redmine being opened.

dLobatog:
Thanks @chris1984 !

PR #4348, opened 2017-02-28
Fixes #18726 - Set :null => true on migrations for Rails 5

sean797:
Currently I'm seeing

`#timestamps` was called without specifying an option for `null`. In Rails 5, this behavior will change to `null: false`. You should manually specify `null: true` to prevent the behavior of your existing migrations from changing.


lzap:
@dLobatog any objections? I'd like to merge.

lzap:
Develop can't be migrated, I am merging this. Updated the commit message. Thanks a lot!

PR #4349, opened 2017-03-01
Fixes #18740 - remove erb extension from host_checkbox.js

amirfefer:


ohadlevy (changes requested):
thanks, overall looks good, can you please include this js in the hosts.js so we can now save a roundtrip to the server?

amirfefer:
@ohadlevy `hosts.js` loads only in new/edit pages, so I don't see any unnecessary roundtrip.

ohadlevy:
thanks @amirfefer

PR #4350, opened 2017-03-01
fixes #18735 - truncate encryption key to match preferred length

domcleal:
Ruby 2.4's OpenSSL bindings raise an ArgumentError during encryption if
the key length exceeds the cipher's configured length (32 bytes with the
default cipher), but the verification hash algorithm still uses the full
length key.

The encryption key is now truncated to the cipher's preferred length
while passing the full key (if supplied) for signatures.

The default new key length has been changed from 40 to 32 bytes matching
the cipher default, but there's no reason to deprecate or force existing
installations to change.

### DIFF
--- a/lib/foreman/util.rb
+++ b/lib/foreman/util.rb
@@ -22,9 +22,10 @@ module Foreman
       SecureRandom.base64(96).tr('+/=', '-_*')
     end
 
-    # recommended to make encryption_key at least 32 bytes. Ex. SecureRandom.hex(20)
+    # recommended to make encryption_key 32 bytes, matching the key length preferred by
+    # AS::MessageEncryptor's default algorithm
     def secure_encryption_key
-      SecureRandom.hex(20)
+      SecureRandom.hex(ActiveSupport::MessageEncryptor.key_len / 2)
     end
   end
 end
--- a/app/models/concerns/encrypt_value.rb
+++ b/app/models/concerns/encrypt_value.rb
@@ -39,7 +39,6 @@ module EncryptValue
 
   def encrypt_field(str)
     return str unless is_encryptable?(str)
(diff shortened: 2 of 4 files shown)


domcleal:
[test foreman] - no job started.

lib/foreman/util.rb
```diff
@@ -22,9 +22,10 @@ def secure_token
       SecureRandom.base64(96).tr('+/=', '-_*')
     end
 
-    # recommended to make encryption_key at least 32 bytes. Ex. SecureRandom.hex(20)
+    # recommended to make encryption_key 32 bytes, matching the key length preferred by
+    # AS::MessageEncryptor's default algorithm
     def secure_encryption_key
-      SecureRandom.hex(20)
+      SecureRandom.hex(ActiveSupport::MessageEncryptor.key_len / 2)
```
lzap, line 28: Nitpic: while touching this, could you replace hex with base64 or anything that is higher entropy? I understand that hex is easy to handle (just divide by two), but why no to be more secure by default? Base64 is still printable, the only drawback is slash char but we do not store this in any kind of filename.

domcleal: Please file a new PR or issue for it, simply changing this to Base64 isn't any more correct than hexadecimal (neither provides a full 8-bit 32-byte range) and changing the format of the encryption key is outside of the scope of this ticket.

lzap: Sure, just for the record, hexadecimal key entropy is substantially smaller (4.000) than upper&lower case alphabet (5.7004). Since base64 is even more (three more characters) than just alphabet, it's approximetly half-times stronger (4 vs 6).

The failure is irrelevant, thanks.

PR #4352, opened 2017-03-02
Fixes #18763 - Change delete button to the secondary button style

junezhang:
From the demo meeting[1], we suggested the delete button to follow the secondary button style, also the default style, please check the attachment as following:
delete-button-2 1

As when user choose to click the delete button, in the system, it has a confirmation with user, so it's not necessary to be different with others.
This patch is aim to fix this issue. 

[1] https://groups.google.com/forum/?fromgroups#!topic/foreman-dev/Gx69q6ZRB3s


ohadlevy:
thanks @junezhang

PR #4354, opened 2017-03-03
Fixes #18783 - refactor role form help

ares:
While working on this area I took the liberty to improve the form help after consulting with @rohoover 
This PR allows us to start moving popover icons to labels (safe mode fix) and actually uses it for multiple select in this form. Since we couldn't add it to label before, the inline notification has it's own popover. Also it turned out that the positioning of it was confusing, when it's on top it makes more sense since you start from there when you're editing.

after:
roles1

before:
roles2

### DIFF
--- a/app/helpers/form_helper.rb
+++ b/app/helpers/form_helper.rb
@@ -307,13 +307,13 @@ module FormHelper
     label = options[:label] == :none ? '' : options.delete(:label)
     label ||= ((clazz = f.object.class).respond_to?(:gettext_translation_for_attribute_name) &&
         s_(clazz.gettext_translation_for_attribute_name attr)) if f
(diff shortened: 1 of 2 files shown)


Rohoover:
bitmoji

tbrisker:
While this improves this form, would it make sense to move all popover help to be in the labels? 
Otherwise this form is inconsistent with all other forms which have the tooltip after the fields.

ares:
Thanks for looking @tbrisker, I'd like to keep it in separate PR. The reason why these two fields are different is because they use two-pane form field where inline help was not possible to achieve. Anyway I opened separate PR with the change so I hope this could be merged as is.

tbrisker:
considering the last round of changes, should this be closed in favor of #4358 as that introduces a more generalized solution for the tooltips?

ares:
Closing as suggested

PR #4355, opened 2017-03-03
Fixes #18722 - fix host parameters in host form

ares:


ares:
If it's acked, it should be cherry-picked to 1.14-stable as it's marked for 1.14.3 in redmine. Not sure if the Label 1.14 is used for that so rather not setting it.

dLobatog:
Works well - I do agree it has to be cp-ed. @ares can you write a test for it? `test/integration/parameters_permissions_test.rb` doesn't cover this well. Thanks

ares:
Test added. Testing this is hell.

ohadlevy:
@ares one of the reasons I think we should slowly move params into react/redux

dLobatog:
Thanks @ares!

PR #4356, opened 2017-03-03
Fixes #18788 - log protected attributes in prod

lzap:
I see no reason not to do this. More info in the ticket.


lzap:
This might log everytime there is an extra param while error is raised only when such a param is accessed I believe. We need to test it if that does not flood logs.

ares:
The Foreman tests failures are unrelated `DashboardIntegrationTest.test_0003_dashboard link hosts in error state`. Works well, it logs forbidden attributes in debug log level. Thanks @lzap, merged as 3582015c6acf4c278c22c9e96c0352a43a4a094b

PR #4357, opened 2017-03-05
fixes #18796 - do not use statistics-pie css as its used by flot

ohadlevy:
flot initalization relay on the above css class, using it here makes
legacy flot to try and initalize on the same div as c3.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -6,7 +6,6 @@
 //= require scoped_search
 //= require bootstrap
 //= require patternfly
-//= require charts
 //= require topbar
 //= require two-pane
 //= require vendor
--- a/app/views/about/index.html.erb
+++ b/app/views/about/index.html.erb
@@ -1,5 +1,5 @@
 <% title _("About") %>
-<%= javascript 'proxy_status', 'about' %>
+<%= javascript 'proxy_status', 'charts', 'about' %>
 
 <div class="row" id="about">
   <div class="col-md-7">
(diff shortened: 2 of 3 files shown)


tbrisker:
@ohadlevy Appears that one of the js tests broke.

gailsteiger:
The modal charts aren't opening.

Remove `//= require charts` from app/assets/javascripts/application.js line 9 (I thought I did this in the original PR) and add back the `statistics-pie` class

Should solve both problems and the test should pass.

ohadlevy:
@gailsteiger I've updated it according to your suggestions, I do believe this will impact plugins that are using charts (/cc @ares for openscap and rex).

I also think using the same css class that is used by flot is confusing.

gailsteiger:
@ohadlevy I don't care much about the css class. `small` has a different meaning when used without `statistics-pie` but things look OK. If `statistics-pie` is removed the failing test should be fixed. I think charts.js should be removed from application.js. Can't the plugins add charts.js where needed as in core?

charts.js adds an event listener to the window object: 

`$(window).on('resize', refreshCharts);`

We should look to retiring that file.

dLobatog:
Tests still fail, set to WoC

gailsteiger:
@dLobatog the js test mentioned by tbrisker above no longer fails.

@ohadlevy Are the failing foreman tests related to this change? I didn't see any connections.

dLobatog:
@gailsteiger Yes, see http://ci.theforeman.org/job/test_develop_pr_core/12358/#showFailuresLink or how the errors are `ReferenceError: Can't find variable: refreshCharts`

gailsteiger:
@dLobatog Thank you very much for the info.

ohadlevy:
@dLobatog updated missing charts.js, the one place that I was confused about was the host form, on submit, we replace the page content instead of redirecting, basically ignoring the html head sections.

gailsteiger:
OK with me

dLobatog:
Thanks @ohadlevy @gailsteiger

PR #4358, opened 2017-03-06
Fixes #18804 - replace inline helps with label helps

ares:
Depends on changes from #4354 
Also for the reviewer, I think we should give enough time to plugin devs to align their code bases. While I can work on some, it's out of my capacity to fix everything. Releasing this with 1.15 introduces a risk that we'd have mixed icons in single form (e.g. rex adding flag to interface, see example below). Therefore I'd suggest to keep this for 1.16

nics

### DIFF
--- a/app/helpers/form_helper.rb
+++ b/app/helpers/form_helper.rb
@@ -302,18 +302,24 @@ module FormHelper
   end
 
   def add_label options, f, attr
+    return ''.html_safe if options[:label] == :none
+
     label_size = options.delete(:label_size) || "col-md-2"
     required_mark = check_required(options, f, attr)
-    label = options[:label] == :none ? '' : options.delete(:label)
-    label ||= ((clazz = f.object.class).respond_to?(:gettext_translation_for_attribute_name) &&
-        s_(clazz.gettext_translation_for_attribute_name attr)) if f
-    label = label.present? ? label_tag(attr, "#{label}#{required_mark}", :class => label_size + " control-label") : ''
+    label = ''.html_safe + options.delete(:label)
+    label += ((clazz = f.object.class).respond_to?(:gettext_translation_for_attribute_name) &&
+      s_(clazz.gettext_translation_for_attribute_name attr).html_safe) if f && label.empty?
+
+    if options[:label_help].present?
+      label += ' '.html_safe + popover("", options[:label_help], options[:label_help_options] || {})
+    end
+    label = label.present? ? label_tag(attr, label.to_s + required_mark.to_s, :class => label_size + " control-label") : ''
     label
   end
 
   def check_required options, f, attr
     required = options.delete(:required) # we don't want to use html5 required attr so we delete the option
-    return ' *' if required.nil? ? is_required?(f, attr) : required
+    return ' *'.html_safe if required.nil? ? is_required?(f, attr) : required
   end
 
   def blank_or_inherit_f(f, attr)
--- a/app/views/lookup_keys/_fields.html.erb
+++ b/app/views/lookup_keys/_fields.html.erb
@@ -14,25 +14,26 @@
     <h2><%= _("Default behavior") %></h2>
     <h6><%= _("Override the default value of the Puppet class parameter.") if is_param%></h6>
     <%= checkbox_f(f, :override, :onchange => 'toggleOverrideValue(this)', :size => "col-md-8",
-                   :help_inline => popover('', _('Whether the class parameter value is managed by Foreman.'))
+                   :label_help => _('Whether the class parameter value is managed by Foreman.')
         ) if is_param %>
     <%= param_type_selector(f, :onchange => 'keyTypeChange(this)') %>
     <%= textarea_f f, :default_value, :value => f.object.default_value_before_type_cast, :size => "col-md-8",
                    :disabled => (is_param && (!f.object.override || f.object.omit)),
                    :input_group_btn => fullscreen_input,
                    :rows => 1,
-                   :help_inline => popover('', _("Value to use when there is no match.")),
+                   :label_help => _("Value to use when there is no match."),
                    :class => "no-stretch #{'masked-input' if f.object.hidden_value?}" %>
     <div class="form-group">
       <%= checkbox_f(f, :omit, :label => _('Omit'),
-                     :help_inline => omit_help,
+                     :label_help => omit_help_body,
+                     :label_help_options => { :title => omit_help_title },
                      :size => "col-md-1", :label_size => "col-md-2", :table_field => true,
                      :onchange => 'toggleOmitValue(this, "default_value")',
                      :disabled => (is_param && !f.object.override)) if is_param %>
       <%= checkbox_f(f, :hidden_value, :label => _('Hidden value'),
                      :class => 'hidden_value_textarea_switch', :onchange => 'toggle_lookupkey_hidden(this)',
                      :checked => f.object.hidden_value?,
-                     :help_inline => popover("", _("Hide all values for this parameter.")),
+                     :label_help => _("Hide all values for this parameter."),
                      :size => "col-md-1", :label_size => "col-md-2", :table_field => true,
                      :disabled => (is_param && !f.object.override)) %>
     </div>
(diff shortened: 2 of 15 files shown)


app/helpers/form_helper.rb
```diff
@@ -304,16 +304,24 @@ def help_inline(inline, error)
   def add_label options, f, attr
     label_size = options.delete(:label_size) || "col-md-2"
     required_mark = check_required(options, f, attr)
-    label = options[:label] == :none ? '' : options.delete(:label)
-    label ||= ((clazz = f.object.class).respond_to?(:gettext_translation_for_attribute_name) &&
-        s_(clazz.gettext_translation_for_attribute_name attr)) if f
-    label = label.present? ? label_tag(attr, "#{label}#{required_mark}", :class => label_size + " control-label") : ''
+    if options[:label] == :none
```
tbrisker: I wonder, why do we even go into this function if `options[:label] == :none`? all we get is an empty string in that case, and in fact the field helper has another option, `option[:no_label]` that does exactly that.

app/helpers/form_helper.rb
```diff
@@ -304,16 +304,24 @@ def help_inline(inline, error)
...
+      label = ''.html_safe
+    else
+      label = ''.html_safe + options.delete(:label)
+      label += ((clazz = f.object.class).respond_to?(:gettext_translation_for_attribute_name) &&
+        s_(clazz.gettext_translation_for_attribute_name attr).html_safe) if f && label.empty?
+    end
+
+    if options[:label_help].present?
+      label += ' '.html_safe + popover("", options[:label_help], options[:label_help_options] || {})
```
tbrisker: `html_safe` isn't needed here on the empty space.

app/helpers/form_helper.rb
```diff
@@ -304,16 +304,24 @@ def help_inline(inline, error)
...
+    end
+    label = label.present? ? label_tag(attr, label.to_s + required_mark.to_s, :class => label_size + " control-label") : ''
     label
   end
 
   def check_required options, f, attr
     required = options.delete(:required) # we don't want to use html5 required attr so we delete the option
-    return ' *' if required.nil? ? is_required?(f, attr) : required
+    return ' *'.html_safe if required.nil? ? is_required?(f, attr) : required
```
tbrisker, line 322: this doesn't need `html_safe`

app/views/lookup_keys/_fields.html.erb
```diff
@@ -14,36 +14,37 @@
...
+                     :label_help => _("Hide all values for this parameter."),
                      :size => "col-md-1", :label_size => "col-md-2", :table_field => true,
                      :disabled => (is_param && !f.object.override)) %>
     </div>
   </fieldset>
   <fieldset>
     <%= collapsing_header _("Optional input validator"), "#optional_input_validators_#{f.object.id}", "collapsed" %>
     <div id="optional_input_validators_<%= f.object.id %>" class="collapse out">
-      <%= alert :class => 'alert-info', :header => _('Note'), :text => _('If ERB is used in a parameter value, the validation of the value will happen during the ENC request. If the value is invalid, the ENC request will fail.') %>
+      <h6><%= _('If ERB is used in a parameter value, the validation of the value will happen during the ENC request. If the value is invalid, the ENC request will fail.') %></h6>
```
tbrisker, line 44: Is this change related?

app/views/lookup_keys/_fields.html.erb
```diff
@@ -53,17 +54,18 @@
...
-                      :title => _("The order in which values are resolved.")).html_safe %> <%= _("Set the order in which values are resolved.") %></h6>
-      <%= textarea_f f, :path, :rows => :auto, :label => _("Order"), :id => 'order', :size => "col-md-8", :onchange => 'fill_in_matchers()', :value => f.object.path, :class => "no-stretch" %>
+      <h6><%= _("Set the order in which values are resolved.") %></h6>
+      <%= textarea_f f, :path, :rows => :auto, :label => _("Order"), :id => 'order', :size => "col-md-8", :onchange => 'fill_in_matchers()', :value => f.object.path, :class => "no-stretch",
+                        :label_help => _("The order in which matchers keys are processed, first match wins.<br> You may use multiple attributes as a matcher key, for example, an order of <code>host group, environment</code> would expect a matcher such as <code>hostgroup = \"web servers\", environment = production</code>").html_safe,
+                        :label_help_options => { :title => _("The order in which values are resolved.") } %>
```
tbrisker: Does it make sense to have a title for the popover that is almost identical to the header for the section?

app/views/nic/_base_form.html.erb
```diff
@@ -4,24 +4,21 @@
...
+              :label_help => _("Device identifier for this interface. This may be different on various platforms and environments, here are some common examples.<br/><ul>" +
                                 "<li>Use the basic name for physical interface identifiers, e.g. <strong>eth0</strong> or <strong>em0</strong> with biosdevname.</li>" +
                                 "<li>For virtual interfaces, use either alias notation (<strong>eth0:1</strong>, name:index) or VLAN notation (<strong>eth0.15</strong>, name.tag).</li>" +
-                                "<li>For bonds it's common to use <strong>bond0</strong> on Linux, <strong>lagg0</strong> on FreeBSD systems.</li></ul>"),
-                      :title => _("Device identifier"),
-                      :rel => 'popover-modal',
-                      :'data-placement' => "top"
-              ),
+                                "<li>For bonds it's common to use <strong>bond0</strong> on Linux, <strong>lagg0</strong> on FreeBSD systems.</li></ul>").html_safe,
+              :label_help_options => { :title => _("Device identifier"), :rel => 'popover-modal', :'data-placement' => "top" },
```
tbrisker: title here also seems redundent

app/views/nic/bmcs/_bmc.html.erb
```diff
@@ -3,15 +3,15 @@
 <%= content_tag :span, :id => 'bmc_fields' do %>
   <%= text_f f, :username, :size => "col-md-8", :label_size => "col-md-3" %>
   <%= password_f f, :password, :unset => action_name == "edit", :size => "col-md-8", :label_size => "col-md-3",
-                    :help_inline => popover('',
-                        _("The BMC password will be used by Foreman to access the host's BMC controller via a BMC-enabled smart proxy, if available.") + ' ' +
+                    :label_help => _("The BMC password will be used by Foreman to access the host's BMC controller via a BMC-enabled smart proxy, if available.<br>").html_safe + ' ' +
```
tbrisker: nitpick: should be `<br/>`, and no need for extra space when breaking line.

tbrisker (changes requested):
Looks good in general, see inline comments.

app/helpers/form_helper.rb
ares: maybe to get a required mark? anyway I'd like to avoid bigger helpers refactoring in this PR :-)

tbrisker: nope, follow the existing code - if it is none then this function just returns an empty string after checking a couple of options.

tbrisker: If you want to not touch other functions you can short-circuit at the top and `return '' if options[:label] == :none` as the first line so nothing changes otherwise

app/helpers/form_helper.rb
ares: That's true only because this helper is called from `add_label` where we append to safe buffer, I think it's a good practice that helper returns safe strings (if they are safe), at least internal rails helpers do. While we don't do it anywhere in core code, I think this is useful.

app/views/lookup_keys/_fields.html.erb
ares: No, but it brings consistency with the role form too, it's the same thing that @rohoover suggested there.

app/views/lookup_keys/_fields.html.erb
ares: I'm happy to remove (even though I just kept how it was before) as long as you won't made me to extract `<h6>` change into separate PR :-)

app/helpers/form_helper.rb
ares: it is, or I need to write it as `label = label + ' '.html_safe + popover("", options[:label_help], options[:label_help_options] || {})`

app/helpers/form_helper.rb
ares: updated

ares:
Thanks @tbrisker, once you confirm, I'll squash two commits and update the message so it closes both issues.

tbrisker (approved):

tbrisker:
LGTM, squash away and I will merge once tests are green.

ares:
Squashed, I closed the original PR.

tbrisker:
Thanks @ares ! Please update the foreman-dev list about this so plugin authors can take advantage of it.

dLobatog:
Awesome work here!

ares:
Thanks @tbrisker, sent with additional PRs to most used plugins.

PR #4359, opened 2017-03-06
Fixes #18764 - Hammer-cli current user password update

dhlavac:


app/controllers/api/v2/users_controller.rb
```diff
@@ -36,21 +36,32 @@ def index
...
-          param :default_organization_id, Integer if SETTINGS[:organizations_enabled]
-          param :auth_source_id, Integer, :required => true
-          param :timezone, ActiveSupport::TimeZone.all.map(&:name), :required => false, :desc => N_("User's timezone")
-          param :locale, FastGettext.available_locales, :required => false, :desc => N_("User's preferred locale")
-          param_group :taxonomies, ::Api::V2::BaseController
+          param_group :user_params
+        end
+      end
+
+      def_param_group :current_user_update do
```
tstrachota: Since the endpoint can be used for updating any user, not just the current one, would you mind to rename to just `user_update`?

app/controllers/api/v2/users_controller.rb
```diff
@@ -36,21 +36,32 @@ def index
...
+        param :user, Hash, :required => true, :action_aware => true do
+          param_group :user_params
+          param :current_password, String, :required => true
```
tstrachota: I think that `current_password` is required only when you perform the request from non-admin account so `:required => true` should go away.

tstrachota: Also a description when this field is required would be handy. Could you add it, please?

tstrachota (changes requested):
Works fine in general. Added some comments about name and description of the new params.

dhlavac:
@tstrachota solved comments please look at PR

tstrachota:
LGTM :+1: @dhlavac

dLobatog:
Thanks @dhlavac @tstrachota !

PR #4361, opened 2017-03-07
Fixes #18767 - keep cloning info

ares:


app/models/role.rb
```diff
@@ -31,6 +31,7 @@ class Role < ActiveRecord::Base
     compare = 'not' if args.first
     where("#{compare} builtin = 0")
   }
+  scope :cloned, -> { where('cloned_from_id IS NOT NULL') }
```
dLobatog: `where.not(:cloned_from_id => nil)` would be slightly better. The reason is that if scopes are chained, there is no reference to the roles table on cloned_from_id. 

```

irb(main):004:0> Role.where.not(:cloned_from_id => nil).to_sql
=> "SELECT \"roles\".* FROM \"roles\" WHERE (\"roles\".\"cloned_from_id\" IS NOT NULL)"
irb(main):005:0> Role.where('cloned_from_id IS NOT NULL').to_sql
=> "SELECT \"roles\".* FROM \"roles\" WHERE (cloned_from_id IS NOT NULL)"
```

app/views/roles/_form.html.erb
```diff
@@ -10,8 +10,14 @@
 
   <div class="tab-content">
     <div class="tab-pane active" id="primary">
+      <% if @role.cloned_from_id.present? && User.current.can?(:view_roles, @role.cloned_from) %>
```
dLobatog: @rohoover I think the mix of styles used here isn't quite good, it looks inconsistent (2 styles to show an info message). Not sure how to fix it, any ideas?

screenshot from 2017-03-08 10-32-18

@ares I'm happy to merge and fix it later.

dLobatog (changes requested):
@ares Thanks - I only have 1 comment about the scope that I think would make sense to get fixed so that we ensure complicated nested scopes keep on working.

👍 to merge after that, but I'll wait until the end of the day in case @rohoover can suggest a way to fix the look of the form better

app/views/roles/_form.html.erb
ares: I consulted this with @rohoover already, the current PR status is the outcome. Note that with the recent `label_help` change the form looks more consistent since the notification is already converted to h5. I just need to do the rebase. I can update the screenshot after the rebase.

app/models/role.rb
ares: oh good catch :exclamation:

ares:
thanks @dLobatog, I rebased the PR, please note the changes in `divider` caused by the rebase

here are current screenshots

in case the role was not cloned
single

in case the role was cloned
both

@rohoover I'd appreciate if you could ACK

dLobatog (approved):

dLobatog:
@ares Looks good to me 😄  thanks! Waiting for @rohoover to ack

Rohoover:
@dLobatog 
bitmoji

PR #4364, opened 2017-03-08
Fixes #18800 - strong params no longer breaks inherited attrs

lzap:
We have two bugs already, this is important issue that prevents from host
creations when Katello is installed. The other one was reported on discovery
and we already have a PR to fix this in discovery, but it's preferred to fix
this in core.

https://github.com/theforeman/foreman_discovery/pull/331

### DIFF
--- a/app/models/host/managed.rb
+++ b/app/models/host/managed.rb
@@ -561,8 +561,9 @@ class Host::Managed < Host::Base
 
   def apply_inherited_attributes(attributes, initialized = true)
     return nil unless attributes
-    #don't change the source to minimize side effects.
-    attributes = hash_clone(attributes).with_indifferent_access
+    #convert possible strong parameters to unsafe hash (filtering out unsafe items) and
+    #clone to minimize side effects
+    attributes = hash_clone(attributes.to_h).with_indifferent_access
 
     new_hostgroup_id = attributes['hostgroup_id'] || attributes['hostgroup_name'] || attributes['hostgroup'].try(:id)
     #hostgroup didn't change, no inheritance needs update.
--- a/test/models/host_test.rb
+++ b/test/models/host_test.rb
@@ -3040,6 +3040,20 @@ class HostTest < ActiveSupport::TestCase
       host = Host.new(:name => "test-host", :hostgroup => hg)
       assert host.environment
     end
+
+    test 'should filter out unpermitted strong parameters' do
+      host = FactoryGirl.build(:host, :managed, :with_hostgroup)
+      attributes = ActionController::Parameters.new({ 'unpermitted_attribute' => 1 })
+      actual_attr = host.apply_inherited_attributes(attributes)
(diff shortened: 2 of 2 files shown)


Klaas-:
I've tested the fix on 1.14.2/katello 3.3 and creating a server via api works without specifying content_facet_attributes

ShimShtein:
:+1: It's the proper fix. It filters out unpermitted attributes and produces clean permitted hash.

lzap:
Added two unit tests to prevent regressions.

test/models/host_test.rb
```diff
@@ -3040,6 +3040,20 @@ def to_managed!
...
+      attributes = ActionController::Parameters.new({ 'unpermitted_attribute' => 1 })
+      actual_attr = host.apply_inherited_attributes(attributes)
+      assert_nil actual_attr['unpermitted_attribute']
+    end
+
+    test 'must not return strong parameters' do
+      host = FactoryGirl.build(:host, :managed, :with_hostgroup)
+      attributes = ActionController::Parameters.new
+      actual_attr = host.apply_inherited_attributes(attributes)
+      refute_equal actual_attr.class, ActionController::Parameters
```
ShimShtein: maybe we should use

``` ruby
refute actual_attr.is_a? ActionController::Parameters
```

lzap: Amended!

app/models/host/managed.rb
```diff
@@ -561,8 +561,9 @@ def hostgroup_inherited_attributes
 
   def apply_inherited_attributes(attributes, initialized = true)
     return nil unless attributes
-    #don't change the source to minimize side effects.
-    attributes = hash_clone(attributes).with_indifferent_access
+    #convert possible strong parameters to unsafe hash (filtering out unsafe items) and
+    #clone to minimize side effects
+    attributes = hash_clone(attributes.to_h).with_indifferent_access
```
ShimShtein, line 566: I am a bit concerned that `attributes` won't have `#to_h` method (if somebody passes a regular `Hash` instead of params).
I am not sure that it's valid use case anyway, so I won't block this PR if it won't be fixed.

timogoebel: I don't see, why this should be a problem.

```
irb
|{2.3.0 :001 > {}.to_h
 => {}
2.3.0 :002 > {:a => :b}.to_h
 => {:a=>:b}
2.3.0 :003 >
```

ShimShtein: right, my bad.

timogoebel (approved):
LGTM, any objections?

timogoebel:
Merged, thanks @lzap and @ShimShtein.

app/models/host/managed.rb
lzap: Yeah I checked also Ruby code, it returns itself (no copy).

PR #4365, opened 2017-03-08
Fixes #18843 - Changed from eager_load to includes in #index

ShimShtein:
Avoiding multiple joins that result in very wide result rows. It leads to big memory footprint in the DB especially on big datasets. This big footprint translates then into DB high CPU usage while processing the query.

### DIFF
--- a/app/controllers/api/v2/hosts_controller.rb
+++ b/app/controllers/api/v2/hosts_controller.rb
@@ -20,7 +20,7 @@ module Api
 
       add_smart_proxy_filters :facts, :features => Proc.new { FactImporter.fact_features }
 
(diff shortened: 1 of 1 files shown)


timogoebel:
@ShimShtein : Please check out the failing tests. Thanks.

ShimShtein:
It seems that in the test the system chooses to implement `includes` as eager loads (using left outer joins) and it causes `:interfaces` relation to load incorrectly.

ShimShtein:
Changed the implementation from `#includes` to `#preload`. This forces Active Record to use a query per association - which is good for big databases, and eliminates the wrong eager_loading bug for small ones. It does impact the performance of small databases, since it's worse than using `#eager_load`, but the impact is minor.

dLobatog:
Thanks @ShimShtein @timogoebel

PR #4314, opened 2017-02-21
Adding VMware vSphere Disk Mode support (Dependent, Independent - P…

sbernhard:
fixes #18815 - Adding VMware vSphere Disk Mode support

PR #4252, opened 2017-02-02
Fixes #12652 - Remove VMware Volume name

chris1984:
Remove the Volume name label under Create Host since at this time its not possible to name VMware volumes.

PR #4366, opened 2017-03-09
Fixes #4509 - VMWare scsi react

matanwerbner:
Adding vmware storage controllers + volumes in react

@timogoebel can you please check it's working with server side? ty


config/webpack.config.js
```diff
@@ -100,7 +100,7 @@ if (production) {
     hot: true
   };
   // Source maps
-  config.devtool = 'inline-source-map';
+  config.devtool = 'eval';
```
ohadlevy: not related to this pr?

webpack/assets/javascripts/react_app/common/colors.scss
```diff
@@ -0,0 +1 @@
+$light-grey: #efefef;
```
ohadlevy: I assume this is from patternfly pallet? https://www.patternfly.org/styles/color-palette/#_

webpack/assets/javascripts/react_app/components/common/forms/CommonForm.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+const CommonForm = (props) => {
+  return (
+      <div className={ `form-group ${props.className || ''}` }>
```
ohadlevy: cant we default the props in the const, e.g.:

```js
const CommonForm = ({className = '',  label = ''}) => { ...
````

File: webpack/assets/javascripts/react_app/components/common/forms/Select.js [hunk 4]
```diff
@@ -0,0 +1,23 @@
...
+      >
+        <option value="">{__('Please select')}</option>
+        {props.options}
+      </select>
+      </CommonForm>
+  );
+};
+
+Select.defaultProps = {
+  className: ''
```
ohadlevy: inconsistent with CommonForm?

webpack/assets/javascripts/react_app/components/hosts/powerStatus/PowerStatus.scss
```diff
@@ -0,0 +1,11 @@
+.host-power-status {
+  &.on {
+    color: #3f9c35;
```
ohadlevy, line 3: not related to the pr.

webpack/assets/javascripts/react_app/components/hosts/powerStatus/index.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import { connect } from 'react-redux';
-import * as HostsActions from '../../../redux/actions/hosts/';
+import * as HostsActions from '../../../redux/actions/hosts/powerStatus';
```
ohadlevy: ditto?

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/disk.scss
```diff
@@ -0,0 +1,21 @@
+@import '../../../../../../common/colors';
```
ohadlevy, line 1: can we add a variable for common path (absolute vs relative?)

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
```diff
@@ -0,0 +1,85 @@
...
+        value={dataStore}
+        onChange={updateDisk.bind(this, 'dataStore')}
+        options={selectableStores}
+      />
+
+      <TextInput
+        value={size}
+        className="text-vmware-size"
+        onChange={updateDisk.bind(this, 'size')}
+        label={__('Size (GB)')}
```
ohadlevy: should we include any basic validation (e.g. size is bigger than 0 and and is integer?)

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/index.js
```diff
@@ -0,0 +1,89 @@
...
+  ControllerTypes,
+  controller
+}) => {
+  const _updateController = (attribute, e) => {
+    const value = e.target.type === 'checkbox' ? e.target.checked : e.target.value;
+
+    updateController({[attribute]: value});
+  };
+
+  const _updateDisk = (diskIndex, attribute, e) => {
```
ohadlevy: whats the difference for this function? comment maybe?

webpack/assets/javascripts/react_app/redux/actions/hosts/storage/vmeware.consts.js
```diff
@@ -0,0 +1,11 @@
+export const defaultConrollerAttributes = {
+  type: 'ParaVirtualSCSIController'
+};
+export const defaultDiskAttributes = {
+  size: '10',
+  dataStore: '',
+  storagePod: '',
+  thinProvision: false,
+  eagerZero: false,
+  name: 'Hard disk'
```
ohadlevy: should this be translated?

webpack/assets/javascripts/react_app/redux/index.js
```diff
@@ -14,3 +14,9 @@ export default createStore(
   window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__(),
   applyMiddleware(...middleware)
 );
+
+export const getNewStore = () => createStore(
+  reducer,
```
ohadlevy: what does this do? :)

webpack/assets/javascripts/react_app/redux/reducers/hosts/powerStatus/powerStatus.fixtures.js
```diff
@@ -1,7 +1,5 @@
 import Immutable from 'seamless-immutable';
-export const initialState = Immutable({
```
ohadlevy: please elaborate on all the structure changes?

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/StorageContainer.consts.js
```diff
@@ -0,0 +1,9 @@
+export const MaxDisksPerController = 15;
+export const MaxControllers = 4;
+export const ControllerTypes = {
+  VirtualBusLogicController: 'Bus Logic Parallel',
```
timogoebel: Can we get these values from the rails model?

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/index.js
```diff
@@ -0,0 +1,89 @@
...
+
+  return (
+    <div className="controller-container">
+      <div className="controller-header">
+        <div className="control-label col-md-2 controller-selected-container">
+          <Checkbox
+            className="cb-controller-is-selected"
+            checked={controller.isSelected}
+            onChange={_updateController.bind(this, 'isSelected')}/>
+          <label>{__('Create SCSI controller')}</label>
```
timogoebel: This checkbox should be checked by default. A VM needs at least one scsi controller or your VM won't have any storage.
I'm wondering if this checkbox is actually useful.

webpack/assets/javascripts/react_app/components/common/forms/Select.js
```diff
@@ -0,0 +1,23 @@
...
+
+const Select = (props) => {
+  return (
+    <CommonForm label={props.label}
+    className={`common-select ${props.className}`}>
+      <select className="form-control"
+              value={props.value}
+              onChange={props.onChange}
+      >
+        <option value="">{__('Please select')}</option>
```
timogoebel: Can we hide this if a user selects something, e.g. make this a real placeholder?

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
```diff
@@ -0,0 +1,85 @@
...
+        label={__('Storage Pod')}
+        value={storagePod}
+        onChange={updateDisk.bind(this, 'storagePod')}
+        options={selectablePods}
+      />
+
+      <Select
+        label={__('Data store')}
+        value={dataStore}
+        onChange={updateDisk.bind(this, 'dataStore')}
```
timogoebel: The data doesn't update when you change the compute resource. We have a couple of VMWare compute resources that all have different datastores.
If I select one compute resource, then another one it still shows the datastores of the first one.

timogoebel: In addition if you create a new volume, the select fields aren't rendered as `select2´. The usage makes a lot of sense, though. We have at least 100 datastores. Select2 helps a lot.

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
```diff
@@ -0,0 +1,85 @@
...
+  const selectablePods = storagePods.map(attribute => {
+    const key = Object.keys(attribute)[0];
+    const value = Object.values(attribute)[0];
+
+    return <option key={key} value={key}>{value}</option>;
+  });
+
+  return (
+    <div className="disk-container">
+      <TextInput
```
timogoebel: This should be read-only.

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
```diff
@@ -0,0 +1,85 @@
...
+
+  const selectablePods = storagePods.map(attribute => {
+    const key = Object.keys(attribute)[0];
+    const value = Object.values(attribute)[0];
+
+    return <option key={key} value={key}>{value}</option>;
+  });
+
+  return (
+    <div className="disk-container">
```
timogoebel: Disk mode parameter is missing. It was introduced in 1d5f6c2ff89d9acde4bb03c6833df79a4ff10483.

timogoebel (changes requested):
@matanwerbner : Thanks for taking over the PR! I did not test provisioning, yet. But I have some comments regarding the react UI.

webpack/assets/javascripts/react_app/components/hosts/powerStatus/index.js
matanwerbner: this is related, i've changed folder structure quite a bit

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/index.js
matanwerbner: they each dispatch a different action - 
one handles changes for the controller object,
the other for disk object.

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/index.js
matanwerbner: this is by design supplied (i will attach), it's only purpose is to decide which controllers get deleted when you click "delete controllers".

webpack/assets/javascripts/react_app/redux/index.js
matanwerbner: this creates a new store, it is very usefull for testing

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/index.js
timogoebel: I think the design is flawed then. It would be better to have a dedicated delete button per controller (we already have a dedicated button per disk). The checkbox is very confusing. I didn't get, that it's related to the `Delete Controller` button.

image

btw: This is how this looks in VMWare. The UI is pretty bad I think, but Foreman users might be familiar with this:

image
(Sorry for the German language, but I guess you get the point)

@Rohoover: Do you agree? Any more comments?

Rohoover: @timogoebel @matanwerbner 

The original design was provided here: https://github.com/theforeman/foreman/pull/3979#issuecomment-276147276

It was done with a low level of effort in mind. The checkboxes work similar to the Foreman tables, where items are selected before actions occur to them.

That said, alternative layouts can be done. Attached is another quick mockup with you feedback. I think it we wanted we can also add the chevrons (accordion) feature shown in the VMWare UI as well if desired.

create controller volumev2

timogoebel: @Rohoover : Thanks. I think with this design it's a lot more clear how to delete a controller. I like that.

I do think the chevrons are a good idea. However I don't think it's a requirement for this PR.
@matanwerbner: If you want to earn some extra points, go ahead.

webpack/assets/javascripts/react_app/components/hosts/powerStatus/PowerStatus.scss
matanwerbner: i've moved powerStatus files to their own dedicated directory inside hosts, thats the relation

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/StorageContainer.consts.js
matanwerbner: yes, can you help put those there? my ruby is very weak

timogoebel: Sure, will do.

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/disk.scss
matanwerbner: this is a bit tricky, i will do a seperate PR on it

webpack/assets/javascripts/react_app/redux/reducers/hosts/powerStatus/powerStatus.fixtures.js
matanwerbner: state is changed to `{ hosts: { powerStatus: {}, storage: {} } }`

webpack/assets/javascripts/react_app/components/common/forms/Select.js
matanwerbner: this is hidden when a user selects another options, it is regular select behavior, same as other options in the select

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
matanwerbner: disk name should be read only? maybe we should no show it at all then..

timogoebel: the name is generated by VMWare. A user cannot change it but should be able to see it when editing a host (or he would edit the wrong disk by accident).

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
matanwerbner: first comment: the json seems to update OK, can you check if maybe it gets lost in the server somehow?
as for select2 - i will add for compute resource

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/index.js
matanwerbner: i'll try to get the chevrons in soon, i have some things on my plate prior to that, so for now i just removed the checkboxes...

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
matanwerbner: currently they will all show "hard disk" since it is taken from consts. should i implement an ajax request to fetch the new volume name from the server?

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
ohadlevy: @matanwerbner I think this is an issue as the react component is already initialized, it wont be mounted again.

afaiu steps to reproduce are:
1. have two different vmware compute resource
2. create a new host, select the first vmware, then change the compute resource to the second vmware

first vmware form will persist.

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
ohadlevy: @matanwerbner I'm OK with doing it in a separate PR, please open an issue for it.

webpack/assets/javascripts/react_app/redux/actions/hosts/storage/vmeware.consts.js
```diff
@@ -0,0 +1,14 @@
+export const defaultConrollerAttributes = {
```
ohadlevy: typo in filename, it should be vmware.const.js

webpack/assets/javascripts/react_app/constants.js
```diff
@@ -7,7 +7,12 @@ export const ACTIONS = {
   NOTIFICATIONS_EXPAND_DRAWER_TAB: 'NOTIFICATIONS_EXPAND_DRAWER_TAB',
   NOTIFICATIONS_MARK_AS_READ: 'NOTIFICATIONS_MARK_AS_READ',
   NOTIFICATIONS_MARKED_AS_READ: 'NOTIFICATIONS_MARKED_AS_READ',
-  NOTIFICATIONS_SET_REQUEST_STATUS: 'NOTIFICATIONS_SET_REQUEST_STATUS'
+  NOTIFICATIONS_SET_REQUEST_STATUS: 'NOTIFICATIONS_SET_REQUEST_STATUS',
+  CONTROLLER_ADDED: 'CONTROLLER_ADDED',
```
ohadlevy: whats the difference to the ```STORAGE_VMWARE_ADD_CONTROLLER``` const?

webpack/assets/javascripts/react_app/components/hosts/powerStatus/PowerStatus.scss
ohadlevy: TBH: based on the size of this PR, please do the restructuring in another PR that we most likely can merge quickly, then you can rebase this pr again.

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
matanwerbner: http://projects.theforeman.org/issues/18889

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
matanwerbner: @timogoebel i believe i've solved this issue, can you pull and check again?

timogoebel:
@matanwerbner : I'm super busy this week. Sorry for the delay. I do want this feature, appreciate your work and will gladly help you. I'll check this out first thing next week. Please bear with me here.

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
```diff
@@ -0,0 +1,80 @@
...
+  });
+  const selectableStores = renderOptions(datastores);
+  const selectablePods = renderOptions(storagePods);
+
+  return (
+    <div className="disk-container">
+      <TextInput
+        value={name}
+        onChange={updateDisk.bind(this, 'name')}
+        label={__('Disk name')}
```
timogoebel: This should be read-only or not even a text field at all. The name is interesting for users but can't actually be changed at all. Please see https://github.com/theforeman/foreman/pull/4252.

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
```diff
@@ -0,0 +1,80 @@
...
+    const key = Object.keys(attribute)[0];
+    const value = Object.values(attribute)[0];
+
+    return <option key={key} value={key}>{value}</option>;
+  });
+  const selectableStores = renderOptions(datastores);
+  const selectablePods = renderOptions(storagePods);
+
+  return (
+    <div className="disk-container">
```
timogoebel: Disk mode select box is missing which was recently added. Please see https://github.com/theforeman/foreman/pull/4314

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
```diff
@@ -0,0 +1,80 @@
...
+      <TextInput
+        value={name}
+        onChange={updateDisk.bind(this, 'name')}
+        label={__('Disk name')}
+      />
+
+      <Select
+        label={__('Storage Pod')}
```
timogoebel: Can we hide this if there aren't any storage pods?

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/index.js
```diff
@@ -0,0 +1,90 @@
...
+  };
+
+  return (
+    <div className="controller-container">
+      <div className="controller-header">
+        <div className="control-label col-md-2 controller-selected-container">
+          <label>{__('Create SCSI controller')}</label>
+        </div>
+        <div className="controller-type-container col-md-4">
+          <select
```
timogoebel: This should be a select2 as well when adding a new controller.

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
timogoebel: And if you select a storage pod we have to disable or hide the datastore select.
You can create a VM either on a storage pod or on a datastore. Storage pod has the higher priority. If a user selects a storage pod, vmware host creation will ignore what was selected as the datastore.

timogoebel (changes requested):
@matanwerbner : Sorry for the delay in reviewing this. All in all, it looks good, but I found some issue while testing this.

First of all I added code to pass the controller_types to react:
https://github.com/timogoebel/foreman/commit/c047ccc7980a1875be9a35018378c405ca502a80

Just squash that in if you want.

Second of all, the code does not pass the values of `datastore` or `storage_pod` in the JSON hash send via post. The fields are just empty strings. That's why vm creation fails right now.

When there are errors creating the VM, the react part of the form is reset.

When you edit a vm, it does not pick up the present values. The form should show the current data from the VM, e.g. if the VM has three volumes it should show three volumes. All fields except the size should be disabled as fog-vsphere (the library we use for vmware management) does not support editing any more parameters of a volume.

In addition, I left some comments inline.

And thanks again for doing this, first react form and a vmware form - that's not trivial 👍

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
matanwerbner: added, do you want to send the possible values the same was as controller_types?

matanwerbner:
@timogoebel controller+disk creation should work properly now, please test.
as for editing an existing controller, i'll need a way to have a controller saved on my local enviroment to work on it, instructions would help a lot

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/StorageContainer.consts.js
```diff
@@ -0,0 +1,8 @@
+export const MaxDisksPerController = 15;
+export const MaxControllers = 4;
+export const InitialSCSIKey = 1000;
+export const getDiskModeTypes = () => [{
+    'persistent': __('Independent - Persistent')},
```
timogoebel: Label should be `Persistent`.  This should also be the default (not "please select") to save some clicks. A user will change this only in rare cases. In addition, this is the VMWare default. E.g. if a user does not choose an option, it will fallback to "Independent"

timogoebel: In addition, this data should come from the ruby model. :-)

When mounting the component, you can access the data like this (untested):

```
# Simple hash
controller_types: compute_resource.disk_mode_types
# Array of hashes (or object in js-speech)
controller_types: compute_resource.disk_mode_types.map {|k, v| {k => v}}
# Might also work, gives you an array of arrays
controller_types: compute_resource.disk_mode_types.to_a
```

You can use the same for the controller_types. Map is the same as in JS. Iterates over an object (ruby object) and returns an array.

timogoebel:
```
irb(main):018:0> cr.disk_mode_types
=> {"persistent"=>"»Persistent«", "independent_persistent"=>"»Independent - Persistent«", "independent_nonpersistent"=>"»Independent - Nonpersistent«"}
irb(main):019:0> cr.disk_mode_types.to_a
=> [["persistent", "»Persistent«"], ["independent_persistent", "»Independent - Persistent«"], ["independent_nonpersistent", "»Independent - Nonpersistent«"]]
irb(main):020:0> cr.disk_mode_types.map {|k, v| {k => v}}
=> [{"persistent"=>"»Persistent«"}, {"independent_persistent"=>"»Independent - Persistent«"}, {"independent_nonpersistent"=>"»Independent - Nonpersistent«"}]
```

timogoebel (changes requested):
@matanwerbner : Thanks. This looks much better and works well.
I found some small issues. I commented inline.
One major issue is that the current state of the VM/the host's compute object is not passed to react. The component is always initialized with an empty state.
That's why a couple of things don't work. E.g. when there are failures in the queue and the page refreshes the react part is initialized empty. And compute profiles don't work. Or editing an existing VM.

timogoebel:
> as for editing an existing controller, i'll need a way to have a controller saved on my local enviroment to work on it, instructions would help a lot

@matanwerbner : I'm not sure what you need exactly. I guess you need to pass the host's compute_object data to react. This is how this can be done:

```ruby
irb(main):004:0> h.compute_object.volumes
=>   <Fog::Compute::Vsphere::Volumes
    server_id="5004cb44-baf2-668b-1b82-2335eaf5ddd9"
    [
                  <Fog::Compute::Vsphere::Volume
        id="6000C295-7559-44f7-8c5c-8a2878d94fce",
        datastore="XX",
        storage_pod=nil,
        mode="persistent",
        size=10485760,
        thin=false,
        eager_zero=nil,
        name="Hard disk 1",
        filename="[XX] agnes-intveld.example.com/agnes-intveld.example.com.vmdk",
        size_gb=10,
        key=2000,
        unit_number=0,
        controller_key=1000
      >,
                  <Fog::Compute::Vsphere::Volume
        id="6000C297-c1eb-c881-0f9b-935a37b13303",
        datastore="XX",
        storage_pod=nil,
        mode="persistent",
        size=15728640,
        thin=false,
        eager_zero=nil,
        name="Hard disk 2",
        filename="[LX_ESX_V7000_RZ1_14] agnes-intveld.example.com/agnes-intveld.example.com_1.vmdk",
        size_gb=15,
        key=2016,
        unit_number=0,
        controller_key=1001
      >,
                  <Fog::Compute::Vsphere::Volume
        id="6000C298-75a0-57e6-2226-36ea7d198039",
        datastore="XX",
        storage_pod=nil,
        mode="persistent",
        size=11534336,
        thin=false,
        eager_zero=nil,
        name="Hard disk 3",
        filename="[XX] agnes-intveld.example.com/agnes-intveld.example.com_2.vmdk",
        size_gb=11,
        key=2017,
        unit_number=1,
        controller_key=1001
      >
    ]
  >
```

and the controllers:

```ruby
irb(main):005:0> h.compute_object.scsi_controllers
=> [  <Fog::Compute::Vsphere::SCSIController
    shared_bus="noSharing",
    type="ParaVirtualSCSIController",
    unit_number=7,
    key=1000,
    server_id=nil
  >,   <Fog::Compute::Vsphere::SCSIController
    shared_bus="noSharing",
    type="ParaVirtualSCSIController",
    unit_number=7,
    key=1001,
    server_id=nil
  >]
```

If you need the data in a special way, let me know and I'll hack something together.

matanwerbner:
@timogoebel where does this code belong?

timogoebel:
@matanwerbner : I've written some code to pass more data to react, please see https://github.com/timogoebel/foreman/commit/4b21e69de79e70b3d12e1e35670f0490504397c8

matanwerbner:
@timogoebel thanks, this structure is much better.
can i ask that we format the config values in a identical way?
currently i have dataStores and storagePods as an array of objects, and diskModeTypes+controllerTypes as an array of arrays :)

timogoebel:
@matanwerbner : Sure. Does that give you what you want?

```
...
controller_types: compute_resource.scsi_controller_types.map { |k, v| {k => v}},
disk_mode_types: compute_resource.disk_mode_types.map { |k, v| {k => v}},
...
```

Should give you an array of objects.

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
```diff
@@ -0,0 +1,85 @@
...
+      {!storagePod &&
+        <Select
+          label={__('Data store')}
+          value={dataStore}
+          onChange={updateDisk.bind(this, 'dataStore')}
+          options={datastores}
+        />}
+
+      <Select
+        label={__('mode')}
```
timogoebel: Should be "Mode".

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
```diff
@@ -0,0 +1,85 @@
...
+        value={mode}
+        onChange={updateDisk.bind(this, 'mode')}
+        options={diskModeTypes}
+      />
+
+      <TextInput
+        value={size}
+        className="text-vmware-size"
+        onChange={updateDisk.bind(this, 'size')}
```
timogoebel: This should update `size_gb`.

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
```diff
@@ -0,0 +1,85 @@
...
+
+      <Select
+        label={__('mode')}
+        value={mode}
+        onChange={updateDisk.bind(this, 'mode')}
+        options={diskModeTypes}
+      />
+
+      <TextInput
+        value={size}
```
timogoebel: This should be `size_gb`. In addition the json passed to VMWare must only contain `size_gb` and not `size`. We can filter that in the rails controller later if you want.

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
```diff
@@ -0,0 +1,85 @@
...
+        />}
+      {!storagePod &&
+        <Select
+          label={__('Data store')}
+          value={dataStore}
+          onChange={updateDisk.bind(this, 'dataStore')}
+          options={datastores}
+        />}
+
+      <Select
```
timogoebel: Can we please drop the "Please select" here?

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
```diff
@@ -0,0 +1,85 @@
...
+        <Select
+          label={__('Storage Pod')}
+          value={storagePod}
+          onChange={updateDisk.bind(this, 'storagePod')}
+          options={storagePods}
+        />}
+      {!storagePod &&
+        <Select
+          label={__('Data store')}
+          value={dataStore}
```
timogoebel: When editing a host, the current value is not picked up. E.g. the select field shows "Please select".

timogoebel (changes requested):
@matanwerbner : Thanks. Some comments inline.

In addition, when editing a host we have to disable some fields (or just show text instead of a select field) :
- Storage Pod
- Datastore
- Mode
- Thin Provision
- Eager Zero

Also adding or deleting harddisks and controllers isn't supported right now. :-(

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
matanwerbner: @timogoebel we have too many naming conventions here, Mode as capital case, eagerZero as camelcase and size_gb as something else.. is there a way we can improve this? eslint gives error if variable names are not camelcased...

timogoebel: @matanwerbner : Sure. I think the main issue is that ruby and js have different naming conventions.
Foreman relies on fog-vsphere to create a VM with volumes. fog-vsphere specifies a volume model that currently has the attributes `size` and `size_gb`. They are backed by the same property.
https://github.com/fog/fog-vsphere/blob/d213ec1562d512f3d8913106d7af4ae1b93bd218/lib/fog/vsphere/models/compute/volume.rb#L26-L33

You can easily use `size` instead of `size_gb` but a user should only see the size in GB. We could also replace this via a more intelligent text field with a spinner.

timogoebel: Some sample data: https://pastebin.com/mGY25KYw

matanwerbner: @timogoebel i understand that there are two properties: size and size_gb, which one of them is editable for the user?
also, thanks for the data.

matanwerbner: @timogoebel the json data i would like to have is the one passed to the mountingService of the shape:
`{
    config: ...
    controllers: ...
   volumes: ...
}`

timogoebel: @matanwerbner : Both `size_gb` and `size` are in theory editable. These parameters change the same value. If you set `size_gb` to `10`, `size` will change as well. A user should only see the size in GB, though.

matanwerbner: @timogoebel that is a bit problematic, we should find a user friendly way to have one field that determines size, it can be presented in GB, but we shouldn't have two editable fields that represent the same field.
also, can you please provide the json data fed into the component? (see above).

timogoebel: @matanwerbner : Sorry, I do want to show only one field to the user. I was just talking about the fog-vsphere api.
I created some randomish-test data: https://gist.github.com/timogoebel/e375ab2aa4d556c9ca3e83919f667830

matanwerbner: @timogoebel this data seems fine, i can load it up and it works as expected in storybook. 
also, i've added a numeric input for size_gb, and is displaying that field only (you can omit the "size" field in the server.

timogoebel:
@matanwerbner : I tried to re-rest this. I now get an error: `Error: Cannot resolve 'file' or 'directory' ../../../../../common/forms/NumbericInput in foreman/webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk`

ohadlevy:
@matanwerbner I looked, and the current state is a rebase hell :) I've fixed most of the issues at https://github.com/ohadlevy/foreman/tree/4366 however it seems like its an old version of your code (the disk actions are emtpy?)

matanwerbner:
@ohadlevy do you mean the redux state?

matanwerbner:
@timogoebel @ohadlevy rebased and fixed the numericInput issue, please run `npm install` on the branch

timogoebel:
@matanwerbner: Thanks.

I gave this a quick spin.
I have a VM with one controller and two volumes. When I edit it, the edit form shows one controller and two volumes. That's good. However it does not pick up the data correctly.
image
I think, this is because the selects do not have a value.

Other comments from above still apply. Could you also please rebase this properly? Thanks.

ohadlevy:
@matanwerbner VMStorageActions is a left over? let me know if you need a hand re-rebasing this pr.

matanwerbner:
re-rebased, should be OK now.

@timogoebel , with the help of @ohadlevy there is a working vmware enviroment. now i have a server error when trying to create a host with one of the datastores.

parameters are: https://pastebin.com/THfMDAib

error in console is this: https://pastebin.com/GDmUAFS5

matanwerbner:
@timogoebel any news on this?

timogoebel:
@matanwerbner: Sorry for the delay. I thought you were checking this out with @ohadlevy. 

The `datastore1` values comes from [here](https://github.com/fog/fog-vsphere/blob/master/lib/fog/vsphere/requests/compute/create_vm.rb#L90). The value should be `Local-Bulgaria` as defined in the params you sent. 

Here is what you can do:
Rename the following in the json:
```
dataStore -> datastore (converted to underscore, this would be data_store. It needs to be datastore.)
size_gb -> sizeGb
controller_key -> controllerKey
```

I just noticed, `normalize_scsi_attributes` method has a `deep_transform_keys { |key| key.to_s.underscore }`. This should convert all camelcase attributes to underscore attributes.

mmoll:
seems like the last rebase went wrong?

timogoebel:
@matanwerbner : I have played with this a bit and did some changes. Could you please check them and incorporate them into the PR? Thanks.
https://github.com/timogoebel/foreman/commit/d0ab16211d6b356d48f87cbe08c5dbe1ce19d875

matanwerbner:
@timogoebel we are making progress!
i have a few issues left to resole:

1. after save, the data is returned with 'size' property and no 'sizegb' (i understand all properties are lowercased now)
2. i will need some property in config object to know that the object was already created and we are in edit mode. something like { config: { exists: true }, controllers: {...}, volumes: {...}} 
3. i do not fully understand what needs to change in edit mode, do we enable editing the different fields? removing and adding controllers and volumes? please specify.

also, please test that host creation is working as expected.

timogoebel:
@matanwerbner:

> we are making progress!

Yep. I love this already. Let's finish it up.

> after save, the data is returned with 'size' property and no 'sizegb' (i understand all properties are lowercased now)

I'll take a look. Thanks. Actually the param should be camel case with fist letter downcased -> `sizeGb`.

> i will need some property in config object to know that the object was already created and we are in edit mode. something like { config: { exists: true }, controllers: {...}, volumes: {...}}

Here you go: https://github.com/timogoebel/foreman/commit/5b827f6ef5c2808e5763bd393d8b921f7d409dfa

> i do not fully understand what needs to change in edit mode, do we enable editing the different fields? removing and adding controllers and volumes? please specify.

To summarize it quickly:
fog-vsphere does not support VM editing very well. Right now, disable everything (regarding storage) except for this:
- sizeGb

matanwerbner:
@timogoebel i've integrated the changes, and disabled the controls like requested.
after handling the 'size' field issue, please take it for a test drive and let me know what needs to improve :)

app/views/compute_resources_vms/form/_volumes.html.erb
```diff
@@ -9,7 +9,7 @@
     <%= field_set_tag _("Storage"), :id => "storage_volumes" do %>
 
       <%= f.fields_for :volumes do |i| %>
-        <%= render :partial => provider_partial(compute_resource, 'volume'), :locals => { :f => i, :compute_resource => compute_resource, :new_host => new_vm, :new_vm => new_vm, :remove_title => _('remove storage volume') }, :layout => "compute_resources_vms/form/#{item_layout}_layout" %>
+      <%= render :partial => provider_partial(compute_resource, 'volume'), :locals => { :f => i, :compute_resource => compute_resource, :new_host => new_vm, :new_vm => new_vm, :remove_title => _('remove storage volume') }, :layout => "compute_resources_vms/form/#{item_layout}_layout" %>
```
timogoebel: Please remove this indentation change.

webpack/assets/javascripts/react_app/components/common/forms/Select.js
```diff
@@ -0,0 +1,50 @@
...
+      onChange,
+      options,
+      disabled
+    } = this.props;
+
+    const innerSelect = (
+      <select disabled={disabled}
+              ref="select"
+              className="form-control" value={value} onChange={onChange}>
+        <option value="">{__('Please select')}</option>
```
timogoebel: Can we get rid of the please select?
Select 2 shows a nice "clear" button, when there is an option value without a text.

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
```diff
@@ -0,0 +1,91 @@
...
+        label={__('Disk Mode')}
+        value={mode}
+        disabled={vmExists}
+        onChange={updateDisk.bind(this, 'mode')}
+        options={diskModeTypes}
+      />
+
+      <NumericInput
+        value={sizeGb}
+        format={v => `${v} GB`}
```
timogoebel: The value is displayed like this: "55 gb". Can we make the GB actually in capital letters?

timogoebel:
@matanwerbner : Thanks. Looks really good. Sorry for the delay. Things are a little crazy right now.

To fix the missing size_gb value when editing a vm:
```diff
diff --git a/app/views/compute_resources_vms/form/vmware/_base.html.erb b/app/views/compute_resources_vms/form/vmware/_base.html.erb
index 7695f0e..46cdeb1 100644
--- a/app/views/compute_resources_vms/form/vmware/_base.html.erb
+++ b/app/views/compute_resources_vms/form/vmware/_base.html.erb
@@ -52,6 +52,6 @@ end %>
     storagePods: vsphere_storage_pods(compute_resource),
     datastores: vsphere_datastores(compute_resource)
   },
-  volumes: f.object.volumes.map { |volume| volume.attributes.deep_transform_keys { |key| key.to_s.camelize(:lower).to_sym }.reject { |k,v| k == :Size } },
+  volumes: f.object.volumes.map { |volume| volume.attributes.merge(:size_gb => volume.size_gb).deep_transform_keys { |key| key.to_s.camelize(:lower).to_sym }.reject { |k,v| k == :Size } },
   controllers: f.object.scsi_controllers
 }.to_json) %>
```

I haven't fully tested this, but I left some comments inline.

I'd like to share a screenshot, how this currently looks to me:

For a new VM:
image

For an existing VM:
image

From a user's perspective, I think the position of the  "Add Controller", "Delete Controller" and "Add Volume" button are a little confusing. And the "Delete Controller" and "Delete Volume" button aren't consistent.
@Rohoover : Do you have any comments to this? I'd love to get some feedback from you.

Oh, and this needs a rebase, now.

timogoebel (changes requested):

webpack/assets/javascripts/react_app/components/common/forms/Select.js
matanwerbner: We are using select2 for everything, please see the componentDidMount on top of this file.
select2 needs a select element to work on...

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
matanwerbner: done.

matanwerbner:
@ohadlevy do you know what these rubocop errors mean?

ohadlevy:
@matanwerbner as discussed on IRC these are due to hound upgrading their rubocop, ignore for now.

webpack/assets/javascripts/react_app/components/common/forms/Select.js
timogoebel: Try adding something like this:

<option></option>

Should give you this:

image

matanwerbner: I'm not sure i'm following...it currently looks like it should, no? https://select2.github.io/examples.html

Rohoover:
@timogoebel 
Sorry for the delay, I was on PTO.

Here is the original design prior to changes:

create controller volume

Then there were iterations:
create controller volumev2


One thought based on the above screen:
The delete buttons should be word buttons vs. one word/one icon. The icon being used isn't used in patternfly.

If we are open to additional layout changes I can iterate further. Perhaps putting the volumes in boxes would offer some visual differentiation and provide better relationships for buttons.

timogoebel:
> If we are open to additional layout changes I can iterate further. Perhaps putting the volumes in boxes would offer some visual differentiation and provide better relationships for buttons.

@Rohoover : I'd be happy to hear some more suggestions from you. 👍  I think, we're almost done with the PR. We just need some finishing touches. While testing this, I just noticed that the form isn't very intuitive to use. Maybe there is some low hanging fruit.

matanwerbner:
@timogoebel perhaps we can merge this in in basic reasonable layout, and later make style optimizations?

timogoebel:
@matanwerbner : Yep. I had the same idea over the weekend. If you rebase this, I'll test this and if nothing comes up, I'll merge. Thanks.

matanwerbner:
@timogoebel rebased it.

timogoebel:
@matanwerbner : There is one failing test. I believe, this should fix it. Haven't tested it though.

```diff
diff --git a/app/helpers/compute_resources_vms_helper.rb b/app/helpers/compute_resources_vms_helper.rb
index 52174423e..edec99c56 100644
--- a/app/helpers/compute_resources_vms_helper.rb
+++ b/app/helpers/compute_resources_vms_helper.rb
@@ -101,7 +101,7 @@ module ComputeResourcesVmsHelper
   end

   def storage_pod_stats(pod)
-    "#{pod.name} (#{_('free')}: #{number_to_human_size(pod.freespace)}, #{_('prov')}: #{number_to_human_size(pod.capacity - pod.freespace)}, #{_('total')}: #{number_to_human_size(pod.capacity)})"
+    "#{pod.name} (#{_('free')}: #{number_to_human_size(pod.freespace.to_i)}, #{_('prov')}: #{number_to_human_size(pod.capacity.to_i - pod.freespace.to_i)}, #{_('total')}: #{number_to_human_size(pod.capacity.to_i)})"
   end

   def available_actions(vm, authorizer = nil)
```

Testing this PR once I get my dev environment working again.

Rohoover:
@timogoebel 

I was working on the design and had a question. Are the volumes editable after creation?

timogoebel:
> Are the volumes editable after creation?

@Rohoover: Once the virtual machine is created (that normally happens when the host is saved for the first time), you can /only/ edit the size of the volume. Nothing more. For now, at least.

timogoebel:
@Rohoover : I took a screenshot for you to show you the "before".

New virtual machine:
image


Existing virtual machine:
image
Foreman can only change the size. All other fields are read-only. I'm wondering if we actually need form field for that.


This is how the whole form looks:

image

timogoebel:
@matanwerbner: I gave this another spin. Some comments:

1. Can we make 'persistent' the default disk mode when adding a new disk? A user should not need to choose the mode explicitly.
2. Please remove the file `app/views/compute_resources_vms/form/vmware/_volume.html.erb`.
3. Please remove the debug json.
4. To remove the "Please select" and use select2 default mode for an empty state, please change this:
```diff
diff --git a/webpack/assets/javascripts/react_app/components/common/forms/Select.js b/webpack/assets/javascripts/react_app/components/common/forms/Select.js
index 68c4309d8..38438b743 100644
--- a/webpack/assets/javascripts/react_app/components/common/forms/Select.js
+++ b/webpack/assets/javascripts/react_app/components/common/forms/Select.js
@@ -31,7 +31,7 @@ class Select extends React.Component {
       <select disabled={disabled}
               ref="select"
               className="form-control" value={value} onChange={onChange}>
-        <option value="">{__('Please select')}</option>
+        <option></option>
         {renderOptions(options)}
       </select>
     );
```

5. The `storagePod` attribute has to be named `storagePod` (capital P).

```diff
diff --git a/webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js b/webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
index 09361dc4c..df2a4365f 100644
--- a/webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
+++ b/webpack/assets/javascripts/react_app/components/hosts/storage/vmware/controller/disk/index.js
@@ -10,7 +10,7 @@ const Disk = ({
   updateDisk,
   name,
   config: { datastores, storagePods, diskModeTypes, vmExists },
-  storagepod,
+  storagePod,
   datastore,
   sizeGb,
   thin,
@@ -41,12 +41,12 @@ const Disk = ({
       {!datastore &&
         <Select
           label={__('Storage Pod')}
-          value={storagepod}
+          value={storagePod}
           disabled={vmExists}
-          onChange={updateDisk.bind(this, 'storagepod')}
+          onChange={updateDisk.bind(this, 'storagePod')}
           options={storagePods}
         />}
-      {!storagepod &&
+      {!storagePod &&
         <Select
           disabled={vmExists}
           label={__('Data store')}
```

matanwerbner:
@timogoebel I've taken care of most of what you've asked, just a couple of questions:
regarding the change in select.js, what is the reason to doing that? dont we want a friendly, localized "please select" text to be shown?
second, what do you mean by debug json?

thanks

timogoebel:
@matanwerbner
> regarding the change in select.js, what is the reason to doing that? dont we want a friendly, localized "please select" text to be shown?

Actually, we don't have that anywhere else. We always use the pattern I suggested. But maybe @Rohoover has some advice here.
It's clear button vs. a "Please select".
image

> second, what do you mean by debug json?

I mean the blue json output under the react form.

matanwerbner:
gotcha. done.

timogoebel:
@matanwerbner: Thanks. I retested this.
Some more comments:

1. When editing a VM, the size is now shown correctly. I posted the fix earlier:
```diff
diff --git a/app/views/compute_resources_vms/form/vmware/_base.html.erb b/app/views/compute_resources_vms/form/vmware/_base.html.erb
index 7695f0e92..46cdeb12b 100644
--- a/app/views/compute_resources_vms/form/vmware/_base.html.erb
+++ b/app/views/compute_resources_vms/form/vmware/_base.html.erb
@@ -52,6 +52,6 @@ end %>
     storagePods: vsphere_storage_pods(compute_resource),
     datastores: vsphere_datastores(compute_resource)
   },
-  volumes: f.object.volumes.map { |volume| volume.attributes.deep_transform_keys { |key| key.to_s.camelize(:lower).to_sym }.reject { |k,v| k == :Size } },
+  volumes: f.object.volumes.map { |volume| volume.attributes.merge(:size_gb => volume.size_gb).deep_transform_keys { |key| key.to_s.camelize(:lower).to_sym }.reject { |k,v| k == :Size } },
   controllers: f.object.scsi_controllers
 }.to_json) %>
```

2. Can we please disable the image button alongside the other form fields when editing a host?

3. The code still shows the please select. :-)

4. I can select a disk of size 0. This doesn't work. Can we prohibit this and make 1 the minimum value?
image

The rest works great! Good job.

matanwerbner:
@timogoebel requested changes have been commited, please check

timogoebel:
@matanwerbner : Thanks. Nicely done.

@ohadlevy made some minor changes in https://github.com/ohadlevy/foreman/commit/541fb96635b22f295887319123d6edd0655c5997 . Could you squash them in _except_ for the precision / decimal stuff for sizeGb? sizeGb has to be an integer. Float does not work.

Unfortunately I found another bug. When you clone a host the scsi and harddisk settings aren't cloned to the new host. :-(

Here is the fix:
```diff
diff --git a/app/models/compute_resources/foreman/model/vmware.rb b/app/models/compute_resources/foreman/model/vmware.rb
index 1b4fd6ee0..38817d624 100644
--- a/app/models/compute_resources/foreman/model/vmware.rb
+++ b/app/models/compute_resources/foreman/model/vmware.rb
@@ -553,6 +553,9 @@ module Foreman::Model
         hsh[index.to_s] = interface_attrs
         hsh
       end
+      vm_attrs[:scsi_controllers] = vm.scsi_controllers.map do |controller|
+        controller.attributes
+      end
       vm_attrs
     end
```

I also tested compute profiles. The volumes don't seem to get saved there. I'll investigate.

timogoebel:
> I also tested compute profiles.

Unfortunately, I found some more serious issues with this.
Compute profiles don't save and update correctly.
Hosts don't update correctly.

Edit: Fix posted below.

ohadlevy:
@matanwerbner I've made a few minor fixes and improvements to the storybook, can you merge https://github.com/ohadlevy/foreman/commit/e36314d2d17c66df53ecff39831aaa95298aef4b into this PR as well? thanks!

ohadlevy:
added storybook at https://rawgit.com/ohadlevy/foreman/vmware-scsi/index.html?selectedKind=Host%20VMWare%20Storage&selectedStory=default%20state%20for%20new%20host&full=0&down=1&left=1&panelRight=0&downPanel=kadirahq%2Fstorybook-addon-actions%2Factions-panel
/cc @Rohoover can you have another look at ^

timogoebel:
@matanwerbner : I have completed my fix for the compute profiles: https://github.com/timogoebel/foreman/commit/b1c47b93da7d8f57a627c3775143f997036e0034 This also contains a tiny test for the scsi controller normalization and @ohadlevy's review fixes. It does not contain ohadlevy/foreman@e36314d.

To summarize:
You need to add
ohadlevy/foreman@e36314d (needs a rebase as a new key was added to the Component's config hash in the other commit)
and
https://github.com/timogoebel/foreman/commit/b1c47b93da7d8f57a627c3775143f997036e0034

matanwerbner:
@timogoebel i'm having difficulties and errors applying your patch, can you rebase/pull from develop?
i've included @ohadlevy 's patch

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/index.js
```diff
@@ -0,0 +1,111 @@
...
+        />
+      );
+    });
+  }
+
+  render() {
+    const {
+      addController,
+      controllers,
+			volumes,
```
ohadlevy: nitpick: whitespace.

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/index.js
```diff
@@ -0,0 +1,111 @@
...
+      updateDisk,
+      removeController,
+      config,
+      volumes
+    } = this.props;
+
+    return controllers.map((controller, idx) => {
+      const controllerVolumes = volumes.filter(
+				v => v.controllerKey === controller.key
+			);
```
ohadlevy: whitespace in here and below

ohadlevy:
replaced by #4676 - thanks @matanwerbner :-)

PR #4367, opened 2017-03-09
Refs #18804 - move spinner help icons

ares:
Also the popover needs to be initialized (not sure when it was broken)

to reproduce
1) go to new host form
2) choose compute resource that allows to customize RAM e.g. libvirt
3) go to Virtual Machine tab
4) see the (i) icon as part of Memory label
5) clicking on it displays the help popover

steps 4 and 5 are fixed by this PR


dLobatog:
@ares  Thanks!

PR #4368, opened 2017-03-09
Fixes #18854 - Search host by usergroup.name

GAZDOWN:
Hi I wrote a fix for 18854. There is still slight issue though, maybe someone can point me on the part of code which should be modified as well. The code is allowing the user to filter based on usergroup.name, but after you use the operator, the autocomplete is not giving any suggestions.

### DIFF
--- a/app/models/concerns/hostext/search.rb
+++ b/app/models/concerns/hostext/search.rb
@@ -97,7 +97,7 @@ module Hostext
         scoped_search :in => :search_users,      :on => :firstname, :complete_value => true, :only_explicit => true, :rename => :'user.firstname',:operators => ['= ', '~ '], :ext_method => :search_by_user
         scoped_search :in => :search_users,      :on => :lastname,  :complete_value => true, :only_explicit => true, :rename => :'user.lastname', :operators => ['= ', '~ '], :ext_method => :search_by_user
         scoped_search :in => :search_users,      :on => :mail,      :complete_value => true, :only_explicit => true, :rename => :'user.mail',     :operators => ['= ', '~ '], :ext_method => :search_by_user
-        scoped_search :in => :search_usergroups, :on => :name,      :complete_value => true, :only_explicit => true, :rename => :'usergroup.name',:operators => ['= ', '~ '], :ext_method => :search_by_usergroup 
+        scoped_search :in => :search_usergroups, :on => :name,      :complete_value => true, :only_explicit => true, :rename => :'usergroup.name',:operators => ['= ', '~ '], :ext_method => :search_by_usergroup
       end
 
       cattr_accessor :fact_values_table_counter
@@ -130,7 +130,7 @@ module Hostext
         hosts = groups.map(&:hosts).flatten
         opts  = hosts.empty? ? "< 0" : "IN (#{hosts.map(&:id).join(',')})"
         {:conditions => " hosts.id #{opts} " }
-      end 
+      end
 
       def search_by_puppetclass(key, operator, value)
         conditions = sanitize_sql_for_conditions(["puppetclasses.name #{operator} ?", value_to_sql(operator, value)])
--- a/app/models/concerns/hostext/search.rb
+++ b/app/models/concerns/hostext/search.rb
@@ -122,13 +122,10 @@ module Hostext
       end
 
       def search_by_usergroup(key, operator, value)
-        clean_key = key.sub(/^.*\./,'')
-        key_name = Usergroup.connection.quote_column_name(clean_key)
-
-        condition = sanitize_sql_for_conditions(["#{key_name} #{operator} ?", value_to_sql(operator, value)])
-        groups = Usergroup.where(condition)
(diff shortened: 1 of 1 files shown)


dLobatog:
@xgazdi Thanks for the PR! It needs rebasing with `develop` and squashing into one commit as the bot comments say. 

I am not sure if enabling autocomplete in this field is going to work. The reason is precisely that it's implemented via an `ext_method`. If you want to try, you need to remove `:only_explicit => true` from the list, and it'd make suggestions, but I don't think it can work with external methods. 

Other than that, it generally looks good, can you write a couple of tests demonstrating the feature? 😄

app/models/concerns/hostext/search.rb
```diff
@@ -119,6 +121,17 @@ def search_by_user(key, operator, value)
...
 
+      def search_by_usergroup(key, operator, value)
+        clean_key = key.sub(/^.*\./,'')
+        key_name = Usergroup.connection.quote_column_name(clean_key)
+
+        condition = sanitize_sql_for_conditions(["#{key_name} #{operator} ?", value_to_sql(operator, value)])
+        groups = Usergroup.where(condition)
+        hosts = groups.map(&:hosts).flatten
+        opts  = hosts.empty? ? "< 0" : "IN (#{hosts.map(&:id).join(',')})"
+        {:conditions => " hosts.id #{opts} " }
```
tbrisker: Are you sure an ext_method is needed at all? 
Even if so, why not join the user table and put the condition on it (with an added condition of `owner_type=Usergroup`) instead of loading possibly tens of thousands of hosts to memory just to pull out their id's and search on that again?

tbrisker (changes requested):
Thanks! Please note comment inline, the method can be improved.

GAZDOWN:
Hello, thanks for all suggestions. Honestly I am not sure yet if I can use scoped_search without external method on this. The reason why I decided to go this way was simply the polymorphic association on owner_id column. I know the search implemented this way isn't the fastest one and I have no issue changing it, if it is possible.

app/models/concerns/hostext/search.rb
```diff
@@ -92,10 +93,11 @@ module Search
...
-        scoped_search :relation => :search_users, :on => :mail,      :complete_value => true, :only_explicit => true, :rename => :'user.mail',     :operators => ['= ', '~ '], :ext_method => :search_by_user
+        scoped_search :in => :search_users,      :on => :login,     :complete_value => true, :only_explicit => true, :rename => :'user.login',    :operators => ['= ', '~ '], :ext_method => :search_by_user,      :alias => :owner
+        scoped_search :in => :search_users,      :on => :firstname, :complete_value => true, :only_explicit => true, :rename => :'user.firstname',:operators => ['= ', '~ '], :ext_method => :search_by_user
+        scoped_search :in => :search_users,      :on => :lastname,  :complete_value => true, :only_explicit => true, :rename => :'user.lastname', :operators => ['= ', '~ '], :ext_method => :search_by_user
+        scoped_search :in => :search_users,      :on => :mail,      :complete_value => true, :only_explicit => true, :rename => :'user.mail',     :operators => ['= ', '~ '], :ext_method => :search_by_user
```
tbrisker, line 99: Note that the latest version of scoped_searched rename `:in` to `:relation`.

app/models/concerns/hostext/search.rb
```diff
@@ -92,10 +93,11 @@ module Search
...
+        scoped_search :in => :search_usergroups, :on => :name,      :complete_value => true, :only_explicit => true, :rename => :'usergroup.name',:operators => ['= ', '~ '], :ext_method => :search_by_usergroup
```
tbrisker, line 100: Since we don't search on other properties of the usergroup, it would make sense to just rename to `usergroup` instead of `usergroup.name`.

app/models/concerns/hostext/search.rb
```diff
@@ -119,6 +121,14 @@ def search_by_user(key, operator, value)
         {:conditions => " hosts.id #{opts} " }
       end
 
+      def search_by_usergroup(key, operator, value)
```
tbrisker, line 124: This  ext_method is not necessary, scoped search will know how to search by the usergroup correctly without it

tbrisker (changes requested):
@GAZDOWN Hi, looks like this needs a rebase against latest develop. Also please squash all of your commits into one. Other then that a few inline comments.

iNecas:
@GAZDOWN are you planning to get back to this PR any time soon?

GAZDOWN:
@iNecas, Hello, I am afraid, that my foreman dev. env. is gone, so I will not be able to provide the fix in any reasonable time.

tbrisker:
Thank you very much for the effort @GAZDOWN! 
I'll close this PR for now so that the issue will return to new state and others may continue work on it.
If you wish to return to this feel free to reopen the PR or create a new one.

PR #2171, opened 2015-02-17
Fixes #9410 - finish script connect to FQDN when no IP set

lzap:
Just an idea for GSS. If folks like it, I can add a unit test and try that.
UNTESTED.

PR #4370, opened 2017-03-13
Fixes #18765 - finish script use FQDN when IP is missing

lzap:
This is resurrect of

https://github.com/theforeman/foreman/pull/2171/files

The old patch did not pass review because I was unable to explain the
motivation. The reason for the fallback is that only cloud compute
resources usually provide IP addresses but virtualization do not. Our
users often associate finish templates with libvirt or VMWare but it
does not work at all. This only applies to externally managed networks
where we don't know the IP (no reservation made by Foreman).

Now there is a question of safety, with incorrect DNS setup, Foreman
could run finish script on incorrect server. Therefore I am sending this
patch for discussion - I think I could make this behavior an opt-it via
setting.

### DIFF
--- a/app/models/concerns/orchestration/ssh_provision.rb
+++ b/app/models/concerns/orchestration/ssh_provision.rb
@@ -44,7 +44,7 @@ module Orchestration::SSHProvision
   end
 
   def setSSHWaitForResponse
-    logger.info "Starting SSH provisioning script - waiting for #{provision_ip} to respond"
+    logger.info "Starting SSH provisioning script - waiting for #{provision_host} to respond"
     if compute_resource.respond_to?(:key_pair) && compute_resource.key_pair.try(:secret)
       credentials = { :key_data => [compute_resource.key_pair.secret] }
     elsif vm.respond_to?(:password) && vm.password.present?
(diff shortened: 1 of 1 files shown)


dLobatog:
I don't think there's a lot to worry here regarding DNS, there are credentials/keypair which could stop even the SSH upload before it runs if it's really the wrong server. 

How about checking `provision_interface.subnet` - see if it matches with the IP that would be used for `provision_interface.fqdn`? In such a case we could warn and stop the finish step saying the DNS record for that fqdn appears to be wrong.

cc @lzap

lzap:
I am not sure iI understand the check you recommend.

lzap:
Rebased @dLobatog can you explain? This is pretty straightforward.

lzap:
I don't think the subnet check is worth the effort, won't help in all cases.

dLobatog (approved):
@lzap Alrighty - let's merge if :green_apple:

dLobatog:
Tests passed everywhere, on 2.3 postgres it looks like there was some Jenkins related failure: 'Caused: java.io.IOException: Remote call on slave04.osuosl.theforeman.org failed'

lzap:
So you want me to rebase?

lzap:
Oh thanks :-)

ares:
setting redmine release to 1.19, please don't forget updating redmine issue after merging

PR #4371, opened 2017-03-14
refs #17463 - require test helper for smart proxy mixin

domcleal:
Tests currently fail as the module is undefined:

    test/controllers/api/v1/smart_proxies_controller_test.rb:5:in `<class:SmartProxiesControllerTest>': uninitialized constant Api::V1::SmartProxiesControllerTest::SmartProxiesControllerSharedTest (NameError)
            from test/controllers/api/v1/smart_proxies_controller_test.rb:3:in `<top (required)>'


dLobatog:
@domcleal Thanks, should've ran tests again since the PR was opened quite a while ago...

dLobatog:
Thanks @domcleal

PR #4373, opened 2017-03-14
Fixes #18892 - Do not show 'undefined method' in error message

xprazak2:


app/controllers/api/base_controller.rb
```diff
@@ -22,6 +22,11 @@ class BaseController < ActionController::Base
       render_error 'standard_error', :status => :internal_server_error, :locals => { :exception => error }
     end
 
+    rescue_from NoMethodError do |error|
+      Foreman::Logging.exception("Action failed", error)
+      render_error 'custom_error', :status => :internal_server_error, :locals => { :message => _('500 Internal Server Error') }
```
dLobatog: Please include the error message in here too - for 500 type errors, it helps to know what broke

xprazak2:
[test foreman]

app/controllers/api/base_controller.rb
dLobatog: Why is this better than showing the actual error..? The response headers already show 500, and the message shows valuable info, but in this case there is no useful info (500 is already in the response headers)

dLobatog (changes requested):
@xprazak2 Sorry, I couldn't understand the purpose, would you mind to explain a bit?

app/controllers/api/base_controller.rb
ares: I think it would be better if we displayed a better error message to user instead of some cryptic internal message. If instead of "500 Internal Server Error" the message was something like "Internal Server Error: the server was unable to finish the request, this might be caused by unavailability of some required service, incorrect API call or a bug on server side. Please find more information in server's production.log." then I'd definitely see the value.

xprazak2:
I changed the error message as suggested.

xprazak2:
[test foreman]

xprazak2:
@dLobatog, could you take a look when you have a spare moment please?

ares:
:+1: works fine. Even though the original issue was already fixed by https://github.com/theforeman/foreman/pull/3851 I think this is a good patch for future when we have a bug in a code.

@dLobatog do you want to give it another look or can I merge?

ares (approved):

dLobatog (approved):

dLobatog (commented):
👍  from me too, @GregSutcliffe / @domcleal  mind to check the error? https://github.com/theforeman/foreman/pull/4373/files#diff-b4f92b4e361ffab4fbb91c68eb3f77bbR27 To me it all reads fine except for "bug on server side" -> "server-side bug".

app/controllers/api/base_controller.rb
```diff
@@ -22,6 +22,14 @@ class BaseController < ActionController::Base
       render_error 'standard_error', :status => :internal_server_error, :locals => { :exception => error }
     end
 
+    rescue_from NoMethodError do |error|
+      Foreman::Logging.exception("Action failed", error)
+      message = _("Internal Server Error: the server was unable to finish the request. ")
+      message << _("This might be caused by unavailability of some required service, incorrect API call or a bug on server side. ")
```
GregSutcliffe: s/might/may/

I also agree with "server-side bug"

app/controllers/api/base_controller.rb
```diff
@@ -22,6 +22,14 @@ class BaseController < ActionController::Base
...
+      message << _("Please find more information in server's production.log.")
```
GregSutcliffe: `production.log` isn't correct in any other environment. Also, this sounds like an instruction (i.e. "go do this") - I presume we actually want to be passive. Perhaps "There may be more information in the server's logs"?

GregSutcliffe:
@dLobatog minor things, +1 overall

xprazak2:
I made changes as suggested.

dLobatog (changes requested):
@xprazak2 Sorry, it seems that tests are failing! 👍 in general

ares:
Tests are green now, all comments were addressed. Works well too. Thanks @xprazak2, merged as f53bd4449d9eca64781fe9420de46eb4ec57139a

PR #4376, opened 2017-03-14
refs #18760 - only build CSV URL from permitted params

domcleal:
Prevents malicious data being injected into the URL options (e.g. host)
and prevents a security exception rendering URLs with non-sanitised
params under Rails 5.0.


ohadlevy (approved):

tbrisker:
Thanks @domcleal !

PR #4377, opened 2017-03-14
Fixes #18898 - "Host" prefix isnt required

sean797:
Really tiny change, but it was annoying me and making that row double height on my screen.


tbrisker:
Test failure unrelated, merging. Thanks @sean797 !

PR #4378, opened 2017-03-15
Fixes #18901 - fix bottom notification dropdown

matanwerbner:
before:
recording

after:
recording 1


dLobatog:
Thanks @matanwerbner !

PR #4379, opened 2017-03-15
Fixes 18853 - Notification marked as read right after click

lizagilman:


tbrisker:
add to PR whitelist

webpack/assets/javascripts/react_app/API.js
```diff
@@ -30,9 +30,6 @@ export default {
       type: 'put',
       dataType: 'json',
       data: data,
-      success: function (response, textstatus, jqXHR) {
```
ohadlevy, line 33: should we keep a success event ? even if its noop ?

webpack/assets/javascripts/react_app/actions/NotificationActions.js
```diff
@@ -33,7 +33,12 @@ export default {
       expand: group
     });
   },
-  markAsRead(url) {
-    API.markNotificationAsRead(url);
+  markAsRead(id, group) {
+    API.markNotificationAsRead(`/notification_recipients/${id}`);
```
ohadlevy: should we store the API here? I think it makes more sense to include the URL in the markNotificationAsRead function and only pass the ID ?

webpack/assets/javascripts/react_app/actions/NotificationActions.js
```diff
@@ -33,7 +33,12 @@ export default {
       expand: group
     });
   },
-  markAsRead(url) {
-    API.markNotificationAsRead(url);
+  markAsRead(id, group) {
```
ohadlevy, line 36: why do we need the group? the ID should be unique across all groups?

ohadlevy:
@matanwerbner @gailsteiger can you review please?

webpack/assets/javascripts/react_app/actions/NotificationActions.js
matanwerbner: I see your point, although searching for a specific ID inside a specific group is faster then searching for an ID on all the groups...

webpack/assets/javascripts/react_app/API.js
lizagilman: it has no use so why should we keep it..?

gailsteiger: This could be a good place to do the presentation update.

Now, the update is done optimistically when the ajax call is made (`NotificationActions.markAsRead`) which is probably OK, but then there should be provision to reverse if the ajax call returns an error.

Alternatively, the update could be made here after success confirmation is received.

webpack/assets/javascripts/react_app/components/notifications/NotificationDrawer.js
```diff
@@ -24,6 +24,7 @@ class NotificationDrawer extends Component {
 
   onChange(actionType) {
     switch (actionType) {
+      case ACTIONS.MARK_AS_READ:
       case ACTIONS.RECEIVED_NOTIFICATIONS: {
         this.setState({ notifications: NotificationsStore.getNotifications() });
```
gailsteiger, line 29: fall through switch statements should not be allowed. If a single action name can't describe both cases, refactor and extract a method which can be called from each case.

webpack/assets/javascripts/react_app/components/notifications/NotificationDrawerToggle.js
```diff
@@ -23,9 +23,9 @@ class NotificationDrawerToggle extends Component {
 
   onChange(actionType) {
     switch (actionType) {
-      case ACTIONS.RECEIVED_NOTIFICATIONS: {
+      case ACTIONS.RECEIVED_NOTIFICATIONS:
+      case ACTIONS.MARK_AS_READ: {
         const notifications = NotificationsStore.getNotifications();
```
gailsteiger, line 28: fall through switch statements should not be allowed. If a single action name can't describe both cases, refactor and extract a method which can be called from each case.

webpack/assets/javascripts/react_app/constants.js
```diff
@@ -7,7 +7,8 @@ export const ACTIONS = {
   NOTIFICATIONS_EXPAND_DRAWER_TAB: 'NOTIFICATIONS_EXPAND_DRAWER_TAB',
   NOTIFICATIONS_MARK_AS_READ: 'NOTIFICATIONS_MARK_AS_READ',
   NOTIFICATIONS_MARKED_AS_READ: 'NOTIFICATIONS_MARKED_AS_READ',
-  NOTIFICATIONS_SET_REQUEST_STATUS: 'NOTIFICATIONS_SET_REQUEST_STATUS'
+  NOTIFICATIONS_SET_REQUEST_STATUS: 'NOTIFICATIONS_SET_REQUEST_STATUS',
+  MARK_AS_READ: 'MARK_AS_READ'
 };
```
gailsteiger, line 12: Why not use `NOTIFICATIONS_MARK_AS_READ` instead of creating new constant?

webpack/assets/javascripts/react_app/API.js
matanwerbner: this is exactly the purpose of this PR, to make immediate update to the state (optimistic concurrency).

webpack/assets/javascripts/react_app/components/notifications/NotificationDrawer.js
matanwerbner: i'm currently working on a PR that refactors everything, so no more switch statements in UI components :)

webpack/assets/javascripts/react_app/components/notifications/NotificationDrawerToggle.js
matanwerbner: please see above.

webpack/assets/javascripts/react_app/API.js
gailsteiger: It's fine to use the optimistic approach but server update errors should be handled.

ohadlevy:
please note the current implementation only reflected change when the timer was reached, I think if we change the state when we get the PUT success response that's OK too, but in that case we should introduce a spinner?

gailsteiger:
@ohadlevy it's really a question of how long the PUT takes. If it's long enough to need a spinner then the optimistic concurrency is best.

webpack/assets/javascripts/react_app/components/notifications/NotificationDrawer.js
gailsteiger: I presume you mean the redux conversion. It will be good to get rid of the `switch` statements. Although formally not code smells, they make me uncomfortable.

Odd that linting didn't pick up the error since we have `"no-fallthrough": "error"` specified

matanwerbner: "no-fallthrough" will error in case you forget to "return" or "break" on a case.

not in cases where multiple cases are handled the same, which is fine and even important and useful in redux.

take for example a loading spinner for the screen, that should show up when any of several actions are fired (LOGIN_REQUEST, FETCH_DATA_REQUEST).

ohadlevy:
this is now being handled by #4401 - thanks @lizagilman

PR #4380, opened 2017-03-15
Fixes #18467, #18737 - search input is not layed on top of notification drawer

lizagilman:


gailsteiger:
As a general comment, always prefer a pure css solution to a javascript solution to style issues.

The patternfly design calls for the notification layer to overlay other elements on the page. In this case, I suggest that you look into setting the `z-index` of `.navbar-outer`  to 3. That might also help with #18737.

There is a peculiar phenomenon in this case resulting from relative layout. When the screen width falls below 1000, the width of the search bar increases and overlays the navigation drawer - something that does not make sense.

Another thing to consider is that the navigation drawer will hide essential action buttons on the far right of the screen when it is displayed.

lizagilman:
@gailsteiger changed z-index as you suggested

gailsteiger:
The `.autocomplete-clear` 'x' is visible on top of the navigation drawer.

image

lizagilman:
updated
the autocomplete results box tops the notifications drawer 
should the drawer be on top of that box?

scrn_sht

ohadlevy:
@lizagilman yes

lizagilman:
@gailsteiger 
indeed helped with #18737

gailsteiger:
@lizagilman glad to know it worked for both bugs

ohadlevy:
@lizagilman if this solves two bugs, please indicate that in the commit, e.g. fixes #12345, #54321...

ohadlevy:
@matanwerbner I assume this will not conflict with #4401 ?

lizagilman:
@ohadlevy 
updated

matanwerbner:
I dont think so.

PR #4381, opened 2017-03-16
Fixes #18264 - change select multiple host actions to post request

amirfefer:
When selecting large amount of hosts for bulk action,
url is getting too long and may cause 414 error.
To avoid that, hosts_ids should send via post request.


app/assets/javascripts/host_checkbox.js
```diff
@@ -120,13 +120,12 @@ function submit_modal_form() {
...
 
 function build_modal(element, url) {
-  var url = url + "?" + $.param({host_ids: $.foremanSelectedHosts});
   var title = $(element).attr('data-dialog-title');
   $('#confirmation-modal .modal-header h4').text(title);
   $('#confirmation-modal .modal-body').empty()
     .append("<div class='modal-spinner spinner spinner-lg'></div>");
   $('#confirmation-modal').modal();
-  $("#confirmation-modal .modal-body").load(url + " #content",
+  $("#confirmation-modal .modal-body").load(url + " #content",{host_ids: $.foremanSelectedHosts},
```
ohadlevy, line 128: Could you please add a few tests for this behavior ?

amirfefer: @ohadlevy: I've added some tests.
@dLobatog: Why you removed the `Not yet reviewed` label?

amirfefer:
@ohadlevy: I've added some tests.
@dLobatog: Why you removed the `Not yet reviewed` label?

dLobatog:
@amirfefer I checked the code but didn't test it, that's why I removed it. It should be ready now with tests - merging.

dLobatog:
Thanks @amirfefer !

PR #4382, opened 2017-03-16
fixes #18929 - add rails-observers dep for TopbarSweeper

domcleal:
0443e85 was in error in removing the dependency, as this sweeper
functionality is also provided by rails-observers.


ohadlevy:
thanks @domcleal

PR #4383, opened 2017-03-16
Adds caching to notification api calls

ohadlevy:
this also removes pagination from the implementation, as that was not
implemented in the UI at all.

this implementation reduce the notification fetching time on
my production system from avg response times of 80ms to 20ms.

### DIFF
--- a/app/controllers/notification_recipients_controller.rb
+++ b/app/controllers/notification_recipients_controller.rb
@@ -5,15 +5,8 @@ class NotificationRecipientsController < Api::V2::BaseController
   before_action :find_resource, :only => [:update, :destroy]
 
   def index
-    @notifications = NotificationRecipient.
-      where(:user_id => User.current.id, :notification_id => Notification.active).
-      order(:created_at).
-      eager_load(:notification, :notification_blueprint)
-
-    render :json => {
-      :notifications => @notifications.paginate(paginate_options).map(&:payload),
-      :total => @notifications.count
-    }
+    payload = UINotifications::CacheHandler.new(User.current.id).payload
+    render :json => payload
   end
 
   def update
--- a/app/models/notification_recipient.rb
+++ b/app/models/notification_recipient.rb
@@ -7,6 +7,7 @@ class NotificationRecipient < ActiveRecord::Base
   validates :user, :presence => true
 
   scope :unseen, -> { where(:seen => false) }
+  after_save :delete_user_cache
 
   def payload
     {
@@ -23,5 +24,10 @@ class NotificationRecipient < ActiveRecord::Base
   def current_user?
     return true unless SETTINGS[:login]
     User.current.id == user_id
+  end
+  private
+
+  def delete_user_cache
+    UINotifications::CacheHandler.new(user_id).clear
   end
 end
--- a/app/services/ui_notifications/cache_handler.rb
+++ b/app/services/ui_notifications/cache_handler.rb
@@ -1,0 +1,65 @@
+module UINotifications
+  class CacheHandler
+    def initialize(user_id = User.current.try(:id))
+      raise(Foreman::Exception, 'must provide user_id') if user_id.nil?
+      @user_id = user_id
+    end
+
+    # JSON Payload
+    def payload
+      result = cache.read(cache_key)
+      if result
+        logger.debug("Cache Hit: notification, reading cache for #{cache_key}")
+        return result
+      end
+
+      result = { notifications: notifications }.to_json
+
+      logger.debug("Cache Miss: notification, writing cache for #{cache_key}")
+      cache.write(cache_key, result, { raw: true, expires_in: cache_expiry })
+
+      result
+    end
+
+    def clear
+      cache.delete(cache_key)
+    end
+
+    private
+    def logger
+      @logger ||= Foreman::Logging.logger('notifications')
+    end
+
+    delegate :cache, to: Rails
+    attr_reader :user_id
+
+    def notifications
+      @notifications ||= NotificationRecipient.
+        where(user_id: user_id, notification_id: Notification.active).
+        order(created_at: :desc).
+        preload(:notification, :notification_blueprint).
+        map(&:payload)
+    end
+
+    def cache_key
+      "notification-#{user_id}".freeze
+    end
+
+    def cache_expiry
+      next_expiry = Notification.
+        active.
+        joins(:notification_recipients).
+        where(notification_recipients: {user_id: user_id}).
+        minimum(:expired_at)
+
+      result = next_expiry.nil? ? default_cache_expiry : next_expiry - Time.now.utc
+
+      logger.debug("Expiring notification cache #{cache_key} in #{result} seconds")
+      result
+    end
+
+    def default_cache_expiry
+      1.hour
+    end
+  end
+end
--- a/test/unit/ui_notifications/cache_handler_test.rb
+++ b/test/unit/ui_notifications/cache_handler_test.rb
@@ -1,0 +1,45 @@
+require 'test_helper'
+
+class UINotificationsCacheHandlerTest < ActiveSupport::TestCase
+  test 'notification should raise without a subject' do
+    assert_raise(Foreman::Exception) { UINotifications::Base.new(nil) }
+  end
+
+  test 'should provide JSON payload' do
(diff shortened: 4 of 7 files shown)


app/controllers/notification_recipients_controller.rb
```diff
@@ -34,4 +26,27 @@ def find_resource
...
+        :total => notifications.count
+      }.to_json
+    end
+  end
+
+  def load_notifications
+    NotificationRecipient.
+      where(:user_id => User.current.id, :notification_id => Notification.active).
+      order(:created_at).
+      includes(:notification, :notification_blueprint).
```
tbrisker: You can use `.preload` or `.joins` to prevent a potentially large double left-outer-join

app/controllers/notification_recipients_controller.rb
```diff
@@ -34,4 +26,27 @@ def find_resource
...
+      {
+        :notifications => notifications,
+        :total => notifications.count
+      }.to_json
+    end
+  end
+
+  def load_notifications
+    NotificationRecipient.
+      where(:user_id => User.current.id, :notification_id => Notification.active).
```
tbrisker: Doesn't this cause an extra inner query on a table you (could be) joining in any case?

app/models/notification_recipient.rb
```diff
@@ -7,6 +7,7 @@ class NotificationRecipient < ActiveRecord::Base
   validates :user, :presence => true
 
   scope :unseen, -> { where(:seen => false) }
+  after_save :delete_user_cache
```
tbrisker, line 10: Is this the only condition that invalidates the cache?

tbrisker (commented):
:+1: in general, a few inline comments.

app/controllers/notification_recipients_controller.rb
ohadlevy: not sure i understand?

notification_id => Notification.scope adds conditions ? (expiry)

app/models/notification_recipient.rb
ohadlevy: thanks - you are correct, I need to invalidate the cache before the notification itself should expire - will fix.

app/controllers/notification_recipients_controller.rb
tbrisker: I mean this is the current query:
```sql
SELECT "notification_recipients".* 
FROM "notification_recipients" 
WHERE "notification_recipients"."user_id" = $1 
AND "notification_recipients"."notification_id" IN (
    SELECT "notifications"."id" 
    FROM "notifications" 
    WHERE (notifications.expired_at >= '2017-03-20 14:57:26.205932'))  
ORDER BY "notification_recipients"."created_at" ASC
```
but if you anyways load the notifications, it makes sense to do a join and just add another condition and reduce the nested query.

ohadlevy:
@tbrisker I've refactored the PR, Could you please re-review? thanks!

app/services/ui_notifications/cache_handler.rb
```diff
@@ -0,0 +1,65 @@
+module UINotifications
+  class CacheHandler
+    def initialize(user_id = User.current)
```
tbrisker: `User.current.id`?

app/controllers/notification_recipients_controller.rb
```diff
@@ -5,15 +5,8 @@ class NotificationRecipientsController < Api::V2::BaseController
...
-    @notifications = NotificationRecipient.
-      where(:user_id => User.current.id, :notification_id => Notification.active).
-      order(:created_at).
-      eager_load(:notification, :notification_blueprint)
-
-    render :json => {
-      :notifications => @notifications.paginate(paginate_options).map(&:payload),
-      :total => @notifications.count
-    }
+    payload = UINotifications::CacheHandler.new(User.current).payload
```
tbrisker: Shouldn't this be `User.current.id`?

app/services/ui_notifications/cache_handler.rb
```diff
@@ -0,0 +1,65 @@
...
+    def cache_key
+      "notification-#{user_id}"
+    end
+
+    def cache_expiry
+      next_expiry = Notification.
+        active.
+        joins(:notification_recipients).
+        where(notification_recipients: {user_id: user_id}).
+        minimum(:expired_at)
```
tbrisker, line 53: would it make sense to just expire the cache at the next expiring notification, regardless of the user, to simplify this lookup?

test/unit/ui_notifications/cache_handler_test.rb
```diff
@@ -0,0 +1,34 @@
...
+    expiry = UINotifications::CacheHandler.new(1).send(:cache_expiry)
+    assert_equal 1.hour, expiry
+  end
+
+  test 'should calculate cache expiry time based on the notification expiry' do
+    user = FactoryGirl.create(:user)
+    blueprint = FactoryGirl.create(:notification_blueprint, :expires_in => 3.hours,
+                              :message => 'this test just executed successfully')
+    notification = FactoryGirl.create(:notification, :notification_blueprint => blueprint, :audience => 'global')
+    FactoryGirl.create(:notification_recipient, :notification =>  notification, user_id: user.id)
```
tbrisker: NP: I'm not opposed to new-style hashes, but at least be consistent with one style per line :smile:

test/unit/ui_notifications/cache_handler_test.rb
```diff
@@ -0,0 +1,34 @@
...
+    assert_equal 'notification-1', UINotifications::CacheHandler.new(1).send(:cache_key)
+  end
+
+  test 'should use default cache expiry time' do
+    expiry = UINotifications::CacheHandler.new(1).send(:cache_expiry)
+    assert_equal 1.hour, expiry
+  end
+
+  test 'should calculate cache expiry time based on the notification expiry' do
+    user = FactoryGirl.create(:user)
```
tbrisker: Any reason to create a new user here rather then just using an existing one?

test/unit/ui_notifications/cache_handler_test.rb
```diff
@@ -0,0 +1,34 @@
...
+    blueprint = FactoryGirl.create(:notification_blueprint, :expires_in => 3.hours,
+                              :message => 'this test just executed successfully')
+    notification = FactoryGirl.create(:notification, :notification_blueprint => blueprint, :audience => 'global')
+    FactoryGirl.create(:notification_recipient, :notification =>  notification, user_id: user.id)
+    expiry = UINotifications::CacheHandler.new(user.id).send(:cache_expiry)
+    assert_equal 0, (3.hour - expiry).round
+  end
+
+  test 'should use cache if it exists' do
+  end
```
tbrisker: So does it use cache if it exists?

tbrisker (changes requested):
Looks better, a few inline comments.

app/services/ui_notifications/cache_handler.rb
ohadlevy: IMHO no, why are you concerned about this query?

ohadlevy:
@tbrisker I think I've addressed all of your comments..
bart-simpson-chalkboard-wallpaper-generator

app/services/ui_notifications/cache_handler.rb
tbrisker: missed this one

app/services/ui_notifications/cache_handler.rb
tbrisker: Might be premature optimization, but if the tables grow significantly this might become a painful join for every cache miss.

ohadlevy:
@tbrisker updated again, also added .freeze on cache key and try in case User.current is nil.

app/services/ui_notifications/cache_handler.rb
```diff
@@ -0,0 +1,65 @@
...
+    def logger
+      @logger ||= Foreman::Logging.logger('notifications')
+    end
+
+    delegate :cache, to: Rails
+    attr_reader :user_id
+
+    def notifications
+      @notifications ||= NotificationRecipient.
+        where(:user_id => user_id, :notification_id => Notification.active).
```
ohadlevy: I'll update the hash rocket syntax after tests will pass/fail.

ohadlevy:
@tbrisker test failures seems unrelated.

tbrisker:
Thanks @ohadlevy !

PR #4385, opened 2017-03-17
Fixes #18949 - user information in logger

iNecas:
After the change, the log messages look like this:

```
2017-03-17T14:36:06 23eeb21c [user:] [app] [D] Setting current user thread-local variable to skoroadmin
2017-03-17T14:36:06 23eeb21c [user:skoroadmin] [sql] [D]   Usergroup Load (0.2ms)  SELECT "usergroups".* FROM "usergroups" INNER JOIN "cached_usergroup_members" ON "usergroups"."id" = "cached_usergroup_members"."usergroup_id" WHERE "cached_usergroup_members"."user_id" = $1  ORDER BY usergroups.name  [["user_id", 4]]
2017-03-17T14:36:06 23eeb21c [user:skoroadmin] [sql] [D]   Role Load (0.2ms)  SELECT DISTINCT "roles".* FROM "roles" INNER JOIN "cached_user_roles" ON "roles"."id" = "cached_user_roles"."role_id" WHERE "cached_user_roles"."user_id" = $1  [["user_id", 4]]
2017-03-17T14:36:06 23eeb21c [user:skoroadmin] [sql] [D]    (0.2ms)  SELECT permissions.name FROM "permissions" INNER JOIN "filterings" ON "permissions"."id" = "filterings"."permission_id" INNER JOIN "filters" ON "filterings"."filter_id" = "filters"."id" WHERE "filters"."role_id" = $1  ORDER BY filters.role_id, filters.id  [["role_id", 3]]
```

### DIFF
--- a/app/models/concerns/foreman/thread_session.rb
+++ b/app/models/concerns/foreman/thread_session.rb
@@ -27,11 +27,12 @@ module Foreman
       def clear_thread
         if Thread.current[:user] && !Rails.env.test?
           Rails.logger.warn("Current user is set, but not expected. Clearing")
-          Thread.current[:user] = nil
+          User.current = nil
         end
         yield
       ensure
-        [:user, :organization, :location].each do |key|
+        User.current = nil
+        [:organization, :location].each do |key|
           Thread.current[key] = nil
         end
       end
--- a/app/models/concerns/foreman/thread_session.rb
+++ b/app/models/concerns/foreman/thread_session.rb
@@ -52,6 +52,11 @@ module Foreman
           end
 
           Rails.logger.debug "Setting current user thread-local variable to " + (o.is_a?(User) ? o.login : 'nil')
+          if o.is_a? User
+            ::Logging.mdc['user'] = o.login
+          else
+            ::Logging.mdc.delete('user')
+          end
           Thread.current[:user] = o
         end
 
--- a/app/models/concerns/foreman/thread_session.rb
+++ b/app/models/concerns/foreman/thread_session.rb
@@ -53,7 +53,7 @@ module Foreman
 
           Rails.logger.debug "Setting current user thread-local variable to " + (o.is_a?(User) ? o.login : 'nil')
           if o.is_a? User
-            ::Logging.mdc['user'] = o.login
+            ::Logging.mdc['user'] = "[user:#{o.login}]"
           else
             ::Logging.mdc.delete('user')
           end
--- a/config/logging.yaml
+++ b/config/logging.yaml
@@ -47,6 +47,7 @@
 #     [%M] method name where the logging request was issued
 #     [%X{request}] request-ID set in HTTP headers, or random UUID
 #     [%X{session}] session ID from cookies, else the request-ID
+#     [%X{user}] information about current user
 #     [%X{string}] variable set using ::Logging.mdc['string'] =
 
 :default:
@@ -55,7 +56,7 @@
   :log_trace: false
   :level: info
   :type: file
-  :pattern: "%d %.8X{session} [user:%X{user}] [%c] [%.1l] %m\n"
+  :pattern: "%d %.8X{session} %X{user} [%c] [%.1l] %m\n"
 
 :production:
   :filename: "production.log"
--- a/config/logging.yaml
+++ b/config/logging.yaml
@@ -55,7 +55,7 @@
   :log_trace: false
   :level: info
   :type: file
-  :pattern: "%d %.8X{session} [%c] [%.1l] %m\n"
+  :pattern: "%d %.8X{session} [user:%X{user}] [%c] [%.1l] %m\n"
 
 :production:
   :filename: "production.log"


iNecas:
I've tweaked this a bit to not show the user info when the user is not set in f74d6a2eac15d2c2c47e276115ec6be6282edabe

ares:
I think this is a good and useful addition. I'd prefer to keep information as small as possible, so I'd suggest `@skoroadmin` instead of `[user:skoroadmin]` and don't print the info if no user is logged in. But I don't have strong opinion. For me it would be helpful when I see the trace and I'm not sure whether the action was run under admin or regular user. Knowing it was not admin gives me pointers to taxonomies or permissions problems and helps reproducing quickly...

ares:
Actually how about `u_regular_user` and `a_admin` for users that are admins?

iNecas:
The prefix looks a bit cryptic to me, what about `regular_user`, vs `admin!`?

Btw. there is one issue with the proposed solution: the info about the user get logged, but unfortunately, we don't have this information at the time the parameters are being logged, which leads to this kind of logs.
```
2017-03-17T15:13:50 23eeb21c  [app] [I] Started GET "/" for 127.0.0.1 at 2017-03-17 15:13:50 +0100
2017-03-17T15:13:50 23eeb21c [app] [I]   Parameters: {}
2017-03-17T15:13:50 23eeb21c  [app] [I] Processing by DashboardController#index as HTML
2017-03-17T15:13:50 23eeb21c [user:skoroadmin] [app] [I]   Rendered dashboard/index.html.erb within layouts/application (88.7ms)
2017-03-17T15:13:50 23eeb21c [user:skoroadmin] [app] [I]   Rendered common/_searchbar.html.erb (5.2ms)
2017-03-17T15:13:50 23eeb21c [user:skoroadmin] [app] [I]   Rendered layouts/_application_content.html.erb (6.3ms)
2017-03-17T15:13:51 23eeb21c [user:skoroadmin] [app] [I]   Rendered home/_user_dropdown.html.erb (5.6ms)
2017-03-17T15:13:51 23eeb21c [user:skoroadmin] [app] [I]   Rendered home/_organization_dropdown.html.erb (4.7ms)
2017-03-17T15:13:51 23eeb21c [user:skoroadmin] [app] [I]   Rendered home/_location_dropdown.html.erb (5.2ms)
2017-03-17T15:13:51 23eeb21c [user:skoroadmin] [app] [I]   Rendered home/_org_switcher.html.erb (11.1ms)
2017-03-17T15:13:51 23eeb21c [user:skoroadmin] [app] [I]   Rendered home/_submenu.html.erb (3.1ms)
2017-03-17T15:13:51 23eeb21c [user:skoroadmin] [app] [I]   Rendered home/_submenu.html.erb (3.4ms)
2017-03-17T15:13:51 23eeb21c [user:skoroadmin] [app] [I]   Rendered home/_submenu.html.erb (2.0ms)
2017-03-17T15:13:51 23eeb21c [user:skoroadmin] [app] [I]   Rendered home/_submenu.html.erb (1.3ms)
2017-03-17T15:13:51 23eeb21c [user:skoroadmin] [app] [I]   Rendered home/_submenu.html.erb (3.0ms)
2017-03-17T15:13:51 23eeb21c [user:skoroadmin] [app] [I]   Rendered home/_topbar.html.erb (135.9ms)
2017-03-17T15:13:51 23eeb21c [user:skoroadmin] [app] [I]   Rendered layouts/base.html.erb (768.2ms)
2017-03-17T15:13:51 23eeb21c  [app] [I] Completed 200 OK in 987ms (Views: 874.0ms | ActiveRecord: 16.0ms)
2017-03-17T15:13:53 23eeb21c  [app] [D]
2017-03-17T15:13:53 23eeb21c  [app] [D]
```

Ideally I wanted to have the user already, when printing the info about the request itself but that doesn't seem to be feasible. Perhaps easier would be just to print the info message about the currently logged user, where we could be also more verbose about the fact, is the user is admin or regular one.

iNecas:
So it would be

```
2017-03-17T15:13:50 23eeb21c [app] [I] Started GET "/" for 127.0.0.1 at 2017-03-17 15:13:50 +0100
2017-03-17T15:13:50 23eeb21c [app] [I] Processing by DashboardController#index as HTML
2017-03-17T15:13:50 23eeb21c [app] [I]   Parameters: {}
2017-03-17T15:13:50 23eeb21c [app] [I] Current user: skoroadmin (regular user)
```

ares:
sounds good to me (or just don't print regular user and only print admin! if it's admin)

iNecas:
I've opened a simple PR that would also help us with both getting the information for admin to know who is performing what on their system as well as for us to easier identify, if there might be issue with permissions (adding the adminsitrator vs. current user info) https://github.com/theforeman/foreman/pull/4393

lzap:
I have a concern from the security sandpoint, username is valuable asset and we should not it give away like that (in logs). At least not in implicit configuration. I suggest to replace this with user ID, I know this is less comfortable when reading logs but security is security. The idea is to lookup the username in database when needed.

iNecas:
We already had the presence of user login in logs for a while: 

```
Parameters: {"utf8"=>"✓", "login"=>{"login"=>"admin", "password"=>"[FILTERED]"}, "commit"=>"Log In"}
```
Another thing is the login doesn't seem to be that secret information in corporate world, especially with ldap in place, I don't think anyone should assume a login to be a secret.

If we wanted to be to precise, we should do filtering on all places but I would not got for it until we have another reliable mechanism to get the data that we have right now only in logs.

lzap:
Yeah the difference is this is with every request. But generally I don't agree with logging form parameters each request in Rails, I had a discussion about it and filed PR, closed with no discussion ;-) I believe this should be DEBUG. Username is considered a sensitive information for sure. You know what, let me fix Rails first, I will file a PR to filter out username from Parameters info message and let's see how it goes.

PR #4388, opened 2017-03-19
Fixes #18762 - Add csv export to reports table

tbrisker:


app/controllers/config_reports_controller.rb
```diff
@@ -1,10 +1,20 @@
...
 
   def index
-    @config_reports = resource_base_search_and_page(:host)
+    respond_to do |format|
+      format.html do
+        @config_reports = resource_base_search_and_page(:host)
+        render :index
+      end
+      format.csv do
+        @config_reports = resource_base_with_search.preload(:host)
```
ohadlevy, line 14: can you please add a test for this endpoint? thanks.

tbrisker: How would it be different then the test for the csv responder?

ohadlevy: it actually tests the fact the server response and go though this controller ? this will catch any issues with the columns or other customization?

tbrisker: I believe so - take a look at https://github.com/theforeman/foreman/blob/develop/test/controllers/concerns/csv_responder_test.rb

ohadlevy: @tbrisker this still means that if there are any customization on the controller (e.g. logic to define csv_columns) they will not be covered by a test (not the actual SQL query from the server etc). or am I missing something?

tbrisker: I'm not sure I understand, what kind of test exactly do you feel is missing here?

tbrisker:
@ohadlevy added test.

ohadlevy (approved):

ohadlevy:
@tbrisker tests are failing, i wonder if this needs a rebase?

tbrisker:
@ohadlevy I think you are confused with the facts csv pr, this one is passing.

ohadlevy:
@tbrisker thanks, test failure is indeed unrelated.

PR #4390, opened 2017-03-19
Fixes #18954 - Add CSV export to facts table

tbrisker:


ohadlevy:
just by logging into the facts view I get:
```
FactValue Load (68885.7ms)  SELECT  "fact_values".* FROM "fact_values" INNER JOIN "fact_names" ON "fact_names"."id" = "fact_values"."fact_name_id" WHERE "fact_names"."ancestry" IS NULL AND (fact_names.name <> '_timestamp')  ORDER BY "fact_values"."value" ASC NULLS FIRST  LIMIT 20 OFFSET 0
```

ohadlevy:
when using it with discovery plugin i get major N+1: (or even N^2)

```
 SELECT  "hosts".* FROM "hosts" WHERE "hosts"."type" IN ('Host::Discovered') AND "hosts"."id" = $1  ORDER BY hosts.created_at DESC LIMIT 1  [["id", 449]]
```

app/models/fact_value.rb
```diff
@@ -17,21 +17,20 @@ class FactValue < ActiveRecord::Base
...
-    eager_load(:fact_name).where("fact_names.name = ?",:_timestamp)
+    joins(:fact_name).where("fact_names.name = ?",:_timestamp)
   }
   scope :my_facts, lambda {
     if !User.current.admin? || Organization.expand(Organization.current).present? || Location.expand(Location.current).present?
       joins_authorized(Host, :view_hosts, :where => Host.taxonomy_conditions)
     end
   }
 
-  scope :required_fields, -> { includes(:host, :fact_name) }
```
tbrisker, line 31: This name is misleading, and not possible to extend in plugins.

tbrisker:
@ohadlevy changed the way we preload the related tables so that they can be overriden in discovery to add the `discoverd_host` relation and prevent the n+1

app/controllers/fact_values_controller.rb
```diff
@@ -24,12 +25,30 @@ def index
       values = values.root_only
     end
 
-    @fact_values = values.no_timestamp_facts.required_fields.paginate :page => params[:page]
+    @fact_values = values.no_timestamp_facts
+
+    respond_to do |format|
+      format.html do
+        @fact_values = @fact_values.preload(related_tables).paginate :page => params[:page]
```
tbrisker, line 32: for a small size (20 rows by default) preloading the related tables is faster then joining.

app/controllers/fact_values_controller.rb
```diff
@@ -24,12 +25,30 @@ def index
...
+        render :index
+      end
+      format.csv do
+        csv_response(@fact_values.joins(related_tables).includes(related_tables))
```
tbrisker, line 36: `joins` is needed to force an inner join instead of a left outer join in the `includes`. When pulling in 1000 records at a time, using an inner join is faster then then preloading.

app/models/fact_value.rb
```diff
@@ -2,7 +2,7 @@ class FactValue < ActiveRecord::Base
   include Authorizable
   include ScopedSearchExtensions
 
-  belongs_to_host
+  belongs_to :host, {:class_name => "Host::Base", :foreign_key => :host_id}
```
tbrisker: Facts are connected to Host::Base. This will allow removing a workaround in discovery that causes n+1 queries on fact pages.

app/services/csv_exporter.rb
```diff
@@ -5,8 +5,10 @@ def self.export(resources, columns)
     Enumerator.new do |csv|
       csv << csv_header(columns)
 
-      resources.reorder(nil).limit(nil).find_each do |obj|
-        csv << CSV.generate_line(columns.map{|c| obj.send(c)})
+      resources.uncached do
```
tbrisker, line 8: uncached is needed to prevent memory leaking when processing a huge amount of records - this disables `ActiveRecord::QueryCache` for the loop, preventing rails from trying to cache potentially millions of rows.

tbrisker (commented):
Updated with a solution for memory leak and correct optimization for both index and csv.

ohadlevy:
@tbrisker :+1: please add a simple functional test that checks the CSV endpoint too.

tbrisker:
Now with 10% more test coverage!

ohadlevy (approved):

tbrisker:
Looks like tests are failing because `Host::Base` doesn't have a hostgroup. Will continue investigation tomorrow.

ohadlevy:
@tbrisker does it make sense to split this pr into two? one for N+1 and another to export CSV?

tbrisker:
@ohadlevy yes, I will do that, as it appears that fixing the n+1 is more complex then I thought at first.

ohadlevy:
Ack pending Jenkins.

ohadlevy:
Any idea why the test failed?

tbrisker:
@ohadlevy looks like unrelated intermittent failure

ohadlevy:
Thanks @tbrisker

PR #4386, opened 2017-03-17
fixes #18950 - enable password field in CR form test

domcleal:
Prior to capybara 2.13.0, the test passed despite filling in a disabled
password field, but a change now prevents this odd assumption. The test
now enables the field via the provided button, also requiring JS.

PR #4393, opened 2017-03-21
Fixes #18949 - info message about current user

iNecas:


iNecas:
Follow-up from https://github.com/theforeman/foreman/pull/4385

dLobatog:
@iNecas Looks good to me - integration tests are failing though, I'll look into it and check if this PR is causing it or there is another reason.

dLobatog:
Seems that this is happening because your branch doesn't have https://github.com/theforeman/foreman/pull/4386/files - merging

dLobatog:
Thanks @iNecas !

PR #4394, opened 2017-03-21
refs #18760 - exclude test routes from permission checks

domcleal:
Running test/controllers/concerns/csv_responder_test.rb and
access_permissions_test in sequence (e.g. in `rake test`) would cause
failures against the "fake" routes introduced by CsvResponderTest as no
permissions were defined.


tbrisker:
Thanks @domcleal !

PR #4395, opened 2017-03-22
fixes #18976 - separate tests with different request formats

domcleal:
When making a JSON format request followed by a request with no explicit
format, Rails 5 is now memorising that it's requesting application/json
instead of resetting to HTML or similar. It shouldn't be assumed that
the state is reset between requests within a test (rails/rails@33db421).


dLobatog:
Interesting, I did not know about this. Thanks @domcleal !

PR #4396, opened 2017-03-22
Fixes #18977 - Do not raise on creating a role in plugin initialization

adamruzicka:
As an alternative to https://github.com/theforeman/foreman-packaging/pull/1561

### DIFF
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -125,7 +125,7 @@ class Role < ActiveRecord::Base
     search = options.delete(:search)
 
     collection = Permission.where(:name => permissions).all
-    raise ArgumentError, 'some permissions were not found' if collection.size != permissions.size
+    raise ::Foreman::PermissionMissingException.new(N_('some permissions were not found')) if collection.size != permissions.size
 
     collection.group_by(&:resource_type).each do |resource_type, grouped_permissions|
       filter = self.filters.build(:search => search)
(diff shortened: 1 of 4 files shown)


ehelms:
Might make more sense to just check if we are in the migration step via `Foreman.in_rake?('db:migrate')` at https://github.com/theforeman/foreman/pull/4396/files#diff-d0644101ad428cdee084ab59f3e9a43cR238

I'd worry this rescue would hide other issues instead of targeting the current specific scenario.

iNecas:
Good point @ehelms !

adamruzicka:
done

iNecas:
So `rake plugin:apipie:cache[%%{gem_name}]` would not hit the issue (https://github.com/theforeman/foreman-packaging/pull/1561/files#diff-67e0a6165416d559ed233a03cbf4dbc1R866) after this change?

ares:
I don't think we should limit ourselves to `Foreman.in_rake?('db:migrate')`. There are other tasks that would fail such as `apipie:cache`, `assets:precompile` etc. I think we should avoid explicit list. I'd rather do the rescue, check whether the cause was missing permission (needs changing ArgumentError to some specific exception) and then check whether we are in any rake task or testing env. If all of that is true, we should print warning but continue working, otherwise we should fail with explaining message like "Can not continue since some permissions were not found, please run rake db:seed and retry"

iNecas:
Sounds good, with one update: I would not touch the original condition that is there (as it could have consequences on the tests, where we don't assume the permissions are seeded this way). +1 for new execption class and catching/logging that in case we are in rake (without specifying which rake it is), otherwise re-raising

iNecas (approved):
Approving the change. Will merge unless @ehelms has something to say

iNecas:
Pending tests failures

iNecas:
Test failures independent, squashing and merging now

PR #4397, opened 2017-03-22
refs #18950 - wait for CR test connection to complete

domcleal:
Should fix intermittent errors such as:

<pre>
ActionController::RoutingError: No route matches [PUT] "/locations"
    lib/middleware/tagged_logging.rb:18:in `call' (ActionController::RoutingError)
/usr/local/rvm/gems/ruby-2.0.0-p481@test_develop_pr_core-1/gems/actionpack-4.2.8/lib/action_dispatch/middleware/debug_exceptions.rb:21
</pre>

Can be reproduced by adding a short (2s) sleep to the end of the test, and a longer sleep prior to the `render` in `ComputeResourcesController#test_connection`, so the DB is cleared while the response is being rendered.


dLobatog:
I couldn't reproduce the error I see on jenkins - however the one in the instructions is fixed. Merging

dLobatog:
Thanks @domcleal !

PR #4399, opened 2017-03-22
fixes #18982 - replace AC::Params#each using one-arg block

domcleal:
ActionController::Parameters#each works differently in Rails 5.0: it no
longer yields an array of [key, value] entries for hashes, only the
key when given a block with arity of one. This method now iterates over
only the values when passed a hash or AC::Parameters instance.

### DIFF
--- a/app/controllers/concerns/api/compatibility_checker.rb
+++ b/app/controllers/concerns/api/compatibility_checker.rb
@@ -4,9 +4,11 @@ module Api
 
     # removes unsupported "nested" flag from "host_parameters_attributes" (both Array and Hash formats supported)
     def check_create_host_nested
-      params[:host] && params[:host][:host_parameters_attributes] && params[:host][:host_parameters_attributes].each do |attribute|
-        attribute = attribute[1] if attribute.is_a? Array
-        Foreman::Deprecation.api_deprecation_warning("Field host_parameters_attributes.nested ignored") unless attribute.delete(:nested).nil?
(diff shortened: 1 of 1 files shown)


app/controllers/concerns/api/compatibility_checker.rb
```diff
@@ -4,9 +4,11 @@ module CompatibilityChecker
 
     # removes unsupported "nested" flag from "host_parameters_attributes" (both Array and Hash formats supported)
     def check_create_host_nested
-      params[:host] && params[:host][:host_parameters_attributes] && params[:host][:host_parameters_attributes].each do |attribute|
-        attribute = attribute[1] if attribute.is_a? Array
-        Foreman::Deprecation.api_deprecation_warning("Field host_parameters_attributes.nested ignored") unless attribute.delete(:nested).nil?
+      if params[:host] && params[:host][:host_parameters_attributes] && (attrs = params[:host][:host_parameters_attributes])
```
tbrisker: Why not `if params[:host] && (attrs = params[:host][:host_parameters_attributes])` ?

tbrisker (changes requested):
Just a minor question inline, other than that seems fine.

app/controllers/concerns/api/compatibility_checker.rb
domcleal: Thanks, PR updated with the superfluous condition removed.

tbrisker (approved):

tbrisker:
Thanks @domcleal !

PR #4400, opened 2017-03-22
Fixes #14160 - Allow boot order to be changed

chris1984:


app/views/compute_resources_vms/form/vmware/_base.html.erb
```diff
@@ -23,6 +23,7 @@ end %>
 <%= checkbox_f f, :memoryHotAddEnabled, { :help_inline => _("Memory hot add lets you add memory resources for a virtual machine while the machine is powered on."), :label => _('Memory hot add'), :label_size => "col-md-2", :disabled => !new_host } %>
 <%= checkbox_f f, :cpuHotAddEnabled, { :help_inline => _("CPU hot add lets you add CPU resources for a virtual machine while the machine is powered on."), :label => _('CPU hot add'), :label_size => "col-md-2", :disabled => !new_host } %>
 <%= checkbox_f f, :add_cdrom, { :help_inline => _("Add a CD-ROM drive to the virtual machine."), :label => _('CD-ROM drive'), :label_size => "col-md-2" } if new_host %>
+<%= checkbox_f f, :boot_order, { :help_inline => _("Use default Foreman behaivor to lock boot order on creation"), :label => _('Hardcode boot order'), :label_size => "col-md-2", :disabled => !new_host } %>
```
ekohl: behaivor is most likely a typo.

timogoebel:
@chris1984 : Thanks. I got it working in https://github.com/timogoebel/foreman/commit/3fe7014de4f76921b2f2ec36606146fcfe57084d.

I'd suggest that instead of the checkbox we add a sortable multiselect, where a user can actually change the boot order. I think, that makes more sense. There are probably users who want to first boot off cdrom, then network and then disk. Or any other combination.

Something like this:
http://select2.github.io/select2/#programmatic  -> "Select2 Drag and Drop Sorting"

Should not be hard to do. What do you think, @chris1984 ?

chris1984:
@timogoebel planning to revisit this, how hard is the jquery to add the webui select box? I have never did anything with the webui before.

timogoebel:
> how hard is the jquery to add the webui select box?

I shouldn't be too difficult I think. We don't want any magic, just a rather default widget. I'd have to do some research to actually know what needs to be done.

chris1984:
@timogoebel can we do a paired programming session then or something, or when you get time can you send me an example. I really am not familiar with jquery at all. I can try to work with one of the guys here using your existing code for the other 2 areas if you want?

timogoebel:
@chris1984 : Some wip code: https://github.com/timogoebel/foreman/commit/94c097e8fe42a37601ee22c1e9a3a04ca1119c65

Issues I have found so far:
- Data gets lost when saving the form with errors
- default values don't work
- jQuery is always messy
- no drag & drop sorting
- I have no idea if the data is passed correctly to vmware. Haven't tried provisioning a vm, yet.

chris1984:
@timogoebel the form looks to be working better but when we use the default foreman style we get an error with boot order being "" invalid and when we use the boot options it does not reflect, but at least the checkbox is staying checked now.

app/models/compute_resources/foreman/model/vmware.rb
```diff
@@ -32,6 +32,14 @@ def capabilities
       [:build, :image]
     end
 
+    def boot_devices
+       {
+         :disk => 'Harddisk',
+         :cdrom => 'CD-ROM',
+         :network => 'Network'
```
timogoebel, line 39: for the sake of completeness we might want to add:
```
:floppy => 'Floppy'
```

app/views/compute_resources_vms/form/vmware/_base.html.erb
```diff
@@ -23,6 +23,8 @@ end %>
 <%= checkbox_f f, :memoryHotAddEnabled, { :help_inline => _("Memory hot add lets you add memory resources for a virtual machine while the machine is powered on."), :label => _('Memory hot add'), :label_size => "col-md-2", :disabled => !new_vm } %>
 <%= checkbox_f f, :cpuHotAddEnabled, { :help_inline => _("CPU hot add lets you add CPU resources for a virtual machine while the machine is powered on."), :label => _('CPU hot add'), :label_size => "col-md-2", :disabled => !new_vm } %>
 <%= checkbox_f f, :add_cdrom, { :help_inline => _("Add a CD-ROM drive to the virtual machine."), :label => _('CD-ROM drive'), :label_size => "col-md-2" } if new_vm %>
+<%= checkbox_f f, :manage_boot_order, { :help_inline => _("Use default Foreman behavior to lock boot order on creation"), :label => _('Hardcode boot order'), :label_size => "col-md-2", :disabled => !new_vm } %>
```
timogoebel, line 26: Is this really needed? If the Boot Order select is empty, it will not enforce any boot order and use VMWare's defaults.

timogoebel:
> when we use the default foreman style we get an error with boot order being "" invalid

@chris1984: I guess, you mean this one:
` RuntimeError: failed to create boot device because "" is unknown` 

This is because the request that is sent to the server is `"boot_order"=>["", "cdrom"],`, when querying select2's value via js console i get the following:

```
$("#host_compute_attributes_boot_order").val();
["cdrom"]
```
I don't have the time to trace this right now. 
Maybe it's this hidden input field:
image
Where does that come from? It stays there even if I deactivate select2: `$('select').select2("destroy");`

dLobatog (changes requested):
Moving to WoC because of the questions by @timogoebel

chris1984:
@timogoebel after talking with @Klaas- it seems there is an issue with sending the correct boot order where VMware will not honor it, it looks to be an issue with VMware he said. Are you fine with going back to the checkbox idea of checked do default Foreman behavior right now and if unchecked just don't send anything so the user can edit the bios boot order in the VM console?

timogoebel:
> Are you fine with going back to the checkbox idea of checked do default Foreman behavior right now and if unchecked just don't send anything so the user can edit the bios boot order in the VM console?

I don't like this at all and I'm not aware of such an issue with VMWare. It's definitely not a clean solution imho. @Klaas-: Mind to explain where you think that there is a bug in VMWare? I don't get it. Is there a VMWare KB for this?

Klaas-:
@timogoebel  the (suspected) bug is within vmware when you force boot order via the vmx file. This is basically what we do at the moment through api. It seems once you force the bootorder you can't do things like deleting a disk (via web client, fat client seems to ignore this problem) or using those vms as templates (what the original issue is about if I understood it correctly). I've reported this issue to vmware but support is not really sure if this is supported by my licenses ... :) In any case I've also involved the fog-vsphere maintainer. He seems to agree to my assessment that this is a bug. Of cause this has nothing to do with what this PR has evolved to - making the boot order configurable is a nice new feature but not setting the bootorder is not really solving the issue that is at the hearth of this.

Klaas-:
plus if we don't set boot order certain things will stop working - like rebuilding a host. so if we work around the issue and make setting the boot order configurable we should warn if not putting network first there or explicitly set boot order while in build mode only (not sure if that is actually supported by vmware though)

timogoebel:
@Klaas-: Thanks for the explanation. I still don't really believe in a checkbox to just disable "bootorder hardcoding". From a user's experience, this still seems like the wrong approach.

chris1984:
@timogoebel I am happy to do w/e will get the PR Merged :) since it will close 3 downstream bugs I have assigned to me. Do you still want to go with the drag and drop or something different?

timogoebel:
@chris1984: I had a talk with @Klaas- earlier today. I still don't fully get the suspected bug to he honest. What is the expected behavior and the actual behavior happening here?

One known restriction with this approach is:
When you set the boot order via foreman, you can't change it in the VM's bios anymore.

@Klaas- said that you cannot delete a hard disk anymore.
That is true. I can confirm this.

Is there anything else?

chris1984:
@timogoebel ok thanks, I will keep trying with what you proposed and send in an updated commit for your review and possible help to debug further.

bkearney:
@chris1984 whats the next step?

chris1984:
@timogoebel do you have some more time to help me with this?

timogoebel:
@chris1984: If you want to continue working on this, please use a different Redmine issue.

chris1984:
@timogoebel I noticed you delete this PR from the upstream issue, since you are working on this should I close this one out?

chris1984:
@timogoebel ignore i just posted that right after you sent that and the page didnt refresh.... ack will do

Klaas-:
@chris1984 the upstream bug is not about changing the bootorder, its about not being able to clone images with the bootorder set. The workaround in fog-vsphere doesn't enable changing of bootorder.

Emil-G:
@chris1984 @timogoebel  Was this prematurely closed? https://bugzilla.redhat.com/show_bug.cgi?id=1398317

Klaas-:
@Emil-G it depends on what you want to achieve; the this PR was abandoned if I recall correctly because the fix for #14160 was achieved differently. Recently there was a restart of the afford to make bootselection possible with vmware #26010 (https://projects.theforeman.org/issues/26010)

PR #4405, opened 2017-03-23
fixes #18719 - replace usage of deprecated Fixnum constant

domcleal:
Most type checks for Fixnum can safely be replaced by Integer, as prior
to Ruby 2.4, Fixnum was a subclass of Integer and so the conditional
holds.


tbrisker:
Thanks @domcleal !

PR #4408, opened 2017-03-23
fixes #19013 - configure boot_retry for vmware guests

ekingme:
fixes #19013 - When provisioning a VMware host and using an ISO, the ISO is sometimes not mounted quick enough to boot.  Setting the Boot Retry option will allow the host to automatically reboot if it is unable to find a bootable medium.  This also partially addresses http://projects.theforeman.org/issues/4533.

### DIFF
--- a/app/models/compute_resources/foreman/model/vmware.rb
+++ b/app/models/compute_resources/foreman/model/vmware.rb
@@ -578,7 +578,8 @@ module Foreman::Model
         :scsi_controller => { :type => scsi_controller_default_type },
         :datacenter => datacenter,
         :firmware => 'automatic',
-        :boot_order => ['network', 'disk']
+        :boot_order => ['network', 'disk'],
+        :boot_retry => 10
       )
     end
 


mmoll:
ok to test

timogoebel:
Basically, this looks good. However, I haven't tested it, yet.

@ekingme: Do you see the issue with the bootdisk plugin?

ekingme:
@timogoebel yes, that's exactly where this is a problem for me. The bootdisk plugun allows an ISO to be generated and automatically attached to a new VMware guest created via foreman for environments where DHCP and TFTP are not an option. The problem for me is the guest boots too quickly, before the image can be attached, and then never tries to reboot after. Enabling this option (supported by fog), allows the host to retry boots.

I tested in my environment (with Foreman 1.13.2) with no issues.

timogoebel:
@ekingme : Thanks for the PR. I'd suggest putting the code in the plugin so it only runs on hosts that are bootdisk provisioned.

@domcleal: Is that fine with you?

You'd have to add the code here: https://github.com/theforeman/foreman_bootdisk/blob/master/app/models/concerns/foreman_bootdisk/compute_resources/vmware.rb#L12

and extend these tests:

https://github.com/theforeman/foreman_bootdisk/blob/master/test/unit/concerns/compute_resources/vmware_test.rb#L35

https://github.com/theforeman/foreman_bootdisk/blob/master/test/unit/concerns/compute_resources/vmware_test.rb#L62

ekingme:
Sorry for the delay.  I was able to test this by adding this option to the bootdisk plugin as suggested and I'll open a PR there soon.

timogoebel:
@ekingme : Ping me, when you have the PR ready. I'll be happy to test it.

PR #3678, opened 2016-07-27
fixes #15863 - add short request UUID to logs

domcleal:
New middleware stores the request ID (either from X-Request-ID or a
UUID generated by Rails) in the logger's thread storage. A short UUID is
now logged by default to minimise size of the log messages.

---

Generates log messages such as:

<pre>
$ egrep "33b38af1.*(Started|Completed)" log/development.log 
2016-07-27T15:38:43 33b38af1 [app] [I] Started GET "/" for 127.0.0.1 at 2016-07-27 15:38:43 +0100
2016-07-27T15:38:45 33b38af1 [app] [I] Completed 200 OK in 1978ms (Views: 1695.6ms | ActiveRecord: 18.5ms)
</pre>

PR #4410, opened 2017-03-24
Fixes #19017 - prefer request id over session id in logging

iNecas:
suggested to use session id if available. Although it seemed as a good
idea at the beginning, it has several disadvanteges:

* it treats UI and API requests differently

* it makes it impossible to differentiate requests made from the session
  at the same time (notifications is one example, but in general, we
  should not assume the user is performing
  on request at a time)

* the usability of this as correlation id is questionable, especially
  with asynchronous actions in place

### DIFF
--- a/config/logging.yaml
+++ b/config/logging.yaml
@@ -55,7 +55,7 @@
   :log_trace: false
   :level: info
   :type: file
-  :pattern: "%d %.8X{session} [%c] [%.1l] %m\n"
+  :pattern: "%d %.8X{request} [%c] [%.1l] %m\n"
 
 :production:
   :filename: "production.log"


iNecas:
@domcleal @ohadlevy @lzap as people that came to conclusion that session id is better for tagged logging (discussed in https://github.com/theforeman/foreman/issues/3678), I would like to re-open this as IMO it limits the usability of the feature as mentioned before, without apparent advantage

lzap:
Honestly, we should have both. In structured logging you don't have any limitation of what's "too long" because you introduce new field. Both make perfect sense. If I had to pick one (until we implement journald support where we can really send both) I would pick request which is more globally appropriate IMHO. This depends really, I think if you are more into UI troubleshooting you pick session, if into backend (myself included) you pick request.

My biggest argument for request would be that session is unusable for backend while it is still quite usable for UI. Therefore IMHO this is RTM, unless folks have a different opinion (e.g. include both).

Edited: Lack of coffee.

PR #4411, opened 2017-03-26
Fixes #12272 - Support multiple certificates in ovirt resource

orrabin:


orrabin:
@lzap as per our conversation, I changed all the newlines in the example certificates to match the UI and changes the split accordingly.

orrabin:
[test foreman]

app/models/compute_resources/foreman/model/ovirt.rb
```diff
@@ -365,9 +365,13 @@ def api_version
       @api_version ||= client.api_version
     end
 
-    def ca_cert_store(cert)
-      return if cert.blank?
-      OpenSSL::X509::Store.new.add_cert(OpenSSL::X509::Certificate.new(cert))
+    def ca_cert_store(certs)
+      return if certs.blank?
+      store = OpenSSL::X509::Store.new
+      certs.split("-----END CERTIFICATE-----").each do |cert|
```
iNecas: You can make the split pattern as part of the result like this:

```ruby
certs.split(/(?=-----BEGIN)/)
```

Also, I'm not sure how big chance is the certs would start with `-----BEGIN TRUSTED CERTIFICATE-----` (I saw that format in my pki directory), but keeping the pattern shorter `-----BEGIN` can potentially cover more cases than the specific version that is there right now.

orrabin:
@iNecas thanks, your pattern works better and I don't need to fix the string after it.

iNecas (approved):
Bonus points for tests :) Waiting for Jenkins

iNecas:
Test failure independent. Merging now. Thanks @orrabin

app/models/compute_resources/foreman/model/ovirt.rb
```diff
@@ -365,9 +365,13 @@ def api_version
       @api_version ||= client.api_version
     end
 
-    def ca_cert_store(cert)
-      return if cert.blank?
-      OpenSSL::X509::Store.new.add_cert(OpenSSL::X509::Certificate.new(cert))
+    def ca_cert_store(certs)
+      return if certs.blank?
+      store = OpenSSL::X509::Store.new
+      certs.split(/(?=-----BEGIN)/).each do |cert|
```
lzap, line 371: Nice patch thanks, out of my curiousity, why this is "positive lookahead" regexp?

iNecas: I not sure I understand the question. It's there to split based on expression, but still include the match inside the split.

lzap: It's not a match, it's just split into array, interate over it. There is no point to have `?=` (lookahead) option, is my point. Unless @orrabin explains. I am trying to learn here, no big deal.

orrabin: @lzap my split removed the part of the certificate I was splitting on: `certs.split("-----END CERTIFICATE-----")` and I needed to add it back so @iNecas suggested the lookahead which didn't remove anything.

PR #4413, opened 2017-03-27
Refs #18001 - Avoid role extending from rake tasks permission

ares:
Loading roles too early causes problems when there's no db yet. Even running rake db:create for production database fails. For development it works but fails later when environment is being loaded for migrations etc.

http://ci.theforeman.org/job/test_plugin_pull_request/3536/database=sqlite3,label=fast,ruby=2.3/consoleFull

### DIFF
--- a/app/services/foreman/plugin/rbac_support.rb
+++ b/app/services/foreman/plugin/rbac_support.rb
@@ -7,6 +7,8 @@ module Foreman
           add_all_permissions_to_role("Manager", all_permissions)
           add_all_permissions_to_role("Viewer", view_permissions)
         end
+      rescue ActiveRecord::StatementInvalid => e
+        Foreman::Logging.exception _("Could not add permissions to Manager and Viewer roles: %s") % e.message, e, :level => :debug
       end
 
       def add_resource_permissions_to_default_roles(resources, opts = {})
@@ -38,6 +40,8 @@ module Foreman
         manager = Role.find_by :name => "Manager"
         return if !manager || resources.empty?
         include_permissions manager, resources, opts
+      rescue ActiveRecord::StatementInvalid => e
+        Foreman::Logging.exception _("Could not add permissions to Manager role: %s") % e.message, e, :level => :debug
       end
 
       def add_resource_permissions_to_viewer(resources, opts)
@@ -46,6 +50,8 @@ module Foreman
         opts[:condition_hash] = { :name => "view_%" }
         return if !viewer || resources.empty?
         include_permissions viewer, resources, opts
+      rescue ActiveRecord::StatementInvalid => e
+        Foreman::Logging.exception _("Could not add permissions to Viewer role: %s") % e.message, e, :level => :debug
       end
 
       def include_permissions(role, resources, opts)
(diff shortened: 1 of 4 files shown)


lzap:
Can you send full exception stacktraces into the logging thing? So they are not dropped. Other than that :+1:

ares:
Thanks @lzap, I added the logging. I also now rescue from specific exception and added the rescue to add_all_permissions_to_default_roles which could fail for same reason but on permissions table. Please take another look.

app/services/foreman/plugin/rbac_support.rb
```diff
@@ -7,6 +7,13 @@ def add_all_permissions_to_default_roles(all_permissions)
           add_all_permissions_to_role("Manager", all_permissions)
           add_all_permissions_to_role("Viewer", view_permissions)
         end
+      rescue ActiveRecord::StatementInvalid => e
+        if Foreman.in_rake? || Rails.env == 'test'
```
lzap: Isn't this supposed to be `Rails.env != 'test'`? Because in test environment we want to rather error out (and definitely not want to flood logs).

app/services/foreman/plugin/rbac_support.rb
```diff
@@ -7,6 +7,13 @@ def add_all_permissions_to_default_roles(all_permissions)
...
+          Rails.logger.warn(_("Could not add permissions to Manager and Viewer roles: %s") % e.message)
+          Rails.logger.debug e.backtrace.join("\n")
```
lzap: Can you use `Foreman::Logging.exception` please here?

lzap (changes requested):
Two comments

ares:
@lzap this turned out to be a bigger problem. I changed the the logging in all 4 methods so that we always just log debug message. Otherwise it would prevent starting `rails console` e.g. in case the development database was not migrated.

The https://github.com/theforeman/foreman/pull/4168 unfortunatelly broke plugins that defines permissions. It was only reproducible on empty db which is why we missed it (katello tests define permissions differently so test_develop didn't catch it either). The problem was that the new `RbacRegistry` returned only permissions that were already stored in db. But in case the db was not ready yet we were unable to return the list of permissions and their resource types which is required by fixtures to properly setup permissions for plugins. Therefore I modified the registry so it keeps all information in memory and does not require db to be operational.

This issue cause plugin PRs to fail. I triggered a test run with this branch and REX PR https://github.com/theforeman/foreman_remote_execution/pull/239  at http://ci.theforeman.org/job/test_plugin_pull_request/3545/ which should be green now (at least it is green locally).

app/services/foreman/plugin/rbac_support.rb
```diff
@@ -85,6 +91,8 @@ def add_all_permissions_to_role(role_name, permissions)
         role = Role.find_by :name => role_name
         return unless role
         include_permissions_for_role role, permissions.map(&:name)
+      rescue ActiveRecord::StatementInvalid => e
+        Foreman::Logging.exception _("Could not extend role '%{name}': %{message}") % {:name => role_name, :message => e.message}, e, :level => :debug
```
lzap, line 95: Not so keen about `debug` level, exception must not be debug at any point.

lzap (approved):
I am merging as tests are broken, discovery passed.

PR #4414, opened 2017-03-27
Fixes #19036 - Add the hiera.yaml file from to foreman-debug collection

bkearney:


lzap:
I have shortened the commit subject, thanks @bkearney. My only concern is if it contains sensitive data, but we already filter out YAML patterns like "password:" or similar, this is very generic file, we can't do much about it.

PR #4415, opened 2017-03-27
Fixes #19039 - Lock plugin roles

xprazak2:


ares:
@xprazak2 katello tests are failing with undefined method `permission_diff' for nil:NilClass /var/lib/workspace/workspace/test_develop_pr_katello/database/postgresql/ruby/2.2/slave/fast/db/migrate/20170327100309_lock_plugin_roles.rb:6:

foreman test failure is unrelated

lib/seed_helper.rb
```diff
@@ -58,5 +58,29 @@ def create_role(role_name, permission_names, builtin, check_audit = true)
       permissions = Permission.where(:name => permission_names)
       create_filters(role, permissions)
     end
+
+    def rename_existing(role, original_name)
```
ares: SeedHelper is generic helper for seeding, while this and `rename_with_free_name` methods are Role specific, their names should mention role in name at least.

app/services/foreman/plugin.rb
```diff
@@ -241,8 +241,11 @@ def role(name, permissions)
       default_roles[name] = permissions
       return false if pending_migrations || Rails.env.test?
       Role.transaction do
-        role = Role.where(:name => name).first_or_create
-        role.add_permissions!(permissions) if role.permissions.empty?
+        role = Role.find_by :name => name
+        role = Role.create :name => name, :origin => self.id unless role
+        role.ignore_locking do |r|
+          r.add_permissions!(permissions)
```
ares: I'd need to double check but this was creating doubled filterings every time the plugin was initialized. See https://github.com/theforeman/foreman/pull/4168/files#r107688522, skipping for existing permission is required I think.

ares (changes requested):

xprazak2:
I renamed the methods to give them more specific names and skipping filtering creation for existing permissions

db/migrate/20170327100309_lock_plugin_roles.rb
```diff
@@ -0,0 +1,15 @@
+class LockPluginRoles < ActiveRecord::Migration
+  def up
+    Foreman::Plugin.all.each do |plugin|
+      plugin.default_roles.each do |role_name, expected_perms|
+        role = Role.find_by :name => role_name
+        next unless role
```
ares: how can this happen? if we skip the locking here, the migration will be marked as done and won't run again so if this is caused by some "db not ready" issue, this would leave the role unlocked

lib/seed_helper.rb
```diff
@@ -58,5 +58,29 @@ def create_role(role_name, permission_names, builtin, check_audit = true)
...
+      num = 1
+      new_name = generate_role_name prefix, original_name, num
+      while Role.find_by :name => new_name
+        new_name = generate_role_name prefix, original_name, num
+        num += 1
+      end
+      role.update_attribute :name, new_name
+    end
+
+    def generate_role_name(prefix, original_name, num)
```
ares: this one was not roles specific so could remain without role in name

db/migrate/20170221195674_tidy_current_roles.rb
```diff
@@ -21,34 +21,10 @@ def up
   def process_existing(original_name, role, permission_names)
     diff = role.permission_diff permission_names
     return if diff.empty?
-    rename_existing role, original_name
+    SeedHelper.rename_existing_role role, original_name
     create_from_seeds original_name, permission_names
   end
 
-  def rename_existing(role, original_name)
```
ares: I wonder whether we shouldn't keep the method copy in here. The problem I see with this is that if later someone changes the implementation in seed helper, he/she would also change a the behavior of a migration that might or might not be applied. This could be easily missed but could result in users having different results.

ares (changes requested):
Adding a [plugin test](test/unit/plugin_test.rb) that verifies the role is locked would be useful. Also we should test that it adds permission but does not duplicate filterings. That could probably be tested by running "role" twice with different permissions sets.

db/migrate/20170327100309_lock_plugin_roles.rb
```diff
@@ -0,0 +1,15 @@
+class LockPluginRoles < ActiveRecord::Migration
+  def up
+    Foreman::Plugin.all.each do |plugin|
```
ares: I think we can't do this by migration, if I add plugin later after this migration was already done, it would have roles unlocked... I think we need to do it at runtime when we're adding permissions to the role

xprazak2: If you add plugin later, roles will get [locked](https://github.com/theforeman/foreman/pull/4415/files#diff-d0644101ad428cdee084ab59f3e9a43cR245) when registering the plugin. Migration just renames the roles that were modified.

ares: I see, so in my case that didn't work. The reason is the role existed even before I added the plugin so it was found and not created.

to reproduce:
1) create role e.g. Discovery Manager
2) rake db:migrate
3) add a discovery plugin
4) Discovery Manager remains unlocked and was not renamed

I agree it might be quite an edge case but we should probably make sure the role is locked when we search for it in `role` DSL

db/migrate/20170221195674_tidy_current_roles.rb
xprazak2: If we want to keep a copy in here, we should also have copies of `rename_with_free` and `generate_name` methods, because the same applies to them. I decided to extract these because copy-pasting 3 methods from different migration seemed ugly.

app/services/foreman/plugin.rb
ares: note for myself: added in new version

db/migrate/20170221195674_tidy_current_roles.rb
ares: yeah, I know it's a duplication of 3 methods, but in this case I think it's better (safer)

db/migrate/20170327100309_lock_plugin_roles.rb
xprazak2: I did not think of this one, nice catch.

xprazak2:
I removed the migration, all changes are now part of DSL for plugin registration. I also do not reuse code from previous migrations. `role` method in DSL [cannot be really tested](https://github.com/theforeman/foreman/blob/develop/app/services/foreman/plugin.rb#L242), but I moved most of registration logic to a separate class which gets a decent test coverage. I will add test for filterings duplicity.

Edit: done.

app/services/foreman/plugin/role_lock.rb
```diff
@@ -0,0 +1,71 @@
...
+        role
+      end
+
+      def rename_and_create(role, original_name, permissions)
+        rename_existing role, original_name
+        create_plugin_role original_name, permissions
+      end
+
+      def create_plugin_role(name, permissions)
+        role = Role.create :name => name, :origin => @plugin_id
```
ares: this silently fails if built role is invalid, we should do `Role.create!` I think (it would fail in `r.add_permissions!` few lines below anyway, but the trace would be hiding the cause

app/services/foreman/plugin/role_lock.rb
```diff
@@ -0,0 +1,71 @@
...
+          r.add_permissions!(permissions)
+        end
+        role
+      end
+
+      def rename_existing(role, original_name)
+        prefix = "Customized"
+        role_name = "#{prefix} #{original_name}"
+        if Role.find_by(:name => role_name)
+          rename_with_free_name role, prefix, original_name
```
ares: could this be combined in single method? It seems that this method covers first iteration where num would be nil, so I'd recommend  improving `generate_name` to that num is optional. This also deduplicates line 48 where `generate_name` could be used.

test/unit/role_lock_test.rb
```diff
@@ -0,0 +1,101 @@
...
+  end
+
+  test "should create plugin role" do
+    name = "Test Manager"
+    refute find_role(name)
+    @role_lock.create_plugin_role name, @permissions
+    role = find_role(name)
+    assert role
+    assert_equal @permissions, role.permissions.pluck(:name).map(&:to_sym)
+    assert_equal @role_lock.plugin_id, role.origin
```
ares, line 54: I'd add `assert role.locked?` just to be sure

test/unit/role_lock_test.rb
```diff
@@ -0,0 +1,101 @@
...
+  test "should process existing role with unchanged permissions list" do
+    role = roles(:create_hosts)
+    original_name = role.name
+    original_id = role.id
+    refute role.locked?
+    @role_lock.process_role original_name, role.permissions.pluck(:name).map(&:to_sym)
+    updated_original = find_role(original_name)
+    assert_equal original_id, updated_original.id
+    assert_equal original_name, updated_original.name
+    assert @role_lock.plugin_id, updated_original.origin
```
ares: I'd add assert role.locked? just to be sure esp. since we checked this on line 76 with refute

test/unit/role_lock_test.rb
```diff
@@ -0,0 +1,101 @@
...
+    @role_lock.process_role role_name, [:view_hosts]
+    new_role = find_role(role_name)
+    assert new_role
+    assert new_role.locked?
+  end
+
+  test "should register role" do
+    name = "Test Manager"
+    registry = Foreman::Plugin::RbacRegistry.new
+    assert registry.role_ids.empty?
```
ares: I'd prefer `assert_empty` to generate better message on failure

test/unit/role_lock_test.rb
```diff
@@ -0,0 +1,101 @@
...
+    @role_lock.register_role name, @permissions, registry
+    refute registry.role_ids.empty?
```
ares: `refute_empty` should work

ares (changes requested):
The PR works well for all combinations I was able to think of. I left few comments in code.

xprazak2:
I use `create!` when creating new roles and squashed the 2 methods in `role_lock`, added assertions to tests and changed `generate_name` method to take optional `num` argument

xprazak2:
[test katello]

xprazak2:
Integration test failures on foreman, failures on katello related to minitest update. There is a [PR](https://github.com/Katello/katello/pull/6807) for that, we'll need to wait for it to get in.

mmoll:
[test foreman] [test katello] then :)

xprazak2:
Integration test failing

ares (changes requested):
Is there a chance you haven't pushed the last version? I don't see comments from last review fixed in the code.

xprazak2:
My apologies, I pushed poorly. The changes should be there now.

tbrisker:
I think this PR can benefit from the recently added `Role.ignore_locking` method (https://github.com/theforeman/foreman/pull/4623/files#diff-7863dd934f73db4d813673b9dec6db0aR81)

xprazak2:
@tbrisker, do you have in mind any specific place it should be used? I think [I take advantage of it ](https://github.com/theforeman/foreman/pull/4415/files#diff-3bb6d4c73cc0823a33afc88ecc16fdfeR13) already..

tbrisker:
@xprazak2 There is a new class-level ignore_locking method, I'll try to add inline where it could be used

app/models/role.rb
```diff
@@ -178,6 +179,12 @@ def ignore_locking
     self
   end
 
+  def self.create_locked!(role_attrs = {})
+    role = Role.create! role_attrs.merge(:modify_locked => true)
```
tbrisker: I think this is not really needed anymore since you could just 
```
Role.ignore_locking do
  Role.create! attrs
end
``` 
where you need to. This method doesn't actually cause the created role to be locked afaict.

app/services/foreman/plugin/role_lock.rb
```diff
@@ -0,0 +1,65 @@
...
+      end
+
+      def rename_and_create(role, original_name, permissions)
+        rename_existing role, original_name
+        create_plugin_role original_name, permissions
+      end
+
+      def create_plugin_role(name, permissions)
+        role = Role.create_locked! :name => name, :origin => @plugin_id
+        role.ignore_locking do |r|
```
tbrisker: This can be replaced with `Role.ignore_locking` and have the creation in the same block

app/services/foreman/plugin/role_lock.rb
```diff
@@ -0,0 +1,65 @@
...
+          r.add_permissions!(permissions)
+        end
+        role
+      end
+
+      def rename_existing(role, original_name)
+        prefix = "Customized"
+        role_name = generate_name prefix, original_name
+        if Role.find_by(:name => role_name)
```
tbrisker: why not `Role.where(:name => role_name).each_with_index`?

tbrisker (commented):
added a few inline comments, but didn't do a complete review yet.

app/services/foreman/plugin/role_lock.rb
xprazak2: Because I know that I will find at most 1 - names are unique. But I do not know, how long I will have to iterate to find unique name, that is why there is a while loop - I try to find a role with newly generated name and if I get none, I can use it to rename.

xprazak2:
I use `Role.ignore_locking` instead of `create_locked!`, which is not needed.

app/services/foreman/plugin/role_lock.rb
```diff
@@ -0,0 +1,65 @@
...
+
+      def initialize(plugin_id)
+        @plugin_id = plugin_id
+      end
+
+      def register_role(name, permissions, role_registry)
+        Role.transaction do
+          role = process_role name, permissions
+          role.ignore_locking do |r|
+            r.add_permissions!(permissions)
```
tbrisker: don't you already add the permissions inside process_role->create_plugin_role? it is called for every case that there is a difference in permissions.

app/services/foreman/plugin/role_lock.rb
```diff
@@ -0,0 +1,65 @@
...
+      end
+
+      def process_role(name, permissions)
+        role = Role.find_by :name => name
+        if role && role.permission_diff(permissions).empty?
+          role.update_attribute :origin, @plugin_id if role.origin.empty?
+        elsif role
+          role = rename_and_create role, name, permissions
+        else
+          role = create_plugin_role name, permissions
```
tbrisker: you can simplify this if statement and get rid of the rename_and_create function, something like:
```ruby
if role && role.permission_diff(permissions).empty?
     role.update_attribute :origin, @plugin_id if role.origin.empty?
     return role
end
rename_existing(role, name) if role
create_plugin_role(name, permissions)
role
```

app/services/foreman/plugin/role_lock.rb
tbrisker: how about something like this to find the last one:
```ruby
Role.where('name like ?%',"Customized #{original_name}").order(:name).last.name
```
Or, instead of adding sequential numbers add a uuid and then there is no risk of collision?

tbrisker (changes requested):
Looks good in general, added a few suggestions to simplify the RoleLock module inline.

xprazak2:
I removed the while loop and cleaned up as suggested.

ares (changes requested):
While testing this, I encountered one problem that would hit us in future. Suppose you have a role created by plugin. Then the plugin adds new permission to this role. It renames the existing locked role to "Customized $name" while keeping it locked and create the new role with new set of permissions.

I think it should update the existing role if it's locked (origin set) and permission set differs. This way we'll always have all locked roles up to date. That means we'd need to remove additional permissions and add new ones.

xprazak2:
I rebased and handle the case when new permissions are added, when permissions are removed is not handled yet.

xprazak2:
Roles are now processed correctly if a permission is removed.

app/models/role.rb
```diff
@@ -146,31 +146,78 @@ def add_permissions(permissions, options = {})
...
+    permissions = Array(permissions)
+    collection = permission_records permissions
+    current_filters = self.filters
+    find_permissions_to_remove collection, current_filters
+  end
+
+  def find_permissions_to_remove(collection, current_filters)
+    collection.group_by(&:resource_type).inject([]) do |memo, (resource_type, grouped_permissions)|
+      memo.concat mark_for_removal(resource_type, grouped_permissions, current_filters)
+      memo
```
ares: this should not be necessary, `concat` returns the new array

app/models/role.rb
```diff
@@ -146,31 +146,78 @@ def add_permissions(permissions, options = {})
...
+    end
+  end
+
+  def mark_for_removal(resource_type, grouped_permissions, current_filters)
```
ares: The name of method is a slightly confusing, I don't see anything marked here, it seems to return either filter or filtering. The name could be something like `filters_and_filterings_for_removal`

app/models/role.rb
```diff
@@ -146,31 +146,78 @@ def add_permissions(permissions, options = {})
...
       grouped_permissions.each do |permission|
+        next if filter.permissions.include?(permission)
         filtering = filter.filterings.build
         filtering.filter = filter
         filtering.permission = permission
       end
     end
   end
 
+  def remove_permissions(permissions)
```
ares: Does this method remove any permission? it seems it only finds permissions to be removed, perhaps rename it to `find_objects_for_permissions_removal`. If my understanding is correct, it returns filters and filterings that should be deleted in order to remove given permissions.

Also here, permissions are expected to be array of names, the argument could be renamed to `permission_names` to be more obvious.

app/models/role.rb
```diff
@@ -146,31 +146,78 @@ def add_permissions(permissions, options = {})
...
+    permissions = Array(permissions)
+    collection = permission_records permissions
+    current_filters = self.filters
+    find_permissions_to_remove collection, current_filters
+  end
+
+  def find_permissions_to_remove(collection, current_filters)
```
ares: I'd maybe keep it in `remove_permissions`, it seems like it's not used anywhere else and it feels like splitting one method into two only to get better score for method complexity

app/models/role.rb
```diff
@@ -146,31 +146,78 @@ def add_permissions(permissions, options = {})
...
-    (current_names - permission_names) | (permission_names - current_names)
+    current_names = permission_symbols
+    extra_permissions(permission_names, current_names) | missing_permissions(permission_names, current_names)
+  end
+
+  def permission_symbols
+    permissions.map(&:name).map(&:to_sym)
+  end
+
+  def extra_permissions(permission_names, current_names = permission_symbols)
```
ares: this seems like a candidate for private method

app/models/role.rb
```diff
@@ -146,31 +146,78 @@ def add_permissions(permissions, options = {})
...
+    current_names - permission_names
+  end
+
+  def missing_permissions(permission_names, current_names = permission_symbols)
```
ares, line 196: this seems like a candidate for private method

app/models/role.rb
```diff
@@ -254,13 +301,26 @@ def not_locked
   end
 
   def find_filter(resource_type, current_filters, search)
-    filter_record = Filter.where(:search => search, :role_id => id).joins(:permissions)
-                          .where("permissions.resource_type" => resource_type).first
+    Filter.where(:search => search, :role_id => id).joins(:permissions)
+          .where("permissions.resource_type" => resource_type).first
+  end
+
+  def filter_for_permission_add(resource_type, current_filters, search)
```
ares, line 296: this seems like a candidate for private method

app/models/role.rb
```diff
@@ -254,13 +301,26 @@ def not_locked
...
     if filter_record
       # add filterings to what we have in memory, not to a newly fetched record
-      current_filters.detect { |fil| fil.id == filter_record.id }
+      find_current_filter current_filters, filter_record
     else
       self.filters.build(:search => search)
     end
   end
+
+  def filter_for_permissions_remove(resource_type, current_filters)
```
ares, line 306: this seems like a candidate for private method

app/models/role.rb
```diff
@@ -254,13 +301,26 @@ def not_locked
...
+    filter_record = find_filter resource_type, current_filters, nil
+    find_current_filter current_filters, filter_record
+  end
+
+  def find_current_filter(current_filters, filter_record)
```
ares, line 311: this seems like a candidate for private method

ares (changes requested):
Couple of smaller comments, the functionality is awesome!

One thing that turned out during tests is that renaming to "Customized $role" and new "$role" creation create audit records. We should probably avoid that.

One issue I found is that when I removed all permissions for a given resource type from some role, Foreman does not start with this trace

```
017-08-15T10:24:04  [app] [W] Could not create role 'Ansible Roles Manager': ERF73-2603 [Foreman::PermissionMissingException]: some permissions were not found
2017-08-15T10:24:04  [app] [E] Cannot continue because some permissions were not found, please run rake db:seed and retry
Exiting
/home/ares/Projekty/Zdrojaky/foreman/app/models/role.rb:191:in `permission_records': ERF73-2603 [Foreman::PermissionMissingException]: some permissions were not found (Foreman::PermissionMissingException)
        from /home/ares/Projekty/Zdrojaky/foreman/app/models/role.rb:166:in `remove_permissions'
        from /home/ares/Projekty/Zdrojaky/foreman/app/models/role.rb:218:in `remove_permissions!'
        from /home/ares/Projekty/Zdrojaky/foreman/app/services/foreman/plugin/role_lock.rb:45:in `block in update_plugin_role_permissions'
        from /home/ares/Projekty/Zdrojaky/foreman/app/models/role.rb:85:in `ignore_locking'
        from /home/ares/Projekty/Zdrojaky/foreman/app/services/foreman/plugin/role_lock.rb:41:in `update_plugin_role_permissions'
        from /home/ares/Projekty/Zdrojaky/foreman/app/services/foreman/plugin/role_lock.rb:22:in `process_role'
        from /home/ares/Projekty/Zdrojaky/foreman/app/services/foreman/plugin/role_lock.rb:13:in `block (2 levels) in register_role'
        from /home/ares/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/activerecord-4.2.9/lib/active_record/connection_adapters/abstract/database_statements.rb:213:in `block in transaction'
        from /home/ares/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/activerecord-4.2.9/lib/active_record/connection_adapters/abstract/transaction.rb:184:in `within_new_transaction'
        from /home/ares/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/activerecord-4.2.9/lib/active_record/connection_adapters/abstract/database_statements.rb:213:in `transaction'
        from /home/ares/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/activerecord-4.2.9/lib/active_record/transactions.rb:220:in `transaction'
        from /home/ares/Projekty/Zdrojaky/foreman/app/services/foreman/plugin/role_lock.rb:12:in `block in register_role'
        from /home/ares/Projekty/Zdrojaky/foreman/app/models/concerns/foreman/thread_session.rb:93:in `as'
        from /home/ares/Projekty/Zdrojaky/foreman/app/models/concerns/foreman/thread_session.rb:99:in `as_anonymous_admin'
        from /home/ares/Projekty/Zdrojaky/foreman/app/services/foreman/plugin/role_lock.rb:11:in `register_role'
        from /home/ares/Projekty/Zdrojaky/foreman/app/registries/foreman/plugin.rb:247:in `role'
        from /home/ares/Projekty/Zdrojaky/foreman_ansible/lib/foreman_ansible/engine.rb:64:in `block (2 levels) in <class:Engine>'
        from /home/ares/Projekty/Zdrojaky/foreman/app/registries/foreman/plugin.rb:66:in `instance_eval'
        from /home/ares/Projekty/Zdrojaky/foreman/app/registries/foreman/plugin.rb:66:in `register'
```
it seems that it only happens if the filter that's being removed contains 2 or more permissions, if the filter has just one, it does not happen and it's deleted correctly

ares:
oh and tests seems to be failing because of the changes

app/models/role.rb
xprazak2: I split this because code climate was complaining about method complexity, I'll put it back together.

xprazak2:
I made private all the methods that I could, renamed the methods and removal should now work without problems.

app/models/role.rb
ares: Feel free to ignore code climate checks if they don't make sense. It's just a shortcut for reviewers so they don't need to run it manually. It doesn't have to be green for PR to be merged.

ares:
tests queued for 3 days :-) [test foreman] [test katello] [test upgrade]

ares (changes requested):
Getting this when starting Foreman after I do any change to existing plugin role.

```
home/ares/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/activerecord-4.2.9/lib/active_record/connection_adapters/postgresql_adapter.rb:637:in `prepare': PG::UndefinedFunction: ERROR:  operator does not exist: character varying = integer (ActiveRecord::StatementInvalid)
LINE 1: ...(*) FROM "permissions" WHERE "permissions"."name" IN (SELECT...
                                                             ^
HINT:  No operator matches the given name and argument type(s). You might need to add explicit type casts.
: SELECT COUNT(*) FROM "permissions" WHERE "permissions"."name" IN (SELECT "permissions"."id" FROM "permissions" INNER JOIN "filterings" ON "permissions"."id" = "filterings"."permission_id" INNER JOIN "filters" ON "filterings"."filter_id" = "filters"."id" WHERE "filters"."role_id" = $1  ORDER BY filters.role_id, filters.id)
        from /home/ares/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/activerecord-4.2.9/lib/active_record/connection_adapters/postgresql_adapter.rb:637:in `prepare_statement'
        from /home/ares/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/activerecord-4.2.9/lib/active_record/connection_adapters/postgresql_adapter.rb:596:in `exec_cache'
        from /home/ares/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/activerecord-4.2.9/lib/active_record/connection_adapters/postgresql_adapter.rb:585:in `execute_and_clear'
        from /home/ares/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/activerecord-4.2.9/lib/active_record/connection_adapters/postgresql/database_statements.rb:160:in `exec_query'
        from /home/ares/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/activerecord-4.2.9/lib/active_record/connection_adapters/abstract/database_statements.rb:356:in `select'
        from /home/ares/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/activerecord-4.2.9/lib/active_record/connection_adapters/abstract/database_statements.rb:32:in `select_all'
        from /home/ares/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/activerecord-4.2.9/lib/active_record/connection_adapters/abstract/query_cache.rb:70:in `select_all'
        from /home/ares/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/activerecord-4.2.9/lib/active_record/relation/calculations.rb:270:in `execute_simple_calculation'
        from /home/ares/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/activerecord-4.2.9/lib/active_record/relation/calculations.rb:227:in `perform_calculation'
        from /home/ares/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/activerecord-4.2.9/lib/active_record/relation/calculations.rb:133:in `calculate'
        from /home/ares/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/activerecord-4.2.9/lib/active_record/relation/calculations.rb:48:in `count'
        from /home/ares/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/activerecord-4.2.9/lib/active_record/relation.rb:258:in `size'
        from /home/ares/Projekty/Zdrojaky/foreman/app/models/role.rb:317:in `permission_records'
        from /home/ares/Projekty/Zdrojaky/foreman/app/models/role.rb:165:in `find_for_permission_removal'
        from /home/ares/Projekty/Zdrojaky/foreman/app/models/role.rb:206:in `remove_permissions!'
        from /home/ares/Projekty/Zdrojaky/foreman/app/services/foreman/plugin/role_lock.rb:47:in `block in update_plugin_role_permissions'
        from /home/ares/Projekty/Zdrojaky/foreman/app/models/role.rb:85:in `ignore_locking'
        from /home/ares/Projekty/Zdrojaky/foreman/app/services/foreman/plugin/role_lock.rb:43:in `update_plugin_role_permissions'
        from /home/ares/Projekty/Zdrojaky/foreman/app/services/foreman/plugin/role_lock.rb:22:in `process_role'
        from /home/ares/Projekty/Zdrojaky/foreman/app/services/foreman/plugin/role_lock.rb:13:in `block (2 levels) in register_role'
```

xprazak2:
Should be fixed now.

app/models/role.rb
```diff
@@ -146,31 +146,66 @@ def add_permissions(permissions, options = {})
...
+
   def permission_diff(permission_names)
-    current_names = permissions.map(&:name).map(&:to_sym)
-    (current_names - permission_names) | (permission_names - current_names)
+    current_names = permission_symbols
+    extra_permissions(permission_names, current_names) | missing_permissions(permission_names, current_names)
+  end
+
+  def permission_symbols
+    permissions.map(&:name).map(&:to_sym)
```
ares: microoptimization: `.map { |p| p.name.to_sym }`

app/services/foreman/plugin/role_lock.rb
```diff
@@ -0,0 +1,72 @@
...
+          Role.transaction do
+            role = process_role name, permissions
+            role_registry.role_ids << role.id
+          end
+        end
+      end
+
+      def process_role(name, permissions)
+        role = Role.find_by :name => name
+        if role && role.origin && !role.permission_diff(permissions).empty?
```
ares: `!role.permission_diff(permissions).empty?` would be more readable as `role.permission_diff(permissions).present?` (unless nil is a value that was important to accept in this case)

ares (changes requested):
Now it works brilliant! I left just two small comments. I see that a lot of methods remain public so I think they should be covered with tests (since other devs can use them). Or perhaps they could be made private. Once this is addressed I think it's ready to be merged.

xprazak2:
I made changes as suggested.

ares:
@xprazak2 seems like something else has change too, tests are broken

app/models/role.rb
```diff
@@ -146,31 +146,66 @@ def add_permissions(permissions, options = {})
...
+    permission_names - current_names
   end
 
   def add_permissions!(*args)
     add_permissions(*args)
     save!
   end
 
+  def remove_permissions!(*args)
+    find_for_permission_removal(*args).map(&:destroy!)
```
ares: I suppose the issue is here, you get the array already in args, if you put asterisk there, it calls `find_for_permission_removal` with multiple arguments while it expects single array

ares (approved):

ares:
I think all @tbrisker's suggestions were also implemented. Thanks @xprazak2, merged as cf8bc039064bb26cc0be1257f1de3730cd08f263. Great addition!

PR #4416, opened 2017-03-28
Fixes #19034 - add task for clearing expired notifications

amirfefer:


ohadlevy:
@amirfefer thanks. TBH I'm in mixed feelings about the module vs a common base class, why did you implement in in a module vs classes?

ohadlevy:
@amirfefer also see test failures.

ohadlevy:
@amirfefer please DRY your code, also i think it makes sense to unify the methods into one that accepts different options instead of one for each scenario. thanks

amirfefer:
@ohadlevy thanks for your review, I've unify the logic to a one `clean!` method.

app/services/ui_notifications/clean_expired.rb
```diff
@@ -0,0 +1,43 @@
+module UINotifications
+  class CleanExpired
+    def initialize(options = {})
+      if options[:blueprint].present?
+        blueprint = NotificationBlueprint.find_by_name(options[:blueprint])
```
ohadlevy: please use find_by(name: ..) syntax instead.

app/services/ui_notifications/clean_expired.rb
```diff
@@ -0,0 +1,43 @@
...
+        raise(Foreman::Exception, 'blueprint name does not exist') if blueprint.nil?
+        @blueprint = blueprint.name
+      end
+      if options[:expired_at].is_a? String
+        parsed_time =  options[:expired_at].to_time
```
ohadlevy: nitpick: extra whitespace after =

app/services/ui_notifications/clean_expired.rb
```diff
@@ -0,0 +1,43 @@
...
+        raise(Foreman::Exception, 'Expired time cannot be greater then now') if parsed_time > Time.now.utc
+      end
+      @expired_at = parsed_time || Time.now.utc
+      @group = options[:group]
+    end
+
+    def clean!
+      Foreman::Logging.logger('notifications').info("Removing all expired notifications")
```
ohadlevy: nitpick, extract to a method.

app/services/ui_notifications/clean_expired.rb
```diff
@@ -0,0 +1,43 @@
...
+      to_delete = Notification.joins(:notification_blueprint)
+                      .where('notifications.expired_at < ?', @expired_at)
+                      .where(blueprint_query, @blueprint)
+                      .where(group_query, @group)
+      @deleted_count = to_delete.size
```
ohadlevy: AFAIR the following line will provide you with a number, so you dont need to query first.

app/services/ui_notifications/clean_expired.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    def deleted_count
+      raise 'cleaner has not cleaned anything yet, run #clean! first' if @deleted_count.nil?
+      @deleted_count
+    end
+
+    private
+
+    def blueprint_query
+      @blueprint.present? ? 'notification_blueprints.name = ?' : nil
```
ohadlevy: IMHO this will fail on SQLITE.

ohadlevy (changes requested):

amirfefer:
@ohadlevy , thanks. I've fixed sqlite errors and added more tests.

app/services/ui_notifications/clean_expired.rb
```diff
@@ -0,0 +1,40 @@
...
+      @group = options[:group]
+    end
+
+    def clean!
+      UINotifications::Base.logger.info("Removing all expired notifications")
+      to_delete = Notification.joins(:notification_blueprint)
+                      .where('notifications.expired_at < ?', @expired_at)
+                      .where(blueprint_query)
+                      .where(group_query)
+      @deleted_count = to_delete.destroy_all.length
```
tbrisker: `destroy_all` instantiates all objects and runs callbacks when destroying. While this is fine for a small amount, this can be quite slow if there is a very large number of records to destroy - in that case it would be faster to use delete_all to directly delete everything from the database (like we do for report expiry for example).

amirfefer: @tbrisker I didn't use `delete_all` because according to [documentation](https://apidock.com/rails/ActiveRecord/Base/delete_all/class), dependent rules defined on associations are not honored, and error may raised.

So to use `delete_all` on notifications I've delete associated `notification_recipients` first.

ohadlevy:
@tbrisker it seems @amirfefer pushed a new version, can you please have another look? thanks!

ohadlevy:
@tbrisker anything left on this one? 
@amirfefer please create a packaging pr to add this to cron

ohadlevy:
[test katello]

app/services/ui_notifications/clean_expired.rb
```diff
@@ -0,0 +1,41 @@
...
+      @group = options[:group]
+    end
+
+    def clean!
+      UINotifications::Base.logger.info("Removing all expired notifications")
+      to_delete = Notification.joins(:notification_blueprint)
+                      .where('notifications.expired_at < ?', @expired_at)
+                      .where(blueprint_query)
+                      .where(group_query)
+      NotificationRecipient.where(notification_id: to_delete.map {|n| n.id }).delete_all
```
tbrisker: you can use something like `NotificationRecipient.where(notification: to_delete).delete_all` to avoid loading all the notifications to memory, and instead it will generate a sub-query based on the `to_delete` query.

app/services/ui_notifications/clean_expired.rb
```diff
@@ -0,0 +1,41 @@
+module UINotifications
+  class CleanExpired
+    def initialize(options = {})
+      if options[:blueprint].present?
+        blueprint = NotificationBlueprint.find_by(name: options[:blueprint])
+        raise(Foreman::Exception, 'blueprint name does not exist') if blueprint.nil?
+        @blueprint = blueprint.name
+      end
+      parsed_time = options[:expired_at].is_a?(String) ? options[:expired_at].to_time : Time.now.utc
+      raise(Foreman::Exception, 'Expired time cannot be greater then now') if parsed_time > Time.now.utc
```
tbrisker: if `options[:expired_at]` then `parsed_time == nil` which will fail at `parsed_time > Time.now.utc`

app/services/ui_notifications/clean_expired.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    def clean!
+      UINotifications::Base.logger.info("Removing all expired notifications")
+      to_delete = Notification.joins(:notification_blueprint)
+                      .where('notifications.expired_at < ?', @expired_at)
+                      .where(blueprint_query)
+                      .where(group_query)
+      NotificationRecipient.where(notification_id: to_delete.map {|n| n.id }).delete_all
+      @deleted_count = to_delete.delete_all
```
tbrisker: Why not just return this value instead of saving it in an instance variable and having another method to get it?

app/services/ui_notifications/clean_expired.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    def deleted_count
+      raise 'cleaner has not cleaned anything yet, run #clean! first' if @deleted_count.nil?
+      @deleted_count
+    end
+
+    private
+
+    def blueprint_query
+      @blueprint.present? ? {:notification_blueprints => {:name => @blueprint }} : nil
```
tbrisker: You could just set the instance variable in the initializer to the query instead of to the user input, and reduce these functions. (also true for the group)

app/services/ui_notifications/clean_expired.rb
```diff
@@ -0,0 +1,41 @@
...
+    def initialize(options = {})
+      if options[:blueprint].present?
+        blueprint = NotificationBlueprint.find_by(name: options[:blueprint])
+        raise(Foreman::Exception, 'blueprint name does not exist') if blueprint.nil?
+        @blueprint = blueprint.name
+      end
+      parsed_time = options[:expired_at].is_a?(String) ? options[:expired_at].to_time : Time.now.utc
+      raise(Foreman::Exception, 'Expired time cannot be greater then now') if parsed_time > Time.now.utc
+      @expired_at = parsed_time
+      @group = options[:group]
```
tbrisker: what if the group doesn't exist?

app/services/ui_notifications/clean_expired.rb
```diff
@@ -0,0 +1,41 @@
+module UINotifications
+  class CleanExpired
+    def initialize(options = {})
+      if options[:blueprint].present?
+        blueprint = NotificationBlueprint.find_by(name: options[:blueprint])
```
tbrisker: if you just need to check if the blueprint exists, you can use `NotificationBlueprint.where(name: options[:blueprint]).exists?` without loading the actual blueprint. Although i'm not sure you have to check if it exists, worst case no notifications will be deleted.

lib/tasks/notifications_cleaner.rake
```diff
@@ -0,0 +1,19 @@
+namespace :notifications do
+  desc 'Clean expired notification
+        Args can be a specific time, a group name or a blueprint name
```
tbrisker: you should give an example of using these arguments here

tbrisker (changes requested):

app/services/ui_notifications/clean_expired.rb
amirfefer: The general idea I took from [Fact_cleaner](https://github.com/theforeman/foreman/blob/develop/app/services/fact_cleaner.rb#L13), It seems  better then just return a deleted count, it stored for later use if needed. (I don't quite sure if in the future clean notifications will occur only as rake task).

amirfefer:
@tbrisker Thanks for your review, I've followed your inline comments, and I've added a test for bad parsing time string. I have one comment about `@deleted_count` instance variable, as you can see in inline comment.

app/services/ui_notifications/clean_expired.rb
```diff
@@ -0,0 +1,30 @@
+module UINotifications
+  class CleanExpired
+    def initialize(options = {})
+      blueprint = options[:blueprint]
```
tbrisker: you can simplify these a bit by using kwargs, something like `def initialize(blueprint: nil, group: nil, expired_at: Time.now.utc)`

app/services/ui_notifications/clean_expired.rb
```diff
@@ -0,0 +1,30 @@
...
+      @blueprint = blueprint.present? ? {:notification_blueprints => {:name => blueprint }} : nil
+      group = options[:group]
+      @group = group.present? ? {:notification_blueprints => {:group => group}} : nil
+      parsed_time = options[:expired_at].is_a?(String) ? options[:expired_at].to_time : Time.now.utc
+      raise(Foreman::Exception, 'Parsing expired time has been failed') if parsed_time.nil?
```
tbrisker: nitpick- should be `'Parsing expired time has failed'`.

ohadlevy:
@amirfefer ping?

amirfefer:
@tbrisker ,Thanks for your kwargs advice :+1: 
Could you please re-review?

tbrisker:
Thanks @amirfefer !

PR #4417, opened 2017-03-28
fixes #19035 - rewrite TopbarSweeper without rails-observers

domcleal:
Moves from the observer object into two mixins, one on the model and one
on the top-level controllers to observe creates/updates/destroys on
monitored models. Replaces rails-observers as it lacks Rails 5 support.


app/services/topbar_sweeper.rb
```diff
@@ -0,0 +1,22 @@
...
+      "tabs_and_title_records-#{id}"
+    end
+
+    def expire_cache_all_users
+      User.unscoped.pluck(:id).each do |id|
+        Rails.cache.delete("views/#{fragment_name(id)}")
+      end
+    end
+  end
+end
```
lzap, line 22: I know you are just getting rid of dependency, but this class is very clunky, it was introduced as part of refactoring in issue 2513 and it's thread unfriendly singleton with unnecessary reference to controller. Method `expire_fragment` which is only available in controller context should be really called from proper place and not from here. Also the mixture of class methods and singleton methods is not ideal. I think it can be all done properly in a controller helper.

app/models/filter.rb
```diff
@@ -140,9 +141,9 @@ def search_condition
     searches.join(' and ')
   end
 
-  def expire_topbar_cache(sweeper)
-    role.users.each      { |u| u.expire_topbar_cache(sweeper) }
-    role.usergroups.each { |g| g.expire_topbar_cache(sweeper) }
+  def expire_topbar_cache
+    role.users.each      { |u| u.expire_topbar_cache }
+    role.usergroups.each { |g| g.expire_topbar_cache }
```
lzap, line 146: Can you use more quicker batch loop in here via `find_each`?

lzap (changes requested):
Overall great, but this is good oportunity to break up the weird singleton class into controller helper or something else.

app/services/topbar_sweeper.rb
domcleal: I'm not sure that I understand this comment. Broken down:

> and it's thread unfriendly singleton with unnecessary reference to controller

One thread can only be processing one request at a time, I'm not sure that thread-safety is a requirement in controllers or helpers like this.

> Method expire_fragment which is only available in controller context should be really called from proper place and not from here. [..] I think it can be all done properly in a controller helper.

Do you want this implemented as a controller callback on any action that modifies any of the affected models? That could be a large number of actions. Keeping the callback on the model (and then this singleton to link the model callback to the calling controller) is likely to ensure consistency better, no matter where a model is changed from.

app/models/filter.rb
domcleal: If the association has already been loaded then `find_each` will produce _additional_ database queries. I'm not sure this is a clear cut optimisation.

app/services/topbar_sweeper.rb
lzap: Yes, I know we are not thread safe, but that does not mean we should avoid cleaner design. But it looks like I missed the fact we are observing models where controllers are not accessible. Then it makes sense, let me do one more read if I find something better.

app/models/filter.rb
lzap: Indeed that's not right.

test/integration_test_helper.rb
```diff
@@ -124,6 +124,13 @@ def set_request_user(user)
...
     create_cookie('test_user', user.login)
   end
+
+  def with_controller_caching(*controller_klasses)
+    controller_klasses.each { |c| c.perform_caching = true }
+    yield
+  ensure
+    controller_klasses.each { |c| c.perform_caching = false }
+  end
 end
```
lzap, line 134: Nice, I just noticed we use `:file_store` for test environment:

    config.cache_store = :file_store, Rails.root.join("tmp", "cache", "paralleltests#{ENV['TEST_ENV_NUMBER']}")

Do you think let's say `ActiveSupport::Cache::MemoryStore` might be more effective? We don't have many of those I assume, just the three here for now. But for the future.

Other than that, RTM.

domcleal: I don't think file or memory cache will make any difference. The Rails cache is used in a number of other, unrelated places in Foreman and libraries, not only this test (this test is exercising fragment caching only).

lzap (approved):

lzap:
Ok then thanks @domcleal !

PR #4418, opened 2017-03-28
Fixes #19044 - filter out :login field from logs

lzap:
This removes login field from INFO logs, password is filtered by default by Rails.

```
Parameters: {"utf8"=>"✓", "login"=>{"login"=>"admin", "password"=>"[FILTERED]"}, "commit"=>"Log In"}
```

In addition, I don't much like what was introduced in

https://github.com/theforeman/foreman/pull/4393

this is unnecessary exposal of username which can be abused for password brute-forcing. In addition to that, we already include this information, but in DEBUG level, meaning we duplicate this.

```
Rails.logger.debug "Setting current user thread-local variable to " + (o.is_a?(User) ? o.login : 'nil')
```

This patch changes this to database ID instead of username. To be honest, I liked the initial proposal of sending this to each individual log file which enables us to easily grep for all logs of particular user, it's essentially another correlation ID:

https://github.com/theforeman/foreman/pull/4385

So I am rewriting this back to this approach. Also I am adding some prefixes to the IDs so it's easily "grepable". My proposal:

* RXXXXXXX - request id
* SXXXXXXX - session id (or RXXXXX if there is no session)
* U12345 - regular user
* A12345 - admin user

In journald this does not make much difference, but in syslog where the only tool we have is grep this can be quite a difference. Example output:

```
11:34:19 rails.1   | 2017-03-28T11:34:19 R556aa  [app] [I] Started GET "/widgets/20" for ::1 at 2017-03-28 11:34:19 +0200
11:34:19 rails.1   | 2017-03-28T11:34:19 R556aa  [app] [I] Processing by DashboardController#show as HTML
11:34:19 rails.1   | 2017-03-28T11:34:19 R556aa  [app] [I]   Parameters: {"id"=>"20"}
11:34:19 rails.1   | 2017-03-28T11:34:19 R556aa A3 [app] [D] Setting current user thread-local variable to admin
11:34:19 rails.1   | 2017-03-28T11:34:19 R556aa A3 [app] [D] Setting current organization thread-local variable to none
11:34:19 rails.1   | 2017-03-28T11:34:19 R556aa A3 [app] [D] Setting current location thread-local variable to none
11:34:19 rails.1   | 2017-03-28T11:34:19 R556aa A3 [app] [I]   Rendered dashboard/_distribution_widget.html.erb (11.7ms)
11:34:19 rails.1   | 2017-03-28T11:34:19 R556aa  [app] [D] Setting current user thread-local variable to nil
11:34:19 rails.1   | 2017-03-28T11:34:19 R556aa  [app] [I] Completed 200 OK in 28ms (Views: 12.1ms | ActiveRecord: 3.1ms)
```

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -193,6 +193,9 @@ module Foreman
     config.log_level = Foreman::Logging.logger_level('app').to_sym
     config.active_record.logger = Foreman::Logging.logger('sql')
 
+    # Rails info log message filtering (password is filtered by default)
+    config.filter_parameters << :login
+
     if config.serve_static_files
       ::Rails::Engine.subclasses.map(&:instance).each do |engine|
         if File.exist?("#{engine.root}/public/assets")


lib/middleware/tagged_logging.rb
```diff
@@ -5,14 +5,15 @@ def initialize(app)
     end
 
     def call(env)
-      ::Logging.mdc['request'] = env['action_dispatch.request_id']
+      request_id = env['action_dispatch.request_id']
```
lzap: One less hash lookup each request, this is optimalization.

app/models/concerns/foreman/thread_session.rb
```diff
@@ -47,14 +48,12 @@ def current
         end
 
         def current=(o)
-          unless o.nil? || o.is_a?(self)
-            raise(ArgumentError, "Unable to set current User, expected class '#{self}', got #{o.inspect}")
-          end
-
-          if o.is_a?(User)
-            Rails.logger.info "Current user: #{o.login} (#{o.admin? ? 'administrator' : 'regular user'})"
```
iNecas: The problem is the `User.current` set after logging the info about request/params, and it's cleared before we log the info about the response: see sample at https://github.com/theforeman/foreman/pull/4385#issuecomment-287367867. This means we would need to rely on some random log messages from the view to see what user was running that. Keeping at least a info message, not mentioning the login itself, but at least an info if it's a regular user or administrator. could help

app/models/concerns/foreman/thread_session.rb
```diff
@@ -27,11 +27,12 @@ module Cleaner
       def clear_thread
         if Thread.current[:user] && !Rails.env.test?
           Rails.logger.warn("Current user is set, but not expected. Clearing")
-          Thread.current[:user] = nil
+          User.current = nil
```
ohadlevy: why?

ohadlevy:
why would you not want to log the user name? would logging a user id be better ? (I'm just trying to figure out how to use the logs afterwards without a lot of pain).

app/models/concerns/foreman/thread_session.rb
lzap: Why what? You prefer to keep `Thread.current` here?

lzap:
Security, user name is not that important, we need to correlate the logs and user ID is enough IMHO.

app/models/concerns/foreman/thread_session.rb
lzap: I know, I can say we can live with it, this will disappear once we start using journald structured logger. It shows only the message by default. The rest is hidden, you can display it explicitly or search for it.

app/models/concerns/foreman/thread_session.rb
iNecas: 👍 for this change: this way have a logic at once place handling setting the user current

ares:
What are the reasons why we should consider user names private information? They are not hidden by ***** in the form, we print them in hammer info/list commands. I know people can mess up and put their password into login field, but that's the only security risk I see. I don't think other applications try to hide this in logs.

lzap:
I briefly discussed this on #security and several good scenarios were identified. Our production.log is world readable file, also users sometimes accidentaly put their password into username which might expose it. My patch still logs users but using database IDs. We can always pair them if needed. But response was: it depends.

fweimer:
@ares Unknown user names could be passwords if the user enters the incorrectly. Historically, this was used as sufficient justification not log user names which are not known to the system.

lzap:
Meaning, if you insist, I can just close.

iNecas:
I'm ok with filtering the login field to avoid password slips in login field, but as I express, I don't support the filtering of the login once the user is authenticated, until we have much better story about auditing all the actions in the system.

Another question is, if the production.log needs/should be world readable?

ares:
I don't insist, but to me it sounds as obstacle for debugging without clear benefit. When someone uses API for getting user info, the username will be logged as part of URL anyway (e.g. Started GET "/api/users/admin"). That's just my 2c, I can live with either :-)

lzap:
Dropped all the code except filtering out :login form field.

iNecas (approved):
Works for me

ohadlevy:
IMHO sometimes it helps to see who is trying and failing to login, I for one would prefer to log the name if it already exists in our db? also, I dont think our log is world readable?

sean797:
@lzap @ohadlevy you're both kind of right. `/var/log/foreman/production.log` is world readable but because `/var/log/foreman/` isn't world executable I cannot read it. We should probably fix that. I've created http://projects.theforeman.org/issues/19232

```
[vagrant@centos7-katello-3-3 ~]$ sudo ls -ld /var/log/foreman/
drwxr-x---. 3 foreman foreman 4096 Apr  7 10:40 /var/log/foreman/
[vagrant@centos7-katello-3-3 ~]$ sudo ls -l /var/log/foreman/production.log
-rw-r--r--. 1 foreman foreman 73200 Apr 10 08:15 /var/log/foreman/production.log
[vagrant@centos7-katello-3-3 ~]$ tail /var/log/foreman/production.log
tail: cannot open ‘/var/log/foreman/production.log’ for reading: Permission denied

lzap:
@ohadlevy that's too much of a complex code imho, if you don't like full filtering I will close.

PR #4419, opened 2017-03-28
Fixes #19047 - change the last report date format

ares:
Current situation - the problem is user can't tell exact time of a report arrival
reports0

Possible solution 1 - very detailed information (actually asked by user)
reports1
Possible solution 2 - short information, contains enough data IMHO, implemented in this PR
reports2
Possible solution 3 - full information after you hover on "1 month ago"
reports3
Possible solution 4 - combination of full / ago in the same column
reports4

@rohoover your input would be appreciated as well as opinions of others of course

### DIFF
--- a/app/helpers/reports_helper.rb
+++ b/app/helpers/reports_helper.rb
@@ -1,7 +1,7 @@
 require 'ostruct'
 module ReportsHelper
   def reported_at_column(record)
-    link_to(_("%s ago") % time_ago_in_words(record.reported_at), config_report_path(record))
+    link_to(l(record.reported_at, :format => :short), config_report_path(record))
   end
 
   def report_event_column(event, style = "")


lzap:
I like (4)

ohadlevy:
I'm in favor of 3, I dont think it makes a lot of difference if its 1.6 or 1.4 months ago, unless i really care, in which at that time I'm ok with an hover.

Rohoover:
I keep asking myself, is the hover extraneous? I would think hovers to be used in instances where more information is to be provided but it cannot be provided in the original text. If we can already fit the full date in the field and by knowing the date you (hopefully) should be able to work out the time since its Last Report why not just put the full date. Unless people don't care about being specific and want generic terms like "1 month ago".

Didn't we also have a conversation about the possibility that if the date goes over X amount of time, lets say a month, that it would start listing the detailed date to avoid situations like solution 3 where a column of "1 month ago" starts to appear forcing a user to hover to figure out more details vs being immediately discoverable?

ares:
> Didn't we also have a conversation about the possibility that if the date goes over X amount of time, lets say a month, that it would start listing the detailed date to avoid situations like solution 3 where a column of "1 month ago" starts to appear forcing a user to hover to figure out more details vs being immediately discoverable?

Yes, it was on different page. The problem here is that sometimes I need to know exact time regardless whether it was 1 hour or 1 day ago. Sometimes I only need to know that there was 1 report in less than 30 minutes ago. I'm not sure people want to always do the math in such case, which is why "x minutes" ago was introduced.

ares:
Given the space we have in table I'm starting to lean towards (4). We could also split that into two columns or make the second information smaller font or non-clickable so the readability is better. Anyway I'll wait for other comments.

timogoebel:
I prefer 3. I don't like 1 & 2.

ares:
So results so far:

(3): 2 votes
(4): 2 votes

we could also reverse (3) by showing the full date and hover the "ago" form. @Rohoover what's your preference? Do you think full date is better and we should let user to do the math?

dLobatog:
@ares I'm in favor of 1, 2 or 4 because it allows me to see that I'm receiving reports at the proper `runinterval` (if using Puppet). With the option 3 it's not as easy to see this.

ekohl:
IMHO the format should be constant length (like 2017-04-12) so you can easily compare different rows.

xprazak2:
As the previous comments suggest, people have different preferences - some want to see exact date, some regular intervals and others maybe something entirely different. How about we leave the date format selection to users? We could start by introducing one setting with 'interval', 'full' and 'short' options covering cases 'current', '1' and '2' (or any other combination), which would allow users to choose what they prefer.

Granted, it will be more work than one line change but the choice will be left to users and we will not have to look for one solution that fits everyone.

iNecas:
Is there any progress/plan here? Just to add another opinion here, it would be nice to have the way we show times of events consistently across all places. Since it's already shown from the conversation that there are different demands on the way we show the time format, based on what I'm searching for, having an ability to switch between them seems like one way to make everyone happy. Another possibility would be to give the ability to change the way we display format on the page, via some JS kung fu (I can't think though, how to visualize the "format swither" to be intuitive.

It seems, so far we use the `time_ago_in_words` in many places, and it seems pretty reasonable unifier. Perhaps having a helper `format_time` (that for start would use time ago in words, with hoovering) and extending that later from one place would make most sense, and would contribute to unifying the way we show the times in general.

ares:
There was a discussion on foreman-dev, the result IIRC was providing 2 helpers, one for relative, one for absolute time. Then unifying in the whole codebase. Setting to WoC until I get time to finish it.

ekohl:
Django has a [date filter](https://docs.djangoproject.com/en/1.11/ref/templates/builtins/#date) and [timesince filter](https://docs.djangoproject.com/en/1.11/ref/templates/builtins/#timesince) which I always though to be very easy to use. The date filter has some

ekohl:
Django has a [date filter](https://docs.djangoproject.com/en/1.11/ref/templates/builtins/#date) and [timesince filter](https://docs.djangoproject.com/en/1.11/ref/templates/builtins/#timesince) which I always though to be very easy to use. The date filter has some localized formats:

> The format passed can be one of the predefined ones DATE_FORMAT, DATETIME_FORMAT, SHORT_DATE_FORMAT or SHORT_DATETIME_FORMAT, or a custom format that uses the format specifiers shown in the table above.

If you implement a similar convention then you just need to agree on where to use DATE/DATETIME or their SHORT_ versions.

ares:
Replaced by #4820

PR #4420, opened 2017-03-29
Fixes #19064 - seed the external auth source

ares:
To test, configure `authorize_login_delegation_auth_source_user_autocreate` setting to some value (e.g. External) which is normally set by the installer. Run rake db:seed. Then go to user group edit form, you should see external usergroups tab. After you click on "+ Add external user group" you should see "EXTERNAL" in auth source selection.


domcleal (changes requested):
Please address the test failures.

ares:
I updated the tests so they should be green now. I also added a test for the seeding and rebased the branch on top of develop.

ares:
Katello tests should be fixed by https://github.com/Katello/katello/pull/6731

domcleal (approved):

domcleal:
Merged, thanks @ares.

PR #4423, opened 2017-03-30
fixes #19085 - remove test for Token#expires

domcleal:
Token expiry is fully tested in HostTest, as the expiry is actually a
concern on the Host::Managed model. This test fails due to datetime
field precision on MySQL databases with Rails 5, and doesn't appear to
have any value - it only tests a default constructor.


ohadlevy:
thanks @domcleal

PR #4424, opened 2017-03-30
fixes #19087 - create host built notification from Host#built

domcleal:
Comparing the host's installed_at timestamp is unreliable on MySQL
databases, as the timestamp field has a limited granularity. Under Rails
5 in a sub-second test, this field will not be changed.

Instead, extend Host#built to create notifications as the original
method is also used to send e-mail notifications and update the
installed_at field.


ohadlevy:
thanks @domcleal

PR #4425, opened 2017-03-31
Fixes #17588 - Remove 1.16 deprecations

dLobatog:


app/models/setting.rb
```diff
@@ -103,17 +103,6 @@ def self.method_missing(method, *args)
     super
   rescue NoMethodError
```
domcleal: No need to keep this empty method.

domcleal (changes requested):
Please ensure the tests pass.

dLobatog:
Tests passing again

test/models/settings/email_setting_test.rb
```diff
@@ -1,72 +1,29 @@
...
-      assert_equal("example@example.com", Setting[:email_reply_address])
-    end
-
-    test 'delivery_method setting value can be a string' do
-      @delivery_settings["delivery_method"] = 'smtp'
-      load_defaults
-      assert_equal(@delivery_settings["smtp_settings"], Setting::Email.delivery_settings)
-    end
+  setup do
+    Setting::Email.stubs(:mailconfig).returns({})
```
domcleal: This method's being removed, no need to stub it now.

test/models/settings/email_setting_test.rb
```diff
@@ -1,72 +1,29 @@
...
-      Setting::Email.stubs(:mailconfig).returns({})
-    end
-
-    test 'email settings should be editable if email.yaml not exist' do
-      Setting[:delivery_method] = :smtp
-      Setting[:smtp_address] = "sm@example.com"
-      Setting[:smtp_authentication] = "plain"
-      assert_equal({"address" => "sm@example.com", "authentication" => "plain"}, Setting::Email.delivery_settings)
-    end
+  test 'email settings should be editable if email.yaml not exist' do
```
domcleal: No need to mention email.yaml here, it's obsolete. (Suggest renaming this test to simply say it tests the output of `delivery_settings` when settings are modified.)

domcleal (changes requested):
There are some more `email.yaml` leftovers - please remove the template file, and entries in `.gitignore` and `config/spring.rb`.

dLobatog:
Updated- thanks for the review @domcleal

domcleal (approved):

domcleal:
Merged, thanks @dLobatog.

PR #4429, opened 2017-04-02
Fixes #19124 - Temp pin for faraday gem

tbrisker:
until https://github.com/lostisland/faraday/issues/679 is resolved.


tbrisker:
New gem was released while this was testing. Closing.

PR #4431, opened 2017-04-03
Fixes #18561 - faster fact deletion on MySQL

lzap:
Looks like the only option to do delete with join. ActiveRecord does not provide any way to construct a SQL query `DELETE ... JOIN ... WHERE`. This (N+1) is better than doing manual SQL I think.

Note that this is a regression introduced by this patch

https://github.com/theforeman/foreman/commit/46e1ea9f704f41afbd0373e4b8fc54313ac6b397

which main purpose was to bring memory consuption down (N+1 was modified to be inner query which works fine on PostgreSQL). This code is different tho, the query is more simple and we do not load all records in memory, just plucking list of IDs for deletion.

### DIFF
--- a/app/services/fact_importer.rb
+++ b/app/services/fact_importer.rb
@@ -61,8 +61,14 @@ class FactImporter
 
   def delete_removed_facts
     ActiveSupport::Notifications.instrument "fact_importer_deleted.foreman", :host_id => host.id, :host_name => host.name, :facts => facts, :deleted => [] do |payload|
-      # deletes all facts using a single SQL query (with inner query)
-      payload[:count] = @counters[:deleted] = FactValue.joins(:fact_name).where(:host => host, 'fact_names.type' => fact_name_class).where.not('fact_names.name' => facts.keys).delete_all
+      delete_query = FactValue.joins(:fact_name).where(:host => host, 'fact_names.type' => fact_name_class).where.not('fact_names.name' => facts.keys)
+      if ActiveRecord::Base.connection.adapter_name.downcase.starts_with? 'mysql'
+        # MySQL does not handle delete with inner query correctly (slow) so we will do two queries on purpose
+        payload[:count] = @counters[:deleted] = FactValue.delete_all(:id => delete_query.pluck(:id))
+      else
+        # deletes all facts using a single SQL query with inner query otherwise
+        payload[:count] = @counters[:deleted] = delete_query.delete_all
+      end
     end
   end
 


app/services/fact_importer.rb
```diff
@@ -61,8 +61,14 @@ def fact_name_class
 
   def delete_removed_facts
     ActiveSupport::Notifications.instrument "fact_importer_deleted.foreman", :host_id => host.id, :host_name => host.name, :facts => facts, :deleted => [] do |payload|
-      # deletes all facts using a single SQL query (with inner query)
-      payload[:count] = @counters[:deleted] = FactValue.joins(:fact_name).where(:host => host, 'fact_names.type' => fact_name_class).where.not('fact_names.name' => facts.keys).delete_all
+      if ActiveRecord::Base.connection.adapter_name.downcase.starts_with? 'mysql'
+        # MySQL does not handle delete with inner query correctly (slow) so we will do N+1 here on purpose
+        ids = FactValue.joins(:fact_name).where(:host => host, 'fact_names.type' => fact_name_class).where.not('fact_names.name' => facts.keys).pluck(:id)
```
tbrisker: Instead of repeating the same query building with a different last method, how about saving the ` FactValue.joins(:fact_name).where(:host => host, 'fact_names.type' => fact_name_class).where.not('fact_names.name' => facts.keys)` part in a variable and just calling a different method on it?

app/services/fact_importer.rb
```diff
@@ -61,8 +61,14 @@ def fact_name_class
 
   def delete_removed_facts
     ActiveSupport::Notifications.instrument "fact_importer_deleted.foreman", :host_id => host.id, :host_name => host.name, :facts => facts, :deleted => [] do |payload|
-      # deletes all facts using a single SQL query (with inner query)
-      payload[:count] = @counters[:deleted] = FactValue.joins(:fact_name).where(:host => host, 'fact_names.type' => fact_name_class).where.not('fact_names.name' => facts.keys).delete_all
+      if ActiveRecord::Base.connection.adapter_name.downcase.starts_with? 'mysql'
+        # MySQL does not handle delete with inner query correctly (slow) so we will do N+1 here on purpose
```
tbrisker: nitpick: this isn't an N+1 but just two queries instead of one.

tbrisker (changes requested):
See a couple of comments inline. Also, did you perform any benchmarks here comparing before vs. after? I'm curious, might this way be faster on other dbs as well?

lzap:
Amended the variable and comment change.

I did testing, results can be found in the ticket/PR, there is a benchmark test available and I was running it on both SQLite3 and PostgreSQL but not on MySQL.

tbrisker:
@yakirgb can you please confirm this fixes the issue for you?
@lzap I meant about if you tried benchmarking using this change on postgres as well, might be faster anyways or about the same speed, in which case i'd prefer using the same code for all dbs.

yakirgb:
@tbrisker i'll test tomorrow the PR:
https://github.com/theforeman/foreman/pull/4431/files
without changes of: 
`ids = FactValue.joins(:fact_name).where(:host => host, 'fact_names.type' => fact_name_class).where.not('fact_names.name' => facts.keys).pluck(:id)`

yakirgb:
@lzap @tbrisker The PR solved the issue.
```
2017-04-05 06:41:23 62b6447f [audit] [I] [ntwd30601.tbl.com] deleted 24 (9.7ms)
2017-04-05 06:41:23 62b6447f [app] [I] Import facts for 'ntwd30601.tbl.com' completed. Added: 0, Updated: 8, Deleted 24 facts
```
Thanks!

tbrisker (approved):

lzap:
@tbrisker Speed is not relevant for PostgreSQL, the patch that introduced this problem in MySQL was mostly about memory. We were loading all records into memory, now we do subselect for PostgreSQL and full load (but only IDs) for MySQL.

Since we run our tests on all db platforms, I'd rather keep the memory footprint as low as possible and keep separate SQLs for both db systems. The PostgreSQL version is most efficient possible.

tbrisker:
Thanks @lzap !

PR #4432, opened 2017-04-03
fixes #19131 - upgrade minitest to latest 5.x

domcleal:
- override minitest's `_` expectation method with the gettext method
- rename `location` let helper, conflicted with a minitest method name
- fix `assert_equal nil, [..]` deprecation warnings, prefer `assert nil`

Contains fixes for tests that leak data or behavioral changes between
tests, as the ordering of test cases is now randomised:

- remove Host/Nic class modifications in orchestration concern test
- move NotificationBlueprint seeding into test transaction, preventing
  DB truncation in integration tests from removing the records
- ensure DBCleaner.start is called when in transaction mode to prevent
  records leaking out of integration tests

### DIFF
--- a/bundler.d/test.rb
+++ b/bundler.d/test.rb
@@ -1,7 +1,7 @@
 group :test do
   gem 'mocha', '~> 1.1'
   gem 'single_test', '~> 0.6'
-  gem 'minitest', '~> 5.1.0'
+  gem 'minitest', '~> 5.1'
   gem 'minitest-optional_retry', '~> 0.0', :require => false
   gem 'minitest-spec-rails', '~> 5.3'
   gem 'ci_reporter_minitest', :require => false
--- a/test/active_support_test_case_helper.rb
+++ b/test/active_support_test_case_helper.rb
@@ -234,4 +234,11 @@ class ActiveSupport::TestCase
       raise "unknown RestClient::Response.create version (#{RestClient.version}, arity #{RestClient::Response.method(:create).arity})"
     end
   end
+
+  # Minitest provides a "_" expects syntax which overrides the gettext "_" method
+  # Override the minitest method and call the original instead for compatibility
+  # with the app's runtime environment.
+  def _(*args)
+    Object.instance_method(:_).bind(self).call(*args)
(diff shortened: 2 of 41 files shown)


domcleal:
Test failures fixed, isolation between existing tests has been improved in a few places (class modifications, incorrect seeding outside of tests, and database_cleaner not being called in some integration tests.)

tbrisker:
@parthaa could you please take a look at the failing Katello test?

dLobatog:
👍 except for the Katello test, thanks @domcleal

domcleal:
Not sure this is waiting on contributor/me then - they can continue to use the current version if they wish, but it's not an issue with this pull request.

dLobatog:
I think there's a tacit agreement we don't break their CI (if not, what's the point of putting it in our checklist?) and if we do, we should have a fix ready right away. Anyway, it's just one test, I'm writing the fix - once the PR is out I'll merge.

[test katello] - I couldn't reproduce it with just `rake test:katello TEST=../katello/test/controllers/api/v2/activation_keys_controller_test.rb`

dLobatog:
I think  https://github.com/Katello/katello/commit/43f1e810cd63bee37fe50fe735d313bec20b1863 and https://github.com/Katello/katello/commit/ed21e2e7554ca50607f4c4f4e364fd1509493e0b changed the test so the fix isn't necessary, let's see if it's green now

dLobatog:
@domcleal Mind to rebase? Katello tests won't run (`NameError: uninitialized constant Katello::ApplicationRecord`) without https://github.com/theforeman/foreman/commit/2e7dfc3d548b91091c613290eeda8101effe5290

domcleal:
PR is rebased.

dLobatog:
Thanks, I'm trying to reproduce the error on Katello::PulpProxyTest but not really succeeding at it with `rake test:katello TEST=../katello/test/lib/tasks/seeds_test.rb` - I suppose I'll have to run the entire suite

dLobatog:
Thanks @domcleal !

PR #4433, opened 2017-04-04
Fixes #19151 - progress_report_id fix for host edit

lzap:
The whole orchestration skipping when unattended setting is false is weird. This patch fixes this hot problem, we need to backport this into 1.14 and 1.15.

### DIFF
--- a/app/models/host/base.rb
+++ b/app/models/host/base.rb
@@ -337,6 +337,10 @@ module Host
       false
     end
 
+    def orchestrated?
+      self.class.included_modules.include?(Orchestration)
+    end
+
     private
 
     def build_values_for_primary_interface!(values_for_primary_interface, args)
--- a/app/views/hosts/_form.html.erb
+++ b/app/views/hosts/_form.html.erb
@@ -92,7 +92,7 @@
       <%= f.hidden_field :uuid if @host.uuid.present? %>
       <% # Required for import action %>
       <%= f.hidden_field :compute_resource_id if @host.new_record? && @host.uuid.present? %>
-      <%= f.hidden_field :progress_report_id, :data=>{:url=>task_path(@host.progress_report_id)} %>
+      <%= f.hidden_field(:progress_report_id, :data=>{:url=>task_path(@host.progress_report_id)}) if @host.orchestrated? %>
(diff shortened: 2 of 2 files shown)


lzap:
I think this part needs more refactoring, some our code requires `progress_report` to be always present, this is worth separate refactor ticket.

app/models/host/managed.rb
```diff
@@ -197,6 +197,8 @@ class Jail < ::Safemode::Jail
     validate :check_if_provision_method_changed, :on => :update, :if => Proc.new { |host| host.managed }
     validates :uuid, uniqueness: { :allow_blank => true }
   else
+    include Orchestration::ProgressReport
```
domcleal: There shouldn't be any orchestration modules present in Host::Managed when not using unmanaged mode. The need for these methods should be removed, or it should be in Host::Managed proper, or Host::Base if applicable.

domcleal (dismissed):
There don't appear to be any tests.

lzap:
I am not writing any test until we come to agreement. This should do it, it looks like the only appearance is in the edit form, other three are in API but that's not the issue.

app/models/host/base.rb
```diff
@@ -312,6 +312,10 @@ def skip_orchestration?
       false
     end
 
+    def orchestraded?
```
dLobatog: typo, orchestrated

dLobatog (changes requested):
@lzap Is this still happening? I can't recall seeing the error, you just have to create a host with `:unattended => false` to reproduce, or what is it?

lzap:
@mhulan why have you mentioned this patch? it is still a valid bug, but should not be blocker unless you have turned off unattended...

ares:
@lzap the release of the redmine issus is set to 1.16 which equals to blocker for 1.16

dLobatog (changes requested):
@lzap This is basically ready, could you check out the typo and re-submit? Ping me and I'll merge for 1.16 RC3

lzap:
Done

dLobatog (approved):

dLobatog:
Cherry-picked to 1.16-stable as 50b6d64325eaf91e36b393ae6822f7414a57e329

PR #4434, opened 2017-04-04
Fixes #16082 - fact_value now belongs_to host_base

lzap:
Replaces `belongs_to_host` with `belongs_to_host_managed` which is more appropriate name. Keeps the old helper for cases when `Host::Base` is needed. There is one case, in NIC::Base, and the other is introduced: FactValue table which really does belong to Host::Base rather than Host::Managed.

Replaces https://github.com/theforeman/foreman/pull/3724


app/models/fact_value.rb
```diff
@@ -2,7 +2,7 @@ class FactValue < ActiveRecord::Base
   include Authorizable
   include ScopedSearchExtensions
 
-  belongs_to_host
+  belongs_to_host_managed
```
tbrisker: Shouldn't this be on `Host::Base`?

app/models/host/hostmix.rb
```diff
@@ -5,6 +5,10 @@ def has_many_hosts(options = {})
     end
 
     def belongs_to_host(options = {})
+      belongs_to :host, {:class_name => "Host::Base", :foreign_key => :host_id}.merge(options)
```
tbrisker: Might not this change break plugins using this method? maybe keep the same method but make `:class_name` a kwarg that defaults to `Host::Managed`, to more clearly indicate this can be overridden when needed?

app/models/concerns/facets/base.rb
```diff
@@ -5,7 +5,7 @@ module Base
     extend ActiveSupport::Concern
 
     included do
-      belongs_to_host
+      belongs_to_host_managed
```
tbrisker, line 8: not sure if this should be managed or base tbh

app/models/token.rb
```diff
@@ -1,6 +1,6 @@
 class Token < ActiveRecord::Base
   validates_lengths_from_database
-  belongs_to_host :foreign_key => :host_id
+  belongs_to_host_managed :foreign_key => :host_id
```
tbrisker, line 3: not directly related but seems to me that `:foreign_key => :host_id` is superfluous here?

tbrisker (changes requested):
See comments inline - not sure this is the best approach for solving this.

app/models/fact_value.rb
lzap: Oh yeah I screwed it up when editing.

app/models/fact_value.rb
```diff
@@ -2,7 +2,7 @@ class FactValue < ActiveRecord::Base
   include Authorizable
   include ScopedSearchExtensions
 
-  belongs_to_host
+  belongs_to_host_base
   belongs_to :fact_name
   delegate :name, :short_name, :compose, :origin, :to => :fact_name
   has_many :hostgroup, :through => :host
```
tbrisker, line 8: Host::Base doesn't have a relation hostgroup

lzap:
Good concern, created deprecation warning for plugins. I "fixed" tests, hostgroup search is no longer possible because FactValue has no hostgroup association anymore. This also will disable filters on them.

FactValue should never have been associated with managed host in the first place, this is a regression but results are:

- you can no longer search by hostgroup in fact values
- you can no longer create permission filters in fact values

If my understanding is correct, then permissions filters will not work for unmanaged hosts already today.

lzap:
Hmmm just to avoid confusion, I initially thought that Host::Base is unmanaged host, but realized this is determined by "managed" flag on Host::Managed. I am closing this patch for now, I think the proper way out of this is to simply drop STI for facets. @ShimShtein how far are we? Is this good timing to start such an effort (1.15 branched off). I can work on discovery part if needed.

ShimShtein:
@lzap unfortunately still pretty far. We are not done with removing puppet out of core, not to speak of other pretty controversial things like build orchestration.

PR #4436, opened 2017-04-04
Fixes #19148 - Add description field to subnets

tbrisker:


timogoebel (approved):
@tbrisker : Thanks. Code looks fine and feature works well! Ready to merge when tests are 💚 .

app/views/subnets/_fields.html.erb
```diff
@@ -1,4 +1,5 @@
 <%= text_f f, :name %>
+<%= text_f f, :description %>
```
ohadlevy: Text area or field?

tbrisker: right, fixed.

timogoebel: Oh, I thought that was on purpose. :-)

ares:
would you mind also opening hammer-cli-foreman PR? Should be similar to https://github.com/theforeman/hammer-cli-foreman/pull/260

ares:
and last idea - would it make sense to make this available in [ENC output](https://github.com/theforeman/foreman/blob/develop/app/models/subnet.rb#L18)? fair to say no :-)

EDIT: [safe mode](https://github.com/theforeman/foreman/blob/develop/app/models/subnet.rb#L95) should be allowed though

timogoebel (changes requested):
@tbrisker : I think, @ares has a point here. Could you please allow safe mode and add this to enc output while you're at it. Can't hurt.

tbrisker:
@ares @timogoebel Thanks, added to enc and safemode (though not sure there will be much use for it) and created pr for hammer as well

timogoebel:
@tbrisker : One thing I found:

image

But I think, this is unrelated to this PR.

timogoebel (approved):
@tbrisker : Thanks, looks good.

tbrisker:
@timogoebel is that after you stretched it or is that how it loaded for you?

timogoebel:
@tbrisker: It's after I changed the size. It looks good when the page loads.

ares (approved):
Thanks, I have no more comments :+1:

PR #4438, opened 2017-04-05
Fixes #19169 - remove image password from audit

ares:


db/migrate/20170405065305_remove_image_password_audit.rb
```diff
@@ -0,0 +1,20 @@
+class RemoveImagePasswordAudit < ActiveRecord::Migration
+  def up
+    Audited::Audit.where(:auditable_type => 'Image').find_in_batches do |batch|
+      batch.each do |audit|
```
tbrisker: Why not just `find_each`?

db/migrate/20170405065305_remove_image_password_audit.rb
```diff
@@ -0,0 +1,20 @@
...
+        audit.audited_changes.delete("password")
```
tbrisker: no point deleting if we destroy the audit (in case it's just the password that's audited) anyways I think?

db/migrate/20170405065305_remove_image_password_audit.rb
```diff
@@ -0,0 +1,20 @@
...
+        if audit.changed?
```
tbrisker: you could use `.where('audited_changes like ?', '%password%')` to get just the audits that have a password change in them

tbrisker (changes requested):
see inline

db/migrate/20170405065305_remove_image_password_audit.rb
ares: no good reason :-)

db/migrate/20170405065305_remove_image_password_audit.rb
ares: well I could add a condition whether there's only 'password' key but I can simply try deleting the key and checking the hash, if there are no more audit changes then we simply destroy the whole record

db/migrate/20170405065305_remove_image_password_audit.rb
```diff
@@ -0,0 +1,16 @@
+class RemoveImagePasswordAudit < ActiveRecord::Migration
+  def up
+    Audited::Audit.where(:auditable_type => 'Image').where('audited_changes like ?', '%password%').find_each do |audit|
+      audit.audited_changes.delete("password")
+      if audit.audited_changes.empty?
+        audit.destroy
```
ohadlevy: i think we should show an audit entry when the password was changed, just dont show the password itself.

db/migrate/20170405065305_remove_image_password_audit.rb
ares: it won't work reliably since it will also detect audits where other attributes contain value password but it limits some records, I'll update it

db/migrate/20170405065305_remove_image_password_audit.rb
ares: so we also need to add more logic so that we don't ignore password attribute and we create custom audit message, something like we did for user password change

ares:
Thanks @tbrisker and @ohadlevy for quick look, I updated the PR so it keeps the audit but only shows that password was changed. Please take another look
audit

app/models/image.rb
```diff
@@ -19,6 +22,19 @@ class Image < ActiveRecord::Base
   scoped_search :relation => :operatingsystem, :on => :id, :rename => "operatingsystem", :complete_enabled => false, :only_explicit => true, :validator => ScopedSearch::Validators::INTEGER
   scoped_search :on => :user_data, :complete_value => {:true => true, :false => false}
 
+  def password_changed_changed?
```
timogoebel, line 25: We have something similar in User's model. Does it make sense to extract this to a concern to not repeat ourselves?

tbrisker: I had a feeling this isn't the only place where we don't want audits to include passwords. big :+1: to extracting to concern for reuse wherever we have a password field (or possibly even monkey patching it into audited as a second step to automatically redact passwords and other "sensitive" fields)

ares: Proper solution is to add a support in audited gem to handle password fields (no monkey patching). This is a security fix so I'd like to avoid touching other code and do the refactoring in here, we might need to bring it back to older versions of Foreman. Btw there is already an upstream issue for this - https://github.com/collectiveidea/audited/issues/54 feel free to upvote.

timogoebel: Ok. Makes sense to live with the duplication to allow easier backporting. 👍

tbrisker (approved):

app/models/image.rb
tbrisker: I'll look into getting this done properly in the audited gem, willing to take this workaround for now.

tbrisker:
@timogoebel @ohadlevy any further comments?

timogoebel:
@tbrisker : No, feel free to merge this at your discretion.

tbrisker:
Thanks @ares !

PR #4439, opened 2017-04-06
fixes #17895 - do not import link-local ipv6 addr fact

timogoebel:
If facter reports a link-local ipv6 address as an interfaces primary ipv6 address, drop the address from the interface and either remove an existing address or simply don't set the address.


tbrisker:
@timogoebel Mind rebasing this on latest develop? looks like katello tests are failing because of that.

timogoebel:
@tbrisker : Rebased.

lzap (approved):
WFM thanks!

PR #4440, opened 2017-04-06
fixes #19192 adds mark all as read notification action

ohadlevy:
This implements missing feature from patternfly ( see http://www.patternfly.org/pattern-library/communication/notification-drawer/#/code/angular)


ohadlevy:
@matanwerbner could you please have  alook? I would need a bit of help writing the tests as it seems my changes didnt break any existing test ;)

webpack/assets/javascripts/react_app/redux/actions/notifications/index.js
```diff
@@ -41,15 +41,15 @@ export const getNotifications = url => dispatch => {
...
   dispatch({
     type: NOTIFICATIONS_MARK_AS_READ,
     payload: {
       group,
-      id
+      ids
     }
   });
-  API.markNotificationAsRead(id);
+  ids.forEach(id => { API.markNotificationAsRead(id); });
```
matanwerbner: i think it would be more efficient to have an api method accepting an array of IDS, instead of making many seperate calls

webpack/assets/javascripts/react_app/redux/actions/notifications/index.js
```diff
@@ -41,15 +41,15 @@ export const getNotifications = url => dispatch => {
   );
 };
 
-export const onMarkAsRead = (group, id) => dispatch => {
+export const onMarkAsRead = (group, ids) => dispatch => {
```
matanwerbner: i think it makes sense to have a seperate action "onMarkGroupAsRead", that will accept group name, and reducer can map matching notifications to seen.
that way no need to change existing action, and testing will be easier.

webpack/assets/javascripts/react_app/components/notifications/drawer/notificationGroup.js
```diff
@@ -39,9 +40,17 @@ export default (
               onClickedLink={onClickedLink}
               key={notification.id}
               notification={notification}
-              onMarkAsRead={onMarkAsRead.bind(this, group)}
+              onMarkAsRead={onMarkAsRead.bind(this, group, [notification.id])}
```
matanwerbner: if we make a seperate action for marking group, this change is will not be required.

ohadlevy: @matanwerbner thats how I started, but it feels like a lot of boilerplate (nearly duplicate code) to have both mark READ as GROUP and mark READ as event..? nearly everything is doubled, the only thing that is not is the API which can change to accept an array of ids instead of a single id?

matanwerbner: first, it's a different API call (markAsReadById vs markAsReadByGroupName).
the Reducer handles it differently, and it is called from a different place, with different parameters.
so i'm not sure there is any duplication here :)

ohadlevy: why is it a different api call? its either calling the existing api multiple times, or simply we change the server api to receive one or more id's ?

matanwerbner: consider the situation where you have a group with 200 notifications, that you want to mark as seen.
if you have pagination, you dont have the ids of all of them, and even if you did making seperate requests would be in-efficient (one request would also make a very nasty sql query).

it would be better to handle things simply and directly - passing the group name to the server and let it do the heavy lifting. for optimistic concurrency, the reducer can handle the notifications that are loaded in the ui state.

if we want to support pagination - there is no other way.

ohadlevy: @matanwerbner pushed a new version,  prior to making any API changes for feedback.

I do think we need to simplify the amount of places we need to pass props around, maybe start using the ... convention ({...this.props})

ohadlevy:
@matanwerbner pushed a new version, prior to making any API changes for feedback.

I do think we need to simplify the amount of places we need to pass props around, maybe start using the ... convention ({...this.props})

webpack/assets/javascripts/react_app/redux/reducers/notifications/index.js
```diff
@@ -25,12 +26,19 @@ export default (state = initialState, action) => {
...
+      )
+    );
+    case NOTIFICATIONS_MARK_GROUP_AS_READ:
       return state.set(
         'notifications',
         state.notifications.map(
-          n => n.id === payload.id ?
-            Object.assign({}, n, {seen: true}) :
-            n
+          n => payload.ids.includes(n.id) ?
```
matanwerbner: better to do `n => n.group === payload.group ?`
that way its more efficient - you dont have to pass around the ids and iterate on them

webpack/assets/javascripts/react_app/components/notifications/drawer/notificationGroup.js
```diff
@@ -42,6 +44,15 @@ export default (
               onMarkAsRead={onMarkAsRead.bind(this, group)}
             />
           ))}
+          <div className="drawer-pf-action">
+            <a
+              className="btn btn-link btn-block"
+              onClick={onMarkGroupAsRead.bind(this, group,
+                unreadNotifications.map(n => {return n.id;}))}
```
matanwerbner: this third parameter is not needed here

ohadlevy:
@matanwerbner updated with new api.

ohadlevy:
@matanwerbner added rails API tests, could you give me a few examples on how to test the mark as read actions? they were lacking from the original pr.

matanwerbner:
checkout notifications.test.js , it has some "e2e" tests, we can easily add a group test.

if you want unit tests, i have not done for notifications but you can see it in powerstatus: reducers and action creators... i feel like the e2e test is best for this sort of feature..

ohadlevy:
@tbrisker mind having a look? thanks.

webpack/assets/javascripts/react_app/API.js
```diff
@@ -17,6 +17,18 @@ export default {
...
+        console.log(jqXHR);
+      }
+    });
+  },
+  markGroupNotificationAsRead(group) {
+    $.ajax({
+      url: `/notification_recipients/group/${group}`,
+      contentType: 'application/json',
+      type: 'PUT',
+      error: function (jqXHR, textStatus, errorThrown) {
```
tbrisker, line 30: Nitpick: I think this can be something like `(jqXHR) => {console.log(jqXHR);}`, although I'm not sure if the XHR object is what we want logged here?

app/controllers/notification_recipients_controller.rb
```diff
@@ -17,6 +17,19 @@ def destroy
...
+    count = NotificationRecipient.
+      joins(:notification_blueprint).
+      where(user_id: User.current.id, seen: false,
+        notification_blueprints: { group: params[:id]}).
+      update_all(seen: true)
+
+    logger.debug("updated #{count} notification recipents as seen for group #{params[:id]}")
+    UINotifications::CacheHandler.new(User.current.id).clear unless count.zero?
+
+    head status: (count.zero? ? 304 : :ok)
```
tbrisker: please use `:not_modified` instead of 304.

config/routes.rb
```diff
@@ -519,5 +519,9 @@
     end
   end
 
-  resources :notification_recipients, :only => [:index, :update, :destroy]
+  resources :notification_recipients, :only => [:index, :update, :destroy] do
+    collection do
+      put 'group/:id' => 'notification_recipients#update_group_as_read'
```
tbrisker: Since the group is sent by name rather then ID, shouldn't this be `'group/:group` (and `params[:group]` in the controller)?

test/controllers/notification_recipients_controller_test.rb
```diff
@@ -100,10 +100,34 @@ class NotificationRecipientsControllerTest < ActionController::TestCase
...
+    assert_response :success
+    put :update_group_as_read, { :id => 'Testing' }, set_session_user
+    assert_response 304
+  end
+
+  test 'invalid group mark as read' do
+    put :update_group_as_read, { :id => 'unknown;INSERT INTO users (user_id, group)' }, set_session_user
+    assert_response 304
+  end
+
```
tbrisker: Maybe add a test that when you have two groups only one gets marked as read?

tbrisker (changes requested):
A few (minor) comments inline. Also still missing tests for front-end.

webpack/assets/javascripts/react_app/API.js
tbrisker: Also, I'm not a big fan of logging errors to the console, that works for developers but users won't check there usually. Maybe create a notification for the error itself?

ohadlevy:
@tbrisker I've addressed all of your comments excluding the console log statement which I assume deserve a separate discussion / PR.
@matanwerbner I'm having a hard time implementing the tests, could you give me a hint on how to continue? I'm looking for:

- [x] end to end tests
- [x] reducer tests
- [ ] api tests?

anything else?
thanks!

matanwerbner:
@ohadlevy you can have a look at the official docs:
http://redux.js.org/docs/recipes/WritingTests.html
as for the e2e, it is in notifications.test.js , i'll talk with you about extending it.

matanwerbner:
@ohadlevy please try this patch?
it should make tests much much cleaner and also remove the annoying console logs during tests.
[fix_notification_test.txt](https://github.com/theforeman/foreman/files/928407/fix_notification_test.txt)

webpack/assets/javascripts/react_app/API.js
matanwerbner: @tbrisker i agree, we should have a unified way of handling server-side errors

ohadlevy:
@matanwerbner thanks, I've included your suggestions + added redux action testings

webpack/assets/javascripts/react_app/API.js
ohadlevy: @matanwerbner @tbrisker I would prefer another PR to handle this?

matanwerbner:
Looks good on my end!

ohadlevy:
@tbrisker mind having another look? test failure seems unrelated.

tbrisker (approved):

tbrisker:
Thanks @ohadlevy !

PR #4441, opened 2017-04-06
Fixes #16082 - Associate FactValues with Host::Base

tbrisker:
To prevent regression in existing search functionality, this also
requires that the hostgroup relation will be moved to the base class as
well. Since Host::Managed inherits from Host::Base, this should not
break anything.

### DIFF
--- a/app/models/fact_value.rb
+++ b/app/models/fact_value.rb
@@ -2,7 +2,7 @@ class FactValue < ActiveRecord::Base
   include Authorizable
   include ScopedSearchExtensions
 
-  belongs_to_host
+  belongs_to :host, {:class_name => "Host::Base", :foreign_key => :host_id}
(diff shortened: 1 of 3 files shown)


tbrisker:
This is part of an attempt at fixing an N+1 on the fact index page and csv export with discovery enabled. See https://github.com/theforeman/foreman_discovery/pull/337 for more details.

lzap:
This looks good to me, there is some work need to be done on discovery side, so I am not merging until we are sure it helped there. Thanks a log, huge help. This was a pain.

lzap (commented):
Yeah works.

ohadlevy:
@lzap anything else pending?

lzap:
As I said, I am going to test if this helps in discovery. Hold on, I will merge after that.

PR #4443, opened 2017-04-07
Fixes #19213 - use provision interface in foreman_url

iNecas:
Before this patch, we were using primary interface for checking on ftfp
proxy and its template feature. Provision interface is more appropriate
for this purposes.


lzap (approved):
Nice and clean. Waiting for tests.

PR #4444, opened 2017-04-07
fixes #14699 - Enable host edit to accept new params

parthaa:
Prior to this commit:
Foreman plugins like Katello/Discovery extended the Host edit page
and added fields of their own.
For example
Katello adds things like Content Views, enviroments etc.

However they were unable to get their attibutes passed to the backend
because the os_selected and other "*_selected"  methods do something
like
attrs = attribute_hash(['operatingsystem_id', 'organization_id',
                        'location_id']);
sending a hard coded list. This makes it impossible for plugins to send
more data when things like OS or Medium or image get selected.

This commit does the following:

Updates the Host edit js's os_select and other functions
to use a globally available map to figure out which attributes to
send to the server when one calls "medium_selected" "os_selected" etc.

This will help plugins like Katello and foreman-discovery extend the
global map and add their parameters.

For example all katello would have to do now is to say something like
window.tfm.hosts.pluginEditAttributes.os.push("content_view_id")
window.tfm.hosts.pluginEditAttributes.os.
push("kickstart_repository_id")

and when os gets selected these attributes will be automatically
get sent to server.

### DIFF
--- a/app/assets/javascripts/host_edit.js
+++ b/app/assets/javascripts/host_edit.js
@@ -347,7 +347,7 @@ function architecture_selected(element){
   var url = $(element).attr('data-url');
   var type = $(element).attr('data-type');
   var attrs = {};
-  attrs[type] = attribute_hash(['architecture_id', 'organization_id', 'location_id']);
+  attrs[type] = attribute_hash(tfm.hosts.getAttributesToPost("architecture"));
   tfm.tools.showSpinner();
   $.ajax({
     data: attrs,
@@ -366,7 +366,7 @@ function os_selected(element){
   var url = $(element).attr('data-url');
   var type = $(element).attr('data-type');
   var attrs = {};
-  attrs[type] = attribute_hash(['operatingsystem_id', 'organization_id', 'location_id']);
+  attrs[type] = attribute_hash(tfm.hosts.getAttributesToPost("os"));
   tfm.tools.showSpinner();
   $.ajax({
     data: attrs,
@@ -419,7 +419,7 @@ function medium_selected(element){
   var url = $(element).attr('data-url');
   var type = $(element).attr('data-type');
   var attrs = {};
-  attrs[type] = attribute_hash(['medium_id', 'operatingsystem_id', 'architecture_id']);
+  attrs[type] = attribute_hash(tfm.hosts.getAttributesToPost("medium"));
   attrs[type]["use_image"] = $('*[id*=use_image]').attr('checked') == "checked";
   $.ajax({
     data: attrs,
@@ -435,7 +435,7 @@ function use_image_selected(element){
   var url = $(element).attr('data-url');
   var type = $(element).attr('data-type');
   var attrs = {};
-  attrs[type] = attribute_hash(['medium_id', 'operatingsystem_id', 'architecture_id', 'model_id']);
+  attrs[type] = attribute_hash(tfm.hosts.getAttributesToPost("image"));
   attrs[type]['use_image'] = ($(element).attr('checked') == "checked");
   $.ajax({
     data: attrs,
--- a/webpack/assets/javascripts/foreman_hosts.js
+++ b/webpack/assets/javascripts/foreman_hosts.js
@@ -1,0 +1,33 @@
+import _ from 'lodash';
+
+let pluginEditAttributes = {
+  architecture: [],
+  os: [],
+  medium: [],
+  image: []
+};
+
+export function registerPluginAttributes(componentType, attributes) {
+  if (pluginEditAttributes[componentType] !== undefined) {
+    pluginEditAttributes[componentType] = _.uniq(
+      pluginEditAttributes[componentType].concat(attributes));
+  }
+}
+
+export function getAttributesToPost(componentType) {
+  const defaultAttributes = {
+    'architecture': ['architecture_id', 'organization_id', 'location_id'],
+    'os': ['operatingsystem_id', 'organization_id', 'location_id'],
+    'medium': ['medium_id', 'operatingsystem_id', 'architecture_id'],
+    'image': ['medium_id', 'operatingsystem_id', 'architecture_id', 'model_id']
+  };
+  let attrsToPost = defaultAttributes[componentType];
+
+  if (attrsToPost === undefined) {
+    return [];
+  }
+  if (pluginEditAttributes[componentType] !== undefined) {
+      attrsToPost = attrsToPost.concat(pluginEditAttributes[componentType]);
+  }
+  return _.uniq(attrsToPost);
+}
--- a/webpack/assets/javascripts/bundle.js
+++ b/webpack/assets/javascripts/bundle.js
@@ -18,6 +18,7 @@ window.tfm = Object.assign(
     sshKeys: require('./foreman_ssh_keys'),
     trends: require('./foreman_trends'),
     hostgroups: require('./foreman_hostgroups'),
+    hosts: require('./foreman_hosts'),
     toastNotifications: require('./foreman_toast_notifications'),
     numFields: require('./jquery.ui.custom_spinners'),
     reactMounter: require('./react_app/common/MountingService'),
--- a/webpack/assets/javascripts/foreman_hosts.test.js
+++ b/webpack/assets/javascripts/foreman_hosts.test.js
@@ -1,0 +1,21 @@
+jest.unmock('./foreman_hosts');
+const hosts = require('./foreman_hosts');
+
+describe('getAttributesToPost', () => {
+    beforeEach(() => {
+      window.tfm = {hosts: {pluginEditAttributes: {}}};
+  });
+
+  it('attribute hash holds the default attributes', () => {
+    let ret = hosts.getAttributesToPost('os');
+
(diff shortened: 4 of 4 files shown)


parthaa:
see https://github.com/Katello/katello/pull/6746/files to understand the ramifications of this change,

lzap (approved):
Correction, discovery itself does not add extra attributes, it's the Katello plugin which is the problem here as during new discovered provisioning (edit form) the katello content bits are missing then. For more read http://projects.theforeman.org/issues/16063

Anyway, while I don't have issues with this solution, I don't feel this is the correct way (global hash) to do things. I will let JavaScript guys to decide, we should perhaps do some kind of a helper class to manage for all forms.

dLobatog:
@lzap Is this problem found on 1.15? If so - I would prefer to get this in to allow provisioning & discovery with Katello on this version. 

@parthaa Can you think of a way to manage this without using a global variable? I don't think it's the best to make a hard dependency that forces https://github.com/Katello/katello/pull/6746/files to run after this JS file.

parthaa:
@dLobatog the only alternative would be to run an "update_form" on that element which sends all params over. I am fine with that approach too. I see Global var as a stop gap until we rewrite this page in react.

parthaa:
@dLobatog more over this change at least provides an opportunity to set more params and is way better than what katello / other plugins have to do to make this work, i.e. overwrite the method itself...

lzap:
@dLobatog the problem in discovery was fixed (see the ticket), but the patch was in katello codebase not in discovery itself.

jlsherrill:
I think you'd either have a global variable, or a global object that hold the variable and provides some function to add/edit items in the variable, both solutions still require the foreman JS to be loaded first.

app/assets/javascripts/host_edit.js
```diff
@@ -1,5 +1,12 @@
 //= require parameter_override
 
+var HOST_EDIT_ATTRIBUTES = {
+  "architecture" : ['architecture_id', 'organization_id', 'location_id'],
```
ohadlevy: nitpick: why dont write it without the strings, e.g.

```js
var HOST.. = {
  architecture: [...],
}
```
then simply access it by
```js
HOST_EDIT_ATTRIBUTES.architecture
```

ohadlevy: also, can we have a test that checks this? ideally even a test that simulate a plugin extending this.

ohadlevy: another comment, we could move this whole file to be managed under webpack, then you would have a tfm.hosts. (or similar) name space instead of an ugly global var.(and would be a lot easier to test).
I'm more than happy to help if you need guidance on how to do that.

matanwerbner: I personally don't think a global variable is such a bad idea... redux can be used, but it might be an overkill.

however, i don't really like the use of "var", and prefer to always use "const", to make operations immutable.

the following:

`attrs[type] = (window.tfm.globals.host_edit_override || HOST_EDIT_ATTRIBUTES["architecture"]);`

seems to me much cleaner and safer. if you push stuff to the dafault attributes, they are not the default anymore and you could have trouble.

ohadlevy: @matanwerbner please note this is not webpack code, hence vanila js.

matanwerbner: actually.... http://caniuse.com/#feat=const

ohadlevy: @parthaa ping?

ohadlevy: @matanwerbner can you provide an example structure as you shown?

matanwerbner: @ohadlevy 
`window.tfm.globals.host_edit_architectures_override = ['architecture_id', 'organization_id']` 
for instance

parthaa:
@ohadlevy before I touch foreman_hosts any further and try to fix formatting errors in the old host_edit js I want to be sure this is the route you want me to take.
All I did was basically move some of the methods from the host_edit.js to foreman_hosts.js so that I can test em.

I also added a "plugin_edit_attributes" to foreman_hosts which could be used by plugins to pass in extra attributes that they need.

parthaa:
@ohadlevy I made a smaller change in the foreman_hosts instead of bigger refactors. I still have the original change in -> https://github.com/parthaa/foreman/commit/538d11599e9495e88ae06b75dfd5ba7d229c8dd6 . It would require lot more changes to make the lint tools happy and bit riskier of a refactor without proper testing. So the present one provides a smaller footprint and a happy medium

ohadlevy:
@parthaa I believe this is what I was thinking of - thanks! adding @matanwerbner for review.

webpack/assets/javascripts/foreman_hosts.js
```diff
@@ -0,0 +1,39 @@
+export let pluginEditAttributes = {
```
matanwerbner: this object is not imported anywhere, it is not used.

this can be replace with a comment on how to use window.tfm.hosts.pluginEditAttributes by supplying arrays.

webpack/assets/javascripts/foreman_hosts.js
```diff
@@ -0,0 +1,39 @@
...
+  architecture: [],
+  os: [],
+  medium: [],
+  image: []
+};
+
+export function getAttributesToPost(selectionType) {
+  let attrsToPost;
+
+  switch (selectionType) {
```
matanwerbner: this can probably be a bit nicer with a dictionary object:
`const defaultAttributes = { 'architecture': ['architecture_id', 'organization_id', 'location_id'] } `etc...

and loadsh 'get'
`return defaultAttributes[selectionType].concat(get(window, `tfm.hosts.pluginEditAttributes[${selectionType}]`) || []);`

please try to use const instead of 'let', you code will be immutable and much safer

webpack/assets/javascripts/foreman_hosts.js
parthaa: @matanwerbner  if you are interested to know how this is used in katello
https://github.com/Katello/katello/pull/6746/files

matanwerbner: @parthaa apologies, i was confused by the unusual exports.
more common is to do something like:
`
 const hosts = {
  pluginEditAttributes: {...},
  getAttributesToPost: (selectionType) => {...}
}
export default hosts
`
not a big deal though.

parthaa:
[test]

parthaa:
[test foreman]

webpack/assets/javascripts/foreman_hosts.js
```diff
@@ -0,0 +1,26 @@
+export let pluginEditAttributes = {
```
tbrisker: I would prefer keeping the list of attributes internal to this module and expose a method that allows registering more attributes, something like `tfm.hosts.registerAttributes('os', ['lifecycle_environment_id', 'content_view_id', 'environment_id', 'content_source_id', 'architecture_id'])` for example.

webpack/assets/javascripts/foreman_hosts.js
```diff
@@ -0,0 +1,26 @@
...
+    'medium': ['medium_id', 'operatingsystem_id', 'architecture_id'],
+    'image': ['medium_id', 'operatingsystem_id', 'architecture_id', 'model_id']
+  };
+  let attrsToPost = defaultAttributes[selectionType];
+
+  if (attrsToPost === undefined) {
+    return [];
+  }
+
+  if (window.tfm.hosts.pluginEditAttributes[selectionType]) {
```
tbrisker: I think there's no need to fully qualify the `pluginEditAttributes` object here?

tbrisker (changes requested):

webpack/assets/javascripts/foreman_hosts.js
matanwerbner: that would imply a state, after registering you are mutating the hosts attributes, and that can cause problems for other consumers.

webpack/assets/javascripts/foreman_hosts.js
matanwerbner: lodash 'get' can help
`return attrsToPost.concat(get(window, `tfm.hosts.pluginEditAttributes[${selectionType}]`) || [])`

webpack/assets/javascripts/foreman_hosts.js
parthaa: @tbrisker I am having trouble getting this method to work.. Can ya paste me a more complete example.

parthaa:
@matanwerbner  @ohadlevy @tbrisker I ve updated according to your suggestions. Please let me know if you have other suggestions.

parthaa:
please look at https://github.com/Katello/katello/pull/6746/files for example usage.

parthaa:
[test katello]

tbrisker (approved):

tbrisker:
LGTM, @matanwerbner any other comments?

matanwerbner:
Looks good, no comments.

tbrisker:
Thanks @parthaa !

PR #4445, opened 2017-04-09
Fixes #16839 - sort by origin is possible

lizagilman:


tbrisker:
add to PR whitelist

test/factories/facts_related.rb
```diff
@@ -1,6 +1,7 @@
 FactoryGirl.define do
   factory :fact_name do
     sequence(:name) {|n| "fact#{n}" }
+    type 'PuppetFactName'
```
ohadlevy: @lizagilman I would suggest to try without this line, as it turns out break many other tests.

app/controllers/fact_values_controller.rb
```diff
@@ -5,13 +5,7 @@ class FactValuesController < ApplicationController
   before_action :setup_search_options, :only => :index
 
   def index
-    base = resource_base
-    begin
-      values = base.my_facts.search_for(params[:search], :order => params[:order])
-    rescue => e
```
dLobatog, line 11: In my opinion this error handling removal is OK - it's from 2011 and I'd prefer not to hide potential exceptions

app/models/fact_value.rb
```diff
@@ -15,6 +15,7 @@ class FactValue < ActiveRecord::Base
   scoped_search :relation => :host,      :on => :name, :complete_value => true, :rename => :host, :ext_method => :search_by_host_or_hostgroup, :only_explicit => true
   scoped_search :relation => :hostgroup, :on => :name, :complete_value => true, :rename => :"host.hostgroup", :ext_method => :search_by_host_or_hostgroup, :only_explicit => true
   scoped_search :relation => :fact_name, :on => :short_name, :complete_value => true, :aliases => ["fact_short_name"]
+  scoped_search :relation => :fact_name, :on => :type, :complete_value => false, :only_explicit => true, :aliases => ["origin"]
```
dLobatog, line 18: Why `complete_value` is false? There are only 3 or 4 options in a typical installation, which makes it a good candidate for autocompletion IMO. Same for `only_explicit`, why?

dLobatog (changes requested):
Works fairly well, thanks @lizagilman - I left a couple of comments about the scoped_search options.

app/models/fact_value.rb
ohadlevy: The type field value is not user friendly, the initial bug was with sort, fixing the search imho is out of scope for this pr?

dLobatog: Oh I see it now - "type = ForemanAnsible::FactName" is definitely bad... this is a good candidate for `ext_method` though.

dLobatog (approved):

dLobatog:
Merging after seeing the type values are things like "ForemanAnsible::FactName" - which shouldn't be displayed to the user. This should use `ext_method` or something similar to avoid this problem.

dLobatog:
Thanks @lizagilman @ohadlevy

PR #4448, opened 2017-04-10
Fixes #19236 - handle situation when host is destroyed oVirt

iNecas:
The issue was introduced in 4269abbd, where we replaced the
`OVIRT::OvirtException` error by `ActiveRecord::RecordNotFound` in
`find_vm_by_uuid method`, but we haven't updated the corresponding
`rescue` block in `destroy_vm`.


dLobatog:
Thanks @iNecas !

lzap:
I am just wondering if the special /404/ workaround was for some old oVirt version like 3.x which is still in use. Whatever, we will find out :-)

PR #4452, opened 2017-04-12
Fixes #19255 - Edit icon update in the setting page

junezhang:
Update the edit icon in setting page, keep consistent with PatternFly. In this PR, I updated the icon by replacing the original Edit icon image.

Please check the Before&After attachment. 
edit


dLobatog:
Thanks @junezhang !

PR #4454, opened 2017-04-12
Refs #13772 - ensure ApplicationRecord is loaded soon enough

iNecas:
In `lib/core_extensions.rb` we define as `per_page` method, that gets
called by `will_paginate` at load time of the models. The issue is
we are using `Setting` model there, that inherits from the
`ApplicationRecord` but the `ApplicationRecord` is not loaded yet at
that time, which leads to Rails not knowings it's abstract class
and therefore expecting it to be a base for STI.

After this patch, we require the `ApplicationRecord` soon enough to make
sure it's present before using any other model.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -50,6 +50,7 @@ end
 # load the corresponding bit of fog
 require 'fog/ovirt' if defined?(::OVIRT)
 
+require_dependency File.expand_path('../../app/models/application_record.rb', __FILE__)
 require_dependency File.expand_path('../../lib/foreman.rb', __FILE__)
 require_dependency File.expand_path('../../lib/timed_cached_store.rb', __FILE__)
 require_dependency File.expand_path('../../lib/foreman/exception', __FILE__)


lzap (approved):
Works fine here, I wonder why this is not in Rails 5. I am unable to grep it there. I will create new Rails 5 application and compare.

iNecas:
The difference is probably that Rails 5 already define the ApplicationRecord as abstract, but I have not checked the code. The second thing is that the reason we hit this is we override the `per_page` method in `core_extensions` and `will_paginate` calls this method too soon: it's a combination of factors that leads to this effect.

lzap:
Yeah I just figured out, did not read your description at all before waving :-)

lzap:
Not the first time we've been bitten by Setting in early initialization code AFAIR. Very good catch @iNecas I wonder how you tracked it down.

tbrisker:
:+1: pending Mr. Jenkins. Nice fix!

tbrisker (approved):

tbrisker:
Thanks @iNecas !

lzap:
Thanks.

PR #4456, opened 2017-04-13
Fixes #19265 - Change empty settings value color

amirfefer:
Red color may indicates an error, therefore grey color is more suitable:
empty_setting


timogoebel (approved):
Thanks, @amirfefer.

PR #4458, opened 2017-04-14
Fixes #19217 - update the close icon to remove icon

junezhang:
In "Add Parameters[1]" and "Add Combinations[2]" pages, we noticed there are "Remove" icons, but currently using the Close icon, based on the icon guideline, we suggest to use the right icon for correct meaning.
Please check the Before&After attachment as below:
close-remove

[1]Create Host -> Parameters -> Add Parameters
[2]Provisioning Template -> Association -> Add Combinations

P.S. When doing the icon updating, I also modified the capitalization matters.

### DIFF
--- a/app/helpers/form_helper.rb
+++ b/app/helpers/form_helper.rb
@@ -338,7 +338,7 @@ module FormHelper
   end
 
   def link_to_remove_fields(name, f, options = {})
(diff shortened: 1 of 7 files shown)


junezhang:
[test foreman]

junezhang:
@orrabin I found the problem of the last PR is caused by the button class updating. So I haven't updated the button style in this PR. Could you help to test this PR, Thanks.

orrabin:
ok to test

orrabin:
Thanks @junezhang!

PR #4459, opened 2017-04-17
Fixes #19288 - word button instead of icon button

lizagilman:
before: 
image

after: 
image

### DIFF
--- a/app/views/fact_values/_fact.html.erb
+++ b/app/views/fact_values/_fact.html.erb
@@ -24,7 +24,7 @@
   <td class='fact-origin'><%= fact_origin_icon(fact.origin) %></td>
   <td><%= fact_from fact %></td>
   <td><% unless fact.compose %>
-          <%= link_to_function(content_tag(:button, icon_text("adjust", "")),
+          <%= link_to_function(content_tag(:span, "View Chart", :class => "btn btn-sm btn-default"),
                                "get_pie_chart('fact_chart-#{fact.fact_name_id}', '#{fact_path(fact.fact_name_id)}')",
                                :title => _("Show distribution chart")) %>
       <% end %>


ohadlevy:
@rohoover any input? thanks.

Rohoover:
Seems long... what do we think of View Chart?

lizagilman:
viewchart

Rohoover:
I like it. Does this work for you @ohadlevy

app/views/fact_values/_fact.html.erb
```diff
@@ -24,7 +24,7 @@
   <td class='fact-origin'><%= fact_origin_icon(fact.origin) %></td>
   <td><%= fact_from fact %></td>
   <td><% unless fact.compose %>
-          <%= link_to_function(content_tag(:button, icon_text("adjust", "")),
+          <%= link_to_function(content_tag(:span, "Distribution Chart", :class => "btn btn-xs btn-default", :style => "font-size: 10px;"),
```
ares: should the font size be specified in an external CSS file?

ares:
Just one small things, otherwise looks good to me.

app/views/fact_values/_fact.html.erb
lizagilman: this is very specific adjustment because the button text here is too long, I think usually there is no need for that
isn't inline style exactly for such cases? should we create a CSS class just for that? 
But anyway with "View Chart" no need for this adjustment

lizagilman:
updated

app/views/fact_values/_fact.html.erb
ares: I'm trying to avoid inline styling, if changes are required later, it's easier for me to look into one css folder. Also if we wanted to reuse CSS variables (SCSS) for font-size, it would not be possible in inline styling. So yes, in such case I'd prefer new CSS class. But I'm probably not the best person to ask :-)

ares:
Thanks @lizagilman, merged as 23e22260dfc23d15ccc8599f599e6800c83b13e2, the before/after screenshot helped to get this in quickly :+1:

PR #4461, opened 2017-04-18
Fixes #19326 - create compute-resource VMware - URL optional

swapab:
Server-side validation for https://github.com/theforeman/hammer-cli-foreman/pull/272

### DIFF
--- a/app/models/compute_resource.rb
+++ b/app/models/compute_resource.rb
@@ -15,7 +15,6 @@ class ComputeResource < ApplicationRecord
   validates :name, :presence => true, :uniqueness => true
   validate :ensure_provider_not_changed, :on => :update
   validates :provider, :presence => true, :inclusion => { :in => proc { self.providers } }
-  validates :url, :presence => true
   scoped_search :on => :name, :complete_value => :true
   scoped_search :on => :type, :complete_value => :true
   scoped_search :on => :id, :complete_enabled => false, :only_explicit => true, :validator => ScopedSearch::Validators::INTEGER
--- a/app/models/compute_resources/foreman/model/vmware.rb
+++ b/app/models/compute_resources/foreman/model/vmware.rb
@@ -7,6 +7,7 @@ module Foreman::Model
     include ComputeResourceCaching
 
     validates :user, :password, :server, :datacenter, :presence => true
+
     before_create :update_public_key
 
     alias_attribute :server, :url
--- a/test/models/compute_resources/ec2_test.rb
+++ b/test/models/compute_resources/ec2_test.rb
@@ -6,6 +6,8 @@ module Foreman
     class EC2Test < ActiveSupport::TestCase
       include ComputeResourceTestHelpers
 
+      should_not validate_presence_of(:url)
+
       should have_one(:key_pair).with_foreign_key('compute_resource_id').
         dependent(:destroy)
 
--- a/test/models/compute_resources/libvirt_test.rb
+++ b/test/models/compute_resources/libvirt_test.rb
@@ -1,8 +1,10 @@
 require 'test_helper'
 require 'models/compute_resources/compute_resource_test_helpers'
 
-class LibvirtTest < ActiveSupport::TestCase
+class Foreman::Model::LibvirtTest < ActiveSupport::TestCase
   include ComputeResourceTestHelpers
+
+  should validate_presence_of(:url)
 
   test "#associated_host matches any NIC" do
     host = FactoryGirl.create(:host, :mac => 'ca:d0:e6:32:16:97')
--- a/test/models/compute_resources/ovirt_test.rb
+++ b/test/models/compute_resources/ovirt_test.rb
@@ -1,8 +1,14 @@
 require 'test_helper'
 require 'models/compute_resources/compute_resource_test_helpers'
 
-class OvirtTest < ActiveSupport::TestCase
+class Foreman::Model:: OvirtTest < ActiveSupport::TestCase
   include ComputeResourceTestHelpers
+
+  should validate_presence_of(:url)
+  should validate_presence_of(:user)
+  should validate_presence_of(:password)
+  should allow_values('http://foo.com', 'http://bar.com/baz').for(:url)
+  should_not allow_values('ftp://foo.com', 'buz').for(:url)
 
   test "#associated_host matches any NIC" do
     host = FactoryGirl.create(:host, :mac => 'ca:d0:e6:32:16:97')
--- a/test/models/compute_resources/vmware_test.rb
+++ b/test/models/compute_resources/vmware_test.rb
@@ -1,6 +1,18 @@
 require 'test_helper'
 
-class VmwareTest < ActiveSupport::TestCase
+class Foreman::Model::VmwareTest < ActiveSupport::TestCase
+  should validate_presence_of(:server)
+  should validate_presence_of(:user)
+  should validate_presence_of(:password)
+  should validate_presence_of(:datacenter)
+  should allow_values('vcenter.example.com', 'vcenter').for(:server)
+
+  test 'error message is added for server attribute' do
+    vmware_cr = FactoryGirl.build(:vmware_cr, :server => nil)
+    vmware_cr.validate
+    assert_includes vmware_cr.errors.full_messages, "Server can't be blank"
+  end
+
   test "#create_vm calls new_vm when network provisioning" do
     interfaces_attributes = { "new_interfaces" => { "type" => "VirtualE1000",   "network" => "network-17", "_delete" => ""},
                               "0"              => { "type" => "VirtualVmxnet3", "network" => "network-17", "_delete" => ""}}
--- a/app/models/compute_resources/foreman/model/libvirt.rb
+++ b/app/models/compute_resources/foreman/model/libvirt.rb
@@ -4,7 +4,7 @@ module Foreman::Model
 
     ALLOWED_DISPLAY_TYPES = %w(vnc spice)
 
-    validates :url, :format => { :with => URI.regexp }
+    validates :url, :format => { :with => URI.regexp }, :presence => true
     validates :display_type, :inclusion => { :in => ALLOWED_DISPLAY_TYPES }
 
     def self.available?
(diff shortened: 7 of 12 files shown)


timogoebel:
@swapab : Please create a Foreman issue for this and link to it in the commit message. Thanks.

app/models/compute_resource.rb
```diff
@@ -77,6 +77,10 @@ def self.providers_requiring_url
     _("Libvirt, oVirt, OpenStack and Rackspace")
   end
 
+  def self.providers_url_mandatory
+    %w(libvirt ovirt openstack rackspace)
```
timogoebel: Can we make this an option of the Model-class, e.g. `Foreman::Model::Ovirt`? It would be easier for plugins to change this.

timogoebel (changes requested):

app/models/compute_resource.rb
lzap: Frankly I'd squash this with `is_eligible_for_url?` method, unless you have specific reason to have it like that.

timogoebel: Yeah, or that if we want to keep it simple.
How about adding `validates :url, :presence => true` to each Foreman::Model class that needs the validation? I think, that's even better than to filter in the superclass?

swapab: @lzap I originally introduced `providers_url_mandatory` to be able to override it in other plugins(example: https://github.com/theforeman/foreman-docker/pull/190)
I had thought of doing it the way @timogoebel suggested https://github.com/theforeman/foreman/pull/4461#discussion_r112439805.

But right now I am trying to find an answer for @domcleal's question: http://projects.theforeman.org/issues/19326#note-2
Do you guys see any reason for making `url` as required attribute for VMware?

---
I prefer to keep url validation per model, will do the change.

swapab:
@timogoebel I have updated the PR to validate `url` in respective compute-resources.
Parallel docker PR: https://github.com/theforeman/foreman-docker/pull/198

Not able to understand why the **upgrade** build is failing.
The console output is not of much help: http://ci.theforeman.org/job/test_develop_pr_upgrade/3433/console

bundler.d/foreman_docker.rb
```diff
@@ -0,0 +1 @@
+gemspec :path => '../foreman-docker'
```
timogoebel: Please remove this.

test/factories/compute_resources.rb
```diff
@@ -49,7 +49,6 @@
       provider 'Vmware'
       user 'vuser'
       password 'vpassword'
-      sequence(:url) { |n| "#{n}.example.com" } # alias for server
```
timogoebel: Why is this necessary? It looks like a good default to me. You cannot use `sequence(:server)` as FactoryGirl doesn't allow you to use an aliased method.

test/models/compute_resources/libvirt_test.rb
```diff
@@ -11,6 +11,18 @@ class LibvirtTest < ActiveSupport::TestCase
     assert_equal host, as_admin { cr.associated_host(iface) }
   end
 
+  describe 'validate' do
+    test 'valid if url is present' do
+      libvirt_cr = FactoryGirl.build(:libvirt_cr)
+      assert libvirt_cr.valid?, 'LibVirt compute resource expected to be valid'
```
timogoebel: You can use `assert_valid`.

test/models/compute_resource_test.rb
```diff
@@ -224,6 +224,11 @@ def setup
     assert_equal host, as_admin { cr.send(:associate_by, 'mac', '00:22:33:44:55:1a') }
   end
 
+  test 'valid if url is blank and provider is not eligible to have URL' do
+    cr = FactoryGirl.build(:vmware_cr)
```
timogoebel: Could you check that instead the `server` attribute is required and a error message is added for `server`. Maybe `shoulda-matchers` help to reduce the boilerplate.

timogoebel (changes requested):
@swapab: Thanks. I left some inline comments.

test/factories/compute_resources.rb
swapab: Done. Restored sequence.

test/models/compute_resources/libvirt_test.rb
swapab: Done. Thanks.

test/models/compute_resource_test.rb
swapab: Yes makes sense. Moved the test cases to model tests `vmware_test.rb`

test/factories/compute_resources.rb
```diff
@@ -49,7 +49,7 @@
...
```
timogoebel: As written above, you cannot use `sequence(:server)` here as FactoryGirl doesn't allow you to use an aliased method. Please use `sequence(:url)` and restore the comment.

timogoebel (changes requested):

test/factories/compute_resources.rb
swapab: Oops. I misread it. Fixed.

timogoebel: Could you please also restore the comment?
```
sequence(:url) { |n| "#{n}.example.com" } # alias for server
```

test/models/compute_resources/ovirt_test.rb
```diff
@@ -13,6 +13,18 @@ class OvirtTest < ActiveSupport::TestCase
     assert_equal host, as_admin { cr.associated_host(vm) }
   end
 
+  describe 'validate' do
+    test 'valid if url is present' do
```
timogoebel: Can we use shoulda matchers to check this to reduce boilerplate?
Please also add a test for the URL regex. Shoulda matchers should be fine.

```ruby
should allow_values('http://foo.com', 'http://bar.com/baz').
    for(:url)

should_not allow_values('http://foo.com', 'buz').
    for(:url)
```

In addition, i think it's safe to add `https://github.com/theforeman/foreman/blob/develop/app/validators/url_schema_validator.rb` and limit the allowed schemes to http and https.

```ruby
validates :url, :url_schema => ['http', 'https']
```

Please add tests (shoulda matchers) for this, too. Positive and Negative should be fine.

timogoebel (changes requested):
@swapab: Thanks. I left some more comments inline. Could you please take another look? Then we're ready to 🚢 .

test/models/compute_resources/ovirt_test.rb
swapab: `:url_schema => ['http', 'https']` should be only applicable for ovirt? Or should be applied for rackspace, openstack as well?

timogoebel: i definitely know, that vmware is either http or http. I don't know for the others for sure.

swapab: Okay, I have added `url_schema` for VMware as well. And have asked on IRC for some help to know about other CRs.

test/models/compute_resources/ec2_test.rb
```diff
@@ -16,6 +16,13 @@ class EC2Test < ActiveSupport::TestCase
         assert_equal host, as_admin { cr.associated_host(iface) }
       end
 
+      describe 'validate' do
```
timogoebel: I believe, this can also be replaced by
```ruby
should validate_presence_of(:url)
```

Ditto for libvirt & rackspace.

timogoebel (changes requested):

test/models/compute_resources/ec2_test.rb
swapab: Thanks. Done.

app/models/compute_resources/foreman/model/vmware.rb
```diff
@@ -6,7 +6,10 @@ class Vmware < ComputeResource
     include ComputeResourceConsoleCommon
     include ComputeResourceCaching
 
-    validates :user, :password, :server, :datacenter, :presence => true
+    validates :user, :password, :datacenter, :presence => true
+    validates :server, :format => { :with => URI.regexp }, :presence => true,
+              :url_schema => ['http', 'https']
```
timogoebel: Sorry for misleading you here, VMWare should /not/ have the `url_schema` validated. My mistake. Could you remove that please?

test/models/compute_resources/vmware_test.rb
```diff
@@ -1,6 +1,19 @@
 require 'test_helper'
 
-class VmwareTest < ActiveSupport::TestCase
+class Foreman::Model::VmwareTest < ActiveSupport::TestCase
+  should validate_presence_of(:server)
+  should validate_presence_of(:user)
+  should validate_presence_of(:password)
+  should validate_presence_of(:datacenter)
+  should allow_values('http://foo.com', 'http://bar.com/baz').for(:server)
+  should_not allow_values('ftp://foo.com', 'buz').for(:server)
```
timogoebel: Please also remove these tests. Sorry again.

timogoebel (changes requested):
@swapab: Sorry, I asked you to add a wrong validation to VMWare. If you could remove that, I think we're ready to 🚢 .

app/models/compute_resources/foreman/model/vmware.rb
swapab: No problem :smile: 
I have updated the PR accordingly

swapab: Q. Should I also remove the `format` ?

timogoebel: @swapab : Yes, please. Basically we want to support simple hostnames. "vcenter.example.com". Extra points for adding a test for that.

```
should allow_values('vcenter.example.com', 'vcenter').for(:server)
```

swapab: Thanks. Done.

timogoebel:
[test katello]

timogoebel:
@swapab : Could you please rebase the PR? This might fix the failing tests. Thanks.

timogoebel (approved):

timogoebel:
Merged, thanks @swapab.

PR #4462, opened 2017-04-18
Fixes #19295 - use new DHCP API

ares:


timogoebel (approved):
Code looks good to me, thanks @ares!

lib/proxy_api/dhcp.rb
```diff
@@ -91,7 +91,7 @@ def set(subnet, args)
     # [+mac+]    : String in coloned sextuplet format
     # Returns    : Boolean status
     def delete(subnet, mac)
-      parse super("#{subnet}/#{mac}")
+      parse super("#{subnet}/mac/#{mac}")
     rescue RestClient::ResourceNotFound
```
dLobatog: Seems like this rescue would never be reached under normal conditions, 404s are ignored and treated like "OK, it's deleted anyway" 

https://github.com/theforeman/smart-proxy/blob/develop/modules/dhcp/dhcp_api.rb#L139

dLobatog (changes requested):
@ares 👍 to me - can you double check the 404 thing? Thanks for fixing this

lib/proxy_api/dhcp.rb
timogoebel: Yes. Good catch, that did change between the old and new API.

ares: right, I removed the rescue

dLobatog (approved):

dLobatog:
Thanks @ares , @timogoebel !

PR #4463, opened 2017-04-18
Fixes 19298 - Use chevron instead of colons for breadcrumbs

lizagilman:
before: 
colons

after:
chevron

### DIFF
--- a/app/helpers/fact_values_helper.rb
+++ b/app/helpers/fact_values_helper.rb
@@ -10,20 +10,22 @@ module FactValuesHelper
     memo       = ''
     name       = name.split(FactName::SEPARATOR).map do |current_name|
       memo = memo.empty? ? current_name : memo + FactName::SEPARATOR + current_name
-      if parent.present? && h(parent.name) == memo
-        current_name
-      else
-        if value_name != memo || value.compose
-          parameters = { :parent_fact => memo }
-          url = host_parent_fact_facts_path(parameters.merge({ :host_id => params[:host_id] || value.host.name }))
-          link_to(current_name, url,
-                  :title => _("Show all %s children fact values") % memo)
+      content_tag(:li) do
+        if parent.present? && h(parent.name) == memo
+          current_name
         else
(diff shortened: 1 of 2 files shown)


dLobatog:
Thanks @lizagilman ! - I modified the commit title to say "Fixes #" as it's what our automation expected.

lizagilman:
oops :flushed:
thanks @dLobatog

domcleal:
The commit message comment should have been addressed before this was merged. If the commit message is not in the correct format, the Redmine ticket will not be linked and checked.

Also, tests were clearly not run by either the author or on CI, as they fail. Please address comments from @theforeman-bot _before_ merging. See https://github.com/theforeman/foreman/pull/4463#issuecomment-294855690, https://github.com/theforeman/foreman/pull/4463#issuecomment-294855691, https://github.com/theforeman/foreman/pull/4463#issuecomment-294855692. Tests fixed in #4466.

ohadlevy:
also please see the regression this pr introduce - http://projects.theforeman.org/issues/19305

ares:
The problem is that some tests are running even for non-whitelisted PRs and it's easy to miss that only 3 checks succeeded instead of 6 if you see green box. Comments by @theforeman-bot are fine but can be easy missed if there's other communication. Maybe the prprocessor could enforce WoC (or whatever new label) until it can see the commit is fully tested.

dLobatog:
I saw the 'green box' and didn't stop to check whether it was 3 or 6 checks that passed, that's why it got through, sorry. As for the extra empty space above and below the name, I' ll figure out how to remove them as clearly they were not here (https://groups.google.com/forum/?fromgroups#!searchin/foreman-dev/meeting$20facts%7Csort:relevance/foreman-dev/b4GjoGqAcxs/j3-4jjqeDAAJ)

ohadlevy:
@dLobatog the spaces are because of the breadcrumbs css class, we might not need this css class at all for this specific usage case.

PR #4466, opened 2017-04-19
refs #19298 - update fact test to find links in ol/li elements

domcleal:


ohadlevy:
test failure unrelated, thanks @domcleal

PR #4467, opened 2017-04-19
Fixes #19309 - Only select first hosts page on select all

tbrisker:


dLobatog:
@tbrisker Can you add an integration test for this? It's not so clear this wouldn't happen again without tests + I think it's a criteria for [RC patches](http://projects.theforeman.org/projects/foreman/wiki/Release_Management#Issue-tracking) - thanks for spotting it so quickly

tbrisker:
I think @amirfefer is planning to rewrite this whole code in ES6/React, I'd rather not spend a long time writing complex tests for this if the code is going to be replaced soon with code that is much easier to test.

dLobatog:
I'll figure out the tests then: that code in React would not go in for 1.15 but this should be in, and I'd rather make sure it works (possibly the integration tests would test the same behavior with React)

dLobatog:
I'll submit a refs PR with something like this:

```
 test 'select all hosts in all pages' do
        Setting[:entries_per_page] = 1
       host = FactoryGirl.create(:host)
        visit hosts_path
        #page.find('#check_all').trigger('click')
        page.execute_script("$('#check_all').trigger('click')")
                                                                                                                                                # Ensure and wait for all hosts to be checked, and that no unchecked hosts remain
        assert page.has_no_selector?('input.host_select_boxes:not(:checked)')                                                           
        # Suggestion dialog shows up
        assert page.has_selector? "#multiple-alert"
        select_all_hosts = "Select all #{Host.unscoped.count} hosts"
        assert page.has_link? select_all_hosts
        click_link select_all_hosts
        assert page.evaluate_script('$("#multiple-alert").data("multiple")')
        Setting[:entries_per_page] = 20
      end
```

dLobatog:
https://github.com/theforeman/foreman/pull/4473

PR #4468, opened 2017-04-19
Fixes #18687 - restore hash format for parameter attributes

ares:


app/controllers/concerns/parameter_attributes.rb
```diff
@@ -6,6 +6,8 @@ def process_parameter_attributes
     return unless model_params
     parameter_params = model_params["#{parameter_class_mapping}_parameters_attributes"]
     return unless parameter_params
+    # parameters may be either Array or indexed Hash
+    parameter_params = parameter_params.values if parameter_params.is_a?(Hash)
```
domcleal: This check should be against ActionController::Parameters, as Hash is the base class only (which will be removed). Else, invert the conditional and check for Array.

domcleal (changes requested):

app/controllers/concerns/parameter_attributes.rb
ares: thanks, I inverted the check

domcleal (approved):

domcleal:
Merged, thanks @ares.

PR #4469, opened 2017-04-19
Fixes #14178 - patternfly toast notifications

matanwerbner:
this is last finishes on @gailsteiger 's previous work, 

tests have been added, and now possible to pass link to be displayed (see foreman_toast.js)

@waldenraines , @ohadlevy what say you?


tbrisker:
@matanwerbner looks like a bad rebase, please rebase your branch against current develop.

webpack/assets/javascripts/foreman_toast_notifications.js
```diff
@@ -0,0 +1,41 @@
+import * as ToastActions from './react_app/redux/actions/toasts';
+import store from '../javascripts/react_app/redux';
+import $ from 'jquery';
+/* eslint-disable max-params  */
+
+export function notify(message, type, link, sticky) {
+  showToast({
+    message,
+		type,
```
ohadlevy: nitpick: indentation is wrong?

Also, no validation or default values ?

webpack/assets/javascripts/react_app/common/MountingService.js
```diff
@@ -2,6 +2,8 @@ import React from 'react';
 import StatisticsChartsList from '../components/charts/StatisticsChartsList';
 import PowerStatus from '../components/hosts/powerStatus/';
 import NotificationContainer from '../components/notifications/';
+import ToastsList from
```
ohadlevy: nitpick: keep in one line?

webpack/assets/javascripts/react_app/common/MountingService.js
```diff
@@ -2,6 +2,8 @@ import React from 'react';
...
+'../components/toast/';
```
ohadlevy: maybe call it toastNotifications? I think people reading the code will not be sure what is the toast directory? :)

webpack/assets/javascripts/react_app/common/makeUUID.js
```diff
@@ -0,0 +1,9 @@
+export default () => {
```
ohadlevy: is it better to use a package like uuid instead?

webpack/assets/javascripts/react_app/components/common/CloseButton.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import Icon from './Icon';
+
+const CloseButton = ({onClick}) => {
+  return (
+    <button className="close" aria-hidden="true" onClick={onClick}>
```
ohadlevy: q: who is actually closing it? is it bootstrap jquery code somewhere?

webpack/assets/javascripts/react_app/components/toast/ToastList.scss
```diff
@@ -0,0 +1,11 @@
+.toast-notifications-list-pf {
```
ohadlevy: nitpick: tabs vs spaces

ohadlevy: also, why do we need to change patternfly css?

webpack/assets/javascripts/react_app/components/toast/ToastsList.test.js
```diff
@@ -0,0 +1,89 @@
...
+
+  it('single hidden message state', () => {
+    const store = mockStore(singleHiddenMessageState);
+    const box = mount(<ToastList store={store} />);
+
+    expect(box.render().html()).toBe(emptyHtml);
+  });
+});
+
+// it('error', () => {
```
ohadlevy: commented out code?

webpack/assets/javascripts/react_app/components/toast/index.js
```diff
@@ -0,0 +1,36 @@
+import React, { Component } from 'react';
+import Toast from './toastListitem/';
+import * as ToastActions from '../../redux/actions/toasts';
+import { connect } from 'react-redux';
+import './ToastList.scss';
```
ohadlevy: why not index.scss?

webpack/assets/javascripts/react_app/components/toast/toastListitem/Toast.consts.js
```diff
@@ -0,0 +1 @@
+export const defaultTimerDelay = 8000;
```
ohadlevy: q: why not a default toast list prop instead?

webpack/assets/javascripts/react_app/components/toast/toastListitem/index.js
```diff
@@ -0,0 +1,83 @@
...
+    const { sticky, dismiss, timerDelay } = this.props;
+
+    if (!sticky) {
+      this.timer = new Timer(dismiss, timerDelay || defaultTimerDelay);
+      this.timer.startTimer();
+    }
+  }
+
+  componentWillUnmount() {
+    if (this.timer) {
```
ohadlevy: cant we move this condition into the clearTimer function?

webpack/assets/javascripts/react_app/components/toast/toastListitem/index.js
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  onMouseEnter() {
+    if (this.timer) {
+      this.timer.clearTimer();
+    }
+  }
+
+  onMouseLeave() {
+    if (this.timer) {
```
ohadlevy: dry

webpack/assets/javascripts/react_app/components/toast/toastListitem/index.js
```diff
@@ -0,0 +1,83 @@
...
+      this.timer.startTimer();
+    }
+  }
+
+	linkMarkup(title, href) {
+		return (
+			<div className="pull-right toast-pf-action">
+        <a href={ href } >
```
ohadlevy: indentation is wrong.

webpack/assets/javascripts/react_app/components/toast/toastListitem/index.js
```diff
@@ -0,0 +1,83 @@
...
+    const { type, link, message, dismiss } = this.props;
+
+    return (
+      <div
+        className={`alert alert-${type} toast-pf`}
+        onMouseEnter={this.onMouseEnter}
+        onMouseLeave={this.onMouseLeave}>
+        <CloseButton onClick={dismiss} />
+				{
+					link && this.linkMarkup(link.title, link.href)
```
ohadlevy: indentation is wrong

webpack/assets/javascripts/react_app/components/toast/toastListitem/index.js
```diff
@@ -0,0 +1,83 @@
...
+}
+Toast.propTypes = {
+  message: React.PropTypes.string.isRequired
+};
+
+Toast.defaultProps = {
+  dismissable: true,
+  visible: true,
+  type: 'success',
+  timerDelay: 8000
```
ohadlevy: overlap with the constant?

webpack/assets/javascripts/react_app/components/toast/toastListitem/index.js
```diff
@@ -0,0 +1,83 @@
...
+    helpers.bindMethods(this, [
+      'onMouseEnter',
+      'onMouseLeave'
+    ]);
+  }
+
+  componentDidMount() {
+    const { sticky, dismiss, timerDelay } = this.props;
+
+    if (!sticky) {
```
ohadlevy: q: i assume this is implemented this way before redux implementation, but just wondering, would it be better to have the timer as an action instead?

webpack/assets/javascripts/react_app/redux/reducers/toasts/index.js
```diff
@@ -0,0 +1,35 @@
...
+const initialState = Immutable({
+  messages: {}
+});
+
+export default (state = initialState, action) => {
+  const { payload } = action;
+
+  switch (action.type) {
+    case TOASTS_ADD: {
+			return state.setIn(
```
ohadlevy: indentation....

can we fix hound to warn about this?

ohadlevy (changes requested):

ohadlevy:
@matanwerbner please also note your rebase is wrong, it still contains unrequired commits from develop.

webpack/assets/javascripts/foreman_toast_notifications.js
matanwerbner: the only field i can think of validating is the "type"... what is the logic for that?

webpack/assets/javascripts/react_app/common/makeUUID.js
matanwerbner: doesn't really matter...one less dep is better imo

webpack/assets/javascripts/react_app/components/common/CloseButton.js
matanwerbner: clicking it is dispatching an action which changes the state of the notification to visible: false, and then it doesn't get rendered

webpack/assets/javascripts/react_app/components/toast/ToastList.scss
matanwerbner: the close button needs some margin from the text, and has annoying outline when clicking.

webpack/assets/javascripts/react_app/components/toast/index.js
matanwerbner: for index.js there is a reason, import will default to it (import './toast/') .
for index.scss it will not happen, so it is better to have a descriptive name.

webpack/assets/javascripts/react_app/components/toast/toastListitem/Toast.consts.js
matanwerbner: i dont like that defaultProps so much, it is easier to see the const imported from the consts file, then to look for it in the default props value section (that is react dependant, and could get deprecated)

webpack/assets/javascripts/react_app/components/toast/toastListitem/index.js
matanwerbner: nope, it will throw if this.timer is not initialized :)

webpack/assets/javascripts/react_app/components/toast/toastListitem/index.js
matanwerbner: how would you do that? an action is an action and a time is a timer.
at first i was eager to make changes to the timer also, but the more i thought about it, the more it makes sense... and it can be reused in other places in the app...

webpack/assets/javascripts/react_app/components/toast/toastListitem/index.js
matanwerbner: it's a one liner :)

matanwerbner:
@ohadlevy rebase again, please check?

app/assets/javascripts/dashboard.js
```diff
@@ -69,10 +69,10 @@ function save_position(path){
         url: path,
         data: {'widgets': positions},
         success: function(){
-            $.jnotify(__("Widget positions successfully saved."), 'success', false);
+            tfm.toastNotifications.notify(__("Widget positions successfully saved."), 'success', { title: 'lalal', href: 'http://www.google.com' });
```
ohadlevy: leftover?

test/integration/parameters_permissions_test.rb
```diff
@@ -1,6 +1,6 @@
 require 'integration_test_helper'
 
-class ParametersPermissionsIntegrationTest < ActionDispatch::IntegrationTest
+class ParametersPermissionsIntegrationTest < IntegrationTestWithJavascript
```
ohadlevy: I would suggest to revert this change.

webpack/assets/javascripts/foreman_toast_notifications.js
```diff
@@ -0,0 +1,40 @@
...
+export function notify(message, type, link, sticky) {
+  showToast({
+    message,
+    type,
+    sticky,
+    link
+  });
+}
+
+function showToast(toast) {
```
ohadlevy: should this function be exported?

webpack/assets/javascripts/react_app/common/makeUUID.js
ohadlevy: patternfly depends on uuid, so i would suggest we use uuid package too.

webpack/assets/javascripts/react_app/components/toastNotifications/ToastList.scss
```diff
@@ -0,0 +1,11 @@
+.toast-notifications-list-pf {
```
ohadlevy: identification is wrong.

webpack/assets/javascripts/react_app/redux/actions/toasts/index.js
```diff
@@ -0,0 +1,20 @@
+import { TOASTS_ADD, TOASTS_HIDE, TOASTS_DELETE } from '../../consts';
+import makeUUID from '../../../common/makeUUID';
+
+export const addToast = toast => {
+  return {
+    type: TOASTS_ADD,
+    payload: {
+      key: makeUUID(),
```
ohadlevy: do we really need uuid here? would timestamp work too?

matanwerbner: It's needed for react implementation - removing adding etc. 
timestamp seems kind of hackish, also not sure if server supports it...

webpack/assets/javascripts/react_app/components/toastNotifications/ToastList.scss
matanwerbner: what do you mean?

webpack/assets/javascripts/react_app/redux/actions/toasts/index.js
ohadlevy: @matanwerbner afaiu server has no function here, this is purely in javascript?

webpack/assets/javascripts/react_app/components/toastNotifications/ToastList.scss
ohadlevy: its 8 or more chars of spaces (or tabs).

webpack/assets/javascripts/foreman_toast_notifications.js
matanwerbner: yes, since it is "required" into window.tfm in bundle.js

test/integration/parameters_permissions_test.rb
matanwerbner: i'm not sure what it does, i can revert if you are sure it's ok

webpack/assets/javascripts/foreman_toast_notifications.js
ohadlevy: I know, I was trying to say in a polite way its not currently exported ;)

webpack/assets/javascripts/react_app/redux/actions/toasts/index.js
matanwerbner: i mean when importing bulk messages from rails they do not have a timestamp

webpack/assets/javascripts/foreman_toast_notifications.js
matanwerbner: ah i get it..
**notify** is exported, and is used in the app. 
showToast is inner implementation both exported and used for import from rails.

ohadlevy: yes, but since we want to allow katello and @waldenraines to use it, i think we should export it too?

matanwerbner: we can, i like notify since it creates the toast message from distinct parameters, which makes it more safe imho.
anyone can use notify by requiring foreman_toast_notifications

ohadlevy: ah, i missed that you added link in there, in that case its fine.

webpack/assets/javascripts/react_app/components/toastNotifications/ToastList.scss
matanwerbner: fixed.

matanwerbner:
@ohadlevy anything else? can we merge?

test/integration/parameters_permissions_test.rb
tbrisker: This changes the driver that runs the test to one that supports JS (phantomjs). If these tests need JS working to pass this change is needed, but I don't think this is related here.

webpack/assets/javascripts/foreman_toast_notifications.js
```diff
@@ -0,0 +1,40 @@
+import * as ToastActions from './react_app/redux/actions/toasts';
+import store from '../javascripts/react_app/redux';
+import $ from 'jquery';
+/* eslint-disable max-params  */
+
+export function notify(message, type, link, sticky) {
```
tbrisker: shouldn't `link` and `sticky` have default values?

webpack/assets/javascripts/foreman_toast_notifications.js
```diff
@@ -0,0 +1,40 @@
...
+  store.dispatch(ToastActions.addToast(toast));
+}
+
+// to accommodate rails flash syntax
+function importFlashMessagesFromRails() {
+  const notifications = $('#notifications').data().flash;
+
+  notifications.forEach(([type, message]) => {
+    // normalize rails flash names
+    if (type === 'danger') {
```
tbrisker: I would move this normalization to the `notify` function in case some plugin uses `danger` instead of `error`

webpack/assets/javascripts/foreman_toast_notifications.js
```diff
@@ -0,0 +1,40 @@
...
+      type = 'error';
+    }
+
+    showToast({
```
tbrisker: If you set default values in `notify` you could simplify this by replacing it with `notify`.

webpack/assets/javascripts/react_app/components/toastNotifications/index.js
```diff
@@ -0,0 +1,36 @@
...
+import * as ToastActions from '../../redux/actions/toasts';
+import { connect } from 'react-redux';
+import './ToastList.scss';
+
+class ToastsList extends Component {
+
+  render() {
+    const { messages, hideToast } = this.props;
+
+    const visibleToasts = Object.keys(messages).reduce((curr, next) => {
```
tbrisker: Why not something like `Object.values(messages).find( (val) => (val.visible))`? It would also simplify the map below for returning the visible notifications.

webpack/assets/javascripts/react_app/redux/reducers/toasts/index.js
```diff
@@ -0,0 +1,28 @@
...
+  switch (action.type) {
+    case TOASTS_ADD: {
+      return state.setIn(['messages', payload.key], payload.message);
+    }
+
+    case TOASTS_HIDE: {
+      return state.setIn(['messages', payload.key, 'visible'], false);
+    }
+
+    case TOASTS_DELETE: {
```
tbrisker: I wonder, is there any point in having both hide and delete? after clicking the X there is no way to show the notification anyways, why not unify this? It will also let us simplify the implementation since we don't have to check visibility at all.

webpack/stories/index.js
```diff
@@ -112,3 +114,29 @@ storiesOf('Power Status', module)
       />
     )
   );
+
+function dismiss() {
+  console.log('callback executed');
```
tbrisker: It makes more sense to me to show this somehow without requiring the console to be open - maybe an alert?

tbrisker (changes requested):

webpack/stories/index.js
gailsteiger: You might want to take a look at the [patternfly-react](https://jgiardino.github.io/patternfly-react/?knob-Label=Danger%20Will%20Robinson%21&selectedKind=Alert&selectedStory=with%20danger&full=0&down=1&left=1&panelRight=0&downPanel=kadirahq%2Fstorybook-addon-actions%2Factions-panel) implementation using an action logger

webpack/assets/javascripts/foreman_toast_notifications.js
matanwerbner: they will be undefined if not sent, which will be evaluated the same as false or null, i cant see any reason to use default values here.

webpack/assets/javascripts/foreman_toast_notifications.js
matanwerbner: my goal is not to use notify, but to export it to consumers.
since there is different logic to notify then that of import (like sticky if type===success, or danger => error) then that logic is separated between the two functions, which then both call the dispatch method.

webpack/assets/javascripts/foreman_toast_notifications.js
matanwerbner: is there a reason anyone will use "danger" when they mean "error"? 
i think a better option is to throw an exception that says "unknown notification type, please use .."

webpack/assets/javascripts/react_app/components/toastNotifications/index.js
matanwerbner: then i would loose the key(uuid) of the message, which i need later on...

webpack/assets/javascripts/react_app/redux/reducers/toasts/index.js
matanwerbner: I agree...

webpack/stories/index.js
matanwerbner: done.

webpack/assets/javascripts/foreman_toast_notifications.js
tbrisker: Some plugins may still use `notify` in their js code, possibly with the wrong class, we don't want to break that ability or cause their notifications to break. To me it also makes sense to have sticky default to `type!=='success'` if not otherwise specified, so that this is consistent across all notifications and not just those that arrive from rails.

webpack/assets/javascripts/foreman_toast_notifications.js
tbrisker: We don't want to break plugins if we don't have to.

webpack/assets/javascripts/react_app/redux/reducers/toasts/index.js
tbrisker: Then please unify them (unless @ohadlevy sees something we're missing here?)

webpack/stories/index.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 
-/* eslint-disable no-unused-vars */
+/* eslint-disable no-unused-vars, no-console */
```
tbrisker: no longer needed

webpack/assets/javascripts/foreman_toast_notifications.js
matanwerbner: shouldn't we give consumers the ability to choose if their message is sticky or not?

webpack/assets/javascripts/foreman_toast_notifications.js
matanwerbner: removed.

app/assets/javascripts/application.js
```diff
@@ -375,16 +355,9 @@ function spinner_placeholder(text){
...
-                  classBackground: "",
-                  sticky: (type != 'success'),
-                  type: type };
-  $.jnotify("</div>" + icon + $(item).text(), options);
-  // jnotify does not support multiple classes passed via classMessage so added via jquery.
-  $('.foreman-alert').addClass("alert alert-" + type);
-  $(item).remove();
+function notify(message, type) {
+  tfm.tools.deprecate('notify', 'tfm.toastNotifications.notify');
+  tfm.toastNotifications.notify({message: message, type: type});
```
tbrisker, line 360: here we pass the parameters as a hash, but the function expects regular parameters.

app/helpers/reactjs_helper.rb
```diff
@@ -1,5 +1,5 @@
 module ReactjsHelper
-  def mount_react_component(name, selector, data)
+  def mount_react_component(name, selector, data = [])
```
tbrisker, line 2: I'm curious, is there any reason not to pass the data directly when mounting the component instead of having to read it from the `data-flash` attribute afterwards?

webpack/assets/javascripts/foreman_toast_notifications.js
tbrisker: Is there any reason not to unify these two functions? all this function does is pass it's parameter to another function...

webpack/assets/javascripts/foreman_toast_notifications.js
```diff
@@ -0,0 +1,30 @@
+import * as ToastActions from './react_app/redux/actions/toasts';
+import store from '../javascripts/react_app/redux';
+import $ from 'jquery';
+/* eslint-disable max-params  */
+
+export function notify(message, type, link, sticky = type !== 'success') {
```
tbrisker: Wee comment above about un-matching function signature. I tend to prefer the hash version when there are many possible parameters, that way it's clear which parameter is which (and you will be able to remove the `eslisnt-disable` above)

webpack/assets/javascripts/react_app/redux/reducers/toasts/index.js
```diff
@@ -0,0 +1,24 @@
...
+const initialState = Immutable({
+  messages: {}
+});
+
+export default (state = initialState, action) => {
+  const { payload } = action;
+
+  switch (action.type) {
+    case TOASTS_ADD: {
+      return state.setIn(['messages', payload.key], payload.message);
```
tbrisker, line 13: shouldn't this be `toasts` instead of `messages`?

webpack/assets/javascripts/react_app/components/toastNotifications/toastListitem/index.js
```diff
@@ -0,0 +1,55 @@
...
+  constructor(props) {
+    super(props);
+    helpers.bindMethods(this, ['onMouseEnter', 'onMouseLeave']);
+  }
+
+  componentDidMount() {
+    const { sticky, dismiss, timerDelay } = this.props;
+
+    if (!sticky) {
+      this.timer = new Timer(dismiss, timerDelay || defaultTimerDelay);
```
tbrisker, line 18: Shouldn't the default delay be part of the Timer component?

webpack/assets/javascripts/react_app/components/toastNotifications/index.js
```diff
@@ -0,0 +1,30 @@
...
+
+class ToastsList extends Component {
+
+  render() {
+    const { messages, deleteToast } = this.props;
+
+    return (
+      <div className="toast-notifications-list-pf">
+        {
+          map(
```
tbrisker, line 15: Any reason not to use `messages.map`?

webpack/assets/javascripts/react_app/components/toastNotifications/toastListitem/Toast.consts.js
```diff
@@ -0,0 +1 @@
+export const defaultTimerDelay = 8000;
```
tbrisker, line 1: Note that currently the delay is 2000, not sure if this is intentional or not

tbrisker (changes requested):

app/helpers/reactjs_helper.rb
ohadlevy: @tbrisker yes, you are right that if we were using flash only kind of notifications, we could just pass the attributes, but that quickly breaks for JS generated events (e.g. dashboard or other users that use the notify function). as you would need to alter the existing notification state.

tbrisker: is it possible to only use this to create the initial state?

webpack/assets/javascripts/foreman_toast_notifications.js
matanwerbner: most of the existing code uses the un-hash version, so we might cause some problems with plugins.
we can solve this by supporting both hash and parameters (check if first parameter is an object etc.)

webpack/assets/javascripts/react_app/components/toastNotifications/toastListitem/index.js
matanwerbner: I dont think so, since other components may want to use the timer with different default delays

webpack/assets/javascripts/react_app/components/toastNotifications/index.js
matanwerbner: messages is not an array... it's a hash. lodash map iterates over hashes

webpack/assets/javascripts/react_app/redux/reducers/toasts/index.js
matanwerbner: i prefer pancakes... seriously, we can name it how we like, but the word toast just feels annoying after a while, and not really software-related

tbrisker: I'm just concerned that we may have other messages in the store in the future and want to prevent possible confusion.

app/assets/javascripts/application.js
matanwerbner: fixed.

webpack/assets/javascripts/react_app/redux/reducers/toasts/toasts.fixtures.js
```diff
@@ -0,0 +1,35 @@
+import Immutable from 'seamless-immutable';
+
+export const initialState = Immutable({
+  messages: {}
+});
+
+export const messageBeforeAdd = {
+  type: 'success',
+  message: 'first message',
+  visible: true
```
tbrisker: `visible` was removed

webpack/assets/javascripts/react_app/redux/reducers/toasts/index.js
tbrisker: note to self: disregard, it's `toasts.messages`.

webpack/assets/javascripts/react_app/components/toastNotifications/index.js
tbrisker: Why not make it an array actually? that would also remove the need for uuids as you could just use the array index as the key.

tbrisker (changes requested):
almost there, just a couple of minor things left.

matanwerbner:
@tbrisker this will actually not work....
See
https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318

Tl;dr list items need to have a unique key that is not their list index

matanwerbner:
@tbrisker anything else?

webpack/assets/javascripts/react_app/components/toastNotifications/index.js
matanwerbner: @tbrisker this will actually not work....
See
https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318

Tl;dr list items need to have a unique key that is not their list index

webpack/assets/javascripts/foreman_toast_notifications.js
```diff
@@ -0,0 +1,25 @@
+import * as ToastActions from './react_app/redux/actions/toasts';
+import store from '../javascripts/react_app/redux';
+import $ from 'jquery';
+/* eslint-disable max-params  */
```
ohadlevy: not required anymore?

webpack/assets/javascripts/react_app/components/toastNotifications/toastListitem/Toast.consts.js
```diff
@@ -0,0 +1 @@
+export const defaultTimerDelay = 2000;
```
ohadlevy: according to http://www.patternfly.org/pattern-library/communication/toast-notifications/ this should be 8 seconds.

webpack/assets/javascripts/react_app/components/toastNotifications/toastListitem/Toast.scss
```diff
@@ -0,0 +1,9 @@
+.toast-pf.alert {
+  .close {
```
ohadlevy, line 2: just wondering, why is this css required? shouldnt it be in patternfly?

ohadlevy:
@matanwerbner I'm also getting the following error when trying out storybook:
```
ERROR in ENOENT: no such file or directory, scandir '/home/ohad/git/foreman/node_modules/node-sass/vendor'
 @ ./webpack/assets/javascripts/react_app/components/toastNotifications/toastListitem/Toast.scss 4:14-168 13:2-17:4 14:20-174

ERROR in ENOENT: no such file or directory, scandir '/home/ohad/git/foreman/node_modules/node-sass/vendor'
 @ ./webpack/assets/javascripts/react_app/components/hosts/powerStatus/powerStatusInner/PowerStatusInner.scss 4:14-185 13:2-17:4 14:20-191

ERROR in ENOENT: no such file or directory, scandir '/home/ohad/git/foreman/node_modules/node-sass/vendor'
 @ ./webpack/assets/javascripts/react_app/components/notifications/notifications.scss 4:14-170 13:2-17:4 14:20-176
```

ohadlevy:
@matanwerbner regarding the scss error a yarn (or npm) upgrade fixed it, i wonder if it my env or rather a new minimum dep somewhere.

ohadlevy:
@matanwerbner please apply the following diff to get foreman tests running:
```diff
diff --git a/test/integration/parameters_permissions_test.rb b/test/integration/parameters_permissions_test.rb
index 6fa0be0..65fbdc4 100644
--- a/test/integration/parameters_permissions_test.rb
+++ b/test/integration/parameters_permissions_test.rb
@@ -53,7 +53,6 @@ class ParametersPermissionsIntegrationTest < ActionDispatch::IntegrationTest
       within parameter_row_css do
         click_link 'Delete'
       end
-      assert page.has_text?('Successfully deleted a_parameter.')
 
       assert page.has_no_css? "tr#common_parameter_#{@visible_global_parameter.id}_row"
     end
@@ -112,7 +111,6 @@ class ParametersPermissionsIntegrationTest < ActionDispatch::IntegrationTest
 
     fill_in "domain_parameter_#{parameter.id}_value", :with => 'new_value'
     click_button 'Submit'
-    assert_includes page.text, "Successfully updated #{domain.name}."
 
     visit edit_domain_path(domain)
     parameter_area = find(:css, "#domain_parameter_#{parameter.id}_value")
```

ohadlevy:
@tbrisker @matanwerbner a question, should we allow to pass html as a message? 
currently AFAIU this implementation disallow it, i could see notifications in the ui (compute resource check connection) with
```html
<p>Test connection was successful</p>
```

ohadlevy:
@matanwerbner also note that with out current "two-pane" the notifications do not pop up correctly, I'm assuming we are using the wrong event, to reproduce, goto configure -> global params and CRUD some objects.

webpack/assets/javascripts/react_app/components/toastNotifications/toastListitem/Toast.scss
matanwerbner: you're right, i'll remove the margin, but the outline still bugs me so i'll keep it

matanwerbner:
@ohadlevy can you elaborate your last comment?

matanwerbner:
@ohadlevy we should never allow un-sanitized html injected into our application, for obvious security reasons.

ohadlevy:
@matanwerbner that's obvious, but in the process we now do not allow the rails application to inject html too (which is currently required / used).

matanwerbner:
@ohadlevy i'll enable this in favor of backward compatibility, i still feel it is a security weakness, and may easily mess up the UI.

webpack/assets/javascripts/react_app/redux/reducers/toasts/toasts.fixtures.js
tbrisker: This still needs removing.

webpack/assets/javascripts/react_app/redux/reducers/toasts/toasts.fixtures.js
```diff
@@ -0,0 +1,35 @@
...
+};
+
+export const stateAfterAdd = Immutable({
+  messages: {
+    1: {
+      type: 'success',
+      message: 'first message',
+      visible: true
```
tbrisker: here too

webpack/assets/javascripts/react_app/redux/reducers/toasts/toasts.fixtures.js
```diff
@@ -0,0 +1,35 @@
...
+    }
+  }
+});
+
+export const stateAfterHide = Immutable({
```
tbrisker: This state is no longer relevant

tbrisker (changes requested):
Just a bit of leftovers. I agree with @ohadlevy - we need to allow the server side to use HTML in notifications, for example to mark part of it as bold, or even just to keep the compatibility with current implementation. The caller has to take care of escaping any user input.

matanwerbner:
@ohadlevy @tbrisker anything else?

tbrisker (approved):

tbrisker:
Thanks @gailsteiger, @ohadlevy and @matanwerbner !

PR #4471, opened 2017-04-20
Fixes #19325 - Adding VMware vSphere SCSI Controller ID Selection to volumes

sbernhard:
Adding VMware vSphere SCSI Controller ID Selection to volumes

### DIFF
--- a/test/controllers/provisioning_templates_controller_test.rb
+++ b/test/controllers/provisioning_templates_controller_test.rb
@@ -161,8 +161,8 @@ class ProvisioningTemplatesControllerTest < ActionController::TestCase
     end
 
     test "pxe menu's labels should be sorted by full hostgroup title" do
-      first = FactoryGirl.build(:hostgroup, :with_parent, :name => "def", :operatingsystem => operatingsystems(:centos5_3), :architecture => architectures(:x86_64), :medium => media(:one))
-      second = FactoryGirl.build(:hostgroup, :with_parent, :name => "abc", :operatingsystem => operatingsystems(:centos5_3), :architecture => architectures(:x86_64), :medium => media(:one))
+      first = FactoryGirl.build(:hostgroup, :parent => FactoryGirl.create(:hostgroup, :name => "parent1"), :name => "def", :operatingsystem => operatingsystems(:centos5_3), :architecture => architectures(:x86_64), :medium => media(:one))
+      second = FactoryGirl.build(:hostgroup, :parent => FactoryGirl.create(:hostgroup, :name => "parent2"), :name => "abc", :operatingsystem => operatingsystems(:centos5_3), :architecture => architectures(:x86_64), :medium => media(:one))
       FactoryGirl.create(:template_combination, :provisioning_template => templates(:mystring2), :hostgroup => second)
       FactoryGirl.create(:template_combination, :provisioning_template => templates(:mystring2), :hostgroup => first)
       ProxyAPI::TFTP.any_instance.expects(:create_default).with(regexp_matches(/^PXE.*/), has_entry(:menu, regexp_matches(/#{first.name}.*#{second.name}/m))).returns(true).times(3)


timogoebel:
@sbernhard : Thanks. Could you please rebase this and squash it into a single commit?

app/views/compute_resources_vms/form/vmware/_volume.html.erb
```diff
@@ -24,3 +24,4 @@
            "true",
            "false" %>
 <%= select_f f, :mode, compute_resource.disk_mode_types, :first, :last, { }, :class => "span5", :label => _("Disk mode"), :label_size => "col-md-2", :disabled => !new_host %>
+<%= select_f f, :controller_key, compute_resource.controller_ids, :first, :last, { }, :class => "span5", :label => _("SCSI Controller ID"), :label_size => "col-md-2" , :disabled => !new_host %>
```
timogoebel: This should be `:disabled => !new_vm`. Dito in the line above.

app/models/compute_resources/foreman/model/vmware.rb
```diff
@@ -575,7 +594,6 @@ def vm_instance_defaults
         :memory_mb  => 768,
         :interfaces => [new_interface],
         :volumes    => [new_volume],
-        :scsi_controller => { :type => scsi_controller_default_type },
```
timogoebel: Why remove this?

app/models/compute_resources/foreman/model/vmware.rb
```diff
@@ -362,16 +372,25 @@ def test_connection(options = {})
...
+      # set the default type first and then receive the configured type
+      scsi_controller_type = scsi_controller_default_type
+      if args[:scsi_controller_type].present?
+        scsi_controller_type = args.delete(:scsi_controller_type)
+      end
+
+      scsi_controllers = Hash.new
+      args[:volumes_attributes].each do |k,v|
+        Rails.logger.info("Will add #{v[:controller_key]} vmware controller with type #{scsi_controller_type}")
+        scsi_controllers[v[:controller_key]] = {:type => scsi_controller_type, :key => v[:controller_key]}
```
timogoebel: Does this support API requests where `:controller_key` is not set?

app/models/compute_resources/foreman/model/vmware.rb
```diff
@@ -362,16 +372,25 @@ def test_connection(options = {})
...
       args = args.deep_symbolize_keys
 
+      # set the default type first and then receive the configured type
+      scsi_controller_type = scsi_controller_default_type
+      if args[:scsi_controller_type].present?
+        scsi_controller_type = args.delete(:scsi_controller_type)
+      end
+
+      scsi_controllers = Hash.new
+      args[:volumes_attributes].each do |k,v|
```
timogoebel: This gives me an `NoMethodError: undefined method `each' for nil:NilClass` when trying to create a new vm though the UI.

timogoebel (changes requested):
@sbernhard : Thanks. I left some comments inline. I'm not really convinced, this adds any actual value for customers as long as it's only possible to manage a single scsi controller with foreman. There is a PR open to fix this. #4366

app/views/compute_resources_vms/form/vmware/_volume.html.erb
sbernhard: Some fields are with new_host and some with new_vm. I can change it to new_vm if you are sure that this is the right one.

app/models/compute_resources/foreman/model/vmware.rb
sbernhard: I'm setting the default type in parse_args as its necessary to set the default controller type for all volumes and not only for one.

app/views/compute_resources_vms/form/vmware/_volume.html.erb
timogoebel: I'm positive `new_vm` is the correct choice here.

app/models/compute_resources/foreman/model/vmware.rb
sbernhard: Didn't tried that right now but from the fog-vsphere code I would say, yes.

sbernhard:
The main reason for this change is this (page 11):
http://www.vmware.com/files/pdf/solutions/sap/sap-solutions-on-vmware-best-practices-guide.pdf

[cite]
Use multiple virtual SCSI controllers for database virtual machines. The use of multiple virtual SCSI
controllers allows the execution of several parallel I/O operations inside the guest OS. It is
recommended to use one 
[/cite]

app/views/compute_resources_vms/form/vmware/_volume.html.erb
sbernhard: In the current github develop branch there are 2 lines of code which are using "new_host" instead of "new_vm". It's the vmware name and the disk mode. Should I open a new redmine issue and fix both?

timogoebel: Yes, that would be wonderful. Everything we can get into core will be greatly appreciated. We (read: I) renamed `new_host` to `new_vm` as 637da2f2 allows you to create a host in foreman on an existing vm. So the naming was wrong.

timogoebel:
@sbernhard : Thanks. I am familiar with that document and that recommendation. That's one reason why we started with #4366. That PR allows a user to add more then one scsi controller to a VM and attach disks to the controllers.

sbernhard:
@timogoebel OK. Then we are working on the same issue from different directions I guess. :-)
Adding SCSI controllers to a VM and then using them in the volume section would be better of course. From the working going on in #4366 you are planning to do this in this way. Therefore, I would be OK to stop working on the #4471 and wait for #4366. Do you agree?

timogoebel:
> Therefore, I would be OK to stop working on the #4471 and wait for #4366. Do you agree?

Yes. That would probably be the best approach.

sbernhard:
Can you then close http://projects.theforeman.org/issues/19325 please?

app/views/compute_resources_vms/form/vmware/_volume.html.erb
sbernhard: See Redmine #19340

timogoebel:
@sbernhard : I'd prefer to keep the issue open as #4366 does not allow to actually change the scsi id of a controller.

PR #4472, opened 2017-04-20
fixes #19317, #10927 - move registries to autoload_once path

domcleal:
Partially reverts 70ce8fe to reinstate autoload_once_paths as the
preferred way to load registry-type objects with class variables
containing data set up during app initialisation (e.g. plugin
registrations, menu items.)

autoload_once is compatible with using require_dependency, which is
preferred to allow Rails to track which constants are loaded, so eager
loading doesn't ever try to load the same file twice (7923c95.)

This moves all registry-type objects to app/registries/ and marks this
path as being autoload_once. The contents of this directory should be
kept at an absolute minimum to ease development.


jlsherrill:
This resolves all the issues for me during testing, and makes sense logically.

tbrisker:
Works well for the most part. However I noticed one registry that is missing - Pagelets. Each autoload resets the pagelets.

domcleal:
@tbrisker thanks, updated the PR with app/services/pagelets/ moved into the registries directory.

tbrisker (approved):
Looks good. will merge once tests pass.

tbrisker:
Thanks @domcleal !

PR #4473, opened 2017-04-20
Refs #19309 - Test for 'select all hosts' for 1 page

dLobatog:


test/integration/host_js_test.rb
```diff
@@ -5,7 +5,7 @@
 class HostJSTest < IntegrationTestWithJavascript
   # intermittent failures:
   #   HostJSTest::edit page.test_0003_correctly override global params
-  extend Minitest::OptionalRetry
+  #extend Minitest::OptionalRetry
```
domcleal, line 8: Why's this commented?

test/integration/host_js_test.rb
```diff
@@ -276,11 +276,11 @@ class HostJSTest < IntegrationTestWithJavascript
   describe "hosts index multiple actions" do
     test 'show action buttons' do
       visit hosts_path
-      page.find('#check_all').trigger('click')
+      #page.find('#check_all').trigger('click')
```
domcleal, line 279: ditto, and why is this changed?

test/integration/host_js_test.rb
```diff
@@ -276,11 +276,11 @@ class HostJSTest < IntegrationTestWithJavascript
...
+      page.execute_script("$('#check_all').trigger('click')")
 
       # Ensure and wait for all hosts to be checked, and that no unchecked hosts remain
       assert page.has_no_selector?('input.host_select_boxes:not(:checked)')
-
```
domcleal, line 283: Unrelated?

test/integration/host_js_test.rb
```diff
@@ -302,6 +302,26 @@ class HostJSTest < IntegrationTestWithJavascript
...
+      # Suggestion dialog shows up once we click on 'check all'
+      assert page.has_selector? "#multiple-alert"
+      select_all_hosts = "Select all #{Host.unscoped.count} hosts"
+      assert page.has_link? select_all_hosts
+      # Multiple is false when we have not clicked "select all"
+      assert page.evaluate_script('$("#multiple-alert").data("multiple")')
+      click_link select_all_hosts
+      # Multiple is false after clicking on "select all"
+      refute page.evaluate_script('$("#multiple-alert").data("multiple")')
+      Setting[:entries_per_page] = 20
```
domcleal, line 323: Superfluous, the database should be reset after each test.

test/integration/host_js_test.rb
```diff
@@ -302,6 +302,26 @@ class HostJSTest < IntegrationTestWithJavascript
...
+      FactoryGirl.create(:host)
+      visit hosts_path
+      page.find('#check_all').trigger('click')
+      # Ensure and wait for all hosts to be checked, and that no unchecked hosts remain
+      assert page.has_no_selector?('input.host_select_boxes:not(:checked)')
+      # Suggestion dialog shows up once we click on 'check all'
+      assert page.has_selector? "#multiple-alert"
+      select_all_hosts = "Select all #{Host.unscoped.count} hosts"
+      assert page.has_link? select_all_hosts
+      # Multiple is false when we have not clicked "select all"
```
domcleal, line 318: True or false? The assertion below checks whether it's true, and later it seems to be false.

domcleal (changes requested):
I think the (integration) test stops short of full coverage, it's more of a unit test by only checking internal state and may be invalidated by internal refactoring - which shouldn't be the case for an integration test.

It probably needs to perform an action and ensure all of the hosts are modified.

mmoll:
what's the status here?

PR #4474, opened 2017-04-20
fixes #19328 - autosign tasks have correct wording

timogoebel:


app/models/concerns/orchestration/puppetca.rb
```diff
@@ -55,7 +55,7 @@ def queue_puppetca_create; end
   def queue_puppetca_update
     # Host has been built --> remove auto sign
     if old.build? && !build?
-      queue.create(:name => _("Delete autosign entry for %s") % self, :priority => 50,
+      queue.create(:name => _("Delete PuppetCA autosign entry for %s") % self, :priority => 50,
```
ShimShtein: We should probably extract this to a method and call it from both places. This way we will have less code duplication.

timogoebel:
@ShimShtein : Thanks. Extracted to a method and added a little test coverage.

domcleal (approved):

domcleal:
Merged, thanks @timogoebel.

PR #4476, opened 2017-04-21
fixes #19344 - sort both expected and actual results in UG test

domcleal:
Sorting of the actual value is by name (`Host::Managed#<=>`), but the
expected value was sorted by ID (order of creation). Expected/actual
arguments are swapped to correct order.

### DIFF
--- a/test/models/usergroup_test.rb
+++ b/test/models/usergroup_test.rb
@@ -60,12 +60,12 @@ class UsergroupTest < ActiveSupport::TestCase
     @h5 = FactoryGirl.create(:host, :owner => @u2)
     @h6 = FactoryGirl.create(:host, :owner => @ug3)
 
-    assert_equal @u1.hosts.sort, [@h1]
-    assert_equal @u2.hosts.sort, [@h2, @h5]
-    assert_equal @u3.hosts.sort, [@h2, @h3, @h6]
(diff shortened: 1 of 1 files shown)


test/models/usergroup_test.rb
```diff
@@ -60,12 +60,12 @@ def populate_usergroups
     @h5 = FactoryGirl.create(:host, :owner => @u2)
     @h6 = FactoryGirl.create(:host, :owner => @ug3)
 
-    assert_equal @u1.hosts.sort, [@h1]
-    assert_equal @u2.hosts.sort, [@h2, @h5]
-    assert_equal @u3.hosts.sort, [@h2, @h3, @h6]
-    assert_equal @u4.hosts.sort, [@h6]
-    assert_equal @u5.hosts.sort, [@h2, @h4, @h6]
-    assert_equal @u6.hosts.sort, []
+    assert_equal [@h1], @u1.hosts.sort
```
tbrisker: Nitpick: no need to sort when there should only be 0/1 results.

domcleal: `sort` removed from the lines comparing 0-1 results. PR updated.

tbrisker:
test failures unrelated.

tbrisker:
Thanks @domcleal !

PR #4477, opened 2017-04-21
fixes #19345 - set host group parent names for consistent sorting

domcleal:
Prevents sequentially generated parent names (e.g. hostgroup9, 10) from
being sorted in the opposite order to the test expectation.


dLobatog:
Thanks @domcleal !

PR #4484, opened 2017-04-25
Fixes #19374 - added check to fix idempotency of facet registration

ShimShtein:


sghai:
I tested the fix and migrate_foreman step succeeded and installer is proceeding ahead.

tbrisker:
Thanks @ShimShtein @sghai !

PR #4485, opened 2017-04-25
fixes #19369 - reports: show puppet compile master

timogoebel:
The commit adds a field `producer` to the report model to identify which server (behind a loadbalancer) compiled a puppet catalog.
This is a simple approach that just saves the hostname in a text db field. Other options would be to introduce a new model for that.

To test this, one would need to extend [foreman-report_v2.rb](https://github.com/theforeman/puppet-foreman/blob/master/files/foreman-report_v2.rb#L75) by

```
report['producer'] = %x{hostname -f}.strip
```

This is how this looks:

image


ares:
[test foreman]

app/views/config_reports/show.html.erb
```diff
@@ -15,6 +15,10 @@
 <%= render 'output', :logs => @config_report.logs%>
 <%= render 'metrics', :status => @config_report.status, :metrics => @config_report.metrics["time"] if @config_report.metrics["time"] %>
 
+<% if @config_report.producer.present? %>
+<p class='ra'> <%= _("Report generated by %s ") % @config_report.producer %> </p>
```
ares, line 19: the html should be probably indented

test/static_fixtures/reports/empty.json
```diff
@@ -1,4 +1,4 @@
-// raw data from Foreman's old parsing of reports, with some readable formtting
+// raw data from Foreman's old parsing of reports, with some readable formatting
```
ares, line 1: :+1:

ares (changes requested):
One small nit in the code. Is there a PR for foreman-report opened? Since the producer is likely a hostname of a server where smart proxy runs, would it make sense to try to map it to this object? E.g. the producer name could be a link to the proxy if it's matched.

After this gets merged, report format [wiki](http://projects.theforeman.org/projects/foreman/wiki/Json-report-format) should be updated.

ares:
One more thought, maybe the reported at and generated by could be merged, e.g. "Reported at 2017-04-05 17:33:15 by puppetmaster.example.com" where the "by ..." would be only appended if produced was set.

PR #4487, opened 2017-04-26
Fixes #19384 - Add tooltip to the "Remove Matcher" icon

junezhang:
Add tooltip to the Remove icon.

### DIFF
--- a/app/views/lookup_keys/_value.html.erb
+++ b/app/views/lookup_keys/_value.html.erb
@@ -22,7 +22,7 @@
   <%end%>
   <td>
     <span class="help-block">
-      <%= link_to_remove_fields('', f) %>
+      <%= link_to_remove_fields('', f, title: _('Remove Matcher')) %>
     </span>
   </td>
 </tr>


ShimShtein:
For me it breaks into two lines:

image

To reproduce:
1. Go to Configure -> Classes
2. Select a class
3. Go to Smart Variables tab.

Let's test it in the meantime:

[test]

tbrisker:
ok to test

tbrisker:
@junezhang I think adding a tooltip might be enough in this case? the form is already fairly crowded. (and IIRC there was a text there before that was removed because it caused line breaks, possibly in different languages)

junezhang:
Sorry for replying late as the May holiday.
If the space is limited and Remove is a common usage, we can use the icon alone. But here, the space is not so limited, we can remove the width of every column, then it can release some spaces, and let the content be self-adaption. You can see my Before vs After attachment, there is a big space between the first column and the second column. How do you think?

p.s, another thing is when I adding the tooltip, I found one thing is that:  `link_to_remove_fields` used by "Add Parameters" and "Add combination" has the tooltip "Remove Parameter", but here in "Add Matcher" it doesn't have the tooltip. I don't know why the same code, one has the tooltip, "Add Matcher" doesn't have.

tbrisker:
@junezhang Thanks, looks good to me when removing the set widths. @ShimShtein - does this fix the issue for you with the line wrap? 
As for the tooltip - you can pass another parameter to `link_to_remove_fields` for the tooltip, otherwise it defaults to "remove parameter". If you want to change that you can use something like `<%= link_to_remove_fields(_('Remove'), f, title: _('Remove Matcher') %>`

tbrisker:
Setting to waiting on contributor in case you want to change the tooltips, let me know if not and I will merge (pending @ShimShtein's approval)

junezhang:
@tbrisker Thanks for replying it, I agree with that the tooltip is enough here, so add the tooltip to the icon "Remove Matcher". Please have a review.

tbrisker:
@junezhang thanks, could you please just correct the commit message to reflect the actual change?

junezhang:
@tbrisker  Have fixed the commit message and PR description, but I don't know why the katello testing is failing. Please have a review.

tbrisker:
Katello failures are unrelated. Merging.

tbrisker:
Thanks @junezhang !

PR #4490, opened 2017-04-27
Fixes #19400 - deffer loading of tables list cache

iNecas:
The BookmarkControllerValidator can get loaded sooner than the tables are
loaded (especially in tests).

### DIFF
--- a/app/validators/bookmark_controller_validator.rb
+++ b/app/validators/bookmark_controller_validator.rb
@@ -1,7 +1,17 @@
 class BookmarkControllerValidator < ActiveModel::EachValidator
-  @@active_record_tables = ActiveRecord::Base.connection.tables.map(&:to_s)
   def validate_each(record, attribute, value)
-    controllers = ["dashboard"] + (@@active_record_tables + Permission.resources.map {|x| x.tableize }).uniq
-    record.errors[attribute] << _("%{value} is not a valid controller") % {:value => value } unless controllers.include?(value)
+    unless self.class.valid_controllers_list.include?(value)
+      record.errors[attribute] << _("%{value} is not a valid controller") % {:value => value }
+    end
+  end
+
+  def self.valid_controllers_list
+    @valid_controllers_list ||= (["dashboard"] +
+      ActiveRecord::Base.connection.tables.map(&:to_s) +
+      Permission.resources.map(&:tableize)).uniq
+  end
+
+  def self.reset_controllers_list
+    @valid_controllers_list = nil
   end
 end
--- a/test/models/bookmark_test.rb
+++ b/test/models/bookmark_test.rb
@@ -37,6 +37,7 @@ class BookmarkTest < ActiveSupport::TestCase
     FactoryGirl.create(:permission, :resource_type => 'ProvisioningTemplate', :name => 'manage_provisioning_templates')
     FactoryGirl.create(:permission, :resource_type => 'MyPlugin', :name => 'view_my_plugins')
     Permission.reset_resources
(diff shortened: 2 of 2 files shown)


app/validators/bookmark_controller_validator.rb
```diff
@@ -1,7 +1,10 @@
 class BookmarkControllerValidator < ActiveModel::EachValidator
-  @@active_record_tables = ActiveRecord::Base.connection.tables.map(&:to_s)
   def validate_each(record, attribute, value)
-    controllers = ["dashboard"] + (@@active_record_tables + Permission.resources.map {|x| x.tableize }).uniq
+    controllers = ["dashboard"] + (self.class.active_record_tables + Permission.resources.map {|x| x.tableize }).uniq
```
tbrisker: would it make sense caching the entire controller list instead of generating it again on the fly each time?

iNecas: I didn't want to change the original behavior too much + the chance that the list will change in the future should be minimal, once the runtime is fully loaded (which was the issue originally, when the file could be loaded before the schema). Otherwise we would always need to go to the database for this, which are unnecessary 99.99999% times.

iNecas: Oh, sry, I misread the question: that's a good call, will update

iNecas: Updated

app/validators/bookmark_controller_validator.rb
```diff
@@ -1,7 +1,13 @@
...
-    record.errors[attribute] << _("%{value} is not a valid controller") % {:value => value } unless controllers.include?(value)
+    unless self.class.valid_controllers_list.include?(value)
+      record.errors[attribute] << _("%{value} is not a valid controller") % {:value => value }
+    end
+  end
+
+  def self.valid_controllers_list
+    @valid_controllers_list ||= (["dashboard"] +
+      ActiveRecord::Base.connection.tables.map(&:to_s) +
+      Permission.resources.map {|x| x.tableize }).uniq
```
tbrisker: nitpick - you can use `.map(&:tableize)` here, but not critical.

iNecas: Thanks tomercop! Updated

tbrisker (approved):
Ack, will merge when :green_apple:

iNecas:
The tests didn't like the additional caching: I will update the tests

iNecas:
Small change to update the change when permissions change, mainly for tests

iNecas:
Tests are green

app/models/permission.rb
```diff
@@ -23,6 +23,7 @@ def self.with_translations
   end
 
   def self.reset_resources
+    BookmarkControllerValidator.reset_controllers_list
```
tbrisker: This function is only called once, from the bookmark test. I would prefer resetting the validator in that test rather then here, `Permission` shouldn't really need to know about the internals of this validator.

tbrisker (changes requested):

tbrisker:
also nitpick - deffer should be defer in commit message.

iNecas:
Updated

iNecas:
[test foreman]

iNecas:
After https://github.com/rubiety/validates_lengths_from_database/pull/19, the tests seem to go green again and it seems to help the remote execution CI as well http://ci.theforeman.org/job/test_plugin_pull_request/3666/

tbrisker:
Thanks @iNecas !

PR #4492, opened 2017-04-27
Fixes #19418 - Correctly route cancel button on host form

tbrisker:


tbrisker:
The current way doesn't work for discovery, nor does it make much sense to me in general.

ShimShtein:
IMHO in discovery case, it would be better to redirect to `/discovered_hosts` instead of specific discovered host instance (`/discovered_hosts/:id`).

@lzap, what do you say?

app/views/hosts/_form.html.erb
```diff
@@ -134,6 +134,6 @@
 
     <%= hidden_field_tag 'bare_metal_capabilities', @host.bare_metal_capabilities %>
     <%= f.hidden_field :overwrite? %>
-    <%= submit_or_cancel f, @host.overwrite?, :cancel_path => @host.new_record? || !@host.name_changed? ? @host : edit_host_path(@host.name_was) %>
+    <%= submit_or_cancel f, @host.overwrite?, :cancel_path => (@host.new_record? ? hosts_path : host_path(:id => @host.id)) %>
```
timogoebel: When clicking the cancel button on an existing host, this gives me `http://localhost:3000/hosts/26` instead of `http://localhost:3000/hosts/raul-gherardi.example.com`. Unfortunately the name change handling in the original code is really necessary.
Simply using `host_path(:id => @host.name_was)` should work in any case. I can't see any sense in the `edit_host_path` in the original code.

timogoebel (changes requested):

app/views/hosts/_form.html.erb
tbrisker: Thanks, fixed

dLobatog:
[test upgrade]

dLobatog (approved):
Tried 3 cases:

* Cancel create new hosts -> Goes to /hosts
* Cancel editing an existing host -> Goes to /hosts/fqdn
* Cancel provisioning a discovered host -> Goes to /discovered_hosts/fqdn (I think this is better than /discovered_hosts btw). 

👍 if [test upgrade] passes, thanks @tbrisker / @timogoebel (your comment got fixed I think) / @ShimShtein

dLobatog:
Thanks @tbrisker !

PR #4479, opened 2017-04-23
Fixes #19305 - fixing spacing issue on facts page

lizagilman:
fixing design issues that came up from #4463

PR #4495, opened 2017-04-30
Fixes #19372 - replace children fact values icon

lizagilman:
before: 
plus_icon

after:
zoom_in

@rohoover what do you think?

### DIFF
--- a/app/helpers/fact_values_helper.rb
+++ b/app/helpers/fact_values_helper.rb
@@ -17,7 +17,7 @@ module FactValuesHelper
 
     if value.compose
       url = host_parent_fact_facts_path(:parent_fact => value_name, :host_id => params[:host_id] || value.host.name)
-      link_to(icon_text('plus-sign','', :title => _('Expand nested items')), url) + ' ' + name
+      link_to(icon_text('angle-down', '', :kind => 'fa', :title => _('Expand nested items')), url) + ' ' + content_tag(:span, name)
     else
       name
     end


ohadlevy:
thanks @lizagilman - @rohoover is that the icon you had in mind?

timogoebel:
Just for the record: I think the plus sign is more suitable for an expand action then a zoom-in icon. I associate the zoom-in icon with a zoom-in action, e.g. in an image.
How about `fa-angle-down` in Patternfly's [TreeGrid Table](http://www.patternfly.org/pattern-library/widgets/#treegrid-table) or [Fixed Height Accordion](http://www.patternfly.org/pattern-library/widgets/#fixed-height-accordion)?

lizagilman:
I think TreeGrid Table may be the right thing here

Rohoover:
@timogoebel is correct. Zoom should be used to zoom in, either a chevron icon or the tree grid table /accordion would be better for expansion.

As another note -- the add icon is problematic because we use it to add things, and we are sometimes using it for expansion. I prefer it mean the former with chevrons for the latter.

ohadlevy:
@lizagilman ping?

dLobatog (dismissed):
Setting to WoC as per @Rohoover & @timogoebel comment - most likely no change is needed

lizagilman:
[test foreman]

app/helpers/fact_values_helper.rb
```diff
@@ -29,8 +29,7 @@ def fact_name(value, parent)
 
     if value.compose
       url = host_parent_fact_facts_path(:parent_fact => value_name, :host_id => params[:host_id] || value.host.name)
-      link_to(icon_text('plus-sign','', :title => _('Expand nested items')), url) + ' ' + content_tag(:span, name)
-    else
+      link_to(icon_text('angle-down','', :kind => 'fa',:title => _('Expand nested items')), url) + ' ' + content_tag(:span, name)   else
```
timogoebel: The `else` should be on a separate line.

timogoebel (changes requested):
@lizagilman : I found a small typo. If this is fixed, we can get this in. Thanks.

lizagilman:
@timogoebel @dLobatog  updated, thanks

timogoebel (changes requested):
@lizagilman : Sorry for yet another comment, but there is a line break after the icon. I think, we should get rid of it.
image

lizagilman:
@timogoebel I think this is related to #4479 
Once both of this PRs be merged the page should look ok

timogoebel:
@lizagilman : Yep. You are right. I tested this with the other patch applied, too. Works great.
Let's merge the other patch first and then rebase this on top.

timogoebel (approved):

timogoebel:
@lizagilman : Would you mind rebasing this?

lizagilman:
@timogoebel : rebased

app/helpers/fact_values_helper.rb
```diff
@@ -17,7 +17,7 @@ def fact_name(value, parent)
 
     if value.compose
       url = host_parent_fact_facts_path(:parent_fact => value_name, :host_id => params[:host_id] || value.host.name)
-      link_to(icon_text('plus-sign','', :title => _('Expand nested items')), url) + ' ' + name
+      link_to(icon_text('angle-down','', :kind => 'fa',:title => _('Expand nested items')), url) + ' ' + content_tag(:span, name)
```
timogoebel: Nitpick: Could you please add some spaces between commas?

```
link_to(icon_text('angle-down', '', :kind => 'fa', :title => _('Expand nested items')), url) + ' ' + content_tag(:span, name)
```

timogoebel (changes requested):
@lizagilman: Thanks. Looks good. One more nitpick and this is ready.

lizagilman:
@timogoebel updated

timogoebel (approved):

timogoebel:
Thanks. 👍  if tests pass.

timogoebel:
Merged, thanks @lizagilman. Good job.

PR #4496, opened 2017-04-30
Fixes #19427 - sort by report time is possible

lizagilman:


app/models/fact_value.rb
```diff
@@ -11,6 +11,7 @@ class FactValue < ApplicationRecord
 
   scoped_search :on => :value, :in_key=> :fact_name, :on_key=> :name, :rename => :facts, :complete_value => true, :only_explicit => true, :ext_method => :search_cast_facts
   scoped_search :on => :value, :default_order => true, :ext_method => :search_value_cast_facts
+  scoped_search :on => :updated_at, :rename => :reported_at, :only_explicit => true
```
tbrisker: Please add `:complete_value => true` so that some examples of searches are shown to users.

tbrisker (changes requested):

app/models/fact_value.rb
lizagilman: updated

tbrisker:
Thanks @lizagilman !

PR #4497, opened 2017-05-02
Fixes #19432 - order host bulk actions alphabetically

ares:


tbrisker:
I think it may make sense to sort this in the view, since I believe some plugins override this method to add other functions.

ares:
I'd like to avoid sorting HTML code that `multiple_actions_select` returns but I can move it to `multiple_actions_select` helper, plugins most likely modify [multiple_actions](https://github.com/theforeman/foreman_remote_execution/blob/8a48e4dc078ea4e97712ee2b9b7754c424590c32/app/helpers/concerns/foreman_remote_execution/hosts_helper_extensions.rb#L7) so that should be enough.

tbrisker (approved):

tbrisker:
Test failures are unrelated. I'm going to give this a bit longer before merging in case someone in @theforeman/core thinks this would be a bad idea to sort the list. Also would be happy for input from @rohoover - should the action list be order in alphabetical order or by placing related actions together?

Rohoover:
@tbrisker 

Tricky question. 

The problem with alphabetical is that you typically do not want Delete stuck in the middle... higher chance of miss-click.

tbrisker:
@Rohoover Right now they are approximately ordered by groups of related actions, this is my list of actions with a few plugins enabled:
screenshot from 2017-05-08 17-19-16
The order could likely be improved, but i'm not sure if alphabetical is the way to go here.

Rohoover:
@tbrisker 

Holy crap. That list is huge.

What are your thoughts about putting them in groups by features? So, "Build Hosts", "Disassociate Hosts" and "Delete Hosts" together.

Or, place them in order of popularity.

ares:
Thanks @Rohoover, grouping could help a bit. I can imagine the following

* Generic
  * Change group
  * Edit parameters
  * Disassociate host
  * Assign organization
  * Assign location
  * Delete hosts
  * Change owner
* Provisioning
  * Build hosts
  * Rebuild config
  * Change power state
* Configuration
  * Change environment
  * Disable notifications
  * Enable notifications
  * Change puppet master
  * Change puppet CA
  * Play ansible roles (could be it's own group)

the question remains though, should items be sorted alphabetically? Both withing the group and the group name. I suppose plugins (e.g. ansible in this case) would be able to add their own groups. Which means we should offer this through plugin DSL. Since that would be a bigger change anyway, is there some other recommendation to deal with such a big list? All of these are bulk actions, meaning you can do $this with bulk of hosts.

Rohoover:
@ares 
I wouldn't want to list them alphabetically and then lose the sub groupings. For example: I wouldn't want anything to come between Disable or Enable Notifications.

tbrisker:
Looks like there isn't an agreement yet as to how to order it, appears that alphabetically isn't better then the current order. marking as reached an impasse until some agreement is reached.

ares:
I think there is an agreement, we should have sub-groups and keep related items together, not sorted alphabetically. I think waiting on contributor would be better representation of the status. I plan to look at that at some point.

xprazak2:
@ares, are there any plans to continue with this?

sbernhard:
Having groups would be nice but this PR ist now about a year old. Sorting the items alphabetically is in any case better than without sorting. Without sorting and using many different plugins the unsorted list is just a mess. And to be honest, the unsorted list with ordered related actions, wasn't recognized by many of our users.

Sort it alphabetically which is one step better than the current state.

timogoebel:
I agree with @sbernhard here. I think alphabetical sorting is way better than random sorting of this list.
Can we merge this and add a follow-up ticket to group this?

sbernhard:
@tbrisker / @ares, how can we proceed with this PR?

ares:
I'd say let's a discussion on discourse explaining the change and then attach a poll to see what people prefer. I personally would vote :-1: now, even though this was my PR. It would be interesting to use menu helpers we have for main menu so plugins could extend this selector in explicit order. Having related options together seems to me now more important than alphabet order.

timogoebel:
> It would be interesting to use menu helpers we have for main menu so plugins could extend this selector in explicit order. Having related options together seems to me now more important than alphabet order.

I concur. I believe we don't need to take a vote here. It should be fairly easy to implement a plugin interface.

hlawatschek:
is there a follow up issue/ticket or discussion on that?

ezr-ondrej:
@hlawatschek please follow up https://community.theforeman.org/t/order-host-bulk-actions-alphabetically/11902/2, kinda related discussion: https://community.theforeman.org/t/ux-improvement-bulk-delete-things-maybe-general-bulk-actions/13702

PR #4498, opened 2017-05-03
Fixes #19436 - Better controller selection on host dropdown

tbrisker:
Always setting to `host` unless the controller is `hostgroup` broke
discovery's usage of the form as it uses the `discoverd_host`
controller.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -209,7 +209,8 @@ module ApplicationHelper
   end
 
   def method_path(method)
-    send("#{method}_#{controller_name}_path")
+    controller = controller_name.start_with?('compute') ? 'hosts' : controller_name
+    send("#{method}_#{controller}_path")
   end
 
   def edit_textfield(object, property, options = {})
--- a/app/helpers/hosts_helper.rb
+++ b/app/helpers/hosts_helper.rb
@@ -482,9 +482,4 @@ module HostsHelper
   def power_status_visible?
     SETTINGS[:unattended] && Setting[:host_power_status]
   end
-
-  def host_or_hostgroup_path(method)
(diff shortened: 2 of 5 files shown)


timogoebel:
@tbrisker : Somehow I don't like the approach either. I'm curious if we can find a way that generates the correct url based on the same criteria for
a) the form action
b) the ajax data

tbrisker:
Since some of the dropdowns were using `method_path` while others had `host_or_hostgroup_path`, I reverted the change to `host_or_hostgroup_path` from 637da2f and moved the controller correction to the `method_path` method. Another option would be to override this method only in the compute_resource_vms controller.
While I agree this is a bit of a hack and would be happy to clean up the host form, this bug is a regression introduced in 1.15 which breaks discovery.

timogoebel:
@tbrisker: 👍  for 1.15 and 👍 for agreeing, that this isn't great and needs more work for 1.16.

dLobatog:
Thanks @tbrisker , setting back to WoC as it's failing tests. [test katello] (I'm not so sure katello failures happened after this)

dLobatog (changes requested):
Looks like architecture_selected_hosts_path isn't working

lzap:
@dLobatog does 1.15 mean RCx or 1.15.1?

tbrisker:
@dLobatog - thanks, I had a typo. fixed now.
@lzap - I believe this is a 1.15.0 blocker.

dLobatog (approved):

dLobatog:
[test upgrade]  - @tbrisker MySQL upgrade job failed, I will retry and merge if green

dLobatog:
Thanks @tbrisker !

PR #4499, opened 2017-05-03
fixes #19437 - remove duplicate Rails.cache.clear calls

domcleal:
Rely on the default cache clear in the AS test helper, which catches
errors from Rails when the file cache doesn't exist yet.


ohadlevy:
[test katello]

tbrisker:
[test katello]

timogoebel:
@tbrisker : Katello tests are broken regardless of this patch. This shouln't keep us from merging this imho.

tbrisker:
Thanks @domcleal !

PR #4500, opened 2017-05-03
Fixes #19475 - Rename test variables to more specific term

bastilian:
An issue occurring with a recent minitest upgrade broke tests.
Renaming variables set with let to a less ambiguous wording is
to work around the issue.

**Note:** CI might be green, but test might not have been run in a way that they are failing. For the tests to fail the foreman-docker plugin needs to be installed to cause a failure (see http://ci.theforeman.org/job/test_plugin_matrix/database=postgresql,ruby=2.0.0,slave=fast/3027/console). To verify the patch best is to run tests locally in develop with the plugin included, which should fail, if the minitest is at the latest. The same tests should pass in this branch with the plugin included as well.

### DIFF
--- a/test/models/image_test.rb
+++ b/test/models/image_test.rb
@@ -17,40 +17,40 @@ class ImageTest < ActiveSupport::TestCase
   end
 
   context "audits for password change" do
-    let(:image) { FactoryGirl.build(:image, :compute_resource => FactoryGirl.create(:compute_resource, :libvirt)) }
+    let(:protected_image) { FactoryGirl.build(:image, :compute_resource => FactoryGirl.create(:compute_resource, :libvirt)) }
 
     test "audit of password change should be saved only once, second time audited changes should not contain password_changed" do
       as_admin do
-        image.password = "newpassword"
-        assert_valid image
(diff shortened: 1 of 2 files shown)


bastilian:
ok to test

bastilian:
[test]

bastilian:
The test failure is unrelated and as mentioned the patch should also be verified manually.

ares:
I did some reasearch and came to the same conclusions. docker-api monkeypatches [Rake::DSL](https://github.com/swipely/docker-api/commit/076cbf5d5b42b5e31f14b62a222be95fabb40983) because of their [DSL](https://github.com/swipely/docker-api/blob/b8c3969795677886afab4a5797dc43eb66e55d15/README.md#rake-task) features. The fix should probably land either in minitest or docker-api but it's not clear whether minitest intentionally includes Rake::DSL. Since it will take time before it gets figured out, I'll merge this. It's harmless since it only renames local test methods. Let's hope we won't have many PRs where devs would reintroduce this name as it would break foreman-docker again.

Thanks @bastilian, merged as 0099710808f3820f7608bb3158d6090579d97bf7

PR #4501, opened 2017-05-03
Require fog subprojects instead of fog

icco:
In response to https://github.com/fog/fog-google/issues/158, update your fog dependency to include modern versions of fog.

### DIFF
--- a/bundler.d/fog.rb
+++ b/bundler.d/fog.rb
@@ -1,3 +1,4 @@
 group :fog do
-  gem 'fog', '1.38.0', :require => false
+  gem 'fog-aws', '1.3.0', :require => false
+  gem 'fog-google', '0.5.2', :require => false
 end


bundler.d/fog.rb
```diff
@@ -1,3 +1,4 @@
 group :fog do
-  gem 'fog', '1.38.0', :require => false
+  gem 'fog-aws', '1.3.0', :require => false
```
domcleal, line 2: This is already in `bundler.d/ec2.rb`.

bundler.d/fog.rb
```diff
@@ -1,3 +1,4 @@
...
+  gem 'fog-google', '0.5.2', :require => false
```
domcleal, line 3: Already in `gce.rb`, would be changed there.

domcleal (changes requested):
The `fog` dependency is required for oVirt, Docker, and possibly OpenNebula support, which aren't available in a separate fog-* gem, so I don't think this dependency can simply be removed currently.

mmoll:
I guess the first step for now would be to only update the fog gem(s) used in Foreman core. Regarding the single providers, I would hope that some @theforeman and @manageiq people could maintain a split-out fog-ovirt, while somebody from @netways could take over a split-out fog-one...

icco:
@mmoll The discussion on that work is at https://github.com/fog/fog/issues/3942 if you'd like to speed that along. Until that work is done, I'll close this PR.

PR #4503, opened 2017-05-04
Add rdoc to dependencies

kotso:
CentOS7 minimal installer fails without it


domcleal:
Per the above comment, this requires an issue number referenced in the commit message. What issue is it fixing precisely, is it the same as http://projects.theforeman.org/issues/15628? If so, you can use that existing ticket number and amend the commit message (`git commit --amend`, `git push -f github`).

mmoll:
@kotso any update here?

mmoll:
closing due to inactivity

PR #4504, opened 2017-05-04
fixes #19463 - disable auditing of factory_girl fixtures

domcleal:
Audit entries are mostly superfluous in the test environment, but can
be enabled by adding the :with_auditing trait. Disabling audits
improves the test:models runtime from 160 to 128 seconds for me.

### DIFF
--- a/test/factories/disable_auditing.rb
+++ b/test/factories/disable_auditing.rb
@@ -1,0 +1,18 @@
+FactoryGirl.define do
+  # Default to disabling auditing when saving new models to reduce database
+  # work and speed up tests. Add the :with_auditing global trait to enable it.
+  trait :with_auditing do
+    before(:create) do |instance, evaluator|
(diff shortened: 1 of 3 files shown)


dLobatog:
Looks good to me, I'm not quite sure why these katello tests are failing but it does not seem caused by this at first glance. @bbuckingham @jlsherrill @thomasmckay @parthaa have you seen any recent test errors like this? I saw https://github.com/theforeman/foreman/pull/4498 had similar errors too

dLobatog (approved):
Will merge after we clarify what's wrong with the Katello tests, thanks @domcleal

jlsherrill:
[test katello]

jlsherrill:
The Katello tests seem to be a transient failure, but one i've only seen about 3 times so far.   I've filed an issue here to track that: http://projects.theforeman.org/issues/19474

tbrisker:
Thanks @domcleal !

PR #4505, opened 2017-05-05
Fixes #17771 - Test if rsync successfully uploaded tarball

gnurag:
This PR checks if rsync executed successfully and only then it prints a success message, along with http url for the uploaded foreman-debug archive.

Upon failure, I've also added a suggestion to upload to an alternate location. Let me know if this suggestion is reasonable, or else I'll remove it.

### DIFF
--- a/script/foreman-debug
+++ b/script/foreman-debug
@@ -377,10 +377,15 @@ fi
 # upload if -u was passed in
 if [ $UPLOAD_DISABLED -eq 0 -a $UPLOAD -eq 1 ]; then
   qprintf "Uploading...\n"
-  rsync $TARBALL rsync://theforeman.org/debug-incoming
-  qprintf "The tarball has been uploaded, please contact us on our mailing list or IRC\n"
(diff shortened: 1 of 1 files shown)


domcleal (changes requested):
The exit code appears to be `0` on error, please change this so it's non-zero when the upload fails.

gnurag:
Thanks, what would be an appropriate exit code? 1, 2 seem reserved. Perhaps 69?

`#define EX_USAGE        64      /* command line usage error */`

domcleal:
I don't see how you get from 1, 2 to 69. The next logical choice would probably be 3. Either way, please make a choice (with justification if necessary) and update the PR for review.

timogoebel:
@domcleal : PR has been updated.

gnurag:
@domcleal Added `exit 3`; please re-review.

domcleal (approved):

domcleal:
@theforeman-bot ok to test

domcleal:
Merged, thanks @gnurag.

PR #4506, opened 2017-05-05
fixes #19464 - rewrite Dashboard::Manager to support autoloading

domcleal:
Replaces class-level widget storage with a static list of built-in
widgets, then appends the registered plugin widgets to get the full
default list of widgets. As a result, the class can safely be reloaded
when modified and is more testable as class-level state is removed.


ohadlevy:
[test foreman]

tbrisker:
Works well. Although this removes the ability to add a permitted template without adding a widget that uses it (as the original idea of that was to allow some flexible user defined widgets with a set template), I'm ok with that as it isn't used anyways, if needed in the future we'll figure it out.

tbrisker:
Thanks @domcleal !

PR #4507, opened 2017-05-05
refs #19317 - move FiltersHelperOverrides to app/registries/

domcleal:
Uses class to hold state, should be kept in the autoload_once registry
directory.


tbrisker:
Thanks @domcleal !

PR #4508, opened 2017-05-05
Fixes #19173 - build notification succeeds with expired token.

dmitri-d:


dmitri-d:
Test failure is a js test timing out on a connection to /organizations.

timogoebel:
@witlessbird : Thanks. Is this really a good idea? If a token is expired, it should be expired, invalid and useless. Is there any real-world scenario, where this is necessary if everything else is set up correctly?

dmitri-d:
@timogoebel: This is an attempt to fix an issue someone experienced irl (token expired after provisioning started, leading to it being forever stuck in build mode). I couldn't think of a better way to fix this; fwiw the change only allows the host to call "/build" and nothing else.

app/controllers/unattended_controller.rb
```diff
@@ -81,25 +83,45 @@ def render_template(type)
...
+    @host = find_host_by_spoof || find_built_host_by_token || find_host_by_ip_or_mac
+    verify_found_host(@host)
+  end
+
+  def verify_found_host(a_host)
+    logger.debug "Found #{@host}" unless host_not_found?(@host) || host_os_is_missing?(@host) || host_os_family_is_missing?(@host)
+  end
+
+  def host_not_found?(a_host)
+    unless a_host
```
timogoebel: How about extracting this to something like

```ruby
def host_value_present?(a_host, value, N_("%{controller}: unable to find a host that matches the request from %{addr}"))
  unless value
     [...]
  end
end

**Comment by timogoebel** [comment]
@witlessbird : Thanks. Fair enough. I did experience something like this myself. However usually when something else was wrong. The host then stays in an install loop. Fun experience. From a user's perspective, this is probably a valid RFE.

**Comment by dmitri-d** [comment]
- extracted value presence check into a separate method.

File: app/controllers/unattended_controller.rb [hunk 2]
```diff
@@ -110,13 +126,25 @@ def find_host_by_spoof
...
+  def token_from_params
     token = params.delete('token')
     return nil if token.blank?
     # Quirk: ZTP requires the .slax suffix
     if (result = token.match(/^([a-z0-9-]+)(.slax)$/i))
-      token, _suffix = result.captures
+      return result[1]
     end
-    Host.for_token(token).first
+    return token
```
timogoebel: superfluous return

timogoebel (changes requested):
@witlessbird : Thanks. I tested this manually. Works great. There is just one nitpick that needs fixing before merge.

dmitri-d:
* removed a redundant "return" keyword in UnattendedController#token_from_params method.

timogoebel (approved):
Test failure is unrelated (JS test).

timogoebel:
Merged, thanks @witlessbird.

PR #4510, opened 2017-05-08
Fixes #18995, #18996 - User visible_environments works w/o taxonomies

dLobatog:


dLobatog:
[test upgrade] - timed out on MySQL.

test/models/user_test.rb
```diff
@@ -990,4 +990,37 @@ def editing_self_helper
...
+      all_but_production = Environment.unscoped.select { |e| e.name != 'production' }
+      assert_equal all_but_production.map(&:name), User.current.visible_environments
+    end
+
+    test 'should show the list of environments visible when only orgs are enabled' do
+      Taxonomy.expects(:locations_enabled).returns(false).at_least_once
+      setup_user 'view', 'environments', 'name = test_env_org1'
+      FactoryGirl.create(:environment, :name => 'test_env_org1',
+                         :organizations => [User.current.organizations.first],
+                         :locations => User.current.locations.all)
```
tbrisker: shouldn't this be `nil` if locations are disabled?

test/models/user_test.rb
```diff
@@ -990,4 +990,37 @@ def editing_self_helper
...
+      assert_equal ['test_env_org1'], User.current.visible_environments
+    end
+
+    test 'should show the list of environments visible when only locs are enabled' do
+      # No need to expect 'organizations_enabled' here - as long as locations_enabled
+      # is true, #visible_environments will check by default on taxable_taxonomies.
+      setup_user 'view', 'environments', 'name = test_env_loc1'
+      FactoryGirl.create(:environment, :name => 'test_env_loc1',
+                         :locations => [User.current.locations.first],
+                         :organizations => User.current.organizations.all)
```
tbrisker: ditto?

test/models/user_test.rb
```diff
@@ -990,4 +990,37 @@ def editing_self_helper
...
+      setup_user 'view', 'environments', 'name = test_env_org1'
+      FactoryGirl.create(:environment, :name => 'test_env_org1',
+                         :organizations => [User.current.organizations.first],
+                         :locations => User.current.locations.all)
+      assert_equal ['test_env_org1'], User.current.visible_environments
+    end
+
+    test 'should show the list of environments visible when only locs are enabled' do
+      # No need to expect 'organizations_enabled' here - as long as locations_enabled
+      # is true, #visible_environments will check by default on taxable_taxonomies.
```
tbrisker: Then why is it needed when only org are enabled?

dLobatog: https://github.com/theforeman/foreman/pull/4510/files#diff-4676c008b11a5480d73d4a6de01e45b9R404 - it's just that the if condition to add in 'taxonomy checks' is `locations_enabled || organizations_enabled`. When 'locations_enabled' is true, like in this case, 'organizations_enabled' will never get called.

test/models/user_test.rb
dLobatog: Yes, but if it's nil then `Environment.unscoped.authorized(:view_environments)` will fail to show them to this user. The point here was to test the `if` condition in the user.rb `visible_environments` change. 

I'll try to flip `SETTINGS[:locations_enabled]` see if I can make it truly simulate what we want to test here.

dLobatog:
@tbrisker Updated to use `SETTINGS[:taxonomy_enabled]` instead, this seems to actually reproduce the condition much better at different levels (including taxonomies authorization layer), instead of simulating it with Mocha expectations.

test/models/user_test.rb
```diff
@@ -990,4 +990,42 @@ def editing_self_helper
...
+      SETTINGS[:locations_enabled] = true
+      SETTINGS[:organizations_enabled] = false
+      setup_user 'view', 'environments', 'name = test_env_loc1'
+      FactoryGirl.create(:environment, :name => 'test_env_loc1',
+                         :locations => [User.current.locations.first])
+      assert_equal ['test_env_loc1'], User.current.visible_environments
+    end
+
+    test 'should show the list of environments visible when both orgs/locs are enabled' do
+      assert_equal ['production'], User.current.visible_environments
```
tbrisker: shouldn't this be all environments?

test/models/user_test.rb
```diff
@@ -990,4 +990,42 @@ def editing_self_helper
...
+    teardown do
+      SETTINGS[:locations_enabled] = true
+      SETTINGS[:organizations_enabled] = true
+    end
+
+    test 'should check for environments permission with orgs/locs disabled' do
+      SETTINGS[:locations_enabled] = false
+      SETTINGS[:organizations_enabled] = false
+      setup_user 'view', 'environments', 'name != production'
+      all_but_production = Environment.unscoped.select { |e| e.name != 'production' }
```
tbrisker, line 1004: Might want to make sure this isn't an empty list

test/models/user_test.rb
dLobatog: There's a bug in the method (already existing prior to this patch, but I'll fix it). Locations and Organizations that have NO orgs or locs will never show up as visible environments, not even for the admin user. The reason is that there's a join/where on taxable_taxonomies that will not allow these to show up

dLobatog:
@tbrisker Thanks for the quick review, you spotted another bug with the previous implementation on https://github.com/theforeman/foreman/pull/4510#discussion_r115224627 - I have fixed it by including all environments without location/organization as visible by the admin user

app/models/user.rb
```diff
@@ -400,12 +400,20 @@ def taxonomy_ids
...
+    if Taxonomy.locations_enabled || Taxonomy.organizations_enabled
+      authorized_scope = authorized_scope.
+        joins(:taxable_taxonomies)
+        .where('taxable_taxonomies.taxonomy_id' => taxonomy_ids[:organizations] + taxonomy_ids[:locations])
+    end
+    result = authorized_scope.distinct.pluck(:name)
+    if User.current.admin
+      # Admin users can also see Environments that do not have any organization or location, even when
+      # organizations and locations are enabled.
+      untaxed_environments = Environment.unscoped.select { |e| e.locations == [] || e.organizations == [] }.map(&:name)
```
tbrisker: This would load all environments to memory, I wonder if it's possible to achieve the same result in another way? if we load everything to memory might as well do all the filtering in memory

dLobatog:
@tbrisker I've just updated this using plain SQL (a subquery to find the environments that have no taxonomies - i.e: only visible to admin) for the filtering. Let me know if that's what you were looking for, thanks again for the review

tbrisker (approved):

tbrisker:
thanks @dLobatog !

PR #4512, opened 2017-05-08
Fixes #10273 - Clone disk from template when preallocate marked

orrabin:
This won't work without a fix in fog.
The `attributes` here: https://github.com/fog/fog-core/blob/master/lib/fog/core/attributes.rb#L69 is not a `HashWithIndifferentAccess`.
As a result the options to use when building the vm: https://github.com/abenari/rbovirt/blob/master/lib/client/vm_api.rb#L58 will contain the keys `clone` and `disks` as strings which will prevent them from being part of the xml that is sent to ovirt: https://github.com/abenari/rbovirt/blob/master/lib/ovirt/vm.rb#L111-L117

### DIFF
--- a/app/models/compute_resources/foreman/model/ovirt.rb
+++ b/app/models/compute_resources/foreman/model/ovirt.rb
@@ -236,7 +236,11 @@ module Foreman::Model
       args[:template] = args[:image_id] if args[:image_id]
 
       sanitize_inherited_vm_attributes(args)
+
+      preallocate_disks(args) if args[:volumes_attributes].present?
+
       vm = super({ :first_boot_dev => 'network', :quota => ovirt_quota }.merge(args))
+
       begin
         create_interfaces(vm, args[:interfaces_attributes])
         create_volumes(vm, args[:volumes_attributes])
@@ -245,6 +249,16 @@ module Foreman::Model
         raise e
       end
       vm
+    end
+
+    def preallocate_disks(args)
+      change_allocation_volumes = args[:volumes_attributes].values.select{ |x| x[:preallocate] == '1' }
+      if args[:template].present? && change_allocation_volumes.present?
+        disks = change_allocation_volumes.map do |volume|
+          { :id => volume[:id], :sparse => 'false', :format => 'raw', :storagedomain => volume[:storage_domain] }
+        end
+        args.merge!(:clone => true, :disks => disks)
+      end
     end
 
     def new_vm(attr = {})
--- a/bundler.d/ovirt.rb
+++ b/bundler.d/ovirt.rb
@@ -1,3 +1,3 @@
 group :ovirt do
-  gem 'fog-ovirt', '~> 0.1.2'
+  gem 'fog-ovirt', '~> 0.1.3'
 end


iNecas:
I'm not sure usage of HashWithIndifferentAccess will be accepted in fog-core, as it has no dependency on ActiveSupport now (probably on purpose).

Where is the place when the symbols that we use as keys become strings?

orrabin:
For the record, the other way to avoid changing for-core as @iNecas suggested is adding `disks` and `clone` to the server attributes: https://github.com/fog/fog/blob/master/lib/fog/ovirt/models/compute/server.rb#L11
which would keep them as symbols while merging attributes and result in the correct xml in rbovirt.

orrabin:
The fog pr has been merged so this can now be tested.

chris1984 (commented):
@orrabin would a katello/foreman nightly be good for testing this pr?

orrabin:
@chris1984 yes nightly is good. This needs to be tested with a compute resource for ovirt 3.6.

app/models/compute_resources/foreman/model/ovirt.rb
```diff
@@ -210,10 +210,23 @@ def start_with_cloudinit(uuid, user_data = nil)
...
       end
+
+      change_allocation_volumes = args[:volumes_attributes].values.select{ |x| x[:preallocate] == '1' }
+      if args[:template].present? && change_allocation_volumes.present?
+        disks = []
+        change_allocation_volumes.each do |volume|
+          disks << {:id => volume[:id], :sparse => false, :format => 'raw' }
+        end
+
+        args.merge!(:clone => true, :disks => disks) unless disks.empty?
```
adamruzicka: Wouldn't it be cleaner if rewritten like this?
```ruby
if args[:template].present? && change_allocation_volumes.present?
  disks = change_allocation_volumes.map do |volume|
    { :id => volume[:id], :sparse => false, :format => 'raw' }
  end
  args.merge!(:clone => true, :disks => disks)
end
```

adamruzicka (approved):
Tested and it works flawlessly 👍 

Left one comment about possibility for refactoring.

orrabin:
@adamruzicka thanks, I fixed your comment.
Edit: we need to find a new ovirt instance to retest, the one I was using seems to be broken now.

iNecas:
Maybe time to revive this?

raulbla:
it does not work with Ovirt 4.1. The api seems to accept clone=true only as a query string parameter to the post request creating the VM.

orrabin:
@raulbla it doesn't work for nfs in 4.1, this is due to a change in ovirt (try it directly in ovirt - create a template and try to create a vm from it and you'll see it will stay on thin provisioning).
This can only be tested with block storage when testing with ovirt 4.1.

raulbla:
@orrabin our environment is using block storage. however,  when we create a new vm from template using the gui and select the clone option it works correctly.

pmoravec:
@orabin it fails also for iscsi / block-based data centers. clone=true argument didnt appear in the API call (patching both this and fog PR 3958).

app/models/compute_resources/foreman/model/ovirt.rb
```diff
@@ -225,6 +230,16 @@ def create_vm(args = {})
       vm
     end
 
+    def preallocate_disks(args)
+      change_allocation_volumes = args[:volumes_attributes].values.select{ |x| x[:preallocate] == '1' }
```
pmoravec, line 255: If I am creating a VM without disks, `args[:volumes_attributes]` is empty (until I something overlooked) and this would raise `NoMethodError: undefined method `values' for nil:NilClass`, am I right?

orrabin: I added this if: https://github.com/theforeman/foreman/pull/4512/files#diff-20d0fb6f34ad993f1199e464ebe97df5R219 so this won't happen.

orrabin:
Tested against 4.1 and to get it to work there I needed changes in rbovirt. I added: https://github.com/abenari/rbovirt/pull/122, once it's merged this can be tested again.

orrabin:
Rebased, this still needs a new version of rbovirt to be tested.

iNecas:
New version of rbovirt released. I guess it would be safest to also specify the version requirement on rbovirt here.

shiramax:
Tested with Red Hat Virtualization Manager Version: 4.1.7.5-0.1.el7
and it worked for me.

iNecas (changes requested):
Needs new fog-ovirt version and bump the dependency on it here. Will merge afterwards.

orrabin:
@iNecas I updated the dependency and rebased.

iNecas (approved):

iNecas:
Thanks @orrabin

PR #4513, opened 2017-05-09
fixes #19479, #10587 notifications now works with two-pane again

ohadlevy:
- notification is now inside the content div, allowing two-pane server
responses to include it.
- refactored all notifications to use notification helpers (notice,
warning and error) instead of direct flash manipulation
- also added the usage of flash now via the helpers, and fixed
it in a couple of places it was used incorrectly (mostly clone rendering
which does not redirect at all).
- fixed consisteny to remove trailing dots from toast messages (as it
seems most strings are without dot suffix).

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -49,7 +49,7 @@ class UsersController < ApplicationController
   def destroy
     @user = find_resource(:destroy_users)
     if @user == User.current
-      notice _("You cannot delete this user while logged in as this user.")
+      warning _("You cannot delete this user while logged in as this user")
       redirect_to :back
       return
     end
--- a/webpack/assets/javascripts/foreman_toast_notifications.js
+++ b/webpack/assets/javascripts/foreman_toast_notifications.js
@@ -13,15 +13,27 @@ export function notify({message, type, link, sticky = isSticky(type)}) {
       }));
 }
 
+export function clear() {
+  store.dispatch(ToastActions.clearToasts());
+}
+
 function importFlashMessagesFromRails() {
-  $('#notifications').data().flash.forEach(([type, message]) => {
+  const notifications = $('#notifications');
+
+  if (notifications.length === 0 ||
+    !notifications.data('flash')) {return;}
+
+  notifications.data('flash').forEach(([type, message]) => {
     notify({message, type});
   });
+  // remove both jquery cache and dom entry to avoid ajax ContentLoad events
+  // reloading our notifications
+  notifications.attr('data-flash', '').removeData();
 }
 
 // load notifications from Rails on ContentLoad
 // to accommodate rails flash syntax
 $(document).on('ContentLoad', () => {
-  store.dispatch(ToastActions.clearToasts());
+  clear();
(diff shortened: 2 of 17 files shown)


tbrisker (approved):
Looks good, lets see that tests pass.

tbrisker:
@ohadlevy looks like one test is failing, could you take a look?

ohadlevy:
@tbrisker fixed failing test, I actually changed a notice alert to warning, hence the failure. thanks

app/controllers/users_controller.rb
```diff
@@ -49,7 +49,7 @@ def update
   def destroy
     @user = find_resource(:destroy_users)
     if @user == User.current
-      notice _("You cannot delete this user while logged in as this user.")
+      warning _("You cannot delete this user while logged in as this user")
```
ohadlevy, line 52: this was changed from notice to warning level.

webpack/assets/javascripts/foreman_toast_notifications.js
```diff
@@ -13,15 +13,26 @@ export function notify({message, type, link, sticky = isSticky(type)}) {
...
+  store.dispatch(ToastActions.clearToasts());
+}
+
 function importFlashMessagesFromRails() {
-  $('#notifications').data().flash.forEach(([type, message]) => {
+  const notifications = $('#notifications');
+
+  if (notifications.length === 0) {return;}
+
+  notifications.data().flash.forEach(([type, message]) => {
```
tbrisker: This seems to fail when there is no flash data, which occurs in some of the tests.

tbrisker (changes requested):
Integration tests are still failing.

johnpmitsch:
This fixes http://projects.theforeman.org/issues/19500 for me

ohadlevy:
@tbrisker updated.

webpack/assets/javascripts/foreman_toast_notifications.js
```diff
@@ -13,15 +13,28 @@ export function notify({message, type, link, sticky = isSticky(type)}) {
...
+  store.dispatch(ToastActions.clearToasts());
+}
+
 function importFlashMessagesFromRails() {
-  $('#notifications').data().flash.forEach(([type, message]) => {
+  const notifications = $('#notifications');
+
+  if (notifications.length === 0 ||
+    !notifications.data() ||
+    !notifications.data().flash) {return;}
```
tbrisker: you can safely combine these two checks to `!notifications.data('flash')` (also works for looping below)

ohadlevy:
@tbrisker test failure unrelated.

ohadlevy:
If I read the error on http://projects.theforeman.org/issues/19500#change-86746 correctly, it seems that .data was undefined?

ohadlevy:
Oh I see, tomorrow than :)

ohadlevy:
@tbrisker updated.

ohadlevy:
tests failures seem unrelated?

tbrisker (approved):

tbrisker:
Thanks @ohadlevy !

PR #4515, opened 2017-05-09
Fixes #17749 - Remove confusing .gz.txt extension from logs

gnurag:
Currently .txt extension is appended to log files that are processed
by foreman-debug and added to the debug archive. This is confusing to
the user.

This PR gets rid of the .txt extension, and also removes .gz/.bz2/xz
extension from log files that have been uncompressed.

### DIFF
--- a/script/foreman-debug
+++ b/script/foreman-debug
@@ -86,21 +86,25 @@ add_file() {
   MIME=$(file -bi "$FILE" | cut -d\; -f1)
   case $MIME in
         application/x-gzip)
-          zcat "$FILE" | sed -r "$FILTER" > "$DIR$FILE.txt"
-	  touch -c -r "$FILE" "$DIR$FILE.txt"
+          OUTFILE="$DIR${FILE/%.gz/}"
+          zcat "$FILE" | sed -r "$FILTER" > "$OUTFILE"
+          touch -c -r "$FILE" "$OUTFILE"
           ;;
         application/x-bzip2)
-          bzcat "$FILE" | sed -r "$FILTER" > "$DIR$FILE.txt"
-	  touch -c -r "$FILE" "$DIR$FILE.txt"
+          OUTFILE="$DIR${FILE/%.bz2/}"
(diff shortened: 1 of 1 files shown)


script/foreman-debug
```diff
@@ -86,21 +86,24 @@ add_file() {
   MIME=$(file -bi "$FILE" | cut -d\; -f1)
   case $MIME in
         application/x-gzip)
-          zcat "$FILE" | sed -r "$FILTER" > "$DIR$FILE.txt"
-	  touch -c -r "$FILE" "$DIR$FILE.txt"
+          OUTFILE=$(echo "$FILE" | sed s/.gz$//)
```
lzap: Bash allows you to replace strings in variables, it's built in feature. Why not to use that.

http://stackoverflow.com/questions/13210880/replace-one-substring-for-another-string-in-shell-script

script/foreman-debug
```diff
@@ -86,21 +86,24 @@ add_file() {
...
+          zcat "$FILE" | sed -r "$FILTER" > "$DIR$OUTFILE"
+          touch -c -r "$FILE" "$DIR$OUTFILE"
```
lzap: I think you can merge $DIR and $OUTFILE into single variable, can you?

lzap (changes requested):
Nice thanks for taking care of these small annoying things. I'd suggest to use shell builtin.

script/foreman-debug
gnurag: @lzap Thanks, tried % flag to match .gz at the end of the line and it worked great. Updating the PR.

script/foreman-debug
gnurag: @lzap Tried to merge `$DIR$FILE` into a single var, please see if it looks okay.

gnurag:
@lzap tried the builtin bash variable substitution syntax. please re-review.

lzap (approved):

lzap:
Well done, works great.

PR #4516, opened 2017-05-09
fixes #19483 - use #admin? method to check inherited admin flags

domcleal:


tbrisker (approved):

tbrisker:
Thanks @domcleal !

ares:
I suppose this should have been cherry-picked to 1.15-stable too since if fixes the issue that was introduced by http://projects.theforeman.org/issues/18995 in 1.15.0. @dLobatog would you consider this for 1.15.2?

PR #4518, opened 2017-05-09
Fixes #19491 - properly i18n notification seed

jlsherrill:


tbrisker:
@jlsherrill looks like you missed a few strings in that file, mind fixing all of them?

jlsherrill:
Whoops, thanks @tbrisker !

tbrisker (approved):

tbrisker:
Thanks @jlsherrill !

ohadlevy:
@lzap a similar fix is probably needed on discovery, I would also suggest this fix goes into 1.15.1 ?

ares:
@dLobatog, @tbrisker I suppose this was not backported to 1.15.1 correct?

jlsherrill:
doesn't look like, i assume we could propose for 1.15.2 if it fits that criteria

PR #4520, opened 2017-05-10
Fixes #18438 - control notifications polling

matanwerbner:
This will allow the developer to control the timing of the notifications polling, or disable it:
on build, it reads the value of NOTIFICATIONS_POLLING enviroment variable, and:\
if not defined, it will default to polling with 10 sec interval.
if defined, it will poll with the supplied value in ms.
if false, polling will be disabled (after first polling request).

### DIFF
--- a/webpack/assets/javascripts/react_app/redux/actions/notifications/index.js
+++ b/webpack/assets/javascripts/react_app/redux/actions/notifications/index.js
@@ -9,7 +9,11 @@ import {
   notificationsDrawer as sessionStorage
 } from '../../../common/sessionStorage';
 import API from '../../../API';
-const getNotificationsInterval = 10000;
+import { isNil } from 'lodash';
+const defaultNotificationsPollingInterval = 10000;
+const notificationsInterval = isNil(process.env.NOTIFICATIONS_POLLING) ?
+  defaultNotificationsPollingInterval :
+  process.env.NOTIFICATIONS_POLLING;
 
 export const getNotifications = url => dispatch => {
   new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


webpack/assets/javascripts/react_app/redux/actions/notifications/index.js
```diff
@@ -9,33 +9,42 @@ import {
   notificationsDrawer as sessionStorage
 } from '../../../common/sessionStorage';
 import API from '../../../API';
-const getNotificationsInterval = 10000;
+import { isNil } from 'lodash';
+const defaultNotificationsPollingInterval = 10000;
+const notificationsInterval = isNil(process.env.NOTIFICATIONS_POLLING) ?
+  defaultNotificationsPollingInterval :
+  process.env.NOTIFICATIONS_POLLING;
```
ohadlevy, line 16: isnt that a string that should be converted into an int?

webpack/assets/javascripts/react_app/redux/actions/notifications/index.js
```diff
@@ -9,33 +9,42 @@ import {
...
-        dispatch({
-          type: NOTIFICATIONS_GET_NOTIFICATIONS,
-          payload: {
-            notifications: response.notifications
-          }
-        });
-        resolve();
-      }, () => reject());
+      API.get(url).then(
+        response => {
```
ohadlevy, line 25: what is the change here? not sure i follow?

webpack/assets/javascripts/react_app/redux/actions/notifications/index.js
matanwerbner: doesnt matter, setTimeout excepts strings as well

webpack/assets/javascripts/react_app/redux/actions/notifications/index.js
matanwerbner: probably just formatting... the relevant change is below with the "if" statement.

webpack/assets/javascripts/react_app/redux/actions/notifications/index.js
xprazak2: Could we leverage [dotenv](https://github.com/theforeman/foreman/blob/develop/config/webpack.config.js#L94-L97) for this?

matanwerbner: yes, but not here. in webpack.config.js... and there the require('dotenv') line should probably come first, before the DefinePlugin (where NOTIFICATIONS_POLLING is defined)

lzap:
To be honest, this could be global setting, users would like to lower it down also, like customers with many users. But this is fine too.

lzap:
In case this is merged as is (I do not have comments looks good to me) please make sure to add this into `/etc/sysconfig/foreman`. It's in here: https://github.com/theforeman/foreman-packaging/blob/rpm/develop/foreman/foreman.sysconfig

matanwerbner:
@ohadlevy anything else?

ohadlevy:
tests failures unrelated - thanks @matanwerbner.
@lzap this "feature" can be only used in development mode, not in production - thanks!

PR #4521, opened 2017-05-10
Fixes #19502 - fix react warnings

matanwerbner:
after upgrading react-dom to 15.1, some warnings started to appear regarding propTypes and react.createClass. this PR fixes them by using prop-types package, and upgrading react-bootstrap to 0.31.0


package.json
```diff
@@ -56,8 +56,9 @@
     "jstz": "~1.0.7",
     "lodash": "~4.15.0",
     "multiselect": "~0.9.12",
+    "prop-types": "^15.5.9",
     "react": "^15.1.0",
```
ohadlevy, line 60: dont we need to bump the react version too then?

matanwerbner: no..why?

ohadlevy: I thought the deprecation warnings just came from 15.5.4 and if we keep the older version features like props-type might conflict? or in other words, i'm uncertain if keeping 15.0 will work with the props-type package?

matanwerbner: since we have caret (^) npm installs automatically 15.5.4.
anyway i dont think it will conflict since we are using an external package here (prop-types)

ohadlevy: in our production builds, we still use 15.3 (see https://github.com/theforeman/foreman-packaging/tree/rpm/develop/nodejs-react) if you are OK with that version I'll ack.

matanwerbner: tested for 15.3.2 and works as expected.

ohadlevy:
thanks @matanwerbner

PR #4522, opened 2017-05-10
Fixes #19504 - Quote puppet class filter in host form

tbrisker:
So that the filter can contain characters such as ":"


timogoebel (approved):

timogoebel:
Merged, thanks @tbrisker. Test failures were unrelated.

PR #4523, opened 2017-05-10
show flash messages in two-pane views

mdellweg:
Copy the flash items from the ajax response to the main DOM before
triggering 'ContentLoad'.

### DIFF
--- a/app/assets/javascripts/two-pane.js
+++ b/app/assets/javascripts/two-pane.js
@@ -144,6 +144,24 @@ function right_pane_content(response){
   var contentId;
   if (handle_redirect(response)) return; //session expired redirect to login
 
+  // move notifications to main document
+  var notification_tag = $('#notifications', response)[0];
+  if (notification_tag) {
+    var dest_array = $($('#notifications')[0]).data('flash');
+    var flash_array = $(notification_tag).data('flash');
+    var length = flash_array.length;
+    var i;
+    $(notification_tag).detach();
+    if (dest_array == "") {
+      dest_array = Array();
+    }
+    for (i = 0; i < length; i++){
+      dest_array.push(flash_array[i]);
+    }
+    $('#notifications').data('flash', dest_array);
+    tfm.toastNotifications.clear();
+  }
+
   if (!$("#content", response).length){
     $('.two-pane-right').html(response);
     $('.two-pane-right form').prepend("<div class='fr close-button'><span class='pficon pficon-close two-pane-close'></span></div>");
--- a/app/assets/javascripts/two-pane.js
+++ b/app/assets/javascripts/two-pane.js
@@ -136,7 +136,6 @@ function hide_columns(){
     $('.table-two-pane').wrap( "<div class='row'><div class='col-md-3 two-pane-left'></div></div>");
   }
   tfm.tools.showSpinner();
-  tfm.toastNotifications.clear()
   $('.two-pane-left').after("<div class='col-md-9 two-pane-right'></div>");
 }
 
--- a/app/views/layouts/_application_content.html.erb
+++ b/app/views/layouts/_application_content.html.erb
@@ -1,6 +1,6 @@
 <div id="main">
+  <%= notifications %>
   <div id="content" class="container">
-    <%= notifications %>
     <% unless @page_header.blank? %>
       <div class="row form-group">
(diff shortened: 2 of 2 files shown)


ohadlevy:
@mdellweg thanks! can you confirm this is needed on latest development? i think its overlapping with #4513 that was merged today?

mdellweg:
@ohadlevy You are right, i missed that. But i think, you should consider this solution anyway, because in case, there is a "#two_pane_content" class object (see around line 160 of two-pane.js), even less context will be copied. Moving the flash messages inside "#content" won't suffice there.
Additionally, i refactored my solution to not duplicate the messages when combined with d28cacf.

app/assets/javascripts/two-pane.js
```diff
@@ -144,6 +144,11 @@ function right_pane_content(response){
   var contentId;
   if (handle_redirect(response)) return; //session expired redirect to login
 
+  // move flash messages from response to main window
+  $('.flash', response).each(function(index, item) {
```
ohadlevy: I'm not sure I'm following, i think that .flash doesnt exists anymore on develop branch?

mdellweg: I see your point. I solved the problem on 1.14-stable branch. I did not expect, that the whole framework had been changed. But at least for 1.14 it solves the problem.
My critique with your solution is based on the fact, that `right_pane_content` loads the `div`-tag with class `content` or preferably class `two-pane-content` if available.
```
if (!$("#content", response).length){
    $('.two-pane-right').html(response);
    $('.two-pane-right form').prepend("<div class='fr close-button'><a class='two-pane-close' href='#'>&times;</a></div>");
    $('.form-actions a').addClass('two-pane-close');
    fix_multi_checkbox();
  } else {
    // response is not a form use the entire page
    if ($('#two_pane_content', response).length) {
      // replace only #two_pane_content if it's present in the response
      contentId = '#two_pane_content';
      two_pane_close();
    } else {
      contentId = '#content';
    }
    $(contentId).replaceWith($(contentId, response));
  }
```
I'm not sure whether this code path is used at all. But in case the AJAX response contains a `two-pane-content`, we would miss the notifications again.
Copying them to the main document, regardless where you find them seems the save way to me. Do you think, on the devel branch, we should copy everything inside `$("#notifications")` into the corresponding tag of the main document?

mdellweg: Trying to translate this to the develop branch, i ended up with someting like this:
```
function right_pane_content(response){
  var contentId;
  if (handle_redirect(response)) return; //session expired redirect to login

  // move notifications to main document
  var notification_tag = $('#notifications', response)[0];
  if (notification_tag) {
    var dest_array = $($('#notifications')[0]).data('flash');
    var flash_array = $(notification_tag).data('flash');
    var length = flash_array.length;
    var i;
    $(notification_tag).detach();
    if (dest_array == "") {
      dest_array = Array();
    }
    for (i = 0; i < length; i++){
      dest_array.push(flash_array[i]);
    }
    $('#notifications').data('flash', dest_array);
    tfm.toastNotifications.clear();
  }

  if (!$("#content", response).length){
<...>
```
If we could be sure, that the earlier notifications were already toasted, it might suffice to just replace the whole `data-flash` attribute.

ohadlevy: I guess I'm missing the value in your change, why is it required? (note that we simply changed the notification div to be inside the content div.

if you think its still required, please update the PR to work against the develop branch. thanks!

mdellweg:
I see a problem, when the response contains a '#two_pane_content' tag, let's say
```
<div id='content'>
  <div id='notifications'>
    <!-- some more or less important notification -->
    <!-- they will not show, because only the table will be transferred to the main dom -->
  </div>
  <table id=`two_pane_content`>
    <!-- only this table will be transferred to the main page after closing the overlay -->
  </table>
</div>
```
I see this in `app/views/compute_resources/show.html.erb:65`.
Transferring the notifications to a common place (possibly outside of '#content') regardless of where you find them in the ajax response seems to me to be the save way, because it is independent of which part of the page you are actually patching.
I will update the PR.

ohadlevy:
I believe this is already resolved in 1.16 - please reopen in case we've missed something.

PR #4524, opened 2017-05-10
Fixes #19506 - reset cache after CR update

ares:


ares:
actually this might desired only if the caching is enabled for the CR

timogoebel:
> actually this might desired only if the caching is enabled for the CR

Yes. One more comment. Should the callback function always return true to not stop other callbacks from running?

ares:
Thanks @timogoebel, could you take another look please?

timogoebel (approved):
@ares: Thanks. Will merge once tests pass.

timogoebel:
Merged, thanks @ares.

PR #4525, opened 2017-05-10
Refs #19512 - use nodejs 6.10 for EL7 tests

mmoll:
also add 4.8, which is the current legacy LTS release and used in Debian/stretch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,9 @@ node_js:
   #- "0.10" # EL6, Trusty, Jessie - sadly, jest does not work with 0.10
   - "4.2" # Xenial
   - "4.6" # Fedora 24
-  - "6.9" # EL 7
+  - "4.8" # legacy LTS, Stretch
+  - "6.9" # initial EPEL 7
+  - "6.10" # current LTS, current EPEL 7
 before_install:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -3,7 +3,9 @@ node_js:
   #- "0.10" # EL6, Trusty, Jessie - sadly, jest does not work with 0.10
   - "4.2" # Xenial
   - "4.6" # Fedora 24
-  - "6.9" # EL 7
+  - "4.8" # legacy LTS, Stretch
+  - "6.9" # initial EPEL 7
```
tbrisker, line 7: I think we can drop this if EPEL now ships with 6.10, no?

tbrisker (commented):
see inline. btw, why did you use `refs` and not `fixes` in the commit message?

.travis.yml
mmoll: I'd like to remove that together with everything non-6.10 when working on raising the minimum requirement as now at least some EL builders don't receive automatic updates and would have 6.9 installed.

mmoll:
> why did you use refs and not fixes

This is only the preparation for requiring 6.10 later

tbrisker:
Thanks @mmoll !

PR #4526, opened 2017-05-12
Fixes #17883 - enforce newer fog-vsphere

ares:
this version escapes the datacenter name properly and therefore fixes the issue


tbrisker:
Thanks @ares !

PR #4528, opened 2017-05-14
Fixes #19296 - Combination remove link shows correct tooltip

orrabin:


app/helpers/form_helper.rb
```diff
@@ -339,7 +339,8 @@ def blank_or_inherit_f(f, attr)
   end
 
   def link_to_remove_fields(name, f, options = {})
-    f.hidden_field(:_destroy) + link_to_function(icon_text('remove', name, :kind => 'pficon'), "remove_fields(this)", options.merge(:title => _("Remove Parameter")))
+    options.merge!(:title => _("Remove Parameter")) unless options[:title].present?
```
tbrisker: The same result could also be achieved using `options[:title] ||= _("Remove Parameter")`, or by `options.reverse_merge(:title => _("Remove Parameter"))`, or by using a kwarg for title with a default value and merging that into the other options.

tbrisker (approved):

PR #4529, opened 2017-05-15
Fixes #19534 - removed excessive encryptor logging

lzap:
I am still keeping the two most important one, when en/decryption fails, but we only log this once.

### DIFF
--- a/app/models/concerns/encrypt_value.rb
+++ b/app/models/concerns/encrypt_value.rb
@@ -11,14 +11,7 @@ module EncryptValue
   end
 
   def is_encryptable?(str)
-    if !encryption_key.present?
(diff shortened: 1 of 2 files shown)


timogoebel:
[test upgrade]

@lzap : Thanks. We should do either this or use a dedicated logger for this. I don't see any value in logging this at all. I'll leave this open for a bit before merging to give people time to comment on this.

tbrisker (approved):

lzap:
I accidentaly pushed some other code into this PR, fixed. Sorry about that.

lzap:
Tests failures unrelated, gem deps issues plus assert

    NoMethodError: undefined method `split' for nil:NilClass

timogoebel:
Merged, thanks @lzap.

PR #4531, opened 2017-05-15
Fixes #19541 - properly mark user <-> mail_notification relation

ShimShtein:
Mark `:inverse_of` on `User` and `UserMailNotifiaction` objects so the relation properties would be set automatically, even for new in-memory instances.

Example:
```
[6] pry(main)> User.new(:login => 'zzz', :mail_notifications => [MailNotification.new]).user_mail_notifications.first.user
=> #<User:0x0000000c16a330
 id: nil,
 login: "zzz",
 firstname: nil,
...
```

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -43,7 +43,7 @@ class User < ApplicationRecord
   has_many :widgets, :dependent => :destroy
   has_many :ssh_keys, :dependent => :destroy
 
-  has_many :user_mail_notifications, :dependent => :destroy
+  has_many :user_mail_notifications, :dependent => :destroy, :inverse_of => :user
   has_many :mail_notifications, :through => :user_mail_notifications
 
   accepts_nested_attributes_for :user_mail_notifications, :allow_destroy => true, :reject_if => :reject_empty_intervals
--- a/app/models/user_mail_notification.rb
+++ b/app/models/user_mail_notification.rb
@@ -1,6 +1,6 @@
 class UserMailNotification < ApplicationRecord
-  belongs_to :user
-  belongs_to :mail_notification
+  belongs_to :user, :inverse_of => :user_mail_notifications
+  belongs_to :mail_notification, :inverse_of => :user_mail_notifications
 
   scope :daily, -> { where(:interval => 'Daily') }
   scope :weekly, -> { where(:interval => 'Weekly') }


domcleal:
What does "fixed" mean? The issue description makes little sense to me, the factory appears to work correctly, and associations should work whether built or created.

ShimShtein:
Current implementation does not populate the `:user` field in `:user_mail_notifications` records. You can see it clearly if you add back `validates :user, :presence => true` validation to `UserMailNotification` model.

domcleal:
> Current implementation does not populate the :user field in :user_mail_notifications records

During build or create? Seems to work fine on a create:

<pre>
> FactoryGirl.create(:user, :with_mail_notification).user_mail_notifications.first
=> #&lt;UserMailNotification:0x0000000b34ed78 id: 3, user_id: 980190966, mail_notification_id: 1018186302, last_sent: nil, interval: nil, created_at: Mon, 15 May 2017 13:05:24 UTC +00:00, updated_at: Mon, 15 May 2017 13:05:24 UTC +00:00, mail_query: nil>
</pre>

build isn't applicable as the user doesn't yet exist. It'll be populated on save.

> You can see it clearly if you add back validates :user, :presence => true validation to UserMailNotification model.

There's no such validation in the model, and there shouldn't be.. if the object is built via the user's association then it'll be populated when _saved_ and the ID is known. The user would fail pre-save validation otherwise, and it wouldn't be possible to create an object with associations.

ShimShtein:
@domcleal, after a small research I can see that the relations are handled properly (even for unsaved models) as long as `:inverse_of` are set on them.

It allows the pre-save validators to run correctly (tested it locally) for unsaved models too.

As a side effect this reduces the number of created objects and decreases the number of SQL queries, although I don't think the effect is too significant.

I have updated the first comment with an example that shows proper handling of unsaved models.

Thank you for making me dig deeper into AR relations! I think this solution is much more elegant than the previous one!

domcleal:
That's fine, thanks. Could you please update the ticket description to show an example of what this should actually fix? It currently describes the earlier "fix" (which is now no longer relevant), not the issue itself.

ShimShtein:
Updated the issue as requested.

domcleal:
Merged, thanks @ShimShtein.

PR #4534, opened 2017-05-18
Fixes #19582 - Alert when puppetclass is not saved due to error

orrabin:
Getting an error from a smart class parameter or smart variable that was not edited and prevents saving is confusing.
This alert is to make sure it's clear the puppet class will not be saved if there are any errors in any of it's parameters/variables.

### DIFF
--- a/app/views/puppetclasses/_form.html.erb
+++ b/app/views/puppetclasses/_form.html.erb
@@ -1,5 +1,8 @@
 <%= form_for @puppetclass, :url => (@puppetclass.new_record? ? puppetclasses_path : puppetclass_path(:id => @puppetclass)) do |f| %>
   <%= base_errors_for @puppetclass %>
+  <% unless @puppetclass.errors.empty? %>
+    <%= alert :class => 'alert-danger', :header => '', :text => _("The class could not be saved because of an error in one of the class parameters.") %>
+  <% end %>
   <ul class="nav nav-tabs" data-tabs="tabs">
     <li class="active"><a href="#primary" data-toggle="tab"><%= _('Puppet Class') %></a></li>
     <li><a href="#smart_class_param" data-toggle="tab"><%= _('Smart Class Parameter') %></a></li>


app/views/puppetclasses/_form.html.erb
```diff
@@ -1,5 +1,8 @@
 <%= form_for @puppetclass, :url => (@puppetclass.new_record? ? puppetclasses_path : puppetclass_path(:id => @puppetclass)) do |f| %>
   <%= base_errors_for @puppetclass %>
+  <% unless @puppetclass.errors.empty? %>
```
orrabin, line 3: should this be only for errors on parameters/variables?

orrabin:
@Rohoover is this the correct place and wording of the inline notification?

image

Rohoover:
@orrabin 
Can the error occur on any of the tabs? Or is it specifically in the Smart Variable Tab?

orrabin:
> Can the error occur on any of the tabs? Or is it specifically in the Smart Variable Tab?

@Rohoover any of the tabs

Rohoover:
@orrabin 

So I think the position is correct because the error affects the whole page.

app/views/puppetclasses/_form.html.erb
```diff
@@ -1,5 +1,8 @@
...
+    <%= alert :class => 'alert-danger', :header => '', :text => _("The puppet class has not been saved because of an existing error in the class or one of it's parameters.") %>
```
ShimShtein: nitpick: IMHO, it should be either `"Puppet class has not..."` or `"The class has not..."`.

ShimShtein (commented):
@ohadlevy, you did some work with error notifications, should it affect this message in some way?

orrabin:
@ShimShtein thanks, I fixed the wording.

app/views/puppetclasses/_form.html.erb
```diff
@@ -1,5 +1,8 @@
 <%= form_for @puppetclass, :url => (@puppetclass.new_record? ? puppetclasses_path : puppetclass_path(:id => @puppetclass)) do |f| %>
   <%= base_errors_for @puppetclass %>
+  <% unless @puppetclass.errors.empty? %>
+    <%= alert :class => 'alert-danger', :header => '', :text => _("The class has not been saved because of an existing error in the class or one of it's parameters.") %>
```
tbrisker: nitpick: its not it's, also I would drop the word "existing"

dLobatog (changes requested):
Set to WOC as per @tbrisker 's comment about the typo, I haven't tried it but it looks fine

orrabin:
@tbrisker thanks, I fixed the typo.

app/views/puppetclasses/_form.html.erb
tbrisker: I don't think there can be any error in the class itself, so if there is any error it has to be in the parameters/variables.

app/views/puppetclasses/_form.html.erb
```diff
@@ -1,5 +1,8 @@
 <%= form_for @puppetclass, :url => (@puppetclass.new_record? ? puppetclasses_path : puppetclass_path(:id => @puppetclass)) do |f| %>
   <%= base_errors_for @puppetclass %>
+  <% unless @puppetclass.errors.empty? %>
+    <%= alert :class => 'alert-danger', :header => '', :text => _("The class has not been saved because of an error in the class or one of its parameters.") %>
```
tbrisker: could you reword this to `The class could not be saved because of an error in one of the class parameters.`?

tbrisker (approved):

tbrisker:
Thanks @orrabin !

PR #4537, opened 2017-05-18
Fixes #19585 - Allow csv export of nested attributes

tbrisker:
This is required for example to export host facet attributes.

### DIFF
--- a/test/unit/csv_exporter_test.rb
+++ b/test/unit/csv_exporter_test.rb
@@ -5,6 +5,9 @@ class CsvExporterTest < ActiveSupport::TestCase
     result = CsvExporter.export(Host::Managed, [:id])
     assert_equal "Id\n", result.next
     assert_equal result.count, Host::Managed.count+1
+    assert_difference('CsvExporter.export(Host::Managed, [:id]).count') do
+      FactoryGirl.create(:host)
+    end
   end
 
   test 'handles empty results correctly' do
@@ -22,6 +25,16 @@ class CsvExporterTest < ActiveSupport::TestCase
     result = CsvExporter.export(Domain.where(:id => id), [:id, :test_method])
     assert_equal "Id,Test Method\n", result.next
     assert_equal "#{id},success!\n", result.next
+    assert_raises StopIteration do
+      result.next
+    end
+  end
+
+  test 'calls nested methods on records' do
+    host = FactoryGirl.create(:host)
+    result = CsvExporter.export(Host::Managed, [:name, 'location.name'])
+    assert_equal "Name,Location.Name\n", result.next
+    assert_equal "#{host.name},#{host.location.name}\n", result.next
     assert_raises StopIteration do
       result.next
     end
--- a/app/services/csv_exporter.rb
+++ b/app/services/csv_exporter.rb
@@ -2,12 +2,14 @@ require 'csv'
 
 module CsvExporter
   def self.export(resources, columns)
+    header = csv_header(columns)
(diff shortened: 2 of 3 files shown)


test/unit/csv_exporter_test.rb
```diff
@@ -5,6 +5,9 @@ class CsvExporterTest < ActiveSupport::TestCase
     result = CsvExporter.export(Host::Managed, [:id])
     assert_equal "Id\n", result.next
     assert_equal result.count, Host::Managed.count+1
+    assert_difference('CsvExporter.export(Host::Managed, [:id]).count') do
```
tbrisker, line 8: not directly related to this but I noticed that previously the test may have had incorrect results as there were no hosts in the database.

ohadlevy:
@tbrisker can you please provide example usage case and output? I'm just not seeing exactly how this is used yet :) - test failure seems unrelated.

tbrisker:
@ohadlevy I need this currently for adding CSV export on content hosts, where I would want to have columns such as `content_facet.applicable_errata.count` for example.

ShimShtein:
Nice feature! Two comments though:
1. (less critical) The auto-generated name looks a bit strange, the upper case for each method is not necessary
2. No exception handling for non-existent methods.

Examples:
```
[6] pry(main)> CsvExporter.export(Host.all, [:name, 'interfaces.first.mac']).map{ |line| line}
=> ["Name,Interfaces.First.Mac\n",
 "tommy-havier.mydomain.com,4e:11:11:11:11:11\n",
 "lucy-randles.mydomain.com,52:54:00:58:55:a7\n",
 "shari-general.mydomain.com,00:1a:4a:16:01:5d\n",
 "client.foreman.test,52:54:00:79:0e:c7\n",
 "matt-knoff.mydomain.com,00:1a:4a:16:01:52\n"]
[7] pry(main)> CsvExporter.export(Host.all, [:name, 'interfaces.mac']).map{ |line| line}
=> ["Name,Interfaces.Mac\n",
 "tommy-havier.mydomain.com,\n",
 "lucy-randles.mydomain.com,\n",
 "shari-general.mydomain.com,\n",
 "client.foreman.test,\n",
 "matt-knoff.mydomain.com,\n"]
```

ohadlevy:
can we rename columns? for example
``` CsvExporter.export(Host.all, [:name, 'interfaces.first.mac'] ``` generates ```Name,Interfaces.Mac```

where I really want mac or mac-1 or something else.

tbrisker:
@ohadlevy @ShimShtein - #4538 is for allowing to rename the headers, I did it in two PRs on purpose to make review easier. As for non-existing methods - I used try on purpose because you may have certain methods only available for some objects, there should be no exceptions thrown, just empty cells when there is no data.

ShimShtein:
I would suggest underscores for default names (I know it is overridable in the other PR).

About non-existing methods: well, I wanted to protect from completely wrong paths (where method names are misspelled e.t.c.), but it can be managed through proper unit testing, so I guess it's not that problematic.

:+1: from me - tested and works.

tbrisker:
@ohadlevy ping?

ohadlevy:
@tbrisker I see the two PR's related in term of interface, I would prefer to finalize the discussion on #4538 first?

tbrisker:
@ohadlevy the two PRs don't depend on each other, which is why I split them in two. Once one is merged I'll rebase the other and make whatever changes are needed.

ohadlevy:
thanks @tbrisker

PR #4538, opened 2017-05-18
Fixes #19588 - Allow setting custom headers in CSV export

tbrisker:


app/services/csv_exporter.rb
```diff
@@ -1,9 +1,10 @@
 require 'csv'
 
 module CsvExporter
-  def self.export(resources, columns)
+  def self.export(resources, columns, header = default_header(columns))
+    raise ArgumentError, "Columns and header row aren't the same length" unless columns.length == header.length
```
ShimShtein: IMHO the message isn't clear enough, maybe something like `"Columns count does not match headers count"`?

ShimShtein (commented):
ACK, works as expected:
```
[5] pry(main)> CsvExporter.export(Architecture.all, [:name, :created_at], ['my awesome name'])
ArgumentError: Columns and header row aren't the same length
from /home/shim/Documents/foreman/foreman/app/services/csv_exporter.rb:5:in `export'
[6] pry(main)> CsvExporter.export(Architecture.all, [:name, :created_at], ['my awesome name', nil])
=> #<Enumerator: ...>
[7] pry(main)> CsvExporter.export(Architecture.all, [:name, :created_at], ['my awesome name', nil]).map { |aa| p aa }
"my awesome name,\n"
"x86_64,2016-09-11 14:24:01 UTC\n"
"i386,2016-09-11 14:24:01 UTC\n"
=> ["my awesome name,\n", "x86_64,2016-09-11 14:24:01 UTC\n", "i386,2016-09-11 14:24:01 UTC\n"]
[8] pry(main)> CsvExporter.export(Architecture.all, [:name, :created_at], ['my awesome name', :another_column]).map { |aa| p aa }
"my awesome name,another_column\n"
"x86_64,2016-09-11 14:24:01 UTC\n"
"i386,2016-09-11 14:24:01 UTC\n"
=> ["my awesome name,another_column\n", "x86_64,2016-09-11 14:24:01 UTC\n", "i386,2016-09-11 14:24:01 UTC\n"]
```

app/services/csv_exporter.rb
ohadlevy: who is the recipient of the error? I would not want to get this in the UI...

tbrisker: This error shouldn't ever occur since the parameters are always hard coded at the moment in the callers. It's more of a defensive programming check to make sure the developers don't miscount the number of columns and create invalid CSVs.

test/unit/csv_exporter_test.rb
```diff
@@ -26,4 +26,15 @@ class CsvExporterTest < ActiveSupport::TestCase
       result.next
     end
   end
+
+  test 'accepts custom column headers' do
+    result = CsvExporter.export(Host::Managed, [:id], ['My Lovely Header'])
```
ohadlevy, line 44: would the following be a better interface?
```ruby
CsvExporter.export(Host::Managed, [:id => ['My Lovely Header'])
```
or something along those lines? is there any other attribute besides rename?

if nothing else, I would prefer ```[[header, title],[header,title]]```..

tbrisker: In most cases there is no need to rename the headers, so I feel that requiring a title for every column would complicate both the parameters and the implementation.

dLobatog (approved):
Seems good to me, I also think requiring a title shouldn't be necessary. Waiting on ACK by @ohadlevy since he expressed concern about the API https://github.com/theforeman/foreman/pull/4538/files#r119603798

ohadlevy (approved):

ohadlevy:
thanks @tbrisker

PR #4539, opened 2017-05-18
Fixes #19416 - LDAPS authentication for FreeBSD

karli-sjoberg:
It says "Fixes" because it´s demanded in the guidelines but I don´t know if it fixes it for OP of Redmine issue. All I know is that this change fixes LDAPS authentication for Foreman running on FreeBSD:)

### DIFF
--- a/app/models/auth_sources/auth_source_ldap.rb
+++ b/app/models/auth_sources/auth_source_ldap.rb
@@ -15,6 +15,7 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
+require 'rbconfig'
 require 'net/ldap'
 require 'timeout'
 
@@ -90,7 +91,11 @@ class AuthSourceLdap < AuthSource
 
   def encryption_config
     return nil unless tls
-    { :method => :simple_tls, :tls_options => { :verify_mode => OpenSSL::SSL::VERIFY_PEER } }
+    if RbConfig::CONFIG['host_os'] =~ /freebsd/i
+      { :method => :simple_tls, :tls_options => { :ca_file => "/usr/local/etc/ssl/cert.pem", :verify_mode => OpenSSL::SSL::VERIFY_PEER } }
+    else
+      { :method => :simple_tls, :tls_options => { :verify_mode => OpenSSL::SSL::VERIFY_PEER } }
+    end
   end
 
   def ldap_con(login = nil, password = nil)


dLobatog (changes requested):
Thanks for the PR @karli-sjoberg , unfortunately I don't have a FreeBSD installation around nor it's one of our supported platforms, so I don't think making special options for it is a way to go.

What changes on FreeBSD that does not allow this file to be added to the system CA bundle? That's what is used by default.

karli-sjoberg:
Hi Daniel!

On Thu, 2017-05-25 at 08:11 -0700, Daniel Lobato García wrote:
> @dLobatog requested changes on this pull request.
> Thanks for the PR, unfortunately I don't have a FreeBSD installation
> around nor it's one of our supported platforms, so I don't think
> making special options for it is a way to go.

This response confuses and disappoints me greatly. It´s just the kind
of thing I would have expected from a proprietary provider but
absolutely _not_ from an open source project built around the fact that
everybody is different, and we adjust accordingly!

Have a look at node.rb[*], which is from where I copied the code that
senses the operatingsystem. Why would you allow the code there but not
here, that is ridiculous?:)

> What changes on FreeBSD that does not allow this file to be added to
> the system CA bundle? That's what is used by default.

"/usr/local/etc/ssl/cert.pem" _is_ the system CA bundle in FreeBSD:)
Just like "/etc/pki/tls/cert.pem" in RedHat or "/etc/ssl/certs/ca-
certificates.crt" in Debian.

This is a very easy fix for everyone that runs Foreman on FreeBSD that
needs LDAPS (why wouldn´t you) and doesn´t break anything for anyone
else.

Best Regards
Karli Sjöberg

[*] https://raw.githubusercontent.com/theforeman/puppet-foreman/master/
files/external_node_v2.rb

mmoll:
@karli-sjoberg the file you changed in this PR here doesn't have any fixed path for any Linux distro, correct?

karli-sjoberg:
Den 25 maj 2017 8:05 em skrev Michael Moll <notifications@github.com>:@karli-sjoberg the file you changed in this PR here doesn't have any fixed path for any Linux distro, correct?Yes, that is correct.

mmoll:
@karli-sjoberg then @dLobatog is correct. Once it's clear how to add the certs in question to the certs store on FreeBSD in a way that works in general, the docs should get updated. I guess the underlying problem is some base/ports problem on FreeBSD.

karli-sjoberg:
On Thu, 2017-05-25 at 11:11 -0700, Michael Moll wrote:
> @karli-sjoberg then @dLobatog is correct. Once it's clear how to add
> the certs in question to the certs store on FreeBSD in a way that
> works in general, the docs should get updated.

Just so we are clear, it´s not about adding certs to a certs
store, "/usr/local/etc/ssl/cert.pem" _is_ the certs store in FreeBSD,
but Foreman doesn´t recognize (or know about) it.

> I guess the underlying problem is some base/ports problem on FreeBSD.

All other applications that uses it have no problem with it, so I
wouldn´t dismiss it so quickly. My feeling is that it lies somewhere in
Foreman or one of it´s gems. That´s where the problem is.

I´m going to dig a little deeper into this.

/K

mmoll:
> My feeling is that it lies somewhere in Foreman or one of it´s gems. That´s where the problem is.

I think it has to do with FreeBSD's ports ruby and its usage of the cert store.

karli-sjoberg:
On Thu, 2017-05-25 at 12:12 -0700, Michael Moll wrote:
> My feeling is that it lies somewhere in Foreman or one of it´s gems.
> That´s where the problem is.
> I think it has to do with FreeBSD's ports ruby and its usage of the
> cert store.

Aha, I found it!

You were right it had to do with ruby. I wrote a quick script that
prints ruby´s "OpenSSL::X509::DEFAULT_CERT_FILE", which printed:
"/usr/local/openssl/cert.pem" and not "/usr/local/etc/ssl/cert.pem".
Adding my CA to this file made LDAPS work with "auth_source_ldap.rb"
unedited.

This pull request can therefore be scrapped and I will update my
documentation and the Redmine issue instead.

Thank you for your insights!

/K

dLobatog:
@karli-sjoberg We do support many systems (including FreeBSD or even Windows) as proxies or clients. (https://theforeman.org/manuals/1.15/quickstart_guide.html#QuickstartGuide lists at the top the systems we officially support with packages etc..) The node.rb file patch you linked is made to allow a Puppet master in FreeBSD contact Foreman as an ENC, it's not part of the Foreman code. 

Anyway, I'm glad it's fixed now! Now it's time to update the docs. I'll add a small section to https://theforeman.org/manuals/1.15/#4.1.1LDAPAuthentication (Trusting SSL certs) including what you said - a script to print the `OpenSSL::X509::DEFAULT_CERT_FILE` to make sure users know what Ruby is using. Thanks for your help!

PR #4540, opened 2017-05-19
Fixes #19496 - added passenger recycler

lzap:
Passenger open-source version does support process restarting on memory threshold. It looks like passenger have reserved SIGUSR1 for graceful restart. A cron job could do the job easily.

I wrote initial version of a script that gracefully terminates Passenger processes consuming more than 2 GB of private RSS memory. Typical consumption of Foreman with Katello and other basic plugins is around 1.2 GB and since only Passenger Enterprise allows to limit maximum amount of RSS memory for processes, this simple script does exactly that.

Motivation is simple - we often introduce bugs in our Rails codebase which performs some eager loading or there are memory leaks in our code or dependencies and Passenger processes can grow up to dozens gigabytes. Unfortunately, there is no other way of getting out other than restarting httpd with passenger. This script could help to avoid situations when production instance starts to swap hard thank to some small regression we introduced. Also administrator will be noticed early via email when this happens so we will keep track of these regressions in production.

Once merged I will follow up with a cron job in foreman package so by default we run it once every hour.

### DIFF
--- a/Rakefile.dist
+++ b/Rakefile.dist
@@ -14,6 +14,7 @@ MANDIR = ENV['MANDIR'] || "#{DATAROOTDIR}/man"
 
 def is_redhat; File.exist?("/etc/redhat-release"); end
 def is_debian; File.exist?("/etc/debian-release"); end
+def cp_p(*a); cp(*a, {:verbose => true, :preserve => true}); end
 
 file BUILDDIR do
   mkdir BUILDDIR
@@ -43,18 +44,18 @@ task :build => [
 
 task :install => :build do |t|
   mkdir_p "#{MANDIR}/man8"
-  cp "#{BUILDDIR}/foreman-rake.8.gz", "#{MANDIR}/man8/"
-  cp "#{BUILDDIR}/foreman-debug.8.gz", "#{MANDIR}/man8/"
-  cp "#{BUILDDIR}/foreman-tail.8.gz", "#{MANDIR}/man8/"
+  cp_p "#{BUILDDIR}/foreman-rake.8.gz", "#{MANDIR}/man8/"
+  cp_p "#{BUILDDIR}/foreman-debug.8.gz", "#{MANDIR}/man8/"
+  cp_p "#{BUILDDIR}/foreman-tail.8.gz", "#{MANDIR}/man8/"
 
   # foreman-tail
   TAILDIR = ENV['TAILDIR'] || "script/foreman-tail.d"
   mkdir_p "#{DATAROOTDIR}/foreman/#{TAILDIR}"
-  cp Dir.glob("#{TAILDIR}/common/*"), "#{DATAROOTDIR}/foreman/#{TAILDIR}/", :verbose => true
+  cp_p Dir.glob("#{TAILDIR}/common/*"), "#{DATAROOTDIR}/foreman/#{TAILDIR}/"
   if is_redhat
-    cp Dir.glob("#{TAILDIR}/redhat/*"), "#{DATAROOTDIR}/foreman/#{TAILDIR}/", :verbose => true
+    cp_p Dir.glob("#{TAILDIR}/redhat/*"), "#{DATAROOTDIR}/foreman/#{TAILDIR}/"
   elsif is_debian
-    cp Dir.glob("#{TAILDIR}/debian/*"), "#{DATAROOTDIR}/foreman/#{TAILDIR}/", :verbose => true
+    cp_p Dir.glob("#{TAILDIR}/debian/*"), "#{DATAROOTDIR}/foreman/#{TAILDIR}/"
   else
     raise "Unsupported system"
   end
@@ -65,6 +66,14 @@ task :install => :build do |t|
 
   # ssh
   mkdir_p "#{DATAROOTDIR}/foreman/.ssh"
+
+  # cron scripts
+  mkdir_p "#{SYSCONFDIR}/cron.hourly"
+  cp_p "script/passenger-recycler.rb", "#{SYSCONFDIR}/cron.hourly/passenger-recycler"
+
+  # configuration files
+  mkdir_p "#{SYSCONFDIR}/foreman"
+  cp_p "script/passenger-recycler.rb.conf.example", "#{SYSCONFDIR}/foreman/passenger-recycler.rb.conf"
 end
 
 task :default => :build
--- a/script/passenger-recycler.rb
+++ b/script/passenger-recycler.rb
@@ -1,0 +1,59 @@
+#!/usr/bin/env ruby
+#
+# Trivial Passenger memory monitor installed via /etc/cron.hourly/. To modify
+# configuration create /etc/foreman/passenger-recycler.rb.conf file.
+#
+
+CONFIG = '/etc/foreman/passenger-recycler.rb.conf'
+load CONFIG if File.readable?(CONFIG)
+ENABLED ||= true
+MAX_PRIV_RSS_MEMORY ||= 2_000_000
+GRACEFUL_SHUTDOWN_SLEEP ||= 120
+KILL_BUSY ||= true
+SEND_STATUS ||= true
+exit 0 unless ENABLED
+
+def running?(pid)
+  return Process.getpgid(pid) != -1
+rescue Errno::ESRCH
+  return false
+end
+
+require 'phusion_passenger'
+require 'phusion_passenger/platform_info'
+require 'phusion_passenger/platform_info/ruby'
+require 'phusion_passenger/admin_tools/memory_stats'
+stats = PhusionPassenger::AdminTools::MemoryStats.new
+unless stats.platform_provides_private_dirty_rss_information?
+  puts "Please run as root or platform unsupported"
+  exit 1
+end
+stats.passenger_processes.each do |p|
+  if p.private_dirty_rss > MAX_PRIV_RSS_MEMORY
+    pid = p.pid.to_i
+    started = `ps -p#{pid} -o start=`.strip rescue '?'
+    status_ps = `ps -p#{pid} -u`
+    status_all = `passenger-status`
+    status_backtraces = `passenger-status --show=backtraces`
+    puts "Terminating #{pid} (started #{started}) with private dirty RSS size of #{p.private_dirty_rss} MB"
+    Process.kill "SIGUSR1", pid
+    sleep GRACEFUL_SHUTDOWN_SLEEP
+    if running?(pid) && KILL_BUSY
+      puts "Process #{pid} is still running, sending SIGKILL"
+      Process.kill "KILL", pid
+      sleep 5
+    end
+    if running?(pid)
+      puts "Process #{pid} still terminating, moving on..."
+    else
+      puts "Process successfully #{pid} terminated"
+    end
+    if SEND_STATUS
+      puts status_ps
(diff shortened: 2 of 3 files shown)


lzap:
@ShimShtein rebased.

Rakefile.dist
```diff
@@ -65,6 +66,14 @@ task :install => :build do |t|
...
   # ssh
   mkdir_p "#{DATAROOTDIR}/foreman/.ssh"
+
+  # cron scripts
+  mkdir_p "#{SYSCONFDIR}/cron.hourly"
+  cp_p "script/passenger-recycler.rb", "#{SYSCONFDIR}/cron.hourly/passenger-recycler"
+
+  # configuration files
+  mkdir_p "#{SYSCONFDIR}/foreman"
+  cp_p "script/passenger-recycler.rb.conf.example", "#{SYSCONFDIR}/foreman/passenger-recycler.rb.conf"
```
lzap, line 76: @domcleal any objections against starting  moving configuration files out of packaging repo into core repo? I believe `Rakefile.dist` is the correct place of this, less work in the `foreman-packaging`.

domcleal: This file isn't in the foreman-packaging repo. If it's not OS specific, it shouldn't be in packaging.

lzap: Yeah, I am introducing this as a new config file, but asking in general - I see we carry lots of configuration files in SPEC when this file might be better home for them.

script/passenger-recycler.rb
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env ruby
+#
+# Trivial Passenger memory monitor installed via /etc/cron.hourly/. To modify
+# configuration create /etc/foreman/passenger-recycler.rb.conf file.
+#
+
+CONFIG = '/etc/foreman/passenger-recycler.rb.conf'
```
ShimShtein, line 7: Do we have to read the configuration from a file? Maybe we can get away with `ENV`?

ShimShtein:
:+1: Works on my machine.

script/passenger-recycler.rb
lzap: That is indeed possible, do you insist on this? I find config file more clear, you can provide comments with explanations and in this case you would need to create wrapper or system wide variables.

ShimShtein: I know that we are using ENV in [foreman-tasks](https://github.com/theforeman/foreman-tasks/blob/master/deploy/foreman-tasks.sysconfig). Plus we can set ENV inline for use cases like cron:

```
GRACEFUL_SHUTDOWN_SLEEP = 1 passenger-recycler.rb
```
I don't have strong opinion here though, it's a question for people with actual administration experience.

domcleal, line 7: This should be a YAML file for consistency, we don't use Ruby source files for configuration.

script/passenger-recycler.rb
```diff
@@ -0,0 +1,59 @@
...
+SEND_STATUS ||= true
+exit 0 unless ENABLED
+
+def running?(pid)
+  return Process.getpgid(pid) != -1
+rescue Errno::ESRCH
+  return false
+end
+
+require 'phusion_passenger'
```
domcleal, line 22: This assumes the Passenger installation's been added to the load path, how will it load from non-rubygems packages, e.g. /usr/share/passenger? Should it also set the path to locations.ini?

script/passenger-recycler.rb
```diff
@@ -0,0 +1,59 @@
...
+require 'phusion_passenger/platform_info'
```
domcleal, line 23: This fails as some environment isn't set up correctly, so the full path to the Passenger installation is unknown:

<pre>
/usr/share/rubygems/rubygems/core_ext/kernel_require.rb:55:in `require': cannot load such file -- /phusion_passenger/utils/tmpio (LoadError)
        from /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:55:in `require'
        from /usr/share/passenger/phusion_passenger.rb:235:in `require_passenger_lib'
        from /usr/share/passenger/phusion_passenger/platform_info.rb:26:in `<top (required)>'
        from /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:55:in `require'
        from /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:55:in `require'
        from passenger-recycler.rb:23:in `<main>'
</pre>

with Passenger 5.0.26. Possibly because `PhusionPassenger.locate_directories` isn't called during initialisation.

Should it also use `require_passenger_lib`?

Rakefile.dist
```diff
@@ -43,18 +44,18 @@ task :build => [
...
   TAILDIR = ENV['TAILDIR'] || "script/foreman-tail.d"
   mkdir_p "#{DATAROOTDIR}/foreman/#{TAILDIR}"
-  cp Dir.glob("#{TAILDIR}/common/*"), "#{DATAROOTDIR}/foreman/#{TAILDIR}/", :verbose => true
+  cp_p Dir.glob("#{TAILDIR}/common/*"), "#{DATAROOTDIR}/foreman/#{TAILDIR}/"
   if is_redhat
-    cp Dir.glob("#{TAILDIR}/redhat/*"), "#{DATAROOTDIR}/foreman/#{TAILDIR}/", :verbose => true
+    cp_p Dir.glob("#{TAILDIR}/redhat/*"), "#{DATAROOTDIR}/foreman/#{TAILDIR}/"
   elsif is_debian
-    cp Dir.glob("#{TAILDIR}/debian/*"), "#{DATAROOTDIR}/foreman/#{TAILDIR}/", :verbose => true
+    cp_p Dir.glob("#{TAILDIR}/debian/*"), "#{DATAROOTDIR}/foreman/#{TAILDIR}/"
```
domcleal, line 58: Why are all of these being modified to preserve ownership etc?

Rakefile.dist
domcleal, line 76: The config file location here is defined by SYSCONFDIR, but the CONFIG path in the script is hardcoded, it should be set at build time.

Rakefile.dist
```diff
@@ -65,6 +66,14 @@ task :install => :build do |t|
 
   # ssh
   mkdir_p "#{DATAROOTDIR}/foreman/.ssh"
+
+  # cron scripts
+  mkdir_p "#{SYSCONFDIR}/cron.hourly"
+  cp_p "script/passenger-recycler.rb", "#{SYSCONFDIR}/cron.hourly/passenger-recycler"
```
domcleal, line 72: Missing some kind of `foreman` prefix, or preferably leave this out and allow packagers to install it in their Foreman-related crontabs, which will be more obvious. This isn't part of Passenger or the default system.

script/passenger-recycler.rb
```diff
@@ -0,0 +1,59 @@
...
+      puts "Process #{pid} still terminating, moving on..."
+    else
+      puts "Process successfully #{pid} terminated"
+    end
+    if SEND_STATUS
+      puts status_ps
+      puts status_all
+      puts status_backtraces
+    end
+    exit 1
```
domcleal, line 56: Why does it exit here after terminating one process?

This is within a loop, what's the point of the loop? It also prints things like "moving on..." above, and would exit with a status 1 whether a process was terminated successfully or not.

domcleal (changes requested):
I'm unsure this is really appropriate to keep in Foreman, it isn't specific to Foreman in any way. Can you also submit it to the Passenger project please?

lzap:
What you mean by it's not specific to Foreman, Passenger is our infrastructure and we need to monitor it. Until now, we did not care. The company behind Passenger has strong business plan based on this feature missing from the community version, I have no plans to fight this upstream. It's also currently heavily focused around Foreman use case, this doesn't aim to be generic tool.

I propose this as a poors man solution to the problem of recycling processes, we see this very often with Katello plugin installed, servers are swapping hard. I believe this would be beneficial to Foreman project. I also want it to keep simple.

script/passenger-recycler.rb
lzap: The extension is pretty clear, I don't see a reason to do YAML here except consistency, which is not enough strong argument for me. YAML is inappropriate format in this case.

script/passenger-recycler.rb
lzap: Thanks for testing with this, I will take a look into it.

script/passenger-recycler.rb
lzap: My idea behind this is only to kill one process at maximum, since I don't want to code rolling restart here - I want to keep it simple. If there's load there are multiple processes so the script kills just one per hour by default.

I could however count processes of each application if that's possible and than always make sure I won't kill them all (leave at least one alive).

Rakefile.dist
lzap: This should not belong here, I thought it's incorrect to not to preserve timestamps. I mean, it is, but I assume you want this out of this patch.

Rakefile.dist
lzap: Good point, definitely needs foreman prefix.

Rakefile.dist
lzap: Good observation, but I don't think it is worth sed-ing it here. But if you insist...

script/passenger-recycler.rb
dLobatog: I also think it makes sense to keep the configuration file with the same format as everything else (YAML). I know it's easy from the developer POV to just run `load(CONFIG)` but it complicates things a bit from the user POV. Also YAML files can be easily modified with our installer, unlike Ruby source..

dLobatog: I would vote for a config file.. foreman_ansible and the Foreman inventory for Ansible can be configured via environment variables. The result is that it leads to confusion to users, "where do I persist this?", "which user should have the environment variable?", "should these variables be global?", etc...

dLobatog (changes requested):
I'm of the same opinion of https://github.com/theforeman/foreman/pull/4540#pullrequestreview-40026587 This looks necessary (given that we can't limit RSS any other way...), but it's a generic tool. Given you said the Passenger project would not approve this in the community version, I'd rather have it in "lzap/passenger-recycler" or "theforeman/passenger-recycler", and Foreman make use of it instead of packaging it together. 

Notice this way you'd be able to push updates and deliver new versions of the recycler if needed, without having to release any new Foreman versions. If it helps convincing you, having https://github.com/theforeman/ldap_fluff as a separate gem has allowed me to deliver fixes to LDAP stuff much more timely and with less hassle than if users had to wait for a patch version.

script/passenger-recycler.rb
lzap: Yeah it's not that much work, I will do it. I'd do it anyway by the way, I will get back to this patch this week :-)

lzap:
Sigh, I was hoping to have smallest and simplest script possible, these are all good points. I will return back to this soon.

iNecas:
@lzap Given the suggestions on this code not being suitable for this repository, are you going to create a new project with it? I wonder if the foreman-maintain could be the place to host this functionality.

lzap:
Thanks all for review, I incorporated all the comments and I am closing it here, this will go into foreman-maintain helper tool git repo: https://github.com/theforeman/foreman_maintain/pull/78

PR #4541, opened 2017-05-19
fixes #19528 - rewrite FiltersHelperOverrides to support autoload

domcleal:
Replaces class-level filter search overrides storage with a query
against Foreman::Plugin, allowing the class to be reloaded and tested
safely. Removes another 'registry' class.


tbrisker:
Thanks @domcleal !

PR #4542, opened 2017-05-19
fixes #18500 - replace quiet_assets with sprockets-rails flag

domcleal:
quiet_assets was merged into sprockets-rails 3.1+ and is no longer
maintained for Rails 5, so it can be removed. Min version hasn't been
increased as it only affects development log verbosity and isn't a hard
dependency. Older versions used in packages will work fine.

The logging gem deliberately doesn't implement the #silence method, so
this is wrapped to fully match the Rails API and make the feature work.

---

Change for quiet_assets is a small one but relies on #4288, so that's also included in this PR (as it's unreviewed). I can rebase this if the other PR is merged.


dLobatog:
Closed as it was merged already in #4646

PR #4544, opened 2017-05-20
Fixes #17377 - update fog(-core) to current versions

mmoll:
With these versions, the problem with digitalocean mentioned in #4029 should not occour.


Gemfile
```diff
@@ -23,7 +23,7 @@ gem 'gettext_i18n_rails', '~> 1.0', '< 1.8.0'
 gem 'rails-i18n', '~> 4.0.0'
 gem 'turbolinks', '~> 2.5'
 gem 'logging', '>= 1.8.0', '< 3.0.0'
-gem 'fog-core', '1.42.0'
+gem 'fog-core', '1.44.2'
```
tbrisker, line 26: Would it maybe make sense to not lock it on a specific version but maybe just a minor or even major?

mmoll: I think it's better to lock to "known good" versions for now and change it, once we got rid of the monolithic "fog" gem.

tbrisker:
Thanks @mmoll !

PR #4545, opened 2017-05-22
Fixes #19612 - CVE-2017-7505 don't expose admin to taxed users

ares:


tbrisker:
@ares Didn't dive much into it but looks like failing tests are related.

ares:
Sadly, we relied on this behavior in some tests and code (notification recipients, authentication when login is disabled). All should be fixed now.

app/models/notification.rb
```diff
@@ -62,8 +62,8 @@ def set_default_initiator
   end
 
   def set_notification_recipients
-    subscribers = subscriber_ids
-    notification_recipients.build subscribers.map{|id| { :user_id => id}}
+    subscribers = User.unscoped.where(:id => subscriber_ids)
+    notification_recipients.build subscribers.map{|user| { :user => user}}
```
ares, line 66: for reviewer (and myself): If we used `:user_id`, rails later tried to load the association when it validates presence of `user` but it respects the default scope defined by taxonomix, therefore it couldn't find the user and silently failed to save the built associated notification recipient. Searching for users like this should be more efficient, since it loads all of them in one query.

tbrisker:
ok to test

tbrisker:
[test foreman]
[test katello]
[test upgrade]
for some reason tests didn't fire...

tbrisker (approved):
LGTM, let's see that tests pass now.

tbrisker:
I think https://github.com/Katello/katello/blob/master/test/katello_test_helper.rb#L91 needs to also be fixed for katello tests to run

dLobatog:
I was able to reproduce and test by:

1. Having a couple of admin users NOT in any organization, and a test user (not admin) with a role which has the "edit_users" permission. 
2. The test user is in the organization "forklift-devel".
3. Without the patch, the test user can see and change properties of the admin users.
4. With the patch, the test user cannot do what I could on step 3. 

If I remove the test user from all organizations, it still works fine and the test user cannot see the admin users.
If I add an admin user to the test user organization 'forklift-devel' - the admin user is now visible.

@ares 👍 from me, if you can rebase & make it green I'm happy to merge for inclusion on 1.15.1

dLobatog:
@iNecas Do katello tests run without https://github.com/seattlerb/minitest/pull/696 ?

iNecas:
I don't see the relation of katello test failures to https://github.com/seattlerb/minitest/pull/696

```
/usr/local/rvm/gems/ruby-2.2.5@test_develop_pr_katello-1/gems/activerecord-4.2.8/lib/active_record/relation/finder_methods.rb:324:in `raise_record_not_found_exception!': Couldn't find User with 'id'=135138680 [WHERE (1=0)] (ActiveRecord::RecordNotFound)
```

ares:
Thanks @dLobatog, katello tests should be fixed by https://github.com/Katello/katello/pull/6800 I triggered tests of both at http://ci.theforeman.org/job/test_katello_pull_request/11873/

tbrisker:
@ares looks like something didn't work on the manual test run, care to take a look? I don't want to merge this only to discover we broke katello somewhere. Other plugins may also need fixes.

ares:
yeah, the jenkins params are not well documented, so I triggered it with wrong values, let's see if [this](http://ci.theforeman.org/job/test_katello_pull_request/11879/) will be better

dLobatog:
@ares I don't think that job got to run 😞 let's wait for https://github.com/Katello/katello/pull/6800 to be merged, it's a tiny PR.

ares:
Katello PR was merged but this https://github.com/theforeman/foreman-tasks/pull/252 should go in too before this is merged. Otherwise one katello task test will be broken. @iNecas confirmed they will merge and release new version soon (I expect it tomorrow)

ares:
Both required PRs are merged, foreman-tasks are released. I think everything would be green now but Foreman tests seems to be broken.

dLobatog:
[test katello]

ares:
I went through all Katello failures, all are instances of `NoMethodError: undefined method `split' for nil:NilClass` which other plugins hit recently, it was caused by using assert_equal while on of the value was nil. I don't think it's related, other PRs fail with the same.

dLobatog:
@ares No worries, I'm keeping track of these. Foreman tests seem fine. Thank you!

PR #4548, opened 2017-05-22
refs #19534 - check EncryptValue class variables are defined

domcleal:
bed4594 added class variables to record logging state, but failed to
check it was defined before accessing them, causing exceptions when
decryption failed. Tests added for this new behaviour.


lzap:
Weird, I tested this in a `irb` session but incorrectly apparently, thanks.

PR #4549, opened 2017-05-22
fixes #19340 - replace all new_host calls with new_vm

timogoebel:
new_host was replaced by new_vm in 637da2f28accc1378f0326d97adcbb2efc5d9808. This cleans up leftovers of new_host.

### DIFF
--- a/app/views/compute_resources_vms/form/vmware/_volume.html.erb
+++ b/app/views/compute_resources_vms/form/vmware/_volume.html.erb
@@ -11,7 +11,7 @@
 <% else %>
   <%= text_f f, :datastore, :class => "span5", :label => _("Data store"), :label_size => "col-md-2", :disabled => true %>
 <% end %>
-<%= text_f f, :name, :class => "col-md-2", :label => _("Name"), :label_size => "col-md-2", :disabled => !new_host, :readonly => true %>
+<%= text_f f, :name, :class => "col-md-2", :label => _("Name"), :label_size => "col-md-2", :readonly => true %>
 <%= text_f f, :size_gb,
            :class       => "col-md-2",
            :label => _("Size (GB)"), :label_size => "col-md-2" %>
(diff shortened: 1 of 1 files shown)


app/views/compute_resources_vms/form/vmware/_volume.html.erb
```diff
@@ -11,7 +11,7 @@
 <% else %>
   <%= text_f f, :datastore, :class => "span5", :label => _("Data store"), :label_size => "col-md-2", :disabled => true %>
 <% end %>
-<%= text_f f, :name, :class => "col-md-2", :label => _("Name"), :label_size => "col-md-2", :disabled => !new_host, :readonly => true %>
+<%= text_f f, :name, :class => "col-md-2", :label => _("Name"), :label_size => "col-md-2", :readonly => true %>
```
tbrisker, line 14: Just to verify - this is no longer disabled on purpose?

timogoebel: `:readonly => true`
should always disable it if i'm not mistaken.

tbrisker: That only disables editing, but the value still gets sent to the server in this case. if `:disabled => true` then the value isn't sent to the server at all. This might still be ok in some cases, just not sure if this is one of them.

timogoebel: I'll double check, but it should be fine. The name is ignored by VMWare anyway as you are not allowed to change a volume's name.

dLobatog (approved):
👍 to merge from me, please make sure submitting this to VMWare won't change anything (or restore the `:disabled` option). Thanks @timogoebel & @tbrisker

timogoebel:
@tbrisker / @dLobatog : I have double checked. Both vm creation and edition still work with this patch. Sorry for the delay.

dLobatog:
Thanks @timogoebel @tbrisker !

PR #4550, opened 2017-05-22
Fixes #19031 - move to patternfly pagination style

lizagilman:
The new renderer renders pagination links to match patternfly pagination style. 

http://www.patternfly.org/pattern-library/navigation/pagination/

image

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -342,6 +342,10 @@ table {
   vertical-align: bottom;
 }
 
+#pagination {
+  margin-top: -6px;
+}
+
 #interfaceModal {
   .modal-dialog {
     min-width: 1000px;
@@ -655,3 +659,8 @@ code.transparent {
     }
   }
 }
+
+.per-page {
+  margin-right: 5px;
+}
+
--- a/app/controllers/architectures_controller.rb
+++ b/app/controllers/architectures_controller.rb
@@ -5,8 +5,7 @@ class ArchitecturesController < ApplicationController
   before_action :find_resource, :only => [:edit, :update, :destroy]
 
   def index
-    base = resource_base.includes(:operatingsystems).search_for(params[:search], :order => params[:order])
-    @architectures = base.paginate(:page => params[:page])
+    @architectures = resource_base_search_and_page.includes(:operatingsystems)
   end
 
   def new
--- a/app/controllers/fact_values_controller.rb
+++ b/app/controllers/fact_values_controller.rb
@@ -5,7 +5,7 @@ class FactValuesController < ApplicationController
   before_action :setup_search_options, :only => :index
 
   def index
-    values = resource_base.my_facts.search_for(params[:search], :order => params[:order])
+    values = resource_base_search_and_page.my_facts
 
     conds = (original_search_parameter || '').split(/AND|OR/i)
     conds = conds.flatten.reject { |c| c.include?('host') }
@@ -23,7 +23,7 @@ class FactValuesController < ApplicationController
 
     respond_to do |format|
       format.html do
-        @fact_values = @fact_values.preload(related_tables).paginate :page => params[:page]
+        @fact_values = @fact_values.preload(related_tables).paginate(:page => params[:page], :per_page => params[:per_page])
         render :index
       end
       format.csv do
--- a/app/helpers/layout_helper.rb
+++ b/app/helpers/layout_helper.rb
@@ -53,41 +53,6 @@ module LayoutHelper
                                                                            :tabindex => '-1' }.deep_merge(options))
   end
 
-  def will_paginate(collection = nil, options = {})
-    options[:class] = "col-md-7"
-    options[:renderer] ||= "WillPaginate::ActionView::BootstrapLinkRenderer"
-    options[:inner_window] ||= 2
-    options[:outer_window] ||= 0
-    options[:previous_label] ||= _('&laquo;')
-    options[:next_label] ||= _('&raquo;')
-    super collection, options
-  end
-
-  def page_entries_info(collection, options = {})
-    html = if collection.total_entries == 0
-             _("No entries found")
-           else
-             if collection.total_pages < 2
-               n_("Displaying <b>%{count}</b> entry", "Displaying <b>all %{count}</b> entries", collection.total_entries) % {:count => collection.total_entries}
-             else
-               _("Displaying entries <b>%{from} - %{to}</b> of <b>%{count}</b> in total") %
-                   { :from => collection.offset + 1, :to => collection.offset + collection.length, :count => collection.total_entries }
-             end
-           end.html_safe
-    html += options[:more].html_safe if options[:more]
-    content_tag(:div, :class => "col-md-5 hidden-xs") do
-      content_tag(:div, html, :class => "pull-left pull-bottom darkgray pagination")
-    end
-  end
-
-  def will_paginate_with_info(collection = nil, options = {})
-    content_tag(:div, :id => "pagination", :class => "row",
-                :data => ({'count' => collection.total_entries, 'per-page' => per_page(collection)})) do
-      page_entries_info(collection, options) +
-        will_paginate(collection, options)
-    end
-  end
-
   def icon_text(i, text = "", opts = {})
     opts[:kind] ||= "glyphicon"
     (content_tag(:span,"", :class=>"#{opts[:kind] + ' ' + opts[:kind]}-#{i} #{opts[:class]}", :title => opts[:title]) + " " + text).html_safe
--- a/app/views/common/_pagination.html.erb
+++ b/app/views/common/_pagination.html.erb
@@ -1,0 +1,26 @@
+<form
+  class='content-view-pf-pagination table-view-pf-pagination paginate' id='pagination'
+  data-count=<%= collection.total_entries %> data-per-page=<%= per_page(collection) %> >
+  <div class='form-group'>
+    <%= select_tag('per_page',
+                   options_for_select(per_page_options,
+                                      params[:per_page] || Setting[:entries_per_page]),
+                   label: _('per page'),
+                   onChange: 'tfm.tools.updateTable(this)',
+                   class: 'pagination-pf-pagesize without_select2 per-page') %>
+    <span><%= _('per page') %></span>
+  </div>
+
+  <div class='form-group'>
+    <span>
+      <span class='pagination-pf-items-current'>
+        <%= collection.offset + 1 %>-<%= collection.offset + collection.length %>
+      </span>
+      <%= _('of') %>
+      <span class='pagination-pf-items-total'>
+        <%= collection.total_entries %>
+      </span>
+    </span>
+    <%= will_paginate(collection, options) %>
+  </div>
+</form>
--- a/config/initializers/will_paginate.rb
+++ b/config/initializers/will_paginate.rb
@@ -3,23 +3,43 @@ require 'will_paginate/array'
 
 module WillPaginate
   module ActionView
-    class BootstrapLinkRenderer < LinkRenderer
+    class PatternflyLinkRenderer < LinkRenderer
       protected
 
-      def html_container(html)
-        tag :div, tag(:ul, html, :class=>"pagination pull-right"), container_attributes
+      def container_attributes
+        super.except(:first_label, :last_label)
       end
 
-      def page_number(page)
-        tag :li, link(page, page, :rel => rel_value(page)), :class => ('active' if page == current_page)
+      def pagination
+        [:back, :pages, :forward]
+      end
+
+      def back
+        tag :ul, first_page + previous_page, class: 'pagination pagination-pf-back'
+      end
+
+      def forward
+        tag :ul, next_page + last_page, class: 'pagination pagination-pf-forward'
+      end
+
+      def pages
+        current_page_input = tag(:input, '', class: 'pagination-pf-page', type: 'text', value: current_page, id: 'pagination1-page')
+        current_page_label = tag(:label, _('Current Page'), class: 'sr-only', for: 'pagination1-page')
+        of_total_pages = tag(:span, _('of ') + tag(:span, total_pages, class: 'pagination-pf-pages'))
+        current_page_input + current_page_label + of_total_pages
+      end
+
+      def first_page
+        previous_or_next_page(current_page == 1 ? nil : 1, @options[:first_label], 'first_page')
+      end
+
+      def last_page
+        previous_or_next_page(current_page == total_pages ? nil : total_pages, @options[:last_label], 'last_page')
       end
 
       def previous_or_next_page(page, text, classname)
-        tag :li, link(text, page || '#'), :class => [classname[0..3], classname, ('disabled' unless page)].join(' ')
-      end
-
-      def gap
-        tag :li, link(super, '#'), :class => 'disabled'
+        classname = [classname, ('disabled' unless page)].join(' ')
+        tag :li, link(text, page || '#'), class: classname
       end
     end
   end
--- a/test/controllers/architectures_controller_test.rb
+++ b/test/controllers/architectures_controller_test.rb
@@ -8,6 +8,8 @@ class ArchitecturesControllerTest < ActionController::TestCase
   basic_index_test
   basic_new_test
   basic_edit_test
+  basic_pagination_per_page_test
+  basic_pagination_rendered_test
 
   def test_new_submit_button_id
     get :new, {}, set_session_user
@@ -72,7 +74,7 @@ class ArchitecturesControllerTest < ActionController::TestCase
   def user_with_viewer_rights_should_fail_to_edit_an_architecture
     setup_user
     get :edit, {:id => Architecture.first.id}
-    assert @response.status == '403 Forbidden'
+    assert_response :forbidden
   end
 
   def user_with_viewer_rights_should_succeed_in_viewing_architectures
--- a/test/controllers/bookmarks_controller_test.rb
+++ b/test/controllers/bookmarks_controller_test.rb
@@ -1,6 +1,13 @@
 require 'test_helper'
 
 class BookmarksControllerTest < ActionController::TestCase
+  setup do
+    @factory_options = [{:controller => "users"}]
+  end
+
+  basic_pagination_per_page_test
+  basic_pagination_rendered_test
+
   test "should get index" do
     get :index, {}, set_session_user
     assert_response :success
--- a/test/controllers/hosts_controller_test.rb
+++ b/test/controllers/hosts_controller_test.rb
@@ -3,6 +3,9 @@ require 'test_helper'
 class HostsControllerTest < ActionController::TestCase
   setup :initialize_host
 
+  basic_pagination_rendered_test
+  basic_pagination_per_page_test
+
   def host_attributes(host)
     known_attrs = HostsController.host_params_filter.accessible_attributes(HostsController.parameter_filter_context)
     host.attributes.except('id', 'created_at', 'updated_at').slice(*known_attrs)
--- a/test/controllers/models_controller_test.rb
+++ b/test/controllers/models_controller_test.rb
@@ -1,6 +1,9 @@
 require 'test_helper'
 
 class ModelsControllerTest < ActionController::TestCase
+  basic_pagination_per_page_test
+  basic_pagination_rendered_test
+
   def test_index
     get :index, {}, set_session_user
     assert_template 'index'
--- a/test/controllers/realms_controller_test.rb
+++ b/test/controllers/realms_controller_test.rb
@@ -1,6 +1,9 @@
 require 'test_helper'
 
 class RealmsControllerTest < ActionController::TestCase
+  basic_pagination_per_page_test
+  basic_pagination_rendered_test
+
   def test_index
     get :index, {}, set_session_user
     assert_template 'index'
--- a/test/controllers/shared/basic_rest_response_test.rb
+++ b/test/controllers/shared/basic_rest_response_test.rb
@@ -36,5 +36,98 @@ module BasicRestResponseTest
         end
       end
     end
+
+    def basic_pagination_per_page_test
+      context 'GET #index' do
+        setup do
+          @entries_per_page = Setting[:entries_per_page] ? Setting[:entries_per_page] : 5
+          FactoryGirl.create_list(get_factory_name, @entries_per_page + 2, *@factory_options)
+        end
+
+        test 'should render correct per_page value' do
+          get :index, {per_page: @entries_per_page + 1}, set_session_user
+          assert_response :success
+          per_page_results = css_select('.pagination-pf-items-current')
+          assert_equal "1-#{@entries_per_page + 1}", per_page_results.first.content.squish
+        end
+
+        test 'should render per page dropdown with correct values' do
+          get :index, {per_page: @entries_per_page + 1}, set_session_user
+          assert_response :success
+          assert_select "[id='per_page']" do
+            assert_select "option[selected='selected'][value='#{@entries_per_page + 1}']"
+            assert_select "option[value='5']"
+            assert_select "option[value='10']"
+            assert_select "option[value='15']"
+            assert_select "option[value='25']"
+            assert_select "option[value='50']"
+          end
+        end
+
+        test 'should display pagination buttons' do
+          get :index, {per_page: @entries_per_page + 1}, set_session_user
+          assert_response :success
+          assert_select "[id=pagination]" do
+            assert_includes @response.body.squish, "pagination pagination-pf-back"
+            assert_includes @response.body.squish, "pagination pagination-pf-forward"
+          end
+        end
+
+        test 'sort links should include per page param' do
+          get :index, {per_page: @entries_per_page + 1}, set_session_user
+          assert_response :success
+          sort_links = css_select('thead a')
+          sort_links.each do |link|
+            assert_includes link['href'], "per_page=#{@entries_per_page + 1}"
+          end
+        end
+      end
+    end
+
+    def basic_pagination_rendered_test
+      context 'GET #index' do
+        setup do
+          @old = Setting[:entries_per_page]
+          FactoryGirl.create(get_factory_name) if @controller.resource_class.count.zero?
+          Setting[:entries_per_page] = @controller.resource_class.count
+        end
+
+        test 'should not render pagination' do
+          get :index, {}, set_session_user
+          assert_response :success
+          refute_includes @response.body, "id=pagination"
+        end
+
+        test 'should render pagination' do
+          FactoryGirl.create(get_factory_name, *@factory_options)
+          get :index, {}, set_session_user
+          assert_response :success
+          assert_select "form[id='pagination']"
+        end
+
+        test 'should not render pagination when no search results' do
+          next if @controller.resource_name == "trend"
+          @request.env['HTTP_REFERER'] = root_url
+          get :index, {search: "name='A98$bcD#67Ef*g"}, set_session_user
+          assert (@response.body.include? "No entries found") ||  @response.body.match(/You are being.*redirected/)
+        end
+
+        teardown do
+          Setting[:entries_per_page] = @old
+        end
+      end
+    end
+  end
+
+  def get_factory_name
+    model = @controller.controller_name.singularize
+    case model
+      when "trend"
+        :trend_os
+      when "subnet"
+        :subnet_ipv4
+      else
+        model.to_sym
+    end
   end
 end
--- a/test/controllers/trends_controller_test.rb
+++ b/test/controllers/trends_controller_test.rb
@@ -1,6 +1,9 @@
 require 'test_helper'
 
 class TrendsControllerTest < ActionController::TestCase
+  basic_pagination_rendered_test
+  basic_pagination_per_page_test
+
   test "should get empty_data page, if no trend counters exist" do
     trend = FactoryGirl.create(:trend_os)
     get :show, { :id => trend.id }, set_session_user
--- a/webpack/assets/javascripts/foreman_tools.js
+++ b/webpack/assets/javascripts/foreman_tools.js
@@ -85,13 +85,17 @@ export function initTypeAheadSelect(input) {
 // handle table updates via turoblinks
 export function updateTable(element) {
   let uri = new URI(window.location.href);
-  let search;
+  let search, perPage;
 
   if (element !== undefined) {
-    search = $(element).find('.autocomplete-input').val();
+    search = $(element).find('.autocomplete-input').val() || $('#search-form').find('.autocomplete-input').val();
     if (search !== undefined) {
       uri.setSearch('search', search.trim());
     }
+  }
+  perPage = $('#per_page').val();
+  if (perPage !== undefined) {
+    uri.setSearch('per_page', perPage);
   }
   /* eslint-disable no-undef */
   Turbolinks.visit(uri.toString());
--- a/app/assets/javascripts/two-pane.js
+++ b/app/assets/javascripts/two-pane.js
@@ -117,7 +117,7 @@ function two_pane_close(){
   $('.table-two-pane tr td').show();
   $('.table-two-pane th').show();
   $("#title_action").show();
-  $('.pagination').show();
+  $('#pagination').show();
   if ($('.two-pane-left').length){
     $('.table-two-pane').unwrap().unwrap();
   }
@@ -131,7 +131,7 @@ function hide_columns(){
   $('.table-two-pane th:nth-child(1)').show();
   $("#title_action").hide();
   $('.two-pane-right').remove();
-  $('.pagination').hide();
+  $('#pagination').hide();
   if ($('.two-pane-left').length == 0){
     $('.table-two-pane').wrap( "<div class='row'><div class='col-md-3 two-pane-left'></div></div>");
   }
--- a/app/controllers/audits_controller.rb
+++ b/app/controllers/audits_controller.rb
@@ -4,7 +4,7 @@ class AuditsController < ApplicationController
   before_action :setup_search_options, :only => :index
 
   def index
-    Audit.unscoped { @audits = resource_base.includes(:user).search_for(params[:search], :order => params[:order]).paginate :page => params[:page] }
+    Audit.unscoped { @audits = resource_base_search_and_page.includes(:user) }
   end
 
   def show
--- a/app/controllers/auth_source_ldaps_controller.rb
+++ b/app/controllers/auth_source_ldaps_controller.rb
@@ -4,7 +4,7 @@ class AuthSourceLdapsController < ApplicationController
   before_action :find_resource, :only => [:edit, :update, :destroy]
 
   def index
-    @auth_source_ldaps = resource_base.all
+    @auth_source_ldaps = resource_base_search_and_page.all
   end
 
   def new
--- a/app/controllers/compute_resources_controller.rb
+++ b/app/controllers/compute_resources_controller.rb
@@ -13,7 +13,7 @@ class ComputeResourcesController < ApplicationController
   end
 
   def index
-    @compute_resources = resource_base.live_descendants.search_for(params[:search], :order => params[:order]).paginate :page => params[:page]
+    @compute_resources = resource_base_search_and_page.live_descendants
   end
 
   def new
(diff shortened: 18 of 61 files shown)


app/helpers/layout_helper.rb
```diff
@@ -54,37 +54,24 @@ def popover(title, msg, options = {})
...
-           end.html_safe
-    html += options[:more].html_safe if options[:more]
-    content_tag(:div, :class => "col-md-5 hidden-xs") do
-      content_tag(:div, html, :class => "pull-left pull-bottom darkgray pagination")
-    end
-  end
-
   def will_paginate_with_info(collection = nil, options = {})
-    content_tag(:div, :id => "pagination", :class => "row",
+    return if Setting[:entries_per_page] > collection.total_entries
```
ohadlevy: could you add a  before / after picture for @rohoover ? also should we show pagination actions when there are no additional pages?

app/views/common/_pagination.html.erb
```diff
@@ -0,0 +1,22 @@
+<div class="form-group">
+  <%= select_tag('per_page',
+                 options_for_select([5, 10, 15, 25, 50], params[:per_page] || Setting[:entries_per_page].to_i),
```
ohadlevy: please include the current ```params[:per_page]``` if its defined and the value of ```Setting[:entries_per_page]``` to the dropdown.

app/views/common/_pagination.html.erb
```diff
@@ -0,0 +1,22 @@
...
+                 label: _('per page'),
+                 onChange: "$(this).parentsUntil('.form').submit()",
```
ohadlevy: please extract it to a function under tfm.tools, this will allow us to add other actions (such as spinner or other events).

config/initializers/will_paginate.rb
```diff
@@ -3,23 +3,42 @@
...
+        tag :ul, first_page+previous_page, :class => "pagination pagination-pf-back"
+      end
+
+      def forward
+        tag :ul, next_page+last_page, :class => "pagination pagination-pf-forward"
+      end
+
+      def pages
+        current_page_input = tag(:input, "", :class => "pagination-pf-page", :type => "text", :value => current_page, :id => "pagination1-page")
+        current_page_label = tag(:label, "Current Page", :class => "sr-only", :for => "pagination1-page")
```
ohadlevy: extract all strings (and below)

ohadlevy (changes requested):
looks good overall thanks @lizagilman 

please refactor tests to take into account the per_page variable as well. I would assume we will need to fix a few controllers (architectures, facts, hosts etc) to accept the per_page too.

app/helpers/layout_helper.rb
lizagilman: before:
before

after:
after

should we display number of total entries?

app/helpers/layout_helper.rb
```diff
@@ -54,37 +54,24 @@ def popover(title, msg, options = {})
...
+    options[:page_links] = false
+    options[:container] = false
+    options[:inner_window] ||= 0
     options[:outer_window] ||= 0
-    options[:previous_label] ||= _('&laquo;')
-    options[:next_label] ||= _('&raquo;')
+    options[:previous_label] ||= icon_text("angle-left", "", :kind => "fa")
+    options[:next_label] ||= icon_text("angle-right", "", :kind => "fa")
+    options[:last_label] ||= icon_text("angle-double-right", "", :kind => "fa")
+    options[:first_label] ||= icon_text("angle-double-left", "", :kind => "fa")
```
tbrisker: instead of having such a long list of defaults, you can do something like:
```ruby
defaults = {renderer: "WillPaginate::ActionView::PatternflyLinkRenderer", page_links: false, ......}
super(collection, defaults.merge(options))
```

app/views/common/_pagination.html.erb
```diff
@@ -0,0 +1,27 @@
...
+    per_page_options = [5, 10, 15, 25, 50]
+    per_page_options << Setting[:entries_per_page].to_i
+    per_page_options << params[:per_page].to_i unless params[:per_page].blank?
+    per_page_options = per_page_options.uniq.sort
+    select_tag('per_page',
+                 options_for_select(per_page_options, params[:per_page] || Setting[:entries_per_page]),
+                 label: _('per page'),
+                 onChange: "$(this).parentsUntil('.form').submit()",
+                 class: "pagination-pf-pagesize without_select2") %>
+  <span><%= _('per page') %></span>
```
tbrisker: Please make this a helper

app/views/common/_pagination.html.erb
```diff
@@ -0,0 +1,27 @@
+<div class="form-group">
+  <%=
+    per_page_options = [5, 10, 15, 25, 50]
+    per_page_options << Setting[:entries_per_page].to_i
+    per_page_options << params[:per_page].to_i unless params[:per_page].blank?
```
tbrisker: note that currently many (most?) ui controllers ignore the `per_page` parameter.

test/controllers/architectures_controller_test.rb
```diff
@@ -80,4 +80,16 @@ def user_with_viewer_rights_should_succeed_in_viewing_architectures
     get :index
     assert_response :success
   end
+
+  def should_render_pagination_when_needed
```
tbrisker: Please add a comment here explaining why this is in the architecture controller

config/initializers/will_paginate.rb
```diff
@@ -3,23 +3,42 @@
...
+        of_total_pages = tag(:span, "of " + tag(:span, total_pages, :class => "pagination-pf-pages"))
+        current_page_input+current_page_label+of_total_pages
```
tbrisker: nitpick: add spaces around `+`

tbrisker (changes requested):

app/views/common/_pagination.html.erb
ohadlevy: actually, most do support it as they use the ```resource_base_search_and_page``` helper. i think some that do not should be fixed (in this pr or another pr before this one) - i find architectures, hosts and facts missing the parameter.

lizagilman:
updated

app/views/common/_pagination.html.erb
lizagilman: @ohadlevy can you take a look - https://github.com/lizagilman/foreman/commit/dd452b9aafcb1bb76d3265ff56a602a89fffc05b ?

app/helpers/layout_helper.rb
```diff
@@ -54,40 +54,35 @@ def popover(title, msg, options = {})
...
-           end.html_safe
-    html += options[:more].html_safe if options[:more]
-    content_tag(:div, :class => "col-md-5 hidden-xs") do
-      content_tag(:div, html, :class => "pull-left pull-bottom darkgray pagination")
-    end
+    defaults = {
+      renderer: "WillPaginate::ActionView::PatternflyLinkRenderer",
+      page_links: false,
+      container: false,
+      inner_window: options[:inner_window] ||= 0,
```
tbrisker: you don't need the ` options[:inner_window] ||= ` part

app/views/common/_pagination.html.erb
```diff
@@ -0,0 +1,24 @@
+<div class="form-group">
+  <%=
```
tbrisker: nitpick - extra white space

app/views/common/_pagination.html.erb
```diff
@@ -0,0 +1,24 @@
...
+
+    select_tag('per_page',
+                 options_for_select(per_page_options || Setting[:entries_per_page]),
```
tbrisker: I don't think the setting is needed here, it will never be used (as `per_page_options` will always return a non-false value)

app/views/common/_pagination.html.erb
```diff
@@ -0,0 +1,24 @@
...
+                 onChange: "tfm.tools.changePerPageNumber()",
+                 class: "pagination-pf-pagesize without_select2") %>
+  <span><%= _('per page') %></span>
+</div>
+
+<div class="form-group">
+  <span>
+    <span class='pagination-pf-items-current'>
+      <%= "%{from}-%{to}" %
+              { :from => collection.offset + 1, :to => collection.offset + collection.length } %>
```
tbrisker: You can get rid of the string interpolation here using something like:
`<%= collection.offset + 1 %>-<%= collection.offset + collection.length %>`

tbrisker (changes requested):
Added some comments inline

app/helpers/layout_helper.rb
lizagilman: @Rohoover

lizagilman:
@tbrisker thanks updated

app/views/common/_pagination.html.erb
lizagilman: updated https://github.com/lizagilman/foreman/commit/dd452b9aafcb1bb76d3265ff56a602a89fffc05b, if there isn't anything missing there, should I add it to this pr?
@ohadlevy

app/helpers/layout_helper.rb
```diff
@@ -54,40 +54,34 @@ def popover(title, msg, options = {})
   end
 
   def will_paginate(collection = nil, options = {})
-    options.merge!(:class=>"col-md-7")
```
ohadlevy: can we move all of the pagination related core into app/helpers/pagination_helper.rb ?

config/initializers/will_paginate.rb
```diff
@@ -3,23 +3,42 @@
...
-      def page_number(page)
-        tag :li, link(page, page, :rel => rel_value(page)), :class => ('active' if page == current_page)
+      def pagination
+        [:back, :pages, :forward]
       end
 
-      def previous_or_next_page(page, text, classname)
-        tag :li, link(text, page || '#'), :class => [classname[0..3], classname, ('disabled' unless page)].join(' ')
+      def back
+        tag :ul, first_page+previous_page, :class => "pagination pagination-pf-back"
```
ohadlevy: nitpick: add whitespace between operators (+) here and below.

config/initializers/will_paginate.rb
```diff
@@ -3,23 +3,42 @@
...
+        current_page_input = tag(:input, "", :class => "pagination-pf-page", :type => "text", :value => current_page, :id => "pagination1-page")
+        current_page_label = tag(:label, _("Current Page"), :class => "sr-only", :for => "pagination1-page")
+        of_total_pages = tag(:span, _("of ") + tag(:span, total_pages, :class => "pagination-pf-pages"))
+        current_page_input + current_page_label + of_total_pages
       end
 
-      def gap
-        tag :li, link(super, '#'), :class => 'disabled'
+      def first_page
+        previous_or_next_page(current_page == 1 ? nil : 1, @options[:first_label], "first_page")
```
ohadlevy: is this a string that is used in the UI? if it is, it should be extracted (and others in this page).

test/controllers/architectures_controller_test.rb
```diff
@@ -80,4 +80,17 @@ def user_with_viewer_rights_should_succeed_in_viewing_architectures
     get :index
     assert_response :success
   end
+
+  # randomly selected architectures controller for the following test
+  def should_render_pagination_when_needed
+    old = Setting[:entries_per_page]
+    Setting[:entries_per_page] = Architecture.count
```
ohadlevy: can you ensure there is at least one ? i see a difference between:
- no entries at all (we should see the patternfly empty state view)
- entries < default pagination size
- entries > default pagination size

test/controllers/architectures_controller_test.rb
```diff
@@ -80,4 +80,17 @@ def user_with_viewer_rights_should_succeed_in_viewing_architectures
...
+    get :index, {}, set_session_user
+    refute_includes @response.body, 'pagination'
+    FactoryGirl.create(:architecture)
+    get :index, {}, set_session_user
+    assert_includes @response.body, 'pagination'
+  ensure
+    Setting[:entries_per_page] = old
+  end
```
ohadlevy: I'm missing tests to the following:
- per_page dropdown exists with the correct values
- selecting per_page redirect to the correct page and has the correct number of entries
- first / next / prev / last links exists
- controller respects the per_page and page params

app/helpers/layout_helper.rb
```diff
@@ -54,40 +54,34 @@ def popover(title, msg, options = {})
...
+      next_label: options[:next_label] ||= icon_text("angle-right", "", :kind => "fa"),
+      last_label: options[:last_label] ||= icon_text("angle-double-right", "", :kind => "fa"),
+      first_label: options[:first_label] ||= icon_text("angle-double-left", "", :kind => "fa")
+    }
+    super(collection, defaults.merge(options))
   end
 
   def will_paginate_with_info(collection = nil, options = {})
-    content_tag(:div, :id => "pagination", :class => "row",
+    return if Setting[:entries_per_page] >= collection.total_entries
```
ohadlevy: @rohoover - if the page entries is smaller the pagination size (e.g. there is no next page) should we display pagination controls - also, currently we use the pagination line to show total entries - if we remove the controls then we need to display that data elsewhere? thanks!

ohadlevy (changes requested):

app/helpers/layout_helper.rb
```diff
@@ -54,40 +54,34 @@ def popover(title, msg, options = {})
...
-    end
+    defaults = {
+      renderer: "WillPaginate::ActionView::PatternflyLinkRenderer",
+      page_links: false,
+      container: false,
+      outer_window: options[:outer_window] ||= 0,
+      previous_label: options[:previous_label] ||= icon_text("angle-left", "", :kind => "fa"),
+      next_label: options[:next_label] ||= icon_text("angle-right", "", :kind => "fa"),
+      last_label: options[:last_label] ||= icon_text("angle-double-right", "", :kind => "fa"),
+      first_label: options[:first_label] ||= icon_text("angle-double-left", "", :kind => "fa")
```
tbrisker: This still needs fixing to remove all the `options[...] ||=` part

app/views/common/_pagination.html.erb
```diff
@@ -0,0 +1,21 @@
+<div class="form-group">
+<%= select_tag('per_page',
```
tbrisker: nitpick: this should be indented, also no need to put each parameter on a separate line.

config/initializers/will_paginate.rb
tbrisker: It's the class name.

config/initializers/will_paginate.rb
```diff
@@ -3,23 +3,42 @@
...
+      end
+
+      def last_page
+        previous_or_next_page(current_page == total_pages ? nil : total_pages, @options[:last_label], "last_page")
+      end
+
+      def previous_or_next_page(page, text, classname)
+        tag :li, link(text, page || '#'), :class => [classname[0..3], classname, ('disabled' unless page)].join(' ')
```
tbrisker: you might want to make the class calculation in a separate line or two using an extra variable to make this a bit more readable

webpack/assets/javascripts/foreman_tools.js
```diff
@@ -80,3 +80,7 @@ export function initTypeAheadSelect(input) {
     width: '400px'
   });
 }
+
+export function changePerPageNumber() {
+ $("#per_page").parentsUntil('.form').submit();
```
tbrisker: I think this should be `.closest('form')`, no?

tbrisker (changes requested):

app/helpers/layout_helper.rb
Rohoover: @ohadlevy 
Continue to show controls but as disabled. This resolves the second issue with line and total entries. Also users have expressly pointed out to me that they do not like appearing and disappearing controls.

lizagilman: @Rohoover 
What should we display for current page and total pages? "1 of 1", current page text box disabled?
Maybe we can display the pagination line without the controls? In this case it looks pretty much the same but without the unnecessary controls.

page entries is smaller the pagination size- 

no_controls_pagination_line

page entries is larger the pagination size - 

pagination_line_with_controls

webpack/assets/javascripts/foreman_tools.js
lizagilman: I agree, but it doesn't seem to work :flushed:

ohadlevy:
@lizagilman i've made a few changes that I think make sense, please have a look at the commit message - thanks https://github.com/theforeman/foreman/commit/d7c66830a83da40d383e97e09f65f1e58b03a70b

lizagilman:
@ohadlevy  ok :+1: 

please also take a look at -https://github.com/lizagilman/foreman/commit/dd452b9aafcb1bb76d3265ff56a602a89fffc05b (if you haven't yet)

lizagilman:
updated 
WIP regarding the following (no need to review again yet) :
- selecting per_page redirect to the correct page and has the correct number of entries
- controller respects the per_page and page params

app/views/common/_pagination.html.erb
```diff
@@ -0,0 +1,28 @@
+<form
+class='content-view-pf-pagination table-view-pf-pagination paginate'
```
ohadlevy: nitpick: indentation.

lizagilman:
updated! @tbrisker

ohadlevy:
[test foreman]

app/controllers/fact_values_controller.rb
```diff
@@ -5,7 +5,8 @@ class FactValuesController < ApplicationController
   before_action :setup_search_options, :only => :index
 
   def index
-    values = resource_base.my_facts.search_for(params[:search], :order => params[:order])
+    #values = resource_base.my_facts.search_for(params[:search], :order => params[:order])
+    values = resource_base_search_and_page.my_facts
```
ohadlevy: remove commented out line.

app/controllers/fact_values_controller.rb
```diff
@@ -23,7 +24,8 @@ def index
 
     respond_to do |format|
       format.html do
-        @fact_values = @fact_values.preload(related_tables).paginate :page => params[:page]
+        @fact_values = @fact_values.preload(related_tables).paginate(:page => params[:page], :per_page => params[:per_page])
+
```
ohadlevy: please remove empty line.

ohadlevy (changes requested):
@lizagilman thanks for this PR! its nearly done :-)

I only request to add one more test, which checks that the sorting keeps the per_page value, for example if I click on sort by name, it should be a link that actually has the per_page= in there.

app/views/common/_pagination.html.erb
```diff
@@ -0,0 +1,27 @@
+<form
+  class='content-view-pf-pagination table-view-pf-pagination paginate' id='pagination'
+  data-count=<%= collection.total_entries %>
+      data-per-page=<%= per_page(collection) %> >
```
tbrisker: nitpick: please remove the extra line breaks here, kind of hard to see where the tag starts and where it ends. Even if you have to break a tag to multiple lines please make sure they are all aligned.

lizagilman:
updated

ohadlevy (approved):

ohadlevy:
@tbrisker mind re-reviewing ? thanks.

app/controllers/architectures_controller.rb
```diff
@@ -5,8 +5,7 @@ class ArchitecturesController < ApplicationController
   before_action :find_resource, :only => [:edit, :update, :destroy]
 
   def index
-    base = resource_base.includes(:operatingsystems).search_for(params[:search], :order => params[:order])
-    @architectures = base.paginate(:page => params[:page])
+    @architectures = resource_base_search_and_page(:operatingsystems)
```
tbrisker: Here and in other places you've replaced `.includes` with `.eager_load` (which is what happens inside `resource_base_search_and_page`), is this on purpose? this may have negative performance effect as `eager_load` will create a `LEFT OUTER JOIN` while includes may either do `eager_load` or `preload`. (see this [post](http://blog.scoutapp.com/articles/2017/01/24/activerecord-includes-vs-joins-vs-preload-vs-eager_load-when-and-where) for an explanation of joins/includes/eager_load/preload)
If this isn't on purpose you could  use `resource_base_search_and_page.includes(:operatingsystems)` to preserve the current implementation.

config/initializers/will_paginate.rb
```diff
@@ -3,23 +3,43 @@
...
+      def first_page
+        previous_or_next_page(current_page == 1 ? nil : 1, @options[:first_label], 'first_page')
+      end
+
+      def last_page
+        previous_or_next_page(current_page == total_pages ? nil : total_pages, @options[:last_label], 'last_page')
+      end
+
+      def previous_or_next_page(page, text, classname)
+        classname = [classname[0..3], classname, ('disabled' unless page)].join(' ')
```
tbrisker: Why do we need `classname[0..3]`?

test/controllers/hosts_controller_test.rb
```diff
@@ -1,6 +1,10 @@
 require 'test_helper'
 
 class HostsControllerTest < ActionController::TestCase
+  setup do
+    @factory_options = :with_hostgroup
```
tbrisker: are you sure this is needed?

tbrisker: also, looks like this controller doesn't have the pagination tests

test/controllers/hostgroups_controller_test.rb
```diff
@@ -1,6 +1,10 @@
 require 'test_helper'
 
 class HostgroupsControllerTest < ActionController::TestCase
+  setup do
+    @factory_options = {:environment => @environment}
```
tbrisker: Is this needed? also looks like there is no pagination test on this controller

test/controllers/models_controller_test.rb
```diff
@@ -1,6 +1,10 @@
 require 'test_helper'
 
 class ModelsControllerTest < ActionController::TestCase
+  basic_pagination_per_page_test
+  basic_pagination_rendered_test
+  basic_pagination_rendered_test
```
tbrisker: no need to run the same test twice

test/controllers/realms_controller_test.rb
```diff
@@ -24,13 +27,13 @@ def test_create_valid
   end
 
   def test_edit
-    get :edit, {:id => Realm.first}, set_session_user
+    get :edit, {:id => Realm.unscoped.first}, set_session_user
```
tbrisker: Is this still needed? or is it a leftover from a previous version?

test/controllers/shared/basic_rest_response_test.rb
```diff
@@ -36,5 +36,108 @@ def basic_edit_test(object_found = nil)
...
+        model = model.last
+      else
+        model = model.first
+      end
+    end
+    model.underscore.to_sym
+  end
+
+  def create_model_instance(factory_name)
+    factory_name = :subnet_ipv4 if factory_name == :subnet
```
tbrisker: Would it be easier to return `:subnet_ipv4` in the `get_factory_name` helper for any `Subnet` model, or just use the generic `:subnet` factory?

test/controllers/shared/basic_rest_response_test.rb
```diff
@@ -36,5 +36,108 @@ def basic_edit_test(object_found = nil)
...
+    end
+  end
+
+  def get_factory_name(model)
+    model = model.to_s
+    return :trend_os if model == "Trend"
+    if model.include? "::"
+      model = model.split("::")
+      if model.first != "Subnet"
+        model = model.last
```
tbrisker: Is `Subnet::Ipv?` the only case where we want the parent class name? I think at least for `Host::Managed` we'd also want `:host` and not `:managed`

test/controllers/shared/basic_rest_response_test.rb
```diff
@@ -36,5 +36,108 @@ def basic_edit_test(object_found = nil)
...
+          sort_links.each do |link|
+            assert_includes link['href'], "per_page=#{entries_per_page + 1}"
+          end
+        end
+      end
+    end
+
+    def basic_pagination_rendered_test
+      context 'GET #index' do
+        old = Setting[:entries_per_page]
```
tbrisker: This should go inside the setup block, and use a local variable (`@old`) so it is accessible in the teardown block.

test/controllers/shared/basic_rest_response_test.rb
```diff
@@ -36,5 +36,108 @@ def basic_edit_test(object_found = nil)
...
+          create_model_instance(factory_name)
+          get :index, {}, set_session_user
+          assert_response :success
+          assert_select "form[id='pagination']"
+        end
+
+        test 'should not render pagination when no search results' do
+          @request.env['HTTP_REFERER'] = root_url
+          get :index, {search: "name='A98$bcD#67Ef*g"}, set_session_user
+          assert (@response.body.include? "No entries found") ||  @response.body.match(/You are being.*redirected/)
```
tbrisker: I don't think redirect is a valid response when there are no results.

test/controllers/trends_controller_test.rb
```diff
@@ -1,6 +1,8 @@
 require 'test_helper'
 
 class TrendsControllerTest < ActionController::TestCase
+  basic_pagination_per_page_test
```
tbrisker: Missing the rendered test

app/assets/stylesheets/application.scss
```diff
@@ -257,6 +257,7 @@ table {
 
 .table-fixed {
   table-layout: fixed;
+  margin: 0px !important;
```
tbrisker: `!important` is often a sign that somewhere else a more specific selector has different style set; if possible, it should be overridden at the more specific selector level where it is needed rather then forced with `!important` which will make it almost impossible to override if needed.

tbrisker (changes requested):
See comment inline, mostly very small issues. Also, it looks like the layout on several pages isn't correct, for example on hostgroups:
image
notice the gap between the table and the pagination, I noticed this on several pages.

Another issue is with the layout on pages that use two-pane view, when opening the two-pane it pushes the pagination to the bottom of the opened pane:
image

And one last thing, once this is merged, some plugins may need fixes to work with this as well, for example tasks seems to not use the pagination helpers correctly leading to this:
image

lizagilman:
@tbrisker working on your comments will push ASAP thanks

app/assets/stylesheets/application.scss
lizagilman: the more specific also has `!important`

test/controllers/hosts_controller_test.rb
lizagilman: needed

test/controllers/realms_controller_test.rb
lizagilman: here this is still needed

lizagilman:
updated

lizagilman:
[test katello]

ohadlevy:
@lizagilman can you please rebase your pr? thanks!

lizagilman:
@ohadlevy rebased

lizagilman:
@tbrisker approve changes? :pray:

app/assets/stylesheets/application.scss
```diff
@@ -342,6 +343,7 @@ table {
   vertical-align: bottom;
 }
 
+
```
tbrisker: nitpick: extra line

test/controllers/architectures_controller_test.rb
```diff
@@ -72,7 +74,7 @@ def setup_user
   def user_with_viewer_rights_should_fail_to_edit_an_architecture
     setup_user
     get :edit, {:id => Architecture.first.id}
-    assert @response.status == '403 Forbidden'
+    assert_equal '403 Forbidden', @response.status
```
tbrisker: you can use the `assert_response :forbidden` helper here

test/controllers/shared/basic_rest_response_test.rb
```diff
@@ -36,5 +36,121 @@ def basic_edit_test(object_found = nil)
...
+      FactoryGirl.create(factory_name, :as_smart_class_param, :puppetclass => puppetclasses(:one), :override => true, :default_value => 'test')
+      return
+    end
+    FactoryGirl.create(factory_name, @factory_options)
+  end
+
+  def create_model_list(factory_name, num)
+    case factory_name
+      when :subnet
+        factory_name = :subnet_ipv4
```
tbrisker: isn't this already handled by the get_factory_name function?

test/controllers/shared/basic_rest_response_test.rb
```diff
@@ -36,5 +36,121 @@ def basic_edit_test(object_found = nil)
...
+          model = model.last
+      end
+    end
+    return :subnet_ipv4 if model.underscore.to_sym == :subnet
+    model.underscore.to_sym
+  end
+
+  def create_model_instance(factory_name)
+    if factory_name == :puppetclass_lookup_key
+      FactoryGirl.create(factory_name, :as_smart_class_param, :puppetclass => puppetclasses(:one), :override => true, :default_value => 'test')
```
tbrisker: Why not just set the `@factory_options` where needed and get rid of this condition and the same one below?

test/controllers/shared/basic_rest_response_test.rb
```diff
@@ -36,5 +36,121 @@ def basic_edit_test(object_found = nil)
...
+    if model.include? "::"
+      model = model.split("::")
+      case model.first
+        when "Subnet", "Host"
+          model = model.first
+        else
+          model = model.last
+      end
+    end
+    return :subnet_ipv4 if model.underscore.to_sym == :subnet
```
tbrisker: you can just check if `model == 'Subnet'`, no need to convert it here to symbol

test/controllers/shared/basic_rest_response_test.rb
```diff
@@ -36,5 +36,121 @@ def basic_edit_test(object_found = nil)
...
+  def get_factory_name(model)
+    model = model.to_s
+    return :trend_os if model == "Trend"
+    if model.include? "::"
+      model = model.split("::")
+      case model.first
+        when "Subnet", "Host"
+          model = model.first
+        else
+          model = model.last
```
tbrisker: I'm just curious, what other models are there that need .last?

test/controllers/shared/basic_rest_response_test.rb
```diff
@@ -36,5 +36,121 @@ def basic_edit_test(object_found = nil)
         end
       end
     end
+
+    def basic_pagination_per_page_test
+      context 'GET #index' do
+        entries_per_page = Setting[:entries_per_page] ? Setting[:entries_per_page] : 5
```
tbrisker: This should be in the setup block, if you need to access it in the tests make it an instance variable.

app/assets/stylesheets/application.scss
```diff
@@ -273,6 +273,7 @@ table {
 
 .table-fixed {
   table-layout: fixed;
+  margin: 0 !important;
```
tbrisker: This shouldn't be part of `.table-fixed` - this class is for forcing column widths which are required for clipping, adding a margin here is unexpected behaviour and might have side effects on other pages that use this class.

tbrisker (changes requested):
A few more minor changes inline. Additionally, looks like two-pane pages still don't work correctly:
image
Notice how the pagination is pushed below the two-pane form instead of remaining attached to the table.

ohadlevy:
[test foreman]

lizagilman:
@tbrisker thanks, 
what are you suggesting with two-pane pages?

ohadlevy:
@lizagilman update https://github.com/theforeman/foreman/blob/develop/app/assets/javascripts/two-pane.js#L134 ?

tbrisker:
@lizagilman the pagination on two pane should stay attached to the bottom of the table and not move when two-pane is open. also, please rebase on latest develop to solve the failing tests.

test/controllers/shared/basic_rest_response_test.rb
lizagilman: All models except this two

test/controllers/shared/basic_rest_response_test.rb
lizagilman: I implemented it in a little different way (with the asterisk), is this better now?

test/controllers/shared/basic_rest_response_test.rb
```diff
@@ -36,5 +36,110 @@ def basic_edit_test(object_found = nil)
...
+          model = model.first
+        else
+          model = model.last
+      end
+    end
+    return :subnet_ipv4 if model == "Subnet"
+    model.underscore.to_sym
+  end
+
+  def create_model_instance(factory_name, *factory_options)
```
tbrisker: I think these two functions are now redundant - all they do is pass their parameters untouched to another function, so we can get rid of them entirely and just use those functions directly.

app/assets/stylesheets/application.scss
lizagilman: We can do `#pagination {margin-top: -6px;}` ,
do you this is right? with the - sign?

lizagilman:
@tbrisker waiting for your feedback regarding -https://github.com/theforeman/foreman/pull/4550/files#diff-f859e2848b07324f808ce1a5ccd9fcba
and - 
outdated comment on basic_rest_response_test.rb line 140 (I replied on your there..

regarding two-pain view I used Ohad's suggestion, hide/ show it

app/assets/stylesheets/application.scss
tbrisker: That would be preferable if that solves the issue on all pages. In general we should try to place css styling on the element that needs it if possible, to avoid side effects.
~You might need to play with css positioning a bit to solve the two-pane issue as well, and that could possibly resolve the spacing issues too.~ Nevermind, saw you fixed it in js.

test/controllers/shared/basic_rest_response_test.rb
```diff
@@ -36,5 +36,110 @@ def basic_edit_test(object_found = nil)
...
+
+        teardown do
+          Setting[:entries_per_page] = @old
+        end
+      end
+    end
+  end
+
+  def get_factory_name(model)
+    model = model.to_s
```
tbrisker: you could just use `@controller.resource_name` instead of converting the class to_s here and elsewhere (just note that it returns lowercase, i.e. `subnet` not `Subnet`). This will also remove the need for the splitting of the model below, e.g. `HostsController.new.resource_name == host` etc.

test/controllers/bookmarks_controller_test.rb
```diff
@@ -1,6 +1,13 @@
 require 'test_helper'
 
 class BookmarksControllerTest < ActionController::TestCase
+  setup do
+    @factory_options = [{:controller => "users"}]
```
tbrisker, line 5: I think the bookmark factory doesn't care about this option

test/controllers/shared/basic_rest_response_test.rb
```diff
@@ -36,5 +36,110 @@ def basic_edit_test(object_found = nil)
         end
       end
     end
+
+    def basic_pagination_per_page_test
+      context 'GET #index' do
+        setup do
+          @entries_per_page = Setting[:entries_per_page] ? Setting[:entries_per_page] : 5
+          factory_name =  get_factory_name(@controller.resource_class)
```
tbrisker: Since you always pass the same parameter which is based on the instance variable which is also available to the function, there is no real need to pass it as a parameter and save the result etc. If you access the controller directly inside `get_factory_name` you can just use `create_model_list(get_factory_name, @entries_per_page + 2, *@factory_options)`

tbrisker (changes requested):
Almost there, don't despair! :smiley:  
I figured out a couple more changes to the test that can simplify the code even further, and a couple of other small comments.

lizagilman:
[test katello]

test/controllers/bookmarks_controller_test.rb
lizagilman: I see that it does..

test/controllers/shared/basic_rest_response_test.rb
lizagilman: looks like `@controller.resource_name` doesn't cover all cases

lizagilman:
[test foreman]

tbrisker (approved):

tbrisker:
ready to merge once tests are green

tbrisker:
Thanks @lizagilman!

ohadlevy:
Awesome! Good job @lizagilman !

PR #4553, opened 2017-05-24
Fixes #19646 - improve inline help

ares:


orrabin:
Maybe instead of breaking the lines with `br` give the tooltip a size and let it break itself?

I tried to reduce the screen width and got this:
image

app/views/nic/_base_form.html.erb
```diff
@@ -63,17 +63,17 @@
 
 <% if SETTINGS[:unattended] %>
   <%= checkbox_f f, :managed,
-                 :label_help => _("Should this interface be managed via DHCP and DNS smart proxy and should it be configured during provisioning?"),
+                 :label_help => _("If managed flag is enabled, external services such as DHCP, DNS and TFTP will be configured according to the provided information.<br/>What services will be updated is determined by the subnet and domain that are selected for this interface. If you wish to manage<br/> only subset of these services, change the subnet and domain proxies configuration. When proxy is set to 'None', the corresponding<br/> service configuration will be skipped.<br/><br/>If manage flag is disabled, none of the service will be configured for this interface, even when it's configured for subnet and domain.<br/><br/>The flag also determines whether the interface should be configured during host provisioning, so that the configuration remains after reboot."),
```
ShimShtein: > If you wish to manage only subset of these services, change the subnet and domain proxies configuration.

It's a bit unclear where to change this configuration, it would be awesome if you provided a link to those settings directly.

ShimShtein (commented):
:+1: Looks nice. Just one inline comment.

app/views/nic/_base_form.html.erb
ares: that would need to read selected subnet and domain from select using JS which I'd like to avoid, or what target did you have in mind?

ares:
> Maybe instead of breaking the lines with br give the tooltip a size and let it break itself?

We tend to use this everywhere so I'm not sure whether it make sense to change styling only for this help. Would it be acceptable to raise a new issue for revisting tooltips formatting and maybe shrinking the size everywhere?

orrabin:
> Would it be acceptable to raise a new issue for revisting tooltips formatting and maybe shrinking the size everywhere?

Sounds good to me.

app/views/nic/_base_form.html.erb
ShimShtein: You can point to subnets and domains pages (which would be something). Bonus points to wiring freeip and domain_selected xhrs to help with setting proper links.

ares:
Thanks @orrabin, opened as http://projects.theforeman.org/issues/19655

app/views/nic/_base_form.html.erb
ares: ok, fair enough, I'll look at this probably next week

ares:
@ShimShtein I added links to subnets and domains list. Changing to current selected value is complicated since we don't have routes in JS. I could hardcode something like '/subnets/$current' but it doesn't feel right.

Since all comments have been addressed, @ShimShtein @orrabin or anyone else, please take another look.

app/views/nic/_base_form.html.erb
```diff
@@ -63,17 +63,17 @@
 
 <% if SETTINGS[:unattended] %>
   <%= checkbox_f f, :managed,
-                 :label_help => _("Should this interface be managed via DHCP and DNS smart proxy and should it be configured during provisioning?"),
+                 :label_help => _("If the Managed flag is enabled, external services such as DHCP, DNS, and TFTP will be configured according to the information provided.<br/>The services to be configured is determined by the subnet and domain that are selected for this interface. If you want to manage<br/> only a subset of these services, change the %{subnet} and %{domain} proxies configuration. When a proxy is set to 'None', the corresponding<br/> service configuration will be skipped.<br/><br/>If the Managed flag is disabled, none of the services will be configured for this interface, even if it is configured for the subnet and domain.<br/><br/>The Managed flag also determines whether the interface should be configured during host provisioning, so that the configuration remains after reboot.") % { :subnet => link_to(_('subnet'), subnets_path, :target => :blank), :domain => link_to(_('domain'), domains_path, :target => :blank)},
```
dLobatog: `The services to be configured *are* determined`.
In my opinion, the part about `When a proxy is set to 'None', the corresponding<br/> service configuration will be skipped.<br/>`  doesn't really add anything, it's to be expected.

app/views/nic/_base_form.html.erb
```diff
@@ -63,17 +63,17 @@
...
 <% end %>
 <%= checkbox_f f, :primary,
-               :label_help => _("The Primary interface is used for constructing the FQDN of the host"),
+               :label_help => _("The Primary interface is used for constructing the FQDN of the host.<br/>This interface is also used for network communication during provisioning, e.g. downloading installation packages."),
                :label_help_options => { :rel => 'popover-modal' },
                :class => :interface_primary,
                :size => "col-md-8", :label_size => "col-md-3" %>
 <%= checkbox_f f, :provision,
-               :label_help => _("The Provisioning interface is used for TFTP of PXELinux (or SSH for image-based hosts)"),
+               :label_help => _("The Provisioning interface is used for TFTP configuration for PXE boot based on this interface MAC address.<br/>For image based provisioning using finish template, for SSH connection we use the provisioning interface IP as a target."),
```
dLobatog: `For image-based provisioning using finish templates, this interface's IP will be used as the SSH target`

dLobatog (changes requested):
Thanks @ares , I think the messages are definitely clearer. The br HTML tags though, muddle up the string for translations. We have examples of including the HTML tags in the strings in Foreman but it's better to avoid it - it's confusing to see such long strings with tags on Transifex. `app/views/ssh_keys/ssh_key_tab,html.erb` or `app/views/key_pairs/index.html.erb` have examples of what I mean.

app/views/nic/_base_form.html.erb
ares: updated

re 'None', from my experience, that's not really expected :-)

app/views/nic/_base_form.html.erb
ares: updated

ares:
I think in some cases, it's better to leave simple `<br/>` in translated string. After breaking the string, translator loses to context and there's a chance of using other wording in the same tooltip. Anyway, I separated them for now, I hope they will be removed soon anyway (reported as http://projects.theforeman.org/issues/19655).

dLobatog (approved):

dLobatog:
Thanks @ares !

PR #4554, opened 2017-05-25
Fixes #19631 - ComputeResource with KeyPair can be removed

dLobatog:
The concern for removing the compute resources with keypairs was trying
to destroy the compute resource too early. Also, this deletion already
happened because of the has_one :dependency => :destroy relation

This fixes the problem and adds tests to ensure the relation works.

### DIFF
--- a/app/models/concerns/key_pair_compute_resource.rb
+++ b/app/models/concerns/key_pair_compute_resource.rb
@@ -23,8 +23,17 @@ module KeyPairComputeResource
     setup_key_pair
   end
 
-  def delete_key_pair(kay_pair_name)
-    delete_key_from_resource(kay_pair_name)
+  def delete_key_from_resource(remote_key_pair = key_pair.name)
+    logger.info "removing key from compute resource #{name} "\
+                "(#{provider_friendly_name}): #{remote_key_pair}"
+    client.key_pairs.get(remote_key_pair).try(:destroy)
+  rescue => e
+    Foreman::Logging.exception(
+      "Failed to delete key pair from #{provider_friendly_name}: #{name}, you "\
+      "might need to cleanup manually: #{e}",
+      e,
+      :level => :warn
+    )
   end
 
   private
@@ -39,16 +48,12 @@ module KeyPairComputeResource
   end
 
   def destroy_key_pair
-    return unless key_pair
-    delete_key_from_resource(key_pair.name)
-    key_pair.destroy!
-  end
-
-  def delete_key_from_resource(key_pair_name)
-    logger.info "removing #{provider_friendly_name}: #{name} key #{key_pair_name}"
-    key = client.key_pairs.get(key_pair_name)
-    key.nil? || key.destroy
-  rescue => e
-    logger.warn "failed to delete key pair from #{provider_friendly_name}: #{name}, you might need to cleanup manually : #{e}"
+    return unless key_pair.present?
+    delete_key_from_resource
+    # If the key pair could not be removed, it will be logged.
+    # Returning 'true' allows this method to not halt the deletion
+    # of the Compute Resource even if the key pair could not be
+    # deleted for some reason (permissions, not found, etc...)
+    true
   end
 end
--- a/test/models/compute_resources/ec2_test.rb
+++ b/test/models/compute_resources/ec2_test.rb
@@ -1,59 +1,39 @@
 require 'test_helper'
 require 'models/compute_resources/compute_resource_test_helpers'
 
-class EC2Test < ActiveSupport::TestCase
-  include ComputeResourceTestHelpers
+module Foreman
+  module Model
+    class EC2Test < ActiveSupport::TestCase
+      include ComputeResourceTestHelpers
 
-  test "#associated_host matches any NIC" do
-    host = FactoryGirl.create(:host, :ip => '10.0.0.154')
-    cr = FactoryGirl.build(:ec2_cr)
-    iface = mock('iface1', :public_ip_address => '10.0.0.154', :private_ip_address => "10.1.1.1")
-    assert_equal host, as_admin { cr.associated_host(iface) }
-  end
+      should have_one(:key_pair).with_foreign_key('compute_resource_id').
+        dependent(:destroy)
 
-  describe "find_vm_by_uuid" do
-    it "raises RecordNotFound when the vm does not exist" do
-      cr = mock_cr_servers(Foreman::Model::EC2.new, empty_servers)
-      assert_find_by_uuid_raises(ActiveRecord::RecordNotFound, cr)
-    end
+      test "#associated_host matches any NIC" do
+        host = FactoryGirl.create(:host, :ip => '10.0.0.154')
+        cr = FactoryGirl.build(:ec2_cr)
+        iface = mock('iface1', :public_ip_address => '10.0.0.154', :private_ip_address => "10.1.1.1")
+        assert_equal host, as_admin { cr.associated_host(iface) }
+      end
 
-    it "raises RecordNotFound when the compute raises EC2 error" do
-      cr = mock_cr_servers(Foreman::Model::EC2.new, servers_raising_exception(Fog::Compute::AWS::Error))
-      assert_find_by_uuid_raises(ActiveRecord::RecordNotFound, cr)
-    end
-  end
+      describe "find_vm_by_uuid" do
+        it "raises RecordNotFound when the vm does not exist" do
+          cr = mock_cr_servers(Foreman::Model::EC2.new, empty_servers)
+          assert_find_by_uuid_raises(ActiveRecord::RecordNotFound, cr)
+        end
 
-  context "key pairs" do
-    setup do
-      @aws_key_pairs = []
-      3.times do |i|
-        @aws_key_pairs << AWSKeyPair.new("foreman-#{i}", "13:01:73:0#{i}")
+        it "raises RecordNotFound when the compute raises EC2 error" do
+          cr = mock_cr_servers(Foreman::Model::EC2.new, servers_raising_exception(Fog::Compute::AWS::Error))
+          assert_find_by_uuid_raises(ActiveRecord::RecordNotFound, cr)
+        end
+      end
+
+      context "key pairs" do
+        test "should be capable of key_pair" do
+          cr = FactoryGirl.create(:ec2_cr)
+          assert_includes(cr.capabilities, :key_pair)
+        end
       end
     end
-
-    test "#get_compute_key_pairs" do
-      cr = FactoryGirl.build(:ec2_cr)
-      key_pair = FactoryGirl.build(:key_pair)
-      cr.key_pair = key_pair
-      Foreman::Model::EC2.any_instance.stubs(:key_pairs).returns(@aws_key_pairs)
-      assert_kind_of(ComputeResourceKeyPair, cr.get_compute_key_pairs.first)
-    end
-
-    test "should be capable of key_pair" do
-      cr = FactoryGirl.create(:ec2_cr)
-      assert_includes(cr.capabilities, :key_pair)
-    end
-  end
-end
-
-# We can't use 'Fog::Compute::AWS::KeyPair' model
-# This class mocks it.
-class AWSKeyPair
-  attr_reader :name, :fingerprint, :private_key
-
-  def initialize(name, fingerprint, private_key = nil)
-    @name = name
-    @fingerprint = fingerprint
-    @private_key = private_key
   end
 end
--- a/test/models/compute_resources/openstack_test.rb
+++ b/test/models/compute_resources/openstack_test.rb
@@ -1,115 +1,122 @@
 require 'test_helper'
 require 'models/compute_resources/compute_resource_test_helpers'
 
-class OpenstackTest < ActiveSupport::TestCase
-  include ComputeResourceTestHelpers
+module Foreman
+  module Model
+    class OpenstackTest < ActiveSupport::TestCase
+      include ComputeResourceTestHelpers
 
-  setup do
-    @compute_resource = FactoryGirl.build(:openstack_cr)
-  end
+      should have_one(:key_pair).with_foreign_key('compute_resource_id').
+        dependent(:destroy)
 
-  teardown do
-    Fog.unmock!
-  end
+      setup do
+        @compute_resource = FactoryGirl.build(:openstack_cr)
+      end
 
-  test "#associated_host matches any NIC" do
-    host = FactoryGirl.create(:host, :ip => '10.0.0.154')
-    iface = mock('iface1', :floating_ip_address => '10.0.0.154', :private_ip_address => "10.1.1.1")
-    assert_equal host, as_admin { @compute_resource.associated_host(iface) }
-  end
+      teardown do
+        Fog.unmock!
+      end
 
-  test "boot_from_volume does not get triggered when a string 'false' is passed as argument" do
-    Fog.mock!
-    @compute_resource.stubs(:key_pair).returns(mocked_key_pair)
-    @compute_resource.expects(:boot_from_volume).never
-    @compute_resource.create_vm(:boot_from_volume => 'false', :nics => [""],
-                                :flavor_ref => 'foo_flavor', :image_ref => 'foo_image')
-  end
+      test "#associated_host matches any NIC" do
+        host = FactoryGirl.create(:host, :ip => '10.0.0.154')
+        iface = mock('iface1', :floating_ip_address => '10.0.0.154', :private_ip_address => "10.1.1.1")
+        assert_equal host, as_admin { @compute_resource.associated_host(iface) }
+      end
 
-  describe "formatting hints" do
-    it "formats well when set to ServerGroupAntiAffinity" do
-      args = {
-        :scheduler_hint_filter => "ServerGroupAntiAffinity",
-          :scheduler_hint_data => {
-            :scheduler_hint_value => "some-uuid"
+      test "boot_from_volume does not get triggered when a string 'false' is passed as argument" do
+        Fog.mock!
+        @compute_resource.stubs(:key_pair).returns(mocked_key_pair)
+        @compute_resource.expects(:boot_from_volume).never
+        @compute_resource.create_vm(:boot_from_volume => 'false', :nics => [""],
+                                    :flavor_ref => 'foo_flavor', :image_ref => 'foo_image')
+      end
+
+      describe "formatting hints" do
+        it "formats well when set to ServerGroupAntiAffinity" do
+          args = {
+            :scheduler_hint_filter => "ServerGroupAntiAffinity",
+              :scheduler_hint_data => {
+                :scheduler_hint_value => "some-uuid"
+              }
           }
-      }
-      desired = {
-        :os_scheduler_hints => {
-          :group => "some-uuid"
-        }
-      }
-      @compute_resource.format_scheduler_hint_filter(args)
-      assert_equal(desired, args)
-    end
-
-    it "formats well when set to ServerGroupAffinity" do
-      args = {
-        :scheduler_hint_filter => "ServerGroupAffinity",
-          :scheduler_hint_data => {
-            :scheduler_hint_value => "some-uuid"
+          desired = {
+            :os_scheduler_hints => {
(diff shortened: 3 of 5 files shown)


app/models/concerns/key_pair_compute_resource.rb
```diff
@@ -5,7 +5,7 @@ module KeyPairComputeResource
     prepend KeyPairCapabilities
     has_one :key_pair, :foreign_key => :compute_resource_id, :dependent => :destroy
     after_create :setup_key_pair
-    after_destroy :destroy_key_pair
+    after_destroy :delete_key_from_resource
```
domcleal: This method may return false on failure (if `key.destroy` returns false, and possibly on `rescue`), halting the callback, which is different behaviour to the commit prior to 59e7136. Should it remain wrapped to ignore failures?

test/models/compute_resources/ec2_test.rb
```diff
@@ -1,59 +1,77 @@
 require 'test_helper'
 require 'models/compute_resources/compute_resource_test_helpers'
+require 'fog/aws'
```
domcleal: Should be unnecessary, it's part of the app init.

test/models/compute_resources/ec2_test.rb
```diff
@@ -1,59 +1,77 @@
...
+          Foreman::Model::EC2.any_instance.stubs(:key_pairs).returns(@aws_key_pairs)
+          assert_kind_of(ComputeResourceKeyPair, cr.get_compute_key_pairs.first)
+        end
+
+        test "should be capable of key_pair" do
+          cr = FactoryGirl.create(:ec2_cr)
+          assert_includes(cr.capabilities, :key_pair)
+        end
+
+        test 'should remote the key pair in the compute resource' do
```
domcleal: 1. s/remote/remove/
1. ... on compute resource deletion
1. Should this be in a test for the concern, rather than in the EC2 test? It isn't specific to EC2.
1. What asserts that the key pair is removed?

test/models/compute_resources/ec2_test.rb
```diff
@@ -1,59 +1,77 @@
...
+          cr = FactoryGirl.create(:ec2_cr)
+          key_pair = FactoryGirl.build(:key_pair)
+          mock_key_pairs = mock()
+          cr.key_pair = key_pair
+          cr.send(:client).expects(:key_pairs).returns(mock_key_pairs)
+          mock_key_pairs.expects(:get).with(key_pair.name).returns(key_pair)
+          cr.destroy
```
domcleal: Not asserted or using `destroy!`, could be failing.

domcleal (changes requested):

test/models/compute_resources/openstack_test.rb
```diff
@@ -1,115 +1,122 @@
...
-  include ComputeResourceTestHelpers
+module Foreman
+  module Model
+    class OpenstackTest < ActiveSupport::TestCase
+      include ComputeResourceTestHelpers
 
-  setup do
-    @compute_resource = FactoryGirl.build(:openstack_cr)
-  end
+      should have_one(:key_pair).with_foreign_key('compute_resource_id').
```
dLobatog, line 9: The only change in this file is the addition of this line

dLobatog:
Updated so that:

* The methods related to key pair are tested in its own concern test instead of EC2.
* The behavior of the callback is protected again (if the key pair cannot be deleted for some reason, still proceed with the compute resource deletion)
* Test asserts the deletion

test/models/concerns/key_pair_compute_resource_test.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+end
+
+class KeyPairComputeResourceTest < ActiveSupport::TestCase
+  setup { Fog.mock! }
+  teardown { Fog.unmock! }
+
+  test 'should remove the key pair on compute resource deletion' do
+    cr = DummyComputeResource.new
+    key_pair = FactoryGirl.build(:key_pair)
```
domcleal: This should also have `.expects(:destroy)`, though it also seems to be destroyed twice (in a quick test) which needs checking. Otherwise nothing asserts that the key is destroyed on deletion.

test/models/concerns/key_pair_compute_resource_test.rb
```diff
@@ -0,0 +1,49 @@
+require 'test_helper'
+
+class DummyComputeResource < ComputeResource
+  include KeyPairComputeResource
+
+  def client
+    @client ||= ::Fog::Compute.new(:provider => 'AWS', :aws_access_key_id => '123', :aws_secret_access_key => 'abcd')
```
domcleal: Nitpick, `include Mocha::API` above and then you can simply use `mock('client')` - much lighter than using Fog.

domcleal (dismissed):

test/models/concerns/key_pair_compute_resource_test.rb
dLobatog: It's destroyed twice because there's an expectation below that returns (wrongly) a Foreman object KeyPair as the result of `client.key_pairs.get` (this should just return an object from fog)

dLobatog:
@domcleal Thanks for looking in detail, I think I've fixed the problem with the expectation happening twice.

`destroy` was being called twice because there's an expectation below that returns (wrongly) a Foreman object KeyPair as the result of client.key_pairs.get (this should just return an object from fog). Now there's an specific mock for the key pair "in fog", and all 3 (ComputeResource, KeyPair in Foreman and key pair in Fog) deletions are verified

app/models/concerns/key_pair_compute_resource.rb
```diff
@@ -23,8 +23,8 @@ def recreate
     setup_key_pair
   end
 
-  def delete_key_pair(kay_pair_name)
-    delete_key_from_resource(kay_pair_name)
+  def delete_key_pair(key_pair_name)
+    delete_key_from_resource(key_pair_name)
```
tbrisker: I'm not sure why we need this function that only calls another function tbh, but in any case `delete_key_from_resource` no longer takes an argument.

dLobatog: It's not needed, I kept it because ComputeResource uses it but it could very well use `delete_key_from_resource(key_pair)`.  I noticed this change (removing the arguments from delete_key_from_resource) broke that call, so thanks for that 😄  Fixing..

dLobatog:
@tbrisker See https://github.com/theforeman/foreman/pull/4554#discussion_r118890034 , it helped me notice a bug and fix it in the latest update.

app/models/concerns/key_pair_compute_resource.rb
```diff
@@ -23,8 +23,12 @@ def recreate
     setup_key_pair
   end
 
-  def delete_key_pair(kay_pair_name)
-    delete_key_from_resource(kay_pair_name)
+  def delete_key_from_resource(remote_key_pair = key_pair)
+    logger.info "removing #{provider_friendly_name}: #{name} key #{remote_key_pair.name}"
+    key = client.key_pairs.get(remote_key_pair.name)
+    key.nil? || key.destroy
```
lzap: Nitpicks: This could be just `client.key_pairs.get(remote_key_pair.name).try(:destroy)`. Also the info log above is a little bit confusing when rendered as `removing openstack: myopenstack key foreman-XYZ". There might be a better wording.

app/models/concerns/key_pair_compute_resource.rb
```diff
@@ -23,8 +23,12 @@ def recreate
...
+  rescue => e
+    logger.warn "failed to delete key pair from #{provider_friendly_name}: #{name}, you might need to cleanup manually : #{e}"
```
lzap: I think this should be proper error with stacktrace, do not rethrow tho, but keep it in logs.

        Foreman::Logging.exception("Failed to delete key pair", e)

test/models/concerns/key_pair_compute_resource_test.rb
```diff
@@ -0,0 +1,50 @@
...
+    key_pair = FactoryGirl.build(:key_pair)
+    cr.key_pair = key_pair
+    mock_key_pairs = mock('mock_key_pairs')
+    fog_key_pair = mock('fog_key_pair')
+    cr.send(:client).expects(:key_pairs).returns(mock_key_pairs)
+    mock_key_pairs.expects(:get).with(key_pair.name).returns(fog_key_pair)
+    key_pair.expects(:destroy)
+    fog_key_pair.expects(:destroy)
+    assert cr.destroy!
+  end
```
lzap: Nitpick: I think all of the mocks stubs should be created using `stubs` except of course `destroy` methods.

test/models/concerns/key_pair_compute_resource_test.rb
```diff
@@ -0,0 +1,50 @@
...
+  class FakeKeyPair
+    attr_reader :name, :fingerprint, :private_key
+
+    def initialize(name, fingerprint, private_key = nil)
+      @name = name
+      @fingerprint = fingerprint
+      @private_key = private_key
+    end
+  end
+end
```
lzap: Nitpick: Whole ordering of this file is reversed, I'd expect inner class first, then testing `get` and then `delete` :-)

lzap (changes requested):
I only have couple of nitpicks but some of them are worth one rebase.

test/models/compute_resources/ec2_test.rb
```diff
@@ -1,59 +1,39 @@
...
-    it "raises RecordNotFound when the vm does not exist" do
-      cr = mock_cr_servers(Foreman::Model::EC2.new, empty_servers)
-      assert_find_by_uuid_raises(ActiveRecord::RecordNotFound, cr)
-    end
+module Foreman
+  module Model
+    class EC2Test < ActiveSupport::TestCase
+      include ComputeResourceTestHelpers
+
+      should have_one(:key_pair).with_foreign_key('compute_resource_id').
```
dLobatog, line 9: The only change in this file is this line addition

dLobatog:
Thanks for the review @lzap , I think I addressed all comments except the one about stubs, not sure what's wrong with the expectations in the test?

test/models/concerns/key_pair_compute_resource_test.rb
```diff
@@ -0,0 +1,50 @@
...
+  test 'should remove the key pair on compute resource deletion' do
+    cr = DummyComputeResource.new
+    key_pair = FactoryGirl.build(:key_pair)
+    cr.key_pair = key_pair
+    mock_key_pairs = mock('mock_key_pairs')
+    fog_key_pair = mock('fog_key_pair')
+    cr.send(:client).expects(:key_pairs).returns(mock_key_pairs)
+    mock_key_pairs.expects(:get).with(key_pair.name).returns(fog_key_pair)
+    key_pair.expects(:destroy)
+    fog_key_pair.expects(:destroy)
```
lzap: Since we've seen double deletion previously, maybe to add `expects(:destroy).once`?

lzap (approved):

lzap:
@dLobatog asked to take a quick look on a call, I thought you are on PTO, did not listen carefully on the meeting @domcleal if you want to take it back I don't mind.

@dLobatog there is nothing wrong with expectations there, it's not incorrect to have it like that.

Looks good to me, I am going to test this now.

test/models/concerns/key_pair_compute_resource_test.rb
lzap: This is the last thing, I will add this during merge.

lzap:
When I attempt to delete a key from my trystack CR, it is erroring out with:

```
[app|I] Started DELETE "/compute_resources/2-trystack/key_pairs/foreman-29a407b63-e0ad-41be-bcbe-5f02db79b313" for 127.0.0.1 at 2017-05-30 18:44:13 +0200
[app|I] Processing by KeyPairsController#destroy as HTML
[app|I]   Parameters: {"authenticity_token"=>"YikIwMaFHq6f1FyRsOxHN8LTkG9U0O78ktO+6P40rMXjEn1LrX+/kxBJnXFFTF0eZdVAr62+ifEJzkH6b3qZMg==", "compute_resource_id"=>"2-trystack", "id"=>"foreman-29a407b63-e0ad-41be-bcbe-5f02db79b313"}
[app|I] Current user: admin (administrator)
[app|D] Setting current user thread-local variable to admin
[app|D] Setting current organization thread-local variable to none
[app|D] Setting current location thread-local variable to none
[app|W] Failed to delete key pair from OpenStack: trystack, you might need to cleanup manually: undefined method `name' for "foreman-29a407b63-e0ad-41be-bcbe-5f02db79b313":String
 | NoMethodError: undefined method `name' for "foreman-29a407b63-e0ad-41be-bcbe-5f02db79b313":String
 | /home/lzap/work/foreman/app/models/concerns/key_pair_compute_resource.rb:28:in `delete_key_from_resource'
 | /home/lzap/work/foreman/app/controllers/key_pairs_controller.rb:30:in `destroy'
 | /home/lzap/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/actionpack-4.2.8/lib/action_controller/metal/implicit_render.rb:4:in `send_action'
 | /home/lzap/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/actionpack-4.2.8/lib/abstract_controller/base.rb:198:in `process_action'
 | /home/lzap/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/actionpack-4.2.8/lib/action_controller/metal/rendering.rb:10:in `process_action'
 | /home/lzap/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/actionpack-4.2.8/lib/abstract_controller/callbacks.rb:20:in `block in process_action'
 | /home/lzap/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:117:in `call'
 | /home/lzap/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:555:in `block (2 levels) in compile'
 | /home/lzap/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:505:in `call'
 | /home/lzap/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:498:in `block (2 levels) in around'
 | /home/lzap/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:313:in `block (2 levels) in halting'
 | /home/lzap/work/foreman/app/controllers/concerns/application_shared.rb:15:in `set_timezone'
 | /home/lzap/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:432:in `block in make_lambda'
 | /home/lzap/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:312:in `block in halting'
 | /home/lzap/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:497:in `block in around'
 | /home/lzap/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:505:in `call'
 | /home/lzap/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:498:in `block (2 levels) in around'
 | /home/lzap/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:313:in `block (2 levels) in halting'
 | /home/lzap/work/foreman/app/models/concerns/foreman/thread_session.rb:32:in `clear_thread'
```

The problem is that when used via controller directly it passes name instead of AR instance.

Please also improve deletion workflow in the controller, it errors out to the logs but returns with success message. The reason for that is the method is used from controller workflow when exception is expected as well as from callbacks when we do not want them. Can you also add tests for CRUD from UI controller and test this path? Frankly, the way we use `delete_key_from_resource` is clunky, but I can't find a better solution.

Also one thing which is mystery for me - ma OpenStack key is showing "Key not connected to any compute resource" message (*), but I can't call method `active` on the instance:

```
[13] pry(main)> ComputeResource.unscoped.last.key_pairs.first
=>   <Fog::Compute::OpenStack::KeyPair
    name="foreman-29a407b63-e0ad-41be-bcbe-5f02db79b313",
    fingerprint="0c:ce:xxxxxxx:12:f4:43:9a",
    public_key="ssh-rsa AAAxxxxxxxxxxxxrYk5r Generated-by-Nova",
    private_key=nil,
    user_id=nil,
    id=nil
  >
[14] pry(main)> ComputeResource.unscoped.last.key_pairs.first.active
NoMethodError: undefined method `active' for #<Fog::Compute::OpenStack::KeyPair:0x0056392b8e3ba8>
```

Can you explain me what "not connected" key means and why is it listed for my OpenStack CR? When it's not connected, why it shows here? I don't follow this.

(*) https://github.com/theforeman/foreman/blob/develop/app/helpers/key_pairs_helper.rb#L3

test/models/concerns/key_pair_compute_resource_test.rb
lzap: Rebase needed, please add `once` then.

app/models/concerns/key_pair_compute_resource.rb
```diff
@@ -23,8 +23,17 @@ def recreate
     setup_key_pair
   end
 
-  def delete_key_pair(kay_pair_name)
-    delete_key_from_resource(kay_pair_name)
+  def delete_key_from_resource(remote_key_pair = key_pair)
+    logger.info "removing key from compute resource #{name} "\
+                "(#{provider_friendly_name}): #{remote_key_pair.name}"
+    client.key_pairs.get(remote_key_pair.name).try(:destroy)
```
lzap: Nil exception when string is passed to the method.

lzap (changes requested):
Deletion from the key list page is broken.

dLobatog:
@lzap Thanks for testing, I've rebased and:

* Added `.once` to the expectations for destroy, although I think that doesn't do anything, expecting things once is the default.
* Fixed the problem where just the name (string) was passed to `delete_key_from_resource`

About `active`: It's not a property of the Fog KeyPair object, but of the `ComputeResourceKeyPair` service, therefore it's normal that Fog throws that error pasted in the comment. 

The controller lists *all* keys available in the ComputeResource. The `active` key is the key that the ComputeResource is currently using - this key is passed onto all vms that it uses. It's automatically setup when you create the ComputeResource - https://github.com/theforeman/foreman/pull/4024/files#diff-397787e68eb0a5119e12f02d73bf2016R34 - since there is a `has_one` relation between KeyPair and ComputeResource, only one KeyPair is active, the rest of the keypairs shown in the `key_pairs` index page are other key pairs available in the resource.

> Please also improve deletion workflow in the controller, it errors out to the logs but returns with success message. 

If by that you mean the ComputeResource deletion, it's "expected" (well.. it's the current behavior) that if you can't delete the KeyPair in the remote resource, that's glossed over and the ComputeResource is deleted just fine (no errors on the UI, only on the logs, warning level). 

I do agree it's a bit clunky and hides maybe too much detail from the user, but I'd like to keep this patch as small as possible, as it's meant to be in 1.15.1. If you want I can take a stab at improving the error messaging when deleting a ComputeResource with KeyPair in another PR if that's ok?

tbrisker (approved):
LGTM, failing integration tests don't seem related. @lzap - any further comments?

lzap:
Test failure unrelated? "Error: Cannot find module 'tough-cookie'"

lzap:
Yeah @tbrisker finished testing right now, removal is being orchestrated correctly now. I stubmled upon a bug that is somehow related but we should fix it in a separate ticket: http://projects.theforeman.org/issues/19917

lzap (approved):

lzap:
Merged thanks!

PR #4555, opened 2017-05-25
Fixes #19706 - don't treat DHCP leases as conflicts

lzap:
Our previous code that should ignore leases was based on existance of hostnames, it used to work but no longer works in recent releases. In smart-proxy, we added explicit flag to indicate if a DHCP record is a reservation or a lease:

https://github.com/theforeman/smart-proxy/pull/532

This patch leverages this and removes the old hostname "hack" to proper test if the existing record is a lease or not. Conflict is not thrown in that case.

### DIFF
--- a/lib/net/dhcp/record.rb
+++ b/lib/net/dhcp/record.rb
@@ -1,6 +1,6 @@
 module Net::DHCP
   class Record < Net::Record
-    attr_accessor :name, :ip, :mac, :network, :nextServer, :filename, :related_macs
+    attr_accessor :name, :ip, :mac, :network, :nextServer, :filename, :related_macs, :type
 
     def initialize(opts = { })
       super(opts)
@@ -8,6 +8,19 @@ module Net::DHCP
       self.mac     = Net::Validations.validate_mac! self.mac
       self.network = Net::Validations.validate_network! self.network
       self.ip      = Net::Validations.validate_ip! self.ip
+      self.type    = opts["type"]
+    end
+
+    def legacy_dhcp_api?
+      type.nil?
+    end
+
+    def lease?
+      type && type == "lease"
+    end
+
+    def reservation?
+      !lease?
     end
 
     def to_s
@@ -41,7 +54,7 @@ module Net::DHCP
 
     # Returns an array of record objects which are conflicting with our own
     def conflicts
-      conflicts = [proxy.record(network, mac), proxy.records_by_ip(network, ip)].flatten.compact.delete_if { |c| c == self || related_macs.include?(c.mac) }
+      conflicts = [proxy.record(network, mac), proxy.records_by_ip(network, ip)].flatten.compact.delete_if { |c| c.lease? || c == self || related_macs.include?(c.mac) }
       @conflicts ||= conflicts.uniq {|c| c.attrs}
     end
 
@@ -57,8 +70,10 @@ module Net::DHCP
 
       # If we're converting an 'ad-hoc' lease created by a host booting outside of Foreman's knowledge,
       # then :hostname will be blank on the incoming lease - if the ip/mac still match, then this
-      # isn't a conflict.
-      to_compare << :hostname if other.attrs[:hostname].present? && attrs[:hostname].present?
+      # isn't a conflict. Only applicable on legacy proxy API without "type" attribute.
+      if legacy_dhcp_api?
+        to_compare << :hostname if other.attrs[:hostname].present? && attrs[:hostname].present?
+      end
 
       # Not all DHCP smart-proxy providers support TFTP filename option (e.g. libvirt).
       to_compare << :filename if other.attrs[:filename].present? && attrs[:filename].present?
@@ -68,9 +83,16 @@ module Net::DHCP
     end
 
     def attrs
-      { :hostname   => hostname, :mac => mac, :ip => ip, :network => network,
-        :nextServer => nextServer, :filename => filename, :name => name,
-        :related_macs => related_macs
+      {
+        :hostname => hostname,
+        :mac => mac,
+        :ip => ip,
+        :network => network,
+        :nextServer => nextServer,
+        :filename => filename,
+        :name => name,
+        :related_macs => related_macs,
+        :type => type
       }.delete_if { |k, v| v.nil? }
     end
   end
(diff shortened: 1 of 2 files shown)


lib/net/dhcp/record.rb
```diff
@@ -10,8 +10,16 @@ def initialize(opts = { })
       self.ip      = Net::Validations.validate_ip! self.ip
     end
 
+    def lease?
+      type && type == "lease"
```
dLobatog: I'm not sure the first `type` (presence check) is doing anything here

lib/net/dhcp/record.rb
```diff
@@ -10,8 +10,16 @@ def initialize(opts = { })
...
+    end
+
+    def reservation?
+      !lease?
```
dLobatog, line 23: Is the record type always available, even in the deprecated proxy API? The reason I'm asking is because all leases will look like reservations as `!lease?` evaluates to true if `type == nil`

dLobatog (changes requested):
Thanks @lzap - I'll test more thoroughly once the proxy part is merged, but it looks good. Can you add some test demonstrating what changed on `test/unit/net/dhcp_test` ?

lib/net/dhcp/record.rb
lzap: Yeah, my thinking was if you setup Foreman with older version of Proxy, it will not return type in the REST API response. Type will be `nil` but it is not necessary here, sure.

lib/net/dhcp/record.rb
lzap: That's intentional, I want responses from deprecated proxy API to be reservations - that's the safest option here. Anoter alternative would be to return `false` in both cases and create helper method `deprecate_api?` that wold be `!(lease? && reservation?)` and use it to execute old code for backward compatibity (the code I removed - comparing hostnames). What you prefer?

dLobatog: I think the 2nd option sounds nicer from the user POV, the behavior of the previous implementation would still work. If you can do that and mark it as deprecated I think it'd be the best.

dLobatog (changes requested):

lzap:
I created new ticket for this problem

http://projects.theforeman.org/issues/19706

because the originally associated one is not fixed by this patch

http://projects.theforeman.org/issues/19634

@dLobatog I still think the original issue you run into needs a change in proxy (do not fail when lease cannot be deleted). Or we can leverage the type field and never attempt to delete a lease from Foreman. But that would be a separate patch.

lzap:
I will get back to this patch shortly, somehow got lost in my inbox. This is very important change to have unfortunately, blocker for discovery.

timogoebel:
@lzap: Please add some tests for this as well.

yoctozepto:
I applied this patch on 1.15.3 in production and it fixed changing IP address after discovery. See http://projects.theforeman.org/issues/19634#note-29 for more details.

lzap:
@timogoebel discovery unit tests are broken due to Rails 5 upgrade at the moment, out of ideas if you have any...

iNecas:
@lzap I gave it some debugging, see https://github.com/theforeman/foreman/pull/4863 if that helps with the issue (it should), not sure if I've broken anything by this change but if we don't try, we never know.

lzap:
Added tests!

lib/net/dhcp/record.rb
lzap: Implemented second option, so previous implementation stays the same. There is a "legacy" flag and that is used in the codebase.

lib/net/dhcp/record.rb
```diff
@@ -41,7 +54,7 @@ def create
 
     # Returns an array of record objects which are conflicting with our own
     def conflicts
-      conflicts = [proxy.record(network, mac), proxy.records_by_ip(network, ip)].flatten.compact.delete_if { |c| c == self || related_macs.include?(c.mac) }
+      conflicts = [proxy.record(network, mac), proxy.records_by_ip(network, ip)].flatten.compact.delete_if { |c| c.lease? || c == self || related_macs.include?(c.mac) }
```
lzap, line 57: This is THE change actually, if it's a lease type of record, filter it out. It is the first statement in the chain, most likely to happen.

lib/net/dhcp/record.rb
```diff
@@ -57,8 +70,10 @@ def ==(other)
 
       # If we're converting an 'ad-hoc' lease created by a host booting outside of Foreman's knowledge,
       # then :hostname will be blank on the incoming lease - if the ip/mac still match, then this
-      # isn't a conflict.
-      to_compare << :hostname if other.attrs[:hostname].present? && attrs[:hostname].present?
+      # isn't a conflict. Only applicable on legacy proxy API without "type" attribute.
+      if legacy_dhcp_api?
+        to_compare << :hostname if other.attrs[:hostname].present? && attrs[:hostname].present?
+      end
```
lzap, line 76: Moved the old hostname-based comparison to legacy branch. Hostname is no longer relevant, it's just a "comment" type of entry in DHCP.

lib/net/dhcp/record.rb
```diff
@@ -68,9 +83,16 @@ def ==(other)
...
+      {
+        :hostname => hostname,
+        :mac => mac,
+        :ip => ip,
+        :network => network,
+        :nextServer => nextServer,
+        :filename => filename,
+        :name => name,
+        :related_macs => related_macs,
+        :type => type
```
lzap, line 95: Only added the type there.

dLobatog (approved):
I was able to reproduce and this fixes the problem just fine and works with old smart proxies as well (treating the DHCP record as reservation). Cherry-picking to 1.15 for 1.15.5 and 1.16. 

Thanks @lzap !

dLobatog:
[test katello] - let's give it a spin, the error doesn't look like it was caused by this PR 
```
07:08:24  > /usr/bin/git fetch --tags --progress git://github.com/theforeman/foreman.git +refs/pull/*:refs/remotes/origin/pr/*
07:08:31 Checking out Revision c83509cab80b72e058253b82debb50535764c287 (detached)
07:08:32 org.eclipse.jgit.errors.MissingObjectException: Missing unknown c83509cab80b72e058253b82debb50535764c287
```

lzap:
Thanks a lot for testing, this one is a blocker for discovery (as discovered hosts always has leases -> conflict). All tests are green. PS: I am late on github, ping me if there is a problem merging this, I would like to hit 1.15.5. @oprazak

dLobatog (commented):
Thanks @lzap !

PR #4556, opened 2017-05-25
Fixes #15861 - increase size of audited_changes column

amirfefer:
In mysql text type is limited to 64K, while in psql there's no limit.
Changing column's type to MEDIUMTEXT, which limited to 16MB,
should solve this issue in mysql

### DIFF
--- a/test/models/concerns/audit_extensions_test.rb
+++ b/test/models/concerns/audit_extensions_test.rb
@@ -46,4 +46,15 @@ class AuditExtensionsTest < ActiveSupport::TestCase
     FactoryGirl.create(:ec2_cr, :with_auditing)
     refute_empty Audit.search_for("type = compute_resource")
   end
+
+  test "audited changes field can be greater then 65K bytes" do
+    prov_template = templates(:mystring)
+    prov_template.template = "0000000000" * 3500
+    as_admin do
+      assert prov_template.save!
+      prov_template.template = "1111111111" * 3500
+      assert prov_template.save!
+    end
+    assert Audit.last.audited_changes.to_s.bytesize > 66000
+  end
 end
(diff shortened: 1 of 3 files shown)


dLobatog (approved):
Thanks @amirfefer - will merge if tests pass

dLobatog:
Check out tests, they seem to fail on mysql @amirfefer

test/models/concerns/audit_extensions_test.rb
```diff
@@ -46,4 +46,13 @@ def setup
     FactoryGirl.create(:ec2_cr, :with_auditing)
     refute_empty Audit.search_for("type = compute_resource")
   end
+
+  test "audit change field can be greater then 65K bytes" do
+    prov_template = templates(:mystring)
+    prov_template.template = "0" * 66000
```
lzap: Nitpick: I think this might be faster a bit:

     prov_template.template = "0000000000" * 6600

```
[root@zzzap katello-3.4.0.rc2]# time ruby -e '"0000000000" * 99999999'

real    0m4.588s
user    0m0.256s
sys     0m0.124s
[root@zzzap katello-3.4.0.rc2]# time ruby -e '"0" * 999999990'

real    0m0.352s
user    0m0.240s
sys     0m0.112s
```

Even faster would be String.new("", capacity: 66000) and appending it in a loop perhaps.

test/models/concerns/audit_extensions_test.rb
```diff
@@ -46,4 +46,13 @@ def setup
...
+    as_admin do
+      assert prov_template.save!
+    end
+    assert_equal 66000, prov_template.template.bytesize
```
tbrisker: shouldn't this check the audited changes?
Also, afaict it doesn't look like the template field itself supports medium text, in fact the only column i found using mediumtext is the log column in the reports table.

tbrisker (changes requested):

amirfefer:
@lzap - thanks for your performance advice :+1: 
@tbrisker  - you're right, I've changed the test a bit
@dLobatog could you re-review?

dLobatog:
Thanks @amirfefer !

PR #4557, opened 2017-05-25
Fixes #19672 - fix bulk actions for plugins

ares:


ares:
When https://github.com/theforeman/foreman/pull/4338 was merged, a new method `multiple_with_filter?` has been introduced in hosts_controller and exposed via helper_method. It's also being used in generic helper `HostsAndHostgroupsHelper#multiple_filter` and it works since most actions are implemented in hosts controller. If plugin adds actions (e.g. foreman_openscap - assign compliance policy), a plugin controller is used and the method is undefined. This method should be probably available globally or the helper should duplicate the logic and check for search param presence. I believe this affects all plugins introducing custom bulk actions and should be fixed in 1.15 branch.

I think this is a candidate for 1.15.1 blocker

app/helpers/application_helper.rb
```diff
@@ -40,6 +40,10 @@ def link_to_function(name, function, html_options = {})
     content_tag(:a, name, html_options.merge(:href => href, :onclick => onclick))
   end
 
+  def multiple_with_filter?
```
dLobatog: Please leave a comment explaining why - as a quick scan will show this is duplicate code to what's in the controller

dLobatog (changes requested):
Seems like an easy fix for plugins, I'm not so sure about it blocking the release though. The method could be implemented in the plugin controller or helper trivially (just like here).

Is there any other plugin affected by this? I suppose it happens with all plugins that call 'find_multiple' which includes Ansible, Chef etc.. 

Since it's so small and convenient I guess it doesn't hurt on 1.15.1 - any downsides of having this in the application controller you can think of?

Thanks @ares

ares:
> Seems like an easy fix for plugins, I'm not so sure about it blocking the release though. The method could be implemented in the plugin controller or helper trivially (just like here).
> Is there any other plugin affected by this? I suppose it happens with all plugins that call 'find_multiple' which includes Ansible, Chef etc..

The 1.15 release has broken all plugins extending bulk actions. Plugins like REX and OpenSCAP don't work properly. To fix it in plugins, we'd need to copy the method to every plugin and do release for each of them. That was the reason I suggested this as a blocker.

> any downsides of having this in the application controller you can think of?

the method name seems confusing for other controllers, I've added a comment

dLobatog (approved):

dLobatog:
Thanks @ares - merged for 1.15.1

PR #4558, opened 2017-05-26
fixes #18200 - don't re-encrypt settings when value is unchanged

domcleal:


lzap:
Looks good, but can you add a test for a case when value is not decryptable? You handle this correctly in the code, but I cannot find coverage for this case, it's important one. If setting is corrupted somehow, we need to be able to overwrite it with the new value (e.g. when algorithm is changed).

domcleal:
Test added to check an unencrypted database value (the point of `is_decryptable?`) is then encrypted when setting the new value.

lzap:
I think I did not understand semantics of `is_decryptable?` but that does not change a thing, you made it correctly. Thanks, waiting for tests.

lzap (approved):
RTM

lzap:
Unrelated failures, thanks @domcleal !

PR #4559, opened 2017-05-29
Fixes #19700 - update rubocop rules

ohadlevy:
The following changes have been made:

- Performance/RedundantMerge:
changes lines such as:
```ruby
not_found_message.merge! :message => options
```
to:
```ruby
not_found_message[:message] = options
```

- converts str.match() to str =~ ()
- remove extra whitespaces etc.

### DIFF
--- a/lib/proxy_api/dhcp.rb
+++ b/lib/proxy_api/dhcp.rb
@@ -22,9 +22,12 @@ module ProxyAPI
 
     def unused_ip(subnet, mac = nil)
       params = {}
-      params.merge!({:mac => mac}) if mac.present?
+      params[:mac] = mac if mac.present?
 
-      params.merge!({:from => subnet.from, :to => subnet.to}) if subnet.from.present? && subnet.to.present?
+      if subnet.from.present? && subnet.to.present?
+        params[:from] = subnet.from
+        params[:to] = subnet.to
+      end
       if params.any?
         params = "?" + params.map{|e| e.join("=")}.join("&")
       else
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -41,3 +41,9 @@ Metrics/ClassLength:
 Performance/FixedSize:
   Exclude:
     - 'test/**/*'
+
+Performance/RedundantMatch:
+  Enabled: true
+
+Performance/RedundantMerge:
+  Enabled: true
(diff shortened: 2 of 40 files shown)


dLobatog (changes requested):
Thanks @ohadlevy ! Take a look at the seeds changes, it looks like tests are failing because of some problem there (including some Katello tests, only 11 of those failures are "supposed to be"  there)

db/seeds.d/08-data.rb
```diff
@@ -1,4 +1,4 @@
-SEEDED_PARTITION_TABLES = [
+SEEDED_PARTITION_TABLES ||= [
```
lzap: Reason for this?

lib/proxy_api/dhcp.rb
```diff
@@ -22,9 +22,12 @@ def subnet(subnet)
...
     def unused_ip(subnet, mac = nil)
       params = {}
-      params.merge!({:mac => mac}) if mac.present?
+      params[:mac] = mac if mac.present?
 
-      params.merge!({:from => subnet.from, :to => subnet.to}) if subnet.from.present? && subnet.to.present?
+      if subnet.from.present? && subnet.to.present?
+        params[:from] = subnet.from
+        params[:to] = subnet.to
+      end
```
lzap, line 30: This might be reasonable case when to use `merge` tho, but I can live with that enforcing.

lzap (dismissed):
Test failures, also I'd suggest setting `MaxKeyValuePairs` to just `1` - I think two lines of code is worth merging into one `merge` call. Defaults to 2.

lib/proxy_api/dhcp.rb
lzap: Also see my review comment, I'd change MaxKeyValuePairs.

db/seeds.d/08-data.rb
ohadlevy: I was trying to address ```foreman/db/seeds.d/08-data.rb:1: warning: already initialized constant SEEDED_PARTITION_TABLES``` I'll revert as its not really related to this PR.

ohadlevy:
Tests are passing.

db/seeds.d/08-data.rb
lzap: Hmm looks we load this twice.

ohadlevy:
@dLobatog can you re-review please? thanks.

dLobatog (approved):

dLobatog:
Thanks @ohadlevy & @lzap !

PR #4560, opened 2017-05-30
Fixes #18855 - Add better error message on invalid taxonomies

xprazak2:
We can't really fix this right now, so I just aim to add better error message. The validation errors appear when a taxonomy is invalid. It is really easy to make a taxonomy invalid, all it takes is to update os from facts.

### DIFF
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -60,9 +60,9 @@ class Role < ApplicationRecord
   taxonomy_join_table = :taxable_taxonomies
   has_many taxonomy_join_table.to_sym, :dependent => :destroy, :as => :taxable
   has_many :locations, -> { where(:type => 'Location') },
-           :through => taxonomy_join_table, :source => :taxonomy
+           :through => taxonomy_join_table, :source => :taxonomy, :validate => false
   has_many :organizations, -> { where(:type => 'Organization') },
-           :through => taxonomy_join_table, :source => :taxonomy
+           :through => taxonomy_join_table, :source => :taxonomy, :validate => false
 
   validates :name, :presence => true, :uniqueness => true
   validates :builtin, :inclusion => { :in => 0..2 }


app/controllers/api/v2/roles_controller.rb
```diff
@@ -57,7 +57,12 @@ def destroy
       param_group :role
       def clone
         @role = @role.clone(role_params)
-        process_response @role.save
+        if @role.save
+          process_success
+        else
+          @role.add_tax_errors
```
tstrachota: Could you move it to `after_validation` callback of `Role`? I guess this error can appear during other operations like update too. Moving it there would cover it.

app/models/role.rb
```diff
@@ -212,4 +217,9 @@ def find_filter(resource_type, current_filters, search)
       self.filters.build(:search => search)
     end
   end
+
+  def invalid_tax_error(tax)
+    msg = "cannot be cloned. Some #{tax} are invalid. Please fix their associations and try again."
+    errors.add(:base, _(msg))
```
tstrachota: I'm afraid gettext won't be able to correctly find the string this way. You need to wrap the string definition with `N_(...)` and extract `#{tax}` into string substitution or define it twice. The latter is probably better because we translate words "organization" and "location" too.

tstrachota (changes requested):
@xprazak2 can this issue affect other taxable resources too? I wonder if it makes sense to implement some generic solution that would improve error messages for all affected resources.

xprazak2:
I moved the code into `after_validation` callback and improved translations. Role is not really taxable, it uses the the taxonomy relations because of filters, but does not include Taxonomix. This does not affect taxable resources (I tested Medium and Domain).

ares (changes requested):
@tstrachota is that the error you were able to reproduce? I reproduced what @xprazak2 described also in UI and not just for cloning. The error happens on any role save if there's invalid organization among selected ones. With the patch, nicer error is displayed but it mentions cloning even when I'm creating new role. There's still ugly error message for org selector (see second screenshot) but that's probably out of scope for this PR.
tax_select1
tax_select2

I think alternative approach might be better. How about just skipping orgs/locs validation for roles? I don't think assigning invalid object should prevent operations on Role. It also solves the selector error message.

Following diff worked for me just fine

```diff
diff --git a/app/models/role.rb b/app/models/role.rb
index 082a82f..07a747e 100644
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -60,9 +60,9 @@ class Role < ApplicationRecord
   taxonomy_join_table = :taxable_taxonomies
   has_many taxonomy_join_table.to_sym, :dependent => :destroy, :as => :taxable
   has_many :locations, -> { where(:type => 'Location') },
-           :through => taxonomy_join_table, :source => :taxonomy
+           :through => taxonomy_join_table, :source => :taxonomy, :validate => false
   has_many :organizations, -> { where(:type => 'Organization') },
-           :through => taxonomy_join_table, :source => :taxonomy
+           :through => taxonomy_join_table, :source => :taxonomy, :validate => false
 
   validates :name, :presence => true, :uniqueness => true
   validates :builtin, :inclusion => { :in => 0..2 }
```

tstrachota:
@ares @xprazak2  +1 for disabling the validations on taxonomy associations.

xprazak2:
I took the alternative approach and disabled the validations.

xprazak2:
Test failure unrelated, js test is failing.

tstrachota:
:+1: fixes the problem for me.

ares (approved):

ares:
Thanks @xprazak2, merged as e620923c07f987b87f6c06299d961d97c184a9ff

PR #4561, opened 2017-05-30
Fixes #18717 - Allow for loadbalanced and multi-homed proxies

sean797:
This adds the concept of 'Hostnames' for Smart Proxies, a Hostname can be assigned 1 or many Smart Proxies. When assigned 1 this could then act as a secondary way of contacting that Smart Proxy. When a Hostname is assigned to 2 or more Smart Proxies this would probably be the Load Balanced address (VIP) for the Smart Proxies.

All client communication to Proxy is now done via the Hostname. Hosts are now assigned a Hostname instead of a Proxy.

In the future, I envision us being able to do things (maybe create DHCP leases or other stuff) to all Smart Proxies a host uses. 

Note, this follows on from where https://github.com/theforeman/foreman/pull/4346 left off, taking into account the discussion on [foreman-dev](https://groups.google.com/d/msg/foreman-dev/7RJ8KSZxlAw/5AcuNSSNBwAJ)


dLobatog (commented):
As far as I understand, this is a solution so that the communication Foreman <-> Proxy goes through one interface (hostname in this case) and Proxy <-> Client uses a different one. 

How does it improve the use case for a Smart Proxy with multiple interfaces versus adding the URLs of the proxy listening in each of the interfaces as a separate Smart Proxy, in the current version of Foreman?  

Can you give some examples of configuraitons that need this setup? For which proxy features? 

The PR title and welcome page mentions load balancing as well, I can't find anything regarding load balancing, how is that done?

My personal feeling so far is that this complicates the Smart Proxy creation a little bit, by introducing a new menu item and automatically creating the Hostname. Maybe an UI that incorporated Hostname creation within the SmartProxy form could fix the issue.

@sean797 thanks for the effort, it's much appreciated when a lot of us don't usually deal with these kind of setups. Let me know about the questions above please

sean797:
> How does it improve the use case for a Smart Proxy with multiple interfaces versus adding the URLs  of the proxy listening in each of the interfaces as a separate Smart Proxy, in the current version of Foreman?

Adding another 'fake' Smart Proxy for an interface doesn't work as Foreman will attempt to make a request to it which will fail and block building a host. For DHCP lease creation for example.

> Can you give some examples of configurations that need this setup? For which proxy features?

Anyone wanting clients to communicate with the Proxy on a different FQDN. Could be a NAT between client <-> Proxy or a Proxy with multiple interfaces. There is probably other situations I cannot think of. This PR is only for Puppet, though I see no reason why it cannot work with other features in the future, for example...
- Pulp (as @stbenjam said [here](https://github.com/theforeman/foreman/pull/4346#pullrequestreview-27597118) "want to make sure all the proxies have the same content" )
- TFTP (Foreman could ensure all proxies have the same files)
- DHCP (foreman can create the leases on both nodes, http://steronius.blogspot.co.uk/2015/10/dhcp-failover-on-rhel-7.html DHCP failover would also require installer changes )

> The PR title and welcome page mentions load balancing as well, I can't find anything regarding load balancing, how is that done?

The user will have to find another way to replicate Puppet SSL certs and modules between any load balanced Puppet CA / Masters. Once they are replicated the client should be able to switch between the Smart Proxies without any issues. I don't think Foreman can help with the replication really. You could use something like https://www.cyberciti.biz/faq/linux-inotify-examples-to-replicate-directories/ for example. I'll probably do a blog post about it, assuming this get merged at some point.
See [here](https://www.draw.io/?lightbox=1&highlight=0000ff&edit=_blank&layers=1&nav=1&title=Untitled%20Diagram.xml#R7VnbcpswEP0az7QPySAE2H603aTtTDOTqWd6eZRBAbUCUSFiu19fCQRYFrGdOBenbR4ctCxC2rPn7MoewFm6es9RnlyxCNOB60SrAXw3cF3gQUf%2BU5Z1bRk6o9oQcxJpp84wJ7%2BxNurn4pJEuDAcBWNUkNw0hizLcCgMG%2BKcLU23G0bNt%2BYoxpZhHiJqW7%2BSSCS1deQGnf0DJnHSvBkE4%2FrOAoU%2FY87KTL9v4MKb6q%2B%2BnaJmLr3RIkERW26Y4MUAzjhjor5KVzNMVWybsNXPXd5xt103x5k45AG3fuAW0VJv%2FZJxnKJML0%2Bsm5AsEyLwPEehGi8l7AM4TURK5QjISz0P5gKv7lwLaHcoMwezFAu%2Bli76ARfq1aybZNBBWnYQuB6sbclG%2BMfaD2nU43bqbufyQm%2B%2BPxBja8M4kimhh4yLhMUsQ%2FSis04rnLGawDGDITfM19%2B0vRp8V4NzXw5%2FYCHWOuNRKZg0dbN%2FYiw3AqoWsTuccs2s5KH20tERiMdYbEBsB51jigS5NWfvi2D16IRztN5wyBnJRLEx87UydFgC3zOx9PwtNOoZO2zapR0EF7AT97rMc6yW8BnnMqKFBajMTGHixHFBfqNF5aCg0ruS3v504L%2BTFkRJnElDKIOOuTSoDCdSKCb6RkqiqEoGihaYTlv6zxhlvHpvIwA7OKJlTK%2BkU4dNsMc7ueOcO54%2FNEKuAbgfzh2QjQu7uSlUWLeodD%2B4enSmhWtGCa7eOJ9%2FUkNcYSfXnVMZaEGY1CLHAvMIdhaSHWKiqoQClqKiIGFjviS0I3FkO0njhsuzc9l7CS7DkWMkFgy2asr9%2FI%2FmPrRyaZ5K6KTpmrOV2pmMQEAV1xdcXsXqqk23K1QoJquUClCqqlj9qVJvYqcZpbLnUOn09PVvO2zALn%2FAG9nlr62Tx9Q%2FYFNsf3kDDy9v%2B%2Fl7AP8exivvRGqkPzYBf%2FQaCf61Ggmcfhz1TGfOORgDs0qewRMqk95eZXNfpbL57gHSpqvEE0ib92TdQ8YyvNU6OKfQOgS2xA2PlLiDo%2B3vaPWQQJTFlf78KnEhqmS9rLs9xsXfJkd3tGsbcjQyxQickBgFFo5Np36nkhQJypUxYqWMy8UzKow7hkYkA9ijMG6PwoweQ2CCVyww3XcVVf%2FWtnMPF5%2BhLT4vcm7xtqqOB43%2ByvIPnJ3%2Bx%2Fdjw%2F%2FK2GRNsJPNUhhB07020uiekDSCR%2F228FV0FD2khi9B6u1T8j5Sg2C0y%2F9oUrs7Dln%2FGqn3fUXpGkj4J0RoW5m%2FEC5KpHw%2BXttnrjcfJs2JLKSlPnalaL3Aby1QO96DZ2mFtjPe7zlswWFPKxTcvxWSw%2B63qjru3Q%2BC8OIP) for my amazing diagram 🤣 


> My personal feeling so far is that this complicates the Smart Proxy creation a little bit, by introducing a new menu item and automatically creating the Hostname. Maybe an UI that incorporated Hostname creation within the SmartProxy form could fix the issue. 

I feel the same, I think we already have enough menu items. Especially when you start adding a couple of plugins. I guess instead of a menu item we could have a button from the Smart Proxies page, maybe later we can get someone with good insights to UI design input before this is merged, providing people are happy with everything else.  

Thanks for taking a look @dLobatog 🍻

sean797:
Unrelated test failures should be fixed now.. Lets check..

[test foreman]
[test katello]

sean797:
Rebased and tests passed, would appreciate if someone can review this 😄

lzap:
I apologize to be late in this discussion, had this tab opened for months :-)

First of all, I struggle to find a reason why Foreman must be aware of multi-homed smart-proxies at all. I mean, a smart proxy (or Capsule) is an entity itself and if one wants to load-balance incoming requests, it should be done there. Foreman should transparently talk to one smart-proxy while on the background that could be two or more.

I understand that if Foreman was aware of this, user could manually pick the proxy per individual host or hostgroup, but in HA environments is this really needed? I'd rather see an approach in extending our smart-proxy codebase to be able to talk to multiple underlying services (DHCP, TFTP) or some changes in underlying services themselves to be HA-aware (Pulp).

sean797:
> First of all, I struggle to find a reason why Foreman must be aware of multi-homed smart-proxies at all. I mean, a smart proxy (or Capsule) is an entity itself and if one wants to load-balance incoming requests, it should be done there. Foreman should transparently talk to one smart-proxy while on the background that could be two or more.

The problem is the users selects a proxy (e.g when creating a host) this is then used to render the hosts templates and the proxy hostname used in the templates is wrong, and breaks provisioning.

> I understand that if Foreman was aware of this, user could manually pick the proxy per individual host or hostgroup, but in HA environments is this really needed? I'd rather see an approach in extending our smart-proxy codebase to be able to talk to multiple underlying services (DHCP, TFTP) or some changes in underlying services themselves to be HA-aware (Pulp).

If the smart proxy was made to be aware of multiple underlying services and interact with them all then that moves the single point of failure to the foreman-proxy service, whereas with this approach the foreman app becomes the single point of failure (which can be setup for HA)

lzap:
I briefly looked at the code and it's a monster patch which I believe is trying to solve one particular deployment: when you have two interfaces for both satellite and capsule. In these enterprise environments it is common to have "administrative" networks separated from "data" networks. I understand the motivation but this makes things utterly complex for other users, maybe 90% of them. We need to find a compromise and we should do this in the RFC thread rather than here with all the code written (this is definitely lot of work in there).

I will be completely honest here, I don't like this in a way it is designed because of the complexity. We see this when users are setting up provisioning (template - os - default template). I'd prefer much lightweight implementation where things stays as-is but enabling multi-home users to do their job. Any idea how to achieve that? Maybe to create "aliases" to each proxy and a switch to use them.

Another idea, if I say that most of these enterprise customers only have two interfaces for each capsule (proxy), is that true? Because if most of them are just two, we can completely avoid introducing new model and relations and create new field for existing proxy called "alias" and build some mechanism around it so clients will use it depending on what user sets in host/hostgroup. That's essentially the workaround which is currently recommended, if I get it right.

Another idea would be to make an assumption that all client communication is http(s), then there is another patch which is adding http proxies (https://github.com/theforeman/foreman/pull/4616). The purpose is a little bit different (its for compute resources) but we could refactor it to have two kinds of these proxies, one could be simply "client proxy". A little bit crazy but it's also not intrusive.

sean797:
> I briefly looked at the code and it's a monster patch which I believe is trying to solve one particular deployment: when you have two interfaces for both satellite and capsule. In these enterprise environments it is common to have "administrative" networks separated from "data" networks. I understand the motivation but this makes things utterly complex for other users, maybe 90% of them. We need to find a compromise and we should do this in the RFC thread rather than here with all the code written (this is definitely lot of work in there).

Thanks for taking a look! :-)
Not just 1 type of deployment, also loadbalaned proxies and when there is a NAT between client and smart proxy.
This has already been discussed in https://github.com/theforeman/foreman/pull/4346 and on [foreman-dev](https://groups.google.com/d/msg/foreman-dev/7RJ8KSZxlAw/5AcuNSSNBwAJ) I don't believe complexity has increase much. AFAIK the only scenario where this adds anything else for the user to do is if they change a proxies hostname, which I don't think many people do. Most users who don't use this feature wont ever have to know about it or do anything with it (unless they change a proxies hostname)

> Another idea, if I say that most of these enterprise customers only have two interfaces for each capsule (proxy), is that true? Because if most of them are just two, we can completely avoid introducing new model and relations and create new field for existing proxy called "alias" and build some mechanism around it so clients will use it depending on what user sets in host/hostgroup. That's essentially the workaround which is currently recommended, if I get it right.

That enables a smart proxy to have 2 interfaces, why limit this to 2? Some people will want more, but i guess for most 2 will be enough for most. Also what about load balanced proxies? for example in the future I hope for foreman to create DHCP entries on multiple smart proxies when required. See https://github.com/theforeman/puppet-foreman_proxy/pull/352

> Another idea would be to make an assumption that all client communication is http(s), then there is another patch which is adding http proxies (#4616). The purpose is a little bit different (its for compute resources) but we could refactor it to have two kinds of these proxies, one could be simply "client proxy". A little bit crazy but it's also not intrusive.

I'm not sure I understand, but not all client communication is http(s), what about DHCP, TFTP, maybe others

lzap:
Thanks, is my understanding correct that foreman will always communicate to smart-proxy using its primary URL and it is up to the administrator to setup other host names (e.g. into HA or cluster via shared storage or some kind of replication)?

Before we go any further with this high impact PR, I would like to hear more opinions namely from @ohadlevy @iNecas @witlessbird @ares @jlsherrill and @stbenjam to name few.

sean797:
> Thanks, is my understanding correct that foreman will always communicate to smart-proxy using its primary URL and it is up to the administrator to setup other host names (e.g. into HA or cluster via shared storage or some kind of replication)?

Correct - So just as today Proxies have a URL which Foreman uses to communicate with it. Depending on features a Smart Proxy cluster might not even require any shared storage or any custom replication as foreman could use Smart Proxies belonging to a Hostname to:
 - Create the DHCP leases on all of them
 - Ensure TFTP files are the same on all of them
 - Puppet CA would require replicating certificates or using shared storage. (maybe even the installer could setup replication using inotify)
 - Named (DNS) could be configured as master/slave, that would probably just require Puppet module changes. 
 - In the Katello world (Pulp) if the Smart Proxys are in the same Orgs, Locs & LC env then they would have the same content. Katello could add a check for that.

All of the above would allow Foreman & Katello clients to hit any Smart Proxy in a cluster for provisioning, on-going config-mgmt & pkg updates (katello), potentially via a VIP or loadblancer.

ares:
I did not do full review but from quick read through I have several comments. First of all, thank you for this PR, it seems like a lot effort was invested into it. Now my question is, why would Foreman need to know all names of the proxy. I guess @lzap had the same concern but I didn't understand the [answer](https://github.com/theforeman/foreman/pull/4561#issuecomment-313684121). It there was a LB that we'd talk to, Foreman wouldn't have to know which proxy handles the request. If we need multihomed proxies and Foreman awareness of which proxy name should be used in particular case is desired, can't we just add multiple proxies with different URLs? Thanks for patience when answering :-)

If it's up to user to setup sharing between multiple proxies I think we should have a decent documentation prior we ship this in any release. Or some checking mechanism that would verify the shared storage works.

sean797:
> If there was a LB that we'd talk to, Foreman wouldn't have to know which proxy handles the request.

True, but if Foreman is aware that a group of smart proxys are a HA cluster, it can use this knowledge as per my previous [comment](https://github.com/theforeman/foreman/pull/4561#issuecomment-319682981) eliminating the need for any shared storage or replication in some cases, simplifying it in all others. Also getting hosts to use the LB would suffer the issues of multihoming as it would require adding a fake proxy, see below.

> If we need multihomed proxies and Foreman awareness of which proxy name should be used in particular case is desired, can't we just add multiple proxies with different URLs?

Maybe.. AFAIK the current problems with that approach are:
- Validation will fail when creating it if the URL isn't reachable by Foreman. (Which it almost never is allowed to, hense the need for multihoming)
- Same problem if you select it when creating a host, Foreman will attempt to talk to it for DHCP leases creation or whatever and boom!

You could probably get around these by putting an entry in /etc/hosts (I haven't tried) but with this LB'ing approach you get multihoming for free. Also in Katello this would create 2 Pulp sync tasks on the same smart proxy, I don't know if this would cause an issue but its not necessary, I'm not sure how other plugins would be affected.

Although loadbalanced and multi-homed smart proxies probably can be done today via adding fake smart proxies and /etc/hosts hacks, to me it all feels very hacky and there are a lot of benefits if Foreman has this information, I've shared the benefits I can think of but I'm sure there are others I cannot.

evgeni:
There are two scenarios that can be (partially) solved by this PR:
* HA: multiple proxies are responsible for the same (virtual) proxy-name and it does not matter to which proxy the client is talking to (imagine some LB between client and proxies)
* multi-homing/multi-nic: the same proxy is reachable by different names and clients can only use one of the available names due to routing/firewalls/whatever

I only have experience with the second, so I'll just explain my wishes for that, and maybe that helps a bit the understanding of the problem.

In bigger environments it is common to separate different network segments by vlans, subnets etc. But at the same time you often do not want your central infrastructure to have a full node in each of those segments. As a solution many environments deploy systems with multiple (virtual) network cards, thus making those systems reachable in multiple segments. Each such network card will have an own IP address and quite often a different hostname. Imagine your proxy being proxy.example.net, but having two additional interfaces, which resolve to proxy.devel.example.net and proxy.qa.example.net (even worse, they will sometimes be called proxydevel.example.net and proxyqa.example.net). It's also often not just two networks, but more than that.

Now in Foreman, we have a lot of places where we render templates that contain the hostname of the proxy. This is to allow the client to connect to the proxy when it needs to. If the client can resolve and reach proxy.example.net from every vlan/subnet, there is nothing to be done and the communication will work fine. But if (and that is sadly common) the client cannot resolve the name in a different segment or can resolve but not reach it, we need a way to provide the client with the correct service name in its segment, or the client won't be able to communicate with us.

This PR should (I did not read every line of the patch) add the said hostname attribute to proxies and allow the user to tell which hostname of a proxy a client should be using. This then allows us to render the right name in the templates and thus the client will talk to us on the right interface.

lzap:
After reading above, specifically @evgeni 's description, I still feel like a simple alias would do this trick much more easily.

For the first HA case, I think sharing storage is the most reliable way of doing HA and if this is not doable or preferred, any other mean of distributing/copying/streaming data would do this. We should not make this a Foreman feature at all costs.

We already have a bit overengineered database association between Templates and Operating Systems and it is causing confusion every single week. That's why we hesitate with this patch so much.

sean797:
> After reading above, specifically @evgeni 's description, I still feel like a simple alias would do this trick much more easily.

Agreed, that's kind of what this does, just using another table, if we were to create an `alias` column on the `smart_proxies` table, as you previously suggested, we would be limiting it to 2 interfaces. I know of 1 enterprise that requires many more (they have 100s on DMZs all with a NAT where the clients in the DMZ have a different hostname for the Proxy, putted a Proxy in each DMZ or even one for every 2 DMZ's would result in a unmanageable number of Proxies), I'm sure there are others in a not to different situation.

> For the first HA case, I think sharing storage is the most reliable way of doing HA and if this is not doable or preferred, any other mean of distributing/copying/streaming data would do this. We should not make this a Foreman feature at all costs.

What costs do you mean?
Shared storage is expensive 💰, makes the fail over procedure more complicated, active/active becomes harder to achieve (most SANs don't support write/write) and if Smart Proxies can be made HA without shared storage it allows them to scale horizontally for redundancy and/or scale.
Other means of distributing/copying/streaming data become more and more unreliable the more the data changes, it'll probably be okay for copying small Puppet SSL Certs but anything larger can become a big problem. 

> We already have a bit overengineered database association between Templates and Operating Systems and it is causing confusion every single week. That's why we hesitate with this patch so much.

Honestly I don't see how that compares.. That's a required step for any user, whereas this would only be required for people wanting to take advantage of the feature. I don't see a better way of implementing these features that users ask for almost everyday.

sean797:
ping @lzap @ares, any more comments? I'd like to start moving this forward :)

Todo list:
- [ ] validate smart proxy certs when adding a hostname
- [ ] validate proxies with the same hostname has the same features

ares:
As discussed yesterday, this will focus on multihoming use case only. We'll need to figure out how to do it less confusing for users. We should make sure that host name never gets out of sync with smart proxy taxonomies, that it's clear that user are choosing hostname of a particular smart proxy etc. Maybe the hostname does not have to exist as a standalone object and we could simply have a list of alternative hostnames as an attribute of a smart proxy. Something like each line of text area being one hostname. That would simplify taxonomies, permissions, menus. After we finalize the implementation details, we should pull our UX experts into the discussion.

ekohl:
> validate smart proxy certs when adding a hostname

@sean797 you should also ensure for services which are served over HTTPS (puppet / puppetca) that they also contain the correct name(s) on the certificate.

I'd also argue that https://github.com/theforeman/smart-proxy/pull/510 becomes much more important.

bkearney:
> As discussed yesterday, this will focus on multihoming use case only. We'll need to figure out how to do it less confusing for users. 

Are we still going to suport hostname being an m to n relationship to smart proxies?

ares:
If we just convert it to an attribute of a smart proxy, it would be still an option to specify the same name for two of them (loadbalancing use case). So while it wouldn't be n:m relation in terms of SQL, you could still have n proxies with m host names. Obviously it's prone to typos if you want to have the same name on two proxies, but we said we're focusing on multihome use case where I think it's fine.

ekohl:
@ares that comes back to the suggestion I had for a proxy. I'm not exactly sure where the scope shall be so here's at least some info what can be exposed.

We can have many services on a proxy that are exposed to systems:
* foreman-proxy https://fqdn:8443 http://fqdn:8000
* puppet https://fqdn:8140
* puppet ca https://fqdn:8140
* tftp udp://ip:69
* dhcp udp://ip:67
* dns udp://ip:53 tcp://ip:53

In case of Katello it's:

* foreman-proxy https://fqdn:9090 (didn't check HTTP)
* pulp https://fqdn:443
* rhsm https://fqdn:8443

Note that the proxy can also proxy provisioning templates so both Foreman and clients talk to it.

For proper multi homing all the HTTPS services should be configurable where they listen but you can start with splitting in 2: where Foreman connects and where clients connect.

bkearney:
> If we just convert it to an attribute of a smart proxy, it would be still an option to specify the same name for two of them (loadbalancing use case). So while it wouldn't be n:m relation in terms of SQL, you could still have n proxies with m host names. Obviously it's prone to typos if you want to have the same name on two proxies, but we said we're focusing on multihome use case where I think it's fine.

My recollection of the discussion was similar, but I understood the following for load balancing:

1) We dont want foreman to set up the load balancer
2) The normal load balancing use case looks like multi-home with the addition that the same name is shared by more than one smart-proxy. 

So, I had assumed we would support that sharing of host names.

sean797:
Okay, ready for review now! I've updated this to include the extra validation, remove the menu item and instead link from the Smart Proxies page. @ares IMHO without the n:m relation it would make for unnecessary complexity getting a list of proxies that share a Hostname, not to mention that this is what relations are for.

Some pictures for people reference: 
screenshot from 2017-10-13 15-11-57
screenshot from 2017-10-13 15-11-31
screenshot from 2017-10-13 15-11-18

lzap:
I still do not like introducing new Hostname model and M:N relationship with SmartProxy model. This new version introduced no change in that. Let me explain.

Today, we have quite nice workaround for multi-homing using a hostgroup/host parameter, users can specify exactly what this patch provides. I would not be against making this a first class citizen in New/Edit Host Form as some kind of "override hostname" flag with just some plain list of possible hostnames.

New model being just a string that can be assigned in different parts of Foreman is fine. In the end, it's just a rendered string in a template what is the issue we are trying to solve here. But new model and  delegating *some* (not all!) responsibilities across, that's another story and recipe for regressions and overall more complexity.

The second problem of load-balanced proxies (we agreed we won't touch for now) can be also solved in different ways, for example we can create some awareness that some proxies are "linked" together and Foreman could prevent orchestrating on those (load balanced) proxies.

ares:
@sean797 could you please rebase? I'd like to test how it works. My concern with hostname being a separate model is the taxonomies can get out of sync easily. I don't think it's addressed here (I just quickly checked the code)

ares:
thanks :-)

sean797:
@ares rebased 👍 I think you're correct, taxonomies work here just like they currently do in every other model, IMO that's another issue

@lzap Honestly as a user I don't see the benefit of choosing a smart proxy and then a hostname, why not make it simpler and choose one? The New Host/Edit form already has LOTS of options, especially with plugins.

While I agree with your first point (about regressions) though I don't think that should stop us moving forward, arguable that's the difference between vanilla open source software and a supported product based on it.

I think your second point is the crux of the issue, I'm trying to solve both issues here as they are quite similar. Multi-homing can be solved in a much simpler way (using alias). We then could create a way to link SPs separately, I think both would achieve the same end goal 🎉  That way the PR(s) would be smaller and probably easier to review, but I think linking them via the thing (Hostnames) that actual links them makes it simpler for the user and this code is already written. If you want to put down an exhaustive list of pros & cons for each implementation we can have a discussion around that.

bkearney:
@lzap isnt the linking exactly what modeling the n:m relationship is doing? I would not expect foreman to do the load balancing, but I would expect foreman to make it easy to provision a host into a load balanced data center.

lzap:
@bkearney not exactly, these are two separate things.

> Honestly as a user I don't see the benefit of choosing a smart proxy and then a hostname

Typical user (no HA/LB) would not do that. Only those who really need this feature.

bkearney:
@lzap, I agree.. you should only select the hostname.. not the smart proxy.

lzap:
Maybe I am over-simplifying but I see this as a "string on a template" problem where we don't need another M:N relationship. Just pick an alternative from a list (1:N relationship to just a "string", can be reused in other parts easily).

sean797:
The N:M relationship is important so:
- Foreman can create a DHCP lease on all nodes (normally only one DHCP server would be active, see https://github.com/theforeman/puppet-foreman_proxy/pull/352)
- Foreman can verify certs on all nodes
- Foreman can verify all nodes are in the same Org, Loc & LifeCycleEnv (katello)
- Foreman can ensure the files they server over TFTP are the same.

There maybe be more benefits with other features.

**EDIT:** Basically anything where there is state on the SP, Foreman can ensure that all the SPs have the same state.

sean797:
@lzap or anyone else, anymore comments or questions? If not can someone review please, I would like to get this merged for 1.17

ekohl:
Right now I'm very busy with just getting nightly working and the dynflow executor so we can get that into 1.16 so I can't do a proper review.

bkearney:
Can anyone else review? This would be great to see in 1.17

app/controllers/api/v1/hostgroups_controller.rb
```diff
@@ -77,6 +78,16 @@ def destroy
...
       end
+
+      private
+
+      def swap_proxy_for_hostname
+        ca_proxy_hostname = SmartProxy.find_by_id(hostgroup_params[:puppet_ca_proxy_id]).try(:hostnames).try(:first)
+        puppet_proxy_hostname = SmartProxy.find_by_id(hostgroup_params[:puppet_proxy_id]).try(:hostnames).try(:first)
+        @hostgroup_params = hostgroup_params.merge(puppet_proxy_hostname_id: puppet_proxy_hostname.try(:id)).
+                                            merge(puppet_ca_proxy_hostname_id: ca_proxy_hostname.try(:id)).
+                                            except(:puppet_ca_proxy_id, :puppet_proxy_id)
```
ares: Does this mean that puppet_proxy_id and puppet_ca_proxy_id parameters are effectively ignored? Also we always seems to find first, if the proxy has more hostnames and second is used, this would always override it to first? (not tested yet)

app/controllers/api/v1/hosts_controller.rb
```diff
@@ -152,6 +153,14 @@ def resource_class
       def forward_request_url
         @host.request_url = request.host_with_port if @host.respond_to?(:request_url)
       end
+
+      def swap_proxy_for_hostname
```
ares: this should be deduplicated with hostgroup controller, perhpas extracted to concern?

app/controllers/api/v2/hostgroups_controller.rb
```diff
@@ -138,6 +139,16 @@ def action_permission
       def allowed_nested_id
         %w(puppetclass_id location_id organization_id)
       end
+
+      def swap_proxy_for_hostname
+        ca_proxy_hostname = SmartProxy.find_by_id(hostgroup_params[:puppet_ca_proxy_id]).try(:hostnames).try(:first)
+        puppet_proxy_hostname = SmartProxy.find_by_id(hostgroup_params[:puppet_proxy_id]).try(:hostnames).try(:first)
+        Foreman::Deprecation.api_deprecation_warning('puppet_ca_proxy_id parameter is deprecated, please use the new puppet_ca_proxy_hostname_id parameter instead') if ca_proxy_hostname
+        Foreman::Deprecation.api_deprecation_warning('puppet_proxy_id parameter is deprecated, please use the new puppet_proxy_hostname_id parameter instead') if puppet_proxy_hostname
```
ares: I think also this method should be deduplicated with v2 hosts controller, you can use `instance_variable_set` and `resource_name` (IIRC) that should return 'host' or 'hostgroup'

also it seems if I specify both puppet_proxy_id and puppet_proxy_hostname_id, the puppet_proxu_id wins and therefore it's first hostname wins which I don't think is expected (not tested yet), it should print a warning that one xor another should be used and use the hostname one I think

app/controllers/api/v2/hostnames_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+          Foreman use Smart Proxy Hostnames to configure Hosts to communicate with a Proxy.
+          Generally you would have 1 Hostname per Smart Proxy, but if you have a Smart Proxy
+          serving multiple networks with multiple interfaces or a Smart Proxy cluster
+          then you will probably want more than 1 per Smart Proxy or 1 shared between 2 Smart Proxies.
+        DOC
+      end
+
+      before_action :find_resource, :only => %w{show update destroy}
+
+      api :GET, '/hostnames', N_("List of hostnames")
```
ares: Should all endpoints be nested under a smart proxy? if hostname can't exist without it, I think /proxy/1/hostnames would make more sense. I don't think users would like to list all hostnames for all proxies but I might be wrong. What do you think?

app/controllers/concerns/foreman/controller/smart_proxy_auth.rb
```diff
@@ -103,7 +103,13 @@ def auth_smart_proxy(proxies = SmartProxy.unscoped.all, require_cert = true)
     end
     return false unless request_hosts
 
-    hosts = Hash[proxies.map { |p| [URI.parse(p.url).host, p] }]
+    hosts = []
+    SmartProxy.unscoped.all.each do |proxy|
```
ares: `include(:hostnames)` to avoid n+1 SQL

app/registries/menu/loader.rb
```diff
@@ -86,6 +86,7 @@ def self.load
           if SETTINGS[:unattended]
             menu.item :compute_resources, :caption => N_('Compute Resources')
             menu.item :compute_profiles,  :caption => N_('Compute Profiles')
+            menu.divider
```
ares: is this still necessary?

app/helpers/shared_smart_proxies_helper.rb
```diff
@@ -1,23 +1,25 @@
 module SharedSmartProxiesHelper
-  def smart_proxy_fields(f, options = {})
+  def smart_proxy_fields(f, hostname = false, options = {})
```
ares: this seems as dangerous change, if someone uses the method and send some options, it would be always considered as hostname is true not getting any deprecation warning and options would be ignored, should we rather define new helper `smart_proxy_hostname_fields` and keep the old with deprecation?

ares (changes requested):
I did an initial read through of the last version. Honestly, I think this is very big change and too risky given the fact where we are with 1.17. Except for inline comments, I think it's still too complicated for what it tries to achieve which is - every host can have different value when you ask what's its puppetmaster or puppet ca hostname. I don't think Hostname should be even taxable as the taxonomy is basically inherited from Smart proxy it belongs to. Wherever we work with hostname, it should take proxy's taxonomies into consideration.

If we want to get this feature into 1.17, I think we should take simpler approach not changing any API. The smart proxy could have alternative hostnames like it does in this PR. The host and hostgroup would have a new fields for every proxy it is linked to. In UI we'd display the selector of all names for a given proxy but only if a proxy would have alternative names. I think we don't even have to complicate it with linking to features. If the proxy has alternative names, we can suppose all features can be accessed on all of these names. Then when we call host.puppetmaster, it would return the selected name or fallback to proxy url. This approach is less intrusive as it does not change existing API and relations. Also it's more consistent with existing host form behavior where you can e.g. select compute profile only if you choose compute resource other than bare metal. It would also mean no change (no risk) for users who do not define any alternative name for proxy. You wouldn't have to deal with separate entity, defining permission just for editing hostnames since it would in fact be only editing the proxy. While it would be saved as full AR object, it wouldn't be available as separate API. It would be one of proxy attribute that we'd split into separate hostnames. In API comma separated, in UI text area, each URL separate line. Personally I don't see a big value in this model and would be fine with keeping the names serialized in one text field in smart proxy. At the end it's just the list of strings. It would still allow user to have also 1 name for 2 proxies and IMHO it's simpler to setup than via linking the proxy to hostname of other proxy.

I hope I'm not missing anything important and I'm happy to explain what I mean in more details. @sean797 if you think you'd be able to change the implementation before 1.17 branching (1 week) I can closely cooperate with you and review it on time.

sean797:
Thanks for taking a look! 👍 

> I think it's still too complicated for what it tries to achieve which is - every host can have different value when you ask what's its puppetmaster or puppet ca hostname

That isn't just what it tries to achieve, I also want Foreman to ensure a cluster of SPs have the same state, therefore allowing clients to use any of the SPs in a cluster and it gets the same "thing" back. (allowing users to put a LB between SPs and clients). After this is merged expect more PRs for DHCP, TFTP and most other features, I didn't want to include those here and make this PR any bigger.

> I don't think Hostname should be even taxable

Makes sense +1

>  I think we should take simpler approach not changing any API. 

Plugin API? I don't think I changed (from the users view) any HTTP API.

> The host and hostgroup would have a new fields 

Does that mean the user would potentially select a proxy and a hostname? IMHO selecting both is not necessary, Whats the benefit of doing it that way?

> would be fine with keeping the names serialized in one text field in smart proxy. At the end it's just the list of strings. It would still allow user to have also 1 name for 2 proxies

How would that work? Aren't m to n relationships for exactly that? What is the benefit of not using a m to n relationship? Remember in future I want Foreman to be able to ensure consistent state between SPs, that means when we create a Host that uses DHCP foreman will create the lease on both SPs that are assigned the hostname. 

-----

I guess what you are suggesting is moving away from our 1 model = 1 API = 1 UI view? Which I don't think makes sense for this, if it becomes an attribute of SP then its kind of weird getting the user to select others SPs that use the same Hostname? Also if a Hostname is shared between SP1 & SP2 and a user only has access to SP1 then should they be able to view/modify/delete the Hostname?

app/controllers/api/v1/hostgroups_controller.rb
sean797: If a user gives a Proxy I don't know what else we can do but assume they want the first hostname? I guess we could compare the URL and Hostnames and fail back to .first if none matches?

app/controllers/api/v2/hostnames_controller.rb
sean797: A GET on `/proxy/1/hostnames` makes sense 👍 
I don't think edit, create or delete do though.

ares:
>> I think we should take simpler approach not changing any API.

> Plugin API? I don't think I changed (from the users view) any HTTP API.

By API I mean the API endpoints you deprecated in v2. Everywhere where proxy is being set, users should now pass the hostname instead. Even those who will continue to have single proxy.

> > The host and hostgroup would have a new fields

> Does that mean the user would potentially select a proxy and a hostname? IMHO selecting both is not necessary, Whats the benefit of doing it that way?

Yes, but only those who define multiple hostnames for the proxy they select. Otherwise the field would not even appear. The benefit is 1) users with no need to have multiple hostnames for proxy (majority I think) are not touched by the change, 2) no existing API change, just extending with a new attribute 3) consistency with the rest of the form

> would be fine with keeping the names serialized in one text field in smart proxy. At the end it's just the list of strings. It would still allow user to have also 1 name for 2 proxies

> > How would that work? Aren't m to n relationships for exactly that? What is the benefit of not using a m to n relationship? Remember in future I want Foreman to be able to ensure consistent state between SPs, that means when we create a Host that uses DHCP foreman will create the lease on both SPs that are assigned the hostname.

It seemed like the alternative hostnames are more like attributes of a proxy rather than first class citizen objects or separate entity. But if you want to share the hostname between two SPs, I'm also fine with keeping it in separate table. That does not change how they should be assigned though. You can still do something like `SmartProxy.hostnames = "a\n\b\nc" and the setter would parse the value coming from text area and create/remove related hostnames as necessary. To avoid typos, it could work similarly to http://jqueryui.com/autocomplete/#multiple-remote

That way you'd be later still able to find all proxies for a given hostname. Although that sounds more like hack to me. Shouldn't the loadbalancer know that write operations needs to be redirected to all backeds, while read can be handled by single one? I'm not sure whether Foreman should know about this.

> I guess what you are suggesting is moving away from our 1 model = 1 API = 1 UI view? Which I don't think makes sense for this, if it becomes an attribute of SP then its kind of weird getting the user to select others SPs that use the same Hostname? Also if a Hostname is shared between SP1 & SP2 and a user only has access to SP1 then should they be able to view/modify/delete the Hostname?

I was more focusing on multihome, not loadbalancing use case. I think we agreed that LB is out of this PR and as @lzap mentioned above, that could be solved differently via linking between proxies. But as stated above, we could keep that as two separate models but the Hostname without full blown UI, without API endpoints and taxonomy association. I don't think we should define any permissions for hostnames either. User should only see hostnames that are assigned to proxies he/she can see. They modification and destroying should also follow rules based on edit/destroy of proxies.

sean797:
**Loadbalancing use case is not out of scope for this PR.** SPs could have an alternative Hostname attribute and then another way of linking SPs, @lzap suggested that already I think https://github.com/theforeman/foreman/pull/4561#issuecomment-341354867, and I asked about pros & cons of doing it that way https://github.com/theforeman/foreman/pull/4561#issuecomment-341374007 already. I think it makes a lot more sense to link them via the thing that is actually common (the shared Hostname). Also @lzap @ares with this implementation what is the "other way" you want to link them? What are the pros & cons you see with that vs this implementation?

Hostnames need their own permissions, ect.. because they are shared between 2 (or more) SPs. Arguable with this approach SPs don't need their own permissions as they can be inherited by the Hostname(s). But then what happens if user has access to Hostname1 but not Hostname2 and the a Proxy uses Hostname1 & Hostname2? Actually I think both SPs and Hostnames need their own permissions, they are their own objects in their own right.

I'm fine with not deprecating API endpoints, and although I think its unnecessary (and even simpler just selecting a Hostname) I'm okay with having users wanting to take advantage of this feature selecting a proxy and then a hostname if necessary. (we can have that discussion & maybe change it in the future, as its just cosmetic) If they don't have multiple Hostnames we can use the `first` Hostname (which is what `swap_proxy_for_hostname` on the API controller does).

ares:
> Also @lzap @ares with this implementation what is the "other way" you want to link them? What are the pros & cons you see with that vs this implementation?

I'll defer to @lzap on this since I haven't thought about this in details. The thing is that with the change I suggested, the hostname would not be taken as a separate entity, therefore it would not be the link. I think using one object for two different things usually bites later when these two things start to have different requirements. If we want to implement load balancing, I'd much rather prefer to have a Load balancer as a first class citizen object, with proper permissions, taxonomy, orchestration on all linked SPs etc. This logic does not seem to belong to something called Hostname.

>  But then what happens if user has access to Hostname1 but not Hostname2 and the a Proxy uses Hostname1 & Hostname2? Actually I think both SPs and Hostnames need their own permissions, they are their own objects in their own right.

That should not really matter. By having edit permission, user can set whatever hostnames for the proxy they want. By removing it from proxy, it only removes the link between SP and hostname, it would not touch the hostname itself as long as there's at least one link. I don't expect anyone to define granular filters for edit_hostnames permissions. Permissions in Foreman are complicated more than they should already. I guess it would end up by user either having view_hostnames or *_hostnames permissions without any further restrictions.

> I'm fine with not deprecating API endpoints, and although I think its unnecessary (and even simpler just selecting a Hostname) I'm okay with having users wanting to take advantage of this feature selecting a proxy and then a hostname if necessary. (we can have that discussion & maybe change it in the future, as its just cosmetic) If they don't have multiple Hostnames we can use the first Hostname (which is what swap_proxy_for_hostname on the API controller does).

But that does not change what I dislike, changing the way how objects are linked, unless I'm missing something. The difference is that this introduces new entity that is now required to work with. See https://github.com/theforeman/foreman/pull/4561/files#diff-4d26af5acbc256436626e7dbe476e3deL242 for example, the logic should still remain the same. If host has puppet ca proxy selected or just puppet proxy then the proxy should be "inherited". At this point, the hostname should not matter yet. Only when you call `puppetmaster` or `puppet_ca_server` methods, they should ask proxy what is the hostname for it. Something like this

```ruby
# Host
def puppetmaster
  puppet_proxy.hostname(self, 'puppet')
end

# this fallsback to puppet_proxy since it's still "inheritted"
def puppet_ca_server
  puppet_ca_proxy.hostname(self, 'puppet_ca')
end

# Proxy
def hostname(host, feature)
  host.public_send("#{feature}_proxy_hostname") || self.hostnames.first.to_s || self.to_s
  # or even better, without the last part, proxy url would be automatically saved as the first hostname on proxy creation
end
```

IMHO there's no reason to stop using this - https://github.com/theforeman/foreman/pull/4561/files#diff-4d26af5acbc256436626e7dbe476e3deR73 but we also need to define `puppet_proxy_hostname_id`. Existing relations changes should not be necessary. Things like this should not be necessary https://github.com/theforeman/foreman/pull/4561/files#diff-2d53894341fd1cf84a823e3bd0855495R566 

The `swap_proxy_for_hostname` would not be necessary since the proxy model should just create first hostname based on it's url. And in UI we'd only display the hostname selector if there are 2 or more hostnames for a given proxy. That way, nothing changes for existing users.

sean797:
>  I think using one object for two different things usually bites later when these two things start to have different requirements.

I could see how creating a 'hacky' link between things that aren't really linked could bite later, but I cant see how linking something by the thing they have in common could ever be an issue.
**EDIT**: could you explain how this could become an issue later?

> If we want to implement load balancing, I'd much rather prefer to have a Load balancer as a first class citizen object, with proper permissions, taxonomy, orchestration on all linked SPs etc. This logic does not seem to belong to something called Hostname.

Creating another object to link them by is IMO not necessary and will only add complexity for users and the codebase.

If you feel strongly about this do you want to take this to foreman-dev to get some more views?

> IMHO there's no reason to stop using this - https://github.com/theforeman/foreman/pull/4561/files#diff-4d26af5acbc256436626e7dbe476e3deR73 but we also need to define puppet_proxy_hostname_id. Existing relations changes should not be necessary. Things like this should not be necessary https://github.com/theforeman/foreman/pull/4561/files#diff-2d53894341fd1cf84a823e3bd0855495R566

This is because Hosts are now assigned a Hostname instead of a Proxy, having a Host be assigned a Proxy and a Hostname is very weird IMO, imagine this scenario.. 

Proxy1 & Proxy2 with the following Hostnames proxy1.example.com, proxy2.example.com & proxy.example.com. I want my Host to use proxy.example.com (the shared Hostname) What Proxy do I select? It isn't clear and IMO almost hacky to tell users "just select anyone", although I understand your point about less risk, I think we can mitigate that by testing this PR properly and aiming to merging this soon after 1.17 branching so we'd have more time to find and fix any issues before 1.18. Would that be okay?

ares:
>> I think using one object for two different things usually bites later when these two things start to have different requirements.

>I could see how creating a 'hacky' link between things that aren't really linked could bite later, but I cant see how linking something by the thing they have in common could ever be an issue.
EDIT: could you explain how this could become an issue later?

If a hostname would be the only thing that is shared, then it's fine. It seems to me you were suggesting hostname would be the object that handles the load balancing too. Like it would be responsible for doing the whole orchestration on all linked proxies. And that I think this is an issue as the Hostname should be just name of the host. In ideal world, every object has only one responsibility. In real world it usually has more. The smaller number of responsibilities, the better. Look at our host god object and how we struggle with breaking it apart with facets. Load balancing is a different thing than specifying multiple hostnames to single proxy.

Until we have clear understanding of what full load balancing support would mean, I think we should avoid mixing it into multi-homing PR. If you say one day we would add $x and $y, I say let's keep these changes for when we start adding $x and $y, even if it means changing what we add by this PR.

>> If we want to implement load balancing, I'd much rather prefer to have a Load balancer as a first class citizen object, with proper permissions, taxonomy, orchestration on all linked SPs etc. This logic does not seem to belong to something called Hostname.

> Creating another object to link them by is IMO not necessary and will only add complexity for users and the codebase.

If we ever agree that Foreman should handle this and do the orchestration of all  load balanced proxies, then I believe we need a new object for this. Otherwise you're mixing that into Hostname. There were concerns raised that Foreman should not handle this. And if it's not being added it in this PR, it should not be influencing this PR.

> If you feel strongly about this do you want to take this to foreman-dev to get some more views?

If we don't agree here and other reviewer won't thumbs up the current way, then I think we need to start designing it from scratch with more devs involved from the beginning and yes, the foreman-dev is the place where to do it.

> This is because Hosts are now assigned a Hostname instead of a Proxy, having a Host be assigned a Proxy and a Hostname is very weird IMO, imagine this scenario..

As a user I'm can much better imagine assigning a proxy than a hostname. After I assign a proxy I can choose a hostname that's available only for this proxy. Otherwise if I have let's say 10 proxies, each with 2 different names, I see 20 hostnames in a selector. I can't even give hostname a name, while I can name my proxies as I want. I don;t understand on what's weird on this workflow, you first choose a proxy and then, only if proxy has more hostnames available, you chose the hostname from list which is limited only to hostnames available for this specific proxy.

> Proxy1 & Proxy2 with the following Hostnames proxy1.example.com, proxy2.example.com & proxy.example.com. I want my Host to use proxy.example.com (the shared Hostname) What Proxy do I select?

You keep arguing with load balance setup. This PR does not implement it, so I think it's irrelevant. Let's focus on multi name feature where I think it makes perfect sense to choose which proxy you want to use for the host and then select on of its hostname if you don't want to use the default one.

If load balancing use case, which is I believe what you described, would benefit from a different workflow, it only confirms these should be two different concepts.

>  I understand your point about less risk, I think we can mitigate that by testing this PR properly and aiming to merging this soon after 1.17 branching so we'd have more time to find and fix any issues before 1.18. Would that be okay?

I think we got to the point where we can't merge this in 1.17 as it is. I don't think it would be simplified to form that was requested in few days, so 1.18 is the only option. I think there were already several reviewers in this PR as well as in the original [foreman-dev](https://groups.google.com/forum/#!msg/foreman-dev/7RJ8KSZxlAw/5AcuNSSNBwAJ) thread who raised concerns about current approach. I don't think the concept has changed from the very beginning.

Sean, I'm sorry for being stubborn but I fell like I'm getting to a dead end. The only way out I see is starting from scratch and separate load balancing from multi home. If you prefer, we can get back to designing but keeping both designs separate.

sean797:
This PR focuses on Load balancing, by a side effect it also allows multi-homing (as that is a requirement of load balancing), please stop focusing on the multi-homing aspect of this, I think we all agree purely multi-homing can be achieved via another (simpler) method.

For the use case I am focusing on, we need the following:

For each Feature (this PR is just Puppet & Puppet CA) a Host is assigned is an object, based on the SPs associated with that object orchestration happens on them (if applicable for that feature). Also that object has a hostname attribute that is == Host.#{feature}. Right now that object is called Hostnames, would it make sense to call that object Loadbalancers?

I can add more commits to do the orchestration part if you like? I would rather do it in another PR so not to inpede with the progress of this PR, but I can make those changes here if required.

I'm not sure I'm sold with calling it Loadbalancer as anyone wanting to do just mutli-homing would create a Loadbalancer with 1 SP assigned, which I think is a little odd as form the uses view. But just cosmetic. (again, multi-homing, is not my primary focus, but I think we should this think about it)

> I think there were already several reviewers in this PR as well as in the original foreman-dev thread who raised concerns about current approach. I don't think the concept has changed from the very beginning.

This approach has changed considerable from the very beginning, see https://github.com/theforeman/foreman/pull/4346 that was just allowing for mutli-homing, but @stbenjam convinced me about load balancing.

Lots of people have raised concerns, mostly around complexity and risk of change, (forgive me anyone that feels that isn't accurate and raise them again please) but no one has given any other suggestions for another way of doing this, either on the PR or foreman-dev, if you have a concern its good to show that but just saying "lets do load balancing another time" or "it can be done another way" without any suggestions is actually really frustrating.

> Sean, I'm sorry for being stubborn but I fell like I'm getting to a dead end. The only way out I see is starting from scratch and separate load balancing from multi home. If you prefer, we can get back to designing but keeping both designs separate.

But these 2 things are not separate, load balancing requires "multi homing support", as clients need to use a different name than the SPs FQDNs. But not every implementation of multi homing will allow for load balancing. Please just think of this as load balancing, its only a side effect that mutli homing will also work.

ares:
> This PR focuses on Load balancing, by a side effect it also allows multi-homing (as that is a requirement of load balancing), please stop focusing on the multi-homing aspect of this, I think we all agree purely multi-homing can be achieved via another (simpler) method.

Ok, then I'll describe how I think load balancing should be implemented and won't concentrate on multihoming. Sorry, from the original discussion, I though you were mainly looking into multihome support. This PR does not contain the orchestration so the load balancing would not work.

> For each Feature (this PR is just Puppet & Puppet CA) a Host is assigned is an object, based on the SPs associated with that object orchestration happens on them (if applicable for that feature). Also that object has a hostname attribute that is == Host.#{feature}. Right now that object is called Hostnames, would it make sense to call that object Loadbalancers?

The orchestration is not implemented here. Before you add it, I think simpler approach for this should be used. We don't need separate LoadBalancer object or Hostname object for this I believe. See below my suggestion.

> I can add more commits to do the orchestration part if you like? I would rather do it in another PR so not to inpede with the progress of this PR, but I can make those changes here if required.

I think we should only merge complete features, regardless of their size. Ideally have a reviewer with which you consult the PR in smaller steps, now you have one :-)

> I'm not sure I'm sold with calling it Loadbalancer as anyone wanting to do just mutli-homing would create a Loadbalancer with 1 SP assigned, which I think is a little odd as form the uses view. But just cosmetic. (again, multi-homing, is not my primary focus, but I think we should this think about it)

It won't be needed and for multihoming we'll still have hostnames.

> but no one has given any other suggestions for another way of doing this, either on the PR or foreman-dev, if you have a concern its good to show that but just saying "lets do load balancing another time" or "it can be done another way" without any suggestions is actually really frustrating.

I'm sorry for contributing to the frustration, but I hope my comments contains suggestions. If not, I'm happy to do a hangout session where I describe better alternative approach.

> But these 2 things are not separate, load balancing requires "multi homing support", as clients need to use a different name than the SPs FQDNs. But not every implementation of multi homing will allow for load balancing. Please just think of this as load balancing, its only a side effect that mutli homing will also work.

Now I'll try to explain how both could be implemented and at the same time the overall change would be less complex and at least for me, easier to understand and use.

**Load balancing**

Hosts would still be linked to smart proxy as they are today. Smart proxy could be marked as a load balancer (just a flag). If it is marked as a load balancer, user would need to link it to other smart proxies that this is a load balancer for. That way, all read operations still work, as the load balancer would behave as any other smart proxy. User would still select a proxy, which in fact would be a load balancer. At the same time they could select also non-balancer proxy. All orchestration operations would have extra condition, `if proxy.load_balanced?` and in that case it would do the operation on all `proxy.load_balanced_proxies` instead of itself. That way the same orchestration code is run but it would just be wrapped in `.each` block.

Pros:

- reusing existing test connection in smart proxy form
- reusing feature detection that we'd see for load balancer
- no change for users without load balancing
- no need to deal with taxonomies across multiple objects
- no need to define new permissions
- no public API changes or deprecations, no object API changes
- existing user workflow remains completely untouched until they mark some proxy as load balancer

Con:
we should be 
- careful about logs and status that we load from proxy, we can get different response every time

**Multihome**

This I tried to described earlier. There would be a list of hostnames for a given proxy. When there's at least 2 hostnames available for the proxy (be it load balancer or not), user can chose hostname for a particular feature after he/she selects this proxy. Hostname in this case does not have to be separate entity with it's own taxonomies association, permissions etc. It's all derived from associated proxy. The relation between proxy and hostname is 1:n, not n:m.

Pros:
- completely isolated from load balancing use case
- the same that I listed for load balancing

Cons:
- ?

I'm happy to explain in more details, hear your feedback, do reviews gradually if it's decided to implement it this or similar way.

sean797:
Thanks for the proposal, I think were back on track with a proper discussion 🎆  I have 1 initial question..

What is the Smart Proxy with a loadbalancer flag set?
 - the loadbalancer 
 - one (or all) of the Smart Proxies in a cluster
 - a fake Smart Proxy that doesnt really exist
 -  something else?

sideangleside:
@ares, with the proposal you have above, how do I, as a user, ensure that the proper hostname is rendered when attempting to provision? Basically, when I select a proxy when creating a new host does the list of smart-proxies include both the load-balanced and non load-balanced proxies?

ares:
@sean797, the URL of a smart proxy with load-balancer enabled flag is the load-balancer URL itself. At the same time, you can see it as a fake proxy that does not really exist. The load-balancer should have the same API as smart proxy as it just proxies all requests to some real proxy right? Also user needs to provide links to other existing proxies so we know which we will run the orchestration on. If those proxies have multiple hostnames, we should be good with primary one, since these links are there only for Foreman orchestration.

@sideangleside exactly, since the load-balancer is just another proxy, it would end up in the same list of smart proxies to select. If this is hard to distinguish, we could add "(LB)" prefix/suffix to it's name when the flag is enabled. And since it's regular proxy, it could have also multiple host names so users could chose one of host names assigned to this load-balancer smart proxy if they have more than one.

sean797:
Okay, that makes sense, my thoughts:

- That would require Foreman does some communication via the load-balancer. ideally I think we should try and limit traffic going via the load-balancer to only required client facing traffic. Some users have security restrictions were they wouldn't allow Foreman to communicate via the load-balancer. I think we should make load-balancing SPs as flexible as we can.
- As a user if I have to add my load-balancer as a Smart Proxy, I feel that is somewhat hacky.
- The user must select proxy and hostname. This is a bad user experience, the user doesn't want to select any more attributes on the Host or Hostgroup form that they otherwise wouldn't need to.
- It worsens the monolithic Host object by linking SPs and Hostname to it. To quote you:
  > our host god object and how we struggle with breaking it apart with facets
- You cannot link a Model to itself so the SPs Bi-directional looped association would have to go via another Model anyway. Very similar to what the Hostname Model does here.
- would result in more joins.
    - Your proposal: Host -> SP & SP <-(via another model)-> SP & Hostname -> SP & Host -> Hostname
    - My proposal: Host -> Hostname & Hostname <-> SPs

Effectively our proposals are somewhat similar, the main difference I see is from the user view I am creating a Hostname and associating Hosts & SPs to it. Whereas under your suggestion the Host would need to be associated to the SPs and the Hostname. You could think of my method as very similar to yours but my Hostname model as a kind of Host facet holding SP and Hostname.

The user would have to select a SP & Hostname for every feature per Host or Hostgroup they have, whereas with my proposal the user could pre define common SPs and Hostname associations with a “Hostname” and just select a “Hostname” per feature per Host or Hostgroup, its unlikely users will have lots of different combinations of SPs & Hostnames, its nice to pre-define them.

I don’t see “no change” as a large Pro, because something has always been done one way doesn’t mean it still fits current requirements. If hosts are assigned a Hostname instead of a SP that is a minor inconvenience for users (hence the deprecation) you could use that argument for any user facing change. Protecting current user work flows is not important to me, we have deprecation's for that, I don't see a benefit in protecting current workflows that make other workflows either impossible or not flexible. IMO we should aim for constant improvement.

A lot of the Pros you have listed are code related, I would like to see some more focus on the User and their experience. I also don't see the benefit of this below? 

> completely isolated from load balancing use case
 
-------------------

Reading back through the discussion I get the impression the issue with my proposal fit into 2 main categories; “Let’s not change existing users workflows for users who don’t want this” & the Hostname Model name.

I think the benefits of having a Host associated with a Hostname is massive. When you configure a client system outside of Foreman you configure it with a Hostname or FQDN to communicate with the server, this Hostname or FQDN is then associated to the server (normally via DNS) if we were to accurately represent this flow in Foreman it would look like: Host -> Hostname(FQDN) -> server(in our case SP).
Obviously the true flow is Host -> Hostname(FQDN) -> IP address -> loadbalancers(maybe) -> servers(SPs), but we do not need IP address because DNS does that for us and there is no benefit of Foreman knowing specifically about a loadbalancer, all the application cares about is that the SPs that a Host might use all hold the same state.
 
I'm open to ideas about the Model name, all it is doing is allowing multiple SPs to be linked together and hold a FQDN attribute that can be rendered in Hosts templates. Maybe `SmartProxyGroup` more accurately decribes what this `Hostname` module does?

ares:
> That would require Foreman does some communication via the load-balancer. ideally I think we should try and limit traffic going via the load-balancer to only required client facing traffic. Some users have security restrictions were they wouldn't allow Foreman to communicate via the load-balancer. I think we should make load-balancing SPs as flexible as we can.

Not sure I understand, only read traffic goes through LB. All write operations go directly to linked smart proxies. If Foreman can't talk to LB, can it talk to SP? How your approach is different in this? I think my proposal also allows to implement any orchestration to happen directly with SPs linked to a given LB but this sounds weird to me. If needed, there can always be more `if self.load_balanced?` but I think we should be a bit opinionated, we've seen how features like the unattended mode, disabled authentication, taxonomies ends up. We can't support every use case.

> As a user if I have to add my load-balancer as a Smart Proxy, I feel that is somewhat hacky.

Perhaps, but I'd really like to avoid doing STI for this or extracting the shared part to concern. Note that such LB proxy in fact behaves as regular proxy. Once can read it's version, features, fetch list of puppet classes, import subnets from it etc. From Foreman point of view, I think it is a proxy. We could simply disable some feature of this proxy that would not make sense, e.g. Logs.

> The user must select proxy and hostname. This is a bad user experience, the user doesn't want to select any more attributes on the Host or Hostgroup form that they otherwise wouldn't need to.

No, that's not true. Only if they have multihome use case as well and the proxy has at least 2 hostnames assigned. And when the proxy has multiple names, I think it's still better to first select the proxy and only then show the dropdown with the list of proxy alternative names. Otherwise you'd have a list of all available hostnames not filtered by proxy. In single plain list you would not see what hostname belongs to what proxy. We'd need at least multilevel dropdown but we don't have anything like that in Foreman.

> It worsens the monolithic Host object by linking SPs and Hostname to it. To quote you:

>> our host god object and how we struggle with breaking it apart with facets

Yes, that'd give proxy extra responsibility. The implementation, that I was thinking of though, is having a load balancer class (not model) that would contain all the LB logic. The proxy would only get the flag and relation to other proxies and that's all. E.g. `SmartProxy.first.load_balancer.balanced_proxies` would create in-memory load balancer object that would return all proxies linked from the original proxy. The same could be used in your Hostname model I guess.

> You cannot link a Model to itself so the SPs Bi-directional looped association would have to go via another Model anyway. Very similar to what the Hostname Model does here.
would result in more joins.

As I tried to explain, this would be 1:n, not n:m, there's no need for 2 LB sharing the same proxy unless I'm missing something? And you can have 1:n on self if needed.

> Your proposal: Host -> SP & SP <-(via another model)-> SP & Hostname -> SP & Host -> Hostname

No, my proposal does not need hostname for LB scenario. Alternative hostname is there only for multihome scenario. In my proposal, I'm not convinced of having hostname as a separate model even. Also I assume these tables won't have hundreds of records. So in fact it's just
`Host -> SP -> SPs` in SQL

> The user would have to select a SP & Hostname for every feature per Host or Hostgroup they have, whereas with my proposal the user could pre define common SPs and Hostname associations with a “Hostname” and just select a “Hostname” per feature per Host or Hostgroup, its unlikely users will have lots of different combinations of SPs & Hostnames, its nice to pre-define them.

I think we can have the same on host group level in my proposal. In my scenario, users who require just LB don't even have to see anything related to multihome.

> I don’t see “no change” as a large Pro, because something has always been done one way doesn’t mean it still fits current requirements. If hosts are assigned a Hostname instead of a SP that is a minor inconvenience for users (hence the deprecation) you could use that argument for any user facing change. Protecting current user work flows is not important to me, we have deprecation's for that, I don't see a benefit in protecting current workflows that make other workflows either impossible or not flexible. IMO we should aim for constant improvement.

Sorry for a bit of FUD but you suggest that change to the workflow is justified by few % of users that would benefit from it? How many load balancing and multihome RFEs have you seen in redmine?

> A lot of the Pros you have listed are code related, I would like to see some more focus on the User and their experience. I also don't see the benefit of this below?

The non code related benefits for me:

* no change for users without load balancing
* no need to deal with taxonomies across multiple objects
* no need to define new permissions

Permissions, taxonomies is definitely user experience. Keeping it to existing objects makes it much more palatable.

The code related pros are also important to me, as I'm one of people who will need to maintain the code. This is the reason why we sometimes say no to the feature, if it does not look like the feature belonging to Foreman domain.

> Reading back through the discussion I get the impression the issue with my proposal fit into 2 main categories; “Let’s not change existing users workflows for users who don’t want this” & the Hostname Model name.

1) yes, changing existing logic which is IMHO not necessary at all,
2) it's really not just the Hostname model name

If I read your answer correctly, you'd only change hostname model name in your PR, is that correct?

sean797:
> Not sure I understand, only read traffic goes through LB. All write operations go directly to linked smart proxies. If Foreman can't talk to LB, can it talk to SP? How your approach is different in this? I think my proposal also allows to implement any orchestration to happen directly with SPs linked to a given LB but this sounds weird to me. 

Yes, foreman would be required to be able to communicate to a foreman-proxy via the load-balancer for status APIs ect.. Whereas with my proposal the Hostname is never used for any communication by Foreman, only used as a way of linking SPs and to render in Templates.

>>  You cannot link a Model to itself so the SPs Bi-directional looped association would have to go via another Model anyway. Very similar to what the Hostname Model does here.
    would result in more joins.

>As I tried to explain, this would be 1:n, not n:m, there's no need for 2 LB sharing the same proxy unless I'm missing something? And you can have 1:n on self if needed.

One of us are for sure.. As i understand your proposal, if I have a LB (which is stored to as a SP in Foreman) I need to link that LB to the other SPs in that Cluster, this is 1:n in terms of 1 load-balancer to many SPs but n:m in terms of SQL as load-balancers and SPs are stored in the same SP Model.

>> Your proposal: Host -> SP & SP <-(via another model)-> SP & Hostname -> SP & Host -> Hostname

>No, my proposal does not need hostname for LB scenario. Alternative hostname is there only for multihome scenario. In my proposal, I'm not convinced of having hostname as a separate model even. Also I assume these tables won't have hundreds of records. So in fact it's just
Host -> SP -> SPs in SQL

You are missing the Hostnames for multi-homing. Its also related to the point above.

> Sorry for a bit of FUD but you suggest that change to the workflow is justified by few % of users that would benefit from it? How many load balancing and multihome RFEs have you seen in redmine?

- I regularly speak to many different users who want these features, I wouldn't have wrote this patch if I didn't know a lot of people want it. Obviously I do tend to speak to users with larger estates.
- 2 of the top 5 (by popularity) YouTube Videos are HA related https://www.youtube.com/channel/UCCo7AZ1oG6TbG0-dwjRqCmw/videos?flow=grid&view=0&sort=p
- every couple of months someone asks about HA or load-balancing (and multi-homing less often) on the mailing lists.
- Looking at last years user survey memcache is the 14th most popular plugin, I'd bet that would be higher if more plugins supported HA, also I think users generally are more interested in HA'ing SP than Foreman itself, as they are client facing.
- @GregSutcliffe have we got any stats regarding https://theforeman.org/2015/12/journey_to_high_availability.html

Anyone else have any other data sources?

> Permissions, taxonomies is definitely user experience. Keeping it to existing objects makes it much more palatable.

Arguable that is more of an issue with the current permission system.

> If I read your answer correctly, you'd only change hostname model name in your PR, is that correct?

Correct, unless you have a proposal that provides the same level of flexibility.

ares:
> Yes, foreman would be required to be able to communicate to a foreman-proxy via the load-balancer for status APIs ect.. Whereas with my proposal the Hostname is never used for any communication by Foreman, only used as a way of linking SPs and to render in Templates.

So this would be loadbalancing only from clients point of view (hosts, vms, whatever, not Foreman itself).   In your proposal, how read operations would be implemented? From which proxy we'd get puppet environments? I think in my suggestion we don't lose the flexibility, the loadbalance proxy can override all the communication flows too. But in this case, I'd reconsider whether loadbalancer should be just flag or rather it should become a child class of SmartProxy.

> One of us are for sure.. As i understand your proposal, if I have a LB (which is stored to as a SP in Foreman) I need to link that LB to the other SPs in that Cluster, this is 1:n in terms of 1 load-balancer to many SPs but n:m in terms of SQL as load-balancers and SPs are stored in the same SP Model.

`smart_proxies` table would have `load_balanced_by_id` linking LB proxy. I don't see the need for n:m table if we agree one SP is always attached to 0 or 1 LB.

> You are missing the Hostnames for multi-homing. Its also related to the point above.

that is only required in rare combination where you need to use LB and need to have a different name for it per host, if you don't setup multiple hostnames for a LB, the primary would be used (stored in the LB SP object itself)

> > Sorry for a bit of FUD but you suggest that change to the workflow is justified by few % of users that would benefit from it? How many load balancing and multihome RFEs have you seen in redmine?

> I regularly speak to many different users who want these features, I wouldn't have wrote this patch if I didn't know a lot of people want it. Obviously I do tend to speak to users with larger estates.

and of course we want to support these users use cases but at the same time without negative effects on all the rest

> 2 of the top 5 (by popularity) YouTube Videos are HA related https://www.youtube.com/channel/UCCo7AZ1oG6TbG0-dwjRqCmw/videos?flow=grid&view=0&sort=p

interesting, based on https://www.theforeman.org/2017/03/2017-foreman-survey-analysis.html#page1 and https://www.theforeman.org/2016/03/2016-forman-survey-analysis.html#diversity the chart "How many nodes do you manage with Foreman" I assume it's around 15% of users.

> every couple of months someone asks about HA or load-balancing (and multi-homing less often) on the mailing lists.
> Looking at last years user survey memcache is the 14th most popular plugin, I'd bet that would be higher if more plugins supported HA, also I think users generally are more interested in HA'ing SP than Foreman itself, as they are client facing.
> @GregSutcliffe have we got any stats regarding https://theforeman.org/2015/12/journey_to_high_availability.html

yes, as all other improvements, I agree we should implement it, it would be a great addition, I just don't want to do such a dramatic change that affects everybody until it's necessary. I think there's an alternative approach that I suggested.

> > Permissions, taxonomies is definitely user experience. Keeping it to existing objects makes it much more palatable.

> Arguable that is more of an issue with the current permission system.

Yes, but we can avoid making it worse. My design seems less prone to these problems.

> > If I read your answer correctly, you'd only change hostname model name in your PR, is that correct?

> Correct, unless you have a proposal that provides the same level of flexibility.

Perhaps I haven't understood your concerns correctly, perhaps I failed to explain my proposal clearly enough, anyways at this point I'm afraid I have to :-1: this PR. Other reviewers can out vote me of course. My concerns in short:

* changes existing objects API which are foundation for users workflows and other plugins without a need for it
* It does not implement load balancing fully, there's no orchestration code that would be required
* it's overkill for just multihoming
* it adds unnecessary complexity because it combines load balancing and multihoming features into a single new object
* I find it confusing from user perspective

If you want to continue the effort I think we should start with the design first, state clear goals and use cases we want to cover and continue with the implementation when there's some positive reaction from devs. Also the UX team should be involved so we make sure, the solution is not confusing for users.

sean797:
> So this would be loadbalancing only from clients point of view (hosts, vms, whatever, not Foreman itself).

Yes, see the original diagram I drew above, there is a link in https://github.com/theforeman/foreman/pull/4561#issuecomment-305520088
[here](https://www.draw.io/?lightbox=1&highlight=0000ff&edit=_blank&layers=1&nav=1&title=Untitled%20Diagram.xml#R7VnbcpswEP0az7QPySAE2H603aTtTDOTqWd6eZRBAbUCUSFiu19fCQRYFrGdOBenbR4ctCxC2rPn7MoewFm6es9RnlyxCNOB60SrAXw3cF3gQUf%2BU5Z1bRk6o9oQcxJpp84wJ7%2BxNurn4pJEuDAcBWNUkNw0hizLcCgMG%2BKcLU23G0bNt%2BYoxpZhHiJqW7%2BSSCS1deQGnf0DJnHSvBkE4%2FrOAoU%2FY87KTL9v4MKb6q%2B%2BnaJmLr3RIkERW26Y4MUAzjhjor5KVzNMVWybsNXPXd5xt103x5k45AG3fuAW0VJv%2FZJxnKJML0%2Bsm5AsEyLwPEehGi8l7AM4TURK5QjISz0P5gKv7lwLaHcoMwezFAu%2Bli76ARfq1aybZNBBWnYQuB6sbclG%2BMfaD2nU43bqbufyQm%2B%2BPxBja8M4kimhh4yLhMUsQ%2FSis04rnLGawDGDITfM19%2B0vRp8V4NzXw5%2FYCHWOuNRKZg0dbN%2FYiw3AqoWsTuccs2s5KH20tERiMdYbEBsB51jigS5NWfvi2D16IRztN5wyBnJRLEx87UydFgC3zOx9PwtNOoZO2zapR0EF7AT97rMc6yW8BnnMqKFBajMTGHixHFBfqNF5aCg0ruS3v504L%2BTFkRJnElDKIOOuTSoDCdSKCb6RkqiqEoGihaYTlv6zxhlvHpvIwA7OKJlTK%2BkU4dNsMc7ueOcO54%2FNEKuAbgfzh2QjQu7uSlUWLeodD%2B4enSmhWtGCa7eOJ9%2FUkNcYSfXnVMZaEGY1CLHAvMIdhaSHWKiqoQClqKiIGFjviS0I3FkO0njhsuzc9l7CS7DkWMkFgy2asr9%2FI%2FmPrRyaZ5K6KTpmrOV2pmMQEAV1xdcXsXqqk23K1QoJquUClCqqlj9qVJvYqcZpbLnUOn09PVvO2zALn%2FAG9nlr62Tx9Q%2FYFNsf3kDDy9v%2B%2Fl7AP8exivvRGqkPzYBf%2FQaCf61Ggmcfhz1TGfOORgDs0qewRMqk95eZXNfpbL57gHSpqvEE0ib92TdQ8YyvNU6OKfQOgS2xA2PlLiDo%2B3vaPWQQJTFlf78KnEhqmS9rLs9xsXfJkd3tGsbcjQyxQickBgFFo5Np36nkhQJypUxYqWMy8UzKow7hkYkA9ijMG6PwoweQ2CCVyww3XcVVf%2FWtnMPF5%2BhLT4vcm7xtqqOB43%2ByvIPnJ3%2Bx%2Fdjw%2F%2FK2GRNsJPNUhhB07020uiekDSCR%2F228FV0FD2khi9B6u1T8j5Sg2C0y%2F9oUrs7Dln%2FGqn3fUXpGkj4J0RoW5m%2FEC5KpHw%2BXttnrjcfJs2JLKSlPnalaL3Aby1QO96DZ2mFtjPe7zlswWFPKxTcvxWSw%2B63qjru3Q%2BC8OIP) it is again.

> In your proposal, how read operations would be implemented? From which proxy we'd get puppet environments?

No change, see this PR, nothing has changed there. AFAICT Foreman currently has no concept of Puppet Environments -> SP mapping, maybe we should do that, for now that would be one of the caveats of setting up your cluster. In Katello this wouldn't be an issue as we can ensure all SPs in a Hostname have the same LifeCycle env and therefore the same Puppet environments.
I already check Certificates and Features.


> smart_proxies table would have load_balanced_by_id linking LB proxy. I don't see the need for n:m table if we agree one SP is always attached to 0 or 1 LB.

That isn't using AR relations is it? If so, a diagram explaining them would be great.

> and of course we want to support these users use cases but at the same time without negative effects on all the rest

I don't agree with most the negative impacts of accurately representing the flow (as I describe at the end of https://github.com/theforeman/foreman/pull/4561#issuecomment-348925149) users are use to outside of Foreman. I would go so far as to say having Hosts assigned to SPs as it doesn't represent that flow is confusing to new users.

> Yes, but we can avoid making it worse. My design seems less prone to these problems.

I don't think we should be forced to implement less flexible and more confusing (at least for the user) implementations because of other issues. We should fix those issues at the source, we shouldn't be forced to workaround them to the detriment of the application.

I don't understand how you can be against the API changes here but are for the interface -> network -> subnet API change in https://github.com/theforeman/foreman/pull/5017#issuecomment-349403711 ?
They are effectively the same thing, adding an extra model in the flow.
interface -> network -> subnet
host -> hostname -> smart proxy

> It does not implement load balancing fully, there's no orchestration code that would be required

I said I would add it in this PR if required. Not valid.

> it's overkill for just multihoming

I said many times its not just multihoming. Not valid.

> it adds unnecessary complexity because it combines load balancing and multihoming features into a single new object

I don't understand how that is a problem? It's effectively using the DRY principle.

> I find it confusing from user perspective

Can you explain exactly why? Sorry if you already have and I missed it.


---------

Okay, I guess we are not going to agree. Would you be able to draw a small diagram explaining your model proposal? and accurately explain the users workflow for clusters, mutihomed and normal SPs and adding a Host with each one.

I'll do the same of my proposal, then write a draft email for foreman-dev and send it to you before hand, I want to make sure I accurately represent your proposal and views.

ares:
> I don't understand how you can be against the API changes here but are for the interface -> network -> subnet API change in #5017 (comment) ?
> They are effectively the same thing, adding an extra model in the flow.
> interface -> network -> subnet
> host -> hostname -> smart proxy

Then I failed in my explanations. In the linked comment I explicitly point out the new requirement of users setting up the network with vlan/mtu even if they don't need to adjust these values. The solution I suggested there is to provide default network so users would not be affected, the default network would be automatically assigned to existing objects. In your proposal, users are enforced to start using hostnames even in case they are not interested in load balancing or multihoming. Also subnet is not such exposed object as a host and proxy. Changes to these are more sensitive, or at least that's my opinion.

For the rest, I will try to provide more detailed write up of what I suggested but I'm afraid I won't make it before New Year.

sean797:
> The solution I suggested there is to provide default network so users would not be affected, the default network would be automatically assigned to existing objects. 

Which is very similar to this PR, in that this also a default Hostname of every Smart Proxy based off the URL. As I understand it most of the points you have put forwarded against this proposal could be used against that Network one also.

> For the rest, I will try to provide more detailed write up of what I suggested but I'm afraid I won't make it before New Year.

Great, that's perfect, thanks! 

Also thanks for taking the time to look into this thoroughly, although we don't agree, I appreciate the time you have invested in this!

sean797:
I've updated this PR to reflect the out come of the [vote](https://community.theforeman.org/t/vote-naming-the-new-smart-proxies-route-group-object/8560) tests pass apart from a mysql migration issue (I'll work out what the issue is soon). So i think this is ready for an initial review :+1: 

Right now this PR does the following:
Creates the new SmartProxyPool model
Makes it possible to use SmartProxyPool for Puppet

What else do we need to do to get the merged? Do we want support for more features in this PR? IMHO we should do them in another PR, this one it already pretty big, though happy to do whatever :-)

sean797:
Closing this since I haven't really got the time (or energy) to spend pushing this forward, convincing and re-convincing people this is the correct approach. If someone else would like to pick this up, I would be forever grateful.

Thanks for all the time everyone has put into the discussions around this. Especially @ares @lzap @iNecas.

lzap:
Hopefully some of your work from this PR can be used in upcoming work which is I hope going to happen.

PR #4563, opened 2017-06-01
Fixes #19728 - set admin user in rake console

lzap:
We have a integration tests that checks for login screen we should be fine. Let's double check tho.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -227,6 +227,7 @@ module Foreman
 
   def self.setup_console
     ENV['IRBRC'] = File.expand_path('../irbrc', __FILE__)
-    puts "For some operations a user must be set, try User.current = User.first"
+    User.current = User.anonymous_console_admin
+    Rails.logger.warn "Console started with '#{User.current.login}' user, call User.current= to change it"
   end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -16,6 +16,7 @@ class User < ApplicationRecord
 
   ANONYMOUS_ADMIN = 'foreman_admin'
   ANONYMOUS_API_ADMIN = 'foreman_api_admin'
+  ANONYMOUS_CONSOLE_ADMIN = 'foreman_console_admin'
 
   validates_lengths_from_database :except => [:firstname, :lastname, :format, :mail, :login]
   attr_accessor :password, :password_confirmation, :current_password
@@ -186,6 +187,10 @@ class User < ApplicationRecord
 
   def self.anonymous_api_admin
     unscoped.find_by_login(ANONYMOUS_API_ADMIN) || raise(Foreman::Exception.new(N_("Anonymous admin user %s is missing, run foreman-rake db:seed"), ANONYMOUS_API_ADMIN))
+  end
+
+  def self.anonymous_console_admin
+    unscoped.find_by_login(ANONYMOUS_CONSOLE_ADMIN) || raise(Foreman::Exception.new(N_("Anonymous admin user %s is missing, run foreman-rake db:seed"), ANONYMOUS_CONSOLE_ADMIN))
   end
 
   # Tries to find the user in the DB and then authenticate against their authentication source
@@ -644,7 +649,7 @@ class User < ApplicationRecord
   end
 
   def hidden_authsource_restricted
-    if auth_source_id_changed? && hidden? && ![ANONYMOUS_ADMIN, ANONYMOUS_API_ADMIN].include?(self.login)
+    if auth_source_id_changed? && hidden? && ![ANONYMOUS_ADMIN, ANONYMOUS_API_ADMIN, ANONYMOUS_CONSOLE_ADMIN].include?(self.login)
       errors.add :auth_source, _("is not permitted")
     end
   end
--- a/db/seeds.d/04-admin.rb
+++ b/db/seeds.d/04-admin.rb
@@ -15,6 +15,17 @@ unless User.unscoped.find_by_login(User::ANONYMOUS_ADMIN).present?
   end
 end
 
+# Anonymous Console Admin is used for console commands etc.
+unless User.unscoped.find_by_login(User::ANONYMOUS_CONSOLE_ADMIN).present?
+  User.without_auditing do
+    user = User.new(:login => User::ANONYMOUS_CONSOLE_ADMIN, :firstname => "Console", :lastname => "Admin")
+    user.admin = true
+    user.auth_source = src_hidden
+    User.current = user
+    raise "Unable to create anonymous console admin user: #{format_errors user}" unless user.save
+  end
+end
+
 # Anonymous API user is used for API access when oauth_map_users is disabled
 # It should be removed and replaced by per-user OAuth tokens (#1301)
 unless User.unscoped.find_by_login(User::ANONYMOUS_API_ADMIN).present?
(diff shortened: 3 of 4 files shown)


ares:
:+1: from me, but I'd like to hear from others too. This solves the problem with taxonomix default scope that disallow to find objects until user is set. Also since the user is admin, we can see objects that are not assigned to any org/loc.

lzap:
Added an extra console_admin for recording purposes.

dLobatog (changes requested):
👍 from me too. I don't really think adding another anonymous admin is a good idea though. It complicates the model a little bit more and introduces another seed for little benefit. Why is `foreman_admin` not good enough for logging purposes?

config/application.rb
```diff
@@ -227,6 +227,7 @@ class Application < Rails::Application
 
   def self.setup_console
     ENV['IRBRC'] = File.expand_path('../irbrc', __FILE__)
-    puts "For some operations a user must be set, try User.current = User.first"
+    User.current = User.anonymous_console_admin
+    puts "Current user is #{User.current.login}, call User.current= to change it"
```
ohadlevy: could we also add a warning here? using this might break your warranty kind of warning? :)

ohadlevy:
@dLobatog I think its better to have a seperate user, if its end up being used by a console session (vs a rake task that we use in cron) I woud like to see that in the audit log.

Maybe we should consider saving the irb history too (and in foreman-debug).

dLobatog (approved):

dLobatog:
I didn't know console was supported in any kind of way, other than just a tool for developers. If it's not supported, anything you do with the console is basically the users' risk and they should know at least to set the user. In any case I'm fine with the change, just wanted to make it as small as possible.

ares:
I don't think console would be supported, but [1.15 changed](https://github.com/theforeman/foreman/pull/3961) the way how console worked before, so for advanced users, I think it would be good to fix in 1.15.z.

lzap:
@dLobatog unfortunately console is the essential tool when debugging katello and there is no reason to hide it or make the experience worse. I rebased and added Rails WARNING log instead.

lzap:
@dLobatog proposing 1.15 backport, this is regression that is difficult to spot (records do not appear in the results when taxonomy is set).

dLobatog:
I do agree the console is useful from the developers perspective (the majority of our users are not even aware of it in my experience - understandably so as it's not even part of Foreman).

> this is regression that is difficult to spot (records do not appear in the results when taxonomy is set).

How can it be a regression when the console itself isn't even a feature meant to be used by Foreman users? It's clearly stated in the docs how to use it (https://theforeman.org/manuals/1.15/#5.6RailsConsole) and set the user. 

I don't think it's a good idea to add yet another admin user to a version that's already released, another migration, especially for something that isn't broken or meant to be used by anyone other than developers even.

Maybe I'm missing something here, which workflows for Foreman users (this includes admins who deploy Foreman) are broken due to not having the console user set by default?

dLobatog:
Either there are more failures on Katello than usual or this PR is causing them, I see 16  http://ci.theforeman.org/job/test_develop_pr_katello/2893/#showFailuresLink vs the usual for the last few days 11

lzap:
Daniel, what's problematic is the wrong behaviour. If you forget to set the current user, with the CVE we fixed in taxonomy few months ago, you suddenly experience hidden records which do not belong to "no taxonomy". It's the behaviour that is misleading, when you are hunting a bug (imagine a person who is giving remote support), this can lead to bad decisions and ultimately bad experience overall.

Please give us strong argument against not creating another admin user, these accounts are locked and it was only introduced for recording purposes - we are able to tell who was doing what on a console (and when).

I don't believe these are relevant, will rebase on top of https://github.com/Katello/katello/pull/6807 once merged.

dLobatog:
On 06/04, Lukáš Zapletal wrote:
> Daniel, what's problematic is the wrong behaviour. If you forget to set the current user, with the CVE we fixed in taxonomy few months ago, you suddenly experience hidden records which do not belong to "no taxonomy". It's the behaviour that is misleading, when you are hunting a bug (imagine a person who is giving remote support), this can lead to bad decisions and ultimately bad experience overall.

I am not sure what records you mean by that. The way taxonomies are
supposed to work in Foreman is:

* Taxable object has taxonomies (e.g: smart proxy assigned to org/loc)
  - Admins see it
  - Users who belong in org/loc (*both*) see it

* Taxable object has no taxonomies assigned (e.g: smart proxy without
  org/loc)
  - Admins see it
  - Users do NOT see it

* Not taxable objects such as Puppet classes can be seen by anyone with
  the permission.

Which of these categories do the "hidden objects" you refer to belong
to? If you have *no user set* at all, you shouldn't be able to see
anything basically (if SETTINGS[:login] == true)

> Please give us strong argument against not creating another admin user, these accounts are locked and it was only introduced for recording purposes - we are able to tell who was doing what on a console (and when).

I don't have a strong argument against it, I think this should be merged
once tests pass. I just don't think it's a regression worth risking any
more unstability on 1.15.1, the docs explain how it should have to set
the user, and explicitely say using the console is dangerous and
shouldn't be done. Only devs are (a little bit ) inconvenienced by
having to set `User.current`, and possibly some advanced user.
That's why I think it's worth merging but not cherry-picking to 1.15.1

> I don't believe these are relevant, will rebase on top of https://github.com/Katello/katello/pull/6807 once merged.

The first 5 tests are not fixed by that PR, they're either newly
introduced in katello/katello or by this PR

  Actions::Katello::Repository::ImportApplicabilityTest.test_0001_runs
  Actions::Pulp::Repository::BackgroundSyncTest.test_sync
  Actions::Pulp::Repository::UpdateScheduleTest.test_update_schedule
  Actions::Pulp::Repository::UpdateScheduleTest.test_disable_schedule
  Katello::Host::RecalculateErrataStatusTest::run.test_0001_plans

>
> --
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/theforeman/foreman/pull/4563#issuecomment-306116707

--
Daniel Lobato Garcia

@dLobatog
blog.daniellobato.me
daniellobato.me

GPG: http://keys.gnupg.net/pks/lookup?op=get&search=0x7A92D6DD38D6DE30
Keybase: https://keybase.io/elobato

lzap:
> Which of these categories do the "hidden objects" you refer to belong
> to? If you have *no user set* at all, you shouldn't be able to see
> anything basically (if SETTINGS[:login] == true)

Consider this typical debugging session:

```
[14] pry(main)> User.current
=> nil
[15] pry(main)> Medium.all.count
=> 0
```

Wait, I don't see any media? What's wrong? Now the user who was already angry when something was not working gets mad. I don't know what is going on here to be honest, some of my media has a taxonomy set, the rest do not. But I still see *nothing* as "none" user.

```
[16] pry(main)> Medium.unscoped.all.count
=> 10
[20] pry(main)> Medium.unscoped.all.collect(&:organization_ids)
=> [[], [], [], [], [], [], [], [], [1], [1]]
```

When admin is set, I always see everything. Plus we have some audit data just for case.

I agree that you don't want this in 1.15.1, how about to make a change (I will do a PR) that will set the current user to standard admin there so there is no migration/seed needed? That will fix this weird regression. Thanks.

lzap:
Opened up https://github.com/theforeman/foreman/pull/4565 as a backport.

lzap:
Rebased, no changes just to trigger tests again. What you think of the split @dLobatog ?

dLobatog:
Error in Katello is solved by https://github.com/Katello/katello/pull/6811 , error in Foreman is unfortunately the usual integration test.

Merging for 1.16, thanks @lzap and @ohadlevy @ares  for joining the debate

PR #4567, opened 2017-06-05
Fixes #19922 - Unpin gettext_i18n_rails

tbrisker:


dLobatog:
Thanks @tbrisker , the Katello failure is fixed here https://github.com/Katello/katello/pull/6811

Note if you want a newer version of the gem (is there any feature you were looking for?), you should update it in https://github.com/theforeman/foreman-packaging/tree/rpm/develop as it's 1.1.0

PR #4569, opened 2017-06-05
Fixes #19930: Skip adding permissions to default_roles during migrate

ehelms:
This breaks during upgrades from my testing with foreman-tasks available. There is the possibility this can be called when the Role or Filter model have not been fully migrated and a plugin that calls the function during initialization triggers an error.

I'd like to have this considered for 1.15.1 to unblock some user upgrade scenarios.

### DIFF
--- a/app/registries/foreman/plugin/rbac_support.rb
+++ b/app/registries/foreman/plugin/rbac_support.rb
@@ -2,6 +2,7 @@ module Foreman
   class Plugin
     class RbacSupport
       def add_all_permissions_to_default_roles(all_permissions)
+        return if Foreman.in_rake?('db:migrate')
         view_permissions = all_permissions.where("name LIKE :name", :name => "view_%")
         Role.transaction do
           add_all_permissions_to_role("Manager", all_permissions)


ehelms:
@ares Mind having a look?

dLobatog:
@ehelms Thanks for the PR, I'm happy to include it on 1.15.1 if it's breaking any upgrades. What do you mean by 

>  Role or Filter model have not been fully migrated 

?

How would I reproduce this?

ehelms:
An example of this, is to go back in migration time to just before (https://github.com/theforeman/foreman/blob/develop/db/migrate/20160818091420_add_override_flag_to_filter.rb) and ensure you have latest foreman-tasks (e.g. 0.9.1) that includes https://github.com/theforeman/foreman-tasks/blob/master/lib/foreman_tasks/engine.rb#L73 such that it is called during plugin initialization. It will call code that relies on the database attribute to be present but is not yet. While this is an older example that happens only during some upgrades, this could happen again.

dLobatog:
Thanks @ehelms !

ares:
@dLobatog did/will this get to 1.15-stable?

PR #4570, opened 2017-06-06
Fixes #19677 - Add settings for local boot templates

xprazak2:


app/models/concerns/orchestration/tftp.rb
```diff
@@ -209,4 +208,10 @@ def each_unique_feasible_tftp_proxy
     end
     results.all?
   end
+
+  def local_boot_template_name(kind)
+    key = "local_boot_#{kind}"
+    host_param = HostParameter.find_by(:reference_id => host.id, :name => key)
```
dLobatog: Best to look in `host.host_params`, in order to get all possible values (from hostgroup, global parameters, inherited parameters, etc...)

test/models/orchestration/tftp_test.rb
```diff
@@ -230,6 +230,37 @@ class TFTPOrchestrationTest < ActiveSupport::TestCase
     end
   end
 
+  test "generate_default_pxe_template_for_pxelinux_localboot_from_setting" do
+    if unattended?
```
dLobatog: Use an early return, `return unless unattended?` to avoid nesting the whole test, feel free to change other tests in this file

test/fixtures/settings.yml
```diff
@@ -348,3 +348,8 @@ attribute72:
   category: Setting::Provisioning
   default: nil
   description: 'Default host owner'
+attribute73:
+  name: local_boot_PXELinux
+  category: Setting::Provisioning
+  default: nil
```
dLobatog: Isn't `PXELinux default local boot` actually the default?

dLobatog (changes requested):
@lzap please take a look as you requested this feature. I think it looks fairly sane.  

I'm not sure about inclusion in 1.15 at this point, the feature is the same as 1.14 (no regressions), it's just that the template is locked. Am I missing something?

Thanks @xprazak2 !

xprazak2:
I changed tests so that they use early return statement, moved the default value into the setting and I fetch the parameter from `host.host_params`

dLobatog (approved):

dLobatog:
👍 to merge when @lzap confirms. I see 6 failures on Katello now, but I think I saw that before on another PR yesterday.. time to fix them.

app/models/concerns/orchestration/tftp.rb
```diff
@@ -209,4 +208,9 @@ def each_unique_feasible_tftp_proxy
     end
     results.all?
   end
+
+  def local_boot_template_name(kind)
+    key = "local_boot_#{kind}"
```
lzap, line 213: Nitpick: I'd prefer `kind.downcase` here but it's not exposed to the user not big deal. Only if you plan a rebase. The same for global name as well.

lzap:
It looks like we attempt to get templates_url from proxy during tests. Not sure if this is in `develop` or you introduced it, but can you fix it as well since you touch this part already?

```
RestClient::Exceptions::OpenTimeout: Timed out connecting to server
    lib/proxy_api/resource.rb:67:in `get'
    lib/proxy_api/template.rb:10:in `template_url'
    lib/foreman/foreman_url_renderer.rb:60:in `foreman_url_from_templates_proxy'
    lib/foreman/foreman_url_renderer.rb:24:in `foreman_url'
    test/unit/foreman/renderer_test.rb:85:in `block in <class:RendererTest>' (RestClient::Exceptions::OpenTimeout)
/usr/local/rvm/gems/ruby-2.2.5@test_develop_pr_core-1/gems/rest-client-2.0.2/lib/restclient/request.rb:731
```

We definitely need to stub `foreman_url` in all tests.

xprazak2:
I use `kind.downcase` for both local boot and global default templates, call to template proxy is stubbed in tests.

test/fixtures/settings.yml
```diff
@@ -348,3 +348,8 @@ attribute72:
   category: Setting::Provisioning
   default: nil
   description: 'Default host owner'
+attribute73:
+  name: local_boot_PXELinux
```
lzap, line 352: You said you use downacase but there's still mixed case key here, I was hoping we get rid of this completely. I believe all or most of our settings are downcase only.

lzap (approved):
But other than that I have no comments, not sure if I find time today to test this tho.

lzap:
I have one issue - after migration the default value for global default template is not set (is blank). It is supposed to be set to "PXELinux global default" (as here https://github.com/theforeman/community-templates/blob/develop/provisioning_templates/PXELinux/pxelinux_global_default.erb - this is from develop and 1.15).

Edit: You have `nil` for the default value, we need to have it set correctly.

lzap:
Another issue - when I try to set "PXELinux global default" in "Global default PXELinux template" I am getting this error:

Value is invalid. No provisioning template with name "PXELinux global default" and kind "pxelinux" found. Consult "Provisioning Templates" page to see what templates are available.

lzap:
I will be wrapping up shortly, I have two comments which are blockers, re-test will be necessary. The code looks good tho @ares

xprazak2:
Settings for both local boot and global templates are in downcase, there should be no more validation errors.

xprazak2:
I tried to address the comments about setting names in downcase and empty default values, but I reverted to my original solution. Here is why:

The settings are set correctly to a default value of `nil` after migration, because we validate if selected template is really in our database. If you try to set the default value to a template name, you have chicken-and-egg problem, because Settings are autocreated on server start. You will not be able to run seeds, because it loads the engine and validation fails, but you need to run the seeds so that validation passes.

I agree that all setting names should be in snake_case, but I see no point of changing the names since this would make us inconsistent with [Katello](https://github.com/Katello/katello/blob/master/app/models/setting/content.rb#L21) and the not even [Foreman](https://github.com/theforeman/foreman/blob/develop/app/models/setting/puppet.rb#L10-L12) sticks to the naming convention.

xprazak2:
Unrelated failures in js tests.

ares:
@lzap does the [comment](https://github.com/theforeman/foreman/pull/4570#issuecomment-310981236) address your requirements?

lzap:
Hmmm I don't believe this works properly.

1) Delete all settings in console
2) Boot up the server, new settings are being created.

Problem A: Head over to Settings - Provisioning, what I see is all settings named `Global default XYZ template` are empty (literally blank). I believe this won't work out of box, we need to make sure it's set automatically for new deployments.

Problem B: Try to set a global default template, getting: Value is invalid. No provisioning template with name "PXEGrub global default" and kind "pxegrub" found. Consult "Provisioning Templates" page to see what templates are available.

lzap:
Wait, looks like I was re-testing commit fc494ea796816279c16a6955bfe9c06e851687c1 lemme see, some git issues over here.

lzap:
Ok problem B solved, but I still see Problem A, but let me explain in more detail.

In your design, you allow blanks for all new settings and if the setting is blank you fallback to default. I think this is confusing for users - what does blank value mean? Even if we document it and modify description, this will still be confusing.

Therefore I suggest one more change - please redesign it to not allow blanks and make sure the correct default values are set using the default value of the Settings constructor. Since templates are now locked, we can rely on fact that they are present and valid so we can set them.

ares:
I 100% agree with empty setting value for default is confusing. On the other hand, it's consistent with the behavior of global pxe templates. @lzap @xprazak2, would it be better to fix it in a separate PR for all such settings?

xprazak2:
The blank value as default is really not great. I suppose I could disable validation if I detect the default value, for global pxe templates as well.

xprazak2:
I added one commit that sets correct templates as default values. Let me know if I should squash or move it to a separate PR.

ares:
It seems tests are broken by the change PxeTemplateNameValidatorTest.test_0001_should be valid when empty

```
Error Message

Expected false to be truthy.
Stacktrace

Expected false to be truthy. (Minitest::Assertion)
/var/lib/workspace/workspace/test_develop_pr_core/database/postgresql/ruby/2.4/slave/fast/test/unit/validators/pxe_template_name_validator_test.rb:15    
```

otherwise I think the change is small enough to be part of this PR, I think it would be good for @lzap to keep it as separate commit for easier review and he could perhaps squash on merge

lzap:
Feel free to amend everything, I don't mind. Thanks.

test/unit/validators/pxe_template_name_validator_test.rb
```diff
@@ -25,4 +25,11 @@ class Validatable
     @item.value = template.name
     assert @item.valid?
   end
+
+  test "should be valid even if default template does not exist" do
+    name = "PXELinux default local boot"
+    Template.where(:name => name).map(&:destroy)
```
lzap: This is little bit clunky, why not to use destroy_all?

lzap (changes requested):
One nitpick, the blank value is still selectable, if you select in the settings, Foreman correctly errors out with "￼￼￼Value is invalid. No provisioning template with ...". Can you perhaps get rid of the blank value completely?

Please make new commit so I can review it separately. I will squash anyway. Thanks. Other than that and the other nitpick we are good to go, great.

lzap:
This should prevent the PR spam: https://github.com/theforeman/prprocessor/pull/64

xprazak2:
I removed the blank values completely and used `destroy_all` in test.

lzap:
Merged thanks! @xprazak2

ares:
thanks for merging @lzap, could you please set the "Release" field in redmine? I suppose this lands in 1.16

ares:
nvm, it has been set by Daniel meanwhile, 1.15.3 it is

lzap:
I am terrible in this, sorry.

PR #4571, opened 2017-06-06
refs #19589 - foreman can now start in production

ohadlevy:
this avoids the following error in production (under passenger):

 Message from application: uninitialized constant Generators::Base (NameError)
 lib/generators/plugin/migration_generator.rb:2:in `<module:Plugin>'
 lib/generators/plugin/migration_generator.rb:1:in `<top (required)>

### DIFF
--- a/lib/generators/plugin/migration_generator.rb
+++ b/lib/generators/plugin/migration_generator.rb
@@ -1,3 +1,5 @@
+require 'rails/generators'
+
 module Plugin
   class MigrationGenerator < Rails::Generators::Base
     class_option :plugin_name, :required => true


dLobatog (approved):
👍 , I'm able to reproduce the failure just by running `RAILS_ENV=production passenger start` and this fixes it. 

It looks like a few Katello tests failed, as far as I know only https://github.com/Katello/katello/pull/6811 should have failed. Let's [test katello] and see if we get a better result now as I'm not sure what happened. 

If only 1 test fails now, I'll merge. Thanks @ohadlevy !

dLobatog:
[test katello]

dLobatog:
Thanks @ohadlevy!

ShimShtein:
IMHO generators should not be loaded into passenger environment at all, especially not inside `eager_load!`.

The generator is loaded because the whole `/lib` directory is eagerly loaded [here](https://github.com/theforeman/foreman/blob/develop/config/application.rb#L106). I would suggest excluding `lib/generators` from eager load paths, since I don't see any value in it.

ohadlevy:
@ShimShtein than in that case - patches are welcomed ;)
also, since a lot of the code that used to live in /lib moved to app/services etc, maybe we should revisit the eager load? whats the different between auto load and eager load in this case?

PR #4572, opened 2017-06-06
Fixes #19934 - Don't limit smart class parameter order length

tbrisker:


db/migrate/20170606115344_change_lookup_key_path_to_text.rb
```diff
@@ -0,0 +1,6 @@
+class ChangeLookupKeyPathToText < ActiveRecord::Migration
+  def change
+    remove_index :lookup_keys, :path
```
tbrisker, line 3: I'm removing the index since indexing text on mysql is problematic and we never search based on the path anyways.

ohadlevy: I was really surprised to see we don't use the index at all, but you are correct...

ohadlevy:
thanks @tbrisker

PR #4574, opened 2017-06-07
Fixes #19764 - Fix Layout/AccessModifierIndentation cop

tbrisker:


dLobatog (approved):
👍 given tests pass (which I guess they'll do with such a small change!)

lzap:
Thanks! Only 5732 to go... ;-)

PR #4575, opened 2017-06-07
Fixes #19952 - Allow view_audit_logs to work

dLobatog:
Every time the /audits page is visited by a non-admin user, it checks
whether the permission 'view_audit_logs' is for the Audit class. In
previous versions of 'audited', this was
Audited::Adapters::ActiveRecord::Audit, but right now it's simply
Audited::Audit.

To test the change, visit the page with an user that just has the
'view_audit_logs' permission. Check how without updating the
authorizer.rb file, /audits will be empty.

### DIFF
--- a/app/services/authorizer.rb
+++ b/app/services/authorizer.rb
@@ -169,7 +169,7 @@ class Authorizer
     return 'Parameter' if klass <= Parameter
 
     case (name = klass.to_s)
-    when 'Audited::Adapters::ActiveRecord::Audit'
+    when 'Audited::Audit'
       'Audit'
     when /\AHost::.*\Z/
       'Host'


tbrisker:
Should this require a change to the minimal audited gem version as well?

dLobatog:
@tbrisker We already did the change, it's what caused this to break:

On versions up until 4.2 of Audited, https://github.com/collectiveidea/audited/blob/3.0-stable/lib/audited/adapters/active_record/audit.rb 

4.3 onwards, it's `Audited::Audit` where that code was moved. https://github.com/collectiveidea/audited/blob/v4.3.0/lib/audited/audit.rb

Since Foreman pins the gem to ~>4.3 , we get the latter class name. https://github.com/theforeman/foreman/blob/develop/Gemfile#L8

The reporter mentions in Redmine the patch fixes the issue for them.

tbrisker (approved):

tbrisker:
Thanks @dLobatog !

PR #4576, opened 2017-06-07
Fixes #19968: Add a rake task to seed hosts

ekohl:
This uses FactoryGirl to seed the database with trivial hosts. It allows developers to easily set up a bunch of hosts to get some realistic database size.

Currently it's very much a proof of concept and meant to get some feedback if this is a good idea. A more realistic generator would create hosts with some properties, realisticly divide them among orgs with domains.

@orrabin @tbrisker this is what I was talking about in the car today.

### DIFF
--- a/lib/tasks/seed.rake
+++ b/lib/tasks/seed.rake
@@ -1,0 +1,150 @@
+require 'factory_bot'
+
+namespace :seed do
+  desc 'Generate various forgeries. This should not be used in production'
+  task :forgeries => %w[forgeries:all]
+
+  namespace :forgeries do
+    task :load_factories => :environment do
+      if Rails.env.production?
+        STDERR.puts "Refusing to seed data in a production environment"
+        exit 1
+      end
+
+      include FactoryBot::Syntax::Methods
+      FactoryBot.find_definitions
+    end
+
+    task :domains => :load_factories do
+      names = ['example.com', 'example.org', 'example.net']
+
+      User.as_anonymous_admin do
+        if SETTINGS[:organizations_enabled]
+          Organization.all.each do |organization|
+            Organization.current = organization
+            # TODO: Select a number of locations? All?
+            locations = SETTINGS[:locations_enabled] ? [] : nil
+            generate(Domain, get_desired(5)) do |count, generator|
+              create(:domain,
+                     :name => build_domain(names, generator),
+                     :organizations => [organization],
+                     :locations => locations)
+            end
+          end
+        else
+          generate(Domain, get_desired(50)) do |count, generator|
+            create(:domain,
+                   :name => build_domain(names, generator))
+          end
+        end
+      end
+    end
+
+    task :hosts => :load_factories do
+      User.as_anonymous_admin do
+        domains = Domain.all
+        if SETTINGS[:organizations_enabled]
+          organizations = Organization.all
+        end
+        operatingsystems = Operatingsystem.all
+        owner = User.anonymous_admin
+
+        generate(Host::Managed, get_desired(100)) do |count, generator|
+          os = operatingsystems.sample
+
+          if SETTINGS[:organizations_enabled]
+            Organization.current = organizations.sample
+            domains = Domain.all
+          end
+
+          create(:host,
+                 :hostname => generator.next_random_name,
+                 :domain => domains.sample,
+                 :operatingsystem => os,
+                 :architecture => os.architectures.sample,
+                 :organization => Organization.current,
+                 :owner => owner)
+        end
+      end
+    end
+
+    task :operatingsystems => :load_factories do
+      User.as_anonymous_admin do
+        build_operatingsystems.each do |os|
+          unless Operatingsystem.find_by(:name => os.name, :major => os.major, :minor => os.minor)
+            os.save!
+            puts "Created #{os}"
+          end
+        end
+      end
+    end
+
+    task :organizations => :load_factories do
+      User.as_anonymous_admin do
+        if SETTINGS[:organizations_enabled]
+          generate(Organization, get_desired(10)) do |count, generator|
+            create(:organization, :name => generator.next_random_name)
+          end
+        end
+      end
+    end
+
+    task :all => [:operatingsystems, :organizations, :domains, :hosts] do
+    end
+
+    def get_desired(default)
+      (ENV['COUNT'] || default).to_i
+    end
+
+    def generate(model, desired)
+      generator = NameGenerator.new
+      current_size = model.all.size
+      step = (ENV['STEP'] || 1).to_i
+
+      print "Found #{current_size}/#{desired} #{model.model_name.plural}. "
+      needed = desired - current_size
+
+      if needed > 0
+        puts "Generating #{needed} #{model.model_name.plural}"
+        needed.times do |count|
+          yield count, generator
+          print '.' if count % step
+        end
+
+        # Put an end of line if we every printed progress
+        puts if needed >= step
+      else
+        puts 'Nothing to do'
+      end
+    end
+
+    def build_operatingsystems
+      architectures = Architecture.all
+
+      [
+        # Red Hat
+        build(:operatingsystem, :name => 'RedHat', :major => '6', :architectures => architectures),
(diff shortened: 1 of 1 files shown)


xprazak2:
This seems like a start 👍. A cleanup task that could remove previously added fake data would be nice as well.

sean797:
Massive +1 to the idea 👍 couple of comments:
- What do you think about checking the environment and if in production requiring an override flag set? (or something along those lines) Just in case an inquisitive user runs it without really knowing what its going to do.
- The name 'hosts', how do you see this working in the future? I assume adding more tasks like 'domains', 'subnets' ect... If so can we have a task like 'all_fake_data' that does everything?

ekohl:
@sean797 yes, that's sort of the idea. This was just a very quick and hacky way to get the concept across but something like an all_fake_data would be the goal. The way I see it is to generate a bunch of domains, ensure a list of operating systems etc. When generating the hosts it should retrieve all those and randomly distribute the hosts between those domains, pick an OS etc. My main goal would be to have some production amount of data in it so you can see how pages scale and the performance. You could also include "weird" data to reproduce customer situations but I don't know what they are.

Safety checks are obviously important. I first considered doing this in a plugin but I believe that'll be a lot of overhead and it's much easier to do this in core.

ekohl:
So I expanded it a bit:
* It checks if it's running in a production environment.
* It will provision operating systems
* It will provision organizations
* It will assign more fields to a host

Questions:
* In most production environments factory_girl won't be available. Will this be a problem?
* What would be required to get this into a mergable shape?

lib/tasks/seed.rake
```diff
@@ -0,0 +1,95 @@
+require 'factory_girl'
+
+namespace :seed do
```
sean797: Tasks like `rake seed:all` & `rake seed:hosts` seems a little generic to me, I might be unsure want I'm really seeding. What about using a different namespace, maybe `seed:forgeries` ?

sean797 (commented):
Added a small comment, I will test it and report back

lib/tasks/seed.rake
ekohl: Oh, I like that idea and I fully agree the current name is too generic. Just something I hadn't gotten around to.

sean797 (dismissed):
Thanks @ekohl Works well! 👍

ohadlevy:
I am trying this pr on develop branch and i can't seem to see the task at all, I'm using rails 5.0 if that matters.

ares:
If you can't see it in `rake -T`, that's because the task does not have any description, try `-TA`

ohadlevy:
using -TA made no difference. also excplictly calling the task name rake
seeds:all seeds:hosts etc did not work (task not found).

ekohl:
@ohadlevy I changed it to ```seed:forgeries``` (al within that ```:hosts```, etc). I rebased it and added a description so it should be in the listing.

ohadlevy:
does it make sense to skip audting for the seeding part or rather keep it so we have audits too?

ohadlevy:
also I've noticed that host names are only short names?

ekohl:
> does it make sense to skip audting for the seeding part or rather keep it so we have audits too?

I'd think auditing is actually nice to have.

> also I've noticed that host names are only short names?

I think I should add some domains too then.

timogoebel:
[test katello]

ekohl:
@ohadlevy I added domains but don't know how to enable auditing. @timogoebel also suggested to add facts but both can be done in later additions.

leewaa:
@ekohl any more progress here ? 

Would be nice to get this one moving.

ohadlevy:
+1

ekohl:
I think there should be some discussion on what good (enough) forgeries are. I'd be happy to rebase this, have people play with it and see if they're happy using it. Since it's a development only tool, it'd be fine to get this in even as a partial feature and iterate on it.

leewaa:
I have used it a few times and it seems to work fine :) 

The only thing is: the first thing I look for using `rake -T` is `rake seed:hosts` since that is exactly whats being seeded. In the context of the project the naming of `forgeries` isn't actually used anywhere so I think `seed:hosts` is better naming.

timogoebel:
> Since it's a development only tool, it'd be fine to get this in even as a partial feature and iterate on it.

I think we should merge this and improve it as we go. It definitely helped us a couple of times already.

ohadlevy (dismissed):
lets just get this green and merged :)

tbrisker:
This needs s/Girl/Bot/g

ekohl:
> This needs s/Girl/Bot/g

I was working on that, but it looks like there's still some weirdness going on with the organization scoping.

tbrisker:
ping @ekohl ?

timogoebel:
@ekohl: We played with this some more, you can find the working result here: https://gist.github.com/kamils-iRonin/f6ca009966c460b6faabe467c57f0ca9

I don't know if that helps, but at least it's out there now. :-)

lib/tasks/seed.rake
```diff
@@ -0,0 +1,139 @@
...
+          create(:organization, :name => generator.next_random_name)
+        end
+      end
+    end
+
+    task :all => [:operatingsystems, :organizations, :domains, :hosts] do
+    end
+
+    def get_desired(default)
+      (ENV['COUNT'] || default).to_i
```
adamruzicka, line 96: For example if I want just 50 hosts (instead of the default 100), setting `COUNT` to 50 would lead to 50 orgs being created. Could something be done about this?

ekohl: I considered that and we could probably get some context so you can use `ORGANIZATION_COUNT` and `DOMAIN_COUNT` by moving it into the `generate` function.

ekohl:
I don't know why but I do get stuck on the domain generation. It doesn't detect the existing domains when orgs are enabled. Perhaps it has something to do with locations and it doesn't deal with that. Given Katello has both enabled by default I think it's good to support that. My knowledge of taxonomies is limited so I'd appreciate help here :)

lucasm-iRonin:
@ekohl there are 2 issues with Foreman / this PR:

1. Creating domains, operating systems etc. runs permissions check in [`after_save` callback](https://github.com/theforeman/foreman/blob/7bd5816a56c7ae9d4fde3816fc44bb1b36b764b4/app/models/concerns/authorizable.rb#L4). Another issue is that it fails silently if permission are not granted. I believe this could be solve'able by replacing `raise ActiveRecord::Rollback` with `ActiveRecord::RecordInvalid` so it could raise the error to the user instead of returning record with empty id. However it would be great to discuss it with someone from the core team why it has been written that way.

2. Current user is not set in the seeds. As described in point no 1, permissions check requires to either have `User.current` or `Thread.current[:ignore_permission_check]` set.

If you add `Thread.current[:ignore_permission_check]` at the beginning of the seed file, it should  create all necessary records for you (I've just tested it and I have host created).

timogoebel:
> Creating domains, operating systems etc. runs permissions check in after_save callback. Another issue is that it fails silently if permission are not granted. I believe this could be solve'able by replacing raise ActiveRecord::Rollback with ActiveRecord::RecordInvalid so it could raise the error to the user instead of returning record with empty id. However it would be great to discuss it with someone from the core team why it has been written that way.

@ares: Do you have an answer for @lucasm-iRonin's question?

ekohl:
Ping :) My rails knowledge is too limited to continue here by myself.

ares:
Sorry for late reply.

ad 1) replacing Rollback with Invalid exception wouldn't help, the after callback rescues both and does not reraise. We could use another exception, but that would change `save` method behavior, which would no longer return false, but raised an exception.

ad 2) just wrap the whole block in ```User.as_anonymous_admin do ... end```, that way you don't have to deal with permissions and taxonomies.

ekohl:
Updated. I think this now does what it's supposed to do and, if others agree, could be merged to be iterated on by other developers.

timogoebel (approved):
Looks good to me, I'd like to follow your suggestion and improve it as we go.

tbrisker (approved):
code looks good, didn't test yet but agree on the approach of starting with this and reiterating. This will make dev setup much smoother.

timogoebel:
Thanks, @ekohl.

Test failures were unrelated.

PR #4577, opened 2017-06-08
fixes #19969 - Find a vlanid for a subnet6 if no subnet4 is present

jovandeginste:
You can have a network interface with an ipv6 subnet configured but
without an ipv4 subnet; in this case, the function `vlanid` for
`interface` will throw an error.

This fix will test for `nil` and fall back to `subnet6`. If both subnets
are `nil`, we return `nil` too.

### DIFF
--- a/app/models/nic/interface.rb
+++ b/app/models/nic/interface.rb
@@ -27,7 +27,21 @@ module Nic
     alias_method :network6, :subnet6_network
 
     def vlanid
-      self.tag.blank? ? self.subnet.vlanid : self.tag
+      # Determine a vlanid according to the following cascading rules:
+      # 1. if the interface has a tag, use that as the vlanid
+      # 2. if the interface has a v4 subnet with a non-blank vlanid, use that
+      # 3. if the interface has a v6 subnet with a non-blank vlanid, use that
+      # 4. if no reasonable vlanid was determined, then return an empty string
+      #
+      # In case the v4 and v6 subnet are both present, they should have the same
+      # vlanid. If they have a different vlanid, this is probably an error in the
+      # user's database. There is no way of determining the real vlanid, so we
+      # pick the v4 one unless it turns out to be blank.
+
+      return self.tag unless self.tag.blank?
+      return self.subnet.vlanid if self.subnet && self.subnet.vlanid.present?
+      return self.subnet6.vlanid if self.subnet6 && self.subnet6.vlanid.present?
+      return ''
     end
 
     def bridge?
--- a/test/fixtures/subnets.yml
+++ b/test/fixtures/subnets.yml
@@ -51,3 +51,11 @@ six:
   network: '2001:db10::'
   mask: 'ffff:ffff:ffff:ffff::'
   dns: three
+
+seven:
+  name: seven
+  type: Subnet::Ipv6
+  network: '2001:db10:7::'
+  mask: 'ffff:ffff:ffff:ffff::'
+  dns: three
+  vlanid: 44
--- a/test/models/nic_test.rb
+++ b/test/models/nic_test.rb
@@ -72,16 +72,37 @@ class NicTest < ActiveSupport::TestCase
 
   test "should delegate subnet attributes" do
     subnet = subnets(:two)
+    subnet6 = subnets(:seven)
     domain = (subnet.domains.any? ? subnet.domains : subnet.domains << Domain.first).first
     interface = FactoryGirl.build(:nic_managed,
                                   :ip => "3.3.4.127",
                                   :mac => "cabbccddeeff",
                                   :host => FactoryGirl.create(:host),
                                   :subnet => subnet,
+                                  :subnet6 => subnet6,
                                   :name => "a" + FactoryGirl.create(:host).name,
                                   :domain => domain)
     assert_equal subnet.network, interface.network
+    assert_equal subnet6.network, interface.network6
     assert_equal subnet.vlanid, interface.vlanid
+    assert_equal '42', interface.vlanid
+  end
+
+  test "should delegate subnet6 attributes if subnet is nil" do
+    subnet = nil
+    subnet6 = subnets(:seven)
+    domain = (subnet6.domains.any? ? subnet6.domains : subnet6.domains << Domain.first).first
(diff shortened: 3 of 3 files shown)


timogoebel:
@jovandeginste : Thanks for finding & fixing this. Do you feel comfortable adding a small test for this?

timogoebel:
@theforeman-bot: ok to test

jovandeginste:
@timogoebel I will see how the tests work

app/models/nic/interface.rb
```diff
@@ -23,7 +23,10 @@ class Interface < Base
     delegate :network, :to => :subnet
 
     def vlanid
-      self.tag.blank? ? self.subnet.vlanid : self.tag
+      return self.tag unless self.tag.blank?
+      return self.subnet.vlanid if self.subnet
```
ares: I think the condition should also verify that `subnet.vlanid.present?`, otherwise if interface is attached to ipv4 subnet without vlan, ipv6 subnet vlan id can't win?

app/models/nic/interface.rb
```diff
@@ -23,7 +23,10 @@ class Interface < Base
...
+      return self.subnet6.vlanid if self.subnet6
```
ares: if `self.subnet6` is present but `self.subnet6.vlanid` is blank, it will return '' instead of nil, it should also probably be a part of a condition

app/models/nic/interface.rb
jovandeginste: @ares while your reasoning is sound code-wise, the `subnet` and `subnet6` _should_ have the same vlanid (if the subnets are present); after all, this is the same connection, only different protocol.

The case I'm solving here, is the one where you don't have a IPv4 `subnet` at all.

With that in mind, should I or should I not check for blank? :-)

Having a blank vlanid for the v4 subnet and non-blank vlanid for v6 subnet is rather a user error in my opinion?

app/models/nic/interface.rb
jovandeginste: What else should be returned if it's blank? `nil`? Previous behaviour probably resulted in an empty string too ...

ares: You're right about previous behavior, I wonder if that was bug or not. The empty string often comes from the UI, when user does not fill in the field and generally considers that as not filled. Therefore I think it should return `nil` for blank strings but I might be wrong.

jovandeginste:
@timogoebel I've been looking at it today, but can't figure out how to run the tests. Is there any documentation available?

app/models/nic/interface.rb
ares: I agree, perhaps this would indicate it slightly better

```ruby
subnet = [self.subnet, self.subnet6].compact
return subnet.vlanid unless subnet.nil?
```

but maybe just a comment explaining what you described here would do :-) Ultimate solution would be to display a warning if user uses two subnets with different vlan tag. But I suppose that's out of scope.

timogoebel:
@jovandeginste:

Yep, basically run `bundle exec ruby -Itest test/unit/example_test.rb`. [The Handbook](https://theforeman.org/handbook.html#UsingtheSpringpreloaderindevelopment) has more details. To run a single test, append `-n /.*somestring.*/`.

app/models/nic/interface.rb
jovandeginste: I think you forgot a `.first` in the end of the first line. But then you actually do exactly the same as I suggested in the PR (but - and that's my opinion of course - less readable).

I do agree I can add a blank-check too, will do that when I figure out how the tests work.

ares: You're right again :-) I tend to agree with the readability, so if you'll touch it because of blank check, just please add a comment explaining the expected setup. Something like if there's a vlanid set, it's expected both subnets should have it set to the same value.

jovandeginste: @ares at the risk of being too verbose, I added some comments and added a blank check. Now I REALLY need to figure out how to get those tests working :-/

ares: feel free to jump on #theforeman-dev IRC channel, I and others might help with the tests

jovandeginste:
@timogoebel tests are added, I hope they don't fail ;-)

mmoll:
please rebase and if possible also add a test for GH-4582 here 👍

jovandeginste:
@mmoll rebased, and extra tests for #4582 are added

app/models/nic/interface.rb
```diff
@@ -27,7 +27,21 @@ class Interface < Base
...
+      # 3. if the interface has a v6 subnet with a non-blank vlanid, use that
+      # 4. if no reasonable vlanid was determined, then return an empty string
+      #
+      # In case the v4 and v6 subnet are both present, they should have the same
+      # vlanid. If they have a different vlanid, this is probably an error in the
+      # user's database. There is no way of determining the real vlanid, so we
+      # pick the v4 one unless it turns out to be blank.
+
+      return self.tag unless self.tag.blank?
+      return self.subnet.vlanid if self.subnet && !self.subnet.vlanid.blank?
```
dLobatog: Nitpick, I think the double negative (! and blank?) can easily be substituted by:

`return self.subnet.vlanid if self.subnet && self.subnet.vlanid.present?`

Similarly the last return is unnecessary

dLobatog (changes requested):
@jovandeginste Minor change requested, it looks good to me, however I wonder why the method is changed to return an empty string (`''`) instead of nil as it did before?

app/models/nic/interface.rb
jovandeginste: I updated the `!...blank?`; however the last `return` returns an empty string (and not a `nil` object). This is to conserve compatibility.

dLobatog (approved):
For the onlookers, I was not sure whether the old method was returning a `nil` or an empty string. 

```ruby
def vlanid
    self.tag.blank? ? self.subnet.vlanid : self.tag 
end
```

Just tried and `self.tag` returns an empty string.

dLobatog:
👍 to merge if tests pass, thanks @jovandeginste !

dLobatog:
Thanks @jovandeginste , @ares, @timogoebel !

PR #4578, opened 2017-06-08
Refs #19588 - Allow passing header to csv responder

tbrisker:


app/services/csv_exporter.rb
```diff
@@ -1,7 +1,8 @@
 require 'csv'
 
 module CsvExporter
-  def self.export(resources, columns, header = default_header(columns))
+  def self.export(resources, columns, header = nil)
+    header ||= default_header(columns)
```
tbrisker, line 5: can't use default value here because the csv responder may pass a `nil` which will cause the default value to be ignored.

dLobatog (approved):

dLobatog:
Thanks @tbrisker !

PR #4579, opened 2017-06-09
Fixes #14692 - Allow user to update its username

rabajaj0509:


tbrisker:
ok to test

app/models/user.rb
```diff
@@ -163,6 +163,10 @@ def internal?
     auth_source.is_a? AuthSourceInternal
   end
 
+  def can_update_username?
+    hidden? or internal? or admin?
```
dLobatog: Nitpick, this is control-flow OR, it should be logic OR (http://www.virtuouscode.com/2010/08/02/using-and-and-or-in-ruby/)

app/controllers/concerns/foreman/controller/parameters/user.rb
```diff
@@ -25,7 +25,7 @@ def user_params_filter
         end
 
         filter.permit do |ctx|
-          if !ctx.editing_self? && (ctx.ui? || ctx.api?)
+          if ctx.can_update_username? && (ctx.ui? || ctx.api?)
```
dLobatog, line 28: Can you write unit/functional tests for these changes?

dLobatog (changes requested):
From the redmine issue, please also include some clause to not allow LDAP users to update their names (users where auth_source.type is AuthSourceLDAP) - 

Users can update their own names if the Auth Source of these users is Internal, but cannot update their own names if the Auth Source is LDAP-based. The reason is simply that they'll lock themselves out if they change the name (Foreman won't propagate these changes to LDAP).

Thanks @rahulbajaj0509 !

app/models/user.rb
```diff
@@ -163,6 +163,14 @@ def internal?
     auth_source.is_a? AuthSourceInternal
   end
 
+  def ldap?
+    auth_source.is_a? AuthSourceLdap
+  end
+
+  def can_update_username?
+    !ldap?
```
timogoebel: I think, it would be better to make this an option of the AuthSource. That way this would use polymorphism.

```ruby
delegate :can_update_username?, :to => :auth_source
```

timogoebel (changes requested):

app/models/user.rb
evgeni: :+1:, esp as LDAP is not the only source we can't update. External (Kerberos) based auth is also quite strict about that ;)

app/controllers/concerns/foreman/controller/parameters/user.rb
```diff
@@ -51,6 +51,10 @@ def editing_self?
       @editing_self
     end
 
+    def can_update_username?
+      User.current.try(:can_update_username?)
```
timogoebel: How about 

```ruby
!!User.current.try(:can_update_username?)
````
to make sure we return a boolean.

File: app/models/auth_source.rb [hunk 5]
```diff
@@ -40,6 +40,14 @@ def auth_method_name
     "Abstract"
   end
 
+  def ldap?
+    AuthSource.is_a? AuthSourceLdap
```
timogoebel: I believe, we can use polymorphism for this as well. Set this to `false` and overwrite the method in `app/models/auth_sources/auth_source_ldap.rb`

timogoebel (changes requested):
Thanks. I left some more minor comments inline.

app/models/auth_source.rb
```diff
@@ -40,6 +40,14 @@ def auth_method_name
     "Abstract"
   end
 
+  def ldap?
```
boaz0: I don't like this method. I think `can_update_username?` is good enough and each `AuthSource` should implement its own `can_update_username?` and here `can_update_username?` will return `true`.

app/models/auth_sources/auth_source_ldap.rb
```diff
@@ -84,6 +84,10 @@ def auth_method_name
     "LDAP"
   end
 
+  def ldap?
+    AuthSource.is_a? AuthSourceLdap
```
dLobatog: Shouldn't this just return true..?

dLobatog (changes requested):
@rahulbajaj0509 Check out tests and a few comments inline, almost there 😄 !

test/models/user_test.rb
```diff
@@ -78,6 +78,16 @@ def setup
     assert_equal user.lower_login, "user"
   end
 
+  test "user can update username" do
+    user = FactoryGirl.create(:user, :auth_source => auth_sources(:internal))
```
kgaikwad: No need to add `:auth_source as :internal` as in [user_related.rb](https://github.com/theforeman/foreman/blob/develop/test/factories/user_related.rb) for `:user` factory, it is assigned to `AuthSourceInternal` only.

kgaikwad (changes requested):
Thanks, please check inline comment.

rabajaj0509:
I do not see any tests failing on my local instance, any idea why so many tests are failing ?

dLobatog (changes requested):
@rahulbajaj0509 Can you try rebasing? This was basically ready but a bunch of tests were failing, let's see which ones were they. Thank you

rabajaj0509:
@dLobatog  there are little changes made in the code since it has been a long time, let me edit the pr accordingly and then I shall rebase. Thanks :)

rabajaj0509:
@dLobatog I think bug got fixed in https://github.com/theforeman/foreman/commit/02d1635cbfca6e96d46b50d27232e152efb58d9f    so we can close this!

PR #4580, opened 2017-06-09
fixes #19977 - Fetch a value for a given parameter name

jovandeginste:
This code will add a function to lookup a parameter by name, and returns
it's value. It will return `nil` if no value was found.

The function is added to the Jail, so it can be called from provisioning
templates.

### DIFF
--- a/lib/foreman/renderer.rb
+++ b/lib/foreman/renderer.rb
@@ -4,12 +4,13 @@ module Foreman
   module Renderer
     class RenderingError < Foreman::Exception; end
     class SyntaxError < RenderingError; end
+    class WrongSubnetError < RenderingError; end
 
     include ::Foreman::ForemanUrlRenderer
 
     ALLOWED_GENERIC_HELPERS ||= [ :foreman_url, :snippet, :snippets, :snippet_if_exists, :indent, :foreman_server_fqdn,
                                   :foreman_server_url, :log_debug, :log_info, :log_warn, :log_error, :log_fatal, :template_name, :dns_lookup,
-                                  :pxe_kernel_options, :save_to_file ]
+                                  :pxe_kernel_options, :save_to_file, :subnet_param, :subnet_has_param? ]
     ALLOWED_HOST_HELPERS ||= [ :grub_pass, :ks_console, :root_pass,
                                :media_path, :param_true?, :param_false?, :match,
                                :host_param_true?, :host_param_false?,
@@ -47,6 +48,17 @@ module Foreman
 
     def host_param_false?(name)
       @host.params.has_key?(name) && Foreman::Cast.to_bool(@host.params[name]) == false
+    end
+
+    def subnet_has_param?(subnet, param_name)
+      validate_subnet(subnet)
+      subnet.parameters.exists?(name: param_name)
+    end
+
+    def subnet_param(subnet, param_name)
+      validate_subnet(subnet)
+      param = subnet.parameters.where(name: param_name).first
+      param.nil? ? nil : param.value
     end
 
     def render_safe(template, allowed_methods = [], allowed_vars = {}, scope_variables = {})
@@ -283,6 +295,10 @@ module Foreman
     def pxe_config
       @kernel = @host.operatingsystem.kernel @host.arch
       @initrd = @host.operatingsystem.initrd @host.arch
+    end
+
+    def validate_subnet(subnet)
+      raise ::Foreman::Renderer::WrongSubnetError.new(N_("'%{object_name}' is a '%{object_class}', expected a subnet.") % { object_name: subnet.to_s, object_class: subnet.class.to_s}) unless subnet.is_a?(Subnet)
     end
   end
 end
--- a/test/unit/foreman/renderer_test.rb
+++ b/test/unit/foreman/renderer_test.rb
@@ -327,6 +327,41 @@ EOS
     assert @renderer.host_param_false?('false_param')
   end
 
+  context 'subnet helpers' do
+    setup do
+      @renderer.host = FactoryGirl.create(:host, :with_puppet)
+      subnets(:one).subnet_parameters.create(name: 'myparam', value: 'myvalue')
+    end
+
+    test 'should have subnet_has_param? helper returning true' do
+      assert @renderer.subnet_has_param?(subnets(:one), 'myparam')
+    end
+
+    test 'should have subnet_has_param? helper returning false' do
+      refute @renderer.subnet_has_param?(subnets(:one), 'my_wrong_param')
+    end
+
+    test 'should have subnet_has_param? helper returning false when subnet is nil' do
+      assert_raises Foreman::Renderer::WrongSubnetError do
+        @renderer.subnet_has_param?(nil, 'myparam')
+      end
+    end
+
+    test 'should render existing subnet param using "subnet_param" helper' do
+      assert_equal @renderer.subnet_param(subnets(:one), 'myparam'), 'myvalue'
+    end
+
+    test 'should not render missing subnet param using "subnet_param" helper' do
+      assert_nil @renderer.subnet_param(subnets(:one), 'my_wrong_param')
+    end
+
+    test 'should throw an error using "subnet_param" helper with nil' do
+      assert_raises Foreman::Renderer::WrongSubnetError do
+        @renderer.subnet_param(nil, 'myparam')
+      end
+    end
+  end
+
   test 'should have host_enc helper' do
     @renderer.host = FactoryGirl.create(:host, :with_puppet)
     assert @renderer.host_enc


app/models/subnet.rb
```diff
@@ -213,6 +213,12 @@ def has_vlanid?
     self.vlanid.present?
   end
 
+  # Given a parameter name, return the (first) value, or nil if nothing found
+  def parameter_for_name(name)
```
dLobatog: Since we have `host_param`, I suppose it'd make sense to have a similar syntax for fetching Subnet parameters, so `subnet_param` - checkout https://github.com/theforeman/foreman/blob/develop/lib/foreman/renderer.rb#L36 or https://github.com/theforeman/foreman/commit/477b6b6b932ee27b2b1c6b2ff39ae6b07fe120e9 if you want to see how it was implemented

@ares @xprazak2 Is it possible to log something here when the parameter is not found, but make it show up on the warning dialog during preview instead of the foreman logs?

dLobatog (changes requested):
Thanks @jovandeginste , I left a comment - also please add tests for this method when you get a chance. Can't believe you had to disable the jail for this 😄 ! OperatingSystem suffers from the same problem if you feel like fixing it on another PR.

app/models/subnet.rb
jovandeginste: @dLobatog is it possible to add a `subnet_param` to `lib/foreman/renderer.rb` ? I see eg. `ALLOWED_HOST_HELPERS`, is this relevant too? Should I add `subnet_param` to that array, or should there be an different implementation here? I don't see it done easily (but I have limited experience with this)

sean797:
I'm not sure I understand why this is necessary, I was under the impression parameters are inherited and doing `@host.param['foo']` would work for a parameter set on a subnet as well as other places?

app/models/subnet.rb
ares: I believe `host_param` contains values from subnets as well, it does the whole classification. There's no `domain_param` or `os_param` either. The `parameter_for_name` is I think equivalent to `host_param` but it only searches for HostParameter objects.

Warnings during preview sounds as a great idea. Could you open RFE for that pls? I can imagine there would be more stuff we can warn about.

ares:
I think @sean797 is right but I'd recommend using (and improving) `host_param('foo')` instead of `@host.param['foo']`

jovandeginste:
@ares @sean797 I don't understand - a host can have multiple subnets (to start), and I need to know the value of `param_x` for every subnet separately (in combination with the subnet); how would I access this through inherited params?

ares:
I see now, I believe `host_param` only takes parameters from subnet of primary interface. I suppose adding `subnet_param` macro would make sense then, it should probably take one argument the subnet you're interested in and second would be the parameter name.

To answer your original question, yes it should be added to rederer.rb like [host_param](https://github.com/theforeman/foreman/commit/477b6b6b932ee27b2b1c6b2ff39ae6b07fe120e9#diff-6632a0077c474a44423343eb597f9d72R33). Since this is not host specific macro, it should be added into `ALLOWED_GENERIC_HELPERS` array. A test for this macro would be also good, an example can be taken from [this](https://github.com/theforeman/foreman/commit/477b6b6b932ee27b2b1c6b2ff39ae6b07fe120e9#diff-abc4e103433897b58061f39899e7e4b0R230). Let us know if you need a help with any of these.

jovandeginste:
@ares got something together, is this reasonable?

jovandeginste:
@ares I guess the failing check is not really related to my code?

test/unit/foreman/renderer_test.rb
```diff
@@ -327,6 +327,56 @@ def setup_safemode_renderer
...
+    subnet = FactoryGirl.create(:subnet_ipv4)
+    subnet.subnet_parameters.create(name: 'myparam', value: 'myvalue')
+    assert @renderer.subnet_has_param?(subnet, 'myparam').present?
+  end
+
+  test 'should have subnet_has_param? helper returning true' do
+    @renderer.host = FactoryGirl.create(:host, :with_puppet)
+    subnet = FactoryGirl.create(:subnet_ipv4)
+    subnet.subnet_parameters.create(name: 'myparam', value: 'myvalue')
+    assert @renderer.subnet_has_param?(subnet, 'myparam') == true
```
ares: no need for `== true`, just `assert @renderer.subnet_has_param?(subnet, 'myparam')`

test/unit/foreman/renderer_test.rb
```diff
@@ -327,6 +327,56 @@ def setup_safemode_renderer
     assert @renderer.host_param_false?('false_param')
   end
 
+  test 'should have subnet_has_param?' do
+    @renderer.host = FactoryGirl.create(:host, :with_puppet)
+    subnet = FactoryGirl.create(:subnet_ipv4)
+    subnet.subnet_parameters.create(name: 'myparam', value: 'myvalue')
+    assert @renderer.subnet_has_param?(subnet, 'myparam').present?
```
ares: since the method returns `true` only, it's fine without `.present?`, which is already covered by the test below, I think this one can be removed

test/unit/foreman/renderer_test.rb
```diff
@@ -327,6 +327,56 @@ def setup_safemode_renderer
...
+    subnet = FactoryGirl.create(:subnet_ipv4)
+    subnet.subnet_parameters.create(name: 'myparam', value: 'myvalue')
+    assert @renderer.subnet_has_param?(subnet, 'myparam') == true
+  end
+
+  test 'should have subnet_has_param? helper returning false' do
+    @renderer.host = FactoryGirl.create(:host, :with_puppet)
+    subnet = FactoryGirl.create(:subnet_ipv4)
+    subnet.subnet_parameters.create(name: 'myparam', value: 'myvalue')
+    assert @renderer.subnet_has_param?(subnet, 'my_wrong_param') == false
```
ares: this line is equivalent to more standard form
`refute @renderer.subnet_has_param?(subnet, 'my_wrong_param')`

test/unit/foreman/renderer_test.rb
```diff
@@ -327,6 +327,56 @@ def setup_safemode_renderer
...
+    @renderer.host = FactoryGirl.create(:host, :with_puppet)
+    subnet = FactoryGirl.create(:subnet_ipv4)
+    subnet.subnet_parameters.create(name: 'myparam', value: 'myvalue')
+    assert @renderer.subnet_has_param?(subnet, 'myparam') == true
+  end
+
+  test 'should have subnet_has_param? helper returning false' do
+    @renderer.host = FactoryGirl.create(:host, :with_puppet)
+    subnet = FactoryGirl.create(:subnet_ipv4)
+    subnet.subnet_parameters.create(name: 'myparam', value: 'myvalue')
```
ares: this line is actually not needed and slows down the test a bit :-) but if you want to make sure it does not load the existing parameter with different name, it's fine to keep it

test/unit/foreman/renderer_test.rb
```diff
@@ -327,6 +327,56 @@ def setup_safemode_renderer
...
+    assert @renderer.subnet_has_param?(subnet, 'my_wrong_param') == false
+  end
+
+  test 'should have subnet_has_param? helper returning false when subnet is nil' do
+    @renderer.host = FactoryGirl.create(:host, :with_puppet)
+    exception = assert_raises Foreman::Renderer::WrongSubnetError do
+      @renderer.subnet_has_param?(nil, 'myparam') == false
```
ares: no need for `== false`, it fails before it even starts to comparison

test/unit/foreman/renderer_test.rb
```diff
@@ -327,6 +327,56 @@ def setup_safemode_renderer
...
+    subnet = FactoryGirl.create(:subnet_ipv4)
+    subnet.subnet_parameters.create(name: 'myparam', value: 'myvalue')
+    assert_equal @renderer.subnet_param(subnet, 'myparam'), 'myvalue'
+  end
+
+  test 'should not render missing subnet param using "subnet_param" helper' do
+    @renderer.host = FactoryGirl.create(:host, :with_puppet)
+    subnet = FactoryGirl.create(:subnet_ipv4)
+    subnet.subnet_parameters.create(name: 'myparam', value: 'myvalue')
+    assert @renderer.subnet_param(subnet, 'my_wrong_param').present? == false
```
ares: same as above, parameter creation seems unnecessary and this line would actually be better as

`assert_nil @renderer.subnet_param(subnet, 'my_wrong_param')`

test/unit/foreman/renderer_test.rb
```diff
@@ -327,6 +327,56 @@ def setup_safemode_renderer
...
+  end
+
+  test 'should not throw an error "subnet_param" helper with nil' do
+    @renderer.host = FactoryGirl.create(:host, :with_puppet)
+    assert_raises NoMethodError do
+      @renderer.subnet_param(nil, 'myparam').present? == false
```
ares: no need for `== false`

lib/foreman/renderer.rb
```diff
@@ -49,6 +50,16 @@ def host_param_false?(name)
       @host.params.has_key?(name) && Foreman::Cast.to_bool(@host.params[name]) == false
     end
 
+    def subnet_has_param?(subnet, param_name)
+      raise ::Foreman::Renderer::WrongSubnetError.new(N_("'%{object_name}' is a '%{object_class}', expected a subnet.") % { object_name: subnet.to_s, object_class: subnet.class.to_s}) unless subnet.is_a?(Subnet)
+      return !subnet.parameters.where(name: param_name).empty?
```
ares: I think it would be slightly more readable without negation and using `exists?` like this
`return `subnet.parameters.exists?(name: param_name)`

it also generates slightly more effective SQL query

lib/foreman/renderer.rb
```diff
@@ -49,6 +50,16 @@ def host_param_false?(name)
...
+    end
+
+    def subnet_param(subnet, param_name)
+      param = subnet.parameters.where(name: param_name).first
+      return param.value if param
```
ares: I don't insist but I think following would be clearer in terms of what the method returns

`param.nil? ? nil : param.value`

ares (changes requested):
I think this is the right direction. I left few comments, mostly cosmetic. Don't worry about the test failure, it's one of the tests that randomly fails and is not related to this PR.

test/unit/foreman/renderer_test.rb
jovandeginste: Yes, however it is more in line with the `== false` ... (?)

test/unit/foreman/renderer_test.rb
jovandeginste: yes, that's what I'm testing...

test/unit/foreman/renderer_test.rb
jovandeginste: okay, good to know!

test/unit/foreman/renderer_test.rb
jovandeginste: (and now I read another comment lower, about refute)

test/unit/foreman/renderer_test.rb
jovandeginste: correct

test/unit/foreman/renderer_test.rb
jovandeginste: again, a bug might always return the first param

lib/foreman/renderer.rb
jovandeginste: I did not know the `exists?` function ... My other guess implementation would have been `!...find(...).nil?`; TIL :-)

lib/foreman/renderer.rb
jovandeginste: I honestly disagree with the readability assessment, but if everyone else in the community thinks this is more readable, I won't care too much :-)

jovandeginste:
@ares @dLobatog again, changes were made :-)

lib/foreman/renderer.rb
```diff
@@ -49,6 +50,16 @@ def host_param_false?(name)
       @host.params.has_key?(name) && Foreman::Cast.to_bool(@host.params[name]) == false
     end
 
+    def subnet_has_param?(subnet, param_name)
+      raise ::Foreman::Renderer::WrongSubnetError.new(N_("'%{object_name}' is a '%{object_class}', expected a subnet.") % { object_name: subnet.to_s, object_class: subnet.class.to_s}) unless subnet.is_a?(Subnet)
+      return subnet.parameters.exists?(name: param_name)
+    end
+
+    def subnet_param(subnet, param_name)
+      param = subnet.parameters.where(name: param_name).first
```
ares: this needs the same check for subnet type, otherwise I'm getting `undefined method `parameters' for "1":String` when I use it incorrectly

ares (changes requested):
One nit, otherwise works like a charm!

lib/foreman/renderer.rb
jovandeginste: I split up the check into a private function and added it to the `subnet_param` function

jovandeginste:
@ares yet again, changes were made :-)

lib/foreman/renderer.rb
```diff
@@ -49,6 +50,17 @@ def host_param_false?(name)
       @host.params.has_key?(name) && Foreman::Cast.to_bool(@host.params[name]) == false
     end
 
+    def subnet_has_param?(subnet, param_name)
+      validate_subnet(subnet)
+      return subnet.parameters.exists?(name: param_name)
```
dLobatog: nitpick, this return is superfluous

dLobatog: nitpick, this return is superfluous

test/unit/foreman/renderer_test.rb
```diff
@@ -327,6 +327,49 @@ def setup_safemode_renderer
...
+  end
+
+  test 'should render existing subnet param using "subnet_param" helper' do
+    @renderer.host = FactoryGirl.create(:host, :with_puppet)
+    subnet = FactoryGirl.create(:subnet_ipv4)
+    subnet.subnet_parameters.create(name: 'myparam', value: 'myvalue')
+    assert_equal @renderer.subnet_param(subnet, 'myparam'), 'myvalue'
+  end
+
+  test 'should not render missing subnet param using "subnet_param" helper' do
```
dLobatog: All of the tests are doing `@renderer.host = FactoryGirl.create(:host, :with_puppet)`, you could move the tests into a `context 'subnet helpers' ` or similar, then put that line into a `setup` block.

Similarly, there is repeated code for `
subnet = FactoryGirl.create(:subnet_ipv4)
subnet.subnet_parameters.create(name: 'myparam', value: 'myvalue')` which could go into another context and setup block. Only the tests that check for `subnet is nil` do not need it.

dLobatog (changes requested):
@jovandeginste Sorry, "one last thing" - please check out the comments in tests and the superfluous return statement. 👍 after that, thank you!

lib/foreman/renderer.rb
jovandeginste: many things in ruby are superfluous, that's part of the appeal :-)

test/unit/foreman/renderer_test.rb
jovandeginste: I ... can't get it working. Please help!

jovandeginste:
@dLobatog fixed the tests up - is this acceptable?

dLobatog (approved):
👍 to merge if tests pass!

dLobatog:
I think @ares  your concern was fixed in https://github.com/theforeman/foreman/pull/4580/files#diff-6632a0077c474a44423343eb597f9d72R301 - merging..

dLobatog:
Thanks @jovandeginste !

jovandeginste:
finally :-)

ares:
@jovandeginste :+1:

PR #4582, opened 2017-06-13
Fixes #19991 - Add network6 to nic model

jovandeginste:
Fixes #19991 - Add network6 to nic model

This adds the method `network6`, delegated to the `subnet6` relation.
Because the `delegate` methods does not support completely renaming a
method, I use a combination of delegate and alias. I updated the
definition for `network` (the v4 equivalent) to be consistent.

Attention: builds further upon #4577, because I extend those tests...

### DIFF
--- a/app/models/nic/interface.rb
+++ b/app/models/nic/interface.rb
@@ -20,7 +20,11 @@ module Nic
 
     validate :alias_subnet
 
-    delegate :network, :to => :subnet
+    delegate :network, :to => :subnet, :prefix => true
+    delegate :network, :to => :subnet6, :prefix => true
+
+    alias_method :network, :subnet_network
+    alias_method :network6, :subnet6_network
 
     def vlanid
       self.tag.blank? ? self.subnet.vlanid : self.tag


app/models/nic/interface.rb
```diff
@@ -20,10 +20,28 @@ class Interface < Base
 
     validate :alias_subnet
 
-    delegate :network, :to => :subnet
+    delegate :network, :to => :subnet, :prefix => true
+    delegate :network, :to => :subnet6, :prefix => true
```
dLobatog, line 24: `network6`? I'm not quite sure how this works if two delegates with the same name point to 2 different associations

app/models/nic/interface.rb
```diff
@@ -20,10 +20,28 @@ class Interface < Base
...
+
+    alias_method :network, :subnet_network
```
dLobatog, line 26: What are these for? Doesn't `:prefix` do this already?

dLobatog (changes requested):
@jovandeginste Thanks, I think the patch makes sense generally, even though it's not being used here it could be used in templates.

I asked a few comments inline,  please check it out. Would you be able to separate the changes from this commit from https://github.com/theforeman/foreman/pull/4577/files ? I think it's only the test that conflicts, so if you can - that'd be the best as we could merge it right away.

app/models/nic/interface.rb
jovandeginste: The `:prefix => true` changes the `delegate` behaviour: instead of a method `network` pointing to `subnet.network`, it creates a method `subnet_network` (still pointing to `subnet.network`). The same is done for `subnet6`: a method `subnet6_network` pointing to `subnet6.network`

app/models/nic/interface.rb
jovandeginste: This will alias `subnet_network` to `network`, so that you can call `self.network` and get `self.subnet_network`, which points to `self.subnet.network`. I know, a lot of indirection :-) The next line does the same for `network6` and `subnet6`.

jovandeginste:
@dLobatog I changed the commit to only add the (new) function; I can add tests later.

dLobatog:
[test]

dLobatog (approved):

tbrisker:
ok to test

dLobatog:
Thanks @jovandeginste - please add the test in the other PR if you can!

PR #4583, opened 2017-06-13
Fixes #13092 - Deploy correct bonding for reprovisioned host

xprazak2:


xprazak2:
Integration tests failing

app/views/unattended/provisioning_templates/snippet/_kickstart_ifcfg_generic_interface.erb
```diff
@@ -40,7 +40,7 @@ DEFROUTE=<%= primary %>
 <%-     end -%>
 <%-   end -%>
 <%- end -%>
-<%- if @interface.virtual? && ((!@subnet.nil? && @subnet.has_vlanid?) || @interface.vlanid.present?) -%>
+<%- if @interface.virtual? && ((!@subnet.nil? && @subnet.has_vlanid?) || (@interface.tag.present? && @interface.vlanid.present?)) -%>
```
ares: please don't change the templates here, send a PR to [community-templates repo](https://github.com/theforeman/community-templates). Once it's merged there, we sync templates to Foreman code base from there. Change like this would be overidden on next sync.

app/views/unattended/provisioning_templates/snippet/_kickstart_networking_setup.erb
```diff
@@ -35,7 +35,7 @@ kind: snippet
 <%-   next if bonded_interfaces.include?(interface.identifier) -%>
 
 <%=   "# #{interface.identifier} interface" %>
-<%=   snippet('kickstart_ifcfg_get_identifier_names', :variables => { :interface => interface }) -%>
+<%=   snippet('kickstart_ifcfg_get_identifier_names', :variables => { :interface => interface, :identifier => interface.identifier }) -%>
```
ares: the same as above, should not be changed in this repo

app/views/unattended/provisioning_templates/snippet/_kickstart_ifcfg_generic_interface.erb
ares: why is this change required? `@interface.vlanid` checks for tag and if it's not set, it loads the default value from subnet, this basically mean that only interfaces with explicit tag set would be considered vlans, while it's more common to have the tag set only for subnet which the vlan interface is attached to

app/views/unattended/provisioning_templates/snippet/_kickstart_networking_setup.erb
ares: This would disable the real identifier detection for all interfaces, not just vlans attached to bonds. I suppose this should be only set if `@host.bond_interfaces.map(&:identifier).include?(interface.attached_to)` (untested)

app/models/host/base.rb
```diff
@@ -174,6 +174,8 @@ def set_interfaces(parser)
 
       parser.interfaces.each do |name, attributes|
         iface = get_interface_scope(name, attributes).try(:first) || interface_class(name).new(:managed => false)
+        # update bond when interface is attached to it
+        update_attached_interfaces(name, attributes) unless attributes['virtual']
```
ares: shouldn't this happen in `set_interface` just like `update_virtuals`? I suppose it should also happen only if identifier changed... maybe it should be combined with the `update_virtuals` method into single one

ares (changes requested):

app/views/unattended/provisioning_templates/snippet/_kickstart_ifcfg_generic_interface.erb
xprazak2: Because if `@subnet.nil? #=> true` and `@interface.tag.blank? #=> true`, then you get [undefined method 'vlanid' for nil](https://github.com/theforeman/foreman/blob/develop/app/models/nic/interface.rb#L30)

xprazak2:
I moved the template changes into community-templates and updating the list of attached interfaces for a bond is now part of `set_interface` method.

app/models/host/base.rb
```diff
@@ -388,7 +388,22 @@ def get_interface_scope(name, attributes, base = self.interfaces)
       end
     end
 
+    def update_attached_interfaces(name, attributes)
+      bond_interfaces.each do |bond|
+        next unless bond.children_mac_addresses.include?(attributes['macaddress'])
+        next if bond.attached_devices_identifiers.include? name
+        iface = interfaces.find_by :mac => attributes['macaddress']
```
ShimShtein: This forces a query to database for each bond, even if interfaces are already loaded. IMHO it would be better to use ruby's `select`:
``` ruby
interfaces.select { |iface| iface.mac == attributes['macaddress'] }
```

config/initializers/active_record_extensions.rb
```diff
@@ -8,11 +8,11 @@ class ActiveRecord::Base
 # Permit safemode template rendering to have basic read-only access over
 # model relations
 class ActiveRecord::AssociationRelation::Jail < Safemode::Jail
-  allow :[], :each, :first, :to_a
+  allow :[], :each, :first, :to_a, :map
```
tbrisker, line 11: I think this could be a very dangerous addition - think about something like `host.interfaces.map(&:destroy)`

config/initializers/active_record_extensions.rb
```diff
@@ -8,11 +8,11 @@ class ActiveRecord::Base
...
 end
 
 class ActiveRecord::Associations::CollectionProxy::Jail < Safemode::Jail
-  allow :[], :each, :first, :to_a
+  allow :[], :each, :first, :to_a, :map
```
tbrisker, line 15: ditto

app/models/host/base.rb
tbrisker: since we only need one `Enumerable#find` would be better then `Enumerable#select`.

tbrisker (dismissed):

config/initializers/active_record_extensions.rb
ShimShtein: We already have `#each`, I don't think `map` gives more power than each. It would be better to blacklist `#destroy`, since it's a dangerous action we don't want.

xprazak2:
I replaces `find_by` with `detect`

xprazak2:
katello failure unrelated

config/initializers/active_record_extensions.rb
tbrisker: this has been addressed by latest safemode release.

app/models/host/base.rb
```diff
@@ -388,7 +388,26 @@ def get_interface_scope(name, attributes, base = self.interfaces)
       end
     end
 
+    def update_attached_interfaces(name, attributes)
```
ares: this should probably operate on iface passed as argument

app/models/host/base.rb
```diff
@@ -388,7 +388,26 @@ def get_interface_scope(name, attributes, base = self.interfaces)
...
+        iface = interfaces.detect { |interface| interface.mac == attributes['macaddress'] }
+        update_bond bond, iface, name
+      end
+    end
+
+    def update_bond(bond, iface, name)
+      if iface && iface.identifier
+        bond.remove_device(iface.identifier)
+        bond.add_device(name)
+        bond.save
```
ares: save should log errors if it fails

app/models/host/base.rb
```diff
@@ -388,7 +388,26 @@ def get_interface_scope(name, attributes, base = self.interfaces)
...
+      end
+    end
+
     def set_interface(attributes, name, iface)
+      # update bond.attached_interfaces when interface is in the list and identifier has changed
+      update_attached_interfaces(name, attributes) unless attributes[:virtual]
```
ares: this should live next to update_virtuals and only be triggered if identifier has changed (and similar conditions like update_virtual), the name could be update_bonds to make it clearer

ares (changes requested):

xprazak2:
I renamed `update_attached_interfaces` to `update_bonds`, passed it the interface that is already found, moved it further down to `update_virtuals` and added logging

mmoll:
@xprazak2 could you rebase?

xprazak2:
Rebased.

mmoll:
[test katello]

ares:
Some of my comments made this not working.

The update_bond needs to be executed before the interface identifier is changed, otherwise bond won't find `children_mac_addresses` because of attached devices search based on current interface identifier. So please, move it back under the comment you forgot to move with the update_bonds line, you had it right before my comments :-) at the same time, we can't do iface.identifier_changed?, at this point it was not changed yet. We can replace it with iface.identifier != name.

The error that was there is though the logger line which uses variables that are not available in that scope.

Since I know how much effort it involves to get the setup, I've sent you a PR with changes that made it working for me. This one is good to go when you merge.

@tbrisker any comments from your side? I'd like to merge when my comment above is resolved.

ares:
here's the fixing PR for the reference https://github.com/xprazak2/foreman/pull/2

ares (approved):

mmoll:
this needs a rebase for the audited pin

ares (changes requested):
One quick rebase to make tests green please :-)

xprazak2:
Rebased.

ares (approved):

ares:
Thanks @xprazak2 and sorry for such long review time... merging!

PR #4584, opened 2017-06-13
fixes #19995 - search now uses turbolinks

ohadlevy:
this avoids full page reloads when using search.

this also adds the urijs package, and while it can be done
without it, I prefered using it in order to be able to
support later on the per_page feature.

### DIFF
--- a/webpack/assets/javascripts/foreman_tools.js
+++ b/webpack/assets/javascripts/foreman_tools.js
@@ -1,4 +1,5 @@
 import $ from 'jquery';
+import URI from 'urijs';
 
 export function showSpinner() {
   $('#turbolinks-progress').show();
@@ -79,4 +80,20 @@ export function initTypeAheadSelect(input) {
     },
     width: '400px'
   });
+}
+
+// handle table updates via turoblinks
+export function updateTable(element) {
+  let uri = new URI(window.location.href);
+  let search;
+
+  if (element !== undefined) {
+    search = $(element).find('.autocomplete-input').val();
+    if (search !== undefined) {
+      uri.setSearch('search', search.trim());
+    }
+  }
+  /* eslint-disable no-undef */
+  Turbolinks.visit(uri.toString());
+  return false;
 }
--- a/app/views/common/_searchbar.html.erb
+++ b/app/views/common/_searchbar.html.erb
@@ -1,4 +1,7 @@
-<%= form_tag send("#{auto_complete_controller_name}_path"), :method => "get",:id =>"search-form" do %>
+<%= form_tag send("#{auto_complete_controller_name}_path"),
+            method: 'get',
+            id: 'search-form',
+            onsubmit: 'return tfm.tools.updateTable(this);' do %>
   <div class="input-group">
     <%= auto_complete_search(:search, params[:search].try(:squeeze," "), :placeholder => _("Filter") + ' ...').html_safe %>
     <span class="input-group-btn">
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "redux-thunk": "^2.2.0",
     "seamless-immutable": "^7.0.1",
     "select2": "~3.5.2-browserify",
+    "urijs": "^1.18.10",
     "uuid": "^3.0.1"
   },
   "scripts": {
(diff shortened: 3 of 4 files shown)


ohadlevy:
before:
before

after:
after

ohadlevy:
@tbrisker, @matanwerbner  any idea how should i test this? thanks.

ohadlevy:
test failures seems unrelated.

ohadlevy:
added tests.

webpack/assets/javascripts/foreman_tools.js
```diff
@@ -80,3 +81,22 @@ export function initTypeAheadSelect(input) {
     width: '400px'
   });
 }
+
+// send get forms through turbolinks
+export function tableUpdated(element) {
+  let uri = new URI(window.location.href);
+  let search;
+
+  if (element !== undefined) {
```
tbrisker: Is there any place the search field doesn't have id `#search` or isn't the child `.autocomplete-input`? I would rather use a sane default then a condition.

boaz0 (approved):
LGTM

webpack/assets/javascripts/foreman_tools.js
boaz0: maybe you can use a ternary operator here:
```
search = (element !== undefined) ? $(element).find('.autocomplete-input') : $('#search')
```
Nevertheless, I am not sure how readable will it be.

ohadlevy: thanks! I wrote it like that as I know https://github.com/theforeman/foreman/pull/4550/files#diff-8bb38b9ee188f2de0368a1fcad1ede41R87 (once it will be merged) will probably use this code path too..

tbrisker:
Looks good and works well, just ~one~ a few inline comment.

webpack/assets/javascripts/foreman_tools.js
```diff
@@ -80,3 +81,22 @@ export function initTypeAheadSelect(input) {
...
+    search = $(element).find('.autocomplete-input');
+  } else {
+    search = $('#search');
+  }
+
+  if (search[0] !== undefined) {
```
tbrisker: can this condition actually occur? that means there is no search element present on the page

webpack/assets/javascripts/foreman_tools.js
ohadlevy: One of the reasons we can't have multiple implementation of search in the same page is because of the ID selector (see roles and filters). I would like to decouple ourself from that pattern and allow using multiple search fields when appropriate.

webpack/assets/javascripts/foreman_tools.js
```diff
@@ -80,3 +81,22 @@ export function initTypeAheadSelect(input) {
     width: '400px'
   });
 }
+
+// send get forms through turbolinks
+export function tableUpdated(element) {
```
tbrisker: maybe name this more descriptively, something like searchTable or something? the name is a bit confusing

webpack/assets/javascripts/foreman_tools.js
ohadlevy: This is maybe premature, but the idea behind this function is to be called when table related data has changed, including per_page pagination change. in rare cases where there is no search for a given table that can happen.

webpack/assets/javascripts/foreman_tools.js
ohadlevy: see my previous comment about this? how would you call it?

ohadlevy:
@tbrisker rebased and addressed your comments.

ohadlevy:
@tbrisker test failure seems unrelated. @dLobatog another one for your list?

webpack/assets/javascripts/foreman_tools.js
```diff
@@ -80,3 +81,19 @@ export function initTypeAheadSelect(input) {
...
+export function updateTable(element) {
+  let uri = new URI(window.location.href);
+  let search;
+
+  if (element !== undefined) {
+    search = $(element).find('.autocomplete-input').val();
+    if (search !== undefined) {
+      uri.setSearch('search', search.trim());
+    }
+  /* eslint-disable no-undef */
```
tbrisker: nitpick: should be on next line

tbrisker (changes requested):
just a tiny nitpick and I'll merge.

ohadlevy:
@tbrisker updated.

tbrisker (approved):

tbrisker:
Thanks @ohadlevy !

tbrisker:
Please make sure to add uri.js to https://github.com/theforeman/foreman-packaging/tree/rpm/develop

PR #4587, opened 2017-06-14
Fixes #19696 - Validate email addresses with extra dots.

rabajaj0509:
This commit adds a regex expression that validates the email addresses for extra dots.

### DIFF
--- a/app/validators/email_validator.rb
+++ b/app/validators/email_validator.rb
@@ -6,7 +6,7 @@ class EmailValidator < ActiveModel::EachValidator
       address = value.split("@")
       encoded_address = address.count == 2 ? Mail::Encodings.decode_encode(address[0],:encode) + '@' + Mail::Encodings.decode_encode(address[1],:encode) : value
       m = Mail::Address.new(encoded_address)
-      r = m.domain.present? && m.address == value
+      r = m.domain.present? && m.address == value && !m.domain.include?('..')
     rescue Mail::Field::ParseError => exception
       Foreman::Logging.exception("Email address is invalid", exception)
       r = false


tbrisker:
ok to test

tbrisker:
@rahulbajaj0509 Looks like tests are failing with this change. Also, since this uses the email validator, please make the changes to the validator itself which you can find in `app/validators/email_validator.rb`.

lzap:
Yep, never test emails with regexp, there can be much more weird characters in emails than you think :-)

rabajaj0509:
@lzap then what do you think could be the solution rather than using a regexp ?

@tbrisker what is your call on using regexp ?

rabajaj0509:
I have gone through this bug carefully and come to the conclusion that this has not been handled by the 'Mail' gem itself. There is no problem as such with our code. 

I have also raised an issue in the 'Mail' gem : 
https://github.com/mikel/mail/issues/1124

rabajaj0509:
After raising an error with the mail gem, it seems that they have left out this feature intentionally for some reason. Therefore i have made the required changes in the code to handle consecutive dots in the e-mail address :)

dLobatog:
@rahulbajaj0509 I would follow the advice at the threads pointed out by the Mail maintainer. These emails can be seen in the wild even if they don't adhere to the RFC:

https://github.com/swiftmailer/swiftmailer/issues/239
https://github.com/swiftmailer/swiftmailer/issues/71

To be fair I would close the issue and the attached bugzilla as won't fix, it's not worth being that restrictive (more so than GMail, for example, try sending emails to .....................youraddress......@gmail.com

dLobatog (commented):
@rahulbajaj0509 I would follow the advice given by the Mail maintainer on the threads he posted. The point is these mail addresses violate the RFC but are common on the wild. Validating them means that in some time we would get reports saying they don't work:

https://github.com/swiftmailer/swiftmailer/issues/71
https://github.com/swiftmailer/swiftmailer/issues/239

I don't think being more restrictive than GMail (try sending an email to y.o.ur.....address....@gmail.com - it will get delivered to -> youraddress@gmail.com just fine) is worth it, especially when these addresses are used by some people. Feel free to close the attached Bugzilla as NOTABUG/WONTFIX

rabajaj0509:
@dLobatog  thanks for your input on this one :)

PR #4588, opened 2017-06-14
Fixes #20013 - Notification singleton polling

matanwerbner:
This fixes a bug with duplicate polling started when the notification component is re-mounted (as a result of turbolinks navigation). the fix is to set the isPolling flag in the state.notifications to true.
also, refactored some of the notifications code, added a test, and upgraded jest to v20!


webpack/assets/javascripts/react_app/components/notifications/index.js
```diff
@@ -8,9 +8,11 @@ import { groupBy, some, isUndefined } from 'lodash';
 
 class notificationContainer extends React.Component {
   componentDidMount() {
-    const { getNotifications, data: { url } } = this.props;
+    const { startNotificationsPolling, data: { url }, isPolling } = this.props;
 
-    getNotifications(url);
+    if (!isPolling) {
+      startNotificationsPolling(url);
```
ohadlevy: shouldn't this be defined at the pulling action? why would the component know about it (or care)?

webpack/assets/javascripts/react_app/components/notifications/index.js
```diff
@@ -8,9 +8,9 @@ import { groupBy, some, isUndefined } from 'lodash';
 
 class notificationContainer extends React.Component {
   componentDidMount() {
-    const { getNotifications, data: { url } } = this.props;
+    const { startNotificationsPolling, data: { url } } = this.props;
 
-    getNotifications(url);
+    startNotificationsPolling(url);
```
ohadlevy, line 13: another question: would it be better to change the react mounter to not mount if the div is already mounted? or in other words, not allow turbolinks or some other ajax call to remount an already mounted component? this way we will not need to face any of these challenges?

matanwerbner:
As TL replaces the html in the body node of the document, a mount will have to happen. Relying on state to tell if polling should start is the only way i can see currently.

PR #4590, opened 2017-06-14
Fixes #17204 - Sort setting UI category tabs consistently

swapab:


sean797:
Thanks @swapab, just want to get someone else recommendation on ordering. @Rohoover can you comment please? This sets "General" first, then the remainder alphabetically.

Here's a screenshot for reference:
screenshot from 2017-06-14 17-56-41

Rohoover:
@sean797 
I can't really comment on the order unless there was data showing most used options from the user. I think the approach you took is reasonable without said information.

swapab:
@sean797 @Rohoover Thanks for the review.
The order was set to give an ideal sense to it as opposed to random one.
Given the fact that in future we would have analytics of _most used options_ it would be easier to shuffle around accordingly.

app/models/setting.rb
```diff
@@ -71,7 +72,15 @@ def self.config_file
...
   def self.live_descendants
-    self.disabled_plugins.default_organization.organization_fact.default_location.location_fact
+    self.disabled_plugins.default_organization.organization_fact.default_location.location_fact.order_by(:category)
+  end
+
+  def self.stick_general_to_first
+    where(:category => sticky_setting).concat(where.not(:category => sticky_setting)).group_by(&:category)
+  end
+
+  def self.sticky_setting
```
sean797: Can you move this to a private method?

app/models/setting.rb
```diff
@@ -71,7 +72,15 @@ def self.config_file
   end
 
   def self.live_descendants
-    self.disabled_plugins.default_organization.organization_fact.default_location.location_fact
+    self.disabled_plugins.default_organization.organization_fact.default_location.location_fact.order_by(:category)
+  end
+
+  def self.stick_general_to_first
```
sean797: Nitpick, can this be `self.stick_general_first` ?

sean797 (changes requested):
Thanks @Rohoover for taking a look 🥇 @swapab, I made 2 small comments, I agree the ordering seems sane, Thanks! 👍

Edit: Also tests are broken, can you take a look?

swapab:
Requested changes were done.
All tests passing.

app/models/setting.rb
swapab: I am not a fan of private class methods. Moved it to a variable in `stick_general_first`.

app/models/setting.rb
```diff
@@ -62,6 +62,7 @@ def validate(record)
   scope :organization_fact, -> { where('name not in (?)', 'organization_fact') unless Taxonomy.enabled_taxonomies.include? 'organizations' }
   scope :default_location, -> { where('name not in (?)', 'default_location') unless Taxonomy.enabled_taxonomies.include? 'locations' }
   scope :location_fact, -> { where('name not in (?)', 'location_fact') unless Taxonomy.enabled_taxonomies.include? 'locations' }
+  scope :order_by, ->(attr) { except(:order).order(attr) }
```
dLobatog, line 65: What is the `except(:order)` for?

app/models/setting.rb
```diff
@@ -71,7 +72,12 @@ def self.config_file
   end
 
   def self.live_descendants
-    self.disabled_plugins.default_organization.organization_fact.default_location.location_fact
+    self.disabled_plugins.default_organization.organization_fact.default_location.location_fact.order_by(:category)
+  end
+
+  def self.stick_general_first
+    sticky_setting = 'Setting::General'
+    where(:category => sticky_setting).concat(where.not(:category => sticky_setting)).group_by(&:category)
```
dLobatog, line 80: The issue mentions the rest of the settings should be ordered alphabetically, where is that?

dLobatog (changes requested):
Thanks @swapab - I see the results are alphabetically ordered but i'm not sure where is that coming, could you please address the 2 questions inline?

app/models/setting.rb
swapab: The `except(:order)` is to remove `default_scope` https://github.com/swapab/foreman/blob/develop/app/models/setting.rb#L57
I feel the `default_scope` is not required and should be replaced with explicit sorting wherever required.

app/models/setting.rb
swapab: The sort query has been added in `Setting::live_descendants`
https://github.com/swapab/foreman/blob/e55c6e4b8a486a34e81bef048853031c890f5bb7/app/models/setting.rb#L75

ohadlevy:
whats the status of this pr? thanks

swapab:
@ohadlevy It needs a re-review.
I have answered @dLobatog's queries https://github.com/theforeman/foreman/pull/4590#pullrequestreview-44315495

dLobatog (approved):

dLobatog:
Sorry for the delay, merged! Thanks @swapab!

swapab:
@dLobatog No problem.Thanks! :smiley:

ares:
Thanks @dLobatog for merging, I set the redmine release to 1.16

PR #4591, opened 2017-06-14
Sync 1.15-stable community templates

ares:


dLobatog (commented):
@ares This will be done right before we release 1.15.2, why doing it now?

ares:
It seems it didn't happen before 1.15.1 but I suppose that's because it was done the same day it was cherry-picked to 1.15-stable in community-repo. I wanted to make sure it will get in this time, but I'm fine closing this if you have that covered.

PR #2230, opened 2015-03-08
Fixes #9591 - Added support for "inherit" state.

ShimShtein:
Changed the UI so it will possible to create a dropdown with empty value and "inherit" option in the same time.
On the server side, I check for existence of the field in the update to check if I need to inherit the value (if the field does not exists - its value will be inherited).
I did not change the notion of "inherit" in hosts, it is more of a "copy" notion - it copies the value from the hostgroup, but does not change it, if the value is changed. The change will be performed on the first update to the host object.

PR #4593, opened 2017-06-14
Fixes #5790 - Add compute resource to hostgroup

xprazak2:


app/helpers/form_helper.rb
```diff
@@ -330,11 +330,11 @@ def check_required options, f, attr
     return ' *'.html_safe if required.nil? ? is_required?(f, attr) : required
   end
 
-  def blank_or_inherit_f(f, attr)
+  def blank_or_inherit_f(f, attr, blank_value = _("no value"))
```
ares: would it be better to use options hash? I'd like to avoid changing the method signature in future when we need one more optional argument.

xprazak2:
I changed the third param into option hash

lzap:
This PR is likely linked to wrong issue named "Specify default value for "Hardware Model" when creating new host".

lzap:
Can you explain the motivation for this change? There is nothing in the description and issue number is perhaps incorrect.

xprazak2:
The issue number is correct, on hosts/index page, the 'Model' column in table shows what compute resource the host is deployed to or its hardware model if it is a Bare metal. This adds compute resource to hostgroup, so users get the default value by selecting a hostgroup with compute resource.

xprazak2:
[test upgrade]

app/helpers/form_helper.rb
```diff
@@ -330,11 +330,11 @@ def check_required options, f, attr
...
   end
 
-  def blank_or_inherit_f(f, attr)
-    return true unless f.object.respond_to?(:parent_id) && f.object.parent_id
+  def blank_or_inherit_f(f, attr, opts = { :inherited_blank_value => _("no value"), :blank_value => true })
+    return opts[:blank_value] unless f.object.respond_to?(:parent_id) && f.object.parent_id
     inherited_value   = f.object.send(attr)
     inherited_value   = inherited_value.name_method if inherited_value.present? && inherited_value.respond_to?(:name_method)
-    inherited_value ||= _("no value")
+    inherited_value ||= opts[:inherited_blank_value]
```
ripcurld0: shouldn't `opts(...)` be wrapped with `_()`?

boaz0 (commented):
shouldn't `opts(...)` be wrapped with `_()`?

boaz0 (commented):
Nevermind, LGTM

tbrisker:
I think this actually resolves http://projects.theforeman.org/issues/5790. While the original issue talks about model field, it seems the user is unaware that this defaults to the compute resource if one is present. When deploying on a compute resource, the hardware model field is not displayed at all, so I'm not sure where exactly they see this field at all? Maybe they see this field before selecting the compute resource to deploy on?

ares:
I asked the original reporter for clarification but I though the "deploy on" was meant. I suggest @xprazak2 changes the issue linked in commit and after we hear from the reported we either mark 19043 as duplicate of 5790 or keep it as separate issue. Thanks!

xprazak2:
I linked this to 5790 instead of 19043.

fwdIT:
Thx for the effort @xprazak2 !
I was also working on this more than a year ago ( http://projects.theforeman.org/issues/13538 with the help of @tbrisker ) but got stuck since the initial code I already had was not compatible with the current code in the dev branch anymore when I wanted to continue this year. And I didn't have the time yet to look further into this while it is something that I can really use :) 👍

xprazak2:
test failures unrelated - js integration tests are failing and katello cannot find a required file

app/helpers/form_helper.rb
```diff
@@ -330,11 +330,11 @@ def check_required options, f, attr
     return ' *'.html_safe if required.nil? ? is_required?(f, attr) : required
   end
 
-  def blank_or_inherit_f(f, attr)
-    return true unless f.object.respond_to?(:parent_id) && f.object.parent_id
+  def blank_or_inherit_f(f, attr, opts = { :inherited_blank_value => _("no value"), :blank_value => true })
```
tbrisker: you can use kwargs here instead of an opts hash

app/views/hostgroups/_form.html.erb
```diff
@@ -27,10 +27,19 @@
...
       <%= hostgroup_puppet_environment_field f %>
-      <%= select_f(f, :compute_profile_id, ComputeProfile.visibles, :id, :name,
+
+      <%= select_f f, :compute_resource_id, accessible_resource(@hostgroup, :compute_resource), :id, :to_label,
+          { :include_blank => blank_or_inherit_f(f, :compute_resource, :inherited_blank_value => _('Bare Metal'), :blank_value => _('Bare Metal')) },
+          { :label => _('Deploy on'),
+            :onchange => 'tfm.hostgroups.computeResourceChanged(this);',
+            :help_inline => :indicator } if SETTINGS[:unattended] %>
+
+      <div id="compute_profile" <%= display?(!@hostgroup.compute_resource_id) %> >
```
tbrisker: I wonder if we should actually hide this if there is no compute resource set - maybe users want to just set the profile at the hostgroup level (as they can already)? 
Also, this isn't displayed if the compute resource is inherited from the parent.

tbrisker (dismissed):
See a couple of inline comments. Additionally, the host form should also have this field show the inherit option like other fields that may be inherited from hs.

app/views/hostgroups/_form.html.erb
xprazak2: I hide this to keep consistent behaviour with hosts form. If users just set the compute profile, then they have 'Bare Metal' as default compute resource and that combination does not make much sense to me.

tbrisker: consider this hypothetical hostgroup heirarchy:
```
large_servers
large_servers/aws
large_servers/vmware
large_servers/ovirt
```
you want all of them to have a large profile, but deploy on different resources.

xprazak2: Yes, but you can achieve it with hiding the field as well: 

```
large_servers
  - deploy on : aws
  - compute profile: large

large_servers/aws
  - deploy on: aws (Inherit parent)
  - compute profile: large (Inherit parent)

large_servers/vmware
  - deploy on: vmware
  - compute profile: large (Inherit parent)
```
Hiding the field prevents you from having:
```
large_servers
  - deploy on: Bare metal
  - compute profile: large
```

but I guess this is just a matter of taste, so I made changes as suggested.

xprazak2:
test failure unrelated - cannot allocate memory to scan for webpack dirs

tbrisker:
Still have a few issues:
1. When selecting a CR in a parent host-group, child cannot override to bare metal
2. On host creation, sometimes selecting a HG that has a CR set doesn't change the CR field on the host form, not sure why it only works for some HGs.
3. The CR input on the host form doesn't have the "Inherit" button like other attributes that can be inherited from the HG:
image

xprazak2:
@tbrisker, thanks for review, I think the issues 2) and 3) may be out of scope. Note that I am not touching the hosts form, I only aim to add one field to hostgroups form and not to fix existing problems with hosts form. But I can try to include fix for host here if you would like me to.

Issue 1) is valid, but not easily fixable. Both `Bare Metal` and `Inherit from parent (whatever)` are blank values and I cannot very well have 2 blank values in one dropdown. To fix this in a nice and consistent way, I can imagine we could have Bare metal as a regular compute resource, but again, this would be a big change with impact on multiple areas.

Let me know how you want me to proceed.

tbrisker:
I disagree that 2 and 3 are out of scope - the whole point of adding the CR to the hostgroup is to allow hosts to inherit them. 
1 might need a different approach, possibly we won't have a choice other then adding a fake bare metal cr - @ShimShtein maybe you have an idea (how we can override a parent hg's cr to baremetal)?

iNecas:
Not diving too much into details, I agree 2 and 3 should be addressed here, unless proven there is something too big and overlaping non-related areas that would make the PR out of control

ShimShtein:
Thanks @tbrisker for tagging me.

> 1) is valid, but not easily fixable. Both Bare Metal and Inherit from parent (whatever) are blank values and I cannot very well have 2 blank values in one dropdown.

Actually, (3) is the solution to this problem :smile:.
This is the whole point of the inherit button: it enables three states: nothing selected, option selected and inheritance required. This is what https://github.com/theforeman/foreman/pull/2230 is for. In order to use this feature, you should add relevant fields to `#hostgroup_inherited_attributes` array and that's pretty much it for the server side. For the client side, you should look at [compute_profile](https://github.com/theforeman/foreman/blob/05dc994fa97f9d92091c92c4965b9489dc39f9ff/app/views/hosts/_form.html.erb#L65) field implementation, it's pretty straight forward.

xprazak2:
I made the following changes:
 * I added the inherit button for compute resource to hosts form, so new hosts can inherit the value (or override it)
 * I added the inherit button to hostgroup, now children can override to bare metal. I changed other dropdowns to use the select with button to have consistent look, we will need additional changes in katello and foreman_openscap

ShimShtein (changes requested):
While playing with this PR, got some strange results:

1. For hostgroup, the inheritance is lost on edit: create parent with compute resource, create child with "inherit" set for CR -> CR is selected correctly, edit the child -> no CR is selected.
2. Less important, but still an issue, on toggling "inherit" clear the value of the select, or put the right one there. Right now it stays on the last selected value, which is a bit confusing.
3. When creating a host, the "Virtual Machine" tab is not updated according to compute resource from the hostgroup on HG selection.
4. Host's CR field does not get updated when selecting an HG with bare metal value: create parent HG with some CR (libvirt for example), create child HG with bare metal CR, go to host creation form, and select parent HG, then change to child. Another test: go to host creation, select some CR (without HG) then select child HG.

xprazak2:
Most quirks related to dropdown behaviour should be fixed now. Even though I added the inherit button, we still cannot override compute resource to 'Bare metal' in child hostgroup:

1) create a parent with a compute resource
2) create a child, try to override to 'Bare metal', save
3) go to edit child, notice the compute resource is not 'Bare metal' but is inherited from parent

Reason for this is that a value is inherited from parent/ancestor if child does not have any value and ancestor does. I think that is also why we do not allow to override to empty value in other dropdowns. The buttons work in hosts form because we do not use the ancestry inheritance but set the values directly from host group.

I'll revert the fields in hostgroups form as they were. I do not think not being able to override to 'Bare metal' is a big deal, but if we want it, we need to introduce a fake bare metal resource.

ShimShtein:
Without actually looking at the code, I think it's a leftover of the old inheritance mechanism - look for the field name in declarations other than `#hostgroup_inherited_attributes` and exclude it from there.

xprazak2:
@ShimShtein, could you please elaborate on your last comment? I am not sure what I should remove from where and I struggle to understand how I would tell from my models when the compute resource is meant to be inherited from parent and when not:

xprazak2:
I reverted back to a simpler solution. I removed the form field buttons from hostgroups form, they work for host, but not for hostgroup. The reason is that we handle host < hostgroup inheritance on our own while hostgroup < hostgroup uses the ancestry gem and there is no way to tell if nil attribute overrides the one from parent to nil or if we want to inherit. 

Currently, no attribute in child hostgroup can be overriden to empty value if the attribute is set in parent to non-empty. So everything should work, you just cannot override to 'Bare metal' in child hostgroup.

dLobatog (dismissed):
@xprazak2 Seems like tests are not running, can you rebase to the current develop?

xprazak2:
I rebased.

ShimShtein:
> I reverted back to a simpler solution.

:+1: @xprazak2. I agree with your decision. Eventually, I think that bare metal will get its own value, but right now, the only way to achieve full inheritance, would be by adding additional boolean field to hostgroup that will store the should_inherit flag. Personally I am against this solution.

xprazak2:
Rebased.

ares:
@ShimShtein does that mean ACK? @tbrisker, do you still require some changes?

ShimShtein (approved):
@ares ACK from me.

db/migrate/20170604082313_add_compute_resource_to_hostgroup.rb
```diff
@@ -0,0 +1,9 @@
+class AddComputeResourceToHostgroup < ActiveRecord::Migration
+  def up
+    add_column :hostgroups, :compute_resource_id, :integer
```
ares: should this define also a foreign key?

ares (changes requested):
Seems to work well, just one comment re foreign keys and then, it should be good to merge.

mmoll:
@xprazak2 what's the status here? this looks 99% ready...

xprazak2:
I added the foreign key and rebased.

ares (approved):

ares:
Test failure unrelated ComputeProfileJSTest.test_0002_create compute profile, if there are not more comments from @tbrisker I guess we can merge.

ares:
Thanks @xprazak2, merging!

PR #4595, opened 2017-06-15
Fixes #16807 - remove permission edit_users for test_mail

amirfefer:
Test email action requires that the user has either `create` or `edit_users`, which is  unnecessary.


ohadlevy:
thanks @amirfefer

PR #1301, opened 2014-03-17
lookup keys in puppet class form visual changes

abenari:
Make lookup keys list in puppet class form look more like the rest of the product, and improve the form usability.

PR #2052, opened 2015-01-07
fixes #8852 - add token-based authentication feature for API

isratrade:

PR #4597, opened 2017-06-16
fixes #1301 - user login with access token

timogoebel:
This PR allows users to do API requests with a personal access token instead of their password.

```
curl -u admin:Vvu5gzVNEUfuXaqDawvepg http://localhost:3000/api/v2/subnets
```

As this is security related: I tried my best to think of every possible way to make this safe, but please do check this PR extra carefully.
One restriction I found: If a user has his account disabled in ldap, he could still login via a personal access token right now.

The bootstrap datepicker i18n is a mess. I tried to find a solution to this, but let's say: It's open for debate.

API implementation is missing for now to make this easier to review.

This is how the UI looks:

image

TODOs:

- [ ] proper i18n
- [ ] ExpiresAt Field Datepicker does not work / validate correctly
- [ ] Permissions are not handled at UI level

and probably more...


Gemfile
```diff
@@ -38,6 +38,8 @@ gem 'deacon', '~> 1.0'
 gem 'webpack-rails', '~> 0.9.8'
 gem 'mail', '~> 2.6'
 gem 'sshkey', '~> 1.9'
+gem 'bootstrap-datepicker-rails'
+gem 'where-or'
```
timogoebel: This is just a backport of rails 5 functionality.

Gemfile
```diff
@@ -38,6 +38,8 @@ gem 'deacon', '~> 1.0'
 gem 'webpack-rails', '~> 0.9.8'
 gem 'mail', '~> 2.6'
 gem 'sshkey', '~> 1.9'
+gem 'bootstrap-datepicker-rails'
```
timogoebel: I know, we don't like gemified assets anymore. But I couldn't get the i18n working with the node version. And this is already packaged for plugins.

app/assets/javascripts/application.js
```diff
@@ -547,3 +547,22 @@ function clearError(field) {
   var error_block = form_group.children(".help-inline").children(".error-message");
   error_block.remove();
 };
+
+// Make sure the correct tab is displayed when loading the page with an anchor,
```
timogoebel: Just moved this code from `proxy_status.js`

ohadlevy: I know I'm annoying, but can me move stuff away from app/assets into webpack and jest? how about tfm.tools.forms.setTab or similar?

app/assets/javascripts/bootstrap_datepicker.js.erb
```diff
@@ -0,0 +1,16 @@
+//= require bootstrap-datepicker
+
+$(document).on('ContentLoad', function () {
```
ohadlevy: I'll try about finding a way to handle the locale, there has to be a better way than js.erb :)

app/controllers/personal_access_tokens_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    token_value = @personal_access_token.generate_token
+    if @personal_access_token.save
+      flash[:personal_access_token] = token_value
+      process_success :success_redirect => edit_user_path(@user, :anchor => "personal_access_tokens")
+    else
+      process_error
+    end
+  end
+
+  def revoke
```
ohadlevy: how about s/revoke/destroy ? this will be more in align with REST and will not require special handing in routes file?

app/helpers/application_helper.rb
```diff
@@ -518,7 +518,7 @@ def user_set?(field)
   end
 
   def notifications
-    content_tag :div, :id => 'notifications', :'data-flash' => flash.to_json.html_safe do
+    content_tag :div, :id => 'notifications', :'data-flash' => flash.to_a.select {|flash| ['notice', 'success', 'info'].include?(flash.first) }.to_json.html_safe do
```
ohadlevy: why is this needed?

app/controllers/personal_access_tokens_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def new
+    @personal_access_token = PersonalAccessToken.new
+  end
+
+  def create
+    @personal_access_token = PersonalAccessToken.new(personal_access_token_params.merge(:user => @user))
+    token_value = @personal_access_token.generate_token
+    if @personal_access_token.save
+      flash[:personal_access_token] = token_value
```
ohadlevy: why are you using flash to transfer the token? it can leak into tmp cache and others? wouldn't an AJAX call with the content be better?

app/helpers/form_helper.rb
```diff
@@ -399,4 +399,24 @@ def ie_multipart_fix
...
+
+  def input_group_addon(content, options = {})
+    content_tag :span, class: 'input-group-addon' do
+      content_tag :span, content, options
+    end
+  end
+
+  def datepicker_f(f, attr, options = {})
+    options = {
+      :class => 'bsdatepicker',
```
ohadlevy: based on http://www.patternfly.org/pattern-library/forms-and-controls/datepicker/#/code/html-css the extra css class is not needed?

app/helpers/users_helper.rb
```diff
@@ -28,4 +28,9 @@ def user_taxonomies_html_options(user)
   def mail_notification_query_builder(mail_notification, f)
     render :partial => "#{mail_notification}_query_builder", :locals => {:f => f, :mailer => mail_notification.name } if mail_notification.queryable?
   end
+
+  def expires_label(date)
+    return _('Never') if date.nil?
+    _("%s ago") % time_ago_in_words(date)
```
ohadlevy: shouldnt expire be in the future ? (maybe its used correctly,  i just find the terms confusing).

app/models/user.rb
```diff
@@ -43,6 +43,7 @@ class User < ApplicationRecord
   has_many :cached_usergroup_members
   has_many :widgets, :dependent => :destroy
   has_many :ssh_keys, :dependent => :destroy
+  has_many :personal_access_tokens, :dependent => :destroy
```
ohadlevy: maybe also add has one current token ?

app/models/user.rb
```diff
@@ -197,14 +198,21 @@ def self.anonymous_console_admin
...
       # user has an authentication method and the authentication was successful
-      if user.auth_source && (attrs = user.auth_source.authenticate(login, password))
+      if api_request && user.authenticate_by_personal_access_token(password)
+        logger.debug("Authenticated user #{user.login} with a Personal Access Token.")
+
+        unless user.auth_source.valid_user?(user.login)
+          logger.debug "Failed to find #{user.login} in #{user.auth_source} authentication source"
+          user = nil
+        end
+      elsif user.auth_source && (attrs = user.auth_source.authenticate(login, password))
```
ohadlevy: I wish all of this logic would be extracted to a class...

app/views/personal_access_tokens/_personal_access_tokens_tab.html.erb
```diff
@@ -0,0 +1,61 @@
+<div class="tab-pane" id="personal_access_tokens">
+  <% if flash[:personal_access_token] %>
+    <div class="alert alert-success">
+      <span class="pficon pficon-ok"></span>
```
ohadlevy: nitpick: we have helpers for icons and alerts

app/views/personal_access_tokens/_personal_access_tokens_tab.html.erb
```diff
@@ -0,0 +1,61 @@
...
+      <strong><%= _('Your New Personal Access Token') %></strong><br />
+      <code id="personal-access-token" style="font-size: 120%;"><%= flash[:personal_access_token] %></code>
+      &nbsp;
+      <a id="personal-access-token-clipboard" data-clipboard-target="#personal-access-token" title="<%= _('Copy to clipboard!') %>">
```
ohadlevy: nice, i remember we had another place we wanted to use it (maybe rex?) can't think of it it atm. /cc @dLobatog @ares ?

app/views/personal_access_tokens/_personal_access_tokens_tab.html.erb
```diff
@@ -0,0 +1,61 @@
+<div class="tab-pane" id="personal_access_tokens">
+  <% if flash[:personal_access_token] %>
```
ohadlevy: why not use a simple class / attribute to identify this? I'm uncertain on the usage of flash?

app/views/personal_access_tokens/_personal_access_tokens_table.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  <thead>
+    <tr>
+      <th class="col-md-5"><%= _('Name') %></th>
+      <th class="col-md-2"><%= _('Created') %></th>
+      <th class="col-md-2"><%= _('Expires') %></th>
+      <th class="col-md-1"><%= _('Actions') %></th>
+    </tr>
+  </thead>
+  <tbody>
+    <% tokens.each do |token| %>
```
ohadlevy: i believe you can do instead:
```ruby
<%= render partial: "token_row", collection: tokens %>
```

app/views/personal_access_tokens/new.html.erb
```diff
@@ -0,0 +1,9 @@
+<% title _('Create Personal Access Token for %s') % @user.name %>
+
+<ol class="breadcrumb">
```
ohadlevy: :+1: maybe worth extracting to a helper?

test/controllers/personal_access_tokens_controller_test.rb
```diff
@@ -0,0 +1,41 @@
+require 'test_helper'
+
+class PersonalAccessTokensControllerTest < ActionController::TestCase
+  let(:token) { FactoryGirl.create(:personal_access_token)  }
+  let(:user) { token.user }
+
+  test 'new' do
```
ohadlevy: you can use https://github.com/theforeman/foreman/blob/develop/test/controllers/shared/basic_rest_response_test.rb and replace some of the tests like at https://github.com/theforeman/foreman/blob/develop/test/controllers/architectures_controller_test.rb#L8

ohadlevy (changes requested):
@timogoebel let me start by saying this is a very welcome change! thanks for your effort on this!

there are two "additional features" in this context that I would like to discuss prior to merging this PR.
1. should we allow to create multiple tokens which different permissions? e.g. some user that wants to use his token in some way but not always with all of his permissions (very common usage case with github for example).

2. should we help the user know where his token is being used? for example i always wanted to add something like https://github.com/blog/1661-modeling-your-app-s-user-session

thanks!

app/assets/javascripts/bootstrap_datepicker.js.erb
timogoebel: Basically, we could define a custom language `foreman` and then use our default means for translations.
http://bootstrap-datepicker.readthedocs.io/en/latest/i18n.html

app/controllers/personal_access_tokens_controller.rb
timogoebel: In the controller, the token is generated only once and only the sha is stored in the database. So I cannot access the data via a ajax call. The controller could save an empty token model and then a ajax call could generate the token, update the database model and return the token in cleartext. Wanted to keep this simple, though.

app/helpers/application_helper.rb
timogoebel: The toast notifications js assumes, that the flash hash only contains `['notice', 'success', 'info']` keys.

app/helpers/form_helper.rb
timogoebel: The class is required for the js initialization. If you use `data-provides` it does not translate the datapicker.

app/controllers/personal_access_tokens_controller.rb
ohadlevy: so why not simply pass that as a @variable to the view? why do you need to use flash?

timogoebel:
@ohadlevy: Thanks for the good review.

> should we allow to create multiple tokens which different permissions? e.g. some user that wants to use his token in some way but not always with all of his permissions (very common usage case with github for example).

Yes. Definitely. I think, it makes a lot of sense to add a scope to a token. Users might just want to grant read-only access to users. However I think, this won't be a trivial feature to implement and we need to be careful not to implement the permissions system a second time. Therefor I consider this out of scope for this PR.

> should we help the user know where his token is being used? for example i always wanted to add something like https://github.com/blog/1661-modeling-your-app-s-user-session

Yep. This definitely makes sense. We could add a "last used" column to the token model that we update on every login. But I'd also do this in a follow up ticket.

app/models/user.rb
timogoebel: I'm not sure, I follow. There isn't _one_ current token, is there?

app/controllers/personal_access_tokens_controller.rb
timogoebel: Yep, but it doesn't actually destroy the record. That's why `destroy` is wrong imho.

app/controllers/personal_access_tokens_controller.rb
timogoebel: @ohadlevy: I need to pass the data from the `personal_access_token` controller (which generates the token) to `users_controller` (which displays the token + index). The `personal_access_token` controller could just render a page that displays the token and the flash would not be necessary. However, I wanted to avoid this kind of page. Another option would be to do all this with client side javascript. But I'm not really a frontend guy, you know? :-)

app/views/personal_access_tokens/new.html.erb
timogoebel: @ohadlevy: What do you think of something like this?
https://github.com/weppos/breadcrumbs_on_rails

ohadlevy:
I would like to ask other reviewers (@ares, @dLobatog, @tstrachota  etc) for your feedback as this is a core sensitive access change.

Gemfile
ohadlevy: could you add a condition that this only runs when using rails < 5 ?

app/models/personal_access_token.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def self.authenticate_user(user, token)
+    self.active.where(:user => user, :token => encrypt_token(user, token)).any?
+  end
+
+  def self.token_salt(user)
+    Digest::SHA1.hexdigest(user.id.to_s)
+  end
+
+  def self.encrypt_token(user, token)
```
ares: this seems to hash the token, perhaps `s/encrypt/hash/` to avoid confusion

app/models/personal_access_token.rb
```diff
@@ -0,0 +1,52 @@
...
+    token_value = SecureRandom.urlsafe_base64(nil, false)
+    self.token = self.class.encrypt_token(user, token_value)
+    token_value
+  end
+
+  def revoke!
+    update!(revoked: true)
+  end
+
+  def expires?
```
ares: seems to be defined in `Expirable` already

app/registries/foreman/access_permissions.rb
```diff
@@ -807,4 +807,13 @@
                    :ssh_keys => [:destroy],
                    :"api/v2/ssh_keys" => [:destroy]
   end
+
+  permission_set.security_block :personal_access_tokens do |map|
+    map.permission :view_personal_access_tokens,
```
ares: this does not seem to be used anywhere, I'd expect this in view where tokens are listed, so we make sure that only user who owns tokens or another user with this permission can see them

app/models/user.rb
```diff
@@ -197,14 +198,21 @@ def self.anonymous_console_admin
   # If the user is not in the DB then try to login the user on each available authentication source
   # If this succeeds then copy the user's details from the authentication source into the User table
   # Returns : User object OR nil
-  def self.try_to_login(login, password)
+  def self.try_to_login(login, password, api_request = false)
```
ares: would it make sense to add options hash instead of third argument? I'd like to avoid situation where we want to add 4th optional param in future

package.json
```diff
@@ -17,6 +17,7 @@
     "babel-preset-es2015": "~6.6.0",
     "babel-preset-react": "^6.5.0",
     "babel-register": "^6.9.0",
+    "clipboard": "^1.7.1",
```
ares: we were able to do this without extra dependency in [plugin](https://github.com/theforeman/foreman_virt_who_configure/blob/fd523b4e06dce034ad232ac7eba24d2ca597385c/app/assets/javascripts/foreman_virt_who_configure/config_copy_and_paste.js), maybe same approach could be taken here or if not, it would be great if you provide similar JS function so others could use it instead of writing `<%= javascript_tag("new Clipboard('#personal-access-token-clipboard');"); %>`

timogoebel:
@ares , @ohadlevy : Thanks for the review.

I started to refactor this to an API-only / React thing. I want to have a react component, that shows a button. When the button is clicked, it shows the new token form and when the form is submitted, it shows success and the new token. The server implementation would just use the APIv2.
Redux and React still cause some headaches, though.
This would get rid of the flash to pass the token between controllers. Would that be a valid approach in your opinion?

ares:
I didn't do the full review, just left few comments here and there. In general, I think this is a good addition.

Comparing this to previous try (#2052) it seems more complete in terms of how user can manage tokens. The downside is that it's not based on standard oauth and is somewhat orthogonal to oauth authentication we already have in API. On the other hand, I like the simplicity of how it can be used from client side. I think we should decide whether we want to continue supporting our limited oauth (uses oauth protocol but with global username/password only) or start using personal access tokens and replace the oauth with token for internal admin account. Right now I find personal access tokens better, but I'd like to hear opinions from others too before saying final word.

What I find slightly confusing is that the token can be used as password. I think it would be nicer if it would be specified as some header, e.g. Foreman-API-Key (that's what redmine does) but I know github does it the same way so feel free to ignore me :-)

More specific to the PR, I think it needs to have API too. I'd recommend introducing it even in this PR, I expect it to be standard API CRUD controller.

tstrachota:
This is very nice feature!

I didn't have a chance to look closer at the code. I did just some testing, which resulted in bunch of notes and questions, most of them about permissions:
1. For some reason copy to clipboard didn't work for me (Firefox 54).
1. "Last used" doesn't get updated
1. Is it intentional that tokens can't be used for UI login? (not sure if there's any usecase for that)
1. What is the exact meaning of "create_personal_access_tokens" permission? When I have it, does it mean that I can create tokens for other users or that I can create tokens just for myself.
1. Does it make sense to have "view_personal_access_tokens" permission when in fact they're not visible from anywhere after you close the flash message?
1. I was able to create a personal token when I only had view_personal_access_token" permission
1. I can see "Revoke" button even when I don't have revoke permission. Clicking on it results in error: "PersonalAccessToken not found Please try to update your request"
1. When are the inactive tokens deleted?
1. I like the breadcrumbs. Since it's the first time we sue them,  @rohoover, do we have any style or usage guide for breadcrumbs?
1. Nitpick: it would be nice to replace "This user has any tokens" with "You don't have any tokens" on the "My account" page

tstrachota:
@ohadlevy ad 1. from https://github.com/theforeman/foreman/pull/4597#pullrequestreview-44718614 :
Limiting permissions of tokens to a subset of user's permissions would be very interesting feature. I think it's ok to just to explore what would be necessary changes, make sure the architecture in this PR is ready for such addition and leave it for future extension.

@ares ad password vs. header:
I'm in favour of going the github way and using tokens just as passwords. It's much easier to use this way and it plays nicely with the cli. Apart from it might not feel 100% clean,  are there any drawbacks of this approach?

ares:
@tstrachota nope, it's just unexpected and slightly confusing but it might be just me... we should make sure it would be well documented, but I don't see it as an obstacle

ohadlevy:
@timogoebel I'm more than happy to help with react part, tbh I'm uncertain if we need redux or not for this usage case, if you have some code I can try to help too.

dLobatog (commented):
Not wanting to be a grinch here, especially since this seems really thoroughly thought and implemented, but I would much rather move our login layer to Devise and rely on community-tested stuff, so we can get things like this https://github.com/lynndylanhurley/devise_token_auth for "free" rather than having to code our implementations (certainly less tested than Devise)

I'm not opposed to this being merged, just wanted to share my thoughts about further improvements to our login layer

Rohoover:
@tstrachota 

PatternFly has an existing breadcrumb https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/breadcrumbs.html

But we're also working on a new breadcrumb style but it more specifically applies to two-pane view removal:

mockup_d1_v4

ohadlevy:
@timogoebel this needs a rebase, also if I can help in any way please let me know..

timogoebel:
@ohadlevy : Rebased and pushed some React Code. My plan is to move the whole Form and Creation to a React Component. The component will then do a request to APIv2 to create the token.
It does not really work at all, I just created some boilerplate code. Next step would be to model the Form in React.
If you want to take a look, I added the component to the storyboard.

app/controllers/personal_access_tokens_controller.rb
ohadlevy: it destroys the REST verb, not the object, imho controller != models

webpack/assets/javascripts/react_app/redux/actions/users/personalAccessTokens/index.js
```diff
@@ -91,6 +91,6 @@ export const getTokens = userId => dispatch =>
     requestAction: USERS_PERSONAL_ACCESS_GET_REQUEST,
     successAction: USERS_PERSONAL_ACCESS_GET_SUCCESS,
     failedAction: USERS_PERSONAL_ACCESS_GET_FAILURE,
-    url: `/api/users/${userId}/personal_access_tokens`,
+    url: `/api/users/${userId}/personal_access_tokens?per_page=9999`,
```
timogoebel: We should improve this in a later PR.

timogoebel:
@ohadlevy: I added your changes into this commit. Thanks for those. I also continued to work on this.
- The "old" rails code is cleaned up.
- Most of the tab content has moved to react.
- When a token is created, it is added to the store and displayed in the form. The token list is not refreshed from the server. I'm unsure if this is the correct approach.
- The tokens are filtered active/inactive.
- Last used works and is displayed correctly.
- When loading the tokens from the server more than 20 tokens are fetched (API pagination).

Could you please re-review.
If others want to test this: Please go ahead. I have a feeling, we're coming to an end here and this is the time for feedback.

TODO-Items:

- [ ] Check "time ago" is translated correctly. (It does not.)
- [ ] Check "time ago" also works with future tense.
- [ ] Check permissions are working correctly.

app/views/personal_access_tokens/_personal_access_tokens_tab.html.erb
```diff
@@ -0,0 +1,9 @@
+<div class="tab-pane" id="personal_access_tokens">
+  <% #if authorized_for(hash_for_new_user_personal_access_token_path(:user_id => @user)) %>
+  <div id="personal_access_token_create"></div>
+  <%= mount_react_component('PersonalAccessToken', '#personal_access_token_create', {
+    'user_id': @user.id,
+    'csrf-token': form_authenticity_token
```
ohadlevy: imho no need to pass it per component, see https://github.com/theforeman/foreman/pull/4795/files#diff-35521d7d9795ca3a7f62a33bb3ceef6aR30

app/views/personal_access_tokens/_personal_access_tokens_tab.html.erb
```diff
@@ -0,0 +1,9 @@
+<div class="tab-pane" id="personal_access_tokens">
+  <% #if authorized_for(hash_for_new_user_personal_access_token_path(:user_id => @user)) %>
```
ohadlevy, line 2: you can pass it as a prop, something like:
```erb
 canCreate: authorizer.can?(:create_tokens)
```

config/routes.rb
```diff
@@ -281,6 +281,11 @@
         get 'auto_complete_search'
       end
       resources :ssh_keys, :only => [:new, :create, :destroy]
+      resources :personal_access_tokens, :only => [] do
+        member do
+          put :revoke
```
ohadlevy: why keep it and not use the api?

webpack/assets/javascripts/react_app/common/MountingService.js
```diff
@@ -33,14 +35,24 @@ export function mount(component, selector, data) {
     StorageContainer: {
       type: StorageContainer,
       markup: <StorageContainer store={store} data={data} />
+    },
+    PersonalAccessToken: {
+      type: PersonalAccessToken,
+      markup: <PersonalAccessToken data={data} />
```
ohadlevy: nitpick: can use {...data} instead to save the props.data.url and just props.url

webpack/assets/javascripts/react_app/components/common/AlertPanel/Alert.consts.js
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable no-throw-literal */
+const okClass = 'alert alert-success';
```
ohadlevy: imho no need to create a new component, can't we reuse Alert? similar changes to https://github.com/theforeman/foreman/pull/4795/files#diff-00d01709f729237a8f790e772209ea9c ?

webpack/assets/javascripts/react_app/components/common/Icon/Icon.consts.js
```diff
@@ -14,13 +14,13 @@ export default type => {
     case 'info':
       return infoIcon;
     case 'warning':
-    case 'danger':
       return warningIcon;
+    case 'danger':
```
ohadlevy, line 18: since this function returns, does this really works?

webpack/assets/javascripts/react_app/components/users/personalAccessTokens/TokenList.js
```diff
@@ -0,0 +1,42 @@
...
+import PanelTitle from '../../common/Panel/PanelTitle';
+import PanelBody from '../../common/Panel/PanelBody';
+
+export default ({ tokens, title, emptyText, revocable = false }) => (
+  <Panel>
+  <PanelHeading>
+  <PanelTitle text={`${title} (${tokens ? tokens.length : ''})`} />
+  </PanelHeading>
+  <PanelBody>
+  {tokens && tokens.length > 0 &&
```
ohadlevy: why ? i though tokens would always be an array based on its initial state ?

webpack/assets/javascripts/react_app/redux/actions/users/personalAccessTokens/index.js
```diff
@@ -0,0 +1,96 @@
...
+    })
+      .then(checkErrors)
+      .then(response => {
+        return response
+          .json()
+          .then(body => ({
+            type: USERS_PERSONAL_ACCESS_TOKEN_FORM_SUCCESS,
+            payload: { body }
+          }))
+          .then(dispatch);
```
ohadlevy: this dispatch is written wrongly - it should be:
```js
.then(body => 
    dispatch(
      {
         type: USERS_PERSONAL_ACCESS_TOKEN_FORM_SUCCESS,
         payload: { body }
       }
    )
)
```

In general, I've generalized this function for form submtions - have a look at https://github.com/theforeman/foreman/pull/4795/files#diff-35521d7d9795ca3a7f62a33bb3ceef6a

webpack/assets/javascripts/react_app/redux/reducers/users/personalAccessTokens/index.js
```diff
@@ -0,0 +1,51 @@
...
+  USERS_PERSONAL_ACCESS_GET_SUCCESS,
+  USERS_PERSONAL_ACCESS_GET_FAILURE
+
+} from '../../../consts';
+import Immutable from 'seamless-immutable';
+
+const initialState = Immutable({
+  isOpen: false,
+  isSuccessful: false,
+  body: '',
```
ohadlevy: I'm having a hard time with this variable name - can we call it something else? what does body means?

ohadlevy (changes requested):
thanks @timogoebel  - another round of reviews.

most likely we should decide if it makes sense or not to merge #4795 first or not, and to add jest tests

webpack/assets/javascripts/react_app/redux/reducers/users/personalAccessTokens/index.js
```diff
@@ -0,0 +1,51 @@
...
+    case USERS_PERSONAL_ACCESS_TOKEN_FORM_BUTTON: {
+      return state.set('isOpen', false).set('isSuccessful', false).set('body', null);
+    }
+
+    case USERS_PERSONAL_ACCESS_TOKEN_FORM_OPENED: {
+      return state.set('isOpen', true);
+    }
+
+    case USERS_PERSONAL_ACCESS_TOKEN_FORM_SUCCESS: {
+      return state.set('isSuccessful', true).set('body', payload.body).set('tokens', [...state.tokens, payload.body]);
```
ohadlevy: should they be sorted some how (similar to how they were sorted by the server?
you can do something like
```js
set('tokens', [...state.tokens, payload.body].sort((a,b) => a.created_at > b.created_at ));
```

app/assets/javascripts/application.js
timogoebel: I think we won't need this anymore. I'll remove this later on.

timogoebel:
@ohadlevy: Pushed a rebased version. Unfortunately, it does not work yet (formSubmit is not executed).

.eslintrc
```diff
@@ -61,10 +61,11 @@
       "tabWidth": 2,
       "ignoreTrailingComments": true,
       "ignoreUrls": true,
+      "ignoreStrings": true,
```
ohadlevy, line 64: are these changes still required?

app/controllers/personal_access_tokens_controller.rb
```diff
@@ -0,0 +1,17 @@
+class PersonalAccessTokensController < ApplicationController
```
ohadlevy: any reason why not to use the API?

app/models/user.rb
```diff
@@ -393,7 +393,10 @@ def editing_self?(options = {})
       options[:user_id].to_i == self.id ||
     options[:controller].to_s == 'api/v2/personal_access_tokens' &&
       options[:action] =~ /show|destroy|index|create/ &&
-      options[:user_id].to_i == self.id
+      options[:user_id].to_i == self.id ||
+    options[:controller].to_s == 'personal_access_tokens' &&
```
ohadlevy: same question - can't we just use the api directly?

app/views/personal_access_tokens/_personal_access_tokens_tab.html.erb
```diff
@@ -0,0 +1,10 @@
...
+  <div id="personal_access_token_create"></div>
+  <%= mount_react_component('PersonalAccessTokens', '#personal_access_token_create', {
+    user_id: @user.id,
```
ohadlevy: nitpick: prefer js anontation for variables, e.g.
```js
{
  userId: @user,
  url: api_user_personal_access_tokens_path(:user_id => @user),
  canCreate: authorized_for(hash_for_api_user_personal_access_tokens_path(:user_id => @user))
}
```
￼

app/views/users/_form.html.erb
```diff
@@ -79,6 +82,10 @@
       <%= render 'ssh_keys/ssh_keys_tab', :f => f, :user => @user %>
     <% end %>
 
+    <% if @editing_self || authorized_for(:controller => 'personal_access_tokens', :action => :view) %>
+      <%= render 'personal_access_tokens/personal_access_tokens_tab', :f => f, :user => @user %>
```
ohadlevy, line 86: would it make sense to manage the entire tab in react?

webpack/assets/javascripts/react_app/components/common/TimeAgo.js
```diff
@@ -0,0 +1,4 @@
+import React from 'react';
+import Moment from 'react-moment';
+
+export default ({ date }) => date && <Moment fromNow={true} parse="YYYY-MM-DD h:mm:s z" date={date} /> || <span>{__('N/A')}</span>;
```
ohadlevy, line 4: does i18n works as expected? /cc @tstrachota

webpack/assets/javascripts/react_app/components/users/personalAccessTokens/Token.js
```diff
@@ -0,0 +1,33 @@
...
+    </td>
+    <td>
+      <TimeAgo date={last_used_at} />
+    </td>
+    <td>
+  {revocable &&
+      <span className="btn btn-sm btn-default">
+        <a
+          data-confirm="Are you sure?"
+          data-id="aid_personal_access_tokens_1_revoke"
```
ohadlevy: this will use jquery delete dialog ("Are you sure") - we should probably have an alternative option?

webpack/assets/javascripts/react_app/components/users/personalAccessTokens/Token.js
```diff
@@ -0,0 +1,33 @@
...
+          rel="nofollow"
+          data-method="put"
```
ohadlevy: this will also do a full page reload - i think we should convert it to a action instead.

webpack/assets/javascripts/react_app/components/users/personalAccessTokens/TokenList.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import Token from './Token';
+import Panel from '../../common/Panel/Panel';
+import PanelHeading from '../../common/Panel/PanelHeading';
+import PanelTitle from '../../common/Panel/PanelTitle';
+import PanelBody from '../../common/Panel/PanelBody';
+
+export default ({ tokens, title, emptyText, revocable = false }) => (
+  <Panel>
+  <PanelHeading>
```
ohadlevy: indentation is off

webpack/assets/javascripts/react_app/components/users/personalAccessTokens/index.js
```diff
@@ -0,0 +1,96 @@
...
+
+    let headerActions;
+
+    if (isSuccessful) {
+      headerActions = (
+        <Alert type="success" onClose={hideForm} title={__('Your New Personal Access Token')}>
+          <code style={{ fontSize: '120%' }}>{body.token_value}</code>
+          &nbsp;
+          <ClipboardButton
+            data-clipboard-text={body.token_value}
```
ohadlevy: does this uses data attributes? (is it jquery based?)

webpack/assets/javascripts/react_app/components/users/personalAccessTokens/index.js
```diff
@@ -0,0 +1,96 @@
...
+            component="a"
+            title={__('Copy to clipboard!')}
```
ohadlevy: imho the ! is not required.

webpack/assets/javascripts/react_app/components/users/personalAccessTokens/index.js
```diff
@@ -0,0 +1,96 @@
...
+          >
+            <i className="fa fa-clipboard" aria-hidden="true" />
```
ohadlevy: please use Icon component?

webpack/assets/javascripts/react_app/components/users/personalAccessTokens/index.js
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    return (
+      <div className="row">
+        <div className="col-md-12">
+         <p>{__('Personal Access Tokens allow you to authenticate API requests without using your password.')}</p>
+         {headerActions}
+         <TokenList tokens={tokens ? tokens.filter(token => token['active?']) : undefined}
+           title={__('Active Personal Access Tokens')}
+           emptyText={__('This user does not have any active Personal Access Tokens.')}
```
ohadlevy: should we not use the PF empty state if no tokens are defined?

webpack/assets/javascripts/react_app/components/users/personalAccessTokens/tokenForm/index.js
```diff
@@ -0,0 +1,63 @@
...
+import Form from '../../../common/forms/Form';
+import TextField from '../../../common/forms/TextField';
+import { required, length, date } from 'redux-form-validators';
+import * as FormActions from '../../../../redux/actions/common/forms';
+
+const validations = {
+  name: [required(), length({ max: 254 })],
+  // TODO: rename this var?
+  // eslint-disable-next-line camelcase
+  expires_at: [date({ unless: value => { return value.expires_at === undefined; }, format: 'yyyy-mm-dd', '>': 'today' })]
```
ohadlevy: does the format depends on locale?

webpack/assets/javascripts/react_app/components/users/personalAccessTokens/tokenForm/index.js
```diff
@@ -0,0 +1,63 @@
...
+};
+
+// Reusable with any other form
```
ohadlevy: probably not needed anymore?

app/controllers/personal_access_tokens_controller.rb
timogoebel: No, this should be done in react as well.

app/views/users/_form.html.erb
timogoebel: I don't see, why we should. What could we gain from this?

webpack/assets/javascripts/react_app/components/common/TimeAgo.js
timogoebel: It does not. Moment needs special treatment.

.eslintrc
timogoebel: I think this change makes a lot of sense.

timogoebel:
@ohadlevy: Fixes some bugs and moved the revoke token feature entirely to react.

.eslintrc
ohadlevy: @danseethaler FYI - he was re-reviewing the linting rules

app/views/users/_form.html.erb
ohadlevy: highlight errors?

webpack/assets/javascripts/react_app/components/common/TimeAgo.js
ohadlevy: /cc @tstrachota as he is working on date time display in react too.

.eslintrc
danseethaler: #4968 Removes the `max-len` rule in favor of the `eslint-plugin-prettier` which will enforce the max-length when it makes sense.

mmoll:
@timogoebel What's the status here? :)

timogoebel:
@mmoll: Nobody reviewed this. I guess we can close the PR. If I have a lot of time, I'll pick this up again and do it with graphql.

PR #4598, opened 2017-06-16
Fixes #20034 - add Organization admin role

ares:


ares:
@dcaplan this might be interesting for you. The name for org admin role is "Organization admin", is that acceptable? The role does not have any description yet, I plan to address that through separate PR, it needs more changes.

app/registries/foreman/plugin/rbac_support.rb
```diff
@@ -34,7 +37,7 @@ def add_permissions_to_default_roles(roles)
 
       def check_role_name_before_extending(role_name)
         msg = "Invalid role name, only 'Viewer' or 'Manager' is allowed to be extended from a plugin"
-        raise Foreman::Exception.new msg if role_name != "Viewer" && role_name != "Manager"
+        raise Foreman::Exception.new msg if role_name != "Viewer" && role_name != "Manager" && role_name != "Organization admin"
```
dLobatog: Looks like it's time to add a constant array `SHIPPED_ROLES` and check whether `role_name` is in there

app/registries/foreman/plugin/rbac_support.rb
```diff
@@ -4,8 +4,10 @@ class RbacSupport
       def add_all_permissions_to_default_roles(all_permissions)
         return if Foreman.in_rake?('db:migrate')
         view_permissions = all_permissions.where("name LIKE :name", :name => "view_%")
+        org_admin_permissions = all_permissions.where("resource_type <> :resource_type", :resource_type => "Organization")
```
dLobatog: Nitpick, prefer to use ActiveRecord helpers, they prevent bugs and allow the query to be optimized `where.not(:resource_type => 'Organization')` -

dLobatog (changes requested):
Thanks, just left a couple of comments @ares - 

I see in the BZ they're asking for a bit more, like having org admins automatically created when the organization is created, like in http://people.redhat.com/~mmccune/create_org_admins.rake 

Is that coming on another PR or can it go in this one? 

Lastly, if you want this on 1.15.2 please add a couple of tests demonstrating the behavior 😄 

All over 👍 - will give it maybe a day after :it's fully ready to merge to allow @dcaplan to voice any concerns

app/registries/foreman/plugin/rbac_support.rb
ares: I agree, but it turns out that this would ignore permissions that has NULL as resource_type (generic permissions), so I needed to add "OR" to the condition, AR in Rails 4 don't have DSL for it and I'd like to avoid using AREL directly.

app/registries/foreman/plugin/rbac_support.rb
ares: I kept the constant commit separate, if we'll cherry-pick this into stable branch we might want to keep it minimal. I can squash if you're OK with taking both to 1.15.2.

ares:
Thanks @dLobatog, I extended tests to cover the functionality, I didn't add explicit tests for internal organization permission, since permission - role associations are hardcoded in fixtures for tests so it would be only testing that I setup fixtures correctly. Please take another look.

I'm not entirely sure which BZ you meant, the redmine issue is not linked to any. I'd like to keep autocreation of role on organization creation to separate PR. There's more to it, e.g. these clones should be locked and auto extended with new permissions as well. I also plan to improve role description and display this in role list but again, as a separate PR.

This is a base step which is required for users who want to delegate organization management though. Users can start cloning this role while we improve the workflow in upcoming versions.

dLobatog (approved):

dLobatog (approved):

dLobatog:
Thanks @ares , merged for 1.15.2

dscaplan:
@ares I ok with Organization Admin as long as it does not get truncated in the various tables where it is managed

PR #4599, opened 2017-06-16
Fixes #20035 - Adds API link to About page

ares:


app/views/about/index.html.erb
```diff
@@ -124,7 +124,8 @@
       <h4><%= _("Support") %></h4>
       <h6><%= _("Documentation") %></h6>
       <ul>
-        <li><%= link_to _('Manual'),documentation_url, :rel => "external" %></li>
+        <li><%= link_to _('API'), apipie_apipie_path %></li>
```
ohadlevy: nitpick: should this be API docs or something more clear?

ares: Yeah... I was not sure about the label, "API documentation" seems like too much for just a list of endpoints. How about "Available API calls" or "API provided by the server" or simply "Server API"?

ohadlevy: I would add it maybe via an hover or an icon? something like
```Automate Foreman via a simple and powerful API```

dLobatog (changes requested):
Setting to WoC as per @ohadlevy 's comment. @ares  Personally I think "API documentation" or "API resources"  is accurate to describe what Apipie provides.

ares:
I liked "API Resources" the most, I also added the hover message as @ohadlevy suggested

dLobatog:
Failure http://ci.theforeman.org/job/test_develop_pr_core/13327/database=postgresql,ruby=2.4,slave=fast/testReport/junit/(root)/ComputeResourceJSIntegrationTest/test_0003_compute_resource_password_isn_t_deleted_when_testing_connection/ should be fixed by https://github.com/theforeman/foreman/pull/4604

dLobatog (approved):

dLobatog:
Thanks @ares !

PR #4603, opened 2017-06-19
fixes #19912 - added Style/WordArray cop

ohadlevy:


tbrisker:
I think for this rule we should enforce the bracket format rather then the percent format, more readable imho. Any other opinions @theforeman/core ? feel free to give thumbs up/down.

timogoebel:
+1 for bracket format

dLobatog (commented):
-1 to the bracket format - I would prefer to stick to https://github.com/bbatsov/ruby-style-guide as much as possible. It's the closest thing to PEP8 we have in Ruby and IMO it's easier to read when all projects adhere to that standard (which is what rubocop is meant to do)

ares:
I vote for keeping `%w()`, `%w[]` and `[]` allowed. I don't think any of these would be significantly better, I'd keep it disabled.

iNecas:
If the purpose of Rubocop is to keep things consistent, that we probably want to have this rule. I would vote for bracket format: the reason would be to be open to the sys-admins community, that not necessary need to know Ruby that well, while know other languages, and `%w[]` thing is pretty specific to Ruby. You can take my vote for it. I will not spend any more time discussing it :) Thanks for letting us to have voice in the decision.

tbrisker:
Since no agreement was reached, it seems that we should allow all styles. Please move the cop to `.rubocop.yml` and rollback any fixes, or close this PR.

PR #4604, opened 2017-06-19
Fixes #20048 - Wait before checking CR password

dLobatog:
The test is trying to check the value for the password field.
However it's failing to get it, possibly because jQuery is still
setting the password.

I see find_field is supposed to wait for Capybara.default_max_wait_time, but there is no apparent reason why the password would not show up. Is there a way to check in Jenkins, saving Rails logs somewhere..?


dLobatog:
I think the problem appears to be that "find('xxxxxx').value" is not correct - `.value` might not [wait until](https://www.varvet.com/blog/why-wait_until-was-removed-from-capybara/) `find` returns - let's use Capybara helpers for that.

tbrisker:
looks like test are failing.

PR #4605, opened 2017-06-19
Fixes #20049 - Retry various integration tests before failing

dLobatog:
Thanks to #15894 we can retry some tests. Some of the failures I'm
seeing I'm having a hard time to reproduce, like
"Capybara::Poltergeist::StatusFailError: Request to
'http://127.0.0.1:41550/compute_profiles/980190962-1-Small' failed to
reach server, check DNS and/or server status - Timed out with no open
resource requests...". I think it's not a bad idea to let these test run
3 times and only count them as failed if they fail 3 times.

### DIFF
--- a/test/integration/about_test.rb
+++ b/test/integration/about_test.rb
@@ -1,6 +1,10 @@
 require 'integration_test_helper'
 
 class AboutIntegrationTest < IntegrationTestWithJavascript
+  # intermittent failures:
+  #   AboutIntegrationTest.test_0002_about page proxies should have version
+  extend Minitest::OptionalRetry
+
   setup do
     ComputeResource.any_instance.expects(:ping).at_least_once.returns([])
     proxy_status = mock('ProxyStatus::Version')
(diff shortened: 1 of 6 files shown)


dLobatog:
Adding http://ci.theforeman.org/job/test_develop_pr_core/13337/database=postgresql,ruby=2.2,slave=fast/testReport/junit/(root)/PuppetclassIntegrationTest/test_0001_edit_page/ which I just saw on https://github.com/theforeman/foreman/pull/4604

dLobatog:
[test upgrade]

>>  hudson.remoting.RequestAbortedException: java.io.IOException: Unexpected termination of the channel

ShimShtein:
:+1: Works locally, one question though: is it worth adding retry policy to all JS tests (`IntegrationTestWithJavascript`)?

dLobatog:
@ShimShtein In principle.. yes, but I would prefer to wait and see the errors, in case we can notice some pattern.

ShimShtein:
ACK then. I have an item on my TODO list to see what can be done to reduce the amount of random test failures, but I am not sure when I'll get to this.

tbrisker:
Thanks @dLobatog !

PR #4606, opened 2017-06-19
adds code of conduct.

ohadlevy:


ohadlevy:
open this up for discussion - /cc @GregSutcliffe

mmoll:
I'm against CoCs and especially the Contributor Convenant one. 👎 

As a very quick readup I pulled http://paul-m-jones.com/archives/6214 from my bookmarks as it gets the point and has some more links.

timogoebel:
I don't think, we need this.
It should be self-evident to be nice and respectful. Both to yourself and others. I try to be nice - even with or without a code of conduct. It should be obvious that one can have a serious discussion about some technical issue and share 🍻 afterwards. Having a code of conduct probably won't hurt, but imho it's pointless.

ares:
What is the motivation for this? IIRC we discussed this few years ago and ended up with - if we need CoC, it should be "don't be stupid".

ohadlevy:
thanks - based on this feedback I retract this PR, the hope was to show that we are open / transparent / positive community etc for all.

dLobatog:
Please notice we have boundaries for how people should treat each other in the Foreman community. https://www.theforeman.org/support.html#CommunityGuidelines 

I don't think we've ever had to action anything other than (very sparingly) ask someone on IRC to be more civil. As this is an open source community, I do my best to ensure the discussion here, in Redmine, or in public spaces, is strictly restricted to the code and the project, without any regards to any of the attributes mentioned in this CoC or anything in general. From what I see, other maintainers also keep it as much code-focused as possible and welcoming to newcomers.

_I sincerely think the following paragraph applies to no one and I hope not to be proven otherwise_

Indeed: if anyone does not feel welcome in the Foreman community, it's probably not one of the committers. If that is the case, you can feel free to contact us - privately or publicly - on the channels described in https://www.theforeman.org/support.html#CommunityGuidelines . We will be happy to hear you and discuss any actions that would improve your situation in the Foreman community.

lzap:
I never read these documents anyway :-)

Edit: With elevated focus on CoC these days (fall 2018) due to discussion on LKML, I believe need to clarify my comment above. What I wanted to say is simply a presence of a document in a git repository means nothing. It's all about us, our community and our behavior and I strongly believe Foreman community is one of the best. We already have a fraction of what can be called CoC in the Support section of our site, that's probably the best place for further expansion.

In the FOSS world it is common to download a tarball or checkout a git repo but never read README, LICENSE or other files. Instead, you go to project site and study documentation and other info there. These files are often distributed with the software and getting installed, people rarely read them. Ability to send a link to a document instead writing arbitrary text is nice in general, but we had just few such an issues for the whole history of the project.

With all that said, I am neutral on this change. I don't mind having *similar* document directly in the git repo, but I think we are lucky that we don't need those for now. I do have some issues with author(s), story and background of this particular document questioning meritocracy to its limits, I'd strongly prefer writing our own one as we go meaning if we have an issue, let's add new rule.

PR #4607, opened 2017-06-21
Fixes #20066 - added safe renderer to global params

ShimShtein:


ohadlevy (changes requested):
Please add a test.. thanks

ShimShtein:
@ohadlevy Done! Tests added.

ohadlevy:
[test upgrade]

test/models/host_test.rb
```diff
@@ -3577,6 +3577,22 @@ def to_managed!
...
+    test 'shows global params' do
+      host = FactoryGirl.build(:host)
+      parameter = FactoryGirl.create(:common_parameter, :name => 'test_param1', :value => 'test_value1')
+
+      assert_equal 'test_value1', host.host_params['test_param1']
+    end
+
+    test 'renders global params if template specified' do
+      host = FactoryGirl.build(:host)
+      parameter = FactoryGirl.create(:common_parameter, :name => 'test_param1', :value => '<%= "test_value1" %>')
```
ohadlevy: nitpick: not that matters much, but i think it would be more interesting to use some more computed value.

ShimShtein:
Now the test uses `@host` local variable for its value.

ohadlevy (approved):
ack pending tests.

dLobatog (commented):
[test upgrade]

dLobatog:
[test upgrade]

ShimShtein:
Upgrade tests are complaining about not finding `audited-activerecord` gem. Seems unrelated to this PR.

app/models/concerns/host_params.rb
```diff
@@ -20,13 +20,15 @@ def clear_host_parameters_cache!
     def host_inherited_params(include_source = false)
       hp = {}
       params = host_inherited_params_objects
+      renderer = SafeRender.new(:variables => { :host => self })
```
ohadlevy, line 23: @ShimShtein for future reference, I think it would make sense to store on params (global, class etc) if the value has <% in it or not, this can probably save a few ms instead of calling the SafeRender by default.

ohadlevy:
test failure not related, thanks @ShimShtein

PR #4609, opened 2017-06-22
Fixes #20072 - Allow thin host list via API

tbrisker:


app/controllers/api/v2/hosts_controller.rb
```diff
@@ -37,6 +37,12 @@ class HostsController < V2::BaseController
       def index
         @hosts = action_scope_for(:index, resource_scope_for_index)
 
+        if params[:thin]
+          @hosts = @hosts.reorder(:name).distinct.pluck(:id, :name)
```
tbrisker, line 42: we have to reorder on a field we have, otherwise if the user passed an order param on something that isn't in the query it will fail.

ohadlevy: should we add some more basic info, such as ip address ?

tbrisker: The original request was for just id and name

beav:
WOW this is fast!!

```
2017-06-22 19:26:28 [app] [I] Started GET "/api/v2/hosts?thin=true&per_page=50000" for ::1 at 2017-06-22 19:26:28 -0400
2017-06-22 19:26:28 [app] [I] Processing by Api::V2::HostsController#index as JSON
2017-06-22 19:26:28 [app] [I]   Parameters: {"thin"=>"true", "per_page"=>"50000", "apiv"=>"v2"}
2017-06-22 19:26:28 [app] [I] Authorized user admin(Admin User)
2017-06-22 19:26:28 [app] [I]   Rendered api/v2/hosts/thin.json.rabl within api/v2/layouts/index_layout (375.5ms)
2017-06-22 19:26:28 [app] [I] Completed 200 OK in 882ms (Views: 377.5ms | ActiveRecord: 352.6ms)
```

dLobatog (approved):
👍 , thanks! Let's improve only if someone requests it, as I think the number of possibilities would be endless. Time for API v3 in GraphQL so that people can request whatever they need? 😄

PR #4610, opened 2017-06-22
 Fixes #20083 - Search by environment for smart class parameter

orrabin:
Also moves seach by override to smart class parameter and renames search
by key in smart variable

### DIFF
--- a/app/models/lookup_keys/puppetclass_lookup_key.rb
+++ b/app/models/lookup_keys/puppetclass_lookup_key.rb
@@ -9,6 +9,7 @@ class PuppetclassLookupKey < LookupKey
   after_validation :reset_override_params, :if => ->(key) { key.override_changed? && !key.override? }
 
   scoped_search :relation => :param_classes, :on => :name, :rename => :puppetclass, :aliases => [:puppetclass_name], :complete_value => true
+  scoped_search :relation => :environments, :on => :name, :rename => :environment, :complete_value => true, :only_explicit => true
 
   scope :smart_class_parameters_for_class, lambda { |puppetclass_ids, environment_id|
                                              joins(:environment_classes).where(:environment_classes => {:puppetclass_id => puppetclass_ids, :environment_id => environment_id})


config/initializers/00_rest_client.rb
```diff
@@ -0,0 +1,7 @@
+require 'rest_client'
```
ohadlevy: left over? or maybe this should be redirected to its own logger (i would say thats a feature by it self).

ponse)
```diff
@@ -0,0 +1,41 @@
+=> [[32m#<OVIRT::StorageDomain:0x00000004d79200[0m
```
ohadlevy: not related, please delete.

iNecas:
Seems like the tests don't like the change for some reason…

orrabin:
@iNecas I was trying to refactor to get the search of override out of `variable lookup key` but there is only one united `lookup key` in api v1 so it broke.
I changed this to only add environment.

app/models/lookup_keys/puppetclass_lookup_key.rb
```diff
@@ -9,6 +9,7 @@ class PuppetclassLookupKey < LookupKey
   after_validation :reset_override_params, :if => ->(key) { key.override_changed? && !key.override? }
 
   scoped_search :relation => :param_classes, :on => :name, :rename => :puppetclass, :aliases => [:puppetclass_name], :complete_value => true
+  scoped_search :relation => :environments, :on => :name, :rename => :environment, :complete_value => true
```
tbrisker: Should this (and possibly the previous one as well) be set to `:only_explicit`? I'm not sure if using free text to search for class or environment is common, and in a large setup this may have a possible performance effect.

iNecas: :+1: on keeping the `:environments` `:only_explicit`. For the `param_classes`, it would mean a change in a behaviour, that is not related directly to the attached issue. I would probably keep it as separate issue, ideally opened by some user that would hit it.

orrabin:
@tbrisker I added `only_explicit`

tbrisker:
Thanks @orrabin !

PR #4613, opened 2017-06-22
Fixes #20087 - Pin rails-observers on Ruby < 2.2.2

dLobatog:
Notice this PR is against 1.11-stable. test_update currently fails with the following message,  this PR fixes it by pinning the gem.

```
Gem::InstallError: rails-observers requires Ruby version >= 2.2.2.
An error occurred while installing rails-observers (0.1.4), and Bundler cannot
continue.
Make sure that `gem install rails-observers -v '0.1.4'` succeeds before
bundling.

In Gemfile:
  audited-activerecord was resolved to 4.2.2, which depends on
    audited was resolved to 4.2.2, which depends on
      rails-observers
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -28,7 +28,7 @@ gem 'net-scp'
 gem 'net-ssh'
 gem 'net-ldap', '>= 0.8.0'
 gem 'activerecord-session_store', '>= 0.1.1', '< 2'
-gem 'rails-observers', '~> 0.1'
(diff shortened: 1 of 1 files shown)


tbrisker:
Ack pending passing upgrade test.

tbrisker (approved):

tbrisker:
Tested locally, works with ruby 2.0. merging.

tbrisker:
Thanks @dLobatog !

PR #4616, opened 2017-06-23
Fixes #12216 - support http proxies for compute resources

jlsherrill:


jlsherrill:
This is based on this RFC: https://github.com/theforeman/rfcs/pull/18

I decided against making it a plugin and just work on integrating it with the compute resources.  foreman-docker and katello both would likely integrate with this.

This is only integrating with EC2 at the moment, but I think other compute resources should be fairly easy (and am happy to do the work here).  I may not have the ability to test some of those, so I may need some help there.  I wanted to open for initial thoughts prior to implementing it for the rest.

app/views/http_proxies/_form.html.erb
```diff
@@ -0,0 +1,29 @@
...
+    <% if show_organization_tab? %>
+      <li><a href="#organizations" data-toggle="tab"><%= _("Organizations") %></a></li>
+    <% end %>
+  </ul>
+
+  <div class="tab-content">
+
+    <div class="tab-pane active" id="primary">
+      <%= text_f   f, :name %>
+      <%= text_f   f, :url, :help_inline => _("Url of the proxy including schema (https://proxy.example.com)") %>
```
ehelms: Nit: URL should be capitalized here

app/views/api/v2/http_proxies/base.json.rabl
```diff
@@ -0,0 +1,3 @@
+object @http_proxy
+
+attributes :id, :name, :url, :port, :username, :password
```
ehelms: Is it common practice to return a password here in a base template?

app/models/http_proxy.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  encrypts :password
+  friendly_id :name
+
+  has_many :compute_resources
+
+  validates :url, :presence => true
+  validates :port, :presence => true
+
+  scoped_search :on => :name
```
ehelms: I could see being allowed to search on URL as well in the case where you remember the proxy domain name but forgot what you named the proxy.

app/views/compute_resources/form/_ec2.html.erb
```diff
@@ -1,3 +1,8 @@
+<%= select_f(f, :http_proxy_id, HttpProxy.all, :id, :name,
```
timogoebel: Nitpick: Maybe move this to a helper to allow for easier inclusions by other compute resources?

app/views/api/v2/http_proxies/base.json.rabl
jlsherrill: good catch, its encrypted in production so it doesn't even make sense to

app/models/http_proxy.rb
jlsherrill: added

mmoll:
[test katello] - abusing this PR to test something general...

jlsherrill:
[test katello]

jlsherrill:
@theforeman/core Any thoughts/ideas here?  The ec2 part of this is ready to go.  

The current codeclimate error seems silly?  i'm not sure if that actually matters

ares:
@jlsherrill feel free to ignore the codeclimate, it was added just as a help for reviewer but it's not mandatory to have this one green, I hope we'll get to review soon

lzap:
Nice, a test button would be great because if the destination port is out of SELinux allowed list of ports than it's denied connection. @ehelms or @ares are you doing review?

jlsherrill:
@lzap I thought about adding a test button, however the user would need to provide some URL to try to connect to (as proxy configurations and firewalls may be restrictive).  This seemed kind of awkward but I'm okay implementing it if that is desired.

lzap:
Hmmm making HTTP GET HEAD / would not do the work? I mean, even 404/500 is a success, DNS or connection error is a failure. But that's nitpick, I just find this useful in the SELinux context. I see users hitting the wall hard, my imagination :-)

bastilian:
@jlsherrill in this case the best to test might be to simply check wether or not the EC2 API is reachable, with a HEAD to amazonaws.com maybe or use an already existing EC2 compute resource to test, but that might be trickier.

app/views/http_proxies/_form.html.erb
```diff
@@ -0,0 +1,29 @@
...
+      <li><a href="#organizations" data-toggle="tab"><%= _("Organizations") %></a></li>
+    <% end %>
+  </ul>
+
+  <div class="tab-content">
+
+    <div class="tab-pane active" id="primary">
+      <%= text_f   f, :name %>
+      <%= text_f   f, :url, :help_inline => _("URL of the proxy including schema (https://proxy.example.com)") %>
+      <%= number_f   f, :port, :help_inline => _("Port to connect to the proxy with.") %>
```
timogoebel: Why is the port not part of the URL?
I think it would be easier to specify a proxy like this: `http://localhost:3128/`

app/models/http_proxy.rb
```diff
@@ -0,0 +1,26 @@
...
+  include Encryptable
+
+  extend FriendlyId
+
+  encrypts :password
+  friendly_id :name
+
+  has_many :compute_resources
+
+  validates :url, :presence => true
```
timogoebel, line 13: This should also validate, that a user entered a correct url with schema.
`http://example.com` should be allowed, `börks` should be invalid.

app/models/compute_resource.rb
```diff
@@ -123,6 +124,10 @@ def to_label
     "#{name} (#{provider_friendly_name})"
   end
 
+  def connection_options
+    http_proxy ? {:proxy => http_proxy.full_url} : {}
```
timogoebel, line 127: If no proxy is set, this falls back to the proxy defined via environment variables. If you don't set a proxy via the UI, Foreman still uses a proxy. I find this confusing. What do you think to explicitly set this to use no proxy if it is not explicitly set?

timogoebel:
I'm testing this right now. Sharing some screenshots in case somebody is interested in how this looks.
 
image

image

image

timogoebel (changes requested):
Thanks. Generally, this looks good and works well. I left some comments inline.

I think, the test connection button needs some more work.
If I configure an invalid proxy, the test connection controller action throws a `Excon::Error::ProxyParse (Proxy is invalid)` exception. This is only logged and the user only sees a 500 Internal Server Error in the UI. This isn't user friendly.

timogoebel:
I also found another bug while testing this: https://github.com/theforeman/foreman/pull/4683

app/views/http_proxies/_form.html.erb
ares: also seems like there are extra whitespaces

app/models/compute_resource.rb
jlsherrill: By doing this, we would no longer be supporting the environment variables, correct?   I would be supportive of that!

jlsherrill: hrmm, i guess if we do this, we'd need to make sure this is implemented for all compute resources?

jlsherrill:
Updated to fix the compute resource test connection as well as add  test connection button to the http proxies create and edit:

image

Also: 
* added url validation
* fixed compute resource test connection issue @timogoebel found

app/assets/javascripts/http_proxy.js
```diff
@@ -0,0 +1,21 @@
+function testConnection(item, url) {
```
timogoebel: I'm afraid, I have to ask you to move this to webpack.

webpack/assets/javascripts/react_app/redux/actions/notifications/index.js
```diff
@@ -11,7 +11,7 @@ import {
 } from '../../../common/sessionStorage';
 import API from '../../../API';
 import { isNil } from 'lodash';
-const defaultNotificationsPollingInterval = 10000;
+const defaultNotificationsPollingInterval = 100000000;
```
timogoebel: Although I feel the need for this change: This is not the place. Please remove this.

test/unit/http_proxy_test.rb
```diff
@@ -0,0 +1,27 @@
...
+class HttpProxiesTest < ActiveSupport::TestCase
+  let(:http_proxy) {FactoryGirl.create(:http_proxy)}
+
+  test 'create' do
+    proxy = HttpProxy.new(:name => 'foobar', :url => "http://someurl:5000")
+
+    assert proxy.save
+  end
+
+  test 'create requires url' do
```
timogoebel: This can be replaced by shoulda-matchers to reduce the boilerplate.
```ruby
class HttpProxiesTest < ActiveSupport::TestCase
  should validate_presence_of(:url)
end
```

test/unit/http_proxy_test.rb
```diff
@@ -0,0 +1,27 @@
...
+    proxy = HttpProxy.new(:name => 'foobar', :username => "myuser")
+
+    refute proxy.save
+  end
+
+  test 'invalid url fails' do
```
timogoebel:
```ruby
class HttpProxiesTest < ActiveSupport::TestCase
  should_not allow_value('börks').for(:url)
end
```

app/models/http_proxy.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    RestClient::Request.execute(method: :head, url: url, proxy: full_url)
+  rescue Excon::Error::Socket => e
+    e.message
+  end
+
+  def valid_url
+      URI(url) if url
+  rescue URI::InvalidURIError => e
+    errors.add(:url, e.message)
```
timogoebel: Does this translate correctly? Maybe a regex or the url schema validator are a better choice.

test/controllers/http_proxies_controller_test.rb
```diff
@@ -0,0 +1,46 @@
+require 'test_helper'
+
+class HttpProxiesControllerTest < ActionController::TestCase
+  let(:model) { FactoryGirl.create(:http_proxy) }
+
+  def test_index
```
timogoebel: You could use https://github.com/theforeman/foreman/blob/develop/test/controllers/shared/basic_rest_response_test.rb and replace some of the tests like at https://github.com/theforeman/foreman/blob/develop/test/controllers/architectures_controller_test.rb#L8

timogoebel (changes requested):
@jlsherrill : Thanks for the good work. This looks really awesome. I just have some more suggestions.

test/controllers/http_proxies_controller_test.rb
```diff
@@ -0,0 +1,42 @@
...
+  def test_show
+    get :edit, { :id => @model.id }, set_session_user
+
+    assert_response :success
+  end
+
+  def test_destroy
+    delete :destroy, { :id => @model.id }, set_session_user
+
+    assert_response 302
```
timogoebel: This should not use the number but a symbol, i.e. `:found`. Same below.

jlsherrill:
@timogoebel finally got some time to update this!

jlsherrill:
Should be able to be a bit more responsive from now on

app/models/compute_resource.rb
timogoebel: Just to complete this discussion: I think, we should do this in a separate PR if we want to do this.

webpack/assets/javascripts/foreman_http_proxies.js
```diff
@@ -0,0 +1,26 @@
+/* global tfm*/
+/* eslint no-undef: "error"*/
```
timogoebel: import $ from 'jquery';
import { notify } from './foreman_toast_notifications';

webpack/assets/javascripts/foreman_http_proxies.js
```diff
@@ -0,0 +1,21 @@
+export function testConnection(item, url) {
+  $('#test_connection_indicator').show();
+  $(item).addClass("disabled");
+  var data = $("form").serialize();
+  $.ajax({
+    url: url,
+    type: 'put',
+    data: data,
+    success: function (result, textstatus, xhr) {
+      tfm.toastNotifications.notify({message: result.message, type: 'success'});
```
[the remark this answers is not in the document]
timogoebel: With the imports above you should be able to just write:

```js
notify({message: result.message, type: 'success'})
```

timogoebel (commented):
@jlsherrill: Thanks. Looks very good so far. Some more comments inline.

webpack/assets/javascripts/foreman_http_proxies.js
jlsherrill: added!  thanks for the tips. have not used webpack before.

timogoebel: I think it's safe now to remove these comments.

timogoebel (approved):

timogoebel:
@jlsherrill: Thanks. This is excellent work. I tested it and could not find any issues. Well done.

PR #4618, opened 2017-06-23
Fixes #19383 - document enabled host attribute

ares:
For some reason, my hammer does not display this even though I see it in /apidoc and at public/apidoc/v2.json and I'm running `hammer -r host create --help`. Anyway I believe it needs documentation like this. The string is copied from UI but if $reviewer thinks it's worth of rephrasing, I'm happy to change it both for API and UI.


dLobatog (approved):

dLobatog:
Thanks @ares

ares:
Thanks, I've set the release to 1.16 in redmine.

PR #4619, opened 2017-06-23
Fixes #20097 - Allow plugins JS to be added to webpack

dLobatog:
At the moment, the only entrypoint for webpack seems to be /webpack.
webpack.config.js should also support code coming from plugins.
Using bundler, we can fetch this directory trivially - a dumb heuristic
could be to add as an entry point the '/webpack' directory of all gem
dependencies that look like "foreman*".

### DIFF
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -11,11 +11,13 @@ var devServerPort = 3808;
 
 // set TARGETNODE_ENV=production on the environment to add asset fingerprints
 var production = process.env.RAILS_ENV === 'production' || process.env.NODE_ENV === 'production';
+const execSync = require('child_process').execSync;
 
 var config = {
   entry: {
     // Sources are expected to live in $app_root/webpack
     'bundle': './webpack/assets/javascripts/bundle.js'
+    'plugin': execSync('./script/plugin_webpack_directories.rb').toString().split('\n');
   },
 
   output: {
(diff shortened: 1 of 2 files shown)


config/webpack.config.js
```diff
@@ -11,11 +11,13 @@ var devServerPort = 3808;
 
 // set TARGETNODE_ENV=production on the environment to add asset fingerprints
 var production = process.env.RAILS_ENV === 'production' || process.env.NODE_ENV === 'production';
+const execSync = require('child_process').execSync;
 
 var config = {
   entry: {
     // Sources are expected to live in $app_root/webpack
     'bundle': './webpack/assets/javascripts/bundle.js'
+    'plugin': execSync('./script/plugin_webpack_directories.rb').toString().split('\n');
```
ohadlevy, line 20: no need for ';' you are inside  an object.

boaz0 (commented):
I am not sure it's in this scope but what about changing `var` to `const` in the right places.
For example, changing `var production` to `const production`. We aren't going to reassign it anymore, are we?

All in all, LGTM

ohadlevy:
@ripcurld0 this is the webpack config file which runs prior to babel, afair it depends on the node version used to support this, I'm uncertain from which version of node es6 syntax is supported.

boaz0:
@ohadlevy I see.

tbrisker:
Closing this in favor of #4625 . Thanks @dLobatog !

PR #4620, opened 2017-06-25
Fixes #19844 - fix Security/Eval cop

tbrisker:


ohadlevy:
:+1: pending fixes to hound comments and tests :)

app/models/operatingsystems/windows.rb
```diff
@@ -14,7 +14,7 @@ def pxe_prefix(arch)
   end
 
   def bootfile arch, type
-    pxe_prefix(arch) + eval("#{self.family}::PXEFILES[:#{type}]")
+    pxe_prefix(arch) + PXEFILES[type.to_sym]
```
dLobatog, line 17: Isn't this missing `self.family.constantize::` before PXEFILES?

app/models/operatingsystems/xenserver.rb
```diff
@@ -27,13 +27,13 @@ def display_family
   end
 
   def bootfile(arch, type)
-    pxe_prefix(arch) + "-" + eval("#{self.family}::PXEFILES[:#{type}]").split("/")[-1]
+    pxe_prefix(arch) + "-" + PXEFILES[type.to_sym].split("/")[-1]
```
dLobatog, line 30: `self.family.constantize::` ditto

app/models/operatingsystems/xenserver.rb
```diff
@@ -27,13 +27,13 @@ def display_family
...
   end
 
   def boot_files_uri(medium, architecture, host = nil)
     raise ::Foreman::Exception.new(N_("Invalid medium for %s"), self) unless media.include?(medium)
     raise ::Foreman::Exception.new(N_("Invalid architecture for %s"), self) unless architectures.include?(architecture)
-    eval("#{self.family}::PXEFILES").values.push(MBOOT).collect do |img|
+    PXEFILES.values.push(MBOOT).map do |img|
```
dLobatog, line 36: `self.family.constantize::` ditto, Xenserver has a family according to operatingsystem.rb

dLobatog (changes requested):
@tbrisker It looks like the family prefixes to PXEFILES are missing on a couple of files, could you check that's correct? Thanks!

app/models/operatingsystems/windows.rb
tbrisker: No, because in this case the family is always `Windows` - Operatingsystem.family just reads the type attribute.

tbrisker:
@dLobatog the `self.family` part was just copy paste from the general operating system code AFACIT. for windows/xenserver the `family` (which is basically an alias for `type`) would be `Windows` or `Xenserver` accordingly and the relevant constant would be just `PXEFILES` without having to fully qualify it.

dLobatog (approved):

dLobatog:
Thanks @tbrisker !

ares:
Thanks for merging, please don't forget to set the release in redmine issue after merging.

PR #4622, opened 2017-06-26
Fixes #20104 - fix_db_cache needs to run as admin

dLobatog:
The rake task 'fix_db_cache' triggers CacheManager, and CacheManager
tries to find roles, user groups, etc.. without any permissions. This
will cause it to fail with an error similar to
https://gist.github.com/52da11cb368ec530bcf0247d3ee38855 .

Many of the actions called by CacheManager, like UsergroupMember.save
will have to find objects that needs permissions to be viewed, hence
CacheManager has to be called "as_admin".

Similarly an user may destroy or save a new UsergroupMember. The cache
needs to be updated with information about *all* user groups in the
system, not only the ones visible to the user making the change.

### DIFF
--- a/app/models/usergroup_member.rb
+++ b/app/models/usergroup_member.rb
@@ -32,12 +32,26 @@ class UsergroupMember < ApplicationRecord
 
   def remove_old_cache_for_old_record
     klass = member_type_changed? ? self.member_type_was.constantize : self.member_type.constantize
-    users = member_id_changed? ? find_all_affected_users_for(klass.find(member_id_was)).flatten : find_all_affected_users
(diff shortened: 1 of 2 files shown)


tbrisker:
@dLobatog Looks like tests are failing

tbrisker:
looks like this is failing when run from `20140219183345_add_taxonomy_searches_to_filter.rb` since the admin role hasn't been seeded yet at that point.

dLobatog (commented):
Updated, it seems that when there are no users available, `User.without_auditing` does a similar job as `as_anonymous_admin`, it skips permissions checking, not sure why though. I took the idea from seeds.rb.

tbrisker:
Strange indeed, but looks like it works.

tbrisker:
Thanks @dLobatog !

PR #4623, opened 2017-06-26
Fixes #19266 - fix_db_cache crashes Rails initialization

dLobatog:
The Filters cache that CacheManager recreates when
fix_db_cache is true, fails when the roles associated with said
filters are locked. To generate the cache, we call `.save` on the filter
object, which is not allowed by Role.

Another issue is that some Filters may have many resource types, which
will also fail to save again to recreate the cache. This behavior is now
autofixed in the cache. [See an example here](https://github.com/theforeman/foreman_remote_execution/blob/master/lib/foreman_remote_execution/engine.rb#L70), the remote_execution plugin creates a role with one filter which mixes many resource types (proxy, hosts, remote execution..)

### DIFF
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -75,6 +75,34 @@ class Role < ApplicationRecord
   scoped_search :on => :builtin, :complete_value => { :true => true, :false => false }
   scoped_search :on => :description, :complete_value => false
 
+  class << self
+    attr_accessor :modify_locked
+
+    def ignore_locking
+      self.modify_locked = true
+      yield
+    ensure
+      self.modify_locked = false
+    end
+
+    # Find all the roles that can be given to a user
+    def find_all_givable
+      all(:conditions => {:builtin => 0}, :order => 'name')
+    end
+
+    # Return the builtin 'Default role' role. If the role doesn't exist,
+    # it will be created on the fly.
+    def default
+      default_role = find_by_builtin(BUILTIN_DEFAULT_ROLE)
+      if default_role.nil?
+        opts = { :name => 'Default role', :builtin => BUILTIN_DEFAULT_ROLE }
+        default_role = create! opts
+        raise ::Foreman::Exception.new(N_("Unable to create the default role.")) if default_role.new_record?
+      end
+      default_role
+    end
+  end
+
   def permissions=(new_permissions)
     add_permissions(new_permissions.map(&:name).uniq) if new_permissions.present?
   end
@@ -108,23 +136,6 @@ class Role < ApplicationRecord
     else
       allowed_permissions.include? action
     end
-  end
-
-  # Find all the roles that can be given to a user
-  def self.find_all_givable
-    all(:conditions => {:builtin => 0}, :order => 'name')
-  end
-
-  # Return the builtin 'Default role' role. If the role doesn't exist,
-  # it will be created on the fly.
-  def self.default
-    default_role = find_by_builtin(BUILTIN_DEFAULT_ROLE)
-    if default_role.nil?
-      opts = { :name => 'Default role', :builtin => BUILTIN_DEFAULT_ROLE }
-      default_role = create! opts
-      raise ::Foreman::Exception.new(N_("Unable to create the default role.")) if default_role.new_record?
-    end
-    default_role
   end
 
   # options can have following keys
@@ -171,6 +182,7 @@ class Role < ApplicationRecord
   end
 
   def locked?
+    return false if self.modify_locked || self.class.modify_locked
     return false unless respond_to? :origin
     origin.present? && builtin != BUILTIN_DEFAULT_ROLE
   end
--- a/app/models/user_role.rb
+++ b/app/models/user_role.rb
@@ -64,7 +64,7 @@ class UserRole < ApplicationRecord
   end
 
   def build_user_role_cache
-    [ self.cached_user_roles.build(:user_id => owner.id, :role_id => role.id) ]
+    [ self.cached_user_roles.build(:user_id => owner_id, :role_id => role_id) ]
   end
 
   def build_user_group_role_cache(owner)
--- a/app/services/cache_manager.rb
+++ b/app/services/cache_manager.rb
@@ -10,7 +10,9 @@ class CacheManager
   end
 
   def self.create_new_filter_cache
-    Filter.all.each(&:save!)
+    Role.ignore_locking do
+      Filter.all.map(&:save!)
+    end
   end
 
   def self.recache!
--- a/test/models/role_test.rb
+++ b/test/models/role_test.rb
@@ -45,50 +45,62 @@ class RoleTest < ActiveSupport::TestCase
     end
   end
 
-  it "should not rename locked role" do
-    manager = Role.find_by :name => "Manager"
-    manager.name = "Renamed Manager"
-    refute manager.save
-    assert_equal "This role is locked from being modified by users.", manager.errors.messages[:base].first
-  end
-
-  it "should rename locked role when required" do
-    manager = Role.find_by :name => "Manager"
-    manager.name = "Renamed Manager"
-    manager.modify_locked = true
-    assert manager.save
-  end
-
-  it "should not modify locked role permissions" do
-    viewer = Role.find_by :name => "Viewer"
-    viewer.add_permissions [:create_hosts]
-    refute viewer.valid?
-    assert_equal "is locked for user modifications.", viewer.errors.messages.values.first.first
-  end
-
-  it "should modify locked role permissions when required" do
-    viewer = Role.find_by :name => "Viewer"
-    viewer.modify_locked = true
-    viewer.add_permissions [:create_hosts]
-    assert viewer.valid?
-  end
-
-  it "should modify role using ignore_locking method" do
-    viewer = Role.find_by :name => "Viewer"
-    viewer.ignore_locking do |role|
-      role.name = "Changed Viewer"
-      role.save!
+  context 'locked roles' do
+    it "should not rename locked role" do
+      manager = Role.find_by :name => "Manager"
+      manager.name = "Renamed Manager"
+      refute manager.save
+      assert_equal "This role is locked from being modified by users.", manager.errors.messages[:base].first
     end
-    assert Role.find_by(:name => "Changed Viewer")
-    refute Role.find_by(:name => "Viewer")
-  end
 
-  it "should be givable even when locked" do
-    user = FactoryGirl.create(:user)
-    role = roles(:manager)
-    assert role.locked?
-    user.roles << role
-    assert user.save
+    it "should rename locked role when required" do
+      manager = Role.find_by :name => "Manager"
+      manager.name = "Renamed Manager"
+      manager.modify_locked = true
+      assert manager.save
+    end
+
+    it "should not modify locked role permissions" do
+      viewer = Role.find_by :name => "Viewer"
+      viewer.add_permissions [:create_hosts]
+      refute viewer.valid?
+      assert_equal "is locked for user modifications.", viewer.errors.messages.values.first.first
+    end
+
+    it "should modify locked role permissions when required" do
+      viewer = Role.find_by :name => "Viewer"
+      viewer.modify_locked = true
+      viewer.add_permissions [:create_hosts]
+      assert viewer.valid?
+    end
+
+    it "should modify role using ignore_locking method" do
+      viewer = Role.find_by :name => "Viewer"
+      viewer.ignore_locking do |role|
+        role.name = "Changed Viewer"
+        role.save!
+      end
+      assert Role.find_by(:name => "Changed Viewer")
+      refute Role.find_by(:name => "Viewer")
+    end
+
+    it "should be givable even when locked" do
+      user = FactoryGirl.create(:user)
+      role = roles(:viewer)
+      assert role.locked?
+      user.roles << role
+      assert user.save
+    end
+
+    # Using 'def' to ensure bypass_locking is disabled after the test
+    def test_allow_to_modify_roles_when_bypass_locking
+      assert roles(:manager).locked?
+      Role.ignore_locking do
+        viewer = roles(:viewer)
+        viewer.add_permissions [:create_hosts]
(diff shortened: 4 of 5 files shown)


dLobatog:
Updated, tests were failing because the original filter could not be deleted in PG

dLobatog (commented):
@ares This PR is basically a fix for what we talked about on IRC - the REX role which contained just 1 filter with 3 different resource types, that breaks "fix_db_cache" calls on upgrade. 

I couldn't find the original bug where that was solved to point to it here though. cc @xprazak2

app/services/cache_manager.rb
```diff
@@ -10,7 +10,39 @@ def self.create_new_permission_cache
...
+    locked_roles.each { |role| role.origin = nil }
+    locked_roles.each { |role| role.modify_locked = true }
+    locked_roles.each(&:save)
+    Filter.all.each do |filter|
+      next if filter.save
+
+      if filter.errors[:permissions] == ['Permissions must be of same resource type']
+        # If the filter was created long time ago, it may have had several
+        # resources of the same type. To fix it, let's split it into different
+        # filters each with its own resource type.
```
tbrisker: Is it possible to do this in a migration? If I understand correctly this is caused by some old filters that weren't migrated correctly and would only need to be done once. I wouldn't expect this to be part of the cache rebuild.

app/services/cache_manager.rb
```diff
@@ -10,7 +10,39 @@ def self.create_new_permission_cache
...
+        end
+
+        # At this point, all permissions are in their own filters by resource type
+        filter.errors.clear
+        filter.destroy
+      end
+    end
+    locked_roles.each_with_index { |role, index| role.origin = origins[index] }
+    locked_roles.each { |role| role.modify_locked = false }
+    locked_roles.each(&:save)
```
tbrisker: Would it make more sense to just run the actual callback that we need for the cache to be fixed rather then having to make this complex workaround to bypass the lock validation? Or maybe change the lock validation to not run when in rake?

tbrisker (changes requested):

app/services/cache_manager.rb
dLobatog: Yes, I could make this a separate issue and make it in a migration, it's not a problem.

app/services/cache_manager.rb
dLobatog: I'm not really sure, but I think calling `.save` is what we want - so that the related objects (Filterings, Permissions, Role, UserRole..) are also running callbacks. At least for `create_new_permission_cache` the point is precisely to run callbacks on all the objects related with the saved UsergroupMember and UserRole. I think `.save` is also future-proof. 

Disabling the lock validation while in Rake isn't ideal, as tests would have to handle that somehow too, not to mention seeds could also accidentally update locked roles, etc...

tbrisker: I still must say I don't care too much for the hack to get around the locking, maybe we need to add a class level bypass as well, kind of like we have for auditing? Then we can have something like:
```ruby
Role.bypass_locks = true
Filter.all.each(&:save!)
Role.bypass_locks = false
```
and we can reuse it if we need in other places without having to manually modify every role twice (origin and modify_locked)

tbrisker:
While this seems to work when running the rake task, in the initializer (if `Setting['fix_db_cache']==true`) this fails for me with:
```
/home/tbrisker/foreman/app/models/user_role.rb:67:in `build_user_role_cache': undefined method `id' for nil:NilClass (NoMethodError)
	from /home/tbrisker/foreman/app/models/user_role.rb:56:in `cache_user_roles!'
	from /home/tbrisker/.rvm/gems/ruby-2.1.10@foreman_gemset/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:432:in `block in make_lambda'
	from /home/tbrisker/.rvm/gems/ruby-2.1.10@foreman_gemset/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:228:in `call'
	from /home/tbrisker/.rvm/gems/ruby-2.1.10@foreman_gemset/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:228:in `block in halting_and_conditional'
	from /home/tbrisker/.rvm/gems/ruby-2.1.10@foreman_gemset/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:506:in `call'
	from /home/tbrisker/.rvm/gems/ruby-2.1.10@foreman_gemset/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:506:in `block in call'
	from /home/tbrisker/.rvm/gems/ruby-2.1.10@foreman_gemset/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:506:in `each'
	from /home/tbrisker/.rvm/gems/ruby-2.1.10@foreman_gemset/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:506:in `call'
	from /home/tbrisker/.rvm/gems/ruby-2.1.10@foreman_gemset/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:92:in `__run_callbacks__'
	from /home/tbrisker/.rvm/gems/ruby-2.1.10@foreman_gemset/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:778:in `_run_save_callbacks'
	from /home/tbrisker/.rvm/gems/ruby-2.1.10@foreman_gemset/gems/activerecord-4.2.8/lib/active_record/callbacks.rb:302:in `create_or_update'
	from /home/tbrisker/.rvm/gems/ruby-2.1.10@foreman_gemset/gems/activerecord-4.2.8/lib/active_record/persistence.rb:142:in `save!'
	from /home/tbrisker/.rvm/gems/ruby-2.1.10@foreman_gemset/gems/activerecord-4.2.8/lib/active_record/validations.rb:43:in `save!'
	from /home/tbrisker/.rvm/gems/ruby-2.1.10@foreman_gemset/gems/activerecord-4.2.8/lib/active_record/attribute_methods/dirty.rb:29:in `save!'
	from /home/tbrisker/.rvm/gems/ruby-2.1.10@foreman_gemset/gems/activerecord-4.2.8/lib/active_record/transactions.rb:291:in `block in save!'
	from /home/tbrisker/.rvm/gems/ruby-2.1.10@foreman_gemset/gems/activerecord-4.2.8/lib/active_record/transactions.rb:351:in `block in with_transaction_returning_status'
	from /home/tbrisker/.rvm/gems/ruby-2.1.10@foreman_gemset/gems/activerecord-4.2.8/lib/active_record/connection_adapters/abstract/database_statements.rb:213:in `block in transaction'
	from /home/tbrisker/.rvm/gems/ruby-2.1.10@foreman_gemset/gems/activerecord-4.2.8/lib/active_record/connection_adapters/abstract/transaction.rb:184:in `within_new_transaction'
	from /home/tbrisker/.rvm/gems/ruby-2.1.10@foreman_gemset/gems/activerecord-4.2.8/lib/active_record/connection_adapters/abstract/database_statements.rb:213:in `transaction'
	from /home/tbrisker/.rvm/gems/ruby-2.1.10@foreman_gemset/gems/activerecord-4.2.8/lib/active_record/transactions.rb:220:in `transaction'
	from /home/tbrisker/.rvm/gems/ruby-2.1.10@foreman_gemset/gems/activerecord-4.2.8/lib/active_record/transactions.rb:348:in `with_transaction_returning_status'
	from /home/tbrisker/.rvm/gems/ruby-2.1.10@foreman_gemset/gems/activerecord-4.2.8/lib/active_record/transactions.rb:291:in `save!'
	from /home/tbrisker/.rvm/gems/ruby-2.1.10@foreman_gemset/gems/activerecord-4.2.8/lib/active_record/relation/delegation.rb:46:in `map'
	from /home/tbrisker/.rvm/gems/ruby-2.1.10@foreman_gemset/gems/activerecord-4.2.8/lib/active_record/relation/delegation.rb:46:in `map'
	from /home/tbrisker/foreman/app/services/cache_manager.rb:9:in `create_new_permission_cache'
	from /home/tbrisker/foreman/app/services/cache_manager.rb:51:in `recache!'
	from /home/tbrisker/foreman/config/initializers/fix_cache.rb:10:in `<top (required)>'
<SNIP>
```

app/services/cache_manager.rb
tbrisker: In any case I don't think you need to modify the origin, just setting modify_lock to true should be enough to prevent the lock validation (and if it doesn't then it should)

dLobatog (commented):
Splitted the PR into https://github.com/theforeman/foreman/pull/4638 and this one, and also I've implemented `bypass_locking` at the class level as suggested by @tbrisker (tests pending, should be simple)

app/models/user_role.rb
```diff
@@ -64,7 +64,7 @@ def cache_user_roles!
   end
 
   def build_user_role_cache
-    [ self.cached_user_roles.build(:user_id => owner.id, :role_id => role.id) ]
+    [ self.cached_user_roles.build(:user_id => owner_id, :role_id => role_id) ]
```
dLobatog, line 67: @tbrisker This is what was missing and why the initializer failed (`.id` can't be called in the middle of a callback if the object is not saved)

dLobatog (commented):
Updated to add a little test for `Role.bypass_locking`

app/models/role.rb
```diff
@@ -75,6 +75,27 @@ class Role < ApplicationRecord
   scoped_search :on => :builtin, :complete_value => { :true => true, :false => false }
   scoped_search :on => :description, :complete_value => false
 
+  class << self
+    attr_accessor :bypass_locking
```
ares: could it perhaps be wrapped in a class method accepting block like the instance method does? that could contain ensure block with resetting the bypass, ideally not setting it to false but to $original_value in case someone has two bypasses nested

and I'd prefer if the name would be consistent with instance variant which is called `ignore_locking`

app/models/role.rb
```diff
@@ -171,6 +175,7 @@ def clone(role_params = {})
   end
 
   def locked?
+    return false if self.class.bypass_locking
```
ares: Why is this needed? the role is still locked hence it should return true, we're just bypassing it. If that's because of filters validations/filters, I think [existing check](https://github.com/dLobatog/foreman/blob/fc614898bea04d0998d01f4411fdb2e0357d39bf/app/models/filter.rb#L226) should be extended

app/services/cache_manager.rb
```diff
@@ -10,7 +10,10 @@ def self.create_new_permission_cache
   end
 
   def self.create_new_filter_cache
-    Filter.all.each(&:save!)
+    Role.bypass_locking = true
+    Filter.all.map(&:save!)
+  ensure
```
ares: if we had a wrapping method, we wouldn't have to remember to do `ensure` whenever we bypass the locking

ares (changes requested):
Mostly looks good, just few questions/comments

app/models/role.rb
dLobatog: If we extend the existing `not_locked` check, it still fails. The reason is that `Filter#role_not_locked` would also need to be aware of `Role.bypass_locking`.

 I don't mind changing the check twice, but making the Filter `role_not_locked` validator check on the state of other class is strange (that check should be in `add_permissions` probably)

app/models/role.rb
dLobatog: The class variable could be `modify_locked` too, and the class method `ignore_locking`. I tried doing this with using a class variable (same method as the instance), but it didn't quite work yet. How would you do it?

tbrisker: I agree it makes sense to have a common name for both the class and instance version of this, as well as a wrapper function that ensures it returns to the original value.

app/models/role.rb
tbrisker: I agree, if we're disabling locking then no roles are locked and it makes sense to return false.

dLobatog (commented):
Updated @tbrisker & @ares , note it works by modifying the `def locked?` method on each role. If I don't modify that, I'm forced to add checks on `role_not_locked` in Filter, and `not_locked` in Role. 

This double validation should probably be refactored in develop (this is supposed to go to 1.15).

Also the test in `cache_manager_test.rb` is going to break tests in REX, Ansible & Discovery, the Filters in tests have multiple resource types (http://i.imgur.com/OqqiaLI.png). I'll take care of that (https://github.com/theforeman/foreman/pull/4638 + fixes to each of the plugins.)

tbrisker (approved):

app/models/role.rb
```diff
@@ -171,6 +182,7 @@ def clone(role_params = {})
   end
 
   def locked?
+    return false if self.class.modify_locked
```
tbrisker: shouldn't this also return false if `self.modify_locked`?

dLobatog: There's a check in https://github.com/theforeman/foreman/pull/4623/files#diff-7863dd934f73db4d813673b9dec6db0aR217 for that already. 

The problem is that if I modify the `not_locked` validation to include `self.class.modify_locked`, it will only prevent name modifications and things like that. In order to prevent adding permissions in addition to all of that, I can either:

* Put it here
or
* Put it in `not_locked` and `Filter#role_not_locked` (at least)

Since (IMO) Filter shouldn't really need to depend on a class variable from Role, I put it here.

tbrisker: I know the check is there, i was wondering if it made sense to move that here as well, so `locked?` will return `false` if ignore locking is used either on the instance or class level.

tbrisker: i.e. 
```ruby
return false if self.modify_locked || self.class.modify_locked
```

dLobatog: If I move it up here, I should be able to remove it in the validator (it's redundant) - I will not, because it's meant for 1.15 though

dLobatog:
@tbrisker Updated `locked?` with the instance-level `modify_locked`

tbrisker:
:+1:  will merge once Mr. Jenkins is done.

mmoll:
If somebody is interested, that dynflow hang is seen sometimes in the katello test, I saved the test.log of that run if anybody wants to have a look...

tbrisker:
Thanks @dLobatog !

PR #4625, opened 2017-06-26
Fixes #20097 - Foreman plugin support for webpack/react

ohadlevy:
How to use:

in a plugin, create a directory called webpack
then inside a file called index.js put something like:

```js
import React from 'react';
import ReactDOM from 'react-dom';
import Icon from 'foremanReact/common/Icon';

const reactNode = document.querySelector('#content');

if (reactNode) {
  ReactDOM.render(
    <Icon type='ok' />,
    reactNode);
};

```
TODOS:
- [ ] support JEST / eslint
- [ ] extend the react mounter, so plugins can add more mounting points and use
them.
- [ ] update storybook to support plugins too
- [ ] make equivalent of rake test

### DIFF
--- a/app/views/layouts/base.html.erb
+++ b/app/views/layouts/base.html.erb
@@ -15,6 +15,7 @@
     <%= yield(:stylesheets) %>
 
     <%= csrf_meta_tags %>
+    <%= javascript_include_tag *webpack_asset_paths('vendor', :extension => 'js'), "data-turbolinks-track" => true %>
     <%= javascript_include_tag *webpack_asset_paths('bundle', :extension => 'js'), "data-turbolinks-track" => true %>
     <%= javascript_include_tag "locale/#{FastGettext.locale}/app", "data-turbolinks-track" => true %>
     <%= javascript_include_tag 'application', "data-turbolinks-track" => true %>
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-var*/
 'use strict';
 
 var path = require('path');
@@ -5,6 +6,7 @@ var webpack = require('webpack');
 var StatsPlugin = require('stats-webpack-plugin');
 var ExtractTextPlugin = require('extract-text-webpack-plugin');
 var CompressionPlugin = require('compression-webpack-plugin');
+var execSync = require('child_process').execSync;
 
 // must match config.webpack.dev_server.port
 var devServerPort = 3808;
@@ -14,12 +16,15 @@ var production =
   process.env.RAILS_ENV === 'production' ||
   process.env.NODE_ENV === 'production';
 
-var config = {
-  entry: {
-    // Sources are expected to live in $app_root/webpack
-    bundle: './webpack/assets/javascripts/bundle.js'
-  },
+var plugins = JSON.parse(execSync('./script/plugin_webpack_directories.rb'));
 
+var config = {
+  entry: Object.assign(
+    {
+      bundle: './webpack/assets/javascripts/bundle.js'
+    },
+    plugins.entries
+  ),
   output: {
     // Build assets directly in to public/webpack/, let webpack know
     // that all webpacked assets start with webpack/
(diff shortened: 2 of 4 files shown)


ohadlevy:
this replaces #4619

app/views/layouts/base.html.erb
```diff
@@ -15,6 +15,7 @@
     <%= yield(:stylesheets) %>
 
     <%= csrf_meta_tags %>
+    <%= javascript_include_tag *webpack_asset_paths('vendor', :extension => 'js'), "data-turbolinks-track" => true %>
```
dLobatog, line 18: I see it's mostly React (the library itself), however it doesn't seem to get my `../foreman_ansible/webpack/index.js` content in it.. https://gist.github.com/dLobatog/96f33a68be7fdd92ba79e8d5d6219496 is what I see

config/webpack.config.js
```diff
@@ -14,12 +16,15 @@ var production =
...
-    bundle: './webpack/assets/javascripts/bundle.js'
-  },
+var plugins = JSON.parse(execSync('./script/plugin_webpack_directories.rb'));
 
+var config = {
+  entry: Object.assign(
+    {
+      bundle: './webpack/assets/javascripts/bundle.js'
+    },
+    plugins.entries
```
dLobatog, line 26: Out of curiosity, where is the part that adds '-core.js' to the name of the config entries? I see `ansiblecore.js` here for example, https://gist.github.com/7b39b650c625aba5d7f4216fa7a0c61d (on foreman_ansible I just have `webpack/index.js`)

script/plugin_webpack_directories.rb
```diff
@@ -0,0 +1,26 @@
+#! /usr/bin/env ruby
+
+require 'bundler'
+require 'json'
+
+PLUGIN_NAME_REGEXP = /foreman*|katello*/
+
+# Only works for local dependencies
```
dLobatog: Do you mean it only works for gems that have the source path set by `:path=>` in the Gemfile?

dLobatog (changes requested):
@ohadlevy Thanks for improving #4619! 

I have been trying it, and it compiles fine.  My problem is that even if I see ../foreman_ansible/index.js on the webpack compile output, however I can't find it neither on myforklifturl:3808, nor on the Firefox/Chromium inspector... I tried to do something stupid like putting `console.log` or `alerts` in there and they don't seem to be loaded either. What could I be missing?

app/views/layouts/base.html.erb
ohadlevy: it wont, you need to include it in your plugin.
e.g, in a page somewhere (either in the layout or a specific page) add:
```erb
+<%= javascript_include_tag *webpack_asset_paths('name of chunk'', :extension => 'js'), "data-turbolinks-track" => true %>
```

config/webpack.config.js
ohadlevy: I think "core" comes from plugins that have it in their names, e.g. foreman-tasks-core, rex-core etc.

have a look at the ruby script output, it defines the chunk name. I'm uncertain why you have the extra word core in there either.

script/plugin_webpack_directories.rb
ohadlevy: actually, you can remove it, it was a copy paste from your original script.

config/webpack.config.js
dLobatog: It makes sense now, it reads foreman_ansible_core first apparently 😄

ohadlevy: just wondering - why do you have a core spec in there to begin with?

ohadlevy:
@dLobatog @waldenraines do you think it makes sense to merge as is? this should unblock basic usage and allow experimenting with packaging efforts?

config/webpack.config.js
dLobatog: @ohadlevy The agent that executes the roles can run on the proxy or on Foreman itself. In fact it's the first plugin that introduced this "_core" gem dependency concept to allow reusing the code to run actions in the proxy or in Foreman.

dLobatog (approved):

dLobatog (changes requested):
@ohadlevy Yeah, I agree - I was able to successfully do basic things like importing `'foremanReact/common/Alert` and using it from a plugin. 

Can you rebase, squash and use the proper commit title to reference the Redmine issue? It should be ready after that I think.

ohadlevy:
@dLobatog updated.

dLobatog (approved):

dLobatog:
Thanks @ohadlevy !

waldenraines:
> @dLobatog @waldenraines do you think it makes sense to merge as is? this should unblock basic usage and allow experimenting with packaging efforts?

I'm a little late to the party but yes 😄  and thanks @ohadlevy!

PR #4626, opened 2017-06-28
Fixes #20134 - Compute Profile for Ovirt loads template data

orrabin:


dLobatog (changes requested):
@orrabin Marking this one as 1.15.2, is there any way to write a small test for this that verifies the value is changed? Thanks!

orrabin:
@dLobatog I can add an integration test for the value change but that would only test the use of this js on the host form and this pr is fixing a bug when using the same js from compute profiles. In my case the value shouldn't change because `templateSelector[0]` is undefined.

dLobatog (approved):

dLobatog:
@orrabin Oh I see it now.. OK - I'm not really sure it makes much sense to test just the selector. Hopefully we can move this to be tested with Jest at some point. Merging...

PR #4630, opened 2017-06-29
Fixes #20143 - changed kind for Preseed default PXEGrub2

lzap:
Otherwise EFI provisioning for Debian/Ubuntu is not possible, 1.15 candidate @dLobato thanks.

### DIFF
--- a/db/migrate/20170629080604_change_preseed_default_pxe_grub2_kind.rb
+++ b/db/migrate/20170629080604_change_preseed_default_pxe_grub2_kind.rb
@@ -1,0 +1,19 @@
+class ChangePreseedDefaultPxeGrub2Kind < ActiveRecord::Migration
+  def update_kind
+    kind = TemplateKind.find_by_name(:PXEGrub2)
+    tmpl = ProvisioningTemplate.unscoped.find_by_name("Preseed default PXEGrub2")
+    tmpl.update_attribute(:template_kind, kind) if tmpl && kind
+  end
+
+  def up
+    if User.unscoped.find_by_login(User::ANONYMOUS_ADMIN).present?
+      User.as_anonymous_admin do
+        update_kind
+      end
+    else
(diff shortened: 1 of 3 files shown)


dLobatog (changes requested):
@lzap Thanks! I noticed one thing that's wrong.. 

` db/seeds.d/02-partition_tables_list.rb` contains `ProvisioningTemplatsList`, and viceversa, `02-provisioning_templates_list.rb` contains the PartitionTables list.

Mind to rename them so each has the right content? The change was introduced in bbb1b90c1dae0aebf20e6ce04b34bc801355423b, as you can see it won't change any behavior.

ares:
Good catch @dLobatog, I'm not sure how that happened, I can rename them in separate PR if you want since I've messed up.

tbrisker:
looks like migrations are failing since there is no admin existing before they finish and the seed is executed, similar to the issue we hit in https://github.com/theforeman/foreman/pull/4622 - you can use the workaround @dLobatog used there, but if we continuously run into this issue in migrations maybe we should fix it in the `as_anonymous_admin` helper?

lzap:
I implemented the hack that @tbrisker recommended, since this goes into stable release I would like to avoid any refactoring. Can you create issue for that? Also what do you recommend, simply to skip if user is not present?

I swapped contents of the two files but in separate commit.

dLobatog (approved):

dLobatog:
@lzap I think the way it is is fine now, thanks 😄 I would not skip if the user is not present, rather use `without_auditing` and ensure the change happens. I'll cherry-pick to 1.15.2 only the first commit, as the problem I mentioned in http://projects.theforeman.org/issues/20192 is not present in that branch even.

dLobatog:
Damn, I accidentally clicked squash -_- I'll ensure only the changes you made on the first commit get on 1.15.2 though

PR #4631, opened 2017-06-29
Fixes #20146 - domain-less server email errors fixed

lzap:
On a server without domain (e.g. `hostname -f` returning just host name) there were email validation errors because domain part was `nil` via Facter. This fixes it - it falls back to hostname if domain is not set.

More than that, our design was weird - there is a config setting `domain` in `settings.yaml` which I believe should override what's set on system but it does not - Facter wins. I refactored the code so it really overrides when it is present.

Also lowered case for the default "Foreman-noreply" email which looks much saner to me.

### DIFF
--- a/config/settings.rb
+++ b/config/settings.rb
@@ -1,5 +1,6 @@
 require_relative 'boot_settings'
 require_relative '../app/services/foreman/version'
+require 'facter'
 
 root = File.expand_path(File.dirname(__FILE__) + "/..")
 settings_file = Rails.env.test? ? 'config/settings.yaml.test' : 'config/settings.yaml'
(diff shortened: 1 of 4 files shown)


config/settings.rb
```diff
@@ -11,6 +12,7 @@
 SETTINGS[:puppetconfdir] ||= '/etc/puppet'
 SETTINGS[:puppetvardir]  ||= '/var/lib/puppet'
 SETTINGS[:puppetssldir]  ||= "#{SETTINGS[:puppetvardir]}/ssl"
+SETTINGS[:domain] ||= Facter.value(:domain) || Facter.value(:hostname)
```
dLobatog: I don't think it's meant to work when the domain is set to the hostname. It certainly avoids the error though...

dLobatog (approved):
@lzap Thanks, this looks like it makes `SETTINGS[:domain]` prevail over Facter, which makes sense. 

Can you add a couple of tests to proof the behavior? Normally that's required for stuff that goes into minor releases unless it's too obvious (e.g the PXEGrub PR wouldn't need them) or difficult to test

boaz0 (approved):
LGTM

tbrisker:
Please rebase, does not merge cleanly.

lzap:
Rebased, @dLobatog I wish I could test but this is Rails initialize code (foreman/config/settings.rb), I'd need to extract this into external class like DefaultSettings and then test that. Is it worth the effort tho? The other two files simply uses the global setting "domain".

dLobatog (approved):
Alright - let's merge to develop then. Thanks @lzap

lzap:
Thanks, not an important bug.

PR #4633, opened 2017-06-30
Fixes #20174 - handle no storage pool gracefully

ares:
This PR fixes the ugly error and as a side effect it:
* refactors how we detect that compute resource supports adding volumes
* allows provisioning vms on libvirt without storage (probably not too useful)
* introduces a way to detect storage errors before we run the provisioning orchestration, could be reused by other CRs that support it

Mainly it fixes 500 on compute profile edit, after applying this patch you can see:
20175_after_profile

Note that it also affects host form, before if the storage pool was missing, you could see following error on VM tab
20174_after_host

Now it only displays warning and it allows you to continue with provisioning
20176_before_host

To reproduce the issue you can simply deactivate your libvirt storage pools e.g. using virt-manager, you don't have to delete them.

### DIFF
--- a/app/models/compute_resources/foreman/model/libvirt.rb
+++ b/app/models/compute_resources/foreman/model/libvirt.rb
@@ -29,7 +29,7 @@ module Foreman::Model
     end
 
     def capabilities
-      [:build, :image]
+      [:build, :image, :new_volume]
     end
 
     def editable_network_interfaces?
@@ -75,7 +75,7 @@ module Foreman::Model
       errors[:base] << e.message
     end
 
-    def new_nic(attr = { })
+    def new_nic(attr = {})
       client.nics.new attr
     end
 
@@ -84,8 +84,15 @@ module Foreman::Model
       new_nic(attr)
     end
 
-    def new_volume(attr = { })
+    def new_volume(attr = {})
+      return unless new_volume_errors.empty?
       client.volumes.new(attr.merge(:allocation => '0G'))
+    end
+
+    def new_volume_errors
+      errors = []
+      errors.push _('no storage pool available on hypervisor') if storage_pools.empty?
+      errors
     end
 
     def storage_pools
@@ -210,7 +217,7 @@ module Foreman::Model
       super.merge(
         :memory     => 768.megabytes,
         :nics       => [new_nic],
-        :volumes    => [new_volume],
+        :volumes    => [new_volume].compact,
         :display    => { :type     => display_type,
                          :listen   => Setting[:libvirt_default_console_address],
                          :password => random_password,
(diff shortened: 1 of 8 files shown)


app/views/compute_resources_vms/form/_volumes.html.erb
```diff
@@ -1,20 +1,24 @@
...
+          <% end %>
 
-      <% if new_vm %>
-        <%= add_child_link '+ ' + _("Add Volume"), :volumes, { :class => "info", :title => _('add new storage volume') } %>
+          <% if new_vm %>
+            <%= add_child_link '+ ' + _("Add Volume"), :volumes, { :class => "info", :title => _('add new storage volume') } %>
+          <% end %>
+        </div>
+      <% else %>
+        <%= alert :text => _('Unable to define volumes because of:') + ' ' + compute_resource.new_volume_errors.join(', '), :close => false, :class => 'alert-warning', :header => '' %>
```
dLobatog: Nitpick, just `because` or even nothing (`Unable to define volumes: no storage pool available on hypervisor`) are more similar to other errors in Foreman.

app/models/compute_resources/foreman/model/libvirt.rb
```diff
@@ -84,8 +84,16 @@ def new_interface(attr = {})
       new_nic(attr)
     end
 
-    def new_volume(attr = { })
-      client.volumes.new(attr.merge(:allocation => '0G'))
+    def new_volume(attr = {})
+      if new_volume_errors.empty?
```
dLobatog: Nitpick, use a guard clause like `return if new_volume_errors.empty?` to make it obvious what happens (the method returnsn il) and avoid having a method wrapped in an if condition

dLobatog (changes requested):
Thanks @ares , please check out the error message, everything else works fine. I tried to create a host without a Storage pool though, but that fails (not the fault of this PR)

ares:
Thanks @dLobatog, I fixed both comments. I wonder what prevented you from creating a host on libvirt without storage. It worked for me locally.

Gemfile
```diff
@@ -9,7 +9,7 @@ gem 'audited', '~> 4.3'
 gem 'will_paginate', '~> 3.0'
 gem 'ancestry', '>= 2.0', '< 4'
 gem 'scoped_search', '~> 4.0'
-gem 'ldap_fluff', '>= 0.3.5', '< 1.0'
+gem 'ldap_fluff', '>= 0.3.5', '< 1.0', :path => '/home/ares/z/ldap_fluff'
```
mmoll: hrm... ;)

mmoll (changes requested):

Gemfile
ares: damn, I'm so used to type `-a --amend`, thanks for spotting

dLobatog (approved):

dLobatog:
Thanks @ares !

PR #4634, opened 2017-06-30
Fixes #20162 - Bring consistency to forms

rabajaj0509:
This commit shall bring consistency to all forms and comply to the patterns according to patternfly.


rabajaj0509:
related bug - https://github.com/theforeman/foreman_discovery/pull/357

rabajaj0509:
removing the 'well' will affect all forms like tasks, etc. therefore we must let it be as is.

Sorry for the inconvenience.

PR #4637, opened 2017-07-02
Fixes #19825 - Fix Performance/TimesMap cop

tbrisker:


dLobatog (approved):
Interesting one

dLobatog:
Thanks @tbrisker !

tbrisker:
@dLobatog Yep, these cases aren't really significant performance wise, but I wanted to enable it so that future code will be checked against this.

PR #4638, opened 2017-07-03
Fixes #20190 - Fix filters with more than one resource_type

dLobatog:
As shown in #19266, some permissions could have been introduced via a
migration in previous versions of Foreman when there was no validation
for having an unique resource_type for each Filter.

### DIFF
--- a/db/migrate/20170703140743_fix_filters_with_multiple_resource_types.rb
+++ b/db/migrate/20170703140743_fix_filters_with_multiple_resource_types.rb
@@ -1,0 +1,31 @@
+class FixFiltersWithMultipleResourceTypes < ActiveRecord::Migration
+  def change
+    filters_to_split = Filter.unscoped.joins(:permissions).group('filters.id', :resource_type).
+                              having('min(resource_type) <> max(resource_type)').pluck(:id)
+    Filter.unscoped.includes(:permissions).references(:permissions).where(:id => filters_to_split).
+      each do |filter|
+      next if filter.valid?
+      if filter.errors[:permissions] == ['Permissions must be of same resource type']
+        # If the filter was created long time ago, it may have had several
+        # resources of the same type. To fix it, let's split it into different
+        # filters each with its own resource type.
+        filter.permissions.group_by(&:resource_type).each do |resource_type, permissions|
+          # Create a new filter for all resource types
+          restricted_resource_type_filter = Filter.new(
+            :role => filter.role,
+            :permissions => permissions,
+            :taxonomy_search => filter.taxonomy_search,
+            :organizations => filter.organizations,
+            :locations => filter.locations,
+            :override => filter.override
+          )
+          restricted_resource_type_filter.save
+        end
+
+        # At this point, all permissions are in their own filters by resource type
+        filter.errors.clear
+        filter.destroy
+      end
+    end
+  end
+end


dLobatog:
[test katello]

db/migrate/20170703140743_fix_filters_with_multiple_resource_types.rb
```diff
@@ -0,0 +1,29 @@
+class FixFiltersWithMultipleResourceTypes < ActiveRecord::Migration
+  def change
+    Filter.unscoped.all.each do |filter|
+      next if filter.valid?
+
+      if filter.errors[:permissions] == ['Permissions must be of same resource type']
+        # If the filter was created long time ago, it may have had several
+        # resources of the same type. To fix it, let's split it into different
+        # filters each with its own resource type.
+        filter.permissions.group_by(&:resource_type).each do |resource_type, permissions|
```
ShimShtein, line 12: It's an N+1 query. I don't know how many filters should be present in one system, but if there are a lot of them this can cause a really slow migration.

dLobatog (commented):
There could be many filters (322 I see on my install), however I assume most of them would be valid. 
Invalid filters like the ones fixed by this migration, that's much more unlikely. The only cases I've seen is if you had installed Remote Execution a long time ago, before Foreman validated "single resource_types" per Filter, you could have invalid filters (which blocked http://projects.theforeman.org/issues/19266)

cc @ShimShtein

ShimShtein:
Just to reduce the amount of queries to minimum:
``` ruby
filters_to_split = Filter.unscoped.joins(:permissions).group('filters.id', :resource_type).having('min(resource_type) = max(resource_type)').pluck(:id)
Filter.unscoped.includes(:permissions).where(:id => filters_to_split).each do |filter_to_fix|
  filter_to_fix.permissions.group_by(&:resource_type).each do |resource_type, permissions|
    # Do work
  end
end
```
First query will get only filters with more than one resource type, and the second one will preload permissions for those filters. This way you should have bounded amount of queries.

ShimShtein:
There is an error in my previous [comment](https://github.com/theforeman/foreman/pull/4638#issuecomment-322031825), it should be `<>`:
``` ruby
filters_to_split = Filter.unscoped.joins(:permissions).group('filters.id', :resource_type).having('min(resource_type) <> max(resource_type)').pluck(:id)
```
Sorry for misleading!

dLobatog (commented):
@ShimShtein Updated, I don't think any systems will have more than 6 filters in this situation though!

dLobatog:
[test foreman]

dLobatog:
[test katello]

ShimShtein:
Started testing it, but the results were not as expected. Found out that there was an error in the first query. We shouldn't group by `resource_type`, otherwise, the results will always be empty. We should remove it from the group clause:
``` ruby
filters_to_split = Filter.unscoped.joins(:permissions).group('filters.id').having('min(resource_type) <> max(resource_type)').pluck(:id)
```

mmoll:
what's the status here?

lzap:
There is similar bugreport for 1.19, @dLobatog do you think it's relevant? https://projects.theforeman.org/issues/24941 What is the status here?

db/migrate/20170703140743_fix_filters_with_multiple_resource_types.rb
lzap: We are talking about dozens of them.

lzap:
The query does not seem to work. I created a filter with two permissions with different resource types:

```ruby
Filter.unscoped.joins(:permissions).group('filters.id', :resource_type).pluck(:id, :resource_type)
[[1, "AuthSource"],
 [1, "Bookmark"]]
```

However the query does not detect it:

```ruby
Filter.unscoped.joins(:permissions).group('filters.id', :resource_type).having('min(resource_type) <> max(resource_type)').pluck(:id)
=> []
```

Honestly, I don't understand the idea of the grouping functions on VARCHAR type, I read that MIN/MAX should work with string values returning "maximum value of expression across all input values" but I am afraid it does not do what it was supposed to do: https://www.postgresql.org/docs/11/functions-aggregate.html

Since nobody complained for such a long time and the patch does not work (for me), I am closing this PR and the issue. Let's reopen when needed.

PR #4639, opened 2017-07-03
fixes #20193 - upgrade storybook to 3.1

ohadlevy:


webpack/stories/index.js
```diff
@@ -1,7 +1,9 @@
 import React from 'react';
 
 /* eslint-disable no-unused-vars */
```
tbrisker: is this comment needed? I don't see any unused vars (I think linkTo was the unused one?)

package.json
```diff
@@ -76,7 +80,7 @@
     "lint": "./node_modules/.bin/eslint -c .eslintrc webpack/ || exit 0",
     "test": "node node_modules/.bin/jest",
     "test:watch": "node node_modules/.bin/jest --watchAll",
-    "storybook": "start-storybook --dont-track -p 6006",
```
tbrisker, line 79: does this mean that tracking is enabled now or did the default change?

.storybook/webpack.config.js
```diff
@@ -1,23 +1,34 @@
...
+          'plugins': [
+            path.join(__dirname, '..', 'node_modules/babel-plugin-transform-object-rest-spread'),
+            path.join(__dirname, '..', 'node_modules/babel-plugin-transform-object-assign')
+          ]
+        }
       },
       {
         test: /\.css$/,
-        loader: 'style-loader!css-loader'
+        loaders: ['style-loader', 'css-loader', 'sass-loader']
```
tbrisker: I think `sass-loader` shouldn't run on `.css` files, only on `.scss`

package.json
```diff
@@ -5,7 +5,12 @@
   "description": "Foreman isn't really a node module, these are just dependencies needed to build the webpack bundle. 'dependencies' are the asset libraries in use and 'devDependencies' are used for the build process.",
   "private": true,
   "devDependencies": {
-    "@kadira/storybook": "^2.5.2",
+    "@storybook/addon-actions": "^3.1.2",
+    "@storybook/addon-info": "^3.1.4",
+    "@storybook/addon-knobs": "3.1.2",
```
tbrisker: Is there a reason this one is locked to a specific version and the other addons are not? 
Also, do we actually use any of the addons other then actions anywhere?

tbrisker (changes requested):

package.json
ohadlevy: sadly that option has been removed, I was unable to find if it actually tracks or not, based on my browser connections it seems like it does not.

ohadlevy:
@tbrisker updated.

tbrisker (approved):

tbrisker:
not sure why, but looks like tests broke completely?

tbrisker:
Ok, looks like tests broke due to #4625, not related to this, opened #4644 to make sure js tests run when webpack config changes. Merging.

tbrisker:
Thanks @ohadlevy !

PR #4640, opened 2017-07-04
Fixes #20201 - Global HTTP proxy for outgoing requests

bastilian:
This introduce two new settings under "General":

* **http_proxy** - proxy address in the format of http://PROXY:PORT/
* **http_proxy_except_list** - array of strings, including wildcards for hosts that should not be proxied (`*.example.com` or `1.1.1.*`)

The extensions for Excon::Connection and Net::HTTP allow setting a proxy on their requests.

It is an additional/supporting feature to #4616 and a final feature to cover most cases requiring various HTTP proxies to route requests correctly.


jlsherrill:
Would it make sense to coordiante this with https://github.com/theforeman/foreman/pull/4616 ?
EDIT: oh i see you mentioned that above

lib/foreman/http_proxy.rb
```diff
@@ -0,0 +1,42 @@
+module Foreman
+  module HTTPProxy
+    def http_proxy
+      Setting::General.find_by_name('http_proxy').try(:value)
```
ares: could be written as `Setting[:http_proxy]` which also uses the cache

lib/foreman/http_proxy.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    def http_proxy_except_list
+      Setting::General.find_by_name('http_proxy_except_list').try(:value)
```
ares: could be written as `Setting[:http_proxy_except_list]` which also uses the cache

lib/foreman/http_proxy.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    def proxy_http_request?(current_proxy, request_host, schema)
+      !http_proxy.nil? && current_proxy.nil? && !request_host.nil? &&
```
ares: http_proxy might be also empty string if user resets the value in UI, `blank`? would probably be a better fit

lib/foreman/http_proxy.rb
```diff
@@ -0,0 +1,42 @@
...
+      schema != 'unix' &&
```
ares: would whitelist be safer? I suppose we only support `http` and `https` schema

lib/foreman/http_proxy/excon_connection_extension.rb
```diff
@@ -0,0 +1,13 @@
+Excon::Connection.class_eval do
+  include Foreman::HTTPProxy
+  alias_method :orig_request, :request
+
+  def request(params, &block)
+    if proxy_http_request?(@data[:proxy], @data[:host], @data[:schema])
+      log_proxied_request(http_proxy, @data[:host])
+      @data[:proxy] = http_proxy
+      setup_proxy
```
ares: Since the proxy can be misconfigured, it raises an Excon error (see trace below), could we somehow better rescue from this state so Foreman does not response with 500? or possibly do some validation check when the setting is saved.

```
| Excon::Error::ProxyParse (Proxy is invalid):
 |   lib/foreman/http_proxy/excon_connection_extension.rb:9:in `request'
 |   app/models/compute_resources/foreman/model/ec2.rb:65:in `regions'
 |   app/models/compute_resources/foreman/model/ec2.rb:74:in `test_connection'
 |   app/controllers/compute_resources_controller.rb:110:in `test_connection'
 |   app/controllers/concerns/application_shared.rb:15:in `set_timezone'
 |   app/models/concerns/foreman/thread_session.rb:32:in `clear_thread'
 |   app/controllers/concerns/foreman/controller/topbar_sweeper.rb:12:in `set_topbar_sweeper_controller'
 |   /home/ares/Projekty/Zdrojaky/katello/lib/katello/params_parser_wrapper.rb:12:in `call'
 |   lib/middleware/catch_json_parse_errors.rb:8:in `call'
 |   lib/middleware/tagged_logging.rb:18:in `call'
```

app/models/setting/general.rb
```diff
@@ -17,7 +17,9 @@ def self.default_settings
       self.set('db_pending_seed', N_("Should the `foreman-rake db:seed` be executed on the next run of the installer modules?"), true, N_('DB pending seed')),
       self.set('proxy_request_timeout', N_("Max timeout for REST client requests to smart-proxy"), 60, N_('Proxy request timeout')),
       self.set('login_text', N_("Text to be shown in the login-page footer"), nil, N_('Login page footer text')),
-      self.set('host_power_status', N_("Show power status on host index page. This feature calls to compute resource providers which may lead to decreased performance on host listing page."), true, N_('Show host power status'))
+      self.set('host_power_status', N_("Show power status on host index page. This feature calls to compute resource providers which may lead to decreased performance on host listing page."), true, N_('Show host power status')),
+      self.set('http_proxy', N_('Set a HTTP(S) proxy for all outgoing connections'), nil, N_('HTTP(S) Proxy')),
```
ares: I think it would deserve more detailed description, e.g. communication with smart proxy seems to be unaffected. Also an example (hostname:port) might be useful. The same for the setting below.

ares (changes requested):
in general looks good, few comments and questions left inline

I think we should document what all communication this affects, could you please put together a list? that helps me testing and we can use it in documentation too.

ares:
Also depending on which one goes in first, I think it would be great if the setting selector could actually use the HttpProxy object introduced in #4616. The selector could be just a select box, similar to how we select e.g. Global default PXEGrub template

lib/foreman/http_proxy/excon_connection_extension.rb
bastilian: The only way that comes to mind for validating the HTTP proxy setting would be to ping an arbitrary URL, which can have other reasons of failing too.

It may be more reasonable to rescue these errors and show a meaningful error message.
Agree?

bastilian:
There is now a validation for the HTTP proxy URL setting. 

I will implement proper errors and notifications for exceptions that concern http proxying and look into integrating with #4616.

lib/foreman/http_proxy/excon_connection_extension.rb
ares: yup, sounds reasonable

bastilian:
@ares Errors are now rescued and re-raised mentioning the fact it was proxied with "Proxied request failed with:".

ares:
[test foreman] [test katello]

ares:
[test foreman]

ares:
@bastilian it seems some Katello tests might be affect by this. It seems that the recorded communication does not match anymore. Could you please take a look at it?
 See http://ci.theforeman.org/job/test_develop_pr_katello/3713/database=postgresql,ruby=2.2,slave=fast/testReport/junit/(root)/Katello__CertsTest/test_verify_ueber_cert_changes/ and http://ci.theforeman.org/job/test_develop_pr_katello/3713/database=postgresql,ruby=2.2,slave=fast/testReport/junit/(root)/Katello__CertsTest/test_verify_ueber_cert_no_change/

the rest seems as current known Katello failures

bastilian:
@ares I wasn't able to reproduce the failures locally. The failing tests pass with the latest develop/master here. Mind kicking [test] again?

bastilian:
@ares nvm. rebased upon develop and pushed.

bastilian:
:green_heart: CI \o/

bastilian:
@ares These are now the real latest changes. CI will fail, with these tests and the fix needs to be done in the katello tests as they override `Settings[]` and that causes problems with the settings introduced in this branch, as they are invoked early. Sorry for the false positive, didn't realize i was working with an outdated branch.

bastilian:
Fix for tests in Katello is opened: https://github.com/Katello/katello/pull/6947

ares:
Thanks @bastilian, testing both PRs together as http://ci.theforeman.org/job/test_katello_pull_request/12521/ (I hope I got all params right :-)
edit: job is green :green_apple: 

3 comments we discussed on IRC
- [x] unification of Proxy vs proxy in settings description
- [x] use Setting.convert_array_to_regexp that we use already for interfaces identifiers
- [x] add a comments why `$!` is required instead of `rescue => e`

Otherwise I think we're ready and I'll try to sync with someone from Katello to merge their part. I suppose Katello PR can be merged even before this one.

bastilian:
@ares made the changes you requested 

> unification of Proxy vs proxy in settings description

both now are lower case

> use Setting.convert_array_to_regexp that we use already for interfaces identifiers

The `http_host_excepted_by_wildcard?` method uses  `convert_array_to_regexp` now and added tests with wildcards that have the asterisk in the middle

> add a comments why $! is required instead of rescue => e

Changed to use `rescue => e`. Couldn't recall why I used $! instead before.

ares:
Foreman tests error unrelated, `HostJSTest::NIC modal window::removing interfaces.test_0001_remove interface` on single combination

Rerun the katello tests with newer version - http://ci.theforeman.org/job/test_katello_pull_request/12531/

ares (approved):

ares:
Katello PR was merged, I suppose it should be green now [test katello] [test foreman]

ares:
@bastilian could you please rebase to make Katello tests green? there was a another failure recently...

bastilian:
@ares rebased. fingers crossed. :)

ares:
[test foreman] [test katello] [test upgrade]

ares:
[test katello]

ares:
All green, thanks @bastilian, merged as 7f8fd14d7a7fb4030bc2e6c36a297b4aa864b7ba, could you please add a documentation for how to use it to our manual at https://github.com/theforeman/theforeman.org ?

lib/foreman/http_proxy/net_http_extension.rb
```diff
@@ -0,0 +1,17 @@
...
+      log_proxied_request(http_proxy, host)
+      @proxy_address = URI.parse(http_proxy)
+      http_proxied_rescue do
+        orig_request(req, body, &block)
+      end
+    else
+      orig_request(req, body, &block)
+    end
+  end
+end
```
lzap, line 17: I have an impression that this does not actually work. Example:

```ruby
Net::HTTP.start(uri.host, uri.port) do |http|
  http.request(...)
end
```

The thing is that method `start` actually opens connection already which fails on networks which don't have route to the destination and method `request` will never be called as the application (or main request) errors out.

I am filing https://projects.theforeman.org/issues/27065 for this.

test/unit/foreman/http_proxy_test.rb
```diff
@@ -0,0 +1,159 @@
+require 'test_helper'
+
+class HTTPProxyTest < ActiveSupport::TestCase
+  class DummyHTTPAdapter
+    include Foreman::HTTPProxy
+  end
```
lzap, line 6: Sigh. Unit tests... ;-)

PR #4642, opened 2017-07-04
fixes #19050 - add Ruby on Rails 5.0 support

mmoll:
Adds configurable support for running on Ruby on Rails 5.0 or 4.2,
defaulting now to 5.0 on Ruby 2.3 or higher. Ruby versions with partial
or no support remain using 4.2 by default. Other points:

1. SETTINGS[:rails] is loaded pre-Rails boot, then post-boot the
   majority of the SETTINGS are loaded depending on the environment.
   settings.yaml.dist will allow packagers to override the default
   version of Rails loaded to match their distribution.
2. rails-controller-testing, record_tag_helper replace extracted Rails
   4.x features.
3. ParamsParser middleware is no longer loaded by default and is
   deprecated, CatchJsonParseErrors is no longer order dependent.
4. quiet_assets is not supported under Rails 5, will be replaced under
   ticket #18500, only affects development logging.


mmoll:
replaces GH-4422

mmoll:
let's see if adding psych is helping with the katello test error

mmoll:
[test katello]

mmoll:
the errors of the katello test are always different ones...

mmoll:
tbh, I think this is good to go and the katello test problems are not related to this PR

ohadlevy:
@jlsherrill @ehelms any objections?

ohadlevy (approved):
:+1: overall, I assume there are going to be minor breakage for plugins but since 4.2 is still the default for ruby version that we mostly ship i don't see this as a major concern.

ehelms:
As long as this doesn't break developer environments go for it. I don't
foresee production environments being updated for a while at this rate.

 1) Will this enforce testing of both stacks for all PRs?
 2) Since Rails 5.1 is out, is there a reason not to do latest and greatest?

On Jul 5, 2017 3:42 AM, "Ohad Levy" <notifications@github.com> wrote:

*@ohadlevy* approved this pull request.

👍 overall, I assume there are going to be minor breakage for plugins but
since 4.2 is still the default for ruby version that we mostly ship i don't
see this as a major concern.

—
You are receiving this because you were mentioned.

<https://github.com/theforeman/foreman/pull/4642#pullrequestreview-47974053>,
or mute the thread

mmoll:
> As long as this doesn't break developer environments go for it.

Even if it would, you can always set to 4.2 also on Ruby 2.3+ (and then fix the problems with 5.0).

> I don't foresee production environments being updated for a while at this rate.

I will set the version to 4.2 for the Debian/Ubuntu versions with Ruby 2.3, because of the plugin situation.

However, after 1.16.0 is out, I'd really like to have this set to 5.0.

> 1) Will this enforce testing of both stacks for all PRs?

Yes, as Ruby 2.3+ will use 5.0 per default.

> 2) Since Rails 5.1 is out, is there a reason not to do latest and greatest?

I didn't try it, but think it won't work OOTB, so let's go one step at a time. 🚶

Also, there are a few (minor) open (non-)issues around it, like GH-4288, which IMHO should get fixed first to ease future updates.

ohadlevy (approved):

ohadlevy:
thanks @mmoll and @domcleal - much appreciated.

PR #4643, opened 2017-07-05
refs #20097 - use relative path in webpack config.

ohadlevy:


dLobatog (approved):
LGTM, it doesn't make a difference on my development environment though, what's the motivation for the change?

Thanks @ohadlevy

PR #4648, opened 2017-07-05
Fixes #19828 - Fix Rails/Blank cop

tbrisker:
I disabled the default UnlessPresent value since sometimes it's clearer
to use `unless attribute.present?` instead of `if attribute.blank?`.


dLobatog:
Thanks @tbrisker !

PR #4649, opened 2017-07-05
Refs #19050 - set defaults Rails version to 4.2

mmoll:
https://github.com/rails/rails/issues/25046 is hitting plugin tests.


ohadlevy:
thanks @mmoll !

PR #4651, opened 2017-07-06
Fixes #20119 - Be more descriptive when validating hostname

boaz0:
@iNecas @tbrisker  PTAL

### DIFF
--- a/lib/net/validations.rb
+++ b/lib/net/validations.rb
@@ -9,6 +9,8 @@ module Net
     HOST_REGEXP ||= /\A(([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])\.)*([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])\z/
     MASK_REGEXP ||= /\A((255.){3}(0|128|192|224|240|248|252|254))|((255.){2}(0|128|192|224|240|248|252|254).0)|(255.(0|128|192|224|240|248|252|254)(.0){2})|((128|192|224|240|248|252|254)(.0){3})\z/
 
+    HOST_REGEXP_ERR_MSG = N_("hostname can contain only lowercase letters, numbers, dashes and dots")
+
     class Error < RuntimeError
     end
 
--- a/app/models/host/base.rb
+++ b/app/models/host/base.rb
@@ -31,7 +31,7 @@ module Host
 
     alias_attribute :hostname, :name
 
-    validates :name, :presence   => true, :uniqueness => true, :format => {:with => Net::Validations::HOST_REGEXP}
+    validates :name, :presence   => true, :uniqueness => true, :format => {:with => Net::Validations::HOST_REGEXP, :message => _(Net::Validations::HOST_REGEXP_ERR_MSG)}
     validate :host_has_required_interfaces
     validate :uniq_interfaces_identifiers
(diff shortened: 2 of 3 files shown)


app/assets/javascripts/host_edit.js
```diff
@@ -591,6 +591,27 @@ $(document).on('change', '.interface_type', function () {
   interface_type_selected(this);
 });
 
+$(document).on('change', '.host_name', function (event) {
```
iNecas: Based on the fact that the Foreman UI tries to slowly move away from the jQuery code, I think it would would save ourselves some work just by not introducing this here right now. For the user, I hope the more desriptive message, that you already did on the model, should be sufficient.

iNecas: It seems Jenkins also hits into some issues in relation to this code http://ci.theforeman.org/job/test_develop_pr_core/13520/

lib/net/validations.rb
```diff
@@ -9,6 +9,8 @@ module Validations
     HOST_REGEXP ||= /\A(([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])\.)*([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])\z/
     MASK_REGEXP ||= /\A((255.){3}(0|128|192|224|240|248|252|254))|((255.){2}(0|128|192|224|240|248|252|254).0)|(255.(0|128|192|224|240|248|252|254)(.0){2})|((128|192|224|240|248|252|254)(.0){3})\z/
 
+    HOST_REGEXP_ERR_MSG = "hostname must start with and contain alphanumerical characters and can contain dashes and dots"
```
tbrisker: Please extract the string for translation here using `N_()`. Also consider perhaps rephrasing to something a bit more concise such as "hostname may only contain letters, numbers, dashes and dots".

tbrisker (changes requested):
Just a minor comment, thanks!

lib/net/validations.rb
boaz0: done ✔️

lib/net/validations.rb
```diff
@@ -9,6 +9,8 @@ module Validations
     HOST_REGEXP ||= /\A(([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])\.)*([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])\z/
     MASK_REGEXP ||= /\A((255.){3}(0|128|192|224|240|248|252|254))|((255.){2}(0|128|192|224|240|248|252|254).0)|(255.(0|128|192|224|240|248|252|254)(.0){2})|((128|192|224|240|248|252|254)(.0){3})\z/
 
+    HOST_REGEXP_ERR_MSG = N_("hostname can contain only small letters, numbers, dashes and dots")
```
tbrisker: please use `lowercase letters` instead of `small letters`

boaz0: sure

tbrisker (approved):

tbrisker:
Good to merge once tests are green.

tbrisker:
Thanks @ripcurld0 !

boaz0:
🎊 my pleasure @tbrisker

PR #4652, opened 2017-07-07
Fixes #20241 - Fix test for mail notifications

dLobatog:
One of the tests in test/unit/tasks/reports_test is testing for
delivery.subject =~ /Configuration Management Summary/
This made sense in develop because since #16159 mail alerts have that
subject. However on the 1.15-stable branch, it still has the old
subject, "Puppet Summary".

Notice this is going ONLY to 1.15-stable.


dLobatog:
Merging to fix 1.15_stable for the release

PR #4653, opened 2017-07-09
Fixes #20248, #20327 - Respect context in CSV export

tbrisker:
Since CSV export is done using an enumerator which is executed by a
different thread to the one that creates it, we need to manually pass in
the context (user, location, organization) to the enumerator.

### DIFF
--- a/app/controllers/hosts_controller.rb
+++ b/app/controllers/hosts_controller.rb
@@ -62,7 +62,7 @@ class HostsController < ApplicationController
         render :index if title && (@title = title)
       end
       format.csv do
-        @hosts = search.includes(included_associations - [:host_statuses, :token, :compute_resource])
+        @hosts = search.preload(included_associations - [:host_statuses, :token])
         csv_response(@hosts)
       end
     end
@@ -941,7 +941,7 @@ class HostsController < ApplicationController
   end
 
   def csv_columns
-    [:name, :operatingsystem, :environment, :model, :hostgroup, :last_report]
+    [:name, :operatingsystem, :environment, :compute_resource_or_model, :hostgroup, :last_report]
   end
 
   def normalize_vm_attributes
--- a/app/helpers/hosts_and_hostgroups_helper.rb
+++ b/app/helpers/hosts_and_hostgroups_helper.rb
@@ -1,12 +1,6 @@
 module HostsAndHostgroupsHelper
   include AncestryHelper
 
-  def model_name(host)
-    name = host.try(:model)
-    name = host.compute_resource.name if host.compute_resource
-    name
-  end
-
   def domain_subnets(type)
     accessible_related_resource(@domain, :subnets, :where => {:type => type})
   end
--- a/app/models/concerns/foreman/thread_session.rb
+++ b/app/models/concerns/foreman/thread_session.rb
@@ -37,6 +37,24 @@ module Foreman
       end
     end
 
+    # This allows getting and setting all current values in case it's needed,
+    # for example to pass to an enumerator that is executed by a separate thread
+    module Context
+      def self.get
+        {
+          :user => User.current,
+          :organization => Organization.current,
+          :location => Location.current
+        }
+      end
+
+      def self.set(user: nil, organization: nil, location: nil)
+        User.current = user
+        Organization.current = organization
+        Location.current = location
+      end
+    end
+
     # include this in the User model
     module UserModel
       extend ActiveSupport::Concern
--- a/app/services/csv_exporter.rb
+++ b/app/services/csv_exporter.rb
@@ -4,9 +4,11 @@ module CsvExporter
   def self.export(resources, columns, header = nil)
     header ||= default_header(columns)
     raise ArgumentError, "Columns and header row aren't the same length" unless columns.length == header.length
+    # need to save the current context as the enumerator is executed by a separate thread
+    context = Foreman::ThreadSession::Context.get
     Enumerator.new do |csv|
+      Foreman::ThreadSession::Context.set(context)
       csv << CSV.generate_line(header)
-
       columns.map!{|c| c.to_s.split('.').map(&:to_sym)}
       resources.uncached do
         resources.reorder(nil).limit(nil).find_each do |obj|
--- a/test/controllers/hosts_controller_test.rb
+++ b/test/controllers/hosts_controller_test.rb
@@ -31,6 +31,18 @@ class HostsControllerTest < ActionController::TestCase
     assert_template 'index'
   end
 
+  test "should get csv index with data" do
+    host = FactoryGirl.create(:host, :with_hostgroup, :with_environment, :on_compute_resource, :with_reports)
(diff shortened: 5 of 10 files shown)


app/controllers/hosts_controller.rb
```diff
@@ -60,7 +60,7 @@ def index(title = nil)
         render :index if title && (@title = title)
       end
       format.csv do
-        @hosts = search.includes(included_associations - [:host_statuses, :token, :compute_resource])
+        @hosts = search.preload(included_associations - [:token])
```
tbrisker: not directly related but took advantage of this PR for a small performance improvement.

app/helpers/hosts_and_hostgroups_helper.rb
```diff
@@ -1,12 +1,6 @@
 module HostsAndHostgroupsHelper
   include AncestryHelper
 
-  def model_name(host)
-    name = host.try(:model)
-    name = host.compute_resource.name if host.compute_resource
-    name
-  end
```
tbrisker, line 8: moved this into the host object so that it can be seen in the csv.

app/services/csv_exporter.rb
```diff
@@ -4,13 +4,17 @@ module CsvExporter
...
 
-      columns.map!{|c| c.to_s.split('.').map(&:to_sym)}
-      resources.uncached do
-        resources.reorder(nil).limit(nil).find_each do |obj|
-          csv << CSV.generate_line(columns.map{|c| c.inject(obj, :try)})
+        columns.map!{|c| c.to_s.split('.').map(&:to_sym)}
+        resources.uncached do
+          resources.reorder(nil).limit(nil).find_each do |obj|
+            csv << CSV.generate_line(columns.map{|c| c.inject(obj, :try)})
+          end
```
tbrisker: Just added spaces with no change due to `User.as` block

test/controllers/hosts_controller_test.rb
```diff
@@ -31,6 +31,18 @@ def host_attributes(host)
     assert_template 'index'
   end
 
+  test "should get csv index with data" do
+    host = FactoryGirl.create(:host, :with_hostgroup, :with_environment, :on_compute_resource)
```
tbrisker: I wanted to test the last report field as well but I think it failed due to the way the report factory works.

tbrisker: nevermind, figured it out.

app/controllers/hosts_controller.rb
```diff
@@ -60,7 +60,7 @@ def index(title = nil)
         render :index if title && (@title = title)
       end
       format.csv do
-        @hosts = search.includes(included_associations - [:host_statuses, :token, :compute_resource])
```
ShimShtein, line 65: why do we need `:host_statuses` preloaded?

app/services/csv_exporter.rb
```diff
@@ -4,13 +4,17 @@ module CsvExporter
   def self.export(resources, columns, header = nil)
     header ||= default_header(columns)
     raise ArgumentError, "Columns and header row aren't the same length" unless columns.length == header.length
+    # we need to pass the current user to the enumerator which is executed by a seperate thread
```
ShimShtein: It took me some time to realize that although `resources` should already be filtered by the user (even though it's in another thread), the actual filter evaluation will be performed in this new thread, so it has to set the user correctly.

Could you please add this info to a comment?

test/factories/host_related.rb
```diff
@@ -174,8 +174,8 @@ def set_nic_attributes(host, attributes, evaluator)
       after(:create) do |host,evaluator|
         evaluator.report_count.times do |i|
           report = FactoryGirl.create(:report, :host => host, :reported_at => (evaluator.report_count - i).minutes.ago)
-          host.last_report = report.reported_at
         end
+        host.update_attribute(:last_report, host.reports.last.reported_at)
```
ShimShtein, line 178: What is the reason for this change?

ShimShtein (changes requested):
It works as expected, a few small comments though.

app/controllers/hosts_controller.rb
tbrisker: you're right, we don't need it right now.

app/services/csv_exporter.rb
tbrisker: the resources are correctly scoped actually, it's only when loading relations that this fails as there is no current user inside the enumerator.

test/factories/host_related.rb
tbrisker: Otherwise the factory generated host doesn't have the last_report attribute set correctly for some reason. also this removes the n+1 of updating the last report for each report.

ShimShtein:
:+1: ACK

tbrisker:
Looks like the other thread vars (org/loc) should also be passed into the enumerator. setting to WoC for now.

tbrisker:
Updated to also pass location and organization into the enumerator.

app/models/concerns/foreman/thread_session.rb
```diff
@@ -37,6 +37,24 @@ def clear_thread
       end
     end
 
+    # This allows getting and setting all current values in case it's needed,
```
ShimShtein, line 40: I can see that there are others that use `Thread.current` to store thread specific context.

May I suggest making the context a bit more generic? It should be pretty easy:
* instead of storing `:user`, `:organization` and `:location` directly in the thread, let's store a new hash there with these keys.
* Modify `*Model` classes to access `Thread.current[:foreman_context][:X]`
* Copying the context from one thread to another would be as simple as deep cloning the context hash.
* Either extend Thread class with `foreman_context` method that will access the new hash, or add this method to `Foreman::ThreadSession` module.

With these changes everyone will be sure that the data is passed to new threads.

ShimShtein (changes requested):
A small suggestion around thread contexts.

app/models/concerns/foreman/thread_session.rb
tbrisker: I think this refactoring, while possibly good in the long run, is outside of the scope of this issue, which is needed for 1.15.3 (currently CSV export provides incorrect/partial information). 
I wouldn't want to modify the existing methods here in a minor release, as they are very critical in many places, therefor I only added functionality.

ShimShtein: OK then. Created a refactor [issue](http://projects.theforeman.org/issues/20390) in redmine, just to make it visible.

ShimShtein:
:+1:  Tested on my machine. Filters by user, location and organization.

dLobatog (approved):
The `compute resource or model` part left me a bit perplexed. My VMs do have a model, but it's definitely not shown in the UI as the `model_name` modified in `hosts_and_hostgroups_helper` shows. That should be fixed at some point though, it's not correct that hosts can only have ONE of model and compute resource.

dLobatog:
Thanks @tbrisker @ShimShtein

PR #4654, opened 2017-07-09
refs #20193 - update to storybook that support webpack3

ohadlevy:
older version of storybook could conflict on the verison of
webpack, and when working in development environment,
the hot live reload might be using webpack 2.x vs 3.x
linked binary (node_modules/.bin/..)


tbrisker:
Thanks @ohadlevy !

PR #4655, opened 2017-07-10
Fixes #20258 - better unique permission types error

lzap:
We see regressions in rex, scap and discovery (and possibly more). Unit tests are randomly failing with:

ActiveRecord::RecordInvalid: Validation failed: Permissions Permissions must be of same resource type

This improves the error message a bit so we see more, also removes the double Permissions words in there.


iNecas:
Thanks @lzap, let's see if it helps us to better understand what the issue is

PR #4657, opened 2017-07-11
Fixes #20274 - Filter fixtures have many resource_types

dLobatog:
Filters cannot have many resource_types - each filter needs to have its
own resource type, e.g: one filter does not accommodate permissions for
"Host" and "DiscoveredHost".

This is enforced by a validator in the model, but our fixtures are able
to completely skip it. The result is that plugins have Filters with
multiple resource types.

This has not been a problem until recently, as we added a test that
checks that the db filter cache can be recreated (CacheManagerTest).
On plugins, this test fails to pass as fixtures are
wrongly created.

To reproduce, make sure your test DB is either Postgres or MySQL, add a
plugin (like foreman_discovery), and try to run the CacheManagerTest

This should unblock plugin tests - REX, Ansible, Discovery.. make the CacheManagerTest because of this issue with fixtures.

### DIFF
--- a/app/models/filter.rb
+++ b/app/models/filter.rb
@@ -201,7 +201,14 @@ class Filter < ApplicationRecord
   # if we have 0 types, empty validation will set error, we can't have more than one type
   def same_resource_type_permissions
     types = self.permissions.map(&:resource_type).uniq
-    errors.add(:permissions, _('must be of same resource type (%s)') % types.join(',')) if types.size > 1
+    errors.add(
+      :permissions,
+      _('must be of same resource type (%s) - Role (%s)') %
+      [
+        types.join(','),
+        self.role.name
+      ]
+    ) if types.size > 1
   end
 
   def not_empty_permissions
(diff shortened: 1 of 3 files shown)


app/models/filter.rb
```diff
@@ -201,7 +201,14 @@ def parenthesize(string)
   # if we have 0 types, empty validation will set error, we can't have more than one type
   def same_resource_type_permissions
     types = self.permissions.map(&:resource_type).uniq
-    errors.add(:permissions, _('must be of same resource type (%s)') % types.join(',')) if types.size > 1
+    errors.add(
```
dLobatog, line 204: This is not strictly needed but it makes debugging easier - if a filter fails to save, I want to know which Role is it associated with (it's easier to reason about, I can tell where a role is coming, like "Discovery Reader", but filters have no name)

lzap: Absolutely, we have so many of these non-informative error messages in our codebase. :+1:

lzap:
Wow thanks a lot. Executed tests against discovery 5 times locally, all green.

Any objections @ares ?

lzap (approved):

rabajaj0509:
@dLobatog, this pr was really informative, thanks 👍

lzap:
Thanks.

PR #4661, opened 2017-07-12
Fixes #19947 - Fix creating bookmarks for global params

xprazak2:


dLobatog (changes requested):
Thanks, this looks good @xprazak2 - could you add a functional test that creates a bookmark for global params using this controller? I plan to include this fix on 1.15.3 and patch release changes should come with tests

xprazak2:
I added a test

dLobatog (approved):

dLobatog:
Thanks @xprazak2 !

PR #4662, opened 2017-07-12
Fixes #20209 - Remove ovirt image from compute profile

orrabin:
The image was disabled in compute profile.
Compute profile does not include the operating system and the possible list of images for a host depends on the operating system so it might not belong in compute profile anyway.
As a result I removed the image from compute profile.


dLobatog (approved):
Thanks @orrabin

dLobatog:
Cherry-picked to 1.15-stable as 113ef7d58cfea2fb2807996503e28259046cd023

PR #4664, opened 2017-07-12
Fixes #17360 - Prevent % string interpolation in scoped_search

kgaikwad:


tbrisker:
@kgaikwad looks good, could you just rephrase the commit message please? something like "Prevent % string interpolation in scoped_search"?

kgaikwad:
@tbrisker ,
Thank you, modified commit message as per you suggested.

test/models/fact_value_test.rb
```diff
@@ -67,6 +67,11 @@ def setup
     refute_empty results
   end
 
+  test 'should return search results for ILIKE operator on host' do
+    results = FactValue.search_for("host ~ abc")
+    assert_empty results
+  end
```
tbrisker: Could you please add another test to make sure that `~` returns results when there are some?

tbrisker (changes requested):

test/models/fact_value_test.rb
kgaikwad: @tbrisker ,
Done.
New test added with `~` operator on `host.hostgroup`.

test/models/fact_value_test.rb
```diff
@@ -67,6 +67,21 @@ def setup
...
+
+  test 'should return search results for ILIKE operator on hostgroup' do
+    host = FactoryGirl.create(:host, :hostgroup => FactoryGirl.create(:hostgroup, :name => 'samplehgp'))
+    FactoryGirl.create(:fact_value, :value => '2.6.9',:host => host,
+                       :fact_name => FactoryGirl.create(:fact_name, :name => 'kernelversion'))
+    search_text = host.hostgroup.to_label
+    results = FactValue.search_for("host.hostgroup ~ #{search_text}")
```
tbrisker: I would try searching on just a part of the hostgroup name to make sure the ILIKE works correctly, something like `ampl`.

test/models/fact_value_test.rb
```diff
@@ -67,6 +67,21 @@ def setup
...
+  test 'should return empty search results for ILIKE operator on host' do
+    results = FactValue.search_for("host ~ abc")
+    assert_empty results
+  end
+
+  test 'should return search results for ILIKE operator on hostgroup' do
+    host = FactoryGirl.create(:host, :hostgroup => FactoryGirl.create(:hostgroup, :name => 'samplehgp'))
+    FactoryGirl.create(:fact_value, :value => '2.6.9',:host => host,
+                       :fact_name => FactoryGirl.create(:fact_name, :name => 'kernelversion'))
+    search_text = host.hostgroup.to_label
```
tbrisker: you already know the hostgroup name (`samplehgp`), no need to get it again.

test/models/fact_value_test.rb
```diff
@@ -67,6 +67,21 @@ def setup
...
+    results = FactValue.search_for("host.hostgroup ~ #{search_text}")
+    assert_equal 1, results.count
+    refute_empty results
```
tbrisker: this assertion is not needed, if `results.count == 1` then it isn't empty.

tbrisker (changes requested):
Thanks @kgaikwad, just a couple of minor comments inline.

test/models/fact_value_test.rb
kgaikwad: Oops my bad.. Sorry again @tbrisker !!
Instead of pushing code in hurry, I need to learn code cross-checking twice before I push it :-).
I will update PR soon. Thank you.

kgaikwad:
@tbrisker ,
Done with the changes.
Please let me know if any other changes required.

tbrisker (approved):

tbrisker:
Thanks @kgaikwad !

PR #4665, opened 2017-07-12
Fixes #20284 - Host saves compute profile storage

orrabin:


iNecas (changes requested):
It seems this is just partly the problem. The real issues seems to be in the fact, that we're leaving the id there even when we unselect the template, so after this patch, although it's not 7 (but some uuid), it still prevents from adding the storage to the VM.

orrabin:
@iNecas thanks, this is because when the template is not selected: https://github.com/theforeman/foreman/blob/develop/app/assets/javascripts/compute_resource.js#L69 does not enter the if and doesn't refresh the storage information.

Also, forcing the code to get there I saw the remove doesn't work: https://github.com/theforeman/foreman/blob/develop/app/assets/javascripts/compute_resource.js#L81

I tested on develop to make sure it's not because I removed the 7 so there are still a few issues to handle here with the js.

app/assets/javascripts/compute_resource.js
```diff
@@ -66,7 +66,7 @@ function testConnection(item) {
 
 function ovirt_templateSelected(item){
   var template = $(item).val();
-  if (template && !item.disabled) {
+  if (!item.disabled) {
```
orrabin, line 69: With no template selected the default template which does have memory and cpu count will return in the result (in the ajax).

The other option is ignoring the blank template and in case of no template removing the values of all the fields in the virtual machine tab.

app/assets/javascripts/compute_resource.js
```diff
@@ -78,7 +78,7 @@ function ovirt_templateSelected(item){
         $('[id$=_cores]').val(result.cores);
         $('#network_interfaces').children('.fields').remove();
         $.each(result.interfaces, function() {add_network_interface(this);});
-        $('#storage_volumes').children('.fields').remove();
+        $('#storage_volumes .children_fields >.fields').remove();
```
orrabin, line 81: This will also remove volumes created manually but otherwise it's a lot more complicated to find if the volume belonged to a template or not and to know which to remove.

Also, it seems the behavior was always to delete all the volumes, the selector just changed at some point and stopped working.

app/assets/javascripts/compute_resource.js
iNecas: I'm ok with this approach

iNecas (approved):

iNecas:
[test foreman]

iNecas:
Thanks @orrabin

ares:
Thanks @iNecas for merging, please don't forget to set the release field in redmine.

PR #4667, opened 2017-07-13
Fixes #20295 - Sanitize webpack_directories json

xprazak2:


ohadlevy:
is the warning messages are STDOUT or STDERR? if its ERR you might just ignore STDERR ?

I didnt try, but maybe:
require('child_process').execSync('command', {stdio:[0,1]});

xprazak2:
I haven't yet found where it is coming from, but ignoring STDERR does not help.

tstrachota:
@xprazak2 the output comes from here https://github.com/theforeman/foreman_openscap/blob/master/foreman_openscap.gemspec when the gem is loaded. I think it should be printed to stderr but it's a different issue.

For the means of sanitization to avoid such failures in future it should be enough to keep `puts` in `plugin_webpack_directories.rb` and always use only the last printed line, because we're sure the json will be printed last.

I wonder if it makes sense to print the cropped lines somewhere, because if they appear it's most likely an error that we shouldn't hide.

xprazak2:
Ahh, that was probably the one place where I did not look, thanks. I'll raise an issue for warnings to go to stderr. I swapped `puts` for `print` because it does not add the newline character, so I can grab the last line directly. With `puts`, the last item is actually an empty string when I do the split:

```js
"some debug output\n{json: true}\n".split("\n") // ["some debug output", "{json: true}", ""]
```

but I can keep `puts` and handle it on js side.

We are not hiding the errors, they still appear when rails server starts.

xprazak2:
I put the `puts` back, I also ignore stderr of the script as well as grab only the last line which is JSON we want if there are multiple lines.

tstrachota:
A comment in the sanitize function explaining why we're actually doing this would be nice. Otherwise :+1: the code looks good and it works nicely!

config/webpack.config.js
```diff
@@ -16,7 +16,18 @@ var production =
   process.env.RAILS_ENV === 'production' ||
   process.env.NODE_ENV === 'production';
 
-var plugins = JSON.parse(execSync(path.join(__dirname, '../script/plugin_webpack_directories.rb')));
+var sanitizeWebpackDirs = function (pluginDirs) {
```
ohadlevy: how about syntax like:
```js
var sanitizeWebpackDirs = function (pluginDirs) {
  var splitDirs = pluginDirs.toString().split('\n').reverse();

  return splitDirs.length > 2 ? splitDirs[1] : pluginDirs;
};

var webpackDirs = execSync(path.join(__dirname, '../script/plugin_webpack_directories.rb'), {
  stdio: ['pipe', 'pipe', 'ignore']
});
var plugins = JSON.parse(sanitizeWebpackDirs(webpackDirs));
```

xprazak2:
I added a comment explaining what the function does and improved syntax.

dLobatog (approved):

dLobatog (changes requested):
@xprazak2 Looks like you need to rebase to make sure the ruby2ruby changes are in, tests should pass then 😄

PR #4668, opened 2017-07-13
Fixes #17811 - Improve error message LDAP fail

dhlavac:
LDAP authentication failures are not reported in an understandable way down the stack


app/controllers/api/base_controller.rb
```diff
@@ -16,8 +16,15 @@ class BaseController < ActionController::Base
     after_action :log_response_body
 
     rescue_from StandardError do |error|
-      Foreman::Logging.exception("Action failed", error)
-      render_error 'standard_error', :status => :internal_server_error, :locals => { :exception => error }
+      case error
+        when Net::LDAP::Error
```
ares: I think this would be better to have as new `rescue_from` block

app/controllers/api/base_controller.rb
```diff
@@ -16,8 +16,15 @@ class BaseController < ActionController::Base
...
+          message = _("Error during LDAP connection :  #{error}")
```
ares: since we caught this here, we have no information to which LDAP server we were trying to connect, maybe rescue this exception earlier where the ldap connection is estabilished and render a new custom exception there with message like `_("Error connection to %{url} LDAP server during authentication: %{error}" % {:url => ..., :error => ...}`

ares (changes requested):

app/controllers/api/base_controller.rb
dhlavac: Yes I know but when I try catch `Net::LDAP::Error`  it doesn't work

dhlavac:
@ares Changes done please look at that

lib/foreman/exception.rb
```diff
@@ -50,6 +50,9 @@ class MaintenanceException < Foreman::Exception
   class PermissionMissingException < Foreman::Exception
   end
 
+  class LdapException < Foreman::Exception
```
ares: Could you please use `Foreman::WrappedException` as a parent for this class? this allows us also to keep the original error and its trace. This means we should move the definition from app/services/foreman/wrapped_exception.rb to this file (or the other way around).

app/models/auth_sources/auth_source_ldap.rb
```diff
@@ -75,6 +75,8 @@ def authenticate(login, password)
     logger.debug "Retrieved LDAP Attributes for #{login}: #{attrs}"
 
     attrs
+  rescue Net::LDAP::Error => error
+    raise ::Foreman::LdapException.new(N_("Error connection to %{name} LDAP server with url: %{url} during authentication: %{error}" % {:url => self.host, :name => self.name, :error => error}))
```
ares: if ::Foreman::LdapException inherits from wrapped exception, the constructor would accept extra argument - the original exception, `error` in this case

the line would look similar to
`raise ::Foreman::LdapException.new(error, N_("Error while connecting to '%{name}' LDAP server at '%{url}' during authentication" % {:url => self.host, :name => self.name}))`

please note I also update the exception text, it's good to wrap values in some delimiter chars.

ares (changes requested):
Thanks, it looks better now. I noticed that `net-ldap` 0.16 replaced `Net::LDAP::LdapError` with `Net::LDAP:Error` which is the source of these problems. Could you please find all places in Foreman where we rescue from  `Net::LDAP::LdapError` and add `Net::LDAP::Error` like you did for authentication?

app/models/auth_sources/auth_source_ldap.rb
```diff
@@ -75,6 +75,8 @@ def authenticate(login, password)
...
   rescue Net::LDAP::LdapError => error
```
ares: this is how it worked with older `net-ldap` version, you improved the error message and if we keep this for compatibility with older net-ldap, I think we can combine both exceptions here with same rescue block

long story short:
```ruby
rescue Net::LDAP::Error, Net::LDAP::LdapError => error
  raise ::Foreman::LdapException.new(error, N_("Error while connecting to '%{name}' LDAP server at '%{url}' during authentication" % {:url => self.host, :name => self.name}))
end
```

should do the trick

dhlavac:
@ares Requested changes done

ares (changes requested):
Works just fine, I think it's ready to merge, could you please just rebase onto latest develop to avoid failures in jenkins?

dhlavac:
@ares  rebased

ares (approved):

ares:
Thanks @dhlavac, merged as 01b7dbb57788dd43032ec707bf02d8771bf2d921

PR #4669, opened 2017-07-13
Refs #20271 - fix templates for new safemode

ares:


db/migrate/20170713173801_extract_to_proc_calls_in_templates.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+  end
+
+  def down
+    # audit can be used for reverting unwanted changes
+    say "skipping parameters macros migration"
+  end
+
+  def convert(content)
+    content.gsub(/<%(.*?)\(?&:([a-zA-Z0-9!?_]+)\)?(.*)%>/, '<%\1{ |item| item.\2 }\3%>')
```
tbrisker, line 48:
```
<%= @host.name %>
in Ruby, you can use syntax like arr.map(&:a) to call a function "a" on an array.
<% end %>
```
although it is an edge case, might want to check there isn't a closing `%> ` between the opening `<%` and the symbol to proc.

ares: Your example works fine because of new lines, the problem would happen if that's on single line, but I'm not sure how to do that with regexp. Any hint?

lzap: You can use lookahead syntax which is supported in Ruby 2.0+. We had an example in our old subnet parser which was recently replaced by new one: https://github.com/lzap/smart-proxy/blob/ce684575ba8777ac4edb68c6b0dfb40fd69f3e3d/modules/dhcp_isc/isc_file_parser.rb#L61-L73

This will work, but do we really need to change all templates? Since we have a locking mechanism now, we can simple deploy fixed ones and ask users to modify their clones in upgrade notes. At least we should do backups, this can completely ruin a production instance in case there is a bug. I feel like letting it to error out is much safer.

ares: I'm afraid it might be hard for users who are not fluent with Ruby to fix templates that they mostly just copied from built-in templates. The production would be ruined anyway, safemode will break such templates. Perhaps we could add an audit record manually (we still need to skip validations in migrations)

tbrisker: Another case that isn't caught well is when there are multiple calls chained, e.g. `@host.interfaces.map(&:name).each(&:downcase)` etc. 
I tend to agree it's quite complex to correctly catch all of the edge cases, might be better to put a  release note explaining clearly how to change them then to modify all of the templates without warning.

ares: perhaps we should provide a script to update templates which would give user a preview before they apply the changes. I'll ask foreman-maintain guys if that would be a good fit, thanks for the feedback, I've updated manual issues in theforeman.org repo and I'm closing this PR

PR #4670, opened 2017-07-13
Fixes #18826 - handled URL generation exception

kgaikwad:
When user without email logged in and try to redirect to `Katello routes`
then it raises exception `UrlGenerationError` instead of redirecting
to user edit path.
With this commit, it will redirect to user edit path as like `Foreman routes`.

### DIFF
--- a/test/models/user_test.rb
+++ b/test/models/user_test.rb
@@ -21,6 +21,7 @@ class UserTest < ActiveSupport::TestCase
   should allow_value('').for(:mail).on(:create)
   should allow_value('é ô à', "C_r'a-z.y( )<,Na=me;>").for(:firstname)
   should allow_value('é ô à', "C_r'a-z.y( )<,Na=me;>").for(:lastname)
+  should allow_value('A$APRocky').for(:login)
   should_not allow_value('The Riddle?').for(:firstname)
   should_not allow_value("it's the JOKER$$$").for(:lastname)
   # Associations


kgaikwad:
[test]

kgaikwad:
I moved [ticket #18826](http://projects.theforeman.org/issues/18826) from Katello to the Foreman project.

xprazak2:
It fixes the error for me, the user is redirected back to edit page.

daviddavis:
Any idea why the exception is being raised in the first place?

kgaikwad:
@daviddavis, @xprazak2

Not only for foreman actions but also for katello,  before_action filter `require_mail` gets executed which is defined in foreman project. 
If current user with no email found, then further this [line](https://github.com/kgaikwad/foreman/blob/develop/app/controllers/application_controller.rb#L82) of code gets executed.

For `edit_user_path(:id => User.current)`, it leads to `ActionController::UrlGenerationError: No route matches` exception. Screenshot is attached in issue tracker for reference.
If we replace named route i.e `edit_user_path` with route string `"/users/#{User.current.id}/edit"`, it works fine.
So either we have to rescue this exception or we can use route string instead of named route.

Please let me know your suggestions.

daviddavis:
Yea, my question is why does edit_user_path raise a UrlGenerationError. Any idea why that's happening?

kgaikwad:
@daviddavis ,

Routes related to `User` resource are defined in Foreman not in Katello. 
As per my understanding this is the reason that Katello engine not recognizes this named route `edit_user_path` and it gives `UrlGenerationError` exception for `edit_user_path`.

If we provide route string `/users/#{User.current.id}/edit` then it simply redirects to this url.

daviddavis:
I see. Any idea of how `before_filter :require_login` works when a user visits a katello page and is not logged in? Is that using a string route? If so, then a string route sounds like the best option to me. I would worry that `UrlGenerationError` would mask other problems elsewhere in the code.

kgaikwad:
@daviddavis ,
Thanks for pointing me towards `:require_login`. 
I have checked it before but I don't know how I skipped `SSO` part. 
Inside `before_filter :require_login`, to get value of login_url in `SSO::Apache` class, main_app is used i.e `main_app.login_users_path`.
Note that main_app is nothing but a `ActionDispatch::Routing::RoutesProxy`.

I have tried for getting `edit_users_path` like `main_app.edit_users_path`,
its working without exception. 
Please correct me if my understanding is wrong somewhere.

daviddavis:
@kgaikwad awesome! looks good to me. Thanks for looking into it.

kgaikwad:
@daviddavis ,
Any idea why this PR is showing in `Closed` state.
I haven't did anything, just pulled out latest changes from `develop` branch by rebasing.

kgaikwad:
@daviddavis ,
Someone is committed the same change & merged it to develop branch few days back.  
[Commit_id](https://github.com/theforeman/foreman/commit/f92cc17393ded44208623070889f6041b8347b09) for reference.
That's why it is showing as closed.

daviddavis:
Ah good catch @kgaikwad. Sorry someone beat you to it 😢

PR #4671, opened 2017-07-14
Fixes #20306 - allow the same key for two users

ares:
If that's accepted, I suppose this should go into 1.15 where we introduced ssh keys.


timogoebel (approved):
LGTM.

PR #4673, opened 2017-07-14
Fixes #20287 - fix plugin assets precompilation

mmoll:
The error was only visible with RAILS_ENV=production


mmoll:
@ehelms @waldenraines could you also have a look here?

iNecas (approved):
I can confirm this works for compiling assets for foreman-tasks and foreman_remote_execution

iNecas:
@dLobatog @ehelms @waldenraines anything against merging this? Plugin assets compilation just doesn't work currently without this fix and I can confirm it works with this patch.

dLobatog (approved):
Tested the difference with and without the patch - works just fine, thanks @mmoll & @iNecas

PR #4674, opened 2017-07-16
Fixes #19808 - Remove useless assignments

tbrisker:


lib/tasks/config.rake
```diff
@@ -101,12 +101,12 @@ BANNER
         value = @key_values[key]
         setting = Setting.find_by_name(key)
         if value == :unset
-          value = nil
+          setting.value = nil
```
tbrisker, line 104: I think this is actually a bug - passing `:unset` had no effect.

app/models/provisioning_template.rb
```diff
@@ -140,7 +140,7 @@ def self.build_pxe_default(renderer)
           Foreman::Logging.exception("Cannot render '#{global_template_name}'", exception)
           error_msgs << "#{exception.message} (#{kind})"
         end
-        return [422, error_msg] unless error_msg.empty?
+        return [:unprocessable_entity, error_msgs.join(', ')] unless error_msgs.empty?
```
tbrisker, line 147: @lzap - Correct me if I understood incorrectly, but looks like `error_msg` was always empty here because it is only set in the other branch of the `if` statement. tbh i'd rather put the `return` inside the `rescue` block that needs to return rather then after it, kind of unexpected in the middle of this `else` block.

app/models/provisioning_template.rb
```diff
@@ -124,13 +124,13 @@ def self.global_template_name_for(kind, renderer)
   end
 
   def self.build_pxe_default(renderer)
-    return [422, _("No TFTP proxies defined, can't continue")] if (proxies = SmartProxy.with_features("TFTP")).empty?
+    return [:unprocessable_entity, _("No TFTP proxies defined, can't continue")] if (proxies = SmartProxy.with_features("TFTP")).empty?
```
tbrisker, line 131: changed all numeric http codes to rails symbols while i'm touching this function.

app/models/provisioning_template.rb
lzap: Yes, legacy code. This will work just fine. Thanks.

lzap:
Of course, tests are failing because they have no longer valid inputs :-)

lzap:
Nice one, two bugs solved. Please review the tests, there are likely incorrect combos now.

dLobatog (changes requested):
Setting to WoC as tests are failing

tbrisker:
@lzap added PXEGrub{,2} default template to the factory, hopefully test will pass now

test/fixtures/templates.yml
```diff
@@ -86,6 +86,22 @@ pxegrub:
...
 
+pxegrub_default:
+   name: PXEGrub global default
+   template: "FOO"
+   template_kind: pxegrub
+   operatingsystems: centos5_3, redhat
+   locked: false
+   type: ProvisioningTemplate
+
+pxegrub2_default:
```
dLobatog, line 97: Since there are now two default templates (for two different `TemplateKind::PXE`), calling `build_pxe_default` will attempt to render two instead of one. https://github.com/theforeman/foreman/blob/develop/test/models/provisioning_template_test.rb#L219 the expectation should be expected twice

dLobatog (changes requested):
@tbrisker Tests still failing but the fix is quite simple, check out the inline comment

test/fixtures/templates.yml
tbrisker: updated. :crossed_fingers:

tbrisker:
test failures seem unrelated, looks like some new gem broke everything. investigating.

tbrisker:
failing tests are #4678

dLobatog:
[test katello] 
[test foreman]

dLobatog (approved):

dLobatog (changes requested):
@tbrisker Please rebase to include your changes to ruby2ruby 😄

tbrisker:
done

dLobatog (approved):

dLobatog:
Thanks!

PR #4675, opened 2017-07-16
Fixes #20312: Use bundler_ext in production to load webpack plugin

ehelms:


ohadlevy:
@ehelms thanks, does it make sense to add a test to see this file return json correctly?

ehelms:
@ohadlevy I added one, lets see what Jenkins does. Testing the bundler_ext route is not really possible since it requires installing a gem that also changes multiple other paths.

test/unit/script/plugin_webpack_directories_test.rb
```diff
@@ -0,0 +1,16 @@
+require 'minitest/autorun'
+require 'json'
+
+class AssociationAuthorizerTest < Minitest::Test
+  def setup
+    @root_dir = File.expand_path('../../../..', __FILE__)
```
ohadlevy: consider using File.join instead (if we ever run on non unix platform).

dLobatog (changes requested):
Setting to WoC based on Hound, `File.join` comment, commit summary length > 65

dLobatog (approved):

dLobatog (approved):
Ready to merge if tests pass

test/unit/script/plugin_webpack_directories_test.rb
```diff
@@ -0,0 +1,15 @@
...
+require 'json'
+
+class PluginWebpackDirectoriesTest < Minitest::Test
+  def setup
+    @root_dir = File.join('../../../..', __FILE__)
+    @script = File.join("#{@root_dir}/script/plugin_webpack_directories.rb", __FILE__)
+  end
+
+  def test_bundler
+    plugin_webpack = `#{@script}`
```
dLobatog: Seems that this is not producing any output in Jenkins, as it's nil in the test that is failing

dLobatog (changes requested):
@ehelms Please check out the test, thanks !

dLobatog (changes requested):
@ehelms Please rebase now that the commit by @ohadlevy was merged, thanks!

ehelms:
Rebased and should pass tests this time around

dLobatog (changes requested):
@ehelms Still failing :/

test/unit/script/plugin_webpack_directories_test.rb
```diff
@@ -0,0 +1,16 @@
+require 'minitest/autorun'
+require 'json'
+
+class PluginWebpackDirectoriesTest < Minitest::Test
+  def setup
+    @root_dir = File.join('../../../..', __FILE__)
```
ohadlevy: nitpicking, but the whole value of using file.join was to not include / or \ so ['..','..','..'....]
also, shouldnt it be 3 times vs 4 ?

if you want to be sure you can probably assert that @root_dir exists is rails root too

ehelms:
@ohadlevy @dLobatog This is now passing tests. Please double check all your requests.

ohadlevy:
overall :+1: but if we add more logic I would prefer to start extracting methods and / or classes etc ...

dLobatog (approved):

dLobatog:
Thanks @ehelms & @ohadlevy !

ares:
I set 1.16 in redmine

PR #4509, opened 2017-05-07
Fixes #19589 - Added plugin migration generator

ShimShtein:
Now plugin creators would be able to generate migrations by invoking

``` sh
rails g plugin:migration <migration name and options> --plugin_name=my_plugin
```

Additional options are available, use `--help` key to display all options.

PR #4676, opened 2017-07-17
Fixes #4509 - VMWare scsi react

ohadlevy:
This pr replaces #4366 as @matanwerbner no longer actively maintain it.

its rebased and minor changes on top of his original PR.


timogoebel:
@ohadlevy:
I found one small issue while testing this, can be fixed with this:

```diff
diff --git a/app/models/compute_resources/foreman/model/vmware.rb b/app/models/compute_resources/foreman/model/vmware.rb
index 996c22ba7..58702ae95 100644
--- a/app/models/compute_resources/foreman/model/vmware.rb
+++ b/app/models/compute_resources/foreman/model/vmware.rb
@@ -29,7 +29,7 @@ module Foreman::Model
     end

     def capabilities
-      [:build, :image, :new_volume]
+      [:build, :image]
     end

     def vms(opts = {})
```

Another issue looks more complex and I'm unsure if this is even related to this PR: When you create a new host and in the form select the Compute Resource fist and then the Organization the scsi part is empty by mistake. Any idea, what could cause this?

ohadlevy:
@timogoebel fixed the first issue:

> Another issue looks more complex and I'm unsure if this is even related to this PR: When you create a new host and in the form select the Compute Resource fist and then the Organization the scsi part is empty by mistake. Any idea, what could cause this?

I assume this happens because the component is already mounted and can't be remounted - this is probably true to most of the entry level mount points... did you get any browser error? I wasn't able to reproduce as I don't have a full vmware infra atm.

timogoebel:
@ohadlevy : I get the following error:
image

ohadlevy:
@timogoebel if you look in the react browser ext (if you have it) can you check if the key is empty of the controller keys are duplicated (e.g. the same value exists twice?)

ohadlevy:
@timogoebel do you have any suggestions for the failing tests?

ohadlevy:
@timogoebel I've noticed that the storybook patches were missing, added those as well.

timogoebel:
@ohadlevy : This should fix the failing testes:

```diff
diff --git a/test/models/compute_resources/vmware_test.rb b/test/models/compute_resources/vmware_test.rb
index 6c04329d5..68b219cab 100644
--- a/test/models/compute_resources/vmware_test.rb
+++ b/test/models/compute_resources/vmware_test.rb
@@ -268,6 +268,10 @@ class VmwareTest < ActiveSupport::TestCase
       ]
       @vm.stubs(:volumes).returns(@volumes)

+      scsi_controller1 = mock('scsi_controller1')
+      scsi_controller1.stubs(:attributes).returns({:type=>"VirtualLsiLogicController", :shared_bus=>"noSharing", :unit_number=>7, :key=>1000})
+      @vm.stubs(:scsi_controllers).returns([scsi_controller1])
+
       @networks = [
         OpenStruct.new(:id => 'dvportgroup-123456', :name => 'Testnetwork')
       ]
@@ -281,7 +285,14 @@ class VmwareTest < ActiveSupport::TestCase
           "0" => { :vol => 1, :size_gb => 4 },
           "1" => { :vol => 2, :size_gb => 4 }
         },
-        :interfaces_attributes => {}
+        :interfaces_attributes => {},
+        :scsi_controllers => [
+          {
+            :type => "VirtualLsiLogicController",
+            :shared_bus => "noSharing",
+            :unit_number => 7,
+            :key => 1000
+        }]
       }
       attrs = @cr.vm_compute_attributes_for('abc')

@@ -309,7 +320,14 @@ class VmwareTest < ActiveSupport::TestCase
           "0" => { :vol => 1, :size_gb => 4 },
           "1" => { :vol => 2, :size_gb => 4 }
         },
-        :interfaces_attributes => {"0"=>{:compute_attributes=>{:network=>"Testnetwork", :type=>"VirtualVmxnet3"}, :mac=>"00:50:56:84:f1:b1"}}
+        :interfaces_attributes => {"0"=>{:compute_attributes=>{:network=>"Testnetwork", :type=>"VirtualVmxnet3"}, :mac=>"00:50:56:84:f1:b1"}},
+        :scsi_controllers => [
+          {
+            :type => "VirtualLsiLogicController",
+            :shared_bus => "noSharing",
+            :unit_number => 7,
+            :key => 1000
+        }]
       }
       attrs = @cr.vm_compute_attributes_for('abc')
```

timogoebel:
@ohadlevy: A small update:

1. The issue when you first select the compute resource and then the organization and it will break the form is still present. :-(
2. There is still one failing test. However I can't reproduce it locally. :-( I'll try my best to figure out what is going on.

ohadlevy:
@timogoebel for 1. can you double check if the empty method works as expected? (e.g. if it really clears the div for you or not). (assuming you rebased etc)

timogoebel:
@ohadlevy: I don't think, the method works as expected. I suspect, that `reactNode` refers to the "new component" and not to the old one.
I tried debugging this like this:
```javascript
     console.log('Before', reactNode.innerHTML);
     reactNode.innerHTML = '';
     console.log('After', reactNode.innerHTML);
```
This gives me:

```
Before
After
```

So, the before is already empty.

ohadlevy:
@timogoebel I believe I fixed the issue by fully unmounting the component (see the 3rd commit) can you verify it solves your issue?

ohadlevy:
@timogoebel also note, that as a side effect, if the data is already in the store, than it will preserve the behavior (just like the rest of the form - e.g. by changing the number of cpu's)

app/controllers/compute_attributes_controller.rb
```diff
@@ -1,12 +1,14 @@
...
   include Foreman::Controller::Parameters::ComputeAttribute
+  include Foreman::Controller::NormalizeScsiAttributes
 
   def new
     @set = ComputeAttribute.new(:compute_profile_id => params[:compute_profile_id].to_i,
                                 :compute_resource_id => params[:compute_resource_id].to_i)
   end
 
   def create
+    normalize_scsi_attributes(compute_profile_params["vm_attrs"]) if compute_attribute_params["vm_attrs"] && compute_attribute_params["vm_attrs"]["scsi_controllers"]
```
ohadlevy: this one is compute_profile_params but the next one is compute_attribute_params. which one is the right one?

ohadlevy: once you confirm, I'll extract to a before_action.

timogoebel:
@ohadlevy : This should fix the final test:

```diff
diff --git a/test/models/compute_resources/vmware_test.rb b/test/models/compute_resources/vmware_test.rb
index 2a9845108..bcfd75ad2 100644
--- a/test/models/compute_resources/vmware_test.rb
+++ b/test/models/compute_resources/vmware_test.rb
@@ -152,6 +152,7 @@ class VmwareTest < ActiveSupport::TestCase
     end

     test "converts scsi_controller_type to hash" do
+      Foreman::Deprecation.expects(:deprecation_warning).once
       attrs_in = HashWithIndifferentAccess.new("cpus"=>"1", "scsi_controller_type"=>"ParaVirtualSCSIController", "interfaces_attributes"=>{}, "volumes_attributes"=>{})
       attrs_out = {:cpus=>"1", :interfaces=>[], :volumes=>[], :scsi_controller=>{:type=>"ParaVirtualSCSIController"}}
       assert_equal attrs_out, @cr.parse_args(attrs_in)
```

timogoebel (approved):
Test failures are unrelated. Manual testing works well. We're good to go!

Thanks @matanwerbner and @ohadlevy!

ares:
Thanks @timogoebel for the hard work and merging, please don't forget to set the release field in redmine after mering a PR :-) I set it to 1.16.

ohadlevy:
also thanks @shlomizadok for the original PR :)

shlomizadok:
🎉  So happy it finally got in !!
Kudos @timogoebel @ohadlevy et. all!

PR #4677, opened 2017-07-17
Refs #17525 - verify domain dns id also on update

ares:
While still not systematic solution to all possible relations in Foreman, this tries to close the gap from https://github.com/theforeman/foreman/pull/4120, except for relying on `dns_id` being the only foreign key, we proactively check for the value. The patch is still small so I think could be cherry-picked to 1.15.3 since the original fix was introduced in 1.15 branch.

### DIFF
--- a/app/controllers/api/v2/domains_controller.rb
+++ b/app/controllers/api/v2/domains_controller.rb
@@ -62,10 +62,12 @@ module Api
 
       def create
         @domain = Domain.new(domain_params)
-        process_response @domain.save
-      rescue ActiveRecord::InvalidForeignKey
-        @domain.errors.add(:dns_id, _('Invalid smart-proxy id'))
-        process_resource_error
+        if verify_proxy_id(domain_params[:dns_id])
+          process_response @domain.save
+        else
+          @domain.errors.add(:dns_id, _('Invalid smart-proxy id'))
+          process_resource_error
+        end
       end
 
       api :PUT, "/domains/:id/", N_("Update a domain")
@@ -73,7 +75,12 @@ module Api
       param_group :domain
 
       def update
-        process_response @domain.update_attributes(domain_params)
+        if verify_proxy_id(domain_params[:dns_id])
+          process_response @domain.update_attributes(domain_params)
+        else
+          @domain.errors.add(:dns_id, _('Invalid smart-proxy id'))
+          process_resource_error
+        end
       end
 
       api :DELETE, "/domains/:id/", N_("Delete a domain")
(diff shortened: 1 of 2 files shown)


ares:
if it's not accepted to 1.15, please let me know so I can create separate issue to track it against different version

lzap (approved):
Looks good, but I will let the 1.15 decision on @dLobatog

app/controllers/api/v2/domains_controller.rb
```diff
@@ -88,6 +95,10 @@ def destroy
       def allowed_nested_id
         %w(subnet_id location_id organization_id)
       end
+
+      def verify_proxy_id(id)
+        id.nil? || SmartProxy.authorized(:view_smart_proxies).find_by_id(id).present?
```
dLobatog, line 100: `id.blank?` checks against empty strings too

dLobatog (changes requested):
👍 , ack this for 1.15 , no problem. 

@ares Please add one or two tests for the new function demonstrating it's behavior to make sure it doesn't break, to keep patch versions as stable as possible I'm following http://projects.theforeman.org/projects/foreman/wiki/Release_Management#Issue-tracking and requiring tests for any change meant to be cherry-picked (this PR fails number 7 in that list)

ares:
Fair enough, I added the tests (similar to existing for creation). Btw other tests revealed some issues with my initial implementation, so the code path is decently covered.

app/controllers/api/v2/domains_controller.rb
ares: But we want only `nil` here, we don't want to verify dns_id if it was not passed as attribute (`domain_parameters[:dns_id]` is `nil`), if user sent empty string we want to trigger the verification and try load the proxy so it fails and tells user that empty string is invalid value.

lzap:
There are failures.

dLobatog (approved):
@ares Could you rebase to bring in the changes from https://github.com/theforeman/foreman/pull/4678  ? I think that's why tests are failing now.

ares:
rebased

lzap:
Capybara::Poltergeist::StatusFailError: Request to 'http://127.0.0.1:37703/variable_lookup_keys' failed to reach server, check DNS and/or server status seems unrelated.

dLobatog:
Thanks @ares !

dLobatog:
Cherry picked to 1.15-stable as d2eece634bf9040e4d9b32a36886b097980e1cae

PR #4678, opened 2017-07-18
Fixes #20325 - Pin ruby2ruby to 2.3.2

tbrisker:
Version 2.4.0 seems to break safe render, pinning for now to last known
good version.


dLobatog:
Thanks @tbrisker!

PR #4680, opened 2017-07-18
fixes #20328 - ensure target div is always empty

ohadlevy:
this solves the problem when a div has already react component on it,
which gets reloaded via some ajax call.


ohadlevy:
@timogoebel i think this should solve the issue you were seeing earlier in #4676

dLobatog (approved):
Thanks @ohadlevy !

PR #4681, opened 2017-07-18
Fixes #20321 - run puppet tasks under admin

ares:
the patch looks scary to review but it only wraps tasks in block like
```
User.as_anonymous_admin do
end
```


mmoll:
adding `?w=1` to the URL makes it easier to review on the GH Web UI 🔍

dLobatog (approved):
@ares Thanks, setting this for 1.15.3

ares:
Rebased to pin down ruby2ruby

ares:
all green, code climate got stuck but is green too - https://codeclimate.com/github/theforeman/foreman/pull/4681

dLobatog (approved):
Merging, thanks @ares !

PR #4684, opened 2017-07-19
Fixes #20345 - add back button on host info page

tbrisker:


app/helpers/hosts_helper.rb
```diff
@@ -301,48 +301,51 @@ def state(s)
...
-                                    }
-              )
-            end
-        ),
-        if host.supports_power?
-          button_group(
-              link_to(_("Loading power state ..."), '#', :disabled => true, :class => 'btn btn-default', :id => :loading_power_state)
+      button_group(
+        link_to(_("Back"), hosts_path, :class => 'btn btn-default')
+      ),
```
tbrisker, line 306: These three lines are the only actual change, the rest is correcting alignment.

lzap:
Aren't back buttons UX evil? I am good with the change tho :-) What's Roxanne's nick again?

ohadlevy:
@tbrisker do you mind adding a before / after image that @rohoover can have a look? thanks!

tbrisker:
Before:
screenshot from 2017-07-20 10-14-49

After:
screenshot from 2017-07-20 10-14-49

dLobatog (commented):
I tend to agree with @lzap , probably it's best to save the search in a cookie or similar so that on future accesses to "All hosts" (clicking the item from the menu) the search is saved. Personally I found back and forward buttons very redundant (all browsers have that..)

lzap:
I noticed this is backport request, if you need this than we can simply merge this and create refactor ticket, that would definitely work for me.

timogoebel:
Just my two cents: I think breadcrumbs would be better than a "Back" button.

tbrisker:
@dLobatog - I wonder if cookies won't be a bit overkill for this? other "back" buttons don't have this functionality.
@timogoebel - I agree, but currently we already have a "back" button in several other pages and this will be consistent with them. Adding breadcrumbs is definitely the long-term solution, but i'd rather be consistent now than add different implementations on different pages or refactor all the other pages for this.

ohadlevy:
@rohoover can you ack/nack please? thanks!

Rohoover:
I'm fine with adding the back button (as a consistent temporary solution) so long as we intend to nix all back buttons for breadcrumbs instead :)

PR #4685, opened 2017-07-20
fixes #20369 - change host creation form tab order

ohadlevy:
see PR screenshot for more info.

### DIFF
--- a/app/views/hosts/_form.html.erb
+++ b/app/views/hosts/_form.html.erb
@@ -11,16 +11,19 @@
     <ul class="nav nav-tabs" data-tabs="tabs">
       <li class="active"><a href="#primary" data-toggle="tab"><%= _('Host') %></a></li>
 
-      <%= render_tab_header_for(:main_tabs, :subject => @host) %>
-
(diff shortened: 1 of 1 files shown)


ohadlevy:
this pr is mostly for discussion /cc @rohoover

image

ohadlevy:
image

timogoebel:
I like it. Seems to be a more natural / workflow oriented order.

dLobatog (commented):
Big +1, this is normally my workflow, why "Do not merge"?

I think the only risk is that plugins that use `deface` could be broken but nothing that can't be fixed before 1.16. Ansible for example tags its tab next to the Puppet tab so it still makes sense.

ares:
:+1: from me too

ohadlevy:
@dLobatog removed DNM and added a redmine issue.

tbrisker:
Thanks @ohadlevy !

ares:
Thanks for merging @tbrisker, please don't forget to set release in redmine.

PR #4199, opened 2017-01-17
Fixes #17992, #18103 - Improve external usergroup errors

dLobatog:
When one submit an user group with external user groups, and this
doesn't work for whatever reason, like:

Net::LDAP::Error - No route to host - connect(2)
LdapFluff::Generic::UnauthenticatedException
LdapFluff::ActiveDirectory::MemberService::UIDNotFoundException,

it should show these errors in the UI and ideally some text explaining
how to solve the issue.

Currently the errors are merely logged to production.log, which leaves
users unable to understand what happened.

This commit makes those errors show up in the UI with a suggestion on
how to fix those we know.

To test you don't need an AuthSourceLDAP, actually just testing with broken IPs and so forth should show the error in the usergroups page.

PR #4686, opened 2017-07-21
Fixes #11153 - handle ldap errors gracefully

ares:
Before (sorry for missing icons):

problem

after:
submit

refresh

### DIFF
--- a/app/controllers/external_usergroups_controller.rb
+++ b/app/controllers/external_usergroups_controller.rb
@@ -7,7 +7,9 @@ class ExternalUsergroupsController < ApplicationController
     if @external_usergroup.refresh
       notice _("External user group %{name} refreshed") % { :name => @external_usergroup.name }
     else
-      warning _("External user group %{name} could not be refreshed") % { :name => @external_usergroup.name }
+      message = _("External user group %{name} could not be refreshed.") % { :name => @external_usergroup.name }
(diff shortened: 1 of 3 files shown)


dLobatog (commented):
@ares How does this differ from https://github.com/theforeman/foreman/pull/4199 ?

ares:
I didn't see that one in action so I can't tell, I was reviewing https://github.com/theforeman/foreman/pull/4668 while I found these issues so I fixed them right away. If you think it's already solved by #4199, I can close it. PR #4199 should be linked to http://projects.theforeman.org/issues/11153/ in that case.

One difference I see from quick reading the code is that this patch registers the error on auth_source_id of the external group so in UI you can see the field marked red. I'm not sure how base errors are displayed on that page.

boaz0 (approved):
LGTM 👍

ares:
[test katello]

dLobatog (changes requested):
@ares I think I covered this error in #4199 , is this still relevant?

ares:
If nothing else, this fixes the issue on refresh button which was not covered. It is still relevant. Removing WOC. The before is still reproducible, you need to be in production env. Someone please review.

dLobatog (approved):

dLobatog:
Thanks @ares !

PR #4690, opened 2017-07-23
Fixes #20385 - fix the role plugin DSL

ares:
After create/update permission started to be enforced in Foreman 1.16,
the Role.where(:name => name).first_or_create can fail silently. The
newly built role is considered a new record which later when we add
permissions and save! it fails. The validation that prevents the
save is uniqueness on name.

### DIFF
--- a/app/registries/foreman/plugin.rb
+++ b/app/registries/foreman/plugin.rb
@@ -243,11 +243,13 @@ module Foreman #:nodoc:
     # Add a new role if it doesn't exist
     def role(name, permissions)
       default_roles[name] = permissions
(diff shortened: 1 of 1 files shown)


iNecas:
Could you add some background info to the commit message? Might help few months later, when something goes wrong again to figure out, what was the cause here.

ares:
here you get, let me know if you have something else in mind

iNecas:
Running some tasks PR test against this test, just to confirm it works :) http://ci.theforeman.org/job/test_plugin_pull_request/3940/

iNecas:
The tests have confirmed this fixing the issue. Thanks @ares

PR #4691, opened 2017-07-23
Fixes #20384 - seed resources before admin exists

ares:
For seed scripts that executes before internal admin exist we need to
ignore creation permission check on resources that includes Authorizable
concern.

### DIFF
--- a/app/models/concerns/authorizable.rb
+++ b/app/models/concerns/authorizable.rb
@@ -2,6 +2,8 @@ module Authorizable
   extend ActiveSupport::Concern
 
   def check_permissions_after_save
+    return true if Thread.current[:ignore_permission_check]
+
     authorizer = Authorizer.new(User.current)
     creation = self.id_changed?
     name = permission_name(creation ? :create : :edit)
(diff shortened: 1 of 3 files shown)


tbrisker:
I'm curious, would it make sense to move the admin seed up in the order so it runs before the others? as far as I can tell the only other seed it depends on is the auth source one.

ares:
Yeah, the problem would be still there for auth sources but roles could be seeded later, permissions too but they don't even include Authorizable, I suppose the file would be called `03-admin.rb` while roles would move to `031-roles.rb`. So do you want me to move it?

chris1984 (commented):
Confirmed this fixed my new dev instance

orrabin:
Works for me as well. @tbrisker waiting for your response on this.

tbrisker:
I'm ok with merging this right now and opening a separate PR to reorder the seeds in a better way. @ares - what do you prefer?

chris1984:
@ares this is blocking new dev installs and imho a critical blocker.

tbrisker:
thanks @ares - decided to merge this as is for now to unblock dev, feel free to open a new pr reordering the seeds.

ares:
thanks @tbrisker, will do

PR #4692, opened 2017-07-23
Fixes #20386 - Allow to identify smart proxy by ip only

agx:
This allows setting trusted_puppet_master_hosts to an IP in the
non-https case. This can e.g. be useful when testing ansible fact
importing from another machine.

### DIFF
--- a/app/controllers/concerns/foreman/controller/smart_proxy_auth.rb
+++ b/app/controllers/concerns/foreman/controller/smart_proxy_auth.rb
@@ -92,6 +92,7 @@ module Foreman::Controller::SmartProxyAuth
       logger.warn "SSL is required - request from #{request.ip}"
     else
       request_hosts = Resolv.new.getnames(request.ip)
+      request_hosts = [request.ip] if request_hosts == []
     end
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/foreman/controller/smart_proxy_auth.rb
```diff
@@ -92,6 +92,7 @@ def auth_smart_proxy(proxies = SmartProxy.unscoped.all, require_cert = true)
       logger.warn "SSL is required - request from #{request.ip}"
     else
       request_hosts = Resolv.new.getnames(request.ip)
+      request_hosts = [request.ip] if request_host == []
```
dLobatog: `request_hosts` - mind to write a test to demonstrate the behavior?

dLobatog (changes requested):
Thanks @agx, makes sense

agx:
@dLobatog added a test now as well

dLobatog (approved):

dLobatog (approved):

dLobatog:
Thanks @agx!

PR #4693, opened 2017-07-24
Fixes #20387 - Handle extra spaces in the search field

rabajaj0509:
This commit shall improve the usability of the search field.

### DIFF
--- a/app/controllers/concerns/foreman/controller/auto_complete_search.rb
+++ b/app/controllers/concerns/foreman/controller/auto_complete_search.rb
@@ -16,6 +16,7 @@ module Foreman::Controller::AutoCompleteSearch
         completed = item.to_s.chomp(part)
         {:completed => CGI.escapeHTML(completed), :part => CGI.escapeHTML(part), :label => item, :category => category}
       end
+      params[:search].strip!
     rescue ScopedSearch::QueryNotSupported => e
       @items = [{:error =>e.to_s}]
     end


boaz0 (approved):
LGTM but I think it would be lovely if there was a test for that change.

timogoebel (changes requested):
Thanks. Please add a test for this and add a dedicated ticket to Foreman's Redmine.

ohadlevy:
I would like to understand where the initial whitespace is coming from.
it seems that when clicking on an auto completion suggestion it adds a whitespace.
I am guessing that was done so that if you already have a query it would include a space char between?

rabajaj0509:
@ohadlevy  as I am testing it now, the bug is no longer reproducible. The search parameter is getting saved without the spaces automatically. Therefore, I am not sure what is to be done here.
The re-factoring pr still holds good https://github.com/theforeman/foreman_discovery/pull/366.

ohadlevy:
@rahulbajaj0509 are you sure? I believe I can reproduce it, did you test discovery without your patched branch?

rabajaj0509:
@ohadlevy, yes, i am not being able to reproduce it. 
discoverysearch
I had given extra spaces before the `cpu_count`. Am i missing something here?

lzap:
So is my understanding correct that this patch:

https://github.com/theforeman/foreman_discovery/pull/366

fixed it?

rabajaj0509:
@lzap nope, i guess there is some change in the gem itself that handled this issue (i am just guessing, i might be wrong). 

theforeman/foreman_discovery#366 is just refactoring, so when we added the welcome page to dicovery hosts (in theforeman/foreman_discovery#20247 ) we `added  model_of_controller`.

lzap:
The thing is that the space is only visible when you editing the field. After save/submit it disappears. Let me ask the reporter in the BZ and then we can close this - I do not think this is worth any effort.

PR #4695, opened 2017-07-25
Fixes #20403 - Only show authorized compute profiles in forms

tbrisker:


app/helpers/application_helper.rb
```diff
@@ -492,10 +492,8 @@ def accessible_resource(obj, resource, order = :name, association: resource)
     list
   end
 
-  def accessible_related_resource(obj, relation, opts = {})
+  def accessible_related_resource(obj, relation, order: :name, where: nil)
     return [] if obj.blank?
-    order = opts.fetch(:order, :name)
-    where = opts.fetch(:where, nil)
```
tbrisker, line 498: not directly related to this but just noticed while i was working on this and cleaned up along the way.

app/models/compute_profile.rb
```diff
@@ -20,5 +20,5 @@ class ComputeProfile < ApplicationRecord
   scoped_search :on => :name, :complete_value => true
   default_scope -> { order('compute_profiles.name') }
 
-  scope :visibles, -> { eager_load(:compute_attributes).where('compute_attributes.id > 0') }
+  scope :visibles, -> { joins(:compute_attributes) }
```
tbrisker: we don't actually use the eager loaded attributes anywhere, this looks like some weird way to turn a left outer join into an inner join.

app/helpers/application_helper.rb
dLobatog: Should be safe to do this https://github.com/search?q=accessible_related_resource&type=Code&utf8=%E2%9C%93

dLobatog (changes requested):
@tbrisker Looks good, can you add a test to ensure this behavior remains? Just ensuring `visible_compute_profiles` does not return ComputeProfiles a user is unauthorized to see should be enough. Thanks!

app/helpers/application_helper.rb
tbrisker: not sure I understand your comment here, the behaviour doesn't change at all, just replace the ops hash into kwargs.

tbrisker:
@dLobatog thanks, with the test i realized i had the wrong permission name there and that we don't have fixtures for compute profile permissions at all.

tbrisker:
test failures seem to be unrelated.

boaz0 (approved):
LGTM

dLobatog (approved):

dLobatog:
Thanks @tbrisker !

PR #4696, opened 2017-07-25
Refs #20384 - add higher granularity and reorder seeds

ares:
Follow up of https://github.com/theforeman/foreman/pull/4691


tbrisker:
@ares looks like tests are failing, it appears it's trying to run the old seeds for some reason.
I am wondering if maybe we should create just the anonymous admin even without an authsource (using `:validate => false`) before any of the seeds run, possibly even in the seeds.rb itself, to make sure that all seeds run as this user and simplify the need to "bootstrap" ourself. 
We could similarly disable auditing completely for all of the seeds instead of doing that in every seed file manually, as I assume we don't care about seeds being audited (although if we run them as a user the problem with auditing i think also goes away).

ares:
> I am wondering if maybe we should create just the anonymous admin even without an authsource 

I'm afraid creating invalid objects would bite us later.

it failed since I forgot to update paths in migrations, it should be green now

adamruzicka (changes requested):
This requires `sed -i 's/02/020/' db/seeds.d/020-roles_list.rb` The tests are failing because of this

https://github.com/ares/foreman/blob/ab500bc28b7526c8c4ea91a6a57de9fa36b1d402/db/seeds.d/020-roles_list.rb#L1

ares:
@adamruzicka I fixed more occurrences, please take another look

adamruzicka (commented):
Looks good to me, waiting for Jenkins to turn green.

adamruzicka:
One more thing to fix
```
LoadError: cannot load such file -- /var/lib/workspace/workspace/test_develop_pr_core/database/postgresql/ruby/2.1/slave/fast/db/seeds.d/17-notification_blueprints.rb
    test/notifications_test_helper.rb:10:in `seed_notification_blueprints'
```

ares:
Thanks, updated. Grep didn't find anything else, so fingers crossed.

adamruzicka:
Sadly, we are still not there

ares:
Finally a reasonable error that actually revealed an issue, not caused by the change though :-) I kept the change in a separate commit. The issue was that we try to create default role if it's missing after any user save. This created audit. I also added skip permissions check as this is a system action that should not be subject of user permission scoping. @adamruzicka could you please take another look? I feel we're getting closer.

adamruzicka (approved):
Looks good to me, katello's tests are red, but there's a fix for that on the way so I guess we can get this in.

ares:
[test katello] just to be sure :-) adding RfM as an indication that non-commiter review was done

iNecas:
It seems we might need a rebase to get the katello tests green

ares:
Rebased and squashed into a single commit.

iNecas:
@ares you already kept the unrelated change in separate commit, as mentioned in https://github.com/theforeman/foreman/pull/4696#issuecomment-328891550, I would recommend keeping it that way, including the new issue describing the issue, that we could align properly to the 1.17 release (as we are now just referencing some other issue already merged)

ares:
@iNecas I resplit it to two separate issues, I keep them in the same PR since the second on requires the first to be fixed. I can open a separate PR for the first if you want.

iNecas:
This works for me. Thanks @ares

PR #4697, opened 2017-07-26
Fixes #20410 - normal user now can send test email

snagoor:
A user with "viewer" role is unable to send a test email previously. Test email would only have triggered if an user has "edit_users" role. 

With this patch, test email can be triggered with "viewer" only role.

### DIFF
--- a/test/controllers/users_controller_test.rb
+++ b/test/controllers/users_controller_test.rb
@@ -402,6 +402,16 @@ class UsersControllerTest < ActionController::TestCase
     assert_equal user.mail, mail.to[0]
   end
 
+  test "test email should be delivered even when user has read only role" do
+    user = users(:one)
+    user.roles = [Role.find_by_name('Viewer')]
+    User.current = user
+    put :test_mail, { :id => user.id, :user => {:login => user.login} }, set_session_user
+    mail = ActionMailer::Base.deliveries.last
+    assert mail.subject.include? "Foreman test email"
+    assert_equal user.mail, mail.to[0]
+  end
+
   context "when user is logged in" do
     test "#login redirects to previous url" do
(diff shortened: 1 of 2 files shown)


ohadlevy:
nice! can you please add a test too :)

snagoor:
@ohadlevy done test case added.

test/controllers/users_controller_test.rb
```diff
@@ -402,6 +402,16 @@ class UsersControllerTest < ActionController::TestCase
     assert_equal user.mail, mail.to[0]
   end
 
+  test "test email should be delivered even when user has read only role" do
+    user = users(:one)
+    user.roles = [Role.find_by_name('Viewer')]
+    User.current = user
```
tbrisker, line 408: Please use FactoryGirl to create a user with only viewer permissions rather then modify the user in fixtures which may have unexpected side effects in other tests.

test/controllers/users_controller_test.rb
```diff
@@ -402,6 +402,16 @@ class UsersControllerTest < ActionController::TestCase
...
+    put :test_mail, { :id => user.id, :user => {:login => user.login} }, set_session_user
```
tbrisker, line 409: `set_session_user`, when not passed a parameter, defaults to run the request as admin. You should user `set_session_user(user)` instead.

tbrisker (changes requested):
Thanks @snagoor ! just a couple of inline comments.

dLobatog (commented):
This looks to me like a duplicate of https://github.com/theforeman/foreman/pull/4595 , we have even tests of the behavior. 

@snagoor Please try again using the develop version and feel free to reopen if you found something wrong.

PR #4699, opened 2017-07-26
Fixes #20418 - Add taxonomies to newly imported puppet env

xprazak2:


app/services/puppet_class_importer.rb
```diff
@@ -179,7 +179,7 @@ def proxy_environments
   end
 
   def to_be_created_environments
-    proxy_environments - Environment.where(:name => proxy_environments).pluck(:name)
+    proxy_environments - Environment.unscoped.where(:name => proxy_environments).pluck(:name)
```
dLobatog, line 182: Why is this necessary? With this, no error is shown to the user - just "No new environments created". Before, this would attempt to create a duplicate environment - but it would show the error "Name already exists (even if not visible to current user)"

dLobatog (changes requested):
@xprazak2 Generally 👍 , I was a bit confused at the change to `to_be_created_environments`, is it necessary?

Thanks!

app/services/puppet_class_importer.rb
xprazak2: I added the `unscoped` exactly because of the error. When you have no environments and you get the error, import just seems broken to the user. You are also leaking information out of taxonomy scope - now you know there is an environment that you are not supposed to see and you know its name.

Import should not try to create duplicates IMHO, because resources are shared. So if there is already an env in db but we do not have it in our taxonomy, then we are not supposed to see it or touch it.

dLobatog (approved):

dLobatog:
[test katello] - will merge if it passes

dLobatog:
Thanks @xprazak2 !

ares:
Thanks @dLobatog for merging, please don't forget to set redmine release field after merge. I've set it to 1.16.0.

PR #4701, opened 2017-07-27
Refs #17204 - Settings tabs concat Rails 5 compatible

dLobatog:
Concatenating the first tab (General) with the rest using 'concat' isn't
compatible with Rails 5, '+' is.

### DIFF
--- a/app/models/setting.rb
+++ b/app/models/setting.rb
@@ -77,7 +77,7 @@ class Setting < ApplicationRecord
 
   def self.stick_general_first
     sticky_setting = 'Setting::General'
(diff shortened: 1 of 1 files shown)


ohadlevy:
confirmed its working - thanks @dLobatog

swapab:
`ActiveRecord#concat` is not supported in rails 5?
https://apidock.com/rails/ActiveRecord/Associations/CollectionAssociation/concat

dLobatog:
@swapab Your link is precisely pointing to all the versions which include it, which is at most 4.2.7

PR #3683, opened 2016-07-28
fixes #10982 - configure RestClient log to 'proxy' logger

domcleal:
Request/response information is sent to the proxy logger, showing the
URLs called, headers, request body etc.

The X-Request-ID header is now also set in the proxy requests to the
same session ID that is logged by default in Foreman. The ID is now a
digest of the original session ID to prevent the user's session ID being
hijacked from the proxy request data.

PR #4702, opened 2017-07-27
fixes #10982 - configure RestClient log to 'proxy' logger

mmoll:
Request/response information is sent to the proxy logger, showing the
URLs called, headers, request body etc.

The X-Request-ID header is now also set in the proxy requests to a
random ID stored in the session, to prevent the user's real session ID
being hijacked from the proxy request data. The ID should be correlated
against the proxy logger.


mmoll:
rebased version of #3683

lzap:
What happened, why closed?

mmoll:
see #3683

PR #4703, opened 2017-07-31
Fixes #20445 - PuppetCA and SSH key labeled as "Create"

dLobatog:


ohadlevy:
@dLobatog thanks, looking good, can you please fix the typo on commit message?

ohadlevy:
@dLobatog should our basic tests catch that? (e.g. basic_new_test or basic_index_test )

dLobatog (commented):
@ohadlevy I don't think it's worth really looking for a button with "Create", or "New", sometimes we want that string ("New Hosts",  "New Puppet Environments " (on the new/changed/deleted page), etc..

ohadlevy (approved):

ohadlevy:
@dLobatog see test failure. this was the test I was talking about (that we should consider adding those tests for puppet ca flows).

dLobatog (commented):
Updated the test

app/views/key_pairs/index.html.erb
```diff
@@ -41,14 +41,14 @@
...
-            <h1><%= _('SSH keys') %></h1>
+            <h1><%= _('SSH Keys') %></h1>
             <p>
               <%= _("When using SSH based provisioning, a special SSH key is created in order to automate the provisioning process.") %><br />
               <%= _("Normally when setting up a Compute Resource, a key pair (private and public) is created for you automatically.") %><br />
               <%= _("However, if it was deleted externally this page helps you to recreate it") %>
              </p>
             <div class="blank-slate-pf-main-action">
-              <%= link_to(_('New SSH key'), compute_resource_key_pairs_path(@compute_resource), :method=> :post, :class => "btn-lg" ) %>
+              <%= link_to(_('Create SSH Key'), compute_resource_key_pairs_path(@compute_resource), :method=> :post, :class => "btn-lg" ) %>
```
ares, line 51: Btw there's a [request](https://bugzilla.redhat.com/show_bug.cgi?id=1480248) to use Add instead of Create at user profile page,  since we are not effectively creating the key. I'm not sure whether the same should apply here... Sorry for double pinging, but @rohoover what's your though on this (and the original request)

Rohoover: This is a content strategy issue in that we need to define what we mean by add, new, create and apply those terms consistently. 
In the case of  SSH keys, if we are Adding and not Creating one, then we should label the button as Add.

ares: Then my question is, wouldn't this be more confusing? People are also not creating domains, they just defining/adding what they've created on their DNS servers. The same goes for subnets that have been created on their routers.

Rohoover: My interpretation of what you wrote is that in the case of domains, they are Adding elements to Create a larger thing.

ares: So does that mean that we should have on some pages "Add Domain" while elsewhere we'd have "Create Host"? To me the line is too thin and I would prefer consistency - Create everywhere. It's less prone to errors and confusion for people, who don't realize this difference.

ohadlevy:
@dLobatog can you please fix it in orgs/locations too (see https://github.com/theforeman/foreman/blob/develop/app/helpers/taxonomy_helper.rb#L44)

xprazak2:
@dLobatog, could you rebase and take a look at suggestions in previous comments pls?

PR #4704, opened 2017-07-31
fixes #20453 - build is enabled for cloned host

timogoebel:


app/controllers/hosts_controller.rb
```diff
@@ -93,6 +93,7 @@ def new
   def clone
     @clone_host = @host
     @host = @clone_host.clone
+    @host.build = true
```
orrabin: What is we clone a host that is unmanaged? Maybe this should copy the build from the host that is being cloned or only set if the host is managed?

dLobatog: Build is already copied without this PR, the issue is that when you clone a Host you would (normally) want `Build` to be true so that the host is actually built, otherwise it can be useless. 

@orrabin You made a good point about unmanaged hosts though, in that case we don't really care and it should be set to false. Otherwise we get an error (https://github.com/theforeman/foreman/blob/develop/app/models/host/base.rb#L507)

@timogoebel Can you add a check for that please?

test/integration/host_test.rb
```diff
@@ -134,5 +134,11 @@ class HostIntegrationTest < ActionDispatch::IntegrationTest
       click_link 'Parameters'
       assert page.has_no_selector?('#params .has-error')
     end
+
+    test 'build mode is enabled' do
+      host = FactoryGirl.create(:host, :managed)
+      visit clone_host_path(host)
+      assert page.has_checked_field?('host_build')
+    end
```
dLobatog: Please add a test for unmanaged too

dLobatog (changes requested):
@timogoebel Makes sense for sure, normally I wouldn't clone hosts without the intention of building them. Please checkout the inline comments, ready to merge after that. Thank you 👍 !

timogoebel:
@orrabin / @dLobatog : Thanks. Updated.

dLobatog (approved):
👍 pending 🍏

timogoebel:
@dLobatog: Thank you. We're ready to 🚢 .

dLobatog:
:shipit:

PR #4705, opened 2017-07-31
Refs #16884 - add minitest stubbing to test helper

ares:


lzap:
Thanks, merged. The original change broke discovery unfortunately, resource type is now "Host" instead "Host::Discovered".

ares:
which makes it a bit more complicated, I'll summarize what needs to be changed in separate email

PR #4706, opened 2017-07-31
"Fixes #20446 - reset page param after search"

lizagilman:


webpack/assets/javascripts/foreman_tools.js
```diff
@@ -87,6 +87,9 @@ export function updateTable(element) {
   let uri = new URI(window.location.href);
   let search, perPage;
 
+  if (getQueryStringValue(uri, 'page') > 1) {
+    uri.setSearch('page', 1);
+  }
```
ohadlevy: won't this always set the page to 1 regardless? afaiu this function is called both from the per_page and the search submit?.

can you write a test to demonstrate the various options?

webpack/assets/javascripts/foreman_tools.js
```diff
@@ -90,6 +90,9 @@ export function updateTable(element) {
   if (element !== undefined) {
     search = $(element).find('.autocomplete-input').val() || $('#search-form').find('.autocomplete-input').val();
     if (search !== undefined) {
+      if (getQueryStringValue(uri, 'page') > 1) {
```
ohadlevy: I dont think thats correct, if i have a search term and then I'll do pagination, wont this always keep me on the first page?

IMHO only when the search has been submitted (not if it has a value) then you reset the page number to 1. (or remove the page attribute from the url)

ohadlevy:
btw: I think we need to reset the page number when changing the per page value as well.

I just changed from 20 to 5, and then moved 10 pages or so, and then i change it back to 15 where page 10 didn't exists anymore.

lizagilman:
updated

lizagilman:
[test foreman]

ohadlevy:
@lizagilman tests are failing.

lizagilman:
@ohadlevy minor change.. updated

lizagilman:
@ohadlevy ping ?

webpack/assets/javascripts/foreman_tools.js
```diff
@@ -85,19 +85,38 @@ export function initTypeAheadSelect(input) {
...
   if (element !== undefined) {
-    search = $(element).find('.autocomplete-input').val() || $('#search-form').find('.autocomplete-input').val();
-    if (search !== undefined) {
-      uri.setSearch('search', search.trim());
+    isSearchForm = (element.id === 'search-form');
+    if ((isSearchForm || element.id === 'per_page') && getQueryStringValue(uri, 'page') > 1) {
+      uri.setSearch('page', 1);
+    }
+    if (isSearchForm) {
+      searchTerm = $(element).find('.autocomplete-input').val() || $('#search-form').find('.autocomplete-input').val();
```
tbrisker: `isSearchForm` means that `element` has id of `search-form`, so that the two conditions are actually checking for the same thing.

webpack/assets/javascripts/foreman_tools.js
```diff
@@ -85,19 +85,38 @@ export function initTypeAheadSelect(input) {
...
+      if (searchTerm) {
```
tbrisker: If there isn't a value shouldn't we still set the search to an empty value?

webpack/assets/javascripts/foreman_tools.js
```diff
@@ -85,19 +85,38 @@ export function initTypeAheadSelect(input) {
...
+        uri.setSearch('search', searchTerm.trim());
+      }
+    }
+    perPage = $('#per_page').val();
+    if (perPage !== undefined) {
+      if (element.id !== 'search-form' && !(uri.hasSearch('search'))) {
```
tbrisker: the first clause here is the same as `!isSearchForm`, and the second I think is redundant - if the uri doesn't have search, there is nothing to remove.

webpack/assets/javascripts/foreman_tools.js
```diff
@@ -85,19 +85,38 @@ export function initTypeAheadSelect(input) {
...
   let uri = new URI(window.location.href);
-  let search, perPage;
+  let searchTerm, perPage, isSearchForm;
 
   if (element !== undefined) {
-    search = $(element).find('.autocomplete-input').val() || $('#search-form').find('.autocomplete-input').val();
-    if (search !== undefined) {
-      uri.setSearch('search', search.trim());
+    isSearchForm = (element.id === 'search-form');
+    if ((isSearchForm || element.id === 'per_page') && getQueryStringValue(uri, 'page') > 1) {
```
tbrisker: why do we need to check if page > 1? if it is 1 setting it to 1 is a noop, and it makes the `getQueryStringValue` function redundant.

tbrisker (changes requested):
See comments inline, looks like the code can be simplified a bit.

lizagilman:
waiting for review

waldenraines:
@tbrisker @ohadlevy anything else to be done here before merging?

webpack/assets/javascripts/foreman_tools.js
tbrisker: I think you missed this comment.

lizagilman:
updated

waldenraines (approved):
Tested this out and it worked for me, thanks @lizagilman!

tbrisker (approved):
Works well, thanks @lizagilman !

PR #4707, opened 2017-07-31
Fixes #20461 - csv_exporter results in empty rows

sjha4:
While trying to extend csv export functionality to katello pages, we observed that the csv_exporter was rendering empty rows in the exported csv.
The root cause is the line "columns.map!{|c| c.to_s.split('.').map(&:to_sym)}" getting executed more than once in the enumerator block causing the column names to be prepended with multiple ' : '

### DIFF
--- a/app/services/csv_exporter.rb
+++ b/app/services/csv_exporter.rb
@@ -4,12 +4,12 @@ module CsvExporter
   def self.export(resources, columns, header = nil)
     header ||= default_header(columns)
     raise ArgumentError, "Columns and header row aren't the same length" unless columns.length == header.length
+    columns.map!{|c| c.to_s.split('.').map(&:to_sym)}
     # need to save the current context as the enumerator is executed by a separate thread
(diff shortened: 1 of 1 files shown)


tbrisker:
While the change itself seems harmless to me, I'm curious why on certain pages the enumerator is created multiple times? This may be an indicator of some deeper issue, since on most pages this works without a problem.

sjha4:
@tbrisker : To put this into context, I tried this for the PR: https://github.com/Katello/katello/pull/6889

xprazak2:
@sjha4, I think this has been fixed already as `map!` was replaced with
[map](https://github.com/theforeman/foreman/blob/develop/app/services/csv_exporter.rb#L13). Could you re-test if the issue still present?

sjha4:
Hey @xprazak2 : Tested this on a recent PR in katello around csv export. I do not see this issue anymore.
I will go ahead and close this PR.

PR #4708, opened 2017-08-01
Fixes #20470 - Audit role filter changes

tbrisker:


ares:
Tests are failing. From testing it seems to only audit the changes of search and taxonomy search. I think the request is for tracking what permissions were added to the filter. I was trying to look how to do it and it does not seem trivial. I think we'd need to add a before_save callback and use audit_comment for reasonable output, since the has_many associations can't be easily audited.

tbrisker:
Pushed with fix to tests and better filter naming (permissions list instead of id). Still doesn't audit when permissions on an existing filter change (like any other many-to-many relationship that isn't audited atm). 
Screenshot of example audit log:
screenshot from 2017-08-14 16-36-58

dLobatog (changes requested):
@tbrisker This needs a rebase, changes look reasonable, thanks !

tbrisker:
@dLobatog thanks, rebased

dLobatog (approved):

dLobatog:
Thanks @tbrisker !

PR #4709, opened 2017-08-01
Refs #19050 - Update Rails 5.0 to 5.0.6

mmoll:
Also, in 8121cd5f the default version was set to 4.2, because of https://github.com/rails/rails/issues/25046 breaking all plugin tests, which should be fixed in Rails 5.0.5.

### DIFF
--- a/test/models/orchestration/dhcp_test.rb
+++ b/test/models/orchestration/dhcp_test.rb
@@ -75,7 +75,9 @@ class DhcpOrchestrationTest < ActiveSupport::TestCase
   test "provision interface DHCP records should contain default filename/next-server attributes for IPv4 tftp proxy" do
     ProxyAPI::TFTP.any_instance.expects(:bootServer).returns('192.168.1.1')
     subnet = FactoryGirl.build(:subnet_ipv4, :dhcp, :tftp)
-    h = FactoryGirl.create(:host, :with_dhcp_orchestration, :with_tftp_dual_stack_orchestration, :subnet => subnet)
+    h = as_admin do
+      FactoryGirl.create(:host, :with_dhcp_orchestration, :with_tftp_dual_stack_orchestration, :subnet => subnet)
+    end
     assert_equal 1, h.provision_interface.dhcp_records.size
     assert_equal 'grub2/grubx64.efi', h.provision_interface.dhcp_records.first.filename
     assert_equal '192.168.1.1', h.provision_interface.dhcp_records.first.nextServer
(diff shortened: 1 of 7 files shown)


mmoll:
I don't have enough time to work on the test errors right now...

test/models/orchestration/dhcp_test.rb
```diff
@@ -75,7 +75,7 @@ def teardown
   test "provision interface DHCP records should contain default filename/next-server attributes for IPv4 tftp proxy" do
     ProxyAPI::TFTP.any_instance.expects(:bootServer).returns('192.168.1.1')
     subnet = FactoryGirl.build(:subnet_ipv4, :dhcp, :tftp)
-    h = FactoryGirl.create(:host, :with_dhcp_orchestration, :with_tftp_dual_stack_orchestration, :subnet => subnet)
+    h = FactoryGirl.create(:host, :with_dhcp_orchestration, :with_tftp_orchestration, :subnet => subnet)
```
mmoll: @timogoebel could you have a look here?

mmoll:
@tbrisker could you have a look the CSV export failures?

mmoll:
The mailer tests seem flakey in my env (there's always one test failing, but always a different one).

mmoll:
if anybody has fixes for the remaining test failyures, please just push into my fork (should work if you're having push permissions to theforeman/foreman) or open a PR to this branch in my fork.

test/models/orchestration/dhcp_test.rb
timogoebel: @mmoll: Does this break anything? Basically, the change is fine. The behaviour should be well enough covered in the TFTP tests. This just tests, that the behaviour doesn't change if TFTP orchestration is enabled on the IPv6 subnet.

mmoll: without this change:
`ActiveRecord::SubclassNotFound: Invalid single-table inheritance type: Subnet::Ipv4 is not a subclass of Subnet::Ipv6`

timogoebel: Ouch, I have a feeling that we should look into that rather than changing the test.

We do some pretty nasty stuff for the subnet classes:
https://github.com/theforeman/foreman/blob/develop/app/models/concerns/foreman/sti.rb &
https://github.com/theforeman/foreman/blob/develop/app/models/subnet.rb#L281-L287

tbrisker:
@mmoll looks like the tests aren't failing due to the version change but rather due to the config change now making them really run in rails 5 - I just manually ran one of the tests on 5.0.4 and it fails as well.

tbrisker:
@mmoll The mail issues seem to be related to the dynflow activejob adapter - I managed to reliably reproduce it by reusing the same seed value, looks like the first test that runs with activemailer fails during the teardown stage due to activejob not recognizing Dynflow::ActiveJob::QueueAdapters::DynflowAdapter as an activejob adapter. @iNecas  - any thoughts?

iNecas:
Seems like @dLobatog had a reason for pinning down the rails to `< 5.0.0` here https://github.com/Dynflow/dynflow/commit/37d02510d69cc3c8fa27a4ba46fe73b512a4d661

iNecas:
Some compatibility work needed to be done for Rails 5 https://github.com/Dynflow/dynflow/pull/249

test/models/orchestration/dhcp_test.rb
mmoll: I did add `as_admin` to the test and it's working now

test/models/setting_test.rb
```diff
@@ -449,6 +449,8 @@ class CustomSetting < Setting
   end
 
   test "create succeeds when cache is non-functional" do
+    # for unknown reasons this fails in the CI environment, but passes locally
+    skip if ActiveRecord::Base.connection_config[:adapter].eql?("postgresql")
```
mmoll, line 453: I did set up a CentOS7 VM with the very same PostgreSQL version, but didn't get the failure.

mmoll:
this is ready 🐎

dLobatog (changes requested):
@mmoll Just merged #4826 , can you rebase to include these changes to make sure Katello tests pass?

mmoll:
:green_heart:

tbrisker (approved):
Tests are green, about time we start testing rails 5 so we stop breaking it.

dLobatog (approved):

tbrisker:
Thanks @mmoll and all others involved!

PR #4710, opened 2017-08-02
Fixes #20483 - seed all permission for core roles

ares:


db/seeds.d/02-roles_list.rb
```diff
@@ -38,38 +40,7 @@ def role_names
...
-       :create_smart_proxies_autosign, :destroy_smart_proxies_autosign,
-       :edit_smart_proxies_puppetca, :destroy_smart_proxies_puppetca,
-       :create_ssh_keys, :destroy_ssh_keys,
-       :create_subnets, :edit_subnets, :destroy_subnets, :import_subnets,
-       :create_usergroups, :edit_usergroups, :destroy_usergroups,
-       :create_users, :edit_users, :destroy_users,
-       :destroy_config_reports, :upload_config_reports,
-       :upload_facts,
-       :create_trends, :edit_trends, :destroy_trends, :update_trends]
+      PermissionsList.permissions.reject { |resource, name| name.start_with?('view_') }.map(&:last) - manage_organizations_permissions
```
dLobatog: There's a mismatch between symbols and strings here. 

```
PermissionsList.permissions.reject { |resource, name| name.start_with?('view_') }.map(&:last) 
```

This returns a bunch of permissions, including `['Organization', 'create_organizations']`.. they are all 2-tuples of strings. `manage_organization_permissions` is an array of symbols. Then:

```
[['Organization', 'create_organizations'], ...] - manage_organizations_permissions
```

resolves to 

```
[['Organization', 'create_organizations']] - [:create_organizations, :edit_organizations...]
```

I'd suggest calling `map { |name, permission| permission.to_sym` on the `PermissionsList.permissions.reject` output.

db/seeds.d/02-roles_list.rb
```diff
@@ -79,16 +50,7 @@ def manage_organizations_permissions
...
-        :view_architectures, :view_authenticators, :view_bookmarks, :view_compute_resources, :view_compute_resources_vms, :view_compute_profiles,
-        :view_config_groups, :view_domains, :view_realms, :view_environments, :view_external_usergroups, :view_external_variables,
-        :view_external_parameters, :view_filters, :view_hostgroups, :view_keypairs,
-        :view_hosts, :view_params, :view_images, :view_locations, :view_media, :view_models, :view_operatingsystems,
-        :view_provisioning_templates, :view_ptables, :view_puppetclasses, :view_smart_proxies, :view_smart_proxies_autosign,
-        :view_smart_proxies_puppetca, :view_subnets, :view_organizations, :view_usergroups, :view_users, :view_config_reports,
-        :view_facts, :view_audit_logs, :view_statistics, :view_tasks, :view_trends, :view_plugins, :view_mail_notifications,
-        :access_dashboard, :view_roles, :view_ssh_keys
-      ]
+      PermissionsList.permissions.select { |resource, name| name.start_with?('view_') }.map(&:last)
```
dLobatog: Same problem as in the other line.

dLobatog (changes requested):
@ares Thanks, there's a discrepancy on the format of permissions here and the format of `PermissionsList.permissions`, please check it out.

As far as I understand this is mostly DRY and ensuring 02-permissions_list is the only source of truth?

db/seeds.d/02-roles_list.rb
ares: there's `map(&:last)` to get only permissions, I'll add `to_sym` and convert it to block

db/seeds.d/02-roles_list.rb
ares: this wasn't a problem, there's `map(&:last)` it worked with adding `seeded_roles` but for consistency I think it will be better to use the same construct as in `base_manage_permissions`

ares:
Updated, @dLobatog please take another look.

ares:
And to answer the question

> As far as I understand this is mostly DRY and ensuring 02-permissions_list is the only source of truth?
 
correct, this will ensure we don't forget adding more permissions again

ares:
[test katello]

dLobatog (approved):

dLobatog:
Thanks @ares !

PR #4712, opened 2017-08-02
Fixes #20487 - Add rebuild_config to hostgroups API

treydock:


app/controllers/api/v2/hostgroups_controller.rb
```diff
@@ -94,12 +94,32 @@ def clone
...
+        results = @hostgroup.recreate_hosts_config(params[:only], params[:children_hosts])
+        failures = []
+        results.each_pair do |host, result|
+          host_failures = result.reject { |key, value| value }.keys.map{ |k| _(k) }
+          failures << "#{host}(#{host_failures.to_sentence})" unless host_failures.empty?
+        end
+        if failures.empty?
+          render_message _("Configuration successfully rebuilt."), :status => :ok
+        else
+          render_message (_("Configuration rebuild failed for: %s." % failures.to_sentence)), :status => :unprocessable_entity
```
ShimShtein: You can use the standard `render_error` function, to render the error, like [here](https://github.com/theforeman/foreman/blob/develop/app/controllers/api/v2/hosts_controller.rb#L156).

ShimShtein (changes requested):
:+1:  works as expected, small comment though.

app/controllers/api/v2/hostgroups_controller.rb
```diff
@@ -94,12 +94,32 @@ def clone
         process_response @hostgroup.save
       end
 
+      api :PUT, "/hostgroups/:id/rebuild_config", N_("Rebuild orchestration config")
+      param :id, :identifier, :required => true
+      param :only, Array, :desc => N_("Limit rebuild steps, valid steps are %{host_rebuild_steps}")
```
iNecas, line 109: When does `%{host_rebuild_steps}"` get replaced by the real values?

treydock: @iNecas I think this is set in config/initializers/apipie.rb:
 https://github.com/theforeman/foreman/blob/develop/config/initializers/apipie.rb#L25

treydock:
@ShimShtein Used render_error in place of render_message.

treydock:
The katello failure looks unrelated to my changes.  I'm unsure how to simplify the code to pass codeclimate.

app/controllers/api/v2/hostgroups_controller.rb
iNecas: You're right, thanks for pointers

iNecas:
The katello test failures are being addressed in https://github.com/theforeman/foreman/pull/4826, let's wait for that to get resolved

iNecas:
It would be also nice to cover the `recreate_hosts_config` method with some unit test to catch further regressions.

treydock:
@iNecas Added some basic tests for `recreate_hosts_config`.

iNecas (approved):
Thanks, works for me. Please rebase, as there were some fixes upstream done to make the katello tests green. I'm ok with merging after that

treydock:
@iNecas I've rebased

iNecas:
Thanks @treydock , test failures unrelated. Merging now

iNecas:
Also thanks to @ShimShtein

PR #4713, opened 2017-08-03
Fixes #20454 - change page by typing page number

lizagilman:


webpack/assets/javascripts/foreman_tools.js
```diff
@@ -97,6 +97,11 @@ export function updateTable(element) {
   if (perPage !== undefined) {
     uri.setSearch('per_page', perPage);
   }
+
+  pageNum = $('#pagination1-page').val();
```
ohadlevy: i might be missing something, but where does the pagination1-page comes from? why is it called "1"?

lizagilman:
updated

lizagilman:
can anyone review? @tbrisker / @timogoebel  maybe ?

config/initializers/will_paginate.rb
```diff
@@ -23,8 +23,8 @@ def forward
       end
 
       def pages
-        current_page_input = tag(:input, '', class: 'pagination-pf-page', type: 'text', value: current_page, id: 'pagination1-page')
-        current_page_label = tag(:label, _('Current Page'), class: 'sr-only', for: 'pagination1-page')
+        current_page_input = tag(:input, '', class: 'pagination-pf-page', type: 'text', value: current_page, id: 'cur-page-num')
```
tbrisker: The id here is `cur-page-num` but below you look for `#cur_page_num` so it fails to find the right element. The tests also use `cur_page_num` which is why they pass.

tbrisker (changes requested):
Just a small typo, see inline

tbrisker:
Pushed the typo correction to your branch, will merge if all green.

tbrisker:
Tests passed, merging. Thanks @lizagilman !

tbrisker (approved):

PR #4715, opened 2017-08-03
fixes #20485 - cache notification indicator in local storage

ohadlevy:
this ensures that the toggle icon is not swapping between different page loads.

long term maybe we should cache the entire notifications

### DIFF
--- a/webpack/assets/javascripts/react_app/common/sessionStorage.js
+++ b/webpack/assets/javascripts/react_app/common/sessionStorage.js
@@ -19,5 +19,7 @@ export const notificationsDrawer = {
   getIsOpened: () => getValue('isDrawerOpen'),
   setIsOpened: value => setValue('isDrawerOpen', value),
   getExpandedGroup: () => getValue('expandedGroup'),
-  setExpandedGroup: value => setValue('expandedGroup', value)
+  setExpandedGroup: value => setValue('expandedGroup', value),
+  getHasUnreadMessages: () => getValue('hasUnreadMessages'),
+  setHasUnreadMessages: value => setValue('hasUnreadMessages', value)
 };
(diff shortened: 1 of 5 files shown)


waldenraines:
I don't know if you want to add a story/test for this but otherwise LGTM.

ohadlevy:
thanks @waldenraines - do you have an idea on how to add a test in this case?

I also have mixed feelings about where/how I update the local cache, the other option was to create a dedicated redux action, but i wasn't sure if it make sense (as its a partial data from data we already have an action for).

waldenraines:
> thanks @waldenraines - do you have an idea on how to add a test in this case?

I don't, sorry.  Maybe it's not needed, I am not sure, just noting the absence of one.

> I also have mixed feelings about where/how I update the local cache, the other option was to create a dedicated redux action, but i wasn't sure if it make sense (as its a partial data from data we already have an action for).

Yeah, I feel like the way you're doing it is good if I'm understanding the alternative correctly.  I would rather have notification related session interactions in the notification reducer itself rather than a notification sessions reducer.

tbrisker:
Looks like we are starting to copy the state one part at a time to the session store. Let's just clone it entirely  and load the initial state when needed from the store, rather then add logic every time we want another bit to stay.

ohadlevy:
@tbrisker while i can see us going forward caching the store, I think its more involved with that (e.g. might require to change store tree structure) and while I can work on that, I think the functionality is solid within this current PR.

I will experiment with caching the store, but its a feature by it self (need to handle login/logout etc) which I prefer to do as a feature and not a side effect of fixing this bug.

ohadlevy:
@tbrisker I tried a bit further, and got it to work, but sadly there are many side effects.

I would prefer to include this patch until we find a better / complete solution for client side state caching.

my branch for caching is at https://github.com/ohadlevy/foreman/tree/persisting-redux-state

ohadlevy:
ping - anything missing on this pr? thanks.

tbrisker:
Thanks @ohadlevy !

PR #4716, opened 2017-08-03
Fixes #20475 - Random DB IPAM with MAC preservation

lzap:
Default implementation of DB IPAM increases numbers which creates lot of chances for race conditions when provisioning multiple hosts.

We should provide a random implementation that will randomize assigned IPs. It's not bullet-proof solution to the issue, but chances are much lower.

The implementation initiates random generator based off MAC address if provided (e.g. bare metal hosts). Same hosts will get the same IP address, unless it was already taken by different host. For compute resources (no MAC address known in advance) this will always return random IP address.


test/unit/ipam_test.rb
```diff
@@ -45,6 +45,49 @@ class IPAMTest < ActiveSupport::TestCase
...
+        :subnet_ipv4, :name => 'my_subnet',
+        :network => '10.0.0.0',
+        :mask => '255.0.0.0',
+        :ipam => IPAM::MODES[:random_db])
+      ipam1 = IPAM::RandomDb.new(:subnet => subnet, :mac => "AA:BB:CC:DD:EE:FF")
+      ipam2 = IPAM::RandomDb.new(:subnet => subnet, :mac => "AA:BB:CC:DD:EE:FF")
+      assert_equal ipam1.suggest_ip, ipam2.suggest_ip
+    end
+
+    test "should find unused IPv6" do
```
timogoebel: If you want to keep that test, you also need to extend `supported_ipam_modes` for IPv6.

test/unit/ipam_test.rb
```diff
@@ -45,6 +45,49 @@ class IPAMTest < ActiveSupport::TestCase
...
+    end
+
+    test "should find the only possible IPv4" do
+      subnet = FactoryGirl.create(
+        :subnet_ipv4, :name => 'my_subnet',
+        :network => '192.168.11.0',
+        :from => '192.168.11.5',
+        :to => '192.168.11.5',
+        :ipam => IPAM::MODES[:random_db])
+      ipam = IPAM::RandomDb.new(:subnet => subnet, :excluded_ips => (2..254).map{|x| "192.168.2.#{x}"})
```
timogoebel: What do you want to test with `excluded_ips` being out of subnet's range.

timogoebel (changes requested):
@lzap: Good job. I like this. I haven't tested this, but some comments from a quick review.

test/unit/ipam_test.rb
lzap: Oh this is a invalid leftover, IPv6 make no sense, there is EUI64 which solves this problem already.

app/services/ipam/base.rb
```diff
@@ -17,6 +17,18 @@ def initialize(opts = {})
       normalize_mac!
     end
 
+    def subnet_range
+      @subnet_range ||= begin
+        subnet_range = IPAddr.new("#{subnet.network}/#{subnet.mask}", subnet.family).to_range
+        # exclude first element - network
```
ekohl, line 24: Isn't the network address a IPv4-only thing?

app/services/ipam/base.rb
```diff
@@ -17,6 +17,18 @@ def initialize(opts = {})
...
+        from = subnet.from.present? ? IPAddr.new(subnet.from) : subnet_range.first(2).last
+        # exclude last element - broadcast
```
ekohl, line 26: Same for broadcast

test/unit/ipam_test.rb
lzap: Another leftover, anyway it is a good idea to test excluded IPs. Will do that as well as test for hard limit of iterations done.

app/services/ipam/base.rb
lzap: I haven't changed that, this was the original code. This is used in IPv6 as well, so I assume yes. @timogoebel ?

app/services/ipam/base.rb
lzap: Ditto.

lzap:
Amended test changes and added better comment.

```diff
## random-ipam-20475...lzap/random-ipam-20475
 M app/services/ipam/random_db.rb
 M test/unit/ipam_test.rb
[lzap@box foreman]$ gd
diff --git a/app/services/ipam/random_db.rb b/app/services/ipam/random_db.rb
index 17c6891a7..1f11fec5f 100644
--- a/app/services/ipam/random_db.rb
+++ b/app/services/ipam/random_db.rb
@@ -6,6 +6,8 @@ module IPAM
       IPAddr.new(generator.rand(subnet_range.first.to_i..subnet_range.last.to_i), subnet.family)
     end
 
+    # Safety check not to spend much CPU time when there are no many free IPs left. This gives up
+    # in about a second on Ryzen 1700 running with Ruby 2.4.
     MAX_ITERATIONS = 100_000
     def suggest_ip
       iterations = 0
diff --git a/test/unit/ipam_test.rb b/test/unit/ipam_test.rb
index 11041bff2..214efdeff 100644
--- a/test/unit/ipam_test.rb
+++ b/test/unit/ipam_test.rb
@@ -67,24 +67,37 @@ class IPAMTest < ActiveSupport::TestCase
       assert_equal ipam1.suggest_ip, ipam2.suggest_ip
     end
 
-    test "should find unused IPv6" do
+    test "should find the only possible IPv4" do
       subnet = FactoryGirl.create(
-        :subnet_ipv6, :name => 'my_subnet',
-        :network => '2001:db8::',
-        :ipam => IPAM::MODES[:db])
+        :subnet_ipv4, :name => 'my_subnet',
+        :network => '192.168.11.0',
+        :from => '192.168.11.5',
+        :to => '192.168.11.5',
+        :ipam => IPAM::MODES[:random_db])
       ipam = IPAM::RandomDb.new(:subnet => subnet)
-      assert_match /^2001:db8::/, ipam.suggest_ip
+      assert_equal '192.168.11.5', ipam.suggest_ip
     end
 
-    test "should find the only possible IPv4" do
+    test "should find the only possible IPv4 with excluded IPs" do
       subnet = FactoryGirl.create(
         :subnet_ipv4, :name => 'my_subnet',
         :network => '192.168.11.0',
         :from => '192.168.11.5',
-        :to => '192.168.11.5',
+        :to => '192.168.11.100',
         :ipam => IPAM::MODES[:random_db])
-      ipam = IPAM::RandomDb.new(:subnet => subnet, :excluded_ips => (2..254).map{|x| "192.168.2.#{x}"})
-      assert_equal '192.168.11.5', ipam.suggest_ip
+      ipam = IPAM::RandomDb.new(:subnet => subnet, :excluded_ips => (1..99).map{|x| "192.168.11.#{x}"})
+      assert_equal '192.168.11.100', ipam.suggest_ip
+    end
+
+    test "should stop trying to find random IPv4 after reasonable time" do
+      subnet = FactoryGirl.create(
+        :subnet_ipv4, :name => 'my_subnet',
+        :network => '10.0.0.0',
+        :mask => '255.0.0.0',
+        :ipam => IPAM::MODES[:random_db])
+      ipam = IPAM::RandomDb.new(:subnet => subnet)
+      ipam.excluded_ips.stubs(:include?).returns(true)
+      assert_nil ipam.suggest_ip
     end
   end
```

app/services/ipam/base.rb
lzap: I think it is safe to subtract two IP addresses from whole subnet when range is not present, it is needed for IPv4 and won't hurt for IPv6. But I can do this only in IPv4 case if you insist.

app/services/ipam/base.rb
timogoebel: Well, it can't hurt to skip these addresses also for IPv6. Generally speaking: Yes, IPv6 doesn't have a broadcast address.

app/services/ipam/base.rb
timogoebel: Let's do this in another PR - if we want to do it.

timogoebel:
@lzap:

```ruby
def self.supported_ipam_modes
 -    [:dhcp, :db, :none]
 +    [:dhcp, :db, :random_db, :none]
    end
```

is also needed for `app/models/subnet/ipv6.rb`

timogoebel (approved):
Code looks fine to me. I just need to test this manually and we're good to go.

timogoebel:
@lzap: While testing this I found some issues.
1. When you enter/change the mac address for an interface, no new IP is suggested automatically.
2. When you have a mac address set, the "Suggest new" link always suggests the same IP. It does not change. I'm not sure, if this is actually a bug in our JS code or not. We first clear the ip input field and then do an xhr with a then empty value as `taken_ips`.

A fix for 1 could look like this as we already have that for IPv6.
```diff
diff --git a/app/assets/javascripts/host_edit.js b/app/assets/javascripts/host_edit.js
index d3d06d2f9..e46e2d002 100644
--- a/app/assets/javascripts/host_edit.js
+++ b/app/assets/javascripts/host_edit.js
@@ -582,8 +583,11 @@ $(document).on('change', '.interface_subnet6', function () {
 });

 $(document).on('change', '.interface_mac', function () {
+  var subnet_select = $(this).closest('fieldset').find('select.interface_subnet');
   var subnet6_select = $(this).closest('fieldset').find('select.interface_subnet6');
+  clearIpField(subnet_select, '.interface_ip');
   clearIpField(subnet6_select, '.interface_ip6');
+  interface_subnet_selected(subnet_select, 'ip');
   interface_subnet_selected(subnet6_select, 'ip6');
 });
```

A fix for 2 could look like this:
```diff
diff --git a/app/assets/javascripts/host_edit.js b/app/assets/javascripts/host_edit.js
index d3d06d2f9..00ad5e682 100644
--- a/app/assets/javascripts/host_edit.js
+++ b/app/assets/javascripts/host_edit.js
@@ -560,8 +560,9 @@ function clearIpField(parent, childclass) {

 function suggestNewClick(element, e, suffix) {
   suffix = suffix || '';
+  var ip = $(element).closest('fieldset').find('.interface_ip' + suffix).val();
   clearIpField(element, '.interface_ip' + suffix);
-  interface_subnet_selected($(element).closest('fieldset').find('select.interface_subnet' + suffix), 'ip' + suffix);
+  interface_subnet_selected($(element).closest('fieldset').find('select.interface_subnet' + suffix), 'ip' + suffix, ip);
   e.preventDefault();
 }

@@ -661,7 +662,7 @@ function interface_domain_selected(element) {
   });
 }

-function interface_subnet_selected(element, ip_field) {
+function interface_subnet_selected(element, ip_field, blacklisted_ip) {
   // mark the selected value to preserve it for form hiding
   preserve_selected_options($(element));

@@ -699,6 +700,12 @@ function interface_subnet_selected(element, ip_field) {
   }).get();
   taken_ips.push(interface_ip.val());

+  if(blacklisted_ip) {
+    taken_ips.push(blacklisted_ip);
+  }
+
+  taken_ips = $.grep(taken_ips, function(element) { return !!element });
+
   var data = {
     subnet_id: subnet_id,
     host_mac: interface_mac.val(),
diff --git a/app/services/ipam/base.rb b/app/services/ipam/base.rb
index 02de7cf15..320b4c698 100644
--- a/app/services/ipam/base.rb
+++ b/app/services/ipam/base.rb
@@ -10,7 +10,7 @@ module IPAM

     def initialize(opts = {})
       @subnet = opts[:subnet]
-      @excluded_ips = opts.fetch(:excluded_ips, [])
+      @excluded_ips = opts[:excluded_ips] || []
       @mac = opts.fetch(:mac, nil)
       @errors = ActiveModel::Errors.new(self)
```

app/services/ipam/base.rb
ekohl: I agree it's not in the scope of this PR and that IPv6 generally won't miss those IPs because the subnets are WAY bigger.

lzap: Serve yourselves, free IPs for everyone! :-)

lzap:
(1) works but (2) only adds one excluded IP and then it reverts back (empty the array). It looks like I need to store the array in global variable perhaps?

lzap:
What is this for by the way? It works without it as well:

    taken_ips = $.grep(taken_ips, function(element) { return !!element });

timogoebel:
> What is this for by the way?

It just cleans up the array sent to Foreman. Basically it's a `['a', '', ''].select(&:present?)`.

lzap:
Pushing a different approach. I agree with (1) but for (2) I have something much straighforward. If user don't like the "randomized" IP addresses, we can simply find another random address instead doing magic with blacklisting. So I simply do not pass MAC address anymore when user clicks the link and it's always random.

There is one important change in how algorithm works - it used to pick a random IP and then iterate, but I realized we want to avoid iterating and simply pick random IPs from the generator. They will always be same sequence for individual MAC addresses. It makes things more simple and safe in concurrent environment.

```diff
diff --git a/app/assets/javascripts/host_edit.js b/app/assets/javascripts/host_edit.js
index d3d06d2f9..895679977 100644
--- a/app/assets/javascripts/host_edit.js
+++ b/app/assets/javascripts/host_edit.js
@@ -561,7 +561,7 @@ function clearIpField(parent, childclass) {
 function suggestNewClick(element, e, suffix) {
   suffix = suffix || '';
   clearIpField(element, '.interface_ip' + suffix);
-  interface_subnet_selected($(element).closest('fieldset').find('select.interface_subnet' + suffix), 'ip' + suffix);
+  interface_subnet_selected($(element).closest('fieldset').find('select.interface_subnet' + suffix), 'ip' + suffix, true);
   e.preventDefault();
 }
 
@@ -582,8 +582,11 @@ $(document).on('change', '.interface_subnet6', function () {
 });
 
 $(document).on('change', '.interface_mac', function () {
+  var subnet_select = $(this).closest('fieldset').find('select.interface_subnet');
   var subnet6_select = $(this).closest('fieldset').find('select.interface_subnet6');
+  clearIpField(subnet_select, '.interface_ip');
   clearIpField(subnet6_select, '.interface_ip6');
+  interface_subnet_selected(subnet_select, 'ip');
   interface_subnet_selected(subnet6_select, 'ip6');
 });
 
@@ -661,7 +664,7 @@ function interface_domain_selected(element) {
   });
 }
 
-function interface_subnet_selected(element, ip_field) {
+function interface_subnet_selected(element, ip_field, skip_mac) {
   // mark the selected value to preserve it for form hiding
   preserve_selected_options($(element));
 
@@ -701,7 +704,7 @@ function interface_subnet_selected(element, ip_field) {
 
   var data = {
     subnet_id: subnet_id,
-    host_mac: interface_mac.val(),
+    host_mac: skip_mac ? "" : interface_mac.val(),
     organization_id: org,
     location_id: loc,
     taken_ips: taken_ips
diff --git a/app/services/ipam/base.rb b/app/services/ipam/base.rb
index 02de7cf15..563299de9 100644
--- a/app/services/ipam/base.rb
+++ b/app/services/ipam/base.rb
@@ -10,8 +10,9 @@ module IPAM
 
     def initialize(opts = {})
       @subnet = opts[:subnet]
-      @excluded_ips = opts.fetch(:excluded_ips, [])
+      @excluded_ips = opts[:excluded_ips] || []
       @mac = opts.fetch(:mac, nil)
+      @mac = nil if @mac.try(:blank?)
       @errors = ActiveModel::Errors.new(self)
 
       normalize_mac!
diff --git a/app/services/ipam/random_db.rb b/app/services/ipam/random_db.rb
index 1f11fec5f..c99924e75 100644
--- a/app/services/ipam/random_db.rb
+++ b/app/services/ipam/random_db.rb
@@ -1,9 +1,14 @@
 module IPAM
   # Internal DB IPAM returning all IPs in random order to minimize race conditions
   class RandomDb < Base
+
+    def initialize(opts = {})
+      super(opts)
+      @generator = Random.new(@mac ? @mac.gsub(':', '').to_i(16) : Random.new_seed)
+    end
+
     def random_ip
-      generator = Random.new(@mac ? @mac.gsub(':', '').to_i(16) : Random.new_seed)
-      IPAddr.new(generator.rand(subnet_range.first.to_i..subnet_range.last.to_i), subnet.family)
+      IPAddr.new(@generator.rand(subnet_range.first.to_i..subnet_range.last.to_i), subnet.family)
     end
 
     # Safety check not to spend much CPU time when there are no many free IPs left. This gives up
@@ -11,9 +16,9 @@ module IPAM
     MAX_ITERATIONS = 100_000
     def suggest_ip
       iterations = 0
-      # start with a random address within the subnet range
-      candidate = random_ip
       loop do
+        # next random IP from the sequence generated by MAC seed
+        candidate = random_ip
         iterations += 1
         break if iterations >= MAX_ITERATIONS
         # try to match it
@@ -22,10 +27,6 @@ module IPAM
           logger.debug("Found #{ip} in #{iterations} iterations")
           return ip
         end
-        # or move to the next IP address but
-        candidate = candidate.succ
-        # do not cross the range boundary and try again
-        candidate = subnet_range.first if candidate > subnet_range.last
       end
       logger.debug("Not suggesting IP Address for #{subnet} as no free IP found in reasonable time (#{iterations} iterations)")
       errors.add(:subnet, _('no random free IP could be found in our DB, enlarge subnet range'))
```

app/services/ipam/random_db.rb
```diff
@@ -0,0 +1,36 @@
+module IPAM
+  # Internal DB IPAM returning all IPs in random order to minimize race conditions
+  class RandomDb < Base
+
```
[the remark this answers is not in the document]
lzap: Jeez fixed.

app/services/ipam/random_db.rb
```diff
@@ -0,0 +1,35 @@
+module IPAM
+  # Internal DB IPAM returning all IPs in random order to minimize race conditions
+  class RandomDb < Base
+    def initialize(opts = {})
+      super(opts)
+      @generator = Random.new(@mac ? @mac.gsub(':', '').to_i(16) : Random.new_seed)
```
timogoebel: I'd prefer:
```ruby
def generator
  @generator ||= ...
end
```
And access it via `generator` instead of `@generator`.

Also `@mac` has an `attr_accessor`.

timogoebel (changes requested):
@lzap: Thanks. I have some small nitpicks, then I'll test this once again. If I don't find anything, I think we're ready to 🚢. 👍

app/services/ipam/random_db.rb
lzap: Allright done.

app/services/ipam/random_db.rb
```diff
@@ -0,0 +1,37 @@
+module IPAM
+  # Internal DB IPAM returning all IPs in random order to minimize race conditions
+  class RandomDb < Base
+    attr_accessor :generator
+
+    def initialize(opts = {})
+      super(opts)
+      self.generator ||= Random.new(mac ? mac.gsub(':', '').to_i(16) : Random.new_seed)
```
timogoebel: @lzap: Sorry, one more thing. You don't need to overwrite the `initialize` method. Just define a new method `generator`:

```
  def generator
     @generator ||= Random.new(mac ? mac.gsub(':', '').to_i(16) : Random.new_seed)
  end
```

timogoebel (changes requested):

lzap:
Done and also rebased.

timogoebel (approved):
Works nicely. :-)

timogoebel:
Merged, thanks @lzap. Nicely done.

PR #4717, opened 2017-08-03
Fixes #20499 - Import ConfigReports asynchronously

dLobatog:
Moves the content of the POST calls to create a report to an
asynchronous job to be run by the ActiveJob executor (Dynflow by
default).

Something like this is what's logged when a report is imported.

```
2017-08-03T15:14:40  [app] [I] processing report for lenovolobato.lobatolan.home
2017-08-03T15:14:40  [app] [D] Report: {"reported_at"=>"2017-08-03 17:14:39 594274", "metrics"=>{"time"=>{"total"=>0, "_aj_hash_with_indifferent_access"=>true}, "_aj_hash_with_indifferent_access"=>true}, "host"=>"lenovolobato.lobatolan.home", "status"=>{"applied"=>0, "failed"=>1, "skipped"=>0, "_aj_hash_with_indifferent_access"=>true}, "logs"=>[{"log"=>{"sources"=>{"source"=>"Gathering Facts", "_aj_hash_with_indifferent_access"=>true}, "messages"=>{"message"=>"{\"msg\": \"Failed to connect to the host via ssh: ssh: connect to host lenovolobato.lobatolan.home port 22: Connection refused\\r\\n\", \"unreachable\": true, \"changed\": false}", "_aj_hash_with_indifferent_access"=>true}, "level"=>"info", "_aj_hash_with_indifferent_access"=>true}, "_aj_hash_with_indifferent_access"=>true}], "_aj_hash_with_indifferent_access"=>true}
2017-08-03T15:14:40 4985e809 [dynflow] [D] ExecutionPlan d4d5e0f6-2c85-43e0-bf87-cb28d7d24bda      pending >>  planning
2017-08-03T15:14:40  [permissions] [D] verifying the transaction by permission edit_hosts for class Host::Managed
2017-08-03T15:14:40  [app] [I] Performed ImportConfigReport from Dynflow(default) in 161.41ms
```

### DIFF
--- a/app/controllers/api/v2/config_reports_controller.rb
+++ b/app/controllers/api/v2/config_reports_controller.rb
@@ -37,10 +37,8 @@ module Api
       param_group :config_report, :as => :create
 
       def create
-        @config_report = ConfigReport.import(params[:config_report], detected_proxy.try(:id))
-        process_response @config_report.errors.empty?
-      rescue ::Foreman::Exception => e
-        render_message(e.to_s, :status => :unprocessable_entity)
+        job = ImportConfigReport.perform_later(params[:config_report], detected_proxy.try(:id))
+        process_success _('Report import has been enqueued as job %s') % job.job_id, :accepted
       end
 
       api :DELETE, "/config_reports/:id/", N_("Delete a report")
--- a/app/jobs/application_job.rb
+++ b/app/jobs/application_job.rb
@@ -1,0 +1,2 @@
+class ApplicationJob < ActiveJob::Base
+end
--- a/app/jobs/import_config_report.rb
+++ b/app/jobs/import_config_report.rb
@@ -1,0 +1,9 @@
+class ImportConfigReport < ApplicationJob
+  def perform(config_report, proxy)
+    ConfigReport.import(config_report, proxy)
+  end
+
+  rescue_from ::Foreman::Exception do |e|
+    logger.warn("Error importing report #{e} - #{self.job_id}")
+  end
+end
--- a/test/controllers/api/v2/reports_controller_test.rb
+++ b/test/controllers/api/v2/reports_controller_test.rb
@@ -3,6 +3,7 @@ require 'controllers/shared/report_host_permissions_test'
 
 class Api::V2::ReportsControllerTest < ActionController::TestCase
   include ::ReportHostPermissionsTest
+  include ::ActiveJob::TestHelper
 
   setup do
     Foreman::Deprecation.expects(:api_deprecation_warning).with(regexp_matches(%r{/reports were moved to /config_reports}))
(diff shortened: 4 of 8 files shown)


app/jobs/import_config_report.rb
```diff
@@ -0,0 +1,9 @@
+class ImportConfigReport < ActiveJob::Base
+  def perform(config_report, proxy)
```
dLobatog: When this runs, `ConfigReport` will print `processing report for lenovolobato.lobatolan.home`, hence why I didn't include a `before_perform` method to log it.

app/jobs/application_job.rb
```diff
@@ -0,0 +1,2 @@
+class ApplicationJob < ActiveJob::Base
```
dLobatog, line 1: Not really needed right now but it could be useful in the future.

test/controllers/api/v2/reports_controller_test.rb
```diff
@@ -19,23 +20,28 @@ def create_a_puppet_transaction_report
 
     def test_create_valid
       User.current=nil
-      post :create, {:report => create_a_puppet_transaction_report }, set_session_user
-      assert_response :success
+      assert_enqueued_with(:job => ImportConfigReport) do
```
dLobatog, line 23: Just a demonstration of how to test that jobs are enqueued

lzap:
We should start actively monitoring queue length, this can easily grow.

app/controllers/api/v2/config_reports_controller.rb
```diff
@@ -37,10 +37,8 @@ def show
       param_group :config_report, :as => :create
 
       def create
-        @config_report = ConfigReport.import(params[:config_report], detected_proxy.try(:id))
-        process_response @config_report.errors.empty?
-      rescue ::Foreman::Exception => e
-        render_message(e.to_s, :status => :unprocessable_entity)
+        job = ImportConfigReport.perform_later(params[:config_report], detected_proxy.try(:id))
```
ohadlevy, line 40: is there any way to bypass rails from parsing the json? in other words, I would prefer the controller to not create report objects at all, and pass this as a blob. /cc @ShimShtein ?

app/jobs/import_config_report.rb
```diff
@@ -0,0 +1,9 @@
+class ImportConfigReport < ApplicationJob
+  def perform(config_report, proxy)
+    ConfigReport.import(config_report, proxy)
+  end
+
+  rescue_from ::Foreman::Exception do |e|
+    logger.info("Error importing report #{e} - #{self.job_id}")
```
ohadlevy: why info? should be at least warn?

app/controllers/api/v2/config_reports_controller.rb
ShimShtein: > is there any way to bypass rails from parsing the json?

I have done it in https://github.com/Katello/katello/pull/6640. In rails 4, we will need to reuse the [param_parser_wrapper](https://github.com/Katello/katello/blob/master/lib/katello/params_parser_wrapper.rb) from Katello and register it properly in the middleware.

dLobatog (commented):
Updated the job message to `warn`. 

About avoiding the parsing of params like Katello does (Candlepin requests are usually way larger than reports), it sounds like premature optimization to me and it's incompatible with Rails 5 middleware so far. https://github.com/Katello/katello/pull/6875

Notice the controller is just parsing config reports JSON just like always, it's not creating new objects or anything. In fact the actual analysis of the JSON is left to Dynflow.

cc @ohadlevy @ShimShtein

ohadlevy:
> Notice the controller is just parsing config reports JSON just like always, it's not creating new objects or anything. In fact the actual analysis of the JSON is left to Dynflow.

I dont think thats true, dynflow serialize it again to the db (when submitting the job) and the worker deserialize it again.
so it actually can increase compute / memory requirements rather then reduce it.

ShimShtein:
> Notice the controller is just parsing config reports JSON just like always, it's not creating new objects or anything.

@dLobatog, Creating hash structure from a long string means creation of string object for each key, and another object for each value. Value object can be either string (most of them), int or hash. For reports it results in a huge delta of objects which puts strain on GC mechanism. Then you are re-serializing hash structure to a single string again (yet another strain on the CPU and possibly a duplicate string in memory).
If you don't really need parsed values in the rails app, it seems to me wasteful to deserialize and then serialize those values again in the web app.

timogoebel:
@dLobatog: This needs a rebase. Setting to WoC.

PR #4718, opened 2017-08-03
Fixes #20500 - Allow GET /api/hostgroups to include parameters in results

treydock:


dLobatog (changes requested):
Great work here @treydock , I don't have any major comments. Could you just add `@parameters = true` to the `update` and `create` methods? Those ones return a Hostgroup object too, which should contain parameters. Also this way both Host and Host groups API are more consistent.

Thanks!

treydock:
@dLobatog Changes made

dLobatog (approved):

dLobatog:
Thank you @treydock !

PR #4719, opened 2017-08-05
Fixes #20511 - Webpack config should look for nested deps

dLobatog:
The current value of resolve.modules in config/webpack.config.js is
"path.join....'node_modules'"".
This value doesn't work when dependencies have been installed using
global style (npm install --global-style true).

In this style, which is what foreman-packaging uses, dependencies of
dependencies get nested like:

node_modules/dependencyA/node_modules/dependencyB

in a tree like structure. foreman-packaging needs to install in this
mode because the npm cache is stored
like a tree.

The solution I've been given in #webpack IRC is to use a string
'node_modules' instead of path.join. The reason is that
webpack purposefully looks for nested dependencies in that case, but not
if you provide the full path. It works for any string, e.g: "browserify" will look for "browserify"/dep1/"browserify"/dep2/"browserify"....

### DIFF
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -50,7 +50,7 @@ var config = {
   resolve: {
     modules: [
       path.join(__dirname, '..', 'webpack'),
-      path.join(__dirname, '..', 'node_modules')
+      'node_modules/'
     ],
     alias: {
       foremanReact:
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "stylelint": "^7.13.0",
     "stylelint-config-standard": "^16.0.0",
     "url-loader": "^0.5.7",
-    "webpack": "^3.0.0",
+    "webpack": "^3.4.1",
     "webpack-dev-server": "^2.5.1"
   },
   "optionalDependencies": {


package.json
```diff
@@ -39,7 +39,7 @@
     "stylelint": "^7.13.0",
     "stylelint-config-standard": "^16.0.0",
     "url-loader": "^0.5.7",
-    "webpack": "^3.0.0",
+    "webpack": "^3.4.1",
```
dLobatog, line 42: `extract-text-webpack` 3.0 is not compatible with older webpack versions (https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/581). I found 3.4.1 is able to compile without any other changes.

config/webpack.config.js
```diff
@@ -50,7 +50,7 @@ var config = {
   resolve: {
     modules: [
       path.join(__dirname, '..', 'webpack'),
-      path.join(__dirname, '..', 'node_modules')
+      'node_modules/'
```
ohadlevy, line 53: @dLobatog  Does this breaks plugin using webpack?

dLobatog: @ohadlevy I'm not sure as I don't have any plugin with npm dependencies, It does break Foreman nightlies or even development if installed with `--global-style true` though. 

If you're worried about removing the line, I think it should be fine to leave it at 
```javascript
path.join(__dirname, '..', 'node_modules'),
'node_modules/'
```
 and have both lines. However this would also NOT work with plugins that have nested dependencies as it'd only look in the first level deps.

dLobatog (commented):
To reproduce:

```
rm -r node_modules
npm install --global-style true
export NODE_ENV=production
node_modules/webpack/bin/webpack.js --bail --config config/webpack.config.js
```

you should see http://koji.katello.org/kojifiles/work/tasks/6666/16666/build.log or similar.

Notice that fixing this is also important for development, not just packaging. 

NPM 3 will put things in the top-level dependencies whenever it can, as soon as it finds it impossible (2 modules depending on incompatible versions of a dependency) it will nest them - and the old webpack config will not figure it out without this patch.

https://docs.npmjs.com/how-npm-works/npm3 for more info.

ehelms (approved):

ares:
Fixes the problem for me, thanks @dLobatog, merged as 8822b551cf934dc25fbc4d0052d3afed602af327

PR #4720, opened 2017-08-06
Fixes #20452 - Keep password value after validation in ovirt CR

orrabin:
Every time a validation error appears the form refreshes and the password is deleted.
Fixing the validation without entering the password again and trying to save will call `update_available_operating_systems` which will try to connect to ovirt using the empty password.
As a result ovirt returns unauthorized.
With this change the password is kept and there is no need to enter it again if there are validation errors.


dLobatog (approved):

dLobatog:
Thanks @orrabin !

PR #4724, opened 2017-08-07
Fixes #18736 - Forced unlimited check for filter with no tax

dhlavac:


app/models/role.rb
```diff
@@ -232,7 +232,7 @@ def self.override_search_operator(operator)
   private
 
   def sync_inheriting_filters
-    self.filters.where(:override => false).each { |f| f.inherit_taxonomies! }
+    self.filters.where(:override => false).each { |f|  f.save }
```
ares: Seems like extra space before f.

ares (changes requested):
Would you mind adding the test for the original issue that updates to role, verifying the filter was updated too?

dhlavac:
@ares Thank you  for review - Changes done please look at that

dhlavac:
[test]

app/models/role.rb
```diff
@@ -232,7 +232,11 @@ def self.override_search_operator(operator)
   private
 
   def sync_inheriting_filters
-    self.filters.where(:override => false).each { |f| f.inherit_taxonomies! }
+    filters_check = []
+    self.filters.where(:override => false).each { |f| filters_check.push(f.save) }
+    errors.add :base, N_('Some of associated filters is invalid which prevented the role to be saved') if filters_check.include? false
+    errors.empty?
+    raise ActiveRecord::Rollback, N_("Unable to submit role: Problem with associated filter") if filters_check.include? false
```
dLobatog: How about showing the error by finding the filter that failed then printing `#{filter.errors}`?

test/models/role_test.rb
```diff
@@ -315,6 +315,19 @@ def test_allow_to_modify_roles_when_bypass_locking
...
+      it 'should forced unlimited check if role or filter have no taxonomies' do
+        @role.organizations = [ @org1 ]
+        @role.save
+        @filter_with_org.reload
+        assert @role.filters.last.unlimited?
+        refute @role.filters.first.unlimited?
+        @role.organizations = []
+        @role.save
+        @filter_with_org.reload
+        assert @role.filters.last.unlimited?
```
dLobatog: Not sure if the assertion for `.last` means anything?

app/models/role.rb
```diff
@@ -232,7 +232,11 @@ def self.override_search_operator(operator)
   private
 
   def sync_inheriting_filters
-    self.filters.where(:override => false).each { |f| f.inherit_taxonomies! }
+    filters_check = []
+    self.filters.where(:override => false).each { |f| filters_check.push(f.save) }
+    errors.add :base, N_('Some of associated filters is invalid which prevented the role to be saved') if filters_check.include? false
```
dLobatog: `One or more of the associated filters are invalid: #{error}` - it's more similar to other errors in Foreman and also gives a hint to the user about what to do.

dLobatog (changes requested):
Thanks @dhlavac - left a few comments

test/models/role_test.rb
dhlavac: In this case when we have 2 filters in role  I only test if something happend with filter with unlimited check by default - after save .. e.g. unlimited check unchcecked

dhlavac:
@dLobatog Thanks for review changes done and left one comment please look at that

dhlavac:
@dLobatog @ares Changes done.

test/models/role_test.rb
ares: I think you could use `@filter_without_org` and `@filter_with_org` variables that are defined above. That deduplicates the logic.

app/models/role.rb
```diff
@@ -232,7 +232,13 @@ def self.override_search_operator(operator)
   private
 
   def sync_inheriting_filters
-    self.filters.where(:override => false).each { |f| f.inherit_taxonomies! }
+    self.filters.where(:override => false).each do |f|
+      unless f.save
+        errors.add :base, N_('One or more of the associated filters are invalid which prevented the role to be saved')
+        errors.empty?
```
ares: What does this line do? It seems like it has no effect?

ares (changes requested):
While the added test is good to have, it does not cover the reason for the fix I think. Could you please add a another test where a filter would be invalid and it would check that role was not saved (save was rollbacked) and contains the validation error? Something like

```ruby
refute role_with_invalid_filters.save
asser rile_with_invalid_filter.errors[:base].includes "...filters are invalid..."
```

app/models/role.rb
dhlavac: How can I create invalid filter ? -> filter with no permissions doesnt work

dhlavac:
@ares How can I create invalid filter ? -> filter with no permissions doesn't work

ares:
> @ares How can I create invalid filter ? -> filter with no permissions doesn't work

looking at filter model, there are several validate definitions that could be broken, e.g. setting search to empty string, assigning 2 permissions each for different resource type, role set to nil, linked role is locked. Some of it should work I hope.

dhlavac:
@ares I cannot find way to create invalid filter

ares:
@dhlavac 

how about this?
```ruby
# if you need unsaved
FactoryGirl.build(:filter, :permissions => Permission.where(:name => ['view_domains', 'view_subnets']))

# if you need persisted
f = $some_filter
f.permissions = Permission.where(:name => ['view_domains', 'view_subnets'])
f.save(:validate => false)
```

dLobatog (approved):
Thanks @dhlavac

PR #4725, opened 2017-08-08
Fixes #20519 - ignored_interface_identifiers are regexps

lzap:
Let me assume we have tests for this.


lzap:
Actually the "wildcard" is correct, we have a replacer for that.

PR #4726, opened 2017-08-08
Fixes #20521 - better subnet prefix label

lzap:
There is a network prefix field which is to input the IP network prefix (ie the short form of the subnet mask /24 instead of 255.255.255.0). 

The description of this field is "Suffix or prefix length for this subnet, e.g. 32" 

Using the term suffix here is incorrect, since it's talking about the network portion of the address. 

Seen on IPv4 and IPv6 pages.

Also, while 32 is probably a bad example, since it is not (or rarely) used as a subnet prefix. 24 is a better example for IPv4 and 64 for IPv6.

### DIFF
--- a/app/views/subnets/_fields.html.erb
+++ b/app/views/subnets/_fields.html.erb
@@ -3,7 +3,7 @@
 
 <%= subnet_type_f f %>
 <%= text_f f, :network, :label => _("Network Address") %>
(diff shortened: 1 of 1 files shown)


lzap:
Just random CI failures.

dLobatog (approved):

dLobatog:
Thanks @lzap

ares:
Setting release 1.16 in redmine...

PR #4728, opened 2017-08-08
Fixes #20529 - Improve the filtering function

xprazak2:


ohadlevy:
[test foreman]

ohadlevy:
[test katello]

xprazak2:
`Capybara::Poltergeist::StatusFailError: Request failed to reach server`, the rest of matrix is green so this seems unrelated to change.

app/assets/javascripts/application.js
```diff
@@ -229,28 +229,32 @@ function filter_by_level(item){
 
   // Note that class names don't map to log level names (label-info == notice)
   if(level == 'info'){
-    $('.label-info').closest('tr').show();
-    $('.label-default').closest('tr').show();
-    $('.label-warning').closest('tr').show();
-    $('.label-danger').closest('tr').show();
+    $('.label-info.result-filter-tag').closest('tr').show();
+    $('.label-default.result-filter-tag').closest('tr').show();
```
kgaikwad: @xprazak2 ,
You have applied a separate class to these label tags in `app/helpers/reports_helper.rb`.
Instead of individual `.label-{tag}.result-filter-tag` selector, only `.result-filter-tag` selector should work here.
i.e $('.result-filter-tag').closest('tr').show();

app/assets/javascripts/application.js
```diff
@@ -229,28 +229,32 @@ function filter_by_level(item){
...
+    $('.label-warning.result-filter-tag').closest('tr').show();
+    $('.label-danger.result-filter-tag').closest('tr').show();
+    $('.label-success.result-filter-tag').closest('tr').show();
   }
   if(level == 'notice'){
-    $('.label-info').closest('tr').show();
-    $('.label-default').closest('tr').hide();
-    $('.label-warning').closest('tr').show();
-    $('.label-danger').closest('tr').show();
+    $('.label-info.result-filter-tag').closest('tr').show();
```
kgaikwad: For this `if-block`, you can use multiple selectors here.
i.e 
```
if(level == 'notice'){
  $('.label-info.result-filter-tag, .label-warning.result-filter-tag, .label-danger.result-filter-tag').closest('tr').show();
  $('.label-default.result-filter-tag, .label-success.result-filter-tag').closest('tr').hide();
}
```
In same way, you can apply for other if-blocks i.e for levels warning/error.

kgaikwad (changes requested):
@xprazak2 , thanks.

Please check for inline corrections.

app/assets/javascripts/application.js
xprazak2: I do not think it would work, the table looks like this:

```html
<table>
  <thead>
    <tr>
      <th>filter by this column</th>
      <th>do not filter by this column</th>
    </tr>
  <thead>
  <tbody>
     <tr>
       <td><span class="label label-info result-filter-tag">value to filter</span></td>
       <td><span class="label label-danger">value I do not filter</span></td>
     </tr>
     <!-- more rows to follow -->
  <tbody>
</table>
```

`label-{tag}` determines the row, but `result-filter-tag` decides the column, so they both work together to filter on one cell in each table row.

xprazak2:
Thanks for review, I improved the style but kept the logic without changes for now.

app/assets/javascripts/application.js
kgaikwad: @xprazak2 ,

Sorry but I am disagree with your explanation that label-{tag} determines the row, but result-filter-tag decides the column. 
When you say $('.label-{tag}.result-filter-tag'), it will give you DOM element which having both these CSS classes i.e `label-{tag}` and `result-filter-tag`.

From shared example, it will give you `span` i.e `<span class="label label-info result-filter-tag">value to filter</span>`

$('.result-filter-tag') will give you all DOM elements which having only `result-filter-tag` CSS class. 
That's why it should work as in your example also it will return `span` element i.e `<span class="label label-info result-filter-tag">value to filter</span>`

xprazak2: > When you say $('.label-{tag}.result-filter-tag'), it will give you DOM element which having both these CSS classes i.e label-{tag} and result-filter-tag.

That is exactly what I want.

Perhaps I should have provided more detailed example as it is not obvious if there is only one row in the table:

```html
<table>
  <thead>
    <tr>
      <th>filter by this column</th>
      <th>do not filter by this column</th>
    </tr>
  <thead>
  <tbody>
     <tr>
       <td><span class="label label-info result-filter-tag">value to filter - Info</span></td>
       <td><span class="label label-success">value I do not filter - Success</span></td>
     </tr>
     <tr>
       <td><span class="label label-warning result-filter-tag">value to filter - Warning</span></td>
       <td><span class="label label-success">value I do not filter - Success</span></td>
     </tr>
     <tr>
       <td><span class="label label-success result-filter-tag">value to filter - Success</span></td>
       <td><span class="label label-warning">value I do not filter - Warning</span></td>
     </tr>
     <!-- more rows to follow -->
  <tbody>
</table>
```

if, for example, `level == 'notice' ` in `filter_by_level` function, then I want to show all the rows that have in the first column label-info, label-warning, or label-danger (first two rows), therefore I want to include them in my selector. Leaving just `$('.result-filter-tag').closest('tr')` will return all the rows, which is not what I want.

ares: Just to provide a background where this example comes from, we hit this in openscap plugin where we have such table and we'd like to use the existing JS function from core. So we need to identify the column (result-filter-tag)

kgaikwad: @xprazak2 ,

You mean that there are some columns of rows which may have ONLY `result-filter-tag ` CSS class but not `label-{tag}` classes i.e label-info, label-warning and 
you want to show only those rows in which first column is having both classes `label-{tag}` and `result-filter-tag `.

That is the case then `$('.label-{tag}.result-filter-tag')` is needed and no modification which suggested by me in code required.

xprazak2: There may be multiple columns in a row that have `label-{tag}`. I want to show only rows that have both `label-{tag}` and `result-filter-tag` classes. I choose on which column with labels to filter by adding `result-filter-tag`.

kgaikwad: LGTM :+1:

ares:
Merged as 14566ba9c87f4bc11d12445c4ab2f313ded9474a, thanks @xprazak2

PR #4729, opened 2017-08-09
Fixes #19529: Use main_app for redirect in case of plugins

ehelms:
Sometimes when entering the require_mail filter, the plugins context
is carried over and the route cannot be matched. Using main_app ensures
that it will resolve the path from the proper context.


dLobatog (approved):

dLobatog:
Thanks @ehelms !

dLobatog:
Cherry-picked to 1.15 as 2ec74f0abbeacb27da82f18ccbc1ec065648f430

PR #4732, opened 2017-08-09
Fixes #20416 - fix UI under email preferences

amirfefer:
Before:
before_notif
After:
notification

I've removed the label, and added placeholder text instead. IMO it looks better, less messy.

### DIFF
--- a/app/assets/stylesheets/users.scss
+++ b/app/assets/stylesheets/users.scss
@@ -1,7 +1,8 @@
 #audit_summary_query_builder {
   display: block;
-  top: -14px;
+  top: -6px;
   left: 20%;
+  width: 92%;
 }
 
 #audit_summary_query_builder::before {
@@ -12,7 +13,7 @@
   border-radius: 0 0 0 10px;
   position: absolute;
   left: -15px;
-  top: -11px;
+  top: -15px;
 }
 
 .mail_notification_query {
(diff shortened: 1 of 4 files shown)


tstrachota:
The code looks good to me and it fixes the problem. 

@rohoover whe we're touching the code, would you mind checking if this is correct way of displaying such information?

If @rohoover approves this layout in general, I have additional proposals on the top of the current change:
- make the filter input narrower and align the right border with the rest of the inputs
- ensure the spacing after input filter is consistent with the rest of the form

Something like this:
screenshot from 2017-08-10 10-59-30

Rohoover:
Looks good. Anyway we can add the inline help text into the help icon with tooltip?

dLobatog (changes requested):
@amirfefer Setting to WoC if you could please move the `help_inline` to `label_help`? There are multiple examples of `label_help` use, the end result should be like https://www.patternfly.org/pattern-library/forms-and-controls/field-level-help/ 

Thanks!

amirfefer:
@Rohoover Is that what you've meant?
email_notification_v2

Rohoover:
Yes! Thank you! 

bitmoji

boaz0 (approved):
LGTM 👍

boaz0:
@dLobatog PTAL

tbrisker (approved):

tbrisker:
Thanks @amirfefer !

PR #4733, opened 2017-08-09
Fixes #20469 - Register react components from plugins

xprazak2:
Plugins can register their own react components so that `MountingService` is aware of them.

```js
//foreman_myplugin/webpack/index.js

import componentRegistry from 'foremanReact/components/componentRegistry';
import MyComponent from './components/MyComponent';
import MyOtherComponent from './components/MyOtherComponent';

componentRegistry.register({ name: 'MyComponent', type: MyComponent });
// store and data attributes are true by default
componentRegistry.register({ name: 'MyOtherComponent', type: MyOtherComponent, store: false, data: false });

// or to register multiple components:
componentRegistry.registerMultiple([
  { name: 'MyComponent', type: MyComponent },
  { name: 'MyOtherComponent', type: MyOtherComponent, store: false, data: false }
]);
```

components from plugins can be imported using an alias, so plugins can use components from other plugins:

```js
// somewhere in foreman_somename/webpack/
import OtherComponent from 'myplugin/OtherComponent';
```

If you are not sure about alias for plugin, run `script/plugin_webpack_directories.rb`

We do not load all js on each page, plugins must make sure their js files are present.
The arguments of helper are ids that plugins use when registering:

```erb
# my_view.erb
<%= webpacked_plugins_js_for :foreman_ansible, :foreman_openscap %>
```

TODO:
* [x] Stylesheets in components are ignored for unknown reason, so Statistics page looks strange
* [x] register multiple components with one call

### DIFF
--- a/app/helpers/reactjs_helper.rb
+++ b/app/helpers/reactjs_helper.rb
@@ -4,4 +4,31 @@ module ReactjsHelper
       "$(tfm.reactMounter.mount('#{name}', '#{selector}', #{data}));".html_safe
     end
   end
+
+  def webpacked_plugins_js_for(*plugin_names)
+    tags = create_tags(plugin_names)
+    return nil if tags.empty?
+    tags.inject(&:concat).html_safe
+  end
+
+  def create_tags(plugin_names)
+    js_tags_for select_requested_plugins(all_webpacked_plugins.map(&:id), plugin_names)
+  end
+
+  def select_requested_plugins(webpacked_plugin_ids, plugin_names)
+    plugin_names.select { |plugin_name| webpacked_plugin_ids.include?(plugin_name) }
+  end
+
+  def all_webpacked_plugins
+    Foreman::Plugin.registered_plugins.values.select do |plugin|
+      File.exist? File.join(plugin.path, 'webpack/index.js')
+    end
+  end
+
+  def js_tags_for(requested_plugins)
+    requested_plugins.map do |plugin|
+      bundle_name = Foreman::Plugin.bundle_name plugin
+      javascript_include_tag(*webpack_asset_paths(bundle_name, :extension => 'js'), "data-turbolinks-track" => true)
+    end
+  end
 end
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -6,7 +6,7 @@ var webpack = require('webpack');
 var StatsWriterPlugin = require("webpack-stats-plugin").StatsWriterPlugin;
 var ExtractTextPlugin = require('extract-text-webpack-plugin');
 var CompressionPlugin = require('compression-webpack-plugin');
-var execSync = require('child_process').execSync;
+var plugins = require('../script/plugin_webpack_directories');
 
 // must match config.webpack.dev_server.port
 var devServerPort = 3808;
@@ -16,18 +16,16 @@ var production =
   process.env.RAILS_ENV === 'production' ||
   process.env.NODE_ENV === 'production';
 
-// If we get multiple lines, then the plugin_webpack_directories.rb script
-// has on the stdout more that just the JSON we want, so we use newline to split and check.
-var sanitizeWebpackDirs = function (pluginDirs) {
-  var splitDirs = pluginDirs.toString().split("\n").reverse();
-
-  return splitDirs.length > 2 ? splitDirs[1] : pluginDirs;
-};
-
-var webpackDirs = execSync(path.join(__dirname, '../script/plugin_webpack_directories.rb'), {
-  stdio: ['pipe', 'pipe', 'ignore']
-});
-var plugins = JSON.parse(sanitizeWebpackDirs(webpackDirs));
+// Create aliases for plugins so that their components are easily accessible.
+// Each alias points to /$path_to_plugin/webpack
+var aliasPlugins  = function (pluginEntries) {
+  var aliases = {};
+  Object.keys(pluginEntries).forEach(function(key) {
+    var pathSplit = pluginEntries[key].split('/');
+    aliases[key] = pathSplit.slice(0, pathSplit.length - 1).join('/');
+  });
+  return aliases;
+}
 
 var config = {
   entry: Object.assign(
@@ -50,13 +48,15 @@ var config = {
   resolve: {
     modules: [
       path.join(__dirname, '..', 'webpack'),
-      'node_modules/'
+      path.join(__dirname, '..', 'node_modules')
     ],
-    alias: {
+    alias: Object.assign({
       foremanReact:
         path.join(__dirname,
-           '../webpack/assets/javascripts/react_app/components')
-    }
+           '../webpack/assets/javascripts/react_app')
+    },
+    aliasPlugins(plugins['entries'])
+    )
   },
 
   module: {
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "webpack-stats-plugin": "^0.1.5"
   },
   "scripts": {
-    "lint": "./node_modules/.bin/eslint -c .eslintrc webpack/ || exit 0",
+    "lint": "./node_modules/.bin/eslint -c .eslintrc webpack/ $(./script/foreman_plugins_eslint.js) || exit 0",
     "test": "node node_modules/.bin/jest",
     "test:watch": "node node_modules/.bin/jest --watchAll",
     "storybook": "start-storybook -p 6006",
--- a/test/helpers/reactjs_helper_test.rb
+++ b/test/helpers/reactjs_helper_test.rb
@@ -1,0 +1,38 @@
+require 'test_helper'
+
+class ReactjsHelperTest < ActionView::TestCase
+  include ReactjsHelper
+
+  def webpack_asset_paths(bundle_name, opts)
+    ["<script src=\"https://foreman.example.com:3808/webpack/#{bundle_name}.js\"></script>"]
+  end
+
+  setup do
+    Foreman::Plugin.register(:foreman_react) {}
+    Foreman::Plugin.register(:foreman_angular) {}
+    @plugins = [Foreman::Plugin.find(:foreman_react), Foreman::Plugin.find(:foreman_angular)]
+    self.stubs(:all_webpacked_plugins).returns(@plugins)
+  end
+
+  test "should select requested plugins" do
+    plugin_names = [:foreman_react, :foreman_backbone]
+    selected_plugins = select_requested_plugins(all_webpacked_plugins.map(&:id), plugin_names)
+    assert selected_plugins.include?(:foreman_react)
+    assert_equal 1, selected_plugins.size
+  end
+
+  test "should create js tags" do
+    tags = js_tags_for [:foreman_react, :foreman_angular]
+    assert_equal 2, tags.size
+  end
+
+  test "should not create js for plugins without webpacked js" do
+    refute webpacked_plugins_js_for(:foreman_meteor)
+  end
+
+  test "should create js for plugins with webpacked js" do
+    res = webpacked_plugins_js_for(:foreman_react, :foreman_angular)
+    assert res.include?('webpack/react.js')
+    assert res.include?('webpack/angular.js')
+  end
+end
--- a/webpack/assets/javascripts/react_app/components/componentRegistry.js
+++ b/webpack/assets/javascripts/react_app/components/componentRegistry.js
@@ -1,0 +1,66 @@
+import React from 'react';
+import PieChart from './common/charts/PieChart/';
+import StatisticsChartsList from './statistics/StatisticsChartsList';
+import PowerStatus from './hosts/powerStatus/';
+import NotificationContainer from './notifications/';
+import ToastsList from './toastNotifications/';
+import StorageContainer from './hosts/storage/vmware/';
+import forEach from 'lodash/forEach';
+import map from 'lodash/map';
+
+const componentRegistry = {
+  registry: {},
+
+  register({ name = null, type = null, store = true, data = true }) {
+    if (!name || !type) {
+      throw new Error('Component name or type is missing');
+    }
+    if (this.registry[name]) {
+      throw new Error(`Component name already taken: ${name}`);
+    }
+
+    this.registry[name] = { type, store, data };
+    return this.registry;
+  },
+
+  registerMultiple(componentObjs) {
+    return forEach(componentObjs, (obj) => {
+      return this.register(obj);
+    });
+  },
+
+  getComponent(name) {
+    return this.registry[name];
+  },
+
+  registeredComponents() {
+    return map(this.registry, (value, key) => {
+      return key;
+    }).join(', ');
+  },
+
+  markup(name, data, store) {
+    const currentComponent = this.getComponent(name);
+
+    if (!currentComponent) {
+      throw new Error(`Component not found:  ${name} among ${this.registeredComponents()}`);
+    }
+    const ComponentName = currentComponent.type;
+
+    return (<ComponentName data={ currentComponent.data ? data : undefined }
+                           store={ currentComponent.store ? store : undefined } />);
+  }
+};
+
+const coreComponets = [
+  { name: 'PieChart', type: PieChart },
+  { name: 'StatisticsChartsList', type: StatisticsChartsList },
+  { name: 'PowerStatus', type: PowerStatus },
+  { name: 'NotificationContainer', type: NotificationContainer },
+  { name: 'ToastNotifications', type: ToastsList, data: false },
+  { name: 'StorageContainer', type: StorageContainer }
+];
+
+componentRegistry.registerMultiple(coreComponets);
+
+export default componentRegistry;
--- a/app/registries/foreman/plugin.rb
+++ b/app/registries/foreman/plugin.rb
@@ -40,6 +40,7 @@ module Foreman #:nodoc:
       attr_reader   :registered_plugins
       attr_accessor :tests_to_skip
       private :new
+      include Foreman::WebpackAssets
 
       def def_field(*names)
         class_eval do
--- a/app/registries/foreman/webpack_assets.rb
+++ b/app/registries/foreman/webpack_assets.rb
@@ -1,0 +1,15 @@
+module Foreman
+  module WebpackAssets
+    def bundle_name(plugin_name)
+      plugin_name.to_s.gsub(/core$/, '').gsub(/-|_|#{remove_bundle_name_plugins}/,'')
+    end
+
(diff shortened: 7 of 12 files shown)


ohadlevy:
Looking good, I hope to try it soon.
For the js code style I suggest running lint or using something like prettier (with eslint)

config/webpack.config.js
```diff
@@ -52,11 +65,22 @@ var config = {
...
-           '../webpack/assets/javascripts/react_app/components')
-    }
+           '../webpack/assets/javascripts/react_app/components'),
+      foremanReactCommon:
+        path.join(__dirname,
+          '../webpack/assets/javascripts/react_app/common'),
+      foremanReactService:
+        path.join(__dirname,
+          '../webpack/assets/javascripts/services'),
+      foremanNodeModules:
```
xprazak2: This is here because of [packaging-related changes](https://github.com/theforeman/foreman/pull/4719). Plugins need to find foreman's node_modules somehow.

```js
import React from 'foremanNodeModules/react';
```

app/views/layouts/base.html.erb
```diff
@@ -17,6 +17,9 @@
     <%= csrf_meta_tags %>
     <%= javascript_include_tag *webpack_asset_paths('vendor', :extension => 'js'), "data-turbolinks-track" => true %>
     <%= javascript_include_tag *webpack_asset_paths('bundle', :extension => 'js'), "data-turbolinks-track" => true %>
+    <% webpacked_plugins_js_tags.each do |item| %>
+      <%= javascript_include_tag *item, "data-turbolinks-track" => true %>
```
ohadlevy: should this be optional? e.g. maybe not load all js on all pages?

config/webpack.config.js
ohadlevy: does it mean that all plugins need to prefix their requirements to foreman node module? that would explain why in my tests i had a very large vendor file and a very large plugin specific file...I could see this being fairly annoying + hard to use npm packages that include common other libs (e.g. a react-date picker or something like that)

webpack/assets/javascripts/react_app/common/componentRegistry.js
```diff
@@ -0,0 +1,84 @@
...
+import NotificationContainer from '../components/notifications/';
+import ToastsList from '../components/toastNotifications/';
+import StorageContainer from '../components/hosts/storage/vmware/';
+import _ from 'lodash';
+
+let componentRegistry = () => {
+    let registry = {
+      PieChart: {
+        type: PieChart,
+        store: true,
```
ohadlevy: can store and data default to true (so you dry)?

webpack/assets/javascripts/react_app/common/componentRegistry.js
```diff
@@ -0,0 +1,84 @@
...
+        data: false
+      },
+      StorageContainer: {
+        type: StorageContainer,
+        store: true,
+        data: true
+      }
+    };
+
+    let register = (name, component) => {
```
ohadlevy: why let vs const?

webpack/assets/javascripts/react_app/common/componentRegistry.js
```diff
@@ -0,0 +1,84 @@
...
+      return _.map(registry, (value, key) => {
+        return key;
+      }).join(', ');
+    };
+
+    let markup = (name, data, store) => {
+      let currentComponent = getComponent(name);
+
+      if (!currentComponent) {
+        throw new Error(`Component not found:  ${name} among ${registeredComponents()}`);
```
ohadlevy: nitpick: extra whitespace between found: and ${name}

ohadlevy (changes requested):
looks like a great start, I've left a few comments inline.

Other things to consider for this pr:
- [ ] tests
- [ ] ability for a plugin to add customer actions/reducers.
- [ ] reuse the same eslint configuration as core by default.

does a factory (such as described at https://medium.com/@SntsDev/the-factory-pattern-in-js-es6-78f0afad17e9) make any sense in this context?

thanks @xprazak2 !

xprazak2:
Changes:
* I turned the registry into a factory as described in the blog post, `data` and `store` attributes are `true` by default, which made code a bit DRYer.

```js
import MyComponent from './MyComponent';
componentRegistry.register({ name: 'MyComponent', type: MyComponent });
```
`registerMultiple` accepts an array of objects with the same format.

* `npm run lint` lints all the webpacked plugins now as well.
* Including everything into all pages seems a bit too much. It will be therefore responsibility of a plugin to request the js files to be included:

```erb
# my_view.erb
<%= webpacked_plugins_js_for :foreman_ansible, :foreman_openscap %>
```
The arguments of helper are ids that plugins use when registering.

* still missing tests, adding actions and better path resolution

config/webpack.config.js
xprazak2: I agree it is a nuisance, but `import React from 'react';` in plugins does not work and I haven't found a better solution yet. I am not sure what you mean about packages that include common other libs, could you explain please?

ohadlevy:
@xprazak2 interesting to follow https://github.com/ManageIQ/manageiq-ui-classic/pull/1944 as well, as they implement plugability in the redux concept, i think this two (this pr and manageiq) will complete each other.

timogoebel:
We gave this a shot. Works nicely.
@ohadlevy: Anything missing?

config/webpack.config.js
ohadlevy: for example, if I import a 3rd party react component (lets say a date picker) then it code will include
```js
import React from 'react';
```

This will:
a. make us pull another copy of react into the plugin node modules
b. will include another copy of react in the plugin bundle.

ohadlevy: /cc @dLobatog - this is probably a side effect we did not expect.

package.json
```diff
@@ -77,7 +77,7 @@
     "webpack-stats-plugin": "^0.1.5"
   },
   "scripts": {
-    "lint": "./node_modules/.bin/eslint -c .eslintrc webpack/ || exit 0",
+    "lint": "./node_modules/.bin/eslint -c .eslintrc webpack/ $(./script/foreman_plugins_eslint.js) || exit 0",
```
ohadlevy, line 80: nice :) we should do the same for rubocop

waldenraines:
@xprazak2 when I run `plugin_webpack_directories.rb` in this branch I get the following:

```
[vagrant@centos7-devel foreman]$  ./script/plugin_webpack_directories.rb
{"entries":{"":"/home/vagrant/katello/webpack/index.js"},"paths":["/home/vagrant/katello/webpack"]}
```

Notice that I now have no way to reference my bundle.  This is also the case in master so I created #4824 but I wouldn't be opposed to closing that if you want to fix this here since our PRs are going to conflict otherwise.  Thoughts?

xprazak2:
@waldenraines I think you can go ahead with #4824, I'll rebase on top of it.

xprazak2:
Plugins can now register their reducers:

```js
import { injectReducer } from 'foremanReact/redux';
import pluginRootReducer from './pluginReducers';

injectReducer('oscap', pluginRootReducer);
```

which makes the store look like this:


But I imagine we would also want to add more states into the lower levels of state tree (for example under 'hosts'), does that make sense? Or are we fine just adding new states to the top level?

timogoebel:
I'm wondering if we can limit the scope of the PR to what we already have so we can get this into 1.16. We can extend on this later.

waldenraines:
> But I imagine we would also want to add more states into the lower levels of state tree (for example under 'hosts'), does that make sense? Or are we fine just adding new states to the top level?

I could definitely see katello wanting to add states to the hosts tree since a content host is just an extension of a host.  I am fine with this being added as a follow up feature.

waldenraines:
@xprazak2 do you mind rebasing please?

ares:
It will need to wait, Ondra is offline this week.

xprazak2:
I am back! I will clean this up, rebase, add tests and move the reducers into a separate PR as Timo suggested if there are no objections. Extending lower levels of state tree will require additional plumbing.

xprazak2:
I rebased, cleaned up the code and added tests.

ohadlevy:
@waldenraines since you need it, mind giving this a review?

config/webpack.config.js
```diff
@@ -52,11 +50,16 @@ var config = {
...
-    alias: {
+    alias: Object.assign({
       foremanReact:
         path.join(__dirname,
-           '../webpack/assets/javascripts/react_app/components')
-    }
+           '../webpack/assets/javascripts/react_app'),
+      foremanNodeModules:
+        path.join(__dirname,
+          '../node_modules')
```
ohadlevy: :+1: this solves an issue of duplicating code in the bundle! but /cc @dLobatog since he introduced this change in the first place.

if you want to visualize the before / after have a look at https://chrisbateman.github.io/webpack-visualizer/

waldenraines:
> @waldenraines since you need it, mind giving this a review?

I'm in the process of doing this but I am running into some trouble.  I'm sure I'm missing something [here](https://github.com/waldenraines/katello/commit/53d0c90e1df2e4dffbeeed61de89dbf3d5b7c94f) but I can't figure out what.

I keep getting: 

```
Uncaught Error: Cannot find module "foremanReactCommon/componentRegistry"
```

Any help would be appreciated and perhaps I can turn this into a how-to for other plugins.

ohadlevy:
@xprazak2 I tried @waldenraines patch above, and it worked (with some fixes to his code) only after I changed this:
```diff
diff --git a/config/webpack.config.js b/config/webpack.config.js
index fc9198c..204841e 100644
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -47,8 +47,7 @@ var config = {
 
   resolve: {
     modules: [
-      path.join(__dirname, '..', 'webpack'),
-      'node_modules/'
+      path.join(__dirname, '..', 'node_modules')
     ],
     alias: Object.assign({
       foremanReact:
```

waldenraines:
One other thing @ohadlevy and I noticed is that the node modules appear to be included twice, once in `bundle.js` and once in `vendor.js`:

```
14:55:16 webpack.1 |         Asset       Size  Chunks                    Chunk Names
14:55:16 webpack.1 |     bundle.js    8.44 MB       0  [emitted]  [big]  bundle
14:55:16 webpack.1 |    katello.js     2.8 kB       1  [emitted]         katello
14:55:16 webpack.1 |     vendor.js    8.86 MB       2  [emitted]  [big]  vendor
14:55:16 webpack.1 |    bundle.css    63.6 kB       0  [emitted]         bundle
14:55:16 webpack.1 |    vendor.css    16.8 kB       2  [emitted]         vendor
14:55:16 webpack.1 | manifest.json  216 bytes          [emitted]

```

ohadlevy:
@waldenraines actually with my fix it doesn't, @dLobatog fix in #4719 has introduced it.

ohadlevy:
just FYI - if you want to inspect the content of the bundles you can use https://chrisbateman.github.io/webpack-visualizer/

waldenraines:
> import componentRegistry from 'foremanReactCommon/componentRegistry';

The correct path is `foremanReact/components/componentRegistry`

waldenraines:
@xprazak2 how would I go about adding my own 3rd party libraries?  I require several not included in foreman for the Red Hat Repositories page.

xprazak2:
@waldenraines: sorry, I was moving things around a lot and forgot to update description (fixed now), it is `'foremanReact/components/componentRegistry'`. 

Including additional 3rd party libraries needs looking into, maybe we could use something like [package-json-merge](https://www.npmjs.com/package/package-json-merge).

waldenraines:
> Including additional 3rd party libraries needs looking into, maybe we could use something like package-json-merge.

I wonder if it would it be possible to detect any plugins and automatically include their node_modules.  That would be the ideal solution from my perspective.

danseethaler:
Hey I just wanted to bring up a new feature in React v16 that relates here. It's called `portals` and it allows you to mount react elements to any random DOM outside of your React tree. One of the main advantages of mounting components this way is that the portal still exists within the structure of the React tree even though it's not a child component in the DOM tree.

You can read more about it [here](https://reactjs.org/docs/portals.html).

React v16.0.0 just got released but it seems really stable from the community feedback and their pre-release testing. Any thoughts or concerns on upgrading to React v16? I'm happy to bring this up elsewhere if it needs more exposure/input.

danseethaler:
@waldenraines pointed out that this would be a possible replacement for the mounting service but probably not the component registering which this PR deals with.

ohadlevy:
@danseethaler 

> Hey I just wanted to bring up a new feature in React v16 that relates here. It's called portals and it allows you to mount react elements to any random DOM outside of your React tree. One of the main advantages of mounting components this way is that the portal still exists within the structure of the React tree even though it's not a child component in the DOM tree.
> 
This feature is very useful, I can think of two different usage cases that we've faced before:
1. modals - modal is usually outside of the dom element, in the past we were not able to fully tests modal due to this limitations - see (https://github.com/theforeman/foreman/pull/4807/files#diff-1d227d564e99df79610168ed38edf408R32)
2. using a parent attribute which updates multiple sub elements which are not in the dom, for example in a legacy table (e.g. hosts) we show the power status of the host, each status is inside a ```<td>``` and we mount it seperatley since we had no common parent - I believe this would speed up the page loading time.

> You can read more about it here.
> 
> React v16.0.0 just got released but it seems really stable from the community feedback and their pre-release testing. Any thoughts or concerns on upgrading to React v16? I'm happy to bring this up elsewhere if it needs more exposure/input.

I think it make sense to wait another week or so, as when i gave it a try there were a lot of sub packages that did not release yet, also i heard that enzyme has been updated with some breaking changes, so we might need to spend some time on that.
our code base is compatible with react 15.6.2, which means it should be compatible with 16.0 too.

danseethaler:
Couldn't agree more @ohadlevy! I installed React@16 yesterday and didn't have any problems but there were a lot of packages that had React@15.x dependencies like you mentioned. Hopefully these packages will be updated shortly and we can move forward.

waldenraines:
@xprazak2 where are we with this?  Would love to see this merged soon.

xprazak2:
I fixed the code climate error, is there anything else I should fix? Do you want me to incorporate the change that prevents node modules to appear twice?

xprazak2:
[test katello]

webpack/assets/javascripts/react_app/components/componentRegistry.js
```diff
@@ -0,0 +1,65 @@
...
+    if (this.registry[name]) {
+      throw new Error(`Component name already taken: ${name}`);
+    }
+
+    this.registry[name] = { type, store, data };
+    return this.registry;
+  },
+
+  registerMultiple(componentObjs) {
+    return _.forEach(componentObjs, (obj) => {
```
ohadlevy: maybe just import foreach instead of the entire lodash, e.g.:

```js
import foreach from 'lodash/foreach'

return foreach(componentObjs, (obj) => {)
```

xprazak2:
I now import just the functions instead of the whole lodash

app/helpers/reactjs_helper.rb
```diff
@@ -4,4 +4,31 @@ def mount_react_component(name, selector, data = [])
       "$(tfm.reactMounter.mount('#{name}', '#{selector}', #{data}));".html_safe
     end
   end
+
+  def webpacked_plugins_js_for(*plugin_names)
```
ohadlevy, line 8: @xprazak2 would you mind adding tests for these helpers?

xprazak2:
I added a couple of tests to cover new helpers.

test/helpers/reactjs_helper_test.rb
```diff
@@ -0,0 +1,39 @@
+require 'test_helper'
+
+class ReactjsHelperTest < ActionView::TestCase
+  include ReactjsHelper
+
+  Foreman::Plugin.register(:foreman_react) {}
+  Foreman::Plugin.register(:foreman_angular) {}
+
+  def webpack_asset_paths(bundle_name, opts)
+    ["<script src=\"https://foreman.example.com:3808/webpack/#{bundle_name}.js\"></script>"]
```
ohadlevy: shouldnt this be on the same url as the server in prod/test?

xprazak2: Does it matter? It is just a stub for test, the method comes from  [webpack-rails](https://github.com/mipearson/webpack-rails/blob/master/lib/webpack/rails/helper.rb#L15) and I cannot use it directly in test because it complains about missing manifest.

ohadlevy:
oh, ,right i didn't read it correctly - thanks.

mmoll:
[test katello]

ohadlevy (approved):

config/webpack.config.js
```diff
@@ -52,11 +50,16 @@ var config = {
       path.join(__dirname, '..', 'webpack'),
```
waldenraines, line 50: Do you mind making the change @ohadlevy mentions in https://github.com/theforeman/foreman/pull/4733#issuecomment-332848660?

waldenraines (changes requested):
What is here worked well for us after @ohadlevy's change in https://github.com/theforeman/foreman/pull/4733#issuecomment-332848660 but I don't see that included in the webpack config yet.  Will approve after that is added.

config/webpack.config.js
ohadlevy: in that case :+1: from my side to add those changes. /cc @dLobatog as it seems we would need to find a different approach to package this.

xprazak2:
I replaced `'node_modules/'` with `path.join(__dirname, '..', 'node_modules')`. I also removed the alias for node_modules as it is no longer needed, now we can use 

```js
import React from 'react'
```

in plugins as well.

ohadlevy:
Thanks  @xprazak2!

config/webpack.config.js
ekohl: It did indeed broke packaging which means our nightlies are now broken. Given the current load on the packaging and release team it might be wise to slow down these changes, at least until 1.16 is out.

dLobatog:
This PR removed the fix in http://projects.theforeman.org/issues/20511, which in turn makes Foreman nightlies unable to build again http://koji.katello.org/kojifiles/work/tasks/5312/35312/build.log 'ModuleNotFoundError: Module not found: Error: Can't resolve 'object-assign' in '/usr/lib/node_modules/react/lib''. I'll fix that on another PR.

ohadlevy:
@dlobatog as mentioned earlier in this thread, the packaging solution conflict with our plugin assets requirement (as it duplicate the libraries in the bundle across plugins).

PR #4730, opened 2017-08-09
Fixes #20536 - sprockets3 syntax in plugin_assets.rake

dLobatog:
sprockets 3 deprecates the use of environment.register_engine. We should
instead register a transformer as per the migration guide
(https://github.com/rails/sprockets/blob/master/guides/extending_sprockets.md#registering-all-versions-of-sprockets-in-processors)

PR #4734, opened 2017-08-09
Fixes #20543 - implement asset_path for plugin_assets

dLobatog:
Sprockets runs on its own context in the task plugin_assets.rake since
 #20287. This means that the asset paths are not visible to it
immediately, the way we provide that functionality is by overriding a
method in Sprockets that gets the paths from the Rails app since we're
on Rake context.

This is provided by https://github.com/petebrowne/sprockets-helpers but our use case is so limited I think it's better to just put it there ourselves. 

You will see a deprecation warning, which is fixed by https://github.com/theforeman/foreman/pull/4730

### DIFF
--- a/lib/tasks/plugin_assets.rake
+++ b/lib/tasks/plugin_assets.rake
@@ -77,6 +77,12 @@ task 'plugin:assets:precompile', [:plugin] => [:environment] do |t, args|
           env.append_path path
         end
 
+        env.context_class.class_eval do
+          def asset_path(path, options = {})
+            ActionController::Base.helpers.asset_path(path, options)
+          end
+        end
+
         env.version = [
           'production',
           config.assets.version,


lib/tasks/plugin_assets.rake
```diff
@@ -77,6 +77,12 @@ task 'plugin:assets:precompile', [:plugin] => [:environment] do |t, args|
           env.append_path path
         end
 
+        env.context_class.class_eval do
+          def asset_path(path, options = {})
+            ActionController::Base.helpers.asset_path(path, options)
+          end
+        end
```
ehelms, line 84: Could this be merged with line 57?

ehelms: Oh, I guess probably not even though we are duplicating environment creation logic.

ehelms: What I don't get is, why in development environment we don't have to do this definition.

dLobatog:
[test katello]

boaz0 (commented):
LGTM

ehelms (approved):

ehelms:
Thanks @dLobatog !

ares:
Thanks @ehelms for merging, please also set the release field in redmine after merging PR. I set it to 1.16.0.

PR #4735, opened 2017-08-10
Fixes #17964 - Add list of search fields in api docs

tstrachota:
This PR extends the default apipie views. To achieve that I had to copy all apipie views into the foreman (even those that remain without any change).

The search fields are stored in action's metadata and are exported also into apidoc.json (can be re-used eg. in hammer help).

Example for hosts index action:
screenshot from 2017-08-10 11-26-50


app/views/layouts/apipie/apipie.html.erb
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE html>
```
dLobatog: Looks the same as https://github.com/Apipie/apipie-rails/blob/master/app/views/layouts/apipie/apipie.html.erb, is it necessary to override it?

dLobatog (changes requested):
Thanks @tstrachota , it works well, but shouldn't the Apipie views belong in apipie directly? The only one we're overriding is _metadata.erb, is it possible to rely on https://github.com/Apipie/apipie-rails/tree/master/app/views/apipie/apipies for the rest? It would be easier if we don't have to maintain this dependency, or just maintain as little as possible (_metadata.erb).

tstrachota:
@dLobatog I agree it would be much better to re-use the unchanged views from apipie-rails, but I'm afraid these lines https://github.com/Apipie/apipie-rails/blob/master/lib/tasks/apipie.rake#L101-L116 prevent from inheriting them.

ohadlevy:
@tstrachota thanks for the PR, I would be much happier if we can find a way to fix apipie vs including the view in our code base.

tstrachota:
I removed the unchanged views. The PR now depends on: https://github.com/Apipie/apipie-rails/pull/552
I'll bump the apipie-rails dependency once the corresponding PR is merged.

tstrachota:
Updated, foreman now holds only the modified apipie views.

dLobatog (changes requested):
@tstrachota Thanks, could you rebase to make tests run? 👍 if tests pass

dLobatog (approved):

iNecas:
Katello test failures related

ohadlevy:
@tstrachota thanks again - do you see any way we can add testing in a way that we can find if an search attribute is removed? It would be great to find a way we are not breaking the "search api".

tstrachota:
I think that we can only check whether a controller's index action documents the search field when its corresponding model is scoped search capable, but we cannot ensure that the controller actually accepts and implements the search field.

ohadlevy:
@tstrachota does it make sense to add a json snapshot of all search options? thinking of how jest/enzyme testing add snapshots and compares them later.. would something like that work?

tstrachota:
@ohadlevy I'm not sure I understand. What should the snapshot exactly contain?
The search docs are currently generated from the scoped search definition so when the scoped search changes, api docs are updated automatically.

ohadlevy:
@tstrachota I understand, what I'm trying to catch is if someone change an existing search, (e.g. removed it) and now the "api" is broken.

having a test fail since the old snapshot of the search options are not the same would ensure the developer understand that he is breaking / changing the API.

tstrachota:
@ohadlevy I get it now. The docs snapshot would help in such case. I think it's a good idea to compare the whole docs, not only the search fields. Diff of yaml exports could hopefully give enough info in quite readable format with minimal efforts.

I also think that we should add the same test for output of `hammer full-help`.

waldenraines:
@tstrachota updates on this?

waldenraines:
[test katello]

tstrachota:
[test katello]

tstrachota:
[test]

tstrachota:
[test]

tstrachota:
Katello tests started to fail again after rebase. Failures doesn't seem to be related but it's probably safer to wait until it's back green.

app/controllers/api/base_controller.rb
```diff
@@ -104,6 +104,21 @@ def api_request?
 
     protected
 
+    def self.add_scoped_search_description_for(resource)
```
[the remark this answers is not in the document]
mmoll: @tstrachota what about this? alternatively, please disable this cop

waldenraines:
> Katello tests started to fail again after rebase. Failures doesn't seem to be related but it's probably safer to wait until it's back green.

Should be okay after https://github.com/Katello/katello/pull/7134 is merged.

waldenraines:
[test katello]

app/views/apipie/apipies/_metadata.erb
```diff
@@ -0,0 +1,10 @@
+<% if meta[:search] %>
+  <%= heading(_('Search fields'), h_level) %>
+  <%= render(:partial => "scoped_search", :locals => { :search => meta[:search] } ) %>
+<% end %>
+
+<% other_meta = meta.except(:search) %>
+<% unless other_meta.blank? %>
+  <%= heading(t('apipie.metadata'), h_level) %>
```
ares, line 8: Shouldn't `t('apipie.metadata)` be rather `_('Apipie metadata')`? it seems like I18n gem that we don't use here, it seems to be used on multiple places in other view too. Or does the i18n load the translation from apipie-rails? How can we ensure the set of languages translated there is the same as those we translate? I think our own translations would make more sense.

app/controllers/api/base_controller.rb
```diff
@@ -104,6 +104,21 @@ def api_request?
...
+      search_fields = resource.scoped_search_definition.fields.map do |k,f|
+        info = { :name => k.to_s }
+        if f.complete_value.is_a?(Hash)
+          info[:values] = f.complete_value.keys
+        else
+          # type is unknown for fields that are delegated to external methods
+          # 'string' is a good guess in such cases
+          info[:type] = f.ext_method.nil? ? f.type.to_s : 'string' rescue ''
```
ares: when is the rescue used? and shouldn't it return 'string'?

ares (changes requested):
Except for hound issue, I left two comments. Otherwise seems to work well and is very useful.

app/controllers/api/base_controller.rb
tstrachota: I came across a case when `f.type` was raising exception from somewhere in AR reflections. It appeared with only relation when used with Katello. Since it really isn't a crucial piece of functionality I gave up and added rescue instead of spending hours digging in the code. It's probably some race condition in loading Katello's host extensions.

I decided to return `''` meaning "We don't know what the type is". But I can change it to `'string'` if you find it better.

app/views/apipie/apipies/_metadata.erb
tstrachota: This translation comes from apipie. The templates are rendered from apipie-rails, so it should be translated fine.

tstrachota:
Updated

tstrachota:
@ares ping

ares:
The only concern with copying so many views is that we will forget updating them whenever they change in apipie-rails. @iNecas does that happen often?

iNecas:
It has not happened in last 3 years. There are some proposals for updates, but again: nothing that would happen too often

tstrachota:
@ares ?

ares (approved):
Ok, based on comments above, nothing blocks this to be merged, thanks @tstrachota, nice addition

PR #4736, opened 2017-08-10
Fixes #20255 - add csv export for hostgroups

amirfefer:


app/models/hostgroup.rb
```diff
@@ -206,6 +206,10 @@ def clone(name = "")
     new
   end
 
+  def hosts_count
+    Host::Managed.authorized.where(:hostgroup => id).size
```
ares: The permission needs to be specified as the first argument for `authorized`, otherwise any filter with resource type Host would be considered. In other words, this code would count in also hosts for which user does not have view permission but he/she has delete permission.

test/controllers/hostgroups_controller_test.rb
```diff
@@ -73,6 +73,13 @@ def setup_user(operation, type = 'hostgroups')
     assert_response :success
   end
 
+  test 'csv export works' do
+    host = FactoryGirl.create(:hostgroup, :with_environment, :with_os)
+    get :index, { :format => 'csv', :search => "name = #{host.hostgroup.name}" }, set_session_user
+    assert_response :success
+    assert_equal 2, response.body.lines.size
```
ares: Should the test verify values for specific columns?

ares (changes requested):
Please look at inline comments, looks good in general.

app/controllers/hostgroups_controller.rb
```diff
@@ -103,6 +112,10 @@ def process_hostgroup
     render :partial => "form"
   end
 
+  def csv_columns
+    [:name, :operatingsystem, :environment, :hosts_count, :children_hosts_count]
```
tbrisker: We don't show os or environment in the ui, so we shouldn't export them in the csv either.

test/controllers/hostgroups_controller_test.rb
```diff
@@ -73,6 +73,13 @@ def setup_user(operation, type = 'hostgroups')
     assert_response :success
   end
 
+  test 'csv export works' do
+    host = FactoryGirl.create(:hostgroup, :with_environment, :with_os)
```
tbrisker: nitpick: this is a hostgroup, so the variable shouldn't be called host

tbrisker (changes requested):

amirfefer:
@tbrisker , @ares thanks for the review, could you please take a second look?

ares (approved):

ares:
My comments were addressed, leaving for Tomer to to confirm his. I didn't test the PR.

tbrisker:
[test foreman]

tbrisker (approved):

app/controllers/hostgroups_controller.rb
```diff
@@ -103,6 +112,10 @@ def process_hostgroup
     render :partial => "form"
   end
 
+  def csv_columns
+    [:name, :hosts_count, :children_hosts_count]
```
tbrisker: `:name` only shows the deepest nested hg name, this should use `:title` to get the full name

tbrisker (changes requested):

amirfefer:
@tbrisker thanks, I've fixed it.

tbrisker (approved):

tbrisker:
Thanks @amirfefer !

PR #4722, opened 2017-08-07
Fixes #20514 - Accept dollar sign($) in login names

avylove:
This will allow dollar signs to be valid characters in login names. This facilitates using Active Directory machine SAMaccountNames as valid user accounts.

PR #4737, opened 2017-08-10
Refs #20514 - Add test for $ in user login

dLobatog:
This was recently added in https://github.com/theforeman/foreman/pull/4722 without tests, didn't want to block the PR from being merged as following up with a test was easy.


iNecas:
Test proves itself, thanks @dLobatog

PR #4739, opened 2017-08-11
Fixes #20559 - print rails log for test failures

lzap:
When there is a test failure, it is usually impossible to find the correct log in logs/test.log file for paritular failed tests. This patch automatically adds Rails logs (including ActiveRecord) to STDOUT for every single test failure, so it's visible. Each log transaction starts with "Rails logs for [name of the test] FAILURE" so it's easy to grep it.

This hooks into `ActiveSupport::TestCase`, so make sure you do not use Minitest::Test directly. We do not do that, there was just a one test which was changed.

Remember, if you run into issues on Jenkins, open full console output and search for "Rails logs" to find the logs for failed tests!

### DIFF
--- a/test/controllers/unattended_controller_test.rb
+++ b/test/controllers/unattended_controller_test.rb
@@ -430,7 +430,6 @@ class UnattendedControllerTest < ActionController::TestCase
   test "should return and log error when template not found" do
     @request.env["REMOTE_ADDR"] = @ub_host.ip
     Host::Managed.any_instance.expects(:provisioning_template).returns(nil)
-    Rails.logger.expects(:error).with(regexp_matches(/unable to find.*template/))
     get :host_template, {:kind => 'provision'}
     assert_response :not_found
   end
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -30,6 +30,36 @@ if ActiveRecord::Base.connection.adapter_name == 'PostgreSQL'
   ActiveRecord::Migration.execute "SET CONSTRAINTS ALL DEFERRED;"
 end
 
+module TestCaseRailsLoggerExtensions
+  def before_setup
+    super
+  ensure
+    @_ext_current_buffer = StringIO.new
+    @_ext_old_logger = Rails.logger
+    @_ext_old_ar_logger = ActiveRecord::Base.logger
+    Rails.logger = Foreman::SilencedLogger.new(ActiveSupport::TaggedLogging.new(Logger.new(@_ext_current_buffer)))
+    ActiveRecord::Base.logger = Rails.logger
+  end
+
+  def after_teardown
+    Rails.logger = @_ext_old_logger if @_ext_old_logger
+    ActiveRecord::Base.logger = @_ext_old_ar_logger if @_ext_old_ar_logger
+    if !self.passed?
+      @_ext_current_buffer.close_write
+      STDOUT << "\n\nRails logs for #{self.name} FAILURE:\n"
+      STDOUT << @_ext_current_buffer.string
+    end
+    super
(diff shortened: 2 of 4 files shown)


lzap:
Do not merge, I see bug - when there are many failures, memory grows like crazy. Not sure why this happens, the stringio should be reallocated every time. Looks like the teardown is not being called.

I canceled the CI job for the record.

lzap:
When there was an exception in test's `teardown` method the chain execution was interrupted. Instead hooking into callbacks, I use prepend to make sure logging teardown method is always called.

lzap:
Allright fixed, tested, pushed.

test/test_helper.rb
```diff
@@ -30,6 +30,38 @@
   ActiveRecord::Migration.execute "SET CONSTRAINTS ALL DEFERRED;"
 end
 
+module TestCaseRailsLoggerExtensions
+  def before_setup
+    super
+    @current_buffer = StringIO.new
```
iNecas: `@current_buffer` sounds too generic and can lead to conflicts with tests (if they used current_buffer for something. Could you thing of more specific name?

test/test_helper.rb
```diff
@@ -30,6 +30,38 @@
...
+    @old_ar_logger = ActiveRecord::Base.logger
+    Rails.logger = Foreman::SilencedLogger.new(ActiveSupport::TaggedLogging.new(Logger.new(@current_buffer)))
+    ActiveRecord::Base.logger = Rails.logger
+  end
+
+  def after_teardown
+    Rails.logger = @old_logger if @old_logger
+    ActiveRecord::Base.logger = @old_ar_logger if @old_ar_logger
+    if @current_buffer.nil?
+      Rails.logger.debug "WARNING: Looks like loggging teardown was removed by a test."
```
iNecas: I don't understand this. What situation we might get into this code?

lzap: This was a leftover when I had this hooked into setup/teardown framework. Removed.

test/test_helper.rb
lzap: Good point, renamed.

lzap:
Pass!

boaz0 (approved):
LGTM 👍

test/controllers/unattended_controller_test.rb
```diff
@@ -427,14 +427,6 @@ class UnattendedControllerTest < ActionController::TestCase
     assert_no_match(%r{https://}, @response.body)
   end
 
-  test "should return and log error when template not found" do
-    @request.env["REMOTE_ADDR"] = @ub_host.ip
-    Host::Managed.any_instance.expects(:provisioning_template).returns(nil)
-    Rails.logger.expects(:error).with(regexp_matches(/unable to find.*template/))
-    get :host_template, {:kind => 'provision'}
-    assert_response :not_found
```
iNecas: Sry, but I have to ask, why we are deleting this test? Is it duplicated somewhere else?

lzap: Yeah I did not comment on that, testing logging makes no sense, unless we are writing a logging framework. It also did not work with the new code.

tbrisker: Perhaps just remove the logger expectation and leave the rest of the test in place?

iNecas: :+1: the error check is a valid test

lzap:
Rebased.

iNecas:
Thanks @lzap. Tested and works flawlessly. This is a great addition I hope will help many devels in the hard times of finding the reasons for test failures.

PR #4743, opened 2017-08-14
Fixes #20589 - update login_at when extlogin is used

ares:


snagoor:
@ares Thanks tested and it works great. Attached screenshots of the same in redmine issue [20589](http://projects.theforeman.org/issues/20589)

app/controllers/users_controller.rb
```diff
@@ -150,6 +150,7 @@ def login_user(user)
     session[:original_uri] = nil
     set_current_taxonomies(user, {:session => session})
     TopbarSweeper.expire_cache
+    user.post_successful_login
```
tbrisker, line 153: This is also called from User.try_to_login right now, so might make sense to remove it from there (otherwise it's called twice for regular login)?

ares: good catch, I moved this just before push and didn't realize it's now called twice

ares:
@tbrisker I pushed a new version, could you please take another look?

tbrisker:
@ares looks like test are failing, can you take a look?

ares:
@tbrisker I fixed broken tests, the first unit test tested nothing since the fixture contained last_login_on, the second didn't make sense after the change, I moved it to controller tests where it belongs. It covers both extlogin (which previously didn't work as expected, since the user session was not set) and regular login. Could you please take another look?

ares:
seems like test are now failing on mysql for some reason, I'll try to reproduce locally and fix it

ares:
reproduced, fixed, pushed, we need to compare times converted to int, comparing to floats can lead to such errors

@tbrisker could you take another look? tests should be green now

ares:
Foreman test failures are unrelated.

dLobatog (approved):
Seems to work well here, thanks @ares @tbrisker

PR #4744, opened 2017-08-14
Fixes #20591 - full ovirt cloud-init support

iNecas:
Currently, the cloud-init integration supports only a sub-set of
cloud-init commands. With `custom_script` support, one can pass
arbitrary cloud-init yaml to the managed host and make sure
the provisioned host will get it unchanged (which is not the case
when passing the user data currently, as the rbovirt parses into
ovirt-specific keys)

After the change, we would pass the cloud-init template to the ovirt as
is via `custom_script` ovirt functionality, and it would get to the
cloud-init directly, without any ovirt pre-processing.

Depends on https://github.com/fog/fog/pull/3961

### DIFF
--- a/app/models/compute_resources/foreman/model/ovirt.rb
+++ b/app/models/compute_resources/foreman/model/ovirt.rb
@@ -207,7 +207,7 @@ module Foreman::Model
     def start_vm(uuid)
       vm = find_vm_by_uuid(uuid)
       if vm.comment.include? "cloud-config"
-        vm.start_with_cloudinit(:blocking => true, :user_data => vm.comment)
+        vm.start_with_cloudinit(:blocking => true, :user_data => vm.comment, :use_custom_script => true)
         vm.comment = ''
         vm.save
       else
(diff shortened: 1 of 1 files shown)


orrabin:
Tested with the fog change and this works well.
fog-ovirt change is now merged.

mmoll:
what's the status here? ☝️

orrabin:
[test katello]

orrabin:
[test foreman]

orrabin:
@iNecas could you please rebase this?

mmoll:
@iNecas ping? :)

iNecas:
Sry for delay, rebased.

iNecas:
@orrabin mind having a look? Also, it seems we don't have cloudinit support in oVirt v4 API. It seems we need to pass a VM object to start method, with `initialization` attribute set with `custom_script` http://www.rubydoc.info/gems/ovirt-engine-sdk/OvirtSDK4/Vm#initialization=-instance_method

iNecas:
Tracking the apiv4 cloudinit support in http://projects.theforeman.org/issues/23370

orrabin (approved):
This was already tested with v3 and worked well. I don't think we should keep it open until support is added for v4 since it already has it's own issue open.

orrabin:
Merging this as per my last comment, fixes for v4 will be added on the fog-ovirt side.

orrabin:
Thanks @iNecas !

PR #4745, opened 2017-08-15
Fixes #20608 - Add index for ssh_keys

mmack:
Adds a index :user_id to ssh_keys


timogoebel:
@theforeman-bot ok to test

timogoebel:
LGTM.
Any comments, @tbrisker?

tbrisker:
:+1: once tests are green, but related issue includes several missing indices - should we add them all at once or would it be best to split the issue into several issues and add them one at a time?

timogoebel:
@tbrisker: I'd suggest adding them one by one. Regardless of that: The other missing indices are actually present, but mysql still seems to do a full table scan. @mmack is still investigating.

timogoebel (approved):

timogoebel:
Merged, thanks @mmack.

PR #4746, opened 2017-08-15
Fixes #20610 - Prevent chrome password autofill

tbrisker:


timogoebel (approved):
@tbrisker: This is lovely and works great.

I do see this JS error, but I have a feeling this is unrelated:
image

ohadlevy:
@timogoebel this seems to be a bug somewhere with yarn, using npm instead (remove your node_modules dir first) works correctly.

from some reason yarn pulls the wrong version of datatables.net regardless and npm doesn't.

timogoebel:
@ohadlevy: Thanks. `npm` doesn't work with our corporate proxy. I guess, it's time to call it a day.

ohadlevy:
thanks @timogoebel added https://github.com/yarnpkg/yarn/issues/4176

timogoebel (approved):

timogoebel:
Merged, thanks @tbrisker. Long overdue, this patch. :-)

tbrisker:
@timogoebel Indeed, last time I tried chrome was still trying to be smarter then me. Turns out in Chrome 44 they finally realized there may be cases that don't need to be auto-filled.

PR #4749, opened 2017-08-17
Fixes #20622 - puppet import does not refresh all statuses

lzap:
It appears that puppet reports cause all statuses to be updated:

https://github.com/theforeman/foreman/blob/develop/app/services/report_importer.rb#L42

On a heavily loaded system this could mean this is being called 100s of times per second. This is a fairly slow operation to be doing when not necessary.

Puppet reports should only refresh statuses that are relevant to them.

### DIFF
--- a/app/models/host/managed.rb
+++ b/app/models/host/managed.rb
@@ -734,8 +734,8 @@ class Host::Managed < Host::Base
     save!(:validate => false)
   end
 
-  def refresh_statuses
-    HostStatus.status_registry.each do |status_class|
+  def refresh_statuses(which = HostStatus.status_registry)
+    which.each do |status_class|
       status = get_status(status_class)
       status.refresh! if status.relevant?
     end
--- a/app/services/report_importer.rb
+++ b/app/services/report_importer.rb
@@ -33,13 +33,14 @@ class ReportImporter
   end
 
   def import
-    start_time = Time.now.utc
-    logger.info "processing report for #{name}"
-    logger.debug { "Report: #{raw.inspect}" }
+    start_time = Time.now
+    logger.debug { "Processing report: #{raw.inspect}" }
     create_report_and_logs
     if report.persisted?
-      logger.info("Imported report for #{name} in #{(Time.now.utc - start_time).round(2)} seconds")
-      host.refresh_statuses
+      imported_time = Time.now
+      host.refresh_statuses(statuses_for_refresh)
+      refreshed_time = Time.now
+      logger.info("Imported report for #{name} in #{(imported_time - start_time).round(2)} seconds, status refreshed in #{(refreshed_time - imported_time).round(2)} seconds")
     end
   end
 
@@ -86,6 +87,10 @@ class ReportImporter
 
   def report_status
     raise NotImplementedError
+  end
+
+  def statuses_for_refresh
+    HostStatus.status_registry
   end
 
   def notify_on_report_error(mail_error_state)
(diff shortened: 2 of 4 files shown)


lzap:
@jlsherrill how about that? In Katello we can now override the method same way to only refresh relevant status(es).

app/services/report_importer.rb
```diff
@@ -33,13 +33,14 @@ def initialize(raw, proxy_id = nil)
...
-    logger.info "processing report for #{name}"
-    logger.debug { "Report: #{raw.inspect}" }
+    start_time = Time.now
+    logger.debug { "Processing report: #{raw.inspect}" }
     create_report_and_logs
     if report.persisted?
-      logger.info("Imported report for #{name} in #{(Time.now.utc - start_time).round(2)} seconds")
-      host.refresh_statuses
+      imported_time = Time.now
+      host.refresh_statuses(statuses_for_refresh)
```
lzap, line 41: This line is the only change in this method, the rest are changes in logging so it produces saner output.

app/services/report_importer.rb
```diff
@@ -88,6 +89,10 @@ def report_status
     raise NotImplementedError
   end
 
+  def statuses_for_refresh
+    HostStatus.status_registry
```
tbrisker, line 93: I wonder, is there a way to infer which status changed from the importer class rather then defaulting to refreshing all of them and having to edit all of the importers manually?

app/models/host/managed.rb
```diff
@@ -734,8 +734,8 @@ def refresh_global_status!
     save!(:validate => false)
   end
 
-  def refresh_statuses
-    HostStatus.status_registry.each do |status_class|
+  def refresh_statuses(which = HostStatus.status_registry)
+    which.each do |status_class|
```
tbrisker, line 738: Is there any situation where one report import needs to update multiple statuses?

ares: I think the idea came from [build substatus](https://github.com/theforeman/foreman/blob/develop/app/models/host_status/build_status.rb#L26) where we wanted to flip the state with any other state update. It seems to be displayed correctly but it wouldn't be reflected in DB if it was not updated on this refresh_statuses call, so we wouldn't be able to search for host which token expired.

now when we have active job, we could have a job to verify this state in Time.now + $provisioning_interval

lzap:
Not really sure why all PG tests failed, rebased. This is ready for review.

app/services/report_importer.rb
lzap: If there is, I'd rather prefer separate PR to be honest :-)

lzap:
@tbrisker mind taking a closer look?

dLobatog (commented):
Marked as 1.15.4 - please label stuff the same as Redmine when it should be cherry-picked

tbrisker (approved):
Ok, looks good. Could you please open a PR leveraging this change for other importers?

tbrisker:
Thanks @lzap !

lzap:
Wait @tbrisker which are you referring to? Can't find any.

ares:
I think this would benefit from the change https://github.com/theforeman/foreman_openscap/blob/b328b6d6e012cc49c87f034fa2db4da349d6d7be/app/controllers/api/v2/compliance/arf_reports_controller.rb#L53

updating http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Extending-host-model would be also good

lzap:
Updated docs, filed: http://projects.theforeman.org/issues/21910

PR #4750, opened 2017-08-17
Fixes #20628 - cloned host/hostgroup should copy root pass

amirfefer:
When cloning host/hostgroup, root pass field should be copied too


iNecas (approved):
Tested and works well, tests are failing for probably unrelated reason, but we should wait until it's green

tbrisker:
Thanks @amirfefer !

PR #4751, opened 2017-08-17
fixes #20636 - update babel-core and loader to newer versions

ohadlevy:


bbuckingham:
ACK.  Confirmed that this resolved the same error on my foreman+katello dev environment.  Thanks!

mmoll:
if this requires RPM updates, please submit these also beforehand, so the build pipeline is not blocked for too long

tbrisker:
Thanks @ohadlevy !

PR #4753, opened 2017-08-18
fixes #20645 - stub templates proxy url call

timogoebel:
Should fix one of the random test failures.


timogoebel:
[test foreman]

tbrisker:
Thanks @timogoebel !

PR #4754, opened 2017-08-18
fixes #20646 - clear host_id during interface cloning

timogoebel:
To reproduce: Clone a Host that has an nic identifier set. The device identifier field will show a validation error: "has already been taken".
The Redmine issue has more details.


timogoebel:
[test foreman]

ohadlevy:
thanks @timogoebel !

PR #4755, opened 2017-08-18
fixes #20649 - add vmware WebMKS console

timogoebel:
This adds support for the VMWare WebMKS SDK Console to Foreman. The "old" VNC console is not supported anymore with vSphere 6.5 and greater. This is a suitable replacement. And it works so much better then the VNC console. And it does not need any firewall changes on the esxi server. All in all: This is great news for users.

Unfortunately, this means adding some stuff to the rails asset pipeline. I created a gem asset for this so hound 🐕 stops barking here.

This works by acquiring a ticket from vSphere using rbvmomi. I chose rbvmomi because fog doesn't support this and pure rbvmomi is faster anyways. The ticket can then be used to access the console via a websocket connection. VMWare ships a JavaScript framework to show the console in the browser.

One caveat is, that the esxi servers need to have a valid ssl certificate or the websocket connection fails. This can be worked around by accessing the esxi server via https and adding a browser side exception for the esxi host.
An alternative to this would be proxying the websocket connection through Foreman. But this is impossible with Passenger to my knowledge.

image

### DIFF
--- a/app/models/compute_resources/foreman/model/vmware.rb
+++ b/app/models/compute_resources/foreman/model/vmware.rb
@@ -28,7 +28,8 @@ module Foreman::Model
     def self.supported_display_types
       {
         'vnc' => _('VNC'),
-        'vmrc' => _('VMRC')
+        'vmrc' => _('VMRC'),
+        'webmks' => _('WebMKS')
       }
     end
 
@@ -517,6 +518,8 @@ module Foreman::Model
       case display_type
       when 'vmrc'
         vmrc_console(vm)
+      when 'webmks'
+        webmks_console(vm)
       else
         vnc_console(vm)
       end
@@ -533,6 +536,25 @@ module Foreman::Model
         :name => vm.name,
         :console_url => build_vmrc_uri(server, vm.mo_ref, client.connection.serviceContent.sessionManager.AcquireCloneTicket),
         :type => 'vmrc'
+      }
+    end
+
+    def webmks_console(vm)
+      ticket = vm.acquire_ticket('webmks')
+      host_url = URI::Generic.build(:scheme => 'https',
+                                    :host   => ticket.host,
+                                    :port   => ticket.port).to_s
+      console_url = URI::Generic.build(:scheme => 'wss',
+                                       :host   => ticket.host,
+                                       :port   => ticket.port,
+                                       :path   => "/ticket/#{ticket.ticket}").to_s
+      {
+        :name => vm.name,
+        :host => ticket.host,
+        :extend_content_security_policy => ["wss://#{ticket.host}"],
+        :host_url => host_url,
+        :console_url => console_url,
+        :type => 'webmks'
       }
     end
 
--- a/app/views/hosts/console/webmks.html.erb
+++ b/app/views/hosts/console/webmks.html.erb
@@ -1,0 +1,36 @@
+<% title "#{@console[:name]}" %>
+<% if webmks_assets_provided? %>
+  <%= stylesheet_link_tag '/webmks/css/wmks-all.css' %>
+  <%= javascript_include_tag '/webmks/wmks.min.js' %>
+  <%= javascript_tag("$(document).on('ContentLoad', function() { tfm.webmksConsole.initConsole() });"); %>
+
+  <%= title_actions(
+    button_group(
+      link_to("Ctrl-Alt-Del", "#", :id => "sendCtrlAltDelButton", :class => "btn btn-default", :disabled => true),
+      link_to(_("Full Screen"), "#", :id => "enterFullScreenButton", :class => "btn btn-default", :disabled => true),
+      if @host
+        link_to_if_authorized(_("Back to host"), hash_for_host_path(:id => @host), :title => _("Back to host"), :class => 'btn btn-default')
+      end
+    )
+  ) %>
+
+  <p><%= (_('The VM is running on host %s.') % link_to(@console[:host], @console[:host_url])).html_safe %></p>
+
+  <div style="text-align: center;"><span class="label label-info" id="wmksStatus"><%= _('Connecting...') %></span></div>
+  <div id="wmksError" style="display: none;"></div>
+  <div id="wmksContainer" style="position: absolute; width: 99%; height: 99%; display: none;" data-url="<%= @console[:console_url] %>"></div>
+<% else %>
+  <div class="blank-slate-pf">
+    <div class="blank-slate-pf-icon">
+      <%= icon_text('screen', '', :kind => 'pficon') %>
+    </div>
+    <h1><%= _('WebMKS Console') %></h1>
+    <p><%= _('The WebMKS console is not available because the required libraries are not installed.') %></p>
+    <p><%= link_to _('Learn more about this in the documentation.'), documentation_url("5.2.9VMwareNotes")%></p>
+    <div class="blank-slate-pf-main-action">
+      <% if @host %>
+        <%= link_to_if_authorized(_("Back to host"), hash_for_host_path(:id => @host), :title => _("Back to host"), :class => 'btn btn-primary btn-lg') %>
+      <% end %>
+    </div>
+  </div>
+<% end %>
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "isomorphic-fetch": "^2.2.1",
     "jquery": "~2.2.4",
     "jquery-flot": "~0.8.3",
+    "jquery-ui": "^1.12.1",
     "jquery-ujs": "~1.2.0",
     "jquery.cookie": "~1.4.1",
     "jstz": "~1.0.7",
--- a/webpack/assets/javascripts/foreman_webmks_console.js
+++ b/webpack/assets/javascripts/foreman_webmks_console.js
@@ -1,0 +1,59 @@
+/* eslint-disable no-console, max-len, no-undef */
+import $ from 'jquery';
+import { showSpinner, hideSpinner, iconText } from './foreman_tools';
+
+require('jquery-ui/ui/widgets/dialog');
+
+let wmks;
+
+export function initConsole() {
+  wmks = WMKS.createWMKS('wmksContainer', {})
+    .register(WMKS.CONST.Events.CONNECTION_STATE_CHANGE, (event, data) => {
+      switch (data.state) {
+        case WMKS.CONST.ConnectionState.CONNECTING:
+          console.log('wmks: connection state change: connecting');
+          showSpinner();
+          $('#wmksStatus').show();
+          $('#wmksError').hide();
+          $('#sendCtrlAltDelButton').prop('disabled', 'disabled');
+          $('#enterFullScreenButton').prop('disabled', 'disabled');
+          break;
+        case WMKS.CONST.ConnectionState.CONNECTED:
+          console.log('wmks: connection state change: connected');
+          hideSpinner();
+          $('#wmksContainer').fadeIn();
+          $('#wmksStatus').hide();
+          $('#wmksError').hide();
+          $('#sendCtrlAltDelButton').removeAttr('disabled');
+          $('#enterFullScreenButton').removeAttr('disabled');
+          break;
+        case WMKS.CONST.ConnectionState.DISCONNECTED:
+          console.log('wmks: connection state change: disconnected');
+          hideSpinner();
+          $('#wmksContainer').hide();
+          $('#wmksStatus').hide();
+          $('#sendCtrlAltDelButton').prop('disabled', 'disabled');
+          $('#enterFullScreenButton').prop('disabled', 'disabled');
+          $('#wmksError').html(`<div class="alert alert-danger alert-dismissable">
+            ${iconText('error-circle-o', '', 'pficon')}
+            <button type="button" class="close" data-dismiss="alert" aria-hidden="true">&times;</button>
+            ${__('The console connection was lost.')}
+            ${__('Please check if you can access the hypervisor host via https by clicking the link above.')}
+            </div>`).fadeIn();
+          break;
+        default:
+      }
+    });
+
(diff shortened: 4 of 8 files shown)


bundler.d/assets.rb
```diff
@@ -8,3 +8,4 @@
 gem 'sass-rails', '~> 5.0'
 gem 'spice-html5-rails', '~> 0.1.5'
 gem 'jquery_pwstrength_bootstrap_4', :require => 'jquery_pwstrength_bootstrap'
+gem 'webmks_sdk-rails', :git => 'https://github.com/timogoebel/webmks_sdk-rails.git'
```
ohadlevy: which kind of license is it? are we sure its compatible? 
I know you already did this work, but does it make more sense to package it as an npm vs a gem?

also, this needs to be published and not use git links.

app/assets/javascripts/webmks.js
```diff
@@ -0,0 +1,43 @@
+//= require webmks_sdk/webmks
+//= require jquery.ui.dialog
```
ohadlevy: oh no, we were trying to get rid of jquery.ui :( any chance we can implement this without it?

app/assets/javascripts/webmks.js
```diff
@@ -0,0 +1,43 @@
...
+
+var wmks;
+$(function () {
+  wmks = WMKS.createWMKS('wmksContainer', {}).register(WMKS.CONST.Events.CONNECTION_STATE_CHANGE, function (event, data) {
+    if(data.state == WMKS.CONST.ConnectionState.CONNECTING) {
```
ohadlevy: consider using switch statement?

app/controllers/concerns/foreman/controller/parameters/compute_resource.rb
```diff
@@ -11,6 +11,7 @@ def compute_resource_params_filter
           :password,
           :provider,
           :set_console_password,
+          :console_type,
```
ohadlevy: should we unify the display_type from libvirt and possibly from ovirt too?

app/models/compute_resources/foreman/model/vmware.rb
```diff
@@ -499,15 +499,57 @@ def clone_vm(raw_args)
...
     def console(uuid)
+      return vnc_console(uuid) if console_type == 'vnc'
+      webmks_console(uuid)
+    end
+
+    def vnc_console(uuid)
       vm = find_vm_by_uuid(uuid)
       raise "VM is not running!" unless vm.ready?
-      #TOOD port, password
-      #NOTE this requires the following port to be open on your ESXi FW
```
ohadlevy: isnt that true anymore? (for vnc usage cases that is?)

app/models/compute_resources/foreman/model/vmware.rb
```diff
@@ -499,15 +499,57 @@ def clone_vm(raw_args)
...
       values = { :port => unused_vnc_port(vm.hypervisor), :password => random_password, :enabled => true }
       vm.config_vnc(values)
       WsProxy.start(:host => vm.hypervisor, :host_port => values[:port], :password => values[:password]).merge(:type => 'vnc')
     end
 
+    def webmks_console(uuid)
+      vm = get_raw_vm(uuid)
+      ticket = webmks_ticket_for_raw_vm(vm)
+      host_url = URI::Generic.build(:scheme => 'https',
```
ohadlevy: just double checking, vcenter can't be deployed on http right?

app/models/compute_resources/foreman/model/vmware.rb
```diff
@@ -499,15 +499,57 @@ def clone_vm(raw_args)
...
+                                    :host   => ticket.host,
+                                    :port   => ticket.port).to_s
+      console_url = URI::Generic.build(:scheme => 'wss',
```
ohadlevy: ditto regarding ws://

app/models/compute_resources/foreman/model/vmware.rb
```diff
@@ -499,15 +499,57 @@ def clone_vm(raw_args)
...
+        :host => ticket.host,
+        :extend_content_security_policy => ["wss://#{ticket.host}"],
+        :host_url => host_url,
+        :console_url => console_url,
+        :type => 'webmks'
+      }
+    end
+
+    # requests a webmks ticket for a vm
+    # uses pure rbvmomi to speed up ticket aquisition
```
ohadlevy: TBH I'm not super excited about this one, this has bitten us in the past multiple times.

I understand its easier to write the code directly in foreman vs add another layer in fog, but it later on it comes back as things you didnt think of (e.g. @orrabin was unable to cache ovirt stuff because of similar "feature" as it ended up pulling ssl layers that can't be serialized etc)

app/views/compute_resources/form/_vmware.html.erb
```diff
@@ -4,9 +4,14 @@
 <% datacenters = list_datacenters f.object%>
 <%= selectable_f(f, :datacenter, datacenters, {}, {:label => _('Datacenter'), :help_inline_permanent => load_datacenters_button_f(f, !datacenters.empty?) })%>
 <%= text_f f, :pubkey_hash, :disabled => true, :label => _("Fingerprint"), :size => "col-md-8" %>
+<%= select_f f, :console_type, f.object.supported_console_types,
+  :first, :last, {}, {
+  :label => _('Console Type'),
+  :help_inline => _('VNC consoles are unsupported on VMware ESXi 6.5 and later.')
```
ohadlevy: can we auto detect the version of vmware ? might be useful later on to decide if we need to take an action (e.g. a user upgraded to 6.5 and now we need to migrate settings to webmks?) /cc @orrabin as she does something very similar for ovirt.

ohadlevy (commented):
@timogoebel looks like a great addition! :1st_place_medal: 

I've outlined a few concerns below.

And as usual, I would prefer to use our newer JS tooling vs extending on rails, I will be very happy to suggest / write alternatives if you are willing to go to that path.

Thanks :)

bundler.d/assets.rb
timogoebel: @ohadlevy: I wasn't sure about the license used. That's why I didn't want to push this to rubygems.
I'm no expert in software licenses. How can we make sure, this works? VMWare just has a EULA that you need to accept to download the software. The text is in the gem.

app/assets/javascripts/webmks.js
timogoebel: Sorry, we need the "dialog" component for this.

app/controllers/concerns/foreman/controller/parameters/compute_resource.rb
timogoebel: Yes, makes sense. I didn't see that option. 😊

app/models/compute_resources/foreman/model/vmware.rb
timogoebel: I just don't think these comments will help anybody and wanted to clean that up. "Always leave the campground cleaner than you found it."

app/models/compute_resources/foreman/model/vmware.rb
timogoebel: I guess, it shouldn't be to hard to add this to fog-vpshere.

app/views/compute_resources/form/_vmware.html.erb
timogoebel: Yeah, we can easily find that out. But it needs a connection to vSphere to work.

```
ComputeResource.find_by_name('Example').send(:client).vsphere_rev
=> "6.0"
```

But I think, this is something for http://projects.theforeman.org/issues/17727

timogoebel:
- Console Selection matches the one we use for libvirt
- Opened a PR to add the Ticket acquisition to fog-vsphere: https://github.com/fog/fog-vsphere/pull/103

```ruby
vm = client.servers.get("host.example.com")
vm.acquire_ticket('webmks')
=>   <Fog::Compute::Vsphere::Ticket
    server_id=nil,
    ticket="c37ae2f7811ecf61",
    host="esxi.example.com",
    port=443,
    ssl_thumbprint="3F:63:E1:AA:56:12:4B:44:85:DA:12:FC:BA:40:56:65:0E:30:A1:A8"
  >
```

timogoebel:
I'd just like to update this PR with the current status. It seems like VMware's license does not allow us to ship the WebMKS SDK with Foreman.

What we could support is allow users to download the required files themselves and place them in a directory on the foreman host. This should be rather easy to do. I'll update this PR asap.

cc: @sbernhard

timogoebel:
Moved the code to webpack. The console sdk now has to be extracted manually and is not bundled.

To test make sure your directory structure looks like this:

```
public/webmks
├── css
│   ├── extended-keypad.css
│   ├── main-ui.css
│   ├── trackpad.css
│   └── wmks-all.css
├── img
│   ├── touch_sprite.png
│   └── touch_sprite_feedback.png
└── wmks.min.js
```

app/controllers/hosts_controller.rb
```diff
@@ -342,15 +343,7 @@ def ipmi_boot
...
-    @encrypt = Setting[:websockets_encrypt]
-    render case @console[:type]
-             when 'spice'
-               "hosts/console/spice"
-             when 'vnc'
-               "hosts/console/vnc"
-             else
-               "hosts/console/log"
-           end
+    super
```
ohadlevy: should it be better to alias method chain instead of super?

timogoebel: I don't think so. Isn't that, what `super` was made for? Calling a method "lower" on the inheritance chain? Mind explaining, why you'd prefer `alias_method_chain`?

ohadlevy:
normally super should be used for class inheritances, alias method chain is
more explicit for mixin, but ymmv ,,,

timogoebel:
Rebased. Test failures are unrelated.

@ohadlevy: Anything more to do here?

timogoebel:
Fixed the hound comments.

timogoebel:
Rebased.

app/models/compute_resources/foreman/model/vmware.rb
```diff
@@ -560,12 +582,8 @@ def pubkey_hash=(key)
       attrs[:pubkey_hash] = key
     end
 
-    def associated_host(vm)
-      associate_by("mac", vm.interfaces.map(&:mac))
-    end
-
     def display_type
-      attrs[:display] || 'vmrc'
+      attrs[:display] || 'webmks'
```
ohadlevy: should this be conditional (e.g. only if its avail?)

app/models/compute_resources/foreman/model/vmware.rb
```diff
@@ -576,6 +594,10 @@ def humanized_display_type
       self.class.supported_display_types[display_type]
     end
 
+    def associated_host(vm)
```
ohadlevy: from some reason this shows as it was moved?

app/views/hosts/console/webmks.html.erb
```diff
@@ -0,0 +1,36 @@
+<% title "#{@console[:name]}" %>
+<% if webmks_assets_provided? %>
+  <%= stylesheet_link_tag '/webmks/css/wmks-all.css' %>
+  <%= javascript_include_tag '/webmks/wmks.min.js' %>
+  <%= javascript_tag("$(document).on('ContentLoad', function() { tfm.webmksConsole.initConsole() });"); %>
+
+  <%= title_actions(
+    button_group(
+      link_to("Ctrl-Alt-Del", "#", :id => "sendCtrlAltDelButton", :class => "btn btn-default", :disabled => true),
```
ohadlevy, line 9: extract string?

app/views/hosts/console/webmks.html.erb
```diff
@@ -0,0 +1,36 @@
...
+      link_to(_("Full Screen"), "#", :id => "enterFullScreenButton", :class => "btn btn-default", :disabled => true),
+      if @host
+        link_to_if_authorized(_("Back to host"), hash_for_host_path(:id => @host), :title => _("Back to host"), :class => 'btn btn-default')
+      end
+    )
+  ) %>
+
+  <p><%= (_('The VM is running on host %s.') % link_to(@console[:host], @console[:host_url])).html_safe %></p>
+
+  <div style="text-align: center;"><span class="label label-info" id="wmksStatus"><%= _('Connecting...') %></span></div>
```
ohadlevy, line 19: afair we have a "ca" class which does the same

ohadlevy: does it make sense to add a spinner?

package.json
```diff
@@ -62,6 +62,7 @@
     "isomorphic-fetch": "^2.2.1",
     "jquery": "~2.2.4",
     "jquery-flot": "~0.8.3",
+    "jquery-ui": "^1.12.1",
```
ohadlevy, line 65: noooo :( projects.theforeman.org/issues/21640

app/models/compute_resources/foreman/model/vmware.rb
timogoebel: I messed up the rebase. Fixed now.

webpack/assets/javascripts/foreman_webmks_console.js
```diff
@@ -0,0 +1,59 @@
...
+let wmks;
+
+export function initConsole() {
+  wmks = WMKS.createWMKS('wmksContainer', {})
+    .register(WMKS.CONST.Events.CONNECTION_STATE_CHANGE, (event, data) => {
+      switch (data.state) {
+        case WMKS.CONST.ConnectionState.CONNECTING:
+          console.log('wmks: connection state change: connecting');
+          showSpinner();
+          $('#wmksStatus').show();
```
ohadlevy, line 16: why not do the entire page in react? can't see a lot of sense in this code?

timogoebel:
I'm closing this PR. The webMKS API is a mess and we have vmrc as a working alternative right now.

PR #4757, opened 2017-08-20
fixes #20652 - moves all compute resource javascript related code under webpack.

ohadlevy:
this PR creates a new namespace under tfm, tfm.computeResource
where relevant functions are expoed under the compute resource
namespace, for example:
```tfm.computeResource.ovirt.termplateSelected``` function

Additional common functions are also exposed under the same namespace:
```tfm.computeResource.providerSelected```
 
I left the host form NIC code out of this PR as it actually requires
code changes, changes in this PR are comsetic (mostly linting) and conversion to ES6.

### CODE REVIEW

**Comment by ohadlevy** [comment]
I missed out openstack js code - added now.

**Comment by ohadlevy** [comment]
@orrabin sadly pushed a new version as i detected a bug in the ec2 code..

**Comment by ohadlevy** [comment]
[test foreman]

**Comment by ohadlevy** [comment]
Please verify on all following compute resources, I've added the one's I've verified my self

- [x] oVirt
- [x] libvirt
- [x] EC2
- [x] VMWare
- [ ] OpenStack

/cc @orrabin @dLobatog @timogoebel

**Comment by orrabin** [comment]
ovirt looks good, I found one issue during testing but I see it in develop as well so I'll open a bug for it to be solved in a different pr.

**Comment by ohadlevy** [comment]
@timogoebel would you have a minute to test this? thanks!

**Comment by ohadlevy** [comment]
@dLobatog would you have time to test this against openstack?

File: webpack/assets/javascripts/compute_resource/vmware.js [hunk 1]
```diff
@@ -0,0 +1,54 @@
...
+    success(request) {
+      request.forEach(({ name }) => {
+        $('<option>').text(name).val(name).appendTo(selectbox);
+      });
+      $(selectbox).select2();
+    }
+  });
+}
+
+function storagePodLoad() {
```
timogoebel: This can be removed including `disableDropdown`.

webpack/assets/javascripts/compute_resource/openstack.js
```diff
@@ -0,0 +1,34 @@
...
+      type: 'post',
+      url,
+      data,
+      complete() {
+        hideSpinner();
+      },
+      error(jqXHR, status, error) {
+        $('#scheduler_hint_wrapper').html(
+          // eslint-disable-next-line no-undef
+          Jed.sprintf(__('Error loading scheduler hint filters information: %s'), error)
```
timogoebel, line 25: Where does this come from? Don't we need to include `Jed`?

timogoebel (changes requested):
VMWare works nicely.

webpack/assets/javascripts/compute_resource/openstack.js
ohadlevy: its currently included from the asset pipelne gettext gem. we would need to break it out in another pr.

ohadlevy:
@timogoebel  pr rebased and updated. thanks.

ohadlevy:
anhone aware of any other compute resource that might be effected? /cc @GregSutcliffe and @dLobatog  ?

webpack/assets/javascripts/compute_resource/openstack.js
timogoebel: How does this work for jest tests, then?

ohadlevy: @timogoebel until now, we used mock or similar functions to not actually call the translation functions.

ohadlevy: @timogoebel see https://github.com/webhippie/gettext_i18n_rails_js/blob/master/lib/assets/javascripts/gettext/all.js for more details. IMHO this should not block this PR.

ohadlevy:
bump on this PR? I believe its ready and would prefer to merge it early in the 0.17 process than later..

tbrisker:
I have some mixed feelings about this PR, on one hand it is definitely a step in the right direction, on the other hand I would prefer if we added some tests and took the time to refactor this where it makes sense. I'm afraid the outcome will be the same old code just located somewhere else with a bit of es6 sugar on top, but I do realize that a complete refactor will take much more right now and this change could potentially make the next step easier. 
Additionally, I agree it is best to merge this early in the release cycle so we have time to find all the issues that we don't see now due to lack of tests.

tbrisker:
Also- if we want to make sure we don't break plugins etc we can leave the original functions in place and have them issue a deprecation before calling the new ones.

ohadlevy:
@tbrisker thanks, I share your concerns, but imho this does qualify as a major step in the right direction. the fact that the code has excplict export vs everything is globally available make it easier to consume and refactor.

Testing the host form imho is a major task, and we could probably spend a lot of time discussing it further. all in all, i would prefer to rewrite it than spend a lot of energy testing what we currently have.

I will add deprecation functions, as i did find a couple of functions used across plugins, will that be enough to merge this? or do you have any specific test in mind that make sense adding?

tbrisker (approved):
I agree testing this would be difficult atm, and would possibly be better to do while refactoring, the concern is that other then manual testing we don't know if we didn't break some edge case now, 
After adding deprecations for the existing functions I suppose we can merge this and keep on improving gradually, hoping this won't cause the code to stagnate in webpack instead of in the assets pipeline.
Unless anyone else objects, once the deprecations are in place i'm ok with merging this.

ohadlevy:
@tbrisker pushed with depredations,  I went over 7 compute resource plugins, and all of them were only using the test connection function, all the rest was custom js or no js at all.

another side benefit to using this approach is to avoid turbolinks full page reload since its now part of the bundle.

ohadlevy:
Test failure unrelated

ohadlevy:
@tbrisker anything else needed? thanks.

tbrisker:
@ohadlevy I'll give it another hour or two in case any other reviewer objects, then merge.

timogoebel (approved):

tbrisker:
Thanks @ohadlevy !

PR #4758, opened 2017-08-20
fixes #20653 - ui notifications for hosts with usergroup owner

timogoebel:
@ohadlevy: This is just a test to demonstrate the error. My suggestion would be to introduce an explicit recipient for a notification so that `subject` and `recipient` aren't automatically related anymore.
Maybe we should allow to pass an arbitrary object as audience. So you can write
```ruby
Notification.create!(
   audience: User.current, # or: Host.find_by_name('server1.example.com').owner
)
```

The test failure happening now:
```
NoMethodError: undefined method `all_users' for #&lt;Host::Managed:0x007f89bdbc91e0&gt;
    app/models/notification.rb:49:in `subscriber_ids'
```

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -4,11 +4,13 @@
 # This class' responsibility is, given a notification blueprint, to determine
 # who are the notification recipients
 class Notification < ApplicationRecord
-  AUDIENCE_USER     = 'user'
-  AUDIENCE_GROUP    = 'usergroup'
-  AUDIENCE_TAXONOMY = 'taxonomy'
-  AUDIENCE_GLOBAL   = 'global'
-  AUDIENCE_ADMIN    = 'admin'
+  AUDIENCE_USER      = 'user'
+  AUDIENCE_USERGROUP = 'usergroup'
+  AUDIENCE_GLOBAL    = 'global'
+  AUDIENCE_ADMIN     = 'admin'
+  # calls notification_recipients_ids on the subject to
+  # determine the recipient user ids
+  AUDIENCE_SUBJECT   = 'subject'
 
   belongs_to :notification_blueprint
   belongs_to :initiator, :class_name => User, :foreign_key => 'user_id'
@@ -20,7 +22,7 @@ class Notification < ApplicationRecord
   validates :notification_blueprint, :presence => true
   validates :initiator, :presence => true
   validates :audience, :inclusion => {
-    :in => [AUDIENCE_USER, AUDIENCE_GROUP, AUDIENCE_TAXONOMY,
+    :in => [AUDIENCE_USER, AUDIENCE_USERGROUP, AUDIENCE_SUBJECT,
             AUDIENCE_GLOBAL, AUDIENCE_ADMIN]
   }, :presence => true
   validates :message, :presence => true
@@ -39,14 +41,10 @@ class Notification < ApplicationRecord
     case audience
     when AUDIENCE_GLOBAL
       User.reorder('').pluck(:id)
-    when AUDIENCE_TAXONOMY
-      subject.user_ids.uniq
-    when AUDIENCE_USER
-      [initiator.id]
     when AUDIENCE_ADMIN
       User.unscoped.only_admin.except_hidden.reorder('').uniq.pluck(:id)
-    when AUDIENCE_GROUP
-      subject.all_users.uniq.map(&:id) # This needs to be rewritten in usergroups.
+    else
+      subject.try(:notification_recipients_ids) || []
     end
   end
 
--- a/app/models/taxonomy.rb
+++ b/app/models/taxonomy.rb
@@ -210,6 +210,10 @@ class Taxonomy < ApplicationRecord
     (self.send("#{type.downcase}_parameters".to_sym).authorized(:view_params) + taxonomy_inherited_params_objects.to_a.reverse!).uniq {|param| param.name}
   end
 
+  def notification_recipients_ids
+    self.subtree.flat_map(&:users).map(&:id).uniq
+  end
+
   private
 
   delegate :need_to_be_selected_ids, :selected_ids, :used_and_selected_ids, :mismatches, :missing_ids, :check_for_orphans,
--- a/app/services/ui_notifications/hosts.rb
+++ b/app/services/ui_notifications/hosts.rb
@@ -2,7 +2,7 @@ module UINotifications
   module Hosts
     class Base < UINotifications::Base
       def deliver!
-        if audience.nil? || initiator.nil?
+        unless subject.owner
           logger.warn("Invalid owner for #{subject}, unable to send notifications")
           # add notification for missing owner
           UINotifications::Hosts::MissingOwner.deliver!(subject)
@@ -14,12 +14,7 @@ module UINotifications
       protected
 
       def audience
-        case subject.owner_type
-        when 'User'
-          ::Notification::AUDIENCE_USER
-        when 'Usergroup'
-          ::Notification::AUDIENCE_GROUP
-        end
+        ::Notification::AUDIENCE_SUBJECT if subject.owner
       end
 
       def initiator
--- a/app/services/ui_notifications/hosts/destroy.rb
+++ b/app/services/ui_notifications/hosts/destroy.rb
@@ -10,10 +10,19 @@ module UINotifications
         Notification.create!(
           initiator: initiator,
           audience: audience,
-          # note we do not store the subject, as the object is being deleted.
+          subject: subject.owner, # note we store the host owner, as the host object is being deleted.
           message: StringParser.new(blueprint.message, {subject: subject}),
           notification_blueprint: blueprint
         )
+      end
+
+      def audience
+        case subject.owner_type
+        when 'User'
+          ::Notification::AUDIENCE_USER
+        when 'Usergroup'
+          ::Notification::AUDIENCE_USERGROUP
+        end
       end
 
       def delete_others
--- a/test/models/notification_test.rb
+++ b/test/models/notification_test.rb
@@ -34,48 +34,49 @@ class NotificationTest < ActiveSupport::TestCase
   end
 
   test 'user notifications should subscribe only to itself' do
-    notification = FactoryGirl.build(:notification,
-                                     :audience => Notification::AUDIENCE_USER)
-    assert_equal [notification.initiator.id], notification.subscriber_ids
+    notification = FactoryGirl.create(:notification,
+                                      :subject => User.current,
+                                      :audience => Notification::AUDIENCE_USER)
+    assert_equal [User.current.id], notification.subscriber_ids
   end
 
   test 'usergroup notifications should subscribe to all of its members' do
     group = FactoryGirl.create(:usergroup)
     group.users = FactoryGirl.create_list(:user,25)
     notification = FactoryGirl.build(:notification,
-                                     :audience => Notification::AUDIENCE_GROUP)
+                                     :audience => Notification::AUDIENCE_USERGROUP)
     notification.subject = group
     assert group.all_users.any?
-    assert_equal group.all_users.map(&:id),
-      notification.subscriber_ids
+    assert_equal group.all_users.map(&:id).sort,
+      notification.subscriber_ids.sort
   end
 
   test 'Organization notifications should subscribe to all of its members' do
     org = FactoryGirl.create(:organization)
     org.users = FactoryGirl.create_list(:user,25)
     notification = FactoryGirl.build(:notification,
-                                     :audience => Notification::AUDIENCE_TAXONOMY)
+                                     :audience => Notification::AUDIENCE_SUBJECT)
     notification.subject = org
     assert org.user_ids.any?
-    assert_equal org.user_ids, notification.subscriber_ids
+    assert_equal org.user_ids.sort, notification.subscriber_ids.sort
   end
 
   test 'Location notifications should subscribe to all of its members' do
     loc = FactoryGirl.create(:location)
     loc.users = FactoryGirl.create_list(:user,25)
     notification = FactoryGirl.build(:notification,
-                                     :audience => Notification::AUDIENCE_TAXONOMY)
+                                     :audience => Notification::AUDIENCE_SUBJECT)
     notification.subject = loc
     assert loc.user_ids.any?
-    assert_equal loc.user_ids, notification.subscriber_ids
+    assert_equal loc.user_ids.sort, notification.subscriber_ids.sort
   end
 
   test 'Global notifications should subscribe to all users' do
     notification = FactoryGirl.build(:notification,
                                      :audience => Notification::AUDIENCE_GLOBAL)
     assert User.count > 0
-    assert_equal User.reorder('').pluck(:id),
-      notification.subscriber_ids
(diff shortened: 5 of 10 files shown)


timogoebel:
I went with overwriting the setter for audience so you can pass a specific User or Usergroup to a notification as well.

ohadlevy:
@timogoebel thanks for the PR.

I'm having a hard time following your logic, could you explain exactly what didn't work ? I'm pretty sure we can extend the notification system to be more tuneable / usable - so any improvement is highly welcomed!

timogoebel:
@ohadlevy: I found this while adding ui notifications to expire hosts. But this is actually a bug in core.

Let's start here with the host build notification:
https://github.com/theforeman/foreman/blob/26c7d10a8ecb28d711f673612171eb32a21720f3/app/models/concerns/hostext/ui_notifications.rb#L10
This sets the subject of the notification to  a host object.
https://github.com/theforeman/foreman/blob/develop/app/services/ui_notifications/hosts.rb#L16-L23 sets the audience to `::Notification::AUDIENCE_GROUP` when the host's owner is a usergroup.
https://github.com/theforeman/foreman/blob/develop/app/models/notification.rb#L49 now fails because subject (a host) does not have a method `all_users`.
I'm wondering: Why does the notification object assume, that the subject is a usergroup when the audience is set to usergroup?

ohadlevy:
> https://github.com/theforeman/foreman/blob/develop/app/models/notification.rb#L49 now fails because subject (a host) does not have a method all_users.
> I'm wondering: Why does the notification object assume, that the subject is a usergroup when the audience is set to usergroup?

This is probably the bug? (odd I do remember writing a test for it - will look again in a sec) but i think the thought the idea was simply to do ``` subject.owner.all_users```

timogoebel:
> but i think the thought the idea was simply to do subject.owner.all_users

Yeah, but that fails if the subject is a Usergroup. Actually for that case, there is a test:

https://github.com/theforeman/foreman/blob/develop/test/models/notification_test.rb#L42-L51

I started to write a mapping class for that, but that didn't go pretty well. Maybe we should create a mixin so that we can just ask every model who should be notified? But then again: Not every object has a single user or a usergroup.

timogoebel:
Another approach for this could be to introduce a new kind of audience: "Audience::HostOwner". This would be an easier implementation.
The Audience User would then just be used if `subject` is of kind `user`. Usergroup would only work for `Usergroup`, etc.
Thoughts?

ohadlevy:
can this be extended to other potential roles to? e.g. Org Admin, or |"Application" owner? overall :+1:

timogoebel:
@ohadlevy: Implemented the solution with a new HostOwner audience.

app/services/ui_notifications/hosts/destroy.rb
```diff
@@ -10,12 +10,21 @@ def create
         Notification.create!(
           initiator: initiator,
           audience: audience,
-          # note we do not store the subject, as the object is being deleted.
+          subject: subject.owner, # note we store the host owner, as the host object is being deleted.
```
ohadlevy, line 13: this means we need to depend: delete on the user/usergroup? (not to delete a notification that belongs to a group that is being deleted)?

app/services/ui_notifications/hosts/destroy.rb
```diff
@@ -10,12 +10,21 @@ def create
...
           message: StringParser.new(blueprint.message, {subject: subject}),
           notification_blueprint: blueprint
         )
       end
 
+      def audience
+        case subject.owner_type
```
ohadlevy, line 20: is destroy the only action this should exists at?

app/models/notification.rb
```diff
@@ -4,11 +4,12 @@
...
-  AUDIENCE_GROUP    = 'usergroup'
-  AUDIENCE_TAXONOMY = 'taxonomy'
-  AUDIENCE_GLOBAL   = 'global'
-  AUDIENCE_ADMIN    = 'admin'
+  AUDIENCE_USER       = 'user'
+  AUDIENCE_GROUP      = 'usergroup'
+  AUDIENCE_TAXONOMY   = 'taxonomy'
+  AUDIENCE_GLOBAL     = 'global'
+  AUDIENCE_ADMIN      = 'admin'
+  AUDIENCE_HOST_OWNER = 'host_owner'
```
ohadlevy: does it make sense to keep both host_owner and user/usergroup? I'm trying to think of usage cases I'll notify a group and I can't (besides the host context).

app/services/ui_notifications/hosts/destroy.rb
timogoebel: yes, because for destroy the subject is not passed to the Notification model as the subject is being deleted.

timogoebel:
@ohadlevy: To be honest, I think I don't like this approach either. Maybe we need a fourth attempt.
What do you think of the following: 
The User and Usergroup model get a method `notification_recipients`? The host model gets that method delegated to the owner.
We remove the notification audiences `User`, `Usergroup`, `Taxonomy` and `HostOwner`. If no audience is passed to the `Notification model` we fall back to asking the subject for the recipient IDs.

```ruby
  def subscriber_ids
     case audience
     when AUDIENCE_GLOBAL
       return User.reorder('').pluck(:id)
     when AUDIENCE_ADMIN
        return User.unscoped.only_admin.except_hidden.reorder('').uniq.pluck(:id)
      end
    end
    subject.try(:notification_recipients) || []
  end
```

This doesn't solve the case for the host delete notification. But maybe it's the correct approach to pass the host as subject as well (even though it is deleted). We can decide in the `Notification` model to not store it in the DB if we want to.

ohadlevy:
> We remove the notification audiences User, Usergroup, Taxonomy and HostOwner. If no audience is passed to the Notification model we fall back to asking the subject for the recipient IDs.

can you expand on the taxonomy usage cases you are thinking of? overall +1

timogoebel:
@ohadlevy: Updated.

app/models/taxonomy.rb
```diff
@@ -210,6 +210,10 @@ def params_objects
     (self.send("#{type.downcase}_parameters".to_sym).authorized(:view_params) + taxonomy_inherited_params_objects.to_a.reverse!).uniq {|param| param.name}
   end
 
+  def notification_recipients_ids
+    user_ids.uniq
```
ohadlevy: Just realized we ignore nested orgs

app/models/notification.rb
```diff
@@ -39,14 +39,10 @@ def subscriber_ids
...
-    when AUDIENCE_TAXONOMY
-      subject.user_ids.uniq
-    when AUDIENCE_USER
-      [initiator.id]
     when AUDIENCE_ADMIN
       User.unscoped.only_admin.except_hidden.reorder('').uniq.pluck(:id)
-    when AUDIENCE_GROUP
-      subject.all_users.uniq.map(&:id) # This needs to be rewritten in usergroups.
+    else
+      subject.try(:notification_recipients_ids) || []
```
ohadlevy, line 47: That's nice, in the future can probably be used for host groups too

ohadlevy (commented):
Looks good on first pass, I'll try to spend more time on it today

app/models/taxonomy.rb
timogoebel: I think we should find a proper way to do this inside the DB. Might be hard with MySQL.

ohadlevy:
@dLobatog mind giving this another look? I think I'm OK with the suggested changes but another pair of eyes++ thanks.

app/models/notification.rb
```diff
@@ -4,11 +4,11 @@
...
 # who are the notification recipients
 class Notification < ApplicationRecord
-  AUDIENCE_USER     = 'user'
-  AUDIENCE_GROUP    = 'usergroup'
-  AUDIENCE_TAXONOMY = 'taxonomy'
-  AUDIENCE_GLOBAL   = 'global'
-  AUDIENCE_ADMIN    = 'admin'
+  AUDIENCE_USER      = 'user'
+  AUDIENCE_USERGROUP = 'usergroup'
+  AUDIENCE_SUBJECT   = 'subject'
```
dLobatog: Adding a comment explaining that `AUDIENCE_SUBJECT` works by implementing a method in the subject `notification_recipients_ids` would be nice. 

Long term I would expect to just call a service which figures out what are the 'notification recipients' of an object, to separate concerns better. I don't think it's required just yet since it's just 2 class implementing this, your call. It'd be simpler to understand without the need of adding a comment even. Something like `NotificationRecipientSearcher.new(subject).recipient_ids` would work, and it could even accept parameters.

app/models/notification.rb
```diff
@@ -4,11 +4,11 @@
 # This class' responsibility is, given a notification blueprint, to determine
 # who are the notification recipients
 class Notification < ApplicationRecord
-  AUDIENCE_USER     = 'user'
```
dLobatog, line 7: Many strings in this class would benefit from `# frozen_string_literal: true`, I don't see any that should be mutated/reinstantiated.

app/models/taxonomy.rb
dLobatog: `self.flat_map(&:subtree).flat_map(&:users).uniq` would return what you want

app/services/ui_notifications/hosts.rb
```diff
@@ -14,12 +14,7 @@ def deliver!
       protected
 
       def audience
-        case subject.owner_type
-        when 'User'
-          ::Notification::AUDIENCE_USER
-        when 'Usergroup'
-          ::Notification::AUDIENCE_GROUP
-        end
+        ::Notification::AUDIENCE_SUBJECT if subject.owner
```
dLobatog, line 17: Nice!

dLobatog: Just curious, why `if subject.owner`? Isn't the audience the same and it would just resolve to nothing?

app/services/ui_notifications/hosts/destroy.rb
dLobatog: @ohadlevy @timogoebel  I think it's the opposite, delete the notifications associated with that User/User group when they are deleted. E.g:

1. Host with owner "Ohad" is deleted.
2. Admin tries to remove user "Ohad" - gets a warning "User has notifications, cannot delete this user"
3. Admin logs in as "Ohad", marks all notifications as read
4. Admin logs back in as "admin" then removes "Ohad".

Whereas we can get by with:

1. Host with owner "Ohad" is deleted.
2. Admin removes user "Ohad" - the notifications are deleted too, they are irrelevant now. Neither the host nor the owner are in Foreman anymore. There's an Audit page if the admin wonders what happened with the host created in step 1.

dLobatog (changes requested):
@timogoebel @ohadlevy Checked it out and left a couple of comments. I think the direction is good, but if possible I'd prefer to see it implemented as a service (centralizes logic, anyone who reads the notification class can find the service) rather than implementing a method in each model (less obvious, logic for the same concept - finding notification recipients - is spread all over and mixed with model definition).

app/models/notification.rb
timogoebel: This "magic comment" is already in the file.

app/services/ui_notifications/hosts.rb
timogoebel: This is a little verbose. It's not really necessary but I think, it's the cleaner solution.

app/services/ui_notifications/hosts/destroy.rb
timogoebel: Sorry, I don't follow what this is about.
When a user that has notifications is deleted, the notifications are deleted as well (`dependant :destroy`).
But to be honest: Isn't this totally unrelated to this PR and worth a separate issue if there is still a mistake?

timogoebel:
Updated. Any more todos? I kind of lost track.

app/services/ui_notifications/hosts/destroy.rb
dLobatog: @timogoebel I agree and I think that's the correct behavior - that's the 2nd part of my message ("Whereas we can get by with...")

app/models/taxonomy.rb
```diff
@@ -210,6 +210,10 @@ def params_objects
     (self.send("#{type.downcase}_parameters".to_sym).authorized(:view_params) + taxonomy_inherited_params_objects.to_a.reverse!).uniq {|param| param.name}
   end
 
+  def notification_recipients_ids
+    self.flat_map(&:subtree).flat_map(&:users).uniq
```
dLobatog: @timogoebel I suppose this was supposed to be `subtree.flat_map(&:users).uniq`? Tests are failing because `self.flat_map` won't work as `self` is not a collection.

dLobatog (changes requested):
@timogoebel Left a comment on the Taxonomy model which I think explains why tests fail currently. 👍 to merge after that, thanks !

timogoebel:
@dLobatog: Thanks. Done.

dLobatog (approved):
Great - thanks @timogoebel !

I see `PG::ObjectInUse: ERROR:  database "test_develop_pr_katello-1-test" is being accessed by other users` which made all Katello tests fail, let's rerun  it... [test katello]

dLobatog:
[test katello]

dLobatog (commented):
@timogoebel Sorry - katello tests were failing due to #4826 not being merged. Could you rebase now that is merged so that we make sure tests pass again with this? Thanks

timogoebel:
Rebased.

test/models/notification_test.rb
```diff
@@ -34,16 +34,17 @@ class NotificationTest < ActiveSupport::TestCase
...
 
   test 'usergroup notifications should subscribe to all of its members' do
     group = FactoryGirl.create(:usergroup)
     group.users = FactoryGirl.create_list(:user,25)
     notification = FactoryGirl.build(:notification,
-                                     :audience => Notification::AUDIENCE_GROUP)
+                                     :audience => Notification::AUDIENCE_USERGROUP)
     notification.subject = group
     assert group.all_users.any?
     assert_equal group.all_users.map(&:id),
```
dLobatog: The IDs on these tests - they were returned in the opposite order which made one test fail (http://ci.theforeman.org/job/test_develop_pr_core/14241/database=postgresql,ruby=2.2,slave=fast/testReport/junit/(root)/NotificationTest/test_0005_Organization_notifications_should_subscribe_to_all_of_its_members/). 

Could you call `.sort` on these IDs so that they're returned in the same order? Seems that this could have happened on any of the other tests modified here too.

dLobatog (changes requested):
@timogoebel Gah, sorry but one test failed due to what I explained inline. I'm fine with fixing the tests by calling `.sort`.

Out of curiosity, do you know why would the IDs be returned in the opposite order only on Ruby 2.2/postgres? If it was random I think we would've been more unlucky and it would have failed in more than 1 out of 9, but it could be.

timogoebel:
@dLobatog: Thanks. I didn't see that locally. Added sorting to the tests.

dLobatog (approved):

dLobatog:
Thanks @timogoebel & @ohadlevy  ! 

Merged and cherry-picked to 1.16-stable as d51d3c9b4f2453f5c1713bfa6da38fec2d67d70e

PR #4759, opened 2017-08-21
Fixes #20654 - different color of Failed Restarts in resources chart

iNecas:


iNecas:
Before:
failed-restarts-before

After:
failed-restarts-after

@Rohoover: any inputs on appropriate color selection?

Rohoover:
@iNecas 
My only thought - and this would be something probably better suited for a future state - is how people with color blindness see this.

iNecas:
Thanks @Rohoover, agreed on keeping the scope of this PR just on this one color: the color-blindness would need to have more thoughts on how the colors are chosen unless we already have some scale we could use.

dLobatog (approved):
@iNecas @Rohoover I'm going to tentatively merge this, in the end the change is better than what we had before and the PR went stale. If any of you feel like updating to a color or scale that's suitable for colorblind people (in case this is not), that's even better 👍 

Thanks to you both!

iNecas:
Ty @dLobatog

PR #4761, opened 2017-08-21
Fixes #20660 - improve proxy features list

ares:
before
before1

after
after1

@Rohoover this is based on your suggestion, any comments?

### DIFF
--- a/app/views/smart_proxies/show.html.erb
+++ b/app/views/smart_proxies/show.html.erb
@@ -52,10 +52,26 @@
           <div class="col-md-4">
             <%= _('Active features') %>
           </div>
+
+          <% @smart_proxy.features.order(:name).pluck(:name).in_groups_of(5, false) do |features| %>
+            <div class="col-md-2">
+              <ul class="">
+                <% features.each do |feature| %>
+                  <li class=""><%= h(feature).gsub(' ', '&nbsp;').html_safe %></li>
+                <% end %>
+              </ul>
+            </div>
+          <% end %>
+        </div>
+
+        <div class="row">
+          <div class="col-md-4"></div>
           <div class="col-md-8">
-            <%= @smart_proxy.features.map(&:name).to_sentence %> <%= refresh_proxy_icon(@smart_proxy, authorizer) %>
+            <%= refresh_proxy_button(@smart_proxy, authorizer) %>
           </div>
         </div>
+        <br>
+
         <div class="row">
           <div class="col-md-4">
             <%= _('Hosts managed') %>
--- a/app/helpers/smart_proxies_helper.rb
+++ b/app/helpers/smart_proxies_helper.rb
@@ -60,9 +60,9 @@ module SmartProxiesHelper
     )
   end
 
-  def refresh_proxy_icon(proxy, authorizer)
-    display_link_if_authorized(icon_text("refresh", "", :kind => "pficon"), hash_for_refresh_smart_proxy_path(:id => proxy).
-                                                     merge(:auth_object => proxy, :permission => 'edit_smart_proxies', :authorizer => authorizer), :method => :put)
+  def refresh_proxy_button(proxy, authorizer)
+    display_link_if_authorized('Refresh features', hash_for_refresh_smart_proxy_path(:id => proxy).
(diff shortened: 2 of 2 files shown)


sean797:
TBH I quite like the old way but if we want to list them as a list when maybe we should look at multiple columns? 

I think the way this PR currently is it introduces a lot of empty/dead space 👎

ares:
I tend to agree, multiple columns though might not help readability for features with various length. Let's see what @Rohoover and others think. Also we might sort it alphabetically while we're touching it?

timogoebel:
I do prefer the "old" way as well. Mostly for the same reasons as @sean797.

👎 from me.

Rohoover:
I would recommend alphabetical in multiple columns. Traditionally, when looking for specific items, lists are easier to scan (think Table of Contents, Grocery List, etc.). When they are listed with commas, you are sort of forced to scan the entire word in order to search for the next word. In a list, you can scan and stop (maybe even at the first letter) and move on if it wasn't what you were looking for.

dLobatog (changes requested):
Setting to WoC as per @Rohoover 's comment, I also think it doesn't look better in the original screenshot. The refresh button changed from a icon to a button with text, is that intentional?

ares:
Ok, I'll see what I can do with the multi columns list

> The refresh button changed from a icon to a button with text, is that intentional?

Yes, I combined these two to avoid rebasing, that was another recommendation of @Rohoover

ares:
new version
after1

@sean797 @timogoebel @Rohoover @dLobatog what do you think?

Rohoover:
@ares 
Can we keep the button always as an element on it's own row beneath the bulleted columns?

I like the multiple columns -- uses the space better.

ares:
@Rohoover like this?
after2

Rohoover:
bitmoji

app/views/smart_proxies/show.html.erb
```diff
@@ -52,10 +52,27 @@
...
             <%= _('Active features') %>
           </div>
+
+
+            <% @smart_proxy.features.order(:name).pluck(:name).in_groups_of(5, false) do |features| %>
+              <div class="col-md-2">
+                <ul class="">
+                  <% features.each do |feature| %>
+                    <li class=""><%= h(feature).gsub(' ', '&nbsp;').html_safe %></li>
+                <% end %>
```
sean797: Indentation is off here

app/views/smart_proxies/show.html.erb
```diff
@@ -52,10 +52,27 @@
...
+
+        <div class="row">
+          <div class="col-md-4"></div>
           <div class="col-md-8">
-            <%= @smart_proxy.features.map(&:name).to_sentence %> <%= refresh_proxy_icon(@smart_proxy, authorizer) %>
+            <%= refresh_proxy_button(@smart_proxy, authorizer) %>
           </div>
         </div>
+
+        <br>
```
sean797: Whats this for?

sean797 (changes requested):

sean797:
2 small comments. Looks much better! 👍

app/views/smart_proxies/show.html.erb
ares: left over from various versions, thanks!

ares:
fixed both, thanks @sean797

sean797 (approved):
thanks @ares

timogoebel:
I think this looks a little squeezed. The new refresh features button needs more space to the number of managed hosts.

image

dLobatog (approved):

dLobatog (changes requested):
@ares Setting to WoC based on @timogoebel 's comment, take a look at that if you can - sounds like some padding is required or similar

app/views/smart_proxies/show.html.erb
ares: so the `<br>` was actually necessary for what @timogoebel reported below

ares:
I reverted to new line, there's now more space again like in the last of my screenshots, any more comments?

dLobatog (approved):

dLobatog:
Thanks @ares & everybody else involved! 😄

PR #4763, opened 2017-08-21
Fixes #20665 - make smart proxy delete button grey

ares:
before:
before
after:
after

### DIFF
--- a/app/helpers/smart_proxies_helper.rb
+++ b/app/helpers/smart_proxies_helper.rb
@@ -55,7 +55,7 @@ module SmartProxiesHelper
       ),
       button_group(
         display_delete_if_authorized(hash_for_smart_proxy_path(:id => proxy).merge(:auth_object => proxy, :authorizer => authorizer),
-                                     :data => {:confirm => _("Delete %s?") % proxy.name}, :class => 'btn btn-danger')
+                                     :data => {:confirm => _("Delete %s?") % proxy.name}, :class => 'btn btn-default')
       )
     )
   end


tbrisker:
I have to say, this seems inconsistent with other delete buttons which are red (or sometimes only red font). Shouldn't we have just one style for delete everywhere?

ares:
> Shouldn't we have just one style for delete everywhere?

yes, and most likely it should be grey unless it is really dangerous. The red label should be removed by https://github.com/theforeman/foreman/pull/4771, on list of smart proxies the delete in table is not red.

I could search for all `btn-danger` in separate PR, but I wanted to concentrate on smart proxy pages first. I think it's in comply with what @Rohoover says about red buttons. That color should only be used for potentially destructive actions.

Rohoover:
@ares 
Correct! 
bitmoji

ohadlevy:
When having content on a proxy/capsule, won't this actually remove the content?

ares:
I didn't see such task definition for proxy but it might be caused by [dependent destroy on capsule lifecycle env](https://github.com/Katello/katello/blob/master/app/models/katello/concerns/smart_proxy_extensions.rb#L31), @bbuckingham could you please clarify or recommend someone who could? The question is, when smart proxy is deleted, does it also delete some data, e.g. content, outside of Foreman DB? That would probably qualify for keeping the Delete button red.

dLobatog (commented):
Setting to WoC - I also believe it should stay red if it removes synced content

bbuckingham:
@ares, I do not see anything in the app code that will affect the content on the smart proxy or other backend services on the foreman, when a proxy with associated lifecycle environments is deleted.  

@johnpmitsch or @jlsherrill, Is there anything that you are aware of that we may be missing?  (e.g. is there a task, script...etc that will delete content outside of foreman when a smart proxy is deleted?)

jlsherrill:
Nope, if you delete a smart proxy with pulp content, the content just sits there on the server.  (In fact clients could still fetch content from it).   It will stay there until you were to re-register the smart proxy and attempt to sync it, at which point any content that doesn't belong on it will be deleted.

ares:
Ok thanks, @jlsherrill my next question is, how complicated/painful/feasible it is to restore the proxy object after it was deleted so it would not delete data during next sync? If there's no easy way, I'm inclined to keep this red. OTOH we should somehow explain user why this is red. Maybe some confirmation dialog with generic messaging about plugins...

jlsherrill:
@ares well, if you delete the smart proxy, there's no way to start a sync.  You would simply, re-run the installer on the smart proxy, or re-create it in the UI, associate the proper lifecycle environments to it, and then run the sync.

If you forgot to associate the correct lifecycle environments and did a sync, then it would delete all the content.  Its not super likely.

ares:
@ohadlevy @Rohoover based on the comment from @jlsherrill, do you think this qualifies for red? I'm not sure but I lean towards making it grey.

Rohoover:
@ares 
I'm leaning towards grey because the action is essentially fixable or reversible. Red to me so something that is permanent in nature with broad implications.

That said, it may be worth putting in some inline messaging somewhere in the flow that tells them these things are happening.

ares:
Thanks, resetting flags then

dLobatog:
Thanks @ares / @Rohoover !

PR #4766, opened 2017-08-21
Fixes #20669 - Append 'clone' to cloned provisioning templates

bastilian:


ares:
Looks good and works fine. One observation during the testing that I was uncertain about is that we display warning that marked fields must be reviewed. If I'm cloning existing template, I suppose it's valid so a clone is now also valid. Should we remove the warning?

warn

bastilian:
@ares The warning is now more confusing than helpful. I've removed it now.

ares:
Thanks @bastilian, works great. Merged as 40c7655100142cf42ce00d537894bfaccce32904

PR #4772, opened 2017-08-21
Fixes #20676 - relabel new button for autosign entry

ares:
before
before3
after
after4

### DIFF
--- a/app/helpers/autosign_helper.rb
+++ b/app/helpers/autosign_helper.rb
@@ -1,5 +1,5 @@
 module AutosignHelper
   def autosign_form
-    button_tag(_("New"), :type => 'button', :class => 'btn btn-primary fr', :data => {:toggle => 'modal', :target =>'#autosignModal'}) if authorized_for(:controller => "SmartProxies::Autosign", :action => :create, :auth_object => @proxy)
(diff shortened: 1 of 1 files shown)


dLobatog (approved):

dLobatog:
Thanks @ares

ohadlevy:
@dLobatog doesnt that conflict with your  #4703 ?

dLobatog:
@ohadlevy I think the disagreement there is about SSH keys, Autosign entries are created both in Foreman and the proxy. If I'm not wrong the discussion is about using "Add" for things that already exist, like SSH keys - creating an SSH key in Foreman does not actually create one using `openssl` or anything like that.

ares:
Thanks for merging @dLobatog, I set a release in redmine to 1.16.0, please don't forget to set it after merge.

dLobatog:
@ares Thanks - I revise daily http://projects.theforeman.org/projects/foreman/activity and set the release of anything that's 'Closed' then, that's why sometimes it's not done right after merge.

ares:
Ok, in your case I'll just set it without reminding you if I see unassigned issue before you get to it, so I can delete the redmine notification. Sorry for bothering... I'm inbox driven :-)

PR #4773, opened 2017-08-21
Fixes #20678 - Improve provisioning templates history

bastilian:
* Make glyphicons the default text color
* Add "No comment provided" when it is empty
* Change link actions to buttons

Before:
gnome-shell-screenshot-ivrg5y

After:
gnome-shell-screenshot-1z054y

### DIFF
--- a/app/assets/stylesheets/patternfly_and_overrides.scss
+++ b/app/assets/stylesheets/patternfly_and_overrides.scss
@@ -361,3 +361,7 @@ a.btn-info span {
 .gridster .gs-w {
   z-index: auto !important;
 }
+
+#history a .glyphicon {
+  color: #4d5258;
+}
--- a/app/views/templates/_history.html.erb
+++ b/app/views/templates/_history.html.erb
@@ -1,0 +1,24 @@
+<% unless history.empty? %>
+  <% history.each do |audit| %>
+    <div class='row'>
+      <div class='col-md-6'>
+        <strong><%= audit_user(audit) %></strong>
+        <p>
+          <% unless audit.comment.blank? %>
+            <%= audit.comment %>
+          <% else %>
+            <em><%= _('No comment provided') %></em>
+          <% end %>
+        </p>
+      </div>
+      <div class='col-md-2 ra'><h6><%= audit_time audit %></h6></div>
+      <div class='col-md-8 audit-content ra'>
+        <%= link_to_function icon_text("restart", _("Revert"), :kind => 'pficon'), "tfm.editor.revertTemplate(this)", :class => 'btn btn-default', :data => {:url => revision_provisioning_templates_url, :version => audit.id } %>
+        <%= link_to icon_text("eye-open", _("Show Diff")), audit_path(audit), :rel => 'external', :class => 'btn btn-default' %>
+      </div>
+    </div>
+  <% end %>
+<% else %>
+  <%= alert(:class => 'alert-info', :header => _('No history found'),
+            :text => _('Save something and try again')) %>
+<% end %>
--- a/app/controllers/templates_controller.rb
+++ b/app/controllers/templates_controller.rb
@@ -2,6 +2,7 @@ class TemplatesController < ApplicationController
   include UnattendedHelper # includes also Foreman::Renderer
   include Foreman::Controller::ProvisioningTemplates
   include Foreman::Controller::AutoCompleteSearch
+  include AuditsHelper
 
   before_action :handle_template_upload, :only => [:create, :update]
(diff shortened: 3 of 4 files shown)


bastilian:
@Rohoover I took your suggestions and this is the outcome. Anything to add that can further improve it?

Rohoover:
@bastilian 
Can you verify the color of the icons? They seem really dark to me compared to icons in other buttons. I think it should be #4d5258

bastilian:
@Rohoover Yep. it was too dark. :) fixed it!
gnome-shell-screenshot-3un04y

timogoebel:
Not a review (yet?), but I this is a huge improvement and looks great. Good job @Rohoover and @bastilian. Thanks.

timogoebel:
@theforeman-bot ok to test

timogoebel:
[test katello]
[test upgrade]

app/assets/stylesheets/patternfly_and_overrides.scss
```diff
@@ -361,3 +361,7 @@ a.btn-info span {
 .gridster .gs-w {
   z-index: auto !important;
 }
+
+#history a .glyphicon {
```
ohadlevy, line 365: any idea why its needed? I was expecting icon css to be defined by patternfly?

app/views/templates/_history.html.erb
```diff
@@ -0,0 +1,25 @@
+<% unless history.empty? %>
+  <% history.each do |audit| %>
```
ohadlevy, line 2: isnt the above line redundant? if its an empty array then each wont do anything.

app/views/templates/_history.html.erb
```diff
@@ -0,0 +1,25 @@
...
+          <% unless audit.comment.blank? %>
+            <%= audit.comment %>
+          <% else %>
+            <em><%= _('No comment provided') %></em>
+          <% end %>
+        </p>
+      </div>
+      <div class='col-md-2 ra'><h6><%= audit_time audit %></h6></div>
+      <div class='col-md-8 audit-content ra'>
+        <%= link_to_function icon_text("retweet", _("Revert")), "tfm.editor.revertTemplate(this)", :class => 'btn btn-default', :data => {:url => revision_provisioning_templates_url, :version => audit.id } %>
```
ohadlevy: question to @Rohoover is "retweet" is the right icon for revert?

app/views/templates/_history.html.erb
bastilian: It wouldn't fallback to "No history found", otherwise it would be redundant, yes.

app/assets/stylesheets/patternfly_and_overrides.scss
bastilian: This would be nice to avoid, but there doesn't seem to be any applicable style that makes the icon dark, while keeping the text the default link color.

app/views/templates/_history.html.erb
bastilian: Maybe the restart icon? image

Rohoover: @bastilian 
Yes Restart is a better icon, since retweet the arrows are moving forward.

bastilian:
@Rohoover & @ohadlevy Icon for reverting is now the restart icon from patternfly:

gnome-shell-screenshot-7y3m5y

ohadlevy:
[test katello]

app/views/templates/_history.html.erb
```diff
@@ -0,0 +1,25 @@
+<% unless history.empty? %>
+  <% history.each do |audit| %>
+    <% next unless audit_template? audit %>
```
ares: since you're touching this area, would you mind fixing this annoying bug that when the only audit in `@history` is not [template change](https://github.com/theforeman/foreman/blob/develop/app/helpers/audits_helper.rb#L71)? In such case we get an empty tab instead of the alert. I suppose this `next` should be moved to [controller](https://github.com/theforeman/foreman/blob/develop/app/controllers/templates_controller.rb#L137) where we'd perform `select` on `@history`

ares:
one little ask, otherwise looks and works good :-)

dLobatog (dismissed):
@bastilian Set to WoC after @ares comment

bastilian:
@dLobatog made the change that @ares requested. the history now gets filtered with to select audits via `audit_template?` and not use it to skip those that are not in the view.

ares (approved):
Works great, waiting for jenkins to confirm.

ares:
Thanks @bastilian, merged as e483e931974027e4e684d7c48d8460b36d657cd5

PR #4775, opened 2017-08-23
fixes #20699 - Rake task to selectively backup tables

parthaa:
This commit adds an optional param "tables" to db:dump task
One can use to this selectively dump data from desired tables.
Its main purpose is to curate debug data from customers for analysis.
For example one can run
```
rake db:dump RAILS_ENV=production tables="foreman_tasks.*, dynflow.*"
```
to get a copy of all the foreman-tasks and dynflow tables

Then on a local db one can choose to just restore those tables
and analyze


dLobatog (approved):
@parthaa Thanks, tried to back up logs and audits and it suceeded.

lzap:
Great patch, next step - integrate with foreman-debug (as an opt-in option).

PR #4778, opened 2017-08-23
Fixes #20708 - fix templates auditing

ares:
This PR changes the auditable_type of audits of templates. In audit logs instead of "Template" you will now see "Partition Table" and "Provisioning Template". Also it allows searching on audits page using "partition_table = aaa" or "provisioning_template = bbb". We conside these classes as main, meaning creation of either will be highlighted with green color and "create" wording instead of blue "added".


iNecas:
Tests seem to fail consistently on this one http://ci.theforeman.org/job/test_develop_pr_core/14017/, let's give it another try, just in case it was something else [test foreman]

ares:
Thanks for looking @iNecas, tests are now fixed, we had the same issues in taxonomies STI, I added a migration to convert "Template" in TaxableTaxonomy table to the right one. I avoided doing the same for audits, since original records might be already deleted so we can't tell what was audit of Ptable and what of ProvisioningTemplate. Could you please take another look?

ares:
Also I kept it in separate commit for easier reviewing, please let me know if you want me to squash.

app/models/provisioning_template.rb
```diff
@@ -4,7 +4,17 @@ class ProvisioningTemplate < Template
   friendly_id :name
   include Parameterizable::ByIdName
 
+  class << self
```
dLobatog, line 7: Not against this by any chance - it's an ugly feature. I wonder though, isn't `Template` an abstract class effectively? As far as I can see it's only instantiated in tests, and according to the link in here,  `self.abstract_class = true` on `Template` would solve the problem too, with less repetition & preventing the class from being instantiated.

dLobatog (changes requested):
@ares Nice job fixing this, a test for `search_for` in Audits would be very nice and allow to verify this doesn't happen again.

app/models/provisioning_template.rb
ares: I'll try that, it would make sense in case it works. I'll let you know how it goes.

ares: Sadly it does not work, it fails on finding the right table name
```
 | ActiveRecord::StatementInvalid (PG::UndefinedTable: ERROR:  relation "provisioning_templates" does not exist
 | LINE 5:                WHERE a.attrelid = '"provisioning_templates"'...
 |                                           ^
 | :               SELECT a.attname, format_type(a.atttypid, a.atttypmod),
 |                      pg_get_expr(d.adbin, d.adrelid), a.attnotnull, a.atttypid, a.atttypmod
 |                 FROM pg_attribute a LEFT JOIN pg_attrdef d
 |                   ON a.attrelid = d.adrelid AND a.attnum = d.adnum
 |                WHERE a.attrelid = '"provisioning_templates"'::regclass
 |                  AND a.attnum > 0 AND NOT a.attisdropped
 |                ORDER BY a.attnum
 | ):
 |   app/controllers/application_controller.rb:231:in `resource_base_with_search'
 |   app/controllers/application_controller.rb:235:in `resource_base_search_and_page'
 |   app/controllers/templates_controller.rb:14:in `index'
 |   app/controllers/concerns/application_shared.rb:15:in `set_timezone'
 |   app/models/concerns/foreman/thread_session.rb:32:in `clear_thread'
 |   app/controllers/concerns/foreman/controller/topbar_sweeper.rb:12:in `set_topbar_sweeper_controller'
 |   lib/middleware/catch_json_parse_errors.rb:8:in `call'
 |   lib/middleware/session_safe_logging.rb:17:in `call'
 |   lib/middleware/tagged_logging.rb:18:in `call'
 | 
```

if I left only `self.table_name` calls it failed to load associations `Association named 'template_kind' was not found on Ptable; perhaps you misspelled it?`, IIRC the we had the same issue parameters STI, which is why we moved them to parent class. I'd like to avoid this here.

ares:
@dLobatog I added the test, rebased and squash the full PR. Could you please take another look?

dLobatog (approved):
@ares Shows the audits properly now, thanks!

mmoll:
FWIW, this is breaking rails5 with i.e. `ActiveRecord::SubclassNotFound: Invalid single-table inheritance type: Template is not a subclass of Ptable`

ares:
Thanks for pointing out, I'll try to look into it next week

ares:
Should be addressed by https://github.com/theforeman/foreman/pull/4822, the default type value set to `Template` causes problems on Rails 5

iNecas:
There was an issue in the migration added in this PR: see https://github.com/theforeman/foreman/pull/4893 for the fix

PR #810, opened 2013-07-23
Fix #2816: Prevent diff viewer on report page from overflowing to be outside the parent modal

skottler:

PR #4779, opened 2017-08-23
Fixes #20710 - fix view diff width

ares:
before:
before

after:
after

see https://github.com/theforeman/foreman/pull/810/files and https://github.com/theforeman/foreman/commit/76ab0628 for previous fixes. If someone has an idea how to prevent this in future, I'm happy to improve this PR (unless it's too complicated :-)

### DIFF
--- a/app/assets/javascripts/reports.js
+++ b/app/assets/javascripts/reports.js
@@ -13,8 +13,7 @@ function show_diff(item){
   $("#diff-modal-editor")
         .css("position","relative")
         .css("padding-top","0")
-        .height('380')
-        .width('588');
+        .height('380px');
 
   var editor = ace.edit("diff-modal-editor");
   editor.setTheme("ace/theme/clouds");


mmoll:
maybe the width can get adjusted/extended a bit in general?

app/assets/javascripts/reports.js
```diff
@@ -14,7 +14,7 @@ function show_diff(item){
         .css("position","relative")
         .css("padding-top","0")
         .height('380')
-        .width('588');
+        .width('568');
```
boaz0: did you try percentages instead of hardcoded value? for example `.width('90%')`

ares:
updated, 70% seems to work fine

tbrisker:
This only seems to work on the first "view diff" click, the second click causes an overflow again. 
I played with it a bit, and looks like the only part of that js that's actually needed is the `.height('380px')`.

ares:
Thanks @tbrisker, seems to work fine. Could you please take another look?

tbrisker:
Thanks @ares !

PR #4780, opened 2017-08-23
Fixes #20711 - GET host interface API will show fqdn attribute

snagoor:
GET API call `/api/hosts/:id/interfaces` will now show `fqdn` attribute


xprazak2:
Works for me:

iNecas:
Thanks @snagoor and @xprazak2

PR #4781, opened 2017-08-23
Fixes #20713 - Show REX option through API calls

snagoor:


timogoebel:
I believe this should go into the Remote Execution Plugin, not in Foreman core.

iNecas:
Agreed, this should go the rex: For the Apipie, we've just added an ability to extend existing parameters from concerns https://github.com/Apipie/apipie-rails#concerns. With extending the rabl templates, it might be a bit more complicated. @ares what was/is your approach addressing the issue in virt-who plugin? Or you've solved it by other means?

iNecas:
For the param attributes, I'm thinking if introducing ability to define rabl partial to be rendered on model layer, and then using them in the rabl template to render additional fields would be the way to go? It might be also on plugin_api level.

iNecas:
For the param attributes, I'm thinking if introducing ability to define rabl partial to be rendered on model layer, and then using them in the rabl template to render additional fields would be the way to go? It might be also on plugin_api level.

timogoebel:
@iNecas: some API similar to the attr_exportable API would be great to have. You can already extend the host rabl via facets. But I think that causes too much overhead sometimes.

This is how Katello does it right now:

https://github.com/Katello/katello/blob/06ae7fd5228c1187113f5b9719e0b0dce4294824/app/controllers/katello/concerns/api/v2/hostgroups_controller_extensions.rb#L42-L47

https://github.com/Katello/katello/blob/016c499f6336725ad58efaea332319b9758a2f77/app/views/katello/api/v2/hostgroups_extensions/show.json.rabl#L1-L3

ares:
@inecas I didn't need to extend rabl templates so far. I did a quick check and didn't find any easy way.

iNecas:
`attr_exportable` variant for this seems like it could work: the only thing I'm not sue is whether there is some more advanced data output that could not be covered by this (e.g. nested data) but if the attribute returned hash or array, even that would be possible. OTOH since the API responses are written in rabl, it seems reasonable to be able to extend it with rabl as well: therefore the partials approach might be worth looking into.

The katello approach doesn't scale to multiple plugins

timogoebel:
@iNecas:

> therefore the partials approach might be worth looking into.

Definitely. When giving that more thought, I believe partials are the way to go as they offer maximum flexibility.

> The katello approach doesn't scale to multiple plugins

I didn't say it was good. I just wanted to collect some use-cases. I personally could use this for https://github.com/theforeman/foreman_expire_hosts/issues/8.

timogoebel:
Closing this for now. Thanks @snagoor.

PR #4783, opened 2017-08-24
Fixes #20728 - Inconsistent behavior of cancel & submit for Hostgroup

kgaikwad:
Inconsistent behavior of cancel & submit regarding search filters on Hostgroup page.

### DIFF
--- a/app/assets/javascripts/host_edit.js
+++ b/app/assets/javascripts/host_edit.js
@@ -146,13 +146,10 @@ function update_capabilities(capabilities) {
 var stop_pooling;
 
 function submit_with_all_params() {
-  var url = $('form').attr('action');
-  if (url.match('hostgroups')) {
-    resource = 'hostgroup';
-  } else {
-    resource = 'host';
-  }
-  capitalized_resource = resource[0].toUpperCase + resource.slice(1);
+  var url = $('form').attr('action'),
+  resource = 'host',
+  page_state_title = page_title_by_resource(resource),
+  url_to_redirect;
   $('form input[type="submit"]').attr('disabled', true);
   stop_pooling = false;
   $('body').css('cursor', 'progress');
@@ -167,12 +164,11 @@ function submit_with_all_params() {
       $('#' + resource + '-progress').hide();
       $('#content').replaceWith($('#content', response));
       $(document.body).trigger('ContentLoad');
-      if ($('[data-history-url]').exists()) {
-        history.pushState(
-          {},
-          capitalized_resource + ' show',
-          $('[data-history-url]').data('history-url')
-        );
+      if (resource == 'host' && $("[data-history-url]").exists()) {
+        url_to_redirect = $("[data-history-url]").data('history-url');
+      }
+      if(url_to_redirect){
+        history.pushState({}, page_state_title, url_to_redirect);
       }
     },
     error: function(response) {
@@ -185,6 +181,15 @@ function submit_with_all_params() {
     },
   });
   return false;
+}
+
+function page_title_by_resource(resource) {
+  var capitalized_resource = resource[0].toUpperCase() + resource.slice(1),
+  page_state_title = capitalized_resource;
+  if (resource == 'host') {
+    page_state_title = capitalized_resource + " show";
+  }
+  return page_state_title;
 }
 
 function clear_errors() {
--- a/app/assets/javascripts/host_edit.js
+++ b/app/assets/javascripts/host_edit.js
@@ -146,10 +146,6 @@ function update_capabilities(capabilities) {
 var stop_pooling;
 
 function submit_with_all_params() {
-  var url = $('form').attr('action'),
-  resource = 'host',
-  page_state_title = page_title_by_resource(resource),
-  url_to_redirect;
   $('form input[type="submit"]').attr('disabled', true);
   stop_pooling = false;
   $('body').css('cursor', 'progress');
@@ -158,17 +154,14 @@ function submit_with_all_params() {
 
   $.ajax({
     type: 'POST',
-    url: url,
+    url: $('form').attr('action'),
     data: serializeForm(),
     success: function(response) {
-      $('#' + resource + '-progress').hide();
+      $('#host-progress').hide();
       $('#content').replaceWith($('#content', response));
       $(document.body).trigger('ContentLoad');
-      if (resource == 'host' && $("[data-history-url]").exists()) {
-        url_to_redirect = $("[data-history-url]").data('history-url');
-      }
-      if(url_to_redirect){
-        history.pushState({}, page_state_title, url_to_redirect);
+      if ($("[data-history-url]").exists()) {
+        history.pushState({}, 'Host show', $("[data-history-url]").data('history-url'));
       }
     },
     error: function(response) {
@@ -180,16 +173,8 @@ function submit_with_all_params() {
       $('form input[type="submit"]').attr('disabled', false);
     },
   });
-  return false;
-}
 
-function page_title_by_resource(resource) {
-  var capitalized_resource = resource[0].toUpperCase() + resource.slice(1),
-  page_state_title = capitalized_resource;
-  if (resource == 'host') {
-    page_state_title = capitalized_resource + " show";
-  }
-  return page_state_title;
+  return false;
 }
 
 function clear_errors() {
--- a/app/views/hostgroups/_form.html.erb
+++ b/app/views/hostgroups/_form.html.erb
@@ -1,5 +1,5 @@
 <%= javascript 'host_edit', 'host_edit_interfaces' %>
-<%= form_for @hostgroup, :html => {:data => {:id => @hostgroup.try(:id), :submit => 'progress_bar' }} do |f| %>
+<%= form_for @hostgroup, :html => {:data => {:id => @hostgroup.try(:id)}} do |f| %>
   <%= base_errors_for @hostgroup %>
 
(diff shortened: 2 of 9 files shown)


app/assets/javascripts/host_edit.js
```diff
@@ -124,13 +124,13 @@ function update_capabilities(capabilities){
 var stop_pooling;
 
 function submit_with_all_params(){
-  var url = $('form').attr('action');
+  var url = $('form').attr('action'),
+  resource = (url.match('hostgroups') ? 'hostgroup' : 'host'),
+  page_state_title = page_title_by_resource(resource),
+  url_to_redirect;
   if (url.match('hostgroups')) {
```
iNecas: since we match `hostgroup` on line 128, we should probably use the `resoruce` variable here in condition

app/views/hostgroups/_form.html.erb
```diff
@@ -1,5 +1,5 @@
 <%= javascript 'host_edit', 'host_edit_interfaces', 'class_edit' %>
-<%= form_for @hostgroup, :html => {:data => {:id => @hostgroup.try(:id), :submit => 'progress_bar' }} do |f| %>
+<%= form_for @hostgroup, :html => {:data => {:id => @hostgroup.try(:id), :submit => 'progress_bar', :"history-url" => session["redirect_to_url_#{controller_name}"] }} do |f| %>
```
iNecas: Shouldn't we you the same method as we have in `cancel_path` later in the form? I guess it should not be called `cancel_path_with_index_search`, as it's not about cancel anymore, but using the same method can ensure we are consistent in the future.

iNecas:
Btw. there is also http://projects.theforeman.org/issues/20105, which talks more about preserving the search between actions. Perhaps we should think about how to do some bigger changes to keep the searches between actions.

app/views/hostgroups/_form.html.erb
kgaikwad: Yes , agreed on using same method. I will replace fetching from session part with this method by renaming it.

kgaikwad:
@iNecas,

Thanks for sharing issue URL. Yes, actually as currently we are storing latest controller action url in session object which may cause inconsistent behavior. I mean if we open edit page and do reload/refresh then it remains on edit page after submitting. I felt it weird but we have this same behavior throughout all actions.

Probably, what we can do is that either using session object only we can preserve this search filters for index action of each controller like we are doing for controllers.

After visiting any index action of controller no matter what actions user can perform ie. delete/edit in end for index we can get latest search from session.
Please let me know your thoughts on this.
After it I will check with Tomer if I can assign above ticket to myself.

iNecas:
The risk with using sessions for those comes, if the user has two tabs with different searches opened at the same time. But we could probably eliminate it by storing the filter into the session only when following the links that will need to use it later for going back. This seems like reasonable compromise for this, but I don't feel too comfortable in UI questions.

kgaikwad:
@iNecas ,

Done with code changes and updated the PR.
I will check on the [issue](http://projects.theforeman.org/issues/20105).

tbrisker:
Perhaps take a look at https://github.com/theforeman/foreman/pull/5820 for a different approach that may be simpler to implement without having to handle all of the redirects in js?

kgaikwad:
@tbrisker,

I have added modifications as per your suggestions.

With these changes, I have removed an additional code for hostgroup form submit from host_edit.js.
I haven't modified the host form submit button's behavior as it redirects to host details page.

kgaikwad:
Any updates on this?

kgaikwad:
@tbrisker, 
Fixed the test case from hostgroup_js_test.

```
expected "/hostgroups" to equal "/hostgroups/new" (Minitest::Assertion)
/var/lib/workspace/workspace/test_develop_pr_core/database/postgresql/ruby/2.5/slave/fast/test/integration/hostgroup_js_test.rb:27    
```

ares:
@kgaikwad could you please do the rebase? @ezr-ondrej will do the review :-)

app/views/hosts/_form.html.erb
```diff
@@ -138,6 +138,11 @@
 
     <%= hidden_field_tag 'bare_metal_capabilities', @host.bare_metal_capabilities, :disabled => true %>
     <%= f.hidden_field :overwrite? %>
-    <%= submit_or_cancel f, @host.overwrite?, :cancel_path => (@host.new_record? ? hosts_path : host_path(:id => @host.name_was)) %>
+    <% resource_prev_url = resource_prev_url_with_search_filters
+      unless resource_prev_url
+        resource_prev_url = @host.new_record? ? hosts_path : host_path(:id => @host.name_was)
+      end
+    %>
+    <%= submit_or_cancel f, @host.overwrite?, :cancel_path => resource_prev_url %>
```
ezr-ondrej: I do not like assigning variables in the `erb`, I would prefer:
```suggestion
    <%= submit_or_cancel f, @host.overwrite?, :cancel_path => resource_prev_url_with_search_filters || (@host.new_record? ? hosts_path : host_path(:id => @host.name_was)) %>
```
If you would disagree, to have the line shorter, I would at least use:
```ruby
resource_prev_url ||= @host.new_record? ? hosts_path : host_path(:id => @host.name_was)
```
instead of the unless statement.

app/assets/javascripts/host_edit.js
```diff
@@ -146,13 +146,10 @@ function update_capabilities(capabilities) {
...
 
 function submit_with_all_params() {
-  var url = $('form').attr('action');
-  if (url.match('hostgroups')) {
-    resource = 'hostgroup';
-  } else {
-    resource = 'host';
-  }
-  capitalized_resource = resource[0].toUpperCase + resource.slice(1);
+  var url = $('form').attr('action'),
```
ezr-ondrej: I would prefer aligning the additional variable names at the same level (after `var` statement) in case we are keeping them.

app/assets/javascripts/host_edit.js
```diff
@@ -146,13 +146,10 @@ function update_capabilities(capabilities) {
...
+  resource = 'host',
```
ezr-ondrej: I am missing the statement witch would decide the resourse. Aren't we missing this later on?
```js
if (url.match('hostgroups')) resource = 'hostgroup';
```

app/assets/javascripts/host_edit.js
```diff
@@ -146,13 +146,10 @@ function update_capabilities(capabilities) {
...
+  page_state_title = page_title_by_resource(resource),
```
ezr-ondrej: As this is used only once, I would prefer using the method right away there (not storing it in variable at all)

app/assets/javascripts/host_edit.js
```diff
@@ -167,12 +164,11 @@ function submit_with_all_params() {
...
-      if ($('[data-history-url]').exists()) {
-        history.pushState(
-          {},
-          capitalized_resource + ' show',
-          $('[data-history-url]').data('history-url')
-        );
+      if (resource == 'host' && $("[data-history-url]").exists()) {
+        url_to_redirect = $("[data-history-url]").data('history-url');
+      }
+      if(url_to_redirect){
```
ezr-ondrej: I believe we can loose this `if` statement and join it safely with the previous one.
So basically keeping the original version only adding `resource == 'host' && ` in the `if` statement.

app/assets/javascripts/host_edit.js
```diff
@@ -187,6 +183,15 @@ function submit_with_all_params() {
   return false;
 }
 
+function page_title_by_resource(resource) {
+  var capitalized_resource = resource[0].toUpperCase() + resource.slice(1),
+  page_state_title = capitalized_resource;
```
ezr-ondrej: I believe we can use only:
```suggestion
  var page_state_title = resource[0].toUpperCase() + resource.slice(1);
```
and drop the `capitalized_resource` variable alltogether, later inside the `if` we would:
```js
page_state_title = page_state_title + ' show';
```
What is perfectly fine statement and can be even shorened to `page_state_title += ' show'`

ezr-ondrej (changes requested):
Looks good @kgaikwad :+1: just some notes for the js part and one for the template.
Mostly nitpicks though, and code stylings, feel free to refuse the ones you do not like :relaxed: 

Just the fixed resource = 'host' seems to be real trouble.

app/assets/javascripts/host_edit.js
kgaikwad: Removed `data-submit => progress_bar` on this [line](https://github.com/theforeman/foreman/pull/4783/commits/46190888489a020c58e2f6f495f095a522e1b38a#diff-db336c69095130c90178c09661f8e4cfR2) for hostgroup that's why here it is not checking for `hostgroup`.

kgaikwad:
> Just the fixed resource = 'host' seems to be real trouble.

@ezr-ondrej,
In original changes, if not matches with `hostgroups` then it takes resource = 'host'. This is the reason I kept it as is. But after my change, I can see that only host form is going to use data `:submit => 'progress_bar'`. I will refactor the JS code.

app/assets/javascripts/host_edit.js
ezr-ondrej: ah, understood, but than we do not need any branching depending on the resource. We can assume 'host' being the only option.

kgaikwad: Updated changes in separate commit

ezr-ondrej (approved):
Looks good, works as expected, fixes the original issue.
Thanks @kgaikwad :+1: merging!

PR #4786, opened 2017-08-25
Fixes #20736 - improve bookmark permissions

ares:
Adds a new role "Bookmark user" granting all bookmarks permissions. This is useful for users to be able to bookmark searches. This also fixes the bug that we show "Bookmark this search" for users without "create_bookmarks" permission.

Steps to reproduce:
1. create a user with Viewer role
2. login as that user, navigate to hosts page
3. use some filter, e.g. name ~ *
4. on search button, click the down arrow and select "Bookmark this search"
5. empty modal window opens, submitting does not work

Testing the fix
6. user should not see "Bookmark this search" item
7. rake db:seed
8. assign Bookmark user role to the same user
9. retry step 4, now the user can see the link and is able to submit the form which creates the bookmark

### DIFF
--- a/app/views/common/_searchbar.html.erb
+++ b/app/views/common/_searchbar.html.erb
@@ -17,8 +17,10 @@
           <li><%= link_to_if_authorized bookmark.name, send("hash_for_#{bookmark.controller}_path").merge(:search => bookmark.query) %></li>
         <% end %>
         <li class="divider"></li>
-        <li><%= link_to_function _('Bookmark this search'), "$('#bookmarks-modal').modal();",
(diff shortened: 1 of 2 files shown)


iNecas:
Tested and works well. Then only thing I'm not sure, if we could call it `Bookmarks User`, as I can use the bookmarks even with just viewer role: I just can't create them. Maybe he could be promoted from Bookmark User to Bookmarks Manager :)

ares:
Make sense :-) in case Bookmark Manager is too big promotion, we could make the new title Bookmark Editor.

ares:
@iNecas role renamed to `Bookmark manager` leaving capitalization consistent with the rest, could you please take another look?

iNecas (approved):

iNecas:
test failures unrelated

PR #4787, opened 2017-08-25
Fixes #20737 - add default description to roles

ares:
it contains #4786 to avoid rebasing conflict later, please only review second commit here

It provides description for built-in roles and displays them on index, since there's no edit page for these roles anymore. I ended up with 2 ways of solving too long descriptions, I prefer the first one
short
long

but I'm happy to hear other ideas, @Rohoover might also help.

### DIFF
--- a/db/seeds.d/02-roles_list.rb
+++ b/db/seeds.d/02-roles_list.rb
@@ -53,5 +53,19 @@ class RolesList
     def view_permissions
       PermissionsList.permissions.select { |resource, name| name.start_with?('view_') }.map { |p| p.last.to_sym }
     end
+
+    def roles_descriptions
+      {
+        Role::MANAGER => 'Role granting all available permissions. With this role, user is able to do everything that admin can except for changing Settings.',
+        Role::ORG_ADMIN => 'Role granting all permissions except for managing organizations. It can be used to delegate administration of specific organization to a user. In order to create such role, clone this role and assign desired organizations',
+        'Edit partition tables' => 'Role granting permissions required for managin partition tables',
+        'View hosts' => 'Role granting permission only to view hosts',
+        'Edit hosts' => 'Role granting permissions to update hosts. For features provided by plugins, you might need to combine this role with roles provided by those plugins',
+        Role::VIEWER => 'Role granting read only access. Users with this role can see all data but can not do any modifications',
+        'Site manager' => 'Role granting mostly view permissions but also permissions required for managing hosts in the infrastructure. Users with this role can update puppet parameters, create and edit hosts, manage installation media, subnets, usergroups and edit existing users.',
+        'Bookmarks user' => 'Role granting permissions for managing search bookmarks. Usually useful in combination with Viewer role. This role also grants the permission to update all public bookmarks.',
+        'Default role' => 'Role that is automatically assigned to every user in the system. Adding a permission grants it to everybody'
+      }
+    end
   end
 end
--- a/db/seeds.d/02-roles_list.rb
+++ b/db/seeds.d/02-roles_list.rb
@@ -3,32 +3,39 @@ require (Rails.root + 'db/seeds.d/02-permissions_list.rb')
 class RolesList
   class << self
     def seeded_roles
-      { Role::MANAGER           => base_manage_permissions + view_permissions + manage_organizations_permissions,
-        Role::ORG_ADMIN         => base_manage_permissions + view_permissions - [:view_organizations],
-        'Edit partition tables' => [:view_ptables, :create_ptables, :edit_ptables, :destroy_ptables],
-        'View hosts'            => [:view_hosts],
-        'Edit hosts'            => [:view_hosts, :edit_hosts, :create_hosts, :destroy_hosts, :build_hosts],
-        Role::VIEWER            => view_permissions,
-        'Site manager'          => [:view_architectures, :view_audit_logs, :view_authenticators, :access_dashboard,
-                                    :view_domains, :view_environments, :import_environments, :view_external_variables,
-                                    :create_external_variables, :edit_external_variables, :destroy_external_variables,
-                                    :view_external_parameters, :create_external_parameters, :edit_external_parameters,
-                                    :destroy_external_parameters, :view_facts, :view_hostgroups, :view_hosts, :view_smart_proxies_puppetca,
-                                    :view_smart_proxies_autosign, :create_hosts, :edit_hosts, :destroy_hosts,
-                                    :build_hosts, :view_media, :create_media, :edit_media, :destroy_media,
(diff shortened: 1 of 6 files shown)


db/seeds.d/02-roles_list.rb
```diff
@@ -53,5 +53,19 @@ def manage_organizations_permissions
     def view_permissions
       PermissionsList.permissions.select { |resource, name| name.start_with?('view_') }.map { |p| p.last.to_sym }
     end
+
+    def roles_descriptions
+      {
+        Role::MANAGER => 'Role granting all available permissions. With this role, user is able to do everything that admin can except for changing Settings.',
```
iNecas: Should the settings be Capitalized?

db/seeds.d/02-roles_list.rb
```diff
@@ -53,5 +53,19 @@ def manage_organizations_permissions
...
+        Role::ORG_ADMIN => 'Role granting all permissions except for managing organizations. It can be used to delegate administration of specific organization to a user. In order to create such role, clone this role and assign desired organizations',
+        'Edit partition tables' => 'Role granting permissions required for managin partition tables',
+        'View hosts' => 'Role granting permission only to view hosts',
+        'Edit hosts' => 'Role granting permissions to update hosts. For features provided by plugins, you might need to combine this role with roles provided by those plugins',
+        Role::VIEWER => 'Role granting read only access. Users with this role can see all data but can not do any modifications',
+        'Site manager' => 'Role granting mostly view permissions but also permissions required for managing hosts in the infrastructure. Users with this role can update puppet parameters, create and edit hosts, manage installation media, subnets, usergroups and edit existing users.',
```
iNecas: I guess it was tough to come up with good description for this one :) Sounds good

db/seeds.d/02-roles_list.rb
```diff
@@ -53,5 +53,19 @@ def manage_organizations_permissions
...
+        'Bookmarks user' => 'Role granting permissions for managing search bookmarks. Usually useful in combination with Viewer role. This role also grants the permission to update all public bookmarks.',
```
iNecas: Do we allow to modify the default role (or any seeded role)? If not, I would rather not talk about adding permissions here

db/seeds.d/02-roles_list.rb
```diff
@@ -53,5 +53,19 @@ def manage_organizations_permissions
     def view_permissions
       PermissionsList.permissions.select { |resource, name| name.start_with?('view_') }.map { |p| p.last.to_sym }
     end
+
+    def roles_descriptions
```
iNecas: Would it be possible to get the descriptions into `seeded_roles` method to have everything at one place?

db/seeds.d/02-roles_list.rb
ares: I wanted to refer to menu item but you're probably right, will change it.

db/seeds.d/02-roles_list.rb
ares: Yeah, I'd love to remove this (and few others) but wanted to keep it for separate PR.

db/seeds.d/02-roles_list.rb
ares: All built-in roles are locked except for Default role which we want to keep editable.

iNecas: Ok then

db/seeds.d/02-roles_list.rb
ares: That would change the `RolesList.seeded_roles` format which might have been used by some plugin, but if you're fine with the change, I can change it and change both places where we use it in Foreman.

iNecas: Are you aware of any such a plugin? Unless there is serious reason for the backward compatibility, I vote for the change

Rohoover:
@ares 
I'm fine with the first option.

db/seeds.d/02-roles_list.rb
ares: grepping all my local plugins didn't find anything, changing this and `SeedHelper#create_role` API

ares:
@iNecas I fixed remaining comments and left it in separate commit, let me know if/when you want me to squash. The description for plugin DSL that we discussed is tracked under http://projects.theforeman.org/issues/19983 and will be fixed after this gets in.

iNecas:
It seems the test failures are related

ares:
I fixed the tests (kept it in the second commit), @iNecas could you please take another look?

db/seeds.d/02-roles_list.rb
```diff
@@ -3,32 +3,38 @@
...
+       'Edit partition tables' => {:permissions => [:view_ptables, :create_ptables, :edit_ptables, :destroy_ptables], :description => 'Role granting permissions required for managin partition tables'},
+       'View hosts' => {:permissions => [:view_hosts],
+                        :description => 'Role granting permission only to view hosts'},
+       'Edit hosts' => {:permissions => [:view_hosts, :edit_hosts, :create_hosts, :destroy_hosts, :build_hosts],
+                        :description => 'Role granting permissions to update hosts. For features provided by plugins, you might need to combine this role with roles provided by those plugins'},
+       Role::VIEWER => {:permissions => view_permissions, :description => 'Role granting read only access. Users with this role can see all data but can not do any modifications'},
+       'Site manager' => {:permissions => [:view_architectures, :view_audit_logs, :view_authenticators, :access_dashboard,
```
iNecas: The indentation  is off here and there are som inconsistent whitespaces. I wonder why rubocop doesnt scream on us?

iNecas: Brw I thought we use spaces after `{ ` in hashes

boaz0 (commented):
LGTM

db/seeds.d/02-roles_list.rb
ares: I intentionally broke the aligning of the hash arrows, I don't think it's good since when you add a key that requires one more space, you need to change all other lines. Rubocop does not scream, since in Ruby this is accepted, this is not puppet.

It might seem a bit off since one key is string starting with `'`, the other is constant but that's aligned correctly.

I will add space after `{`, it's not consistent in the app but I generally prefer it.

ares:
@iNecas updated white spaces, could you please take another look?

db/seeds.d/02-roles_list.rb
iNecas: I haven't meant the puppet style, purely the spaces (thanks, that's one of the few cops I would like to see around), yes, the ' and constants caused them to look misaligned.

iNecas (approved):
Waiting for tests

iNecas:
Thanks @ares

PR #4788, opened 2017-08-25
Fixes #20739 - use unique user label in select

ares:
With this PR `multiple_checkboxes` can use custom method for generating object label, it was hardcoded to `to_label` so for users, the items were not unique

before
before

after
after

### DIFF
--- a/app/helpers/form_helper.rb
+++ b/app/helpers/form_helper.rb
@@ -85,7 +85,7 @@ module FormHelper
         hidden_fields += f.hidden_field(attr_ids, :multiple => true, :value => unauthorized_value, :id=>'')
       end
       hidden_fields + f.collection_select(attr_ids, authorized.sort_by { |a| a.to_s },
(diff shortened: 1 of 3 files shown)


iNecas:
Works for me: the only question: why don't we just update the `to_label` method to make it unique? Any use of that method I've missed. I see the `select_title` is there from year 2010, so perhaps we could revisit if it's actually needed?

iNecas:
Works for me: the only question: why don't we just update the `to_label` method to make it unique? Any use of that method I've missed. I see the `select_title` is there from year 2010, so perhaps we could revisit if it's actually needed?

ares:
to_label is used for name of current user in top right corner, this method is used in host owner selector, so I believe both make sense

iNecas:
ok then, Katello tests failures unrelated. Thanks @ares

ares:
Thanks for quick merge

PR #4791, opened 2017-08-28
Fixes #20763 - Reword "Create SSH key" to "Add SSH key"

bastilian:
Add Button and page to add the key:

gnome-shell-screenshot-6y8s5y

gnome-shell-screenshot-0zo35y

### DIFF
--- a/app/controllers/api/v2/ssh_keys_controller.rb
+++ b/app/controllers/api/v2/ssh_keys_controller.rb
@@ -31,7 +31,7 @@ module Api
         end
       end
 
-      api :POST, "/users/:user_id/ssh_keys", N_("Create an SSH key for a user")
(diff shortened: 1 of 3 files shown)


boaz0 (approved):
LGTM

iNecas:
Should we also change the wording in the API docs? https://github.com/theforeman/foreman/blob/develop/app/controllers/api/v2/ssh_keys_controller.rb#L34

bastilian:
@iNecas yeah, i think that make sense. Will change that. Not sure if renaming the attribute would make sense too, keeping in mind to keep backwards compatibility.

bastilian:
@iNecas changed it now too.

timogoebel (approved):

timogoebel:
Merged, thanks @bastilian.

PR #4793, opened 2017-08-29
Fixes #20783 - Move Inline messaging regarding user saving first within "Notification" fieldset

bastilian:
The warning looked out of place, moving it above all content would not have been right either.

gnome-shell-screenshot-dut25y

Within a Notification fieldset, that will also appear when there are notifications to set, gives it context and the warning can stay where it is.

gnome-shell-screenshot-6iev5y

I've also remove the dismissable, since it will be gone once the user is saved and notifications can be set.

### DIFF
--- a/app/views/users/_form.html.erb
+++ b/app/views/users/_form.html.erb
@@ -61,13 +61,16 @@
         </div>
       <% end %>
       </br>
-      <% if @user.new_record? %>
-        <%= alert :class => 'alert-warning', :header => '', :text => _("Please save the user first before assigning mail notifications.") %>
-      <% else %>
-        <%= field_set_tag _("Notifications") do %>
+
+      <%= field_set_tag _("Notifications") do %>
+        <% if @user.new_record? %>
+          <%= alert :class => 'alert-warning', :header => '', :text => _("Please save the user first before assigning mail notifications."), :close => false %>
+        <% elsif !@user.user_mail_notifications.empty? %>
           <%= f.fields_for :user_mail_notifications do |mail_form| %>
             <%= render :partial => "users/mail_notifications", :locals => { :f => mail_form }  %>
           <% end %>
+        <% else %>
+          <%= alert :class => 'alert-info', :header => '', :text => _("Notifications can't be assigned to this user."), :close => false %>
         <% end %>
       <% end %>
     </div>


bastilian:
@Rohoover Mind taking a look and see if that is better than simply having moved the notification up to the top of the content?

Rohoover:
Does this warning appear on the checkbox or the Submit?

bastilian:
@Rohoover the warning is shown when a user is about to be created, which needs to happen first in order to assign notifications. Once the user is saved the warning is replaced by the emails they can receive.

gnome-shell-screenshot-gig25y

The dismissing would only have removed the warning, but not replace it with the above.

Rohoover:
@bastilian 

I'm torn because typically notifications occur at the top, but I think this will be okay here.

orrabin:
Before giving the user any permissions the notifications section is empty when editing:
image
Should the Notifications header be there when there are no notifications possible?
Maybe there should be another warning about permissions?

amirfefer:
Maybe the entire tab shouldn't be visible when there are no permissions?

Rohoover:
I think having the info inline box would be helpful perhaps telling the user that they should contact their administrator if they are not seeing what they expect.

bastilian:
@orrabin I've added an info box when there are no notifications for that user. The info reads "There are no mail notifications to be assigned to this user.". /cc @Rohoover

gnome-shell-screenshot-0vmm6y

Rohoover:
Is the notification assignment active or future?

"There are no mail notifications to be assigned to this user."
or
"There are no mail notifications assigned to this user."

Whichever makes more sense for this context.

orrabin:
@Rohoover the message means this user is missing view permissions, as a result it makes no sense to notify about changes to something the user can't see so we hide the notification options.
Out of the two messages I think "to be assigned" makes more sense in this case but if that is not clear enough maybe something like:
"There are no mail notifications available to assign to this user."
or
"Notifications can't be assigned to this user" ?

Rohoover:
@orrabin 
Could we say "You do not have permission to view mail notifications."

orrabin:
@Rohoover no because the person editing the user does have permission, the user that is being edited does not.

Rohoover:
@orrabin 
Thanks for the clarification.

I prefer "Notifications can't be assigned to this user."

bastilian:
@orrabin & @Rohoover the wording is now changed:

screenshot-20170920065028-1044x161

iNecas:
Based on the conversation. It seems there is no more things to update here. Thanks @bastilian, @Rohoover and @orrabin

PR #4794, opened 2017-08-29
Fixes #18492 - add pf vertical navigation

amirfefer:
### Gif Illustration
aug 29 2017 2-48 pm 1

### Roxanne mockup
[Foreman.-.Vertical.Nav.pdf](https://github.com/theforeman/foreman/files/1260248/Foreman.-.Vertical.Nav.3.pdf)

### Todo list:
- [x] Tests
- [x] Add turbolinks support for menu links (pf js breaks it)
- [x] Fix broken styles
- [x]  Change navbar png colors


Rohoover:
Here is the latest mockup with icons. I will work on getting all the icon names.

foremannav_recommended2

sean797:
The color on the left hand side of `app/assets/images/navbar.png` doesn't look right IMO, its too bright. I think it should match `$nav-pf-vertical-active-bg-color` ?

Overall I quite like this, though I am yet to test it 👍 Thanks!

Rohoover:
@sean797 
Agreed... too bright.

Here are the colors of the masthead going from left to right:

#00729F
#0A517F
#024D6C

app/helpers/home_helper.rb
```diff
@@ -1,8 +1,8 @@
 module HomeHelper
-  def render_menu(menu_name)
+  def render_vertical_menu(menu_name)
     authorized_menu_actions(Menu::Manager.items(menu_name).children).map do |menu|
       items = authorized_menu_actions(menu.children)
-      render "home/submenu", :menu_items => items, :menu_title => _(menu.caption), :menu_name => menu.name if items.any?
+      render "home/vertical_menu", :menu_items => items, :menu_title => _(menu.caption), :menu_icon => menu.icon, :menu_name => menu.name if items.any?
```
timogoebel, line 5: I think the caption is translated twice right now.

image

app/views/home/_org_switcher.html.erb
```diff
@@ -1,11 +1,21 @@
...
-      <%= render 'home/organization_dropdown' if SETTINGS[:organizations_enabled] %>
-      <%= content_tag(:li,"",:class=>"divider") if SETTINGS[:organizations_enabled] && SETTINGS[:locations_enabled] %>
       <%= render 'home/location_dropdown' if SETTINGS[:locations_enabled] %>
     </ul>
   </li>
+<% end %>
+<% if SETTINGS[:organizations_enabled] %>
+    <li class="dropdown org-switcher" >
+      <a href="#" class ="dropdown-toggle nav-item-iconic" data-toggle="dropdown" >
+        <span data-toggle="tooltip" data-placement="bottom" title=<%= _("Organization") %>><%= org_title %></span><span class="caret"></span></a>
```
timogoebel: I'd remove the tooltip. Looks off to me.
image

app/views/home/_user_dropdown.html.erb
```diff
@@ -1,5 +1,5 @@
 <li id='notifications_container' class="drawer-pf-trigger dropdown"> </li>
-<li class="dropdown menu_tab_dropdown">
+<li class="dropdown">
```
timogoebel: image

If I click on the notification bell and miss it by a pixel, the user menu opens. I think there should be some "unclickable" space between the two actions.

timogoebel (changes requested):
Overall, looks good and a lot more Foreman, than the initial proposal. 👍 
The content jumps to the right, when you click a link in the menu. That feels weird. I also left some small comments inline.
We also have to check every page for any display errors. E.g. the actions row in the facts index table is broken now.

sean797:
Another thing to note. If I click the org or location buttons then select a menu item, like 'Hosts' I'm left with the org or location drop down behind it, looking a little odd.

screenshot from 2017-08-30 09-35-05

The notifications drawer looks quite different, in an annoying position and a lot bigger.

**now:**
image

**develop branch:**
image

GregSutcliffe:
Awesome stuff @amirfefer - I'll get this deployed on a public demo box as soon as I can (today or Monday) and we can start getting feedback from the community (as promised in the demo :P).

Rohoover:
>  Another thing to note. If I click the org or location buttons then select a menu item, like 'Hosts' I'm left with the org or location drop down behind it, looking a little odd.

Agreed. Org/Location dropdown should be above all else from both a visual perspective and in consideration of it's relationship to the left menu.

amirfefer:
I've fixed issues noted above, enabled turbolinks from menu, changed navbar colors,  added tests and rebased. 
@timogoebel, @sean797, @Rohoover, @ohadlevy   could you have a second look ?

sep 13 2017 10-51 am

GregSutcliffe:
https://vertnav.demo.theforeman.org/ has the PR applied, and I'm sorting out making it useful to people to test with... admin credential will follow :P

timogoebel:
@GregSutcliffe: I hope a valid SSL certificate will follow as well :-)

GregSutcliffe:
@timogoebel if i get chance, you know how messy changing the certs can be :)

amirfefer:
When removing PF `container-pf-nav-pf-vertical` class  from the main content, it  removes the weird resizing after menu clicks, but  breaks hamburger resizing as well:
sep 13 2017 5-01 pm

Rohoover:
Can we ensure the active secondary nav type is white? It's hard for me to tell from the gif. Thanks!

GregSutcliffe:
@Rohoover et al - if you want to play with it, hit up https://vertnav.demo.theforeman.org, login is admin/vertnav - sorry about the self-signed cert, will try to put a proper one in at some point

bkearney:
Comments on  https://vertnav.demo.theforeman.org:

 * Is the default setting of the menu size (small|large) cookie based, or setting based?
 * The left padding in the content window is too large
 * The click and "push out" of the content window that Greg mentioned in annoying.
 * Clickin on the menu shows a blue bar that does not seem to do anything.
 * I get no indication of where I am, but I dont know if I care.
 * I assumed shrinkng the size of the winfow would cause the menu to go into the hamburger menu. It does eventually, but it seems kinda slow.
 * Can we see the org selector on this?
 * The notification box seems too far to the right.
 * The puppet envrionments page is a bit wonky since the button inclues teh host name. 
 * Is there any option to shrink the header as well? I could handle the entire thing in "small mode" to be 50% smaller.
 * The content title seems a bit thin now.

Rohoover:
Additional Comments:

- The current area of the application you are on should remain highlighted in the navigation. When you click on any of it's children, the page loads and the highlight disappears.
- When clicking on a main nav item, but not a child (which ultimately takes you somewhere), the highlight appears and remains for that item even if that is not what content is loaded.
- Chevrons are missing for menus that expand.
- "Clickin on the menu shows a blue bar that does not seem to do anything." - if this is in reference to the vertical blue bar next to the menu item, that is to indicate that is what content area is currently loaded. 
- "Is there any option to shrink the header as well?" - I think with the org/loc switcher in there the height may be a benefit.
- "The click and "push out" of the content window that Greg mentioned in annoying." - agreed and not necessary.
- Something I noticed but not totally related is that the page titles could be better on a background bar (grey).

bkearney:
> The current area of the application you are on should remain highlighted in the navigation. When you click on any of it's children, the page loads and the highlight disappears.

I am not seeing this behaviour

> When clicking on a main nav item, but not a child (which ultimately takes you somewhere), the highlight appears and remains for that item even if that is not what content is loaded.

This is by design?

> Chevrons are missing for menus that expand.
> "Clickin on the menu shows a blue bar that does not seem to do anything." - if this is in reference to the vertical blue bar next to the menu item, that is to indicate that is what content area is currently loaded.
Yes, but I only see that when I am clickin on the menu. Once focus is removed, the highlight goes away.

> "Is there any option to shrink the header as well?" - I think with the org/loc switcher in there the height may be a benefit.

@GregSutcliffe can we get that loaded?

> Something I noticed but not totally related is that the page titles could be better on a background bar (grey).

Yes, that would help

Rohoover:
>I am not seeing this behaviour
>
For example: when I click on Content> Puppet Environments, the highlighted is gone after the content loads. I logged out and back in and I'm still seeing this.

- As a sidenote - if the default page is Hosts, that menu item should be highlighted upon login.

> This is by design?
> 
Sorry if I was not clear, that is not by design. It should not be highlighted because though you clicked on the main nav item you ultimately didn't go anywhere as all the actual links are in the child menu.

ares (dismissed):
Based on what I see on vertnav demo (thanks @GregSutcliffe!) tt seems that the PR drops existing id HTML attributes of menu items. This is imporant for testing and we should make sure we keep them.

omaciel:
Thank you @ares unique IDs are very important for UI automation

bkearney:
@omaciel Is there value if the ids remain the same as in 1.16? For example, monitor_menu ?

omaciel:
@bkearney yes, we could then continue using the same locators.

bkearney:
@Rohoover the ogselector is now up. This brings up three points:

1) I Still like the new nav
2) Why is location to the left of org? I assumed I would read Org and then the location inthat org
3) I still think the padding is too much on the nav. I would reduce it a bit more to reclaim more content space.

Rohoover:
> Why is location to the left of org? I assumed I would read Org and then the location inthat org
> 
I'm fine with switching them.

> I still think the padding is too much on the nav. I would reduce it a bit more to reclaim more content space.
> 

I'm looking to see if other applications have customized the padding at all.

amirfefer:
@bkearney, @Rohoover  @ares :
* main content left padding is smaller now
* The click and "push out" of the content window  is fixed
* when shrinking the window, the vertical menu goes into the hamburger menu (about 768px and less),  as well as org/loc switcher menu (@Rohoover I'd like your opinion about that) , as you can see in the following gif:
sep 19 2017 11-18 pm

* the current area of the application  remains highlighted in the navigation 
* when clicking on main nav item, no highlight appears
* I switched between org and loc locations
*  the menu items unique ID are exist 

> Is the default setting of the menu size (small|large) cookie based, or setting based?

based on local storage, with it similar to cookie (patternfly behavior)

> Is there any option to shrink the header as well? I could handle the entire thing in "small mode" to be 50% smaller.

I'd prefer to remain it just like that ( or shrink it slightly)  because the org&loc switchers,

###  main nav items highlighting & "click and push out" :
sep 19 2017 10-44 pm

bkearney:
> @bkearney, @Rohoover @ares :
> main content left padding is smaller now

If the demo has been updated, then it still looks too large. @Rohoover, what is your suggestion?

> The click and "push out" of the content window is fixed
This seems like it has been updated. I still see a wierd flicker with the left menu is smaller, but it looks better.

> when shrinking the window, the vertical menu goes into the hamburger menu (about 768px and less), as well as org/loc switcher menu (@Rohoover I'd like your opinion about that) 

When I shrink the window the org switcher and the right hand stuff (notifications and logout) seem to disappear.

> the menu items unique ID are exist
Awesome. Same as before?


>> Is there any option to shrink the header as well? I could handle the entire thing in "small mode" to be 50% smaller.
> I'd prefer to remain it just like that ( or shrink it slightly) because the org&loc switchers,

I will defer to @Rohoover 

All and all, I like it.

@GregSutcliffe  and @orrabin any feedback from the surveys? /me would like to see this in.

amirfefer:
[test foreman]

app/assets/javascripts/application.js
```diff
@@ -24,9 +24,17 @@ $(document).on("page:fetch", tfm.tools.showSpinner)
 $(document).on("page:change", tfm.tools.hideSpinner)
 
 $(function() {
+  $().setupVerticalNavigation(false);
```
ohadlevy: should those function calls be part of the top_bar js file instead ?

app/assets/javascripts/application.js
```diff
@@ -24,9 +24,17 @@ $(document).on("page:fetch", tfm.tools.showSpinner)
...
   $(document).trigger('ContentLoad');
 });
 
+$(document).on('page:fetch', function(){
+  // a workaround to enable turbolinks works with pf vertical navigation
```
ohadlevy: ditto

app/assets/javascripts/topbar.js
```diff
@@ -1,31 +1,8 @@
 $(function() {
-  mark_active_menu();
```
ohadlevy: can we move this file to webpack?

app/assets/stylesheets/base.scss
```diff
@@ -7,6 +7,41 @@ body {
   height: 100%;
 }
 
+.navbar-brand-text {
+  padding-left: 5px;
+  margin-top: 7px !important;
+}
+
+.navbar-pf-vertical {
```
ohadlevy: why not move all of the navigation related css to the top_bar css file? (or maybe we should rename it from top bar to navigation etc).

app/helpers/home_helper.rb
```diff
@@ -43,24 +43,38 @@ def menu_item_tag(item)
...
             else
-              _("Any Context")
+              _("Any Location")
+            end
+    title
+  end
+
+  def org_title
+    org_current = truncate(Organization.current.to_label) if Organization.current
+    title = if org_current
```
ohadlevy: any reason to not make it a oneliner?
```ruby
def tax_title(org = Orgainzation.current)
  return _('Any ...) unless org
  truncate(org.to_label)
end
```

app/registries/menu/loader.rb
```diff
@@ -38,8 +38,26 @@ def self.load
         end
       end
 
+      Manager.map :location do |menu|
+        menu.sub_menu :loc, :caption => N_('Location'), :icon => 'fa fa-globe' do
+          menu.item :any_loc, :caption => N_('Any Location'), :url_hash => {:controller => '/locations', :action => 'clear'}
+          Location.all.each do |location|
+            menu.item location.title.to_sym, :caption => location.title,   :url_hash => {:controller => '/locations', :action => 'select', :id =>  location.id }
```
ohadlevy: can you please test it with 50 locations or so?

test/integration/top_bar_test.rb
```diff
@@ -8,63 +8,54 @@ def setup
 
   test "top bar links" do
     visit root_path
-    within("div.navbar-outer") do
+    within("nav.navbar-pf-vertical") do
       assert page.has_link?("Foreman", :href => "/")
     end
-    within("div.navbar-inner") do
```
ohadlevy, line 14: why remove? I would expect the links to still be there?

ohadlevy (changes requested):
thanks @amirfefer I've added a few inline comments.

Can you please also add tests when starting in mobile mode? (e.g. you open foreman from mobile and not resize to mobile view)

ohadlevy:
navigation and user actions do not seems to be inline: see 
image

ohadlevy:
in some cases, the menu remains expanded, in order to reproduce i simply zoomed in multiple times (until i got into mobile view) and then unzoom and clicked ctrl-0 (to go back to 100%) and the current page was expanded (e.g. the host tab and all sub actions).

ohadlevy:
for the org switcher please:

1. remove the org (and location) header
2. consider moving the actual selection of orgs into the same dropdown - not sure why we need two?
3. maybe move the manage actions to the top (in case you have a scroll?)
4. should we consider fixing http://projects.theforeman.org/issues/21068 ?

/cc @Rohoover ?

image

ohadlevy:
should it still animate on selection (note how the content is pushed to the right), i find it a bit odd that it does ?
recorded

ohadlevy:
when hamburger is enabled (e.g. minimized view), and refreshing the screen, you can see it starting in full screen and channging back to minimize view which 
recorded

ohadlevy:
probably related to an earlier issue, but once opening notifications, you can't close it
recorded

amirfefer:
> when hamburger is enabled (e.g. minimized view), and refreshing the screen, you can see it starting in full screen and channging back to minimize view which

> in some cases, the menu remains expanded, in order to reproduce i simply zoomed in multiple times (until i got into mobile view) and then unzoom and clicked ctrl-0 (to go back to 100%) and the current page was expanded (e.g. the host tab and all sub actions).


those  behaviors  exist in [pf vertical nav demo](https://www.patternfly.org/pattern-library/navigation/vertical-navigation/vertical-navigation.html)  too, bugs?

amirfefer:
notifications is in better position now :  
notification_1

amirfefer:
> remove the org (and location) header

IMO In some situations it better to keep this header. First, we have now two switchers instead of just one, and at the beginning  users may be confused which one is what. Second, sometimes loc/org titles may be ambiguous, so it may help 
> consider moving the actual selection of orgs into the same dropdown - not sure why we need two?

agree
> maybe move the manage actions to the top (in case you have a scroll?)

as well as loc/org cleaner? ("Any Location"..)?

amirfefer:
This PR also makes foreman more convenient with mobile
[foreman on mobile demo](https://drive.google.com/file/d/0B3qMz05xKX_MVm9yVTBsSVZ5Q0E/view)
as you can see org/loc switchers were moved to hamburger menu

amirfefer:
@ohadlevy, @timogoebel
I've fixed issues from the comments above,  moved navigation js to webpack & opened pf [issue](https://github.com/patternfly/patternfly/issues/767),

app/helpers/home_helper.rb
```diff
@@ -43,24 +43,32 @@ def menu_item_tag(item)
...
-              org_current
-            elsif loc_current
-              loc_current
-            else
-              _("Any Context")
-            end
-    title
+  def menu_secondary_item(item)
+    html_options = {:id => "menu_item_#{item.name}"}
+    html_options['data-no-turbolink'] = true if !item.turbolinks
```
ohadlevy, line 48: is this equiv to:
```ruby
  html_options['data-no-turbolink'] = !!item.turbolinks
```

app/helpers/application_helper.rb
```diff
@@ -366,7 +366,7 @@ def avatar_image_tag(user, html_options = {})
       image_tag(gravatar_image(user.mail),
                 html_options.merge!(gravatar_html_options))
     else
-      image_tag('user.jpg', html_options)
```
ohadlevy, line 369: please also delete this image.

webpack/assets/javascripts/foreman_navigation.js
```diff
@@ -0,0 +1,46 @@
+import $ from 'jquery';
+
+$(document).on('page:fetch', () => {
```
ohadlevy: please change this to a function and call it from the current event on page:fetch in application.js - it would make it much easier to find in the future.

webpack/assets/javascripts/foreman_navigation.js
```diff
@@ -0,0 +1,46 @@
...
+  // a workaround to enable turbolinks works with pf vertical navigation
+  $.fn.setupVerticalNavigation.self = undefined;
+  $(document).off('mouseenter.pf.tertiarynav.data-api', '.secondary-nav-item-pf');
+  $(document).off('mouseleave.pf.tertiarynav.data-api', '.secondary-nav-item-pf');
+});
+
+function markActiveMenu() {
+  let link = `[href='${window.location.pathname}']`;
```
ohadlevy: please use const vs let if you are not modifying the variable.

webpack/assets/javascripts/foreman_navigation.js
```diff
@@ -0,0 +1,46 @@
...
+  let currentLoc = $('#location-dropdown .nav-item-iconic').text().trim();
+  let currentOrg = $('#organization-dropdown .nav-item-iconic').text().trim();
+
+  $('.list-group-item.secondary-nav-item-pf ').has(link).addClass('active');
+  $(`.nav-pf-secondary-nav .list-group-item:contains("${currentLoc}")`).addClass('active');
+  $(`.nav-pf-secondary-nav .list-group-item:contains("${currentOrg}")`).addClass('active');
```
ohadlevy: please add tests for those usage cases.

webpack/assets/javascripts/react_app/components/notifications/notifications.scss
```diff
@@ -1,6 +1,6 @@
 #notifications_container {
   position: static;
-  margin-top: 2px;
+  margin-top: -1px;
```
ohadlevy, line 3: please make sure the notification drawer opens below the top bar, it currently opens in the middle of the bar
image

ohadlevy (changes requested):

amirfefer:
@ohadlevy: 
> consider moving the actual selection of orgs into the same dropdown - not sure why we need two?

location

/cc @Rohoover ?

amirfefer:
@ohadlevy , I've added tests and fixed your comments above

Rohoover:
@amirfefer @bkearney @ohadlevy 

Sorry for the delay, I was on-site at customers (who by the way, like the new nav).

> I'd prefer to remain it just like that ( or shrink it slightly) because the org&loc switchers,

> 

Agreed - also the masthead is used for additional icons, perhaps in the future a help icon. I checked around with other applications using vert nav and none of them seem to adjust the height of the masthead so I'm reluctant to want to make an exception here. 

> consider moving the actual selection of orgs into the same dropdown - not sure why we need two?

> 

Agreed.

webpack/assets/javascripts/foreman_navigation.js
```diff
@@ -0,0 +1,29 @@
+import $ from 'jquery';
+
+function markActiveMenu() {
+  let link = `[href='${window.location.pathname}']`;
```
ohadlevy: please use const instead of let.

waldenraines:
@Rohoover thoughts on this scrollbar?

screen shot 2017-09-27 at 12 58 23 pm
screen shot 2017-09-27 at 12 58 37 pm

Rohoover:
If we could get styling closer to this, that would be better:

screen shot 2017-09-27 at 1 09 04 pm
screen shot 2017-09-27 at 1 14 06 pm

waldenraines:
> If we could get styling closer to this, that would be better:

What you are seeing there is OSX doing some magic.  If you were to plug in a mouse or go to Settings ->General  and change "Show Scroll Bars" to always (or plug in a mouse) you would see the same ugly scrollbars as in my screenshot.

I haven't looked at styling scrollbars since CSS2 but it was a real pain then, not sure how easy it is to do now.

Rohoover:
Yea, I figured I had some kind of styling happening, I'm just hoping we can more mimic the styling I see vs. what you see.

waldenraines:
Here are a couple of PRs to get the other icons showing up:

https://github.com/Katello/katello/pull/6975
https://github.com/theforeman/foreman-docker/pull/199

screen shot 2017-09-27 at 1 39 32 pm

waldenraines:
@amirfefer can we make a click on a menu item close the menu?  In katello we don't load a new page when you click a menu item so the menu stays open after a click:

untitled

Even when clicking on an item in the foreman menu it still stays open after clicking it's just not as noticeable since you subsequently transition to a new page (with a full page reload).

app/assets/javascripts/application.js
```diff
@@ -19,11 +18,15 @@
 //= require editable/bootstrap-editable
 //= require editable/rails
 
-$(document).on("page:fetch", tfm.tools.showSpinner)
+$(document).on("page:fetch", function() {
+  tfm.tools.showSpinner();
+  tfm.nav.activateNav();
```
ohadlevy: nitpick for naming, ```tfm.nav.active()`` could be enough.

app/assets/javascripts/application.js
```diff
@@ -19,11 +18,15 @@
...
+});
 
 $(document).on("page:change", tfm.tools.hideSpinner)
 
 $(function() {
+  tfm.nav.initNavigation();
```
ohadlevy: ditto, ```tfm.nav.init``` is probably enough.

app/views/home/_navbar.html.erb
```diff
@@ -0,0 +1,11 @@
+<div id="vertical-nav" class="nav-pf-vertical
+     nav-pf-vertical-with-sub-menus
+     nav-pf-persistent-secondary">
+  <!-- menu -->
```
ohadlevy: nitpick: not sure if you wanted it in the resulting html, but if its just a template comment, erb comments should be 
```erb
<%# comment goes here %>
```

app/views/home/_vertical_menu.html.erb
```diff
@@ -0,0 +1,33 @@
...
+  </a>
+  <div id=<%= "#{menu_name}-secondary" %> class="nav-pf-secondary-nav">
+
+  <div class="nav-item-pf-header">
+  <a class="secondary-collapse-toggle-pf", data-toggle="collapse-secondary-nav" ></a>
+    <%= content_tag :span, _(menu_title) %>
+  </div>
+    <ul class="list-group">
+      <% authorized_menu_actions(menu_items).each do |item| %>
+          <% case item %>
```
ohadlevy: indentation here is wrong, should be 2 spaces.

webpack/assets/javascripts/foreman_navigation.test.js
```diff
@@ -0,0 +1,39 @@
+jest.unmock('./foreman_navigation');
+const $ = require('jquery');
+const nav = require('./foreman_navigation');
```
ohadlevy: please use es6 syntax, for example:
```js
import $ from 'jquery';
import initNavigation from './foreman_navigation';
```

ohadlevy (changes requested):

amirfefer:
@waldenraines :  I fixed the issue with the secondary menu,  now it should close after a  menu item click.
@ohadlevy: I fixed your comments above
also some survey comments have been fixed (i.e firefox scrolling issue)

app/assets/stylesheets/navigation.scss
```diff
@@ -0,0 +1,89 @@
+@import "colors";
```
ohadlevy: not so interesting question, but can / should we move this file to webpack too?

app/registries/menu/toggle.rb
```diff
@@ -1,7 +1,9 @@
 module Menu
   class Toggle < Node
-    def initialize(name, caption)
+    attr_accessor :icon
+    def initialize(name, caption, icon)
       @caption = caption
+      @icon = icon || ""
```
ohadlevy, line 6: @Rohoover in the rare case that a plugin did not define a menu icon, should we have a default icon or the default should be not to display any Icon?

Also, i saw in other projects that they use the ```>> ``` as a way to show there is a sub menu, should we use that? I personally found it noisy (for sure in the top level).

app/views/home/_navbar.html.erb
```diff
@@ -0,0 +1,10 @@
+<div id="vertical-nav" class="nav-pf-vertical
```
ohadlevy, line 1: this file should be 2 spaces.

webpack/assets/javascripts/foreman_navigation.test.js
```diff
@@ -0,0 +1,39 @@
...
+            href="/locations/13/select"><span class="list-group-item-value">
+            Loc2</span></a></li>
+         </ul>
+       </div>
+      </div>`;
+
+    init();
+    expect($('.nav-pf-secondary-nav .list-group-item:contains("Loc1")')
+      .hasClass('active')).toBe(true);
+  });
```
ohadlevy: please add more tests for init, activate and markActiveMenu functions.

waldenraines:
> @waldenraines : I fixed the issue with the secondary menu, now it should close after a menu item click.

Thanks, it's working well now.  The only comment I have is that it would be great to fix the scrolling issue in https://github.com/theforeman/foreman/pull/4794#issuecomment-332587467 but I think it is fine to merge this and fix that later if you don't want to delay this PR further.

app/registries/menu/toggle.rb
Rohoover: I think a default icon would be good. I don't know if PatternFly defines what that icon is. 

In the demo and in my mockup, I incorporate the ">" for sub menus. I believe I commented earlier in the thread that it's missing. I think it's worth keeping because it differentiates between those items with and without a sub menu.

ohadlevy: @Rohoover So to sum it up:
1. scrollbar should be consistent across browsers? (Tbh that's sounds like a patternfly bug).
2. Missing › string on non links (including the top level navigation that are all links to sub menus)?

Anything else?

amirfefer: @Rohoover 
All the top level navigation items have a sub menu, so IMHO the ">" a bit noisy on them, especially in collapse mode.

app/assets/stylesheets/navigation.scss
amirfefer: patternfly uses variables from this file, so I don't quite sure if moving it works well.

amirfefer:
@ohadlevy:  I've added more tests
@waldenraines:  /cc [#4794 (comment) ](https://github.com/theforeman/foreman/pull/4794#discussion_r142245356)

app/views/home/_navbar.html.erb
ohadlevy: @amirfefer bump^

amirfefer: @ohadlevy  already done

app/registries/menu/toggle.rb
waldenraines: > scrollbar should be consistent across browsers? 

I don't think we should block this PR on the scrollbar issue for three reasons:

1. My experience with styling scrollbars consistently is that it's difficult (although perhaps CSS3 makes this easier)
1. Perhaps styling the scrollbar isn't even the correct solution here.  My point is that we should spend a little bit of time coming up with a correct solution instead of rushing into a solution.
1. You have to have a pretty small screen to see this issue.  I had to take my height down to around 600px to see it on the content menu (which is our longest menu)

> (Tbh that's sounds like a patternfly bug).

I would also be interested to know how patternfly as a whole plans on addressing this issue.

TL;DR: I would be fine with opening a follow up issue to address the scrollbar.

waldenraines (approved):
I don't require any more changes other than what's been noted by other reviewers so I am going to go ahead and mark this PR as approved to hopefully help get this merged soon.

amirfefer:
[test katello]

waldenraines:
@timogoebel @ohadlevy @ares what is left to do here before we can merge?

app/registries/menu/toggle.rb
ohadlevy: Ping @rohoover I believe these are the only remaining issues before we can merge this?

Rohoover: Yes, I believe so.

ohadlevy: @rohoover - I'm uncertain how to continue, the scrollbar seems like a pf bug, and I'm uncertain if we should block on it.
secondly, since we have no third level naviation, it seems that we don't need a ">" symbol? is that right?

what would you suggest we do ? thanks.

Rohoover: I don't think we need to block on scroll bar, it's more a visual issue than functionality. As walden pointed out, it's probably an extreme use case that would even see it.

The chevron should be present on any menu that has a sub.

ohadlevy: thanks @Rohoover.
@amirfefer can you please
1. open a bug at pf for the scroll issue (that we track it).
2. add the missing chevrons ?

bittner:
I really adore the new layout, congratulations for the great achievement!

There's one little thing, though, I love from the current design that would obviously go lost with the changes as of now: **The nice overlap** of the The Foreman hat.

the-lovely-foreman-hat
That was *the* feature that made me fall in love with The Foreman when I first saw the user interface. Can you re-integrate this design-feature? I'd really really miss it. :cry:

sean797:
@bittner +1 though im not sure how that would work with the collapsed menu:

screenshot from 2017-10-13 16-30-33
Vs
screenshot from 2017-10-13 16-30-23






-----------------------------
The Org, Loc & user menu are slightly misaligned, should be alligned to the top bar IMO
screenshot from 2017-10-13 16-27-55

sean797:
Also the colours used here doesn't quite match what is used on https://theforeman.org/. They should be consistent IMO.

screenshot from 2017-10-13 16-53-59

Rohoover:
@sean797 

Unfortunately the pattern doesn't accommodate logos outside of the defined area. I will take this feedback to the PatternFly team. 

Regarding colors - I pulled the colors from the Foreman application, so the intent is to match. If they don't, we should definitely adjust them.

app/registries/menu/toggle.rb
amirfefer: @ohadlevy  what about a default icon? should we remain it blank for now?

amirfefer:
@ares 
> I don't see html ids in vertnav demo, probably it was not updated. If ids were restored, I have no more comments

ids exists, as you can see here:
screenshot from 2017-10-15 01-21-41

@ohadlevy I've added missing chevrons

amirfefer:
- Current color scheme :
color_scheme_3

- Vertical menu background color is matching with website background:
color_scheme_1

- Secondary menu background color is matching with website background:
color_scheme_2

timogoebel:
I prefer the "current color scheme" for what it's worth.

ohadlevy:
thanks @amirfefer - great job :-)

waldenraines:
Yay! 🎉  Thanks for all the work on this everyone!

ohadlevy:
I didn't want to block this PR further, since nothing that was left was actionable in this pr (e.g. color scheme, default icon for plugin menu items, patternfly vertical nav bugs etc)-  please open a new issue instead.

ares:
oh wow! what a change, one would almost forgot to set the release in redmine after merge :-) great work everyone who participated on this!

PR #4795, opened 2017-08-29
fixes #20838 - adds redux based forms support to foreman

ohadlevy:
this pr adds basic client side validation forms using redux-form and redux-form-validations.

this specific PR implements the bookmark creation workflow.


ohadlevy:
extracted storybook related changes into #4805 - once merged it will reduced the amount of change on this PR.

ohadlevy:
@timogoebel I believe this PR is in ready to review/merge, would be interesting to see if you rebase #4597 easily on top?

timogoebel:
@ohadlevy: Excellent. I'll test and review this tomorrow if time permits.

webpack/assets/javascripts/react_app/redux/actions/common/forms.js
```diff
@@ -0,0 +1,78 @@
...
+    return response.json().then(body => {
+      throw fieldErrors(body);
+    });
+  }
+  throw new SubmissionError({
+    _error: [`${__('Error submitting data:')} ${response.status} ${__(response.statusText)}`]
+  });
+};
+
+const getcsrfToken = () => $('meta[name=csrf-token]').attr('content');
```
timogoebel: I don't think, this is the proper way to do this. Isn't there a way to do this without jQuery? How about passing this similar to the provider in the `MountingService`?

webpack/assets/javascripts/react_app/redux/actions/common/forms.js
```diff
@@ -0,0 +1,78 @@
...
+            dispatch({
+              type: `${item.toUpperCase()}_FORM_SUBMITTED`,
+              payload: { item, body }
+            })
+          )
+          .then(() =>
+            dispatch(
+              addToast({
+                type: 'success',
+                message: __(item) + __(' was successfully created')
```
timogoebel: I can see developers wanting to overwrite this.
Shouldn't we use a placeholder and something similar to `sprintf` in the translated text instead of this concatenation?

timogoebel (commented):
This looks pretty good as a first start into react based forms.
I have some minor comments, but I think we can get this in pretty fast to have a base for expansion. 👍

webpack/assets/javascripts/react_app/redux/actions/common/forms.js
ohadlevy: without jquery:
```js
document.querySelector('meta[name="csrf-token"]').content
```
i think passing it to the mount service is a lot of boilerplate code? TBH I'm not sure if its needed to begin with? i didnt re-read the auth code, but it might not even be need in some form?

ohadlevy: also, passing it as a prop in mounting sevice feels like its boilerplate code, especially since its already on the dom.

webpack/assets/javascripts/react_app/redux/actions/common/forms.js
ohadlevy: sure, i can even see that not every form submit you would not want a toast, would you mind suggesting the alternative you want?

webpack/assets/javascripts/react_app/components/common/Alert/index.js
```diff
@@ -32,10 +33,14 @@ export default ({
       onMouseEnter={onMouseEnter}
       onMouseLeave={onMouseLeave}
     >
+    {title && <strong>{title}</strong>}
+    {title && <br />}
```
waldenraines, line 37: Why is this needed?  Isn't this following http://www.patternfly.org/pattern-library/communication/toast-notifications/?

webpack/assets/javascripts/react_app/components/common/forms/Actions.js
```diff
@@ -0,0 +1,23 @@
...
+export default ({ onCancel, disabled = false, submitting = false }) => {
+  return (
+    <div className="clearfix">
+      <div className="form-actions">
+        <Button className="btn-default" disabled={disabled} onClick={onCancel}>
+          {__('Cancel')}
+        </Button>
+
+        <Button className="btn-primary" type="submit" disabled={disabled || submitting}>
+          {__('Submit')} &nbsp;
```
waldenraines: One thing we ran into in katello when doing something very similar to this in angular was that after abstracting the form controls away like this we ended up needing several variations on them.  I wonder if we should make this configurable with sane defaults.

webpack/assets/javascripts/react_app/components/common/forms/Form.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import Actions from './Actions';
+import Alert from '../Alert';
+
+export default ({
+  className = 'form-horizontal well',
```
waldenraines, line 6: Why is the form being put into a `.well`?

webpack/assets/javascripts/react_app/redux/actions/common/forms.js
waldenraines: +1 for using the non-jquery method.  It would be great if we could eventually remove jQuery entirely 😁

waldenraines (commented):
I still need to test this out, I was only able to look at the code this review.

webpack/assets/javascripts/react_app/components/common/forms/Form.js
timogoebel: ... because the ruby form helpers already do it that way.

https://github.com/theforeman/foreman/blob/16bd5715afbfdbea20f653435973908d395a8f7a/app/helpers/form_helper.rb#L284-L291

webpack/assets/javascripts/react_app/components/common/forms/Actions.js
tstrachota: Large versatility could probably be easily achieved by adding simple `SubmitButton` and `CancelButton` components and provide them as default children for the Actions component.

ohadlevy:
@tstrachota I dont have much capacity for this but would welcome the change, would you mind sending a PR to my branch?

webpack/assets/javascripts/react_app/components/common/Alert/index.js
ohadlevy: @waldenraines I suggest for us to start using patternfly-react and replace our components with theirs.

waldenraines: Agreed, see http://projects.theforeman.org/issues/21130.

webpack/assets/javascripts/react_app/components/common/forms/Form.js
waldenraines: I don't see patternfly using `.well` anywhere on their [forms](https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/form.html).  Can we remove this from both ruby form helpers and from this implementation?

ohadlevy:
@timogoebel What do you think on merging this PR? can we address the validation string extraction  (already a different redmine issue at http://projects.theforeman.org/issues/21134) in a follow up PR?

ohadlevy:
rebased.

webpack/assets/javascripts/react_app/components/common/forms/Form.js
waldenraines: @ohadlevy thoughts on this?

ohadlevy: @waldenraines TBH its been like that every since we started with bootstrap, I don't have a strong opinion one way or another, but i don't think it should be done in this pr, if we want to change it, then it should be across the app and not only in this PR.

waldenraines: I'm okay with that, see http://projects.theforeman.org/issues/21171

webpack/assets/javascripts/react_app/components/common/forms/Actions.js
waldenraines: Yeah, that's even better.  I'm fine doing that when needed to avoid premature optimization.

waldenraines (approved):
All my concerns have either been addressed or we have created new issues for them in redmine.  I am fine merging this.

timogoebel:
> can we address the validation string extraction (already a different redmine issue at http://projects.theforeman.org/issues/21134) in a follow up PR?

@ohadlevy: I would have preferred to handle the client side validation in a separate PR. But I'm not against merging this. I don't have much capacity to do a final review now, but I'll try my best.

webpack/assets/javascripts/react_app/components/common/Alert/Alert.test.js
```diff
@@ -0,0 +1,26 @@
...
+import Adapter from 'enzyme-adapter-react-15';
+configure({ adapter: new Adapter() });
+import toJson from 'enzyme-to-json';
+import { shallow } from 'enzyme';
+
+import React from 'react';
+import Alert from './';
+
+describe('alerts', () => {
+  it('alert can include a title', () => {
```
timogoebel: why not make this a sentence?
`it('can include a title', ..`

webpack/assets/javascripts/react_app/redux/actions/common/forms.js
timogoebel: I think, we can stick for this for now. But we should not concat the two stings in this way. Can we do it with a placeholder? `%s was successfully created.`
Also mind the dot after the sentence.

webpack/assets/javascripts/react_app/components/common/forms/Actions.js
timogoebel: The `&nbsp` should go into the if condition below, i.e. only be added when the spinner is shown.
Looks a little off:
image

webpack/assets/javascripts/react_app/components/common/forms/Actions.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import Button from '../../common/forms/Button';
+import { simpleLoader } from '../Loader';
+
+export default ({ onCancel, disabled = false, submitting = false }) => {
+  return (
+    <div className="clearfix">
+      <div className="form-actions">
+        <Button className="btn-default" disabled={disabled} onClick={onCancel}>
```
timogoebel, line 9: I think, these buttons could use some space in between.

timogoebel (changes requested):
@ohadlevy: Thanks. I think this looks good. I left some small comments inline.
I think, this is ready to merge. i18n is still missing, however we can do this in a separate PR. That would be a requirement for #4807, though.

ohadlevy:
@timogoebel address all of your comments, please review again - thanks!

ohadlevy:
@timogoebel tests are green :-)

webpack/assets/javascripts/react_app/components/common/Alert/Alert.test.js
```diff
@@ -0,0 +1,26 @@
...
+import React from 'react';
+import Alert from './';
+
+describe('alerts', () => {
+  it('can include a title', () => {
+    const wrapper = shallow(<Alert title="hello" type="success" />);
+
+    expect(toJson(wrapper)).toMatchSnapshot();
+  });
+  it('alert can accept childrens', () => {
```
timogoebel: ... it can accept children ;-) But never mind.

timogoebel (approved):
... while I'm not a react expert, this looks like a pretty good start for me. 👍 and thanks, @ohadlevy.

ohadlevy:
[test upgrade]

ohadlevy:
@timogoebel its all yours :)

timogoebel:
Merged, thanks @ohadlevy.
Please do rebase your other PR. :-)

PR #4796, opened 2017-08-30
Fixes #20800 - Comply with Patternfly login page recommendations

bastilian:
To comply more with the Patternfly recommendations for login pages described on http://www.patternfly.org/pattern-library/application-framework/login-page/#/design this:

* Moves alerts inline above form

gnome-shell-screenshot-bcem5y

gnome-shell-screenshot-3ido5y

* Makes the submit button not show a "loading" state (but is still disabled, to prevent resubmitting)

gnome-shell-screenshot-qu725y

The described warning on caps-lock is not yet implemented, as it is more a nice to have.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -516,9 +516,22 @@ module ApplicationHelper
   end
 
   def notifications
-    content_tag :div, :id => 'notifications', :'data-flash' => flash.to_json.html_safe do
+    content_tag :div, :id => 'notifications', :'data-flash' => flash_notifiations.to_json.html_safe do
       mount_react_component('ToastNotifications', '#notifications')
     end
+  end
+
+  def flash_notifiations
+    flash.select { |key, _| key != 'inline' }
+  end
+
+  def flash_inline
+    flash['inline'] || {}
+  end
+
+  def alert_class(type)
+    type = :danger if type == :error
+    "alert-#{type}"
   end
 
   def current_url_params(permitted: [])
--- a/app/views/users/login.html.erb
+++ b/app/views/users/login.html.erb
@@ -6,6 +6,17 @@
         <%= image_tag("login_logo.png") %>
       </div><!--/#brand-->
     </div>
+    <% if flash_inline.any? %>
+      <div class="col-sm-12">
+        <% flash_inline.each do |type, message| %>
+          <div class="alert <%= alert_class(type) %>">
+            <% header = _('Success!') if type == :success %>
+            <%= alert_header(header || "", alert_class(type)) %>
+            <%= message %>
+          </div>
+        <% end %>
+      </div>
+    <% end  %>
     <div class="col-sm-7 col-md-6 col-lg-5 login">
       <%= form_tag login_users_path, :id => "login-form", :class=>"form-horizontal" do %>
           <div class="form-group">
@@ -24,7 +35,7 @@
           </div>
           <div class="form-group">
             <div class="col-xs-offset-8 col-xs-4 submit">
-              <%= submit_tag(_("Log In").html_safe, :class => "btn btn-primary", :data => { :disable_with => _("Please wait...") }) %>
+              <%= submit_tag(_("Log In").html_safe, :class => "btn btn-primary", :data => { :disable_with => _("Log In") }) %>
             </div>
           </div>
       <% end %>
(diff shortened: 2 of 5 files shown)


app/views/users/login.html.erb
```diff
@@ -6,6 +6,17 @@
         <%= image_tag("login_logo.png") %>
       </div><!--/#brand-->
     </div>
+    <% if flash_inline.inspect %>
```
iNecas: wouldn't this always be true?

iNecas: Should be perhapse `flash_inline.present?`

app/helpers/application_helper.rb
```diff
@@ -521,6 +521,19 @@ def notifications
     end
   end
 
+  def flash_notifiations
```
iNecas, line 524: Where is this used?

app/views/users/login.html.erb
```diff
@@ -6,6 +6,17 @@
...
+      <div class="col-sm-12">
+        <% flash_inline.each do |type, message| %>
+          <div class="alert <%= alert_class type %>">
```
iNecas: `alert_class type` or `alert_class(type)`, just for sake of consistency

iNecas:
I'm not very keen on having two types of flash messages. Alternatively, we could keep the messages in `success` or `error`, and having `flash[:show_inline]` as a flag to decide if we want to show it the standard way or inline. This way we would not need to duplicate the alert methods.

app/views/users/login.html.erb
bastilian: Changed it to `.any?` overlooked that as it wouldn't show any notification anyways when looping.

app/helpers/application_helper.rb
bastilian: It was supposed to be a helper to only get non-inline flashes. Removed it now.

app/views/users/login.html.erb
bastilian: corrected.

bastilian:
@iNecas With having them under [:inline] the chance is lower for them popping also up in other places where flash is used, as most only care about the default ones (:notice, :error, ...) and need no special treatment for rejecting inline ones. The warnings should not be duplicates, they should either be inline or not.

iNecas:
Isn't there only one such place currently? Are we expecting to have this change in the future? If we used helper instead of directly `flash[:success]` (probably that's what you had in mind with the unused helper), the filtering could be done there. It would look better to me that way, but will not too forceful, if others don't see the value doing it a bit differently

ohadlevy:
Don't you get some browser error? based on https://github.com/theforeman/foreman/blob/develop/app/helpers/application_helper.rb#L518
and https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/foreman_toast_notifications.js#L20 I would expect some failure or unnecessary js code being called?

bastilian:
@ohadlevy Toast notifications don't seem to pick up on the notifications that are in inline and doesn't complain at least here. I'm going to filter them out though. ...Now i remember where i wanted to use that method I removed. :D

bastilian:
@ohadlevy The inline notifications are now filtered out. Is there any other place where these could cause issues?

ohadlevy:
@rohoover mind having a look? Do we need to adjust the colors to align to foreman colors?

Rohoover:
@ohadlevy 
For the inline notifications or the page itself?

ohadlevy:
Well, we have the option to change both it needed...I was referring to the
alerts color vs the background color.

Rohoover:
@ohadlevy 
Since these are acting as inline notifications I don't think we need to change the colors.

ohadlevy:
@bastilian mind having a look at the test failures?

bastilian:
@ohadlevy The test is fixed.

bastilian:
:/ I got this error locally as well, thinking it is an issue with npm and packages, but the same fails CI now. http://ci.theforeman.org/job/test_develop_pr_core/14137/database=postgresql,ruby=2.1,slave=fast/testReport/junit/(root)/HostJSTest__hosts%20index%20multiple%20actions/test_0001_show_action_buttons/

ohadlevy:
@bastilian can you reproduce locally?

bastilian:
@ohadlevy Yes, but it only happens in tests, not in development in the browser.

bastilian:
@ohadlevy rebased and CI is now green.

ohadlevy:
thanks @bastilian

PR #4798, opened 2017-08-30
Fixes #19116 - Remove 1.17 deprecations

dLobatog:


dLobatog:
Depends on https://github.com/Katello/katello/pull/6928 being merged first.

dLobatog (commented):
@ares @xprazak2 https://github.com/theforeman/foreman/pull/3701 and https://github.com/theforeman/foreman/pull/3983 were meant to remove `host.info`. Now we're in a situation where it can't be removed as the providers refactoring relies completely on it. 

I realize you don't want people to call `host.info` in templates, and use `host_enc` exclusively (already voiced my concerns on the mailing list, won't fight that fight again). 

This leaves us with few options:

1. Leave host.info as is. Raise an error when called from anywhere but host_enc. Keeping a reference to a file (`/renderer\.rb.*host_enc/`) is not great for maintainability and is confusing if you don't know the "background story".

2. ..? Any other suggestions?

ares:
can't we just remove it from safe mode whitelist?

mmoll:
this was done in GH-5051, or is there still something missing?

dLobatog:
#5051 did it

PR #4799, opened 2017-08-30
Fixes #20705 - display paginations section on audits page correctly

lizagilman:


iNecas:
@lizagilman please provide the screenshots of before and after changes for the UI updates, it helps with review a lot. Doing this for now from the testing:

Before:
audits-before

After:
audits-after

The positioning still doesn't feel right to me. Curious  about @Rohoover ideas

ares:
@iNecas note that the audit page should change so this pagination might be in right direction. I think we agreed with @Rohoover that audit list should be regular table.

iNecas:
Shouldn't we do it right away then?

ares:
I guess it's up to @lizagilman, I can imagine it as a separate PR as this fixes pagination, which should be brought to 1.16. Converting to table will be followed by discussion about what the table should contain, how it should look like etc

iNecas:
My only concern is we will merge this while putting converting the table to a backlog which could lead to have the page in half done state for unknown amount of time. Therefore I would like to see either the pagination fitting better (with the least amont of effort) in the page or finishng the whole table.

lizagilman:
How about this until we"ll switch to table -
image

app/views/common/_pagination.html.erb
```diff
@@ -1,5 +1,5 @@
 <form
-  class='content-view-pf-pagination table-view-pf-pagination paginate' id='pagination'
+  class='col-md-12 <%= "audits-pagination" if @page_header == "Audits" %> content-view-pf-pagination table-view-pf-pagination paginate' id='pagination'
```
iNecas: The page header might change based on localization. Could we override the color in css using dome css selector rather than doing if here?

iNecas (commented):
That works much better visualy. Let's find the best way to achieve it

dLobatog (changes requested):
Setting to WoC based on @iNecas 's comment

lizagilman:
updated
@iNecas not exactly what you suggested but at least got rid of the `if ` ...

iNecas:
Works for me. Deferring to @amirfefer to have a final word on this or whether he has other ideas to do it differently.

tbrisker (approved):
WFM. Getting the pagination to display properly with the current layout is a separate issue to changing the layout completely which will be done in the future. @rohoover - can you please ACK the layout in the latest screen shot for an intermediate stage until the entire page layout is modified?

Rohoover:
I think it's fine as an intermediate solution. Thanks!

iNecas:
There are conflicts on the branch: I'm ok with merging afterwards, unless anyone opposes it before this happens

iNecas (approved):

dLobatog (approved):

dLobatog:
Thanks @lizagilman !

PR #4801, opened 2017-08-31
Refs #15954 - pin public_suffix for Ruby 2.0

mmoll:


mmoll:
This is for 1.15 only, as 1.16 doesn't support Ruby 2.0 anyway

dLobatog:
Thanks @mmoll !

PR #4802, opened 2017-08-31
fixes #20820 - set ajax vars for cr host import

timogoebel:
When importing a host from a compute resource, the "ajax fields" architecture / os / os media etc. don't work correctly because they rely on e.g. `@architecture` being set. This moves the code from the hosts controller to a shared concern so the methods are available for `hosts_controller` and `compute_resources_vms_controller`.


dLobatog (approved):
Thanks @timogoebel , this definitely improves the import flow

PR #4803, opened 2017-08-31
fixes #20822 - vm: show link to associated host

timogoebel:
This changes the title actions for the compute resources vm show page.

Before:
image
(Associate VM doesn't work when the VM is already associated to a host)

After (with associated host):
image
Associated is gone now. Link to host was added.

After (without associated host):
image
Import button was added.

### DIFF
--- a/app/helpers/compute_resources_vms_helper.rb
+++ b/app/helpers/compute_resources_vms_helper.rb
@@ -114,6 +114,20 @@ module ComputeResourcesVmsHelper
     end
   end
 
+  def common_available_actions(vm, authorizer = nil)
+    actions = []
+    actions << vm_delete_action(vm, authorizer)
+    actions << vm_console_action(vm)
+    host_action = vm_host_action(vm)
+    if host_action
+      actions << host_action
+    else
+      actions << vm_import_action(vm, :class => 'btn btn-default')
+      actions << vm_associate_action(vm)
+    end
+    actions
+  end
+
   def openstack_available_actions(vm, authorizer = nil)
     actions = []
     if vm.state == 'ACTIVE'
@@ -125,12 +139,13 @@ module ComputeResourcesVmsHelper
       actions << vm_power_action(vm, authorizer)
     end
 
-    actions << vm_import_action(vm)
-    actions << vm_delete_action(vm, authorizer)
+    (actions + common_available_actions(vm, authorizer)).compact
   end
 
   def default_available_actions(vm, authorizer = nil)
-    [vm_power_action(vm, authorizer), vm_delete_action(vm, authorizer)]
+    actions = []
+    actions << vm_power_action(vm, authorizer)
+    (actions + common_available_actions(vm, authorizer)).compact
   end
 
   def vpc_security_group_hash(security_groups)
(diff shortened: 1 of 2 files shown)


app/helpers/compute_resources_vms_helper.rb
```diff
@@ -114,6 +114,20 @@ def available_actions(vm, authorizer = nil)
...
 
+  def common_available_actions(vm, authorizer = nil)
+    actions = []
+    actions << vm_delete_action(vm, authorizer)
+    actions << vm_console_action(vm)
+    host_action = vm_host_action(vm)
+    if host_action
+      actions << host_action
+    else
+      actions << vm_import_action(vm, :class => 'btn btn-default')
```
ares, line 125: is importing supported by all compute resource providers?

ares:
I haven't tested it yet, but I like how it moves link from view to actions helper. I'm not sure atm whether all compute resources support import. Chances are they do but I think ovirt didn't have this button for some reason (might be another bug).

app/helpers/compute_resources_vms_helper.rb
timogoebel: Yep. Just the amount of automatically filled attributes differs.
We have tests for every CR in core: https://github.com/theforeman/foreman/blob/develop/test/unit/compute_resource_host_importer_test.rb

timogoebel:
> Chances are they do but I think ovirt didn't have this button for some reason (might be another bug).

@ares: I don't have ovirt available for testing, but according to the code there should be a button: https://github.com/theforeman/foreman/blob/80dc430ed4b45794ff25c514ad03a5d3d45d2135/app/helpers/compute_resources_vms_helper.rb#L210

dLobatog (approved):
Tested with EC2, Libvirt, OpenStack successfully - I assume @timogoebel 's using VMWare so I'm going to go with merge. 

Thanks @timogoebel!

dLobatog:
@rohoover I was checking the first image in this PR and the order and color of the buttons strikes me as odd, but I didn't want to block it because of that. The 2 things I don't like are:

* "Dangerous" actions should go last, not first. I think both power off and delete qualify for having a button
* "Console" shouldn't be the blue button here, it's like a call for action which IMO is wrong. It should be "Associate VM", which is almost always a positive thing to do. "Host" or "Import" shouldn't be blue either.

🚥  green light for this?

timogoebel:
@dLobatog & @Rohoover : I think a lot of users don't even know this page exists, but I saw that as well. It needs more love than this PR.
I think, the issue with the button colors is similar to the host page: That kind of looks very odd to me as well and breaks the layout with some plugins installed:
image

Maybe we can find a better solution for both pages.

Rohoover:
Sorry for the delayed response guys! Just saw this...

I agree with all your sentiments. What are next steps for this if not handled in this PR.

timogoebel:
@Rohoover: What do you think, if we would put some of these buttons into a "drawer".
E.g. there is only one button called "Actions" and that reveals more actions like scheduling a remote job etc.?
But I guess, we should open a Redmine for any follow-up tasks.

Rohoover:
I'm not against that idea. If my memory serves correct, we do this on other pages... with main actions broken out separate.

PR #4804, opened 2017-09-03
Fixes #13599 - Wrapped exception truncate moved to client

ShimShtein:
Removed message truncation from server side, and moved it to client side - the message is truncated with ellipsis. On mouse over full message is presented as popover.

This commit required a small change to `data-ajax-url` handler in client to enable proper tooltip presentation in HTML loaded by AJAX.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -73,6 +73,7 @@ function onContentLoad(){
       if ($(this).data('on-complete')){
         _.get(window, $(this).data('on-complete')).call(null, this, status);
       }
+      tfm.tools.activateTooltips(this)
     });
   });
 
--- a/lib/foreman/exception.rb
+++ b/lib/foreman/exception.rb
@@ -52,7 +52,7 @@ module Foreman
       super unless @wrapped_exception.present?
 
       cls = @wrapped_exception.class.name
-      msg = @wrapped_exception.message.try(:truncate, 90)
+      msg = @wrapped_exception.message
       super + " ([#{cls}]: #{msg})"
     end
   end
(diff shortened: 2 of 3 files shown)


app/assets/javascripts/application.js
```diff
@@ -73,6 +73,7 @@ function onContentLoad(){
       if ($(this).data('on-complete')){
         _.get(window, $(this).data('on-complete')).call(null, this, status);
       }
+      tfm.tools.activateTooltips(this)
```
ohadlevy, line 76: maybe just default to loading on content load (instead of null in the call function above)?

ShimShtein: The function above is a user-defined hook - each consumer can define a method to call on complete. I don't see a reason why tooltips code should be broken for AJAX, unless the user specifies a hook explicitly.

ShimShtein:
[test foreman]

lib/foreman/exception.rb
```diff
@@ -52,7 +52,7 @@ def message
       super unless @wrapped_exception.present?
 
       cls = @wrapped_exception.class.name
-      msg = @wrapped_exception.message.try(:truncate, 90)
+      msg = @wrapped_exception.message
       super + " ([#{cls}]: #{msg})"
```
lzap, line 56: Nitpick - string `Foreman::WrappedException` is very long and it's eating valuable "visible space" on page, I suggest to drop showing class name if it is one of `Exception`, `RuntimeException`, `Foreman::Exception` or `Foreman::WrappedException`. Also the class name could be after the actual message.

A change would be necessary here and in the wrapped exception as well I think.

ShimShtein: I would prefer not changing format of the message, since it can be parsed by user scripts and this will break it.
Maybe we can omit wrapping messages that inherit from `ForemanException` completely - just pass through the original message - it already contains meaningful information.
What do you think @lzap?

lzap: Omitting wrapping is killing it's main purpose, please no. Let's keep the message as is then.

lzap (approved):
LGTM, I have no comment on the JS part tho.

iNecas:
@ohadlevy good to merge?

app/assets/javascripts/application.js
ohadlevy: this might lead to activeTooltips running multiple times, but I don't think that should block the pr. :+1: from my side.

iNecas:
Thanks @ShimShtein

PR #4806, opened 2017-09-04
Fixes #20835 - added time into foreman-debug

lzap:
This patch will add

    [SECONDS] Verbose message

so we can easily measure what is slow in foreman-debug. Only appears when providing `-v` option.


dLobatog (approved):
Thanks @lzap !

PR #4807, opened 2017-09-04
fixes #20839 - react implementation for bookmark dropdown

ohadlevy:


ohadlevy:
this PR includes commit from #4795, once merged I'll rebase.

ohadlevy:
@timogoebel I believe this is now ready too.

ohadlevy:
before

recorded

after
after

webpack/assets/javascripts/react_app/components/bookmarks/form/index.js
```diff
@@ -0,0 +1,59 @@
...
+        disabled={submitting}
+        submitting={submitting}
+        error={error}
+      >
+        <TextField
+          name="name"
+          type="text"
+          required="true"
+          label={__('Name')}
+          validate={[required(), length({ max: 254 })]}
```
ares, line 40: Out of curiosity, shoukd we duplicate validation definitions or should we try ti generate it here based on what's defined in model? While this one is easy, my concern is with validation on uniqueness or custom validation messages. Is the plan to support it in UI too or should we leave it up to server to validate it on save?

webpack/assets/javascripts/react_app/components/bookmarks/form/index.js
```diff
@@ -0,0 +1,59 @@
...
+        onSubmit={handleSubmit(this.submit)}
+        onCancel={onCancel}
+        disabled={submitting}
+        submitting={submitting}
+        error={error}
+      >
+        <TextField
+          name="name"
+          type="text"
+          required="true"
```
ares, line 38: Can we somehow reuse the existing helper that determines whether the field is requires based on model validation?

webpack/assets/javascripts/react_app/components/bookmarks/form/index.js
```diff
@@ -0,0 +1,59 @@
...
+          label={__('Name')}
```
ares, line 39: Could this use model human attribute names?

ares (commented):
Thanks @ohadlevy, it looks like a great start. I quickly read through, I don't feel for a real review, but I had few question about how to better integrate existing definitions.

webpack/assets/javascripts/react_app/components/bookmarks/form/index.js
ohadlevy: since this is purely data driven, we could and we should probably do that, but i thought it should not be forced in the same pr.

ohadlevy: but take note that we don't need necessarily to map rails models -> form elements.

webpack/assets/javascripts/react_app/components/bookmarks/form/index.js
ohadlevy: the way it works now is that we try to submit, if we get back a 422 error, then we use the json error message and add it to the form - see https://github.com/theforeman/foreman/pull/4807/files#diff-35521d7d9795ca3a7f62a33bb3ceef6aR20 and a test https://github.com/theforeman/foreman/pull/4807/files#diff-af7b70459125c57d18f34cd03710da9cR71

ohadlevy: and a demo of that:

recorded

webpack/assets/javascripts/react_app/components/bookmarks/form/index.js
ohadlevy: I'm not sure tbh, what would be the difference?

ares: avoid duplication of strings which helps in those rare cases we change some names, human names are used also in rendering validation messages so it would be easy to forget about

webpack/assets/javascripts/react_app/components/bookmarks/form/index.js
ares: if this is building a foundation for new generation of forms, I'd prefer to have it solved from the first day, otherwise we'd build technical debt to change it in all forms later

webpack/assets/javascripts/react_app/components/bookmarks/form/index.js
ares: ok that's nice. So we have two levels of validation, which I think is correct, but I'd like to have single definition if possible.

config/routes.rb
```diff
@@ -110,7 +110,7 @@
       end
     end
 
-    resources :bookmarks, :except => [:show] do
+    resources :bookmarks, :except => [:show, :new, :create] do
```
timogoebel, line 113: How about using`only` over `except`?

webpack/assets/javascripts/react_app/components/bookmarks/index.js
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  render() {
+    const { controller, url, showModal, modalClosed, canCreate } = this.props;
+
+    return showModal ? (
+      <SearchModal show={showModal} controller={controller} url={url} onHide={modalClosed} />
+    ) : (
+      <Dropdown pullRight id={controller}>
+        <Dropdown.Toggle />
```
timogoebel, line 45: I can not close the drop-down anymore once it's open on the bookmark index page.

webpack/assets/javascripts/react_app/components/bookmarks/form/index.js
timogoebel, line 40: This isn't translated correctly.

app/views/common/_searchbar.html.erb
```diff
@@ -8,39 +8,14 @@
...
-      </div>
-    </div><!-- /.modal-content -->
-  </div><!-- /.modal-dialog -->
-</div><!-- /.modal -->
+<%= mount_react_component('BookmarkContainer', '#bookmark-form', {
+          controller: auto_complete_controller_name,
+          url: api_bookmarks_path,
+          canCreate: authorizer.can?(:create_bookmarks),
+          documentationUrl: documentation_url("4.1.5Searching")
+        }.to_json) %>
```
timogoebel: Maybe we should do the `.to_json` in `mount_react_component` if it's passed a Hash.

webpack/assets/javascripts/react_app/components/bookmarks/form/index.js
```diff
@@ -0,0 +1,59 @@
...
+          validate={[required(), length({ max: 254 })]}
+        />
+        <TextField
+          name="query"
+          type="textarea"
+          required="true"
+          label={__('query')}
```
timogoebel: Capitalize `query` ?

timogoebel (commented):
... some comments as an initial review.

config/routes.rb
ohadlevy: only: [:index, :update, :edit, :destroy] is better? (don't really care tbh)

app/views/common/_searchbar.html.erb
ohadlevy: if we do that, its fine, but in another PR please :)

config/routes.rb
timogoebel: I think we should eventually reduce this to `only: [:index]`. But except is fine for now as well.

webpack/assets/javascripts/react_app/components/bookmarks/index.js
ohadlevy: this happens only the bookmark page right? any idea why? (can't find anything specific for this page).

webpack/assets/javascripts/react_app/components/bookmarks/Bookmark.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import URI from 'urijs';
+import { MenuItem } from 'react-bootstrap';
+
+function followLink() {
+  let uri = new URI(window.location.href);
+
+  uri.setSearch('search', this.trim());
+  /* eslint-disable no-undef */
+  Turbolinks.visit(uri.toString());
```
waldenraines: Is it worth adding this dependency to turbolinks if we are going to remove it?

webpack/assets/javascripts/react_app/components/common/Icon/Icon.consts.js
```diff
@@ -4,6 +4,7 @@ const infoIcon = 'pficon pficon-info';
 const warningIcon = 'pficon pficon-warning-triangle-o';
 const errorIcon = 'pficon pficon-error-circle-o';
 const closeIcon = 'pficon pficon-close';
+const questionSign = 'glyphicon glyphicon-question-sign';
```
waldenraines, line 8: From the patternfly [icons page](https://www.patternfly.org/styles/icons/):

>  Bootstrap Glyphicons are also included in PatternFly, but are not recommended for use

Maybe use `fa fa-question`?

webpack/assets/javascripts/react_app/components/common/forms/Button.js
```diff
@@ -1,10 +1,10 @@
 import React from 'react';
 
-export default ({ className, onClick, children, disabled = false }) => {
+export default ({ className, onClick, children, type = 'button', disabled = false }) => {
```
waldenraines: 👍 as not defaulting to `button` can cause a lot of problem in forms for cancel buttons, etc.

webpack/assets/javascripts/react_app/components/common/forms/Form.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import Actions from './Actions';
+import Alert from '../Alert';
+
+export default ({
+  className = 'form-horizontal well',
```
waldenraines: I mentioned this on the forms PR but I'm not a fan of putting our forms into wells.  It's not [recommended by patternfly](https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/form.html) and it looks odd to me.

waldenraines (changes requested):

webpack/assets/javascripts/react_app/components/bookmarks/Bookmark.js
ohadlevy: yes, I don't think we are going to remove it any time soon, also its consistent with the rest of the page (e.g. when you click on search, or when you use the pagination controllers) - I can't see why clicking on the bookmark should do anything different.

webpack/assets/javascripts/react_app/components/common/Icon/Icon.consts.js
ohadlevy: that's fine - I was keeping the original behavior.

webpack/assets/javascripts/react_app/components/common/forms/Button.js
ohadlevy: also the name would be very confusing ;)

webpack/assets/javascripts/react_app/components/common/forms/Form.js
ohadlevy: as i replied, lets discuss this outside the scope of this pr.

waldenraines: 👍

timogoebel:
[test foreman]

timogoebel:
@ohadlevy: Thanks a lot for this. Just a small comment. I'll do the rest inline.

The Dropdown has now a spacing, which should be removed.

Before:
image

After:
image

ohadlevy:
rebased

waldenraines:
[test foreman]

ohadlevy:
@ehelms this PR is failing due to the fact that we are unable to find isomorphic-fetch npm package, yet that package is a dependent of fbjs which is a dependent of react. so should we now expose every package we use as a top level package in package.json? i find it a bit strange that we can't leverage existing assets rather to repackage them? /cc @dLobatog @ekohl

ekohl:
@ohadlevy to me it makes sense that if you use it in your app then you specify it as a dependency. Since https://github.com/theforeman/foreman/blob/59e99bf095d432c725fd0efa6ee0e48d6f4a3539/webpack/assets/javascripts/react_app/redux/actions/common/forms.js#L2 does use it, it should be in package.json. What makes it logical for me is that fbjs could decide to drop it and then the application would break.

waldenraines:
[test]

webpack/assets/javascripts/react_app/components/bookmarks/index.js
```diff
@@ -0,0 +1,60 @@
...
+class BookmarkContainer extends React.Component {
+  componentDidMount() {
+    const { url, controller, getBookmarks } = this.props;
+
+    getBookmarks(url, controller);
+  }
+
+  onClick({ preventDefault }) {
+    preventDefault;
+    this.props.showModal ? this.props.modelClosed() : this.props.modalOpened();
```
ShimShtein: Typo: `modelClosed` should be `modalClosed`

ohadlevy:
rebased.

waldenraines:
Other than another rebase and a few minor requested changes what else needs to be done here before this can be merged?

ohadlevy:
rebased.
@danseethaler there a few linting issues which I'm not sure how to resolve, would appreciate if you can look at the hound failures please.

@waldenraines the missing part is that currently validation errors are in english only.

webpack/assets/javascripts/react_app/components/bookmarks/form/index.js
```diff
@@ -0,0 +1,63 @@
...
+import { reduxForm } from 'redux-form';
+import { connect } from 'react-redux';
+import { required, length } from 'redux-form-validators';
+import Form from '../../common/forms/Form';
+import TextField from '../../common/forms/TextField';
+import * as FormActions from '../../../redux/actions/common/forms';
+
+class BookmarkForm extends React.Component {
+  // if we got here, its already post client validations
+  submit({ name, query, publik }, dispatch, props) {
```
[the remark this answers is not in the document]
danseethaler: This should be extracted out of the component to a separate function if it's not tied to the component itself. i.e. not using `this`.

webpack/assets/javascripts/react_app/components/bookmarks/index.js
```diff
@@ -0,0 +1,58 @@
...
+
+class BookmarkContainer extends React.Component {
+  componentDidMount() {
+    const { url, controller, getBookmarks } = this.props;
+
+    getBookmarks(url, controller);
+  }
+
+  onClick() {
+    this.props.showModal ? this.props.modalClosed() : this.props.modalOpened();
```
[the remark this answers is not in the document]
danseethaler: This will be fixed by not using a ternary and just using a regular if/else.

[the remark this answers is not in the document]
ohadlevy: Ok... but why is this syntax is wrong?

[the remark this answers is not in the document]
danseethaler: Ternary operators are actually for assignment - not really for shorthand if/else (even though they're often used that way). They actually return a value and the linting rule says you're ignoring the return value (since it's not being assigned/used) which could be a logic error.

ohadlevy:
thanks @danseethaler - updated based on your comments.

ohadlevy:
[test foreman]

ohadlevy:
rebased - not 100% sure what to do with code climate comments ATM.

ares:
Code climate comments don't have to be green, it's just handy for reviewers to see the scan results.

webpack/assets/javascripts/react_app/common/reduxFormI18n.js
```diff
@@ -0,0 +1,7 @@
+import Validators from 'redux-form-validators';
+// import forEach from 'lodash/forEach';
+
+// extract all default strings in redux-form-validations
+// forEach(Validators.messages, ({ defaultMessage }) => window.n__(defaultMessage));
+
+Validators.formatMessage = ({ defaultMessage }) => __(defaultMessage);
```
ohadlevy: @waldenraines here is a poor mean solution to the problem.

I didn't address the variable translation, e.g. ideas welcome :)
```»is too long (maximum is {count, number} {count, plural, one {character} other {characters}})«```

**Comment by ohadlevy** [comment]
rebased, @sharvit mind reviewing (in light of your recent forms work?) thanks!

File: webpack/assets/javascripts/react_app/components/bookmarks/form/index.js [hunk 16]
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
+import { reduxForm } from 'redux-form';
+import { connect } from 'react-redux';
+import { required, length } from 'redux-form-validators';
+import Form from '../../common/forms/Form';
```
sharvit, line 5: I don't see a reason to use this `Form` component,
it doesn't give such a benefit and takes away flexibility.

* Submit and Cancel buttons should be in the `<Modal.Footer>`
* Submit and Cancel buttons should be in reverse order when inside `<Modal>`
* You should use `react-bootstrap`/`patternfly-react` `<Form horizontal></Form>`
* The form has some background that shouldn't be exist

http://www.patternfly.org/pattern-library/forms-and-controls/buttons-on-forms/
http://rawgit.com/sharvit/patternfly-react/storybook/forms/index.html?knob-Show%20Modal=true&knob-Show%20Feedback%20Icon=true&knob-Show%20Labels=false&knob-Show%20Help=true&knob-Show%20Loading=false&knob-Disabled=false&knob-Label=Danger%20Will%20Robinson%21&selectedKind=Forms&selectedStory=Modal%20Form&full=0&down=1&left=1&panelRight=0&downPanel=storybooks%2Fstorybook-addon-knobs

webpack/assets/javascripts/react_app/components/bookmarks/form/index.js
```diff
@@ -0,0 +1,65 @@
...
+
+    return (
+      <Form
+        onSubmit={handleSubmit(submit)}
+        onCancel={onCancel}
+        disabled={submitting}
+        submitting={submitting}
+        error={error}
+      >
+        <TextField
```
sharvit, line 35: I don't see the benefit of having a `TextField` >> `Field` >> `renderField` >> `CommonForm`.
I think it is unnecessary complexity.
Why won't `Forms` use `Field`>>`CommonForm`?

I start seeing the benefit of exporting form fields components from `patternfly-react` so they will replace the `CommonForm`.

This is how it should be done using `react-bootstrap`/`patternfly-react` components:
```js
<Form horizontal>
  <FormGroup controlId="name" disabled={false} validationState="error">
    <Col componentClass={ControlLabel()} sm={3}>
      Name
    </Col>
    <Col sm={9}>
      <FormControl type="text" disabled={false} />
      <HelpBlock>Error here...</HelpBlock>
    </Col>
  </FormGroup>
</Form>
```

http://rawgit.com/sharvit/patternfly-react/storybook/forms/index.html?knob-Validation%20State=error&knob-Show%20Modal=true&knob-Show%20Feedback%20Icon=false&knob-Right%20aligned=false&knob-Show%20Labels=false&knob-Show%20Help=true&knob-Stacked=false&knob-Show%20Loading=false&knob-Disabled=false&knob-Style=default&knob-Label=Danger%20Will%20Robinson%21&selectedKind=Forms&selectedStory=Horizontal%20Form&full=0&down=1&left=1&panelRight=0&downPanel=storybooks%2Fstorybook-addon-knobs

webpack/assets/javascripts/react_app/components/bookmarks/index.js
```diff
@@ -0,0 +1,62 @@
...
+
+  render() {
+    const {
+      controller, url, showModal, modalClosed, canCreate,
+    } = this.props;
+
+    return showModal ? (
+      <SearchModal show={showModal} controller={controller} url={url} onHide={modalClosed} />
+    ) : (
+      <Dropdown pullRight id={controller}>
```
sharvit, line 44: Will be easier and cleaner to use `DropdownButton`
http://rawgit.com/sharvit/patternfly-react/storybook/dropdown-buttons/index.html?knob-Label=Danger%20Will%20Robinson%21&knob-Style=default&selectedKind=Button&selectedStory=DropdownButton&full=0&down=1&left=1&panelRight=0&downPanel=storybooks%2Fstorybook-addon-knobs

webpack/assets/javascripts/react_app/components/common/DocumentationLink/index.js
```diff
@@ -0,0 +1,15 @@
...
+import { MenuItem } from 'patternfly-react';
+import Icon from '../Icon';
+
+function onClickedLink(e) {
+  e.preventDefault();
+  window.open(this, '_blank');
+}
+
+export default ({ href, id = 'documentationLink' }) => (
+  <MenuItem key="documentationUrl" id={id} href={href} onClick={onClickedLink.bind(href)}>
```
sharvit: tbh, I prefer the `onClickedLink` inside the component and without `bind`, without `this` and with ES2015 syntax.

webpack/assets/javascripts/react_app/components/bookmarks/form/index.js
ohadlevy: @sharvit thanks, comments below:
> I don't see a reason to use this Form component, it doesn't give such a benefit and takes away flexibility.

One of the goals of using this, is to reduce the differences across forms, if you look at the implementation you would see its something pretty trivial to copy paste, and give us the ability to change the internal implementation if like in the future without changing all the forms.

> Submit and Cancel buttons should be in the <Modal.Footer>

True, I've noticed it as well, but it didnt seem to make a significant difference, and IMHO the point above is more important.

> Submit and Cancel buttons should be in reverse order when inside <Modal>

Why?

> You should use react-bootstrap/patternfly-react <Form horizontal></Form>

I'll have a look  once its part of pf release?

> The form has some background that shouldn't be exist

? do you mean the well css class?

webpack/assets/javascripts/react_app/components/bookmarks/form/index.js
ohadlevy: > I start seeing the benefit of exporting form fields components from patternfly-react so they will replace the CommonForm.

Once all of the features around validation errors, tool tips etc are in pf, than :+1:

webpack/assets/javascripts/react_app/components/bookmarks/index.js
ohadlevy: is it merged already? also note that this is part of a button (where the rest of the button is implemented by rails)

webpack/assets/javascripts/react_app/components/common/DocumentationLink/index.js
ohadlevy: mind providing an example?

sharvit:
```js
import React from 'react';
import { MenuItem } from 'patternfly-react';
import Icon from '../Icon';

export default ({ href, id = 'documentationLink' }) => {
  const handleClick = (e) => {
    e.preventDefault();
    window.open(href, '_blank');
  };

  return (
    <MenuItem key="documentationUrl" id={id} href={href} onClick={handleClick}>
      <Icon type="question-sign" className="icon-black" />
      {` ${__('Documentation')}`}
    </MenuItem>
  );
};
```

webpack/assets/javascripts/react_app/components/bookmarks/index.js
sharvit: You right, it won't work without the `title` prop :+1:

webpack/assets/javascripts/react_app/components/bookmarks/form/index.js
sharvit: > One of the goals of using this, is to reduce the differences across forms, if you look at the implementation you would see its something pretty trivial to copy paste, and give us the ability to change the internal implementation if like in the future without changing all the forms.

The form component doesn't do much, just adding buttons and errors to your form.
I think that in forms, the right way to reduce differences is to have components that integrate well with each other instead of having 1 component that handles everything.
For example, in some cases you would want to use different buttons.

>> Submit and Cancel buttons should be in the <Modal.Footer>
>
> True, I've noticed it as well, but it didnt seem to make a significant difference, and IMHO the point above is more important.
>
>> Submit and Cancel buttons should be in reverse order when inside
>
> Why?

Only because patternfly documented it:
http://www.patternfly.org/pattern-library/forms-and-controls/buttons-on-forms/

>> You should use react-bootstrap/patternfly-react
>
> I'll have a look once its part of pf release?

Ok, but it is possible to use it from `react-bootstrap`. Then when pf-react will merge it will just be a matter of replacing the import statements.

>> The form has some background that shouldn't be exist
>
> ? do you mean the well css class?

Yes

webpack/assets/javascripts/react_app/components/bookmarks/form/index.js
sharvit: :+1:

ohadlevy:
@timogoebel I think its ready to re-testing... any chance you can another look?

webpack/assets/javascripts/react_app/components/bookmarks/form/index.js
ohadlevy: it is now :)

ohadlevy:
open / closed as travis tests were not running from some reason.

ohadlevy:
And it's green too :)

webpack/assets/javascripts/react_app/common/reduxFormI18n.js
waldenraines: @ohadlevy that is the same direction I was going in for this as well and I never got the variables to work either.  I'm sure there's a better way but I am not sure what it is.

webpack/assets/javascripts/react_app/components/bookmarks/form/index.js
waldenraines: > Only because patternfly documented it:
> http://www.patternfly.org/pattern-library/forms-and-controls/buttons-on-forms/

I believe I remember hearing that they (or someone else) did some research and found that confirmation buttons on the right was more common and user friendly (even though we deal heavily in linux and linux tends to have them on the left).

waldenraines: > > ? do you mean the well css class?
> 
> Yes

Can we please take this opportunity to remove the `.well`?

timogoebel:
@ohadlevy: I tried to test this, but it always blew up in my face :-(

When I click on "Bookmark this search":

```
Uncaught TypeError: Cannot read property 'contains' of undefined
    at contains.js:17
    at Modal.focus (Modal.js:557)
    at Modal.onShow (Modal.js:486)
    at Modal.componentDidMount (Modal.js:230)
    at commitLifeCycles (react-dom.development.js:11505)
    at commitAllLifeCycles (react-dom.development.js:12294)
    at HTMLUnknownElement.callCallback (react-dom.development.js:1299)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:1338)
    at invokeGuardedCallback (react-dom.development.js:1195)
    at commitAllWork (react-dom.development.js:12415)
(anonymous) @ contains.js:17
focus @ Modal.js:557
onShow @ Modal.js:486
componentDidMount @ Modal.js:230
commitLifeCycles @ react-dom.development.js:11505
commitAllLifeCycles @ react-dom.development.js:12294
callCallback @ react-dom.development.js:1299
invokeGuardedCallbackDev @ react-dom.development.js:1338
invokeGuardedCallback @ react-dom.development.js:1195
commitAllWork @ react-dom.development.js:12415
workLoop @ react-dom.development.js:12687
callCallback @ react-dom.development.js:1299
invokeGuardedCallbackDev @ react-dom.development.js:1338
invokeGuardedCallback @ react-dom.development.js:1195
performWork @ react-dom.development.js:12800
batchedUpdates @ react-dom.development.js:13244
performFiberBatchedUpdates @ react-dom.development.js:1646
stackBatchedUpdates @ react-dom.development.js:1637
batchedUpdates @ react-dom.development.js:1651
batchedUpdatesWithControlledComponents @ react-dom.development.js:1664
dispatchEvent @ react-dom.development.js:1874
react-dom.development.js:8305 The above error occurred in the <Modal> component:
    in Modal (created by Modal)
    in Modal
    in Unknown (created by BookmarkContainer)
    in BookmarkContainer (created by Connect(BookmarkContainer))
    in Connect(BookmarkContainer)
    in Provider

Consider adding an error boundary to your tree to customize error handling behavior.
You can learn more about error boundaries at https://fb.me/react-error-boundaries.
```

Also: This button still has a superfluous spacing.
image

Could you please take a look?

ohadlevy:
Did you do npm install ?

timogoebel:
> Did you do npm install ?

Yes. Sure. Twice.

timogoebel:
btw: The margin comes from here:

```
/* line 46, /Users/f0218175/foremandev/foreman/app/assets/stylesheets/navigation.scss */
.dropdown .caret {
  margin-left: 4px;
}
```

ohadlevy:
Thanks, I'm currently afk but will have a look later on this evening. How
about storybook?

timogoebel:
@ohadlevy: Thanks. https://github.com/react-bootstrap/react-bootstrap/issues/1747 might have some clues about what this is about.

ohadlevy:
@timogoebel I think I've addressed the issue (needed a newer version of react-overlay). the css fix is not the one afaiu?

timogoebel:
Thanks. This fixes the issue for me.
If we can fix the CSS, I believe, this is ready to go from my side.

This fixes the issue for me:
```diff
diff --git a/app/assets/stylesheets/navigation.scss b/app/assets/stylesheets/navigation.scss
index 8f557d132..bc107a47c 100644
--- a/app/assets/stylesheets/navigation.scss
+++ b/app/assets/stylesheets/navigation.scss
@@ -43,8 +43,10 @@
   }
 }

-.dropdown .caret {
-  margin-left: 4px;
+.navbar {
+  .dropdown .caret {
+    margin-left: 4px;
+  }
 }
```

One more remark: I'd prefer if the "Submit / Cancel" button were actually part of the modal footer (it was that way before this PR, now it is part of the modal body), but I can live with the current approach for now.

ohadlevy:
@timogoebel thanks, I've updated the CSS.

BTW: do you think its worth adding a spinner to the dropdown while bookmark are being loaded? if you do, I dont mind adding it as a follow up PR (in case you are planning to merge this :))

webpack/assets/javascripts/react_app/components/bookmarks/index.js
timogoebel: This seems to be fixed now. :-)

timogoebel (approved):
Looks very good and the new spinner is also a good addition. I see a lot of benefits in moving the UI to react. This is a good step in the right direction.

👍 from me

timogoebel:
Merged, thanks @ohadlevy.

And thanks for helping with reviewing this @ares, @sharvit, @waldenraines @ShimShtein and @danseethaler.

danseethaler:
Great work @ohadlevy and @timogoebel!

waldenraines:
🏁 :clinking_glasses: 🎉 💯

PR #4808, opened 2017-09-04
Fixes #20098 - Resolve template_used for image provisioning

dLobatog:
At the moment, the method Hostext::OperatingSystem#template_kinds is
unable to handle image-based provisioning. If there is an image, it will
not be able to find it as 'self.compute_attributes' does not exist for
new hosts.

Creating the host always worked well, and the templates can be found
after clicking the 'submit' button. However, when you click on
'Resolve', you get an error 'undefined method []' as compute_attributes
does not exist.

This is fixed by ensuring the controller includes compute_attributes in
the 'temporary' host created to check the templates this.

To reproduce:

 - Click on Hosts > New host
 - Select a Compute Resource
 - Select an OS, choose 'Image based' as provisioning method
 - Select an image
 - Click on 'Resolve' - you will see "No templates were configured"
 - After applying the patch, you should see the finish/userdata template

### DIFF
--- a/app/controllers/hosts_controller.rb
+++ b/app/controllers/hosts_controller.rb
@@ -935,7 +935,10 @@ class HostsController < ApplicationController
     # kind of host. For example 'is_owned_by', even if it's in host_params,
     # should be ignored by Host::Discovered or any other Host class that does
     # not have that attribute
-    host_attributes = host.class.attribute_names
+    host_attributes = host.class.attribute_names.dup
+    if host_params["compute_attributes"].present?
+      host_attributes << 'compute_attributes'
+    end
     host_params.select do |k,v|
        host_attributes.include?(k) && !k.end_with?('_ids')
     end.except(:host_parameters_attributes)
--- a/test/controllers/hosts_controller_test.rb
+++ b/test/controllers/hosts_controller_test.rb
@@ -1263,38 +1263,47 @@ class HostsControllerTest < ActionController::TestCase
     assert_template 'review_before_build'
   end
 
-  test 'template_used returns templates with interfaces' do
-    @host.setBuild
-    nic=FactoryGirl.create(:nic_managed, :host => @host)
-    attrs = host_attributes(@host)
-    attrs[:interfaces_attributes] = nic.attributes.except 'updated_at', 'created_at', 'attrs'
-    ActiveRecord::Base.any_instance.expects(:destroy).never
-    ActiveRecord::Base.any_instance.expects(:save).never
-    xhr :put, :template_used, {:provisioning => 'build', :host => attrs, :id => @host.id }, set_session_user
-    assert_response :success
-    assert_template :partial => '_provisioning'
-  end
+  describe '#template_used' do
+    setup do
+      @host.setBuild
+      ActiveRecord::Base.any_instance.expects(:destroy).never
+      ActiveRecord::Base.any_instance.expects(:save).never
+      @attrs = host_attributes(@host)
+    end
 
-  test 'template_used returns templates with host parameters' do
-    @host.setBuild
-    attrs = host_attributes(@host)
-    attrs[:host_parameters_attributes] = {'0' => {:name => 'foo', :value => 'bar', :id => '34'}}
-    ActiveRecord::Base.any_instance.expects(:destroy).never
-    ActiveRecord::Base.any_instance.expects(:save).never
-    xhr :put, :template_used, {:provisioning => 'build', :host => attrs }, set_session_user
-    assert_response :success
-    assert_template :partial => '_provisioning'
-  end
+    test 'returns templates with interfaces' do
+      nic=FactoryGirl.create(:nic_managed, :host => @host)
+      @attrs[:interfaces_attributes] = nic.attributes.except 'updated_at', 'created_at', 'attrs'
+      xhr :put, :template_used, {:provisioning => 'build', :host => @attrs, :id => @host.id }, set_session_user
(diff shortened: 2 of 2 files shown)


test/controllers/hosts_controller_test.rb
```diff
@@ -1263,38 +1263,47 @@ class Host::Valid < Host::Managed; end
...
+    end
+
+    test 'does not save has_many relations on existing hosts' do
+      @attrs[:config_group_ids] = [config_groups(:one).id]
+      xhr :put, :template_used, {:provisioning => 'build', :host => @attrs, :id => @host.id }, set_session_user
+      assert_response :success
+      assert_template :partial => '_provisioning'
+    end
+
+    test 'shows templates for image provisioning' do
```
dLobatog, line 1296: This is the only bit that was added. The rest is merely refactoring a few lines of each `template_used` related test into a `setup` block.

timogoebel:
Tests are currently failing. I think, the reason for this is:

`compute_attributes` is defined here and therefore only available for a `Host::Managed`:
https://github.com/theforeman/foreman/blob/05dc994fa97f9d92091c92c4965b9489dc39f9ff/app/models/concerns/orchestration/compute.rb#L8

dLobatog:
Not sure how come tests could be failing for many controllers who have nothing to do with this. I have a suspicion changing `@attrs` is affecting other controllers, why this variable is not cleared after running the test? I'm not quite sure yet.

lzap:
I can confirm misbehavior locally, it randomly fails tests depending on how Minitest orders execution.

lzap:
Weird, was trying to trial and error to rule out particular test, but don't understand what is wrong.

app/controllers/hosts_controller.rb
```diff
@@ -936,6 +936,7 @@ def host_attributes_for_templates(host)
     # should be ignored by Host::Discovered or any other Host class that does
     # not have that attribute
     host_attributes = host.class.attribute_names
+    host_attributes << 'compute_attributes' if host.managed
```
dLobatog: Just introduced this - let's see if it makes a difference (I was unable to reproduce the failures consistently)

dLobatog (commented):
I can consistently get tests to pass when I remove the tests I added (no changes to the controller part) - and they fail when I add them back. So there's got to be something in my tests that make the others fail.

test/controllers/hosts_controller_test.rb
```diff
@@ -1263,38 +1263,49 @@ class Host::Valid < Host::Managed; end
...
-    ActiveRecord::Base.any_instance.expects(:destroy).never
-    ActiveRecord::Base.any_instance.expects(:save).never
-    xhr :put, :template_used, {:provisioning => 'build', :host => attrs, :id => @host.id }, set_session_user
-    assert_response :success
-    assert_template :partial => '_provisioning'
-  end
+  describe '#template_used' do
+    setup do
+      @host.setBuild
+      ActiveRecord::Base.any_instance.expects(:destroy).never
```
ohadlevy, line 1269: dont you need to remove this mocking once the tests are done? this might remain after the test is done?

app/controllers/hosts_controller.rb
```diff
@@ -936,6 +936,9 @@ def host_attributes_for_templates(host)
     # should be ignored by Host::Discovered or any other Host class that does
     # not have that attribute
     host_attributes = host.class.attribute_names
+    if Foreman::Cast.to_bool(params[:host][:managed])
+      host_attributes << 'compute_attributes'
```
iNecas: This would cause the `compute_attributes` item to be added to `Host.class.attribute_names` every time we call `host_attributes_for_templates`. You need to use `dup`

dLobatog: That was the key - thanks @iNecas ! 🤕

app/controllers/hosts_controller.rb
```diff
@@ -935,7 +935,10 @@ def host_attributes_for_templates(host)
     # kind of host. For example 'is_owned_by', even if it's in host_params,
     # should be ignored by Host::Discovered or any other Host class that does
     # not have that attribute
-    host_attributes = host.class.attribute_names
+    host_attributes = host.class.attribute_names.dup
```
dLobatog, line 938: This was the key (not using .dup here) - thanks again @iNecas

test/controllers/hosts_controller_test.rb
dLobatog: Doesn't look like that's the case, other tests are able to `.destroy` just fine after this, the expectations are automatically cleared after each of the test.

iNecas:
The test failure seems intermittent but I have not seen the original failures so I can't tell for 100%. @dLobatog can you confirm the tests issues got resolved. If yes, I can merge this one

ares:
@inecas this looks like standard failures, before that, nearly all tests combinations were red. This one failure says "Cannot allocate memory". I'd say merge at will.

dLobatog:
@iNecas yeah, originally some tests would fail after adding `compute_attributes`  to `host.class.attribute_names` unknowingly. This one looks like a one-off (http://ci.theforeman.org/job/test_develop_pr_core/14218/database=sqlite3,ruby=2.1,slave=fast/testReport/junit/(root)/PluginWebpackDirectoriesTest/test_bundler/), but feel free to run it again if you want

iNecas:
I'm ok with just your confirmation as I wasn't sure about that. Thanks @dLobatog

PR #4809, opened 2017-09-05
Fixes #20846 - Include babel-polyfill to support ES6 Map

bastilian:
Currently CI develop is failing[1], since it's browser does not support ES6 Maps.
Including 'babel-polyfill' will include Map support and test should pass again.

[1] http://ci.theforeman.org/view/Core/job/test_develop/database=postgresql,ruby=2.1,slave=fast/lastBuild/testReport/(root)/AboutIntegrationTest/test_0001_about_page/

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,7 @@
 {
   "presets": ["es2015", "react"],
   "plugins": [
-    "transform-object-rest-spread"
+    "transform-object-rest-spread",
+    "transform-runtime"
   ]
 }
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "babel-loader": "^7.1.1",
     "babel-plugin-transform-object-assign": "^6.8.0",
     "babel-plugin-transform-object-rest-spread": "^6.8.0",
+    "babel-plugin-transform-runtime": "^6.23.0",
     "babel-polyfill": "^6.13.0",
     "babel-preset-es2015": "~6.6.0",
     "babel-preset-react": "^6.5.0",
@@ -46,6 +47,7 @@
     "phantomjs-prebuilt": "^2.1.0"
   },
   "dependencies": {
+    "babel-runtime": "^6.26.0",
     "brace": "^0.10.0",
     "c3": "^0.4.11",
     "datatables.net": "~1.10.12",


webpack/assets/javascripts/foreman_hosts.js
```diff
@@ -1,4 +1,5 @@
 import _ from 'lodash';
+import 'babel-polyfill';
```
ohadlevy: IMHO this is not the right way to handle this, as it will actually load babel code into the browser, instead I think you should be using a babel transform plugin, e.g. like https://babeljs.io/docs/plugins/transform-runtime. this is consistent with other language enhancements we've added to babel/webpack. also it could be reused across the code base and not limited to this file.

bastilian: I will look into using transform-runtime for this.

bastilian:
@ohadlevy it now is included via runtime-transform.

.babelrc
```diff
@@ -1,6 +1,12 @@
 {
   "presets": ["es2015", "react"],
   "plugins": [
-    "transform-object-rest-spread"
+    "transform-object-rest-spread",
+    ["transform-runtime", {
```
ohadlevy: don't we need it also in package.json ?

.babelrc
```diff
@@ -1,6 +1,12 @@
...
+      "helpers": false,
+      "polyfill": true,
```
ohadlevy: this is the default, probably not needed to add it.

.babelrc
```diff
@@ -1,6 +1,12 @@
...
+      "regenerator": false,
+      "moduleName": "babel-runtime"
```
ohadlevy: this is the default too, probably not needed to add it.

bastilian:
@ohadlevy i added the dependencies to package.json, they were already dependencies from another package. And it is now simply included with defaults, that include polyfill.

ohadlevy:
thanks, pending mr. jenkins.

bastilian:
@ohadlevy: @dLobatog mentioned to not introduce any dependencies, therefore it might be better to rewrite out the need for Map instead.

bastilian:
@ohadlevy There is now #4811, should we close this now?

PR #4810, opened 2017-09-05
refs #19554 - fixes linting and ES6 changes

ohadlevy:
this is a follow up to 805420d


ohadlevy:
replaced in favor of #4811

PR #4811, opened 2017-09-05
Refs #19554 - Rewrite to use an Object and not Map

bastilian:
Map is not yet fully supported in all browsers and
would require additional dependencies to be added.

Using a Object with lodash functions achieves the
same and does not require any additional
dependencies.

This also includes fixes from #4810. An unsquashed history can be found here: https://github.com/bastilian/foreman/commits/fix/ref-19554


ohadlevy:
thanks @bastilian !

PR #4812, opened 2017-09-05
Fixes #20858 - pxe_loader in GET api/v2/hostgroups/:id

kgaikwad:


dLobatog (approved):
Thanks @kgaikwad , #4814 should take care of the failing tests.

PR #4813, opened 2017-09-06
Fixes #20840 - show error on invalid name for any parameter

kgaikwad:
Updating any type of Parameter such as subnet_parameter/host_parameter with invalid name
doesn't show any error. 
With this commit, it is now showing error on parameters with invalid name.

### DIFF
--- a/app/helpers/common_parameters_helper.rb
+++ b/app/helpers/common_parameters_helper.rb
@@ -69,6 +69,10 @@ module CommonParametersHelper
   end
 
   def authorized_resource_parameters(resource, type)
-    resource.send(type).authorized(:view_params) + resource.send(type).select(&:new_record?)
+    parameters_by_type = resource.send(type)
+    parameter_ids_to_view = parameters_by_type.authorized(:view_params).map(&:id)
+    resource_parameters = parameters_by_type.select { |p| parameter_ids_to_view.include?(p.id) }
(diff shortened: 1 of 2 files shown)


dLobatog (changes requested):
Thanks @kgaikwad! I was able to see the error on Subnet parameters clearly. On the develop ranch, no error is shown even in the UI.

Just 2 comments:

1 - Please try to include screenshots for helper/view changes, it makes easier to review. 
2 - Can you add an integration test? This would ensure we always see the error even if this code changes. You can see an example of how to check for errors in integration tests in https://github.com/theforeman/foreman/blob/develop/test/integration/hostgroup_test.rb#L44. Feel free to just test it on Subnet parameters or Host parameters, or both, but at least one 😄

kgaikwad:
@dLobatog, 

Thank you for your review and for sharing reference test file to add integration test.
To avoid code duplication, I have added integration test only for subnet_parameters as in case of parameters, both having same functionality .

For reference, I am adding screen-shots for before-after changes:

1. Behaviour before adding changes:
no_error_on_invalid_name_with_spaces_on_edit_subnet_before_change

2. Behaviour after adding changes:
error_on_invalid_name_with_space_on_edit_subnet_after_change

dLobatog (approved):
Great job, thanks @kgaikwad !

PR #4817, opened 2017-09-07
Fixes #19730 - Add vendor to provisioning templates

dLobatog:
Before unlocking a provisioning template, a warning is shown on the
screen. This warning should contain the vendor that provided the
template (normally Foreman or Katello).

This works for new installations. For existing installations, I'm not quite sure how can we add this flag. Matching by name with templates in app/views/ unattended would certainly NOT work - maybe matching by content and name would. Is it worth the effort?

screenshot from 2017-09-07 18-39-11

### DIFF
--- a/db/seeds.d/07-provisioning_templates.rb
+++ b/db/seeds.d/07-provisioning_templates.rb
@@ -6,9 +6,10 @@ ProvisioningTemplate.without_auditing do
     contents = File.read(File.join("#{Rails.root}/app/views/unattended/provisioning_templates", input[:source]))
 
     if (t = ProvisioningTemplate.unscoped.find_by_name(input[:name])) && !SeedHelper.audit_modified?(ProvisioningTemplate, input[:name])
-      if t.template != contents
+      if t.template != contents || t.vendor != 'Foreman'
         t.template = contents
         t.locked = true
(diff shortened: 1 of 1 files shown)


db/seeds.d/07-provisioning_templates.rb
```diff
@@ -9,6 +9,7 @@
       if t.template != contents
```
ares: If you change it to `if t.template != contents || t.vendor != 'Foreman'` it updates existing templates too even if their content is the same. We're updating their content anyway so it's safe to consider these are shipped by Foreman.

ares (changes requested):
one small suggestion, it seems to work fine with it for my DB

dLobatog (commented):
@ares Updated to modify the template `if vendor != 'Foreman'`, that will update existing templates belonging to the list of seeded templates. 

If you want I can also add a check "if locked" too, so that if anyone decided to unlock a 'vendored' template, we don't overwrite it

ares (approved):

ares:
Thanks @dLobatog, merging.

ares:
Merged as f646ff568c4a050baf50d726584447da28986669

PR #4818, opened 2017-09-08
Fixes #19742 - cache hosts for fact values

ares:
When we generate fact values hash we always loaded a host for a given
fact. This in combination with taxonomies queryies that are
automatically generated caused roughly 6N+1 issue. We can preload all
hosts that are required and use this in-memory cache for generating the
hash.

### DIFF
--- a/app/models/fact_value.rb
+++ b/app/models/fact_value.rb
@@ -79,10 +79,12 @@ class FactValue < ApplicationRecord
   end
 
   def self.build_facts_hash(facts)
+    hosts = Host.where(:id => facts.group_by(&:host_id).keys).all
+
     hash = {}
     facts.each do |fact|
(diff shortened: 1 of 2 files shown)


dLobatog (approved):
@ares 👍 if tests pass, preloading the hosts helps a lot.

dLobatog:
Thanks @ares !

dLobatog:
Cherry-picked as 440726488 and 82ab8c8469b5370bac9870db985c678f0cbcdef6
 to 1.15 and 1.16

lzap:
Do you think this has some positive impact of fact import performance?

ares:
@lzap absolutely, see http://projects.theforeman.org/issues/19742#note-9 for more info

PR #4819, opened 2017-09-08
fixes #20882 - show template kind names from plugins

timogoebel:
Before:
image

After:
image


iNecas:
test failures related

timogoebel:
@iNecas:
There is a German saying: "If it takes long enough, it will be all right in the end.". Finally fixed the tests.

iNecas (approved):

PR #4820, opened 2017-09-08
Fixes #19047 - unify date formats

ares:
This introduces two new helpers for printing date in absolute and
relative format. Helper for absolute accepts also a parameter to choose
between long and short variant. The long one includes the year. The date
also has a title so you see the other format after hovering mouse
cursor.

---
replaces #4419, updated all pages, see all affected on screenshots, I'll update foreman plugin writing instructions and open issues or PRs to plugins after this gets in

pinging people who participated in previous discussions: @iNecas @ekohl @timogoebel @dLobatog @ohadlevy @lzap and of course @Rohoover, thumbs up/down appreciated, full review from one of you is enough :-)

reports
dates1
dashboard
dates2
proxy logs
dates3
proxy log detal
dates4
puppet ca expiration
dates5
audits
dates6
facts
dates7
hosts
dates8
report detail
dates9
trends index
dates10

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -46,6 +46,35 @@ module ApplicationHelper
     params.key?(:search)
   end
 
+  # this helper should be used to print date time in absolute form
+  # it will also define a title with relative time information
+  # it supports two formats :short and :long
+  # example of long is February 12, 2021 17:13
+  # example of short is Aug 31, 12:52
+  def date_time_absolute(time, format = :short)
+    raise ArgumentError, "unsupported format '#{format}', use :long or :short" unless %w(long short).include?(format.to_s)
+
+    content_tag :span, :title => date_time_relative_value(time) do
+      date_time_absolute_value(time, format)
+    end
+  end
+
+  # this helper should be used to print date time in relative form, e.g. "10 days ago",
+  # it will also define a title with absolute time information
+  def date_time_relative(time)
+    content_tag :span, :title => date_time_absolute_value(time, :long) do
+      date_time_relative_value(time)
+    end
+  end
+
+  def date_time_absolute_value(time, format = :short)
+    l(time, :format => format)
+  end
+
+  def date_time_relative_value(time)
+    (time > Time.now.utc ? _('in %s') : _('%s ago')) % time_ago_in_words(time)
+  end
+
   protected
 
   def contract(model)
--- a/app/helpers/fact_values_helper.rb
+++ b/app/helpers/fact_values_helper.rb
@@ -1,6 +1,6 @@
 module FactValuesHelper
   def fact_from(record)
-    _("%s ago") % time_ago_in_words(record.updated_at)
+    date_time_relative(record.updated_at)
   rescue
     _("N/A")
   end
--- a/app/helpers/hosts_helper.rb
+++ b/app/helpers/hosts_helper.rb
@@ -78,7 +78,7 @@ module HostsHelper
   end
 
   def last_report_column(record)
-    time = record.last_report? ? _("%s ago") % time_ago_in_words(record.last_report): ""
+    time = record.last_report? ? date_time_relative_value(record.last_report) : ""
     link_to_if_authorized(time,
                           hash_for_host_config_report_path(:host_id => record.to_param, :id => "last"),
                           last_report_tooltip(record))
@@ -86,11 +86,12 @@ module HostsHelper
 
   def last_report_tooltip(record)
     opts = { :rel => "twipsy" }
+    date = record.last_report.nil? ? '' : date_time_absolute_value(record.last_report) + ", "
     if @last_report_ids[record.id]
-      opts["data-original-title"] = _("View last report details")
+      opts["data-original-title"] = date + _("view last report details")
     else
       opts.merge!(:disabled => true, :class => "disabled", :onclick => 'return false')
-      opts["data-original-title"] = _("Report Already Deleted") unless record.last_report.nil?
+      opts["data-original-title"] = date + _("report already deleted") unless record.last_report.nil?
     end
     opts
   end
--- a/app/views/trends/index.html.erb
+++ b/app/views/trends/index.html.erb
@@ -35,5 +35,5 @@
 </table>
 <%= will_paginate_with_info @trends %>
 <% unless TrendCounter.unconfigured? %>
-  <%= _("Last updated %s ago") % (time_ago_in_words TrendCounter.order(:created_at).last.created_at) %>
+  <%= (_("Last update: %s") % (date_time_relative(TrendCounter.order(:created_at).last.created_at)) ).html_safe %>
 <% end %>
--- a/app/assets/javascripts/proxy_status/logs.js
+++ b/app/assets/javascripts/proxy_status/logs.js
@@ -30,7 +30,7 @@ function activateLogsDataTable() {
     autoWidth: false,
     columnDefs: [{
       render: function ( data, type, row ) {
-        return new Date(data * 1000).toLocaleString();
+        return data;
       },
       width: "15%",
       targets: 0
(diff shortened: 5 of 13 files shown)


app/helpers/application_helper.rb
```diff
@@ -46,6 +46,37 @@ def multiple_with_filter?
...
+  # it will also define a title with absolute time information
+  def foreman_relative_date_time(time)
+    content_tag :span, :title => foreman_absolute_date_time_value(time, :long) do
+      foreman_relative_date_time_value(time)
+    end
+  end
+
+  def foreman_absolute_date_time_value(time, format = :short)
+    if format == :short
+      l(time, :format => "%b %d, %H:%M")
```
ekohl: Is there no localized short format?

ares: not the one we wanted, that would result in "08 Sep 14:42", the difference is in comma and order

```
Sep 08, 14:42                      # this format
08 Sep 14:42                       # short format
```

ekohl: ```08 Sep 14:42``` looks like a EU localized format and for a NL format I'd expect that, not ```Sep 08, 14:42```.

iNecas: I have not seen using `foreman_` as a prefix for methods. Do we want to introduce this here? What about just `date_time_absolute` and `date_time_relative`?

iNecas: Sry for hi-jacking the thread, will avoid that in future

ares: @ekohl ok makes sense, I'll change it to short which should be localized
@iNecas the idea was to indicate this is the foreman recommended way which should be used in plugins but if you or anyone else prefers, I can remove the prefix

ares:
@inecas @ekohl updated per your suggestions, could you please take another look?

app/helpers/fact_values_helper.rb
```diff
@@ -1,6 +1,6 @@
 module FactValuesHelper
   def fact_from(record)
-    _("%s ago") % time_ago_in_words(record.updated_at)
+    date_time_absolute(record.updated_at)
```
ekohl: Is this an explicit choice to go from relative to absolute?

app/views/trends/index.html.erb
```diff
@@ -35,5 +35,5 @@
 </table>
 <%= will_paginate_with_info @trends %>
 <% unless TrendCounter.unconfigured? %>
-  <%= _("Last updated %s ago") % (time_ago_in_words TrendCounter.order(:created_at).last.created_at) %>
+  <%= (_("Last updated %s") % (date_time_relative(TrendCounter.order(:created_at).last.created_at)) ).html_safe %>
```
ekohl: Composing strings is generally a bad idea in translations because different languages might have different grammar.

app/helpers/application_helper.rb
```diff
@@ -46,6 +46,35 @@ def multiple_with_filter?
...
+
+  # this helper should be used to print date time in relative form, e.g. "10 days ago",
+  # it will also define a title with absolute time information
+  def date_time_relative(time)
+    content_tag :span, :title => foreman_absolute_date_time_value(time, :long) do
+      foreman_relative_date_time_value(time)
+    end
+  end
+
+  def foreman_absolute_date_time_value(time, format = :short)
```
iNecas: Still some methods that use the Foreman prefix, `data_time_absolute_value` would work for me.

app/helpers/fact_values_helper.rb
ares: yes, the reasoning is the same as for reports/hosts page, the recommendation for tables is use the absolute date since the relative information for 20 lines usually ends up as "x days ago" which does not provide too much info

with facts it might vary a bit more but I think absolute date here makes more sense

ares:
@iNecas @ekohl I pushed new version that should address your comments, please take another look

app/helpers/hosts_helper.rb
```diff
@@ -78,19 +78,20 @@ def flags_for_nic(nic)
   end
 
   def last_report_column(record)
-    time = record.last_report? ? _("%s ago") % time_ago_in_words(record.last_report): ""
+    time = record.last_report? ? foreman_absolute_date_time_value(record.last_report): ""
```
iNecas: Missed rename: this causes the tests to fail

ares:
Thanks @iNecas, sorry I was in hurry. Pushed with new names (also on other places)

ohadlevy:
one general question is if we should use js based time displaying for the
relative strings, e.g. 1 minute ago could easily be updated to 2 minutes
ago if a user stays on that screen for multiple minutes.

we have been looking into using moment.js in multiple places (e.g.
https://github.com/theforeman/foreman/pull/4597) and I think going forward
it would make sense to consolidate on one implementation that also fits
react generated date time strings.

ekohl (approved):
@ares consider my approval a more general approval of the diff I see. Core contributors should still do a proper code in depth review and see if you haven't missed anything.

@ohadlevy I think it does make sense to move presentation to javascript and relative strings are an excellent example of that. Another example could be of changing long to short on smaller screens. That said, I'm not sure it should be part of this PR. This unification is good in itself and is easily refactored later.

ares:
foreman tests unrelated 
* HostJSTest::NIC modal window::editing interfaces.test_0001_click on edit opens modal
* HostJSTest::NIC modal window::editing interfaces.test_0002_ok button updates overview table

in single combination

@ohadlevy for the js js based time displaying for the relative strings, I like the idea but would prefer to keep it as a seperate PR, could you please open RFE? Thanks to this PR, it will be a change in single method but we should first make sure these new helpers are used in all plugins. Then we can easily iterate on further improvements.

iNecas (approved):
Tested and works well, Also works for me. We are still pending @timogoebel (even though he made the 👍 on one comment) @dLobatog @lzap and @Rohoover that were mentioned in the original comment. What is the threshold for merging this?

iNecas:
Btw. if we can agree on this quickly, it would be great to have this in 1.16, so that the plugins can start using it while keeping the compatibility with the last stable release. Otherwise, I, as a plugin maintainer, would not want to start relying on it until the branching time of 1.17

timogoebel:
@iNecas: Well, I'm not too fond of this change, to be honest. Especially for reports, I think, the relative time is more important than the absolute time. I did a quick poll with my team and all agreed that the relative time is what they want to know. If the report came in in the last hour, you immediately know all is well. Only if you see something like "a day ago" you know, something is wrong and you need to investigate.

I don't care enough to block this, but I still don't think this is a change for the better in all places.

What I do like are the helpers, that's a great addition for plugins.

ekohl:
Big 👍 for getting the helpers in for 1.16 so the plugins can start using them. I also see the value of relative times because like @timogoebel said, you usually care about the last day and then the relative values are easy to read.

timogoebel:
> you usually care about the last day and then the relative values are easy to read.

I do agree, that a tooltip with the absolute time is missing. But the helpers would fix that.

ekohl:
Would it be an idea to merge it now with 1-to-1 conversion to helpers and discuss the relative/absolute change in a separate PR?

ares:
Ok, I thought we had this discussion already but rereading foreman-dev and previous PR it's clear everyone has a different opinion :-) I reverted relative date format in hosts, reports and facts table so this can get in ASAP. I can bring that up on next ux demo or reopen foreman-dev thread after this is in.

iNecas:
Sounds like a plan

iNecas:
Test failures unrelated. Merging this now so that we have the the helpers available in 1.16

iNecas:
@ares please update the description of the related issue to better match what was actually done (+ it will probably need to be unlinked from the BZ, as this change is not doing what it claims to fix in the BZ)

iNecas:
Also sending an email to foreman-dev about availability of the new helpers might be a good idea

iNecas:
Cherry-picked to 1.16

Rohoover:
Perhaps a future state could include a user preference for display format?

I talked to a customer last week who wanted the absolute time, not relative.

PR #4822, opened 2017-09-11
Refs #20708 - reset default templates type

ares:


ares:
The linked issue caused issues on Rails 5. It turns out that we configure default `type` value to `Template` which causes problems. After removing it seems to work fine.

mmoll:
fixes the problem for me 👍

mmoll (approved):

iNecas:
[test foreman ] [test katello]

iNecas:
[test foreman]

dLobatog (approved):
👍 if tests pass, what problems did it cause on Rails 5?

mmoll:
@dLobatog see https://github.com/theforeman/foreman/pull/4778#issuecomment-328283576

tbrisker:
If there are any with type `Template` shouldn't we change them as well? I'm not sure if there are, but since it was the default and now is invalid any existing ones should be updated.

ares:
there should not be such templates, STI in rails 4 changes the type before save (or earlier) but rails 5 seems to have some additional check just during initialization

iNecas:
I consider all questions as resolved now. Thanks @ares

PR #4824, opened 2017-09-12
Fixes #20902: ensure 'katello' is webpack bundle name.

waldenraines:
We were removing the bundle name for katello and thus making it
difficult to reference the katello bundle.  This commit changes the
script to ensure that the katello bundle name is retained.

http://projects.theforeman.org/issues/20902

### DIFF
--- a/script/plugin_webpack_directories.rb
+++ b/script/plugin_webpack_directories.rb
@@ -12,11 +12,12 @@ else
 end
 
 PLUGIN_NAME_REGEXP = /foreman*|katello*/
+REMOVE_BUNDLE_NAME_PLUGINS = /foreman*/
 
 config = { entries: {}, paths: [] }
 specs.each do |dep|
   # skip other rails engines that are not plugins
-  # TOOD: Consider using the plugin registeration api?
+  # TODO: Consider using the plugin registration api?
   if gemfile_in
     next unless dep =~ PLUGIN_NAME_REGEXP
     dep = dep.to_spec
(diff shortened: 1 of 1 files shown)


script/plugin_webpack_directories.rb
```diff
@@ -25,11 +26,11 @@
   end
   path = "#{dep.to_spec.full_gem_path}/webpack"
   entry = "#{path}/index.js"
-  # some plugings share the same base directory (tasks-core and tasks, REX etc)
+  # some plugins share the same base directory (tasks-core and tasks, REX etc)
   # skip the plugin if its path is already included
   next if config[:paths].include?(path)
   if File.exist?(entry)
-    bundle_name = dep.name.gsub(/-|_|#{PLUGIN_NAME_REGEXP}/,'')
+    bundle_name = dep.name.gsub(/-|_|#{REMOVE_BUNDLE_NAME_PLUGINS}/,'')
```
waldenraines, line 33: The real question here is should we be doing this even for foreman?  I would argue that the bundle name of foreman should probably be 'foreman' and not ''.

ohadlevy:
thanks @waldenraines overall :+1: can you please add a test at https://github.com/theforeman/foreman/blob/develop/test/unit/script/plugin_webpack_directories_test.rb ?

script/plugin_webpack_directories.rb
waldenraines: @ohadlevy thoughts on this?

dLobatog (changes requested):
@waldenraines Mind to rebase with latest develop? It would make the katello tests pass - as they need a change recently merged

script/plugin_webpack_directories.rb
waldenraines: In other words I would prefer to just remove this line.  I think it makes sense for foreman to reference its own bundle as 'foreman' just like everything else.

waldenraines:
> @waldenraines Mind to rebase with latest develop? It would make the katello tests pass - as they need a change recently merged

Done.

dLobatog (approved):

waldenraines:
This is going to conflict with https://github.com/theforeman/foreman/pull/4733 so we should decide which is going first.

script/plugin_webpack_directories.rb
ohadlevy: @waldenraines do you mind if we just merge as is to unblock katello? changing the bundle name might have other impact on webpack configuration etc which i think we can look into in another pr?

dLobatog (approved):
I'll be merging this to unblock Katello, the `bundle_name` thing can be resolved in other PR if needed.

PR #4475, opened 2017-04-20
Fixes #19332 - Add welcome page when missing

amirfefer:
Some pages are missing a welcome page.
I don't quite sure about the chosen icons.

PR #4825, opened 2017-09-12
Fixes #20896 - Remove help button from ptable

shiramax:
I moved everything in the partition table help page to here:
https://github.com/theforeman/theforeman.org/pull/944

### DIFF
--- a/app/views/ptables/index.html.erb
+++ b/app/views/ptables/index.html.erb
@@ -1,7 +1,6 @@
 <% title _("Partition Tables") %>
 <% title_actions new_link(_("Create Partition Table")),
-                 documentation_button('4.4.4PartitionTables'),
-                 help_button %>
+                 documentation_button('4.4.4PartitionTables') %>
 
 <table class="<%= table_css_classes 'table-two-pane table-fixed' %>">
   <thead>


timogoebel:
@shiramax: Thanks, but why are you removing the welcome page?

shiramax:
@timogoebel , According to Bug #20896, the button 'help' should be removed and the content of the welcome page should be moved to the documentation.

timogoebel:
Actually, we provide the welcome page for every Model in Foreman and I'd like to keep it that way. I do agree, that the page has too much detailed information. But I'd still like to keep the page. Can we keep the page but shorten the text?
@bastilian: As you opend the ticket: Do you concur?

bastilian:
@shiramax & @timogoebel Keeping the page and shortening would be better, yes. When opening the issue I had not realized that it was a welcome page. Since this only appears when there are no saved partition tables it might be enough for the page to follow the examples of other welcome pages and have a short intro a link to the documentation and the create button, like the hosts one:

gnome-shell-screenshot-15cc6y

shiramax:
@bastilian  @timogoebel  
The welcome page of the partition table will not appear because we seed the partition tables before.

ares:
It can appear easily if users change organization or for non admin if they are not assigned to any,

shiramax:
@ares @bastilian @timogoebel  , what do you think ?
image

i'm not sure this icon should in in the partition table as well.

amirfefer:
I already worked on this [here](https://github.com/theforeman/foreman/pull/4475)
If you wish to split this up be my guest :)

shiramax:
@amirfefer , Whatever you decide is fine by me :)

amirfefer:
@shiramax 
You can just remove help button from index page in this PR.
I'm taking care of welcome page.

shiramax:
@amirfefer ,
 Done.

app/views/ptables/index.html.erb
```diff
@@ -1,7 +1,6 @@
 <% title _("Partition Tables") %>
 <% title_actions new_link(_("Create Partition Table")),
-                 documentation_button('4.4.4PartitionTables'),
-                 help_button %>
+                 documentation_button('4.4.4PartitionTables')%>
```
adamruzicka: There should probably be a space before the closing `%>`

adamruzicka (approved):
Inlined one small nitpick, otherwise looks goods to me

shiramax:
@adamruzicka  Done, thanks.

timogoebel (approved):

timogoebel:
Merged, thanks @shiramax.

PR #4826, opened 2017-09-12
Fixes #20908 - postpone initialization of Dynflow runtime

iNecas:
In #18618, the initialization code was moved from foreman-tasks to
foreman. There was however one thing done differently, where the
initialization of dynflow world was moved directly to the time
`Rails.application.dynflow` was called. This is not right, as there are
situations where plugins go and call `Rails.application.dynflow.config`
before the whole stack is prepared, which leads to database connection
being opened too early, which can cause the whole lot of issues (as
we've hit currently with Katello).

There was also change in the signature of `Dynflow::Rails#initialize`
method, that needs to be updated.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -246,11 +246,26 @@ module Foreman
         if defined?(ForemanTasks)
           ForemanTasks.dynflow
         else
-          Dynflow::Rails.new(Foreman::Dynflow::Configuration.new)
+          ::Dynflow::Rails.new(nil, Foreman::Dynflow::Configuration.new)
         end
       @dynflow.require!
-      @dynflow.initialize!
       @dynflow
+    end
+
+    config.after_initialize do
+      dynflow = Rails.application.dynflow
+      dynflow.eager_load_actions!
+      dynflow.config.increase_db_pool_size
+
+      unless dynflow.config.lazy_initialization
+        if defined?(PhusionPassenger)
+          PhusionPassenger.on_event(:starting_worker_process) do |forked|
+            dynflow.initialize! if forked
+          end
+        else
+          dynflow.initialize!
+        end
+      end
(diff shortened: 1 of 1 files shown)


config/application.rb
```diff
@@ -246,12 +246,27 @@ def dynflow
         if defined?(ForemanTasks)
           ForemanTasks.dynflow
         else
-          Dynflow::Rails.new(Foreman::Dynflow::Configuration.new)
+          ::Dynflow::Rails.new(nil, Foreman::Dynflow::Configuration.new)
```
iNecas, line 249: `Dynflow::Rails` was resolving to `Foreman::Dynlfow::Rails` which was not existing: the after_initialize block actually discovered this issue as it seems this code-path was not run in tests before.

iNecas:
And all the tests are green again :)

config/application.rb
dLobatog: Thanks, I had to do something similar in https://github.com/theforeman/foreman/pull/4717/, can confirm it fixes the problem

dLobatog (approved):

dLobatog:
Thanks @iNecas - this should go into 1.16 I suppose?

iNecas:
yes please

dLobatog:
Cherry-picked as 7ad59f2a3048f8f81d6221f7ec87bd48d478110b

config/application.rb
iNecas: Yes, this should fix the issue you were facing there + making it work with later dynflow version as well that came with https://github.com/Dynflow/dynflow/commit/ff129a6b92eb209a82e51d177034b6f9187bff19#diff-56fbff8c6eb846b55588d78db914924aR10

PR #4827, opened 2017-09-13
Refs #20098 - Remove save expectation for templates_used

dLobatog:
In develop, the original #20098 change worked well because #20708 was
merged before. However in 1.15-stable, there are two tests that call
.save on an ActiveRecord model (Audited). This makes tests fail:

HostsControllerTest::#template_used.test_submit_multiple_rebuild_config_optimistic
HostsControllerTest::#template_used.test_submit_multiple_rebuild_config_pessimistic

http://ci.theforeman.org/view/Core/job/test_1_15_stable/database=sqlite3,ruby=2.2,slave=fast/34/

This PR goes straight to 1.15.4 - no need for this on develop.

### DIFF
--- a/test/controllers/hosts_controller_test.rb
+++ b/test/controllers/hosts_controller_test.rb
@@ -1252,7 +1252,6 @@ class HostsControllerTest < ActionController::TestCase
     setup do
       @host.setBuild
       ActiveRecord::Base.any_instance.expects(:destroy).never
-      ActiveRecord::Base.any_instance.expects(:save).never
       @attrs = host_attributes(@host)
     end
 


dLobatog (commented):
Tests fail because they're running this change against develop

tbrisker (approved):
Makes sense.

iNecas:
Tests are failing consistently here, seems like it's related to the change

dLobatog:
@iNecas Try to run tests on 1.15-stable - this change is running against develop, but it should only be merged against 1.15-stable

iNecas:
Are you sure it's running against develop, from what I see in the jenkins logs + https://github.com/theforeman/foreman-infra/blob/ff79e70c618caa8476ac017a4ab6d2d47fcd86bd/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/scm/foreman.yaml, it seems we just checkout the sha of the PR, so provided you rebased this against 1.15-stable, it seems that's what is the jenkins running it against as well. Can you point me to the place in our jenkins job, that would prove me wrong (such as place that we would to cherry-picking to the develop or doing other way of merging)

iNecas:
Since only postgres tests are failing, maybe there is additional fix needed for functional tests as well

dLobatog:
@iNecas Definitely, I suppose I didn't open the Postgres failures yesterday on 1_15_stable (http://ci.theforeman.org/job/test_1_15_stable/database=postgresql,ruby=2.1,slave=fast/35/)

dLobatog:
Seems that the failures could come from #4750

dLobatog:
Got it: #4750 was cherry-picked, but it depends on #4720 to work:

``` password_f f, :root_pass, :help_inline => _("Password must be 8 characters or more").html_safe,
                  :unset => action_name == "edit", :keep_value => action_name == "clone"
```

relies on `keep_value` being handled by the helper. That is introduced in here (https://github.com/theforeman/foreman/pull/4720/files#diff-a1d6431830287bc2d6a5c744edbf1eb7R20) , which I will cherry-pick to 1.15.

dLobatog (commented):
@iNecas 

I cherry-picked what caused the integration test failures (4aba258ef68b9dbefda50b62fa39da3ee3cb372e).
 
I'll rebase this one and run tests again, then it ought to be green.

dLobatog:
[test katello]

iNecas:
The katello tests are running katello master, so we will not get the tests green. Would be nice to run the job manually against 3.4 branch, but I don't see a way to do that in jenkins: just few params exposed for rerun

iNecas:
Thanks @dLobatog

PR #4828, opened 2017-09-13
Fixes #20900 - Combine Template editor and template field

bastilian:
To resolve confusion around the template field being required and the template editor, which actually fulfills that this moves the file import into the toolbar as a feature for the editor, that can be turned on and removes the "second" template field.

Before:
gnome-shell-screenshot-oz3n6y 2

After (in action):

template-import-demo

### DIFF
--- a/app/assets/stylesheets/editor.scss
+++ b/app/assets/stylesheets/editor.scss
@@ -8,7 +8,7 @@
   width: 100%;
 }
 
-.navbar-nav.navbar-right.navbar-editor:last-child {
+.navbar-nav.navbar-right.navbar-editor {
   margin-right: 0;
 
   li {
--- a/app/views/common_parameters/_form.html.erb
+++ b/app/views/common_parameters/_form.html.erb
@@ -6,7 +6,7 @@
 
       <div class="col-md-9">
         <div class="editor-container">
-          <%= render :partial => 'editor/toolbar', :locals => {:show_preview => false} %>
+          <%= render :partial => 'editor/toolbar', :locals => { :show_preview => false, :show_import => false} %>
           <% masked = @common_parameter.hidden_value? ? 'masked-input' : '' %>
           <%= f.text_area(:value,
                           :class => "form-control value editor_source #{masked}",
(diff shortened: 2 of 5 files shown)


ares:
I think it's better, though still feels a bit unintuitive. @Rohoover any thoughts? The feature is that user can select a file on his hard drive and it's content is placed into the editor. We used to have "file upload" element. 

Maybe add a space between this button and the rest, like the fullscreen button has? The other three buttons are related to editing the content. The new button replaces the whole content.

ares:
Foreman tests failures seems to be related

bastilian:
@ares fixed what made CI fail. 

I'm not 100% happy with how it looks, but it's a start that doesn't go too overboard. Separating the button from the rest and maybe even as the first button?

Rohoover:
Yes, I agree separating the button would be good. I also agree that overall this is an improvement.

ares:
@bastilian I don't see any changes in the latest commit, should there be some? After more thinking, I'd additionally suggest renaming "Import from File" to just "Import" or keeping it long but also changing the fullscreen button icon to text "Fullscreen".

bastilian:
@ares the change was only minor, to fix the parameters use of the toolbar[1]. The latest push does contain a change that is visible:

The button to import is now first and the fullscreen button is now text:

gnome-shell-screenshot-9kz66y


Ps.: @ares To my PRs there is always a source branch, if the branch has a '-PR' suffix. In this case the 'fix/20900' [2]

[1] https://github.com/bastilian/foreman/commit/95b8569c4e65f56747a183b90ae8611944d2d1c5
[2] https://github.com/theforeman/foreman/compare/develop...bastilian:fix/20900

ares:
I think it looks good. I think users edit existing (default) templates more often than import new ones so maybe the Import button should be after preview. This order makes sense if you want to create a template from scratch but as said, it's probably less frequent action. Also I can imagine people switch between input and preview often so having these buttons in center requires user to focus on where they click. If input would be the left most button, it's easier to click it (I think).

@Rohoover thoughts about the order or some more ideas? If not, after the order change, I'd merge.

Rohoover:
@ares 
I don't have any more feedback unless we had some data on use to determine button order.

ares:
@bastilian if you could please move import between preview and full screen (keeping the spacing) and then it's good to merge.

bastilian:
@ares the button is now after the preview button:
gnome-shell-screenshot-me0a7y

ares:
Great job @bastilian, merged as 7b34bf7dc727948b8811127343dd7a9463f1d47b!

PR #4830, opened 2017-09-14
Fixes #20920 - Prevent boot issue with Rake.application

tbrisker:


tbrisker:
Looks like in Rails 5, `Rake` is defined always but `Rake.application` isn't, so booting fails when checking if we're in rake or not.

iNecas:
Tested and works well, wiaint for the tests to finish

iNecas:
Test failure intermittent. Thanks @tbrisker

PR #4833, opened 2017-09-14
Fixes #10726 - NIC types are immutable

tstrachota:
- Adds validation of changing NIC types
- Fences out using default type in update actions of Hosts and Interfaces API controller (UI handles this situation with disabling the type select box)

### DIFF
--- a/app/controllers/api/v2/hosts_controller.rb
+++ b/app/controllers/api/v2/hosts_controller.rb
@@ -331,15 +331,15 @@ Return the host's compute attributes that can be used to create a clone of this 
           end
           # map interface types
           params[:interfaces_attributes] = params[:interfaces_attributes].map do |nic_attr|
-            interface_attributes(nic_attr)
+            interface_attributes(nic_attr, allow_nil_type: host.nil?)
           end
         end
         params = host.apply_inherited_attributes(params) if host
         params
       end
 
-      def interface_attributes(params)
-        params[:type] = InterfaceTypeMapper.map(params[:type])
+      def interface_attributes(params, allow_nil_type: false)
+        params[:type] = InterfaceTypeMapper.map(params[:type]) if params.has_key?(:type) || allow_nil_type
         params
       end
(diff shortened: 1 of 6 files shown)


iNecas:
please rebase for katello tests to run

app/controllers/api/v2/hosts_controller.rb
```diff
@@ -331,15 +331,15 @@ def host_attributes(params, host = nil)
...
           end
         end
         params = host.apply_inherited_attributes(params) if host
         params
       end
 
-      def interface_attributes(params)
-        params[:type] = InterfaceTypeMapper.map(params[:type])
+      def interface_attributes(params, skip_nil_type: false)
+        params[:type] = InterfaceTypeMapper.map(params[:type]) unless skip_nil_type && !params.has_key?(:type)
```
iNecas: The condition translates to `if !skip_nil_type || params.has_key?(:type)` which means, that the the line would be skipped, if the `skip_nil_type` is set, which is not the same as if the `param[:type]` was nil. It doesn't break the behavior, but might cause issues later. Changing the order to `if params.has_key?(:type) || !skip_nil_type` (+ using if instead of unless here, as it's hard to reason about double negation in complex conditions - at least for me). Or even better would be having `allow_nil_type` with condition  `if params.has_key?(:type) || allow_nil_type`

iNecas (changes requested):
Tested and works well: one comment about `skip_nil_type` conditions

tstrachota:
What you wrote makes sense @iNecas. Updated.

iNecas (approved):
Works for me, please rebase so that the katello tests get green hopefully as well

tstrachota:
I thought I have rebased, I'm sorry @iNecas, rebased now.

tstrachota:
@iNecas the katello tests are green now.

iNecas:
Thanks @tstrachota

PR #4834, opened 2017-09-14
Fixes #20890 - changes SSH key validation messages

shiramax:


tstrachota:
:+1: Tested on hammer and it works well:
```
> hammer user ssh-keys add --user john --name Key --key "wrong format" 
Could not add SSH Key:
  Key is not a valid public ssh key
  Key must be in OpenSSH public key format.
  Fingerprint could not be generated
  Length could not be calculated
```

boaz0 (approved):
:+1 SGTM

app/models/ssh_key.rb
```diff
@@ -21,7 +21,7 @@ class SshKey < ApplicationRecord
   validates :key,
     :presence => true,
     :ssh_key => true,
-    :format => { with: /\A(ssh|ecdsa)-.*\Z/ }
+    :format => { with: /\A(ssh|ecdsa)-.*\Z/, message: N_('must be in OpenSSH public key format.') }
```
ares: the dot should not be part of the message, see the screen below

ares (changes requested):
in UI, the message is composed with other validation messages, so it does not work well with sentences

key

shiramax:
@ares , Thanks , I removed the dot.

ares (approved):

ares:
Thanks @shiramax, merged as 4a96211d33c261655c887bbb6c787e38d883200a

PR #4835, opened 2017-09-14
Fixes #20931 - unify parameters and templates renderer

iNecas:
Before this patch, we used separate renderer for the parameters. This
caused issues when introducing more dsl methods for the templates that
we want to use in the parameters as well. Also, by the time of writing
the original SafeRenderer, there was no Renderer available.

After doing so, I was hitting issue with recursion, as
`Renderer.host_param` was using `@host.params` while the @host.params
was trying to render the params, that could include the `host_param`
call again, which lead to endless loop. After simplifying the
`HostParams` methods, it is possible to achieve this.

I also noticed that we were rendering only the inherited params and not
the non-inherited once. I've fixed this bug also as part of this patch.

Last but not least, due to the fact that we were using the
host_inherited_params for multiple purposes, we were rendering it also
when showing the host form, which was not even needed.


app/models/concerns/host_params.rb
```diff
@@ -6,39 +6,57 @@ module HostParams
...
                    :source => source,
                    :safe_value => param.safe_value }
         options[:source_name] = param.associated_label if source != 'global'
-        hp.update(Hash[param.name => include_source ? options : value])
+        hash[param.name] = options
+      end
+    end
+
+    def host_params_hash
+      return @cached_host_params_hash if @cached_host_params_hash
```
ares: I think this is equivalent
`@cached_host_params_hash ||= inherited_params_hash.merge(non_inherited_params_hash)`

app/services/parameter_safe_render.rb
```diff
@@ -0,0 +1,38 @@
+# Class to parse ERB with or without Safemode rendering. Needs a set
+# of variables, usually something like:
```
ares: how do I pass such set? it seems that only host is required in initializer and it's hardcoded in `render_string`

app/models/concerns/host_params.rb
```diff
@@ -6,39 +6,57 @@ module HostParams
...
     end
 
     def host_params
-      return cached_host_params unless cached_host_params.blank?
-      hp = host_inherited_params
-      # and now read host parameters, override if required
-      host_parameters.authorized(:view_params).each { |p| hp[p.name] = p.value }
-      @cached_host_params = hp
+      return @cached_host_params_rendered if @cached_host_params_rendered
+      renderer = ParameterSafeRender.new(self)
```
ares: if `renderer` was cached, could it reuse existing `host_param` method? it seems that it would be more dry

app/services/parameter_safe_render.rb
```diff
@@ -0,0 +1,38 @@
...
+    when Hash
+      object.merge(object) { |k, v| render_object v }
+    else
+      # Don't know how to parse this, send it back
+      object
+    end
+  end
+
+  def render_string(string)
+    render_safe(string, ::Foreman::Renderer::ALLOWED_HOST_HELPERS, :host => @host)
```
ares: This does not allow access to other, non-host related macros, is there a reason to that? I can imagine 
access to `snippet` macro, `foreman_server_url` macro and possibly others would be useful. So perhaps using ALLOWED_HELPERS would be better?

app/services/parameter_safe_render.rb
```diff
@@ -0,0 +1,38 @@
+# Class to parse ERB with or without Safemode rendering. Needs a set
+# of variables, usually something like:
+#   @allowed_vars = { :host => @host }
+# so that <%= @host.name %> has the right @host variable
+#
+class ParameterSafeRender
+  include Foreman::Renderer
```
ares: including `UnattendedHelper` from `app/helpers/unattended_helper.rb` instead would also make `root_pass`, `grub_pass` and `ks_console` work

ares (changes requested):
@iNecas it works great and fixes all you listed but there are still some potential improvements, please see inline comments

app/models/concerns/host_params.rb
iNecas: Yes, this is result after further refactoring, will update

app/models/concerns/host_params.rb
iNecas: Yes, the renderer is state-less. Will extract it to separate method and cached it

app/services/parameter_safe_render.rb
iNecas: Residual from the previous implementation. Will update

app/services/parameter_safe_render.rb
iNecas: Sounds good

app/services/parameter_safe_render.rb
iNecas: The main reason was https://github.com/theforeman/foreman/blob/398cb5b2d27e71ca14c4aebbbc5acbd16baa9a59/lib/foreman/renderer.rb#L218, where for non-host related methods, we use some tweaking based on the request params, and it would not work as expected. I can include it there though.

iNecas:
Comments should be addressed

ares:
I'm sorry it has slipped off, could you please rebase, when you have a chance?

iNecas:
Rebased

mmoll:
[test katello]

ares (approved):

ares:
Thanks @iNecas, works nicely! Merging.

PR #4837, opened 2017-09-16
Fixes #20949 - Replace deprecated AM::Errors usage

mmoll:


mmoll:

mmoll:
[test katello]

tbrisker:
looks like there are quite a few other cases where we use `errors[]=` (or `errors[]<<` which i am not sure is still valid either), could you please take a look?

mmoll:
At least everything that gets hit by any tests should be covered. Rails 5.1 does have these deprecated methods removed and I don't see anything still failing there pointing to anything missed here.

tbrisker:
Ok, I'll merge this for now and hopefully we won't see any failures further on in areas not covered by tests. Thanks @mmoll and @domcleal !

PR #4838, opened 2017-09-16
Fixes #20950 - Use 'distinct' in AR finders

mmoll:


mmoll:

iNecas (changes requested):
Based on grep, I've found one more https://github.com/theforeman/foreman/blob/develop/app/models/puppetclass.rb#L104

mmoll:
updated, test failure unrelated.

tbrisker:
Thanks @mmoll and @domcleal !

PR #4843, opened 2017-09-16
Fixes #20955 - Remove conditions to {destroy, delete}_all calls

mmoll:


mmoll:

mmoll:
test failures look unrelated

mmoll:
[test foreman]

iNecas (changes requested):
There is also https://github.com/theforeman/foreman/blob/develop/app/controllers/compute_resources_controller.rb#L12 that should be updated accordingly I think

iNecas:
There is also https://github.com/theforeman/foreman/blob/develop/app/controllers/settings_controller.rb#L10 and https://github.com/theforeman/foreman/blob/develop/app/models/host/managed.rb#L912-L913

mmoll:
updated, tests are green 💚

iNecas (approved):

iNecas:
Thanks @mmoll and @domcleal

PR #4844, opened 2017-09-16
Fixes #20956 - Don't pass class constants to AR conditions

mmoll:


mmoll:

app/models/host/base.rb
```diff
@@ -181,7 +181,7 @@ def set_interfaces(parser)
 
       ipmi = parser.ipmi_interface
       if ipmi.present?
-        existing = self.interfaces.where(:mac => ipmi[:macaddress], :type => Nic::BMC).first
+        existing = self.interfaces.where(:mac => ipmi[:macaddress], :type => Nic::BMC.name).first
```
tbrisker, line 184: not that it makes much difference, but why not just `"Nic::BMC"`?

tbrisker (commented):
Looks good, I just hope we aren't missing any other locations here.

app/models/host/base.rb
mmoll: I guess it was done for the sake of diff-minimizing, but IMHO using the Constant instead of a string is a tad nicer.

tbrisker: most other places currently pass the string, but WFM, not worth the effort..

tbrisker:
Thanks @mmoll and @domcleal !

PR #4845, opened 2017-09-16
Fixes #20957 - Replace alias_method_chain with Module prepend

mmoll:
Deprecated in Rails 5.0 and will be removed in 5.1. Some instances of
classes overwriting existing methods can be handled with `super`, other
concerns or modules are changed to use prepend instead of include.

Note: no ActiveSupport::Concern support for prepends, so load class
methods the pure Ruby way.

### DIFF
--- a/app/controllers/concerns/foreman/controller/smart_proxy_auth.rb
+++ b/app/controllers/concerns/foreman/controller/smart_proxy_auth.rb
@@ -14,14 +14,21 @@ module Foreman::Controller::SmartProxyAuth
       before_action(:only => actions) { require_smart_proxy_or_login(options[:features]) }
       attr_reader :detected_proxy
 
-      define_method(:require_ssl_with_smart_proxy_filters?) do
-        if [actions].flatten.map(&:to_s).include?(self.action_name)
-          false
-        else
-          require_ssl_without_smart_proxy_filters?
-        end
+      cattr_accessor :smart_proxy_filter_actions
+      self.smart_proxy_filter_actions ||= []
+      self.smart_proxy_filter_actions.push(*actions)
+
+      prepend SmartProxyRequireSsl
+    end
+  end
+
+  module SmartProxyRequireSsl
+    def require_ssl?
+      if [self.smart_proxy_filter_actions].flatten.map(&:to_s).include?(self.action_name)
+        false
+      else
+        super
       end
-      alias_method_chain :require_ssl?, :smart_proxy_filters
     end
   end
 
--- a/app/controllers/unattended_controller.rb
+++ b/app/controllers/unattended_controller.rb
@@ -4,13 +4,7 @@ class UnattendedController < ApplicationController
   layout false
 
   # We dont require any of these methods for provisioning
-  FILTERS = [:require_login, :session_expiry, :update_activity_time, :set_taxonomy, :authorize]
-  FILTERS.each do |f|
-    define_method("#{f}_with_unattended") do
-      send("#{f}_without_unattended") if preview?
-    end
-    alias_method_chain f, :unattended
-  end
+  skip_before_action :require_login, :session_expiry, :update_activity_time, :set_taxonomy, :authorize, :unless => Proc.new { preview? }
 
   before_action :set_admin_user, :unless => Proc.new { preview? }
   # We want to find out our requesting host
--- a/app/models/concerns/fog_extensions/openstack/server.rb
+++ b/app/models/concerns/fog_extensions/openstack/server.rb
@@ -1,13 +1,12 @@
 module FogExtensions
   module Openstack
     module Server
-      extend ActiveSupport::Concern
-
-      included do
-        alias_method_chain :security_groups, :no_id
-        attr_reader :nics
-        attr_accessor :boot_from_volume, :size_gb, :scheduler_hint_filter
-        attr_writer :security_group, :network # floating IP
+      def self.prepended(base)
+        class << base
+          attr_reader :nics
+          attr_accessor :boot_from_volume, :size_gb, :scheduler_hint_filter
+          attr_writer :security_group, :network # floating IP
+        end
       end
 
       def to_s
@@ -43,10 +42,9 @@ module FogExtensions
       end
 
       # the original method requires a server ID, however we want to be able to call this method on new instances too
-      def security_groups_with_no_id
+      def security_groups
         return [] if id.nil?
-
-        security_groups_without_no_id
+        super
       end
 
       def boot_from_volume
--- a/app/models/subnet.rb
+++ b/app/models/subnet.rb
@@ -7,7 +7,7 @@ class Subnet < ApplicationRecord
   BOOT_MODES = {:static => N_('Static'), :dhcp => N_('DHCP')}
 
   include Authorizable
-  include Foreman::STI
+  prepend Foreman::STI
   extend FriendlyId
   friendly_id :name
   include Taxonomix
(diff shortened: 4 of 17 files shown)


mmoll:

test/controllers/hosts_controller_test.rb
```diff
@@ -1267,7 +1267,6 @@ class Host::Valid < Host::Managed; end
     setup do
       @host.setBuild
       ActiveRecord::Base.any_instance.expects(:destroy).never
-      ActiveRecord::Base.any_instance.expects(:save).never
```
mmoll: bit unsure how to handle this.

app/models/concerns/fog_extensions/openstack/server.rb
```diff
@@ -1,13 +1,12 @@
...
   module Openstack
     module Server
-      extend ActiveSupport::Concern
-
-      included do
-        alias_method_chain :security_groups, :no_id
-        attr_reader :nics
-        attr_accessor :boot_from_volume, :size_gb, :scheduler_hint_filter
-        attr_writer :security_group, :network # floating IP
+      def self.prepended(base)
```
timogoebel, line 4: How about creating a custom `ActiveSupport::Concern` implementation so you can do

```ruby
prepended do

end
```

?

mmoll: you mean something similar to https://gitlab.com/gitlab-org/gitlab-ee/blob/master/config/initializers/0_as_concern.rb ?

timogoebel: Yeah. Exactly. But should be a separate PR I think.

mmoll: What's the opinion of other core devs here? Personally I'd rather like to see this PR get in now to speed things up, but no hard feelings either way.

timogoebel: I agree. My suggestion should not block this PR at all.

app/controllers/concerns/foreman/controller/smart_proxy_auth.rb
```diff
@@ -14,14 +14,20 @@ def add_smart_proxy_filters(actions, options = {})
...
-      define_method(:require_ssl_with_smart_proxy_filters?) do
-        if [actions].flatten.map(&:to_s).include?(self.action_name)
-          false
-        else
-          require_ssl_without_smart_proxy_filters?
+      cattr_accessor :smart_proxy_filter_actions
+      self.smart_proxy_filter_actions ||= []
+      self.smart_proxy_filter_actions.push(*actions)
+
+      extension = Module.new do
```
iNecas: It seems we don't need a dynamic anonymous module now, as there is nothing in the module block that would require a closure. I would prefer having standard module (something like `SmartProxyRequireSsl`) that would define this method and we would use that prepending. If nothing else, it's easier to see what modules are extending the classes (compating to anonymous modules)

app/controllers/unattended_controller.rb
```diff
@@ -4,12 +4,24 @@ class UnattendedController < ApplicationController
   layout false
 
   # We dont require any of these methods for provisioning
-  FILTERS = [:require_login, :session_expiry, :update_activity_time, :set_taxonomy, :authorize]
-  FILTERS.each do |f|
-    define_method("#{f}_with_unattended") do
-      send("#{f}_without_unattended") if preview?
```
iNecas, line 10: Couldn't we use:

```ruby
skip_before_action :require_login, :session_expiry, :update_activity_time, :set_taxonomy, :authorize, :unless => Proc.new { preview? }
```

app/models/subnet.rb
```diff
@@ -279,12 +279,11 @@ def subnet_for(ip)
     end
 
     # This casts Subnet to Subnet::Ipv4 if no type is set
-    def new_with_default_type(*attributes, &block)
+    def new(*attributes, &block)
       type = attributes.first.with_indifferent_access.delete(:type) if attributes.first.is_a?(Hash)
-      return Subnet::Ipv4.new_without_cast(*attributes, &block) if self == Subnet && type.nil?
```
iNecas, line 284: Wasn't there the `new_without_cast` for a reason there? Why wasn't there just `Subnet::Ipv4.new(*attributes, &block)` before?

test/controllers/hosts_controller_test.rb
iNecas: Where does the `save` call comes from?

iNecas (changes requested):

app/models/subnet.rb
mmoll: @timogoebel do you recall your original reason? :)

test/controllers/hosts_controller_test.rb
mmoll: app/models/concerns/foreman/sti.rb:26

app/controllers/unattended_controller.rb
mmoll: seems to work 👍

app/models/subnet.rb
mmoll: I think that's also because of the change in app/models/concerns/foreman/sti.rb

app/controllers/concerns/foreman/controller/smart_proxy_auth.rb
mmoll: I'm not fully getting it, could you lend me a hand there? ✋

app/models/subnet.rb
iNecas: As long as the tests are green, I'm ok with that, just wondering if there isn't something hidden around this.

app/controllers/concerns/foreman/controller/smart_proxy_auth.rb
iNecas: You would define separate module next to the `ClassMethods` like this;

```ruby
     module SmartProxyRequireSsl
        def require_ssl?
          if [self.smart_proxy_filter_actions].flatten.map(&:to_s).include?(self.action_name)
            false
          else
            super
          end
        end
      end
```
And then, here you would just do `prepend SmartProxyRequireSsl`

test/controllers/hosts_controller_test.rb
iNecas: So the two tests that have this issue are https://github.com/theforeman/foreman/blob/develop/test/controllers/hosts_controller_test.rb?utf8=%E2%9C%93#L1324 and https://github.com/theforeman/foreman/blob/develop/test/controllers/hosts_controller_test.rb?utf8=%E2%9C%93#L1336 and it's just because before this change, the `save` call would not be detected, because the actual method was `save_with_type` :D So we were actually calling the save there, and this expectation was not just catching it.

I think the issue was introduced by refactoring in https://github.com/theforeman/foreman/commit/b3fb472c60eeeb085859de58d771a5fec7de7ba3, where the expects were extracted to before filter without realizing there are some tests that actually call save (and we were not hitting issues just because of the aliasing).

We should move the expects just to the tests that really expect nothing to be saved: the explicitness would also help with readability of the expacations IMO

mmoll: Shouldn't the stuff from `describe '#template_used'` be isolated from the `test_submit_multiple_rebuild_config_*` tests?

iNecas: Yes, it seems those two are actually not related to the rest of the context. +1 for getting them out of `template_used`

mmoll: These two are not inside that `describe` block, so I have no idea why the `expects(:save).never` is triggered there

mmoll: ok, I think I got this...

mmoll:
:green_heart:

iNecas (approved):

iNecas:
Thanks @mmoll

PR #4852, opened 2017-09-17
Fixes #20966 - Added fast return to host power api

ShimShtein:
If the host does not support power actions (no BMC and the VM does not support it - return `unprocessable_entity` and exit with a message.

### DIFF
--- a/app/controllers/api/v2/hosts_controller.rb
+++ b/app/controllers/api/v2/hosts_controller.rb
@@ -251,6 +251,10 @@ Return the host's compute attributes that can be used to create a clone of this 
       param :power_action, String, :required => true, :desc => N_("power action, valid actions are (on/start), (off/stop), (soft/reboot), (cycle/reset), (state/status)")
 
       def power
+        unless @host.supports_power?
+          return render_error :custom_error, :status => :unprocessable_entity, :locals => { :message => _('Power operations are not enabled on this host.') }
+        end
+
         valid_actions = PowerManager::SUPPORTED_ACTIONS
         if valid_actions.include? params[:power_action]
           render :json => { :power => @host.power.send(params[:power_action]) }, :status => :ok


app/controllers/api/v2/hosts_controller.rb
```diff
@@ -252,6 +252,10 @@ def disassociate
 
       def power
         valid_actions = PowerManager::SUPPORTED_ACTIONS
+        unless @host.supports_power?
+          return render_message(_('This host does not support power operations'), :status => :unprocessable_entity)
```
iNecas: shouldn't this use `render_error`?

app/controllers/api/v2/hosts_controller.rb
```diff
@@ -252,6 +252,10 @@ def disassociate
 
       def power
         valid_actions = PowerManager::SUPPORTED_ACTIONS
```
tbrisker: while we're short circuiting, might as well prevent this variable assignment.

app/controllers/api/v2/hosts_controller.rb
tbrisker: the recipient expects a json response AFAIR, so we should keep that format. See below where we error on an unknown action. Also, might want to make sure the ui knows to handle this correctly and show the correct message to the user for the host (it might, i didn't test)

tbrisker (changes requested):

app/controllers/api/v2/hosts_controller.rb
ShimShtein: Hammer and the UI are OK with this message:

``` sh
$ hammer host status --id 1
»This host does not support power operations«
```

``` html
<span data-reactroot="" class="fa fa-power-off host-power-status na" title="»Power operations are not enabled on this host.«"></span>
```

Thanks for reminding me of the UI - completely forgot about it.

ShimShtein: `render_error` will generate a custom page, I still want the response to remain simple. I want hammer and UI to be able to parse the response.

ShimShtein:
Rebased and moved the assignment.

app/controllers/api/v2/hosts_controller.rb
tbrisker: wait, that isn't the message the server sent.

ShimShtein: Right! It comes from a different [controller](https://github.com/theforeman/foreman/blob/fb31c4264a762cd07f22dea0fb6592d620e4bfac/app/controllers/hosts_controller.rb#L277) for the UI. Sorry for misleading.

tbrisker: Could we use the same string for consistency and to save one string from translators?

ShimShtein:
Rebased and copied the string from UI controller.

tbrisker (approved):

tbrisker:
LGTM, any other comments @iNecas ?

app/controllers/api/v2/hosts_controller.rb
iNecas: `render_error` doesn't have to do anything with UI. The difference is how the json output looks like, but hammer should be able to process both. It feels to me still that this is more of an error state than just a message, although, we are using the `render_message` and `render_error` inconsistently  here, which I guess would need a new API version to unify. If we wanted to unify the way we render this kind of messages, I would go for render_error layout probably. But leaving the decision up to you: just be aware it has nothing to do with UI and hammer should be able to process the error.

ShimShtein:
Changed to `render_error` as per @iNecas comment. Indeed hammer knows how to work with it too.

iNecas:
Thanks @ShimShtein and @tbrisker

PR #4853, opened 2017-09-18
Fixes #20940 - Interfaces shows extra fields in API /api/hosts/id

snagoor:
This fix is to have a consistency between REST API `/api/hosts/:id` and `/api/hosts/:id/interfaces` where many important flags such as `managed`, `virtual`  are hidden in the former call and shows additional details in interfaces API call.

### DIFF
--- a/app/views/api/v2/interfaces/base.json.rabl
+++ b/app/views/api/v2/interfaces/base.json.rabl
@@ -1,6 +1,6 @@
 object @interface
 
-attributes :id, :name, :ip, :ip6, :mac, :fqdn, :identifier, :primary, :provision
+attributes :id, :name, :ip, :ip6, :mac, :fqdn, :identifier, :managed, :primary, :provision, :virtual, :created_at, :updated_at, :domain_id, :domain_name, :subnet_id, :subnet_name
 node :type do |i|
   next if i.is_a? Symbol
   i.class.humanized_name.downcase


snagoor:
After the fix 
~~~
    "interfaces": [
        {
            "id": 5,
            "name": "client.example.com",
            "ip": "192.168.121.65",
            "ip6": null,
            "mac": "52:54:00:a5:d6:b7",
            "fqdn": "client.example.com",
            "identifier": "ens3",
            "managed": true,
            "primary": true,
            "provision": true,
            "virtual": false,
            "created_at": "2017-08-07 05:39:33 UTC",
            "updated_at": "2017-08-07 05:39:37 UTC",
            "domain_id": null,
            "domain_name": null,
            "subnet_id": null,
            "subnet_name": null,
            "type": "interface"
        }
~~~

dLobatog (changes requested):
Thanks @snagoor , how about introducing a test that checks that the response for api/v2/hosts/interfaces and api/v2/hosts have the same attributes in the interfaces array? That way we ensure the inconsistency will not happen anymore.

tstrachota:
@snagoor would you mind to rebase?

I think that we should fix this one by including `interfaces/main.json.rabl` from `hosts/show.json.rabl` rather than by moving attributes from `base` to `main`. Since it's a detail info request for just one host it should be ok to provide full details. Interfaces are important aspect.

amitkarsale:
@snagoor any blocker you are facing over the PR?

amitkarsale:
On 1-0-1 conversation with @snagoor will take over this PR. Closing this PR and creating a new one.

amitkarsale:
Oops, I just realized I dont have close PR permissions. ;)
 @tstrachota or @dLobatog can you close this PR.

PR #4854, opened 2017-09-18
Fixes #19519 - Prevent autocomplete on LDAP account

tbrisker:


app/views/auth_source_ldaps/_form.html.erb
```diff
@@ -27,7 +27,7 @@
                     { :label => _('Server type'), :onchange => 'change_ldap_server_type()' } %>
     </div>
     <div class="tab-pane" id="account">
-      <%= text_f f, :account, :help_inline =>_("Use this account to authenticate, <i>optional</i>").html_safe, :label_help => account_help_data[@auth_source_ldap.server_type], :data => { :help => account_help_data } %>
+      <%= text_f f, :account, :help_inline =>_("Use this account to authenticate, <i>optional</i>").html_safe, :label_help => account_help_data[@auth_source_ldap.server_type], :data => { :help => account_help_data }, :autocomplete => 'new-account' %>
```
dLobatog, line 30: I can't get to reproduce the issue, I do have my credentials saved in Firefox yet they're not autocomplete. Anyway this looks like it would do it, but does it autocomplete for `new-account`? 

What do you think of setting `:html => {:autocomplete => "off"}` at the `form_for` so that it's disabled in the entire form?

dLobatog (changes requested):
Thanks @tbrisker, I wasn't quite sure about how to reproduce but it looks fine, can you check my suggestion?

app/views/auth_source_ldaps/_form.html.erb
tbrisker: This happens to me in chrome as well. Chrome ignores `:autocomplete => "off"` sadly. however when we use 'new-account' chrome thinks this is a new account name and doesn't try to autofill it. we use a similar approach for passwords.

dLobatog (approved):
Fair enough, apparently Chrome devs suggest the workaround even.. https://bugs.chromium.org/p/chromium/issues/detail?id=370363#c7 

Thanks @tbrisker , merging

PR #4859, opened 2017-09-19
Fixes #21024 - Reorder AR associations in order of use

mmoll:


mmoll:

iNecas:
Makes sense. Thanks @mmoll

PR #4861, opened 2017-09-20
Fixes #21026 - Clear cached notifications properly

bastilian:
Using `delete_all` for has_many does not trigger callbacks
on the associated objects, `dependent: :destroy` does.

Clearing the cache needs to happen anytime a notification
gets created, modified and destroyed, `after_commit` will
be triggered on all o these occasions.

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -15,7 +15,7 @@ class Notification < ApplicationRecord
   belongs_to :notification_blueprint
   belongs_to :initiator, :class_name => 'User', :foreign_key => 'user_id'
   belongs_to :subject, :polymorphic => true
-  has_many :notification_recipients, :dependent => :delete_all
+  has_many :notification_recipients, :dependent => :destroy
   has_many :recipients, :class_name => 'User', :through => :notification_recipients, :source => :user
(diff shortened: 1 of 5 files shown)


mmoll:
please rebase :)

iNecas:
It looks good to me, but would rather have @ohadlevy or @tbrisker confirmation this is how it's supposed to work

iNecas:
test failures unrealted

app/models/notification.rb
```diff
@@ -15,8 +15,8 @@ class Notification < ApplicationRecord
   belongs_to :notification_blueprint
   belongs_to :initiator, :class_name => 'User', :foreign_key => 'user_id'
   belongs_to :subject, :polymorphic => true
-  has_many :notification_recipients, :dependent => :delete_all
-  has_many :recipients, :class_name => 'User', :through => :notification_recipients, :source => :user
+  has_many :notification_recipients, :dependent => :destroy
+  has_many :recipients, :class_name => User, :through => :notification_recipients, :source => :user
```
mmoll: this should be `:class_name => 'User'`

tbrisker (approved):

tbrisker:
Thanks @bastilian !

PR #4863, opened 2017-09-20
Fixes #21037 - use Host::Base as default for Host module

iNecas:
With an exception to create and new methods, that for backward
compatibility should redirect to Host::Managed.

Otherwise, there are issues in audited gem, that uses the `Host`
type in `auditable_class` attribute in the auditable polymorphic
association.

### DIFF
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -1,8 +1,8 @@
 module Host
   def self.method_missing(method, *args, &block)
-    type = "Host::Managed"
     case method.to_s
     when /create/, 'new'
+      type = "Host::Managed"
       if args.empty? || args[0].nil? # got no parameters
         #set the default type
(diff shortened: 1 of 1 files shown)


iNecas:
See http://ci.theforeman.org/job/test_plugin_matrix/3760/database=postgresql,ruby=2.3,slave=fast/testReport/junit/(root)/Api__V2__DiscoveredHostsControllerTest/test_auto_provision_with_wrong_disabled_org_sucess/ for the errors I'm talking about

iNecas:
The fight of converting all of the usages of `Host.method` is too big for me to fight. A smaller fix should be possible for the auditing to work properly and not cause the errors https://github.com/theforeman/foreman/pull/4864

PR #4864, opened 2017-09-20
Fixes #21037 - use Host::Base as auditable_type

iNecas:
Otherwise, we are hitting "Invalid single-table inheritance type:
Host::Discovered is not a subclass of Host::Managed" when converting the
host from discovered to managed.

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -64,9 +64,13 @@ module Api
 
       return resource_class.where(nil) unless scope
 
-      association = resource_class.reflect_on_all_associations.find {|assoc| assoc.plural_name == parent_name.pluralize}
+      association = resource_class.reflect_on_all_associations.detect {|assoc| assoc.plural_name == parent_name.pluralize}
       #if couldn't find an association by name, try to find one by class
-      association ||= resource_class.reflect_on_all_associations.find {|assoc| assoc.class_name == parent_name.camelize}
+      association ||= resource_class.reflect_on_all_associations.detect {|assoc| assoc.class_name == parent_name.camelize}
+      if association.nil? && parent_name == 'host'
+        association = resource_class.reflect_on_all_associations.detect {|assoc| assoc.class_name == 'Host::Base'}
+      end
+      raise "Association not found for #{parent_name}" unless association
       result_scope = resource_class_join(association, scope)
       # Check that the scope resolves before return
       result_scope if result_scope.to_a
--- a/app/helpers/audits_helper.rb
+++ b/app/helpers/audits_helper.rb
@@ -1,5 +1,5 @@
 module AuditsHelper
-  MAIN_OBJECTS = %w(Host Hostgroup User Operatingsystem Environment Puppetclass Parameter Architecture ComputeResource ProvisioningTemplate ComputeProfile ComputeAttribute
+  MAIN_OBJECTS = %w(Host::Base Hostgroup User Operatingsystem Environment Puppetclass Parameter Architecture ComputeResource ProvisioningTemplate ComputeProfile ComputeAttribute
                     Location Organization Domain Subnet SmartProxy AuthSource Image Role Usergroup Bookmark ConfigGroup Ptable)
 
   # lookup the Model representing the numerical id and return its label
@@ -129,6 +129,8 @@ module AuditsHelper
 
   def audited_type(audit)
     type_name = case audit.auditable_type
+                  when 'Host::Base'
+                    'Host'
                   when 'HostClass'
                     'Puppet Class'
                   when 'Parameter'
@@ -161,6 +163,7 @@ module AuditsHelper
   private
 
   def main_object?(audit)
+    return true if MAIN_OBJECTS.include?(audit.auditable_type)
     type = audit.auditable_type.split("::").last rescue ''
     MAIN_OBJECTS.include?(type)
   end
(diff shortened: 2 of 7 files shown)


iNecas:
Discovery tests running here http://ci.theforeman.org/job/test_plugin_matrix/3768/

iNecas:
@lzap this should help you with the "Invalid single-table inheritance type: Host::Discovered is not a subclass of Host::Managed" issue. The other test failures should be just regular rails5 deprecation warnings.

app/controllers/api/base_controller.rb
```diff
@@ -67,6 +67,10 @@ def parent_scope
       association = resource_class.reflect_on_all_associations.find {|assoc| assoc.plural_name == parent_name.pluralize}
       #if couldn't find an association by name, try to find one by class
       association ||= resource_class.reflect_on_all_associations.find {|assoc| assoc.class_name == parent_name.camelize}
+      if association.nil? && parent_name == 'host'
+        association = resource_class.reflect_on_all_associations.find {|assoc| assoc.class_name == 'Host::Base'}
```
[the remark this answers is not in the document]
iNecas: whatever

iNecas:
In case anyone interested: all green :)

lzap:
@iNecas dude you kick ass! Lemme take a look, may I ask for double review @mhulan or anyone

lzap (approved):
Cool, WFM. The find vs detect was not related, wasn't it?

iNecas:
No, it was hound barking

app/models/concerns/audit_extensions.rb
```diff
@@ -103,8 +103,8 @@ def ensure_username
 
   def fix_auditable_type
     # STI Host class should use the stub module instead of Host::Base
-    self.auditable_type = "Host"          if auditable_type =~  /Host::/
-    self.associated_type = "Host"         if associated_type =~ /Host::/
+    self.auditable_type = "Host::Base"          if auditable_type =~  /Host::/
+    self.associated_type = "Host::Base"         if associated_type =~ /Host::/
```
ares: this is exactly the reason why we shouldn't align code with whitespaces :-) could you please remove them?

ares (changes requested):
The problem now is that any host audit is logged as change to Host/Base (see the screenshot). If for some reason we need this, we should make sure, it's presented as a Host. It should be also listed here https://github.com/theforeman/foreman/blob/develop/app/helpers/audits_helper.rb#L2, https://github.com/theforeman/foreman/blob/develop/app/helpers/audits_helper.rb#L118 and some other places in audits helper.

There are other places that would need to be updated, https://github.com/theforeman/foreman/blob/develop/app/views/audits/show.html.erb#L6

Would it be better to take the same approach as we did for templates STI? See https://github.com/theforeman/foreman/pull/4778 for what I mean.

host

iNecas:
Screenshot missing

iNecas:
I don't think https://github.com/theforeman/foreman/pull/4778 is the same, as in host case we actually change the type of the class as we go (from Discovered to Managed), while in templates, this is immutable. I will try to update the missed places.

iNecas:
Nm for the screenshots: I've seen it by myself: updated + addressed other comments.

lzap:
LGTM @ares ?

ares:
Sorry to keep asking... how does this happen in the first place? I don't see `audited` definition in Host::Discovered nor Host::Base, it's in Host::Managed only. AFAIK there's no Managed -> Discovered transition. I don't want to block it, just want to understand, what we're trying to solve here. This basically means any subclass of Host will be audited as generic Host. No way to search audits based on host type. I know we have only managed and discovered hosts so far, but before we commit to this, I'd like to know why :-)

iNecas:
That's definitely valid reason to ask.

The problem is, when saving the audit records for the host, the audit record is being created first, and as part of the process, the auditable record (the host) is being loaded again from the database: and this is where the issue comes, as the auditable_type was marked as `Host` which translates to `Host::Managed` and the `Host::Discovered` was set in the database, which lead to the error `"Invalid single-table inheritance type: Host::Discovered is not a subclass of Host::Managed`.

Since we allow transitions of types for host record, I think it makes sense to keep the auditable_type as `Host::Base`. Honestly, the Host should not be an STI in first place and discovered host should be just a record on it's own: but that's outside of scope for this PR

lzap:
@ares hey are we good to go?

app/helpers/audits_helper.rb
```diff
@@ -115,7 +115,7 @@ def audited_icon(audit)
 
     type   = audited_type(audit)
     symbol = case type
-               when "Host"
+               when "Host::Base"
```
ares: this is `'Host'` actually, since audited_type converts it

ares (changes requested):
Just one tiny thing re icons, otherwise seems to work fine, good to merge afterwards

ares (approved):

ares:
Ready for merge pending tests

iNecas:
Tests are green

lzap:
Thanks a bunch folks, really appreciated for unblocking discovery tests.

ares:
I set the release to 1.17, @lzap if you want this to be cherry-picked to 1.16, please updated the redmine issue. If it's accepted, then you should also do the cherry-pick to 1.16-stable.

kotyara85:
Facing the same issue on 1.17. Can't provision hosts unless I'm an admin. https://projects.theforeman.org/issues/23869

PR #4865, opened 2017-09-21
Fixes #15286 - New API for auth sources

dhlavac:


mmoll:
[test foreman]

mmoll:
[test foreman]

tbrisker:
Tests are failing, setting to WoC

dhlavac:
Some test still failing

app/models/auth_sources/auth_source_external.rb
```diff
@@ -1,4 +1,7 @@
 class AuthSourceExternal < AuthSource
+  include Taxonomix
+  scoped_search :on => :name, :complete_value => :true
```
ares: seems to be also defined in the AuthSource parent class, do we need to defined it here again?

app/models/auth_sources/auth_source_internal.rb
```diff
@@ -1,4 +1,6 @@
 class AuthSourceInternal < AuthSource
+  scoped_search :on => :name, :complete_value => :true
```
ares: seems to be also defined in the AuthSource parent class, do we need to defined it here again?

app/registries/foreman/access_permissions.rb
```diff
@@ -33,15 +33,19 @@
     ajax_actions = [:test_connection]
     map.permission :view_authenticators, {:auth_source_ldaps => [:index, :show],
                                              :"api/v1/auth_source_ldaps" => [:index, :show],
-                                             :"api/v2/auth_source_ldaps" => [:index, :show]
+                                             :"api/v2/auth_source_ldaps" => [:index, :show],
+                                             :"api/v2/auth_sources" => [:index],
```
ares: this seems to be missing show action, which is why tests failed

db/migrate/20171121111333_change_auth_source_resource_type.rb
```diff
@@ -0,0 +1,9 @@
+class ChangeAuthSourceResourceType < ActiveRecord::Migration
+  def up
+    Permission.where(:resource_type => 'AuthSourceLdap').update_all(:resource_type => 'AuthSource')
+  end
+
+  def down
+    Permission.update_all(:resource_type => 'AuthSourceLdap')
```
ares: this seems dangerous :-) shouldn't it be limited by some `where`?

test/controllers/api/v2/auth_source_externals_controller_test.rb
```diff
@@ -0,0 +1,43 @@
...
+    assert_response :success
+  end
+
+  test 'taxonomies can be set' do
+    put :update, { :id => auth_sources(:external).to_param,
+                   :organization_names => [taxonomies(:organization1).name],
+                   :location_ids => [taxonomies(:location1).id] }
+    show_response = ActiveSupport::JSON.decode(@response.body)
+    assert_response :success
+    assert_equal taxonomies(:location1).id, show_response['locations'].first['id']
```
ares: I suppose this test fails because of strong params, I think attributes (except for id) must be wrapped in the resource hash, e.g.
```
 put :update, { :id => auth_sources(:external).to_param, :auth_source_external => {
                  :organization_names => [taxonomies(:organization1).name],
                  :location_ids => [taxonomies(:location1).id]
                 }
               }
```

ares (changes requested):

dhlavac:
@ares  Thanks for review, changes done ... Migration was too dangerous

ares:
katello tests failing with 
```
14:29:04 ++ grep -Po '(\d+\.)+\d+' /var/lib/workspace/workspace/test_develop_pr_katello/database/postgresql/ruby/2.2/slave/fast/plugin + /lib/katello/version.rb
14:29:04 grep: /var/lib/workspace/workspace/test_develop_pr_katello/database/postgresql/ruby/2.2/slave/fast/plugin: Is a directory
14:29:04 grep: +: No such file or directory
14:29:04 grep: /lib/katello/version.rb: No such file or directory
14:29:04 + VERSION=
```

but all tests succeeded so I don't think it's related

app/controllers/api/v2/auth_source_externals_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+
+      def_param_group :auth_source_external do
+        param :auth_source_external, Hash, :required => true, :action_aware => true do
+          param :name, String, :required => true
+          param_group :taxonomies, ::Api::V2::BaseController
+        end
+      end
+
+      api :PUT, "/auth_source_externals/:id/", N_("Update an external authentication source")
+      param :id, String, :required => true
```
ares: I think `id` should be `:identifier` type, not a String

app/controllers/api/v2/auth_source_internals_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+  module V2
+    class AuthSourceInternalsController < V2::BaseController
+      before_action :find_resource, :only => %w{show}
+
+      api :GET, "/auth_source_internals/", N_("List internal authentication sources")
+      api :GET, '/locations/:location_id/auth_source_internals/',
+          N_('List internal authentication sources per location')
+      api :GET, '/organizations/:organization_id/auth_source_internals/',
+          N_('List internal authentication sources per organization')
+      param_group :taxonomy_scope, ::Api::V2::BaseController
```
ares: internal auth source does not include `Taxonomix` so it can't be associated with any org or loc, this and also the endpoints under /locations/ and /organizations above should not be here

app/controllers/api/v2/auth_sources_controller.rb
```diff
@@ -0,0 +1,17 @@
+module Api
+  module V2
+    class AuthSourcesController < V2::BaseController
+      api :GET, "/auth_sources/", N_("List all authentication sources")
+      api :GET, '/locations/:location_id/auth_sources/',
+          N_('List authentication sources per location')
+      api :GET, '/organizations/:organization_id/auth_sources/',
```
ares: this can be tricky, since internal auth source does not support taxonomies, not all records returned here can be scoped like this, I don't expect there to be more than several auth sources so we I think we can drop these routes

app/controllers/api/v2/auth_sources_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+          N_('List authentication sources per organization')
+      param_group :taxonomy_scope, ::Api::V2::BaseController
```
ares: same as above

app/controllers/api/v2/auth_sources_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+      param_group :search_and_pagination, ::Api::V2::BaseController
+
+      def index
+        @auth_sources = AuthSource.except_hidden.to_a.delete_if { |a| a.to_label.nil? }
```
ares: I suppose you added this because of `to_label` method could return nil. I think we can just remove the return from there, it dates back to commit 6874bbd96a from 2010 and I don't think it's still relevant. Every auth source needs to have type. So if you remove it from [AuthSource](https://github.com/theforeman/foreman/blob/develop/app/models/auth_source.rb#L44-L47), you don't have to do `to_a.delete_if {}` here at all.

test/controllers/api/v2/auth_source_externals_controller_test.rb
```diff
@@ -0,0 +1,45 @@
...
+    assert_not_nil assigns(:auth_source_externals)
+    auth_source_externals = ActiveSupport::JSON.decode(@response.body)
+    refute auth_source_externals.empty?
+  end
+
+  test "should show auth_source_external" do
+    get :show, { :id => auth_sources(:external).to_param }
+    assert_response :success
+    show_response = ActiveSupport::JSON.decode(@response.body)
+    assert !show_response.empty?
```
ares: either `refute show_response.empty?` or even better `refute_empty show_response`, this would generate more meaningful message on failure

test/controllers/api/v2/auth_source_externals_controller_test.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  test "should update auth_source_external" do
+    put :update, { :id => auth_sources(:external).to_param, :auth_source_external => valid_attrs }
+    assert_response :success
```
ares: it would be good to change the name to e.g. "new_name" (the fixture name is External), load the resource based on id that we got as response and then assert_equal the 'new_name' and resource.name. This would verify the update works as expected. This test currently only verifies it returns 200 which would accept even empty update action :-)

test/controllers/api/v2/auth_source_externals_controller_test.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  test 'allow access to auth source external objects' do
+    setup_user('view', 'authenticators')
+    auth_sources(:external).organizations = User.current.organizations
+    auth_sources(:external).locations = User.current.locations
+    get :show, { :id => auth_sources(:external).to_param }
+    assert_response :success
+  end
+
+  test 'taxonomies can be set' do
```
ares: :+1: crucial test to have

test/controllers/api/v2/auth_source_externals_controller_test.rb
```diff
@@ -0,0 +1,45 @@
...
+    put :update, { :id => auth_sources(:external).to_param,
+                   :auth_source_external => {
+                     :organization_names => [taxonomies(:organization1).name],
+                     :location_ids => [taxonomies(:location1).id] }
+    }
+    show_response = ActiveSupport::JSON.decode(@response.body)
+    assert_response :success
+    assert_equal taxonomies(:location1).id, show_response['locations'].first['id']
```
ares: I think it would be slightly better if we loaded the resource and verify the associations using assert_include so we don't rely on order of orgs using `first`
```
auth_source = AuthSource.find(show_response['id'])
assert_include auth_source.locations, taxonomies(:location1)
assert_include auth_source.organizations, taxonomies(organization1)
```

test/controllers/api/v2/auth_source_internals_controller_test.rb
```diff
@@ -0,0 +1,18 @@
...
+    assert_not_nil assigns(:auth_source_internals)
+    auth_source_internals = ActiveSupport::JSON.decode(@response.body)
+    refute auth_source_internals.empty?
+  end
+
+  test "should show auth_source_internal" do
+    get :show, { :id => auth_sources(:internal).to_param }
+    assert_response :success
+    show_response = ActiveSupport::JSON.decode(@response.body)
+    assert !show_response.empty?
```
ares: it would be good to assert we got the right record, e.g. by
```
assert_equal auth_sources(:internal).name, show_response['name']
```
or by using id

test/controllers/api/v2/auth_sources_controller_test.rb
```diff
@@ -0,0 +1,10 @@
+require 'test_helper'
+
+class Api::V2::AuthSourcesControllerTest < ActionController::TestCase
+  test "should get index" do
+    get :index, { }
+    assert_response :success
+    auth_sources = ActiveSupport::JSON.decode(@response.body)
+    assert !auth_sources.empty?
```
ares: I think it would be good to verify we get various auth source types, e.g. by something like

```
assert_includes ids, auth_sources(:internal).id
```

ares (changes requested):
Leaving few more cosmetic comments but it does not prevent me from testing so I'll continue with that. Don't worry, we're getting there :-)

app/views/api/v2/auth_source_internals/show.json.rabl
```diff
@@ -0,0 +1,7 @@
+object @auth_source_internal
+
+extends "api/v2/auth_source_internals/main"
+
+node do |auth_source_internal|
+  partial("api/v2/taxonomies/children_nodes", :object => auth_source_internal)
```
ares: internal don't seem to support taxonomies, this results in {} sent for both org and loc ids

app/controllers/api/v2/auth_source_externals_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+      param_group :auth_source_external
+
+      def update
```
ares: I see the update but there's no create, should it be added?

ares: actually nevermind, I think the last agreement was that we'll support only one external auth source defined through seeds

app/controllers/api/v2/auth_source_externals_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+  module V2
+    class AuthSourceExternalsController < V2::BaseController
+      include Foreman::Controller::Parameters::AuthSourceExternal
+
+      before_action :find_resource, :only => %w{show update}
+
+      api :GET, "/auth_source_externals/", N_("List external authentication sources")
+      api :GET, '/locations/:location_id/auth_source_externals/',
+          N_('List external authentication sources per location')
+      api :GET, '/organizations/:organization_id/auth_source_externals/',
```
ares, line 11: this does not seem to work, seems the route for this endpoint is not defined, the same for location and most likely all auth source types, either remove it or add routes

ares:
I know you already mentioned this but as part of the testing under non-admin I found an issue that searching based on organization_id does not work. Therefore I'm unable to create LDAP auth source. The response I'm getting

```json
{
    "error": {
        "message": "Field 'organization_id' not recognized for searching!",
        "class": "ScopedSearch::QueryNotSupported"
    }
}
```

app/models/permission.rb
```diff
@@ -32,6 +32,7 @@ def self.resource_name(klass)
     return 'ComputeResource' if klass <= ComputeResource
     return 'Subnet' if klass <= Subnet
     return 'Parameter' if klass <= Parameter
+    return 'AuthSource' if klass.superclass <= AuthSource
```
ares: why we call `.superclass` here? The `<=` operator means AuthSource or child of it (any level of inheritance), should just `klass` do?

ares (changes requested):
I left one comment re superclass and opened a PR with the fix we discussed this morning at https://github.com/dhlavac/foreman/pull/1

app/models/permission.rb
dhlavac: Thanks for comment .. changed to klass only

dhlavac:
@ares  thanks for PR, changes done but maybe we need to add some method for adding internal auth sources because now scoped_search shows only external and ldap authsources

dhlavac:
@ares refactored - rails 5.1

app/controllers/api/v2/auth_source_externals_controller.rb
ares: resolved now

ares (approved):
ready from my point of view, would you mind squashing to single commit? @jyejare any updates from your testing?

app/controllers/api/v2/auth_source_externals_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+      param_group :search_and_pagination, ::Api::V2::BaseController
+
+      def index
+        @auth_source_externals = resource_scope_for_index
+      end
+
+      api :GET, "/auth_source_externals/:id/", N_("Show an external authentication source")
+      param :id, :identifier, :required => true
+
+      def show
```
jyejare, line 23: Declared but not defined?

app/controllers/api/v2/auth_source_internals_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+      before_action :find_resource, :only => %w{show}
+
+      api :GET, "/auth_source_internals/", N_("List internal authentication sources")
+      param_group :search_and_pagination, ::Api::V2::BaseController
+
+      def index
+        @auth_source_internals = resource_scope_for_index
+      end
+
+      api :GET, "/auth_source_internals/:id/", N_("Show an internal authentication sources")
```
jyejare: s/sources/source/

app/controllers/api/v2/users_controller.rb
```diff
@@ -15,6 +15,7 @@ class UsersController < V2::BaseController
 
       api :GET, "/users/", N_("List all users")
       api :GET, "/auth_source_ldaps/:auth_source_ldap_id/users", N_("List all users for LDAP authentication source")
+      api :GET, "/auth_source_external/:auth_source_external_id/users", N_("List all users for external authentication source")
```
jyejare: s/auth_source_external/auth_source_externals/

jyejare (changes requested):
ACK pending doubts/bugs and comments:

Following Tests are performed:
```
Tests:

1. GET to list Auth Source Externals is accessible from API
2. GET for specific Auth source external is acessible by id from API
3. PUT for Auth Source externals works wrt taxonomies from API
4. GET for Auth Source Externals is accessible through locations and organizations by API
5. GET to list Auth Source Internals is accessible from API
6. GET for specific Auth source internal is acessible by id from API
7. GET to list Auth Sources is accessible from API
8. GET for Auth Sources is accessible through locations and organizations by API
```

Doubts/Bugs: 
```
1. The created and updated date is same after PUT
2. GET for specific auth source internals is not returning location and organization details but its there while listing internal auth sources
3. No PUT for internal auth sources defined
4. The total number of auth sources in GET is showing wrong count
```

ares:
> 1. The created and updated date is same after PUT

PUT should modify updated_at if there was at least one attribute that has changed, e.g. name

> 2. GET for specific auth source internals is not returning location and organization details but its there while listing internal auth sources

internal sources should not have org/loc so it should not be listed in list

> 3. No PUT for internal auth sources defined

that is correct, it's system only auth source

> 4. The total number of auth sources in GET is showing wrong count

Just reproduced, not that this only happens for non-admin user, e.g. org-admin. This seems to be caused by the fact the controller uses `view_auth_source_internals` which is not defined. @dhlavac you need to add following to internal auth sources controller like you did for ldaps.

```ruby
def controller_permission
  'authenticators'
end
```

that unfortunatelly won't fix the original issue though, since non-admin user can't access resources not assigned to any org/loc. Which is always the case for internal auth source. Because of the nature of this resource, I think it's acceptable.

jyejare:
@ares , Just forgot to mention that these tests have been performed from 'admin' user.

app/controllers/api/v2/auth_source_externals_controller.rb
ares: that's fine, https://github.com/theforeman/foreman/pull/4865/files/92509bf2ca2ae5ea6229c9cadf01850b90ef7a74#diff-8d00a4cc43156275141995eb368a3cc2R6 is the only thing this method does which is why it's empty

dhlavac:
@jyejare @ares Thanks for review - changes done

@jyejare 
1) works good for me - as Marek mention you need to change at least one attribute
2) done 
3) that is correct
4) Total number is not same as listed items - that's true because there is one more authentication source - Hidden in that way it return right value  - what do you think @ares

jyejare:
@dhlavac , @ares , 

I retested:

1. The Created and updated datetime is not being changed when org and loc is updated but it changes when name is updated. (Low Priority)
2. It is now fixed
3. Gotcha
4. @ares, awaiting for your reply on this as the behavior still persists.

ares:
@dhlavac for the 4, you just need to make sure that controllers' resource scope is also ignoring hidden auth source, see https://github.com/theforeman/foreman/blob/develop/app/controllers/api/v2/base_controller.rb#L55 for reasoning

```diff
diff --git a/app/controllers/api/v2/auth_sources_controller.rb b/app/controllers/api/v2/auth_sources_controller.rb
index afb3a1fdd..23b2fb315 100644
--- a/app/controllers/api/v2/auth_sources_controller.rb
+++ b/app/controllers/api/v2/auth_sources_controller.rb
@@ -16,6 +16,10 @@ module Api
       def controller_permission
         'authenticators'
       end
+
+      def resource_scope(*args)
+        super.except_hidden
+      end
     end
   end
 end
```

then it seems to work fine and it's the last point that block this PR

dhlavac:
@ares thanks .. chnges done

jyejare:
@dhlavac  @ares,

Updates:

1. The updated_at is getting changed but its wrong date. I updated external_auth_source with org and loc today... but its giving 5 days old date ```"updated_at":"2018-03-15 07:53:06 UTC```
4. Its now fixed and I see only internal auth source is not listed. And the total number of auth_sources is correct.

That said, The point 1 is still need to fix.

jyejare:
@dhlavac 

As per discussion with @ares that point 1 issue seems to be generic with all the entities and hence need to raise new issue for that sort of behavior.

So, ACK to this PR as all the doubts/buggy behaviors has been taken care.

ares:
Thanks @dhlavac, merging! And thanks @jyejare for testing.

app/models/auth_source.rb
```diff
@@ -27,8 +43,8 @@ class AuthSource < ApplicationRecord
 
   validates :name, :presence => true, :uniqueness => true, :length => { :maximum => 60 }
 
-  scope :non_internal, -> { where("type NOT IN (?)", ['AuthSourceInternal', 'AuthSourceHidden']) }
-  scope :except_hidden, -> { where('type <> ?', 'AuthSourceHidden') }
+  scope :non_internal, -> { where("auth_sources.type NOT IN (?)", ['AuthSourceInternal', 'AuthSourceHidden']) }
+  scope :except_hidden, -> { where('auth_sources.type <> ?', 'AuthSourceHidden') }
```
tbrisker, line 47: Just FYI, for future knowledge - you can now use `where.not(type: 'AuthSourceHidden')` or `where.not(type: ['AuthSourceInternal', 'AuthSourceHidden'])` to generate the same queries in a cleaner way

dhlavac: Thanks, good to know

PR #4866, opened 2017-09-21
Fixes #21055 - use _id attributes for host taxonomy validator

iNecas:
Otherwise, the error messages are not visible.

Also log the host error messages as `error` instead of `info`, as it
should be more appropriate for this case.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -284,7 +284,7 @@ class ApplicationController < ActionController::Base
       end
     end
 
-    logger.info "Failed to save: #{hash[:object].errors.full_messages.join(', ')}" if hash[:object].respond_to?(:errors)
+    logger.error "Failed to save: #{hash[:object].errors.full_messages.join(', ')}" if hash[:object].respond_to?(:errors)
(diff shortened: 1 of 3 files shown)


iNecas:
Btw. I was able to reproduce the issue (resp. to hit the validator errors), only when I forced to always trigger the validator in https://github.com/theforeman/foreman/blob/1.15.4/app/validators/belongs_to_host_taxonomy_validator.rb#L16, Perhaps the way we show the list of subnets has improved in time

```ruby
record.errors.add(attribute_name, _("is not defined for host's %s") % _(taxonomy))
```

Without the patch, the the error is not visible in the form at all, which can be quite frustrating
for the user.

rex-show-taxonomy-validator-issues

tbrisker:
@orrabin - sounds similar to some issue you had IIRC, care to take a look?

mmoll:
what's the status here?

iNecas:
Still waiting for review: I've rebased to retrigger the jenkins, but there were no conflicts, so it should be still valid in the original form.

mmoll (approved):

mmoll:
merged, díky @iNecas!

PR #4867, opened 2017-09-23
Fixes #21084 - Always Use Rails 5.0, Drop Rails 4.2

mmoll:


config/settings.yaml.example
```diff
@@ -26,9 +26,9 @@
 # errors in your browser
 :webpack_dev_server_https: false
 
-# Ruby on Rails version (available: 4.2, 5.0)
-# Defaults to 4.2 on Ruby 2.0-2.2, or 5.0 on Ruby 2.3+
-#:rails: 4.2
+# Ruby on Rails version
+# Defaults to 5.0
+#:rails: 5.0
```
tbrisker, line 31: Do we actually still want this setting and all it's related code if we only support one version?

mmoll: I find it very useful e.g. on the road from 5.0 to 5.1.

mmoll:
As Katello is not yet Rails 5 compatible that failure is expected, 💚 otherwise.

iNecas:
@bbuckingham do you have any estimates on Katello being Rails 5 ready?

Before we go the path that as not turning back, are we all set with the approach we're taking with packaging efforts, issues with SCL etc?

bbuckingham:
@iNecas, @jlsherrill and I talked briefly about Katello RAILS 5 support today and our goal is to complete the updates in the current sprint.  That said, it may be good to have a short buffer past that (in case there is anything unexpected).  The current sprint ends on Oct 9.

mmoll:
@bbuckingham what's the current status in Katello land?

mmoll:
[test katello]

jlsherrill:
Should we wait until we at least have some of the rails 5 packaged for production?

mmoll:
@jlsherrill the plan was to merge this and work towards Rails 5.1

ehelms:
I believe that while this is a good step for development, this will hard break nightly builds and leave us in a state of brokenness that can last for anywhere from 1 day to 3-4 months. This is not intended to be hyperbole. We know that this will break nightly production builds. Which means that until we established a production Rails environment, any number of other changes may go into the codebase that results in compounding breakages. In the past, these compounding breakages have required many developer hours to dig us out of and often have demoralized the folks working on it. Especially given that it could have been avoided in this case.

TLDR; What value do we gain from this if we can't at the same time deliver this value to users?

mmoll:
I'll open an updated 5.1 PR for interested parties to take over...

mmoll:
reopening to keep this around. Even with a direct jump to 4.2->5.1 RPM-wise, having this as a separate change for a limited time would help updating plugins to be 5.0 only compatible before going to 5.1.

mmoll:
@ehelms could you have a look into the Katello failure? It seems `db:drop` does error out now when there's no DB to drop.

ehelms:
[test katello]

ehelms:
@mmoll I opened https://github.com/theforeman/foreman-infra/pull/397 to help fix this.

mmoll:
@ehelms getting further now, you might want to add a `config/as_deprecation_whitelist.yaml` in katello

jlsherrill:
[test katello]

mmoll:
katello is green now after the turbolinks update 💚

ohadlevy (approved):

ohadlevy:
I assume that once this is merged nightly will be broken. Can I have ack/nack ? /Cc @ehelms

ohadlevy:
ok, I'm going to take the blame to pushing this forward - thanks @mmoll for your efforts!!

PR #4870, opened 2017-09-25
Fixes #21088 - Update default release for JS deprecation

tbrisker:


ohadlevy (approved):

iNecas:
There is one deprecation call https://github.com/theforeman/foreman/blob/eda35acd6de206f7dd1fb50af5a990376c58deee/app/assets/javascripts/application.js#L358 that was introduced in 1.16 (https://github.com/theforeman/foreman/commit/d91a0e4d219560000c5cc9c145650bab2039ba3b). Do we want to push that one version forward or should we remove it before merging this one?

tbrisker:
@iNecas thanks, removed that function in a separate commit, no point pushing it another version.

iNecas (approved):

webpack/assets/javascripts/foreman_users.js
```diff
@@ -1,5 +1,6 @@
 import $ from 'jquery';
 import _ from 'lodash';
+import { notify } from './foreman_toast_notifications';
```
ohadlevy, line 3: I dont think this is enough as the function signature has changed.

tbrisker: it will just pass an undef to the missing arguments which i think is fine (this is the same as we passed before in the deprecated function)

tbrisker: or actually looking at it now i think you may be right

iNecas:
Tested and works well. Thanks @tbrisker and @ohadlevy

PR #4871, opened 2017-09-25
dnm - load dynflow via Procfile in development

ohadlevy:
I was always bothered by the fact that dynflow works in the same
process as foreman (in dev mode), this patch allows to load dynflow
as a seperate proccess instead.

the motiviation is to stop loading all of the gems all the time in
both foreman and dynflow, where they are not really required in many
cases, for example, dynflow doesn't need all of our UI/API, doesn't
require mime types or safemode etc - all of which consume a lot of
memory and adds a lot of startup time.

I'm pretty sure my implementation is wrong, and also looking
for ack/nack/comments in general for this approch.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -3,3 +3,4 @@
 rails: [ -n "$RAILS_STARTUP" ] && env PRY_WARNING=1 $RAILS_STARTUP || [ -n "$BIND" ] && bin/rails server -b $BIND || env PRY_WARNING=1 bin/rails server
 # you can use WEBPACK_OPTS to customize webpack server, e.g. 'WEBPACK_OPTS='--https --key /path/to/key --cert /path/to/cert.pem --cacert /path/to/cacert.pem' foreman start '
 webpack: ./node_modules/.bin/webpack-dev-server --config config/webpack.config.js $WEBPACK_OPTS
+dynflow: bin/rake dynflow:executor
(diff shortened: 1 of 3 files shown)


iNecas:
I like the idea of dynflow not loading everything. Not sure if I understand this PR correctly though: the goal is to start with starting it separately and when move some parts out to avoid situation, where it would work in development (because of the all-in-one) but would break in production?

One of the benefits of having the dynflow executor in the same process is that it allows even the code reloading to work, which has been pretty neat, especially when working more on the backend part.

I would probably start the other way around: the first step to have the dependencies out, if then thinking about moving it out from the devel environment (which I'm not even sure is necessary at this phase).

ShimShtein:
> it allows even the code reloading to work

We can mitigate it, Sinatra has a reloader: http://www.sinatrarb.com/contrib/reloader.

iNecas:
I'm not sure the sinatra realoader would work, as we don't go over sinatra when we trigger the run. We could probably do it via doing the reload if needed at every execution plan execution start in development. Anyway, I would like to have some benefits to having to have another process running in development, before we proceed with this change.

ohadlevy:
I'm not planning to work on this, closing for now.

lzap:
Late to the party, but started thread on this: https://community.theforeman.org/t/on-spawning-threads-in-our-rails-app/8650

PR #4872, opened 2017-09-26
Fixes #21104 - Recognize br-ex etc as bridge interfaces

Ntr0:
Extends the bridge regex pattern to allow bridges named after purpose
like br-ex, br-tun as documented in openstack dvr scenario.


boaz0 (commented):
SGTM 👍

lzap:
Thanks.

PR #4873, opened 2017-09-26
Refs #19050 - Use Rails 4.2 on Ruby 2.3

mmoll:
This unbreaks Ubuntu/xenial and Debian/stretch nightlies


dLobatog (commented):
@mmoll Does this not affect rpm nightly too?

mmoll:
AFAICT no, but I don't watch these closely

mmoll:
After seeing this again, nightly RPMs are using Ruby 2.2 now, so no change.

dLobatog (approved):
@mmoll Thanks. Let's merge to allow debs to at least test our nightlies and then we can figure out what to do next if it affects RPM.

PR #4875, opened 2017-09-26
Fixes #21108 - Add host details to custom error

rabajaj0509:


rabajaj0509:
This fix helps the reboot api in foreman discovery print readable error messages : 
https://github.com/theforeman/foreman_discovery/pull/380

lzap:
I don't think this is necessary, see my discovery comment.

PR #4876, opened 2017-09-26
Fixes #21112 - fix loading of the preview host collection

iNecas:
With the previous version, every form show registered new callback,
every one of them triggering a separate ajax request to the
/hosts/preview_host_collections.

### DIFF
--- a/app/assets/javascripts/templates.js
+++ b/app/assets/javascripts/templates.js
@@ -1,0 +1,3 @@
+$(document).on('ContentLoad', function () {
+  tfm.tools.initTypeAheadSelect($('#preview_host_id'));
+});
--- a/app/views/editor/_toolbar.html.erb
+++ b/app/views/editor/_toolbar.html.erb
@@ -14,7 +14,6 @@
     </div>
 
     <% if show_preview %>
-      <%= javascript_tag("$(document).on('ContentLoad', function() { tfm.tools.initTypeAheadSelect($('#preview_host_id')) });"); %>
       <span id="preview_host_selector" style="display:none">
         <%= text_field_tag :preview_host_id, nil,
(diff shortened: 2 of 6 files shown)


adamruzicka (approved):
Tested and the issue is gone.

timogoebel:
Basically, 👍. I have some questions, though.
Does this work with Remote Execution?
We try to not add any more js assets outside of webpack. That's why there is a JS-tag in the first place. Is there another approach to this that doesn't rely on rails' asset pipeline?

iNecas:
The rex update is here https://github.com/theforeman/foreman_remote_execution/pull/275. I tried to find an example of similar registering of `on("ContentLoad")` but all that I've found were done via rails' asset pipeline.

I can imagine having `tfm.templates.registerHooks` or something like that, that we would call from javascript instead of using the `javascript` helper. On the other hand, since this is only for registering and the we don't actually define any new functions inside the new asset, I would not see this as a big issue.

iNecas:
Given it seems we don't have a good answer for now, can we keep it in the current form?

timogoebel (approved):
Fine with me.
@ohadlevy: As you are the React advocate? What do you think?

ohadlevy:
I would prefer to not add more js files in the asset pipeline, as it forces a full page reload (turbolinks).

Why do we need to run this on content load vs when loading the page for the first time?
if just running the function on page load, you wont have multiple registrations.
a poor man solution would be to add a css class that doesn't match the selector so it will not be matched in a second time.

tbrisker:
I agree with @ohadlevy, and while looking into this I happen to have fixed it - so I opened a new PR.

iNecas:
Closing in favor of https://github.com/theforeman/foreman/pull/4950

PR #4878, opened 2017-09-27
Fixes #21119 - set taxonomies in API

ares:
This sets Organization.current and Location.current for API calls the same way we do it in UI. If users has one org available, it defaults to it. When user has more then one org available, the one that will be used can be selected by setting default org for the user. If sessions are used (e.g. hammer auth login) then the selection is persisted during the session.


mbacovsky:
@akofink, any unwanted impact on Katello API/CLI?

mbacovsky:
@ares I did some testing and it fixed the issue for me. However I'm not able to use the admin user properly, especially with no organization to mimic "Any Organization". I'll need to do more testing and compare with unpatched instance.

akofink:
The hammer tests seem to run fine with an updated apipie cache using this PR.

iNecas:
@akofink are the hammer tests run with non-admin user?

akofink:
Sadly, no.

iNecas:
I guess this should not make things worse, as the non-admin user should not be able to load any taxable object without taxonomies set. So this should just make it better. But would be curious about results from trying it out.

mbacovsky:
I did some more testing to reveal the differences. My test env consists from two VMs with fresh Katello nightly build. One is with the patch and one is without.

I created new org, cloned org admin role for the new org and created new user with the new role:
```
hammer organization create --name testorg
hammer role clone --name "Organization admin" --new-name "TestOrg Admin" --organizations testorg
hammer user create --login toadmin --mail "admin@test.org" --auth-source-id 1 --password toadmin --organizations testorg --locations "Default Location" --roles "TestOrg Admin"
```
Here the first difference occurred. On the PATCHED box the new user was in incorrect orgs (notice the Default Organization):
```
hammer user info --login toadmin
Id:                    19
Login:                 toadmin
...
Roles:                 
    Default role
    TestOrg Admin
...
Locations:             
    Default Location
Organizations:         
    Default Organization
    testorg
...
```
I've corrected it to match the UNPATCHED box with `hammer organization remove-user --name "Default Organization" --user toadmin`. Later I found that the `admin` user has the `Default on Login` set to the `Default Organization` which probably was added to the `toadmin`'s orgs. This is unwanted and should not override non-nil api params. Same behavior was seen in non-admin user. 

I've created a first test user as an admin:
```
hammer user create --login touser1 --mail "user1@test.org" --auth-source-id 1 --password touser1 --organizations testorg --locations "Default Location"
```
The PATCHED box had the same issue with extra org again, corrected.

The I created another test user as the new `toadmin` user:
```
hammer -u toadmin -p toadmin user create --login touser2 --mail "user2@test.org" --auth-source-id 1 --password touser2
```
This is the test scenario for this PR and worked well on the PATCHED box. Failed on the clean one (as expected). I've created second test user as an admin there to match the first box.

Listing of the users as the `toadmin` user worked well on both boxes:
```
hammer -u toadmin -p toadmin user list
---|---------|------|----------------|-------|------------|--------------
ID | LOGIN   | NAME | EMAIL          | ADMIN | LAST LOGIN | AUTHORIZED BY
---|---------|------|----------------|-------|------------|--------------
19 | toadmin |      | admin@test.org | no    |            | Internal     
20 | touser1 |      | user1@test.org | no    |            | Internal     
22 | touser2 |      | user2@test.org | no    |            | Internal     
---|---------|------|----------------|-------|------------|--------------
```
*Note: Interestingly the `admin` user is missing from the list while in the UI it is present. But both boxes behaved the same so it is not related to this PR. I'll create new issue for that.*

Trying to do the same list as an admin user failed on the PATCHED box:
```
hammer user list --organization testorg --location "Default Location"
Organization with id 5 not found
```
This is caused again by the `Default On Login` being set. When the option was removed (only possible in the UI or API, not in hammer) the output on the PATCHED box matched the UNPATCHED one. 

It seems `Organization.current` takes precedence over the API params which may lead to unexpected results and needs to be addressed before this is merged

mbacovsky:
Would it make sense and fix the issue to not set <taxonomy>.current when <taxonomy>_id{s} is set in API params? This way it would work more like a default value and probably closer to how UI works.

mbacovsky:
I've found out that the problem was partly solved by fixing `set_taxonomy_scope` [1] which expects Organization.current and Location.current to be unset. With it fixed I'm able to list testorg's users as an admin user and the result matches the UI. 

I'm looking more into whether it fixes other  

[1] https://github.com/theforeman/foreman/blob/698e916ce208b5040b83a908a058c83c94d158ee/app/controllers/concerns/api/taxonomy_scope.rb#L17-L21

mbacovsky:
I've noticed, that the `admin` user was not part of the testorg's users listed because he is not assigned to the org. In UI it is printed in API it is not. Both versions make sense but we should pick one for both interfaces

ares:
Second commit can be isolated so I extracted it as https://github.com/theforeman/foreman/pull/4916
Third one will be covered separately, the redmine issue is http://projects.theforeman.org/issues/21343

ares:
Katello PR opened as https://github.com/Katello/katello/pull/7008, tests with it running as http://ci.theforeman.org/job/test_katello_pull_request/12735/

app/controllers/concerns/application_shared.rb
```diff
@@ -21,4 +21,66 @@ def set_timezone
   def current_permission
     [action_permission, controller_permission].join('_')
   end
+
+  def set_taxonomy
+    user = User.current
+    return if user.nil?
+
+    ['location', 'organization'].each do |taxonomy|
+      return unless Taxonomy.enabled?(taxonomy.to_sym)
```
iNecas: `s/return/next/` ?

app/controllers/concerns/application_shared.rb
```diff
@@ -21,4 +21,66 @@ def set_timezone
...
+
+      available = user.send("my_#{taxonomy.pluralize}")
+      determined_taxonomy = nil
+
+      # this keeps "any context" for admins unless they ask for specific context via parameters
+      if api_request? && user.admin?
```
iNecas: After this change, as an admin, you can't set `any context` in case you have set a default taxonomy. I recommend extracting this method to two: for the ui and for the api behavior, as I believe it would make it more obvious that there is such an issue

iNecas:
What is the limitation of the current fix? What not to expect to work?

app/controllers/concerns/api/taxonomy_scope.rb
```diff
@@ -1,26 +0,0 @@
-module Api
-  module TaxonomyScope
```
iNecas: GIven the wide usage of this module in the plugins, wouldn't make sense to keep it around for a release, just to print a deprecation warning?

ares: yeah I was thinking the same when I opened a PR against REX yesterday... I suppose we should keep it as noop?

iNecas: Yes, noop with deprecation message in `included` block

app/controllers/concerns/application_shared.rb
ares: I suppose by default you mean hammer defaults, not user's default organization. For admins, that should be ignored in this version completely. Org admin always has any context and can set specific context by using --organization-id and --location-id params.

Any context would normally be accessible by `--organization-id nil` but hammer currently can't set nil value. I'm not sure if extracting to a separate method would help, I wanted to keep all the logic on one place since it's "if condition hell" already. Jumping between methods does not add much readability.

EDIT: sry I was confused by the line you added the comment to, I was able to reproduce this in UI, the issue is that we fallback to default taxonomy with every request, while we should do this only on login for UI, defaulting makes sense for API requests

ares:
> What is the limitation of the current fix? What not to expect to work?

The last version keeps UI untouched, if user has default taxonomy set, they are in this context after login, they can still choose any context or specific org/loc. Non-admin user with single org can choose Any organizations but it always redirects him to his only (first) organization.

In API, for admin this works as it did before. Admin always work in any context, their default context is ignored, they can enforce specific context by using `--organization-id` and `--location-id` in hammer (organization_id and location_id params). Then can create resources in any scope by using `--organization-ids 1,3`.

For org admin (or users without view and assign orgs permissions in general) this is a bit different. Such user can't use `--organization-id` since they don't have view permission. For listing resources, the default organization is applied (keeping it to any if user has 2 or more orgs). In such case users can limit specific resorces by using search, e.g.

`hammer domain list --search "organization_id = 1"`

using 

`hammer domain list --organization-id 1`

would not work, that only works for users with view_organizations permissions.

For creation it works similarly. The resource is created in user's context. Since the user can't set context explicitly by using --organization-id, they need to have default organization set. Otherwise creation would fail with "Organization ids Invalid organizations selection, you must select at least one of yours and have 'assign_organizations' permission.". They can't even specify `--organization-ids` for the same reason. That is the limitation for org admins of multiple orgs.

This works assuming current user has default organization set

`hammer domain create --name 'test.com' --location-ids 2`

note that locations can be set if user has view_locations / assign_locations permissions which is the case for org admins. In case the user does not have any default org, the command above fails.

Following command fails even if user is assigned in orgs 1 and 3, this is the limitation, user must not use this parameter and then need to rely on default organization.

`hammer domain create --name 'test.com' --location-ids 2 --organization-ids 1,3`

Ivan, while the patch works now, the method is ugly. I'll try to cover some behavior by integration tests and refactor it a bit. I'd appreciate early feedback on behavior and limitations without waiting on refactoring. Thanks!

app/controllers/concerns/application_shared.rb
```diff
@@ -21,4 +21,77 @@ def set_timezone
...
+          end
+        else
+          determined_taxonomy = nil
+        end
+      # if there was a session set, try to load the taxonomy from there
+      elsif request.session["#{taxonomy}_id"].present?
+        determined_taxonomy = available.where(:id => request.session["#{taxonomy}_id"]).first
+        # warn user if taxonomy stored in session does not exist and delete it from session (probably taxonomy has been deleted meanwhile)
+        if determined_taxonomy.nil?
+          warning _("#{taxonomy.capitalize} you had selected as your context has been deleted") unless api_request?
```
iNecas: There are already 3 conditions that check for `api_request` here. And I would argue there should be 4, as via API, I would expect for this case also to receive `not_found` instead of having a default selected. Please reconsider having two methors: for UI and for API, if nothing else, it would reduce this checks on the `api_request`, which make it quite hard to understand.

ares:
new run with Katello at http://ci.theforeman.org/job/test_katello_pull_request/12750/

ares:
And one more run http://ci.theforeman.org/job/test_katello_pull_request/12751/

iNecas:
There are some tests failing on katello side now :/

iNecas:
The failure on foreman side is unrelated.

ares:
I'll take a look at Katello tests, it looks like the organization used in test is not among available for user.

ares:
I was right, I pushed fixes to affected tests. This patch therefore also fixes Katello issue that was bypassing user's context. Rerun tests together at http://ci.theforeman.org/job/test_katello_pull_request/12755/

app/controllers/concerns/application_shared.rb
```diff
@@ -21,4 +21,82 @@ def set_timezone
...
+      nil
+    end
+  end
+
+  def find_session_taxonomy(taxonomy, user)
+    available = user.send("my_#{taxonomy.pluralize}")
+    determined_taxonomy = available.where(:id => request.session["#{taxonomy}_id"]).first
+    # warn user if taxonomy stored in session does not exist and delete it from session (probably taxonomy has been deleted meanwhile)
+    if determined_taxonomy.nil?
+      warning _("#{taxonomy.capitalize} you had selected as your context has been deleted") unless api_request?
```
iNecas: Translating interpolated string, please use `format` instead

ares: good catch, fixed and tested the warning prints correctly, I also added integration tests for this path which revealed one more issue, keeping it in separate commit so you can see better

ares:
commit message will be squashed, rerun with katello http://ci.theforeman.org/job/test_katello_pull_request/12759/

iNecas (approved):
No other comments from me. I'm approving the PR. Waiting for somebody on Katello side to confirm readiness of merging https://github.com/Katello/katello/pull/7008

mbacovsky:
I did some hammer testing for the following situations:
Admin user:
  - in one org, in two orgs, (both default org set and unset)
  - create, list, info on domain (foreman resource) and product (katello resource)
Everything worked as expected

Org admin user:
  - in one org (both default org set and unset)
 - in two orgs with default organization set
  - create, list, info on domain (foreman resource) and product (katello resource)
Everything worked but `domain list --organization-id <id>` failed with `Organization not found by id '<id>'` which was expected and is limitation of the fix.
 
Common user represented by user with Manager role:
  - in one org (both default org set and unset)
  - create, list, info on domain (foreman resource) and product (katello resource)
Everything worked as expected.

LGTM from me.

Note: I was hitting issue when creating products. The occurrence pattern is not obvious and it seems unrelated to this patch. Bug with same symptoms was identified as a race condition and fixed in 6.2
https://bugzilla.redhat.com/show_bug.cgi?id=1356622. It is probably caused by bulk creation during the automated tests, will investigate it further and report as a separate issue.
Edit: I was able to reproduce the issues on clean Katello, so it is surely not related to this fix

iNecas:
Thanks @ares and @mbacovsky

PR #4879, opened 2017-09-27
Fixes #16363 - fix taxable user deletion

ares:
Reproducing steps:

as admin:
1. create a user and assign him some organization
2. create user group and assign this new user to this user group
3. switch context to this org, do not remain in Any Context where all works
4. try deleting the user
5. see the error

with this fix, the user gets deleted

technical details:
belongs_to respects the default scope which unfortunately is used by taxonomies, therefore when member is being loaded, it fails resulting in member being nil. When it's later passed to `find_all_affected_users_for`, it fails since `nil` is unknown member type. In this case we need to skip taxonomies verification. Sadly we can't use any of common tricks since this is a polymorphic associations.

`belongs_to :member, -> { unscope(:where) }, :polymorphic => true` does not work since it also reset the where defined by polymorphic part, taxonomix does not define any named scope, just `where("id IN ...")`

### DIFF
--- a/app/models/usergroup_member.rb
+++ b/app/models/usergroup_member.rb
@@ -1,5 +1,25 @@
 class UsergroupMember < ApplicationRecord
+  # the belongs to would apply default scope limiting the search by taxonomies but we need to delete
+  # all members, regardless of their taxonomies
+  #
+  # we can't use custom scope definition in belongs_to because it's polymorphic, unscope(:where) would reset
+  # the member_type condition so we need to override how we search for members
+  module OverrideMemberAssociation
+    def member
+      case member_type
+        when 'User'
+          User.unscoped { super }
+        when 'Usergroup'
+          Usergroup.unscoped { super }
+        else
+          raise ArgumentError, "Unknown member type #{member_type}"
+      end
+    end
+  end
+
   belongs_to :member, :polymorphic => true
+  prepend OverrideMemberAssociation
+
   belongs_to :usergroup
 
   before_validation :ensure_no_cycle


app/models/usergroup_member.rb
```diff
@@ -1,5 +1,20 @@
...
+    case member_type
+      when 'User'
+        User.unscoped { member_without_unscoped }
+      when 'Usergroup'
+        Usergroup.unscoped { member_without_unscoped }
+      else
+        raise "Unknown member type #{member_type}"
+    end
+  end
+  alias_method_chain :member, :unscoped
```
mmoll: `alias_method_chain` will be gone in Rails 5.1, see 3896497

tstrachota:
:+1: for both the code and functionality. I tested the patch today and it works nicely.

app/models/usergroup_member.rb
```diff
@@ -1,4 +1,24 @@
...
+    # we can't use custom scope definition in belongs_to because it's polymorphic, unscope(:where) would reset
+    # the member_type condition so we need to override how we search for members
+    def member
+      case member_type
+        when 'User'
+          User.unscoped { super }
+        when 'Usergroup'
+          Usergroup.unscoped { super }
+        else
+          raise "Unknown member type #{member_type}"
```
timogoebel: ... `Foreman::Exception.new()` ?

timogoebel (changes requested):
@ares: Looks good, just one small comment.

app/models/usergroup_member.rb
tstrachota: Good catch. We raise it as `ArgumentError` on two other locations in this file. Whatever exception type we use, we should be consistent.

ares: I agree, I changed the exception to ArgumentError since this is an internal exception for devs, not really for users.

timogoebel (approved):

timogoebel:
@ares: Can you rebase please to fix the test failures? Thanks.

ares:
rebased

mmoll:
test failures related ;)

ares:
ah I accidentally pushed older version from a different computer, should be good now

timogoebel:
@ares: Before I merge this, one final question, just to make sure: Is there a case when `member_type ` might be `nil`?

ares:
no, I'm not aware such case, it's always user or user group

timogoebel:
Merged, thanks @ares and @tstrachota.

I'll put this on 1.17, but feel free to put it in 1.16 at your own discretion.

ares:
Thanks, I think it's fine to keep for 1.17, it's not 1.16 regression.

PR #4880, opened 2017-10-01
Fixes #21162 - Handles any error thrown while connecting via ssh

juliovalcarcel:
Removes unnecessary rescue/debug messages and handles all exceptions the same until the timeout occurs. This ensures that we attempt to connect until the complete timeout occurs.

### DIFF
--- a/app/services/foreman/provision/ssh.rb
+++ b/app/services/foreman/provision/ssh.rb
@@ -77,6 +77,32 @@ class Foreman::Provision::SSH
         Timeout.timeout(8) do
           ssh.run('pwd')
         end
+      rescue Errno::ECONNREFUSED
+        logger.debug "Connection refused for #{address}, retrying"
+        sleep(2)
+        retry
+      rescue Errno::EHOSTUNREACH
+        logger.debug "Host unreachable for #{address}, retrying"
+        sleep(2)
+        retry
+      rescue Net::SSH::Disconnect
+        logger.debug "Host dropping connections for #{address}, retrying"
+        sleep(2)
+        retry
+      rescue Net::SSH::ConnectionTimeout
+        logger.debug "Host timed out for #{address}, retrying"
+        sleep(2)
+        retry
+      rescue Net::SSH::AuthenticationFailed
+        logger.debug "Auth failed for #{username} at #{address}, retrying"
+        sleep(2)
+        retry
+      rescue Timeout::Error
+        retry
+      rescue IOError
+        logger.debug "net-ssh threw an IOError, retrying"
+        sleep(2)
+        retry
       rescue => e
         logger.info "An error occured while connecting before timeout occured \"#{e.inspect}\", retrying"
         logger.debug "Full stacktrace of exception: \n  #{e.backtrace.join("\n  ")}"
--- a/app/services/foreman/provision/ssh.rb
+++ b/app/services/foreman/provision/ssh.rb
@@ -72,21 +72,19 @@ class Foreman::Provision::SSH
   end
 
   def initiate_connection!
-    begin
-      Timeout.timeout(360) do
-        begin
-          Timeout.timeout(8) do
-            ssh.run('pwd')
-          end
-        rescue => e
-          logger.debug "Error occured while connecting \"#{e.inspect}\", retrying"
-          sleep(2)
-          retry
+    Timeout.timeout(360) do
+      begin
+        Timeout.timeout(8) do
+          ssh.run('pwd')
         end
+      rescue => e
+        logger.debug "Error occured while connecting \"#{e.inspect}\", retrying"
+        sleep(2)
+        retry
       end
-    rescue => e
-      Foreman::Logging.exception("Error connecting over SSH, reached max timeout of 360 seconds", e)
     end
+  rescue => e
+    Foreman::Logging.exception("Error connecting over SSH, reached max timeout of 360 seconds", e)
   end
 
   def ssh
--- a/app/services/foreman/provision/ssh.rb
+++ b/app/services/foreman/provision/ssh.rb
@@ -81,14 +81,13 @@ class Foreman::Provision::SSH
         rescue => e
           logger.debug "Error occured while connecting \"#{e.inspect}\", retrying"
           sleep(2)
-          retry 
+          retry
         end
       end
     rescue => e
       Foreman::Logging.exception("Error connecting over SSH, reached max timeout of 360 seconds", e)
     end
   end
-
 
   def ssh
     Fog::SSH.new(address, username, options.merge(:timeout => 4))
--- a/app/services/foreman/provision/ssh.rb
+++ b/app/services/foreman/provision/ssh.rb
@@ -78,13 +78,14 @@ class Foreman::Provision::SSH
           ssh.run('pwd')
         end
       rescue => e
-        logger.debug "Error occured while connecting \"#{e.inspect}\", retrying"
+        logger.info "An error occured while connecting before timeout occured \"#{e.inspect}\", retrying"
+        logger.debug "Full stacktrace of exception: \n  #{e.backtrace.join("\n  ")}"
         sleep(2)
         retry
       end
     end
-  rescue => e
-    Foreman::Logging.exception("Error connecting over SSH, reached max timeout of 360 seconds", e)
+  rescue Timeout::Error
+    Foreman::Logging.exception("Error connecting over SSH, reached max timeout of 360 seconds")
   end
 
   def ssh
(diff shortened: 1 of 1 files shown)


app/services/foreman/provision/ssh.rb
```diff
@@ -72,39 +72,24 @@ def logger
   end
 
   def initiate_connection!
-    Timeout.timeout(360) do
-      begin
-        Timeout.timeout(8) do
-          ssh.run('pwd')
+    begin
```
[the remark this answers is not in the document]
juliovalcarcel: I am not super familar with ruby, but I think this may be a false positive as I need the nested begin/rescue's to handle to two nexted Timeouts. But if I am incorrect let me know and I will gladly fix it.

dLobatog (commented):
[test]
ok to test

app/services/foreman/provision/ssh.rb
```diff
@@ -77,32 +77,14 @@ def initiate_connection!
...
+        logger.debug "Error occured while connecting \"#{e.inspect}\", retrying"
         sleep(2)
         retry
-      rescue Timeout::Error
-        retry
-      rescue => e
-        Foreman::Logging.exception("SSH error", e)
       end
     end
+  rescue => e
```
dLobatog: If it reached the timeout, it should only show `rescue Timeout::Error`.

app/services/foreman/provision/ssh.rb
```diff
@@ -77,32 +77,14 @@ def initiate_connection!
...
-        logger.debug "Host dropping connections for #{address}, retrying"
-        sleep(2)
-        retry
-      rescue Net::SSH::ConnectionTimeout
-        logger.debug "Host timed out for #{address}, retrying"
-        sleep(2)
-        retry
-      rescue Net::SSH::AuthenticationFailed
-        logger.debug "Auth failed for #{username} at #{address}, retrying"
+      rescue => e
```
dLobatog: This will hide other exceptions which may be caused by Foreman itself, the list of exceptions we had here was there in purpose so that other exceptions don't `retry` but fail and the backtrace is shown.

dLobatog (commented):
@juliovalcarcel Thanks for the patch. I've read [21162](http://projects.theforeman.org/issues/21162) but I don't understand how this is fixing it, could you elaborate a bit on how that's happening?

juliovalcarcel:
@dLobatog When creating a new host via the EC2 compute resource it was failing to SSH with a "SSH Error: IOError steam closed" which then caused provisioning to fail entirely when in reality the host it created was able to be SSH to from Foreman with no issue. After more digging I concluded that Foreman attempts to SSH too quickly before the instance is created and provisioned and the default centos user is added with foreman's public key. Looking at /var/log/secure in 22162  you can see that the first attempt to connect is Foreman then we see the centos user being added then (For a reason I am unsure of) the ssh daemon restarts. Doing more digging into net-ssh told me that the net-ssh has limitations that it can't handle when a the ssh server is restarted and will throw an IOError (Which was what was causing my issue). 

This fixes my issue by not only handling the IOError inside the inner `rescue => e` but any other issues that could occur. As a user if the initiate_connection method fails for any reason I always want it to retry until the timeout. It signifies to me that for some reason foreman cannot ssh to the host in 5 mins which means I need to fix something. Even if it failed for any other issue I would always want it to retry inside of the timeout limit. 

I just updated to make the inner log message an info log for easy debugging as the exception that is thrown clearly indicates the issue at play if it was one of the ones we were catching before and I just added a debug log which will print the stacktrace if debug is on, but without throwing an error and failing so in the 1% case it's not one of the set of exceptions identified we can still diagnose it. Also I changed the outer rescue to only handle Timeout::Error

app/services/foreman/provision/ssh.rb
```diff
@@ -77,32 +77,15 @@ def initiate_connection!
...
+        logger.info "An error occured while connecting before timeout occured \"#{e.inspect}\", retrying"
+        logger.debug "Full stacktrace of exception: \n  #{e.backtrace.join("\n  ")}"
         sleep(2)
         retry
-      rescue Timeout::Error
```
iNecas: Do we want to log also the `Timeout::Error`?

juliovalcarcel: By doing rescue e and then logging the exception class in the message we will know if it was a timeout or not

iNecas: Logging the message is fine, but I would discourage logging the backtrace  for this. The problem I see is, we have a set of known exceptions, and we don't need a backtrace from those (including this). Than we have another set of exceptions, that we don't expect: I'm not sure we should retry from there, but more importnantly: we should log the exception with `error ` level from the unknown ones. Therefore, I would be fine with listing the known ones in one rescue block, pretty much the same as you do now (and including the `IOError` + having the `resuce => e` there, with more verbose logging (in `error` level): I think I could live with retry there as well. Other thoughts anyone?

ares: I think retry is acceptable but we need to limit how many times we try retry. +1 to the rest.

iNecas: Number of retries is limited by the outer timeout block

lzap: Never ever eat backtraces, I do not agree with @iNecas here and I would like to see it. If you think that could be too veerbose, just make sure it is logged only once (the first try). In the end, it's debug level (turned off by default).

ares: If the log contains trace of last exception, I'm ok with that. Especially if that's expected exception from which we recover by retry. But I don't insist :-)

iNecas: I think debug trace for known exceotions is ok: what I'm advocating for though is logging unknown exceptions AND backtrace with error log level

lzap: If any exception is not rethrown, it must be logged. Previously we had particular exceptions and it was ok to swallow them, not now when we can miss a generic exception and not have stracktrace in logs.

juliovalcarcel: As a user, what I care about is that I can SSH into my box in this case I don't care what error occurs I just want it to retry for 5 minutes until it gets in. If it does time out its easy enough to enable debug logging and try it again. Theoretically (in this case for ssh) if the issue happens the first time and I turn on debug logs to get the stack trace I will get the same issue again. And in most cases the issue will not be on Foreman's side it will be something else with the server it is SSHing into.

lzap:
Frankly after reading the issue I think the proper solution would be to add IO "closed stream" exception handling and not go with generic.

juliovalcarcel:
I reverted back all but the minimum changes I think should be added. It does a rescue on IOError and then for any unknown exceptions it will still catch them and try again if it hasn't timed out yet while still logging the error.

app/services/foreman/provision/ssh.rb
```diff
@@ -99,10 +99,19 @@ def initiate_connection!
...
       rescue Timeout::Error
         retry
+      rescue IOError
+        logger.debug "net-ssh threw an IOError, retrying"
+        sleep(2)
+        retry
       rescue => e
-        Foreman::Logging.exception("SSH error", e)
+        logger.info "An error occured while connecting before timeout occured \"#{e.inspect}\", retrying"
+        logger.debug "Full stacktrace of exception: \n  #{e.backtrace.join("\n  ")}"
```
lzap, line 108: Please use `Foreman::Logging.exception("An error occured while connecting before timeout occured", e)` instead, this is a helper which formats exceptions properly in logs.

app/services/foreman/provision/ssh.rb
```diff
@@ -99,10 +99,19 @@ def initiate_connection!
         retry
       rescue Timeout::Error
         retry
+      rescue IOError
+        logger.debug "net-ssh threw an IOError, retrying"
```
lzap, line 103: Please do not swallow exceptions, name the exception and properly log it.

    Foreman::Logging.exception("IO error occurred", exception)

app/services/foreman/provision/ssh.rb
```diff
@@ -99,10 +99,19 @@ def initiate_connection!
...
       rescue => e
-        Foreman::Logging.exception("SSH error", e)
+        logger.info "An error occured while connecting before timeout occured \"#{e.inspect}\", retrying"
+        logger.debug "Full stacktrace of exception: \n  #{e.backtrace.join("\n  ")}"
+        sleep(2)
+        retry
       end
     end
+  rescue Timeout::Error
+    Foreman::Logging.exception("Error connecting over SSH, reached max timeout of 360 seconds")
```
lzap, line 114: And here it should be regular `logger.warn` call instead exception.

lzap (changes requested):
We don't want to have exceptions swallowed, except the last warning about giving up.

lzap:
Ping? Interested in getting this into the codebase?

PR #4882, opened 2017-10-02
Fixes #21170 - move password strength meter logic to react

amirfefer:
`password_strength.js` uses jquery_pwstrength_bootstrap wrapper 
I've added `react-password-strength` - react component,  which renders password input with integrated password strength meter by using [zxcvbn](https://github.com/dropbox/zxcvbn) to calculate a password strength score.
In addition, I moved `verify password` logic to wepback (it was  in `password_strength` file as well)
Tests are missing for now.

_after update:_
Try online with [Stroybook](http://rawgit.com/amirfefer/foreman/storybook-my-feature/index.html?selectedKind=Password%20Strength%20Meter&selectedStory=With%20no%20error%20&full=0&down=1&left=1&panelRight=0&downPanel=storybook%2Factions%2Factions-panel) 

_before update:_
oct 3 2017 4-14 pm - edited 1


amirfefer:
@rohoover What do you say about the color scheme?

Rohoover:
I think it jumps from orange to green. Maybe a yellow in between for 

Weak - Red
Normal - Orange
Medium -Yellow
Strong -Green
Very Strong - Dark Green

app/views/users/_form.html.erb
```diff
@@ -41,8 +40,13 @@
         <% if User.current == @user  %>
           <%= password_f f, :current_password, :label => _('Current password'), :placeholder => '' %>
         <% end %>
-        <%= password_f f, :password, :label => _('Password'), :error => @user.errors[:password_hash] %>
-        <%= password_f f, :password_confirmation, :label => _('Verify') %>
+        <div class="clearfix" id="password_react"></div>
```
waldenraines: Is the `.clearfix` needed here?

webpack/assets/javascripts/react_app/components/passwordStrength/index.js
```diff
@@ -0,0 +1,31 @@
...
+import './password_strength.scss';
+
+class PasswordStrength extends React.Component {
+
+  render() {
+    const {className, id, name, error} = this.props.data;
+
+    return (
+      <div className={`form-group ${error.length > 0 ? 'has-error' : ''}`}>
+        <label className="col-md-2 control-label"> { __('Password') } </label>
```
waldenraines: This label should include a `for` attribute to tie it to the input below for accessibility.

webpack/assets/javascripts/react_app/components/passwordStrength/password_strength.scss
```diff
@@ -0,0 +1,10 @@
+.ReactPasswordStrength-input {
+  width: 100% !important;
```
waldenraines: Can we style this without the `!important`s?  Maybe add additional classes or elements to the selector if you need more specificity.

waldenraines (changes requested):

webpack/assets/javascripts/react_app/components/passwordStrength/index.js
ohadlevy: I suggest to use the CommonForm component instead of redefining it here (https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/components/common/forms/CommonForm.js)

webpack/assets/javascripts/react_app/components/passwordStrength/index.js
```diff
@@ -0,0 +1,31 @@
...
+            minLength={0}
+            minScore={2}
+            scoreWords={[__('Weak'), __('Medium'), __('Normal'), __('Strong'), __('Stronger')]}
+            inputProps={{autoComplete: 'new-password', name: name,
+                         id: id, className: className}}
+          />
+        </div>
+        { error.length > 0 &&
+        <span className="help-block help-inline">
+          <span className="error-message"> {error[0]} </span>
```
ohadlevy: could there be more than one error? should we only display the first one?

amirfefer:
@waldenraines  I've mentioned your comments.
@ohadlevy, I've switched verify password validation logic from jquery to react

I've changed color pattern to @Rohoover  suggestion

waldenraines (approved):
Looks good to me!

amirfefer:
@Rohoover 
thoughts about color scheme , text placement ?
oct 19 2017 11-28 pm - edited

ares:
Copying out notes from UX demo with @Rohoover about this:

> Password authorization - Weak/Normal/Strong visualization would not pass 528 compliance for those visually impaired. The colors associated with each strength level seems mismatched. For example: Normal is red. A red status as well as form field highlight indicates an error. Weak is grey, and also possibly not high enough contrast for the visually impaired. The solution may be to simplify the password strength to text only, “Weak Password, Strong Password”. Also - greater thought or discussion could be had around “Normal” or “Strong”. If a “Normal” indicator is shown, does this actually change user behavior in the same way “Weak” does? Perhaps only “weak” is needed.

While I think most of it is addressed and it's a great improvement already, I think there are some other things that could be considered while it's being touched.
* If messages include icons, it would solve the coloring issue.
* fonts seems to be different for pass match and pass strength (italic, different size)
* "too short" should be perhaps "Too short" to be consistent
* does medium password still get accepted?

Also when this gets in, please update/close http://projects.theforeman.org/issues/20740 which tracked this originally :-)

amirfefer:
Thanks @ares 
I've re-styled it a bit:
oct 20 2017 2-39 pm - edited

webpack/assets/javascripts/react_app/components/passwordStrength/index.js
```diff
@@ -0,0 +1,51 @@
...
+    const username = document.getElementById('user_login').value;
+
+    return (
+      <div>
+        <CommonForm label={__('Password')} touched={true} error={error}>
+          <ReactPasswordStrength
+            changeCallback={(data) => {this.setState({password: data.password});}}
+            minLength={6}
+            minScore={2}
+            userInputs={[username]}
```
amirfefer: @ohadlevy, `user_inputs` array adds strings to [zxcvbn](https://github.com/dropbox/zxcvbn) dictionary.  its meant for user inputs from other fields of the form, so this PR  doesn't break username reference.

ohadlevy: this only works if the username is entered prior to the initializing the component, you would need to bind to the form via js not via a react mounter props.

ohadlevy (changes requested):
Please:
1. switch to redux instead of local state
2. bind to the user input in the form instead of receiving it as a prop.
3. add tests :-)
4. would be to vet it with @Rohoover hoover again 

thanks!

webpack/assets/javascripts/react_app/components/passwordStrength/index.js
amirfefer: @ohadlevy  , the username value isn't a prop from react mounter, I'm using the DOM  dynamically when rendering, as you can see in line 25.

ohadlevy: that only reads it when the component renders, it should be an event of when it changes....

Rohoover:
I think the color bar is a big improvement. Also - the text is good because it provides something other than just a color state. I'm a little concerned the the yellow color may not provide enough contrast, but this solution is great overall so I'm not worried if it can't be addressed immediately.

ohadlevy:
@amirfefer can you please rebase?

amirfefer:
@ohadlevy, I've rebased, could you have another look?

ohadlevy:
@sharvit can you please review?

webpack/assets/javascripts/react_app/components/user/passwordStrength/index.js
```diff
@@ -0,0 +1,99 @@
...
+  }
+}
+const mapStateToProps = state => ({
+  password: state.passwordStrength.password.value,
+  matchMessage: state.passwordStrength.password.match,
+});
+
+const mapDispatchToProps = dispatch => bindActionCreators(
+  {
+    changePassword: changePass,
```
sharvit: Can't it just be `changePassword` everywhere?
I would prefer tp name it as `updatePassword`.

webpack/assets/javascripts/react_app/components/user/passwordStrength/index.js
```diff
@@ -0,0 +1,99 @@
...
+      matchMessage,
+    } = this.props;
+
+    return (
+      <div>
+        <CommonForm label={__('Password')} touched={true} error={error}>
+          <ReactPasswordStrength
+            changeCallback={(data) => {
+              changePassword(data.password);
+            }}
```
sharvit: Can be shorter
```js
changeCallback={({ password }) => changePassword(password)}
```

webpack/assets/javascripts/react_app/components/user/passwordStrength/index.js
```diff
@@ -0,0 +1,99 @@
...
+import {
+  changePass,
+  passwordMatch,
+} from '../../../redux/actions/user/passwordStrength';
+import './password_strength.scss';
+
+class PasswordStrength extends React.Component {
+  constructor(props) {
+    super(props);
+    this.checkRetype = this.checkRetype.bind(this);
```
sharvit: Should use the helper here
https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/common/helpers.js#L2-L7

webpack/assets/javascripts/react_app/components/user/passwordStrength/index.js
```diff
@@ -0,0 +1,99 @@
...
+    this.state = {
+      password: '',
+      error: props.data.verify.error,
+    };
```
sharvit: I don't see where you are using this state,
maybe it should be remove because you mapping redux state to props?

webpack/assets/javascripts/react_app/components/user/passwordStrength/index.js
```diff
@@ -0,0 +1,99 @@
...
+            changeCallback={(data) => {
+              changePassword(data.password);
+            }}
+            minLength={6}
+            minScore={2}
+            userInputs={
+              userInputIds.length > 0
+                ? userInputIds.map(input => document.getElementById(input).value)
+                : []
+            }
```
sharvit, line 41: I am not sure i understand the meaning of that, can you explain please?

Also, haven't found this `#user_login` all over the project, what is it exactly?

```js
:userInputIds => ['user_login'],
```

webpack/assets/javascripts/react_app/components/user/passwordStrength/index.js
```diff
@@ -0,0 +1,99 @@
...
+      error: props.data.verify.error,
+    };
+  }
+  checkRetype(event) {
+    if (event.target.value !== this.props.password) {
+      this.props.passwordMatch(false);
+    } else {
+      this.props.passwordMatch(true);
+    }
+  }
```
sharvit: I don't think it should be a part of the component to decide if the passwords are match.

What do you think about:
```js
// actions
updatePassword, updatePasswordConfirmation

// state
{ password: '', passwordConfirmation: '', match: true }

// reducer
// { payload: { password: 'pass' } } or
// { payload: { passwordConfirmation: 'pass '} }
case PASSWORD_STRENGTH_PASSWROD_UPDATED:
case PASSWORD_STRENGTH_PASSWROD_CONFIRMATION_UPDATED:
  state = state.merge(payload);
  return state.set('match', state.password === state.passwordConfirmation);
```

webpack/assets/javascripts/react_app/components/user/passwordStrength/password_strength.scss
```diff
@@ -0,0 +1,75 @@
...
+}
+
+#password_react {
+  .form-group.has-error .ReactPasswordStrength-input {
+    border: 1px solid #c00;
+  }
+
+  .is-strength-0 {
+    .ReactPasswordStrength-strength-desc {
+      color: #a30000;
```
sharvit: Can you use variables for those colors?

webpack/assets/javascripts/react_app/components/user/passwordStrength/password_strength.scss
```diff
@@ -0,0 +1,75 @@
+.ReactPasswordStrength {
+  .ReactPasswordStrength-input {
+    width: 100%;
+    font-size: 12px;
+    font-family: 'Open Sans', Helvetica, Arial, sans-serif;
```
sharvit, line 7: We really need to change to font here? Isn't it the default font?

webpack/assets/javascripts/react_app/redux/consts.js
```diff
@@ -24,3 +24,5 @@ export const NOTIFICATIONS_MARK_GROUP_AS_READ =
   'NOTIFICATIONS_MARK_GROUP_AS_READ';
 export const NOTIFICATIONS_POLLING_STARTED =
   'NOTIFICATIONS_POLLING_STARTED';
+export const PASSWROD_CHANGED = 'PASSWROD_CHANGED';
+export const PASSWORD_MATCHED = 'PASSWORD_MATCHED';
```
sharvit: I think it is better to use here a more contextual names like `PASSWORD_STRENGTH_PASSWROD_CHANGED`

webpack/stories/index.js
```diff
@@ -1,3 +1,4 @@
+
```
sharvit: Just remove this line, there is no updates for this file...

webpack/assets/javascripts/react_app/components/user/passwordStrength/index.js
```diff
@@ -0,0 +1,99 @@
...
+            tooShortWord={__('Too short')}
+            scoreWords={[
+              __('Weak'),
+              __('Medium'),
+              __('Normal'),
+              __('Strong'),
+              __('Very strong'),
+            ]}
+            inputProps={{
+              autoComplete: 'new-password',
```
sharvit: What is this `autoComplete: 'new-password'`?
Can you please explain?

sharvit (changes requested):

webpack/assets/javascripts/react_app/components/user/passwordStrength/password_strength.scss
amirfefer: The component default font is different from foreman default

webpack/assets/javascripts/react_app/components/user/passwordStrength/index.js
amirfefer: Array of input ids to add to `zxcvbn's` dictionary.
This way, the password score is decreased if it composite from other inputs such as username.

amirfefer:
@sharvit Thanks for your review. could you have a second look?

sharvit:
@amirfefer nice work!

1. I think that your actions should be `updatePassword` and `updatePasswordConfirmation` because this is what the user can actually do.
2. Since there is no inner state, your component can be a dummy/functional component.

amirfefer:
@sharvit  Thanks again !
about making it a functional component,  well, we'll lose react life cycle rendering and because we're using redux here, i think it better to keep it as a class component.
The other way it to split it to two components: a stateless functional component, and a container component which connect to redux. imo those changes are redundant.

amirfefer:
@ohadlevy, @waldenraines , @sharvit 
Could we merge this?

sharvit (approved):
Good to me! 👍

ohadlevy:
@amirfefer please rebase.

webpack/assets/javascripts/react_app/components/user/passwordStrength/passwordStrength.test.js
```diff
@@ -0,0 +1,62 @@
...
+import thunk from 'redux-thunk';
+import configureMockStore from 'redux-mock-store';
+import PasswordStrength from './';
+import {
+  initialState,
+  passwordMatch,
+  passwordNotMatched,
+} from './passwordStrength.fixtures';
+
+configure({ adapter: new Adapter() });
```
ohadlevy: these setup are no longer need (see webpack/test_setup.js)

webpack/assets/javascripts/react_app/components/user/passwordStrength/passwordStrength.test.js
```diff
@@ -0,0 +1,62 @@
...
+
+jest.unmock('./');
+global.__ = jest.fn();
```
ohadlevy: not required anymore

webpack/assets/javascripts/react_app/components/user/passwordStrength/passwordStrength.test.js
```diff
@@ -0,0 +1,62 @@
...
+const mockStore = configureMockStore([thunk]);
+
+function setup(verify, error, state) {
+  const store = mockStore(state);
+
+  return mount(<PasswordStrength
```
ohadlevy, line 18: would shallow work here too?

webpack/assets/javascripts/react_app/components/user/passwordStrength/index.js
```diff
@@ -0,0 +1,88 @@
...
+              __('Weak'),
+              __('Medium'),
+              __('Normal'),
+              __('Strong'),
+              __('Very strong'),
+            ]}
+            inputProps={{
+              name,
+              id,
+              className: `${className}`,
```
ohadlevy: do you really need the string interrelation here? I think the following would just work:
```js
inputProps={{  name,  id, className }}
```

webpack/assets/javascripts/react_app/components/user/passwordStrength/index.js
```diff
@@ -0,0 +1,88 @@
...
+            }}
+          />
+        </CommonForm>
+        {verify && (
+          <CommonForm
+            label={__('Verify')}
+            touched={true}
+            error={matchMessage ? verify.error : __('Password do not matched')}
```
ohadlevy: The original string was ```passwords do not match```

ohadlevy (changes requested):
Initial review looks good - I've added a few comments inline.

Please also adds action/reducers tests.

Thanks!

webpack/assets/javascripts/react_app/components/user/passwordStrength/passwordStrength.test.js
amirfefer: shallow doesn't render inner components therefore the resulted snapshot lacks in inner content

amirfefer:
@ohadlevy  thanks, I've fixed and rebased.  would you mind take another look?

mmoll:
@amirfefer Is the test failure in the foreman test related to these changes?

ares:
ping @amirfefer

amirfefer:
I've fixed the users integration test, 
Apparently the react component did not render at all via capybara.

mmoll:
@amirfefer please rebase

I guess the travis test "failures" also need fixing/updating...

amirfefer:
Rebased.
plus I've updated jest snapshot.

ohadlevy:
@amirfefer anything left here?

webpack/assets/javascripts/react_app/common/colors.scss
```diff
@@ -1 +1,7 @@
 $pf-black-200: #ededed;
+$pf-red-200: #a30000;
```
ohadlevy, line 2: can we use this from the patternfly package? @sharvit ?

sharvit: When my css pr will merge you would be able to import variables:
https://github.com/theforeman/foreman/pull/5212/files#diff-f430da9fad5e785cb1f815fd719fdc5aR1

ohadlevy:
@amirfefer can you please also include
```difff
diff --git a/config/webpack.vendor.js b/config/webpack.vendor.js
index aed3a4bde..e6e2f023a 100644
--- a/config/webpack.vendor.js
+++ b/config/webpack.vendor.js
@@ -9,6 +9,7 @@ module.exports = [
   'react-ellipsis-with-tooltip',
   'react-numeric-input',
   'react-onclickoutside',
+  'react-password-strength',
   'patternfly-react',
   'react-redux',
   'redux',
```

amirfefer:
@ohadlevy - done :)

ohadlevy (approved):

amirfefer:
[test foreman]

sharvit (approved):
LGTM!

amirfefer:
@ohadlevy tests are green :)

PR #4883, opened 2017-10-02
Fixes #21172 - load template variables to partition tables

iNecas:


app/models/host/managed.rb
```diff
@@ -314,12 +314,20 @@ def handle_realm
     end
   end
 
+  def disk_layout_template
+    if disk.present?
+      { name: 'Custom disk layout', template: disk }
+    elsif ptable.present?
+      { name: ptable.name, template: ptable.layout }
+    end
```
tbrisker, line 341: Not very significant but maybe cache this hash to prevent multiple object allocations on each call to `diskLayout` etc?

lib/foreman/renderer.rb
```diff
@@ -274,7 +274,7 @@ def jumpstart_attributes
     end
 
     def kickstart_attributes
-      @dynamic   = @host.diskLayout =~ /^#Dynamic/ if (@host.respond_to?(:disk) && @host.disk.present?) || @host.ptable.present?
+      @dynamic   = @host.disk_layout_template[:template] =~ /^#Dynamic/ if @host.disk_layout_template
```
tbrisker: `.start_with('#Dynamic')` is faster and more readable then regex matching.

app/models/host/managed.rb
```diff
@@ -314,12 +314,20 @@ def handle_realm
...
+  end
+
   # returns the host correct disk layout, custom or common
   def diskLayout
+    raise 'Neither disk nor partition table defined for host' unless disk_layout_template
```
tbrisker: Shouldn't we raise a `ForemanException` (or some other meaningful exception) rather then a `RuntimeError` in these cases?
Also, do we `rescue` these exceptions anywhere?

tbrisker (changes requested):
See a few inline comments.

mmoll:
what's the status here?

app/models/host/managed.rb
iNecas: I would prefer not to cache until proved worth it.

lib/foreman/renderer.rb
iNecas: updated

app/models/host/managed.rb
iNecas: I suppose the exception is capture when rendering the provisioning template (similarly as other exceptions are), What about `Foreman::Renderer::RenderingError`?

iNecas:
Rebased and addressed the comments

iNecas:
Rebased and tests are green. @tbrisker mind looking again at this long-opened PR?

tbrisker:
@timogoebel  - care to take a look in the context of the renderer refactoring?

tbrisker (dismissed):
code makes sense, did not test

app/models/hostgroup.rb
```diff
@@ -113,8 +113,15 @@ def self.title_name
...
 
+  def disk_layout_template
+    if ptable.present?
+      { name: ptable.name, template: ptable.layout }
+    end
+  end
+
   def diskLayout
-    ptable.layout.tr("\r", '')
+    raise 'Partition table not defined for hostgroup' unless disk_layout_template
```
timogoebel: Should this raise a `Foreman::Renderer::RenderingError`?

timogoebel (changes requested):
I think we should raise consistent exceptions. Happy to rebase the template refactoring PR on top of this.

How is the circular dependency resolved here?

No offense, but this isn't pretty. ;-)

app/models/hostgroup.rb
iNecas: Updated

iNecas:
This patch should have nothing to do with circular dependency: the only thing it does (as far as I remember) is making sure it loads the same variables as the the provisioning template.

timogoebel:
I don't remember what the exact problem was, but I believe `load_template_vars` would load `@ diskLayout` and that would render the partition table template. That again will load `@diskLayout` and so on and so forth.

@kamils-iRonin: Do you remember?

kamils-iRonin:
We had a problem with the loading scope variables in the `diskLayout` method and  circular dependency, but I don't think it is related to this PR.

timogoebel (approved):
I'm not too happy with raising template errors in host or hostgroup model, but I believe we can easily refactor this when we introduce template scopes.

tbrisker:
[test katello]

app/models/host/managed.rb
tbrisker: Now that I read the code again I (think I) understand why I added this comment a year ago - `diskLayout` calls this function 3 times, so each call this function is executed again, instead of saving the result in a variable in that function. That said, I don't think this is crucial and won't block on it.

tbrisker (approved):
Took me a bit to understand my old comment, but I'm fine with ignoring it. Thanks @iNecas for coming back to this after such a long time!

PR #4887, opened 2017-10-03
Fixes #21179: Upgrade to enzyme@3.0.0

danseethaler:
Enzyme needs to be upgraded to support react@16. In upgrading I
found that using jest's snapshot feature would greatly improve
several of the tests that were in place. I've changed these
tests to use snapshots and added enzyme-to-json to support
snapshot testing with enzyme wrappers. I've also added 
enzyme-adapter-react-15 as a required adapter for react
in enzyme@3.0.0. This will be changed to enzyme-adapter-react-16
once we upgrade to react@16.

http://projects.theforeman.org/issues/21179

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,9 @@
     "compression-webpack-plugin": "~0.3.1",
     "css-loader": "^0.23.1",
     "dotenv": "^2.0.0",
-    "enzyme": "^2.4.1",
+    "enzyme": "^3.0.0",
+    "enzyme-adapter-react-15": "^1.0.0",
+    "enzyme-to-json": "^3.0.1",
     "eslint": "^3.3.1",
     "eslint-plugin-react": "^6.1.2",
     "expose-loader": "~0.6.0",
@@ -80,6 +82,7 @@
     "lint": "./node_modules/.bin/eslint -c .eslintrc webpack/ || exit 0",
     "test": "node node_modules/.bin/jest",
     "test:watch": "node node_modules/.bin/jest --watchAll",
+    "test:current": "node node_modules/.bin/jest --watch",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
     "deploy-storybook": "storybook-to-ghpages"
--- a/webpack/assets/javascripts/react_app/components/hosts/storage/vmware/StorageContainer.test.js
+++ b/webpack/assets/javascripts/react_app/components/hosts/storage/vmware/StorageContainer.test.js
@@ -1,3 +1,8 @@
+// Configure Enzyme
+import { configure } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-15';
+configure({ adapter: new Adapter() });
+
 import React from 'react';
 import { mount } from 'enzyme';
 import { vmwareData, hiddenFieldValue } from './StorageContainer.fixtures';
@@ -27,7 +32,7 @@ describe('StorageContainer', () => {
       wrapper.render().find('.disk-container').length
     ).toEqual(1);
 
-    wrapper.find('.btn-add-disk').simulate('click');
+    wrapper.find('button.btn-add-disk').simulate('click');
 
     expect(
       wrapper.render().find('.disk-container').length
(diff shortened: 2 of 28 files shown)


danseethaler:
I went through all the `.test.js` files and added the needed [enzyme configuration](http://airbnb.io/enzyme/docs/guides/migration-from-2-to-3.html#configuring-your-adapter). After adding the configuration I added snapshots for the tests where it made sense.

I tweaked a few tests to fix enzyme@3.0.0 breaking changes explained [here](http://airbnb.io/enzyme/docs/guides/migration-from-2-to-3.html#element-referential-identity-is-no-longer-preserved) and [here](http://airbnb.io/enzyme/docs/guides/migration-from-2-to-3.html#children-now-has-slightly-different-meaning).

package.json
```diff
@@ -22,7 +22,9 @@
     "compression-webpack-plugin": "~0.3.1",
     "css-loader": "^0.23.1",
     "dotenv": "^2.0.0",
-    "enzyme": "^2.4.1",
+    "enzyme": "^3.0.0",
+    "enzyme-adapter-react-15": "^1.0.0",
```
ohadlevy, line 26: I assume that once we upgrade react we just need to remove the adapter?

webpack/assets/javascripts/react_app/API.js
```diff
@@ -18,7 +18,7 @@ export default {
       data: data,
       error: function (jqXHR, textStatus, errorThrown) {
         /* eslint-disable no-console */
-        console.log(jqXHR);
+        // console.log(jqXHR);
```
ohadlevy: ?

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/StorageContainer.test.js
```diff
@@ -27,7 +32,7 @@ describe('StorageContainer', () => {
       wrapper.render().find('.disk-container').length
     ).toEqual(1);
 
-    wrapper.find('.btn-add-disk').simulate('click');
+    wrapper.find('.btn-add-disk').at(0).at(0).simulate('click');
```
ohadlevy: why do these change? is it a wrong selector?

ohadlevy (commented):
looks good, please have a look at the test failures, minor comments inline

package.json
danseethaler: Yeah actually we'll upgrade it to enzyme-adapter-react-16.

webpack/assets/javascripts/react_app/API.js
danseethaler: I'll uncomment this - not sure what the comments are for. They were obscuring my test results.

webpack/assets/javascripts/react_app/components/hosts/storage/vmware/StorageContainer.test.js
danseethaler: Yeah, the old api only returned [one result](http://airbnb.io/enzyme/docs/guides/migration-from-2-to-3.html#element-referential-identity-is-no-longer-preserved) when a `.find` matched multiple elements. `wrapper.find('.btn-add-disk')` now returns both elements that match:

```jsx
    <Button className="btn-add-disk" disabled={false} onClick={[Function]} active={false} block={false} bsStyle="default" bsClass="btn">
      <button disabled={false} onClick={[Function]} type="button" className="btn-add-disk btn btn-default">
        Add volume
      </button>
    </Button>
    
    
    <button disabled={false} onClick={[Function]} type="button" className="btn-add-disk btn btn-default">
      Add volume
    </button>
```

I'm not sure if the outer wrapper and the inner button are both supposed to have the `btn-add-disk` class. If so maybe we could use a different selector. Thoughts?

danseethaler: I changed the selector to `button.btn-add-disk` and that seems to do the trick.

waldenraines (approved):
This all looks fine to me.

ohadlevy:
Thanks!

PR #4890, opened 2017-10-04
Fixes #11825 - use before_create for template associations

iNecas:
We hit an issue in foreman_remote_execution, when trying to add
has_many association to taxonomies, after the after_create defined,
due to the behavior described in
https://github.com/rails/rails/issues/12180#issuecomment-29543330

### DIFF
--- a/app/models/taxonomy.rb
+++ b/app/models/taxonomy.rb
@@ -8,7 +8,10 @@ class Taxonomy < ApplicationRecord
   serialize :ignore_types, Array
 
   belongs_to :user
+
+  before_create :assign_default_templates
   after_create :assign_taxonomy_to_user
+  before_validation :sanitize_ignored_types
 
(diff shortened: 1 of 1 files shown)


iNecas:
This was originally opened as https://github.com/theforeman/foreman/pull/2723, but closed for inactivity. In the mean time, we were bit by this again, as we were removing our old workaround from rex that was causing rails 5 incompatibilities.

The original PR was stuck on request for writing a test for this: I'm not too keen on doing this, as it would mean I would need to create a new model just for tests and assign it to a taxonomy. What I did however, was moving the validation call before the association definition, which means that the `after_create` would behave the same way for core as it does for plugins.

iNecas:
bump

tbrisker (approved):

tbrisker:
Thanks @iNecas !

PR #4891, opened 2017-10-05
Fixed #21204 - introduced global_setting helper function

lzap:
Locked templates made very difficult to change to discovery in global default templates, user needs to unlock 2+2+2 (PXELinux, Grub, Grub2 template + discovery snippet) and make necessary changes just to change the default menu entry. This can be done via global setting.

This patch adds such a helper function, I am doing changes in community templates to leverage this new function.

Security standpoint - I can't think of any setting that should not be rendered. We do have a root password, but that is already available using the existing helper. Please review with that in mind.

### DIFF
--- a/app/models/setting/provisioning.rb
+++ b/app/models/setting/provisioning.rb
@@ -59,6 +59,8 @@ class Setting::Provisioning < Setting
       self.set('dns_conflict_timeout', N_("Timeout for DNS conflict validation (in seconds)"), 3, N_('DNS conflict timeout')),
       self.set('clean_up_failed_deployment', N_("Foreman will delete virtual machine if provisioning script ends with non zero exit code"), true, N_('Clean up failed deployment')),
       self.set('name_generator_type', N_("Random gives unique names, MAC-based are longer but stable (and only works with bare-metal)"), 'Random-based', N_("Type of name generator"), nil, {:collection => Proc.new {NameGenerator::GENERATOR_TYPES} }),
+      self.set('default_pxe_item_global', N_("Default PXE (PXELinux, PXEGrub, PXEGrub2) menu item in global template - 'local', 'discovery' or custom"), 'local', N_("Default PXE global template entry")),
+      self.set('default_pxe_item_local', N_("Default PXE (PXELinux, PXEGrub2) menu item in local template - 'local_chain_hd0', 'local' or custom"), 'local_chain_hd0', N_("Default PXE local template entry")),
       self.set(
         'excluded_facts',
         N_("Exclude pattern for all types of imported facts (rhsm, puppet e.t.c.). Those facts won't be stored in foreman's database. You can use * wildcard to match names with indexes e.g. macvtap*"),
--- a/lib/foreman/renderer.rb
+++ b/lib/foreman/renderer.rb
@@ -8,22 +8,66 @@ module Foreman
     class HostUnknown < RenderingError; end
     class HostParamUndefined < RenderingError; end
     class HostENCParamUndefined < RenderingError; end
+    class FilteredGlobalSettingAccessed < RenderingError; end
 
     include ::Foreman::ForemanUrlRenderer
 
-    ALLOWED_GENERIC_HELPERS ||= [ :foreman_url, :snippet, :snippets, :snippet_if_exists, :indent, :foreman_server_fqdn,
-                                  :foreman_server_url, :log_debug, :log_info, :log_warn, :log_error, :log_fatal, :template_name, :dns_lookup,
-                                  :pxe_kernel_options, :save_to_file, :subnet_param, :subnet_has_param? ]
-    ALLOWED_HOST_HELPERS ||= [ :grub_pass, :ks_console, :root_pass,
-                               :media_path, :match,
-                               :host_param_true?, :host_param_false?,
-                               :host_param, :host_param!, :host_puppet_classes, :host_enc ]
+    ALLOWED_GENERIC_HELPERS ||= [
+      :foreman_url,
+      :snippet, :snippets,
+      :snippet_if_exists,
+      :indent,
+      :foreman_server_fqdn, :foreman_server_url,
+      :log_debug, :log_info, :log_warn, :log_error, :log_fatal,
+      :template_name,
+      :dns_lookup,
+      :pxe_kernel_options,
+      :save_to_file,
+      :subnet_param, :subnet_has_param?,
+      :global_setting
+    ]
+    ALLOWED_HOST_HELPERS ||= [
+      :grub_pass,
+      :ks_console,
+      :root_pass,
+      :media_path,
+      :match,
+      :host_param_true?, :host_param_false?,
+      :host_param, :host_param!,
+      :host_puppet_classes,
+      :host_enc,
+    ]
 
     ALLOWED_HELPERS ||= ALLOWED_GENERIC_HELPERS + ALLOWED_HOST_HELPERS
 
     ALLOWED_VARIABLES ||= [ :arch, :host, :osver, :mediapath, :mediaserver, :static,
                             :repos, :dynamic, :kernel, :initrd, :xen,
                             :preseed_server, :preseed_path, :provisioning_type, :template_name ]
+
+    ALLOWED_GLOBAL_SETTINGS ||= [
+      :administrator,
+      :proxy_request_timeout,
+      :http_proxy,
+      :http_proxy_except_list,
+      :email_reply_address,
+      :safemode_render,
+      :manage_puppetca,
+      :ignored_interface_identifiers,
+      :remote_addr,
+      :token_duration,
+      :dns_conflict_timeout,
+      :name_generator_type,
+      :default_pxe_item_global,
+      :default_pxe_item_local,
+      :puppet_interval,
+      :outofsync_interval,
+      :default_puppet_environment,
+      :modulepath,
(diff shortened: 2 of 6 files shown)


lzap:
Also added "default_pxe_item_local" so both global and local menu items can be now configured.

lib/foreman/renderer.rb
```diff
@@ -233,6 +252,10 @@ def load_template_vars
       @template_url = params['url']
     end
 
+    def global_setting(name, default = nil)
+      SETTINGS[name.to_sym] || default
```
ares: I believe this was rejected in past because setting contain sensitive information, such as root password for newly provisioned hosts. I suppose we'd need to add a whitelist here. Also the || won't work for settings that has value `false`, it would always return default value. If you're going to fix this, I'd also suggest renaming `global_setting` to `setting`, there's no other setting...

ares: Also note that settings page is only available for admins, there's no permission for this. Making it available through template give ability to everyone who can render (preview) template, to see values.

lzap: So how about renaming it to "setting", creating whitelist, fixing default value and allowing override via host params, so people can still override this per host/hostgroup.

lzap: Or maybe is cleaner just to use host/hostgroup parameter for this. That would drop this complexity. But to be honest I really like using settings in templates.

lzap: The reason for that is this enables users (or admins specifically) to make discovery the default options. This can't be done with host/hostgroup variables.

ares: If that is for admins only, then I think we can drop the host/hg params. Whitelist, default value and renaming is still required I think. I think the whitelist should contain nearly all settings and (sorry to say that), there should be a plugin DSL for extending it. I suppose discovery plugin would require it.

lzap:
Added whitelist, covered everything with tests, added default list of allowed global options which I think are useful and safe. I did not add those which I thought are not much useful. No host/hostgroup override available, I think this is cleaner.

lzap:
@ares this is ready for review and merge

iNecas (approved):

lib/foreman/renderer.rb
```diff
@@ -233,6 +277,11 @@ def load_template_vars
       @template_url = params['url']
     end
 
+    def global_setting(name, default = nil)
+      raise(FilteredGlobalSettingAccessed, _('Global setting "%s" is not accessible in safe-mode') % name) if Setting[:safemode_render] && !ALLOWED_GLOBAL_SETTINGS.include?(name.to_sym)
+      SETTINGS[name.to_sym] || default
```
iNecas: You should use `Setting`, not `SETTINGS`, the `SETTINGS` are just for things that come from settings file

iNecas: Also, you should check on `blank?`, instead of false value. Otherwise, `boolean` values, such as  `safemode_render`, would not work.

iNecas (changes requested):
Note for me: don't approve PRs without testing, even when there are unit tests that seem to prove it works.

See comments inline.

lzap:
Done, rebased.

dLobatog (dismissed):
Needs to be rebased, please @lzap

lzap:
Rebased.

lib/foreman/renderer.rb
```diff
@@ -232,6 +275,11 @@ def load_template_vars
       @template_url = params['url']
     end
 
+    def global_setting(name, default = nil)
+      raise(FilteredGlobalSettingAccessed, _('Global setting "%s" is not accessible in safe-mode') % name) if Setting[:safemode_render] && !ALLOWED_GLOBAL_SETTINGS.include?(name.to_sym)
+      Setting[name.to_sym].blank?  ? default : Setting[name.to_sym]
```
ares: extra space between questionmarks

app/models/setting/provisioning.rb
```diff
@@ -59,6 +59,8 @@ def self.default_settings
       self.set('dns_conflict_timeout', N_("Timeout for DNS conflict validation (in seconds)"), 3, N_('DNS conflict timeout')),
       self.set('clean_up_failed_deployment', N_("Foreman will delete virtual machine if provisioning script ends with non zero exit code"), true, N_('Clean up failed deployment')),
       self.set('name_generator_type', N_("Random gives unique names, MAC-based are longer but stable (and only works with bare-metal)"), 'Random-based', N_("Type of name generator"), nil, {:collection => Proc.new {NameGenerator::GENERATOR_TYPES} }),
+      self.set('default_pxe_item_global', N_("Default PXE (PXELinux, PXEGrub, PXEGrub2) menu item in global template - 'local', 'discovery' or custom"), 'local', N_("Default PXE global template entry")),
```
ares, line 62: These two settings seems to have limited number of options, could you define the list as the last param? that would in UI render select box instead of free form text. For an example, see one line above.

ares (changes requested):
I'll start testing now, the code looks good except for two small things, please take a look.

app/models/setting/provisioning.rb
```diff
@@ -59,6 +59,8 @@ def self.default_settings
...
+      self.set('default_pxe_item_local', N_("Default PXE (PXELinux, PXEGrub2) menu item in local template - 'local_chain_hd0', 'local' or custom"), 'local_chain_hd0', N_("Default PXE local template entry")),
```
ares, line 63: how custom is supposed to work? looking at [here](https://github.com/theforeman/community-templates/pull/430/files#diff-3413ebe6ecab1f605ee7e4ffa84c9759) I'm unable to specify any other value that local/discovery or is that the limitation for grub1?

if custom values are desired, I guess my comment above about limiting the option does not make sense...

lib/foreman/renderer.rb
```diff
@@ -232,6 +275,11 @@ def load_template_vars
       @template_url = params['url']
     end
 
+    def global_setting(name, default = nil)
```
ares: what's the point of having `default` if setting already provide default? when 2 defaults are useful?

app/models/setting/provisioning.rb
ares: feel free to ignore this one, the more I played with it, the more sense it made

app/models/setting/provisioning.rb
lzap: Grub1 only supports numbers for that option, so there is a small helper array that is then converted to number. I assume that users will sync the array in cloned templates, or simply delete whole thing which does not make much sense in read/write (cloned) template.

I wanted consistent naming in settings across all loaders.

lib/foreman/renderer.rb
lzap: The default of default (the option :-) ) is used in templates to have sane default if users does not opt-in and leaves it blank (that's what we ship Foreman with). What do you recommend instead?

lib/foreman/renderer.rb
lzap: This is the only change.

lzap:
Rebased, one line changed.

iNecas (approved):
No other comments from me

lib/foreman/renderer.rb
ares: I still don't get it, if I use `global_settings(:some, 'value')`, I'll get value back if there's no `Setting[:some]`. But in that case, why would I call the `global_settings` in first place? I'd just use `value`. Or I'd do `global_settigs(:root_pass) || 'change_me'` instead of `global_settigns(:root_pass, 'change_me')`

lzap:
Thanks @ares @dLobatog are we good to go?

ares:
@lzap I'd still like to understand the value of default parameter of `global_setting` but then it's good from my side

lzap:
While I don't insist on the default option for the helper, I think it might be useful - it's a generic helper after all which can be used for more things. The helper-default will be used when setting is `blank` meaning that template writers can still provide some default option when user decided to delete the value from settings for some reason. This enables you to add "third state", you can't set `nil` Ruby value for example via settings, this allows you to create an option "and set to blank for XYZ" and then you can use this mechanism appropriately to set it to `nil` or any other value - for example an empty array or hash. This can make template code easier to read for example.

So it makes sense, not in this context, but for the future.

lib/foreman/renderer.rb
```diff
@@ -232,6 +275,11 @@ def load_template_vars
...
+      raise(FilteredGlobalSettingAccessed, _('Global setting "%s" is not accessible in safe-mode') % name) if Setting[:safemode_render] && !ALLOWED_GLOBAL_SETTINGS.include?(name.to_sym)
+      Setting[name.to_sym].blank? ? default : Setting[name.to_sym]
```
ares: this does not work well with false (valid value) which is considered blank, I guess the behavior should differ if settings_type is boolean

ares:
Thanks @lzap, I'm fine with default value now. During testing I found it does not work well with boolean though. After fixing that, thumbs up.

lzap:
Amended the boolean thing, rebased.

ares (approved):
Thanks @lzap, works fine now.

lzap:
Thanks, @ares or @dLobatog mind merging https://github.com/theforeman/theforeman.org/pull/1013 discovery docs, so I can carry on with PR that documents this new feature? Thanks.

PR #4892, opened 2017-10-05
Refs #21005 - Allow select_action_button to set button as primary

bastilian:
For #21005 it is required the dropdown button is set as primary, this is a small refactor and adds the option `primary`. Example: 
gnome-shell-screenshot-8hsl7y

Also see: https://github.com/theforeman/foreman_ansible/pull/101


dLobatog (approved):
Thanks @bastilian !

PR #4893, opened 2017-10-05
Fixes #21206 - Fix templates taxonomies migration

iNecas:
In 0969b62, we added a migration to change the type of taxable
taxonomies. Unfortunatelly, it was not working right
as migratios don't run with any user and Ptable.all returned
empty set. We need to make sure to unscope when loading the data.


ares:
good catch, the commit for the reference is 0969b6274ce6c7e80a35510dbfe131c98a6178d6, lgtm, not tested yet

iNecas:
Thanks, updated the commit message with the right reference to the original commit.

ares:
Thanks @iNecas, works well, merged as https://github.com/theforeman/foreman/commit/e91844d873efa139ce02f0a4d7ca2887fc01e432

PR #4894, opened 2017-10-05
Fixes #21223: add experimental UI menu to foreman.

waldenraines:
Add an experimental UI setting to foreman so we can show new UI pages
while maintaining the old ones. This will allow us to showcase our new
UIs for the users and then eventually remove the old pages and replace
them with the new ones once they have been vetted.

http://projects.theforeman.org/issues/21223

### DIFF
--- a/app/registries/menu/loader.rb
+++ b/app/registries/menu/loader.rb
@@ -92,6 +92,10 @@ module Menu
           end
         end
       end
+
+      Manager.map :labs_menu do |menu|
+        menu.sub_menu :lab_features_menu, :caption => N_('Lab Features'), :icon => 'fa fa-flask'
+      end
     end
   end
 end
--- a/app/views/home/_navbar.html.erb
+++ b/app/views/home/_navbar.html.erb
@@ -3,7 +3,13 @@
   nav-pf-persistent-secondary">
   <ul class="list-group">
     <%= render_vertical_menu :top_menu %>
+
+    <% if Setting[:lab_features] %>
+      <%= render_vertical_menu :labs_menu %>
+    <% end %>
+
     <%= render_vertical_menu :admin_menu %>
+
     <%= render "home/vertical_taxonomies", :taxonomy_type => "location", :icon_name => "fa fa-globe" if SETTINGS[:locations_enabled]  %>
     <%= render "home/vertical_taxonomies", :taxonomy_type => "organization", :icon_name => "fa fa-building" if SETTINGS[:organizations_enabled]  %>
   </ul>
--- a/app/views/home/_topbar.html.erb
+++ b/app/views/home/_topbar.html.erb
@@ -14,12 +14,12 @@
   <ul class="nav navbar-nav navbar-left navbar-iconic navbar-utility">
     <%= render "home/org_switcher" %>
   </ul>
-      <% if User.current %>
-          <nav class="collapse navbar-collapse">
-            <div id="turbolinks-progress"><div class="spinner spinner-inverse"></div></div>
-            <ul class="nav navbar-nav navbar-right navbar-iconic navbar-utility">
-            <%= render 'home/user_dropdown' if SETTINGS[:login] %>
-            </ul>
-        </nav>
-      <% end %>
+  <% if User.current %>
+      <nav class="collapse navbar-collapse">
+        <div id="turbolinks-progress"><div class="spinner spinner-inverse"></div></div>
+        <ul class="nav navbar-nav navbar-right navbar-iconic navbar-utility">
+        <%= render 'home/user_dropdown' if SETTINGS[:login] %>
+        </ul>
+    </nav>
+  <% end %>
 </nav>
--- a/app/models/setting/general.rb
+++ b/app/models/setting/general.rb
@@ -21,7 +21,8 @@ class Setting::General < Setting
       self.set('login_text', N_("Text to be shown in the login-page footer"), nil, N_('Login page footer text')),
       self.set('host_power_status', N_("Show power status on host index page. This feature calls to compute resource providers which may lead to decreased performance on host listing page."), true, N_('Show host power status')),
       self.set('http_proxy', N_('Sets a proxy for all outgoing HTTP connections.'), nil, N_('HTTP(S) proxy')),
-      self.set('http_proxy_except_list', N_('Set hostnames to which requests are not to be proxied'), [], N_('HTTP(S) proxy except hosts'))
+      self.set('http_proxy_except_list', N_('Set hostnames to which requests are not to be proxied'), [], N_('HTTP(S) proxy except hosts')),
+      self.set('lab_features', N_("Whether or not to show a menu to access experimental lab features (requires reload of page)"), false, N_('Show Experimental Labs'))
     ]
   end
 


timogoebel:
It's unclear what the goal of this Setting is. Can you provide more context? Will it stay forever? Do we want to hide it in production setups? Do we 🚢 this in a release?

ares:
copying from the redmine issue

> Add an experimental UI setting to foreman so we can show new UI pages while maintaining the old ones.
> This will allow us to showcase our new UIs for the users and then eventually remove the old pages and replace them with the new ones once they have been vetted.

but I'd also appreciate more context or example of such page in Foreman core or more than one plugin. If that would be for one plugin only, perhaps it should be plugin's setting.

waldenraines:
> It's unclear what the goal of this Setting is. Can you provide more context? Will it stay forever? Do we want to hide it in production setups? Do we 🚢 this in a release?

The goal of this setting is to provide a way for users to use new UI before we release it when we have larger changes that replace an entire page (or pages).  One way to think of this is as a poor man's a/b test.

Hiding this in production setups could make sense but I could also see it being fine to keep it around if any user is interested to see the direction the UI will be going.  I personally think it's fine to ship this with a release.

> but I'd also appreciate more context or example of such page in Foreman core or more than one plugin. If that would be for one plugin only, perhaps it should be plugin's setting.

An example in Foreman core would be when we start work to replace the create host flow with a wizard.  This will be a large change and will likely not share any code with the existing implementation.  Having both UIs around at the same time enables us to ensure we have feature parity as well as having a way to compare the two and ensure we have the support of the community before replacing the existing UI.

timogoebel:
Thanks, let's try this.

waldenraines:
@timogoebel @ares anything else I need to do here before we can merge this?

app/views/home/_topbar.html.erb
```diff
@@ -41,6 +41,14 @@
           <%= render_menu :top_menu %>
         </ul>
         <ul class="nav navbar-nav navbar-admin navbar-right navbar-collapse collapse" id="menu2">
+          <% if Setting[:experimental_ui]%>
```
timogoebel: Nitpick: This is missing a space.

app/views/home/_topbar.html.erb
```diff
@@ -41,6 +41,14 @@
...
+              <li class="dropdown menu_tab_dropdown" id="react_root">
```
timogoebel: Why does this need `id="react_root"`?

timogoebel (changes requested):
@waldenraines: Thanks. Left some small comments.

ares:
I have no more comments, :+1: when Timo's comments are fixed

app/views/home/_topbar.html.erb
waldenraines: Because that's the ID we will reference in order to mount the react application.  See https://github.com/Katello/katello/pull/6992/files#diff-52f1696d7dae32f43202243e041b3af8R7 for an example.

ohadlevy: are we expecting only one? in that case we should probably call id="experimental_ui" instead of react.

waldenraines: Changed the name of the ID.  What do you mean by "only one"?  Only one what?

waldenraines:
@timogoebel addressed your comments.

I also updated this PR to use the vertical navigation that was merged today.

waldenraines:
[test katello]
[test foreman]

waldenraines:
[test foreman]

app/views/home/_navbar.html.erb
```diff
@@ -4,6 +4,14 @@
   <ul class="list-group">
     <%= render_vertical_menu :top_menu %>
     <%= render_vertical_menu :admin_menu %>
+    <% if Setting[:experimental_ui] %>
+        <li class="list-group-item secondary-nav-item-pf" id="experimental_ui">
```
ohadlevy: would this actually work this way? if you are using the react mounter, i would assume we would need to use a content div and not the menu li?

timogoebel:
@waldenraines: Sorry to make this so hard, but I still don't get this. Why don't you just add?

```
<%= render_vertical_menu :labs_menu %>
```

If we do it like that, this would work like any other menu in Foreman and e.g. plugins can extend this easily.

I'd also suggest naming this something other then "Experimaltal UI". How about "Lab Features"? We might want to use this for something else then UI changes.

danseethaler:
@timogoebel the reason is that we need a mounting point for the React app. React needs to setup and control the navigation items so when an nav item is clicked React and respond and update the page (React is all a single-page app, no page reload).

If there's something I'm missing I'd love to hear of a better way to integrate this. Either way React has to create the navigation UI in order to control it.

timogoebel:
>  (React is all a single-page app, no page reload).

Did we already decide to make Foreman a single page app? My impression was, that this was not the goal. Not, that I'm against doing that, but I have a feeling we need a rough plan where we want to go.

> If there's something I'm missing I'd love to hear of a better way to integrate this. Either way React has to create the navigation UI in order to control it.

Ok, that's fair. However, I'm wondering how the switch will look like when we want to add the first react page to a real menu item.

waldenraines:
> Did we already decide to make Foreman a single page app? My impression was, that this was not the goal. Not, that I'm against doing that, but I have a feeling we need a rough plan where we want to go.

That is our plan with katello (katello is already a single page app for the most part) and I think eventually within Foreman as well.  This is open to debate but I would love to see us move away from turbolinks and rails templates to a single page app (hereon abbreviated SPA) that uses foreman APIs.

> Ok, that's fair. However, I'm wondering how the switch will look like when we want to add the first react page to a real menu item.

I think it will be similar to how we do it in katello currently.  We have a [whitelist of pages](https://github.com/katello/katello/blob/master/engines/bastion_katello/lib/bastion_katello/engine.rb) in rails that are handled by angular.  I'm no rails expert so there may be a better way of doing this but the gist of it is that we want to load the SPA layout when the URL matches one of the SPA pages.  From there the SPA takes over the control of the routing and does not do a full page reload when visiting other "pages" of the SPA.

waldenraines:
> @waldenraines: Sorry to make this so hard, but I still don't get this. Why don't you just add?
> 
> <%= render_vertical_menu :labs_menu %>
> If we do it like that, this would work like any other menu in Foreman and e.g. plugins can extend this easily.

I can do that instead, no worries.

> I'd also suggest naming this something other then "Experimaltal UI". How about "Lab Features"? We might want to use this for something else then UI changes.

That's a good idea.  I can also rename the setting to match.

waldenraines:
@timogoebel updated.

app/registries/menu/loader.rb
```diff
@@ -92,6 +92,12 @@ def self.load
           end
         end
       end
+
+      Manager.map :labs_menu do |menu|
+        menu.sub_menu :lab_features_menu, :caption => N_('Lab Features'), :icon => 'fa fa-flask' do
+          menu.item :dashboard,         :caption => N_('Dashboard')
```
waldenraines: @timogoebel is there a better way to do this?  I couldn't get the menu to show up without including at least one `menu.item`.

app/views/layouts/base.html.erb
```diff
@@ -45,6 +45,9 @@
     <!--[if lte IE 9]>
       <%= _('You are using an unsupported browser.') %>
     <![endif]-->
+    <% if Setting[:lab_features] %>
+      <div id="reactRoot"></div>
```
waldenraines: Necessary as a mounting point for the react SPA.

waldenraines:
@timogoebel I think I have addressed all of your requested changes.  Let me know if you have any other feedback.

app/registries/menu/loader.rb
timogoebel: I'm wondering if we should add a default menu item in the first place. If plugins add a "lab feature", they can extend this with their own menu item that actually makes sense.

waldenraines: I was thinking that by doing this it makes it obvious which items are in beta (or even alpha) state.  If we just put them into their own menu then we would need some other way of letting users know that these pages may not work 100% and/or may have bugs.

timogoebel: @waldenraines: Well, yes. I get the idea. But why should we add "Dashboard" by default? It's not in beta or alpha state.

waldenraines: Oh I see, sorry I must have misread the above.

I could not get the menu item to display at all without including this.  I suppose I could remove this and leave the menu item empty and not displaying at all for now until someone adds something to it.  It will just make it more difficult to test.

timogoebel:
@waldenraines: I tested this with a plugin:

```ruby
          # add menu entry
          menu :labs_menu, :dlm_locks,
               url_hash: { controller: :'dlmlocks', action: :index },
               caption: 'Distributed Locks',
               parent: :lab_features_menu
```

image

Works fine. If you'd kindly remove the Dashboard link, I think this is ready to 🚢

waldenraines:
> Works fine. If you'd kindly remove the Dashboard link, I think this is ready to 🚢

@timogoebel updated.

timogoebel (approved):

waldenraines:
[test]

waldenraines:
> Some files could not be reviewed due to errors:

Does anyone have any thoughts on this?

waldenraines:
[test]

ohadlevy:
@waldenraines can you please rebase to fix the test failures?

waldenraines:
> @waldenraines can you please rebase to fix the test failures?

Done.

timogoebel:
Merged, thanks @waldenraines.

lzap:
Nice idea.

PR #4897, opened 2017-10-06
Fixes #18422 - Expose dynflow worker pool size in config

adamruzicka:


iNecas (approved):
Tested and works well. Letting others to review but ok to merge from my point

dLobatog (approved):
Thanks @adamruzicka , out of curiosity why the db_pool_size has a + 5 ? Will merge after knowing that (could be a typo)

adamruzicka:
Dynflow has pool_size set to 5 and db_pool_size set to pool_size + 5 by default. My understanding was that the db_pool should be large enough to have a connection for each worker and some more just in case. Did I get that right, @iNecas ?

iNecas:
Yes. That's the reason: one connection in pool for each worker + additional for non-worker connections to database.

dLobatog:
Thanks @iNecas too

PR #4899, opened 2017-10-09
Fixes #21213 - don't fail on nil time stamps

ares:
that was behavior of a `time_column` helper removed in d738d2c0, this will need to be cherry-picked to 1.16


dLobatog (approved):
Thanks @ares !

PR #4900, opened 2017-10-09
Fixes #21246 - unify extlogout and login footer text

ares:


ares:
To reproduce, set something in Settings -> General -> Login page footer text
before the patch:
before

after the patch:
after

dLobatog (approved):
Thanks @ares !

dLobatog:
If tests are green I'll merge

PR #4902, opened 2017-10-09
Refs #20800 - Align "session timed out" warning with login messages

bastilian:
The session timed out message was missed for #4796. It is now also inline as the other two:

gnome-shell-screenshot-sejn7y

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -169,16 +169,25 @@ class ApplicationController < ActionController::Base
     flash_message(:error, message, now)
   end
 
-  def inline_error(message, now = false)
-    flash[:inline] = { :error => CGI.escapeHTML(message) }
-  end
-
-  def inline_success(message, now = false)
-    flash[:inline] = { :success => CGI.escapeHTML(message) }
-  end
-
   def warning(message, now = false)
     flash_message(:warning, message, now)
+  end
+
+  def inline_error(message)
+    inline_flash_message(:error, message)
+  end
+
+  def inline_success(message)
+    inline_flash_message(:success, message)
+  end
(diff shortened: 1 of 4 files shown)


chris1984 (approved):

app/controllers/application_controller.rb
```diff
@@ -169,16 +169,25 @@ def error(message, now = false)
...
+    flash_message(:warning, message, now)
   end
 
-  def inline_success(message, now = false)
-    flash[:inline] = { :success => CGI.escapeHTML(message) }
+  def inline_error(message)
+    inline_flash_message(:error, message)
   end
 
-  def warning(message, now = false)
```
ohadlevy, line 180: should we deprecate this method instead? I'm uncertain if anyone is using it. also for consistency, what about other helper methods (error, notice etc).

ares: ApplicationShared#set_taxonomy uses it for sure, quick grep shows some plugins do too, see foreman-docker/app/controllers/containers_controller.rb

ohadlevy:
What's the status of this pr? Thanks

bastilian:
@ohadlevy only to be merged, i suppose. Just fixed a minor mistake, which removed the session expired log. Deprecating the `warning` method seems to be unlikely since plugins seem to rely on it, or @ares?

tbrisker (approved):
LGTM, thanks @bastilian ! please open cherry-pick PR to 1.17-stable as well

tbrisker:
[test foreman]

tbrisker:
Failures are regular intermittent ones with capybara, merging, Thanks @bastilian !

PR #4903, opened 2017-10-09
Fixes #21253 - fix latest events widget

akofink:


dLobatog (commented):
@tbrisker Could you check this out? I'm not quite sure by looking at the git history why this was so complicated before. Thanks!

ares:
Looking at ee008293f8396da52b4c2fa7870bf4114df1b25b that introduced it, the workaround was required because of older rails bug. @tbrisker covered it with tests which are green so I believe this is no longer necessary. I tested this also on older stack and it fixes the last event widget. Latest errata is still broken but that most likely needs to be fixed in katello, @akofink do you plan to look at that one too? The linked BZ mentions both widgets so after we merge this, it would be marked as fixed... 

Tested on

- foreman 1.15.4
- katello 3.4.5
- rails 4.2.6

If noone objects, I'll merge this.

akofink:
https://github.com/Katello/katello/pull/6995 and http://projects.theforeman.org/issues/21254 are for the errata widget. I just added 21254 to the BZ. Feel free to merge! Thanks!

timogoebel (approved):

timogoebel:
Merged, thanks @akofink, @ares, @dLobatog.

PR #4904, opened 2017-10-10
Fixes #21266 - enabled all PXE loaders for SUSE

lzap:
We intentionally enable only those loaders we know they are working. Looks like we missed SUSE OS.


dLobatog (approved):
Thanks @lzap

PR #4905, opened 2017-10-10
Fixes #21269 - fixed dns_lookup with address resolution

lzap:
Although this was covered with tests, network things are stubbed for obvious reasons so we did not catch this. Method resolver.getname requires string and not IPAddress instance.


dLobatog (approved):
Thanks @lzap

dLobatog:
cp-ed to 1.15-stable & 1.16-stable

PR #4907, opened 2017-10-11
Fixes #21299 - Make user_id required

mbacovsky:
user_id is part of the resource URL and needs to be mandatory.


timogoebel:
Redmine Ticket was moved to Foreman.

timogoebel (approved):
Looks good, thanks @mbacovsky.

PR #4908, opened 2017-10-11
Fixes #21295 - load the routes before we load the controllers

iNecas:
After upgrade from Rails 4.1 to Rails 4.2 there was significant
drop on rails startup in production. I have tracked it down to
several changes in Rails, but it might also be affected by the
fact the we use the url_helpers on more places than older Foreman
versions.

The reason of the slowness is the `define_method` calls on a module for
every named route we have. The `define_method` is more expensive
the more classes include the module that gets the new method.

The solution of this problem is to load the routes before we
load the controllers (including the eager loading) so that the modules
get defined new methods before the controllers include it in their code.

The original after_finalize loading is still here, but should not get
run again, as it callse `execute_if_updated` and nothing new usually
gets updated there.

To test, one can try in devel setup:

    time echo exit | bundle exec rails c production

or in production:

    service httpd restart; time curl -k https://localhost

In devel setup, I got from 30s to 10s, in production with Katello
and other plugins (it gets more and more visible with more plugins
around), I got from 1m 15s to 15s.

Please not `foreman-rake console` is not affected, as the rake
environment turns off the eager loading.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -253,6 +253,27 @@ module Foreman
       @dynflow
     end
 
+    # We need to mount the sprockets engine before we use the routes_reloader
+    initializer(:mount_sprocket_env, :before => :sooner_routes_load) do
+      if config.assets.compile
+        app = Rails.application
+        if Sprockets::Railtie.instance.respond_to?(:build_environment)
+          app.assets = Sprockets::Railtie.instance.build_environment(app, true)
+        end
+        routes.prepend do
+          mount app.assets => app.config.assets.prefix
+        end
+      end
+    end
+
+    # We use the routes_reloader before the to_prepare and eager_load callbacks
+    # to make the routes load sooner than the controllers. Otherwise, the definition
+    # of named routes helpers in the module significantly slows down the startup
+    # of the application. Switching the order helps a lot.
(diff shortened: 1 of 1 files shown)


tbrisker:
[test foreman] [test katello] [test upgrade] 
not sure why these didn't run...

tbrisker:
[test foreman] looks like pg is failing, rerunning to see if it's related or not

iNecas:
There is an issue with the assets pipeline after this change, I'm looking currently on what's going on

iNecas:
I figured that the sprockets server gets mounted in the after_finalize phase as well, and the routes_reloader was not picking it up after that. It's not very pretty, but since the significant impact this has on the startup of the production environment, I would like to rather fix it by making sure to load it sooner that suffer with the slow start.

ehelms (approved):
From an assets precompile and plugin precompile aspect this works and looks good.

iNecas:
Tests are green now

mccun934:
anyone with merge rights can take a look at this one? looks good to go

ares:
Tested, this is huge speed boost, thanks @iNecas! Merged as c2e58b1b275e0107de4089e15adc8b6203e696b1

lzap:
Bravo!

PR #4909, opened 2017-10-12
Fixes #21312 - components for formatting dates

tstrachota:
This patch reuses components from [react-intl](https://github.com/yahoo/react-intl) and extends the component registry to provide an intl wrapper (react-intl requires the component to be wrapped in IntlProvider). That could be used also if we decide that react-intl is the way to for translating our whole react stack.

**TODO:**
- [x] tests for the components


webpack/assets/javascripts/react_app/components/common/Date/AbsoluteDate.js
```diff
@@ -0,0 +1,40 @@
...
+      title = this.props.intl.formatRelative(date);
+
+      return (
+        <span title={title}>
+          <FormattedDate value={date}
+            day='2-digit'
+            month='short'
+            hour='2-digit'
+            minute='2-digit'
+            hour12={false} />
```
waldenraines: I know that most of the world uses 24hr time but to set this to false is saying that we are okay with actively showing US users a time that they may not readily understand.

Shouldn't this be updated by whatever intl library we are using based on the user's locale?

webpack/assets/javascripts/react_app/components/common/Date/RelativeDate.js
```diff
@@ -0,0 +1,42 @@
...
+
+    if (this.props.date) {
+      date = this.props.date;
+
+      title = this.props.intl.formatDate(date, {
+        day: '2-digit',
+        month: 'short',
+        hour: '2-digit',
+        minute: '2-digit',
+        hour12: false
```
waldenraines: This comment also applies to the `AbsoluteDate` class above but I would think that all of these should be set based on the user's locale and not hardcoded.

waldenraines (changes requested):
I'm wondering about the date formatting.  For instance I see the format `Oct 16, 13:27` in several places.  I think that we should include the year because in some places (sync plans, for instance) the date may be in the past and it may not be clear which year it is referring to.  

See also my other comments about displaying locale specific dates as well.

danseethaler:
Aside from the couple of configuration things mentioned I think this is great. I really like the approach of utilizing the new register components to make the helper components available.

tstrachota:
Thanks @waldenraines your notes make sense. I'll remove the hardcoded 24 hour format.

My initial idea was to get as close as possible to the date format that we currently sue in the foreman. But your point about the year is valid, we need more components. I looked at the date and time formats that we use in both projects.

**Katello uses:**

filter    | example for en_US       | usage notes
----------|-------------------------|-------
shortDate | 9/3/10                  | used for errata issued/updated dates and subscription start/end dates, used 12x
short     | 9/3/10 12:05 PM         | various other places, used 23x
medium    | Sep 3, 2010 12:05:08 PM | various other places, used 10x
longDate  | September 3, 2010       | used only at one place (content view last published)


**Foreman uses:**

format | example for en_US       | usage notes
-------|-------------------------|-------
long   | February 12, 2021 17:13 | didn't find a place where it would be used
short  | Aug 31, 12:52           |

My assumptions are:
- we need a component for showing just a date (without time)
- in most of the cases it's fine to show the date and time without year
- we still need a component for showing date, time and year for actions that are less frequent
- `longDate` is used only once and can be replaced by date
- it's not crutial to display seconds in the dates

Based on that I propose to add following components/formats:

component     | example for en_US
--------------|---------------------------
LongDateTime  | October 13, 2017, 1:54 PM
ShortDateTime | Oct 13, 1:54 PM
Date          | 9/3/2010

What do you think? The rails helpers in foreman can probably just mount the new components to avoid nuances between the dates rendered form js and at the server side.

@rohoover I'd also like to know your opinion on that.

webpack/assets/javascripts/react_app/components/common/Date/AbsoluteDate.js
```diff
@@ -0,0 +1,37 @@
...
+      date = this.props.date;
+      title = this.props.intl.formatRelative(date);
+
+      return (
+          <FormattedDate value={date}
+            day="2-digit"
+            month="short"
+            hour="2-digit"
+            minute="2-digit" />
+        </span>
```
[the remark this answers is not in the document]
tstrachota: Fixed, I pushed wrong commit by mistake.

waldenraines:
> What do you think?

@tstrachota that seems like a good analysis and makes sense to me 👍

Rohoover:
@tstrachota 

Looks good. 

Just as a note, a lot of users I've talked to would like to be able to set a preference. Not something for this PR, but worth mentioning.

ares:
@tstrachota all above makes sense to me, just one thing

> it's not crutial to display seconds in the dates

 I think there are some use cases, perhaps at job invocation, seconds matter

waldenraines:
@tstrachota updates on this?  I think we have reached a decision on date formatting so we should be good to proceed.

tstrachota:
Updated:
- added more components according to the discussion earlier in this PR
- components reflect user's timezone
- rails helpers now use the components
- small refactoring of component registry to provide a single entry point for HOC wrappers

mmoll:
this needs a rebase :)

app/helpers/application_helper.rb
```diff
@@ -51,23 +51,17 @@ def multiple_with_filter?
   # it supports two formats :short and :long
   # example of long is February 12, 2021 17:13
   # example of short is Aug 31, 12:52
-  def date_time_absolute(time, format = :short)
```
ohadlevy, line 54: can you please extract all of the date/time methods to dedicated helper?

app/helpers/application_helper.rb
```diff
@@ -78,8 +72,25 @@ def date_time_relative_value(time)
...
   protected
 
+  def generate_date_id
+    @@date_counter ||= 0
+    @@date_counter += 1
+    "datetime_#{@@date_counter}"
+  end
+
+  def mount_date_component(component, time, seconds)
+    date_id = generate_date_id
```
ohadlevy: can't we use something like a timestamp instead of class vars?

webpack/assets/javascripts/react_app/common/i18n.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import {injectIntl, IntlProvider, addLocaleData} from 'react-intl';
+
+let langAttr = document.getElementsByTagName('html')[0].getAttribute('lang') || 'en';
```
ohadlevy: should be const?

webpack/assets/javascripts/react_app/components/common/Date/Date.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { FormattedDate } from 'react-intl';
+import { timezone } from '../../../common/i18n';
+
+// 9/3/2010
```
ohadlevy: left over?

webpack/assets/javascripts/react_app/components/common/Date/Date.js
```diff
@@ -0,0 +1,42 @@
...
+class LongDateTime extends React.Component {
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    let date, title;
+
+    if (this.props.data.date) {
+      date = this.props.data.date;
```
ohadlevy: please use const date instead of the above let

ohadlevy (commented):
thanks, a few minor comments inline from a quick review.

How do we extract the strings from react-int? in case there are new strings/languages that we support?

ohadlevy:
also, rebase and tests for new components are missings :)

app/helpers/application_helper.rb
tstrachota: I can do that but I'm not sure if some plugins aren't using it already. How about foreman-openscap. Is it used there @ares ?

webpack/assets/javascripts/react_app/components/common/Date/Date.js
tstrachota: Not really. I planned to keep the date format with each component just to give people better clue about what is the output. But I agree it deserves some more explanation.

ohadlevy: once you add jest snapshots, we could see the output ;)

webpack/assets/javascripts/react_app/components/common/Date/LongDateTime.js
```diff
@@ -0,0 +1,46 @@
...
+    super(props);
+  }
+
+  render() {
+    let date, title, seconds;
+
+    if (this.props.data.date) {
+      date = this.props.data.date;
+      title = this.props.intl.formatRelative(date);
+      seconds = this.props.data.seconds ? '2-digit' : undefined;
```
sharvit: These variables should be defined inside the `if` block as consts

webpack/assets/javascripts/react_app/components/common/Date/RelativeDateTime.js
```diff
@@ -0,0 +1,46 @@
...
+
+      title = this.props.intl.formatDate(date, {
+        day: '2-digit',
+        month: 'short',
+        hour: '2-digit',
+        minute: '2-digit',
+        year: 'numeric',
+        timeZone: timezone
+      });
+
```
sharvit: These variables should be defined inside the `if` block as consts

webpack/assets/javascripts/react_app/components/common/Date/ShortDateTime.js
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  render() {
+    let date, title, seconds;
+
+    if (this.props.data.date) {
+      date = this.props.data.date;
+      title = this.props.intl.formatRelative(date);
+      seconds = this.props.data.seconds ? '2-digit' : undefined;
+
```
sharvit: These variables should be defined inside the `if` block as consts

webpack/assets/javascripts/react_app/components/componentRegistry.js
```diff
@@ -4,9 +4,92 @@ import StatisticsChartsList from './statistics/StatisticsChartsList';
...
+import LongDateTime from './common/Date/LongDateTime';
+import ShortDateTime from './common/Date/ShortDateTime';
+import Date from './common/Date/Date';
 import StorageContainer from './hosts/storage/vmware/';
 import forEach from 'lodash/forEach';
 import map from 'lodash/map';
+import { withIntlProvider } from '../common/i18n';
+import { Provider } from 'react-redux';
+
+function getDisplayName(Component) {
```
sharvit: This function is also used in `common/i18n.js` can we reuse it?

webpack/assets/javascripts/react_app/components/componentRegistry.js
```diff
@@ -4,9 +4,92 @@ import StatisticsChartsList from './statistics/StatisticsChartsList';
...
+        );
+      }
+    }
+    PropDataMapper.displayName = `PropDataMapper(${getDisplayName(WrappedComponent)})`;
+
+    return PropDataMapper;
+  };
+}
+
+class WrapperFactory {
```
sharvit: I believe that the WrapperFactory together with the functions above should be separate into a different file

sharvit (changes requested):

tstrachota:
Updated according to comments and tests added.

sharvit (approved):

ohadlevy:
@tstrachota can you rebase please? also note the test failures.

ohadlevy:
@ares any feedback on this PR?

ares:
:+1: from me, I'm happy the method usage didn't change much, @tstrachota would you mind adding the third param to how to write a plugin wiki after this is merged?

tstrachota:
Rebeased. I need to investigate what went wrong with the tests.

@ares sure, I can update the docs.

ohadlevy:
What about i18n?

On Nov 8, 2017 5:48 PM, "Hound" <notifications@github.com> wrote:

*@houndci-bot* commented on this pull request.
------------------------------

In webpack/assets/javascripts/bundle.js
<https://github.com/theforeman/foreman/pull/4909#discussion_r149708455>:

> @@ -25,5 +27,10 @@ window.tfm = Object.assign(window.tfm || {}, {
   numFields: require('./jquery.ui.custom_spinners'),
   reactMounter: require('./react_app/common/MountingService'),
   editor: require('./foreman_editor'),
-  nav: require('./foreman_navigation')
+  nav: require('./foreman_navigation').
+  componentRegistry: componentRegistry,

Parsing error: Unexpected token, expected ,

—
You are receiving this because you commented.

<https://github.com/theforeman/foreman/pull/4909#pullrequestreview-75144700>,
or mute the thread

tstrachota:
Ok, tests are green now.
The previous issue was a typo while resolving rebase conflicts, @ohadlevy . I'm sorry.

waldenraines (approved):
This LGTM and worked where tested (also tested with katello) but will allow someone else more familiar with foreman the final approval.

ares:
This needs a rebase

tstrachota:
Rebased

waldenraines:
@ares @ohadlevy any other comments here?

ares:
I'm trying to test but I can't install dependencies, I always get this, anyone knows?

```
npm WARN react-addons-test-utils@15.6.2 requires a peer of react-dom@^15.4.2 but none is installed. You must install peer dependencies yourself.

npm ERR! Cannot read property '0' of undefined

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/ares/.npm/_logs/2017-12-07T21_18_53_635Z-debug.log
```

EDIT: nvm, looks unrelated, can reproduce now even after resetting back to develop, suggestions welcome though :-)

waldenraines:
> EDIT: nvm, looks unrelated, can reproduce now even after resetting back to develop, suggestions welcome though :-)

There's always the big hammer (`rm -fr node_modules; npm i`).  Sometimes npm gets confused it seems.

ohadlevy:
@ares were you able to review this pr?

app/helpers/application_helper.rb
ares: It's already used, which should not prevent the extraction. I don't understand why it should be in separate helper though? This is application wide helper method. If it was extracted to separate module, I'd still like to include it everywhere.

ares:
Today, npm was in a good mood so I was able to test. All works well, including timezone and locale. If there are no more comments, I'm happy to merge.

ares (approved):

ares:
Turns out this adds 15MB of strings because of intl dependency, @tstrachota will take a look whether we can keep intl out of this, if not, we'll revert this commit before 1.17 branching. We should have decision by the end of the day. @xprazak2 FYI

PR #4911, opened 2017-10-12
Fixes #21317 - graceful handling of VM association error

iNecas:
Instead of immediate failure, just log the issue and proceed with the
import.

### DIFF
--- a/app/controllers/api/v2/compute_resources_controller.rb
+++ b/app/controllers/api/v2/compute_resources_controller.rb
@@ -148,19 +148,14 @@ module Api
       api :PUT, "/compute_resources/:id/associate/", N_("Associate VMs to Hosts")
       param :id, :identifier, :required => true
       def associate
-        @hosts = []
-        if @compute_resource.respond_to?(:associated_host)
-          @compute_resource.vms(:eager_loading => true).each do |vm|
-            if Host.for_vm(@compute_resource, vm).empty?
-              host = @compute_resource.associated_host(vm)
-              if host.present?
-                host.associate!(@compute_resource, vm)
-                @hosts << host
-              end
-            end
-          end
+        if @compute_resource.supports_host_association?
+          associator = ComputeResourceHostAssociator.new(@compute_resource)
+          associator.associate_hosts
+          @hosts = associator.hosts
+          render 'api/v2/hosts/index', :layout => 'api/v2/layouts/index_layout'
+        else
+          render_message(_('Associating VMs is not supported for this compute resource'), :status => :unprocessable_entity)
         end
-        render 'api/v2/hosts/index', :layout => 'api/v2/layouts/index_layout'
       end
 
       api :PUT, "/compute_resources/:id/refresh_cache/", N_("Refresh Compute Resource Cache")
--- a/app/controllers/compute_resources_controller.rb
+++ b/app/controllers/compute_resources_controller.rb
@@ -42,19 +42,26 @@ class ComputeResourcesController < ApplicationController
   end
 
   def associate
-    count = 0
-    if @compute_resource.respond_to?(:associated_host)
-      @compute_resource.vms(:eager_loading => true).each do |vm|
-        if Host.for_vm(@compute_resource, vm).empty?
-          host = @compute_resource.associated_host(vm)
-          if host.present?
-            host.associate!(@compute_resource, vm)
-            count += 1
-          end
-        end
+    if @compute_resource.supports_host_association?
+      associator = ComputeResourceHostAssociator.new(@compute_resource)
+      associator.associate_hosts
+      messages = []
+      if associator.hosts.empty?
+        messages << _('No VMs matched any host.')
+      else
+        messages << n_('%s VM was associated to a host.', '%s VMs were each associated to hosts.', associator.hosts.count) % associator.hosts.count
       end
+      if associator.fail_count > 0
+        messages << n_('%s VM failed while processing: check logs for more details.',
+                       '%s VMs failed while processing: check logs for more details.',
+                       associator.fail_count) % associator.fail_count
+        process_error(:error_msg => messages.join(' '))
+      else
+        process_success(:success_msg => messages.join(' '))
+      end
+    else
+      process_error(:error_msg => 'Associating VMs is not supported for this compute resource.')
     end
-    process_success(:success_msg => n_("%s VM associated to a host", "%s VMs associated to hosts", count) % count)
   end
 
   def update
--- a/app/models/compute_resource.rb
+++ b/app/models/compute_resource.rb
@@ -361,6 +361,10 @@ class ComputeResource < ApplicationRecord
     true
   end
 
+  def supports_host_association?
+    respond_to?(:associated_host)
+  end
+
   protected
 
   def client
--- a/app/services/compute_resource_host_associator.rb
+++ b/app/services/compute_resource_host_associator.rb
@@ -1,0 +1,30 @@
+class ComputeResourceHostAssociator
+  attr_accessor :compute_resource, :hosts, :fail_count
+
+  def initialize(compute_resource)
+    @hosts = []
+    @fail_count = 0
+    self.compute_resource = compute_resource
+  end
+
+  def associate_hosts
+    compute_resource.vms(:eager_loading => true).each do |vm|
+      if Host.for_vm(compute_resource, vm).empty?
+        associate_vm(vm)
+      end
+    end
+  end
+
+  private
+
+  def associate_vm(vm)
+    host = compute_resource.associated_host(vm)
+    if host.present?
+      host.associate!(compute_resource, vm)
+      @hosts << host
+    end
+  rescue StandardError => e
+    @fail_count += 1
+    Foreman::Logging.exception("Could not associate VM #{vm}", e)
+  end
+end
--- a/test/unit/compute_resource_host_associator_test.rb
+++ b/test/unit/compute_resource_host_associator_test.rb
@@ -1,0 +1,38 @@
+require 'test_helper'
+
+class ComputeResourceHostAssociatorTest < ActiveSupport::TestCase
+  let(:associator) do
+    ComputeResourceHostAssociator.new(compute_resource)
+  end
+  let(:compute_resource) { FactoryBot.create(:ec2_cr) }
+  let(:vm1) { stub('vm1', :identity => Foreman.uuid) }
+  let(:vm2) { stub('vm2', :identity => Foreman.uuid) }
+  let(:vm3) { stub('vm3', :identity => Foreman.uuid) }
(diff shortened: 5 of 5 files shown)


app/models/compute_resource.rb
```diff
@@ -360,6 +360,31 @@ def image_exists?(image)
...
   end
 
+  def associate_hosts_support?
+    respond_to? :associated_host
+  end
+
+  def associate_hosts
+    ret = { :hosts => [], :fail_count => 0 }
+    return ret
+    vms(:eager_loading => true).each do |vm|
```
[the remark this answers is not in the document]
iNecas: Ups: residual from testing. Fixed. Good dog!

iNecas:
I've took @dLobatog's patch from https://bugzilla.redhat.com/show_bug.cgi?id=1496680#c7 and polished it a bit.

timogoebel:
While we're touching this, how about extracting this to a Service Class? I think the result of `associate_hosts` method is pretty unintuitive and could be done better. Just a suggestion, I'm fine with this as well.

iNecas:
Makes sense: before I jump on it, do you have some idea on:

1. name for the class
2. how it would be used from the controller?

I can think of something, but if you already have some thoughts, I can incorporate it in the patch

app/controllers/api/v2/compute_resources_controller.rb
```diff
@@ -149,18 +149,12 @@ def available_security_groups
...
-          @compute_resource.vms(:eager_loading => true).each do |vm|
-            if Host.for_vm(@compute_resource, vm).empty?
-              host = @compute_resource.associated_host(vm)
-              if host.present?
-                host.associate!(@compute_resource, vm)
-                @hosts << host
-              end
-            end
-          end
+        if @compute_resource.associate_hosts_support?
```
timogoebel: Nitpick: Perhaps `@compute_resource.supports_associate_hosts?`?

app/controllers/api/v2/compute_resources_controller.rb
```diff
@@ -149,18 +149,12 @@ def available_security_groups
...
+          @hosts = @compute_resource.associate_hosts[:hosts]
+          render 'api/v2/hosts/index', :layout => 'api/v2/layouts/index_layout'
+        else
+          render_message(_('Associating vms is not supported for this compute resource'), :status => :unprocessable_entity)
```
timogoebel: Capitalize VMs: `Associating VMs is not supported for this compute resource`?

timogoebel (commented):
@iNecas. Cool. How about something like this:

```ruby
associater = ComputeResourceHostAssociater.new(
  :compute_resource => self || @compute_resource, # if you want to call it in the controller or not
).associate(vms)
if associater.errors?
  puts _("Errors during associate: %s") % associater.errors.full_messages.to_sentence
else
  puts _("%s Hosts were associated. You are awesome.") % associater.count
end
# Or even add the possibility to grab a list of successful / failed hosts so we could store them in a table in the future.
```

Definitely not the scope of this PR, but: All in all, I think this would be a good job for foreman-tasks. Can we do something similar with ActiveJob? So we can actually show a progress bar or similar visualization? At least for VMWare, this call is pretty slow.

ares:
Switching to WoC

iNecas:
I believe I've addressed all the comments

timogoebel:
@iNecas: Looks good to me, thanks. Well done. Now that we have the class isolated, would you mind adding some basic test coverage? :-) Thanks.

iNecas:
It's already implicitly covered in the v2 compute resources controller, but I can be more explicit, yes

app/services/compute_resource_host_associator.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  private
+
+  def associate_vm(vm)
+    host = compute_resource.associated_host(vm)
+    if host.present?
+      host.associate!(compute_resource, vm)
+      @hosts << host
+    end
+  rescue => e
```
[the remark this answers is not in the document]
iNecas: I'm sorry doggy, but we don't have this cop enabled and I feel quite strongly this is the right approach for this code. Thanks for your input. Take care ☮️

iNecas:
Tests added. Squashed. Rebased. Should be ready for final review/merge.

app/controllers/compute_resources_controller.rb
```diff
@@ -42,19 +42,20 @@ def edit
...
-          if host.present?
-            host.associate!(@compute_resource, vm)
-            count += 1
-          end
-        end
+    associator = ComputeResourceHostAssociator.new(@compute_resource)
+    if associator.supported_compute_resource?
+      associator.associate_hosts
+      messages = []
+      messages << _('%s VMs associated to hosts') % associator.hosts.count
```
timogoebel: '12 VMs were each associated to a host.' ?

app/controllers/compute_resources_controller.rb
```diff
@@ -42,19 +42,20 @@ def edit
...
+      if associator.fail_count > 0
+        messages << _('%s VMs failed while processing: check logs for more details') % associator.fail_count if associator.fail_count > 0
```
timogoebel: The `if associator.fail_count > 0` is redundant here.

app/services/compute_resource_host_associator.rb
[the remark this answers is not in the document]
timogoebel: rescue "StandardError" should get rid of this.

app/services/compute_resource_host_associator.rb
```diff
@@ -0,0 +1,36 @@
...
+    @fail_count += 1
+    logger.error "Could not associate VM #{vm}"
```
timogoebel: `Foreman::Logging.exception("Could not associate VM #{vm}", e)` ?

app/controllers/api/v2/compute_resources_controller.rb
```diff
@@ -148,19 +148,15 @@ def available_security_groups
...
-          @compute_resource.vms(:eager_loading => true).each do |vm|
-            if Host.for_vm(@compute_resource, vm).empty?
-              host = @compute_resource.associated_host(vm)
-              if host.present?
-                host.associate!(@compute_resource, vm)
-                @hosts << host
-              end
-            end
-          end
+        if associator.supported_compute_resource?
```
timogoebel: Nit: I think, I'd make this method part of the compute resource: `@compute_resource.supports_host_association?` or something.

app/services/compute_resource_host_associator.rb
```diff
@@ -0,0 +1,36 @@
+class ComputeResourceHostAssociator
+  attr_accessor :compute_resource, :hosts, :fail_count
+  delegate :logger, :to => :Rails
+
+  def initialize(compute_resource)
```
timogoebel: This class will generate a lot of sql queries. We should probably open a ticket to make this more efficient once this is merged.

timogoebel (changes requested):
@iNecas: Thanks. This looks a lot better than before! 👍
I left some comments inline.

I do think, the association isn't very efficiently implemented but this should do for now. Let's just remember to open a ticket to improve this if possible.

app/services/compute_resource_host_associator.rb
iNecas: http://projects.theforeman.org/issues/22172

iNecas:
All valid inputs, comments should be addressed now.

iNecas:
All green, should be ready for squash&merge

timogoebel (approved):

app/controllers/compute_resources_controller.rb
```diff
@@ -42,19 +42,20 @@ def edit
...
+      messages = []
+      messages << _('%s VMs were each associated to hosts') % associator.hosts.count
+      if associator.fail_count > 0
+        messages << _('%s VMs failed while processing: check logs for more details') % associator.fail_count
+        process_error(:error_msg => messages.join(', '))
+      else
+        process_success(:success_msg => messages.join(', '))
       end
+    else
+      process_error(:error_msg => 'Associating vms is not supported for this compute resource')
```
timogoebel: Nit: "Associating VMs..." (capitalize VMs)

app/controllers/compute_resources_controller.rb
```diff
@@ -42,19 +42,20 @@ def edit
...
-          if host.present?
-            host.associate!(@compute_resource, vm)
-            count += 1
-          end
-        end
+    if @compute_resource.supports_host_association?
+      associator = ComputeResourceHostAssociator.new(@compute_resource)
+      associator.associate_hosts
+      messages = []
+      messages << _('%s VMs were each associated to hosts') % associator.hosts.count
```
timogoebel: This does not handle the case for 1 VM (pluralize). Also this is missing a trailing ".".
Can we add a special message when no VM at all was associated?

"No VMs matched any host in Foreman, no VMs were associated."

image

timogoebel (changes requested):
@iNecas: Thanks, almost there. I found some very small things while testing this manually. If they are fixed, we're ready to 🚢.

iNecas:
Updated, rebased, squashed

app/controllers/compute_resources_controller.rb
```diff
@@ -42,19 +42,24 @@ def edit
...
+      associator = ComputeResourceHostAssociator.new(@compute_resource)
+      associator.associate_hosts
+      messages = []
+      if associator.hosts.empty?
+        messages << _('No VMs matched any host.')
+      else
+        messages << n_('%s VM was associated to a host.', '%s VMs were each associated to hosts.', associator.hosts.count) % associator.hosts.count
+      end
+      if associator.fail_count > 0
+        messages << _('%s VMs failed while processing: check logs for more details.') % associator.fail_count
```
timogoebel: This should also handle singular and plural correctly.

app/controllers/compute_resources_controller.rb
```diff
@@ -42,19 +42,24 @@ def edit
...
+        process_error(:error_msg => messages.join(', '))
```
timogoebel: Now, that the messages have a dot, they cannot be joined by ', '. This will lead to '., '.

timogoebel (changes requested):
@iNecas: Sorry for being a smartass here, but could you take a quick look at the two wording comments. Thanks.

app/controllers/compute_resources_controller.rb
iNecas: Updated

app/controllers/compute_resources_controller.rb
iNecas: Updated

ares:
katello test failures seems unrelated

timogoebel (approved):

timogoebel:
Merged, thanks @iNecas!

PR #4912, opened 2017-10-13
Fixes #21325 - Updating custom facts with booleans through api

patchon:
As described in the issue, currently it's not possible to upload facts with
false booleans to foreman through its api. This commit simply fixes this by
renaming the function that handles the "normalization" of the incomming facts
hash.


patchon:
Not really sure why the bot complains about the commit message. It seems to me that the topic is exactly 50 chars and the lines of descriprions wraps around ~50-55. 

But maybe I'm missing something ?

PR #4916, opened 2017-10-16
Fixes #21342 - fix core roles seeding

ares:
Core roles should always be up to date based on what's defined in seed files.
Users have no other way to update locked roles in newer Foreman versions.

PR #4917, opened 2017-10-16
Fixes #21343 - support multiple orgs supported for non-admin users

ares:
- [x] needs rebase after https://github.com/theforeman/foreman/pull/4878 is merged
- [x] this needs to get in https://github.com/theforeman/foreman/pull/4916
- [x] add integration tests for various combinations

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -278,6 +278,11 @@ module Api
 
     def not_found_if_nested_id_exists
       allowed_nested_id.each do |obj_id|
+        # this method does not reliably work when you have multiple parameters and some of them can be nil
+        # find_nested_object in such case returns nil (since org and loc can be nil for any context),
+        # but it detects other paramter which can have value set
+        # therefore we always skip these
+        next if [ 'organization_id', 'location_id' ].include?(obj_id)
         if params[obj_id].present?
           not_found _("%{resource_name} not found by id '%{id}'") % { :resource_name => obj_id.humanize, :id => params[obj_id] }
           return
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -348,6 +348,9 @@ module Api
       end
 
       return nil if parent_name.nil? || parent_class.nil?
+      # for admin we don't want to add any context condition, that would fail for hosts since we'd add join to
+      # taxonomy table without any condition, inner join would return no host in this case
+      return nil if User.current.admin? && [ Organization, Location ].include?(parent_class) && parent_id.blank?
       # for taxonomies, nil is valid value which indicates, we need to search in Users all taxonomies
       return [parent_name, User.current.my_organizations] if parent_class == Organization && parent_id.blank?
       return [parent_name, User.current.my_locations] if parent_class == Location && parent_id.blank?
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -348,6 +348,10 @@ module Api
       end
 
       return nil if parent_name.nil? || parent_class.nil?
+      # for taxonomies, nil is valid value which indicates, we need to search in Users all taxonomies
+      return [parent_name, User.current.my_organizations] if parent_class == Organization && parent_id.blank?
+      return [parent_name, User.current.my_locations] if parent_class == Location && parent_id.blank?
+
       parent_scope = scope_for(parent_class, :permission => "#{parent_permission(action_permission)}_#{parent_name.pluralize}")
       parent_scope = select_by_resource_id_scope(parent_scope, parent_class, parent_id)
       [parent_name, parent_scope]
--- a/app/controllers/concerns/application_shared.rb
+++ b/app/controllers/concerns/application_shared.rb
@@ -78,12 +78,10 @@ module ApplicationShared
     taxonomy.classify.constantize.send 'current=', determined_taxonomy
   end
 
-  def store_default_taxonomies(user)
-    ['location', 'organization'].each do |taxonomy|
-      default_taxonomy = find_default_taxonomy(user, taxonomy)
-      set_current_taxonomy(taxonomy, default_taxonomy)
-      store_taxonomy(taxonomy, default_taxonomy)
-    end
+  def store_default_taxonomy(user, taxonomy)
+    default_taxonomy = find_default_taxonomy(user, taxonomy)
+    set_current_taxonomy(taxonomy, default_taxonomy)
+    store_taxonomy(taxonomy, default_taxonomy)
   end
 
   # we want to be explicit to keep this readable
--- a/app/controllers/concerns/application_shared.rb
+++ b/app/controllers/concerns/application_shared.rb
@@ -34,15 +34,17 @@ module ApplicationShared
       determined_taxonomy = nil
 
       if api_request? # API request
-        if user.admin? && !params.has_key?("#{taxonomy}_id") # admin always uses any context, they can limit the scope with explicit parameters such as organization_id(s)
-          determined_taxonomy = nil
-        elsif params.has_key?("#{taxonomy}_id") # admin and non-admin who specified explicit context
-          determined_taxonomy = available.where(:id => params["#{taxonomy}_id"]).first
+        if params.has_key?("#{taxonomy}_id") # admin and non-admin who specified context explicitly
+          if params["#{taxonomy}_id"].blank? # the key is present and explicitly set to nil which indicates "any" context
+            determined_taxonomy = nil
+          else
+            determined_taxonomy = available.where(:id => params["#{taxonomy}_id"]).first
 
-          # in case admin asked for taxonomy that does not exist or is not accessible, we reply with 404
-          if determined_taxonomy.nil?
-            not_found _("%{taxonomy} with id %{id} not found") % { :taxonomy => taxonomy.capitalize, :id => params["#{taxonomy}_id"] }
-            return false
+            # in case admin or non-admin asked for taxonomy that does not exist or is not accessible, we reply with 404
+            if determined_taxonomy.nil?
+              not_found _("%{taxonomy} with id %{id} not found") % { :taxonomy => taxonomy.capitalize, :id => params["#{taxonomy}_id"] }
+              return false
+            end
           end
         elsif request.session["#{taxonomy}_id"].present? # non-admin who didn't not specify explicit context
           determined_taxonomy = find_session_taxonomy(taxonomy, user)
--- a/app/controllers/concerns/application_shared.rb
+++ b/app/controllers/concerns/application_shared.rb
@@ -40,38 +40,49 @@ module ApplicationShared
           else
             determined_taxonomy = available.where(:id => params["#{taxonomy}_id"]).first
 
-            # in case admin or non-admin asked for taxonomy that does not exist or is not accessible, we reply with 404
+            # in case user asked for taxonomy that does not exist or is not accessible, we reply with 404
             if determined_taxonomy.nil?
               not_found _("%{taxonomy} with id %{id} not found") % { :taxonomy => taxonomy.capitalize, :id => params["#{taxonomy}_id"] }
               return false
             end
           end
-        elsif request.session["#{taxonomy}_id"].present? # non-admin who didn't not specify explicit context
-          determined_taxonomy = find_session_taxonomy(taxonomy, user)
-        else # non-admin user who didn't specify explicit context and does not have anything in session
+        elsif session.has_key?("#{taxonomy}_id") # session with taxonomy explicitly set to id or nil (any contexxt)
+          if session["#{taxonomy}_id"].present?
+            determined_taxonomy = find_session_taxonomy(taxonomy, user) # specific id
+          else
+            determined_taxonomy = nil # explicitly set any context
+          end
+        else # user who didn't specify explicit context and does not have anything in session
           determined_taxonomy = find_default_taxonomy(user, taxonomy)
         end
       else # UI request
-        if request.session["#{taxonomy}_id"].present?
+        if session["#{taxonomy}_id"].present?
           determined_taxonomy = find_session_taxonomy(taxonomy, user)
         elsif !user.admin? && available.count == 1
           determined_taxonomy = available.first
         end
       end
 
-      store_taxonomy(taxonomy, determined_taxonomy) if determined_taxonomy.present?
+      set_current_taxonomy(taxonomy, determined_taxonomy)
+      store_taxonomy(taxonomy, determined_taxonomy) unless api_request?
     end
   end
 
+  # determined_taxonomy can be nil, which means any context
   def store_taxonomy(taxonomy, determined_taxonomy)
+    # session can't store nil, so we use empty string to represent any context
+    session["#{taxonomy}_id"] = determined_taxonomy.try(:id) || ''
+  end
+
+  def set_current_taxonomy(taxonomy, determined_taxonomy)
     taxonomy.classify.constantize.send 'current=', determined_taxonomy
-    request.session["#{taxonomy}_id"] = determined_taxonomy.id
   end
 
   def store_default_taxonomies(user)
     ['location', 'organization'].each do |taxonomy|
       default_taxonomy = find_default_taxonomy(user, taxonomy)
-      store_taxonomy(taxonomy, default_taxonomy) if default_taxonomy.present?
+      set_current_taxonomy(taxonomy, default_taxonomy)
+      store_taxonomy(taxonomy, default_taxonomy)
     end
   end
 
@@ -83,7 +94,7 @@ module ApplicationShared
 
     if default_taxonomy.present? && available.include?(default_taxonomy)
       default_taxonomy
-    elsif available.count == 1
+    elsif available.count == 1 && !user.admin?
       available.first
     else
       # no available default taxonomy and user is either admin or user with more taxonomies, nil represents "Any Context"
@@ -93,11 +104,16 @@ module ApplicationShared
 
   def find_session_taxonomy(taxonomy, user)
     available = user.send("my_#{taxonomy.pluralize}")
-    determined_taxonomy = available.where(:id => request.session["#{taxonomy}_id"]).first
+    determined_taxonomy = available.where(:id => session["#{taxonomy}_id"]).first
     # warn user if taxonomy stored in session does not exist and delete it from session (probably taxonomy has been deleted meanwhile)
     if determined_taxonomy.nil?
-      warning _("%s you had selected as your context has been deleted") % taxonomy.capitalize unless api_request?
-      request.session["#{taxonomy}_id"] = nil
+      if api_request?
+        not_found _("%{taxonomy} stored in session with id %{id} not found") % { :taxonomy => taxonomy.capitalize, :id => params["#{taxonomy}_id"] }
+        return false
+      else
+        warning _("%s you had selected as your context has been deleted") % taxonomy.capitalize unless api_request?
+      end
+      session.delete("#{taxonomy}_id")
       determined_taxonomy = find_default_taxonomy(user, taxonomy)
     end
     determined_taxonomy
--- a/app/controllers/concerns/application_shared.rb
+++ b/app/controllers/concerns/application_shared.rb
@@ -109,7 +109,7 @@ module ApplicationShared
         not_found _("%{taxonomy} stored in session with id %{id} not found") % { :taxonomy => taxonomy.capitalize, :id => params["#{taxonomy}_id"] }
         return false
       else
-        warning _("%s you had selected as your context has been deleted") % taxonomy.capitalize unless api_request?
+        warning _("%s you had selected as your context has been deleted") % taxonomy.capitalize
       end
       session.delete("#{taxonomy}_id")
       determined_taxonomy = find_default_taxonomy(user, taxonomy)
--- a/app/models/concerns/taxonomix.rb
+++ b/app/models/concerns/taxonomix.rb
@@ -146,8 +146,9 @@ module Taxonomix
 
   def set_current_taxonomy
     if self.new_record? && self.errors.empty?
-      self.locations     << Location.current     if add_current_location?
-      self.organizations << Organization.current if add_current_organization?
+      # we need to use _ids methods so that DirtyAssociations is correctly saved
+      self.location_ids += [ Location.current.id ] if add_current_location?
+      self.organization_ids += [ Organization.current.id ] if add_current_organization?
     end
   end
 
--- a/app/models/taxonomy.rb
+++ b/app/models/taxonomy.rb
@@ -27,6 +27,9 @@ class Taxonomy < ApplicationRecord
   has_many :subnets, :through => :taxable_taxonomies, :source => :taxable, :source_type => 'Subnet'
 
   validate :check_for_orphans, :unless => Proc.new {|t| t.new_record?}
+  # the condition for parent_id != 0 is required because of our tests, should validate macros fill in attribute with values and it set 0 to this one
+  # which would lead to an error when we ask for parent object
+  validate :parent_id_does_not_escalate, :if => Proc.new { |t| t.ancestry_changed? && t.parent_id != 0 && t.parent.present? }
   validates :name, :presence => true, :uniqueness => {:scope => [:ancestry, :type], :case_sensitive => false}
 
   def self.inherited(child)
@@ -240,5 +243,12 @@ class Taxonomy < ApplicationRecord
   def assign_taxonomy_to_user
     return if User.current.nil? || User.current.admin
     TaxableTaxonomy.create(:taxonomy_id => self.id, :taxable_id => User.current.id, :taxable_type => 'User')
+  end
+
+  def parent_id_does_not_escalate
+    unless User.current.can?("edit_#{self.class.to_s.underscore.pluralize}", self.parent)
+      errors.add :parent_id, _("Missing a permission to edit parent %s") % self.class.to_s
+      return false
+    end
   end
 end
--- a/app/views/taxonomies/_form.html.erb
+++ b/app/views/taxonomies/_form.html.erb
@@ -53,7 +53,8 @@
     <div class="tab-content stacked-content col-md-9">
       <div class="tab-pane active" id="primary">
         <%= base_errors_for taxonomy %>
-        <%= select_f(f, :parent_id, taxonomy.class.completer_scope(nil).where("id NOT IN (#{taxonomy.subtree_ids.join(',')})").order(:title), :id, :title, { :include_blank => true },
+        <% taxonomy_name = taxonomy.class.to_s.underscore %>
+        <%= select_f(f, :parent_id, taxonomy.class.completer_scope(nil).authorized("edit_#{taxonomy_name.pluralize}").where("id NOT IN (#{taxonomy.subtree_ids.join(',')})").order(:title), :id, :title, { :include_blank => true },
                      { :label => _('Parent'), :onchange => 'parent_taxonomy_changed(this);',
                        :help_inline => :indicator,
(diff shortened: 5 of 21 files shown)


iNecas:
@mbacovsky could you have your tests running against this patch as well?

ares:
Hold your horses, it might still change. Let's sync on Monday.

mbacovsky:
@ares let me know when the right time commes

ares:
I pushed the version that I think should work in any possible and impossible combination :-) @mbacovsky could you please run your tests against it? @iNecas I'm happy to go over all required changes with you, I kept them in separate commits in hope it will make reviewing easier.

ares:
needs revisiting broken tests

app/views/taxonomies/_form.html.erb
```diff
@@ -53,7 +53,8 @@
     <div class="tab-content stacked-content col-md-9">
       <div class="tab-pane active" id="primary">
         <%= base_errors_for taxonomy %>
-        <%= select_f(f, :parent_id, taxonomy.class.completer_scope(nil).where("id NOT IN (#{taxonomy.subtree_ids.join(',')})").order(:title), :id, :title, { :include_blank => true },
+        <% taxonomy_name = taxonomy.class.to_s.underscore %>
+        <%= select_f(f, :parent_id, taxonomy.class.completer_scope(nil).authorized("edit_#{taxonomy_name.pluralize}").where("id NOT IN (#{taxonomy.subtree_ids.join(',')})").order(:title), :id, :title, { :include_blank => true },
```
ares, line 57: this should also become a validation on taxonomy level

ares: added as a validation in last commit

iNecas:
The code looks very good, bonus points for extensive integration tests. It seems the last commit you added broke the tests http://ci.theforeman.org/job/test_develop_pr_core/15195/database=postgresql,ruby=2.2,slave=fast/testReport/junit/(root)/TaxonomyTest/test___Taxonomy_should_validate_that__name_is_case_insensitively_unique_within_the_scope_of__ancestry_and__type__/, I'm going to test the code now, but I don't expect much additional comments.

ares:
Thanks @iNecas, for some reason, the shoulda macro `should validate_uniqueness_of(:name).scoped_to(:ancestry, :type).case_insensitive` creates the taxonomy object with `parent_id` set to 0, which obviously causes problem when we call `.parent`. I tried to rewrite the test but didn't find a way. So I added extra condition to the if to skip it for objects with parent_id = 0, we should see those only in test env. Not nice but I didn't find a better way. We could also disable the test or rewrite it with factories and test the whole AR chain.

iNecas:
I'm ok with rhe current way: just put a small comment to the code on why
we're doing the check: for future generations to leave some clue

ares:
@iNecas anything else? I'd update it and squash if not.

app/models/taxonomy.rb
```diff
@@ -27,6 +27,7 @@ class Taxonomy < ApplicationRecord
   has_many :subnets, :through => :taxable_taxonomies, :source => :taxable, :source_type => 'Subnet'
 
   validate :check_for_orphans, :unless => Proc.new {|t| t.new_record?}
+  validate :parent_id_does_not_escalate, :if => Proc.new { |t| t.ancestry_changed? && t.parent_id != 0 && t.parent.present? }
```
iNecas: Please add a comment on why we need the `parent_id != 0` here.

iNecas:
Testing against my Sat 6.3 instance now. Feel free to update and squash in the mean time

ares:
@iNecas squashed with also the comment explaining `!= 0`

iNecas:
Tested and works well: I've asked @mbacovsky for running his tests against that, but it's on a good path. We should have results tomorrow.

app/controllers/concerns/application_shared.rb
```diff
@@ -33,15 +33,17 @@ def set_taxonomy
...
+          else
+            determined_taxonomy = available.where(:id => params["#{taxonomy}_id"]).first
 
-          # in case admin asked for taxonomy that does not exist or is not accessible, we reply with 404
-          if determined_taxonomy.nil?
-            not_found _("%{taxonomy} with id %{id} not found") % { :taxonomy => taxonomy.capitalize, :id => params["#{taxonomy}_id"] }
-            return false
+            # in case admin or non-admin asked for taxonomy that does not exist or is not accessible, we reply with 404
+            if determined_taxonomy.nil?
+              not_found _("%{taxonomy} with id %{id} not found") % { :taxonomy => taxonomy.capitalize, :id => params["#{taxonomy}_id"] }
```
mbacovsky: In our API I believe it is more common to issue 422 (unprocessable entity) in such case

mbacovsky:
I've run my hammer tests and found out that creating user Does not what I'd expected:
```
# hammer --username admin --password changeme user create --admin true --mail some.user@email.com --password passwd --auth-source-id 1 --locations "Default Location" --login test_admin_with_one_org  --organizations org_1_00001
User [test_admin_with_one_org] created
# hammer user info --login test_admin_with_one_org
Id:                    14
Login:                 test_admin_with_one_org
Name:                   
Email:                 some.user@email.com
Admin:                 yes
Last login:            
Authorized by:         Internal
Effective admin:       yes
Locale:                default
Timezone:              default
Description:           
Default organization:  
Default location:      
Roles:                 
    Default role
User groups:           

Inherited User groups: 

Locations:             
    Default Location
Organizations:         
    Default Organization
    org_1_00001
Created at:            2017/11/27 23:48:35
Updated at:            2017/11/27 23:48:35
```
The user's organizations contain "Default Organization" that I did not requested.

This was sent to the API and looks okay on the Hammer side:
```
[ INFO 2017-11-27 18:55:04 API] POST /api/users
[DEBUG 2017-11-27 18:55:04 API] Params: {
    "user" => {
                   "login" => "test_admin_with_one_org_2",
                    "mail" => "some.user@email.com",
                   "admin" => true,
                "password" => "***",
          "auth_source_id" => 1,
            "location_ids" => [
            [0] 2
        ],  
        "organization_ids" => [
            [0] 3
        ]   
    }   
}  
```

iNecas:
Did you user had a default organization? I believe that's how it's meant to be. If you dont want the user to use the default org, you need to pass nil value to the `organization_id`, or using the id you want, in `organization_id`. The goal, as @ares  described this, is to have the behavior close to ui one

app/controllers/concerns/application_shared.rb
ares: I'm not sure I understand, do you return 422 when object can't be found?

note that `not accessible` means privileges escalation and for unauthorized user we should return 404 to avoid information leak about not accessible org ids

mbacovsky: If I understand it correctly 404 would suite the situation where the API URI is not correct (e.g. when the non-existing org_id is part of the path) and the resource can't be found. If I get it right here it is about validity of params. The URI exists and the params use correct syntax though semantically wrong.

Excerpt from https://www.keycdn.com/support/422-unprocessable-entity/
```
The 422 (Unprocessable Entity) status code means the server 
understands the content type of the request entity 
(hence a 415(Unsupported Media Type) status code is inappropriate), 
and the syntax of the request entity is correct 
(thus a 400 (Bad Request) status code is inappropriate) 
but was unable to process the contained instructions.
```

ares:
@mbacovsky @iNecas This commit adds the validation that we discussed yesterday, also it defines the new API param for all resources. Let's see how many tests would be affected by the change.

app/models/concerns/taxonomix.rb
```diff
@@ -145,12 +156,17 @@ def scope_by_taxable_ids(scope)
   end
 
   def set_current_taxonomy
-    if self.new_record? && self.errors.empty?
```
iNecas: After this change, the `set_current_taxonomy` would not even be used.

app/models/concerns/taxonomix.rb
```diff
@@ -145,12 +156,17 @@ def scope_by_taxable_ids(scope)
...
-      self.organizations << Organization.current if add_current_organization?
+    if self.new_record? && self.errors.empty? && self.organization_ids.blank?
+      self.location_ids += [ Location.current.id ] if add_current_location?
+      self.organization_ids += [ Organization.current.id ] if add_current_organization?
     end
   end
 
+  def validate_taxonomies_selection
+    errors.add :organization_ids, _('The selection must include current organization') if Organization.current && !self.organization_ids.include?(Organization.current.try(:id))
+    errors.add :organization_ids, _('The selection must include current location') if Location.current && !self.location_ids.include?(Location.current.try(:id))
```
iNecas: `s/organization_ids/locations_ids/`

iNecas:
I'm ok with the second change: however, it seems there could be some challenges with the tests + hammer support, therefore, I would suggest merging everything up to the last working commit, and addressing this as a separate issue of raising an error, vs. silently adding current taxonomies, when saving objects.

mbacovsky:
@iNecas, @ares are you able to list hosts in any context using API? 
```
   curl -X GET -s -k -u admin:changeme "https://centos7-katello-tax-tests.example.com/api/v2/hosts?organization_id&location_id" | python -mjson.tool
```
my admin user is the default one set as :
```
# hammer user info --login admin
Id:                    4
Login:                 admin
Name:                  Admin User
Email:                 root@example.com
Admin:                 yes
Last login:            2017/11/30 00:00:23
Authorized by:         Internal
Effective admin:       yes
Locale:                default
Timezone:              default
Description:           
Default organization:  Default Organization
Default location:      
Roles:                 
    Default role
User groups:           

Inherited User groups: 

Created at:            2017/11/28 10:34:01
Updated at:            2017/11/28 10:34:08
```

ares:
@mbacovsky seems to work for me, note that you should have = after parameteres, in your case it should be "https://centos7-katello-tax-tests.example.com/api/v2/hosts?organization_id=&location_id="

mbacovsky:
@ares, interesting. With the `=` I'm getting empty strings instead of `nil`s. Neither works for me though.

```
 2017-11-30T11:57:21 e27ad33b [app] [I] Started GET "/api/v2/hosts?organization_id&location_id" for 192.168.121.44 at 2017-11-30 11:57:21 +0000
2017-11-30T11:57:21 e27ad33b [app] [I] Processing by Api::V2::HostsController#index as JSON
2017-11-30T11:57:21 e27ad33b [app] [I]   Parameters: {"organization_id"=>nil, "location_id"=>nil, "apiv"=>"v2"}
```

```
2017-11-30T11:56:22 de994ce9 [app] [I] Started GET "/api/v2/hosts?organization_id=&location_id=" for 192.168.121.44 at 2017-11-30 11:56:22 +0000
2017-11-30T11:56:22 de994ce9 [app] [I] Processing by Api::V2::HostsController#index as JSON
2017-11-30T11:56:22 de994ce9 [app] [I]   Parameters: {"organization_id"=>"", "location_id"=>"", "apiv"=>"v2"}
```

app/controllers/concerns/application_shared.rb
mbacovsky: After the discussion I see the code is handling taxonomies in path as well as in params and there is no good way to distinguish them. 404 for both is okay then.

mbacovsky:
@ares this fixed the listing of hosts in API and the org/loc selector in UI behaves a bit better now. 

As a non-admin (two orgs, one default, one loc one default):
 - can select Any org in the selector
 - can not select Any loc in the selector even though it is in the list. After selecting Any loc it changes both org and loc to the defaults

As an admin: (one org (default), one loc (no default)):
 - selecting Any org or Any loc in the selector also changes the other one to the default. Interesting is, that after 2 or three attempts it is possible to have both 'Any' selected properly. 

I'm going to update my hammer tests

mbacovsky:
@ares, some more observations:
- admin user with one org and no default org has set any org as current context, feels odd as an exception, but acceptable
- I'm still hitting the issue that in the UI the context is silently changed (your theory about notifications seems right). It is annoying and needs to be fixed
- It is hard to debug. When in Hammer (sometimes in the UI too) I'm not getting results I expect it is not easy to spot what is wrong. I can imagine it will be pain for users and support. I'd suggest to:
   - add more logging on server side - current org/loc with INFO level, maybe also where it comes form in DEBUG (session/default/one taxonomy/api params)
   - would be great if current taxonomy could get to the API response somehow. So that we can show it in hammer. Would putting it in response headers too evil?
   - might be useful to add some tool/call to analyze the taxonomies e.g.  `permission analyze --resource-type domain --id 27 --as-user-id 23` showing what relevant perms the user has, in what taxonomies the resource and the user is, what would be the current context. I've been doing this for two days manually which could have influenced my feeling about necessity of such tool ;)

Finally my tests passed. There were some (non-trivial) changes in hammer needed. I need to review them and add tests and test against 6.3. All and all, good to go from me when the notifications issue is fixed.

Note: I'm still seeing issues like this when creating product:
```
               create product
                  hammer --username user_two_orgs_default_00015 --password pas ...    [command #85]
                  Could not create the product:
                    Task 753bfdb5-6a68-4419-85e6-a1a80a97adc0: ActiveRecord::RecordNotUnique: PG::Error: ERROR:  duplicate key value violates unique constraint "index_katello_pools_on_cp_id"
                    DETAIL:  Key (cp_id)=(4028f9ac6006c63001602dd744c600e1) already exists.
                    : INSERT INTO "katello_pools" ("cp_id", "created_at", "updated_at", "subscription_id", "quantity", "start_date", "end_date", "consumed", "pool_type") VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9) RETURNING "id"
                  [FAIL]  returns ok
               update product
                  hammer --username user_two_orgs_default_00015 --password pas ...    [command #86]
                  Could not update the product:
                    Required lock is already taken by other running tasks.
                    Please inspect their state, fix their errors and resume them.
                  
                    Required lock: read
                    Conflicts with tasks:
                    - https://centos7-katello-tax-tests.pichi.example.com/foreman_tasks/tasks/753bfdb5-6a68-4419-85e6-a1a80a97adc0
                  [FAIL]  returns ok
```
Despite the message the product is created (at least in the foreman DB). Does not seem to be related to this taxonomies work. I'm investigating it

ares:
@tstrachota @mbacovsky the last commit implements what we discussed today, API does not store taxonomy into session, we distinguish between no session and session with explicit context (note we can't store nil into session, so we use empty string as value)

mbacovsky:
@ares the hammer tests look good. I'm not able to login in the UI though. Though the patch applied cleanly, db:seed went well, hammer seems to work the UI doesn't. 
```
undefined local variable or method `determined_taxonomy' for #<UsersController:0x0000000e5bd408>
```
I'll retry again in the morning.

app/controllers/concerns/application_shared.rb
```diff
@@ -33,42 +33,55 @@ def set_taxonomy
...
+  def set_current_taxonomy(taxonomy, determined_taxonomy)
     taxonomy.classify.constantize.send 'current=', determined_taxonomy
-    request.session["#{taxonomy}_id"] = determined_taxonomy.id
   end
 
   def store_default_taxonomies(user)
     ['location', 'organization'].each do |taxonomy|
       default_taxonomy = find_default_taxonomy(user, taxonomy)
-      store_taxonomy(taxonomy, default_taxonomy) if default_taxonomy.present?
+      set_current_taxonomy(taxonomy, determined_taxonomy)
```
mbacovsky: @ares, seems `determined_taxonomy` is not defined

mbacovsky: works well when changed to `default_taxonomy`

mbacovsky (changes requested):
My existing Hammer tests are green. The Katello duplicity error was present before this patch.
Taxonomy selector in the UI works well now and I'm no longer seeing errors when opening resources.
When the copy/paste error in `store_default_taxonomies` is fixed I'm ready to ACK.

mbacovsky:
@ares, is moving resource from org to org valid use case? For resources which can belong only to single org it is not possible (host, katello resources).

In CLI the --organization-id is used to set the context and we have no --new-organization-id to set new org. Before applying the patch I was able to change org in host via hammer.
In UI e.g. in host the org selector is grayed out. It behaved the same before the patch was applied

ares:
@mbacovsky you can't do it for host, only through a bulk action that fixes mismatches (or fails on mismatch), I think it does not make sense for Katello either. Looking at few pages, it does not seem to be supported. If you were able to change org through hammer before and it did not check taxonomy mismatches of linked resources I'd say that was a bug. I guess that was in fact bug of API.

app/controllers/concerns/application_shared.rb
ares: yeah, I pushed the fixing commit yesterday evening, I did amend so please reapply the last patch, I was hoping to fix it before you realize :-) I see there are still some test failures so I'll look into that today, it should be small change

ares:
I have pushed the commit that should fix the rest of the tests. @mbacovsky would you mind rerunning your tests on this? I think we can consider this final is foreman tests turns green. @iNecas do you need some hand with reviewing of recent changes? I think the only major change since last discussions was that we don't store session for API requests. For that we need to explicitly separate between any context and no organization set in session. We started to store empty string for "any" context into session. Nil can't be stored so we need empty string so `session.has_key?(:organization_id)` works to distinguish. This now also work for pages like /statistics or notifications requests, because they read current taxonomies from UI session.

mbacovsky:
@ares, tests look good, UI as well. By incident I found a weird behaviour:
```
$ hammer domain list --organization "Default Organization"
```
works well. For comparsion the params as recorded in the server log:
```
2017-12-13T10:09:46 05a76949 [app] [I] Started GET "/api/domains?organization_id=1&page=1&per_page=1000" for 192.168.121.44 at 2017-12-13 10:09:46 +0000
2017-12-13T10:09:46 05a76949 [app] [I] Processing by Api::V2::DomainsController#index as JSON
2017-12-13T10:09:46 05a76949 [app] [I]   Parameters: {"organization_id"=>"1", "page"=>"1", "per_page"=>"1000", "apiv"=>"v2", "domain"=>{}}
...
2017-12-13T10:09:46 05a76949 [app] [I] Current user: foreman_admin (administrator)
2017-12-13T10:09:46 05a76949 [app] [D] Setting current user thread-local variable to foreman_admin
2017-12-13T10:09:46 05a76949 [app] [D] Setting current user thread-local variable to nil
2017-12-13T10:09:46 05a76949 [app] [I] Authorized user admin(Admin User)
2017-12-13T10:09:46 05a76949 [app] [I] Current user: admin (administrator)
2017-12-13T10:09:46 05a76949 [app] [D] Setting current user thread-local variable to admin
2017-12-13T10:09:46 05a76949 [app] [D] Setting current location thread-local variable to none
2017-12-13T10:09:46 05a76949 [app] [D] Setting current organization thread-local variable to Default Organization
2017-12-13T10:09:46 05a76949 [app] [D] Setting current location thread-local variable to none
2017-12-13T10:09:46 05a76949 [app] [D] Setting current organization thread-local variable to Default Organization
...
```

While
```
$ hammer domain list --organization "Default Organization" --organization NIL
```
returns 
```
[ERROR 2017-12-13T09:57:29 API] 404 Not Found
[DEBUG 2017-12-13T09:57:29 API] {
    "message" => "Organization not found by id '1'"
}
```
In the server log:
```
 2017-12-13T09:57:29 bb8438c3 [app] [I] Processing by Api::V2::DomainsController#index as JSON
2017-12-13T09:57:29 bb8438c3 [app] [I]   Parameters: {"location_id"=>nil, "organization_id"=>"1", "page"=>"1", "per_page"=>"1000", "apiv"=>"v2", "domain"=>{}}
....
2017-12-13T09:57:29 bb8438c3 [app] [I] Current user: foreman_admin (administrator)
2017-12-13T09:57:29 bb8438c3 [app] [D] Setting current user thread-local variable to foreman_admin
2017-12-13T09:57:29 bb8438c3 [app] [D] Setting current user thread-local variable to nil
2017-12-13T09:57:29 bb8438c3 [app] [I] Authorized user admin(Admin User)
2017-12-13T09:57:29 bb8438c3 [app] [I] Current user: admin (administrator)
2017-12-13T09:57:29 bb8438c3 [app] [D] Setting current user thread-local variable to admin
2017-12-13T09:57:29 bb8438c3 [app] [D] Setting current location thread-local variable to none
2017-12-13T09:57:29 bb8438c3 [app] [D] Setting current organization thread-local variable to Default Organization
2017-12-13T09:57:29 bb8438c3 [app] [D] Setting current location thread-local variable to none
2017-12-13T09:57:29 bb8438c3 [sql] [D]   CACHE (0.0ms)  SELECT  "taxonomies".* FROM "taxonomies" WHERE "taxonomies"."type" IN ('Organization') AND "taxonomies"."id" = $1  ORDER BY "taxonomies"."title" ASC LIMIT 1  [["id", "1"]]
2017-12-13T09:57:29 bb8438c3 [app] [D] Setting current organization thread-local variable to Default Organization
2017-12-13T09:57:29 bb8438c3 [app] [I] Filter chain halted as :find_optional_nested_object rendered or redirected
2017-12-13T09:57:29 bb8438c3 [app] [I] Completed 404 Not Found in 18ms (Views: 0.2ms | ActiveRecord: 2.2ms)
```
Any idea what could be wrong?

iNecas:
@ares I didn't get the purpose of 06dc2d6d4d639a5cdb1d9112cb6fa6521e08a328 much, so some more light on why it is needed would help (ideally via some real-time discussion)

For 9d4b07279321e0073d700b8425f554aa554526c3, I wonder if we can rely on all of these affected requests being api calls? Wouldn't be safer to use `request.xhr?` condition there?

ares:
@mbacovsky what does this command suppose to do?

`$ hammer domain list --organization "Default Organization" --organization NIL`

I see for some reason hammer in this case sends `location_id` nil so I wonder if that's the cause for some reason? Is this a regression from previous version? I could imagine that it might happen because organization is not assigned to a location or something like that, perhaps we should add unscoped to place where we load org, but that would mean it was broken before too.

ares:
@iNecas let's do this tomorrow, just so you have something to think about during the night :-)

06dc2d6 - for routes such as /organization/1/hosts. we enforce the authorization of via `User.current.my_organization` on line below. That works fine but not in case where you send `organization_id` parameter as empty string or nil. In that case, it would add no condition to AR scope but would add an INNER JOIN on taxable_taxonomies table. That fails to find any host since hosts have direct association, they don't go through taxable_taxonomies. For other resources, it worked but adding the join does not make sense. If we ask for any context, we just load data from any context (for admin only of course).

9d4b072 - I think the behavior should be the same for all API requests, including e.g. domain list. The newly added failure in case the the taxonomy was not found is safer. Imagine you want to list resources from org A (coming from session), previously, it would display objects from all your orgs. You'd get false result without any warning. It would get worse with calls that change the data. Also this should be very rare case. Btw we plan to add separate API endpoint that will allow you to switch current context explicitly, from hammer it might look like `hammer set-current-org --id 1`. That will be the only way how to modify session through API.

app/controllers/concerns/application_shared.rb
```diff
@@ -92,11 +103,16 @@ def find_default_taxonomy(user, taxonomy)
...
+    determined_taxonomy = available.where(:id => session["#{taxonomy}_id"]).first
     # warn user if taxonomy stored in session does not exist and delete it from session (probably taxonomy has been deleted meanwhile)
     if determined_taxonomy.nil?
-      warning _("%s you had selected as your context has been deleted") % taxonomy.capitalize unless api_request?
-      request.session["#{taxonomy}_id"] = nil
+      if api_request?
+        not_found _("%{taxonomy} stored in session with id %{id} not found") % { :taxonomy => taxonomy.capitalize, :id => params["#{taxonomy}_id"] }
+        return false
+      else
+        warning _("%s you had selected as your context has been deleted") % taxonomy.capitalize unless api_request?
```
iNecas: The `unless api_request?` is tautology, as we are in the `else` branch of `if api_request?`

iNecas:
On Wed, Dec 13, 2017 at 6:49 PM, Marek Hulán <notifications@github.com>
wrote:

> @iNecas <https://github.com/inecas> let's do this tomorrow, just so you
> have something to think about during the night :-)
>
> 06dc2d6
> <https://github.com/theforeman/foreman/commit/06dc2d6d4d639a5cdb1d9112cb6fa6521e08a328>
> - for routes such as /organization/1/hosts. we enforce the authorization of
> via User.current.my_organization on line below. That works fine but not
> in case where you send organization_id parameter as empty string or nil.
> In that case, it would add no condition to AR scope but would add an INNER
> JOIN on taxable_taxonomies table. That fails to find any host since hosts
> have direct association, they don't go through taxable_taxonomies. For
> other resources, it worked but adding the join does not make sense. If we
> ask for any context, we just load data from any context (for admin only of
> course).
>

I think I got it now. Understanding verification question: we do this only
for admin because that's the only case when somebody can see something from
different organization without needing to be assigned to it. Right?


> 9d4b072
> <https://github.com/theforeman/foreman/commit/9d4b07279321e0073d700b8425f554aa554526c3>
> - I think the behavior should be the same for all API requests, including
> e.g. domain list. The newly added failure in case the the taxonomy was not
> found is safer. Imagine you want to list resources from org A (coming from
> session), previously, it would display objects from all your orgs. You'd
> get false result without any warning. It would get worse with calls that
> change the data. Also this should be very rare case. Btw we plan to add
> separate API endpoint that will allow you to switch current context
> explicitly, from hammer it might look like hammer set-current-org --id 1.
> That will be the only way how to modify session through API.
>
Although my original question was about something else, thanks for
explanation. I agree with everything I understand in this. I will ask about
the things
I don't understand tomorrow. In general, I think we are close to close this
one out.

mbacovsky:
@ares sorry for the confusion, totally my fault. I copy-pasted wrong command. The right failing command is : 
```
hammer -d domain list --organization "Default Organization" --location NIL
```
which matches the logs I've attached before. The use case is to set any location for admin user, which make sense when the admin has default location.

ares:
@inecas you were right indeed with the current context no longer necessary for notifications request as it's set in UI session now, I remember I reverted it but not completely. Last commit removes it completely and it still works fine. Statistics page fix will be reverted separately, though it might make sense to keep it there for the case, when we want to render statistics chart for other than current context. For notifications that probably does not make sense.

app/controllers/api/base_controller.rb
```diff
@@ -278,6 +278,11 @@ def find_nested_object
 
     def not_found_if_nested_id_exists
       allowed_nested_id.each do |obj_id|
+        # this method does not reliably work when you have multiple parameters and some of them can be nil
+        # find_nested_object in such case returns nil (since org and loc can be nil for any context),
+        # but it detects other paramter which can have value set
+        # therefore we always skip these
+        next if [ 'organization_id', 'location_id' ].include?(obj_id)
```
iNecas, line 285: I think the method `find_optional_nested_object` should know, which param was used for searching for the resource, and then checking just on presence of that resource. I'm ok to accept this change only, if there will going to be additional PR that is going to fix this better way. This is too big of a hack for me.

ares: This never worked before reliably. If you look how much we use @nested_obj in our code, it shouldn't be a big concern. We can't tell what parameter was supposed to be used if the nested obj is nil, so the whole logic would need to be reimplemented. Also this code supports only one parameter for nesting while controllers specify multiple allowed nested ids. We should later remove org and loc from allowed nested ids from all controllers, but that is potentially change for all plugins. We should redesign how we use nested routes in API v3. And ideally drop this code completely.

ares:
[test foreman] foreman failures unrelated, all the rest is green

mbacovsky (approved):
All my concerns were fixed, hammer tests work as expected. @ares, thanks for your effort and patience ;)

iNecas (approved):
All of my comments are addressed. I would still suggest revisiting the `find_option_nested_object` method, as I think we are in agreement here that this method, especially in combination with `not_found_if_nested_id_exists`, doesn't work as it should.

Anyway, the changes themselves bring huge value to the taxonomies user and I don't see another reason why holding this from the users.

iNecas:
Thanks @ares and @mbacovsky!

PR #4918, opened 2017-10-16
Fixes #21325 - Updating custom facts with booleans

patchon:
As described in the issue, currently it's not possible to upload facts
with false booleans to foreman through its API. This commit simply fixes
this by not removing a fact-value that is of type boolean false.

### DIFF
--- a/app/services/structured_fact_importer.rb
+++ b/app/services/structured_fact_importer.rb
@@ -1,7 +1,7 @@
 class StructuredFactImporter < FactImporter
   def normalize(facts)
-    # Remove empty values first, so nil facts added by normalize_recurse imply compose
-    facts = facts.select { |k, v| v.present? }
+    # Make sure to remove empty values, unless it's a bool false,
+    facts = facts.reject {|k, v| v.nil? unless v.class == FalseClass }
     normalize_recurse({}, facts)
   end
 


patchon:
Any input on this would be appreciated :octocat:

app/services/structured_fact_importer.rb
```diff
@@ -1,7 +1,7 @@
 class StructuredFactImporter < FactImporter
   def normalize(facts)
-    # Remove empty values first, so nil facts added by normalize_recurse imply compose
-    facts = facts.select { |k, v| v.present? }
+    # Make sure to remove empty values, unless it's a bool false,
+    facts = facts.reject {|k, v| v.nil? unless v.class == FalseClass }
```
ares, line 4: I think you could leave the original code, just change `v.present?` to `!v.nil?`. The `present?` method returns false for things like nil, false, [] which in this case is wrong. I suppose only nils should be skipped.

ares (changes requested):
You don't need to check for nil explicitly and verify FalseClass at the same time, but it's a valid bug and a fix otherwise :-)

app/services/structured_fact_importer.rb
patchon: Hm, I'm far from an ruby expert, but using !v.nil? seems to not take care of "empty" values (that is "empty strings"). 

Following example should explain,  
```
# Running irb on a CentOS/RHEL foreman server 
$ > scl enable tfm irb 

irb(main): > require 'active_support/core_ext/object/blank'
```

Consider following structure, 
```

facts = {'bool_true' => true, 'bool_false' => false, 'struct' => {'struct_bool_true' => true, 'struct_bool_false' => false, 'struct_arr_content' => ['a','b'] } , 'arr_empty' => [] , 'nil_nil' => nil, 'empty_str' => '', "empty_hash" => {}}
```

With your suggestion, it would become, 
```

facts.select {|k, v| !v.nil? } 
=> {"bool_true"=>true, "bool_false"=>false, "struct"=>{"struct_bool_true"=>true, "struct_bool_false"=>false, "struct_arr_content"=>["a", "b"]}, "arr_empty"=>[], "empty_str"=>"", "empty_hash"=>{}}
```

Which wouldn't be correct as it allows for "empty strings". The same with my suggestion, which is funny, since I thought it was the correct fix, but I'm now realizing, its not ;) 

The correct fix, if I understand this right, should be, 
```

facts.reject {|k, v| v.blank? unless v.class == FalseClass } 
```

Which results in, 

```
=> {"bool_true"=>true, "bool_false"=>false, "struct"=>{"struct_bool_true"=>true, "struct_bool_false"=>false, "arr_content"=>["a", "b"]}}
```

Which is what we want, no? 
Everything that is "considered empty" (but not bool false), is removed. 

With that being said, I do realize that without further modification, *this code still allows* keys inside a "structured fact" to be "empty strings", which it probably shouldn't. 

Consider the following for example, 

```
facts = {'bool_true' => true, 'bool_false' => false, 'struct' => {'struct_bool_true' => true, 'struct_bool_false' => false, 'struct_empty_str' => '', 'struct_arr_content' => ['a','b'], 
'struct_arr_empty' => [] } , 'arr_empty' => [] , 'nil_nil' => nil, 'empty_str' => '', "empty_hash" => {}}
```
Which results in, 

```
=> {"bool_true"=>true, "bool_false"=>false, "struct"=>{"struct_bool_true"=>true, "struct_bool_false"=>false, "struct_empty_str"=>"", "struct_arr_content"=>["a", "b"], "struct_arr_empty"=>[]}}
```

However, that may be out of scope for my limited ruby-skills, and someone with better knowledge should probably write the code that handles that kind of input =)

ares: I was under impression we also want to import empty strings, empty arrays, empty arrays. Now when I think about it more, we have no way to represent empty arrays or hashes because of the way how we store data. I'm not sure whether empty strings are valuable information or not.

So you are perhaps right, in such case I'd simplify your code to

```ruby
facts = facts.reject {|k, v| v.blank? && v.class.is_a?(FalseClass) }
```

or keep the positive variant

```ruby
facts = facts.select { |k, v| v.present? || v.is_a?(FalseClass) }
```

rhoml: Hey @patchon any chance this can be updated?

PR #4919, opened 2017-10-16
fixes #21331 - http proxies: name is required

timogoebel:
image

cc: @ohadlevy

### DIFF
--- a/app/models/http_proxy.rb
+++ b/app/models/http_proxy.rb
@@ -10,6 +10,8 @@ class HttpProxy < ApplicationRecord
 
   has_many :compute_resources
 
+  validates :name, :presence => true
+
   validates :url, :presence => true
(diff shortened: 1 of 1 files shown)


ohadlevy (approved):

ohadlevy:
thanks @timogoebel

ares:
Thanks @ohadlevy for merging, please don't forget to set the release in redmine. I set it to 1.17 for you.

sean797:
I think it been mentioned before, but we should automate that to he next major release. I’ll take a look if I have time.

ares:
@sean797 that would be awesome, the prprocessor would need to find out what version is the last one and ideally add a comment to github of what it set. Note that prprocessor is not triggered after PR is merged, the issue is updated once redmine sees the change in the repo. Since it was not trivial change, no one has fixed it yet.

PR #4922, opened 2017-10-16
fixes #21330 - http proxies: test connection clears toasts

timogoebel:
Removed the form validation from the test connection button. Added a short connection timeout so the user's UI does not get blocked for ages when no proxy is configured.
This also clears the toast notifications when a user clicks the test connection button.

cc: @ohadlevy , @jlsherrill

### DIFF
--- a/app/models/http_proxy.rb
+++ b/app/models/http_proxy.rb
@@ -26,9 +26,13 @@ class HttpProxy < ApplicationRecord
   end
 
   def test_connection(url)
-    raise self.errors.full_messages.join("\n") unless self.valid?
-
-    RestClient::Request.execute(method: :head, url: url, proxy: full_url)
+    RestClient::Request.execute(
+      method: :head,
+      url: url,
+      proxy: full_url,
+      timeout: 5,
+      open_timeout: 5
+    )
   rescue Excon::Error::Socket => e
     e.message
   end
--- a/webpack/assets/javascripts/foreman_http_proxies.js
+++ b/webpack/assets/javascripts/foreman_http_proxies.js
@@ -1,11 +1,12 @@
 import $ from 'jquery';
-import { notify } from './foreman_toast_notifications';
+import { notify, clear } from './foreman_toast_notifications';
 
 export function testConnection(item, url) {
   let data = $('form').serialize();
 
   $('#test_connection_indicator').show();
   $(item).addClass('disabled');
+  clear();
   $.ajax({
     url: url,
     type: 'put',


ohadlevy:
TBH: I'm not happy with the fact that errors are shown via toats, but it is better than what we currently have, would you mind breaking the issue into two, so we can split this fix vs long term fix to put it in the form instead?

timogoebel:
@ohadlevy: added the validation inside the toasts again.

ohadlevy:
@timogoebel I'm not sure I'm seeing any change? maybe I wasnt clear to begin with.

Atm - regardless of form field in question, if there is a validation error - it shows up in the toast notification instead of highlighting the field in question like a regular form would do - if you click submit, you would see the error in the right place - so i kind of find that annoying and would like it to be consistent.

having said that - your PR already makes a big difference and I think it make sense to include it, but leave the other issue open so it can also be resolved (so test connection errors are only shown in the toast - and probably if the form is not valid, we wont do test connection to begin with).

jlsherrill:
@ohadlevy I would clarify that I used toast notifications as i was basing it off "ldap authentication" which also uses toast notifications (and also does not clear them)

timogoebel:
@ohadlevy:

My first version removed
```
raise self.errors.full_messages.join("\n") unless self.valid?
```

This line causes validation errors to show up in the toast notification.
To be consistent, my suggestion was to remove that. To validate the form, a user would need to save it first. Test connection would just test the connection but not validate the form.

After you asked me to split that up, I added that line again.
To be honest, I still don't know what you want me to change. :-)

ohadlevy:
@jlsherrill thanks - I understand where you are coming from, we should probably fix them both :) to me the first thing that bothered me was the fact that form errors were showing in the toast.

ohadlevy:
@rohoover what is your take on this? thanks!

timogoebel:
@ohadlevy: Shouldn't we merge this (clear toasts) and look into the other issue (toasts show form validation errors) in a separate ticket?

jlsherrill:
@ohadlevy yeah, in retrospect i agree 👍

ohadlevy:
@timogoebel this is exactly what I'm saying! :-)

for clarity - this is what I'm seeing: 
recorded

ohadlevy:
@timogoebel if you remove this line - I'll merge it
```rb
raise self.errors.full_messages.join("\n") unless self.valid?
```

Rohoover:
I would prefer validation to occur on the form instead of in the toast.

timogoebel:
@ohadlevy: Removed the line causing the form to be validated in the toast notification.

ohadlevy:
@timogoebel ping?

timogoebel:
Rebased to stop the 🐕 from barking.

PR #4923, opened 2017-10-16
fixes #18004 - increase fact value size

timogoebel:
fact_values's value column is 1073741824 bytes for postgres, but for mysql it's only 65535 bytes. This PR matches the max length to be 1073741824 bytes for every database type.

Also adds a validation to the model.

### DIFF
--- a/app/models/fact_value.rb
+++ b/app/models/fact_value.rb
@@ -35,6 +35,7 @@ class FactValue < ApplicationRecord
   scope :with_roots, -> { joins(:fact_name) }
   scope :root_only, -> { with_roots.where(:fact_names => {:ancestry => nil}) }
 
+  validates_lengths_from_database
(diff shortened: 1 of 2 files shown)


timogoebel:
[test upgrade]
[test foreman]
[test katello]

timogoebel:
Fails on postgres with `The limit on text can be at most 1GB - 1byte.`. Need to investigate.

tbrisker:
Does it make sense to have 1GB fact values? I understand in some cases 64K might not be enough, but I'd expect 16MB-1 (`mediumtext` on MySQL) to be enough.

timogoebel:
@tbrisker: Fair enough, changed to `16.megabytes - 1`.

tbrisker:
Thanks @timogoebel !

PR #4925, opened 2017-10-17
Fixes #21137 - Register reducers from plugins

xprazak2:
This allows us to register reducers from plugins and hook into different levels of our state tree.

```js
// foreman_plugin/webpack/index.js
import { injectReducer } from 'redux-injector';
import pluginState from './reducers';

injectReducer('hosts.foremanPlugin', pluginState);
```
State tree should be just a simple object (with uncombined reducers in leaves) if we want the states to be extensible. Using `combineReducers` will work as usual, but no one will be able to hook into the states that were combined manually.

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,6 +80,7 @@
     "redux": "^3.6.0",
     "redux-form": "^7.0.3",
     "redux-form-validators": "^2.0.1",
+    "redux-injector": "^0.1.0",
     "redux-logger": "^2.8.1",
     "redux-thunk": "^2.2.0",
     "seamless-immutable": "^7.0.1",
--- a/webpack/assets/javascripts/react_app/redux/index.js
+++ b/webpack/assets/javascripts/react_app/redux/index.js
@@ -1,8 +1,8 @@
 import createLogger from 'redux-logger';
 import thunk from 'redux-thunk';
-import { applyMiddleware, createStore } from 'redux';
-
-import reducer from './reducers';
+import { applyMiddleware } from 'redux';
+import { createInjectStore } from 'redux-injector';
+import rootState from './reducers';
 
 let middleware = [thunk];
 
@@ -10,8 +10,8 @@ if (process.env.NODE_ENV !== 'production' && !global.__testing__) {
   middleware = [...middleware, createLogger()];
 }
 
-const _getStore = () => createStore(
-  reducer,
+const _getStore = () => createInjectStore(
+  rootState,
   window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__(),
   applyMiddleware(...middleware),
 );
--- a/webpack/assets/javascripts/react_app/redux/reducers/hosts/index.js
+++ b/webpack/assets/javascripts/react_app/redux/reducers/hosts/index.js
@@ -1,8 +1,7 @@
-import { combineReducers } from 'redux';
 import storage from './storage';
 import powerStatus from './powerStatus';
 
-export default combineReducers({
+export default {
   storage,
   powerStatus,
-});
+};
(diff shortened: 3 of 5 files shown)


waldenraines (approved):
@xprazak2 thanks, this looks good to me and worked for my test of creating a simple store.  Was easy to use, nice approach.

ohadlevy:
Indeed it does look nice, two concerns I have are:
1. We can't use ng-redux (if we want to start moving some of the current angular code to redux).
2. We can't customize the store to use local storage.

Maybe we should consider sending a pr for those as it's indeed a nice solution

xprazak2:
I rebased, regarding the concerns:

1. ng-redux currently creates a new store, hooking into the existing one is [on the roadmap](https://github.com/angular-redux/ng-redux/issues/19). The fix will not be a one-liner, but perhaps we could try to move it along.

2. there are already multiple npm modules that connect store to (not just) local storage. Before moving forward with local storage, I think we need to decide our use cases, how often we want to save into local storage and whether whole store or just selected state subtrees should be saved.

xprazak2:
Failing [test](https://travis-ci.org/theforeman/foreman/jobs/293031465) hints at possible complications further down the road. I made [a change](https://github.com/randallknutson/redux-injector/pull/12) that should turn the js tests green.

danseethaler:
Any work on this recently? Seems like the builds have become more stable. Probably worth rebasing and testing again. We're going to need this feature is Katello asap.

danseethaler:
Just saw the [PRs](https://github.com/randallknutson/redux-injector/pulls/xprazak2) submitted by @xprazak2. Looks like redux-injector has been abandoned. The code is very light and it seems like it would be worth just pulling into our code base. Thoughts?

waldenraines:
> Looks like redux-injector has been abandoned.

What makes you say that? I briefly perused the repo and didn't see anything about it being abandoned.

xprazak2:
Current status: I rebased, but I expect tests using snapshots to fail. I submitted a patch to redux-injector 21 days ago that should take care of it, but my PR is without a response from maintainer. I did not see any mentions that the repo is abandoned/deprecated either, the oldest PR has been opened since February without any comments from maintainer. Possible ways to move this forward:

1) wait to get a response on [proposed change](https://github.com/randallknutson/redux-injector/pull/12) and get it into redux-injector
2) fork redux-injector into theforeman
3) pull it directly into our codebase

ares:
I vote for 2, it seems the project does not move anywhere since last release in [7 Nov 2016](https://github.com/randallknutson/redux-injector/commits/master). I suppose we could call it tfm-redux-injector.

It does not specify license :-/

danseethaler:
IMO there is so little code it's worth just pulling into our code base. I have an example that doesn't manipulate the store (or snapshots) and allows registering reducers async.

https://github.com/danseethaler/foreman/commit/97a0158b2420665f14e0cbd3a900239dc8aa0b91

If we did use this approach here's a gist of what the plugin implementation would look like https://gist.github.com/danseethaler/d745921bebbe5f696c4a316e38d8879f

I'm certainly up for whatever works and is simplest while also hoping to get something merged soon.

danseethaler:
Just to clarify, I'm suggesting that we don't use `redux-injector` at all and instead just write a solution ourselves. The link to my branch in my last comment is an idea of how that would work. I've tested it out with Katello it works great.

I think this is the simplest approach and I don't see any downsides but I'm certainly open to other ideas and/or concerns.

xprazak2:
The whole `redux-injector` is ~ 50 lines and I do not expect our own solution to be bigger than that, so having our own solution seems feasible. Plus, we would not have to introduce another dependency.

#5014 is the minimal working solution, so if @danseethaler is willing to reopen, we can get it in and move from there.

danseethaler:
Thanks @xprazak2, I've reopened #5014 and I'm adding some comments there.

xprazak2:
Closing in favour of #5014

PR #4927, opened 2017-10-18
Fixes #21202 - Add close button to notifications drawer

sharvit:
The notifications panel can get close now by clicking on the new x (close) button

https://trello.com/c/CGasKNOB/61-21202-add-close-button-to-notifications-drawer

### DIFF
--- a/webpack/assets/javascripts/react_app/components/notifications/notifications.scss
+++ b/webpack/assets/javascripts/react_app/components/notifications/notifications.scss
@@ -36,6 +36,29 @@
       position: static;
     }
 
+    .drawer-pf-close,
+    .drawer-pf-toggle-expand {
+      color: inherit;
+      cursor: pointer;
+      line-height: inherit;
+      padding: 2px 10px;
+      position: absolute;
+
+      &:hover,
+      &:focus {
+        color: inherit;
+        text-decoration: none;
+      }
+    }
+
+    .drawer-pf-close {
+      right: 0;
+    }
+
+    .drawer-pf-toggle-expand {
+      left: 0;
+    }
+
     .panel-group {
       flex: 1;
       display: flex;
--- a/webpack/assets/javascripts/react_app/components/notifications/__snapshots__/notifications.test.js.snap
+++ b/webpack/assets/javascripts/react_app/components/notifications/__snapshots__/notifications.test.js.snap
@@ -44,6 +44,378 @@ exports[`notifications empty state 1`] = `
 />
 `;
 
+exports[`notifications should close the notification box when click the close button 1`] = `
+<Connect(notificationContainer)
+  data={
+    Object {
+      "url": "/notification_recipients",
+    }
+  }
+  store={
+    Object {
+      "dispatch": [Function],
+      "getState": [Function],
+      "replaceReducer": [Function],
+      "subscribe": [Function],
+      Symbol(observable): [Function],
+    }
+  }
+>
+  <notificationContainer
+    data={
+      Object {
+        "url": "/notification_recipients",
+      }
+    }
+    expandGroup={[Function]}
+    expandedGroup={null}
+    hasUnreadMessages={true}
+    isDrawerOpen={true}
+    isPolling={true}
+    isReady={true}
+    notifications={
+      Object {
+        "React devs": Array [
+          Object {
+            "actions": Object {},
+            "created_at": "2017-02-23T05:00:28.715Z",
+            "group": "React devs",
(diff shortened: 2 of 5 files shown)


ohadlevy:
[ok to test]

ohadlevy:
[test upgrade]

ohadlevy:
@sharvit thanks for the PR. I think there is a bit of CSS mismatch, for example, the link on hover should not add an underline, compare for example to https://www.patternfly.org/pattern-library/communication/notification-drawer/notification-drawer-vertical-nav.html

sharvit:
Hi @ohadlevy,

I checked the css issue and found a weird behavior happen when you are working with the inspect element.
If you put your mouse on the navbar element at the inspect element tab in the debugging tools, then it will cause the underline.
It may cause many other different problems.

I made a video that show how it happen:
close-button-underline-bug

webpack/assets/javascripts/react_app/components/notifications/notifications.scss
```diff
@@ -36,6 +36,27 @@
...
+    }
+
+    .drawer-pf-close {
+      right: 0;
+    }
+
+    .drawer-pf-toggle-expand {
+      left: 0;
+    }
+
```
waldenraines, line 61: @ohadlevy @sharvit is this the styling that had to be pulled in manually?

waldenraines (commented):
I have one small question that is included in this review, otherwise this looks good to me and worked as intended.

webpack/assets/javascripts/react_app/components/notifications/notifications.scss
sharvit: @waldenraines - yes
we just copied them from here:
https://github.com/patternfly/patternfly/blob/master/src/less/notifications-drawer.less#L74-L98

I also saw that the `&:focus` is missing so i just added it with another commit.

waldenraines: Perhaps we want to either add a comment above these lines or put them into an `overrides.scss` or something so we know to remove them once http://projects.theforeman.org/issues/21139 is resolved?

ohadlevy:
The whole file should be removed once we upgrade

waldenraines:
> The whole file should be removed once we upgrade

Ah okay, I will make a note in the redmine.

webpack/assets/javascripts/react_app/components/notifications/notifications.scss
sharvit: As @ohadlevy say, the whole file should be removed after the upgrade.

It doesn't look like `patternfly` is working on a `bootstrap-v4` version, they do have a `v4.0.0-rc.1` tag but it is outdated and it is based on `boostrap-v3`.
https://github.com/patternfly/patternfly/tree/v4.0.0-rc.1

I think that a better way will be to move all the patternfly related files from webpack assets to the app assets.
Then we should split the scss to `patternfly-versions-gap` and `patternfly-extensions`

We should update `patternfly-versions-gap` every time we facing issues of outdated code.
Then we should remove the relevant code when updating `patternfly-sass`, the file itself should contain only `less to scss` converted code with comments that contain the version number and lead to the original `patternfly` code.

ohadlevy: @sharvit breaking down assets sound like a good plan, can you do it please in a separate PR? 

Is there any reason to move it from webpack back to the asset pipeline? I don't see much value in that since our long term goal is to minimize the usage of the asset pipeline.
the only reason I can think of if it hurts any existing themes / branding people might be using /cc @ShimShtein

sharvit: @ohadlevy - Yes, I am going to focus on that right now.

The only reason I don't want to have a patternfly code in the webpack assets is because patternfly itself is not a part of the webpack assets. it will make the "world of patternfly" more readable because all the imports related to patternfly will be in one place.

sharvit:
It was the last commit.
you can review and merge if everything is fine.

ohadlevy:
test failure not related - thanks @sharvit

Rohoover:
Close button should have a blue highlight (found here):

https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/notification-drawer-vertical-nav.html#0

sharvit:
@Rohoover 
I created another pr to fix this issue:
https://github.com/theforeman/foreman/pull/4952

PR #4929, opened 2017-10-19
fixes #21374 - audit host interfaces

timogoebel:
Adds auditing to host interfaces. This is required to e.g. see when the IP address of a host changes.

image

### DIFF
--- a/app/helpers/audits_helper.rb
+++ b/app/helpers/audits_helper.rb
@@ -15,6 +15,7 @@ module AuditsHelper
       else
         label = change.to_s == "[encrypted]" ? _(change.to_s) : change.to_s
     end
+    label = _('[empty]') unless label.present?
     if truncate
       label = label.truncate(50)
     else
@@ -31,11 +32,17 @@ module AuditsHelper
       when 'Puppet Class'
         (id_to_label audit.audited_changes.keys[0], audit.audited_changes.values[0]).to_s
       else
-        name = audit.auditable_name.blank? ? audit.revision.to_label : audit.auditable_name
-        name += " / #{audit.associated_name}" if audit.associated_id && !audit.associated_name.blank?
+        name = if audit.auditable_name.blank?
+                 revision = audit.revision
+                 (revision.respond_to?(:to_audit_label) && revision.to_audit_label) || revision.to_label
+               else
+                 audit.auditable_name
+               end
+        name += " / #{audit.associated_name}" if audit.associated_id && !audit.associated_name.blank? && type_name != 'Interface'
         name
     end
-  rescue
+  rescue StandardError => exception
+    Foreman::Logging.exception("Could not render audit_title", exception)
     ""
   end
 
@@ -116,15 +123,17 @@ module AuditsHelper
     type   = audited_type(audit)
     symbol = case type
                when "Host"
-                 'hdd'
+                 {:icon => 'server', :kind => 'pficon'}
                when "Hostgroup"
-                 'tasks'
+                 {:icon => 'server-group', :kind => 'pficon'}
+               when 'Interface'
+                 {:icon => 'network', :kind => 'pficon'}
                when "User"
-                 'user'
+                 {:icon => 'user', :kind => 'fa'}
                else
-                 'cog'
+                 {:icon => 'cog', :kind => 'fa'}
              end
-    content_tag(:b, icon_text(symbol, type, :class => 'icon-white'), :class => style)
+    content_tag(:b, icon_text(symbol[:icon], type, :class => 'icon-white', :kind => symbol[:kind]), :class => style)
   end
 
   def audited_type(audit)
@@ -143,6 +152,8 @@ module AuditsHelper
                     'Override Value'
                   when 'Ptable'
                     'Partition Table'
+                  when /^Nic/
+                    'Interface'
                   else
                     audit.auditable_type
                 end
--- a/app/models/concerns/audit_extensions.rb
+++ b/app/models/concerns/audit_extensions.rb
@@ -16,6 +16,7 @@ module AuditExtensions
         :host => 'Host::Base',
         :hostgroup => 'Hostgroup',
         :image => 'Image',
+        :interface => 'Nic::Base',
         :location => 'Location',
         :medium => 'Medium',
         :os => 'Operatingsystem',
@@ -48,6 +49,7 @@ module AuditExtensions
     belongs_to :search_ptables, :class_name => 'Ptable', :foreign_key => :auditable_id
     belongs_to :search_os, :class_name => 'Operatingsystem', :foreign_key => :auditable_id
     belongs_to :search_class, :class_name => 'Puppetclass', :foreign_key => :auditable_id
+    belongs_to :search_nics, -> { where('audits.auditable_type LIKE ?', "Nic::%") }, :class_name => 'Nic::Base', :foreign_key => :auditable_id
 
     scoped_search :on => [:username, :remote_address], :complete_value => true
     scoped_search :on => :audited_changes, :rename => 'changes'
@@ -65,6 +67,9 @@ module AuditExtensions
     scoped_search :relation => :search_hostgroups, :on => :name, :complete_value => true, :rename => :hostgroup, :only_explicit => true
     scoped_search :relation => :search_hostgroups, :on => :title, :complete_value => true, :rename => :hostgroup_title, :only_explicit => true
     scoped_search :relation => :search_users, :on => :login, :complete_value => true, :rename => :user, :only_explicit => true
+    scoped_search :relation => :search_nics, :on => :name, :complete_value => true, :rename => :interface_fqdn, :only_explicit => true
+    scoped_search :relation => :search_nics, :on => :ip, :complete_value => true, :rename => :interface_ip, :only_explicit => true
+    scoped_search :relation => :search_nics, :on => :mac, :complete_value => true, :rename => :interface_mac, :only_explicit => true
 
     before_save :fix_auditable_type, :ensure_username, :ensure_auditable_and_associated_name
     before_save :filter_encrypted, :if => Proc.new {|audit| audit.audited_changes.present?}
@@ -103,10 +108,11 @@ module AuditExtensions
 
   def fix_auditable_type
     # STI Host class should use the stub module instead of Host::Base
-    self.auditable_type = "Host::Base" if auditable_type =~  /Host::/
+    self.auditable_type = "Host::Base" if auditable_type =~ /Host::/
     self.associated_type = "Host::Base" if associated_type =~ /Host::/
     self.auditable_type = auditable.type if ["Taxonomy", "LookupKey"].include?(auditable_type) && auditable
     self.associated_type = associated.type if ["Taxonomy", "LookupKey"].include?(associated_type) && associated
+    self.auditable_type = auditable.type if auditable_type =~ /Nic::/
   end
 
   def ensure_auditable_and_associated_name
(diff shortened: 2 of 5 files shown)


ares:
Looking forward to test this one. I wonder if STI works correctly here, recently we fixed templates where we had this case. From looking at screens it looks like a great addition!

timogoebel:
> I wonder if STI works correctly here

@ares: Me too, after I read your comment. I think it should work fine but if you have any hints or I missed anything: Please do let me know :-)

app/helpers/audits_helper.rb
```diff
@@ -116,15 +121,17 @@ def audited_icon(audit)
...
                when "Hostgroup"
-                 'tasks'
+                 'server-group'
+               when 'Interface'
+                 'network'
                when "User"
                  'user'
                else
-                 'cog'
+                 'settings'
```
ares: this changes the icon for all other audits, I don't disagree with the change but is it the best icon to represent other audits?

before
before

after
after

app/models/concerns/audit_extensions.rb
```diff
@@ -65,6 +67,8 @@ def self.auditable_type_complete_values
     scoped_search :relation => :search_hostgroups, :on => :name, :complete_value => true, :rename => :hostgroup, :only_explicit => true
     scoped_search :relation => :search_hostgroups, :on => :title, :complete_value => true, :rename => :hostgroup_title, :only_explicit => true
     scoped_search :relation => :search_users, :on => :login, :complete_value => true, :rename => :user, :only_explicit => true
+    scoped_search :relation => :search_nics, :on => :name, :complete_value => true, :rename => :interface, :only_explicit => true
+    scoped_search :relation => :search_nics, :on => :ip, :complete_value => true, :rename => :interface_ip, :only_explicit => true
```
ares: could we also get a definition for mac please?

app/models/concerns/audit_extensions.rb
```diff
@@ -65,6 +67,8 @@ def self.auditable_type_complete_values
     scoped_search :relation => :search_hostgroups, :on => :name, :complete_value => true, :rename => :hostgroup, :only_explicit => true
     scoped_search :relation => :search_hostgroups, :on => :title, :complete_value => true, :rename => :hostgroup_title, :only_explicit => true
     scoped_search :relation => :search_users, :on => :login, :complete_value => true, :rename => :user, :only_explicit => true
+    scoped_search :relation => :search_nics, :on => :name, :complete_value => true, :rename => :interface, :only_explicit => true
```
ares: `interface = host.example.com` is slightly confusing, maybe better name would be interface_hostname or interface_fqdn?

app/models/nic/base.rb
```diff
@@ -62,6 +62,8 @@ class Base < ApplicationRecord
 
     belongs_to_host :inverse_of => :interfaces, :class_name => "Host::Base"
 
+    audited associated_with: :host
```
ares, line 65: we should ignore password (bmc interfaces), ideally it should use the same trick as e.g. compute_resource/image//user, see
https://github.com/theforeman/foreman/blob/develop/app/models/user.rb#L15
and https://github.com/theforeman/foreman/commit/518a2593744a7cdeaf3088f7fcf595b7cf6bd6f0 for an example

in this case we could just ignore it and hope that https://github.com/collectiveidea/audited/pull/339 will be finished soon

ares: pass_exposed

ares (changes requested):
Overall it's a great addition! I found few nits which should be easy to address.

One issue that I see is that if I create a nic without DNS name and identifier, the audit does not have a clickable name so I can't navigate to audit details. We should probably define auditale_name in Nic::Base that would fall back to some 'unnamed' string.

no_link

One bigger question is, should we differentiate interfaces types? Right now we store everything as `Nic::Base` so we can't search based on interface type. E.g. we can't find changes on BMC interfaces easily.

app/helpers/audits_helper.rb
ares: perhaps @Rohoover could suggest better generic icon or confirm this one is ok

timogoebel: Basically, I changed the icons to patternfly recommendations. If there are other suggestions, I'm happy to change them.

Rohoover: Would it be worth aligning these icons to the section of the application it affected/was initiated in... so anything host related would use the host icon found in the menu, Content would use the content icon. Just a thought.

ares: That's what it does but we have specific incons only for ~3 objects so far. This one is default if none specific icon was set. Sometimes there's not a good icon, e.g. how would we represent things like domain, foreman proxy, operating system etc.

Rohoover: Well for Operating System, it's found in the Host menu, so perhaps use the Host icon until/unless we are able to find something else.

app/models/nic/base.rb
timogoebel: @ares: This is how this looks for me:

image

timogoebel:
@ares: Thanks for the review. I think I addressed all your comments. Mind to take another look?

app/models/nic/base.rb
dLobatog: @timogoebel @ares Have you tried the latest version? I also see the password encrypted (like @timogoebel screenshot, or this if going from "no pass" to "some pass").
screenshot from 2017-11-13 10-56-32

app/views/audits/show.html.erb
```diff
@@ -4,6 +4,10 @@
                                         hash_for_host_path(:id => @audit.auditable.to_param).merge(:auth_object => @audit.auditable, :auth_action => 'view'),
                                         :title => _("Host details"),
                                         :class => 'btn btn-info') if @audit.auditable_type == 'Host::Base' && @audit.auditable %>
+<%= title_actions link_to_if_authorized(_("Associated Host"),
+                                        hash_for_host_path(:id => @audit.associated.to_param).merge(:auth_object => @audit.associated, :auth_action => 'view'),
+                                        :title => _("Host details"),
+                                        :class => 'btn btn-info') if @audit.auditable_type.match(/^Nic/) && @audit.associated_type == 'Host::Base' && @audit.associated %>
```
dLobatog: I think just `btn-default` (grey instead of dark blue) suffices, clicking on "Associated host" isn't really the main action a user should be doing here, rather just going back to check other audits.

screenshot from 2017-11-13 11-00-25

cc @Rohoover your thoughts?

dLobatog (changes requested):
@timogoebel IMO this worked pretty well after a few tests, and the new icons / labels for interface audits look slick 😄 

This won't pass tests without a rebase because of the FactoryGirl -> FactoryBot rename. After rebasing, I think it should be merged for 1.17. Any qualms @ares ?

timogoebel:
* rebased
* changed the "associated host" button from primary to default
* empty values are not rendered as `[empty]`
image

app/helpers/audits_helper.rb
```diff
@@ -32,10 +33,11 @@ def audit_title(audit)
         (id_to_label audit.audited_changes.keys[0], audit.audited_changes.values[0]).to_s
       else
         name = audit.auditable_name.blank? ? audit.revision.to_label : audit.auditable_name
-        name += " / #{audit.associated_name}" if audit.associated_id && !audit.associated_name.blank?
+        name += " / #{audit.associated_name}" if audit.associated_id && !audit.associated_name.blank? unless type_name == 'Interface'
```
tbrisker: Could you please combine this condition into the same if? i.e. 
```ruby
if audit.associated_id && !audit.associated_name.blank? && type_name != 'Interface'
```

timogoebel: Ouch, sure.

ares:
It looks like test failures are related, @timogoebel could you please take a look?

timogoebel:
Fixed the tests.

```diff
diff --git a/app/helpers/audits_helper.rb b/app/helpers/audits_helper.rb
index d7c155a2b..ff3a8d496 100644
--- a/app/helpers/audits_helper.rb
+++ b/app/helpers/audits_helper.rb
@@ -32,7 +32,12 @@ module AuditsHelper
       when 'Puppet Class'
         (id_to_label audit.audited_changes.keys[0], audit.audited_changes.values[0]).to_s
       else
-        name = audit.auditable_name.blank? ? audit.revision.to_label : audit.auditable_name
+        name = if audit.auditable_name.blank?
+                 revision = audit.revision
+                 revision.try(:to_audit_label) || revision.to_label
+               else
+                 audit.auditable_name
+               end
         name += " / #{audit.associated_name}" if audit.associated_id && !audit.associated_name.blank? && type_name != 'Interface'
         name
     end
diff --git a/app/models/nic/base.rb b/app/models/nic/base.rb
index de82eb28f..95accd371 100644
--- a/app/models/nic/base.rb
+++ b/app/models/nic/base.rb
@@ -199,7 +199,7 @@ module Nic
       ip_value.present? && public_send(subnet_field).contains?(ip_value)
     end

-    def to_label
+    def to_audit_label
       return "#{name} (#{identifier})" if name.present? && identifier.present?
       return "#{mac} (#{identifier})" if mac.present? && identifier.present?
       [mac, name, identifier, _('Unnamed')].detect(&:present?)
```

dLobatog (approved):

dLobatog:
@ares OK to merge this now then? Now it shows [empty] so I think it solves the problem you mentioned in your review.

ares:
> @ares OK to merge this now then? Now it shows [empty] so I think it solves the problem you mentioned in your review.

no it is not OK, empty means the password was empty, but try to set some new password for BMC nic, see my screen, tested with latest version, to reproduce
1. have a host with bmc device
2. edit it's nic, set new password
3. save the host
4. see audit list

password

app/models/nic/base.rb
ares: I wonder how our steps differ, but with I'm still able to reproduce with the latest version, I believe we should fix it the same way as I linked in previous comment at least to be consistent with other passwords in other objects

app/helpers/audits_helper.rb
```diff
@@ -31,11 +32,17 @@ def audit_title(audit)
       when 'Puppet Class'
         (id_to_label audit.audited_changes.keys[0], audit.audited_changes.values[0]).to_s
       else
-        name = audit.auditable_name.blank? ? audit.revision.to_label : audit.auditable_name
-        name += " / #{audit.associated_name}" if audit.associated_id && !audit.associated_name.blank?
+        name = if audit.auditable_name.blank?
+                 revision = audit.revision
+                 revision.try(:to_audit_label) || revision.to_label
```
ares: I suppose you added this method to avoid problems in orchestration tests, but not every model has `to_audit_label` method defined, try would fail with `NoMethodError` here, it does not render 500 since there's the rescue below. We should not use exceptions for program flow, I think it's better to first verify the object `respond_to?(:to_audit_label)` or define it in application record (does not feel right).

timogoebel:
@ares: Thanks for testing this again. Unfortunately, I am unable to reproduce this.
In the screenshot you provided, the password is not shown as encrypted. It is for me.

image

The encrypted value should be filtered in the audit record. This is happening here:
https://github.com/theforeman/foreman/blob/1f6a9b2c84ae629dda55330b7a4d8f9ed4e948dc/app/models/concerns/audit_extensions.rb#L88-L97

I guess, that this is not working correctly in your development environment. But I have a feeling, this is worth a separate Redmine Issue & PR.

ares:
ok, that's fair, please allow me some time to dig into this a bit deeper, I hope to get to it in 1 or 2 days, if not, I won't block this PR

ares (approved):

ares:
Sorry for false positive, my encryption key was not correctly configured, I messed up my ENV variable setup. After I fixed it, I see it works fine. As an apologize, I'm going to merge this :-)

Thanks @timogoebel, very nice addition!

PR #4930, opened 2017-10-19
fixes #21394 - user login with access token api

timogoebel:
API-only implementation for https://github.com/theforeman/foreman/pull/4597

The UI is written using react and will be added in a separate PR.


app/models/auth_source.rb
```diff
@@ -60,6 +60,11 @@ def can_set_password?
   def update_usergroups(login)
   end
 
+  # Does the user exist?
+  def valid_user?(name)
+    true
```
ohadlevy: why default to true? (vs false?)

app/models/concerns/expirable.rb
```diff
@@ -0,0 +1,15 @@
...
+  included do
+    scope :expired, -> { where('expires_at <= ?', Time.current) }
+  end
+
+  def expires?
+    expires_at.present?
+  end
+
+  def expired?
+    expires? && expires_at <= Time.current
```
ohadlevy: Time.utc ?

.rubocop_todo.yml
```diff
@@ -298,7 +298,7 @@ Metrics/LineLength:
 # Offense count: 461
 # Configuration parameters: CountComments.
 Metrics/MethodLength:
-  Max: 169
+  Max: 172
```
ohadlevy: we should decrease it  - no increase it

app/models/personal_access_token.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  scope :active, -> { where(revoked: false).where("expires_at >= ? OR expires_at IS NULL", Time.current) }
+  scope :inactive, -> { where(revoked: true).or(where("expires_at < ?", Time.current)) }
+
+  attr_accessor :token_value
+
+  def self.authenticate_user(user, token)
+    tokens = self.active.where(:user => user, :token => encrypt_token(user, token))
+    return false unless tokens.any?
+    tokens.update_all(:last_used_at => Time.current, :updated_at => Time.current)
```
ohadlevy: could there be more than one active token?

ohadlevy (commented):
some questions in line - good job overall!

app/models/personal_access_token.rb
timogoebel: There should only be one token given that the tokes are unique.

app/models/auth_source.rb
timogoebel: You're right. False is the safer assumption.

app/models/personal_access_token.rb
ohadlevy: in that case I would change the query to ```find_by(user: user, token: ..)```

.rubocop_todo.yml
timogoebel: Well, this is because of `db/seeds.d/020-permissions_list.rb`. I think, refactoring this is out of scope of this PR.
Actually, I have to increase this even further. HoundCI does not seem to catch this correctly.

timogoebel:
@ohadlevy: Thanks. Addressed your comments (except the rubocop one :-P)

Gemfile
```diff
@@ -53,6 +53,7 @@ gem 'ruby2ruby', '2.3.2'
 gem 'dynflow', '>= 0.8.29', '< 1.0.0'
 gem 'daemons'
 gem 'get_process_mem'
+gem 'where-or' if SETTINGS[:rails] == '4.2'
```
ohadlevy, line 56: /cc @mmoll this will need to be removed as part of the rails 5.x migration too.

app/models/auth_sources/auth_source_external.rb
```diff
@@ -5,4 +5,11 @@ def auth_method_name
     "EXTERNAL"
   end
   alias_method :to_label, :auth_method_name
+
+  # assumes every user is valid
+  # as we do not do any authentication ourselves
+  def valid_user?(name)
+    return false unless name.present?
```
ohadlevy: why not simply name.present?

app/models/auth_sources/auth_source_internal.rb
```diff
@@ -10,6 +10,11 @@ def auth_method_name
   end
   alias_method :to_label, :auth_method_name
 
+  def valid_user?(name)
+    return false unless name.present?
+    self.users.unscoped.find_by_login(name).present?
```
ohadlevy:
```ruby
name && self.users.unscoped.find_by_login(name).present?
```

timogoebel:
@ohadlevy: Thanks again. Updated again.

test/models/user_test.rb
```diff
@@ -1033,4 +1033,29 @@ def editing_self_helper
...
+
+    test 'user can api login via personal access token' do
+      assert_nil token.last_used_at
+      assert_equal user, User.try_to_login(user.login, token_value, true)
+      assert_not_nil token.reload.last_used_at
+    end
+
+    test 'user can not ui login via personal access token' do
+      assert_nil User.try_to_login(user.login, token_value, false)
+    end
```
ohadlevy: 1. can you add a test that expired token doesnt work when loggin in?
2. since we allow to login to the UI using the token, should we show a message if the token has expired?
3. should we allow to use tokens for the ui?
4. should we log somewhere that a user has logged via token vs regular password?

timogoebel: I think, it does not make any sense to allow users to login to the UI with the token. It should be used for automated actions only. But if you can think of any use case, we can change that for sure.
I added the test for the expired token, will push in a second.
We already log in `User.try_to_login` when the user was authenticated with a personal access token.

ohadlevy: @timogoebel thanks, I missread your try to login method - thanks.

test/models/user_test.rb
```diff
@@ -1033,4 +1033,43 @@ def editing_self_helper
...
+        assert_nil User.try_to_login(user.login, expired_token_value, true)
+      end
+
+      test 'token is validated' do
+        token
+        assert_nil User.try_to_login(user.login, 'invalid', true)
+      end
+    end
+
+    context 'ui login' do
```
ohadlevy, line 1069: since this is actually logged in, I would appericiate if you can add an actual functional test that tries to login to both API and UI, since in the case we ever change it in the future, these tests might not break, but login/logout might.

timogoebel:
@ohadlevy : Added three tests:

UsersController -> Test that login via personal access token is not allowed
Api::V2 -> Tests that auth is possible via internal SSO with both a password and a personal access token

ohadlevy (approved):

timogoebel:
[test upgrade]

timogoebel:
[test foreman]
[test katello]

ohadlevy:
thanks @timogoebel

app/models/personal_access_token.rb
```diff
@@ -0,0 +1,62 @@
+class PersonalAccessToken < ActiveRecord::Base
```
ares, line 1: Shouldn't this inherit from ApplicationRecord?

app/models/personal_access_token.rb
```diff
@@ -0,0 +1,62 @@
...
+    token.update(last_used_at: Time.current.utc)
+    true
+  end
+
+  def self.token_salt(user)
+    Digest::SHA1.hexdigest(user.id.to_s)
+  end
+
+  def self.encrypt_token(user, token)
+    Digest::SHA1.hexdigest([token, token_salt(user)].join)
```
ares, line 34: Hashing is not an encryption, the method name is a bit misleading

ares:
Sorry for being late to the party. I found 2 nits, would it be possible to revisit these? Also do you plan to add support to hammer-cli-foreman? If not, could you please open the redmine issue?

PR #4931, opened 2017-10-19
Fixes #21396 remove using 'user.jpg' file'

amirfefer:


app/helpers/application_helper.rb
```diff
@@ -384,8 +384,8 @@ def gravatar_image(email)
   end
 
   def gravatar_html_options
-    { :onerror => "this.src='#{path_to_image('user.jpg')}'",
-      :alt     => _('Change your avatar at gravatar.com') }
+    { :onerror => "$(this).addClass('fa fa-user')",
+      :alt     => '' }
```
ohadlevy, line 388: why drop the comment? if gravatar is enabled users should know how to use it?

amirfefer: alternative text renders when error occurred while loading the image.
Now we're using icon instead, so we can use `title`attribute instead.
anyway, that text wasn't been visible for users before

ohadlevy: ACK - how hard would it be to write a test for this?

ohadlevy:
test failure unrelated, thanks @amirfefer

PR #4932, opened 2017-10-19
Fixes #21136 - Change all labels to Title Case

shiramax:


shiramax:
[test upgrade]

app/helpers/form_helper.rb
```diff
@@ -316,8 +316,8 @@ def add_label options, f, attr
     label_size = options.delete(:label_size) || "col-md-2"
     required_mark = check_required(options, f, attr)
     label = ''.html_safe + options.delete(:label)
-    if label.empty? && f.try(:object) && ((clazz = f.object.class).respond_to?(:gettext_translation_for_attribute_name))
-      label = s_(clazz.gettext_translation_for_attribute_name attr).html_safe
```
tbrisker: Why are you removing the `gettext_translation_for_attribute_name`? This is used to provide context for translations. e.g. - `Template.gettext_translation_for_attribute_name(:name) => "Template|Name"`, which is then parsed by gettext to provide the correct translation for the context (since it is possible that attributes that have the same name in one language might have different names in another depending on the context)

shiramax:
@tbrisker changed it, thanks. 
@Rohoover , 
This fix turns all the labels to capital in all languages, I'm not sure if it's correct behavior, What do you think?

shiramax:
[test foreman]

shiramax:
[test foreman]

shiramax:
[test katello]

Rohoover:
Yes, I believe this is correct. Thanks!

app/helpers/form_helper.rb
```diff
@@ -316,8 +316,8 @@ def add_label options, f, attr
...
+    if label.empty? && f.try(:object) && ((clazz =f.object.class).respond_to?(:gettext_translation_for_attribute_name))
```
boaz0: a small nit: missing a space in `(clazz =f.object.class)`. It should be `(clazz = f.object.class)`

shiramax:
@ripcurld0 , 
Done, thanks.

dLobatog (approved):
Thanks @shiramax !

PR #4937, opened 2017-10-23
Fixes #21425 - host selection alert should refer to param per_page

amirfefer:
Multiple host selection alert only mention settings[:per_page],
though pagination has been changed, now the user can change per_page via select box

This may lead to miscount hosts amount :
hosts_selection

### DIFF
--- a/app/views/hosts/_list.html.erb
+++ b/app/views/hosts/_list.html.erb
@@ -3,7 +3,7 @@
 <table class="<%= table_css_classes 'table-fixed' %>">
   <thead>
     <tr>
-      <th class="ca" width="40px"><%= check_box_tag "check_all", "", false, { :onclick => "toggleCheck()", :'check-title' => _("Select all items in this page"), :'uncheck-title'=> _("items selected. Uncheck to Clear") } %></th>
+      <th class="ca" width="40px"><%= check_box_tag "check_all", "", false, { :onchange => "toggleCheck()", :'check-title' => _("Select all items on this page"), :'uncheck-title'=> _("items selected. Uncheck to Clear") } %></th>
       <% if power_status_visible? %>
         <th class="ca" width="5%"><%= _('Power') %></th>
       <% end %>
--- a/test/integration/host_js_test.rb
+++ b/test/integration/host_js_test.rb
@@ -28,6 +28,38 @@ class HostJSTest < IntegrationTestWithJavascript
     SETTINGS[:organizations_enabled] = true
   end
 
+  describe 'multiple hosts selection' do
+    setup do
+      @entries = Setting[:entries_per_page]
+      FactoryBot.create_list(:host, 2)
+    end
+
+    teardown do
+      Setting[:entries_per_page] = @entries
+    end
+
+    test 'hosts counter should refer to per_page value first (max prespective)' do
+      Setting[:entries_per_page] = 2
+      visit hosts_path(per_page: 3)
+      check 'check_all'
+      assert page.has_text?(:all, "All 3 hosts on this page are selected")
+    end
+
+    test 'hosts counter should refer to per_page value first (min prespective)' do
+      Setting[:entries_per_page] = 3
+      visit hosts_path(per_page: 2)
+      check 'check_all'
+      assert page.has_text?(:all, "All 2 hosts on this page are selected")
+    end
+
+    test 'hosts counter should refer to setting- entries_per_page when there is no per_page value' do
+      Setting[:entries_per_page] = 3
+      visit hosts_path()
+      check 'check_all'
+      assert page.has_text?(:all, "All 3 hosts on this page are selected")
+    end
+  end
+
   describe 'edit page' do
     test 'class parameters and overrides are displayed correctly for strings' do
       host = FactoryBot.create(:host, :with_puppetclass)
--- a/app/assets/javascripts/host_checkbox.js
+++ b/app/assets/javascripts/host_checkbox.js
@@ -207,12 +207,14 @@ function update_counter() {
   if ($.foremanSelectedHosts)
     $(".select_count").text($.foremanSelectedHosts.length);
   var title = "";
-  if (item.attr("checked"))
-    title = $.foremanSelectedHosts.length + " - " + item.attr("uncheck-title");
+  if (item.prop('checked'))
+    title = pagination_metadata().per_page + " - " + item.attr("uncheck-title");
   else
     title = item.attr("check-title");
 
   item.attr("data-original-title", title );
-  item.tooltip();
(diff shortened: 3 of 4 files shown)


amirfefer:
[foreman test]

ohadlevy:
@amirfefer can you please add a test? thanks.

amirfefer:
@ohadlevy  thanks for your review.
I've added some tests

amirfefer:
[test upgrade]

amirfefer:
[test katello]

amirfefer:
[test katello]

ohadlevy:
@amirfefer please rebase - FactoryGirl was renamed to FactoryBot

amirfefer:
Thanks @ohadlevy, I've rebased.

app/views/hosts/_list.html.erb
```diff
@@ -3,7 +3,7 @@
 <table class="<%= table_css_classes 'table-fixed' %>">
   <thead>
     <tr>
-      <th class="ca" width="40px"><%= check_box_tag "check_all", "", false, { :onclick => "toggleCheck()", :'check-title' => _("Select all items in this page"), :'uncheck-title'=> _("items selected. Uncheck to Clear") } %></th>
+      <th class="ca" width="40px"><%= check_box_tag "check_all", "", false, { :onchange => "toggleCheck()", :'check-title' => _("Select all items in this page"), :'uncheck-title'=> _("items selected. Uncheck to Clear") } %></th>
```
tbrisker: While you're here, a couple of minor points:
1. check-title should be "Select all items on this page"
2. uncheck-title is never shown (perhaps we should just show "Toggle all items on this page"?) and also should be rephrased.
3. When clicking the tooltip isn't removed until another click.

I won't block on this, but perhaps it's time to make the checkboxes a react component? looks like a lot of the js code is quite old and over-complicated.

test/integration/host_js_test.rb
```diff
@@ -22,6 +22,34 @@ class HostJSTest < IntegrationTestWithJavascript
     SETTINGS[:organizations_enabled] = true
   end
 
+  describe 'multiple hosts selection' do
+    setup do
+      FactoryBot.create(:host)
+      FactoryBot.create(:host)
```
tbrisker: you can use `FactoryBot.create_list` instead to create multiple objects of the same kind at once.

test/integration/host_js_test.rb
```diff
@@ -22,6 +22,34 @@ class HostJSTest < IntegrationTestWithJavascript
...
+    end
+
+    test 'hosts counter should refer to per_page value first (max prespective)' do
+      Setting[:entries_per_page] = 2
```
tbrisker: This needs to be reset after every test to the original value, or you may have unexpected results in other tests.

test/integration/host_js_test.rb
```diff
@@ -22,6 +22,34 @@ class HostJSTest < IntegrationTestWithJavascript
...
+      visit hosts_path(per_page: 3)
+      check 'check_all'
+      assert page.has_text?(:all, "All 3 hosts on this page are selected")
```
tbrisker, line 45: if you are creating only 2 hosts how are you sure there are more then 3?  I'm assuming this works now because of the hosts fixture, but if that ever changes it will break this test in an unexpected way. Please create enough hosts for the test you need.

tbrisker (changes requested):
A few minor comments inline, thanks!

test/integration/host_js_test.rb
amirfefer: It's not related to hosts fixtures, one host is created before each test (as you can see at line 17).

amirfefer:
@tbrisker thanks for your review:

-  uncheck-title is now visible 
-  tooltip triggers when hovering and disappears after a click 
-  tests have a bit changed

tbrisker (changes requested):
Almost there @amirfefer, the only issue still remaining is if we select hosts on more then one page, then the hover shows e.g. "40 items selected..." (which is the correct number) while the notice shows "all 20 hosts on this page...".

amirfefer:
@tbrisker - I think we should consider making that page nonpersistent, by removing the cookie logic.
this cookie has created to handle multiple selections, (which the new `select all host` implementation solves) and  may cause confusion and hosts misselection

tbrisker:
@amirfefer I agree, do you want to do that as a separate PR or part of this one? 
@rohoover what do you think? currently we allow selection on multiple pages, which was introduced before we added the option to select all pages, so that people could choose more then $per_page hosts for one action. with the select all pages option, i think that could lead to unexpected results with selecting on one page, then changing to the next page, selecting some other hosts and expecting the bulk action to only apply to the hosts on the current page.

amirfefer:
@tbrisker, I prefer to separate this to another PR.

Rohoover:
I'm a little confused on the problem. Is it that the host counter is reporting the wrong number because of two conflicting implementations of multi-select?

amirfefer:
@tbrisker what do you say? should we merge this, and remove cookie implementation in another PR?

test/integration/host_js_test.rb
```diff
@@ -22,6 +22,37 @@ class HostJSTest < IntegrationTestWithJavascript
     SETTINGS[:organizations_enabled] = true
   end
 
+  describe 'multiple hosts selection' do
+    setup do
+      FactoryBot.create_list(:host, 2)
+    end
+
+    teardown do
+      Setting[:entries_per_page] = settings(:attributes16).value
```
tbrisker: Please save the previous value in the setup and restore it in the teardown, so that it is clear what happens and it isn't dependent on a fixture

waldenraines:
@amirfefer ping

amirfefer:
Rebased.
@tbrisker could we merge this ?

tstrachota:
[test katello]

tstrachota:
[test foreman]

tstrachota:
I tested the PR and apart from @tbrisker's comment about wrong count when there are some hosts selected on another page it works fine. I prefer fixing it in a separate PR.

To avoid users' confusion we could display the notice even when there's just one host selected and keep it displayed when the pages are switched.

@tbrisker if the tests run fine, are there any other blockers that would prevent from merging this?

tbrisker (approved):
LGTM, will merge once tests are green.

tstrachota:
Hmm, there are some katello failures. Not sure if related.

tbrisker:
@amirfefer can you please rebase on latest develop?

amirfefer:
@tbrisker 
rebased :)

tbrisker:
Merged, thanks @amirfefer !

PR #4938, opened 2017-10-23
Fixes #21135 - Add tooltip for overridden flag icon

shiramax:
image

### DIFF
--- a/app/views/puppetclasses/_form.html.erb
+++ b/app/views/puppetclasses/_form.html.erb
@@ -50,7 +50,7 @@
             <% @puppetclass.class_params.includes(:environments, :environment_classes, :lookup_values).each do |key| %>
               <li data-used-environments=<%= key.environments.map(&:to_s).to_json %> >
                 <a data-toggle="tab" id="pill_<%= key.to_param %>" href="#<%= key.to_param %>" title="<%= key.key.size > 43 ? key : "" %>">
-                  <div class="clip"><%= icon_text((key.override ? "flag": ""), key.to_s.tr('_',' '), :kind => 'fa') %></div>
+                  <div class="clip"><%= icon_text((key.override ? "flag": ""), key.to_s.tr('_',' '), :kind => 'fa', :title => _('Overridden')) %></div>
                 </a>
               </li>
             <% end %>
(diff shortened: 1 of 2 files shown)


app/views/puppetclasses/_form.html.erb
```diff
@@ -50,7 +50,7 @@
             <% @puppetclass.class_params.includes(:environments, :environment_classes, :lookup_values).each do |key| %>
               <li data-used-environments=<%= key.environments.map(&:to_s).to_json %> >
                 <a data-toggle="tab" id="pill_<%= key.to_param %>" href="#<%= key.to_param %>" title="<%= key %>">
-                  <div class="clip"><%= icon_text((key.override ? "flag": ""), key.to_s.tr('_',' '), :kind => 'fa') %></div>
+                  <div class="clip"><%= icon_text((key.override ? "flag": ""), key.to_s.tr('_',' '), :kind => 'fa', :title=>"Overridden") %></div>
```
ohadlevy: please make sure you extract the string, thanks :)

shiramax:
@ohadlevy , 
Done, thanks.

ohadlevy:
@orrabin can you review please?

orrabin:
lgtm. I think we should change the tooltip on the key to only appear for long keys, that would make it easier to find the flag tooltip but that could be addressed in a separate pr.

ohadlevy:
in that case, can you add a test please?

shiramax:
@orrabin, 
I created the issue: http://projects.theforeman.org/issues/21543 thanks.
@ohadlevy ,
Which test do you want to add?

ohadlevy:
@shiramax something like the fact that the string is getting shorter if its longer than...?

shiramax:
@ohadlevy , 
Thanks, I think it should be addressed in a different PR, 
I already created this PR: https://github.com/theforeman/foreman/pull/4971 .

dLobatog (changes requested):
Thanks @shiramax , I think @ohadlevy 's comment is already addressed in https://github.com/theforeman/foreman/pull/4971 , but the test isn't there. 

Can you update this PR to include a test for that? It doesn't have to be an integration test, you can simply extract the `key.size > 43` to a method in `app/helpers` and unit test the method.

shiramax:
@dLobatog , @ohadlevy ,
I added the tests, Thanks.

dLobatog (approved):

dLobatog:
Thanks @shiramax !

PR #4939, opened 2017-10-24
Fixes #21442 - Load puppetclass params on new host form

tbrisker:
This was caused by a Rails 5 change in ActionView that skips blank
data attributes [1] causing the JS code used to serialize the form[2]
to send `undefined` instead of `null` which was expected previously.

[1] https://github.com/rails/rails/commit/aeee438bf120f44a4402250fa50ef530f3c830a8
[2] https://github.com/theforeman/foreman/blob/develop/app/assets/javascripts/host_edit.js#L204

### DIFF
--- a/app/controllers/puppetclasses_controller.rb
+++ b/app/controllers/puppetclasses_controller.rb
@@ -59,8 +59,8 @@ class PuppetclassesController < ApplicationController
   private
 
   def get_host_or_hostgroup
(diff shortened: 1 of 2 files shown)


mmoll:
please rebase :)

ares:
Thanks @tbrisker, good detective job :-) Fixes the issue for me, merged as 30216c362a6c3fcbf17990fe7bd7223dd226a416

PR #4942, opened 2017-10-24
Fixes #21133 - Change the color of pficon-info to black

shiramax:
@Rohoover,
please let me know what do you feel about this:
image

I wasn't sure if you wanted to change all the icons to black.

### DIFF
--- a/app/assets/stylesheets/patternfly_and_overrides.scss
+++ b/app/assets/stylesheets/patternfly_and_overrides.scss
@@ -1,6 +1,7 @@
 @import "colors";
 
 $navbar-default-bg: $primary_color;
+$pf-blue-500: #00659c;
 
 .control-label {
   text-align: left;
@@ -247,7 +248,11 @@ div.form-group .glyphicon-info-sign {
   color: black;
 }
 
-div.form-group .pficon-info {
+form .pficon-info {
+  color: $pf-blue-500;
+}
+
+form .fa-info-circle {
   color: black;
 }
 
--- a/app/helpers/lookup_keys_helper.rb
+++ b/app/helpers/lookup_keys_helper.rb
@@ -102,7 +102,8 @@ module LookupKeysHelper
     popover('', description.prepend(warnings[:text]),
             :data => { :placement => 'top' },
             :title => _("Original value info"),
-            :icon => warnings[:icon])
+            :icon => "info-circle",
+            :kind => "fa")
   end
 
   def lookup_key_description(lookup_key, matcher, inherited_value)


Rohoover:
Can we use this color?
pf-black-700 #4d5258
Thanks!

shiramax:
we already using this color in the edit interface screen: 
image

Rohoover:
What color are we using on the text?

So I'm trying to find out what PF recommends for icons that have actions (such as the inline help). It's not quite clear.

One option could be to use pf-black  #030303 for text and then a blue highlight for the info icon (pf-blue  #0088ce)

What I'm seeing currently is grey text with black icons which I think looks accidental vs. intentional.

Thoughts?

shiramax:
@Rohoover , thanks for the input 
is this OK? 
image

and also, what to do you feel about this page: 
image

The first purpose of this issue was to change the icon to black... so should it be a difference between the Omit icon and the second info icon? 

thanks.

Rohoover:
The info help icon is a specific use case, if the icon in the inline field is the same, then it should probably remain blue.

@ares What do you think? I'd like to come to a consensus on the text color/icon color since it should be fairly universal throughout the application.

ares:
IMHO the blue color attracts too much attention, users primary action should not be reading the inline help, it's there only in case it's not clear e.g. for first time users. So I'd keep it the same color as the label has. For puppet class param value, I think it should have the same color as icons after the field (fullscreen) and perhaps should be also positioned there. Since this is not really and inline help, the icon should be changed to something that represents hierarchy. If I'm not mistaken it shows more information about where the value was derived from.

Thoughts?

Rohoover:
@ares 

I was able to get confirmation that inline help icons should be blue, that said, PatternFly has a variety of blue colors, so maybe a good compromise is to pick a blue that stands out less. 

Such as:
pf-blue-500 #00659c
pf-blue-600 #004368

As to your second point. If that icon is not inline help, then yes - let's change it. I looked up some icons:
fa-sitemap
fa-cogs

However, I don't necessarily have the best grasp on the definition of that button.

ares:
ok, over to you @shiramax :-)

orrabin:
@ares is hierarchy the best fit here? the information displays the description and where the parameter came from without showing the hierarchy order so it could be confusing to see fa-sitemap there. fa-cogs looks like settings to me which could be a good fit but it can't be edited so that might be misleading.

ares:
@orrabin you're right, I suppose the best icon would be with meaning "source", this is how it looks after clicking 
source2 

it contains additional info (which is why info icon was used I suppose) but elsewhere it means inline help

this is how it looks for other params
source

@Rohoover any icon suggestion for these tooltips? maybe some magnifier? That's used search search elsewhere though. Paperclip? Anyway, here's the list we can use http://fontawesome.io/icons/

Rohoover:
One thought I had was that if it's just extra info - we can use the same icon. Both effectively are additional info. 

If we want to do something similar but slightly different:
fa-info
fa-info-circle

Or here is one that is perhaps more metaphorical:
fa-puzzle-piece

ares:
I'm fine with second info icon or puzzle (even though that would most likely be used for plugin) or even keeping the same info icon after we listed all possible options.

shiramax:
Ok, so I changed the color of pf-info icon to pf-blue-500, and the replace the pf-ifcon in puppet parameters to fa-info-circle :

image

@Rohoover , @ares , what do you think?

shiramax:
[test katello]

Rohoover:
I'm fine with it if you are Marek.

ares:
Looks good, I haven't tested yet but 👍 based on screenshot. I'll get to testing later or someone else can pick it.

dLobatog (approved):
Thanks @shiramax @ares @Rohoover

PR #4943, opened 2017-10-24
Refs #21424 - Pin factory_girl_rails to ~> 4.8.0

ekohl:
Version 4.9.0 emits a deprecation warning which breaks the CI tests.
Rather than upgrade the stable branch it's easier and less disruptive to
pin.

### DIFF
--- a/bundler.d/test.rb
+++ b/bundler.d/test.rb
@@ -9,7 +9,7 @@ group :test do
   gem 'show_me_the_cookies', '~> 3.0', :require => false
   gem 'database_cleaner', '~> 1.3', :require => false
   gem 'launchy', '~> 2.4'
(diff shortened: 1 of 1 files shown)


ares:
Isn't it solved by https://github.com/theforeman/foreman/pull/4935 ? or do we need to fix all plugins at the same time?

ekohl:
Note that this is against 1.16-stable and not develop.

ares:
Oh, I didn't realize, makes sense :-)

timogoebel:
Makes sense. Does this need a redmine ticket?

mmoll:
"Refs #21424"?

ekohl:
Updated with the reference

tbrisker (approved):

tbrisker:
Thanks @ekohl !

PR #4946, opened 2017-10-25
Fixes #21448 - execute npm run lint from context of plugins

tstrachota:
Moves linting of plugins to a postlint script similar to how we do npm install for plugins.

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,8 @@
     "webpack-stats-plugin": "^0.1.5"
   },
   "scripts": {
-    "lint": "./node_modules/.bin/eslint -c .eslintrc webpack/ script/ $(./script/foreman_plugins_eslint.js) || exit 0",
+    "lint": "./node_modules/.bin/eslint -c .eslintrc webpack/ script/ || exit 0",
+    "postlint": "node ./script/npm_lint_plugins.js",
     "test": "node node_modules/.bin/jest",
     "test:watch": "node node_modules/.bin/jest --watchAll",
     "test:current": "node node_modules/.bin/jest --watch",
--- a/script/foreman_plugins_eslint.js
+++ b/script/foreman_plugins_eslint.js
@@ -1,3 +1,0 @@
-#!/usr/bin/env node
-
-process.stdout.write(require('./plugin_webpack_directories').getPluginDirs().paths.join(' '));
--- a/script/npm_lint_plugins.js
+++ b/script/npm_lint_plugins.js
@@ -1,0 +1,23 @@
+'use strict';
+/* eslint-disable no-var */
+
+var fs = require('fs');
+
+var childProcess = require('child_process');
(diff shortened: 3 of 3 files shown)


xprazak2:
It works for katello, but it only lints the index.js file and not the whole webpack folder:

```
[vagrant@centos-devel foreman]$ npm run lint

> TheForemanDevDeps@1.17.0 lint /home/vagrant/projects/foreman
> eslint -c .eslintrc webpack/ script/ || exit 0

// output ommited
✖ 3 problems (3 errors, 0 warnings)

> TheForemanDevDeps@1.17.0 postlint /home/vagrant/projects/foreman
> node ./script/npm_lint_plugins.js

> katello@1.0.0 lint /home/vagrant/projects/katello
> eslint --fix webpack/index.js || exit 0

// output ommited
✖ 6 problems (6 errors, 0 warnings)
```

I think you want the lint command in plugins to use the foreman's config.

All plugins must also have the lint task defined in their package.json, because we run this from their directory and not from foreman.

tstrachota:
@xprazak2 I updated the script and it checks whether the lint task is defined now.

Imho it's correct that a plugin decides what files should be checked.

xprazak2:
👍 works now, plugins without a `lint` task are skipped so there are no errors.

Ok, we'll say that it is a plugin's responsibility to add a `lint` task to its package.json.

waldenraines (approved):
This works great for me, thanks @tstrachota!

tstrachota:
Tests seem to have failed because of `factory_girl` dependency. I triggered them again.

tstrachota:
Tests are green. Any other comments to the code?

waldenraines (approved):
LGTM

dLobatog (approved):
Works great, thanks @tstrachota 

Could you add a section to the [Foreman handbook](https://github.com/theforeman/theforeman.org/blob/gh-pages/handbook.md) about plugins & linting?

danseethaler:
Sorry, I'm late to the conversation. I have a few questions about the purpose and scope of this change. Let me know if I'm missing something obvious 😊

What's the purpose of linting all the foreman plugins at once? Shouldn't each plugin manage their own linting process in their own repository? I think it would be great to try and standardize on linting rules but this seems like it would simply run the individual lint scripts in each isolated repo. I'm not sure what we gain from that.

When would `npm run lint` be executed? As far as I understand `npm run lint` is never called as part of any automated processes associated with PRs. We are only relying HoundCI to call out any linting failures. We would definitely run it locally before a commit but in that situation the linting status of plugins is irrelevant.

IMHO relying on a linting process that requires all files to be in harmony with the current style rules can make it difficult to make changes to the rules. In #4997 I recommend upgrading to `eslint-config-airbnb` but only updating the react components when we work on them. This allows us to keep our rules up to date without having to spend days researching and refactoring our react code.

The code here looks great - just trying to get a better understanding of our process. Thanks for any feedback!

PR #4947, opened 2017-10-25
Fixes #21447 - The login page should be based on patternfly

sharvit:
The `login.scss` used to be a compiled version of the patternfly `login.scss`
https://github.com/patternfly/patternfly-sass/blob/master/assets/stylesheets/patternfly/_login.scss

**This change is breaking the `foreman_theme_satellite` project so we will need to update the relevant files there.**

**Before the changes:**
screenshot-2017-10-25 login 7
screenshot-2017-10-25 login 6
screenshot-2017-10-25 login 5
screenshot-2017-10-25 login 4

**After the changes:**
screenshot-2017-10-25 login 3
screenshot-2017-10-25 login 2
screenshot-2017-10-25 login 1
screenshot-2017-10-25 login

Notice that the alert styling changed to the way that patternfly designed them for the login page:
https://github.com/patternfly/patternfly-sass/blob/master/assets/stylesheets/patternfly/_login.scss#L52-L55

### DIFF
--- a/app/assets/stylesheets/login.scss
+++ b/app/assets/stylesheets/login.scss
@@ -1,101 +1,54 @@
 @import "colors";
+@import "bootstrap/mixins";
 
-@media (min-width: 768px) {
-  .login-page #brand {
-    position: relative;
-    top: -52px;
+.login-pf {
+  body {
+    background: $login-bg-color;
+
+    .login-page {
+      height: 100%;
+      background: $login-bg-color;
+
+      #brand {
+        top: -52px;
+        height: 40px;
+
+        img {
+          height: 84px;
+          margin: 0;
+        }
+      }
+
+      .container {
+        padding-top: 0;
+
+        .form-horizontal {
+          .control-label {
+            padding-left: 20px;
+          }
+        }
+      }
+    }
   }
 }
 
-.login-page #brand img {
-  display: block;
-  height: 84px;
-  margin: 0;
-  max-width: 100%;
-}
-
-@media (min-width: 768px) {
-  .login-page #brand img {
-    text-align: left;
-  }
-}
-
-.login-page {
-  height: auto;
-  background: $primary_color;
-  background-size: 100% auto;
-  color: #fff;
-}
-
-.login-page .container {
-  background-color: rgba(0, 0, 0, 0.3);
-  clear: right;
-  padding-bottom: 40px;
-  width: auto;
-}
-
-@media (min-width: 768px) {
-  .login-page .container {
-    bottom: 13%;
-    padding-left: 80px;
-    position: absolute;
-    width: 100%;
+// This fix the transparent background issue in patternfly
+// When patternfly-sass will fix this issue we should remove this code
+// https://github.com/patternfly/patternfly/issues/804
+.login-pf .login-page .container {
+  .alert-success {
+    @include alert-variant($alert-success-bg, $alert-success-border, $alert-success-text);
   }
 
-  .login-page {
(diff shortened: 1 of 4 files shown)


amirfefer:
@sharvit 
Could you please follow Patternfly login page [recommendation](http://www.patternfly.org/pattern-library/application-framework/login-page/#/design)?
alert messages look different a bit.

sharvit:
@amirfefer 
I just opened an issue in patternfly:
https://github.com/patternfly/patternfly/issues/804

Their implementation is different than their examples.

ohadlevy:
[ok to test]

sharvit:
Hi @Rohoover,
can you please share your point of view about this issue with patternfly?

Thanks.

app/assets/stylesheets/login.scss
```diff
@@ -1,101 +1,32 @@
 @import "colors";
 
-@media (min-width: 768px) {
-  .login-page #brand {
-    position: relative;
-    top: -52px;
+.login-pf {
```
waldenraines: These styles aren't provided in the patternfly css? Or is this another occurrence of us having to manually copy over styling because of the out of date patternfly-sass project?

app/views/users/login.html.erb
```diff
@@ -1,4 +1,5 @@
 <% content_for :title, _('Login') %>
+<span id="badge"></span>
```
waldenraines, line 2: I don't see this id used in this PR, is this `<span>` necessary?

sharvit: @waldenraines 
There is no use for `span#badge` as you mention unless you install the `foreman_theme_satellite`.
Then it should put the RedHat logo img inside the `span#badge`.

Removing the `#badge` cause this error because patternfly use it as a placeholder:
screenshot-2017-10-26 login

The old version used media queries in order to override this behavior and fix this issue.
I believe we should go with the patternfly implementation and keep it as a placeholder so we have less css code.

**It means that I will need to update the `foreman_theme_satellite` project to inject the img inside the `span#badge` and not together with it before the `#login` like it does now:**

app/assets/stylesheets/login.scss
sharvit: `patternfly-sass` is up to date in this case.
The styles here represent the differences between our design to patternfly design which is a bit different.
For example, patternfly using a background image, the brand positioning and size are different and some paddings inside the container are different.

waldenraines (approved):

Rohoover:
@sharvit My old notes on this don't apply as it looks like things have changed.

sharvit:
Patternfly created a new pr to fix this issue:
https://github.com/patternfly/patternfly/pull/811

I am going to update the login page to use the traditional alerts.

sharvit:
@waldenraines, @Rohoover, @ohadlevy 
I took this opportunity to try shorter alerts, what do you think about it?

screenshot-2017-10-29 login 1
screenshot-2017-10-29 login

Rohoover:
I would stick with patternfly and the full size alerts.

waldenraines:
> I took this opportunity to try shorter alerts, what do you think about it?

I honestly like the way the shorter alerts look but I think we should be consistent with patternfly recommendations and the rest of the application.

sharvit:
Sounds reasonable.

You can review and merge, no future commit will come.

ohadlevy:
@waldenraines - please re-review.

waldenraines (approved):
Looks good to me, thanks @sharvit!

PR #4948, opened 2017-10-25
Refs #21424 - Pin factory_girl_rails to unblock CI

tbrisker:
This is a temporary workaround until we merge the change to the name
which requires changes in all plugins using factory_girl to factory_bot
in order to unblock CI.


tbrisker:
Looks like only 2.4 failed, and it seems to be unrelated- an issue with loading fog/core, possibly a fluke.

tbrisker:
[test foreman] just to be sure...

tbrisker:
yep, as I thought, 2.4 failure was a fluke, passing now.

PR #4953, opened 2017-10-25
Fixes #21462: Update babel preset to `env`

danseethaler:
Updating to the current recommended babel-preset-env to replace
the es20xx presets. This gives greater control over build targets
and makes ongoing maintenance easier.

http://projects.theforeman.org/issues/21462

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,5 @@
 {
-  "presets": ["es2015", "react"],
+  "presets": ["env", "react"],
   "plugins": [
     "transform-object-rest-spread"
   ]
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -61,7 +61,7 @@ var config = {
         options: {
           'presets': [
             path.join(__dirname, '..', 'node_modules/babel-preset-react'),
-            path.join(__dirname, '..', 'node_modules/babel-preset-es2015')
+            path.join(__dirname, '..', 'node_modules/babel-preset-env')
           ],
           'plugins': [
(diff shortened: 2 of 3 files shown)


danseethaler:
For reference: https://babeljs.io/env/

ohadlevy:
thanks @danseethaler

ekohl:
After trying to fix the build issue this created we packaged nodejs-babel-preset-env but when adding the dependency we see:

> Module build failed: ReferenceError: [BABEL] /builddir/build/BUILD/foreman-1.17.0/webpack/assets/javascripts/bundle.js: Unknown option: /usr/lib/node_modules/babel-preset-env/lib/index.js.__esModule. Check out http://babeljs.io/docs/usage/options/ for more info

Is this some minification gone wrong? Any help would be appreciated.

Full build log at http://koji.katello.org/kojifiles/work/tasks/5054/45054/build.log

danseethaler:
@ekohl I'm not quite sure what the issue would be but I'll share a couple thoughts.
- the link provided in the error message isn't quite right - it should be pointing here: http://babeljs.io/docs/usage/api/#options
- It sounds like there's an `__esModule` option being passed that isn't valid. Any idea where that might be coming from?
- Have you tried clearing out all the npm dependencies and running a fresh `npm install`?

ekohl:
@ehelms discovered that the babel-core version we have packaged was too old. Looks like npm/webpack doesn't verify the dependencies before starting so we should include checks for that in our build process.

PR #4954, opened 2017-10-25
Fixes #21465 - load owners from any context

ares:
similar fix like in https://github.com/theforeman/foreman/pull/4879, please see the redmine issue for more details

### DIFF
--- a/app/models/concerns/hostext/ownership.rb
+++ b/app/models/concerns/hostext/ownership.rb
@@ -15,6 +15,20 @@ module Hostext
       validate :owner_taxonomies_match, :if => Proc.new { |host| host.owner.is_a?(User) }
     end
 
+    def owner
+      case owner_type
+        when 'User'
+          User.unscoped { super }
+        when 'Usergroup'
+          Usergroup.unscoped { super }
+        when nil, ''
+          # not set yet, usually the case for new records
+          nil
+        else
+          raise ArgumentError, "Unknown member type #{owner_type}"
+      end
+    end
+
     # method to return the correct owner list for host edit owner select dropbox
     def is_owned_by
       owner.id_and_type if owner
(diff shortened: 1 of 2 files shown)


tbrisker:
Please rebase to get tests to run

dLobatog (commented):
Lots of "Unknown member type: " errors, the error should only be raised when trying to set (create/update) the attribute, not any time owner is called, correct? cc @ares

dLobatog (changes requested):
Back to WoC

ares:
> Lots of "Unknown member type: " errors, the error should only be raised when trying to set (create/update) the attribute, not any time owner is called, correct? cc @ares

no, owner_type should be known whenever we read owner, but there are valid cases when owner_type is unknown (nil) and we try to read it, e.g. when we build a new instance and we try to auto suggest the new owner for it

I've added a case branch for this case which should fix the tests.

mmoll (dismissed):
test failures related

ares:
reruning tests after rebase, I expect failures that I'll need to look into, this became a bigger problem for discovery when refreshing facts

ares:
[test foreman] [test upgrade] tests failures were unrelated, @dLobatog would you mind taking another look?

dLobatog (approved):
@ares Changes make sense to me, tests failures are unrelated -> Merging

xprazak2:
It takes care of the issue where `owner` method cannot find the real owner. However, I found out that you can unknowingly switch owner when editing a host in UI.

1) log in as admin
2) create john as an admin with no location/organization
3) assign john as and owner of a some-host.example.com (host is in Default Organization/Default Location)
4) create beth with manager role in Default Organization/Default Location
5) log out, log in as beth
6) edit some-host.example.com, go to 'Additional Information' tab
7) notice john is not set as owner (dropdown has a blank value set) and is not available in dropdown selection
8) Hit the submit, revisit 'Additional Information' again
9) beth is now owner of some-host

This is caused by the fact that beth is not an admin and cannot see [users outside of her taxonomies](https://github.com/theforeman/foreman/blob/develop/app/views/hosts/_form.html.erb#L126), so john is not included in dropdown selection.

Not sure if we want to handle it here or fix it separately as a UI issue.

ares:
we should handle it as a separate issue, would you mind opening it in redmine? I guess we need to add extra check, similar to what we have for taxonomies that user can't see in the form

xprazak2:
I created a redmine [issue](https://projects.theforeman.org/issues/22396)

PR #4955, opened 2017-10-26
Fixes #21131 - put help text in inline help

shiramax:


xprazak2:
Tooltip works for me:





The ticket also says something about changing the 'Reset Puppet Environment to match selected Content View' to 'Reset Puppet Environment', did we decide to keep the text as is?

dLobatog (approved):
Thanks @shiramax - 

@xprazak2 I'm going to go with merging already as the fix for the text has to be done in Katello

PR #4956, opened 2017-10-26
Fixes #21456 - fix input-field fullscreen mode

amirfefer:
vertical navigation overrides part of input field in fullscreen mode, as you can see here:
fullscreen_field

### after:
oct 26 2017 4-37 pm


tbrisker:
Thanks @amirfefer !

PR #4957, opened 2017-10-27
Fixes #21496 - Properly fetches permission records

parthaa:
The permission records method used to wrongly error out when the
permissions were nested inside an array. Even though the permission list
was returned properly the comparison to raise the PermissionMissing
exception used the wrong logic.

### DIFF
--- a/test/models/role_test.rb
+++ b/test/models/role_test.rb
@@ -229,6 +229,10 @@ class RoleTest < ActiveSupport::TestCase
       @role = FactoryGirl.build(:role, :permissions => [])
     end
 
+    it 'should fetch the appropriate permissions' do
+      assert_equal 2, @role.send(:permission_records, [[@permission1.name, @permission2.name]]).size
+    end
+
     it "should build filters with assigned permission" do
       @role.add_permissions [@permission1.name, @permission2.name.to_sym]
       assert @role.filters.all?(&:unlimited?)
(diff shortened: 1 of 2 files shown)


ehelms (approved):

test/models/role_test.rb
```diff
@@ -229,6 +229,10 @@ def test_allow_to_modify_roles_when_bypass_locking
       @role = FactoryGirl.build(:role, :permissions => [])
     end
 
+    it 'should fetch the appropriate permissions' do
+      assert_equal 2, role.permission_records([[@permission1.name, @permission2.name]]).size
```
ehelms: Should be `@role`

test/models/role_test.rb
```diff
@@ -229,6 +229,10 @@ def test_allow_to_modify_roles_when_bypass_locking
       @role = FactoryGirl.build(:role, :permissions => [])
     end
 
+    it 'should fetch the appropriate permissions' do
+      assert_equal 2, @role.permission_records([[@permission1.name, @permission2.name]]).size
```
ehelms: Since permission_records is a private method, you need to use `send(:permission_records, [[@permission1.anem.....`

parthaa:
@ehelms sorry for the delay . I tested more thoroughly this time..

ares:
Thanks @parthaa and @ehelms, fixes the problem. Merged as 6a1f0e1adcea166c15785b05989a97462fb7ec9b

PR #4959, opened 2017-10-29
Fixes #21454 - fix editor in fullscreen mode

amirfefer:
Vertical navigation overrides part of the ace editor:
editor_bug

after: 
oct 29 2017 2-35 pm

### DIFF
--- a/app/assets/stylesheets/editor.scss
+++ b/app/assets/stylesheets/editor.scss
@@ -16,7 +16,8 @@
   }
 }
 
-.fullscreen {
+.editor-container.fullscreen {
+  height: 82vh !important;
   .navbar-editor {
     margin: 0;
   }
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/base.scss
```diff
@@ -483,9 +483,9 @@ table {
 }
 
 .fullscreen {
```
amirfefer: I've changed this section already in #4956,
so I'll rebase this if #4956 will be merged first

app/assets/stylesheets/editor.scss
```diff
@@ -17,6 +17,7 @@
 }
 
 .fullscreen {
+  height: 82vh !important;
```
tbrisker: This causes all `.fullscreen` elements to get this height, overriding the value set in the `base.scss` file.

tbrisker (changes requested):

amirfefer:
@tbrisker Thanks for your review,
I've changed the css  to be more specific to editor

tbrisker (approved):
Thanks, works as expected.

tbrisker:
Thanks @amirfefer !

PR #4960, opened 2017-10-29
Fixes #21500 - dismiss notifications

sharvit:
The notification drawer should get close
when clicking outside of the box

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "react-bootstrap": "^0.31.0",
     "react-dom": "^16.0.0",
     "react-numeric-input": "^2.0.7",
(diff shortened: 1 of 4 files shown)


ohadlevy:
@rohoover - is that a desired behavior ?

iNecas:
Personally I would expect this behavior, it felt like a bug to ne from the beginning

Rohoover:
It's supposed to remain open. The idea behind this was so that when an event occurred you could compare what the drawer was telling you to what content was on your screen. I'm not personally opposed to having it close, and potentially leaving it open later when the functionality becomes more robust.

ohadlevy:
thanks @sharvit!

PR #4961, opened 2017-10-30
Fixes #21238 - Settings - show defaults on hover

swapab:
Defaults visible only when default and actual value are different

### DIFF
--- a/test/models/setting_test.rb
+++ b/test/models/setting_test.rb
@@ -600,4 +600,14 @@ class SettingTest < ActiveSupport::TestCase
     refute_empty sorted_list
     assert sorted_list.keys.exclude?(sticky_setting)
   end
+
+  test 'full_name_with_default should include default value' do
+    setting = FactoryBot.build(:setting, :name => 'foo', :value => 'baz', :default => 'boo', :description => 'test foo', :full_name => 'Foo Name')
+    assert_equal('Foo Name (Default: boo)', setting.full_name_with_default)
+  end
+
+  test 'full_name_with_default without default value' do
+    setting = FactoryBot.build(:setting, :name => 'foo', :default => '', :description => 'test foo', :full_name => 'Foo Name')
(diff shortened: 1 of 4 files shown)


dLobatog (changes requested):
@swapab I think as an user I would be less confused if I saw "default: xxxx" on hover in *all* cases, not only when the current value is something (not the default). Code-wise - 5 :star: :smile:

ccing @rohoover if you could give your opinion

test/models/setting_test.rb
```diff
@@ -600,4 +600,14 @@ def sticky_setting
     refute_empty sorted_list
     assert sorted_list.keys.exclude?(sticky_setting)
   end
+
+  test 'full_name_with_default should include default value' do
+    setting = Setting.create(:name => 'foo', :value => 'baz', :default => 'boo', :description => 'test foo', :full_name => 'Foo Name')
```
dLobatog: Can you try `Factory.build` to avoid creating an object in db (slower)?

swapab:
@dLobatog +1
I had thought of the same :smile: Let user's see default settings by default on hover.

test/models/setting_test.rb
swapab: Agreed. Done.

Rohoover:
I'm good with that.

swapab:
Why didn't the Jenkins triggered a new job after a new `git push`? New Foreman build particularly.

tbrisker:
[test foreman]

tbrisker:
@swapab can you rebase please on latest develop so tests pass?

dLobatog (approved):

dLobatog (dismissed):
Thanks @swapab , please rebase and push to get tests running again. I'll merge if 🍏

swapab:
@dLobatog rebased and it's :green_apple: !

tbrisker (approved):

tbrisker:
Thanks @swapab !

PR #4962, opened 2017-10-30
Fixes #21502 - Allow internal users in group with external groups

dLobatog:
Adding Users to User Groups with External User Groups does not
work. Currently the only feedback the user gets is that Users get
automatically removed after the form is submitted - no warning or
mention.

In order not to confuse the user or make the user waste their time
adding users, we should display a warning when there are External User
Groups linked to a User Group.

------------------------------------

This should contain tests and should not be merged as is, just opening it to keep track of it and maybe get early feedback

### DIFF
--- a/app/views/usergroups/_form.html.erb
+++ b/app/views/usergroups/_form.html.erb
@@ -9,9 +9,22 @@
   </ul>
   <div class="tab-content">
     <div class="tab-pane active" id="primary">
+      <% if @usergroup.external_usergroups.any? %>
+        <h5>
+          <%= _('Only internal (non-LDAP) users can be added manually. LDAP users
+                are automatically synced from the External Groups list.') %>
+        </h5>
+        <h5>
+          <%= _('To refresh the list of users, click on the tab "External
+                groups" then "Refresh".') %>
+        </h5>
+        <hr>
+      <% end %>
       <%= text_f f, :name %>
-      <%= multiple_checkboxes f, :usergroups, @usergroup, Usergroup, :label => _("User Groups") %>
-      <%= multiple_checkboxes f, :users, @usergroup, User.except_hidden, :label => _("Users"), :object_label_method => :select_title %>
+      <%= multiple_checkboxes f, :usergroups, @usergroup, Usergroup,
+        { :label => _("User Groups") } %>
+      <%= multiple_checkboxes f, :users, @usergroup, User.except_hidden,
+        :label => _("Users"), :object_label_method => :select_title %>
     </div>
     <div class="tab-pane" id="roles">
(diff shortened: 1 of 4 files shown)


ares (commented):
Can't we just skip users that has auth source set to INTERNAL? That way you could combine both internal and ldap users in a single user group.

dLobatog (commented):
@ares I've made it skip internal users, and it works well in the UI case, there is a warning and the user know about this behavior. The original bug mentions how this behavior is disconcerting in hammer too - should we add a warning in hammer for this command or can you think of any other option? Returning extra data in the API response doesn't seem quite right to me.

dLobatog:
Updated to include a warning in the API description and allow to add/remove internal users freely.

ares (changes requested):
The functionality works great. To be consistent with other pages, the warning should not be created using alert. We replaced it with simple `<h5>` separated with `<hr>`, see https://github.com/theforeman/foreman/blob/develop/app/views/roles/_form.html.erb#L13-L23 for an example.

Otherwise ready to merge.

dLobatog (commented):
@ares Updated to use h5 instead of alert, check out the screenshot

dLobatog:
screenshot from 2017-11-14 07-27-09

app/views/usergroups/_form.html.erb
```diff
@@ -9,9 +9,21 @@
   </ul>
   <div class="tab-content">
     <div class="tab-pane active" id="primary">
+      <% if @usergroup.external_usergroups.any? %>
+        <h5>
+          <%= _('Only internal (non-LDAP) users can be added manually. LDAP users
+                are automatically synced from the External Groups list.') %>
+          <br>
```
dLobatog: I've split this into 2 strings so that it's easier to translate.

ares:
@dLobatog it seems something went wrong and there FactoryGirl again in usergroup_test :-) 

http://ci.theforeman.org/job/test_develop_pr_core/15077/database=sqlite3,ruby=2.2,slave=fast/testReport/junit/(root)/UsergroupTest__external%20usergroups/test_0006_internal_auth_source_users_remain_after_refresh/

ares:
One more tiny thing, in role there's spacing between two warning. I'm not sure if here we can consider it as two separate warnings but I think it would be more readable if both are separate `<h5>` and it would look more consistent with role form.

0label
1label

dLobatog:
@ares Updated to use FactoryBot & use 2 h5, it looks like this now:

screenshot from 2017-11-14 09-38-01

ares:
Ack pending jenkins

ares (approved):

ares:
Thanks @dLobatog, merging!

PR #4963, opened 2017-10-30
Fixes #21503 - Import puppet classes when organzations changed

xprazak2:


app/services/puppet_class_importer.rb
```diff
@@ -254,10 +254,10 @@ def load_classes_from_json(blob)
 
   def add_classes_to_foreman(env_name, klasses)
     env         = find_or_create_env env_name
-    new_classes = klasses.map { |k| Puppetclass.where(:name => k[0]).first_or_create }
+    new_classes = klasses.map { |k| Puppetclass.unscoped.where(:name => k[0]).first_or_create }
```
dLobatog: Why unscoped? Doesn't this run on the user's context so it should be scoped by `User.current`?

app/services/puppet_class_importer.rb
```diff
@@ -330,11 +330,16 @@ def find_existing_foreman_classes(klasses = [])
...
   def find_or_create_env(env)
-    Environment.where(:name => env).first || Environment.create!(:name => env,
+    user_visible_environment(env) || Environment.create!(:name => env,
                                                                  :organizations => User.current.my_organizations,
                                                                  :locations => User.current.my_locations)
   end
 
+  def user_visible_environment(env)
+    return unless User.current.visible_environments.include? env
+    Environment.unscoped.find_by :name => env
```
dLobatog, line 342: Is this disregarding the User permissions completely?

dLobatog (changes requested):
@xprazak2 Found a couple of issues in the file I didn't understand, also it seems that tests fail because 2 classes are created on import. Thanks for the PR!

app/services/puppet_class_importer.rb
xprazak2: Running this in a scope is exactly the problem. When a puppet class with given name already exists in different scope, it fails to find it but hits the uniqueness constraint when saving a new record.

app/services/puppet_class_importer.rb
xprazak2: This is a variation of the problem above, it errors even for admin.

1) create 2 organizations, A and B
2) import puppet env as admin in context of A - it gets imported to B as well
3) unassign the newly imported env from B
4) in context of B, try to import the env you just unassigned

It uses `unscoped` only to avoid hitting the uniqueness constraint, user does not know it re-used the existing record. On the outside it looks like it was just created.

I realize this is not ideal and does not solve all the issues with imports, but at leas it should fix the complete application crashes.

dLobatog: Fair enough

app/services/puppet_class_importer.rb
dLobatog: Oh that's a good catch! Can you put it in a comment here so that we're reminded in case we modify this code again?

dLobatog (changes requested):
Thanks @xprazak2  - I think just adding a comment with your explanation about why unscoped is used in `add_classes_to_foreman` would be great. After that I'm happy to merge

xprazak2:
I added a comment explaining the unscoped searches

dLobatog (commented):
Tons of errors in the integration tests, they don't look related to this so let's trigger them again..

dLobatog (approved):

dLobatog:
[test foreman]

dLobatog (changes requested):
I don't see these errors on other PRs yet, it could be that develop is broken but so far I only see it here, could you ping me when this is passing / or with any update about this? @xprazak2

dLobatog (approved):

dLobatog:
Thanks @xprazak2 , seems like it went away. Maybe it was a problem in the new Jenkins slave or something

PR #4964, opened 2017-10-30
Fixes #16835 - Allow implicit search with org_id

tbrisker:
Previously, any resource with many taxonomies would fail when using
implicit search when passing organization_id parameter to the api. This
is due to scoped_search incorrectly adding the taxonomy table to search
on, which was conflicting with the current taxonomy scope. Removing the
implicit search on taxonomies fixes the issue while still allowing users
to explicitly search on location or organization if they want to
(although it makes no sense to use both options).


dLobatog (approved):
Works well, thanks @tbrisker !

PR #4966, opened 2017-10-30
Fixes #21353 - users can edit login if they have permissions

shiramax:


shiramax:
[test foreman]

shiramax:
[test foreman]

shiramax:
[test foreman]

Jenkins aborted the previous test

shiramax:
[test]

shiramax:
[test]

iNecas:
[test katello] [test upgrade]

app/controllers/concerns/foreman/controller/parameters/user.rb
```diff
@@ -42,15 +42,20 @@ def user_params
   end
 
   class Context < Foreman::ParameterFilter::Context
-    def initialize(type, controller_name, action, editing_self)
+    def initialize(type, controller_name, action, editing_self,user_currently_edited)
```
iNecas: nitpick: missing space after comma `, user_currently_edited)`

app/controllers/api/v2/users_controller.rb
```diff
@@ -8,7 +8,7 @@ class UsersController < V2::BaseController
       include Api::Version2
 
       wrap_parameters User, :include => user_params_filter.accessible_attributes(
-        Foreman::Controller::Parameters::User::Context.new(:api, controller_name, nil, false)) +
+        Foreman::Controller::Parameters::User::Context.new(:api, controller_name, nil, false, @user)) +
```
iNecas: Isn't the `@user` always nil here?

iNecas:
@ares in the issue, you mention we also should allow this only for internal auth source. Should we check it here as well, or do we rely on the edit_user permission?

ares:
I think changing the login for non internal admin is potentially dangerous. If I change my login @ares to @inecas and you login later, you'll get the permission set I have. So I'd prefer to add a check, yes.

dLobatog (changes requested):
@shiramax Could you add a couple of tests demonstrating the behavior? Thanks!

shiramax:
@ares , @dLobatog  
External Users might have permissions to change their login too.
Are you sure you we want to deny all external users the ability to change their login?

ares:
> External Users might have permissions to change their login too.
> Are you sure you we want to deny all external users the ability to change their login?

Adding @ekohl and @evgeni since they took part in original IRC discussion. I can imagine that as long as user has access to 2 external user account, he or she would be able to take over any other account. 

I think this should apply: Once users have edit_users permission, it's fine if they update other users including their own login. Internal users without this permission should be able to change their login. External users without this permission should not be able to update their login.

evgeni:
While I don't think allowing external users to change their login is particularly useful, I don't see it as a security issue.

Let's assume the following scenario:
* Foreman is configured with an external LDAP auth source
* In LDAP there are two users: Alice and Bob
* Only Alice exists as a user in Foreman (otherwise Alice→Bob rename in Foreman would fail)
* Alice renames the user to Bob

On next login, Alice can't login anymore, as she would need to enter Bob as the user and also Bob's LDAP password, which she does not know. Login as Alice does not work, as there is no such user in Foreman anymore.

So the only really useful scenario to rename an external user would be when the external provider also renames the user, and that is usually something that is not allowed.

ekohl:
> I think this should apply: Once users have edit_users permission, it's fine if they update other users including their own login. Internal users without this permission should be able to change their login. External users without this permission should not be able to update their login.

That makes sense to me. Changing your own username is something you rarely see in applications and starting with requiring the ```edit_users``` permission is good.

Another thing to consider is that users could accidentally break their logins because they assumed they could requiring an admin to fix it.

ares:
> On next login, Alice can't login anymore, as she would need to enter Bob as the user and also Bob's LDAP password, which she does not know. Login as Alice does not work, as there is no such user in Foreman anymore.

If Alice change her name to Bob and triggers external user group syncs or wait for cron to do it, she would be granted all Bob's user groups which can grant additional permissions since user group can be linked to role.

evgeni:
Ah, dang, did not think about the currently alive session. Makes sense.

ekohl:
@shiramax 
> Are you sure you we want to deny all external users the ability to change their login?

I think with the above reasoning I think we are sure we want to deny that.

app/controllers/api/v2/users_controller.rb
shiramax: should I change it to nil?

shiramax:
@ares , @evgeni , @ekohl , @dLobatog 
Done, please review.
thanks!

app/views/users/_form.html.erb
```diff
@@ -20,8 +20,8 @@
 
   <div class='tab-content'>
     <div class='tab-pane active' id='primary'>
-
-      <%= text_f f, :login, :disabled => f.object.login == 'admin' %>
+      <% can_edit_login =   User.current.can?(:edit_users, @user) || ( f.object.id == User.current.id  && User.current.internal? ) %>
```
dLobatog: Please move this logic to the User model (as a validation) & test it, that'd help ensuring this doesn't break in the future (methods in views are prone to break without anyone noticing)

app/views/users/_form.html.erb
```diff
@@ -20,8 +20,8 @@
...
+      <%= text_f f, :login, :disabled => !can_edit_login, :label_help => !can_edit_login ? "The user can not edit the login":" " %>
```
dLobatog: This ought to be translatable: `_("The user can not edit the login")`.

How about "Only internal users with enough permissions can edit their login". I'm not too sure of the wording though.. (cc @ehelms @GregSutcliffe mind to double check?)

test/controllers/concerns/parameters/user_test.rb
```diff
@@ -3,7 +3,12 @@
 class UserParametersTest < ActiveSupport::TestCase
   include Foreman::Controller::Parameters::User
 
-  let(:context) { Foreman::Controller::Parameters::User::Context.new(:api, 'user', 'update', editing_self) }
+  def setup
+    @auth_source_ldap = FactoryBot.create(:auth_source_ldap)
```
dLobatog: This will create an `n` & user before every single test, which is slow (commits to the db). How about creating a `context` to test this specific feature or just creating the auth source where needed? 

Same for user & `let(:context)`.

test/controllers/concerns/parameters/user_test.rb
```diff
@@ -34,6 +39,27 @@ class UserParametersTest < ActiveSupport::TestCase
       params = ActionController::Parameters.new(:user => {:login => 'another'})
       assert_includes(as_admin { self.class.user_params_filter.filter_params(params, context) }, 'login')
     end
+
+    test "permits login attribute for external users with permissions" do
```
dLobatog: It shouldn't allow changing the `login` attribute for Users that are not Internal, see https://github.com/theforeman/foreman/pull/4966#issuecomment-343183640 https://github.com/theforeman/foreman/pull/4966#issuecomment-343427867 , it may have security implications and it's useless unless the User changing the `login` can also change their name in LDAP.

test/controllers/concerns/parameters/user_test.rb
```diff
@@ -34,6 +39,27 @@ class UserParametersTest < ActiveSupport::TestCase
...
+    end
+
+    test "block login attribute for external users" do
+      params = ActionController::Parameters.new(:user => {:login => 'another'})
+      as_user (FactoryBot.create(:user, :auth_source => @auth_source_ldap)) do
+        assert_not_includes (self.class.user_params_filter.filter_params(params, context)), 'login'
+      end
+    end
+
+    test "permits login attribute fo internal user with permissions" do
```
dLobatog: typo, fo -> for

test/controllers/concerns/parameters/user_test.rb
```diff
@@ -34,6 +39,27 @@ class UserParametersTest < ActiveSupport::TestCase
       params = ActionController::Parameters.new(:user => {:login => 'another'})
       assert_includes(as_admin { self.class.user_params_filter.filter_params(params, context) }, 'login')
     end
+
+    test "permits login attribute for external users with permissions" do
+      params = ActionController::Parameters.new(:user => {:login => 'another'})
+      as_user (FactoryBot.create(:user, :auth_source => @auth_source_ldap, :roles => [roles(:manager)])) do
```
dLobatog: `.build` or `.build_stubbed` is probably good enough in every new test here. That would only create the object in memory rather than making the roundtrip to the database & commit (so it'd be faster).

dLobatog (changes requested):
@shiramax Just left a few comments, thanks @shiramax

app/views/users/_form.html.erb
GregSutcliffe: Hmm, tricky one. Perhaps "Only permitted local users can edit their login"? Not sure about loca vs internal - local seems more "friendly" but possible to misinterpret?

GregSutcliffe: Another idea, two tests with two messages. If it's an external user, display "External users cannot edit their login", and if it's an internal user, check perms and display "You do not have permission to edit your login" (or whatever we use elsewhere for not having the right permissions".

dLobatog (approved):
Thanks @shiramax

app/views/users/_form.html.erb
```diff
@@ -20,8 +20,8 @@
 
   <div class='tab-content'>
     <div class='tab-pane active' id='primary'>
-
-      <%= text_f f, :login, :disabled => f.object.login == 'admin' %>
+      <%  msg =  @user.internal? ? "": "External user login can not be changed" %>
```
iNecas, line 23: Seems like this introduced a regression when creating a new user http://projects.theforeman.org/issues/21849

shiramax:
@iNecas, sorry for that, here is a fix : 
https://github.com/theforeman/foreman/pull/5052

PR #4967, opened 2017-10-31
Fixes #21519 - Prevent stored XSS on fact charts

tbrisker:


tbrisker:
Once accepted this should be cherry-picked into 1.16-stable and possibly also 1.15-stable if we do another release.

ohadlevy:
@tbrisker - thanks! does it make sense to add tests ?

tbrisker:
@ohadlevy not sure if there is a good way to do that, the new charts already sanitize the text correctly, the old js code should go all away once remaining charts are migrated to react. I added the escaping on the backend as well as a backup measure in case flot.js somewhere doesn't escape correctly or in case i missed something in the awful mess that is there right now.

dLobatog (approved):
Looks good to me - I'm not sure it's worth adding tests for checking each of the escapes here. Maybe checking that an HTML tag as part of a fact name is escaped in some page as an integration test would be good, but I'm not too adamant, your call @ohadlevy @tbrisker

ohadlevy:
thanks @tbrisker

PR #4968, opened 2017-10-31
Fixes #21522: Add prettier to foreman

danseethaler:
Adding prettier and formatting all js files under weback/ along
with js/json config files.

http://projects.theforeman.org/issues/21522


tbrisker (changes requested):
Looks like the prettier settings seem to mismatch the eslint settings - please make sure that they are consistent with one another, preferably using the current style unless there is a good reason to change it to avoid a huge chunk of changes making following history more difficult.
Also, looks like you mistakingly added a vscode config file.
TBH, I still don't understand the benefit of using both eslint and prettier - why have two tools that essentially have the same effect of enforcing code style?

danseethaler:
I realize this is a massive commit but wanted to get the discussion going. The only change that is not a formatting change was adding `prettier` as a dependency, adding a `format` script to `package.json` and changing the line length in `.eslintrc` from 100 to 80 to match prettier defaults.

danseethaler:
Hey @tbrisker - thanks for hopping in here. I'll work to get everything in alignment. The purpose of prettier is to format your code in a consistent fashion where eslint will callout errors in your code based on the rules you configure.

Prettier is designed to be opinionated and very low configuration. Consistent formatting will reduce git diffs and makes it easier to program without having the mental overhead of formatting as you go.

danseethaler:
@tbrisker and good catch on the `.vscode` file. I'll remove that.

ohadlevy:
@tbrisker having used prettier myself (on my editor) for a while now, I can see the benefit in using it.
besides making the code more consistent in terms of style, after this patch, it would also introduce less changes as the styling code is consistent.
additionally, i see it as an anti bike shedding tool.

overall, :+1: from my side (once all of the hound / eslint issues are resolved).

@danseethaler how do you plan to enforce the style? is there a github hook or a hound configuration that will complain if its not styled correctly?

danseethaler:
@ohadlevy I believe there is a hound configuration [see this pr](https://github.com/houndci/linters/pull/111). Who is currently configuring hound for foreman?

I would also like to figure out why the hound issues are not same as the issues listed when running `npm run lint` locally.

danseethaler:
@ohadlevy @tbrisker thoughts on removing the `max-len` rule in favor of enforcing where possible via prettier? Seems a lot cleaner than adding `eslint-disable` on lines that can't be broken up. We would set the prettier setting to something like 80 characters and let it work it's magic.

tbrisker:
@danseethaler hound is configured [here](https://github.com/theforeman/foreman/blob/develop/.hound.yml).
I'm guessing that any discrepancy between local eslint and hound is due to different versions locally and on hound - hound has `"eslint": "^4.9.0"` while we have `"eslint": "^3.3.1"`, so i'm guessing the defaults are a bit different and that's what it falls back to for rules not defined in our .eslintrc.
I have a feeling I'm still missing some advantage to prettier - how are we not doing the same checks twice now, once with prettier and once with eslint? I understand maybe for a new project prettier is easier since it has less config and is more opinionated, but right now our .eslintrc is opinionated and requires 0 config.
also, we don't have hound configured now to run prettier so we wouldn't know if someone breaks the prettier config, just if the eslint breaks. another problem is if some default changes in either of them, we would have to change the other to be consistent rather then just use one tool with one config.

ohadlevy:
@tbrisker actually, there is a eslint plugin for prettier, so that probably the right way.

from a usage preservative, prettier has more complete rules when it comes to how to write the code, not just white spaces etc, the fact its well integrated into editors etc makes me more productive, and i think that's the important point here, I suggest you try using it.

tbrisker:
@ohadlevy I would be happy to converge on one tool if possible then maintain two that do very similar things - if prettier is better then eslint, lets drop eslint and move to prettier completely. I'm not sure tbh that prettier has more rules then eslint [[1](https://eslint.org/docs/rules/)] or better integration in editors, but i'm willing to take your word on it.
Also, we should update hound to enforce prettier, not all developer would use it i suppose and we should have gating to make sure all new code follows the same style.

danseethaler:
@tbrisker to be clear prettier does not replace eslint. Prettier is designed to format your code regardless of the initial code format. There is very little configuration in prettier which should match your eslint settings. I'd recommend giving the [readme](https://github.com/prettier/prettier) a look.

@ohadlevy I'm working on getting that plugin going. Thanks.

tbrisker:
@danseethaler looks to me like the whole formatting rules part of eslint is currently duplicated by prettier, so other then automatically fixing the formatting (that can also be done with `eslint --fix` which i'm sure also has editor integration), how is it better? 
we already have 0 configuration, opinionated rules - they are defined in `.eslintrc`, and i'm hesitant to change 1000 lines of code just to conform with prettier's opinion on how code should look like - as long as it's consistent across the code base, i couldn't care less about whitespace around parentheses etc.

danseethaler:
@tbrisker I think you're right on about duplicating the formatting rules in `.eslintrc`. Getting down to zero/minimal config means removing most of those `.eslintrc` rules in favor of the prettier defaults. I'm working on that now.

The reason prettier is better is because of the way it works to reformat your code regardless of the input format. As I mentioned this is all explained in the README but the main idea is that prettier parses your input file into an AST and then rewrites the format based on opinionated standard rules. This will help us by minimizing the config in `.eslintrc` and by simply having more comprehensive formatting.

It's also well suited for JSX and even supports other languages like css and json.

If you haven't used it I could see why it would be hard to grasp what's better about it. But if you give it a try I think you'll be pleased.

tbrisker:
I did look at the README, and I also tried it on my side - when running it in my editor i get completely different output (i'm assuming my editor uses different default settings to yours). TBH it doesn't really matter to me how it does the parsing behind the scenes to fix it, eslint can also do the same with --fix. I still haven't gotten an answer on why their opinion of standard rules is better then the opinion we currently have in place.
Also, I did not get an answer on how to gate this? Not all developers use editors that will autoformat the code, and we should enforce consistent style.

danseethaler:
@tbrisker alright I've simplified the PR for now down to just the `package.json` and the `.eslintrc` to make it clear what I'm proposing. If you `npm run format` prettier will format all the files that match the blob `./!(app|coverage)/**/*.{js,json,css,scss}` which may need to be modified depending on what files we want formatted. With the changes I made to the `.eslintrc` file there are only a couple of camelcase errors that show up which need to be fixed by just adding the `estlint-disable camelcase` as necessary. I'll do this before I submit a final version of this PR.

The enforcement will come from the `eslint-plugin-prettier` settings so hound will bark if there's an error. I've removed some of the styling settings from the `.eslintrc` file and IMO we could reduce this file far more by just extending some recommended settings but what we have is fine for now.

The bottom line with prettier is that it gives you more consistent formatting than you can get with just eslint. `eslint --fix` can do some things but not as much as prettier. It's a widely accepted javascript tool and works great with existing eslint setups.

danseethaler:
Alright I think I've got everything to the right setup. I'm finding the hound errors are invalid and don't show up when running locally. I think this could have to do with changing the `.eslintrc` file and updating files at the same time. In each case the current code is correct. I also updated the `eslint` package like you recommended @tbrisker - thanks for catching that.

Let me know if anyone else has any feedback on this. Thanks.

test/integration/support/poltergeist_onload_extensions.js
```diff
@@ -1,8 +1,9 @@
-var disableAnimationStyles = '-webkit-transition: none !important;' +
```
ohadlevy: please dont change any of the files that are not under eslint control.

danseethaler: Ah yeah that must be a carry-over. I'll get that removed.

danseethaler:
It looks as though prettier is acting different when run by hound than when running locally. It looks like the line length is longer in hound for some reason. Still trying to figure that out.

danseethaler:
Well I'm feeling less confident that @houndci-bot is going to be able to correctly call out prettier errors.

There's an interesting thread in the [foreman-dev](https://groups.google.com/forum/#!searchin/foreman-dev/houndci%7Csort:date/foreman-dev/L-sQd5J__Ks/8bUYDZKTAgAJ) about possibly moving the lint task back into Jenkins. There is some discussion there about how @houndci-bot incorrectly tags PRs. Although the inline comments are nice for small things, if you have a lot of issues (like this PR) the inline comments make it cumbersome to read.

I've also played around with changing out our `.eslintrc` for a simple extension of the airbnb config. I think this would standardize and simplify everything but I don't think we should update all affected files in the PR. I think we should update the `.eslintrc` and let the files be updated as they come. In this case, if we did move back to Jenkins we would simply need a linting test that only ran on the files changed in the PR.

danseethaler:
I've added [this issue](https://github.com/houndci/hound/issues/1461) to the HoundCI repo about prettier inconsistencies.

danseethaler:
Closing this PR for now. I've added [another PR](https://github.com/theforeman/foreman/pull/4997#issue-272721534) to get the `eslint-config-airbnb` in place. Once that's finished and we can get a solution to the HoundCI/Prettier issue I will open another PR for prettier.

PR #4971, opened 2017-11-01
Fixes #21543 - Tooltip will appear only if the key is too long

shiramax:
If the key is too long, a tooltip will appear: 
image

If this key is short, then no tooltip will appear
image


dLobatog (approved):
Thanks @shiramax !

PR #4976, opened 2017-11-02
Fixes #21565 - Orchestration can now optionally be enabled during unattended operations

fabianvf:
http://projects.theforeman.org/issues/21565

I wasn't really sure what tests would look like for a change like this, would appreciate some guidance on that front.

### DIFF
--- a/app/models/host/managed.rb
+++ b/app/models/host/managed.rb
@@ -294,10 +294,10 @@ class Host::Managed < Host::Base
   def import_facts(facts, source_proxy = nil)
     # Facts come from 'existing' attributes/infrastructure. We skip triggering
     # the orchestration of this infrastructure when we create a host this way.
-    skip_orchestration! if SETTINGS[:unattended]
+    skip_orchestration! if SETTINGS[:unattended] && !SETTINGS[:enable_orchestration_on_fact_import]
     super
   ensure
-    enable_orchestration! if SETTINGS[:unattended]
+    enable_orchestration! if SETTINGS[:unattended] && !SETTINGS[:enable_orchestration_on_fact_import]
   end
 
   # Request a new OTP for a host
--- a/app/models/setting/provisioning.rb
+++ b/app/models/setting/provisioning.rb
@@ -46,6 +46,7 @@ class Setting::Provisioning < Setting
       self.set('access_unattended_without_build', N_("Allow access to unattended URLs without build mode being used"), false, N_('Access unattended without build')),
       self.set('manage_puppetca', N_("Foreman will automate certificate signing upon provision of new host"), true, N_('Manage PuppetCA')),
       self.set('ignore_puppet_facts_for_provisioning', N_("Stop updating IP address and MAC values from Puppet facts (affects all interfaces)"), false, N_('Ignore Puppet facts for provisioning')),
+      self.set('enable_orchestration_on_fact_import', N_("Enable host orchestration on puppet fact import. This could cause serious performance issues as the number of hosts increase"), false, N_('Enable orchestration on puppet fact import')),
       self.set('ignored_interface_identifiers', N_("Ignore interfaces that match these values during facts importing, you can use * wildcard to match names with indexes e.g. macvtap*"), IGNORED_INTERFACES, N_('Ignore interfaces with matching identifier')),
       self.set('ignore_facts_for_operatingsystem', N_("Stop updating Operating System from facts"), false, N_('Ignore facts for operating system')),
       self.set('ignore_facts_for_domain', N_("Stop updating domain values from facts"), false, N_('Ignore facts for domain')),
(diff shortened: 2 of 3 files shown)


iNecas:
I guess https://github.com/theforeman/foreman/blob/8c6bc83ee15cf137c1e43a6d7a7df8d66965b44c/test/models/host_test.rb#L161 can be used for inspiration

app/models/host/managed.rb
```diff
@@ -294,7 +294,7 @@ def handle_ca
   def import_facts(facts, source_proxy = nil)
     # Facts come from 'existing' attributes/infrastructure. We skip triggering
     # the orchestration of this infrastructure when we create a host this way.
-    skip_orchestration! if SETTINGS[:unattended]
+    skip_orchestration! if SETTINGS[:unattended] and not SETTINGS[:unattended_orchestration]
```
iNecas: Prefer `&&` over `and` (there is slight difference in operator preferrence). Also `!` is generally preferred in Foreman codebase.

iNecas:
ok to test

fabianvf:
@iNecas thanks for the testing tip, very helpful

test/models/host_test.rb
```diff
@@ -163,6 +163,41 @@ def teardown
     end
   end
 
+  test "should trigger orchestration when importing facts if unattended_orchestration is true" do
```
iNecas: Please use `context` block, and `setup` and `teardown` for setting up the settings: the reason is, that if something fails during the the test method, the setting to the original value would not happen

app/models/setting/provisioning.rb
```diff
@@ -46,6 +46,7 @@ def self.default_settings
       self.set('access_unattended_without_build', N_("Allow access to unattended URLs without build mode being used"), false, N_('Access unattended without build')),
       self.set('manage_puppetca', N_("Foreman will automate certificate signing upon provision of new host"), true, N_('Manage PuppetCA')),
       self.set('ignore_puppet_facts_for_provisioning', N_("Stop updating IP address and MAC values from Puppet facts (affects all interfaces)"), false, N_('Ignore Puppet facts for provisioning')),
+      self.set('unattended_orchestration', N_("Enable host orchestration during unattended operations"), false, N_('Enable unattended orchestration')),
```
lzap: This is not good description, this flag for example will break discovery plugin.

lzap (changes requested):
This is more about fact import than anything unattended. Please figure out a different name and description which will be more clear. Also performance impact of this setting are huge - fact importing is probably the most important API in terms of performance.

I think this flag will be dangerous and can make troubleshooting more difficult. This needs to have clear name, description and documentation in our guide.

app/models/setting/provisioning.rb
fabianvf: Interesting, I definitely don't want to break discovery... The motivation for this is that I just want my hosts (which are all provisioned from discovered hosts) to keep DNS in sync with the host, which is updated by the puppet fact uploads. I knew there were potential performance issues but I wasn't aware that it would break any existing behaviors. Is there a better way I can go about this so that it won't affect the discovery plugin?

lzap: Ok I take it back, it does not break discovery. Go head but I think the name and description is a little bit confusing, this has something to do with fact importing.

app/models/host/managed.rb
```diff
@@ -294,7 +294,7 @@ def handle_ca
   def import_facts(facts, source_proxy = nil)
     # Facts come from 'existing' attributes/infrastructure. We skip triggering
     # the orchestration of this infrastructure when we create a host this way.
-    skip_orchestration! if SETTINGS[:unattended]
+    skip_orchestration! if SETTINGS[:unattended] && !SETTINGS[:unattended_orchestration]
     super
   ensure
     enable_orchestration! if SETTINGS[:unattended]
```
lzap: The same change must be on this line as well.

lzap (approved):

lzap (changes requested):
* The same change must be on this line as well
* Figure out better name somehing like: Orchestrate during fact import

app/models/setting/provisioning.rb
fabianvf: do you think something like `enable_orchestration_on_fact_import` would be more fitting?

fabianvf: With a big old performance disclaimer in the description

lzap: That sounds reasonable to me, don't need to be big, just ordinary description about what is going on :-)

fabianvf:
@lzap pushed up changes

lzap (approved):

lzap:
Thanks.

ezr-ondrej:
@fabianvf in https://github.com/theforeman/foreman/pull/7749#discussion_r440806762 has been raised an concern this actually doesn't work anymore. And original issue seems to be resolved by other means. Could you please share if this is still an interesting feature for you, or we can drop the setting?

PR #4979, opened 2017-11-05
fixes #21579 - pagelets for host show

timogoebel:
This PR adds a pagelets extension point for the host show page (just two more lines). Also, fixes indentation.

Any chance this can still get into 1.16?

Tested with 

```ruby
       extend_page('hosts/show') do |context|
          context.add_pagelet :main_tabs,
            :name => N_('Omaha'),
            :partial => 'hosts/omaha_tab',
            :onlyif => proc { |host| host.omaha_facet }
        end
```

which leads to:

image

### DIFF
--- a/app/views/hosts/show.html.erb
+++ b/app/views/hosts/show.html.erb
@@ -20,6 +20,7 @@
         </tr>
       </tbody>
     </table>
+
     <ul id="myTab" class="nav nav-tabs">
       <li class="active"><a href="#properties" data-toggle="tab"><%= _('Properties') %></a></li>
       <li><a href="#metrics" data-toggle="tab"><%= _('Metrics') %></a></li>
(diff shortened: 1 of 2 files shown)


xprazak2:
👍 works as expected

iNecas (approved):
I think I could live with that :)

ohadlevy:
:+1: mind adding a test for the interface?

timogoebel:
@ohadlevy: Added a small test that checks that a pagelet is actually included in the show page. Pagelets itself are already covered pretty well.

ohadlevy:
thanks @timogoebel !

ares:
I set release to 1.17 in redmine, @dLobatog what do you think? @timogoebel asked for 1.16, it's not a regression and it's not even a bug fix. It seem pretty safe though.

PR #1234, opened 2014-02-12
fixes #4315 Return standard error responses for host power management vi...

mchesler:
...a API

PR #4980, opened 2017-11-05
Fixes #21580 - normalize compute attributes

tstrachota:
Normalizes compute attributes in API show actions. For backwards compatibility reasons this patch adds a new rabl node instead of replacing 'vm_attrs'.

### DIFF
--- a/app/models/compute_resource.rb
+++ b/app/models/compute_resource.rb
@@ -364,7 +364,25 @@ class ComputeResource < ApplicationRecord
     respond_to?(:associated_host)
   end
 
+  def normalize_vm_attrs(vm_attrs)
+    vm_attrs
+  end
+
   protected
+
+  def memory_gb_to_bytes(memory_size)
+    memory_size.to_s.gsub(/[^0-9]/, '').to_i * 1.gigabyte
+  end
+
+  def to_bool(value)
+    ['1', 'true'].include?(value.to_s.downcase) unless value.nil?
+  end
+
+  def slice_vm_attributes(vm_attrs, fields)
+    fields.inject({}) do |slice, f|
+      slice.merge({f => (vm_attrs[f].to_s.empty? ? nil : vm_attrs[f])})
+    end
+  end
 
   def client
     raise ::Foreman::Exception.new N_("Not implemented")
--- a/app/models/compute_resources/foreman/model/ec2.rb
+++ b/app/models/compute_resources/foreman/model/ec2.rb
@@ -106,6 +106,28 @@ module Foreman::Model
       client.images.get(image).present?
     end
 
+    def normalize_vm_attrs(vm_attrs)
+      normalized = slice_vm_attributes(vm_attrs, ['flavor_id', 'availability_zone', 'subnet_id', 'image_id', 'managed_ip'])
+
+      normalized['flavor_name'] =  self.flavors.detect { |f| f.id == normalized['flavor_id'] }.try(:name)
+      normalized['subnet_name'] =  self.subnets.detect { |f| f.subnet_id == normalized['subnet_id'] }.try(:cidr_block)
+      normalized['image_name'] = self.images.find_by(:uuid => vm_attrs['image_id']).try(:name)
+
+      group_ids = vm_attrs['security_group_ids'] || []
+      group_ids = group_ids.select { |gid| gid != '' }
+      normalized['security_groups'] = group_ids.map.with_index do |gid, idx|
+        [idx.to_s, {
+          'id' => gid,
+          'name' => self.security_groups.detect { |g| g.group_id == gid }.try(:name)
+        }]
+      end.to_h
+
+      normalized
+    rescue Fog::Compute::AWS::Error => e
+      Foreman::Logging.exception("Unhandled EC2 error", e)
+      {}
+    end
+
     private
 
     def subnet_implies_is_vpc?(args)
--- a/app/models/compute_resources/foreman/model/libvirt.rb
+++ b/app/models/compute_resources/foreman/model/libvirt.rb
@@ -197,6 +197,37 @@ module Foreman::Model
       vm_attrs
     end
 
+    def normalize_vm_attrs(vm_attrs)
+      normalized = slice_vm_attributes(vm_attrs, ['cpus', 'memory', 'image_id'])
+
+      normalized['image_name'] = self.images.find_by(:uuid => vm_attrs['image_id']).try(:name)
+
+      volume_attrs = vm_attrs['volumes_attributes'] || {}
+      normalized['volumes_attributes'] = volume_attrs.each_with_object({}) do |(key, vol), volumes|
+        volumes[key] = {
+          'capacity' => memory_gb_to_bytes(vol['capacity']).to_s,
+          'allocation' => memory_gb_to_bytes(vol['allocation']).to_s,
+          'format_type' => vol['format_type'],
+          'pool' => vol['pool_name']
+        }
+      end
+
+      interface_attrs = vm_attrs['nics_attributes'] || {}
+      normalized['interfaces_attributes'] = interface_attrs.each_with_object({}) do |(key, nic), interfaces|
+        interfaces[key] = {
+          'type' => nic['type'],
+          'model' => nic['model']
+        }
+        if nic['type'] == 'network'
+          interfaces[key]['network'] = nic['network']
+        else
+          interfaces[key]['bridge'] = nic['bridge']
+        end
+      end
+
+      normalized
+    end
+
     protected
 
     def client
--- a/app/models/compute_resources/foreman/model/ovirt.rb
+++ b/app/models/compute_resources/foreman/model/ovirt.rb
@@ -373,6 +373,37 @@ module Foreman::Model
       attrs[:public_key] = key
     end
 
+    def normalize_vm_attrs(vm_attrs)
+      normalized = slice_vm_attributes(vm_attrs, ['cores', 'interfaces_attributes', 'memory'])
+      normalized['cluster_id'] = vm_attrs['cluster']
+      normalized['cluster_name'] = self.clusters.detect { |c| c.id == normalized['cluster_id'] }.try(:name)
+
+      normalized['template_id'] = vm_attrs['template']
+      normalized['template_name'] = self.templates.detect { |t| t.id == normalized['template_id'] }.try(:name)
+
+      cluster_networks = self.networks(:cluster_id => normalized['cluster_id'])
+
+      interface_attrs = vm_attrs['interfaces_attributes'] || {}
+      normalized['interfaces_attributes'] = interface_attrs.inject({}) do |interfaces, (key, nic)|
+        interfaces.update(key => { 'name' => nic['name'],
+                                'network_id' => nic['network'],
+                                'network_name' => cluster_networks.detect { |n| n.id == nic['network'] }.try(:name)
+                              })
+      end
+
+      volume_attrs = vm_attrs['volumes_attributes'] || {}
+      normalized['volumes_attributes'] = volume_attrs.inject({}) do |volumes, (key, vol)|
+        volumes.update(key => { 'size' => memory_gb_to_bytes(vol['size_gb']).to_s,
+                                'storage_domain_id' => vol['storage_domain'],
+                                'storage_domain_name' => storage_domains.detect { |d| d.id == vol['storage_domain'] }.try(:name),
+                                'preallocate' => to_bool(vol['preallocate']),
+                                'bootable' => to_bool(vol['bootable'])
+                              })
+      end
+
+      normalized
+    end
+
     protected
 
     def bootstrap(args)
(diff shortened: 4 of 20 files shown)


tstrachota:
@ares may I ask you to review this PR?

timogoebel:
@tstrachota: I believe, the commit message is missing a ' - ' -> 'fixes #1234 - my message'.

tstrachota:
@timogoebel you're right. updated.

app/models/compute_resource.rb
```diff
@@ -360,8 +360,20 @@ def image_exists?(image)
...
   end
 
+  def normalize_vm_attrs(vm_attrs)
+    vm_attrs
+  end
+
   protected
 
+  def memory_gb_to_bytes(memory_size)
+    memory_size.to_s.gsub(/[^0-9]/, '').to_i * (1024**3)
```
ares: `(1024**3)` could be perhaps replaces with `1.gigabyte`, see https://github.com/theforeman/foreman/blob/develop/lib/core_extensions.rb#L137-L159

app/models/compute_resources/foreman/model/ec2.rb
```diff
@@ -106,6 +106,31 @@ def image_exists?(image)
       client.images.get(image).present?
     end
 
+    def normalize_vm_attrs(vm_attrs)
+      vm_attrs['flavor_id'] = nil if vm_attrs['flavor_id'] == ''
```
ares: `if vm_attrs['flavor_id'] == ''` would be slightly more readable as `if vm_attrs['flavor_id'].empty?`, same applies to the rest of the PR

app/models/compute_resources/foreman/model/ec2.rb
```diff
@@ -106,6 +106,31 @@ def image_exists?(image)
...
+      vm_attrs['image_name'] = self.images.where(:uuid => vm_attrs['image_id']).first.try(:name)
+
+      group_ids = vm_attrs.delete('security_group_ids') || []
+      group_ids = group_ids.select { |gid| gid != '' }
+      vm_attrs['security_groups'] = group_ids.map.with_index do |gid, idx|
+        [idx.to_s, {
+          'id' => gid,
+          'name' => self.security_groups.detect { |g| g.group_id == gid }.try(:name)
+        }]
+      end.to_h
```
ares, line 123: why `.to_h` when it later returns `vm_attrs`?

app/models/compute_resources/foreman/model/libvirt.rb
```diff
@@ -197,6 +197,28 @@ def vm_compute_attributes_for(uuid)
...
+      vm_attrs['volumes_attributes'].each do |key, vol|
+        vol['capacity'] = memory_gb_to_bytes(vol['capacity']).to_s
+        vol['allocation'] = memory_gb_to_bytes(vol['allocation']).to_s
+        vol['pool'] = vol.delete('pool_name')
+      end
+      vm_attrs['interfaces_attributes'] = vm_attrs.delete('nics_attributes') || {}
+      vm_attrs['interfaces_attributes'].each do |key, nic|
+        # Remove the value for 'bridge' when the type is 'network' and vice versa
+        # The value isn't used anyway. It gets stored only when user first selects the bridge
+        # and then switches the type to 'network'.
```
ares: would this be consistent when we use the same format for creation? how would we differentiate between networks and bridges?

app/models/compute_resources/foreman/model/ovirt.rb
```diff
@@ -319,6 +319,26 @@ def public_key=(key)
       attrs[:public_key] = key
     end
 
+    def normalize_vm_attrs(vm_attrs)
+      vm_attrs['cluster_id'] = vm_attrs.delete('cluster')
+      vm_attrs['cluster_name'] = self.clusters.detect { |c| c.id == vm_attrs['cluster_id'] }.try(:name)
+
+      vm_attrs['template_id'] = vm_attrs.delete('template')
```
ares: this would create template_name key with nil value if no template is found, in ec2 adapter we don't configure id attributes if strings are empty, should this be consistent everywhere?

app/models/compute_resources/foreman/model/vmware.rb
```diff
@@ -562,6 +563,63 @@ def vm_compute_attributes(vm)
       vm_attrs
     end
 
+    def normalize_vm_attrs(vm_attrs)
+      vm_attrs['cores_per_socket'] = vm_attrs.delete('corespersocket')
+      vm_attrs['memory'] = vm_attrs.delete('memory_mb').to_i * 1024
```
ares: ` vm_attrs.delete('memory_mb').megabytes` should work I think

ares (changes requested):
This looks like a great start! Do you also have some plan how to document how attributes are structured somewhere? What are guidelines to add new attributes? And how could we prevent of breaking this again. I see `normalize_vm_attrs` operates on incoming parameter so if CR introduces new parameter, it would simply appear in `normalized_vm_attrs` without any gating. That could cause new inconsistencies quickly. Should we rather build new hash from scratch?

tstrachota:
@ares changed the methods and they return a new hash now. I also added some tests that check for new attribute names and instruct the developer what to do.

There's still one todo: I need to get a real oVirt instance to test network names.

app/models/compute_resources/foreman/model/vmware.rb
```diff
@@ -1,5 +1,6 @@
 require 'fog_extensions/vsphere/mini_servers'
 require 'foreman/exception'
+require 'rbvmomi'
```
timogoebel: Is this always available?

app/models/compute_resources/foreman/model/ec2.rb
```diff
@@ -106,6 +106,27 @@ def image_exists?(image)
...
+      group_ids = group_ids.select { |gid| gid != '' }
+      normalized['security_groups'] = group_ids.map.with_index do |gid, idx|
+        [idx.to_s, {
+          'id' => gid,
+          'name' => self.security_groups.detect { |g| g.group_id == gid }.try(:name)
+        }]
+      end.to_h
+
+      normalized
+    rescue Fog::Compute::AWS::Error
```
timogoebel: This exception should at least be logged.

app/models/compute_resources/foreman/model/vmware.rb
```diff
@@ -562,6 +563,71 @@ def vm_compute_attributes(vm)
...
+      normalized['cluster_name'] = vm_attrs['cluster']
+      normalized['cluster_name'] = nil if normalized['cluster_name'].empty?
+
+      if normalized['cluster_name']
+        normalized['resource_pool_id'] = self.resource_pools(:cluster_id => normalized['cluster_name']).detect { |p| p.name == vm_attrs['resource_pool'] }.try(:id)
+      end
+      normalized['resource_pool_name'] = vm_attrs['resource_pool']
+      normalized['resource_pool_name'] = nil if normalized['resource_pool_name'].empty?
+
+      normalized['guest_name'] = self.guest_types[vm_attrs['guest_id']]
```
timogoebel, line 622: I think the guest_name parameter should be the real name (or label) of the guest_id. And this parameter should be named `guest_id`.

test/models/compute_resources/compute_resource_test_helpers.rb
```diff
@@ -19,9 +19,108 @@ def mock_cr_servers(cr, servers)
...
+      attr_before => ''
+    }
+    normalized = cr.normalize_vm_attrs(vm_attrs)
+
+    refute(normalized.has_key?(attr_before))
+    assert(normalized.has_key?(attr_after))
+    assert_nil(normalized[attr_after])
+  end
+
+  ALLOWED_VM_ATTR_NAMES = %W(
```
timogoebel: If plugins use this helper, they will might have a hard time extending this constant.

app/models/compute_resources/foreman/model/vmware.rb
```diff
@@ -562,6 +563,71 @@ def vm_compute_attributes(vm)
...
+
+      normalized['hardware_version_id'] = vm_attrs['hardware_version']
+      normalized['hardware_version_name'] = vm_hw_versions[vm_attrs['hardware_version']]
+
+      normalized['memory_hot_add_enabled'] = to_bool(vm_attrs['memoryHotAddEnabled'])
+      normalized['cpu_hot_add_enabled'] = to_bool(vm_attrs['cpuHotAddEnabled'])
+      normalized['add_cdrom'] = to_bool(vm_attrs['add_cdrom'])
```
timogoebel, line 629: This is an arbitrary parameter we introduced and should only be set during vm creation. I would omit it here.

timogoebel (commented):
@tstrachota: Thanks. Writing this patch must have been a trial of patience. I left some remarks that came to my mind while reading the code.

ares:
resetting to WoC, @tstrachota could you take a look?

app/models/compute_resources/foreman/model/vmware.rb
tstrachota: It's fog-vsphere's dependency

app/models/compute_resources/foreman/model/ec2.rb
tstrachota: Ack, I'll add logging

app/models/compute_resources/foreman/model/vmware.rb
tstrachota: Are you sure with `guest_id`? I'm not sure I understand your comment, but `guest_id` gets normally copied over and `guest_name` contains description from the `guest_types()` mapping.

app/models/compute_resources/foreman/model/vmware.rb
tstrachota: Does it mean it can't be saved into attributes of a compute profile? AFAIK it's available in the compute profile UI. If we're removing it from here, we should remove it from the profile UI as well.

test/models/compute_resources/compute_resource_test_helpers.rb
tstrachota: Makes sense, I'll update it.

tstrachota:
@ares @timogoebel updated

tstrachota:
@timogoebel do you have any other comments? May I ask you to check the PR again?

app/models/compute_resources/foreman/model/vmware.rb
timogoebel: It can be saved into a compute profile, no worries here. I just wanted to say, that it's not an attribute that can be read from VM data.

app/models/compute_resources/foreman/model/vmware.rb
timogoebel: yes, but do we always ship the gem in production? We have a [metapackage for vmware](https://github.com/theforeman/foreman-packaging/blob/3879d30037d099a48f8480e91cd0f029c883a71e/packages/foreman/foreman/foreman.spec#L554-L564), that might not be installed. It adds the [fog-vsphere dependency](https://github.com/theforeman/foreman/blob/develop/bundler.d/vmware.rb).

app/models/compute_resources/foreman/model/vmware.rb
timogoebel: I'm not sure, anymore. I guess I must have misread the code. Carry on! :-)

tstrachota:
[test katello]

mmoll:
@tstrachota please rebase :)
@timogoebel could you have another look here?

tstrachota:
Rebased

mmoll:
[test katello]

mmoll:
foreman test failues unrelated

timogoebel:
@mmoll, @tstrachota: Sorry guys, this is huge and I just don't have the time to properly review this right now. If somebody else could jump in, I'd really appreciate this.

xprazak2:
@tstrachota, could you take a look at the test failures pls?

tstrachota:
Updated, tests should be fixed now.

tstrachota:
@xprazak2 @ares the tests are green now.
The vmWare part was reviewed by @timogoebel, the other parts were checked at least code-wise. 

Since the change touches all supported compute resources it will be extremely difficult to test every corner of the PR. On the other hand it's not changing the existing behavior. It adds new attributes and therefore I think the risks of breakages are lower. I also believe the code coverage is good.

We've branched 1.18 so I think it's the best time to merge this even though it wasn't tested (from reviewers side) with all CRs. What do you think? Do you have any further comments?

mmoll (approved):
I concur with @tstrachota

timogoebel (approved):
Yep, let's get this in and improve it as we go.

ares (approved):
Given I'm the last one, I have the honor to merge it :-) Thanks @tstrachota @mmoll and @timogoebel

timogoebel:
This caused test failures in core (`NameError: uninitialized constant RbVmomi...`). Can you please take a look, @tstrachota? Thanks.

https://ci.theforeman.org/job/puppetdb_foreman-pull-request/11/database=mysql,label=fast,ruby=2.4/testReport/junit/(root)/Foreman__Model__VmwareTest___normalize_vm_attrs/test_0013_finds_hardware_version_name/

tstrachota:
@timogoebel this https://github.com/theforeman/foreman/pull/5659 should fix it.

PR #4982, opened 2017-11-06
Fixes #21584 - set taxonomies for chart requests

ares:


app/services/statistics/base.rb
```diff
@@ -29,5 +31,17 @@ def search
     def metadata
       {:id => id, :title => title, :url => url, :search => search}
     end
+
+    private
+
+    def build_url(url, org, loc)
+      return nil if url.nil?
+
```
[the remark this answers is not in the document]
ohadlevy: maybe use rails url helpers ?

ohadlevy:
tests please? :)

ares:
Sure, will add tests and fix broken ones :-)

app/services/statistics/base.rb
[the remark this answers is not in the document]
ares: The only usable helper I found is `to_query` on hash, so I changed it to use it.

ares:
@ohadlevy I fixed the code so it does not add unnecessary ? at the end or url which fixed broken tests, also added new tests for covering taxonomy parameters, would you mind taking another look? Thanks

tbrisker:
@ares can you rebase please? i think test failures are due to old tree

ares:
yeah, that's most likely the cause, rebased

ares:
all green now, @tbrisker or @ohadlevy would you mind resuming the review?

app/services/statistics/base.rb
```diff
@@ -3,11 +3,13 @@ class Base
...
 
     def initialize(options = {})
-      @id       = options[:id]
-      @title    = options[:title]
-      @search   = options[:search]
-      @url      = options[:url]
+      @id = options[:id]
+      @title = options[:title]
+      @search = options[:search]
+      @url = build_url(options[:url], options[:organization_id], options[:location_id])
```
tbrisker: I don't see anywhere we actually provide `:url` in the options, perhaps we should just pass the parameters in the url directly? at a higher point in the stack we can actually use `statistics_url(id: id, organization_id: Organization.current.try(:id), location_id: Location.current.try(:id)` to generate the url using rails helpers instead of building it ourselves.

app/services/statistics/base.rb
```diff
@@ -19,7 +21,7 @@ def id
     end
 
     def url
-      @url || "statistics/#{id}"
+      @url || build_url("statistics/#{id}", @organization_id, @location_id)
```
tbrisker: Why are we building the url again here? shouldn't it already be set by the initializer?

app/services/statistics.rb
```diff
@@ -1,16 +1,18 @@
...
+      CountHosts.new(:count_by => :compute_resource, :title => _("Compute Resource Distribution"), :search => "compute_resource=~VAL~", :organization_id => org_id, :location_id => loc_id),
+      CountFacts.new(:count_by => :processorcount, :unit => Nn_('%s core', '%s cores'), :title => _("Number of CPUs"), :search => "facts.processorcount=~VAL1~", :organization_id => org_id, :location_id => loc_id),
+      CountFacts.new(:count_by => :manufacturer, :title => _("Hardware"), :search => "facts.manufacturer~~VAL~", :organization_id => org_id, :location_id => loc_id),
+      CountNumericalFactPair.new(:count_by => :memory, :title => _("Average Memory Usage"), :organization_id => org_id, :location_id => loc_id),
+      CountNumericalFactPair.new(:count_by => :swap, :title => _("Average Swap Usage"), :organization_id => org_id, :location_id => loc_id),
+      CountPuppetClasses.new(:id => :puppetclass, :title => _("Class Distribution"), :search => "class=~VAL1~", :organization_id => org_id, :location_id => loc_id)
     ]
 
     if SETTINGS[:locations_enabled]
```
tbrisker, line 16: Shouldn't we also pass these parameters to these two charts? e.g. so i can see distribution of sub-taxonomies or see a distribution of current org in different locations.

tbrisker (changes requested):
See comments inline, I think we could simplify the code a bit.

app/services/statistics.rb
ares: yeah, makes sense, I've added also loc id for location distribution which will result in 100% chart but that's hopefully the more reasonable than keeping it in any context

app/services/statistics/base.rb
ares: The problem is when you pass blank url to initializer, in that case the build_url returns nil, I think it hit me in tests. I think we rarely set some explicit url and just build it based on id when `url` method is called but we should keep supporting custom URLs.

app/services/statistics/base.rb
ares: Higher in the stack would mean somewhere where we have view context, the list of charts is in Statistics module though. I can use url_helpers since we don't need to specify host, so we can use `statistics_path` instead of `statistics_url`. But `statistics_path` does not accept additional parameters so I needed to use `url_for`. That seems to work but it is quite a change, therefore I'm keeping it in separate in case we want to keep this patch small, e.g. for backporting to older version. Specifying url option is now required.

app/services/statistics.rb
ares: the fix is part of second commit

app/services/statistics/base.rb
ares: this would be resolved in the second commit since the url would become mandatory

ares:
Thanks @tbrisker, please take another look

app/services/statistics.rb
```diff
@@ -1,26 +1,30 @@
...
     if SETTINGS[:organizations_enabled]
-      charts << CountHosts.new(:count_by => :organization, :title => _("Organization Distribution"), :search => "organization=~VAL~")
+      charts << CountHosts.new(:count_by => :organization, :title => _("Organization Distribution"), :search => "organization=~VAL~", :url => build_url('organization'))
     end
 
     charts
   end
+
+  def self.build_url(id)
+    Rails.application.routes.url_helpers.url_for :controller => 'statistics', :action => 'show', :id => id, :only_path => true, :organization_id => Organization.current.try(:id), :location_id => Location.current.try(:id)
```
tbrisker: Nice find on the url_helpers that we can use here without the view, but if we aren't pushing it up the stack to the controller, we can leave it at the `Statistics::Base` level. My main concern was with building the url ourselves which is why I suggested pushing it up to where we have the url helpers. 
I suggest setting it in the `Statistics::Base` initializer unless a url is explicitly specified to reduce the extra parameter passing the same id we already have, so something like:
```ruby
class Base
...
  def initilize(options = {})
...
    @url = options[:url] || build_url
...
  end
...
  def build_url
    Rails.application.routes.url_helpers.url_for :controller => 'statistics', :action => 'show', :id => @id, :only_path => true, :organization_id => Organization.current.try(:id), :location_id => Location.current.try(:id)
  end
...
```

This way we preserve the ability to pass a custom url if needed (that we don't do now), but default to always including the taxonomies in the default url. (and also reduce the diff even more in case backporting is needed)

tbrisker (changes requested):

app/services/statistics.rb
ares: all right, I think it's nice to be able to render a chart for other than current org/loc so I don't load the current context inside the statistic but rather pass it to the initializer, but the rest of the logic is now back

ares:
Thanks @tbrisker, PR updated, could you please take another look?

app/services/statistics/base.rb
```diff
@@ -18,16 +20,22 @@ def id
...
     def metadata
       {:id => id, :title => title, :url => url, :search => search}
     end
+
+    private
+
+    def build_url
+      taxonomies = {}
+      taxonomies[:location_id] = @location_id if @location_id
+      taxonomies[:organization_id] = @organization_id if @organization_id
```
tbrisker: you don't actually need to do this, if you pass a `nil` value for a parameter to the `url_for` helper that parameter will be ignored.

app/services/statistics/base.rb
```diff
@@ -18,16 +20,22 @@ def id
...
+
+      Rails.application.routes.url_helpers.url_for({:controller => 'statistics', :action => 'show', :id => id, :only_path => true}.merge(taxonomies))
```
tbrisker: this should be `:id => @id` I think

app/services/statistics/base.rb
```diff
@@ -1,13 +1,15 @@
...
-      @id       = options[:id]
-      @title    = options[:title]
-      @search   = options[:search]
-      @url      = options[:url]
+      @id = options[:id]
+      @title = options[:title]
+      @search = options[:search]
       @count_by = options[:count_by]
+      @organization_id = options[:organization_id]
+      @location_id = options[:location_id]
```
tbrisker, line 11: Maybe we should also default here to current taxonomies just to simplify the call for the default case?

tbrisker (changes requested):

app/services/statistics/base.rb
ares: That does not make sense to me. This is a class that should not know anything about "what is current context". When you generate a statistic, you know for which context you want it.

Also this would effectively mean that if you explicitly set empty org/loc (any context), the class would fallback to current org/loc which seems inconsistent with the usage where in one org you render a chart for another org.

app/services/statistics/base.rb
ares: right, that's no longer required, fixed

app/services/statistics/base.rb
ares: Why? if there's no id set explicitly, it should fallback `count_by`, if you want I can refactor the method so we do `@id = options[:id] || count_by.to_s` to initializer so `id` would remain pure reader. But it seems unrelated and I'm trying to keep this small so it can be backported.

ares:
Thanks @tbrisker, another round?

ares:
oh wait, I dropped one fix from previous commit, I'll repush in a minute

ares:
new version pushed

app/services/statistics/base.rb
tbrisker: Ok, makes sense to still allow explicitly setting `nil`. 
Technically, the statistics service shouldn't be aware of the context either, maybe we should pass that from the controller to `Statistics.charts`?

app/services/statistics/base.rb
tbrisker: right, missed that. technically i would prefer that be done in the initiallizer but I won't hold this PR on that.

app/services/statistics/base.rb
ares: Yup, that sounds reasonable, even though the Statistics service itself as it stands today is not much reusable since it hardcodes the charts for specific page. Fixed.

ares:
@tbrisker time for another round :-)

tbrisker (approved):
Works well, looks good, thanks @ares !

PR #4987, opened 2017-11-07
fixes #21592 - uses headless chrome instead of phantomjs

ohadlevy:
it seems that we can also cut down the default waiting time
which hopefully will speed up our tests.

### DIFF
--- a/test/integration_test_helper.rb
+++ b/test/integration_test_helper.rb
@@ -4,10 +4,10 @@ require 'rails/test_help'
 require 'mocha/mini_test'
 require 'capybara/rails'
 require 'factory_bot_rails'
-require 'capybara/poltergeist'
 require 'show_me_the_cookies'
 require 'database_cleaner'
 require 'active_support_test_case_helper'
+require 'selenium/webdriver'
 require 'minitest/retry'
 Minitest::Retry.use!
 
@@ -15,21 +15,20 @@ Minitest::Retry.on_consistent_failure do |klass, test_name|
   Rails.logger.error("DO NOT IGNORE - Consistent failure - #{klass} #{test_name}")
 end
 
-Capybara.register_driver :poltergeist do |app|
-  opts = {
-    # To enable debugging uncomment `:inspector => true` and
-    # add `page.driver.debug` in code to open webkit inspector
-    # :inspector => true
-    :js_errors => true,
-    :timeout => 60,
-    :extensions => ["#{Rails.root}/test/integration/support/poltergeist_onload_extensions.js"],
-    :phantomjs => File.join(Rails.root, 'node_modules', '.bin', 'phantomjs')
-  }
-  Capybara::Poltergeist::Driver.new(app, opts)
+Selenium::WebDriver::Chrome.driver_path = File.join(Rails.root, 'node_modules', '.bin', 'chromedriver')
+Capybara.register_driver :selenium_chrome do |app|
+  capabilities = Selenium::WebDriver::Remote::Capabilities.chrome(
+    chromeOptions: { args: %w(headless disable-gpu) }
+  )
+
+  Capybara::Selenium::Driver.new app,
+    browser: :chrome,
+
+    desired_capabilities: capabilities
 end
 
-Capybara.default_max_wait_time = 30
-Capybara.javascript_driver = :poltergeist
+Capybara.default_max_wait_time = 5
+Capybara.javascript_driver = :selenium_chrome
 
 class ActionDispatch::IntegrationTest
   # Make the Capybara DSL available in all integration tests
@@ -211,6 +210,7 @@ class ActionDispatch::IntegrationTest
 
   def login_admin
     SSO.register_method(TestSSO)
+    visit '/'
     set_request_user(:admin)
   end
 
--- a/app/views/nic/_hidden_layout.html.erb
+++ b/app/views/nic/_hidden_layout.html.erb
@@ -1,6 +1,6 @@
 
 <div class="hidden" id="interfaceHidden<%= local_assigns[:id] %>" data-interface-id="<%= local_assigns[:id] %>">
-  <%= field_set_tag "", :id => "interface", :'data-url' => new_interface_path do %>
+  <%= field_set_tag "", :id => "interface + #{local_assigns[:id]}", :'data-url' => new_interface_path do %>
     <%= f.hidden_field(:_destroy, :class => :destroyFlag, :value => (local_assigns[:id].nil? ? 1 : 0)) %>
     <%= yield %>
   <% end %>
--- a/bundler.d/test.rb
+++ b/bundler.d/test.rb
@@ -7,12 +7,12 @@ group :test do
   gem 'ci_reporter_minitest', :require => false
   gem 'capybara', '~> 3.0', :require => false
   gem 'puma', :require => false
+  gem 'selenium-webdriver', :require => false
   gem 'show_me_the_cookies', '~> 4.0', :require => false
   gem 'database_cleaner', '~> 1.3', :require => false
   gem 'launchy', '~> 2.4'
   gem 'factory_bot_rails', '~> 4.5', :require => false
   gem 'rubocop-checkstyle_formatter', '~> 0.2'
-  gem 'poltergeist', '>= 1.18.0', :require => false
   gem 'shoulda-matchers', '~> 3.0'
   gem 'shoulda-context', '~> 1.2'
   gem 'as_deprecation_tracker', '~> 1.4'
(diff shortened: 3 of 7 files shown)


waldenraines (dismissed):
Nice change, approved pending green tests.

ohadlevy:
@mmoll @ekohl any idea why the test fail? I'm installing chromedriver via npm (search from chromedriver in  http://ci.theforeman.org/job/test_develop_pr_core/14979/database=postgresql,ruby=2.2,slave=fast/consoleFull) and I'm using the relative path at https://github.com/theforeman/foreman/pull/4987/files#diff-7a6aaeae51145fd37e71db58212fd18bR15) which of course works locally. how can I debug / troubleshoot this?

ekohl:
According to https://github.com/SeleniumHQ/selenium/wiki/ChromeDriver it expects chrome to be in ```/usr/bin/google-chrome```. You're just installing the driver and I think we need to look at a package installation of chrome. A quick search for RPMs doesn't show me any for centos7 so I'll need to dive into it. Since it's selenium it should also support firefox which may work better on centos7 slaves. Need to dive into it a bit to figure out our options.

ohadlevy:
I was under the assumption that I was override it - (its possible via
https://sites.google.com/a/chromium.org/chromedriver/capabilities#TOC-Using-a-Chrome-executable-in-a-non-standard-location
)

also, we could consider setting up a symlink?

ekohl:
But the entire chrome browser isn't on NPM, only the chrome driver for selenium. One reason I can think of why chrome isn't packaged for EL7 is that it uses some kernel features for sandboxing that the EL7 kernel doesn't support, but this is speculation from my side. It may be that the Fedora packages do work. One path we can choose is to get Fedora slaves for the UI tests but I'd prefer to try a more generic solution first with Firefox.

ohadlevy:
On Wed, Nov 8, 2017 at 11:08 AM, Ewoud Kohl van Wijngaarden <
notifications@github.com> wrote:

> But the entire chrome browser isn't on NPM, only the chrome driver for
> selenium. One reason I can think of why chrome isn't packaged for EL7 is
> that it uses some kernel features for sandboxing that the EL7 kernel
> doesn't support, but this is speculation from my side. It may be that the
> Fedora packages do work. One path we can choose is to get Fedora slaves for
> the UI tests but I'd prefer to try a more generic solution first with
> Firefox.
>
>
>
Are you suggesting we will drop this pr in favor of firefox based
approach?

ekohl:
That would be my suggestion to start with. I have experience with using Firefox + Selenium on CentOS7 and generally that works well. If the setup code is generic enough we can support chrome via an env var for users who prefer that on their desktops.

Note that Firefox and xvfb are already installed on our slaves but the job config may need to be modified. Looks like the 1.11-stable job uses it.

ohadlevy:
On Wed, Nov 8, 2017 at 11:14 AM, Ewoud Kohl van Wijngaarden <
notifications@github.com> wrote:

> That would be my suggestion to start with. I have experience with using
> Firefox + Selenium on CentOS7 and generally that works well. If the setup
> code is generic enough we can support chrome via an env var for users who
> prefer that on their desktops.
>
> Note that Firefox and xvfb are already installed on our slaves but the job
> config may need to be modified. Looks like the 1.11-stable job uses it.
>

the fact is that currently the tests are: a. slow b. inaccurate as they
represent an old browser env (which imho is no longer relevant to our users
- e.g. IE). I would like to unblock / improve our test infrastructure.
whats the timeline you had in mind to implement this? thanks.

ekohl:
That makes total sense, but I haven't thought about it. We have a few priorities to balance no. I think that Rails 5.1 SCL + Updated Ruby is also a high priority blocker for developers but the fixing the Dynflow service for the 1.16 release is my top priority. Perhaps this is small enough to do it before the Rails 5.1 effort but I'll get back to you on that.

ohadlevy:
another option is to run it on travis - as its already configured to support chrome.
the main problem is that currently travis works on one language at a time (e.g. ruby or node) and not both.

ekohl:
Travis has matrix builds. See https://github.com/theforeman/puppet-foreman/blob/master/.travis.yml for how we use separate RVMs. Pretty sure you can choose a different language in the ```include```.

ohadlevy:
I failed to make this work (tried at
https://github.com/ManageIQ/manageiq-ui-classic/pull/2540) without success.

ekohl:
It appears you're right: https://github.com/travis-ci/travis-ci/issues/4090

ehelms:
The Travis configuration is already running node based tests. Where does the Ruby part come into play?

ohadlevy:
@ehelms chrome integration tests run against the rails server.... you need to be able to execute ```rake jenkins:integration```

ehelms:
One broader issue I will point out is that our current standard devel environment via Forklift is based on using CentOS 7 given that is the production RPM distro we support. This change would imply that developers could no longer run these tests within the local environment.

Has anyone tried installing the 'Fedora' only Chrome on EL7? From what I read, when they dropped it they dropped it intending to drop for EL6 but dropped it for all. Should be a fairly easy test on a Forklift environment to configure:

```
[google-chrome]
name=google-chrome - \$basearch
baseurl=http://dl.google.com/linux/chrome/rpm/stable/\$basearch
enabled=1
gpgcheck=1
gpgkey=https://dl-ssl.google.com/linux/linux_signing_key.pub
```

Install chrome and try this PR out. If that works, then we can add this to our slaves no?

ehelms:
@ohadlevy ahh so these are for the older JS tests not the new ones coming with the React changes? Or even those requires a running Rails server?

ohadlevy:
> ahh so these are for the older JS tests not the new ones coming with the React changes? Or even those requires a running Rails server?

this PR aim is to replace phantomjs with headless chrome- the motivation is to:
1. reduce integeration test time
2. use a more modern infrastructure - phantomjs seems like a less of an ideal tool today, we need polyfills etc as its emulated fairly older browsers.
(I must admit I don't know if we have an official minimum browser defined anywhere).

regarding the tests itself:
1. phatomjs and chrome are both doing full end to end tests using an automated browser - you require a 
full rails server.
2. webpack based code tests (using jest) do not require a browser - they run purely in travis today using jsdom.

ehelms:
So testing this out, I was able to install Chrome on EL7 without issues (running the tests now) using that repository. From what I read, they "pulled" Chrome from EL6 due to incompatibility and in the process pulled it from all all EL despite it still working just fine.

Assuming tests pass, I'd opt for adding this repository to our slaves and allowing the use of Chrome here.

ekohl:
@ehelms I think that's fine.

ohadlevy:
@ehelms whats the next step ?

dLobatog (commented):
@ohadlevy @ehelms @ekohl I think a good strategy to move forward is:

* Installing chrome alongside firefox/phantomjs in our slaves
* Making sure tests pass on Jenkins in this PR 
* Remove phantomjs/firefox

I've submitted https://github.com/theforeman/foreman-infra/pull/381, as soon as that is merged please trigger the tests here again to see if they pass. I'd be thrilled to see a `default_max_wait_time=5` working 😄

dLobatog:
[test foreman]

dLobatog (commented):
PR from foreman-infra got merged, hopefully chrome is available on the slaves that will run the tests

ohadlevy:
[test foreman]

dLobatog (commented):
From the tests, `chromedriver` would also be needed on Jenkins.

```
Selenium::WebDriver::Error::WebDriverError:  Unable to find chromedriver. Please download the server from http://chromedriver.storage.googleapis.com/index.html and place it somewhere on your PATH. More info at https://github.com/SeleniumHQ/selenium/wiki/ChromeDriver....
```

dLobatog:
[test foreman]

dLobatog (changes requested):
@ohadlevy Looks like a few tests are failing

ohadlevy:
@dLobatog rebased.

mmoll:
[test foreman]

mmoll:
@ohadlevy please rebase

mmoll:
@ohadlevy a lot of tests fail and I have the feeling a lot of tests are run with this new configuration the first time after a long while... how can that be?

dLobatog:
[test foreman]

dLobatog (changes requested):
@ohadlevy I think this needs rebasing so tests pass, could you try that? Thanks!

mmoll:
rebased, no change.

ohadlevy:
@dLobatog @mmoll - can we upgrade the chrome version?

mmoll:
Currently not, as it was agreed to use the chromium version coming in EPEL.

test/integration_test_helper.rb
```diff
@@ -213,6 +212,7 @@ def database_cleaner_strategy
 
   def login_admin
     SSO.register_method(TestSSO)
+    visit '/'
```
mmoll, line 213: while this fixes some tests, some are also failing because of this (e.g. `TEST=test/integration/fact_value_test.rb`)

mmoll:
I think the changes itself are OK, somebody:tm: would need to have a look into the problem with the cookies vs. visting a page first and then fix a few remaining problems.

ohadlevy:
closing in favor of #5770

PR #4989, opened 2017-11-08
Fixes #21602 - Add scoped_search on host.id for reports

xprazak2:


app/models/report.rb
```diff
@@ -19,6 +19,7 @@ class Report < ApplicationRecord
   def self.inherited(child)
     child.instance_eval do
       scoped_search :relation => :host,        :on => :name,  :complete_value => true, :rename => :host
+      scoped_search :relation => :host,        :on => :id,    :complete_value => true, :rename => :host_id
```
tbrisker: Why not search on the `host_id` column? it would save a potentially painful join with the host table. Also, please make this only explicit.

tbrisker (changes requested):

xprazak2:
I changed to search on `reports.host_id` column.

app/models/report.rb
```diff
@@ -27,6 +27,7 @@ def self.inherited(child)
       scoped_search :relation => :hostgroup,   :on => :title, :complete_value => true, :rename => :hostgroup_title
 
       scoped_search :on => :reported_at, :complete_value => true, :default_order => :desc, :rename => :reported, :only_explicit => true
+      scoped_search :on => :host_id,     :complete_value => true, :only_explicit => true
```
tbrisker: @ares made a good comment on the similar GH-4993 - please make complete_value false so it doesn't auto-suggest all host ids.

tbrisker (changes requested):

xprazak2:
I changed it to `:complete_value => false`

tbrisker (approved):

tbrisker:
Thanks @xprazak2

ares:
Setting release to 1.17

PR #4990, opened 2017-11-08
Fixes #21600 - react-bootstrap Panel

sharvit:
Remove the Panel implementation and switch to use react-bootstrap
implementation of panels.

http://projects.theforeman.org/issues/21600


tbrisker (approved):
Tested, works well, thanks @sharvit !

PR #4994, opened 2017-11-09
Fixes #21611 - Improve *_name lookup speed

tbrisker:


app/controllers/api/v2/hosts_controller.rb
```diff
@@ -20,7 +20,12 @@ class HostsController < V2::BaseController
 
       add_smart_proxy_filters :facts, :features => Proc.new { FactImporter.fact_features }
 
-      add_scope_for(:index) { |base_scope| base_scope.preload([:host_statuses, :compute_resource, :hostgroup, :operatingsystem, :interfaces, :token, :owner]) }
+      add_scope_for(:index) do |base_scope|
+        base_scope.preload([:host_statuses, :compute_resource, :hostgroup, :operatingsystem,
+                            :interfaces, :token, :owner, :model, :environment, :location,
+                            :organization, :image, :compute_profile, :realm, :architecture,
+                            :ptable, :medium, :puppet_proxy, :puppet_ca_proxy])
```
tbrisker, line 27: There were a bunch of N+1s here that were hidden because of the way we loaded `*_name`

iNecas:
```
# without patch:
time bundle exec bin/hammer host list --page 1 --per-page 1000 
2.26s user 0.17s system 11% cpu 20.710 total

# with patch:
bundle exec bin/hammer host list --page 1 --per-page 1000
2.25s user 0.15s system 18% cpu 13.171 total
```
I've also looked at the memory footprint and haven't seen any negative (or positive) impact. Also tested searching by architecture, just ot make sure scoped search was not influenced in any way. Seems good to go to me.

iNecas (approved):

iNecas:
Giving chance to somebody else to merge, or speak up if they feel something should be updated.

dLobatog (commented):
Thanks @tbrisker - I'm OK with merging but could you explain the changes in `has_many_common`? Just want to understand why are they needed and what do they do (how they change the query).

tbrisker:
@dLobatog In the current implementation, every time a *_name(s) method is called, we figure out the class of that association, run a query on that class to load the associated object(s), and then get their names. 
However, ActiveRecord already gives us a direct reference to that object directly. If the object has already been loaded to memory (e.g. because of a `preload` in the query) this will save an extra query against the DB. If it hasn't, AR will automatically generate the query by itself so there is no difference.
Additionally, this saves a couple of small object allocations (a string and block iterator in the many version, an int and string in the single version) which might have a minor affect when loading 1000s of hosts, and most importantly, makes the code simpler to read.

dLobatog (approved):

dLobatog:
Thanks @tbrisker @iNecas

PR #3667, opened 2016-07-21
Fixes #9016 - improved fact loading performance

lzap:
This change improves fact loading performance by order of magnitude on
SQLite mostly thanks to explicit transaction. An extra optimization was
done for the most important "update" method call which is mostly used
during fact uploads.

Updating of 500 existing facts is faster from 14.8 seconds to 2.7
seconds with this patch.

The patch utilizes find_each method with works in batches rather than creating
instance for each individual model object. More reading at:

http://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_each

PR #4996, opened 2017-11-09
Fixes #21599 - Improved fact importing mechanism to deal with names

ShimShtein:
Now the list of fact names will be calculated beforehand and saved to the database. It will also consider that the name could be added from other thread.

### DIFF
--- a/app/services/fact_importer.rb
+++ b/app/services/fact_importer.rb
@@ -39,17 +39,23 @@ class FactImporter
 
   # expect a facts hash
   def import!
+    # This function uses its own transactions that should not be included
+    # in the transaction that handles fact values
+    ensure_fact_names
+
     ActiveRecord::Base.transaction do
       delete_removed_facts
       update_facts
       add_new_facts
     end
 
-    if @error
-      Foreman::Logging.exception("Error during fact import for #{@host.name}", @error)
-      raise ::Foreman::WrappedException.new(@error, N_("Import of facts failed for host %s"), @host.name)
-    end
+    report_error(@error) if @error
     logger.info("Import facts for '#{host}' completed. Added: #{counters[:added]}, Updated: #{counters[:updated]}, Deleted #{counters[:deleted]} facts")
+  end
+
+  def report_error(error)
+    Foreman::Logging.exception("Error during fact import for #{@host.name}", error)
+    raise ::Foreman::WrappedException.new(error, N_("Import of facts failed for host %s"), @host.name)
   end
 
   # to be defined in children
@@ -59,11 +65,56 @@ class FactImporter
 
   private
 
-  attr_reader :host, :facts
+  attr_reader :host, :facts, :fact_names, :fact_names_by_id
+
+  def fact_name_class_name
+    @fact_name_class_name ||= fact_name_class.name
+  end
+
+  def ensure_fact_names
+    initialize_fact_names
+
+    missing_keys = facts.keys - fact_names.keys
+    add_missing_fact_names(missing_keys)
+  end
+
+  def add_missing_fact_names(missing_keys)
+    missing_keys.sort.each do |fact_name|
+      # create a new record and make sure it could be saved.
+      name_record = fact_name_class.new(fact_name_attributes(fact_name))
+
+      ActiveRecord::Base.transaction(:requires_new => true) do
+        begin
+          name_record.save!(:validate => false)
+        rescue ActiveRecord::RecordNotUnique
+          name_record = fact_name_class.find_by!(name: fact_name)
+        end
+      end
+
+      # make the new record available immediately for other fact_name_attributes calls
+      @fact_names[fact_name] = name_record
+    end
+  end
+
+  def initialize_fact_names
+    name_records = fact_name_class.unscoped.where(:name => facts.keys, :type => fact_name_class_name)
+    @fact_names = {}
+    @fact_names_by_id = {}
+    name_records.find_each do |record|
+      @fact_names[record.name] = record
+      @fact_names_by_id[record.id] = record
+    end
+  end
+
+  def fact_name_attributes(fact_name)
+    {
+      name: fact_name
+    }
+  end
 
   def delete_removed_facts
     ActiveSupport::Notifications.instrument "fact_importer_deleted.foreman", :host_id => host.id, :host_name => host.name, :facts => facts, :deleted => [] do |payload|
-      delete_query = FactValue.joins(:fact_name).where(:host => host, 'fact_names.type' => fact_name_class.name).where.not('fact_names.name' => facts.keys)
+      delete_query = FactValue.joins(:fact_name).where(:host => host, 'fact_names.type' => fact_name_class_name).where.not(:fact_name => fact_names.values)
       if ActiveRecord::Base.connection.adapter_name.downcase.starts_with? 'mysql'
         # MySQL does not handle delete with inner query correctly (slow) so we will do two queries on purpose
         payload[:count] = @counters[:deleted] = FactValue.where(:id => delete_query.pluck(:id)).delete_all
@@ -77,21 +128,16 @@ class FactImporter
   def add_new_fact(name)
     # if the host does not exist yet, we don't have an host_id to use the fact_values table.
     method = host.new_record? ? :build : :create!
-    fact_name = find_or_create_fact_name(name, facts[name])
+    fact_name = fact_names[name]
     host.fact_values.send(method, :value => facts[name], :fact_name => fact_name)
   rescue => e
     logger.error("Fact #{name} could not be imported because of #{e.message}")
     @error = e
   end
 
-  # value is used in structured importer to identify leaf nodes
-  def find_or_create_fact_name(name, value = nil)
-    fact_name_class.where(:name => name, :type => fact_name_class.name).first_or_create!
-  end
-
   def add_new_facts
     ActiveSupport::Notifications.instrument "fact_importer_added.foreman", :host_id => host.id, :host_name => host.name, :facts => facts do |payload|
-      facts_to_create = facts.keys - db_facts.pluck('fact_names.name')
+      facts_to_create = facts.keys - db_facts.pluck(:fact_name_id).map { |name_id| fact_names_by_id[name_id].name }
       facts_to_create.each { |f| add_new_fact(f) }
       payload[:added] = facts_to_create
       payload[:count] = @counters[:added] = facts_to_create.size
@@ -123,6 +169,6 @@ class FactImporter
   end
 
   def db_facts
-    host.fact_values.joins(:fact_name).where("fact_names.type = '#{fact_name_class}'")
+    host.fact_values.where(:fact_name => fact_names.values)
   end
 end
--- a/app/services/structured_fact_importer.rb
+++ b/app/services/structured_fact_importer.rb
@@ -1,17 +1,24 @@
 class StructuredFactImporter < FactImporter
   def normalize(facts)
-    # Remove empty values first, so nil facts added by normalize_recurse imply compose
+    # Remove empty values first, so nil facts added by flatten_composite imply compose
     facts = facts.select { |k, v| v.present? }
-    normalize_recurse({}, facts)
+    facts = flatten_composite({}, facts)
+
+    original_keys = facts.keys.to_a
+    original_keys.each do |key|
+      fill_hierarchy(facts, key)
+    end
+
+    facts
   end
 
   # expand {'a' => {'b' => 'c'}} to {'a' => nil, 'a::b' => 'c'}
-  def normalize_recurse(memo, facts, prefix = '')
+  def flatten_composite(memo, facts, prefix = '')
     facts.each do |k, v|
       k = prefix.empty? ? k.to_s : prefix + FactName::SEPARATOR + k.to_s
       if v.is_a?(Hash)
         memo[k] = nil
-        normalize_recurse(memo, v, k)
+        flatten_composite(memo, v, k)
       else
         memo[k] = v.to_s
       end
@@ -19,26 +26,46 @@ class StructuredFactImporter < FactImporter
     memo
   end
 
+  # ensures that parent facts already exist in the hash.
+  # Example: for fact: "a::b::c", it will make sure that "a" and "a::b" exist in
+  # the hash.
+  def fill_hierarchy(facts, child_fact_name)
+    facts[child_fact_name] = nil unless facts.key?(child_fact_name)
+
+    parent_name = parent_fact_name(child_fact_name)
+    fill_hierarchy(facts, parent_name) if parent_name
+  end
+
   def preload_fact_names
     # Also fetch compose values, generating {NAME => [ID, COMPOSE]}, avoiding loading the entire model
     Hash[fact_name_class.where(:type => fact_name_class).reorder('').pluck(:name, :id, :compose).map { |fact| [fact.shift, fact] }]
   end
 
-  def find_or_create_fact_name(name, value = nil)
-    if name.include?(FactName::SEPARATOR)
-      parent_name = /(.*)#{FactName::SEPARATOR}/.match(name)[1]
-      parent_fact = find_or_create_fact_name(parent_name, nil)
-      fact_name = parent_fact.children.where(:name => name, :type => fact_name_class.to_s).first
-    else
-      parent_fact = nil
-      fact_name = fact_name_class.where(:name => name, :type => fact_name_class.to_s).first
-    end
+  def ensure_fact_names
+    super
 
-    if fact_name
-      fact_name.update_attribute(:compose, value.nil?) if value.nil? && !fact_name.compose?
-    else
-      fact_name = fact_name_class.create!(:name => name, :type => fact_name_class.to_s, :compose => value.nil?, :parent => parent_fact)
-    end
-    fact_name
+    composite_fact_names = facts.map do |key, value|
+      key if value.nil?
+    end.compact
+
+    affected_records = fact_name_class.where(:name => composite_fact_names, :compose => false).update_all(:compose => true)
+
+    # reload name records if compose flag was reset.
+    initialize_fact_names if affected_records > 0
+  end
+
+  def fact_name_attributes(name)
+    attributes = super
+    fact_value = facts[name]
+    parent_fact_record = fact_names[parent_fact_name(name)]
+
+    attributes[:parent] = parent_fact_record
+    attributes[:compose] = fact_value.nil?
+    attributes
+  end
+
+  def parent_fact_name(child_fact_name)
+    split = child_fact_name.rindex(FactName::SEPARATOR)
+    child_fact_name[0, split]
   end
 end
--- a/test/benchmark/memory/puppet_fact_importer_benchmark.rb
+++ b/test/benchmark/memory/puppet_fact_importer_benchmark.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+require "benchmark/memory/memory_benchmark_helper"
+
+FactName.transaction do
+  (0..100000).each do |x|
+    FactName.connection.execute "INSERT INTO fact_names (name) values ('rand_fact_name_#{x}')"
+  end
+end
+
+class StructuredFactImporter
+  def fact_name_class
+    FactName
+  end
+end
+
+def generate_facts(total, unique_names = 0, structured_names = 0)
+  facts = Hash[(1..total).map{|i| ["fact_#{i}", "value_#{i}"]}]
+  (total..total+unique_names).map{|i| facts["fact_#{i}_#{Foreman.uuid}"] = "value_#{i}"}
+  (total..total+structured_names).map{|i| facts[(["f#{i}"] * (i % 10)).join('::') + i.to_s] = "value_#{i}"}
+  facts
+end
+
+Rails.logger.level = Logger::ERROR
+
+facts = generate_facts(200, 50, 25)
+User.current = User.unscoped.find(1)
+
+host = FactoryBot.create(:host, :name => "benchmark-#{Foreman.uuid}")
+
+with_chosen_profiler do
+  StructuredFactImporter.new(host, facts).import!
+end
--- a/test/unit/fact_importer_test.rb
+++ b/test/unit/fact_importer_test.rb
@@ -137,6 +137,31 @@ class FactImporterTest < ActiveSupport::TestCase
       assert_equal 1, importer.counters[:updated]
       assert_equal 1, importer.counters[:added]
     end
+
+    test "importer handles fact name additions" do
+      facts = { 'duplicate_name' => 'some_value' }
+      importer = FactImporter.new(host, facts)
+      importer.stubs(:fact_name_class).returns(FactName)
+
+      # redefine fact_name_attributes for this instance, so it will create
+      # the fact name before it's actually created by the importer
+      def importer.fact_name_attributes(fact_name)
+        ActiveRecord::Base.transaction(:requires_new => true) do
+          dup_record = FactoryBot.build(:fact_name, :name => fact_name)
+          # save duplicate record in isolated transaction
+          dup_record.save!
+        end
+
+        {
+          name: fact_name
+        }
+      end
+
+      importer.import!
+
+      name_record = FactName.find_by(name: 'duplicate_name')
+      assert_not_nil name_record
+    end
   end
 
   def default_import(facts)
--- a/test/benchmark/benchmark_helper.rb
+++ b/test/benchmark/benchmark_helper.rb
@@ -1,7 +1,6 @@
 require 'benchmark/ips'
 
 require File.expand_path('../../../config/environment', __FILE__)
-require 'factory_bot_rails'
 
 unless Rails.env.production? && !Rails.configuration.database_configuration["production"]["migrate"]
   puts "Rais must be in production and database must have migrations turned off!"
@@ -18,6 +17,7 @@ EOS
 end
 
 load "#{Rails.root}/db/schema.rb"
+require 'factory_bot_rails'
 
 def foreman_benchmark
   GC.start
--- a/test/benchmark/memory/memory_benchmark_helper.rb
+++ b/test/benchmark/memory/memory_benchmark_helper.rb
@@ -1,0 +1,31 @@
+require "benchmark/benchmark_helper"
+require 'memory_profiler'
+require "stackprof"
+
+def with_memory_profiler
+  report = MemoryProfiler.report do
+    yield
+  end
+  report.pretty_print
+end
+
+def with_stackprof
+  StackProf.run(mode: :object, raw: true, out: '/tmp/stackprof_objects.dump', interval: 1) do
+    yield
+  end
+  puts '/tmp/stackprof_objects.dump dump created, please use "stackprof --text /tmp/stackprof_objects.dump" to investigate'
+end
+
+def with_chosen_profiler(&block)
+  case (ENV['PROFILER'] || '').downcase
+  when 'memory_profiler'
+    profiler_method = :with_memory_profiler
+  when 'stackprof'
+    profiler_method = :with_stackprof
+  else
+    puts 'Set "PROFILER" to either "memory_profiler" or "stackprof"'
+    exit 1
+  end
+
+  send(profiler_method, &block)
+end


ShimShtein:
Test failures seem unrelated (webpack)

app/services/fact_importer.rb
```diff
@@ -40,26 +40,58 @@ def initialize(host, facts = {})
...
+
+    missing_keys = facts.keys - fact_names.keys
+    add_missing_fact_names(missing_keys)
+  end
+
+  def add_missing_fact_names(missing_keys)
+    missing_keys.sort.each do |fact_name|
+      name_record = fact_name_class.new(fact_name_attributes(fact_name))
+
+      name_record = fact_name_class.where(name: fact_name).first unless name_record.save
```
timogoebel: 1. consider `find_by(name: fact_name)`
2. why not do it the other way around? e.g. check if the FactName exists. If not, create it. Is this the attempt to prevent the race condition? I don't get, why you create the record first in memory, save it and if that fails you try to load it from the database. If this is a check, if the record exists, I think it's not ideal. What happens if saving fails for another reason then that the record already exists?

test/unit/fact_importer_test.rb
```diff
@@ -137,6 +137,27 @@ def fact_name_class
...
+        FactoryBot.create(:fact_name, :name => fact_name)
+
+        {
+          name: fact_name
+        }
+      end
+
+      importer.import!
+
+      name_record = FactName.where(name: 'duplicate_name').first
```
timogoebel: `.find_by()`?

app/services/structured_fact_importer.rb
```diff
@@ -1,44 +1,73 @@
...
-      fact_name.update_attribute(:compose, value.nil?) if value.nil? && !fact_name.compose?
-    else
-      fact_name = fact_name_class.create!(:name => name, :type => fact_name_class.to_s, :compose => value.nil?, :parent => parent_fact)
-    end
-    fact_name
+    composite_fact_names = facts.map do |key, value|
+      key if value.nil?
+    end.compact
+
+    affected_records = FactName.where(:name => composite_fact_names, :compose => false).update_all(:compose => true)
```
timogoebel: `fact_name_class.where`?

timogoebel (changes requested):
@ShimShtein: Thanks. Not a full review (yet), but my first thoughts while scanning the code.

app/services/fact_importer.rb
ShimShtein: > Is this the attempt to prevent the race condition?

Yes, it is. I have already loaded everything that was in the DB on the first attempt (in the `initialize_fact_names` function). Here I am dealing with new fact names, so optimistically they all should be saved successfully. In the (rare?) case of race condition, I want to reuse the record from the DB.

> What happens if saving fails for another reason then that the record already exists?

You are right, I'll add `validate!` to ensure the record could be saved before the actual save attempt. Then after reloading, I could check that the record exists and throw if it doesn't - it will eliminate other failures.

ShimShtein:
@timogoebel Addressed all issues that you have noticed.

lzap:
I am going to perform benchmark with branch. Can you please edit `test/benchmark/benchmark_helper.rb` file and move line `require 'factory_bot_rails'` to line 20 after `load` statement? This is likely a regression after ForemanBot version bump. Otherwise there is a missing "user" table error, factories must initialize after DB is loaded up apparently now.

Anyway, I see huge boost in all our micro benchmarks. This is benchmark run with develop:

```
Calculating -------------------------------------
PuppetFactImporter (200) - 0 UN 0 SN
                          4.246  (±23.6%) i/s -     40.000  in  10.028825s
PuppetFactImporter (200) - 0 UN 25 SN
                          3.259  (± 0.0%) i/s -     33.000  in  10.200658s
PuppetFactImporter (200) - 50 UN 0 SN
                          3.610  (± 0.0%) i/s -     36.000  in  10.007642s
PuppetFactImporter (200) - 50 UN 25 SN
                          2.828  (± 0.0%) i/s -     29.000  in  10.282379s
PuppetFactImporter (500) - 0 UN 0 SN
                          1.844  (± 0.0%) i/s -     18.000  in  10.019928s
PuppetFactImporter (500) - 0 UN 25 SN
                          1.630  (± 0.0%) i/s -     17.000  in  10.480939s
PuppetFactImporter (500) - 50 UN 0 SN
                          1.731  (± 0.0%) i/s -     18.000  in  10.432680s
PuppetFactImporter (500) - 50 UN 25 SN
                          1.522  (± 0.0%) i/s -     16.000  in  10.539485s
StructuredFactImporter (200) - 0 UN 0 SN
                          4.165  (± 0.0%) i/s -     41.000  in  10.009779s
StructuredFactImporter (200) - 0 UN 25 SN
                          3.206  (± 0.0%) i/s -     32.000  in  10.046382s
StructuredFactImporter (200) - 50 UN 0 SN
                          3.514  (± 0.0%) i/s -     36.000  in  10.266660s
StructuredFactImporter (200) - 50 UN 25 SN
                          2.786  (± 0.0%) i/s -     28.000  in  10.071512s
StructuredFactImporter (500) - 0 UN 0 SN
                          1.903  (± 0.0%) i/s -     19.000  in  10.188262s
StructuredFactImporter (500) - 0 UN 25 SN
                          1.684  (± 0.0%) i/s -     17.000  in  10.145329s
StructuredFactImporter (500) - 50 UN 0 SN
                          1.794  (± 0.0%) i/s -     18.000  in  10.070417s
StructuredFactImporter (500) - 50 UN 25 SN
                          1.563  (± 0.0%) i/s -     16.000  in  10.270270s
Memory stats
Total objects allocated: 136_477_034
Total heap pages allocated: 3_679
```

What I see is performance increase across all tests consistently by around 50% that is a huge boost.

```
Calculating -------------------------------------
PuppetFactImporter (200) - 0 UN 0 SN
                          9.244  (±10.8%) i/s -     85.000  in  10.015161s
PuppetFactImporter (200) - 0 UN 25 SN
                          6.944  (±14.4%) i/s -     68.000  in  10.001575s
PuppetFactImporter (200) - 50 UN 0 SN
                          8.172  (±12.2%) i/s -     82.000  in  10.117128s
PuppetFactImporter (200) - 50 UN 25 SN
                          6.279  (± 0.0%) i/s -     63.000  in  10.095479s
PuppetFactImporter (500) - 0 UN 0 SN
                          4.701  (±21.3%) i/s -     45.000  in  10.100685s
PuppetFactImporter (500) - 0 UN 25 SN
                          3.972  (± 0.0%) i/s -     39.000  in  10.002376s
PuppetFactImporter (500) - 50 UN 0 SN
                          4.376  (± 0.0%) i/s -     44.000  in  10.147283s
PuppetFactImporter (500) - 50 UN 25 SN
                          3.739  (± 0.0%) i/s -     38.000  in  10.232179s
StructuredFactImporter (200) - 0 UN 0 SN
                          7.407  (±13.5%) i/s -     72.000  in  10.021418s
StructuredFactImporter (200) - 0 UN 25 SN
                          5.595  (± 0.0%) i/s -     56.000  in  10.146865s
StructuredFactImporter (200) - 50 UN 0 SN
                          6.471  (± 0.0%) i/s -     65.000  in  10.097904s
StructuredFactImporter (200) - 50 UN 25 SN
                          5.181  (± 0.0%) i/s -     52.000  in  10.083821s
StructuredFactImporter (500) - 0 UN 0 SN
                          4.019  (±24.9%) i/s -     39.000  in  10.138915s
StructuredFactImporter (500) - 0 UN 25 SN
                          3.502  (± 0.0%) i/s -     35.000  in  10.131383s
StructuredFactImporter (500) - 50 UN 0 SN
                          3.800  (± 0.0%) i/s -     38.000  in  10.078096s
StructuredFactImporter (500) - 50 UN 25 SN
                          3.308  (± 0.0%) i/s -     33.000  in  10.034249s
Memory stats
Total objects allocated: 145_434_839
Total heap pages allocated: 3_682
```

We see slightly more (< 10%) object allocations but the speedup is reasonable. Note these are microbenchmarks running on sqlite3 (inmemory), in practice the benefit can be different. It should be possible to run the benchmark against postgresql database as well, I haven't tried that. If anyone want to give it a try, share numbers.

FYI to start benchmark do:

    FOREMAN_NO_DEPRECATIONS=1 DISABLE_SPRING=true RAILS_ENV=production be ruby -Itest test/benchmark/puppet_fact_importer_benchmark.rb

This looks very promising!

app/services/fact_importer.rb
```diff
@@ -40,26 +40,62 @@ def initialize(host, facts = {})
...
+
+  def add_missing_fact_names(missing_keys)
+    missing_keys.sort.each do |fact_name|
+      # create a new record and make sure it could be saved.
+      name_record = fact_name_class.new(fact_name_attributes(fact_name))
+      name_record.validate!
+
+      # if the record could not be saved, then it should already be in the database
+      # throw an exception if it isn't
+      name_record = fact_name_class.find_by!(name: fact_name) unless name_record.save
```
lzap: But we have an explicit database transaction started, SQL server will make sure we never see the new fact_name until end of the transaction. If this is here to solve `duplicate key value violates unique constraint` that will not work, at least on ACID compatible transaction engine (PostgreSQL). The other two databases we support that's a different story, but overall I am not big fan of this approach.

app/services/fact_importer.rb
```diff
@@ -40,26 +40,62 @@ def initialize(host, facts = {})
...
+
+  def ensure_fact_names
+    initialize_fact_names
+
+    missing_keys = facts.keys - fact_names.keys
+    add_missing_fact_names(missing_keys)
+  end
+
+  def add_missing_fact_names(missing_keys)
+    missing_keys.sort.each do |fact_name|
```
lzap, line 82: Is the `sort` really needed for anything? It's slow operation.

app/services/fact_importer.rb
```diff
@@ -40,26 +40,62 @@ def initialize(host, facts = {})
...
+
+      # make the new record available immediately for other fact_name_attributes calls
+      @fact_names[fact_name] = name_record
+    end
+  end
+
+  def initialize_fact_names
+    name_records = fact_name_class.where(:name => facts.keys, :type => fact_name_class.to_s).to_a
+    @fact_names = Hash[name_records.map{ |record| [record.name, record]}]
+  end
```
lzap: Hmmm but this is eager loading of quite a big table. On some instances, we see millions of records in this table (usually when there are dynamic NIC interfaces reported by facter). We removed this from our codebase to prevent eager loading years ago, this essentially returns it.

app/services/fact_importer.rb
```diff
@@ -86,7 +122,7 @@ def add_new_fact(name)
 
   # value is used in structured importer to identify leaf nodes
   def find_or_create_fact_name(name, value = nil)
-    fact_name_class.where(:name => name, :type => fact_name_class.name).first_or_create!
+    fact_names[name]
   end
```
lzap: The method should have been renamed then.

app/services/structured_fact_importer.rb
```diff
@@ -1,44 +1,73 @@
...
+      {
+        :parent => parent_fact_record,
+        :compose => fact_value.nil?
+      })
+  end
+
+  def parent_fact_name(child_fact_name)
+    #this regex could be optimized for better performance
+    /(.*)#{FactName::SEPARATOR}/.match(child_fact_name).try(:[], 1)
   end
```
lzap: For sure put it outside of method so it does not need to be recompiled every run.

app/services/fact_importer.rb
```diff
@@ -40,26 +40,62 @@ def initialize(host, facts = {})
...
+      name_record = fact_name_class.find_by!(name: fact_name) unless name_record.save
+
+      # make the new record available immediately for other fact_name_attributes calls
+      @fact_names[fact_name] = name_record
+    end
+  end
+
+  def initialize_fact_names
+    name_records = fact_name_class.where(:name => facts.keys, :type => fact_name_class.to_s).to_a
+    @fact_names = Hash[name_records.map{ |record| [record.name, record]}]
```
tbrisker: You can use `fact_name_class.where(:name => facts.keys, :type => fact_name_class.to_s).group_by(&:name)` to reduce object allocations and clean up the code

lzap (dismissed):
Well I have mixed feelings, because previously I was working on removing eager loading from the codebase here, so less allocations is done:

https://github.com/theforeman/foreman/pull/3667/files#diff-671e5385f8ec83d391c87d7c292604edR119

Allocation counts are the back, numbers are similar:

https://github.com/theforeman/foreman/pull/3667#issuecomment-247603866

Frankly, I start to think that the explicit transaction is the root cause of the problem. If there is no transaction, the code should see changes immediately running into less conflicts.

It was added because it performed exceptionally well on sqlite3 (10x faster) but it did not give better numbers on PostgreSQL. I think we can remove it to see if that helps instead doing this eager loading of *all* fact names (again this can be millions of records loaded into memory every single fact upload request).

lzap:
Additional comment on the explicit transaction - this is my theory we need to confirm: without transaction a process will less likely try to create existing name because there is much smaller window to do this, basically just in the find_or_create method (I think it is not atomic did not go deep into Rails codebase).

But with explicit transaction everything is different, default PostgreSQL behavior is to "freeze" current state for each individual client connection, which can be as long as several seconds (our facter code is still quite slow, usually around one second). Two processes are more likely to see missing fact name, thus this can lead into rollback at the end.

What I am trying to say - let's try to remove transactions from there.

app/services/fact_importer.rb
ShimShtein: It's for `structured_fact_importer` - I want to make sure `a` is created before `a::b`.

app/services/fact_importer.rb
ShimShtein: Default transaction isolation level for postgres is `read committed`, which means other transactions are committed and will be read by `find_by!`.

app/services/fact_importer.rb
ShimShtein: I am not loading the whole table, only the names that exist in the uploaded facts hash (notice the `where(:name => facts.keys)` filter.

app/services/fact_importer.rb
ShimShtein: It is used by inheriting classes, that's why I didn't rename it.

ShimShtein:
> Well I have mixed feelings, because previously I was working on removing eager loading from the codebase here, so less allocations is done

I think I could refactor my code to store only id's in the `fact_names` hash, the problem is with the `find_or_create_fact_name` method. Inheritors are expecting to get an ActiveRecord out of this method, and they are expecting that this method will be queried for each fact name.
There is an option to call them all at once before the import procedure, but it will look ugly in the code base.

ShimShtein:
> On some instances, we see millions of records in this table (usually when there are dynamic NIC interfaces reported by facter)

I am planning on another PR that will filter out these kinds of facts.

ShimShtein:
extracted separator regex.

lzap:
Do you have a reproducer that will prove this patch actually helps? Because I really think it's the transaction which is causing these rollbacks rather than anything in the codebase. Without it the only problematic part should be `find_or_create` (https://apidock.com/rails/v4.0.2/ActiveRecord/Relation/find_or_create_by) which is known to be not atomic, but we can do several retries instead which should do the trick.

app/services/fact_importer.rb
lzap: Then extract it into method which you override in structured fact, so plain does not sort? Just an optimization idea.

app/services/fact_importer.rb
lzap: Yes, that's my concern right there. Find will only see it after commit of the other transaction, which creates one second (or more) window. See line 42 (`ActiveRecord::Base.transaction do`) - that's the problematic block. It's too large.

app/services/fact_importer.rb
lzap: Yeah, sorry, that was the case in old version as well. Hosts with thousands of facts will be all loaded. Current version uses batch processing. Is it really necessary to have that loaded in hash? Can't you process this with `find_each` as well?

app/services/fact_importer.rb
lzap: Not sure what you mean, please rename it and make a deprecated alias or something.

ShimShtein:
The `validate!` was a mistake. I am working on fixing it, stay tuned :smile:

app/services/fact_importer.rb
lzap: To me it looks like the only place you use this is one iteration at `missing_keys = facts.keys - fact_names.keys` and then in find_or_create. But I understand that your main motivation for the patch to speed it up is to actually load fact names into memory. If you confirm, then let's accept that and carefully test this.

lzap:
Shim, I think your patch does not solve the problem, but it does not make it worse and seems to speed up processing. Symptoms described in the ticket are that PostgreSQL is encountering blocked transactions causing some operations during fact import to fail:

1) deadlock detected
2) duplicate key value violates unique constraint "index_fact_names_on_name_and_type" on simulatenous host register
3) Process xxx waits for ShareLock on transaction yyy; blocked by process zzz.
4) current transaction is aborted, commands ignored until end of transaction block

Those errors are usually accompanied with SQL SELECT, UPDATE or DELETE statements on fact_name and they can be triggered by any fact operation (facts upload, rhsm register/refresh/fact update).

I believe this PR is just a generic improvement, therefore I suggest you to create new RedMine issue for that and turn this PR (or file another one) into much more simple patch that I just verified it helps. It can also be easily backported if we decide to do so. Let me explain:

I believe this was introduced by adding SQL transaction to the fact processing. This was added in order to speed up fact processing and it showed improvement with sqlite3 and it had no effect on PostgreSQL, but we did not notice this regression under heavy load.

https://github.com/theforeman/foreman/blob/develop/app/services/fact_importer.rb#L42-L46

I am able to verify with this script that simply commenting out the transaction block (keeping the statements there) helps:

https://gist.github.com/lzap/c84107e245250bf287f09c4f87d8b67c

With the explicit SQL transaction block which was added sometime around Foreman 1.13-1.14 my instances are hitting one of the four problems above. Depending on how you update script, you can hit various issues (all workers update same host, all workers updates same facts names or all workers creates random fact name/values).

Without SQL transaction, my instance settles down with the very same processing times and I am not running into any concurrency issues *at all*. I was testing this with up to 8 concurrent processes.

*You must test this on production instance with PostgreSQL, doing this on sqlite3 is irrelevant.*

ares:
I think splitting this into 2 redmine issues makes sense. Or at least two PRs, commenting transaction would be "Fixes #21599" the second would be "Refs #21599".

just for the reference, it was added in 1.14 by https://github.com/theforeman/foreman/commit/46e1ea9f704f41afbd0373e4b8fc54313ac6b397, the original issue is http://projects.theforeman.org/issues/9016

ShimShtein:
OK Retained the 50% boost, but the names are inserted in isolated transactions as a first step.
Made some improvements to existing queries to eliminate `fact_names` join, since we already have the fact names in memory - we were the ones to push them.

Theoretically, we can avoid retaining `FactName` records completely, but it can cause incompatibility with current plugins. We can do it separately IMHO.

lzap:
Can you please provide some numbers around object allocations as well? I'd be interested in grow rate when you process hosts with thousands of facts: http://projects.theforeman.org/issues/9016

lzap:
Shim, I created PR for the referenced RedMine issue: http://projects.theforeman.org/issues/21599

It's here: https://github.com/theforeman/foreman/pull/4996

I suggest a separate one for this PR, I understand this patch as a generic improvement of parsing in terms of speed. The current version does not fix the concurrency issue, it is just less likely to hit it I believe because your version is faster (smaller transaction window).

lzap:
Damn I did not see the new version, let me take a look.

app/services/fact_importer.rb
```diff
@@ -86,12 +141,12 @@ def add_new_fact(name)
...
   # value is used in structured importer to identify leaf nodes
   def find_or_create_fact_name(name, value = nil)
-    fact_name_class.where(:name => name, :type => fact_name_class.name).first_or_create!
+    fact_names[name]
   end
 
   def add_new_facts
     ActiveSupport::Notifications.instrument "fact_importer_added.foreman", :host_id => host.id, :host_name => host.name, :facts => facts do |payload|
-      facts_to_create = facts.keys - db_facts.pluck('fact_names.name')
+      facts_to_create = facts.keys - db_facts.pluck(:fact_name_id).map { |name_id| fact_names_by_id[name_id].name }
```
lzap: So the key thing here is that you are getting rid of one join, right?

app/services/fact_importer.rb
lzap: This must be really renamed, the name is very misleading. Let's create an alias for that.

app/services/fact_importer.rb
```diff
@@ -39,31 +39,86 @@ def initialize(host, facts = {})
...
+      @fact_names[record.name] = record
+      @fact_names_by_id[record.id] = record
+    end
+  end
+
+  def fact_name_attributes(fact_name)
+    {
+      name: fact_name
+    }
+  end
```
lzap, line 113: Is really a method call needed for this? I'd rather put this inline.

app/services/fact_importer.rb
```diff
@@ -39,31 +39,86 @@ def initialize(host, facts = {})
...
+
+  def initialize_fact_names
+    name_records = fact_name_class.where(:name => facts.keys, :type => fact_name_class.to_s).to_a
+    @fact_names = {}
+    @fact_names_by_id = {}
+    name_records.each do |record|
+      @fact_names[record.name] = record
+      @fact_names_by_id[record.id] = record
+    end
+  end
```
lzap: Please use streaming API `find_each` (https://apidock.com/rails/ActiveRecord/Batches/find_each) and avoid `to_a` completely. That is N object allocations saved per request (where N = number of facts). Also pre-allocate target hash to some reasonable size, I suggest 50.

app/services/fact_importer.rb
```diff
@@ -39,31 +39,86 @@ def initialize(host, facts = {})
...
+      end
+
+      # make the new record available immediately for other fact_name_attributes calls
+      @fact_names[fact_name] = name_record
+    end
+  end
+
+  def duplicate_name_error?(name_record)
+    name_record.errors.size == 1 && name_record.errors.details[:name].first.try(:[], :error) == :taken
+  end
```
lzap: Nice solution indeed, kudos. But frankly, my opinion on this is that the whole sub-transaction (and savepoint) is not needed if we decide to ditch transactions. Parsing errors and rollbacking on purpose is not ideal way to do things - rollbacks are supposed to be for error states and not to solve concurrency issues. There are 

Current SQL standards support atomic operation to create or find a record in database (e.g. WHERE NOT EXISTS or  ON CONFLICT (index) DO NOTHING introduced in 9.5) , the problem here is Rails which does not support it. This is kinda hacky and I can't tell how this will behave in real-word environments - we might see transaction log to grow quickly or something unexpected.

Here is a proposal - how about adding a flag (settings.yaml not db setting) where user can opt-out transactions completely, so the explicit transaction and this block would have been disabled in this mode. We can cover both with unit/integration tests easily (just run them twice with and without). This will be safe to merge.

app/services/fact_importer.rb
lzap: I see now, still a quite expensive step for this request. Is it possible to avoid this? Let's say you won't construct `missing_keys` and iterate through `facts.keys` directly making sure that parents gets created first. Just an idea, this would also save up to N allocations and then sort.

lzap (changes requested):
How about making this optional, @ares what is your opinion?

app/services/fact_importer.rb
ShimShtein: > Let's say you won't construct missing_keys and iterate through facts.keys directly making sure that parents gets created first.

1. `missing_keys` is a list of strings (that could be frozen if you wish), and it should be relatively small - most of fact reports do not introduce new facts (repeating reports for the same host is a good example). Especially when I intend to filter out "randomized" names like virtual interfaces.
2. Since I want to make sure parents are created before children, it means sorting the whole `facts.keys` which is much larger than the 'missing_keys' collection.

app/services/fact_importer.rb
ShimShtein: > Current SQL standards support atomic operation to create or find a record in database

Unfortunately it's not SQL standard, and will lock us to PG (especially a quite advanced version), which IMHO should be avoided (even at a cost of couple ms at each request).

The main problem here is not the statement that will fail (even without a transaction), but the connection that will be rendered useless, since until the connection is released it will be in a failed state. One of the errors you were seeing was indeed about trying to save a record from such connection. This way I keep the damage contained to the single statement.

As a general statement, I really want to get the whole facts processing out of foreman into a separate process - it will be a lot slimmer than spawning a new full blown rails app for each fact report. Maybe even consider a different storage type for this kind of information. Although it's a long term goal, and I don't think this PR is the place to start such a discussion.

app/services/fact_importer.rb
ShimShtein: All records will still exist in memory in the end of the loop - they are referenced by both hashes.

app/services/fact_importer.rb
ShimShtein: Yes, it is overridden in `structured_fact_importer` to add more fields to fact name table (hierarchy for example).

app/services/fact_importer.rb
ShimShtein: I see that most important plugins (katello and discovery) are not using this method. We can try and remove this method completely - maybe with an exception message that will point out to the new way of adding fact names.

app/services/fact_importer.rb
ShimShtein: Right. It's a byproduct of having all fact names already cached in memory.

ShimShtein:
@lzap I have improved the memory footprint to something more or less equal to the previous version. Also added the memory benchmark test.

I was thinking about plucking only the name and id for fact names (instead of retaining the whole `FactName` model), but in structured fact importer we are using ancestry gem that will fetch the record, if we supply `:parent_id => fact_name_id` statement.

lzap:
Roman tested just removing the transaction block and it was able to still reproduce the issue, we don't have numbers about rates (I expect them to appear less likely), but it is still there. Let me take a look on your patch again today. https://bugzilla.redhat.com/show_bug.cgi?id=1474293

lzap:
@ShimShtein can you share outputs from benchmarks here for comparison?

app/services/fact_importer.rb
lzap: Yeah I thought so. Thanks.

app/services/fact_importer.rb
```diff
@@ -39,31 +39,90 @@ def initialize(host, facts = {})
...
+
+      ActiveRecord::Base.transaction(:requires_new => true) do
+        begin
+          unless name_record.save(:validate => false)
+            throw ActiveRecord::RecordInvalid.new(name_record) unless duplicate_name_error?(name_record)
+
+            # if the record could not be saved, then it should already be in the database
+            # throw an exception if it isn't
+            name_record = fact_name_class.find_by!(name: fact_name)
+          end
```
lzap: As per our call, this block can go now.

test/benchmark/memory/puppet_fact_importer_benchmark.rb
```diff
@@ -0,0 +1,36 @@
...
+
+facts = generate_facts(200, 50, 25)
+User.current = User.unscoped.find(1)
+host = FactoryBot.create(:host, :name => "benchmark-#{Foreman.uuid}")
+# binding.pry
+# report = MemoryProfiler.report do
+StackProf.run(mode: :object, raw: true, out: '/tmp/stacprof_objects.dump', interval: 1) do
+  StructuredFactImporter.new(host, facts).import!
+end
+# report.pretty_print
```
lzap: As per discussion, please remove `pry` and make this a switch or something.

lzap (changes requested):
Ok looks good, let's keep the transactions for now. Please remove the block and I am good to test.

ShimShtein:
Removed unnecessary save handling from `add_missing_fact_names`.

Added memory profiling benchmark.

ShimShtein:
OK Here are results of running the same benchmark on develop branch VS this patch.

TL;DR: Total amount of allocated memory and objects decreased, but the amount of retained memory and objects slightly increased:

```
# Before patch:
Total allocated: 63532480 bytes (899053 objects)
Total retained:  3036079 bytes (28250 objects)

# After patch:
Total allocated: 55261835 bytes (771221 objects)
Total retained:  3675785 bytes (33373 objects)
```
Attached full reports FYI

[after.txt](https://github.com/theforeman/foreman/files/1492549/after.txt)
[before.txt](https://github.com/theforeman/foreman/files/1492550/before.txt)

app/services/fact_importer.rb
lzap: But you can still avoid calling `to_a` it will create a useless array of pointers for no reason.

Anyway, I insist on not creating expensive AR objects here. For `fact_names_by_id` you can only store name, because the only reference you do is `fact_names_by_id[X].name` and there is no reason to store everything.

For `fact_name` you can also avoid storing whole AR and just store ID, but it needs to be proofed by unit tests. You can do this:

```ruby
    fact_name_id = fact_names[name]
    host.fact_values.send(method, :value => facts[name], :fact_name_id => fact_name_id)
```

Another use is here which you can also rewrite to faster:

    host.fact_values.where(:fact_name_id => fact_names.values)

And the last use can be also rewritten to:

    parent_fact_record_id = fact_names[parent_fact_name(name)]
    attributes[:parent_id] = parent_fact_record_id

This will be I believe much faster. Then you can use batch processing and avoid AR completely, also you can pluck only columns you are interested in (name, id).

app/services/fact_importer.rb
```diff
@@ -39,31 +39,84 @@ def initialize(host, facts = {})
 
   # expect a facts hash
   def import!
+    ensure_fact_names
```
lzap: Can you leave a comment this is out of the transaction on purpose please?

app/services/fact_importer.rb
```diff
@@ -39,31 +39,84 @@ def initialize(host, facts = {})
...
-      raise ::Foreman::WrappedException.new(@error, N_("Import of facts failed for host %s"), @host.name)
-    end
+    report_error(@error) if @error
     logger.info("Import facts for '#{host}' completed. Added: #{counters[:added]}, Updated: #{counters[:updated]}, Deleted #{counters[:deleted]} facts")
   end
 
+  def report_error(error)
+    Foreman::Logging.exception("Error during fact import for #{@host.name}", error)
+    raise ::Foreman::WrappedException.new(error, N_("Import of facts failed for host %s"), @host.name)
+  end
```
lzap, line 59: This was extracted but never used, can you leave it inline if this was not intentional?

app/services/fact_importer.rb
lzap: You can now delete `duplicate_name_error` which is not used anymore.

app/services/structured_fact_importer.rb
```diff
@@ -1,44 +1,75 @@
...
+    parent_fact_record = fact_names[parent_fact_name(name)]
+
+    attributes[:parent] = parent_fact_record
+    attributes[:compose] = fact_value.nil?
+    attributes
+  end
+
+  def separator_regex
+    #this regex could be optimized for better performance
+    @separator_regex ||= /(.*)#{FactName::SEPARATOR}/
```
lzap: Let's remove this comment and replace this with more than twice faster method using rindex:

```ruby
require 'benchmark'
r = /(.*)::/
Benchmark.bm do |x|
    x.report(:regexp) { 990000.times { a = 'a::b::c::d::test::longer_test::e'.match(r)[0] } }
    x.report(:index) { 990000.times {
      a = 'a::b::c::d::test::longer_test::e'
      split = a.rindex(':')
      a = a[0,split-1] if split
    }}
end

[lzap@box foreman]$ ruby test.rb 
       user     system      total        real
regexp  0.940000   0.000000   0.940000 (  0.938746)
index  0.380000   0.000000   0.380000 (  0.384979)
```

You can get rid of regexp completely and replace this with three lines of code.

lzap (changes requested):
I still have some optimization comments. If you think that getting rid of AR instances is not doable, get rid of useless `to_a` at least and optimize the regexp please.

app/services/fact_importer.rb
ShimShtein: It is extracted to reduce ABC metric of `import!`

ShimShtein:
@lzap, I think I have answered/fixed all your concerns.

ares:
Since #5007 has been merged and is linked to the same redmine issue and branching date is getting closer, @ShimShtein could you please create a new redmine ticket to track these improvements?

ares:
also it seems it needs rebasing after the transaction was dropped

ShimShtein:
The branch was deleted unintentionally, the PR was reopened as #5030.

PR #4997, opened 2017-11-09
Fixes #21624 - eslint update to airbnb-config

danseethaler:
After having some discussion on the [foreman-dev](https://groups.google.com/forum/#!topic/foreman-dev/RIVmSdcpA7Q) and working with our `eslint` setup here's what I'm proposing.

### Integrate airbnb-base
We integrate the `eslint-config-airbnb-base` and remove all of our custom rules (which I've done in this PR). We also update all of our existing files (in the `webpack/` and `script/` directories) to follow those rules as much as possible. Any rules that we aren't upgrading to match we turn off in the `.eslintrc` for now.

There are two rules I had to turn off because of [this issue](https://github.com/houndci/linters/issues/62) in Hound. It's bummer because they are really helpful rules (`import/no-unresolved` and `import/extensions`) but for the time being Hound will through erroneous errors (I've always wanted to say that) if we leave them in.

### Upgrade to full airbnb-config
The next step will be another PR which upgrades to the full `eslint-config-airbnb` which includes `eslint-plugin-react` and `eslint-plugin-jsx-a11y`. This PR will not include file updates to match the rules but will assume that files will be made compliant whenever they are worked on. It would be nice if this PR also added an npm script to lint just the files that are staged in git. This will serve as a helper to check anything before HoundCI gets to it.

### Integrate eslint-plugin-prettier
Hopefully by this time we'll have figured out a solution to [this Hound issue](https://github.com/houndci/hound/issues/1461) and will be able to introduce `eslint-plugin-prettier`.

Here's the Redmine: http://projects.theforeman.org/issues/21624


package.json
```diff
@@ -89,6 +91,7 @@
   "scripts": {
     "lint": "./node_modules/.bin/eslint -c .eslintrc webpack/ script/ || exit 0",
     "postlint": "node ./script/npm_lint_plugins.js",
+    "lint:local": "./node_modules/.bin/eslint -c .eslintrc webpack/ script/ || exit 0",
```
ohadlevy: whats the difference to lint task?

danseethaler: This is just something I've been using locally to lint only foreman (skips the `postlint` script). I have some questions about the `postlint` process explained [here](https://github.com/theforeman/foreman/pull/4946#issuecomment-343533540) but I'm happy to remove this if it's not helpful to others.

ohadlevy:
Isn't it the same as the lint command above it?

ohadlevy:
@danseethaler can you rebase please?

danseethaler:
@ohadlevy yeah it's the same except that it skips the `postlint` script. Like I mentioned I'm not quite sure I understand the purpose of the `postlint` process (see [my question here](https://github.com/theforeman/foreman/pull/4946#issuecomment-343533540)) but I'll remove `lint:local` for now.

danseethaler:
[test foreman]

danseethaler:
@ohadlevy or anyone else have any comments before we merge?

webpack/assets/javascripts/react_app/components/common/Alert/Alert.consts.js
```diff
@@ -1,10 +1,11 @@
 /* eslint-disable no-throw-literal */
+/* eslint-disable no-throw-literal */
```
ohadlevy: Duplicate line?

webpack/assets/javascripts/foreman_navigation.test.js
```diff
@@ -69,21 +70,13 @@ describe('initNavigation', () => {
...
            Org1 <span class="caret"></span>
          </a>
-       </li>         
+       </li>
        ${menuWithoutTaxonomies}
      </div>`;
   it('Mark current location/organization as active in vertical menu', () => {
     init();
-    expect(
-      $('.nav-pf-secondary-nav .list-group-item:contains("Loc1")').hasClass(
```
ohadlevy, line 78: Any idea why this gets changed? Afair this is already a pritter output?

ohadlevy (changes requested):
Minor comment thanks

tbrisker (commented):
Are you sure all the changes in the `script` directory are compatible with all our builders? the code there isn't transpiled by babel so we should be very careful that all of it is backwards-compatible enough for the oldest node version we use in build time

webpack/assets/javascripts/react_app/components/common/Alert/Alert.consts.js
danseethaler: Good catch.

webpack/assets/javascripts/foreman_navigation.test.js
danseethaler: Yeah, it's changed because the eslint rule for max-len defined [here](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/style.js#L179) in the airbnb-base config says that strings are ignored. And since there are strings in this line it is ignored. I don't really agree with that setup but that's why we should get pretty integrated into our eslint config.

danseethaler:
@tbrisker thanks for bringing that up. I can see three builds in TravisCI starting a node version 6. Are there any older versions you are aware of? Node v6 supports `const` and arrow functions and the like so I think we should be good, but I'm not familiar with our build process so I can't say for certain.

ohadlevy (approved):

PR #4998, opened 2017-11-10
Fixes #21626 - "Domain Users" are not a valid AD group

dLobatog:
'Domain Users' is a special group in AD. This group users' cannot
be queried through regular LDAP, it can only be seen on the Windows
AD UI.

This confuses people who think that could add this group but they
find that no users are found after adding this group.

This PR adds a warning when you try to do that, so that hopefully we
don't get more bug reports about this.

### DIFF
--- a/app/models/external_usergroup.rb
+++ b/app/models/external_usergroup.rb
@@ -10,6 +10,7 @@ class ExternalUsergroup < ApplicationRecord
   validates :name, :auth_source, :usergroup, :presence => true
   validate :hidden_authsource_restricted
   validate :in_auth_source?, :if => Proc.new { |eu| eu.auth_source.respond_to?(:valid_group?) }
+  validate :domain_users_forbidden
 
   def refresh
     return false unless auth_source.respond_to?(:users_in_group)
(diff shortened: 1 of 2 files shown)


app/models/external_usergroup.rb
```diff
@@ -37,14 +38,26 @@ def users
   private
 
   def in_auth_source?(source = auth_source)
-    errors.add :name, _("is not found in the authentication source") unless source.valid_group?(name)
-  rescue Net::LDAP::Error, Net::LDAP::LdapError => e
-    errors.add :auth_source_id, _("LDAP error - %{message}") % { :message => e.message }
+#    errors.add :name, _("is not found in the authentication source") unless source.valid_group?(name)
+#  rescue Net::LDAP::Error, Net::LDAP::LdapError => e
+#    errors.add :auth_source_id, _("LDAP error - %{message}") % { :message => e.message }
```
ares: I suppose this should not be commented

ares (dismissed):
A great addition, I suppose the commented lines causes the test failures, setting WoC

app/models/external_usergroup.rb
dLobatog: Whoa - indeed, I did comment them to test in any auth source

dLobatog:
@ares Updated removing the commented validator

tbrisker (approved):

tbrisker:
Thanks @dLobatog !

PR #4999, opened 2017-11-10
Fixes #21629 - display error for taxonomy selectors

ares:
before:
a_no_error

after:
a_error

reproducing steps provided in redmine issue, the issue is that we passed 'organizations' as attr, in field helper we do f.object.errors[attr] so it needs to be 'organization_ids'


tbrisker:
@ares can you take a look at the test failure? seems related.

ares:
Thanks @tbrisker, I udpated the helper so it can deal with both ways of attr specification. All looks green now.

tbrisker (approved):
Tested, works well

tbrisker:
Thanks @ares !

PR #5000, opened 2017-11-10
Fixes #21630 - Add colocation for storybook stories

danseethaler:
This moves the storybooks stories to colocate with the
corresponding component file. This makes it a little
nicer to manage and follows the pattern in the patternfly-react
repository.

http://projects.theforeman.org/issues/21630

### DIFF
--- a/webpack/stories/index.js
+++ b/webpack/stories/index.js
@@ -1,104 +1,12 @@
-import React from 'react';
-
-import { storiesOf } from '@storybook/react';
-import { action } from '@storybook/addon-actions';
-
 require('../assets/javascripts/bundle');
 require('../../app/assets/javascripts/application');
 require('../../app/assets/stylesheets/base.scss');
-import ChartBox from '../assets/javascripts/react_app/components/statistics/ChartBox';
-import PieChart from '../assets/javascripts/react_app/components/common/charts/PieChart';
-import mockData from './data/charts/donutChartMockData';
-import PowerStatusInner from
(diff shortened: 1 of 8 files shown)


webpack/stories/index.js
```diff
@@ -1,104 +1,9 @@
-import React from 'react';
+import { configure } from '@storybook/react';
 
-import { storiesOf } from '@storybook/react';
-import { action } from '@storybook/addon-actions';
+const req = require.context('../assets/javascripts/react_app', true, /.stories.js$/);
 
-require('../assets/javascripts/bundle');
-require('../../app/assets/javascripts/application');
-require('../../app/assets/stylesheets/base.scss');
```
danseethaler: I'm not sure if these three lines are necessary. I don't see any changes when they're removed.

ohadlevy: the idea behind it was to reuse the current foreman asset pipline(rails) js and css. AFAIR they are used in a few places. (this would allow a difference between the stories and the app)

ohadlevy (changes requested):
great addition - it anonyed me for a while :) thanks for fixing that - if you could re-include the current rails application.js and base css I'll merge :)

danseethaler:
> great addition - it anonyed me for a while :) thanks for fixing that - if you could re-include the current rails application.js and base css I'll merge :)

Awesome - thanks @ohadlevy! Those lines are back in there 👍🏼

ohadlevy (approved):

PR #5001, opened 2017-11-13
Fixes #22038 - Fix color for link in action button

bastilian:
This addresses the issue introduced by allowing primary action buttons, which does not set the right color when links are used inside. Like the import roles button in foreman_ansible.

Before:

gnome-shell-screenshot-h06q9y

gnome-shell-screenshot-ukgf9y

After:
gnome-shell-screenshot-edoc9y
gnome-shell-screenshot-bpat9y

### DIFF
--- a/app/assets/stylesheets/patternfly_and_overrides.scss
+++ b/app/assets/stylesheets/patternfly_and_overrides.scss
@@ -377,6 +377,12 @@ a.btn-info span {
   }
 }
 
-.container-pf-nav-pf-vertical.hide-nav-pf {
-  display: none !important;
+span.btn-action a {
+  display: block;
+  margin: -2px -6px;
+  padding: 2px 6px;
+}
+
+span.btn-action.btn-primary a {
+  color: #fff;
 }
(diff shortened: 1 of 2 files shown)


boaz0 (approved):
LGTM minus the comments by houndci-bot

tbrisker:
@bastilian from the screenshots it looks like the issue is in the ansible plugin, i think it might be a matter of incorrectly using the class - you can just give the various `btn` classes to the `a` element, instead of wrapping an `a` in a `btn`

bastilian:
@tbrisker up until now the `select_action_button` has not been used as `primary`, the import button is (as far as I've seen) the first use of it. The use of an `a` inside of the `btn` is also in https://github.com/theforeman/foreman/blob/eb0834f5007f60493278b499e27eceb9244e87e1/app/helpers/puppetclasses_and_environments_helper.rb#L20-L25. It would make sense to change this and allow setting links on the `btn` rather than having to use a's. It might be more effort and maybe makes sense to separate it from this.

tbrisker (approved):

tbrisker:
Ok, makes sense I guess. Please rebase and update the issue number to an issue in foreman project on redmine.

dLobatog (approved):
I will change the issue number to http://projects.theforeman.org/issues/22038 upon merge, thanks @bastilian @tbrisker

PR #5002, opened 2017-11-13
Fixes #21643 - fix and move nfs_visibilty.js to webpack

amirfefer:
nfs section does not become visible when choosing a required nfs os family in installation media form.
This occurs due to broken script file - nfs_visibilty.js, which apparently doesn't work since moving to a select box from checkbox to choose an os family.
I re-wrote the logic and moved it to wepback


webpack/assets/javascripts/foreman_medium.js
```diff
@@ -0,0 +1,9 @@
+import $ from 'jquery';
+
+export function nfsVisibility(osFamily, nfsRequired) {
+    if (nfsRequired[osFamily.value]) {
+      $('#nfs-section').show();
+    } else {
+      $('#nfs-section').hide();
+    }
```
tbrisker: you can make this a one liner:
```js
$('#nfs-section').toggle(nfsRequired[osFamily.value]);
```

app/views/media/_form.html.erb
```diff
@@ -17,18 +16,15 @@
       <%= text_f f, :name %>
       <%= text_f f, :path, :size => "col-md-8", :help_block => _("The path to the medium, can be a URL or a valid NFS server (exclusive of the architecture).
           for example <em>http://mirror.centos.org/centos/$version/os/$arch</em> where <strong>$arch</strong> will be substituted for the host's actual OS architecture and <strong>$version</strong>, <strong>$major</strong> and <strong>$minor</strong> will be substituted for the version of the operating system. Solaris and Debian media may also use <strong>$release</strong>.").html_safe %>
-        <span id="nfs-section" <%= display?(!@medium.operatingsystems.map(&:require_nfs_access_to_medium).any?) %>>
+        <span id="nfs-section" <%= display?(!@medium.operatingsystems.map{ |obj| obj.class.require_nfs_access_to_medium }.any?) %>>
```
tbrisker: This won't work if the medium has solaris os_family but no operating systems assigned.
Why not also leverage the `required_nfs_list` helper here?
p.s. you can use `any?` with a block instead of `.map{..}.any?`

app/helpers/mediums_helper.rb
```diff
@@ -1,3 +1,9 @@
 module MediumsHelper
   include PtablesHelper
+
+  def required_nfs_list
+    Hash[Operatingsystem.families.map do |key|
+      [key, key.constantize.require_nfs_access_to_medium]
+    end].to_json
+  end
```
tbrisker: Why not just return a list of all families requiring it and check if the selected family is in it or not?

tbrisker (changes requested):
See comments inline.
Also, IIRC there was some dead css code you can get rid of while you're here.

amirfefer:
@tbrisker , Thanks for your review, could you have a second look?

app/helpers/mediums_helper.rb
```diff
@@ -1,3 +1,12 @@
 module MediumsHelper
   include PtablesHelper
+
+  def required_nfs_list
+    Operatingsystem.families.select {|family| family.constantize.require_nfs_access_to_medium }
+  end
+
+  def required_nfs?
+    required_nfs_list.include?(@medium.os_family) ||
+    @medium.operatingsystems.any?{ |obj| obj.class.require_nfs_access_to_medium }
```
tbrisker: The second part of this check is redundant, you can only add operating systems that match the family

tbrisker (changes requested):
Just one minor comment, other than that works well.

amirfefer:
@tbrisker , Thanks for your review,
could you have a second look?

tbrisker (approved):
Looks good, will merge once tests pass.

tbrisker:
Thanks @amirfefer !

PR #5003, opened 2017-11-14
Fixes #20891 - Remove hostname from /hosts/x/facts results

shiramax:


dLobatog (changes requested):
@shiramax Thanks for the PR!

Notice that `fact_values_controller.rb` and `FactValue.build_facts_hash` are serving 2 routes, the general one (`api/v2/fact_values`) and the particular one 

This may be correct for the particular case (`api/v2/hosts/:host_id/facts`) because hosts have unique key-value pairs. 

In the general case, this does not work, because multiple hosts will have different fact values. E.g: 

* Host A and Host B. 
* Host A has a fact "memory_mb", value "2048"
* Host B has the same fact "memory_mb", value "4092"

With this change, the output of `api/v2/fact_values` changes from:

```json
{ "hostA" : { "memory_mb" : 2048 },  "hostB" : { "memory_mb" : 4092 },  }
```

to 

```json
{ "memory_mb" : 4092 }
```

Please ensure in tests that both routes still work correctly after this patch, you're on the right path as you identified the method causing the discrepancy between `api/v2/hosts/:host_id/interfaces` and `api/v2/hosts/:host_id/facts`

timogoebel:
Please also add tests for the api controller (both routes) that actually check the response json content and structure. I'll give you an example:

```
context '#index' do
      test 'should show dlmlocks' do
        dlmlock = FactoryBot.create(:dlmlock, :host => host1)
        get :index
        assert_response :success
        body = ActiveSupport::JSON.decode(@response.body)
        results = body['results']
        assert results
        entry = results.detect { |entry| entry['id'] == dlmlock.id }
        assert entry
        assert_equal dlmlock.name, entry['name']
        assert_equal dlmlock.type, entry['type']
        assert_equal dlmlock.enabled, entry['enabled']
        assert_equal dlmlock.host_id, entry['host_id']
      end
    end
```

While this patch makes the API response consistent with other endpoints, we should properly deprecate the old behavior imho.

shiramax:
@dLobatog , thanks 
Done

@timogoebel 
I didn't really understand what's the problem with the current tests

timogoebel:
@shiramax: Tests are ok. I just wanted to make sure, the json structure is actually tested.

app/models/fact_value.rb
```diff
@@ -81,13 +81,16 @@ def self.count_each(fact, options = {})
     output
   end
 
-  def self.build_facts_hash(facts)
+  def self.build_facts_hash(facts, add_host)
```
dLobatog: In order to allow this method to be backwards compatible (or if any plugin is using it, making sure it still works), please change it to

```ruby
def self.build_facts_hash(facts, add_host=true)
```

IMO `include_hostname` is a better name for it but don't take my word for it..

app/models/fact_value.rb
```diff
@@ -81,13 +81,16 @@ def self.count_each(fact, options = {})
...
     hosts = Host.where(:id => facts.group_by(&:host_id).keys).all
```
dLobatog: This `.where` call is going to be called even when `add_host` is passed, even though it's not used, can you ensure it's only used when needed?

dLobatog (changes requested):
Thanks for the update @shiramax , just a few minor comments inline.

test/controllers/api/v1/fact_values_controller_test.rb
```diff
@@ -18,15 +18,15 @@ def setup
     get :index, {:host_id => @host.name }
     assert_response :success
     fact_values   = ActiveSupport::JSON.decode(@response.body)
-    expected_hash = FactValue.build_facts_hash(FactValue.where(:host_id => @host.id))
+    expected_hash = FactValue.build_facts_hash(FactValue.where(:host_id => @host.id), false)
```
timogoebel: On second thought, I think these tests should not rely on calling `FactValue.build_facts_hash` to get the expected value. Could you please change them to an explicit check?
If `FactValue.build_facts_hash` returns wrong data, the tests will still pass.

shiramax: thanks, 
shouldn't the explicit check will be to build the hash exactly the same as in build_facts_hash?

timogoebel: I would not build or generate the hash at all, but provide it completely hardcoded.

app/models/fact_value.rb
```diff
@@ -81,13 +81,16 @@ def self.count_each(fact, options = {})
...
-  def self.build_facts_hash(facts)
-    hosts = Host.where(:id => facts.group_by(&:host_id).keys).all
-
+  def self.build_facts_hash(facts, include_host = true)
     hash = {}
     facts.each do |fact|
-      hash[hosts.detect {|h| h.id == fact.host_id}.to_s] ||= {}
-      hash[hosts.detect {|h| h.id == fact.host_id}.to_s].update({fact.name.to_s => fact.value})
+      if include_host
+        hosts = Host.where(:id => facts.group_by(&:host_id).keys).all
```
dLobatog: This is making a `.where` call for every fact, we don't need this either, just once at the beginning is fine. Remember it's only needed when `include_host`

dLobatog (changes requested):
Tiny change needed to avoid making `Host.where` calls on every fact, appreciate the quick feedback cycle, thanks @shiramax

ShimShtein:
@dLobatog Is it OK to change API response? I think we shouldn't do it (at least try not to...). It will break this API consumers. IMHO it would be better to create a new endpoint for the new response.

timogoebel:
@ShimShtein: I think, you are right as I wrote before. The least we can do is document the change in the upgrade nodes.
Just a short story from running Foreman in production: When we added Katello to our environment, the API response for the v1 API changed. We still use the v1 API in our deployment tooling to figure out on what servers our applications are running. So, installing Katello broke the deployment for the most part of our applications. So I had 100 angry developers who could not deploy their applications and blamed Foreman for that. I think we could easily avoid something like this.

shiramax:
@dLobatog , since you open the redmine issue, what do you think? should we close this PR?

dLobatog (changes requested):
@timogoebel @ShimShtein 

It's true changing the API is a bad practice, my problem with this endpoint in particular, (api/v2/hosts/:id/facts) is that is inconsistent with /interfaces, /organizations, etc.. or any other API on the host level - which is a bug.

I appreciate your concern, so maybe what we can do is offer both options in the same response, and add a deprecation warning explaining the endpoint will have the same format as all other host-level endpoints for 1.18. For example: 

HostA - only has one fact "memory_mb": "2000"

The format until 1.18 would be:

```
{
  "total": 1,
  "subtotal": 1,
  "page": 1,
  "per_page": 9999999,
  "search": " host = hosta.example.lan",
  "sort": {
    "by": null,
    "order": null
  },
  "results": 
{ 
"memory_mb": "2000", 
"hosta.example.lan": { "memory_mb": "2000" } 
} 
```

Does this sound reasonable to you @shiramax ? We have an internal API to deprecate these things easily - https://theforeman.org/handbook.html#APIusabilityandversioning - apologies for how long merging this is taking 😅

ShimShtein:
@dLobatog, :+1: for your "why not both" suggestion, small request though:

> We have an internal API to deprecate these things easily 

I don't think it's wise to write a deprecation message each time this API is requested. The user has no option of disabling it, even if he does not use the deprecated property.

shiramax:
@ShimShtein , @dLobatog 
Done, please review again:)

ShimShtein:
@shiramax Tests need to be changed too.

shiramax:
@ShimShtein, can you please review?

app/controllers/api/v1/fact_values_controller.rb
```diff
@@ -17,7 +17,8 @@ def index
           no_timestamp_facts.
           search_for(*search_options).paginate(paginate_options).
           preload(:fact_name, :host)
-        render :json => FactValue.build_facts_hash(values.all)
+        has_host = !params[:host_id].blank?
```
ShimShtein: `params[:host_id].present?` looks better.

app/models/fact_value.rb
```diff
@@ -81,13 +81,15 @@ def self.count_each(fact, options = {})
     output
   end
 
-  def self.build_facts_hash(facts)
-    hosts = Host.where(:id => facts.group_by(&:host_id).keys).all
-
+  def self.build_facts_hash(facts, has_host = false)
     hash = {}
     facts.each do |fact|
+      hosts = Host.where(:id => facts.group_by(&:host_id).keys).all
```
ShimShtein: This line should be outside of `each`, otherwise it will be executed once per each fact.

ShimShtein (changes requested):

iNecas:
@ShimShtein @timogoebel @dLobatog anything else to add?

app/controllers/api/v2/fact_values_controller.rb
```diff
@@ -14,7 +14,8 @@ def index
           no_timestamp_facts.
           search_for(*search_options).paginate(paginate_options).
           preload(:fact_name, :host)
-        @fact_values = FactValue.build_facts_hash(values.all)
+        has_host = !params[:host_id].blank?
```
ShimShtein: I think `present?` looks better here.
``` ruby
has_host = params[:host_id].present?
```

app/models/fact_value.rb
```diff
@@ -81,13 +81,16 @@ def self.count_each(fact, options = {})
...
   end
 
-  def self.build_facts_hash(facts)
-    hosts = Host.where(:id => facts.group_by(&:host_id).keys).all
-
+  def self.build_facts_hash(facts, has_host = false)
     hash = {}
+    hosts = Host.where(:id => facts.group_by(&:host_id).keys).all  if has_host
     facts.each do |fact|
+      hosts = Host.where(:id => facts.group_by(&:host_id).keys).all unless has_host
```
ShimShtein: This line should be outside of each, otherwise it will be executed once per each fact.
I can see that the condition is unnecessary too - you are performing the same query whether `has_host` is true or not.

app/models/fact_value.rb
```diff
@@ -81,13 +81,16 @@ def self.count_each(fact, options = {})
     output
   end
 
-  def self.build_facts_hash(facts)
-    hosts = Host.where(:id => facts.group_by(&:host_id).keys).all
-
+  def self.build_facts_hash(facts, has_host = false)
```
ShimShtein: I think we need more descriptive name than `has_host`. I would suggest `single_host`.

ShimShtein (changes requested):
Some changes still needed.

timogoebel:
Looks good to me, thanks. :-)

app/models/fact_value.rb
```diff
@@ -81,13 +81,15 @@ def self.count_each(fact, options = {})
     output
   end
 
-  def self.build_facts_hash(facts)
-    hosts = Host.where(:id => facts.group_by(&:host_id).keys).all
-
+  def self.build_facts_hash(facts, single_host = false)
     hash = {}
+    hosts = Host.where(:id => facts.group_by(&:host_id).keys).all
```
ShimShtein: We can improve performance quite a lot if we don't load fact value objects to memory at all:
``` ruby
hosts = Host.where(:id => facts.select(:host_id).uniq).all
```

ShimShtein (changes requested):

shiramax:
@ShimShtein, thanks, i changed it, 
could you please explain where in this line I load fact value objects to memory? 
hosts = Host.where(:id => facts.group_by(&:host_id).keys).all

Thanks!!

ShimShtein:
@shiramax, `facts` comes from [fact_values_controller](https://github.com/theforeman/foreman/blob/6b7bdf68ac3fbc254dd4911652c2ff2c344d63b7/app/controllers/api/v2/fact_values_controller.rb#L12), and is a scope. In the old code, you have invoked `group_by` which is a method from `Enumerable` that causes actual query to the DB, hence loads all the `FactValue` objects into the memory.
Easiest way to check this, is enable SQL logging and running the query:

``` ruby
# set the logger
ActiveRecord::Base.logger = Logger.new(STDOUT)

Host.where(:id => FactValue.all.group_by(&:host_id).keys).all
# DEBUG -- :   FactValue Load (1.2ms)  SELECT "fact_values".* FROM "fact_values"
# DEBUG -- :   Host::Managed Load (0.8ms)  SELECT "hosts".* FROM "hosts" WHERE "hosts"."type" IN ('Host::Managed') AND 1=0

Host.where(:id => FactValue.unscoped.select(:host_id).distinct).all
# DEBUG -- :   Host::Managed Load (1.9ms)  SELECT "hosts".* FROM "hosts" WHERE "hosts"."type" IN ('Host::Managed') AND "hosts"."id" IN (SELECT DISTINCT "fact_values"."host_id" FROM "fact_values")
```

Also notice that it should be `distinct` and not `uniq`, sorry for misleading.

shiramax:
@ShimShtein please review :)

1
```diff
@@ -0,0 +1,13 @@
+Fixes #20891 - Remove hostname from /hosts/x/facts results
```
ShimShtein: I think it's not intended to be here

app/models/fact_value.rb
```diff
@@ -81,13 +81,15 @@ def self.count_each(fact, options = {})
     output
   end
 
-  def self.build_facts_hash(facts)
-    hosts = Host.where(:id => facts.group_by(&:host_id).keys).all
-
+  def self.build_facts_hash(facts, single_host = false)
     hash = {}
+    hosts = Host.where(:id => FactValue.unscoped.select(:host_id).distinct).all
```
ShimShtein: This selects *all* hosts that have fact values, does it make sense?

ShimShtein (changes requested):
Still needs to be fixed. Especially the query, it looks a bit odd.

shiramax:
@ShimShtein ,
Thanks, Done.
Please review again :)

shiramax:
[ test foreman]

shiramax:
[test foreman]
[test katello]

app/models/fact_value.rb
```diff
@@ -81,13 +81,19 @@ def self.count_each(fact, options = {})
     output
   end
 
-  def self.build_facts_hash(facts)
-    hosts = Host.where(:id => facts.group_by(&:host_id).keys).all
-
+  def self.build_facts_hash(facts, host_id = nil)
     hash = {}
+    if host_id
+      hosts = Host.where(:id => host_id).empty? ? Host.where(:name => host_id) : Host.where(:id => host_id)
```
ShimShtein: @tbrisker, can you think of a better way for selecting a host either by its name or id?

ShimShtein (changes requested):
You have two files that shouldn't be in this commit - `1` and `db/development.zip`. Please remove both.

app/models/fact_value.rb
tbrisker: There are two options I think of - ```Host::Base.friendly.find(host_id)``` that relies on friendly_id gem, or ```Host::Base.where(id: host_id).or(Host::Base.where(name: host_id))``` that uses general activerecord methods.

app/controllers/api/v2/fact_values_controller.rb
```diff
@@ -15,7 +15,8 @@ def index
           no_timestamp_facts.
           search_for(*search_options).paginate(paginate_options).
           preload(:fact_name, :host)
-        @fact_values = FactValue.build_facts_hash(values.all)
+        host_id = params[:host_id] || nil
```
tbrisker: You don't need to specify this here, if there is no such param then `params[:host_id]` will already return nil.

app/models/fact_value.rb
```diff
@@ -81,13 +81,19 @@ def self.count_each(fact, options = {})
...
+    else
+      hosts =  Host.where(:id =>facts.map(&:host_id).uniq)
```
tbrisker: this will load all facts to memory just to calculate the ids, you can do this in the db using `facts.distinct.pluck(:host_id)`.

tbrisker (changes requested):

shiramax:
@tbrisker , the line you suggested: 
 hosts =Host.where(:id => facts.distinct.pluck(:host_id))
caused the tests to fail, i can't remmember the error, but is it possible that this line can't run on all DB's?

shiramax:
@tbrisker, found the failures :  
http://ci.theforeman.org/job/test_develop_pr_core/17338/

xprazak2:
Seems like all tests are green now. @ShimShtein, @tbrisker, @dLobatog, could you re-review pls?

app/models/fact_value.rb
```diff
@@ -81,13 +81,10 @@ def self.count_each(fact, options = {})
...
-    hosts = Host.where(:id => facts.group_by(&:host_id).keys).all
-
-    hash = {}
-    facts.each do |fact|
-      hash[hosts.detect {|h| h.id == fact.host_id}.to_s] ||= {}
-      hash[hosts.detect {|h| h.id == fact.host_id}.to_s].update({fact.name.to_s => fact.value})
+  def self.build_facts_hash(values, host_id = nil)
+    hash  = values.group_by(&:host_name).transform_values!{|val| val.map{|v| [v.fact_name_name, v.value]}.to_h}
+    if host_id
+      hash  = hash.merge! hash.values[0]
```
ShimShtein: This line causes `no implicit conversion of nil into Hash` error, if there is no host present. Can you please add a test for this case too?

ShimShtein (changes requested):

shiramax:
@ShimShtein Done, please review again :)

ShimShtein:
[test foreman]

ShimShtein (dismissed):

ShimShtein:
[test katello]

dLobatog (dismissed):
Thanks @shiramax @ShimShtein @tbrisker

app/models/fact_value.rb
```diff
@@ -81,13 +81,10 @@ def self.count_each(fact, options = {})
...
 
-  def self.build_facts_hash(facts)
-    hosts = Host.where(:id => facts.group_by(&:host_id).keys).all
-
-    hash = {}
-    facts.each do |fact|
-      hash[hosts.detect {|h| h.id == fact.host_id}.to_s] ||= {}
-      hash[hosts.detect {|h| h.id == fact.host_id}.to_s].update({fact.name.to_s => fact.value})
+  def self.build_facts_hash(values, host_id = nil)
+    hash  = values.group_by(&:host_name).transform_values!{|val| val.map{|v| [v.fact_name_name, v.value]}.to_h}
```
tbrisker: nitpick - spacing here and in the tests seems to be a bit off

app/models/fact_value.rb
```diff
@@ -81,13 +81,10 @@ def self.count_each(fact, options = {})
...
+    if host_id
+      hash.merge! hash.values[0] if hash.any?
```
tbrisker: you can combine the two conditions here

tbrisker (changes requested):

app/models/fact_value.rb
shiramax: @tbrisker, can you please point the problematic line?

shiramax:
[ test katello ]

shiramax:
[test katello]

shiramax:
@dLobatog @ShimShtein @tbrisker , please review :)

tbrisker (approved):
lgtm

ShimShtein:
Merged.
Thanks @shiramax!

PR #5004, opened 2017-11-14
fixes #21657 - remove facets when host is removed

timogoebel:
Without this, you have to manually delete the facets from the database. There is no automatic cleanup.

cc: @ShimShtein

### DIFF
--- a/app/services/facets/entry.rb
+++ b/app/services/facets/entry.rb
@@ -4,7 +4,7 @@ module Facets
       :api_single_view, :api_list_view,
       :api_param_group_description, :api_param_group, :api_controller,
       :tabs,
-      :compatibility_properties
+      :compatibility_properties, :dependent
 
     def initialize(facet_model, facet_name)
       @compatibility_properties = []
@@ -78,6 +78,13 @@ module Facets
     #  host.foo # => 'bar'
     def template_compatibility_properties(*property_symbols)
       @compatibility_properties = property_symbols
+    end
+
+    # Specify what should happen with the facet when the host object is deleted.
+    # [+:destroy+] The facet object will be deleted by calling the destroy record.
+    # [+:delete+] The facet will be deleted directly from the database without calling their destroy method.
+    def set_dependent_action(value)
+      @dependent = value
     end
 
     def load_api_controller
--- a/test/unit/facet_test.rb
+++ b/test/unit/facet_test.rb
@@ -134,6 +134,36 @@ class FacetTest < ActiveSupport::TestCase
       saved_host.update({'test_facet_attributes' => { 'my_attribute' => "val"}})
       assert_not_nil saved_host.test_facet
     end
+
+    test 'facet is not removed when associated host is deleted' do
+      as_admin do
+        saved_host = FactoryBot.create(:host)
+        facet = saved_host.build_test_facet
+        saved_host.save!
+        assert facet.id
+        saved_host.destroy!
+        assert TestFacet.find_by(id: facet.id)
+      end
+    end
+  end
+
+  context 'managed host facet dependent destroy behavior' do
+    setup do
+      Facets.register TestFacet do
+        set_dependent_action :destroy
+      end
+    end
+
+    test 'facet is removed when associated host is deleted' do
+      as_admin do
+        saved_host = FactoryBot.create(:host)
+        facet = saved_host.build_test_facet
+        saved_host.save!
+        assert facet.id
+        saved_host.destroy!
+        refute TestFacet.find_by(id: facet.id)
+      end
+    end
   end
 
   context 'inside db:migrate task' do
--- a/app/models/concerns/facets/managed_host_extensions.rb
+++ b/app/models/concerns/facets/managed_host_extensions.rb
@@ -35,7 +35,7 @@ module Facets
       #    host.foo # => will call Host.my_facet.foo
       def register_facet_relation(klass, facet_config)
         klass.class_eval do
-          has_one facet_config.name, :class_name => facet_config.model.name, :foreign_key => :host_id, :inverse_of => :host
+          has_one facet_config.name, :class_name => facet_config.model.name, :foreign_key => :host_id, :inverse_of => :host, :dependent => facet_config.dependent
           accepts_nested_attributes_for facet_config.name, :update_only => true, :reject_if => :all_blank
           if Foreman.in_rake?("db:migrate")
             # To prevent running into issues in old migrations when new facet is defined but not migrated yet.
--- a/test/unit/facet_configuration_test.rb
+++ b/test/unit/facet_configuration_test.rb
@@ -12,10 +12,10 @@ class FacetConfigurationTest < ActiveSupport::TestCase
   module TestHelper
   end
 
-  teardown do
-    Host::Managed.cloned_parameters[:include].delete(:test_model)
-    Host::Managed.cloned_parameters[:include].delete(:test_facet)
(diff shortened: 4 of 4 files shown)


timogoebel:
I see a lot of test failures like
```
PG::UndefinedTable: ERROR:  relation "test_classes" does not exist...
```

I believe, the tests aren't properly isolated. Will take a look.

ShimShtein:
@jlsherrill , @thomasmckay Any objections? Could it cause issues to Katello?

jlsherrill:
In katello we preferred this route, as deleting facets without deleting backend (candlepin or pulp) entities can lead to issues.  The idea being that deletion is directed through our foreman-task orchestration which properly removes those objects from the backend services before removing the facets and finally the host. 

One question though, how are you able to delete the host without deleting the facet first?  Are you missing a foreign key constraint on the host_id column on your facet?  It should prevent that.  

This change won't completely break katello (as i said the user ends up going through dynflow tasks when deleting a host), but may make it harder to detect if this breaks in the future.  Some sort of more official facet delete api may be nice, as i could see it useful for other things such as a puppet facet (where you may want to cleanup the puppet ca).

sean797:
Do you think we should clear up any ~~fact~~ facets belonging to hosts that no long exist?

ShimShtein:
> Do you think we should clear up any fact belonging to hosts that no long exist?

We already [do this](https://github.com/theforeman/foreman/blob/develop/app/models/host/base.rb#L17)

sean797:
Sorry I mean facets. This will stop it happening anymore, but should there be a migration to clear any stale facets already in the database?

ShimShtein:
@timogoebel, I think we should make it configurable - we can add another configuration option to `services/facets/entry.rb` and add `:dependent => destroy` accordingly.

timogoebel:
@ShimShtein: Sounds like a good idea. Do you (as the "facet guy") by any chance want to take this over? :-)

timogoebel:
@ShimShtein: Added a configuration parameter & tests.

ShimShtein:
[test foreman]

timogoebel:
Rebased to (hopefully) fix the tests.

boaz0 (approved):
LGTM 👍

ShimShtein:
[test foreman]

timogoebel:
Rebased.

ShimShtein:
@timogoebel Tests are still red

timogoebel:
*sights* pushed another attempt. I have a feeling, the configuration tests weren't isolated enough. I stubbed `Facets::ManagedHostExtensions.register_facet_relation` now. Hope, that helps.

iNecas:
It probably didn't

timogoebel:
Another attempt to fix the tests by reloading the host class following the suggestion by @ShimShtein.

mmoll:
please rebase :)

timogoebel:
Rebased.

timogoebel:
[test foreman]

ShimShtein:
[test foreman]

boaz0 (approved):
LGTM 💯

tbrisker:
[test foreman]

ShimShtein:
[test]

timogoebel:
I actually have a feeling, the test failure could be related to the reloading of the host object. But I haven't had time to investigate.

timogoebel:
I changed my opinion about the test failures being related. Rebased. Hopefully, tests are 🍏 now.

mmoll:
@timogoebel please rebase :)

timogoebel:
Rebased.

app/services/facets/entry.rb
```diff
@@ -80,6 +80,14 @@ def template_compatibility_properties(*property_symbols)
       @compatibility_properties = property_symbols
     end
 
+    # Specify what should happen with the facet when the host object is deleted.
+    # [+:destroy+] The facet object will be deleted by calling the destroy record.
+    # [+:delete+] The facet will be deleted directly from the database without calling their destroy method.
+    def dependent(value = nil)
+      @dependent = value if value.present?
```
tbrisker: This method looks like a combination of getter and setter, why not just use `attr_accessor`? Or name it in a way that has clear meaning, like `set_dependent_action`?

tbrisker (changes requested):
I managed to reliably reproduce the test failures. They seem to be caused by the reloading of `host/managed.rb` in the facet test - commenting out those lines, or even just the one in `teardown` makes audit tests pass. Are you sure they are needed?
To reproduce locally: 
`SEED=30219 ruby -I.:test -e "ARGV.each{|f| require f}" test/unit/facet_test.rb test/models/concerns/audit_extensions_test.rb`

timogoebel:
@tbrisker: Thanks for taking a look. :-)

> Are you sure they are needed?

I just pushed a version with these lines commented out. Let's see what that breaks.

ShimShtein:
@tbrisker facets pollute the `Host::Managed` class with new relations. Those relations cannot be undone, so if you are adding a facet for test only, you end up with unwanted relation methods for all other tests. This is what started this whole mess.

timogoebel:
Tests passed without the reload code. I just pushed a version that renames the DSL to use `set_dependent_action`. I hope, we're good to go now. *sights*

timogoebel:
@tbrisker, @ShimShtein: Tests are green now. 🌲

tbrisker (dismissed):
Works for me, any more comments @ShimShtein ?

test/unit/facet_test.rb
```diff
@@ -14,12 +14,6 @@ class FacetTest < ActiveSupport::TestCase
     Facets.stubs(:configuration).returns(@config)
   end
 
-  teardown do
```
ShimShtein: Without those lines, cloning `Host::Managed` will try to clone `test_facet`. I don't think it's desirable.

After this test Host::Managed will contain all properties related to `test_facet` and `namespaced_facet`

timogoebel:
I re-added the teardown lines for the tests.

tbrisker:
@timogoebel I think you forgot to push, and I think it will still fail (it did when i was testing on the older version). Perhaps we need to find some other way to remove the association that doesn't include reloading the model.

timogoebel:
> Perhaps we need to find some other way to remove the association that doesn't include reloading the model.

To be honest, I'm out of ideas. I could try to not add the association in the first place by stubbing it, but otherwise: I have no idea how to proceed here. Suggestions are very welcome.

I think I did push my changes.
@tbrisker: What is missing?

ShimShtein:
I am not sure it's a good one, but maybe we can use an inherited throwaway class here?
Something along the lines:

``` ruby
class FacetTest < ActiveSupport::TestCase
  class TestHost < Host::Managed
  end
end
```
I am not sure it will work, and what ActiveRecord trickery we will need to invoke, but at least it will not pollute the original `Host::Managed` class.

leewaa:
Nothing else is missing here ? Looks ok to me and tests are green.

No other objections to this one ?

Just needs a rebase

timogoebel:
Rebased.

Can we get this in? I don't think rewriting the way we test facets is actually within the scope of this issue. We should probably do it at some time, but we can do that later I believe.

mmoll (approved):

tbrisker (approved):
LGTM, Thanks @timogoebel and everyone involved in reviewing :)

timogoebel:
Thanks, guys.

I updated the documentation to include this: http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#section-26

PR #5005, opened 2017-11-14
fixes #21658 - add pagelet extension points to hostgroup form

timogoebel:


tbrisker (commented):
Code looks good, I'd maybe change the commit message though to from "facet extension points" to "pagelets" since this isn't specific to facets though.
@ShimShtein any other comments?

timogoebel:
Fixed the commit message.

test/controllers/hostgroups_controller_test.rb
```diff
@@ -264,4 +267,19 @@ def setup_user(operation, type = 'hostgroups')
       end
     end
   end
+
+  context 'with pagelets' do
+    setup do
+      @controller.prepend_view_path File.expand_path('../../static_fixtures', __FILE__)
+      Pagelets::Manager.add_pagelet('hostgroups/edit', :main_tabs,
```
ShimShtein: Since the usage would be to both edit and new screens, I think we should test the more realistic behavior - registering to `hostgroup/_form` and consuming from both new and edit screens.
[Here](https://github.com/theforeman/foreman/blob/develop/config/initializers/puppet.rb#L2) is an example of such use for hosts form.

ShimShtein (changes requested):
Overall looks great, could you please create/find an issue to move puppet fields outside the hostgroup too? 

@timogoebel, one small caveat - if you don't refactor current tabs as consumers, you could not control the order of tabs. Your added tabs would always be between parameters and taxonomies.

timogoebel:
> one small caveat - if you don't refactor current tabs as consumers, you could not control the order of tabs. Your added tabs would always be between parameters and taxonomies.

@ShimShtein: I know. I wanted this to match the behavior of the host form for now.

> could you please create/find an issue to move puppet fields outside the hostgroup too

Sure. I stole a ticket from you for this 😈 
http://projects.theforeman.org/issues/18350

Changed the tests to your suggestions.

tbrisker (approved):
Looks good, thanks @timogoebel !

ares:
setting version 1.17.0 in redmine

PR #5006, opened 2017-11-15
Fixes #21669 - Use $major on CentOS media, $version deprecated

dLobatog:
As we can see, CentOS fixed their repos so that
http://mirror.centos.org/centos/7 returns the latest version we need for
vmlinuz, etc..
For a long time they only offered these files under specific version
dirs, but now they have deprecated this and we should just refer to
$major, not $major.$minor
See the notice at http://mirror.centos.org/centos/7.3.1611/readme

I'm not 100% sure we should change in a migration "CentOS mirror", so let me know if you think that'd be OK to fix previous installations like that.

### DIFF
--- a/db/migrate/20171121082256_update_centos_installation_media.rb
+++ b/db/migrate/20171121082256_update_centos_installation_media.rb
@@ -1,0 +1,9 @@
+class UpdateCentosInstallationMedia < ActiveRecord::Migration
+  def change
+    Medium.unscoped.where(
+      :name => 'CentOS mirror',
+      :os_family => 'Redhat',
+      :path => 'http://mirror.centos.org/centos/$version/os/$arch'
+    ).update_all(:path => 'http://mirror.centos.org/centos/$major/os/$arch')
+  end
+end
--- a/db/seeds.d/100-installation_media.rb
+++ b/db/seeds.d/100-installation_media.rb
@@ -3,7 +3,7 @@ os_suse = Operatingsystem.unscoped.where(:type => "Suse") || Operatingsystem.uns
 # Installation media: default mirrors
 Medium.without_auditing do
   [
-    { :name => "CentOS mirror",        :os_family => "Redhat",  :path => "http://mirror.centos.org/centos/$version/os/$arch" },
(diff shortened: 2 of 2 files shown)


tbrisker (commented):
Makes sense, I think we should migrate as well since users' provisioning will be broken if they didn't update it, at least if the original value hasn't been updated.

dLobatog (commented):
Thanks for the review @tbrisker , I've included the migration now

db/migrate/20171121082256_update_centos_installation_media.rb
```diff
@@ -0,0 +1,13 @@
+class UpdateCentosInstallationMedia < ActiveRecord::Migration
+  def change
+    centos_media = Medium.unscoped.where(
+      :name => 'CentOS mirror',
+      :os_family => 'Redhat',
+      :path => 'http://mirror.centos.org/centos/$version/os/$arch'
+    )
```
tbrisker: you can use `.update_all(:path, 'http://mirror.centos.org/centos/$major/os/$arch')` to update them all in one shot without iteration.

ares: unless you want to add a condition to only update media with unchanged path, that might be useful for users who changed it

tbrisker: we already select only those that have the old path in the where clause so if the user changed it it wouldn't update

ares: right

dLobatog (commented):
@tbrisker Updated to use `update_all` instead of iterating

tbrisker (approved):
Ack, will merge once test pass

tbrisker:
Thanks @dLobatog !

tbrisker:
@dlobatog I've set release 1.17 for now, but I wonder if we should cherry-pick to 1.16? If I understand correctly without this CentOS provisioning will be broken?

lzap:
But this is only relevant for CentOS 7, there are still supported older versions of CentOS.

lzap:
Oh I can see it there is a symlink. Okay.

PR #5008, opened 2017-11-15
Fixes #21675 - OpenStack create host tab fails to load

dLobatog:
Currently the tab is not loading as there are a few methods that are not
being overridden properly by the FogExtension.

After fixing that and adding the metadata key the tab loads.

Before:
screenshot from 2017-11-15 13-35-34

After:

screenshot from 2017-11-15 14-13-29

----------------------------------

Note there is some problem in OpenStack now that you cannot set the availability zone to "none" - I can fix that on another PR.

screenshot from 2017-11-15 14-17-01

### DIFF
--- a/app/models/compute_resources/foreman/model/openstack.rb
+++ b/app/models/compute_resources/foreman/model/openstack.rb
@@ -228,7 +228,7 @@ module Foreman::Model
     end
 
     def vm_instance_defaults
-      super.merge(:key_name => key_pair.name)
+      super.merge(:key_name => key_pair.name, :metadata => {})
     end
 
     def assign_floating_ip(address, vm)
--- a/app/models/concerns/fog_extensions/openstack/server.rb
+++ b/app/models/concerns/fog_extensions/openstack/server.rb
@@ -1,13 +1,9 @@
 module FogExtensions
   module Openstack
     module Server
-      def self.prepended(base)
-        class << base
-          attr_reader :nics
-          attr_accessor :boot_from_volume, :size_gb, :scheduler_hint_filter
-          attr_writer :security_group, :network # floating IP
-        end
-      end
+      attr_reader :nics
+      attr_writer :security_group, :network # floating IP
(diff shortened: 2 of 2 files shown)


app/models/concerns/fog_extensions/openstack/server.rb
```diff
@@ -48,11 +44,11 @@ def security_groups
       end
 
       def boot_from_volume
-        attr[:boot_from_volume]
+        attributes[:boot_from_volume]
```
dLobatog, line 47: This change comes from fog (attr to attributes)

app/models/compute_resources/foreman/model/openstack.rb
```diff
@@ -228,7 +228,7 @@ def volume_client
     end
 
     def vm_instance_defaults
-      super.merge(:key_name => key_pair.name)
+      super.merge(:key_name => key_pair.name, :metadata => {})
```
dLobatog, line 231: Just a new key required to create the vm in fog..

app/models/concerns/fog_extensions/openstack/server.rb
```diff
@@ -1,13 +1,9 @@
...
   module Openstack
     module Server
-      def self.prepended(base)
-        class << base
-          attr_reader :nics
-          attr_accessor :boot_from_volume, :size_gb, :scheduler_hint_filter
-          attr_writer :security_group, :network # floating IP
-        end
-      end
+      attr_reader :nics
```
dLobatog, line 4: I think these changes are needed now just because we're OK with Ruby 2.0+, on 2.4 (at least) they are required.

dLobatog (commented):
Integration test failures are unrelated

lzap (approved):
Tested with TryStack (OpenStack RDO Liberty on x86/RHEL)

ares:
Thanks for merging @lzap, please don't forget to set the release in redmine after merge. I set it to 1.17.0

PR #5009, opened 2017-11-15
fixes #21676 - select vmware network for subnet

timogoebel:
The [Redmine ticket](http://projects.theforeman.org/issues/21676) contains the business case for this.
This PR uses a fancy machine learning algorithm 🔥 (`grep`, patent pending) to preselect the VMWare network that matches the subnet.

### DIFF
--- a/app/assets/javascripts/host_edit.js
+++ b/app/assets/javascripts/host_edit.js
@@ -655,7 +655,11 @@ function interface_domain_selected(element) {
           select = subnet6_options
         }
         if (select) {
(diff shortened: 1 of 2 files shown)


dLobatog (approved):
Basic use case described in the Redmine ticket works well, the 2nd field can be changed by the user if this gets it wrong at some point. Thanks @timogoebel

ThoTischner:
This does not work as expected.
We tried to rename the foreman subnet, to match exactly the vmware network.
VLAN ID also matches, but it doesn`t work either.

mmoll:
@disappear89 please create an issue in Redmine: https://projects.theforeman.org/projects/foreman/issues/new

PR #5010, opened 2017-11-15
Fixes #21654 - create registry for importers

rabajaj0509:
The facts impoter must have been initialized when the server restarts
as it is a part of the puppet.rb(https://github.com/theforeman/foreman/blob/0b6ec11eea13927de6ae3697d704e70d6206d15d/config/initializers/puppet.rb#L15) but it does not therefore i have added it manually.

This might not be the accurate way but hosts are getting discovered with this change.

### DIFF
--- a/app/models/host/base.rb
+++ b/app/models/host/base.rb
@@ -142,10 +142,10 @@ module Host
       facts[:domain] = facts[:domain].downcase if facts[:domain].present?
 
       type = facts.delete(:_type)
-      importer = FactImporter.importer_for(type).new(self, facts)
+      facts_importer = Foreman::Plugin.fact_importer_registry.get(type).new(self, facts)
       telemetry_observe_histogram(:importer_facts_import_duration, facts.size, type: type)
       telemetry_duration_histogram(:importer_facts_import_duration, 1000, type: type) do
-        importer.import!
+        facts_importer.import!
       end
 
       save(:validate => false)
--- a/app/registries/foreman/plugin.rb
+++ b/app/registries/foreman/plugin.rb
@@ -21,6 +21,7 @@ require_dependency 'foreman/plugin/rbac_support'
 require_dependency 'foreman/plugin/report_scanner_registry'
 require_dependency 'foreman/plugin/report_origin_registry'
 require_dependency 'foreman/plugin/medium_providers_registry'
+require_dependency 'foreman/plugin/fact_importer_registry'
 
 module Foreman #:nodoc:
   class PluginNotFound < Foreman::Exception; end
@@ -39,6 +40,7 @@ module Foreman #:nodoc:
   class Plugin
     @registered_plugins = {}
     @tests_to_skip = {}
+    @fact_importer_registry = Plugin::FactImporterRegistry.new
     @report_scanner_registry = Plugin::ReportScannerRegistry.new
     @report_origin_registry = Plugin::ReportOriginRegistry.new
     @medium_providers = Plugin::MediumProvidersRegistry.new
@@ -48,7 +50,7 @@ module Foreman #:nodoc:
       attr_reader   :registered_plugins
       attr_accessor :tests_to_skip, :report_scanner_registry,
         :report_origin_registry, :medium_providers,
-        :graphql_types_registry
+        :graphql_types_registry, :fact_importer_registry
       private :new
 
       def def_field(*names)
@@ -151,6 +153,10 @@ module Foreman #:nodoc:
       @renderer_variable_loaders = []
       @ping_extension = nil
       @status_extension = nil
+    end
+
+    def fact_importer_registry
+      self.class.fact_importer_registry
     end
 
     def report_scanner_registry
--- a/app/services/fact_importer.rb
+++ b/app/services/fact_importer.rb
@@ -5,21 +5,23 @@ class FactImporter
   attr_reader :counters
 
   def self.importer_for(type)
-    importers[type.to_s]
+    Foreman::Deprecation.deprecation_warning('2.2', 'Use FactImporterRegistry#get or Foreman::Plugin.importer_registry.get methods instead of FactImporter.importer_for')
+    Foreman::Plugin.fact_importer_registry.get(type)
   end
 
   def self.importers
-    @importers ||= {}.with_indifferent_access
+    Foreman::Deprecation.deprecation_warning('2.2', 'Use FactImporterRegistry#importers or Foreman::Plugin.importer_registry.importers method instead of FactImporter.importers')
+    Foreman::Plugin.fact_importer_registry.importers
   end
 
   def self.register_fact_importer(key, klass, default = false)
-    importers.default = klass if default
-
-    importers[key.to_sym] = klass
+    Foreman::Deprecation.deprecation_warning('2.2', 'Use FactImporterRegistry#register or Foreman::Plugin.importer_registry.register method instead of FactImporter.register_fact_importer')
+    Foreman::Plugin.fact_importer_registry.register(key, klass, default)
   end
 
   def self.fact_features
-    importers.map { |_type, importer| importer.authorized_smart_proxy_features }.compact.flatten.uniq
+    Foreman::Deprecation.deprecation_warning('2.2', 'Use FactImporterRegistry#fact_features or Foreman::Plugin.importer_registry.fact_features method instead of FactImporter.fact_features')
+    Foreman::Plugin.fact_importer_registry.fact_features
   end
 
   def self.support_background
--- a/config/initializers/puppet.rb
+++ b/config/initializers/puppet.rb
@@ -12,7 +12,7 @@ Pagelets::Manager.with_key "hosts/_form" do |mgr|
     :priority => 100
 end
 
-FactImporter.register_fact_importer :puppet, PuppetFactImporter, true
+Foreman::Plugin.fact_importer_registry.register(:puppet, PuppetFactImporter, true)
 FactParser.register_fact_parser :puppet, PuppetFactParser, true
 
 # The module should be included after the class is constructed,
--- a/test/fact_importer_test_helper.rb
+++ b/test/fact_importer_test_helper.rb
@@ -7,7 +7,7 @@ module FactImporterIsolation
   end
 
   def allow_transactions_for_any_importer
-    FactImporter.importers.values.each do |importer|
+    Foreman::Plugin.fact_importer_registry.importers.values.map(&:constantize).each do |importer|
       allow_transactions_for(importer.any_instance)
     end
   end
--- a/test/unit/plugin/importer_registry_test.rb
+++ b/test/unit/plugin/importer_registry_test.rb
@@ -1,0 +1,38 @@
+require 'test_helper'
+
+class FactImporterRegistry < ActiveSupport::TestCase
+  class TestClass
+  end
+
+  def test_register
+    fact_importer = Foreman::Plugin::FactImporterRegistry.new
+    fact_importer.register(:test_class, TestClass, false)
+    assert_equal fact_importer.importers[:test_class], "FactImporterRegistry::TestClass"
+  end
+
+  def test_register_default
+    fact_importer = Foreman::Plugin::FactImporterRegistry.new
+    fact_importer.register(:test_class, TestClass, true)
+    assert_equal fact_importer.importers[:test_class], "FactImporterRegistry::TestClass"
+    assert_equal fact_importer.importers[:random_key], "FactImporterRegistry::TestClass"
+  end
+
+  def test_get
+    fact_importer = Foreman::Plugin::FactImporterRegistry.new
+    fact_importer.register(:test_class, TestClass, false)
+    assert_equal fact_importer.get(:test_class), TestClass
+  end
+
+  def test_get_default
+    fact_importer = Foreman::Plugin::FactImporterRegistry.new
+    fact_importer.register(:test_class, TestClass, true)
+    assert_equal fact_importer.get(:random_key), TestClass
+  end
+
+  def test_fact_features
+    fact_importer = Foreman::Plugin::FactImporterRegistry.new
+    fact_importer.register(:test_class, TestClass, true)
+    FactImporterRegistry::TestClass.expects(:authorized_smart_proxy_features).returns([:test_feature])
+    assert_equal :test_feature, fact_importer.fact_features.first
+  end
+end
--- a/app/controllers/api/v2/hosts_controller.rb
+++ b/app/controllers/api/v2/hosts_controller.rb
@@ -15,7 +15,7 @@ module Api
       check_permissions_for %w{power boot}
       before_action :process_parameter_attributes, :only => %w{update}
 
-      add_smart_proxy_filters :facts, :features => Proc.new { FactImporter.fact_features }
+      add_smart_proxy_filters :facts, :features => Proc.new { Foreman::Plugin.fact_importer_registry.fact_features }
 
       add_scope_for(:index) do |base_scope|
         base_scope.preload([:host_statuses, :compute_resource, :hostgroup, :operatingsystem,
--- a/app/registries/foreman/plugin/fact_importer_registry.rb
+++ b/app/registries/foreman/plugin/fact_importer_registry.rb
@@ -1,0 +1,24 @@
+module Foreman
+  class Plugin
+    class FactImporterRegistry
+      attr_accessor :importers
+
+      def initialize
+        @importers = {}.with_indifferent_access
+      end
+
+      def register(key, klass, default = false)
+        @importers.default = klass.to_s if default
+        @importers[key.to_sym] = klass.to_s
+      end
+
+      def get(key)
+        @importers[key].constantize
+      end
+
+      def fact_features
+        @importers.values.map { |importer| importer.constantize.authorized_smart_proxy_features }.compact.flatten.uniq
+      end
+    end
+  end
+end
--- a/test/unit/fact_importer_test.rb
+++ b/test/unit/fact_importer_test.rb
@@ -13,13 +13,14 @@ class FactImporterTest < ActiveSupport::TestCase
   end
 
   let(:host) { FactoryBot.create(:host) }
+  let(:fact_importer_registry) { Foreman::Plugin.fact_importer_registry }
 
   test "default importers" do
-    assert_includes FactImporter.importers.keys, 'puppet'
-    assert_equal PuppetFactImporter, FactImporter.importer_for(:puppet)
-    assert_equal PuppetFactImporter, FactImporter.importer_for('puppet')
-    assert_equal PuppetFactImporter, FactImporter.importer_for(:whatever)
-    assert_equal PuppetFactImporter, FactImporter.importer_for('whatever')
+    assert_includes fact_importer_registry.importers.keys, 'puppet'
+    assert_equal PuppetFactImporter, fact_importer_registry.get(:puppet)
+    assert_equal PuppetFactImporter, fact_importer_registry.get('puppet')
+    assert_equal PuppetFactImporter, fact_importer_registry.get(:whatever)
+    assert_equal PuppetFactImporter, fact_importer_registry.get('whatever')
   end
 
   test 'importer API defines background processing support' do
@@ -28,15 +29,15 @@ class FactImporterTest < ActiveSupport::TestCase
 
   context 'when using a custom importer' do
     setup do
-      FactImporter.register_fact_importer :custom_importer, CustomImporter
+      fact_importer_registry.register :custom_importer, CustomImporter
     end
 
     test ".register_custom_importer" do
-      assert_equal CustomImporter, FactImporter.importer_for(:custom_importer)
+      assert_equal CustomImporter, fact_importer_registry.get(:custom_importer)
     end
 
     test 'importers without authorized_smart_proxy_features return empty set of features' do
-      assert_equal [], FactImporter.importer_for(:custom_importer).authorized_smart_proxy_features
+      assert_equal [], fact_importer_registry.get(:custom_importer).authorized_smart_proxy_features
     end
 
     context 'importing facts' do


rabajaj0509:
@ShimShtein can you please review this pr ?

rabajaj0509:
This is the reason for adding it manually:
screenshot from 2017-11-15 21-30-40

Above, we can see only `{"rhsm"=>Katello::RhsmFactImporter}` but in the rake console we can see : 
screenshot from 2017-11-15 21-33-29

ShimShtein (changes requested):
It looks like `FactImporter` class got reloaded at some point. I don't have enough context how did you get there. Could you please share a repro for this?

Basically it beats the purpose of preparing puppet to be extracted out of the core.

rabajaj0509:
@ShimShtein Steps to reproduce: 
1.try to discover a host with discovery using libvert.
2.See the server response, it will display a message - undefined method `new` for nil:NIlClass

This happens because the ` FactImporter` is not registered, as soon as you register it(as done in this pr) it gets fixed and hosts get discovered.

iNecas:
I believe this issue is describing the problem http://projects.theforeman.org/issues/21573, the solution should be moving the importers variable somewhere in registers

rabajaj0509:
@iNecas this(http://projects.theforeman.org/issues/21573) is exactly the problem we are facing.

ShimShtein:
@rahulbajaj0509, After testing it, it looks like a reloading problem. I have tried discovering a host, then adding a space to `Host::Base` and them discover a host again. It succeeded on the first discovery and failed with the mentioned error on second.
I have found a good resource that describes the code responsible for the issue on [our wiki](http://projects.theforeman.org/projects/foreman/wiki/Registry_pattern). IMHO we need to change importer registry according to that article.

rabajaj0509:
@ShimShtein are you suggesting we remove the `fact_importer` file from the `app/services` and instead write a new file in the `app/registries` as suggested in the wiki? Am i getting it right here?

ShimShtein:
I suggest splitting fact importer into two - the [registry part](https://github.com/theforeman/foreman/blob/1f3470f78ffcb414ffb979736b7533dbaa54fbf8/app/services/fact_importer.rb#L5-L31) that should go into `app/registries` and the importer class itself, that should stay in `services`.

iNecas:
👍 on Shim's suggestion: just keep in mind to store the class as string and doing constantize when needing the class, otherwise weird things happen on reload

lzap:
If this affects discovery, please prepare patch for it. Thanks!

iNecas:
@rahulbajaj0509 are you working on the proper fix for this or should we put it on our iteration?

rabajaj0509:
@iNecas hey, I am working on this. I am new to the registries concept, if you could help me out figuring the right fix for this then it would be great :)

iNecas:
@rahulbajaj0509 just ask whatever you would like to know :) I suggest getting inspiration from the registers we already have around https://github.com/theforeman/foreman/tree/1f3470f78ffcb414ffb979736b7533dbaa54fbf8/app/registries

rabajaj0509:
@iNecas I guess this needs to be fixed asap and I am taking a little too long for this. Mind fixing it?

iNecas:
I have quite a few high prio things on my plate at the moment. Perhaps @ShimShtein could have a look at the proper fix?

iNecas:
@rahulbajaj0509 I remember we were talking about you still looking into this. Any updates since the last check?

rabajaj0509:
@iNecas sorry had completely forgotten about this one, have done the initial work but I will get it done, first thing on my plate now.

bastilian:
@rahulbajaj0509 you could also delegate the registration for importers to the registry from `Foreman::Plugin` like here: https://github.com/bastilian/foreman/blob/8bd79a1af650e904fcf8c97b22faef39e3841fd9/app/registries/foreman/plugin.rb#L112-L115

Plugins then can register their scanner in the plugin registration, like for the report scanners:
```ruby
    register_report_scanner ForemanAnsible::AnsibleReportScanner
```

You might need to alias the register method or rename it, to not collide with other methods.

app/registries/foreman/plugin/importer_registry.rb
```diff
@@ -0,0 +1,16 @@
+module Foreman
+  class Plugin
+    class ImporterRegistry
```
ShimShtein: Nice solution! It is very clean, I like it. One disadvantage though: it's really hard to isolate in tests. Imagine a test that does:

``` ruby

test 'sets default provider' do
  Foreman::Plugin::ImporterRegistry.register :foo, 'FooClass', true

  # some assertions
end
```

Every test that is executed after this one, and relies on default fact importer will fail and it would be really hard to trace back the error.

I prefer @bastilian's [solution](https://github.com/theforeman/foreman/pull/5221/files#diff-e912f1d36ab9771d3896cf47c09a39f6R42). In that solution, I can easily stub registry accessor method with a new instance of the registry and modify it as much as I like. After the test is done, the original accessor will be set and everything will work as usual.

app/services/fact_importer.rb
```diff
@@ -2,20 +2,6 @@ class FactImporter
...
-  end
-
-  def self.register_fact_importer(key, klass, default = false)
-    importers.default = klass if default
-
-    importers[key.to_sym] = klass
-  end
-
   def self.fact_features
     importers.map { |_type, importer| importer.authorized_smart_proxy_features }.compact.flatten.uniq
```
ShimShtein: This method will not work, it should be either part of the registry, or reference the registry correctly.

@stbenjam, I am not familiar with Smart Proxy Authorization, what is your opinion? Is it a part of the registry, or just a service method that should stay here?

config/initializers/puppet.rb
```diff
@@ -12,7 +12,7 @@
     :priority => 100
 end
 
-FactImporter.register_fact_importer :puppet, PuppetFactImporter, true
+Foreman::Plugin::ImporterRegistry.register :puppet, 'PuppetFactImporter', true
```
ShimShtein: Can we leave the ability to register by actual constant (as opposed to string). It will prevent typos - we are relying on ruby that the constant actually exists this way.

I don't want to change the inner implementation (that stores strings), just the outer interface. We can convert the constant to string inside the `register` method.

ShimShtein (changes requested):
All in all I like it very much, a couple of nitpicks though.

app/services/fact_importer.rb
stbenjam: Probably part of the registry.  The hosts controller needs to know the smart proxy features for all of the registered importers.

config/initializers/puppet.rb
rabajaj0509: I think we should register by using a string, 2 reasons:
1) If we use the original constant then it will get loaded at the beginning, even when it is not called for.
2) In the registry, I have used `@importers[key].constantize` which will take care of constantizing the string.

app/registries/foreman/plugin.rb
```diff
@@ -37,10 +38,13 @@ class PluginRequirementError < Foreman::Exception; end
   class Plugin
     @registered_plugins = {}
     @tests_to_skip = {}
+    @importer_registry = Plugin::ImporterRegistry.new
     @report_scanner_registry = Plugin::ReportScannerRegistry.new
 
+
     class << self
       attr_reader   :registered_plugins
+      attr_accessor :tests_to_skip,:importer_registry
```
bastilian: This should rather be:
```ruby
      attr_accessor :tests_to_skip, :importer_registry
 				    :report_scanner_registry
```

config/initializers/puppet.rb
```diff
@@ -12,7 +12,7 @@
     :priority => 100
 end
 
-FactImporter.register_fact_importer :puppet, PuppetFactImporter, true
+Foreman::Plugin.importer_registry.register_importer :puppet, 'PuppetFactImporter', true
```
bastilian: You should be able to omit `importer_registry` and call `register_importer` on `Foreman::Plugin` directly, as the method will be delegated there.

bastilian (changes requested):

config/initializers/puppet.rb
rabajaj0509: I tried doing that but it doesnt allow me to do so. I get an error: `/home/vagrant/foreman/config/initializers/puppet.rb:15:in <top (required)>: undefined method 'register_importer' for Foreman::Plugin:Class (NoMethodError)`

bastilian: Ah! I'm confusing this with when registering from a plugin. The delegation is then available as it's within an Plugin instance. That you can use in the Katello PR.

app/registries/foreman/plugin.rb
```diff
@@ -37,10 +38,12 @@ class PluginRequirementError < Foreman::Exception; end
   class Plugin
     @registered_plugins = {}
     @tests_to_skip = {}
+    @importer_registry = Plugin::ImporterRegistry.new
     @report_scanner_registry = Plugin::ReportScannerRegistry.new
 
     class << self
       attr_reader   :registered_plugins
+      attr_accessor :tests_to_skip,:importer_registry, :report_scanner_registry
```
bastilian: The `attr_accessor` call below can go and there should be a space after the `,` :)

bastilian (changes requested):

config/initializers/puppet.rb
rabajaj0509: So i guess `Foreman::Plugin.importer_registry.register_importer` will stay as is here?

bastilian: :+1:

app/registries/foreman/plugin.rb
```diff
@@ -37,11 +38,13 @@ class PluginRequirementError < Foreman::Exception; end
   class Plugin
     @registered_plugins = {}
     @tests_to_skip = {}
+    @importer_registry = Plugin::ImporterRegistry.new
     @report_scanner_registry = Plugin::ReportScannerRegistry.new
 
     class << self
       attr_reader   :registered_plugins
       attr_accessor :tests_to_skip, :report_scanner_registry
+      attr_accessor :tests_to_skip, :importer_registry, :report_scanner_registry
```
bastilian: Line 46 can go, as line 47 includes the two attributes from 46 as well.

app/registries/foreman/plugin/importer_registry.rb
```diff
@@ -0,0 +1,23 @@
...
+      attr_accessor :importers 
+
+      def initialize
+        @importers = {
+          :puppet => 'PuppetFactImporter'
+        }.with_indifferent_access
```
bastilian: `with_indifferent_access` may not be necessary. It looks only the `register` method accesses the default via dot-notation to set a default, which could be changed and not require indifferent access.

app/registries/foreman/plugin/importer_registry.rb
```diff
@@ -0,0 +1,23 @@
...
+      end
+
+      def register(key, klass, default = false)
+        @importers.default = klass if default
```
bastilian: Changing to `@importers['default']` here would make indifferent_access not needed.

bastilian (changes requested):

bastilian:
@rahulbajaj0509 Just a few minor changes. :)

One more thing. We have, since https://github.com/theforeman/foreman/pull/5186 the origin attribute on reports, and report scanner. Maybe we can look into combining fact-parser, -importer and report-scanner, importer, etc. to Origins and have incoming data identified where it comes from on first point of entry and run them with all the appropriate and registered parser, scanner, importer for that origin only.

app/registries/foreman/plugin/importer_registry.rb
ShimShtein: @bastilian, it's a real [property](https://docs.ruby-lang.org/en/2.0.0/Hash.html#method-i-default) in `Hash`. It sets the value that will be returned if a key is not present in hash:

``` ruby
@importers.default = 'PuppetFactImporter'

@importers['zzz']
# => 'PuppetFactsImporter'
```

app/registries/foreman/plugin/importer_registry.rb
ShimShtein: We can skip the `indifferent_access` only if we agree on the type of keys - either `:puppet` or `'puppet'`. The problem is that from code it's better to work with symbols (`registry.get(:puppet)`), but  the request uses string as parameters.

app/registries/foreman/plugin/importer_registry.rb
bastilian: @ShimShtein I had no idea about `default`. Shouldn't we in that case avoid it even more?

rabajaj0509: @bastilian  agreed, I think we should avoid the `default`. Will still wait for @ShimShtein on this.

rabajaj0509:
@bastilian so what do you suggest, should I raise a pr in your branch? or you will do the needful? its your call, I have no issues :)

bastilian:
@rahulbajaj0509 We will refactor registries at a later point. Let's first see this PR is merged. :) The duplicate `attr_accessor` should be resolved, the other to I'll leave up to @ShimShtein to decide whether or not to use indifferent access.

app/registries/foreman/plugin/importer_registry.rb
```diff
@@ -0,0 +1,26 @@
+module Foreman
+  class Plugin
+    class ImporterRegistry
+      attr_accessor :importers 
+
+      def initialize
+        @importers = {}
+        @importers.default = 'PuppetFactImporter'
```
ShimShtein: Can we move the default setting to initializer? As a long-term goal, I want to move puppet to a plugin and then this line would not be possible completely.

app/registries/foreman/plugin/importer_registry.rb
```diff
@@ -0,0 +1,26 @@
...
+      end
+
+      def register(key, klass, default = false)
```
ShimShtein: What about adding an option to register a class as a constant, not only as string:

```ruby
importers.register(:puppet, PuppetFactImporter, true)
```
It's more readable, less prone for typos and more intuitive.
To support this, you don't have to store constants in the registry:

``` ruby
def register(key, klass, default = false)
  klass = klass.to_s
  #...
end
```

app/registries/foreman/plugin/importer_registry.rb
```diff
@@ -0,0 +1,26 @@
...
+        @importers[key.to_sym] = klass
+      end
+      alias register_importer register
+
+      def get(key)
+        @importers[key].constantize
+      end
+
+      def fact_features
+        importers.map { |_type, importer| importer.authorized_smart_proxy_features }.compact.flatten.uniq
```
ShimShtein: I think there is `constantize` missing here, `importers` has strings as values, not classes.

app/registries/foreman/plugin.rb
```diff
@@ -37,11 +38,12 @@ class PluginRequirementError < Foreman::Exception; end
   class Plugin
     @registered_plugins = {}
     @tests_to_skip = {}
+    @importer_registry = Plugin::ImporterRegistry.new
     @report_scanner_registry = Plugin::ReportScannerRegistry.new
 
     class << self
       attr_reader   :registered_plugins
-      attr_accessor :tests_to_skip, :report_scanner_registry
+      attr_accessor :tests_to_skip, :importer_registry, :report_scanner_registry
```
ShimShtein: nitpick: can we rename the attribute to `fact_importers`? The same goes for the instance accessor. I think it will be more readable:

``` ruby
Foreman::Plugin.fact_importers.get(:puppet)

Foreman::Plugin.register :foreman_puppet do
  fact_importers.register(:puppet, PuppetFactImporter, true)
end
```

Then we can remove the `delegate` from here and the `alias` from the registry class. They makes me a bit uncomfortable, and I think they won't be needed if we name the accessors better.

ShimShtein (changes requested):
We are getting closer, a couple of minor changes, but the whole idea looks really great!

mmoll:
also, this needs a rebase

app/registries/foreman/plugin/importer_registry.rb
rabajaj0509: @ShimShtein  so the idea here was that when i `get` the importer i constantize them. So imho, there is not need to do constantise here.

ShimShtein: If I read the code correctly, you are iterating over `importers` - this is the internal storage, so it keeps values as strings. This means `importer` parameter will have a string value. This means effectively you are calling `#authorized_smart_proxy_features` on string.

app/models/host/base.rb
```diff
@@ -144,6 +144,9 @@ def import_facts(facts, source_proxy = nil)
         importer.import!
       end
 
+      facts_importer = Foreman::Plugin.fact_importer.get(type).new(self, facts)
```
ShimShtein: There is already an `importer` defined a couple of [lines before](https://github.com/rahulbajaj0509/foreman/blob/823e960a27b13b86ab9f596d1a8f383a0c607941/app/models/host/base.rb#L141)

config/initializers/puppet.rb
```diff
@@ -12,7 +12,6 @@
     :priority => 100
 end
 
-FactImporter.register_fact_importer :puppet, PuppetFactImporter, true
```
ShimShtein, line 15: Where do you plan to register `PuppetFactImporter`, I couldn't find it in the PR?

ShimShtein (changes requested):
Can we add a test for `ImporterRegistry`?

I think it will make registry's interface a bit more clear.

app/models/host/base.rb
lzap: What Shim says, this must be imported just once. Also use telemetry wrappers.

lzap:
Please create new ticket for the issue, this is a refactoring that has nothing to do with RM 21654 which reports nil exception when Rails reloads importing stack.

app/registries/foreman/plugin/importer_registry.rb
```diff
@@ -0,0 +1,25 @@
...
+      def get(key)
+        @importers[key].constantize
+      end
+
+      def fact_features
+        importers.map { |_type, importer| importer.authorized_smart_proxy_features }.compact.flatten.uniq.constantize
+      end
+    end
+  end
+end
```
lzap: I don't understand this - why plugins would create registries? Isn't this job of Foreman core to offer such a service?

lzap:
Rahul, can you explain what is the motivation of this refactoring? Discovery simply uses PuppetImporter from core. We use unmodified version of it, why the patch then?

ShimShtein:
> Rahul, can you explain what is the motivation of this refactoring? Discovery simply uses PuppetImporter from core. We use unmodified version of it, why the patch then?

@lzap, I think I can elaborate on this:

The problem is that in the old (pre-PR) code importers registry is kept in `FactImporter` class that gets reloaded. Once the class is reloaded, it looses its default importer registration that point to puppet. See [my comment](https://github.com/theforeman/foreman/pull/5010#issuecomment-344946676) for more information.

> I don't understand this - why plugins would create registries? Isn't this job of Foreman core to offer such a service?

The PR adds a single registry to `Plugin` class (as a class level variable). Now every plugin can access that singleton and subscribe their own fact importer. Since `Plugin` class is not reloaded, the registry will retain records even after reloads.

ShimShtein:
@rahulbajaj0509, tests are failing.

lzap:
Thanks, I've updated https://projects.theforeman.org/issues/21654 keep going!

rabajaj0509:
@lzap @ShimShtein can you test this pr now? no tests are failing.

app/registries/foreman/plugin.rb
```diff
@@ -38,13 +39,16 @@ class PluginRequirementError < Foreman::Exception; end
   class Plugin
     @registered_plugins = {}
     @tests_to_skip = {}
+    @fact_importer = Plugin::ImporterRegistry.new
     @report_scanner_registry = Plugin::ReportScannerRegistry.new
     @report_origin_registry = Plugin::ReportOriginRegistry.new
```
lzap, line 45: Nitpick - naming convention `_importer` vs `_registry`.

lzap (commented):
LGTM

ShimShtein:
@rahulbajaj0509, getting closer...

- [x] merge conflicts need resolving
- [x] lzap's comment
- [x] my [comment](https://github.com/theforeman/foreman/pull/5010/files#r178622929)

app/registries/foreman/plugin/importer_registry.rb
```diff
@@ -0,0 +1,24 @@
...
+        @importers.default = klass.to_s if default
+        @importers[key.to_sym] = klass
+      end
+
+      def get(key)
+        @importers[key].constantize
+      end
+
+      def fact_features
+        @importers.map { |_type, importer| importer.authorized_smart_proxy_features }.compact.flatten.uniq.constantize
```
ShimShtein: Instead of `@importers.map` we can use `@importers.values`

app/registries/foreman/plugin/importer_registry.rb
```diff
@@ -0,0 +1,24 @@
...
+    class ImporterRegistry
+      attr_accessor :importers
+
+      def initialize
+        @importers = {}
+      end
+
+      def register(key, klass, default = false)
+        @importers.default = klass.to_s if default
+        @importers[key.to_sym] = klass
```
ShimShtein: We need to store string value of `klass` here too, as we do with default value.

ShimShtein (changes requested):
@rahulbajaj0509 I still miss tests for the registry. Can you please add a couple for `register`and 'get' with all combinations of specifying class, symbol and string as parameters. A couple of tests for `fact_features` with different features combinations reported by importers would be nice too.

ShimShtein:
@rahulbajaj0509, what's the status here? I remember vaguely that we decided to add some tests. In the meantime we need another rebase.

lzap:
Guys this is a *nasty* bug in discovery and I hate it so much. Breaks my development workflow, I would  *love* to see this merged.

ares:
rebase would be a good start :-)

ShimShtein:
[test foreman]

ShimShtein:
@rahulbajaj0509, almost there. The code looks good, no complaints there. Test failures are related though: we have a mechanism that ensures that there is no active transaction open when facts are imported (it was a nasty behavior with nested transactions in postgres). In tests, we need to disable this check, but we want to know which tests will be affected by the behavior, since it causes unpredictable behavior in other tests. We disable it in [fact_importer_test_helper.rb](https://github.com/theforeman/foreman/blob/develop/test/fact_importer_test_helper.rb). We need to adjust the behavior of `allow_transactions_for_any_importer` to our new registry.

@jlsherrill, heads up for Katello. When this one will be merged, Katello will have to modify its fact importer registration in [engine.rb](https://github.com/Katello/katello/blob/e29d5da8207cbaee4cdc2c8334543799ec05abf0/lib/katello/engine.rb#L162).

P.S. @rahulbajaj0509, now that "how to create a plugin" documentation lives in this repo, I would like to see this feature documented there. I will not block this PR (since the original behavior is not documented either), as long as you promise adding documentation to it later.

lzap:
Any progress here?

rabajaj0509:
Actually, I tried but I am not able to understand the way the tests are to be fixed here! @ShimShtein  actually explained to me personally too but it doesn't work for me. I guess I need to get a look t it again.

ShimShtein:
@rahulbajaj0509, ping me, if you need help.

ares:
@rahulbajaj0509 thanks for your effort here, do you think you'll be able to get back to this in foreseeable feature? We constantly try to keep PR queue up to date and clean stale PRs, so we'd suggest closing this in 14 days from now if we don't hear back from you. Of course it can be easily reopened later.

rabajaj0509:
Yes, I am working on this one. I was not able to understand the tests failure and still am struggling with the same. @ShimShtein is helping me though. 

@ares Lets keep this pr open.

rabajaj0509:
@ShimShtein does https://github.com/theforeman/foreman/pull/5010/files#diff-a721a18fc92632bd598c97a24cc5133eR10 make sense? Tests are still failing though!

test/fact_importer_test_helper.rb
```diff
@@ -7,7 +7,7 @@ def allow_transactions_for(importer)
   end
 
   def allow_transactions_for_any_importer
-    FactImporter.importers.values.each do |importer|
+    Foreman::Plugin.fact_registry.values.each do |importer|
```
ShimShtein:
```suggestion
    Foreman::Plugin.fact_registry.importers.values.each do |importer|
```

rabajaj0509:
Once this is merged,  @jlsherrill we would need to merge this https://github.com/Katello/katello/pull/8368  :)

ShimShtein:
[test foreman]

test/fact_importer_test_helper.rb
```diff
@@ -7,7 +7,7 @@ def allow_transactions_for(importer)
   end
 
   def allow_transactions_for_any_importer
-    FactImporter.importers.values.each do |importer|
+    Foreman::Plugin.fact_registry.importers.values.each do |importer|
```
rabajaj0509: @ShimShtein should I do something like `Foreman::Plugin.fact_registry.importers.values.map(&:constantize).each do |importer|` ?

ShimShtein: Yes, makes sense.

ShimShtein:
@rahulbajaj0509 tests are still failing. I can't see the relevant log lines, can you run those tests locally and see what is the underlying exception?

mmoll:
@rahulbajaj0509 What's the status here?

rabajaj0509:
@mmoll hey, there are a few test cases failing here, I will try to resolve this next week.

rabajaj0509:
[test foreman]

rabajaj0509:
@ShimShtein most of the test cases are failing because of the chrome version
```
Selenium::WebDriver::Error::SessionNotCreatedError: session not created: Chrome version must be between 71 and 75...
```

mmoll:
@rahulbajaj0509 please rebase.

rabajaj0509:
@ShimShtein list of possible errors:

1.
```
[MinitestRetry] retry 'test_0001_index page' count: 2,  msg: Selenium::WebDriver::Error::UnknownError: unknown error: cannot find Chrome binary
    test/integration_test_helper.rb:275:in `login_admin'
    test/test_helper.rb:68:in `before_setup'
    test/integration_test_helper.rb:304:in `before_setup'
```

2.
```
LocationIntegrationTest#test_0001_index page has notice if nil hosts = [MinitestRetry] retry 'test_0001_index page has notice if nil hosts' count: 1,  msg: ActionView::Template::Error: Could not load compiled manifest from /home/vagrant/foreman/public/webpack/manifest.json - have you run `rake webpack:compile`? (original error No such file or directory @ rb_sysopen - /home/vagrant/foreman/public/webpack/manifest.json)
    app/views/layouts/base.html.erb:13:in `_c7e1967e070c90de82470a9b67846424'
    app/views/layouts/application.html.erb:5:in `_99c9f09852aebadcdf29bd6e3598d865'
    app/controllers/concerns/foreman/controller/taxonomies_controller.rb:24:in `block (2 levels) in index'
    app/controllers/concerns/foreman/controller/taxonomies_controller.rb:21:in `index'
    app/controllers/concerns/foreman/controller/timezone.rb:10:in `set_timezone'
    app/models/concerns/foreman/thread_session.rb:32:in `clear_thread'
    app/controllers/concerns/foreman/controller/topbar_sweeper.rb:12:in `set_topbar_sweeper_controller'
    /home/vagrant/katello/lib/katello/middleware/event_daemon.rb:10:in `call'
    lib/foreman/middleware/telemetry.rb:10:in `call'
    lib/foreman/middleware/catch_json_parse_errors.rb:9:in `call'
    lib/foreman/middleware/logging_context_session.rb:22:in `call'
    lib/foreman/middleware/logging_context_request.rb:11:in `call'
    /home/vagrant/katello/lib/katello/prevent_json_parsing.rb:12:in `call'
    test/integration/location_test.rb:11:in `block in <class:LocationIntegrationTest>'
```

3.
```
[MinitestRetry] retry 'test_0014_should reject subnet with mismatched taxonomy in host' count: 2,  msg: ActiveRecord::RecordNotFound: Couldn't find Setting with [WHERE "settings"."name" = $1]
    app/models/setting.rb:103:in `[]='
    /home/vagrant/katello/app/models/katello/concerns/location_extensions.rb:21:in `reset_settings'
    test/models/nic_test.rb:131:in `block (2 levels) in <class:NicTest>'
    test/models/nic_test.rb:123:in `each'
    test/models/nic_test.rb:123:in `block in <class:NicTest>'
```

4.
```
FilterTest::with taxnomies#test_0001_filter is not automatically scoped to any taxonomies = [MinitestRetry] retry 'test_0001_filter is not automatically scoped to any taxonomies' count: 1,  msg: ActiveRecord::RecordNotFound: Couldn't find Setting with [WHERE "settings"."name" = $1]
    app/models/setting.rb:103:in `[]='
    /home/vagrant/katello/app/models/katello/concerns/location_extensions.rb:21:in `reset_settings'
    test/factories/disable_auditing.rb:13:in `block (3 levels) in <top (required)>'
    test/factories/disable_auditing.rb:13:in `block (2 levels) in <top (required)>'
    test/models/filter_test.rb:105:in `block (3 levels) in <class:FilterTest>'
    test/active_support_test_case_helper.rb:86:in `as_user'
    test/active_support_test_case_helper.rb:92:in `as_admin'
    test/models/filter_test.rb:102:in `block (2 levels) in <class:FilterTest>'
    test/test_helper.rb:68:in `before_setup'
```

5.
```
[MinitestRetry] retry 'test_0001_fetching locations attributes' count: 2,  msg: ActiveRecord::RecordNotFound: Couldn't find Setting with [WHERE "settings"."name" = $1]
    app/models/setting.rb:103:in `[]='
    /home/vagrant/katello/app/models/katello/concerns/location_extensions.rb:21:in `reset_settings'
    test/factories/disable_auditing.rb:13:in `block (3 levels) in <top (required)>'
    test/factories/disable_auditing.rb:13:in `block (2 levels) in <top (required)>'
    test/graphql/queries/locations_query_test.rb:30:in `block in <class:LocationsQueryTest>'
    test/test_helper.rb:68:in `before_setup'
```

6.
```
Api::V2::DomainsControllerTest::user with resource permissions scoped to 2 organizations#test_0009_user can create domain but current context is always preselected as organization_ids = [MinitestRetry] retry 'test_0009_user can create domain but current context is always preselected as organization_ids' count: 1,  msg: ActiveRecord::RecordNotFound: Couldn't find Setting with [WHERE "settings"."name" = $1]
    app/models/setting.rb:103:in `[]='
    /home/vagrant/katello/app/models/katello/concerns/location_extensions.rb:21:in `reset_settings'
    test/factories/disable_auditing.rb:13:in `block (3 levels) in <top (required)>'
    test/factories/disable_auditing.rb:13:in `block (2 levels) in <top (required)>'
    test/controllers/api/v2/domains_controller_test.rb:209:in `setup'
```

7.
```
[MinitestRetry] retry 'test_0017_selected_or_inherited_ids for inherited taxonomy' count: 2,  msg: ActiveRecord::RecordNotFound: Couldn't find Setting with [WHERE "settings"."name" = $1]
    app/models/setting.rb:103:in `[]='
    /home/vagrant/katello/app/models/katello/concerns/location_extensions.rb:21:in `reset_settings'
    test/models/shared/taxonomies_base_test.rb:263:in `block (2 levels) in <module:TaxonomiesBaseTest>'
```

8.
```
[MinitestRetry] retry 'test_0001_tabs are present' count: 2,  msg: ActionView::Template::Error: Could not load compiled manifest from /home/vagrant/foreman/public/webpack/manifest.json - have you run `rake webpack:compile`? (original error No such file or directory @ rb_sysopen - /home/vagrant/foreman/public/webpack/manifest.json)
    app/views/layouts/base.html.erb:13:in `_c7e1967e070c90de82470a9b67846424'
    app/views/layouts/application.html.erb:5:in `_99c9f09852aebadcdf29bd6e3598d865'
    app/controllers/concerns/foreman/controller/timezone.rb:10:in `set_timezone'
    app/models/concerns/foreman/thread_session.rb:32:in `clear_thread'
    app/controllers/concerns/foreman/controller/topbar_sweeper.rb:12:in `set_topbar_sweeper_controller'
    /home/vagrant/katello/lib/katello/middleware/event_daemon.rb:10:in `call'
    lib/foreman/middleware/telemetry.rb:10:in `call'
    lib/foreman/middleware/catch_json_parse_errors.rb:9:in `call'
    lib/foreman/middleware/logging_context_session.rb:22:in `call'
    lib/foreman/middleware/logging_context_request.rb:11:in `call'
    /home/vagrant/katello/lib/katello/prevent_json_parsing.rb:12:in `call'
    test/integration_test_helper.rb:57:in `assert_new_button'
    test/integration/host_test.rb:31:in `block (2 levels) in <class:HostIntegrationTest>'
```

9.
```
/home/vagrant/foreman/.vendor/ruby/2.5.0/gems/childprocess-3.0.0/lib/childprocess/abstract_process.rb:188:in `assert_started': process not started (ChildProcess::Error)
Traceback (most recent call last):
	9: from -e:1:in `<main>'
	8: from /opt/rh/rh-ruby25/root/usr/share/rubygems/rubygems/core_ext/kernel_require.rb:59:in `require'
	7: from /opt/rh/rh-ruby25/root/usr/share/rubygems/rubygems/core_ext/kernel_require.rb:59:in `require'
	6: from /home/vagrant/foreman/.vendor/ruby/2.5.0/gems/selenium-webdriver-3.142.6/lib/selenium/webdriver/common/platform.rb:150:in `block in exit_hook'
	5: from /home/vagrant/foreman/.vendor/ruby/2.5.0/gems/selenium-webdriver-3.142.6/lib/selenium/webdriver/common/service.rb:123:in `stop'
	4: from /home/vagrant/foreman/.vendor/ruby/2.5.0/gems/selenium-webdriver-3.142.6/lib/selenium/webdriver/common/service.rb:123:in `ensure in stop'
	3: from /home/vagrant/foreman/.vendor/ruby/2.5.0/gems/selenium-webdriver-3.142.6/lib/selenium/webdriver/common/service.rb:176:in `stop_process'
	2: from /home/vagrant/foreman/.vendor/ruby/2.5.0/gems/selenium-webdriver-3.142.6/lib/selenium/webdriver/common/service.rb:193:in `process_exited?'
	1: from /home/vagrant/foreman/.vendor/ruby/2.5.0/gems/childprocess-3.0.0/lib/childprocess/unix/process.rb:31:in `exited?'
/home/vagrant/foreman/.vendor/ruby/2.5.0/gems/childprocess-3.0.0/lib/childprocess/abstract_process.rb:188:in `assert_started': process not started (ChildProcess::Error)
```

I am not sure what causes the actual error but these are the ones that I am getting on my local instance

mmoll:
asset compilation fails:
```
# bundle exec rake webpack:compile RAILS_ENV=production
rake aborted!
TypeError: no implicit conversion of nil into String
/home/jenkins/workspace/test_develop_pr_core/database/postgresql-integrations/ruby/2.5/slave/fast/config/initializers/assets.rb:66:in `file?'
/home/jenkins/workspace/test_develop_pr_core/database/postgresql-integrations/ruby/2.5/slave/fast/config/initializers/assets.rb:66:in `block (4 levels) in <top (required)>'
/home/jenkins/workspace/test_develop_pr_core/database/postgresql-integrations/ruby/2.5/slave/fast/config/initializers/assets.rb:66:in `each'
/home/jenkins/workspace/test_develop_pr_core/database/postgresql-integrations/ruby/2.5/slave/fast/config/initializers/assets.rb:66:in `detect'
/home/jenkins/workspace/test_develop_pr_core/database/postgresql-integrations/ruby/2.5/slave/fast/config/initializers/assets.rb:66:in `block (3 levels) in <top (required)>'
/home/jenkins/workspace/test_develop_pr_core/database/postgresql-integrations/ruby/2.5/slave/fast/config/initializers/assets.rb:57:in `each'
/home/jenkins/workspace/test_develop_pr_core/database/postgresql-integrations/ruby/2.5/slave/fast/config/initializers/assets.rb:57:in `block (2 levels) in <top (required)>'
/usr/local/rvm/gems/ruby-2.5.1@test_develop_pr_core-2/gems/activesupport-5.2.1/lib/active_support/lazy_load_hooks.rb:69:in `block in execute_hook'
/usr/local/rvm/gems/ruby-2.5.1@test_develop_pr_core-2/gems/activesupport-5.2.1/lib/active_support/lazy_load_hooks.rb:62:in `with_execution_control'
/usr/local/rvm/gems/ruby-2.5.1@test_develop_pr_core-2/gems/activesupport-5.2.1/lib/active_support/lazy_load_hooks.rb:67:in `execute_hook'
/usr/local/rvm/gems/ruby-2.5.1@test_develop_pr_core-2/gems/activesupport-5.2.1/lib/active_support/lazy_load_hooks.rb:52:in `block in run_load_hooks'
/usr/local/rvm/gems/ruby-2.5.1@test_develop_pr_core-2/gems/activesupport-5.2.1/lib/active_support/lazy_load_hooks.rb:51:in `each'
/usr/local/rvm/gems/ruby-2.5.1@test_develop_pr_core-2/gems/activesupport-5.2.1/lib/active_support/lazy_load_hooks.rb:51:in `run_load_hooks'
/usr/local/rvm/gems/ruby-2.5.1@test_develop_pr_core-2/gems/railties-5.2.1/lib/rails/application/finisher.rb:75:in `block in <module:Finisher>'
/usr/local/rvm/gems/ruby-2.5.1@test_develop_pr_core-2/gems/railties-5.2.1/lib/rails/initializable.rb:32:in `instance_exec'
/usr/local/rvm/gems/ruby-2.5.1@test_develop_pr_core-2/gems/railties-5.2.1/lib/rails/initializable.rb:32:in `run'
/usr/local/rvm/gems/ruby-2.5.1@test_develop_pr_core-2/gems/railties-5.2.1/lib/rails/initializable.rb:61:in `block in run_initializers'
/usr/local/rvm/gems/ruby-2.5.1@test_develop_pr_core-2/gems/railties-5.2.1/lib/rails/initializable.rb:60:in `run_initializers'
/usr/local/rvm/gems/ruby-2.5.1@test_develop_pr_core-2/gems/railties-5.2.1/lib/rails/application.rb:361:in `initialize!'
/usr/local/rvm/gems/ruby-2.5.1@test_develop_pr_core-2/gems/railties-5.2.1/lib/rails/railtie.rb:190:in `public_send'
/usr/local/rvm/gems/ruby-2.5.1@test_develop_pr_core-2/gems/railties-5.2.1/lib/rails/railtie.rb:190:in `method_missing'
/home/jenkins/workspace/test_develop_pr_core/database/postgresql-integrations/ruby/2.5/slave/fast/config/environment.rb:5:in `<top (required)>'
/usr/local/rvm/gems/ruby-2.5.1@test_develop_pr_core-2/gems/activesupport-5.2.1/lib/active_support/dependencies.rb:287:in `require'
/usr/local/rvm/gems/ruby-2.5.1@test_develop_pr_core-2/gems/activesupport-5.2.1/lib/active_support/dependencies.rb:287:in `block in require'
/usr/local/rvm/gems/ruby-2.5.1@test_develop_pr_core-2/gems/activesupport-5.2.1/lib/active_support/dependencies.rb:253:in `load_dependency'
/usr/local/rvm/gems/ruby-2.5.1@test_develop_pr_core-2/gems/activesupport-5.2.1/lib/active_support/dependencies.rb:287:in `require'
/usr/local/rvm/gems/ruby-2.5.1@test_develop_pr_core-2/gems/railties-5.2.1/lib/rails/application.rb:337:in `require_environment!'
/usr/local/rvm/gems/ruby-2.5.1@test_develop_pr_core-2/gems/railties-5.2.1/lib/rails/application.rb:520:in `block in run_tasks_blocks'
/usr/local/rvm/gems/ruby-2.5.1@test_develop_pr_core-2/gems/rake-13.0.1/exe/rake:27:in `<top (required)>'
/usr/local/rvm/gems/ruby-2.5.1@test_develop_pr_core-2/bin/ruby_executable_hooks:24:in `eval'
/usr/local/rvm/gems/ruby-2.5.1@test_develop_pr_core-2/bin/ruby_executable_hooks:24:in `<main>'
Tasks: TOP => webpack:compile => environment
(See full trace by running task with --trace)
```

mmoll:
The core tests still have 20 errros in:
- test/unit/fact_importer_test.rb (easy one)
- test/controllers/api/v2/hosts_controller_test.rb(fact import tests)

ShimShtein:
@rahulbajaj0509, try and solve the first one. My guess would be that other errors are related to this one.

My gut feeling about it, is that puppet plugin is not registered correctly. Try and see if it is registered and that puppet fact importer is set as the default one. I think it is causing other tests to fail too, since the system expects puppet fact importer as the default for the system.

ShimShtein:
@rahulbajaj0509, the errors seem related.

test/unit/plugin/importer_registry_test.rb
```diff
@@ -0,0 +1,31 @@
...
+    fact_importer.register(:test_class, TestClass, false)
+    assert_equal fact_importer.get(:test_class), TestClass
+  end
+
+  def test_get_default
+    fact_importer = Foreman::Plugin::ImporterRegistry.new
+    fact_importer.register(:test_class, TestClass, true)
+    assert_equal fact_importer.get(:random_key), TestClass
+  end
+end
```
ShimShtein, line 38: Can you please add a test for `fact_features` method?
Something along the lines of
``` ruby
  def test_get_default
    fact_importer = Foreman::Plugin::ImporterRegistry.new
    fact_importer.register(:test_class, TestClass, true)
    TestClass.any_instance.expects(:authorized_smart_proxy_features).returns([:test_feature])
    assert_equal :foo, fact_importer.fact_features.first
  end
```

ShimShtein (changes requested):

rabajaj0509:
@ShimShtein looks good?

ShimShtein:
[test foreman]

mmoll:
asset precompilation still fails.

rabajaj0509:
@ShimShtein, error facing during asset pre-compilation! As mentioned by you on IRC, this is caused by plugin definition without path (puppet plugin in our case).

config/initializers/assets.rb
```diff
@@ -58,11 +58,12 @@ class << self
         # Manifests may be stored under the engine installation or under the Foreman app root, then
         # either with just the plugin name or with hyphens replaced with underscores.
         possible_manifests = [plugin.path, app.root].map do |root_dir|
+          next if root_dir.empty?
           [File.join(root_dir, "public/assets/#{plugin.id}/#{plugin.id}.json"),
            File.join(root_dir, "public/assets/#{plugin.id.to_s.tr('-', '_')}/#{plugin.id.to_s.tr('-', '_')}.json")]
         end.flatten
```
ShimShtein: You can use `compact` instead of `next if...`
```suggestion
        end.flatten.compact
```

ShimShtein (dismissed):
@tbrisker, ACK from me, feel free to merge this one once the branching is done

tbrisker:
Katello test seems related:
```
NoMethodError: undefined method `register_fact_importer' for FactImporter:Class
/home/jenkins/workspace/test_develop_pr_katello/database/postgresql/ruby/2.5/slave/fast/plugin/lib/katello/engine.rb:155:in `block in <class:Engine>'

**Comment by rabajaj0509** [comment]
@tbrisker that failure is expected. Once this PR is merged then katello PR: https://github.com/Katello/katello/pull/8368 should be merged.

**Comment by tbrisker** [comment]
@rahulbajaj0509 there are many other plugins using the same methods (a quick search only in the foreman organization for `register_fact_importer` shows 4 such, and there are likely others as well). Please deprecate them and change the implementation to use the new registry rather than removing them completely. That way plugin maintainers have time to update their plugins for the new usage instead of breaking once this is merged.

**Comment by rabajaj0509** [comment]
@tbrisker makes sense, there are 7 plugins that would require this change. Let me raise PRs for all of them and get back to you :) thanks for the insight!

**Comment by tbrisker** [comment]
@rahulbajaj0509 thanks for opening PRs for the plugins, I'm sure that will be appreciated by their maintainers. 
However there may be other custom plugins users are using that aren't on github. They deserve a deprecation warning for a couple of foreman releases before we break their plugins. Please keep the old methods in this PR with a deprecation warning and change their implementation to use the new registry.

**Comment by rabajaj0509** [comment]
Related PRs opened:
foreman_ansible: https://github.com/theforeman/foreman_ansible/pull/328
foreman_omaha: https://github.com/theforeman/foreman_omaha/pull/44
foreman_salt: https://github.com/theforeman/foreman_salt/pull/132
foreman_chef: https://github.com/theforeman/foreman_chef/pull/94
katello: https://github.com/Katello/katello/pull/8368

@tbrisker yes, makes sense. I will add a deprecation warning for a couple of foreman releases before we break their plugins.

File: app/services/fact_importer.rb [hunk 24]
```diff
@@ -4,24 +4,6 @@ class FactImporter
...
-  def self.register_fact_importer(key, klass, default = false)
-    importers.default = klass if default
-
-    importers[key.to_sym] = klass
-  end
-
-  def self.fact_features
-    importers.map { |_type, importer| importer.authorized_smart_proxy_features }.compact.flatten.uniq
-  end
-
```
tbrisker: These should all get deprecation warnings and change implementation to use the regstry.

tbrisker (changes requested):
Thanks for opening all of these PRs! 
Setting to WOC until the deprecations are implemented

app/services/fact_importer.rb
rabajaj0509: @ShimShtein @tbrisker all methods in the registry are instance methods where as these methods are class methods. Not sure how to change implementation to use the registry! I have tried something out but it is a wide guess.

ShimShtein:
@rahulbajaj0509 could you please rebase and fix hound?

lzap:
I haven't followed the conversation here but where are we guys? This is biting me quite often when I work with discovery in development mode :-)

app/services/fact_importer.rb
```diff
@@ -4,22 +4,28 @@ class FactImporter
   delegate :logger, :to => :Rails
   attr_reader :counters
 
+  def deprecation_warning
+    Foreman::Deprecation.deprecation_warning('2.0', 'Use Importer Registry to register or obtain facts for a plugin')
```
tbrisker: Please use a specific deprecation for each method indicating what method should replace it. Also, the deprecation warning should be for 2.2, not 2.0.

app/registries/foreman/plugin.rb
```diff
@@ -49,6 +51,8 @@ class << self
       attr_accessor :tests_to_skip, :report_scanner_registry,
         :report_origin_registry, :medium_providers,
         :graphql_types_registry
+      attr_accessor :fact_registry
```
tbrisker: Any particular reason not to list the `:fact_registry` with the other `attr_accessor`s?

app/registries/foreman/plugin.rb
```diff
@@ -39,6 +40,7 @@ class PluginRequirementError < Foreman::Exception; end
   class Plugin
     @registered_plugins = {}
     @tests_to_skip = {}
+    @fact_registry = Plugin::ImporterRegistry.new
```
tbrisker: This is a bit confusing - is this a fact registry or an importer registry? I would recommend naming the class and the variable wit the same name.

app/services/fact_importer.rb
```diff
@@ -4,22 +4,28 @@ class FactImporter
...
   def self.importer_for(type)
-    importers[type.to_s]
+    deprecation_warning
+    @importers.get(type)
   end
 
   def self.importers
-    @importers ||= {}.with_indifferent_access
+    deprecation_warning
+    @importers ||= Foreman::Plugin::ImporterRegistry.new
```
tbrisker: This should use the existing registry, not create a new one:
```suggestion
    Foreman::Plugin.fact_registry.importers
```

app/services/fact_importer.rb
```diff
@@ -4,22 +4,28 @@ class FactImporter
...
   attr_reader :counters
 
+  def deprecation_warning
+    Foreman::Deprecation.deprecation_warning('2.0', 'Use Importer Registry to register or obtain facts for a plugin')
+  end
+
   def self.importer_for(type)
-    importers[type.to_s]
+    deprecation_warning
+    @importers.get(type)
```
tbrisker:
```suggestion
    Foreman::Plugin.fact_registry.get(type)
```

app/services/fact_importer.rb
```diff
@@ -4,22 +4,28 @@ class FactImporter
...
+    deprecation_warning
+    @importers ||= Foreman::Plugin::ImporterRegistry.new
   end
 
   def self.register_fact_importer(key, klass, default = false)
-    importers.default = klass if default
-
-    importers[key.to_sym] = klass
+    deprecation_warning
+    @importers.register(key, klass, default)
```
tbrisker: Foreman::Plugin.fact_registry.
```suggestion
    Foreman::Plugin.fact_registry.register(key, klass, default)
```

app/services/fact_importer.rb
```diff
@@ -4,22 +4,28 @@ class FactImporter
...
   end
 
   def self.fact_features
-    importers.map { |_type, importer| importer.authorized_smart_proxy_features }.compact.flatten.uniq
+    deprecation_warning
+    @importers.fact_features
```
tbrisker:
```suggestion
    Foreman::Plugin.fact_registry.fact_features
```

config/initializers/puppet.rb
```diff
@@ -12,7 +12,9 @@
...
+Foreman::Plugin.register :foreman_puppet do
+  fact_registry.register(:puppet, PuppetFactImporter, true)
+end
```
tbrisker: IIUC, the changes to `config/initializers/assets.rb` and `config/environments/test.rb` are are due to this - since you are registering a plugin that isn't really a plugin to get to the registry. 
You can instead access the registry directly with:
```suggestion
Foreman::Plugin.fact_registry.register(:puppet, PuppetFactImporter, true)
```
Or take an approach like other registries (e.g. the report origin registry), of hard coding puppet in the default values.

tbrisker (changes requested):
See comments inline.

config/environments/test.rb
```diff
@@ -72,6 +72,7 @@
   # compatibility, allow them to override it by adding an ASDT configuration file.
   config.after_initialize do
     Foreman::Plugin.all.each do |plugin|
+      next if plugin.path.nil?
```
tbrisker: I believe this is no longer needed?

config/initializers/assets.rb
```diff
@@ -57,10 +57,10 @@ class << self
       Foreman::Plugin.all.each do |plugin|
         # Manifests may be stored under the engine installation or under the Foreman app root, then
         # either with just the plugin name or with hyphens replaced with underscores.
-        possible_manifests = [plugin.path, app.root].map do |root_dir|
+        possible_manifests = [plugin.path, app.root].compact.map do |root_dir|
           [File.join(root_dir, "public/assets/#{plugin.id}/#{plugin.id}.json"),
            File.join(root_dir, "public/assets/#{plugin.id.to_s.tr('-', '_')}/#{plugin.id.to_s.tr('-', '_')}.json")]
-        end.flatten
+        end.flatten.compact
```
tbrisker: these too?

app/services/fact_importer.rb
```diff
@@ -5,21 +5,23 @@ class FactImporter
   attr_reader :counters
 
   def self.importer_for(type)
-    importers[type.to_s]
+    Foreman::Deprecation.deprecation_warning('2.2', 'Use ImporterRegistry#get method instead of #importer_for')
+    Foreman::Plugin.fact_registry.get(type)
```
tbrisker: should these all be `importer_registry` now?

tbrisker (changes requested):

app/services/fact_importer.rb
```diff
@@ -5,21 +5,23 @@ class FactImporter
   attr_reader :counters
 
   def self.importer_for(type)
-    importers[type.to_s]
+    Foreman::Deprecation.deprecation_warning('2.2', 'Use ImporterRegistry#get method instead of #importer_for')
```
tbrisker: Just a little clarification so people know exactly what needs to be changed:
```suggestion
    Foreman::Deprecation.deprecation_warning('2.2', 'Use ImporterRegistry#get or Foreman::Plugin.importer_registry.get methods instead of FactImporter.importer_for')
```
Same applies to the other methods below as well.

app/registries/foreman/plugin.rb
```diff
@@ -39,6 +40,7 @@ class PluginRequirementError < Foreman::Exception; end
   class Plugin
     @registered_plugins = {}
     @tests_to_skip = {}
+    @importer_registry = Plugin::ImporterRegistry.new
```
tbrisker: Now that I think more about it, it makes sense to me to call it `fact_importer_registry`/`FactImporterRegistry` just so it's clear what is being imported - there are other things that could be imported into foreman. Sorry for not thinking about it earlier.

tbrisker (changes requested):
Almost there! just a couple of small comments left.

app/registries/foreman/plugin.rb
rabajaj0509: No problem, i will make the changes right away! 
@ShimShtein told me about it and we thought it makes sense to have it called as `FactImporterRegistry` but it becomes a little bigger for a class name. But now that we are all fine with the name, I will go with it :) thanks @tbrisker

app/services/fact_importer.rb
rabajaj0509: Sure

tbrisker (approved):
Thanks @rahulbajaj0509, @ShimShtein and all who took part in reviews!
@rahulbajaj0509 - can you please retrigger the tests in all the plugins once this is merged? they should be passing now.

rabajaj0509:
Thank you @ShimShtein @tbrisker @bastilian for holding on this PR till the end!  @lzap sorry for such a long wait on this :) I hope this issue does not hit us while working on Discovery any more \o/

lzap:
Works great with discovery! Thanks.

PR #5011, opened 2017-11-16
Fixes #21682 - Add Domain to Compute Resource API (OpenStack)

dLobatog:
The Domain field was added to the Compute Resources table in #12054 as
part of the Openstack v3 support. This field needs to appear in the
APIdoc to be able to use it and in the RABL to read it.


tbrisker (changes requested):
I think you also need to permit the param in `Foreman::Controller::Parameters::ComputeResource`

tbrisker (approved):
Oops, my bad, was on an older tree. Code looks good but can't test it.

dLobatog (commented):
To test this PR it isn't really necessary to have OpenStack, just create a mock one in the console and ensure the attribute can be seen/updated through the API

tbrisker:
Tested, works fine. Thanks @dLobatog !

PR #5012, opened 2017-11-16
Fixes #21693 - Honor intermediate certificates in ssl_cert

tbe:
This is needed to establish SSL communication with smart-proxy if:
- SSL client authentication is enabled on smart-proxy
- the client certificate is not signed by a root CA and needs
  intermediate certificates

This fixes #21693

### DIFF
--- a/lib/proxy_api/resource.rb
+++ b/lib/proxy_api/resource.rb
@@ -1,4 +1,5 @@
 require 'securerandom'
+require 'rest-client-patch'
 
 module ProxyAPI
   class Resource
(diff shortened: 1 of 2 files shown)


ares:
Thanks for deep investigation followed by the patch, I was involved on IRC channel. Would it also make sense to file a PR against rest-client so we don't have to keep the monkey patch forever? :-)

dLobatog (commented):
[test foreman]
[test katello]
[test upgrade]

@tbe Could you provide a fake cert or something to test this? I'm not sure how exactly to test & there are no tests (understandably, this should be a patch to rest-client really). Thanks!

tbrisker:
ok to test

xprazak2:
Setting to WoC as a part of this might go to rest-client and testing/verification steps are needed.

PR #5013, opened 2017-11-16
Fixes #21699 - fix vertical navigation transition animation glitch

amirfefer:
transitions between pages via the vertical menu in collapse mode look a bit strange, as you can see here: 
nov 17 2017 12-12 am

transitions should render more fluently:
nov 17 2017 12-09 am

### DIFF
--- a/app/assets/stylesheets/patternfly_and_overrides.scss
+++ b/app/assets/stylesheets/patternfly_and_overrides.scss
@@ -377,3 +377,6 @@ a.btn-info span {
   }
 }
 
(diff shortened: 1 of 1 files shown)


tbrisker:
Doesn't seem to work:
recorded

amirfefer:
@tbrisker it works when a successfully turbolink page load occurs.
if there's a full page load (such as when a new js file is required) - no effect.

tbrisker (approved):

tbrisker:
Thanks @amirfefer !

PR #5014, opened 2017-11-17
Fixes #21702 - Add reducer registration to redux

danseethaler:
This enables registering reducers from plugins asynchronously. I put up a gist [here](https://gist.github.com/danseethaler/d745921bebbe5f696c4a316e38d8879f) to see the way it would be used in Katello.

The basic idea is to add the `registerReducer` function which allows you to add registered reducers and then updates the store using redux's [replaceReducer](https://redux.js.org/docs/api/Store.html#replaceReducer) api.

http://projects.theforeman.org/issues/21702

### DIFF
--- a/webpack/assets/javascripts/react_app/common/MountingService.js
+++ b/webpack/assets/javascripts/react_app/common/MountingService.js
@@ -5,6 +5,8 @@ import React from 'react';
 import store from '../redux';
 import componentRegistry from '../components/componentRegistry';
 
+export { default as registerReducer } from '../redux/reducers/registerReducer';
+
 export function mount(component, selector, data) {
   const reactNode = document.querySelector(selector);
 
--- a/webpack/assets/javascripts/react_app/components/notifications/notifications.test.js
+++ b/webpack/assets/javascripts/react_app/components/notifications/notifications.test.js
@@ -5,7 +5,7 @@ import React from 'react';
 import configureMockStore from 'redux-mock-store';
 import thunk from 'redux-thunk';
 
-import { getStore } from '../../redux';
+import { generateStore } from '../../redux';
 import API from '../../API';
 
 import {
@@ -87,7 +87,7 @@ describe('notifications', () => {
   });
 
   it('full flow', () => {
-    const wrapper = mount(<Notifications data={componentMountData} store={getStore()} />);
+    const wrapper = mount(<Notifications data={componentMountData} store={generateStore()} />);
 
     wrapper.find('.fa-bell').simulate('click');
     expect(wrapper.find('.panel-group').length).toEqual(1);
@@ -98,7 +98,7 @@ describe('notifications', () => {
   });
 
   it('mark group as read flow', () => {
-    const wrapper = mount(<Notifications data={componentMountData} store={getStore()} />);
+    const wrapper = mount(<Notifications data={componentMountData} store={generateStore()} />);
     const matcher = '.drawer-pf-action a.btn-link';
 
     wrapper.find('.fa-bell').simulate('click');
@@ -113,12 +113,12 @@ describe('notifications', () => {
     window.location.replace = jest.fn();
     failResponse = { status: 401 };
 
-    mount(<Notifications data={componentMountData} store={getStore()} />);
+    mount(<Notifications data={componentMountData} store={generateStore()} />);
     expect(global.location.replace).toBeCalled();
   });
 
   it('should avoid multiple polling on re-mount', () => {
-    const store = getStore();
+    const store = generateStore();
     const spy = jest.spyOn(API, 'get');
 
     mount(<Notifications data={componentMountData} store={store} />);
@@ -128,7 +128,7 @@ describe('notifications', () => {
   });
 
   it('should close the notification box when click the close button', () => {
-    const wrapper = mount(<Notifications data={componentMountData} store={getStore()} />);
+    const wrapper = mount(<Notifications data={componentMountData} store={generateStore()} />);
     const closeButtonSelector = '.drawer-pf .drawer-pf-title a.drawer-pf-close';
 
     wrapper.find('.fa-bell').simulate('click');
@@ -141,7 +141,7 @@ describe('notifications', () => {
   it('should close the notification box when click outside of the box', () => {
     const wrapper = mount(<div>
       <div className="something-outside" />
-      <Notifications data={componentMountData} store={getStore()} />
+      <Notifications data={componentMountData} store={generateStore()} />
                           </div>);
 
     wrapper.find('.fa-bell').simulate('click');
--- a/webpack/assets/javascripts/react_app/redux/index.js
+++ b/webpack/assets/javascripts/react_app/redux/index.js
@@ -2,7 +2,7 @@ import createLogger from 'redux-logger';
 import thunk from 'redux-thunk';
 import { applyMiddleware, createStore } from 'redux';
 
-import reducer from './reducers';
+import reducers from './reducers';
 
 let middleware = [thunk];
 
@@ -10,12 +10,13 @@ if (process.env.NODE_ENV !== 'production' && !global.__testing__) {
   middleware = [...middleware, createLogger()];
 }
 
-const _getStore = () => createStore(
-  reducer,
-  window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__(),
-  applyMiddleware(...middleware),
-);
+export const generateStore = () =>
+  createStore(
+    reducers,
+    window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__(),
+    applyMiddleware(...middleware),
+  );
 
-export default _getStore();
+const store = generateStore();
 
-export const getStore = _getStore;
+export default store;
--- a/webpack/assets/javascripts/react_app/redux/reducers/index.js
+++ b/webpack/assets/javascripts/react_app/redux/reducers/index.js
@@ -4,9 +4,14 @@ import hosts from './hosts';
 import notifications from './notifications/';
 import toasts from './toasts';
 
-export default combineReducers({
-  statistics,
-  hosts,
-  notifications,
-  toasts,
-});
+export function combineReducersAsync(asyncReducers) {
+  return combineReducers({
+    statistics,
+    hosts,
+    notifications,
+    toasts,
+    ...asyncReducers,
+  });
+}
+
+export default combineReducersAsync();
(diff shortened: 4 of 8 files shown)


webpack/assets/javascripts/react_app/common/MountingService.js
```diff
@@ -5,6 +5,8 @@ import React from 'react';
 import store from '../redux';
 import componentRegistry from '../components/componentRegistry';
 
+export { default as store, registerReducer } from '../redux';
```
danseethaler: I'm not sure if this is the best place to add these exports to expose them to plugins. I just took a shortcut since `foremanReact/common/MountingService` was already available.

danseethaler:
This PR would imply that all plugins have access to manipulate anything in the global store. It may be worth some discussion on name-spacing store properties. I used `katello_reducers` [here](https://gist.github.com/danseethaler/d745921bebbe5f696c4a316e38d8879f#file-reducers-js-L11) which seems like a simple way to get started.

danseethaler:
Closed in favor of #4925 

Hopefully we can get that figured out and merged soon.

webpack/assets/javascripts/react_app/redux/index.js
```diff
@@ -1,21 +1,31 @@
...
+    combineReducersAsync(),
+    window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__(),
+    applyMiddleware(...middleware),
+  );
+
+  store.asyncReducers = {};
+  return store;
+}
+
+export function registerReducer(store, name, asyncReducer) {
```
xprazak2: number of arguments do not correspond to what is in the [example gist](https://gist.github.com/danseethaler/d745921bebbe5f696c4a316e38d8879f)

Worked as expected when I changed it to:

```js
const store = configureStore();

export function registerReducer(name, asyncReducer) {
// no changes here
}

export default store;
```

xprazak2:
Looking good, I left just one comment. It would be great if we could also extend the lower levels of state tree as [discussed](https://github.com/theforeman/foreman/pull/4733#issuecomment-329809970) but I would not block this because of it.

webpack/assets/javascripts/react_app/redux/index.js
danseethaler: Correct, I refactored a bit to remove the need to expose the store. It should be a smaller api now with just the name and the reducer.

webpack/assets/javascripts/react_app/redux/index.js
```diff
@@ -2,20 +2,27 @@ import createLogger from 'redux-logger';
...
 if (process.env.NODE_ENV !== 'production' && !global.__testing__) {
   middleware = [...middleware, createLogger()];
 }
 
-const _getStore = () => createStore(
-  reducer,
-  window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__(),
-  applyMiddleware(...middleware),
-);
+export const generateStore = () =>
```
danseethaler, line 13: I renamed this to `generateStore` since it's actually generating a new store each time. The main application store is exported as the default but test files use `generateStore` to generate a unique store for the test suite.

waldenraines (approved):
LGTM

danseethaler:
@xprazak2 based on my understanding the `replaceReducer` api basically enforces [slice reducers](https://redux.js.org/docs/basics/Reducers.html#splitting-reducers) which namespaces the state that each registered reducer manages. So I don't think it's currently possible to allow plugin reducers to interact with the store created by foreman.

I'm not sure what the use cases are for shared state management but if we can make it work it's actually sort of nice to have the plugins isolated so there's no confusion about who is managing what parts of the store.

Here is a short discussion I found just for reference.
https://github.com/reactjs/redux/issues/2170#issuecomment-269355344

xprazak2:
I retested and it works nicely.

@danseethaler from what I know, the `replaceReducer` works as you describe

Regarding the use cases, now we can do:



where `content` is added from a plugin. I was thinking that katello might want to add `content` under the already existing `hosts` node in the state tree:



But having plugins contained to their own state slice will definitely prevent them from clashing with each other.

danseethaler:
Anyone opposed to this or want more time to review it?

danseethaler:
@ohadlevy mind taking a look?

danseethaler:
@amirfefer @sharvit mind taking a look at this?

sharvit:
I think you can separate to more files and get rid of the `reducers/index.js`

```js
// staticReducers.js
export const staticReducers = { /* static reducers here */ };

// registerAsyncReducer.js
const registerAsyncReducer = {};

export const registerAsyncReducer = (name, asyncReducer) => {
  asyncReducers[name] = asyncReducer;

  const reducers = combineReducers({ ...staticReducers, ...asyncReducers });
  store.replaceReducer(reducers);
};

// store.js
const reducers = combineReducers(staticReducers);

const generateStore = () =>
  createStore(
    reducers,
    window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__(),
    applyMiddleware(...middleware),
  );

export const store = generateStore();
```

danseethaler:
@sharvit thanks for taking a look!

I think it makes sense to have the `combineReducersAsync` function in the `reducers` directory. We could pull it into `redux/index.js` where the store is exported but having it under `reducers/index.js` is much more common from what I've seen. We'd also need to add a `staticReducers.js` file since we don't have that right now.

Any particular benefit you're targeting?

sharvit:
@danseethaler the benefit is that it is easier to understand the code that way.

By reading the file names I can get a basic understanding of what they deliver:
`reducers/index.js` - exports list of reducers
`store/index.js` - exports store
`staticReducers.js` - exports list of static reducers

It dosen't make seanse to import `reducers/index.js` and run it as a function that recive reducers object.
as long as we don't have combinedReducers exported as const anymore, i don't see a reason to have `reducers/index.js`.

danseethaler:
@sharvit I see what you're saying. Take a look at the updated files. It's now exporting the reducers from the `reducers/index.js` file but also exposing the `combineReducersAsync` function to be used in the `registerAsyncReducer.js` file.

I still prefer the reducers to be exported from `reducers/index.js` instead of running `combineReducers` in the `store.js` file.

sharvit (approved):
@danseethaler it looks much better now  :+1:

webpack/assets/javascripts/react_app/components/notifications/notifications.test.js
```diff
@@ -5,7 +5,7 @@ import React from 'react';
 import configureMockStore from 'redux-mock-store';
 import thunk from 'redux-thunk';
 
-import { getStore } from '../../redux';
+import { generateStore } from '../../redux';
```
ohadlevy, line 8: whats the rational to rename the method?

ohadlevy (changes requested):
one nitpick inline, can you add tests as well? thanks!

webpack/assets/javascripts/react_app/components/notifications/notifications.test.js
danseethaler: I renamed this to `generateStore` since it's actually generating a new store each time. The main application store is exported as the default but test files use `generateStore` to generate a unique store for the test suite.

webpack/assets/javascripts/react_app/redux/reducers/index.js
```diff
@@ -4,9 +4,14 @@ import hosts from './hosts';
...
 
-export default combineReducers({
-  statistics,
-  hosts,
-  notifications,
-  toasts,
-});
+export function combineReducersAsync(asyncReducers) {
+  return combineReducers({
+    statistics,
```
ohadlevy, line 9: does it make sense to use the new api for all redcuers?

webpack/assets/javascripts/react_app/redux/reducers/index.js
```diff
@@ -4,9 +4,14 @@ import hosts from './hosts';
...
+    hosts,
+    notifications,
+    toasts,
+    ...asyncReducers,
+  });
+}
+
+export default combineReducersAsync();
```
sharvit, line 17: Does it work? Shouldn't it be:
```js
export default combineReducersAsync({});
# or
export function combineReducersAsync(asyncReducers = {}) { ... }
```

webpack/assets/javascripts/react_app/redux/reducers/index.js
sharvit: Maybe
```js

const staticReducers = { ... };

export const combineReducersAsync = (asyncReducers = {}) =>
    combineReducers({ ...staticReducers, ...asyncReducers });

export default combineReducers(staticReducers);
```

webpack/assets/javascripts/react_app/redux/reducers/index.js
danseethaler: Yup - it is a little surprising but the spread operator doesn't mind if the value it's spreading is undefined. It will just skip it.

webpack/assets/javascripts/react_app/redux/reducers/index.js
danseethaler: I actually like the idea of having a single way that all reducers are registered and taking out the work `async` out everywhere and just call it `addReducer`. That would make it trivial to add a `removeReducer` later on if we ever wanted it.

danseethaler: I'll refactor a bit and submit an updated version.

danseethaler: Alright I worked through the code some more and I think we should stick with what we have. Although it would be nice to have everything using the same api, it just makes more sense in the code to have static reducers registered, well... statically.
If all reducers were registered via the `addReducer` method we would still need to import all the static reducers but then we'd need to call the `addReducer` method on each one which is not as declarative as what we have now.

@sharvit mentioned moving the static reducers into their own file. I think what we have is a really common practice and the `asyncReducers` variable name implies that the others are static.

Let me know if there's any difference of opinion there.

ohadlevy (approved):

ohadlevy:
thanks @danseethaler !

sharvit:
You kept both files `addReducer.js` and `registerReducer.js`

danseethaler:
@sharvit good catch! It's pretty much orphaned - goes well with the empty `utils.js` file in the same directory :)

I must have added that in at some point and forgot to remove it.

danseethaler:
I guess the remedy is create an issue and a quick PR? Do we want that `utils.js` file in there?

PR #5015, opened 2017-11-21
Fixes #21718 - Show errors properly when creating new oVirt CR

adamruzicka:
We try to retrieve available operating systems in an after_validation
hook. Trying to retrieve them makes no sense if the URL isn't
considered valid by us.

### DIFF
--- a/app/models/compute_resources/foreman/model/ovirt.rb
+++ b/app/models/compute_resources/foreman/model/ovirt.rb
@@ -397,6 +397,7 @@ module Foreman::Model
     private
 
     def update_available_operating_systems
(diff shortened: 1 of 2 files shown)


iNecas:
Makes sense, would you mind adding some check to https://github.com/theforeman/foreman/blob/edd5310fc97828d306a06ecca861543def61cac8/test/unit/foreman/model/ovirt_test.rb#L24 to cover this?

adamruzicka:
- modified the test

iNecas (approved):

iNecas:
Thanks @adamruzicka

PR #5018, opened 2017-11-21
Fixes #21725 - Move test config to setup file

danseethaler:
This moves the redundant test setup code to a single file. Thanks @waldenraines for pointing this out!

http://projects.theforeman.org/issues/21725

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,7 +129,8 @@
       "script"
     ],
     "setupFiles": [
-      "raf/polyfill"
+      "raf/polyfill",
+      "./webpack/test_setup.js"
     ]
   }
 }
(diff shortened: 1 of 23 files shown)


ohadlevy (approved):

package.json
```diff
@@ -129,7 +129,8 @@
       "script"
     ],
     "setupFiles": [
-      "raf/polyfill"
+      "raf/polyfill",
+      "./webpack/test_setup.js"
```
ohadlevy, line 133: would it make more sense to call it something like ```.jest_setup.js``` or ```.jest_helper.js```?

danseethaler: I think it makes sense as is since it could relate to another test framework if we ever moved away from jest. It's also under the `jest` property in `package.json`.

Thoughts?

ohadlevy: :+1: great idea ;)

tbrisker (approved):
DRY 🥇 
Good to merge in my opinion, just not sure what you decided on the name of the file?

danseethaler:
@tbrisker @ohadlevy I think the consensus is to leave the file name as is since it's framework agnostic.

PR #5021, opened 2017-11-22
Fixes #20929 - Added fact name filtering on import

ShimShtein:
Added a setting that will filter out fact names, so they won't be recorded into `fact_names` and `fact_values` tables at all.

### DIFF
--- a/app/models/setting.rb
+++ b/app/models/setting.rb
@@ -207,12 +207,14 @@ class Setting < ApplicationRecord
     end
   end
 
-  def self.regexp_expand_wildcard_string(string)
-    "\\A#{Regexp.escape(string).gsub('\*', '.*')}\\Z"
+  def self.regexp_expand_wildcard_string(string, options = {})
+    prefix = options[:prefix] || '\A'
+    suffix = options[:suffix] || '\Z'
+    prefix + Regexp.escape(string).gsub('\*', '.*') + suffix
   end
 
-  def self.convert_array_to_regexp(array)
-    Regexp.new(array.map {|string| regexp_expand_wildcard_string(string) }.join('|'))
+  def self.convert_array_to_regexp(array, regexp_options = {})
+    Regexp.new(array.map {|string| regexp_expand_wildcard_string(string, regexp_options) }.join('|'))
   end
 
   def has_readonly_value?
--- a/app/models/setting/provisioning.rb
+++ b/app/models/setting/provisioning.rb
@@ -58,7 +58,13 @@ class Setting::Provisioning < Setting
       self.set('use_shortname_for_vms', N_("Foreman will use the short hostname instead of the FQDN for creating new virtual machines"), false, N_('Use short name for VMs')),
       self.set('dns_conflict_timeout', N_("Timeout for DNS conflict validation (in seconds)"), 3, N_('DNS conflict timeout')),
       self.set('clean_up_failed_deployment', N_("Foreman will delete virtual machine if provisioning script ends with non zero exit code"), true, N_('Clean up failed deployment')),
-      self.set('name_generator_type', N_("Random gives unique names, MAC-based are longer but stable (and only works with bare-metal)"), 'Random-based', N_("Type of name generator"), nil, {:collection => Proc.new {NameGenerator::GENERATOR_TYPES} })
+      self.set('name_generator_type', N_("Random gives unique names, MAC-based are longer but stable (and only works with bare-metal)"), 'Random-based', N_("Type of name generator"), nil, {:collection => Proc.new {NameGenerator::GENERATOR_TYPES} }),
+      self.set(
+        'excluded_facts',
+        N_("Exclude pattern for all types of imported facts (rhsm, puppet e.t.c.). Those facts won't be stored in foreman's database. You can use * wildcard to match names with indexes e.g. macvtap*"),
+        default_excluded_facts,
+        N_('Exclude pattern for facts stored in foreman')
+      )
     ] + default_global_templates + default_local_boot_templates
   end
 
@@ -100,5 +106,9 @@ class Setting::Provisioning < Setting
       templates = Proc.new { Hash[ProvisioningTemplate.unscoped.of_kind(pxe_kind).map { |tmpl| [tmpl.name, tmpl.name] }] }
       yield pxe_kind, templates
     end
+  end
+
+  def self.default_excluded_facts(ignored_interfaces = IGNORED_INTERFACES)
+    ignored_interfaces
   end
 end
--- a/app/services/fact_cleaner.rb
+++ b/app/services/fact_cleaner.rb
@@ -2,11 +2,7 @@ class FactCleaner
   delegate :logger, :to => :Rails
 
   def clean!
-    to_delete = find_leaf_orphaned_facts
-    # we also delete all composes that either don't have any descendant or all of them are to be deleted
-    to_delete += find_compose_orphaned_facts(to_delete)
-    delete!(to_delete)
-    @deleted_count = to_delete.count
+    @deleted_count = delete_excluded_facts + delete_orphaned_facts
     self
   end
 
@@ -17,9 +13,17 @@ class FactCleaner
 
   private
 
+  def delete_orphaned_facts
+    to_delete = find_leaf_orphaned_facts
+    # we also delete all composes that either don't have any descendant or all of them are to be deleted
+    to_delete += find_compose_orphaned_facts(to_delete)
+    delete!(to_delete)
+    to_delete.uniq.count
+  end
+
   def find_leaf_orphaned_facts
     result = []
-    logger.info "Searching for leaf orphaned facts..."
+    logger.debug "Searching for leaf orphaned facts..."
     FactName.leaves.includes(:fact_values).find_in_batches do |batch|
       result += batch.select do |fact|
         fact.fact_values.empty?
@@ -29,7 +33,7 @@ class FactCleaner
   end
 
   def find_compose_orphaned_facts(found)
-    logger.info "Searching for compose orphaned facts..."
+    logger.debug "Searching for compose orphaned facts..."
     FactName.composes.find_in_batches do |batch|
       found += batch.select do |compose|
         compose.descendants.all? { |fact| found.include?(fact) }
@@ -39,11 +43,38 @@ class FactCleaner
   end
 
   def delete!(to_delete)
-    logger.info "#{to_delete.size} facts will be removed"
+    logger.debug "#{to_delete.size} facts will be removed"
     to_delete.each do |fact|
       logger.debug { "Removing fact #{fact}" }
       fact.destroy
     end
-    logger.info "Cleanup of facts finished!"
+    logger.debug "Cleanup of facts finished!"
+  end
+
+  def find_excluded_facts
+    logger.debug "Searching for facts that match excluded pattern..."
+    fact_name_ids = []
+
+    excluded_facts_regex = FactImporter.excluded_facts_regex
+
+    FactName.unscoped.find_in_batches(:batch_size => 100) do |names_batch|
+      names_batch.select! { |fact_name| fact_name.name.match(excluded_facts_regex) }
+      fact_name_ids += names_batch.map(&:id)
+    end
+
+    fact_name_ids.uniq
+  end
+
+  def delete_excluded_facts
+    fact_names = find_excluded_facts
+
+    # delete related fact values first
+    logger.debug "Removing values of excluded facts..."
+    deleted_values = FactValue.unscoped.where(:fact_name_id => fact_names).delete_all
+    logger.debug "Removed #{deleted_values} associated fact value records."
+    logger.debug "Removing excluded fact names..."
+    deleted_names = FactName.unscoped.where(:id => fact_names).delete_all
+    logger.debug "Removed #{deleted_names} fact name records."
+    deleted_names
   end
 end
--- a/app/services/fact_importer.rb
+++ b/app/services/fact_importer.rb
@@ -30,6 +30,16 @@ class FactImporter
     []
   end
 
+  def self.excluded_facts_regex
+    Setting.convert_array_to_regexp(
+      Setting[:excluded_facts],
+      {
+        :prefix => '(\A|.*::|.*_)',
+        :suffix => '(\Z|::.*)'
+      }
+    )
+  end
+
   def initialize(host, facts = {})
     @error    = false
     @host     = host
@@ -170,10 +180,14 @@ class FactImporter
   end
 
   def normalize(facts)
-    # convert all structures to simple strings
-    facts = Hash[facts.map {|k, v| [k.to_s, v.to_s]}]
-    # and remove empty values
-    facts.keep_if { |k, v| v.present? }
+    normalized_facts = {}
+    facts.each do |k, v|
+      key = k.to_s
+      val = v.to_s
+
+      normalized_facts[key] = val unless val.empty? || key.match(excluded_facts)
+    end
+    normalized_facts
   end
 
   def db_facts
@@ -186,5 +200,9 @@ class FactImporter
 
   def save_name_record(name_record)
     name_record.save!(:validate => false)
+  end
+
+  def excluded_facts
+    @excluded_facts ||= FactImporter.excluded_facts_regex
   end
 end
--- a/test/fact_importer_test_helper.rb
+++ b/test/fact_importer_test_helper.rb
@@ -34,3 +34,151 @@ module FactImporterFactoryStubber
     @instance_stubs ||= []
   end
 end
+
+module FactsData
+  class FlatFacts
+    def filter
+      ['ignore*', '*_bad', 'filter']
+    end
+
+    def good_facts
+      {
+        'just_a_fact' => 'hello',
+        'nofilter' => 'hello'
+      }
+    end
+
+    def ignored_facts
+      {
+        'ignored_fact' => 'will_not_show',
+        'fact_bad' => 'will_not_show',
+        'filter' => 'will_not_show',
+        'ignore' => 'will_not_show',
+        '_bad' => 'will_not_show'
+      }
+    end
+  end
+
+  class RhsmStyleFacts
+    def filter
+      ['ignore*', '*_bad', 'filter']
+    end
+
+    def good_facts
+      {
+        'something::something_else' => 'hello'
+      }
+    end
+
+    def ignored_facts
+      {
+        'ignored_fact::something' => 'will_not_show',
+        'something::ignored_fact' => 'will_not_show',
+        'something::ignored_fact::something_else' => 'will_not_show',
+        'fact_bad::something' => 'will_not_show',
+        'something::fact_bad' => 'will_not_show',
+        'something::fact_bad::something_else' => 'will_not_show',
+        'filter::something' => 'will_not_show',
+        'something::filter' => 'will_not_show',
+        'something::filter::something_else' => 'will_not_show'
+      }
+    end
+  end
+
+  class FlatPuppetStyleFacts
+    def filter
+      ['ignore*', '*_bad', 'filter']
+    end
+
+    def good_facts
+      {
+        'something_filter_something_else' => 'hello',
+        'fact_bad_something' => 'hello',
+        'filter_something' => 'hello'
+      }
+    end
+
+    def ignored_facts
+      {
+        'ignored_fact_something' => 'will_not_show',
+        'something_ignored_fact' => 'will_not_show',
+        'something_ignored_fact_something_else' => 'will_not_show',
+        'something_fact_bad' => 'will_not_show',
+        'something_filter' => 'will_not_show'
+      }
+    end
+  end
+
+  class HierarchicalPuppetStyleFacts
+    def filter
+      ['ignored*']
+    end
+
+    def good_facts
+      {
+        :good => 'hello',
+        :common_ancestor => {
+          :good_subtree => {
+            :good_key => 'hello'
+          }
+        }
+      }
+    end
+
+    def ignored_facts
+      {
+        :common_ancestor => {
+          :ignored_subtree => {
+            :key1 => 'will_not_show',
+            :key2 => 'will_not_show'
+          }
+        },
+        :empty_ancestor => {
+          :ignored_key => 'will_not_show'
+        },
+        :ignored_value => 'will_not_show'
+      }
+    end
+
+    def flat_result
+      {
+        "good"=>"hello",
+        "common_ancestor::good_subtree::good_key"=>"hello",
+        "common_ancestor::good_subtree"=>nil,
+        "common_ancestor"=>nil
+      }
+    end
+  end
+
+  class DefaultInterfacesFacts
+    def filter
+      Setting[:excluded_facts]
+    end
+
+    def good_facts
+      {
+        'macaddress_virbr4' => 'fe:54:00:59:4e:bf',
+        'mtu_virbr4' => '1500',
+        'net::interface::virbr4::ipv6_address::link' => 'fe80::fc54:ff:fe59:4ebf',
+        'net::interface::virbr4::ipv6_address::link_list' => 'fe80::fc54:ff:fe59:4ebf',
+        'net::interface::virbr4::ipv6_netmask::link' => '64',
+        'net::interface::virbr4::ipv6_netmask::link_list' => '64',
+        'net::interface::virbr4::mac_address' => 'FE:54:00:59:4E:BF',
+        'net::interface::virbr4::permanent_mac_address' => 'Unknown'
+      }
+    end
+
+    def ignored_facts
+      {
+        'macaddress_vnet4' => 'fe:54:00:59:4e:bf',
+        'mtu_vnet4' => '1500',
+        'net::interface::vnet4::ipv6_address::link' => 'fe80::fc54:ff:fe59:4ebf',
+        'net::interface::macvtap1::ipv6_address::link_list' => 'fe80::fc54:ff:fe59:4ebf',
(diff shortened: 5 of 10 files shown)


app/models/setting/provisioning.rb
```diff
@@ -47,6 +47,7 @@ def self.default_settings
       self.set('manage_puppetca', N_("Foreman will automate certificate signing upon provision of new host"), true, N_('Manage PuppetCA')),
       self.set('ignore_puppet_facts_for_provisioning', N_("Stop updating IP address and MAC values from Puppet facts (affects all interfaces)"), false, N_('Ignore Puppet facts for provisioning')),
       self.set('ignored_interface_identifiers', N_("Ignore interfaces that match these values during facts importing, you can use * wildcard to match names with indexes e.g. macvtap*"), IGNORED_INTERFACES, N_('Ignore interfaces with matching identifier')),
+      self.set('ignored_fact_names', N_("Ignore fact names that match these values during facts importing, you can use * wildcard to match names with indexes e.g. macvtap*"), IGNORED_INTERFACES, N_('Ignore fact names with matching identifier')),
```
lzap: Can you tell the user this includes puppet, rhsm and all facts in general?

app/services/fact_importer.rb
```diff
@@ -116,13 +116,18 @@ def update_facts
...
   def normalize(facts)
+    fact_pairs = facts.map {|k, v| [k.to_s, v.to_s]}
+    # filter out matching fact names and empty values
+    fact_pairs.reject! {|k, v| v.empty? || k.match(ignored_fact_names)}
     # convert all structures to simple strings
-    facts = Hash[facts.map {|k, v| [k.to_s, v.to_s]}]
-    # and remove empty values
-    facts.keep_if { |k, v| v.present? }
+    Hash[fact_pairs]
   end
```
lzap: I think it is a good opportunity to clean this method - this looks really ugly. Create pairs array, reject it, then create hash from it. I think creating copy of hash and then deleting filtered entries should be faster (leaving some buckets unused but faster).

    facts.select { |key, value| key.to_s.match(ignored_fact_names) }

This should do it, not tested tho.

db/migrate/20171122092336_add_ignored_fact_names_setting.rb
```diff
@@ -0,0 +1,14 @@
...
+        'ignored_fact_names',
+        N_("Ignore fact names that match these values during facts importing, you can use * wildcard to match names with indexes e.g. macvtap*"),
+        Setting[:ignored_interface_identifiers],
+        N_('Ignore fact names with matching identifier'))) if Setting[:ignored_interface_identifiers]
+  end
+
+  def down
+    Setting.where(:name => :ignored_fact_names).destroy_all
+  end
+end
```
lzap: One thing - this is a great opportunity to clean those facts from db. There are two choices:

A) Implicit - simply delete them in the migration. Can be dangerous for existing users.

B) Explicit - create a rake task and we can ask customer to run it (first show the facts, then confirm deletion). This will be safe enough for everybody.

It would be very nice to have the same rake task for NICs.

lzap (changes requested):
We need this patch so hard, thank you so much :-)

app/models/setting/provisioning.rb
ares: The default value and second part of description seems wrong :-)

lzap: Well it is correct, but I agree that the constant should be renamed in this case.

ares: I didn't know there fact with names like macvtap*, perhaps fact names from newer facter. In such case, that's fine and just the constant should be named differently, it could be set to IGNORED_INTERFACES values though.

ShimShtein:
Rebased, renamed, refactored and extended already existing `facts:clean` rake task

lzap:
Marek pointed out on one important issue - fact names are different from interface names. 

```
interfaces => enp30s0,lo,redhat0,virbr0,virbr1,virbr2,virbr0_nic,virbr1_nic,virbr2_nic
ipaddress6_enp30s0 => fd5a:f687:b576:0:xxxx:f6ff:fe3e:aa52
ipaddress_enp30s0 => 192.168.x.13
macaddress_enp30s0 => 88:d7:f6:xx:xx:52
mtu_enp30s0 => 1500
netmask_enp30s0 => 255.255.255.0
network_enp30s0 => 192.168.1.0
```
I think we should be more explicit to prevent filtering out unwanted entries, therefore I suggest to create separate starting list. How about this:

```ruby
 IGNORED_FACTS =  IGNORED_INTERFACES.map{|i| "_#{i}"}
```

This will at least prefix all entries with underscore, which will match what we want.

test/unit/fact_cleaner_test.rb
```diff
@@ -48,6 +48,19 @@ class FactCleanerTest < ActiveSupport::TestCase
     assert_include remaining_of(root_6_fact), root_6_fact
   end
 
+  test "it cleans excluded facts" do
+    cleaner.stubs(:excluded_facts).returns(['ignored*', '*bad'])
+    cleaner.stubs(:delete_orphaned_facts).returns(0)
```
lzap: This method (`delete_orphaned_facts`) needs a test.

app/models/setting/provisioning.rb
```diff
@@ -48,6 +48,7 @@ def self.default_settings
       self.set('ignore_puppet_facts_for_provisioning', N_("Stop updating IP address and MAC values from Puppet facts (affects all interfaces)"), false, N_('Ignore Puppet facts for provisioning')),
       self.set('enable_orchestration_on_fact_import', N_("Enable host orchestration on puppet fact import. This could cause serious performance issues as the number of hosts increase"), false, N_('Enable orchestration on puppet fact import')),
       self.set('ignored_interface_identifiers', N_("Ignore interfaces that match these values during facts importing, you can use * wildcard to match names with indexes e.g. macvtap*"), IGNORED_INTERFACES, N_('Ignore interfaces with matching identifier')),
+      self.set('excluded_facts', N_("Exclude pattern for all types of imported facts (rhsm, puppet e.t.c.). Those facts won't be stored in foreman's database. You can use * wildcard to match names with indexes e.g. macvtap*"), IGNORED_INTERFACES, N_('Exclude pattern for facts stored in foreman')),
```
lzap: Consider splitting this into separate default list with underscores (see my other comment).

app/services/fact_cleaner.rb
```diff
@@ -2,11 +2,8 @@ class FactCleaner
   delegate :logger, :to => :Rails
 
   def clean!
-    to_delete = find_leaf_orphaned_facts
-    # we also delete all composes that either don't have any descendant or all of them are to be deleted
-    to_delete += find_compose_orphaned_facts(to_delete)
-    delete!(to_delete)
-    @deleted_count = to_delete.count
+    @deleted_count = delete_orphaned_facts
+    @deleted_count += delete_excluded_facts
```
lzap: This could be single line.

app/services/fact_cleaner.rb
```diff
@@ -46,4 +51,29 @@ def delete!(to_delete)
     end
     logger.info "Cleanup of facts finished!"
   end
+
+  def find_excluded_facts
+    logger.info "Searching for facts that match excluded pattern..."
```
lzap: Please `debug`. Frankly, I don't think this is much useful debug line anyway.

app/services/fact_cleaner.rb
```diff
@@ -46,4 +51,29 @@ def delete!(to_delete)
...
+  def delete_excluded_facts
+    fact_names = find_excluded_facts
+
+    # delete related fact values first
+    logger.info "Removing values of excluded facts..."
+    FactValue.unscoped.where(:fact_name_id => fact_names).delete_all
+    logger.info "Removing excluded fact names..."
+    FactName.unscoped.where(:id => fact_names).delete_all
+    fact_names.count
+  end
```
lzap: This should be all `debug` lines.

db/migrate/20171122092336_add_ignored_fact_names_setting.rb
```diff
@@ -0,0 +1,14 @@
...
+        'excluded_facts',
+        N_("Ignore fact names that match these values during facts importing, you can use * wildcard to match names with indexes e.g. macvtap*"),
+        Setting[:ignored_interface_identifiers],
+        N_('Ignore fact names with matching identifier'))) if Setting[:ignored_interface_identifiers]
+  end
+
+  def down
+    Setting.where(:name => :excluded_facts).destroy_all
+  end
+end
```
lzap: Is migration necessary? I thought settings are automatically created during boot time.

lzap (changes requested):
Still have some comments.

test/unit/fact_cleaner_test.rb
ShimShtein: `delete_orphaned_facts` was extracted out of `clean!`. All other tests are actually testing it.

app/services/fact_cleaner.rb
```diff
@@ -39,11 +43,42 @@ def find_compose_orphaned_facts(found)
...
+      sql_pattern = exclude_pattern.gsub('*', '%')
+      # for facts that look like ignore_me::something
+      fact_name_ids += FactName.unscoped.where('name LIKE ?', sql_pattern).pluck(:id)
+      fact_name_ids += FactName.unscoped.where('name LIKE ?', sql_pattern + '::%').pluck(:id)
+      # for facts that look like common::ignore_me::something
+      fact_name_ids += FactName.unscoped.where('name LIKE ?', '%::' + sql_pattern).pluck(:id)
+      fact_name_ids += FactName.unscoped.where('name LIKE ?', '%::' + sql_pattern + '::%').pluck(:id)
+    end
+    fact_name_ids
+  end
```
lzap: Is this method new in the last commit? I haven't seen it. Anyway, this will generate *huge* SQL query and what's the worst - the server will need to run it for each individual fact upload. The cost for this consistency is way too high, I suggest to move this into rake task that can be run from cron or something once a week. This has potential to slow down fact import, we also haven't tested performance of this on MySQL.

app/services/fact_cleaner.rb
```diff
@@ -39,11 +43,42 @@ def find_compose_orphaned_facts(found)
...
-    logger.info "Cleanup of facts finished!"
+    logger.debug "Cleanup of facts finished!"
+  end
+
+  def find_excluded_facts
+    logger.debug "Searching for facts that match excluded pattern..."
+    fact_name_ids = []
+
+    excluded_facts.each do |exclude_pattern|
+      sql_pattern = exclude_pattern.gsub('*', '%')
```
lzap: Use faster `tr` instead of `gsub`. Also, this is vulnerable to SQL injection.

app/services/fact_cleaner.rb
```diff
@@ -39,11 +43,42 @@ def find_compose_orphaned_facts(found)
...
+
+  def delete_excluded_facts
+    fact_names = find_excluded_facts
+
+    # delete related fact values first
+    logger.debug "Removing values of excluded facts..."
+    FactValue.unscoped.where(:fact_name_id => fact_names).delete_all
+    logger.debug "Removing excluded fact names..."
+    FactName.unscoped.where(:id => fact_names).delete_all
+    fact_names.count
```
lzap: Seriously, we should not be doing regular cleaning for every request, but pull this to external process. This is way too many operations.

app/models/setting/provisioning.rb
```diff
@@ -101,4 +107,9 @@ def self.map_pxe_kind
       yield pxe_kind, templates
     end
   end
+
+  def self.default_excluded_facts(ignored_interfaces = IGNORED_INTERFACES)
+    # By default ignore facts from both forms: "a::b::macvtap01::d::e" and "a => b => macvtap01 => ..."
+    ignored_interfaces.flat_map { |i| ["*::#{i}", i] }
+  end
```
lzap: I don't understand, the list is enriched with `*::` here, and then in `find_excluded_facts` method, so is it added twice? I prefer just to have it in one place, the latter seems to be a better fit for this I think.

test/unit/fact_cleaner_test.rb
```diff
@@ -48,6 +48,48 @@ class FactCleanerTest < ActiveSupport::TestCase
...
+    good_facts = [
+      FactoryBot.create(:fact_name, :name => 'not_ignored'),
+      FactoryBot.create(:fact_name, :name => 'not_ignored::something'),
+      FactoryBot.create(:fact_name, :name => 'common::not_ignored'),
+      FactoryBot.create(:fact_name, :name => 'common::not_ignored::something'),
+      FactoryBot.create(:fact_name, :name => 'ignored_shouldnt'),
+      FactoryBot.create(:fact_name, :name => 'ignored_shouldnt::something'),
+      FactoryBot.create(:fact_name, :name => 'common::ignored_shouldnt'),
+      FactoryBot.create(:fact_name, :name => 'common::ignored_shouldnt::something')
+    ]
```
lzap: I must insist on real world facts, because (I haven't tested but) I believe your code does not work for real facts. Please include the following real facts I just captured from my server:

```
interfaces => em1,lo
ipaddress6_em1 => 2620:52:0:800:2247:ffff:ffff:95a0
ipaddress_em1 => 10.8.0.1
macaddress_em1 => 20:47:47:89:95:bb
mtu_em1 => 1500
netmask_em1 => 255.255.224.0
network_em1 => 10.8.0.0
dhcp_servers::em1 => 10.10.10.11
```

Again, I haven't tested myself, but I think your mathing code won't match most of those.

db/migrate/20171122092336_add_ignored_fact_names_setting.rb
lzap: What is the purpose of the migration?

lzap (dismissed):
I have two big concerns. First, the deletion must not happen during fact upload, but via an external rake task (cron). This way, we can even make this more simple and simply iterate over all fact names via Ruby loop and filter them out using single regexp, you are trying to work in two "modes" (SQL and regexp) which makes the code too complex.

Second, you are not testing with real sample data. I think the code either does not match real world samples, or filters out what's not required. I think the pattern should be with underscore to match `macaddress_em1` (pattern `*_em1` or `dhcp-server::em1` (pattern `::em1`).

app/services/fact_cleaner.rb
lzap: I realized this method is only called from clean! method, so that is fine.

app/services/fact_cleaner.rb
lzap: Correction: No vulnerability :-)

app/services/fact_cleaner.rb
lzap: Disregard this comment.

app/models/setting.rb
```diff
@@ -207,12 +207,14 @@ def self.create!(opts)
     end
   end
 
-  def self.regexp_expand_wildcard_string(string)
-    "\\A#{Regexp.escape(string).gsub('\*', '.*')}\\Z"
+  def self.regexp_expand_wildcard_string(string, options = {})
+    prefix = options[:prefix] || '\A'
+    suffix = options[:suffix] || '\Z'
+    "#{prefix}#{Regexp.escape(string).gsub('\*', '.*')}#{suffix}"
```
lzap: Nitpick: This could be simple string concat.

app/services/fact_cleaner.rb
```diff
@@ -17,9 +13,17 @@ def deleted_count
 
   private
 
+  def delete_orphaned_facts
+    to_delete = find_leaf_orphaned_facts
+    # we also delete all composes that either don't have any descendant or all of them are to be deleted
+    to_delete += find_compose_orphaned_facts(to_delete)
+    delete!(to_delete)
+    to_delete.count
+  end
```
lzap: This should be:

    delete!(find_compose_orphaned_facts(find_leaf_orphaned_facts))

as the method returns number of rows deleted.

app/services/fact_cleaner.rb
```diff
@@ -39,11 +43,44 @@ def find_compose_orphaned_facts(found)
...
-    logger.info "Cleanup of facts finished!"
+    logger.debug "Cleanup of facts finished!"
+  end
+
+  def find_excluded_facts
+    logger.debug "Searching for facts that match excluded pattern..."
+    fact_name_ids = []
+
+    excluded_facts.each do |exclude_pattern|
+      sql_pattern = exclude_pattern.gsub('*', '%')
```
lzap: Method `tr` is still much faster.

app/services/fact_cleaner.rb
```diff
@@ -39,11 +43,44 @@ def find_compose_orphaned_facts(found)
...
+
+  def delete_excluded_facts
+    fact_names = find_excluded_facts
+
+    # delete related fact values first
+    logger.debug "Removing values of excluded facts..."
+    FactValue.unscoped.where(:fact_name_id => fact_names).delete_all
+    logger.debug "Removing excluded fact names..."
+    FactName.unscoped.where(:id => fact_names).delete_all
+    fact_names.count
```
lzap: Would be better to actually return number of deleted rows as it gets logged. Best thing would be to add debug lines for both number of rows planned to delete and number which actually was deleted.

test/unit/fact_cleaner_test.rb
```diff
@@ -48,6 +48,49 @@ class FactCleanerTest < ActiveSupport::TestCase
...
   end
 
+  test "it cleans excluded facts" do
+    cleaner.stubs(:excluded_facts).returns(['ignored*', '*bad'])
+    cleaner.stubs(:delete_orphaned_facts).returns(0)
+    good_fact_name = FactoryBot.create(:fact_name, :name => 'good_fact')
+    ignored_fact_name = FactoryBot.create(:fact_name, :name => 'ignored01')
+    bad_fact_name = FactoryBot.create(:fact_name, :name => 'empty_bad')
+    FactoryBot.create(:fact_value, :fact_name => good_fact_name)
+    FactoryBot.create(:fact_value, :fact_name => ignored_fact_name)
```
lzap, line 63: I still do not see testing of "macaddress_em1" kind of fact. Frankly, I was hoping for real-world interface names rather than these artificial ones :-)

test/unit/fact_importer_test.rb
```diff
@@ -55,6 +55,73 @@ def fact_name_class
...
+        'something::ignored_fact' => 'will_not_show',
+        'something::ignored_fact::something_else' => 'will_not_show',
+        'fact_bad::something' => 'will_not_show',
+        'something::fact_bad' => 'will_not_show',
+        'something::fact_bad::something_else' => 'will_not_show',
+        'filter::something' => 'will_not_show',
+        'something::filter' => 'will_not_show',
+        'something::filter::something_else' => 'will_not_show',
+        'something::something_else' => 'hello'
+      }
```
lzap: Can you also add edge cases: "ignored_fact" and "fact_bad" (without any `::` characters). I am not sure if RHSM generates it, but it's good to be safe.

test/unit/fact_importer_test.rb
```diff
@@ -55,6 +55,73 @@ def fact_name_class
...
+
+      importer = FactImporter.new(nil, facts)
+      actual_facts = importer.send(:facts)
+
+      assert_not_nil actual_facts['something::something_else']
+      assert_equal 1, actual_facts.count
```
lzap: Can you also assert that all passed facts have "hello" in its value? Just to be sure it filtered it correctly :-) In all tests please.

test/unit/fact_importer_test.rb
```diff
@@ -55,6 +55,73 @@ def fact_name_class
...
+        # facter < 3.0 style facts: macaddress_virbr0
+        'ignored_fact_something' => 'will_not_show',
+        'something_ignored_fact' => 'will_not_show',
+        'something_ignored_fact_something_else' => 'will_not_show',
+        'something_fact_bad' => 'will_not_show',
+        'something_filter' => 'will_not_show',
+        'something_filter_something_else' => 'hello',
+        'fact_bad_something' => 'hello',
+        'filter_something' => 'hello'
+      }
```
lzap: Also edge cases: "ignore" and "_bad" and "filter" names should all be filtered. Although it might not make sense, users could try to filter some facts with this.

test/unit/structured_fact_importer_test.rb
```diff
@@ -78,6 +78,40 @@ class StructuredFactImporterTest < ActiveSupport::TestCase
...
+          },
+          :good_subtree => {
+            :good_key => 'good_value'
+          }
+        },
+        :empty_ancestor => {
+          :ignored_key => 'some_val'
+        },
+        :ignored_value => 'some_val'
+      }
```
lzap: Honestly, I'd appreciate common testing value style across all those tests - either "yes" or "no" or whatever values so these can be clearly identified :-)

I think it's very important to make sure both cleaning task and filtering here works the same - can you move the fact testing data (all three types) into some helper and reuse it in both tests? This way we can be sure if someone makes change in the future, failing test will catch the other workflow.

lzap (changes requested):
Overall no blockers but I don't want to rush this. Can you improve tests, I am still missing one and I have a proposal to share testing data across both two workflows to prevent regressions.

ShimShtein:
Finally! The tests are green. Had to give up cleaner's DB code, there is no way to escape `LIKE` query so it will work on all databases.

app/services/fact_cleaner.rb
```diff
@@ -39,11 +43,38 @@ def find_compose_orphaned_facts(found)
...
     end
-    logger.info "Cleanup of facts finished!"
+    logger.debug "Cleanup of facts finished!"
+  end
+
+  def find_excluded_facts
+    logger.debug "Searching for facts that match excluded pattern..."
+    fact_name_ids = []
+
+    excluded_facts_regex = FactImporter.new(nil, {}).send(:excluded_facts)
```
lzap: Uh. Oh.

app/services/fact_cleaner.rb
```diff
@@ -39,11 +43,38 @@ def find_compose_orphaned_facts(found)
...
+    fact_names = find_excluded_facts
+
+    # delete related fact values first
+    logger.debug "Removing values of excluded facts..."
+    deleted_values = FactValue.unscoped.where(:fact_name_id => fact_names).delete_all
+    logger.debug "Removed #{deleted_values} associated fact value records."
+    logger.debug "Removing excluded fact names..."
+    deleted_names = FactName.unscoped.where(:id => fact_names).delete_all
+    logger.debug "Removed #{deleted_names} fact name records."
+    deleted_names
```
lzap, line 78: I am usually against debug messages like "I am doing something..." the "Removed" lines are enough to tell you it has been processed.

app/services/fact_importer.rb
```diff
@@ -187,4 +191,14 @@ def ensure_no_active_transaction
...
+
+  def excluded_facts
+    @excluded_facts ||= Setting.convert_array_to_regexp(
+      Setting[:excluded_facts],
+      {
+        :prefix => '(\A|.*::|.*_)',
+        :suffix => '(\Z|::.*)'
+      }
+    )
+  end
```
lzap, line 207: Okay, this can be moved to settings, why not? The other line with send really hurts :-)

test/fact_importer_test_helper.rb
```diff
@@ -34,3 +34,119 @@ def instance_stubs
...
+        'something::ignored_fact::something_else' => 'will_not_show',
+        'fact_bad::something' => 'will_not_show',
+        'something::fact_bad' => 'will_not_show',
+        'something::fact_bad::something_else' => 'will_not_show',
+        'filter::something' => 'will_not_show',
+        'something::filter' => 'will_not_show',
+        'something::filter::something_else' => 'will_not_show'
+      }
+    end
+  end
```
lzap, line 86: One could say this belongs to Katello, but I really think it should be in core. Thanks.

app/models/setting/provisioning.rb
```diff
@@ -101,4 +107,8 @@ def self.map_pxe_kind
       yield pxe_kind, templates
     end
   end
+
+  def self.default_excluded_facts(ignored_interfaces = IGNORED_INTERFACES)
+    ignored_interfaces
+  end
```
lzap, line 113: Can you optionally add couple of tests to test IGNORED_INTERFACES constant actually? We are not sure we ship with default value that just works. A single test with different names "Ipaddress_vnet1" should do it "Ipaddress:vnet1".

lzap (changes requested):
One ugly line and few nitpicks.

app/services/fact_cleaner.rb
ShimShtein: It really helps to identify slow queries - you will see the "starting.." message, but not the "finished" one.

app/services/fact_importer.rb
ShimShtein: Since we don't have a special settings class, I prefer keeping it with the importer. Refactored it to a class method though.

test/fact_importer_test_helper.rb
ShimShtein: Katello is only an example of this kind of facts. Since this notation is supported by core, I think it should be tested in core.

app/services/fact_cleaner.rb
lzap: Fair enough, although AR time is reported for each request by Rails framework and then there is SQL server tracking, but I buy that. We miss TRACE level so much.

app/services/fact_importer.rb
lzap: As long as the send method is gone, okay :-)

lzap (approved):
Thanks, merging.

lzap:
Katello test failures seem irrelevant.

PR #5024, opened 2017-11-22
Fixes #21737 - move about page to react

amirfefer:
About page uses 3 different old js files in order to check the status of compute resources and smart proxies.
I've extracted this status rendering to a react component `<About>`, in addition, I've united the UI style (by using Icons instead of labels).

### new components from **pf-react**:
- [x]  Empty State
- [x]  Tabs
- [x]  Table

###  ~~Todo~~
- [x]  tests
### updated gif (March) :
After

### before:
nov 22 2017 6-10 pm


app/views/about/index.html.erb
```diff
@@ -16,105 +15,28 @@
...
-                  <td><%=h proxy.features.map(&:name).to_sentence %></td>
-                  <td><div class="proxy-show-status"><%= spinner %></div></td>
-                  <td><div class="proxy-version"><%= spinner %></div></td>
-                </tr>
-              <% end %>
-              </tbody>
-            </table>
-          <% end %>
+          <%= data = @smart_proxies.map {|proxy| {:id => {:name => proxy.name, :id => proxy.id},
+                                                  :features => h(proxy.features.map(&:name).to_sentence)}} %>
```
ohadlevy: ```h``` is probably no longer required (it's the default).

app/views/about/index.html.erb
```diff
@@ -16,105 +15,28 @@
...
-                  <tr>
-                    <td class="ellipsis"><%= link_to(compute.name, compute) %></td>
-                    <td><%= compute.provider_friendly_name %></td>
-                    <td><div class="compute-status" data-url=<%= ping_compute_resource_path(compute) %>><%= spinner %></div></td>
-                  </tr>
-                <% end %>
-                </tbody>
-              </table>
-            <% end %>
+            <%= data = @compute_resources.map {|compute| {:id => {:name => compute.name, :id => compute.id},
```
ohadlevy: when there are none, you get a generic message ````No data found``` - we should use the PF empty state instead.

config/routes.rb
```diff
@@ -404,7 +404,7 @@
           post 'template_selected'
           post 'cluster_selected'
           get 'resource_pools'
-          post 'ping'
+          get 'ping'
```
ohadlevy: doesnt this break CR test connection?

webpack/assets/javascripts/react_app/common/helpers.js
```diff
@@ -6,4 +6,12 @@ export default {
     });
   },
   noop: Function.prototype, // empty function
+  urlBuilder(controller, action, id = undefined) {
+    return `/${controller}/${id ? `${id}/` : ''}${action}`;
+  },
+  getQueryParam(query) {
+    const urlParams = new URLSearchParams(window.location.search);
```
ohadlevy: how is this different from the urijs lib that we currently use? - also does this work across browsers?

webpack/assets/javascripts/react_app/components/about/compute/index.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import Status from '../../common/status';
+import Table from '../../common/table';
+import helpers from '../../../common/helpers';
+
+class AboutComputeTable extends React.Component {
+  render() {
+    const columns = [
+      {
+        Header: 'Name',
```
ohadlevy: please extract all strings

webpack/assets/javascripts/react_app/components/about/compute/index.js
```diff
@@ -0,0 +1,45 @@
...
+        accessor: 'id',
+        Cell: props => (
+          <a href={helpers.urlBuilder('compute_resources', '', props.value.id)}>
```
ohadlevy: you can import urlBruilder directly, e.g.
```js
import { urlBuilder } from '../helpers';
```

webpack/assets/javascripts/react_app/components/about/provider/index.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import Table from '../../common/table';
+import Icon from '../../common/Icon';
```
ohadlevy: PF-react now has icon component - maybe we should consider using that instead and remove our own?

webpack/assets/javascripts/react_app/components/common/table/index.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import ReactTable from 'react-table';
+import 'react-table/react-table.css';
+import helpers from '../../../common/helpers';
+
+const SETTING_PER_PAGE = 10; // TODO: use foreman per_page settings
```
ohadlevy: current UI default is 20

webpack/assets/javascripts/react_app/components/common/table/index.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import ReactTable from 'react-table';
+import 'react-table/react-table.css';
```
ohadlevy: table style doesn't look like patternfly...

webpack/assets/javascripts/react_app/components/common/table/index.js
```diff
@@ -0,0 +1,54 @@
...
+  render() {
+    const { columns, data, showPagination } = this.props;
+    const rows = this.isEmpty() ? BLANK_ROWS : data.length;
+
+    return (
+      <ReactTable
+        className="-striped"
+        defaultPageSize={
+          showPagination
+            ? Number(helpers.getQueryParam('per_page')) || Math.min(rows, SETTING_PER_PAGE)
```
ohadlevy: why deal with pagination in this PR if there is no code that uses it yet?

webpack/assets/javascripts/react_app/components/common/table/index.js
```diff
@@ -0,0 +1,54 @@
...
+            : rows
+        }
+        showPagination={showPagination}
+        data={data}
+        columns={columns}
+        // i18n
+        previousText={__('Previous')}
```
ohadlevy: if you are adding pagination support - it should match pf

webpack/assets/javascripts/react_app/components/common/table/index.js
```diff
@@ -0,0 +1,54 @@
...
+        nextText={__('Next')}
+        pageText={__('page')}
+        ofText={__('of')}
+        rowsText={__('rows')}
+        noDataText={__('No data found')}
```
ohadlevy: should probably use empty state optionally (in case the table is the main part of the view).

webpack/assets/javascripts/react_app/components/common/table/index.js
```diff
@@ -0,0 +1,54 @@
...
+        getTdProps={(state, rowInfo, column) => {
+          const rowData = rowInfo ? rowInfo.original[column.id] : undefined;
+
+          return {
+            title:
+              typeof rowData === 'string' && rowData.length > TOOLTIP_SIZE
+                ? rowData
+                : undefined,
+            style: {
+              textAlign: 'center',
```
ohadlevy: why pass style here?

webpack/assets/javascripts/react_app/components/componentRegistry.js
```diff
@@ -8,6 +8,10 @@ import PowerStatus from './hosts/powerStatus/';
 import NotificationContainer from './notifications/';
 import ToastsList from './toastNotifications/';
 import StorageContainer from './hosts/storage/vmware/';
+import AboutComputeTable from './about/compute';
```
ohadlevy: TBH: I am not too happy about having so many entry level components, is there any chance we can take over multiple tabs in one hoc ? this will allow to have all of the "logic" in one place vs across multiple components

webpack/assets/javascripts/react_app/redux/actions/status/index.js
```diff
@@ -0,0 +1,12 @@
+import { STATUS_REQUEST, STATUS_SUCCESS, STATUS_FAILURE } from '../../consts';
```
ohadlevy, line 1: shouldnt these const be more specific? e.g. HOST_STATUS or COMPUTE_RESOURCE_STATUS?

webpack/assets/javascripts/react_app/redux/reducers/status/index.js
```diff
@@ -0,0 +1,19 @@
...
+import { STATUS_REQUEST, STATUS_SUCCESS, STATUS_FAILURE } from '../../consts';
+
+const initialState = Immutable({});
+
+export default (state = initialState, action) => {
+  const { payload } = action;
+
+  switch (action.type) {
+    case STATUS_REQUEST:
+      return state;
```
ohadlevy: why? shouldn't this set the state so that the request has started?

webpack/assets/javascripts/react_app/redux/reducers/status/index.js
```diff
@@ -0,0 +1,19 @@
...
+    case STATUS_SUCCESS:
+      return state.set(payload.id, { ...payload });
```
ohadlevy: I'm a bit worried that ID's can override each other, while I've seen you've prefixed it, the fact is that these actions are used for multiple objects, can confuse a future developer?

ohadlevy (changes requested):
nice @amirfefer - first review round..

ohadlevy:
@amirfefer also note the lack of tests + current failing tests.

sharvit:
@amirfefer It's is great to see more and more code moving to react.
I would love to see the whole `about/index.html.erb` moving to react, including the tabs and the layout.
Then, you will need to add only 1 component to the registry (instead of 4).

You can use `react-bootstrap` for the tabs:
https://react-bootstrap.github.io/components.html#tabs

The row and col can be replaced with:
https://react-bootstrap.github.io/components.html#grid

Those components will be apart of `patternfly-react` in the future so it will be easy to move them later, probably by only change the import statement.

app/views/about/index.html.erb
```diff
@@ -16,105 +15,28 @@
...
-                  <th><%= _("Features") %></th>
-                  <th><%= _("Status") %></th>
-                  <th><%= _("Version") %></th>
-                </tr>
-              </thead>
-              <tbody>
-              <% @smart_proxies.each do |proxy| %>
-                <tr class="proxy-show"  data-url="<%= ping_smart_proxy_path(proxy) %>">
-                  <td><%= link_to_if_authorized proxy.name, hash_for_smart_proxy_path(proxy) %></td>
-                  <td><%=h proxy.features.map(&:name).to_sentence %></td>
```
ohadlevy, line 35: I wonder how translation worked here before?

webpack/assets/javascripts/react_app/common/helpers.js
amirfefer: It becomes a standard interface, most of the browsers support it, and there's a [polyfill](https://github.com/WebReflection/url-search-params)

amirfefer:
@ohadlevy 
I've changed the table library to [react-bootstrap-table](http://allenfang.github.io/react-bootstrap-table/) for an easy integration with pf styling. (`react-table` doesn't use the regular table, th, td tr tags)
I've created a new Tabs component, so there's only one registered component (instead of 4)
I've implemented a new `empty state` component that renders when no data instead of a `no data` string.

empty state component:
empty_state

bootsrap & pf table style:
new_table_style

ohadlevy:
@amirfefer i think you forgot to push your new implementation.

regardless, I would suggest to send PR's to patternfly-react repo with all of pf related components (tabs, emptystate etc)

webpack/assets/javascripts/react_app/redux/actions/status/index.js
amirfefer: I've extracted it out of the tree, something like this:

STATUS
    -COMPUTE_RESOURCE
        [-X...] 
    -SMART_PROXY
        [-Y...]

app/assets/stylesheets/base.scss
```diff
@@ -250,7 +250,7 @@ select {
 
 table {
   clear: both;
-  margin-bottom: 6px !important;
+  //margin-bottom: 6px !important;
```
ohadlevy: ?

app/assets/stylesheets/base.scss
```diff
@@ -335,7 +335,7 @@ table {
 }
 
 #pagination {
-  margin-top: -6px;
+  margin-top: -19px !important;
```
ohadlevy: why change style for all pagination?

app/helpers/about_helper.rb
```diff
@@ -0,0 +1,31 @@
+module AboutHelper
+  def plugins
```
ohadlevy, line 2: if you are creating a controller per action, then I would expect AboutPluginController with only index action.

having said that, I would default to use the API first and only if it doesnt fit create a UI end point

app/views/about/index.html.erb
```diff
@@ -1,122 +1,13 @@
...
-                  <td><%= plugin.version %></td>
-                </tr>
-              <% end %>
-              </tbody>
-            </table>
-          <% end %>
-        </div>
-      </div>
+      <div id="tabs"></div>   
+      <%= mount_react_component('About', '#tabs', {:compute => compute_resources, :proxy => proxies,
```
ohadlevy: probably makes sense to combine it under a method, so in theory we can expand it later vs hard code it. e.g.

```ruby
def about_data
  {:compute => compute_resources, :proxy => proxies,...
end
```

package.json
```diff
@@ -68,12 +68,13 @@
     "lodash": "~4.15.0",
     "multiselect": "~0.9.12",
     "patternfly": "^3.29.5",
-    "patternfly-react": "^0.8.1",
+    "patternfly-react": "^0.11.0",
```
ohadlevy: all package upgrades seems not related?

webpack/assets/javascripts/react_app/common/helpers.js
ohadlevy: should we replace than the urijs?

webpack/assets/javascripts/react_app/components/about/compute/index.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import Status from '../../common/status';
+import Table from '../../common/table';
+import helpers from '../../../common/helpers';
+
+class AboutComputeTable extends React.Component {
+  render() {
+    const columns = [
+      {
+        header: 'Name',
```
ohadlevy: string not extracted?

webpack/assets/javascripts/react_app/components/about/index.js
```diff
@@ -0,0 +1,25 @@
...
+import AboutProxyTable from './proxies';
+
+class About extends React.Component {
+  render() {
+    const {
+      compute, proxy, provider, plugin,
+    } = this.props.data;
+    const tabs = [__('Smart Proxies'), __('Available Providers'), __('Compute Resources'), __('Plugins')];
+
+    return <ForemanTabs id='about_tabs' tabs={tabs} >
```
ohadlevy: this looks really close to how react-router works.. I wonder if we should combine the two? (so if you refresh you come back to the same tab)

webpack/assets/javascripts/react_app/components/common/emptyState/index.js
```diff
@@ -0,0 +1,63 @@
...
+        </div>
+        <h1>{header}</h1>
+        <p>{description}</p>
+        <p>
+          Learn more about this{' '}
+          <a href={docUrl}>on the documentation</a>.
+        </p>
+        <div className="blank-slate-pf-main-action">
+          <button
+            onClick={e => window.Turbolinks.visit(action.url)}
```
ohadlevy: wouldnt turbolink catch these regardless?

webpack/assets/javascripts/react_app/components/common/emptyState/index.js
```diff
@@ -0,0 +1,63 @@
...
+    return (
+      <div className="blank-slate-pf " id={id}>
+        <div className="blank-slate-pf-icon">
+          <Icon type='pf' name={icon || 'add-circle-o'}
+          />
+        </div>
+        <h1>{header}</h1>
+        <p>{description}</p>
+        <p>
+          Learn more about this{' '}
```
ohadlevy: not extracted..

ohadlevy (changes requested):
I ran out of time, but here is some feedback

webpack/assets/javascripts/react_app/components/about/compute/index.js
amirfefer: extracted already in `<Table>` component

webpack/assets/javascripts/react_app/components/about/index.js
amirfefer: sounds great, though this PR becomes too large, maybe should we combine react-router in a different PR?

package.json
amirfefer: I've upgraded in order to use Icon component

app/helpers/about_helper.rb
amirfefer: I've created About controller namespace, inspired by [this](http://jeromedalbert.com/how-dhh-organizes-his-rails-controllers/) post.

ohadlevy: in that case why do you need the helper than?
also.. can't we simply use the existing api?

app/views/about/plugins_empty_state/index.json.rabl
```diff
@@ -0,0 +1,6 @@
+object false
+node do |u|
+  { :header => N_("Plugin"), :description => N_("Plugins are tools to extend and modify the functionality of Foreman. Plugins offer custom functions and features so that each user can tailor their environment to their specific needs."),
```
ohadlevy: this makes me wonder why not simply store this inside a component ? e.g. whats the value of making an api call to get static values?

package.json
```diff
@@ -72,6 +72,7 @@
     "raf": "^3.4.0",
     "react": "^16.0.0",
     "react-bootstrap": "^0.31.0",
+    "react-bootstrap-table": "^4.1.5",
```
ohadlevy: I'm in mixed feelings about this library - didn't  you prefer the other one? :)

webpack/assets/javascripts/react_app/components/about/compute/index.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import Status from '../../common/status';
+import Table from '../../common/table';
+import helpers from '../../../common/helpers';
+
+class AboutComputeTable extends React.Component {
```
ohadlevy: nitpick: please call it compute resource (no point in shorting the name imho)

package.json
amirfefer: The other library doesn't use bootstrap table classes, and more then that it doesn't use tables tags at all , so it would be difficult to add bootstrap and patternfly style.

amirfefer:
@ohadlevy , @sharvit , @waldenraines 

I updated this PR to use these **patternfly-react** components:
- Table
- Empty state
- Tabs

about - edited 1

amirfefer:
### Performance:
post react:
screenshot from 2018-03-10 20-25-27
pre react:
screenshot from 2018-03-10 20-29-28

amirfefer:
I've added couple of tests.

test/integration/about_test.rb
```diff
@@ -14,31 +14,17 @@ class AboutIntegrationTest < IntegrationTestWithJavascript
...
+    assert page.has_link?("Plugins", :href => "#")
     assert page.has_link?("Foreman Users", :href => "http://groups.google.com/group/foreman-users")
     assert page.has_link?("Foreman Developers", :href => "http://groups.google.com/group/foreman-dev")
     assert page.has_link?("issue tracker", :href => "http://projects.theforeman.org/projects/foreman/issues")
     assert page.has_link?("Wiki", :href => "http://projects.theforeman.org")
     assert page.has_link?("Ohad Levy", :href => "mailto:ohadlevy@gmail.com")
-    assert page.has_content?("Version")
-  end
-
-  test "about page proxies should have version" do
```
amirfefer, line 30: already tested [here](https://github.com/theforeman/foreman/pull/5024/files#diff-15b6dd3ae736678ae8923d366096fea5R49)

test/integration/about_test.rb
```diff
@@ -14,31 +14,17 @@ class AboutIntegrationTest < IntegrationTestWithJavascript
 
   test "about page" do
     assert_index_page(about_index_path,"About", nil, false, false)
-    wait_for_ajax
```
amirfefer: `axios` breaks `wait_for_ajax` (We are using axios instead of jquery in react components)
it's redundant here.

webpack/assets/javascripts/react_app/components/about/plugin/schema.js
```diff
@@ -0,0 +1,44 @@
...
+
+export const columns = [
+  {
+    property: 'name',
+    header: {
+      label: 'Name',
+      formatters: [headerFormat],
+    },
+    cell: {
+      formatters: [cellFormat],
```
tstrachota: Plugin names used to be links. Did you remove it by intention?

webpack/assets/javascripts/react_app/components/common/table/index.js
```diff
@@ -0,0 +1,40 @@
...
+  </PfTable.Cell>
+);
+
+class Table extends React.Component {
+  isEmpty() {
+    return this.props.data.length === 0;
+  }
+
+  render() {
+    const { columns, data } = this.props;
```
waldenraines: Thoughts on renaming `data` to `rows`?  I think this would make more sense since that's what these are.

webpack/assets/javascripts/react_app/components/common/emptyState/index.js
```diff
@@ -0,0 +1,70 @@
...
+        </PfEmptyState.Help>
+        <PfEmptyState.Action>
+          <Button href={action.url} bsStyle="primary" bsSize="large">
+            {action.title}
+          </Button>
+        </PfEmptyState.Action>
+        {secondayActions && (
+          <PfEmptyState.Action secondary>
+            {secondayActions.map(item => (
+              <Button href={action.url} title="Perform an action">
```
waldenraines: This string isn't translated.

webpack/assets/javascripts/react_app/components/about/plugin/index.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import Table from '../../common/table';
+import { columns } from './schema';
+
+const AboutPluginTable = (props) => {
+  const emptyStateData = {
+    header: __('Plugin'),
+    description: __('Plugins are tools to extend and modify the functionality of Foreman. Plugins offer custom functions and features so that each user can tailor their environment to their specific needs.'),
+    docUrl: 'https://www.theforeman.org/manuals/1.15/index.html#Plugins',
```
waldenraines: Don't we also need a documentation link text here?  I think we want the ability to customize the text of this link.

webpack/assets/javascripts/react_app/components/about/plugin/index.js
```diff
@@ -0,0 +1,20 @@
...
+    action: { title: __('Get a plugin'), url: 'https://projects.theforeman.org/projects/foreman/wiki/List_of_Plugins' },
```
waldenraines: Using url here will cause a full page reload instead of a transition if using `react-router`.  How do you feel about adding a `to` property here for use with `react-router`?  Obviously we would also need to include react-router into core so I'm fine if we tackle this outside of this PR.

webpack/assets/javascripts/react_app/components/about/proxies/schema.js
```diff
@@ -0,0 +1,78 @@
...
+      formatters: [headerFormat],
+    },
+    cell: {
+      formatters: [ellipsisFormat],
+    },
+  },
+  {
+    property: 'id',
+    header: {
+      label: 'Status',
```
waldenraines: Should these labels be translated?

webpack/assets/javascripts/react_app/components/about/compute/schema.js
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import ConnectedStatus from '../../common/status';
+import { headerFormat, cellFormat } from '../../common/table';
+import helpers from '../../../common/helpers';
+
+export const columns = [
```
waldenraines, line 6: Any thoughts to how this will work with selectable columns?

waldenraines (changes requested):
Used the table/empty state wrapper and schema in my work on https://github.com/Katello/katello/pull/7230.  

I think I like the approach and just have a few comments.

webpack/assets/javascripts/react_app/components/about/compute/schema.js
amirfefer: It should work,
we can follow patternfly-react [example](https://github.com/patternfly/patternfly-react/blob/master/src/components/Table/__mocks__/mockClientPaginationTable.js)

webpack/assets/javascripts/react_app/components/about/plugin/index.js
waldenraines: Or perhaps accept components for the actions instead?  That would make this more flexible at the cost of some repetition.

webpack/assets/javascripts/react_app/components/common/tabs/index.js
```diff
@@ -0,0 +1,33 @@
...
+
+const TabsWrapper = (props) => {
+  const { tabs, children } = props;
+  return (
+    <TabContainer id="basic-tabs-pf" defaultActiveKey={0}>
+      <div>
+        <Nav bsClass="nav nav-tabs nav-tabs-pf" justified>
+          {tabs.map((tab, index) => (
+            <NavItem key={index} eventKey={index}>
+              {__(tab)}
```
tstrachota: This is translated already in https://github.com/theforeman/foreman/pull/5024/files#diff-26f92dbe854c0b8737e453ea192ebe2fR12
Translations should be handled outside of components if possible, so remove this one, please.

webpack/assets/javascripts/react_app/components/common/emptyState/index.js
tstrachota: I guess that "Perform an action" is just copy leftover anyway.

webpack/assets/javascripts/react_app/components/common/emptyState/index.js
```diff
@@ -0,0 +1,70 @@
...
+  componentDidMount() {
+    const { data, uri, model } = this.props;
+    const url = model
+      ? helper.urlBuilder(this.props.model, 'empty_state')
+      : uri;
+
+    if (data) {
+      this.setState({ data });
+    } else {
+      API.get(url).then(
```
tstrachota: Why do we need to fetch the data from server? Does it contain anything else than translated strings?
Since a rabl template needs to be created anyway, can't we just drop the server side and create a wrapper empty state component for each model and save one request? It would also help with keeping the code at one place.

app/helpers/about_helper.rb
tstrachota: Using an api endpoint would definitely be more systemic. Such endpoint doesn't exist at the moment. We can only list proxies and compute profiles separately. How about moving it to a new api endpoint? It would help with http://projects.theforeman.org/issues/3036.

But I understand it's adds some work and there are probably issues with higher priority. So if you don't have space for moving it to API, I can live with the current state ;)

webpack/assets/javascripts/react_app/components/common/emptyState/index.js
```diff
@@ -0,0 +1,70 @@
...
+      secondayActions,
+    } = this.state.data;
+
+    return (
+      <PfEmptyState>
+        <PfEmptyState.Icon type="pf" name={icon || 'add-circle-o'} />
+        <PfEmptyState.Title>{header}</PfEmptyState.Title>
+        <PfEmptyState.Info>{__(description)}</PfEmptyState.Info>
+        <PfEmptyState.Help>
+          For more information please see{' '}
```
tstrachota: This needs to be translated

webpack/assets/javascripts/react_app/components/common/emptyState/index.js
```diff
@@ -0,0 +1,70 @@
...
+          <a href={docUrl}>{__('documentation')}</a>
```
tstrachota: I'm afraid this is going to cause issues with translating languages with different word order. It effectively prevents from putting the word "documentation" anywhere else in the sentence.

webpack/assets/javascripts/react_app/components/common/emptyState/index.js
```diff
@@ -0,0 +1,70 @@
...
+      docUrl,
+      action,
+      secondayActions,
+    } = this.state.data;
+
+    return (
+      <PfEmptyState>
+        <PfEmptyState.Icon type="pf" name={icon || 'add-circle-o'} />
+        <PfEmptyState.Title>{header}</PfEmptyState.Title>
+        <PfEmptyState.Info>{__(description)}</PfEmptyState.Info>
```
tstrachota: This should be translated outside of the component.

app/controllers/application_controller.rb
```diff
@@ -36,6 +36,12 @@ def welcome
     not_found
   end
 
+  def empty_state
```
tstrachota: Loading the empty state setup from the server side seems unnecessarily complicated to me. See my comment in the empty state component please.

tstrachota (changes requested):
Thanks @amirfefer. It looks good visually and I like the status components.

Table labels need to be tranlsated. There's also couple of translation issues in the empty state component.

What benefits does loading the empty state settings from the server bring?

webpack/assets/javascripts/react_app/components/about/plugin/index.js
amirfefer: good point, I added prop `documentationText` with a default value of Documentation

webpack/assets/javascripts/react_app/components/common/emptyState/index.js
ohadlevy: there is also https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/components/common/DocumentationLink/index.js maybe we can combine it somehow?

webpack/assets/javascripts/react_app/components/about/compute/index.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import Table from '../../common/table';
+import { columns } from './schema';
+
+const AboutComputeTable = props => (
+      <Table
```
ohadlevy: inconsistent spacing?

webpack/assets/javascripts/react_app/components/about/plugin/index.js
ohadlevy: we can also use Turbolinks.visit(url) as we do in many other places.

waldenraines:
@amirfefer updates on this?  Would like to get this in soon since it contains change that I am using in https://github.com/Katello/katello/pull/7230

amirfefer:
@waldenraines : ready for another review :)

amirfefer:
@tstrachota, thanks for your review.
I tend to agree with the empty state API endpoint, it's a bit redundant.

webpack/assets/javascripts/react_app/components/common/emptyState/index.js
```diff
@@ -0,0 +1,43 @@
...
+    header,
+    description,
+    customDocumentation,
+    documentationLabel = 'For more information please see',
+    documentationButton = 'Documentation',
+    docUrl,
+    action,
+    secondayActions,
+  } = props;
+  const defaultDocumantion = __(`${documentationLabel} <a href=${docUrl}>${documentationButton}</a>`);
```
tstrachota: This is not going to work correctly. The string gets interpreted first so you'll get a string `"For more information..."` but translation files will contain key `"${documentationLabel} <a href=${docUrl}>${documentationButton}</a>"` and that's not going to match.

We need to use `sprintf` (for instance like here https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/common/reduxFormI18n.js#L12)

webpack/assets/javascripts/react_app/components/common/emptyState/index.js
```diff
@@ -0,0 +1,43 @@
...
+import { EmptyState as PfEmptyState, Button } from 'patternfly-react';
+
+const EmptyState = (props) => {
+  const {
+    icon = 'add-circle-o',
+    header,
+    description,
+    customDocumentation,
+    documentationLabel = 'For more information please see',
+    documentationButton = 'Documentation',
```
tstrachota: Are these defaults used anywhere? If yes, they should be translated as well.

tstrachota (changes requested):
Two more translation-related comments. Apart from that it works nicely!

waldenraines (approved):
Looks good to me.  Everything I tested worked and I also reused the table and empty state components successfully. 

@amirfefer please fix the build and hopefully we can get this merged soon.

tstrachota:
IMHO there are still 2 translation issues that deserve fixing - see my previous comments.

waldenraines:
> IMHO there are still 2 translation issues that deserve fixing - see my previous comments.

Yeah, for sure, I didn't mean this was approved without those changes but including them.

webpack/assets/javascripts/react_app/components/common/emptyState/index.js
```diff
@@ -0,0 +1,43 @@
...
+    docUrl,
```
waldenraines: I still would prefer to see this as an object like the action.  We should be able to customize the text of this button instead of it always being "Documentation".

Something like this:

```javascript
documentation: {
  title: __('Learn more about adding Subscriptions to Allocations'), 
  url: 'http://redhat.com' 
},
```

waldenraines (changes requested):

webpack/assets/javascripts/react_app/components/common/emptyState/index.js
waldenraines: Nitpick, spelling is incorrect, should be documentation not documantion.

webpack/assets/javascripts/react_app/components/common/table/index.js
```diff
@@ -0,0 +1,40 @@
...
+        bordered
+        hover
+        columns={columns}
+      >
+        <PfTable.Header />
+        <PfTable.Body
+          rows={rows}
+          rowKey={({ rowData, rowIndex }) => rowIndex}
+        />
+      </PfTable.PfProvider>
```
waldenraines, line 36: Do we want to include pagination with this?  Right now if you have pagination on your table and trigger the empty state you get this:

screen shot 2018-03-27 at 11 38 50 am

Was thinking maybe you could pass the pagination component into this directly so that way you maintain control of it from within your component.  In other words:

```javascript
const paginationRow = (
  <PaginationRow
    viewType="table"
    itemCount={subscriptions.pagination.total}
    pagination={subscriptions.pagination}
    onChange={onPaginationChange}
  />
);

<Table
  rows={subscriptions.results}
  columns={columns}
  emptyState={emptyStateData()}
  paginationRow={paginationRow}
/>
```

Then all this component would do is include the pagination within the ternary operator.

waldenraines (changes requested):

webpack/assets/javascripts/react_app/components/common/table/index.js
waldenraines: I'm also okay with adding this in a future PR.

tstrachota: I'm leaning towards fixing it in a future PR. It's not a blocker for this one that has been open for a long time already.

I'd propose a bit different solution. The empty state design is quite constant while the table implementation can vary a lot (inline edits, provider setup, etc.). I think that it would be more flexible to refactor (and maybe rename) the component so that it accepts children with default being a table. But as I've said, I prefer to solve it outside of this PR.

waldenraines: > I'd propose a bit different solution. The empty state design is quite constant while the table implementation can vary a lot (inline edits, provider setup, etc.). I think that it would be more flexible to refactor (and maybe rename) the component so that it accepts children with default being a table. But as I've said, I prefer to solve it outside of this PR.

I also prefer this solution and am okay updating katello once we have it done.

ohadlevy:
[test foreman]

mmoll:
This needs at least a rebase :point_up:
Also, have a look into the failing travis test.

tstrachota:
@amirfefer ping, could you rebase, please?

xprazak2:
@amirfefer, could you rebase please?

tstrachota:
@amirfefer ping

amirfefer:
It was among the latter in the priority tasks list, 
I'll rebase and  have a look on it soon.

amirfefer:
@tstrachota , @waldenraines 
Rebased, and ready for a another look :)

webpack/assets/javascripts/react_app/components/common/table/index.js
```diff
@@ -0,0 +1,40 @@
...
+  isEmpty() {
+    return this.props.rows.length === 0;
+  }
+
+  render() {
+    const { columns, rows, emptyState } = this.props;
+    return this.isEmpty() ? (
+      <EmptyState {...emptyState} />
+    ) : (
+      <PfTable.PfProvider
```
boaz0, line 24: Hi @amirfefer, this looks good but what if I want to use `components` in `<PfTable.PfProvider>` - I am not sure I can do that through `Table`. In `Hardware Models` I am using it to set how the heading and cells will look like.

tstrachota:
There's minor conflict again. I wasn't fast enough with the final review:( I'm sorry for that. Please rebase @amirfefer

webpack/assets/javascripts/react_app/components/common/table/index.js
tstrachota: This has been solved in Katello: https://github.com/Katello/katello/blob/master/webpack/move_to_foreman/components/common/table/components/Table.js That component is based on @amirfefer 's solution and is going to be moved to foreman soon. AFAIR the interfaces should be compatible.
So, if you don't mind @ripcurld, I wouldn't block merging this PR by that. It contains lot of other good stuff, has been open for quite some time and is quite ready to be merged.

amirfefer:
@tstrachota  - rebased :)

webpack/assets/javascripts/react_app/components/common/table/index.js
boaz0: @tstrachota I didn't mean to block :smiley:  but just to raise a concern I have in [`Hardware Models`](https://github.com/theforeman/foreman/pull/5755).

webpack/assets/javascripts/react_app/common/EmptyStates.js
```diff
@@ -0,0 +1,38 @@
+export const computeResource = () => ({
+  header: __('Compute Resource'),
+  description: __('Foreman supports creating and managing hosts on a number of virtualization and cloud services - referred to as “compute resources” - as well as bare metal hosts.'),
+  documentation: {
+    // eslint-disable-next-line no-undef
+    url: `https://www.theforeman.org/manuals/${VERSION}/index.html#5.2ComputeResources`,
```
tstrachota, line 6: I found out that the docs link (and the other too) leads to 404. `VERSION` is `1.19.0` but the manual url has to contain only `1.19`. We probably need both `VERSION` and `VERSION_MAJOR` or something similar, maybe an object containing multiple version formats and with `toString()` defined.

webpack/assets/javascripts/react_app/components/common/emptyState/index.js
```diff
@@ -0,0 +1,46 @@
...
+
+  return (
+    <PfEmptyState>
+      <PfEmptyState.Icon type="pf" name={icon} />
+      <PfEmptyState.Title>{header}</PfEmptyState.Title>
+      <PfEmptyState.Info>{description}</PfEmptyState.Info>
+      {url &&
+        <PfEmptyState.Help>
+         {label}
+          <a href={url}> {buttonLabel} </a>
```
tstrachota, line 26: Nitpick:
this indend doesn't put any space between the text and the link. Spaces in `<a>` work but they look weird on hover:
screenshot from 2018-07-09 15-57-09

Please indent correctly and remove spaces form the anchor. `{" "}` should help with preserving the space in between.

tstrachota (changes requested):
I found two more issues on the empty states. Hopefully they'll be easy to fix.
The rest looks good and worked fine for me.

mmoll:
@amirfefer could you rebase and fix the last two small comments from @tstrachota? :)

mmoll:
what's the status here?

tstrachota:
This needs a rebase again.

waldenraines:
@amirfefer any updates on this?

amirfefer:
@waldenraines  - basically i'd like to use table from this [PR](https://github.com/theforeman/foreman/pull/5755), will re-prioritize this soon :)

mmoll:
@amirfefer what's the status here?

mmoll:
any news here? :)

ares:
@amirfefer this needs a bigger rebase now, could you please let us know if you plan to look at this soon? If there's no activity before next triage (2 weeks), we'd lean towards closing this. It would be unfrotunate as there were just 2 small things to finish this, but of course it can be reopened later if your availability for this PR changes.

ares:
Unfortunatelly there was no reply for 2 weeks so we are closing this PR now. Please reopen if you're still interested in getting this in.

PR #5025, opened 2017-11-23
Fixes #21757 - Move dynflow executor to script/dynflowd

ekohl:
This matches with foreman-{rake,tail} which are both in this directory and installed in the same way. It also saves us a rename in packaging.


iNecas (approved):
Makes sense

iNecas:
The katello failures are caused by https://github.com/theforeman/foreman-infra/pull/390 and not this PR. Merging. Thanks @ekohl. Should this be also cherry-picked to 1.16? I'm a bit lost right now what's the plan regarding this and 1.16 in general.

ekohl:
@iNecas right now I'm also unclear about the 1.16 situation and for now I just focus on getting it into nightly.

PR #5027, opened 2017-11-24
 fixes #21762 - Add disk options to oVirt provider 

bagasse:
Add two disk options for oVirt provider:
* The possibility to choose disk interface (VirtIO, VirtIO-SCSI, IDE)
* The possibility to tell to oVirt to wipe the disk when deleted

Tested on oVirt 4.1 with APIv3


dLobatog (approved):
Thanks @bagasse, works well on 4.1.6 too

PR #5030, opened 2017-11-27
Fixes #21776 - Improved fact importing to deal with names

ShimShtein:
Now the list of fact names will be calculated beforehand and saved to the database. It will also consider that the name could be added from other thread.

This is the work from https://github.com/theforeman/foreman/pull/4996. It just got a new issue ID.

### DIFF
--- a/app/services/fact_importer.rb
+++ b/app/services/fact_importer.rb
@@ -39,15 +39,23 @@ class FactImporter
 
   # expect a facts hash
   def import!
-    delete_removed_facts
-    update_facts
-    add_new_facts
+    # This function uses its own transactions that should not be included
+    # in the transaction that handles fact values
+    ensure_fact_names
 
-    if @error
-      Foreman::Logging.exception("Error during fact import for #{@host.name}", @error)
-      raise ::Foreman::WrappedException.new(@error, N_("Import of facts failed for host %s"), @host.name)
+    ActiveRecord::Base.transaction do
+      delete_removed_facts
+      update_facts
+      add_new_facts
     end
+
+    report_error(@error) if @error
     logger.info("Import facts for '#{host}' completed. Added: #{counters[:added]}, Updated: #{counters[:updated]}, Deleted #{counters[:deleted]} facts")
+  end
+
+  def report_error(error)
+    Foreman::Logging.exception("Error during fact import for #{@host.name}", error)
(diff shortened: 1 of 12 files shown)


lzap (approved):
Thanks I have no comments, good work. Can you please share output of all benchmarks just for record purposes? Ideally if you have development setup with postgresql, because sqlite3 numbers were a bit irrelevant the last time. Paste it here and we are good to merge.

lzap:
FYI there are relevant failures on CI.

ShimShtein:
Since unit tests are running in transaction by default, added an option to suppress transaction check in facts importer.

ShimShtein:
Hopefully the tests will be green this time.
Here are the benchmark results:
[before.txt](https://github.com/theforeman/foreman/files/1530593/before.txt)
[after.txt](https://github.com/theforeman/foreman/files/1530594/after.txt)

ShimShtein:
[test foreman]

Test failure unrelated, let's retest.

lzap:
Thanks, LGTM the only thing is Katello tests do fail and it seems relevant. Are you planning to do a PR into Katello? Or is this an issue with the code?

lzap:
Benchmark shows good results on performance, but allocations raised from 99 millions to 145 millions. I'd expect little bit less, but this is acceptable.

ShimShtein:
Fixed tests for both Katello and Discovery.

lzap:
Hmmm I am not sure if the price you are paying for `ensure_no_active_transaction` is really worth it. Is the whole tests breakage just because you want to be sure names are added outside of the transaction, or is there another reason I don't see?

ShimShtein:
> Is the whole tests breakage just because you want to be sure names are added outside of the transaction

Yes, tests are run in transaction by default and the newly added exception is triggered. IMHO it's worth fixing the tests as long as the user will not encounter unexpected behavior down the road. To be more precise - I don't have a problem with running name additions inside a transaction, I do have a problem with fact names added concurrently while I am in a transaction. I think we should guard this place in code explicitly, since we don't know exactly how it will be used in plugins/at a later time.

The reason behind this is that failed `INSERT` statement marks the whole outer transaction as failed, so no more queries will be allowed (not even `SELECT`).

lzap:
Merged, this breaks all plugins which rely on fact importing code, I was able to find: Katello, Discovery (both has PRs) and Foreman Salt: @stbenjam @mmoll @ares to name some commiters. Setting redmine this time!

PR #5031, opened 2017-11-27
Fixes #21781 - added getenforce to generic OS debug

lzap:
A quick merge, as simple as that.

### DIFF
--- a/script/foreman-debug
+++ b/script/foreman-debug
@@ -286,6 +286,7 @@ if [ $NOGENERIC -eq 0 ]; then
   add_files /var/log/messages*
   add_files /var/log/audit/audit.log*
   add_files /var/log/syslog*
+  add_cmd "getenforce" "selinux_state"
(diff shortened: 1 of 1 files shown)


script/foreman-debug
```diff
@@ -286,6 +286,7 @@ if [ $NOGENERIC -eq 0 ]; then
   add_files /var/log/messages*
   add_files /var/log/audit/audit.log*
   add_files /var/log/syslog*
+  add_cmd "getenforce" "selinux_enforce"
```
sean797: not necessary but IMHO `selinux_state` file would be better.

```sh
[root@katello ~]# getenforce 
Permissive

sean797 (approved):

script/foreman-debug
lzap: Done thanks

dLobatog (approved):
Thanks @lzap & @sean797 !

PR #5034, opened 2017-11-29
Fixes #21637 - fallback to default locale when I18n translation is missing

iNecas:
This makes the I18n behavior closer to the gettext one, and prevents
some ugly 'missing_translation' messages in the UI (for example in
apipie, that uses the I18n backend for it's strings).


dLobatog (approved):
Thanks @iNecas !

PR #5035, opened 2017-11-29
Fixes #21805 - don't list default role

ares:
to reproduce
```
hammer user info --id 4

Id:                    4
...
Roles:                 
    Default role         # this should not be here, we don't display it in UI either, it's system role
    Viewer
```


dLobatog (approved):
@ares 👍 when 🍏

ares:
@dLobatog and we just got :green_apple: :-)

PR #5037, opened 2017-11-30
Fixes #21816 - Start dynflowd on dev enviroment

dLobatog:


timogoebel:
I'd vote for running dynflow in a separate process using `foreman` / the Procfile. That better mirros how this works in production.

config/application.rb
```diff
@@ -279,7 +279,7 @@ def dynflow
       dynflow.eager_load_actions!
       dynflow.config.increase_db_pool_size
 
-      unless dynflow.config.lazy_initialization
+      unless Rails.env.production?
```
iNecas, line 282: This is not the right condition: in production, the lazy initialization is actually false: the question, if the world will have an executor, depends on something else (search for `remote?` method). I think this condition should just go away completely

iNecas:
I like the current approach in development better, but I'm biased. I think it's worth asking other devs (including katello folks) about that. Anyway, I don't think it's actually relevant to this particular patch.

config/application.rb
ohadlevy: @timogoebel I share your opinion - see https://github.com/theforeman/foreman/pull/4871

iNecas: We also need to check, whether we are in rake (uness in the worker rake), to prevent the issues with db drop, that jenkins is hitting: sry for not being more verbose: Writing at phone at the moment.

lzap:
I know I am coming late to the discussion, but dynflow is multi-threaded code and that's unacceptable in our thrad-usafe codebase. Procfile is the only reasonable way of running dynflow. Also, in production we must make sure there is just one thread per instance - this is how it worked for years and we must not change this for now. I just started thread on this: https://community.theforeman.org/t/on-spawning-threads-in-our-rails-app/8650

PR #5038, opened 2017-11-30
Fixes #21444 - change subitems headers style in vertical navigation

amirfefer:
I've decreased font size, it shouldn't be the same as the main header size.

screenshot from 2017-11-30 15-25-06

before changing:
screenshot from 2017-11-30 15-25-33

### DIFF
--- a/app/views/home/_vertical_menu.html.erb
+++ b/app/views/home/_vertical_menu.html.erb
@@ -15,8 +15,9 @@
         <% when Menu::Item %>
           <%= menu_secondary_item item %>
         <% when Menu::Divider %>
-          <%= content_tag(:div, "", :class => "divider") %>
-          <%= content_tag(:span, _(item.caption), :class => "nav-item-pf-header") if item.caption %>
+          <%= content_tag(:li, :class => "list-group-item") do %>
+            <%= content_tag(:span, _(item.caption), :class => "vertical-sub-header") if item.caption %></a>
+          <% end %>
         <% when Menu::Toggle %>
           <%= content_tag :li, :class => "dropdown-submenu" do %>
             <%= link_to(_(item.caption), "#", :class => "dropdown-toggle", :"data-toggle" => "dropdown") %>
--- a/app/assets/stylesheets/patternfly_and_overrides.scss
+++ b/app/assets/stylesheets/patternfly_and_overrides.scss
@@ -380,3 +380,12 @@ a.btn-info span {
 .container-pf-nav-pf-vertical.hide-nav-pf {
   display: none !important;
 }
+
+.vertical-sub-header {
+  line-height: 25px;
+  font-size: 12px;
+  margin-left: 20px;
+  padding: 4px 0 2px;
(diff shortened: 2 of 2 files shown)


app/views/home/_vertical_menu.html.erb
```diff
@@ -15,8 +15,9 @@
         <% when Menu::Item %>
           <%= menu_secondary_item item %>
         <% when Menu::Divider %>
-          <%= content_tag(:div, "", :class => "divider") %>
-          <%= content_tag(:span, _(item.caption), :class => "nav-item-pf-header") if item.caption %>
+          <%= content_tag(:li, :class => "list-group-item") do %>
+           <%= content_tag(:span, _(item.caption), :class => "vertical-sub-header") if item.caption %></a>
```
ares: I see it's not the code that is related to this PR but does _(item.caption) really work? shouldn't it be N_(item.caption)?

ares:
I like it, @rohoover :+1: or :-1: :question:

app/views/home/_vertical_menu.html.erb
sharvit: I noticed a small indentation problem, need to add another space at line 19

sharvit (changes requested):

app/views/home/_vertical_menu.html.erb
amirfefer: The string is extracted in `Loader.rb`already, so I guess it's ok to use `_()` here.
There is another translation issue, which I solved [here](https://github.com/theforeman/foreman/pull/5049)

waldenraines:
@amirfefer @ares @rohoover what's the status of this? Can we merge?

waldenraines (approved):

Rohoover:
I'm good with it.

app/views/home/_vertical_menu.html.erb
```diff
@@ -15,8 +15,9 @@
         <% when Menu::Item %>
           <%= menu_secondary_item item %>
         <% when Menu::Divider %>
-          <%= content_tag(:div, "", :class => "divider") %>
-          <%= content_tag(:span, _(item.caption), :class => "nav-item-pf-header") if item.caption %>
+          <%= content_tag(:li, :class => "list-group-item") do %>
+            <%= content_tag(:span, menu_title, :class => "vertical-sub-header") if item.caption %></a>
```
ares: menu_title should probably be item.caption, otherwise I'm getting wrong headers, see 
screenshot_20171212_123301

ares (changes requested):
It looks alright, but there's a regression breaking headers, good to go in after fixing.

amirfefer:
@ares Thanks,
could you have another look?

ares (approved):
looks good and works well

ares:
Thanks @amirfefer, merging! Katello tests failures are unrelated, they would turn green after rebasing but since they ran all successfully before the rake task failed, I guess we can merge.

PR #5039, opened 2017-11-30
fixes #21823 - adds foreman user agent to RSS loading

ohadlevy:


app/services/ui_notifications/rss_notifications_checker.rb
```diff
@@ -70,5 +72,19 @@ def rss_notification_blueprint
...
+      http = Net::HTTP.new(uri.host, uri.port)
+      http.use_ssl = true
+      http.verify_mode = OpenSSL::SSL::VERIFY_NONE
+      request = Net::HTTP::Get.new(uri.request_uri)
+      request.initialize_http_header({"User-Agent" => rss_user_agent})
+      http.request(request).body
+    end
+
+    def rss_user_agent
+      "Foreman #{SETTINGS[:version]} RSS notifications"
```
ekohl: https://tools.ietf.org/html/rfc7231#section-5.5.3 suggests the common way is to separate the product and version with a slash. I'm also not sure about the RSS notifications part of it but if you do, convention is to put it in braces: ```Foreman/#{SETTINGS[:version]} (RSS notifications)```

app/services/ui_notifications/rss_notifications_checker.rb
```diff
@@ -70,5 +72,19 @@ def rss_notification_blueprint
     def notification_already_exists?(item)
       !!Notification.unscoped.find_by_message(item.summary)
     end
+
+    def load_rss_feed
+      uri = URI.parse(@url)
+      http = Net::HTTP.new(uri.host, uri.port)
+      http.use_ssl = true
+      http.verify_mode = OpenSSL::SSL::VERIFY_NONE
```
ekohl: Please do verify SSL or at least provide an option to disable it (but default to verification). We should assume correct SSL stacks and not trust incorrect ones.

app/services/ui_notifications/rss_notifications_checker.rb
```diff
@@ -70,5 +72,19 @@ def rss_notification_blueprint
     def notification_already_exists?(item)
       !!Notification.unscoped.find_by_message(item.summary)
     end
+
+    def load_rss_feed
+      uri = URI.parse(@url)
+      http = Net::HTTP.new(uri.host, uri.port)
```
ekohl: If you use this instead of the ```Net::HTTP.start(uri.host, uri.port) do |http|``` form, does it correctly close the connection after this function or will it wait for the garbage collection?

app/services/ui_notifications/rss_notifications_checker.rb
```diff
@@ -70,5 +72,19 @@ def rss_notification_blueprint
...
+      http.use_ssl = true
+      http.verify_mode = OpenSSL::SSL::VERIFY_NONE
+      request = Net::HTTP::Get.new(uri.request_uri)
+      request.initialize_http_header({"User-Agent" => rss_user_agent})
```
ekohl: An option to opt-out of changing this should be considered.

app/services/ui_notifications/rss_notifications_checker.rb
```diff
@@ -70,5 +72,19 @@ def rss_notification_blueprint
     def notification_already_exists?(item)
       !!Notification.unscoped.find_by_message(item.summary)
     end
+
+    def load_rss_feed
+      uri = URI.parse(@url)
+      http = Net::HTTP.new(uri.host, uri.port)
+      http.use_ssl = true
```
ekohl: The documentation at https://ruby-doc.org/stdlib-2.4.2/libdoc/net/http/rdoc/Net/HTTP.html#class-Net::HTTP-label-HTTPS suggests this is no longer needed if you just pass in the uri rather than the ```uri.request_uri```.

app/services/ui_notifications/rss_notifications_checker.rb
```diff
@@ -1,5 +1,7 @@
 require 'rss'
 require 'date'
+require "net/https"
```
ekohl: https://ruby-doc.org/stdlib-2.4.2/libdoc/net/http/rdoc/Net/HTTP.html#class-Net::HTTP-label-HTTPS states
> In previous versions of Ruby you would need to require 'net/https' to use HTTPS. This is no longer true.

ekohl (changes requested):

app/services/ui_notifications/rss_notifications_checker.rb
ohadlevy: it should close the connection once you call the http.request.body

ohadlevy:
@ekohl updated.

ohadlevy:
@dLobatog mind having a look? thanks :)

ekohl (approved):
I'll leave it to the core team to review further.

ohadlevy:
in the web server logs, the following will be observed:
```
"GET /feed.xml HTTP/1.1" 200 83237 "-" "Foreman/1.17.0-develop (RSS notifications)"
```

boaz0 (approved):
LGTM 🏆

app/services/ui_notifications/rss_notifications_checker.rb
ares: I don't think `http.request.body` closes the connection, but it looks like HTTP#request [runs](http://ruby-doc.org/stdlib-1.9.3/libdoc/net/http/rdoc/Net/HTTP.html) the request in a start using the block. Though I agree, using start with block would be more readable and ruby-like.

ares (changes requested):
Tested, works well. @ohadlevy while the code really closes the HTTP connection, would you mind changing it to `start` with block so it's clear from first read? Then it's good to get in.

ares:
Also it needs rebase now.

ohadlevy:
> I don't think http.request.body closes the connection, but it looks like HTTP#request runs the request in a start using the block. Though I agree, using start with block would be more readable and ruby-like.

@ares what would you suggest I change? I didnt think do a block and return inside the block would be better?

(rebased regardless so tests are executed).

ares:
> @ares what would you suggest I change? I didnt think do a block and return inside the block would be better?

not sure why you'd need to return, see https://github.com/ohadlevy/foreman/pull/10, works just fine, the docs says `#start` [returns the return value of a block](https://docs.ruby-lang.org/en/2.0.0/Net/HTTP.html#method-i-start)

app/services/ui_notifications/rss_notifications_checker.rb
```diff
@@ -70,5 +72,18 @@ def rss_notification_blueprint
     def notification_already_exists?(item)
       !!Notification.unscoped.find_by_message(item.title)
     end
+
+    def load_rss_feed
+      uri = URI.parse(@url)
+      Net::HTTP.start(uri.host, uri.port, :use_ssl => uri.scheme == 'https') do |http|
+        request = Net::HTTP::Get.new(uri.request_uri)
+        request.initialize_http_header({"User-Agent" => rss_user_agent})
+        http.request(request).body
```
ekohl: Should there be any error checking for incorrect HTTP status codes? It's possible the host is unreachable, misconfigured or a proxy is refusing it.

ohadlevy:
@ares would you squash on merge?

ares:
I'll be happy to squash, I sent another PR at https://github.com/ohadlevy/foreman/pull/11 to address @ekohl's comment. @ohadlevy could you please merge that one as well? I guess then it's ready to be merged.

ekohl (approved):

ares (approved):

ares (changes requested):
the change has broken tests, I'll send another addition

ares:
tests fixed at https://github.com/ohadlevy/foreman/pull/12, the new code needed more stubbing

ares:
now Foreman tests failed on HostgroupJSTest on a single combination, retriggering [test foreman], once it's green I'll squash and merge

ares (approved):

ohadlevy:
foreman tests are :green_apple: - katello failures seem unrelated?

ares:
Thanks @ohadlevy, merging. @GregSutcliffe @orrabin you might be interested in this.

PR #5040, opened 2017-11-30
Fixes #21824 - use /var/tmp for debugs and cleaning

lzap:
Looks like the trap is not being called on exit, we need to call it explicitly.

Second, we need to make sure everything is going in `/var/tmp` instead `/tmp` (tmpfs file system).

I added an extra check around cleanup recusrive removal, because user can actually supply the target directory via `-d` option - please review and test carefully. We must prevent from doing `rm -rf /` here when `-d /` is passed.

### DIFF
--- a/script/foreman-debug
+++ b/script/foreman-debug
@@ -19,9 +19,9 @@ This program can be used on Foreman instances, Smart Proxy instances or
 backend services separately.
 
 OPTIONS:
-  -d DIR  Directory to place the tarball in (default /tmp/foreman-XYZ)
+  -d DIR  Directory to place the tarball in (default /var/tmp/foreman-XYZ)
   -g      Skip generic info (CPU, memory, firewall etc.)
-  -a      Do not generate a tarball from the resulting directory
+  -a      Keep directory and do not generate a compressed tarball
   -s SIZE Maximum log set size in MB (current+rotated files, defaults to 10 MB)
   -j PRG  Filter with provided program when creating a tarball
   -p      Additionally print password patterns being filtered out
@@ -246,13 +246,28 @@ else
 fi
 printv "Determined $OS distribution"
 
+clean_temp() {
+  # prevent from removing anything user-defined
+  if [[ "$NOTAR" -eq 0 && "$DIR" == /var/tmp* ]]; then
+    printv "Cleaning $DIR"
+    rm -rf "$DIR"
+  fi
+  printv "Cleaning $TMPDIR"
+  rm -rf "$TMPDIR"
+}
+
+install_clean_trap() {
+  trap clean_temp EXIT
+}
+
 if [ -z "$DIR" ]; then
-  DIR=$(mktemp -d foreman-debug-XXXXX -p /tmp)
-  [ "$NOTAR" -eq 0 ] && trap "rm -rf $DIR" EXIT
+  DIR=$(mktemp -d foreman-debug-XXXXX -p /var/tmp)
 else
   [ ! -d "$DIR" ] && mkdir -p "$DIR"
 fi
-printv "Directory $DIR created"
+export TMPDIR=$(mktemp -d foreman-debug-auxtmp-XXXXX -p /var/tmp)
+install_clean_trap
+printv "Created $DIR and $TMPDIR"
 
 TARBALL="$DIR.tar$EXTENSION"
 
@@ -291,6 +306,7 @@ if [ $NOGENERIC -eq 0 ]; then
   add_cmd "ausearch -m AVC -m USER_AVC -m SELINUX_ERR || grep AVC /var/log/audit/audit.log" "selinux_denials.log"
   if [ -f /usr/sbin/selinuxenabled ] && /usr/sbin/selinuxenabled; then
     add_cmd "sepolgen-ifgen &>/dev/null && audit2allow -Ra || audit2allow -a" "selinux_audit2allow"
+    add_cmd "audit2allow -Ra || audit2allow -a" "selinux_audit2allow"
     add_cmd "semodule -l" "selinux_modules"
     add_cmd "semanage boolean -l" "selinux_booleans"
     add_cmd "semanage fcontext -l" "selinux_fcontext"
@@ -360,6 +376,8 @@ if [ -d "/usr/share/foreman/script/foreman-debug.d" ]; then
     if [ -x "$extension" ]; then
       printv "Processing extension $extension"
       source "$extension" 2>/dev/null
+      # install cleanup trap again just in case it got overwritten
+      install_clean_trap
     fi
   done
 fi


script/foreman-debug
```diff
@@ -246,9 +246,18 @@ else
 fi
 printv "Determined $OS distribution"
 
+clean_temp() {
+  [ "$NOTAR" -eq 1 ] && exit 0
+  # prevent from removing anything user-defined
```
iNecas: I wonder if we should delete the whole `/var/tmp` vs. `/var/tmp/foreman-debug-*`

lzap: Sorry what?

lzap: My goal here is to doublecheck user did not input "/" that could lead to disaster.

iNecas: Sry, read that wrong. Scratch that

lzap: Thanks can you unblock the review then?

lzap:
Easy merge, easy to test. I remember @rahulbajaj0509 was doing some foreman-debug work recently? :-)

script/foreman-debug
```diff
@@ -246,9 +246,18 @@ else
...
+  [[ "$DIR" != /var/tmp* ]] && exit 0
+  printv "Cleaning $DIR"
+  rm -rf $DIR
```
iNecas: Missing quotes `"$DIR"`, just to be on a safe side

script/foreman-debug
```diff
@@ -402,4 +411,5 @@ fi
 
 printv "Finished in $SECONDS seconds"
 
+clean_temp
```
iNecas: Why calling it here? Don't we want to clean the things up only, when we create the temporary dirs by ourselves? If the user provides their own dir, we should probably keep it for them.

Given we register the fucntion using `trap`, it would actually get called twice.

script/foreman-debug
iNecas: Reading this again, it would actually delete user-defined thing, if that was in `/var/tmp`

iNecas (changes requested):
Sry, the last time I was looking on that, I was on phone, not suitable for full review. Some comments inline, hopefully more useful than last time :)

script/foreman-debug
lzap: I don't know why, but trap does not call on script exit. I could not figure it out how to do it, if you can please teach me.

script/foreman-debug
lzap: Added.

script/foreman-debug
lzap: The script keeps the dir if NOTAR is submitted, otherwise it deletes it if it's in /var/tmp. So it actually works the way you like for the most part.

lzap:
Amended.

ares:
@lzap I think rebasing might make upgrade tests green

lzap:
Debug script has no tests, can someone merge manually? I don't want to put more stress on Jenkins just because of that grey button above.

script/foreman-debug
```diff
@@ -246,9 +246,18 @@ else
 fi
 printv "Determined $OS distribution"
 
+clean_temp() {
+  [ "$NOTAR" -eq 1 ] && exit 0
+  # prevent from removing anything user-defined
+  [[ "$DIR" != /var/tmp* ]] && exit 0
+  printv "Cleaning $DIR"
+  rm -rf "$DIR"
+  exit 0
```
iNecas: This will override the original exit code of the code, try:

```bash
#!/usr/bin/bash

function clean_temp () {
  echo "cleaning"
  exit 0
}
trap clean_temp EXIT SIGHUP SIGINT SIGTERM
exit 10
```

Why we are doing it?

script/foreman-debug
iNecas: I have not seen this behavior: was testing this on rhel7 and the trap was working just fine. With this around, it was printing the 'cleaning' twice in the output, which is quite odd. Could you check again we really need this? I would be curious about the reproducer.

script/foreman-debug
iNecas: Scratch my previous comments. Anyway, one thing: you're calling `exit` that affects the exit code of the whole script: return should be used instead I guess.

script/foreman-debug
lzap: Oh good catch, it was a bug.

script/foreman-debug
iNecas: Could you re-test need for this: I have not seen an evidence that this actually needs to be called.

lzap:
This is now ready.

iNecas:
One question still unanswered

lzap:
Removed the line.

iNecas:
So, I've gave it one more try and here are more observations:

1. I've found why it seemed the trap doesn't work at the end https://github.com/Katello/katello-packaging/blob/KATELLO-3.5/katello/katello-debug.sh#L130-L135 - katello overrides the trap by it's own method.

2. the `trap SIGHUP SIGINT SIGTERM` is IMO wrong, as we actually prevent the script to exit (unless calling `exit` explicitly, which is hard to do, as we don't know what exit code to use. My tests showed that `trap EXIT` should work just fine, and the trap is called even when the script is interrupted by `SIGINT`

lzap:
Aaaaaah yeah! Good job finding the `trap` in Katello. Also I agree on signals. Amended change, now I trap only `EXIT` signal and also I am creating `AUXTMP` directory which is also properly cleaned. I will make a PR in Katello to use it instead creating own one.

ares:
[test foreman] getting "ArgumentError: let 'klass' cannot override a method in Minitest::Spec. Please use another name." on all combinations

lzap:
After discussion I am pushing new version that works via TMPDIR. I haven't tested on real setup, but I can confirm it should work.

```
[root@zzzap ~]# mktemp -d
/tmp/tmp.c8UjxXHR8U
[root@zzzap ~]# export TMPDIR=/tmp/test
[root@zzzap ~]# mktemp -d
/tmp/test/tmp.NJn24Ky5rt
```

lzap:
@ares wrong tab? :-) looking at your screen, not surprised.

ares:
nope, this time I hit the right tab :-) tests are still red, rebase should help

lzap:
Did not get it, rebased!

ekohl (approved):
Untested but looks correct

lzap:
Tested works fine:

```
[root@winter ~]# curl -sk https://raw.githubusercontent.com/lzap/foreman-packaging/debug-clean-tmp-21824/katello/katello/katello-debug.sh -o /usr/share/foreman/script/foreman-debug.d/katello-debug.sh
[root@winter ~]# curl -sk https://raw.githubusercontent.com/lzap/foreman/debug-clean-tmp-21824/script/foreman-debug -o /usr/sbin/foreman-debug 
[root@winter ~]# foreman-debug
```

lzap:
Found **blocker bug** when testing, the task "plugin_list" gets stuck, there is some issue with terminal, perhaps it is waiting for some input or something. When I execute it separately via `foreman-rake plugin:list` it's fast (few seconds to boot Rails). Since this smells a lot, I'd like to remove "plugin_list" debug task in this patch, because list of plugins is carried in three other tasks:

```
  add_cmd "gem list" "gem_list"
  add_cmd "scl enable $SCLNAME 'gem list'" "gem_list_scl"
  add_cmd "bundle --local --gemfile=/usr/share/foreman/Gemfile" "bundle_list"
```

What you think?

lzap:
Confirmed, `foreman-rake plugin:list` step in foreman-debug always fails (gets stuck), something is waiting for input, not sure what this is because it works when I start this on regular console. But I removed the line because you can easily get list of plugins from either RPM package list or SCL GEM list. This makes debug a bit faster, because we are not booting Rails at all now. Please test:

    curl -sk https://raw.githubusercontent.com/lzap/foreman-packaging/debug-clean-tmp-21824/katello/katello/katello-debug.sh -o /usr/share/foreman/script/foreman-debug.d/katello-debug.sh
    curl -sk https://raw.githubusercontent.com/lzap/foreman/debug-clean-tmp-21824/script/foreman-debug -o /usr/sbin/foreman-debug 

Compare before/after to see how much faster this is:

    time foreman-debug -v

script/foreman-debug
```diff
@@ -290,7 +305,8 @@ if [ $NOGENERIC -eq 0 ]; then
   add_cmd "ausearch -m AVC -m USER_AVC -m SELINUX_ERR | head -n 100" "selinux_first_denials.log"
   add_cmd "ausearch -m AVC -m USER_AVC -m SELINUX_ERR || grep AVC /var/log/audit/audit.log" "selinux_denials.log"
   if [ -f /usr/sbin/selinuxenabled ] && /usr/sbin/selinuxenabled; then
-    add_cmd "sepolgen-ifgen &>/dev/null && audit2allow -Ra || audit2allow -a" "selinux_audit2allow"
+    [[ -f /var/lib/sepolgen/interface_info ]] || sepolgen-ifgen &>/dev/null
```
iNecas: I would recommend keeping this change in separate issue

script/foreman-debug
```diff
@@ -347,7 +363,6 @@ add_files /etc/{sysconfig,default}/foreman
 add_files /etc/{sysconfig,default}/libvirt*
 add_files /etc/sysconfig/pgsql
 add_files /var/lib/pgsql/data/pg_log/*
-add_cmd "foreman-rake plugin:list" "plugin_list"
```
iNecas: As much as I like this removal, I think we should keep this in separate commit and new issue. I know it looks bureaucratic, but it's very useful later to apply the policy of one issue per commit.

script/foreman-debug
lzap: There's no time, removed.

script/foreman-debug
lzap: There's no time, removed.

lzap:
Amended the two changes which are not relevant.

lzap:
Filed https://github.com/theforeman/foreman/pull/5215 and https://github.com/theforeman/foreman/pull/5216

iNecas:
I wanted to merge with small updated that was discussed offline, but was not allowed to merge that due to protected branch. Pr with updates opened here https://github.com/theforeman/foreman/pull/5220

iNecas (approved):

iNecas:
Closing in favor of https://github.com/theforeman/foreman/pull/5220

PR #5042, opened 2017-11-30
Fixes #21826 - rss notifications are duplicated

boaz0:
From my understanding `item.summary` isn't passed to the created `Notification` which made it weird why it is used when checking if it exists.

cc @ohadlevy

### DIFF
--- a/app/services/ui_notifications/rss_notifications_checker.rb
+++ b/app/services/ui_notifications/rss_notifications_checker.rb
@@ -68,7 +68,7 @@ module UINotifications
     end
 
     def notification_already_exists?(item)
-      !!Notification.unscoped.find_by_message(item.summary)
+      !!Notification.unscoped.find_by_message(item.title)
     end
   end
 end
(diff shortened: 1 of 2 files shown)


ohadlevy:
thanks @ripcurld0! @dLobatog mind having a look please?
@ripcurld0 would you be able to add a test to cover this issue? :)

boaz0:
@ohadlevy sure, I am working on this as we speak.
Although it looks like there was a test case for that: https://github.com/theforeman/foreman/blob/develop/test/unit/ui_notifications/rss_notifications_checker_test.rb#L11-L17 
but to be honest, I am not sure what it exactly does.

boaz0:
TIL how to mock and stub in Ruby.

ares (approved):
Fixes the issue

ares:
Thanks @ripcurld0, fixes the silly issue indeed, merging!

boaz0:
Thank you @ares @ohadlevy

PR #5044, opened 2017-12-01
fixes #21832 - permission test: do not assume required login

timogoebel:
Makes the tests pass when a controller does not require any login, e.g. something like this:

https://github.com/theforeman/foreman_userdata/blob/3f8bef74ba43c091444989fc6a5085a52ca644f3/app/controllers/userdata_controller.rb#L2-L14

### DIFF
--- a/test/unit/shared/access_permissions_test_base.rb
+++ b/test/unit/shared/access_permissions_test_base.rb
@@ -21,10 +21,13 @@ module AccessPermissionsTestBase
           controller = "#{r.defaults[:controller]}_controller".classify.constantize
           filters    = controller.send(:_process_action_callbacks)
 
-          # Pass if the controller deliberately only permit admins (e.g. SettingsController)
-          if filters.select { |f| f.filter == :require_admin }.empty?
-            assert_not_equal [], Foreman::AccessControl.permissions.select { |p| p.actions.include? path }
-          end
+          # Pass if the controller only permit admins (e.g. SettingsController)
+          next if filters.any? { |f| f.filter == :require_admin }
+
+          # Pass if the controller deliberately skips login requirement
+          next if controller < ApplicationController && filters.select { |f| f.filter == :require_login }.empty?
+
+          assert_not_empty Foreman::AccessControl.permissions.select { |p| p.actions.include? path }
         end
       end
     end


test/unit/shared/access_permissions_test_base.rb
```diff
@@ -21,10 +21,15 @@ def check_routes(app_routes, skipped_actions)
...
           filters    = controller.send(:_process_action_callbacks)
 
-          # Pass if the controller deliberately only permit admins (e.g. SettingsController)
-          if filters.select { |f| f.filter == :require_admin }.empty?
-            assert_not_equal [], Foreman::AccessControl.permissions.select { |p| p.actions.include? path }
-          end
+          permission_required = true
+
+          # Pass if the controller only permit admins (e.g. SettingsController)
+          permission_required = false if filters.select { |f| f.filter == :require_admin }.any?
```
tbrisker: you can use `filters.detect{}` instead of `filters.select{}.any?`

test/unit/shared/access_permissions_test_base.rb
```diff
@@ -21,10 +21,15 @@ def check_routes(app_routes, skipped_actions)
...
+
+          # Pass if the controller deliberately skips login requirement
+          permission_required = false if controller < ApplicationController && filters.select { |f| f.filter == :require_login }.empty?
```
tbrisker: not sure why here we check for `ApplicationController` but not for the other one?
Also we could just check both in one run using `f.filter.in? [:require_login, :require_admin]`

tbrisker (changes requested):

test/unit/shared/access_permissions_test_base.rb
timogoebel: The API controllers do not have a `require_login` action as they don't inherit from `ApplicationController`. That's why I had to split that up. The second check is only valid if the first condition is true.

test/unit/shared/access_permissions_test_base.rb
tbrisker: sorry, `filters.any?{}` instead of `filters.select{}.any?` so that the loop stops once that filter is found.

test/unit/shared/access_permissions_test_base.rb
tbrisker: oh, I see, makes sense.

test/unit/shared/access_permissions_test_base.rb
```diff
@@ -21,10 +21,15 @@ def check_routes(app_routes, skipped_actions)
           controller = "#{r.defaults[:controller]}_controller".classify.constantize
           filters    = controller.send(:_process_action_callbacks)
 
-          # Pass if the controller deliberately only permit admins (e.g. SettingsController)
-          if filters.select { |f| f.filter == :require_admin }.empty?
-            assert_not_equal [], Foreman::AccessControl.permissions.select { |p| p.actions.include? path }
-          end
+          permission_required = true
```
tbrisker: maybe just return early instead of having a variable? that way you wouldn't need to do the second check if the first already set the variable to `false`?

tbrisker (changes requested):

test/unit/shared/access_permissions_test_base.rb
timogoebel: hm... an early return gives me "LocalJumpError: unexpected return"... I guess we can't use that. :-(

timogoebel:
@tbrisker: Thanks for your comments. Apparently, we can't use an early return in a test. So I stuck with a variable but included it in the if condition.

test/unit/shared/access_permissions_test_base.rb
tbrisker: OK, figured it out - test are blocks, not functions, so to return early you need to use `next` instead of `return`

timogoebel: Duh, makes sense. Thanks.

timogoebel:
@tbrisker: Changed accordingly. Thanks again.

tbrisker (approved):
LGTM, will merge when 💚

tbrisker:
Thanks @timogoebel !

PR #5047, opened 2017-12-03
Fixes #21846 - Notification drawer outside click

sharvit:
Fix a bug in the notification drawer

Notification drawer should get closed properly when
clicking outside the drawer

I removed the `react-onclickoutside` package and implemented it inside the component.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,6 @@
     "react-bootstrap": "^0.31.0",
     "react-dom": "^16.0.0",
     "react-numeric-input": "^2.0.7",
-    "react-onclickoutside": "^6.6.2",
     "react-redux": "^5.0.2",
     "redux": "^3.6.0",
     "redux-form": "^7.0.3",
--- a/webpack/assets/javascripts/react_app/components/notifications/__snapshots__/notifications.test.js.snap
+++ b/webpack/assets/javascripts/react_app/components/notifications/__snapshots__/notifications.test.js.snap
@@ -1,24 +1,14 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`notifications empty state 1`] = `
-<OnClickOutside(notificationContainer)
-  eventTypes={
-    Array [
-      "mousedown",
-      "touchstart",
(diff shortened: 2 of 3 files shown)


ohadlevy (commented):
What is the difference between the implentations? E.g. why didn't it work before and can we test it better now? Thanks

sharvit:
@ohadlevy - I can't really put my finger on a specific issue with `react-onclickoutside`.
Especially because it feels like the library is stable and commonly used.
I got a strong feeling it is somehow related to turbolinks.

* I could also notice that `react-onclickoutside` keep firing up the outside click event when the drawer is closed
* At the dashboard, when you toggle the auto refresh, it will immediately make the bug happen (turbolinks?)

Maybe it happen because `react-onclickoutside` attach a uid to the component at the constructor.

Implementation differences:
* I assign the events to the document at the capturing phase.
* I unassign the events after closing the drawer
* I am using `this.containerRef.contains(e.target)` to check whether the click is inside or outside the component while `react-onclickoutside` use a `while loop` from the clicked node to the root node (until no more parents) and use a special class name to indicate about the container element. https://github.com/Pomax/react-onclickoutside/blob/master/src/dom-helpers.js

About the tests, after I looked into the snapshot file I can see that the tests didn't really work and the drawer stayed open.
In my implementations, it is also happening because it uses the `document.addEventListener` while the enzyme doesn't attach the components to the dom as I think I understand.

ohadlevy:
@sharvit thanks.

1. it feels like this is a generic problem - if we don't want to send a pr to the original author, maybe we should consider creating a component that simply wraps it (so it can be reused)?
2. how about adding tests? maybe in this case an integration test is needed?

sharvit:
@ohadlevy thanks for your feedback.

1. Will create a wrapper component, make a lot of sense.
2. will try to add an integration test

sharvit:
@ohadlevy i choose to close this pr and continue here https://github.com/theforeman/foreman/pull/5083

Thanks for pushing me to dig deeper :+1:

PR #5048, opened 2017-12-03
fixes #3763 - extend rabl templates

timogoebel:
This should be considered a proposal on how it would be possible for plugins to extend a rabl template to add more attributes to APIv2 responses.

This is still missing tests and inline documentation. I'd love to hear some feedback on this first.

It can be used like this:

lib/foreman_expire_hosts/engine.rb
```ruby
Foreman::Plugin.register :foreman_expire_hosts do
  [...]
  extend_rabl_template 'api/v2/hosts/main', 'api/v2/hosts/expiration'
end
```

app/views/api/v2/hosts/expiration.json.rabl:
```ruby
attribute :expired_on
```

### DIFF
--- a/config/initializers/rabl_init.rb
+++ b/config/initializers/rabl_init.rb
@@ -35,9 +35,42 @@ module Foreman
       # otherwise return super since v1 has object root (config.include_child_root = true)
       super
     end
+
+    # Lists all extension templates that are defined in plugins
+    def template_extensions_from_plugins
+      return [] unless @_options[:template]
+      Foreman::Plugin.all.flat_map {|plugin| plugin.rabl_template_extensions(@_options[:template]) }
+    end
+
+    # extends the current template with templates defined in plugins
+    def load_extensions_from_plugins
+      template_extensions_from_plugins.each do |extension|
+        extends extension unless @_settings[:extends].map { |e| e[:file] }.include?(extension) || @_options[:template] == extension
+      end
+    end
+
+    def render(*args, &block)
+      load_extensions_from_plugins
+      super
+    end
+  end
+
+  module RablTemplateHandlerExt
+    # This extends the rabl code to store the template path
+    # while initializing the engine.
+    # This allows us to find extension template from plugins
+    # that match the template name
+    def call(template)
+      source = template.source
+
+      %{ ::Rabl::Engine.new(#{source.inspect}, :template => '#{template.virtual_path}').
+                                          apply(self, assigns.merge(local_assigns)).
+                                                        render }
+    end
   end
 end
 Rabl::Engine.send(:prepend, Foreman::RablEngineExt)
+ActionView::Template::Handlers::Rabl.singleton_class.send(:prepend, Foreman::RablTemplateHandlerExt)
 
 Rabl.configure do |config|
   # Commented as these are defaults
--- a/app/registries/foreman/plugin.rb
+++ b/app/registries/foreman/plugin.rb
@@ -117,6 +117,7 @@ module Foreman #:nodoc:
       @smart_proxies = {}
       @controller_action_scopes = {}
       @dashboard_widgets = []
+      @rabl_template_extensions = {}
     end
 
     def after_initialize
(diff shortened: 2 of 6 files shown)


ohadlevy:
:+1:  overall, lets just be careful not to crate N+1 issues, AFAIR @ShimShtein already solved a similar problem before with the ability to enhanced controller eager loading?

config/initializers/rabl_init.rb
```diff
@@ -35,9 +35,37 @@ def data_name(data_token)
...
+      load_extensions_from_plugins
+      super
+    end
+  end
+
+  module RablTemplateHandlerExt
+    def call(template)
+      source = template.source
+
+      %{ ::Rabl::Engine.new(#{source.inspect}, :template => '#{template.virtual_path}').
```
iNecas: Why do we need this?

ShimShtein:
Indeed, N+1 is already [solved for hosts](https://github.com/theforeman/foreman/blob/develop/app/controllers/api/v2/hosts_controller.rb#L23). BTW, facets are added to RABL automatically.

config/initializers/rabl_init.rb
```diff
@@ -35,9 +35,37 @@ def data_name(data_token)
...
+                                          apply(self, assigns.merge(local_assigns)).
+                                                        render }
+    end
   end
 end
 Rabl::Engine.send(:prepend, Foreman::RablEngineExt)
+ActionView::Template::Handlers::Rabl.singleton_class.send(:prepend, Foreman::RablTemplateHandlerExt)
```
ShimShtein, line 73: Did you consider using RABL's `partial` call as I do [with facets](https://github.com/theforeman/foreman/blob/4d563b215b72081bb7b63f292bd297438192e462/app/views/api/v2/hosts/main.json.rabl#L56)?

ShimShtein (changes requested):

config/initializers/rabl_init.rb
timogoebel: ... to pass the template name (`template.virtual_path`) to the Engine so we can use it in `template_extensions_from_plugins `. An alternative to this would be `@_options[:template] || context_scope.virtual_path`.

config/initializers/rabl_init.rb
timogoebel: I believe we would get similar results. What do you think could be gained from this?

ShimShtein: Easier to read, less monkey patching?

timogoebel: Ok, so you are proposing to explicitly add extension points to every template. I thought you were talking about `extends` vs. `partial`. I prefer this approach as it's way more flexible.

ShimShtein: I see that we already overload the `Engine` class anyway.

ShimShtein (approved):

ShimShtein:
Maybe we should move facets to use this framework too, if it's already there - less explicit code in host's rabl.

I mean changing the [facet definition](https://github.com/theforeman/foreman/blob/a4bf52491fe5eeceb8606a540e25bc9ceea8a7ba/app/services/facets/entry.rb#L51) to:
``` ruby
    def api_view(view_templates)
      extend_api_template 'api/v2/hosts/main', view_templates[:single]
      extend_api_template 'api/v2/hosts/list', view_templates[:list]
    end
```

timogoebel:
@ShimShtein: Sure, we can do that. But I'd prefer to do this in another PR. It's just a nice to have imho.

All right, I'll take this as a go. Thanks for the fast replies, appreciate it. I'll continue working on this and will let you know once this is ready for a full review.

config/initializers/rabl_init.rb
iNecas: Any downside of the non-monkey patching version? That would be the option of my choice

timogoebel: I agree that this monkey patching is not ideal, but I am not sure if `context_scope.virtual_path` is always the same as the template name. I think, it just refers to the controller + controller action. That is equal to the template name in most cases, but not always. Or am I missing something here?

iNecas: I see now why this is needed. Thanks for explanation

timogoebel:
@ShimShtein, @iNecas et all:

* Renamed the plugin extension point
* Added some inline code comments
* Added tests

Happy to hear more comments. If we could get this into 1.17 this would be awesome for plugin authors. :-)

ShimShtein:
Thanks @timogoebel!

Could you please update the [wiki page](http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin) accordingly?

PR #5051, opened 2017-12-04
Fixes #21851 - Remove deprecations for 1.17

xprazak2:


ares (dismissed):
There's one more deprecation in seeds https://github.com/theforeman/foreman/blob/develop/db/seeds.rb#L16

otherwise seems correct and I guess we can merge it before branching itself

xprazak2:
I missed that one, fixed now.

app/models/concerns/host_common.rb
```diff
@@ -154,12 +154,10 @@ def crypt_root_pass
   end
 
   def param_true?(name)
-    Foreman::Deprecation.renderer_deprecation('1.17', __method__, 'host_param_true?')
```
tbrisker: shouldn't this method be removed completely if it's deprecated rather then the deprecation?

tbrisker (changes requested):
Also js deprecations need to be updated, any method using `tfm.tools.deprecate` should set the release explicitly and the default release for deprecation should be updated to 1.19.

app/models/concerns/host_common.rb
xprazak2: This method is deprecated only for use in templates, we can still use it internally.

xprazak2:
The only `tfm.tools.deprecate` I found are those for [compute resource](https://github.com/theforeman/foreman/blob/develop/app/assets/javascripts/compute_resource.js) which are targeted for 1.18

xprazak2:
[test katello]

app/models/concerns/host_common.rb
tbrisker: Do we? I couldn't find anywhere we do, no point in keeping around unused methods that were deprecated (unlike a few other deprecations this one wasn't only in the renderer). Also should be removed from host and hostgroup jail and tests.

app/models/concerns/host_info_extensions.rb
```diff
@@ -22,11 +22,6 @@ module HostInfoExtensions
 
   # This is the base method that gathers information from all providers.
   def info
-    renderer_regex = /renderer\.rb.*host_enc/
-    unless caller.first.match(renderer_regex) || caller[1].match(renderer_regex)
-      Foreman::Deprecation.renderer_deprecation('1.17', __method__, 'host_enc')
-    end
```
tbrisker, line 28: Should this be removed from the jail as well since it shouldn't be used in templates any more?

app/models/concerns/host_params.rb
```diff
@@ -9,7 +9,6 @@ module HostParams
     attr_reader :cached_host_params
 
     def params
-      Foreman::Deprecation.renderer_deprecation('1.17', __method__, 'host_param') unless caller.first =~ /renderer\.rb.*host_param/
```
tbrisker, line 12: ditto

app/models/hostgroup.rb
```diff
@@ -96,8 +96,8 @@ def self.search_by_config_group(key, operator, value)
 
   class Jail < Safemode::Jail
     allow :name, :diskLayout, :puppetmaster, :operatingsystem, :architecture,
-      :environment, :ptable, :url_for_boot, :params, :puppetproxy, :param_true?,
```
tbrisker, line 99: `:params` here is a different function than the one that was deprecated

tbrisker (approved):
LGTM, let's see what the butler says. @ares any other comments?

ares:
LGTM too, I guess no plugin used `param_true?`, foreman tests are failing though :-/ setting WoC

xprazak2:
Failing on foreman_tasks using `audit_modified?` instead of `SeedHelper.audit_modified?`

xprazak2:
[test katello]

tbrisker:
Thanks @xprazak2 and @ares !

PR #5054, opened 2017-12-05
WIP: Fixes #21860 - Update jquery to version 3

sharvit:
Things that block us from upgrading to jquery v3:
- [ ] [flot](https://github.com/flot/flot) - charts package
    We should be able to get ride of it soon and use react charts once it will be supported by `patternfly-react`
    https://github.com/patternfly/patternfly-react/pull/103
- [ ] `jquery-ui` - we are currently using `jquery-ui-rails@4.x` (latest is 6.x) which bring the `jquery-ui@1.10.4` (latest is 1.12.1)
    - [ ] `jquery-ui/autocomplete` - We will be able to get rid of it once http://projects.theforeman.org/issues/21604 is done
    - [ ] `jquery-ui/spinner` - We will be able to get rid of it once http://projects.theforeman.org/issues/21918 is done


tbrisker:
One thing I noticed broke with this is jquery-ui (in the autocomplete). other js might also be broken, i did not go through all the pages and I feel we don't have good coverage of everything to trust just tests.

tbrisker:
I would also suggest testing some of the older js libs we use such as gridster and flot, iirc when i tried upgrading a year or so ago some of them broke. Perhaps try using the jquery-migration library to get warnings for everything that will break and go over all the pages (i'm fairly certain there is a lot of stuff that will break, e.g. `attr` instead of `prop` etc).

sharvit:
* Updated `jquery-ui-rails` to v6 (using `jquery-ui@1.12.1`).
* Added the `jquery-migration`.

WIP:
* Looking for more packages that getting broken.
* Moving over the deprecated warnings from `jquery-migrate`

tbrisker:
@sharvit I think it would be better to completely get rid of jquery-ui in core rather then update it. Katello uses some widgets that break when upgrading the core version.
see also projects.theforeman.org/issues/21640

sharvit:
Yes @tbrisker, i updated the pr description.
Basically it is only about to finish those features and we ready to remove jquery-ui
http://projects.theforeman.org/issues/21604
http://projects.theforeman.org/issues/21918

dLobatog (changes requested):
@sharvit Is this relevant anymore now that we use axios?

sharvit:
@dLobatog Yes, we are still using jQuery in many places.

In my vision, i want to get rid of jQuery entirely.
Before that, it would be nice to update jQuery to it's latest version (which I'm still having troubles to update to).

xprazak2:
@sharvit, could you rebase pls?

PR #5055, opened 2017-12-05
Fixes #21867 - Restore context when leaving tax wizard

xprazak2:


app/controllers/concerns/application_shared.rb
```diff
@@ -24,6 +24,7 @@ def current_permission
 
   def set_taxonomy
     user = User.current
+    TopbarSweeper.expire_cache
```
ares: I think this would be better to be moved one line above, it does not depend on user, `expire_cache` has it's own check for User.current being nil

ares (changes requested):
one minor thing, then it's good to go, indeed fixes the problem

ares (approved):

ares:
Thanks @xprazak2, merging!

PR #5058, opened 2017-12-06
Fixes #21882 - extract yes/no values in settings

amirfefer:


tbrisker (approved):
test failure unrelated. Thanks @amirfefer !

PR #5060, opened 2017-12-06
Fixes #21878 - Audit Puppet Environment

swapab:


test/models/environment_test.rb
```diff
@@ -20,4 +20,14 @@ class EnvironmentTest < ActiveSupport::TestCase
   test 'should create environment with the name "new"' do
     assert FactoryBot.build(:environment, :name => 'new').valid?
   end
+
+  context 'audited' do
+    test 'on creation on of a new environment' do
+      environment = Environment.new(:name => 'new_env')
```
tbrisker: Please use FactoryBot to ensure isolation between tests.

tbrisker (dismissed):
Just a minor comment, thanks!

test/models/environment_test.rb
swapab: I initially tried that but seems like FactoryBot doesn't creates `audits`.
```
[1] pry(#<EnvironmentTest::audited>)> environment = FactoryBot.create(:environment)
=> #<Environment:0x000000065cbd58
 id: 687036938,
 name: "environment1",
 created_at: Thu, 07 Dec 2017 17:22:14 UTC +00:00,
 updated_at: Thu, 07 Dec 2017 17:22:14 UTC +00:00,
 katello_id: nil>
[2] pry(#<EnvironmentTest::audited>)> environment.valid?
=> true
[3] pry(#<EnvironmentTest::audited>)> environment.audits
=> []
```

tbrisker: Audits on factories are disabled by default, but can be enabled using the `:with_auditing` option, e.g. `FactoryBot.create(:environment, :with_auditing)`

ares (approved):
ack pending jenkins

test/models/environment_test.rb
swapab: @tbrisker Thanks for the input. `:with_auditing` trait is nice. And great, auditing is disabled by default (`test/factories/disable_auditing.rb`)

ares:
Thanks @swapab, merging!

PR #5061, opened 2017-12-06
Fixes #21883 - remove Current user INFO messages

lzap:
This patch removes this

```
2017-12-04 04:00:16  [app] [I] Current user: foreman_admin (administrator)
2017-12-04 04:00:16  [app] [I] Current user: foreman_admin (administrator)
2017-12-04 04:00:16  [app] [I] Current user: foreman_admin (administrator)
2017-12-04 04:00:16  [app] [I] Current user: foreman_admin (administrator)
2017-12-04 04:00:16  [app] [I] Current user: foreman_admin (administrator)
2017-12-04 04:00:23  [app] [I] Current user: foreman_admin (administrator)
2017-12-04 04:00:23  [app] [I] Current user: foreman_admin (administrator)
```

it will still be logged into DEBUG level. User login and admin flag are now always logged per each individual log line and this can be turned on to have along every line via changing logging pattern.

### DIFF
--- a/app/models/concerns/foreman/thread_session.rb
+++ b/app/models/concerns/foreman/thread_session.rb
@@ -70,10 +70,16 @@ module Foreman
           end
 
           if o.is_a?(User)
-            Rails.logger.info "Current user: #{o.login} (#{o.admin? ? 'administrator' : 'regular user'})"
+            user = o.login
+            type = o.admin? ? 'admin' : 'regular'
+            if o.hidden?
+              Rails.logger.debug("Current user set to #{user} (#{type})")
+            else
+              Rails.logger.info("Current user set to #{user} (#{type})")
+            end
           end
-          Rails.logger.debug "Setting current user thread-local variable to " + (o.is_a?(User) ? o.login : 'nil')
-          ::Logging.mdc['user_login'] = o.try(:login)
+          ::Logging.mdc['user_login'] = o&.login
+          ::Logging.mdc['user_admin'] = o&.admin? || false
           Thread.current[:user] = o
         end
 
@@ -116,7 +122,7 @@ module Foreman
             raise(ArgumentError, "Unable to set current organization, expected class '#{self}', got #{organization.inspect}")
           end
 
-          Rails.logger.debug "Setting current organization thread-local variable to #{organization || 'none'}"
+          Rails.logger.debug "Current organization set to #{organization || 'none'}"
           org_id = organization.try(:id)
           ::Logging.mdc['org_id'] = org_id if org_id
           Thread.current[:organization] = organization
@@ -153,7 +159,7 @@ module Foreman
             raise(ArgumentError, "Unable to set current location, expected class '#{self}'. got #{location.inspect}")
           end
 
-          Rails.logger.debug "Setting current location thread-local variable to #{location || 'none'}"
+          Rails.logger.debug "Current location set to #{location || 'none'}"
           loc_id = location.try(:id)
           ::Logging.mdc['loc_id'] = loc_id if loc_id
           Thread.current[:location] = location


app/models/concerns/foreman/thread_session.rb
```diff
@@ -69,10 +69,15 @@ def current=(o)
             raise(ArgumentError, "Unable to set current User, expected class '#{self}', got #{o.inspect}")
           end
 
-          if o.is_a?(User)
-            Rails.logger.info "Current user: #{o.login} (#{o.admin? ? 'administrator' : 'regular user'})"
+          Rails.logger.debug?
+            if o.is_a?(User)
```
tbrisker: how about checking something like `if o.is_a?(User) && !o.hidden?`? this would prevent logging the system accounts.

ares: In such case, please also change the log message, otherwise we'll get report why some users are logged as info and why some as debug. Especially when we'll have syslog in place.

app/models/concerns/foreman/thread_session.rb
```diff
@@ -69,10 +69,15 @@ def current=(o)
             raise(ArgumentError, "Unable to set current User, expected class '#{self}', got #{o.inspect}")
           end
 
-          if o.is_a?(User)
-            Rails.logger.info "Current user: #{o.login} (#{o.admin? ? 'administrator' : 'regular user'})"
+          Rails.logger.debug?
```
ares: This should be written as 

ˋˋˋ
Rails.logger.debug do 
  ... this happens only in debug log level
end
```

**Review by ares — CHANGES_REQUESTED** [review]
I think the current user is important information so I'd prefer to keep it as info all the time. From the log you attached to the case, it's not clear why we have multiple records for single action. I'd suggest looking into why that happened and fiximg the cause instead not logging it on info level. As long as we have one entry per request, I think it should be part of production.log by default.

**Comment by lzap** [comment]
I rebased, here is my reasoning.

1) We already log taxonomy with debug level, my change makes all logging consistent. I also made all three logging messages consistent strings: _Current XXX set to YYY_.

2) Having something an "important" info does not give you the right to put this into info level, it's just fine having it in debug level I think.

3) I do not think that hiding logging lines for hidden user is a good way to go.

4) In syslog structured logging (my upcoming work for next year) we will have user ID logged with every single request by default (so we can easily filter them out), therefore these debug messages will be less relevant in the future

**Comment by ares** [comment]
I'm mostly fine with your arguments, I'm happy to defer to others re debug vs info level, though I think it should be same level as incoming params, which is info.

But I think this PR does not solve the cause of original problem reported in redmime issue. Why it's logged multiple times in a row? Is that just some specific action? We should deduplicate it to fix that issue. Or is the log just what you grepped out? It seems the timestamp is same for 5 logs, so I assume it was the same request.

**Comment by lzap** [comment]
I am not trying to solve multiple occurrences of this log, but simply the fact is is present. I believe it is coming from migration or rake task, this is no an ordinary request log.

**Comment by ares** [comment]
@iNecas mind to share your view on this? I think you recently extended current user information logging.

**Comment by lzap** [comment]
I also added statements to prevent double logging, but please keep in mind these will only work if we have nested block like:

```ruby
User.as_anonymous_admin do
  User.as_anonymous_admin do
    # blah
  end
end
```

Non-nested uses (and we have many of them) will still log set and unset messages.

tbrisker:
Why do we need to log internal system user context switches? Users shouldn't care about what happends inside the code, only when a real user does some action.

iNecas (changes requested):
I agree with @ares, that this is an important piece of information we don't have a way how to get otherwise, was added on purpose and it's covered by customers request (http://projects.theforeman.org/issues/18949).
I agree this whould be `info`.

I agree with @tbrisker that for internal admin, it's not that hot. Therefore, we should either:

1. debug if we are in rake task
2. debug if `user.hidden? || user.internal?`. bonus points for making the `hidden?` and `internal?` methods more effective (as currently, it would mean that we would always need to load the auth source from db, which, although not too bad, can be suboptimal for checking on every context switch.

lzap:
Rebased @iNecas @ares @tbrisker 

The patch solves two issues.

* prevents from multiple logging of context switch when the value did not actually change
* moves this from INFO to DEBUG level (log lines already carries user in extra field)

iNecas:
@lzap there is no information in the extra field about the fact, if the user is regular or admin: since it's really useful info, I still vote for keeping that `info`.

Also, is this actually fixing the multiple logging issue?, I still see this:

```
2018-06-26T18:02:14 [D|app|] Current user set to foreman_admin (admin)
2018-06-26T18:02:14 [D|app|] <TrueClass> true
2018-06-26T18:02:14 [D|app|] Current user unset
2018-06-26T18:02:14 [D|app|] <TrueClass> true
2018-06-26T18:02:14 [D|app|] Current user set to foreman_admin (admin)
2018-06-26T18:02:14 [D|app|] <TrueClass> true
2018-06-26T18:02:14 [D|app|] Current user unset
2018-06-26T18:02:14 [D|app|] <TrueClass> true
2018-06-26T18:02:14 [D|app|] Registering 4 assets for plugin foreman-tasks precompilation
2018-06-26T18:02:15 [D|app|] Current user set to foreman_admin (admin)
2018-06-26T18:02:15 [D|app|] <TrueClass> true
2018-06-26T18:02:15 [D|app|] Current user unset
2018-06-26T18:02:15 [D|app|] <TrueClass> true
2018-06-26T18:02:15 [D|app|] Current user set to foreman_admin (admin)
2018-06-26T18:02:15 [D|app|] <TrueClass> true
2018-06-26T18:02:15 [D|app|] Current user unset
2018-06-26T18:02:15 [D|app|] <TrueClass> true
2018-06-26T18:02:15 [D|app|] Registering 7 assets for plugin foreman_remote_execution precompilation
2018-06-26T18:02:15 [D|app|] Current user set to foreman_admin (admin)
2018-06-26T18:02:15 [D|app|] <TrueClass> true
2018-06-26T18:02:15 [D|app|] Current user unset
2018-06-26T18:02:15 [D|app|] <TrueClass> true
```

The thing is probably, that we set and unset the user several times during the rake task, so the additional check doesn't actually filter out anything.

app/models/concerns/foreman/thread_session.rb
```diff
@@ -69,10 +69,15 @@ def current=(o)
             raise(ArgumentError, "Unable to set current User, expected class '#{self}', got #{o.inspect}")
           end
 
-          if o.is_a?(User)
-            Rails.logger.info "Current user: #{o.login} (#{o.admin? ? 'administrator' : 'regular user'})"
+          Rails.logger.debug do
```
iNecas: FYI: this line logs `[D|app|] <TrueClass> true`. However, even when updated, I'm not sure this PR actually fixes anything.

iNecas:
I see https://github.com/theforeman/foreman/pull/5061#issuecomment-350244870, so this is known issue.

What about logging the internal users as `debug` bug regular users still as `info`? Or even not logging the switches to internal users?

iNecas:
Basically what @tbrisker says in https://github.com/theforeman/foreman/pull/5061#issuecomment-350671720

app/models/concerns/foreman/thread_session.rb
lzap: Oh this was meant to be `debug? do`.

iNecas:
So, to wrap up my preference: info for non-internal users, debug for the internal one.

lzap:
Ok vastly simplified, this is now as simple as INFO -> DEBUG.

iNecas:
I still suggest `info` for non-internal users.

lzap:
There is no excuse for this now IMHO, when we have structured logging. If you think we need that, file PR into installer to set pattern to include this bit of information with every single log line. We already do this for session/request id.

lzap:
You know what, I added internal check it's not big deal if you really want this. Just one extra line per request hopefully.

iNecas:
AFAIK structured logging doesn’t give usthe info whether the user is admin or a regular one

lzap:
I actually added this, was a good point. It's there now. I am still seeking formal approval :-)

Test failure because of https://github.com/theforeman/foreman/pull/5757

app/models/concerns/foreman/thread_session.rb
```diff
@@ -69,11 +69,17 @@ def current=(o)
             raise(ArgumentError, "Unable to set current User, expected class '#{self}', got #{o.inspect}")
           end
 
-          if o.is_a?(User)
-            Rails.logger.info "Current user: #{o.login} (#{o.admin? ? 'administrator' : 'regular user'})"
+          if Rails.logger.debug? && o.is_a?(User)
```
iNecas: This would still prevent anything showing in info. You probably want to remove the condition on `debug?`

lzap:
Oh right fixed.

app/models/concerns/foreman/thread_session.rb
```diff
@@ -70,10 +70,16 @@ def current=(o)
           end
 
           if o.is_a?(User)
-            Rails.logger.info "Current user: #{o.login} (#{o.admin? ? 'administrator' : 'regular user'})"
+            user = o.login
+            type = o.admin? ? 'admin' : 'regular'
+            if o.internal?
```
iNecas: Actually, you want to check on `hidden?`, as internal means it's not coming from LDAP.

lzap:
Holy, rebased.

iNecas (approved):
Thanks @lzap

iNecas:
@ares no other comments from  me, tested and works well this time.

ares (commented):
Ok, I have no further comments, @iNecas since you did the full review, mind to merge?

iNecas:
@ares  Github blocks the merge, if some review is red: therefore I need you to approve it first.

ares:
I was trying to reset my review with previous comment, but it seems new review "comment" only does not override red one. Approving now, sorry.

ares (approved):

iNecas:
Thanks @lzap

PR #5062, opened 2017-12-06
Fixes #21881 - turn off notification logger in example YAML

lzap:
Second part of this change would be in Puppet.

### DIFF
--- a/config/settings.yaml.example
+++ b/config/settings.yaml.example
@@ -45,7 +45,9 @@
 #    :type: syslog
 
 # Individual logging types can be toggled on/off here
-#:loggers:
+:loggers:
+  :notifications:
+    :enabled: false
 #  :app:
 #    :enabled: true
 #  :audit:
@@ -60,7 +62,6 @@
 #    :enabled: false
 #  :templates:
 #    :enabled: true
-#  :notifications:
 #  :background:
 #    :enabled: true
 #  :dynflow:


iNecas (changes requested):
works for me

iNecas (approved):

lzap:
Tests are green, but the more important part of this is to get this disabled in production. @ekohl I looked into the Puppet template and this is way too complicated for me, would you give me a hand? TY!

ares:
Thanks for looking into this. I don't care that much about example config file, but in the configuration we deploy, I think all loggers should be enabled, those that we believe are less important should log on debug level. It's hard to tell users, please enable debug log level, enable sql logger, permissions logger and notification logger and send us the log. In slightly worse case, we can keep them disabled but they need to be at least listed there. I wish foreman-debug could enable all loggers.

lzap:
We do have foreman-admin tool, logger module can enable/disable debug levels, but not individual loggers. The good thing is those loggers are configurable via puppet installer today, so users can easily change their mind.

Logging a several every second for each user session is insane setting. And we can't easily get rid of Rails INFO logs per request. If you don't like it, I will just close this one then.

lzap:
Ok I am closing this one, this was misunderstanding. I want to get rid of this:

```
[app|I] Started GET "/notification_recipients" for 127.0.0.1 at 2017-12-11 14:39:20 +0100
[app|I] Processing by NotificationRecipientsController#index as JSON
[app|I] Completed 200 OK in 3ms (Views: 0.1ms | ActiveRecord: 0.1ms)
```

These are all provided by Rails and can't be turned off.

PR #5063, opened 2017-12-06
Refs #21238 - Translate setting tooltip

tbrisker:
Also display default for array type settings.

### DIFF
--- a/app/models/setting.rb
+++ b/app/models/setting.rb
@@ -286,7 +286,7 @@ class Setting < ApplicationRecord
   # End methods for loading default settings
 
   def full_name_with_default
-    "#{full_name} (Default: #{default})"
+    _("%{full_name} (Default: %{default})") % {full_name: _(full_name), default: default}
   end
 
   private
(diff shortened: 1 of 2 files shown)


app/models/setting.rb
```diff
@@ -286,7 +286,7 @@ def self.column_check(opts)
   # End methods for loading default settings
 
   def full_name_with_default
-    "#{full_name} (Default: #{default})"
+    _("%{full_name} (Default: %{default})") % {full_name: _(full_name), default: default}
```
akofink, line 289: do you need to have the translation for `_(full_name)` if the whole thing is translated?

akofink (commented):
Seems fine to me. Just one question.

app/models/setting.rb
tbrisker: yes, the full_name is extracted separately ins the setting definition. the translation for the whole string is actually just to get the default translated in context correctly

akofink (approved):
APT!

boaz0 (approved):
LGTM

timogoebel (approved):

timogoebel:
Merged, thanks @tbrisker!

And thanks @akofink & @ripcurld0.

PR #5065, opened 2017-12-07
Fix all Rails 5.0 deprecations

mmoll:


mmoll:
depends on GH-4867 (and includes it)

xprazak2:
I tested, works. Nicely done!

app/models/role.rb
```diff
@@ -289,8 +289,10 @@ def allowed_actions
   end
 
   def check_deletable
-    errors.add(:base, _("Cannot delete built-in role")) if builtin?
-    errors.empty?
+    if builtin?
+      errors.add(:base, _("Cannot delete built-in role")) if builtin?
```
ohadlevy: no need for another if condition?

mmoll: I'd say no, according to the tests and the naming of the method. ;)

ohadlevy: but you have the same if a line above?

mmoll: fixed.

mmoll:
Katello failures seem due to https://github.com/Katello/katello/pull/7103 not yet beeing merged there, the rest is green, so this is ready to go, IMHO

mmoll:
[test katello]

mmoll:
all green now 💚

ohadlevy (approved):

ohadlevy:
thank you @mmoll! and of course credit to @domcleal  as well.

PR #5066, opened 2017-12-07
Fixes #21748 - Removed gravatar images

dmitri-d:
Removed gravatar images as gravatar urls use MD5 hashes

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -371,21 +371,9 @@ module ApplicationHelper
   def avatar_image_tag(user, html_options = {})
     if user.avatar_hash.present?
       image_tag("avatars/#{user.avatar_hash}.jpg", html_options)
-    elsif Setting[:use_gravatar] && user.mail.present?
-      image_tag(gravatar_image(user.mail),
-                html_options.merge!(gravatar_html_options))
     else
       icon_text("user #{html_options[:class]}", "", :kind => "fa")
     end
-  end
-
-  def gravatar_image(email)
-    "#{request.protocol}secure.gravatar.com/avatar/#{Digest::MD5.hexdigest(email.downcase)}?d=mm&s=30"
-  end
-
-  def gravatar_html_options
-    { :onerror => "$(this).addClass('fa fa-user')",
-      :alt     => '' }
   end
 
   def readonly_field(object, property, options = {})
--- a/db/migrate/20171208113210_remove_use_gravatar_setting.rb
+++ b/db/migrate/20171208113210_remove_use_gravatar_setting.rb
@@ -1,0 +1,10 @@
+class RemoveUseGravatarSetting < ActiveRecord::Migration[4.2]
+  def up
+    Setting.where(:name => 'use_gravatar').delete_all
+  end
+
+  def down
+    Setting.create!(:name => 'use_gravatar', :description => N_("Foreman will use gravatar to display user icons"),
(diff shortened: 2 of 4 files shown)


timogoebel:
@witlessbird: Thanks. I don't really care for the gravatar feature, but why isn't disabling this setting enough to comply with FIPS? Can we add a feature toggle "FIPS enabled yes/no" that just disabled this code?

https://github.com/theforeman/foreman/blob/8195b2557b5db1b9a6489e6da6b394e0cec56630/app/models/setting/general.rb#L17

Anyway, if we remove this completely, the setting has to be removed as well. I think, this needs to be done with a db a migration.

dmitri-d:
- Removed 'use_gravatar' setting from the db.

> why isn't disabling this setting enough to comply with FIPS? Can we add a feature toggle "FIPS enabled yes/no" that just disabled this code?

My thinking is that re-enabling gravatar images on a system with fips enabled will lead to a crash (with minimal logging). Adding a check for fips mode isn't worth the trouble in my opinion.

dmitri-d:
Anyone?

app/views/users/index.html.erb
```diff
@@ -19,7 +19,7 @@
   <tbody>
     <% @users.each do |user| %>
       <tr>
-        <td class="display-two-pane ellipsis"><%= avatar_image_tag user, :class => "avatar" %> <%=link_to_if_authorized user.login, hash_for_edit_user_path(:id => user).merge(:auth_object => user, :authorizer => authorizer) %></td>
+        <td class="display-two-pane ellipsis"><%= user.mail %> <%=link_to_if_authorized user.login, hash_for_edit_user_path(:id => user).merge(:auth_object => user, :authorizer => authorizer) %></td>
```
iNecas: The email here looks weird + it's a duplicate info, as we already have the email column.

avatar-4

app/helpers/application_helper.rb
```diff
@@ -368,26 +368,6 @@ def action_buttons(*args)
     end
   end
 
-  def avatar_image_tag(user, html_options = {})
-    if user.avatar_hash.present?
-      image_tag("avatars/#{user.avatar_hash}.jpg", html_options)
```
iNecas: The avatar_hash is actually coming from `ldap` and there is probably no reason removing this path.

app/views/users/index.html.erb
iNecas: And since the ldap avatar support should IMO stay, I think we can leave this untouched for now.

dmitri-d:
* restored avatar_image_tag helper

db/migrate/20171208113210_remove_use_gravatar_setting.rb
```diff
@@ -0,0 +1,10 @@
+class RemoveUseGravatarSetting < ActiveRecord::Migration
```
iNecas: Since Foreman moved to rails 5.1 in the meantime, this needs to be:

```ruby
class RemoveUseGravatarSetting < ActiveRecord::Migration[4.2]
```

iNecas (changes requested):
Tested and works well. The only request is to rebase and update the syntax for migration definition

dmitri-d:
* rebased
* updated migration for Rails 5.1

iNecas (approved):

iNecas:
Thanks @witlessbird

PR #5068, opened 2017-12-10
Fixes #11389 - Remove API v1

tbrisker:
API v1 has been deprecated since Foreman 1.9 - over 2 years ago.
This is more then enough time for anyone needing to migrate to v2.
Even though the original issue requests extraction to a plugin, I do not
see any point in investing the time and effort to maintain a long
deprecated API.


boaz0 (commented):
SGTM

lzap:
I am all for dropping it but I remember @ohadlevy preferred putting this into plugin. But frankly, I only agree with this if there is someone committed to maintain the plugin for several releases. Anyway possible plugin is not part of this PR, so +1 for me.

tbrisker:
If anyone wishes to make a plugin they are more then welcome too - that doesn't mean we can't drop it from core until it exists. TBH I don't think there would be much demand for supporting this api considering how old v2 is already - and anyone still using v1 is most likely also still using an older version of Foreman.

tbrisker:
@xprazak2 As the release manager for 1.17 you should make the call if this should go in before or after branching.

lzap (approved):
I agree let's merge it.

xprazak2:
I have no objections against merging this before 1.17 branches, it does not affect packaging as far as I can see.

tbrisker:
@xprazak2 if we want to try and maintain semver and merge this into 1.17 i think that means 1.17 should be 2.0.

ares:
I think there was no clear agreement on 1.17 being 2.0. I was hoping to have a big warning that api v1 is being dropped in 1.17 release notes and merge this just after branching.

ares:
this needs rebasing now

tbrisker:
@ares thanks, rebased

mmoll (approved):
Let's merge this, now that 1.17 is branched.

timogoebel (approved):

timogoebel:
In case anybody cares to put this into a plugin, let me point you to a starter: https://github.com/unorthodoxgeek/foreman_v1

tbrisker:
@mmoll @timogoebel @lzap @ares - looks like there is consensus to drop it, who wants to do the honors? I added the task of indicating in 1.17 docs that it is the last version supporting apiv1 - https://github.com/theforeman/theforeman.org/issues/974

timogoebel:
Merged, thanks @tbrisker.

Now we have plenty of room for a new graphql api! 👍

PR #5070, opened 2017-12-11
Fixes #21922 - Change memory to Megabytes in compute profile

shiramax:


webpack/assets/javascripts/compute_resource/ovirt.js
```diff
@@ -57,7 +57,7 @@ export function instanceTypeSelected(item) {
       url,
       data: `instance_type_id=${instanceType}`,
       success(result) {
-        $('[id$=_memory]').val(result.memory);
+        $('[id$=_memory]').val(result.memory / 1024 / 1024).trigger('change');
```
tbrisker: instead of triggering `change` we should init all spinners after the ajax completes in `reloadOnAjaxComplete`: 
```javascript
tfm.numFields.initAll()
```

tbrisker (changes requested):
Please also add integration tests for this so we don't break it again (e.g. when we replace the jquery_ui spinners with react fields).

webpack/assets/javascripts/compute_resource/ovirt.js
ohadlevy: doesnt it make more sense to agree on one unit for all compute resources? then we can have a consistent api and drive alignment in rails vs js?

shiramax:
@tbrisker , Done.

boaz0 (approved):
LGTM 👍

ohadlevy:
I had a comment which got lost in the review - my main question is shouldnt we agree / align the size units on the server side? it will ensure api compatibility (e.g. across compute resources)

tbrisker (approved):
@ohadlevy I agree we should normalize these values in the api and handle the needed adjustments on the server side, but this is outside the scope of this PR.
Tested and works well, thanks @shiramax !

PR #5076, opened 2017-12-12
Fixes #21952: opt in to jquery select plugins.

waldenraines:
Instead of assuming all selects should use jquery select plugins
we should instead use an opt in model.

http://projects.theforeman.org/issues/21952

### DIFF
--- a/app/helpers/form_helper.rb
+++ b/app/helpers/form_helper.rb
@@ -67,10 +67,8 @@ module FormHelper
     end
     authorized = AssociationAuthorizer.authorized_associations(associations.reorder(nil), klass).all
 
-    # select2.js breaks the multiselects disabled items location
-    # http://projects.theforeman.org/issues/12028
     html_options["class"] ||= ""
-    html_options["class"] += " without_select2"
+    html_options["class"] += " jquery-multiselect"
     html_options["class"].strip!
 
     unauthorized = selected_ids.blank? ? [] : selected_ids - authorized.map(&:id)
@@ -125,6 +123,13 @@ module FormHelper
     html_options[:disabled] = true if disable_button_enabled
 
     html_options[:size] = 'col-md-10' if html_options[:multiple]
+
+    unless html_options[:disable_select2]
+      html_options["class"] ||= ""
+      html_options["class"] += " jquery-select2"
+      html_options["class"].strip!
+    end
+
     field(f, attr, html_options) do
       addClass html_options, "form-control"
 
@@ -172,6 +177,13 @@ module FormHelper
 
   def selectable_f(f, attr, array, select_options = {}, html_options = {})
     html_options[:size] = 'col-md-10' if html_options[:multiple]
+
+    unless html_options[:disable_select2]
+      html_options["class"] ||= ""
+      html_options["class"] += " jquery-select2"
+      html_options["class"].strip!
+    end
+
     field(f, attr, html_options) do
       addClass html_options, "form-control"
       f.select attr, array, select_options, html_options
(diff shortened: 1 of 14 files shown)


app/helpers/form_helper.rb
```diff
@@ -172,6 +172,13 @@ def time_zone_select_f(f, attr, default_timezone, select_options = {}, html_opti
 
   def selectable_f(f, attr, array, select_options = {}, html_options = {})
     html_options[:size] = 'col-md-10' if html_options[:multiple]
+
+    unless html_options[:disable_select2] do
```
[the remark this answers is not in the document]
waldenraines: Oops, thanks @houndci-bot

waldenraines:
[test foreman]

dLobatog (commented):
@waldenraines Why shouldn't it remain opt-out? I remember the first time we did a demo to Patternfly folks and @rohoover they praised them - from an UX point of view I use them all the time.

waldenraines:
> @waldenraines Why shouldn't it remain opt-out? I remember the first time we did a demo to Patternfly folks and @Rohoover they praised them - from an UX point of view I use them all the time.

It will still be opt-out from the rails perspective (as long as you use a method in `form_helper.rb`) but it is bad practice to assume all `<select>`s should be using this jquery library.  This actively breaks every new select in react.

There is no pattern for this in patternfly either.

waldenraines:
[test katello]

waldenraines:
[test foreman]

waldenraines:
[test]

app/helpers/form_helper.rb
```diff
@@ -67,10 +67,8 @@ def multiple_selects(f, attr, associations, selected_ids, options = {}, html_opt
     end
     authorized = AssociationAuthorizer.authorized_associations(associations.reorder(nil), klass).all
 
-    # select2.js breaks the multiselects disabled items location
-    # http://projects.theforeman.org/issues/12028
     html_options["class"] ||= ""
-    html_options["class"] += " without_select2"
+    html_options["class"] += " jquery-multiselect"
```
ohadlevy, line 71: not sure I understand? without_select2 class avoids using select2, I was under the assumption you want to move from a black list (when someone adds the without_select2) to a whitelist of when we use select2?

ohadlevy:
@waldenraines did you see https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/components/common/forms/Select.js ?

app/helpers/form_helper.rb
waldenraines: Yes, but I wanted to keep the form generating code the same so as to minimize changes in the application.  Basically I just want normal selects that I add with react to not automatically get the select2/multiselect classes.

waldenraines:
> @waldenraines did you see https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/components/common/forms/Select.js ?

But wouldn't that still get both the problematic selects applied to it?

waldenraines:
[test]

waldenraines:
> [test]

I think these tests are likely failing in develop as well because they expect an id of `#s2id_host_hostgroup_id` and the id in the application itself is `#s2d_autogen3` unless there is some other magic that I'm not seeing here.  

@dLobatog @ohadlevy thoughts?

waldenraines:
> @dLobatog @ohadlevy thoughts?

@dLobatog @ohadlevy ping

beav:
this patch fixed an issue I hit when testing another PR

waldenraines:
Closing in favor of https://github.com/theforeman/foreman/pull/5253 for now.

PR #5078, opened 2017-12-13
Refs #21312 - limit locales bundled from intl packages

tstrachota:
Restricts locale files in packages `intl` and `react-intl` only to the subset of languages supported in the Foreman to avoid huge webpack bundles.


tstrachota:
@ares @tbrisker may I ask you for the review?

ohadlevy:
@tstrachota whats the size difference ? also note that currently we only load the active language vs all the languages at the same time, do you think we can  break it down to each language and load that conditionally?

config/webpack.config.js
```diff
@@ -23,6 +23,8 @@ var resolveModules = [  path.join(__dirname, '..', 'webpack'),
                         'node_modules/',
                      ].concat(pluginUtils.pluginNodeModules(plugins));
 
+var supportedLocales = 'ca|de|en|es|fr|gl|it|ja|ko|pl|pt|ru|sv|zh'
```
ohadlevy: is there an easy way to maintain this list (instead of hard code?)
also, any idea why the intl js code is so big? which kind of strings are in it to begin with?

tbrisker:
in development this reduced the bundle size from 68MB to 20MB, didn't test production. This is a big save any case, but I still think it may be good if possible to split these from the main bundle and build a bundle for every language so that the users will only need to download one language instead of 14, similar to what we do for legacy translations.

config/webpack.config.js
tstrachota: We could probably `ls ./locale/` and use names of the directories there.

tbrisker: perhaps get a list of subfolders in `locale/` and join by `|`?

tstrachota:
I updated the webpack config to detect locales from `./locale/`.

I agree with adding per-language chunks. I'll need to invest more time into playing with how to do that though. So we might want to merge this improvement first and work on splitting the bundle with priority but independently.

config/webpack.config.js
```diff
@@ -23,6 +24,11 @@ var resolveModules = [  path.join(__dirname, '..', 'webpack'),
                         'node_modules/',
                      ].concat(pluginUtils.pluginNodeModules(plugins));
 
+// Detect supported locales from folders in ./locale/
+const localeDir = path.join(__dirname, '..', 'locale')
+const localeSubdirs = fs.readdirSync(localeDir).filter(f => fs.statSync(path.join(localeDir, f)).isDirectory());
+const supportedLocales = localeSubdirs.map(d => d.split('_')[0]).join('|')
```
tbrisker: This produces duplicates for languages that appear twice such as `en` or `zh`

tstrachota: Good catch, updated

ares:
I guess we're close so we could postpone reverting deadline for tomorrow. Branching won't happen tonight based on my info.

ares:
So we had a discussion with @tstrachota and here's our conclusions:

1) using intl for date/time format localization is long term a good idea, supported by all modern browsers, intl library that causes the big bundle is only used for legacy browsers and with this patch, we only bundle date strings which adds "only" 0.5 MB to ~1 MB we had before intl was merged, right now, the bundle.js in develop has 18.6 MB for vendor.js, see more info at https://gist.github.com/tstrachota/7beca280b3fa944463df1e434902bec2
2) the internationalization of strings such as "seconds", "months", "ago" is provided by modern browser intl API and it sounds reasonable to rely on that, not trying to mix it into our gettext
3) further internationalization of JS stack is subject of discussion, so far the plan is to continue extracting strings using `__()` but a discussion about other solutions could be held, it's orthogonal to this PR, my personal opinion is that no matter what we chose, the native intl browser support should be used where applicable
4) loading of just the language that is required would be good addition, if you think it's a hard dependency of this, we should revert the original PR as it will take longer time, if you're fine with adding this later, we'd prefer to merge this PR and keep dates components in for 1.17

@ohadlevy @tbrisker please let us know what you think and what you prefer in terms of point 4

ohadlevy (changes requested):
@tstrachota can you please look into extracting the intl package to its own chunk? this will allow us conditionally to load the js file instead of always loading it into memory on all browsers.

tstrachota:
I added a commit with prototype of including intl and its translations dynamically from separate chunks.

But I'm struggling with getting the tests run. I'm getting "SyntaxError: Unexpected token import" :-/
Webpack compilation works fine.

ohadlevy:
import is an es6 syntax that webpack support,I'm guessing that you would
need to find the similar jest plugin to enable it for testing?

tstrachota:
Ehm, ok, it was jest not updating cached `.babelrc`... should be ok now.

ares:
What's the status now?  Is this ready or should we revert the original commit to unblock nightlies?

ohadlevy:
@ares the tests currently fail, not sure how far we are at but it does seem pretty close

tstrachota:
Although it looked promising and webpack built separate chunks for intl, I hit another issue. The intl chunk doesn't get loaded (hence the failing tests).
When I tested it I found out that import actually errors out with:
```
Uncaught (in promise) TypeError: Cannot read property 'call' of undefined
    at __webpack_require__ (bootstrap d2619090cf6a83b9d987:712)
    at fn (bootstrap d2619090cf6a83b9d987:117)
    at <anonymous>
```
I experimented with various dynamic loading styles (`require.ensure()`, `require()`) but with the same result :-/ 
So, no luck with the dynamic imports yet. We can still merge at least the first commit with the contexts that reduce the bundle size.

mmoll:
On my x86_64 machine, the first commit reduced the memsize for the respective nodejs process to about 40%, that should suffice to let the DEB builds on 32 bit archs succeed again.

ares:
@ohadlevy what is your preference? Thanks

ohadlevy:
@ares @tstrachota I don't think it make a lot of sense to include in 1.17, I prefer we revert it and potentially reapply the smaller version to develop after branching 1.17-stable.
I would also think it make sense to consider revisiting moment.js for date / time as its a standalone library that is translated well.

tstrachota:
I'm closing this PR. I'll include the code in the new one that I'm going to open.

PR #5079, opened 2017-12-13
Fixes #21120 - orchestration tasks are added only once

lzap:
Previously, our orchestration task engine was accepting multiple instances of same tasks. Foreman is then orchestrating various actions multiple times and this misbehavior is with us for years. All our backend services were idempotent, until we added PXELoader flag which caused DHCP to error out with conflicts, because filename option treated separately in ISC DHCP.

Without this patch, DHCP orchestration of host edit had four steps:

* delete DHCP entry
* delete DHCP entry
* add DHCP entry
* add DHCP entry

This patch will not add orchestration tasks with the same name twice. If you grep in our codebase for `queue.create` you can see that we don't have any task with same name, so `name` is good field to check for. I considered checking `action` but it contain active record reference which changes over time.

Now, `name` is a free form string and it is also being translated, so it is not ideal. There might be cases when multiple tasks must be scheduled. In that case, new field was introduced called `id` and if that's supplied (symbol or string), it is considered as the identifier for searching for duplicite entries.

With the patch, it's now correct:

* delete DHCP entry
* add DHCP entry

If needed, I can also add `id` field to all core tasks and add a deprecation warning for plugins which does not provide it so plugin authors have time to fix their code until we enforce presence of id.

### DIFF
--- a/app/services/orchestration/queue.rb
+++ b/app/services/orchestration/queue.rb
@@ -4,9 +4,9 @@ module Orchestration
   # Represents tasks queue for orchestration
   class Queue
     attr_reader :items, :name
-    STATUS = %w[ pending running failed completed rollbacked conflict canceled]
+    STATUS = %w[pending running failed completed rollbacked conflict canceled]
 
-    delegate :count, :empty?, :to => :items
+    delegate :count, :size, :empty?, :to => :items
     delegate :to_json, :to => :all
     delegate :to_s, :to => :name
 
@@ -17,8 +17,14 @@ module Orchestration
 
     def create(options)
       options[:status] ||= default_status
-      Rails.logger.debug "Enqueued task '#{options[:name]}' to '#{name}' queue"
-      items << Task.new(options)
+      new_task = Task.new(options)
+      if items.include? new_task
+        # Two tasks with same :name are not allowed. Use two different :id options for multiple instances.
+        Rails.logger.debug "Task '#{new_task.id || new_task.name || ''}' already in '#{name}' queue"
+      else
+        Rails.logger.debug "Enqueued task '#{new_task.id || new_task.name || ''}' to '#{name}' queue"
+        items << new_task
+      end
     end
 
     def delete(item)
@@ -26,11 +32,20 @@ module Orchestration
     end
 
     def find_by_name(name)
-      items.each {|task| return task if task.name == name}
+      items.detect {|task| task.name == name}
+    end
+
+    def find_by_id(id)
+      string_id = id.to_s
+      items.detect {|task| task.id == string_id}
     end
 
     def all
       items.sort
+    end
+
+    def task_ids
+      all.map(&:id)
     end
 
     def clear
--- a/app/services/orchestration/task.rb
+++ b/app/services/orchestration/task.rb
@@ -1,12 +1,15 @@
 module Orchestration
   class Task
-    attr_reader :name, :status, :priority, :action, :timestamp
+    attr_reader :id, :name, :status, :priority, :action, :timestamp, :created
 
     def initialize(opts)
       @name      = opts[:name]
+      @id        = opts[:id].try(:to_s) || @name
       @status    = opts[:status]
       @priority  = opts[:priority] || 0
       @action    = opts[:action]
+      @created   = opts[:created] || Time.now.to_f
+      raise("action must be present for task '#{name}'") if action.nil?
       update_ts
     end
 
@@ -24,11 +27,12 @@ module Orchestration
     end
 
     def to_s
-      "#{name}\t #{priority}\t #{status}\t #{action}"
+      tid = id.nil? ? '' : " (#{id})"
+      "#{name}#{tid}\t #{priority}\t #{status}\t #{action}"
     end
 
     def as_json(options = {})
-      { :name => name, :timestamp => timestamp, :status => status, :priority => priority }
+      { :id => id, :name => name, :timestamp => timestamp, :status => status, :priority => priority, :created => created }
     end
 
     private
@@ -39,7 +43,13 @@ module Orchestration
 
     # sort based on priority
     def <=>(other)
+      return self.created <=> other.created if self.priority == other.priority
       self.priority <=> other.priority
+    end
+
+    def ==(other)
+      return false unless other.is_a?(Task)
+      self.id == other.id
     end
   end
 end
--- a/test/models/orchestration/dhcp_test.rb
+++ b/test/models/orchestration/dhcp_test.rb
@@ -194,12 +194,20 @@ class DhcpOrchestrationTest < ActiveSupport::TestCase
   end
 
   test "new host should create a dhcp reservation" do
-    h = FactoryBot.build(:host, :with_dhcp_orchestration)
+    h = FactoryBot.build(:host, :with_dhcp_orchestration, :mac => "aa:bb:cc:dd:ee:ff")
     assert h.new_record?
-
     assert h.valid?
-    assert_equal h.queue.items.count {|x| x.action.last == :set_dhcp }, 1
-    assert h.queue.items.select {|x| x.action.last == :del_dhcp }.empty?
+    assert_equal ["dhcp_create_aa:bb:cc:dd:ee:ff"], h.queue.task_ids
+    assert_equal :set_dhcp, h.queue.find_by_id("dhcp_create_aa:bb:cc:dd:ee:ff").action.last
+  end
+
+  test "new host with multiple validations should create a single dhcp reservation" do
+    h = FactoryBot.build(:host, :with_dhcp_orchestration, :mac => "aa:bb:cc:dd:ee:ff")
+    assert h.new_record?
+    assert h.valid?
+    assert h.valid?
+    assert_equal ["dhcp_create_aa:bb:cc:dd:ee:ff"], h.queue.task_ids
+    assert_equal :set_dhcp, h.queue.find_by_id("dhcp_create_aa:bb:cc:dd:ee:ff").action.last
   end
 
   test "queue_dhcp doesn't fail when mac address is blank but provided by compute resource" do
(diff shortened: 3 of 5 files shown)


app/services/orchestration/queue.rb
```diff
@@ -17,16 +17,30 @@ def initialize(name = "Unnamed")
...
       @items.delete item
     end
 
     def find_by_name(name)
-      items.each {|task| return task if task.name == name}
+      items.find {|task| task.name == name}
+    end
+
+    def find_by_id(id)
+      items.find {|task| task.id == id}
```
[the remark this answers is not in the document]
lzap: I knew it! Damn.

test/models/orchestration/dhcp_test.rb
```diff
@@ -239,9 +239,9 @@ def teardown
     end
     h.ip = h.ip.succ
     assert h.valid?
-    # 1st is creation from factory, 2nd is triggered by h.valid?
-    assert_equal 2, h.queue.items.count {|x| x.action == [ h.primary_interface, :set_dhcp ] }
-    # and also one deletion (of original creation)
+    # TODO replace this with h.queue.find_by_action - I want these changes to be more explicit for the initial review,
```
ares: I guess you can now replace it :-) makes sense to me

lzap: Sure, anything else before I rebase?

lzap: @ares can you confirm that the changes in this test (2 -> 1) are correct? I want just double check this, I don't see any reason to have these, I think it was simply verified because of the design. I guess "makes sense" is that, want to be sure. Possibly breaking change it is.

lzap: I will implement "IDs" (see initial description) which will make this much easier. Matching via actions is not possible because NIC/Host are not kept in memory for whole time.

app/services/orchestration/queue.rb
```diff
@@ -38,6 +52,10 @@ def clear
       true
     end
 
+    def count
```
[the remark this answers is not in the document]
ares: size alias would also be nice

app/services/orchestration/task.rb
```diff
@@ -24,11 +26,12 @@ def status?(symbol)
     end
 
     def to_s
-      "#{name}\t #{priority}\t #{status}\t #{action}"
+      tid = id.nil? ? '' : " (#{id})"
```
ares, line 30: woud it be better to rather have `(no id)` or something in case there's no id set? hopefully that would indicate, that task author should consider setting id

package.json
```diff
@@ -7,7 +7,7 @@
   "devDependencies": {
     "@storybook/addon-actions": "^3.2.12",
     "@storybook/react": "^3.2.12",
-    "@storybook/storybook-deployer": "^2.0.0",
+    "@storybook/storybook-deployer": "^2.2.0",
```
ares: changes in this file seems unrelated?

app/services/orchestration/task.rb
```diff
@@ -41,5 +44,10 @@ def update_ts
     def <=>(other)
       self.priority <=> other.priority
     end
+
+    def ==(other)
+      return self.id == other.id if self.id && other.id
+      self.name == other.name
```
ares: I didn't dig into details yet but my understanding is that we only compare id or name if id is missing, originally we discussed we should also look at all attributes/parameters, is it no longer necessary?

ares (changes requested):
I left several comments from initial review

should this PR also extend existing orchestration and set id for tasks?

app/services/orchestration/task.rb
lzap: Yeah, I actually plan to do DEPRECATED warning in case of missing id, that should be explicit enough I think.

app/services/orchestration/task.rb
lzap: I can't compare parameters, there is usually whole object passed in (host, nic) that is modified. It's either ID (recommended) or name (deprecated, will be removed in 1.18).

lzap:
Let me know what you think @ares and @ohadlevy - the new version:

* Adds new key called `id` which is now required and used as ident to recognize dupes. When not present, deprecation warning is shown and `name` is used as a fallback, which is fine for most cases (but for NICs this can be problem - I already added `id` there).
* Tasks are now preserved in stable order - what comes first with same priority comes out. On my Ruby the sorting algo seems to be stable, but I cannot guarantee other versions, thus I added flag `created` and made sure tasks are preserved in this order after sort.
* Added some new helper methods to orchestration queue which makes tests cleaner.
* Added new unit test for orchestration queue.
* Rewrote dhcp orchestration test with new approach which uses `id` instead of clunky testing of actions.

The last step (test rerwrite) revealed also that we have wrong order for new hosts which are also being edited. The correct order would be: create new entry, edit the entry (delete, create). But we have it the other way around, because create is priority 10 and edit is 5/6. I think since this patch is touching lots of things and adds the stable sort, I propose to change the create DHCP priority from 10 to 3.

To test the misbehavior I am trying to fix just edit Orchestration::Queue#create in a way that multiple entries are allowed, then run tests and see:

```
Failure:
DhcpOrchestrationTest#test_0017_when an existing host triggers a 'rebuild', its dhcp records should be updated if no dhcp records are found [test/models/orchestration/dhcp_test.rb:287]:
--- expected
+++ actual
@@ -1 +1 @@
-["dhcp_remove_aa:bb:cc:dd:ee:f1", "dhcp_create_aa:bb:cc:dd:ee:f1"]
+["dhcp_remove_aa:bb:cc:dd:ee:f1", "dhcp_create_aa:bb:cc:dd:ee:f1", "dhcp_create_aa:bb:cc:dd:ee:f1"]

Failure:
DhcpOrchestrationTest#test_0014_when an existing host changes its ip address, its dhcp records should be updated [test/models/orchestration/dhcp_test.rb:245]:
--- expected
+++ actual
@@ -1 +1 @@
-["dhcp_remove_aa:bb:cc:dd:ee:ff", "dhcp_create_aa:bb:cc:dd:ee:ff"]
+["dhcp_remove_aa:bb:cc:dd:ee:ff", "dhcp_create_aa:bb:cc:dd:ee:ff", "dhcp_create_aa:bb:cc:dd:ee:ff"]

Failure:
DhcpOrchestrationTest#test_0011_new host with multiple validations should create a single dhcp reservation [test/models/orchestration/dhcp_test.rb:209]:
--- expected
+++ actual
@@ -1 +1 @@
-["dhcp_create_aa:bb:cc:dd:ee:ff"]
+["dhcp_create_aa:bb:cc:dd:ee:ff", "dhcp_create_aa:bb:cc:dd:ee:ff"]

... etc
```

Not all test cases will fail, obviously only those when we call `valid?` multiple times. I created an extra test when I explicitly call this twice to check for this case.

I also created alternative backport-friendly version of this approach that can go to 1.15/1.16 because this is too much of a change I think. See the referenced PR. When testing this, be aware that smart-proxy from develop currently ignores filename option and does not serialize it: http://projects.theforeman.org/issues/21975

app/services/orchestration/task.rb
lzap: New version has `id` required now, deprecation warning raised when it's missing. I want to keep both ID and name here just to be clear about this.

app/services/orchestration/task.rb
lzap: In the next Foreman version, ID will be required. The fallback on name won't be used anymore. We can't easily compare data as it contains ActiveRecord objects.

package.json
lzap: Will remove, thanks.

test/models/orchestration/dhcp_test.rb
ares: I believe this is correct, I believe I bumped those numbers just because `valid?` added new item to queue

app/services/orchestration/task.rb
timogoebel: Can't we generate the id automatically by default, based on the task name, action class name, action method, action object id, priority, ... maybe using a hash?

This seems to be a lot of change effort in plugins just to detect duplicates.
Why doesn't something as simple as `[task1, task2, task3].uniq` work?
Or clearing the queue(s) before new tasks are added to make sure they are only added once?

lzap:
@ares let me know and I will rebase, after branching I'd love to have this merged so we have more testing.

app/services/orchestration/task.rb
ares: I'm not sure how many plugins extend the orchestration but if there are some, we should update them too so they pass id, otherwise it won't happen until we hard enforce it since all plugins would rely on defaulting to task name or action method. Since errors caused by duplicated tasks in queue are hard to debug, I'd prefer 500 unless that means updating all foreman plugins. If we talk about ~5, I think it's worth it

app/services/orchestration/task.rb
```diff
@@ -39,7 +45,14 @@ def update_ts
...
     # sort based on priority
     def <=>(other)
+      return self.created <=> other.created if self.priority == other.priority
       self.priority <=> other.priority
     end
+
+    def ==(other)
+      return false unless other.is_a?(Task)
+      return self.id == other.id if self.id && other.id
+      self.name == other.name
```
ares: if `id` is now required, why we fallback to name?

package.json
ares: still here

app/services/orchestration/task.rb
ares: now I get it after reading your last comment, can't we then use name in initializer as an id if id was not passed as argument? therefore we wouldn't need fallback here (and potentially other places)

ares:
left few comments, I guess you can rebase, also see @timogoebel's questions, but from my point of view, this is going right way

app/services/orchestration/task.rb
lzap: Absolutely good idea that's also faster!

lzap:
Rebased, I know that @ares is preparing FOSDEM demo now, @timogoebel might be away too. I would love to get that merged soon as we branched off 1.17 and this is quite a change. Maybe @dLobatog, @tbrisker or @ShimShtein can review and test this? Thanks a bunch

test/models/orchestration/dhcp_test.rb
```diff
@@ -326,7 +328,7 @@ def teardown
     bmc.mac = next_mac(bmc.mac)
     assert h.valid?
     assert bmc.valid?
-    assert_equal 1, h.queue.items.count {|x| x.action == [ h.primary_interface, :set_dhcp ] }
+    assert_equal 2, h.queue.items.count {|x| x.action == [ h.primary_interface, :set_dhcp ] }
```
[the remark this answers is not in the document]
lzap: Rebased, jeeeeez.

test/models/orchestration/dhcp_test.rb
```diff
@@ -326,9 +326,9 @@ def teardown
...
     assert_equal 1, h.queue.items.count {|x| x.action.last == :del_dhcp }
-    assert_equal 1, bmc.queue.items.count {|x| x.action == [ bmc,     :set_dhcp ] }
+    assert_equal 1, bmc.queue.items.count {|x| x.action == [ bmc, :set_dhcp ] }
```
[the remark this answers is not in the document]
lzap: I fixed this already, local rubocop is okay with the fix. Vanish!

test/models/orchestration/dhcp_test.rb
```diff
@@ -326,10 +326,10 @@ def teardown
...
     assert h.valid?
     assert bmc.valid?
-    assert_equal 1, h.queue.items.count {|x| x.action == [ h.primary_interface, :set_dhcp ] }
-    assert_equal 1, h.queue.items.count {|x| x.action.last == :del_dhcp }
-    assert_equal 1, bmc.queue.items.count {|x| x.action == [ bmc,     :set_dhcp ] }
-    assert_equal 1, bmc.queue.items.count {|x| x.action == [ bmc.old, :del_dhcp ] }
+    assert_equal 2, h.queue.items.count {|x| x.action == [ h.primary_interface, :set_dhcp ]}
+    assert_equal 1, h.queue.items.count {|x| x.action.last == :del_dhcp}
+    assert_equal 1, bmc.queue.items.count {|x| x.action == [ bmc, :set_dhcp ]}
+    assert_equal 1, bmc.queue.items.count {|x| x.action == [ bmc.old, :del_dhcp ]}
```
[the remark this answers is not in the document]
lzap: Are u fing king me?

app/services/orchestration/queue.rb
```diff
@@ -17,22 +17,37 @@ def initialize(name = "Unnamed")
...
+      string_id = id.to_s
+      items.detect {|task| task.id == string_id}
     end
 
     def all
       items.sort
     end
 
+    def task_ids
+      all.map{|x| x.id}
```
ares: is better to be written as `all.map(&:id)`

app/services/orchestration/task.rb
```diff
@@ -1,12 +1,17 @@
...
 
     def initialize(opts)
       @name      = opts[:name]
+      @id        = opts[:id].try(:to_s) || @name
       @status    = opts[:status]
       @priority  = opts[:priority] || 0
       @action    = opts[:action]
+      @created   = opts[:created] || Time.now.to_f
+      raise("action must be present for task '#{name}'") if action.nil?
+      # this kills tests, need to update other tests before
```
ares: what tests you had in mind? and why action is now required? I mean is it refactoring so we fail faster if action is missing?

app/services/orchestration/task.rb
```diff
@@ -1,12 +1,17 @@
...
+      #Foreman::Deprecation.deprecation_warning('1.19', "Orchestration task '#{name}' is missing required :id field") if @id.nil?
```
ares: should this still be commented?

ares (changes requested):
I tested on nightly prod-like setup, orchestration still works great (also with several popular plugins). 3 nits in code and then good to be merged!

app/services/orchestration/task.rb
lzap: I removed those, my initial idea was to fail fast with deprecation warning to enforce having `id` but I think nobody will notice this in test output, it's better I think to have sane default (use name if id is not set) and go with that.

app/services/orchestration/queue.rb
lzap: Fixed, rebased.

lzap:
```diff
diff --git a/app/services/orchestration/queue.rb b/app/services/orchestration/queue.rb
index 0e6575d48..c330b1614 100644
--- a/app/services/orchestration/queue.rb
+++ b/app/services/orchestration/queue.rb
@@ -45,7 +45,7 @@ module Orchestration
     end
 
     def task_ids
-      all.map{|x| x.id}
+      all.map(&:id)
     end
 
     def clear
diff --git a/app/services/orchestration/task.rb b/app/services/orchestration/task.rb
index 32f94cb0b..7fe4b442c 100644
--- a/app/services/orchestration/task.rb
+++ b/app/services/orchestration/task.rb
@@ -10,8 +10,6 @@ module Orchestration
       @action    = opts[:action]
       @created   = opts[:created] || Time.now.to_f
       raise("action must be present for task '#{name}'") if action.nil?
-      # this kills tests, need to update other tests before
-      #Foreman::Deprecation.deprecation_warning('1.19', "Orchestration task '#{name}' is missing required :id field") if @id.nil?
       update_ts
     end
```

app/services/orchestration/task.rb
lzap: The comment "this kills" was meant for line 14, deprecated exceptions actually make tests to fail. Re: action - since all attributes are read only, it makes no sense to create such a task?

ares (approved):

ares:
All is good but jenkins seems to be failing on unrelated problem with phantomjs. I'd rather wait with merging until it's resolved.

ares:
[test foreman]

ares:
[test katello]

ares:
Thanks @lzap, merging!

PR #5080, opened 2017-12-13
Fixes #21962 - switch jquery ajax to axios

amirfefer:
jquery is a enormous library, and there is no point to include it only for ajax calls.
Instead, it is better to use dedicated library for async API calls.
Axios is well known library, and well maintained, with easy mocks implementations for testing, and fits to redux async actions.

- [x] Add axios
- [x] Add mocks implementations for testing
- [x] refactor `ajaxRequestAction` to use axios
- [x]  refactor API  `get` function  to use axios
- [x]  add API `post, put and delete` functions 
- [x]  refactor legacy tests

### DIFF
--- a/webpack/assets/javascripts/react_app/API.js
+++ b/webpack/assets/javascripts/react_app/API.js
@@ -1,37 +1,51 @@
-import $ from 'jquery';
+import axios from 'axios';
+
+const getcsrfToken = () => {
+  const token = document.querySelector('meta[name="csrf-token"]');
+
+  return token ? token.content : '';
+};
+
+axios.defaults.headers.common['X-Requested-With'] = 'XMLHttpRequest';
+axios.defaults.headers.common['X-CSRF-Token'] = getcsrfToken();
 
 export default {
-  get(url) {
-    $.ajaxPrefilter((options, originalOptions, jqXHR) => {
-      // eslint-disable-next-line no-param-reassign
-      jqXHR.originalRequestOptions = originalOptions;
-    });
-    return $.getJSON(url);
-  },
-  markNotificationAsRead(id) {
-    const data = JSON.stringify({ seen: true });
-
-    $.ajax({
-      url: `/notification_recipients/${id}`,
-      contentType: 'application/json',
-      type: 'put',
-      dataType: 'json',
-      data,
-      error(jqXHR, textStatus, errorThrown) {
-        /* eslint-disable no-console */
-        console.log(jqXHR);
-      },
+  get(url, headers = {}, params = {}) {
+    return axios.get(url, {
+      headers,
+      params,
     });
   },
-  markGroupNotificationAsRead(group) {
-    $.ajax({
-      url: `/notification_recipients/group/${group}`,
-      contentType: 'application/json',
-      type: 'PUT',
-      error(jqXHR, textStatus, errorThrown) {
-        /* eslint-disable no-console */
-        console.log(jqXHR);
+  put(url, data = {}, headers = {}) {
+    return axios.put(
+      url, data,
+      {
+        headers,
       },
-    });
+    );
+  },
+  post(url, data = {}, headers = {}) {
+    return axios.post(
+      url, data,
+      {
+        headers,
+      },
+    );
+  },
+  delete(url, headers = {}) {
+    return axios.delete(
+      url,
+      {
+        headers,
+      },
+    );
+  },
+  patch(url, data = {}, headers = {}) {
+    return axios.patch(
+      url, data,
+      {
+        headers,
+      },
+    );
   },
 };
--- a/webpack/assets/javascripts/react_app/components/notifications/notifications.fixtures.js
+++ b/webpack/assets/javascripts/react_app/components/notifications/notifications.fixtures.js
@@ -83,11 +83,11 @@ export const stateWithUnreadNotifications = immutable({
   },
 });
 
-export const serverResponse = `{ "notifications":[
+export const serverResponse = `{"data": { "notifications":[
   {"id":1,"seen":true,"level":"info","text":null,"created_at":"2017-02-23T05:00:28.715Z",
   "group":"React devs","actions":{}},
   {"id":2,"seen":false,"level":"info","text":null,"created_at":"2017-02-23T05:00:28.715Z",
-  "group":"React devs","actions":{}}]}`;
+  "group":"React devs","actions":{}}]}}`;
 
 export const emptyHtml =
   '<div id="notifications_container">' +
--- a/webpack/assets/javascripts/react_app/mockRequests.js
+++ b/webpack/assets/javascripts/react_app/mockRequests.js
@@ -1,0 +1,20 @@
+import axios from 'axios';
+import MockAdapter from 'axios-mock-adapter';
+
+const mock = new MockAdapter(axios);
+const methods = {
+  GET: 'onGet',
+  POST: 'onPost',
+  PUT: 'onPut',
+  DELETE: 'onDelete',
+};
+
+export const mockRequest = ({
+  method = 'GET',
+  url,
+  data = null,
+  status = 200,
+  response = null,
+}) => mock[methods[method]](url, data).reply(status, response);
+
+export const mockReset = () => mock.reset();
--- a/webpack/assets/javascripts/react_app/redux/actions/bookmarks/bookmarks.fixtures.js
+++ b/webpack/assets/javascripts/react_app/redux/actions/bookmarks/bookmarks.fixtures.js
@@ -9,7 +9,12 @@ export const initialState = immutable({
   },
 });
 
-export const requestData = { url: '/api/bookmarks', controller: 'hosts' };
+export const requestData = {
+  url: '/api/bookmarks',
+  searchRegex: /\/api\/bookmarks\?search=.{13}hosts/,
+  controller: 'hosts',
+  response: { results: bookmarks },
+};
 
 const requestAction = {
   type: types.BOOKMARKS_REQUEST,
@@ -19,7 +24,10 @@ const requestAction = {
 export const onFailureActions = [
   requestAction,
   {
-    payload: { error: {}, item: { controller: 'hosts' } },
+    payload: {
+      error: new Error('Request failed with status code 422'),
+      item: { controller: 'hosts' },
+    },
     type: types.BOOKMARKS_FAILURE,
   },
 ];
--- a/webpack/assets/javascripts/react_app/redux/actions/bookmarks/bookmarks.test.js
+++ b/webpack/assets/javascripts/react_app/redux/actions/bookmarks/bookmarks.test.js
@@ -1,6 +1,5 @@
 import configureMockStore from 'redux-mock-store';
 import thunk from 'redux-thunk';
-import $ from 'jquery';
 import * as actions from './index';
 import * as types from '../../consts';
 import {
@@ -9,36 +8,42 @@ import {
   onFailureActions,
   onSuccessActions,
 } from './bookmarks.fixtures';
-import { bookmarks } from '../../reducers/bookmarks/bookmarks.fixtures';
 import API from '../../../API';
+import { mockRequest, mockReset } from '../../../mockRequests';
 
 const middlewares = [thunk];
 const mockStore = configureMockStore(middlewares);
+const store = mockStore(initialState);
+
+afterEach(() => {
+  store.clearActions();
+  mockReset();
+});
 
 describe('bookmark actions', () => {
   it('should handle failure to load bookmarks', () => {
-    const store = mockStore(initialState);
-    const { url, controller } = requestData;
+    const { url, controller, searchRegex } = requestData;
 
-    store.dispatch(actions.getBookmarks(url, controller));
-    expect(store.getActions()).toEqual(onFailureActions);
-  });
-  xit('should load bookmarks', () => {
-    const store = mockStore(initialState);
-    const { url, controller } = requestData;
-
-    $.ajax = jest.fn(() => {
-      const ajaxMock = $.Deferred();
-
-      ajaxMock.resolve({ results: bookmarks });
-      return ajaxMock.promise();
+    mockRequest({
+      searchRegex,
+      status: 422,
     });
+    return store.dispatch(actions.getBookmarks(url, controller))
+      .then(() => expect(store.getActions()).toEqual(onFailureActions));
+  });
+  it('should load bookmarks', () => {
+    const {
+      url, controller, response, searchRegex,
+    } = requestData;
 
-    store.dispatch(actions.getBookmarks(url, controller));
-    expect(store.getActions()).toEqual(onSuccessActions);
+    mockRequest({
+      searchRegex,
+      response,
+    });
+    return store.dispatch(actions.getBookmarks(url, controller))
+      .then(() => expect(store.getActions()).toEqual(onSuccessActions));
   });
   it('should load bookmarks with correct search url', () => {
-    const store = mockStore(initialState);
     const { url, controller } = requestData;
     const spy = jest.spyOn(API, 'get');
     const expectedURL = '/api/bookmarks?search=controller%3Dhosts&per_page=100';
--- a/webpack/assets/javascripts/react_app/redux/actions/common/forms.js
+++ b/webpack/assets/javascripts/react_app/redux/actions/common/forms.js
@@ -1,7 +1,6 @@
 // eslint-disable-next-line import/no-extraneous-dependencies
-import fetch from 'isomorphic-fetch';
 import { SubmissionError } from 'redux-form';
-
+import API from '../../../API';
 import { addToast } from '../toasts';
 
 const fieldErrors = ({ error }) => {
@@ -14,29 +13,14 @@ const fieldErrors = ({ error }) => {
   return new SubmissionError(errors);
 };
 
-const checkErrors = (response) => {
-  if (response.ok) {
-    return response;
-  }
-  if (response.status === 422) {
+const onError = (error) => {
+  if (error.response.status === 422) {
     // Handle invalid form data
-    return response.json().then((body) => {
-      throw fieldErrors(body);
-    });
+    throw fieldErrors(error.response.data);
   }
   throw new SubmissionError({
-    _error: [`${__('Error submitting data:')} ${response.status} ${__(response.statusText)}`],
+    _error: [`${__('Error submitting data:')} ${error.response.status} ${__(error.response.statusText)}`],
   });
-};
-
-const getcsrfToken = () => {
-  const token = document.querySelector('meta[name="csrf-token"]');
-
-  if (token) {
-    return token.content;
-  }
-  // fail gracefully when no token is found
-  return '';
 };
 
 const verifyProps = (item, values) => {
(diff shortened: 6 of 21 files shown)


webpack/assets/javascripts/react_app/API.js
```diff
@@ -1,6 +1,18 @@
 import $ from 'jquery';
```
ohadlevy: probably should drop this line (not using jquery?)

webpack/assets/javascripts/react_app/API.js
```diff
@@ -1,6 +1,18 @@
...
+import axios from 'axios';
+
+const onSuccess = response => response.data;
+const onError = error => Promise.reject(error);
 
 export default {
+  ajax(url) {
```
ohadlevy: why expose an "ajax" endpoint? (instead of replacing get, post put etc?)

webpack/assets/javascripts/react_app/API.js
```diff
@@ -1,6 +1,18 @@
 import $ from 'jquery';
+import axios from 'axios';
+
+const onSuccess = response => response.data;
+const onError = error => Promise.reject(error);
```
ohadlevy: whats the rational for us to recreate all exceptions? e.g. I would expect a library to have predefined good exceptions for regular http failure codes vs we have to handle them ourselves?

webpack/assets/javascripts/react_app/API.js
amirfefer: ajax is a temporarily endpoint, i'll merge it with get soon

webpack/assets/javascripts/react_app/API.js
```diff
@@ -1,37 +1,55 @@
...
+const getcsrfToken = () => {
+  const token = document.querySelector('meta[name="csrf-token"]');
+
+  if (token) {
+    return token.content;
+  }
+  // fail gracefully when no token is found
+  return '';
+};
+const onSuccess = response => response.data;
```
sharvit: Doesn't it mean we are losing all other response fields?
Why consumers can't just use 
```js
get(url).then(({data}) => ... );
```

webpack/assets/javascripts/react_app/API.js
```diff
@@ -1,37 +1,55 @@
...
+const onError = (error) => {
+  /* eslint-disable no-console */
+  console.log(error);
```
sharvit: Do we really want console.log here?

webpack/assets/javascripts/react_app/API.js
```diff
@@ -1,37 +1,55 @@
...
+        'X-Requested-With': 'XMLHttpRequest',
+        'X-CSRF-Token': getcsrfToken(),
+      },
       },
-    });
+    )
+      .then(onSuccess)
+      .catch(onError);
   },
 };
```
sharvit: I think we will also need a `delete` request.

sharvit (changes requested):

webpack/assets/javascripts/react_app/API.js
amirfefer: No it doesn't, data is the whole respond's json body, the only fields we're losing here are headers and axios config, which I don't see any use of them right now.

webpack/assets/javascripts/react_app/API.js
amirfefer: oh snap...  no we don't :) I forgot to delete it

webpack/assets/javascripts/react_app/API.js
amirfefer: sure we do, though right now there's no  use for it

webpack/assets/javascripts/react_app/API.js
sharvit: Some API's deciding to return data in headers, mostly when it comes to access tokens.

webpack/assets/javascripts/react_app/API.js
```diff
@@ -1,37 +1,40 @@
...
-        /* eslint-disable no-console */
-        console.log(jqXHR);
+  post(url, data = {}, headers = {}) {
+    return axios.post(
+      url, data,
+      {
+        headers: Object.assign(defaultHeaders, headers),
       },
-    });
+    );
```
ohadlevy: worth adding delete / patch for completion?

webpack/assets/javascripts/react_app/components/notifications/notifications.fixtures.js
```diff
@@ -83,11 +83,11 @@ export const stateWithUnreadNotifications = immutable({
   },
 });
 
-export const serverResponse = `{ "notifications":[
+export const serverResponse = `{"data": { "notifications":[
```
ohadlevy, line 86: why? I don't assume the server response changed?

webpack/assets/javascripts/react_app/mockRequests.js
```diff
@@ -0,0 +1,28 @@
...
+}) => {
+  mockSuccessRequests.forEach((request) => {
+    mock.onGet(request.searchRegex).reply(200, request.response);
+  });
+  mockFailureRequests.forEach((request) => {
+    mock.onGet(request.searchRegex).reply(request.statusCode);
+  });
+};
+
+export const mockPostRequests = ({
```
ohadlevy: we should also have a put for notification drawer mark as read updates?

webpack/assets/javascripts/react_app/redux/actions/common/forms.test.js
```diff
@@ -1,8 +1,7 @@
-import nock from 'nock';
```
ohadlevy, line 1: should probably remove it from package.json too

webpack/assets/javascripts/react_app/redux/actions/notifications/index.js
```diff
@@ -25,27 +25,26 @@ const getNotifications = url => (dispatch) => {
 
   if (isDocumentVisible) {
     API.get(url)
-      .done(onGetNotificationsSuccess)
-      .fail(onGetNotificationsFailed)
-      .always(triggerPolling);
-  } else {
-    // document is not visible, keep polling without api call
-    triggerPolling();
```
ohadlevy: you skipped all of this functionality on purpose ? (skipping if document is not visible and setting up polling?)

webpack/assets/javascripts/react_app/redux/actions/statistics/statistics.fixtures.js
```diff
@@ -1,3 +1,30 @@
+import { mockGetRequests } from '../../../mockRequests';
+
+const mockSuccessRequests = [
+  {
+    path: '/hosts/:id/power',
+    searchRegex: '/hostss/1/power',
```
ohadlevy: typo

ohadlevy (changes requested):

webpack/assets/javascripts/react_app/components/notifications/notifications.fixtures.js
amirfefer: axios returns a new object, the body response encapsulates in data. 
I've extracted it before, as you can see in outdated changes,  though @sharvit  thinks it better to keep the whole axios response object for later use.

amirfefer:
@ohadlevy Thanks for your review
could you have a second look?

amirfefer:
@ohadlevy I've rebased it with the bookmarks changes, could you have another look?

amirfefer:
@danseethaler Would you mind taking a look?

webpack/assets/javascripts/react_app/API.js
```diff
@@ -1,37 +1,51 @@
...
+    return axios.post(
+      url, data,
+      {
+        headers,
       },
-    });
+    );
+  },
+  delete(url, headers = {}) {
+    return axios.post(
```
ohadlevy: wrong action? (here and patch too)

webpack/assets/javascripts/react_app/redux/actions/bookmarks/bookmarks.fixtures.js
```diff
@@ -10,16 +28,20 @@ export const initialState = immutable({
...
 const requestAction = {
   type: types.BOOKMARKS_REQUEST,
   payload: { controller: 'hosts' },
 };
 
 export const onFailureActions = [
-  requestAction,
   {
-    payload: { error: {}, item: { controller: 'hosts' } },
+    type: types.BOOKMARKS_REQUEST,
```
ohadlevy: Why not use the fixtures like other tests?

webpack/assets/javascripts/react_app/redux/actions/common/forms.js
```diff
@@ -49,36 +33,21 @@ const verifyProps = (item, values) => {
...
-          .then(() =>
-            dispatch(addToast({
-              type: 'success',
-              // eslint-disable-next-line no-undef
-              message: Jed.sprintf('%s was successfully created.', __(item)),
-            }))))
-      .catch((error) => {
-        throw error;
-      });
+    API.post(url, values)
```
ohadlevy: the reason i kept the method in the function header, is that we would like to use the same code path for updating an existing form too.

webpack/assets/javascripts/react_app/redux/actions/common/forms.test.js
```diff
@@ -33,14 +29,10 @@ describe('form actions', () => {
...
   it('on failed response raise exception', () => {
-    nock('http://localhost')
-      .post('/api/resource')
-      .reply(500, { message: 'oh snap' });
-
     const store = mockStore({ resources: [] });
 
     return store
-      .dispatch(submitForm({ values: { a: 1 }, url: 'http://localhost/api/resource', item: 'Resource' }))
+      .dispatch(submitForm({ values: { a: 1 }, url: '/api/500', item: 'Resource' }))
```
ohadlevy: the url was /api/resource and the return was 500, reading this feels like its the wrong url?

webpack/assets/javascripts/react_app/redux/actions/common/forms.test.js
```diff
@@ -49,31 +41,23 @@ describe('form actions', () => {
...
     return store
-      .dispatch(submitForm({ values: { a: 1 }, url: 'http://localhost/api/resource', item: 'Resource' }))
+      .dispatch(submitForm({ values: { a: 1 }, url: '/api/404', item: 'Resource' }))
       .then(() => {
         throw new Error('Should not hit this then block - test was set up incorrectly');
       })
       .catch((error) => {
         expect(error).toBeInstanceOf(SubmissionError);
-        expect(error.errors).toEqual({ _error: ['Error submitting data: 404 Not Found'] });
+        expect(error.errors._error[0]).toMatch(/Error submitting data: 404*/);
```
ohadlevy, line 78: can we get the actual error code + text?

webpack/assets/javascripts/react_app/redux/actions/notifications/index.js
```diff
@@ -13,7 +13,7 @@ import {
 } from '../../../common/sessionStorage';
 import API from '../../../API';
 
-const defaultNotificationsPollingInterval = 10000;
+const defaultNotificationsPollingInterval = 500000;
```
ohadlevy: wrong entry?
for future reference, in your development env just use something like
```
NOTIFICATIONS_POLLING=1000000 foreman start
```

webpack/assets/javascripts/react_app/redux/actions/notifications/index.js
```diff
@@ -25,27 +25,29 @@ const getNotifications = url => (dispatch) => {
...
-  function onGetNotificationsSuccess(response) {
+  function onGetNotificationsSuccess({ data }) {
     dispatch({
       type: NOTIFICATIONS_GET_NOTIFICATIONS,
       payload: {
-        notifications: response.notifications,
+        notifications: data.notifications,
       },
     });
+    triggerPolling();
```
ohadlevy: why did you add it here too? there is no always/done kind of call back ?

webpack/assets/javascripts/react_app/redux/actions/statistics/statistics.test.js
```diff
@@ -2,20 +2,26 @@ import configureMockStore from 'redux-mock-store';
 import thunk from 'redux-thunk';
 import immutable from 'seamless-immutable';
 
-import { requestData, onFailureActions } from './statistics.fixtures';
+import { failedRequestData, successRequestData, onFailureActions, onSuccessActions } from './statistics.fixtures';
 
 import * as actions from './index';
 
 const mockStore = configureMockStore([thunk]);
```
ohadlevy, line 9: do you need two stores? (mockStore and Store)

sharvit (approved):
LGTM

danseethaler:
@amirfefer happy to take a look but will be a bit tied up until the 26th 👍

webpack/assets/javascripts/react_app/redux/actions/statistics/statistics.test.js
amirfefer: `mockStore` is for store's configuration with a middleware, where `store` is for initialization.
(implementation taken from [documentation](http://arnaudbenard.com/redux-mock-store/))

webpack/assets/javascripts/react_app/redux/actions/notifications/index.js
amirfefer: there is no a special call back for always / done, though we could use `then` after catch.

webpack/assets/javascripts/react_app/redux/actions/bookmarks/bookmarks.fixtures.js
amirfefer: shouldn't we use const types?

ohadlevy: you should, but if you look at the diff, than it was using it from the requestAction fixture.

webpack/assets/javascripts/react_app/redux/actions/common/forms.test.js
amirfefer: it looks like  `axios mock adapter`does not return `statusText` property (only status code).

amirfefer:
@ohadlevy thanks again,  would you mind to take another look?

mmoll:
[test foreman]

danseethaler (approved):
This looks really good @amirfefer! Thanks for your work 👍

webpack/assets/javascripts/react_app/redux/actions/bookmarks/bookmarks.fixtures.js
```diff
@@ -10,16 +28,17 @@ export const initialState = immutable({
 });
 
 export const requestData = { url: '/api/bookmarks', controller: 'hosts' };
+export const failRequestData = { url: '/api/bookmarks', controller: 'error' };
```
ohadlevy: controller should be a real controller? not sure i understand this fixture?

amirfefer:
@ohadlevy I've changed bookmarks fixtures ambiguous error.
I tend to keep the global mock infrastructure

waldenraines:
@ohadlevy any other requests here or can we merge once the tests are green?

waldenraines:
[test katello]

waldenraines:
@amirfefer can you fix codeclimate please?

amirfefer:
@sharvit , @ohadlevy : i've changed the mock requests from global mocking to per test mocking.
could you have another look?

sharvit (changes requested):
@amirfefer when i ran the tests, all of them are passing but there is one test that receive error (and still passing):
https://gist.github.com/sharvit/559dbeec83239a9fc9bd77e39dc2c579

amirfefer:
@sharvit Thanks. I've fixed that test failure

sharvit (approved):
Thanks @amirfefer! Everything is working and looking great!

amirfefer:
@ohadlevy  Could we merge this?

webpack/assets/javascripts/react_app/redux/actions/bookmarks/bookmarks.test.js
```diff
@@ -9,36 +8,42 @@ import {
...
+    const { url, controller, searchRegex } = requestData;
 
-    store.dispatch(actions.getBookmarks(url, controller));
-    expect(store.getActions()).toEqual(onFailureActions);
+    mockRequest({
+      searchRegex,
+      status: 422,
+    });
+    return store.dispatch(actions.getBookmarks(url, controller))
+      .then(() => expect(store.getActions()).toEqual(onFailureActions));
```
priley86, line 32: maybe there is a better way to avoid promises? I like the approach here overall (and it seems that axios is preferrable to fetch) but I see this bit as a potential limitation to writing more fully scoped tests...will follow up in an email shortly to you guys.

webpack/assets/javascripts/react_app/redux/actions/notifications/index.js
```diff
@@ -19,25 +19,25 @@ const getNotifications = url => (dispatch) => {
 
   if (isDocumentVisible) {
     API.get(url)
-      .done(onGetNotificationsSuccess)
-      .fail(onGetNotificationsFailed)
-      .always(triggerPolling);
+      .then(onGetNotificationsSuccess)
+      .catch(onGetNotificationsFailed)
```
priley86, line 23: same thing here based on comment above... lot of promises...

webpack/assets/javascripts/react_app/redux/actions/bookmarks/bookmarks.test.js
sharvit: @priley86 +1
Checkout this gist:
https://gist.github.com/sharvit/cf658a8494758be8062a63121b925fdc

This is kind of a solution for unit-testing i tried to implement.
We have been discussing about it quit a lot at the last month.
We feel we should migrate the tests soon to this style and start writing integration tests.

amirfefer: This approach is suggested by redux ( Async Action Creators part):
https://redux.js.org/docs/recipes/WritingTests.html
I agree that we should have some test conventions, especially when it involves integration between react component and redux, I've started to work on that kind of integration helper.

webpack/assets/javascripts/react_app/redux/actions/notifications/index.js
amirfefer: What is wrong with promises in API calls? is there a better way?
The second promise here is like `.always` in jquery: 
https://github.com/axios/axios/issues/792

webpack/assets/javascripts/react_app/redux/actions/bookmarks/bookmarks.test.js
priley86: Again we have same dilemma....

```
export const submitForm = ({
  item, url, values, method = 'post',
}) => {
  verifyProps(item, values);
  return dispatch =>
    API[method](url, values)
      .then(({ data }) => {
        dispatch({
          type: `${item.toUpperCase()}_FORM_SUBMITTED`,
          payload: { item, data },
        });
```
... I am chaining promises/chaining actions.... do you not prefer it to be `gen.next()` ?

priley86: I <3 the use of Jest snapshots though.. maybe we can incorporate it with Saga too... just a thought.

webpack/assets/javascripts/react_app/redux/actions/notifications/index.js
priley86: I think the syntax here is much cleaner, personally, but will leave it to you to decide it ;)

https://redux-saga.js.org/docs/advanced/Testing.html

webpack/assets/javascripts/react_app/redux/actions/bookmarks/bookmarks.test.js
amirfefer: +1 on jest snapshots,
I think we should have a deeper look on saga, particularly if its benefit worth the migration from thunk.  I don't sure if we have complicated API requests (which require chaining promises or so) in foreman, testing would be easier though.

amirfefer: leaving saga aside, any thoughts on redux-promise middlewere?
It could answer your concern with promises and yet simpler to migrate.

ohadlevy: where does this leave this PR? I'm not sure i fully understand its status?

ohadlevy (approved):
I believe this is a good step forward, we still have an unresolved discussion around how to test async actions, but imho its outside the scope of this PR. thanks @amirfefer

webpack/assets/javascripts/react_app/redux/actions/bookmarks/bookmarks.test.js
amirfefer: I don't think it relevant to this PR, @priley86  has mentioned good points that we should look into.

PR #5081, opened 2017-12-13
Fixes #21926 - Ensure uniqueness on image names and uuids

tbrisker:
Image names should be unique per CR and OS. UUID uniqueness is already
enforced by ActiveRecord but not in the DB, which could cause conflicts
in multi-proccess environments. Also cleaned up a bit of the image tests
and made them all use FactoryBot instead of fixtures.

### DIFF
--- a/db/migrate/20171213161035_add_indexes_on_images.rb
+++ b/db/migrate/20171213161035_add_indexes_on_images.rb
@@ -1,0 +1,25 @@
+class AddIndexesOnImages < ActiveRecord::Migration[5.0]
+  def up
+    deduped = Image.group(:name, :compute_resource_id, :operatingsystem_id).pluck('min(id)')
+    Image.where.not(id: deduped).each_with_index do |image, i|
+      say "Found images with duplicate name #{image.name} on Compute Resource #{ComputeResource.unscoped.find(image.compute_resource_id)}!"
+      say "Duplicates have been renamed, please check your setup to resolve the duplication."
+      image.update_column(:name, "#{image.name}(Duplicate #{i})")
+    end
+
+    deduped = Image.group(:uuid, :compute_resource_id).pluck('min(id)')
+    Image.where.not(id: deduped).each_with_index do |image, i|
+      say "Found images with duplicate uuid #{image.uuid} on Compute Resource #{ComputeResource.unscoped.find(image.compute_resource_id)}!"
+      say "Duplicates' uuids have been modified, please check your setup to resolve the duplication."
+      image.update_column(:uuid, "#{image.uuid}(Duplicate #{i})")
+    end
+
+    add_index :images, [:name, :compute_resource_id, :operatingsystem_id], name: 'image_name_index', unique: true
+    add_index :images, [:uuid, :compute_resource_id], name: 'image_uuid_index', unique: true
+  end
+
+  def down
+    remove_index :images, name: 'image_name_index'
+    remove_index :images, name: 'image_uuid_index'
+  end
+end
--- a/test/factories/compute_resources.rb
+++ b/test/factories/compute_resources.rb
@@ -65,9 +65,9 @@ FactoryBot.define do
 
   factory :image do
     sequence(:name) { |n| "image#{n}" }
-    uuid Foreman.uuid
+    uuid { Foreman.uuid }
     username 'root'
-    compute_resource
+    association :compute_resource, factory: :libvirt_cr
     operatingsystem
     architecture
   end
--- a/app/models/image.rb
+++ b/app/models/image.rb
@@ -12,7 +12,8 @@ class Image < ApplicationRecord
   before_validation :set_password_changed, :if => Proc.new { |audit| audit.password_changed? }
 
   validates_lengths_from_database
-  validates :username, :name, :operatingsystem_id, :compute_resource_id, :architecture_id, :presence => true
+  validates :username, :operatingsystem_id, :compute_resource_id, :architecture_id, :presence => true
+  validates :name, :presence => true, :uniqueness => {:scope => [:compute_resource_id, :operatingsystem_id]}
   validates :uuid, :presence => true, :uniqueness => {:scope => :compute_resource_id}
   validate :uuid_exists?
 
--- a/test/models/image_test.rb
+++ b/test/models/image_test.rb
@@ -4,20 +4,37 @@ class ImageTest < ActiveSupport::TestCase
   should have_many(:hosts).dependent(:nullify)
 
   test "image is invalid if uuid invalid" do
-    resource = compute_resources(:one)
-    image = resource.images.build(:name => "foo", :uuid => "bar")
+    image = FactoryBot.build(:image, :uuid => "bar")
     ComputeResource.any_instance.stubs(:image_exists?).returns(false)
     image.valid? #trigger validations
     assert image.errors.messages.keys.include?(:uuid)
   end
 
+  test "image name is unique per resource and os" do
+    image1 = FactoryBot.create(:image)
+    image2 = FactoryBot.build(:image, name: image1.name)
+    assert image2.valid?
+    image2.compute_resource = image1.compute_resource
+    assert image2.valid?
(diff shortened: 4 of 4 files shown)


db/migrate/20171213161035_add_indexes_on_images.rb
```diff
@@ -0,0 +1,23 @@
...
+    Image.where.not(id: deduped).each_with_index do |image, i|
+      say "Found images with duplicate name #{image.name} on Compute Resource #{image.compute_resource}!"
+      say "Duplicates have been renamed, please check your setup to resolve the duplication."
+      image.update_column(:name, "#{image.name}(Duplicate #{i})")
+    end
+    deduped = Image.group(:uuid, :compute_resource_id).pluck('min(id)')
+    Image.where.not(id: deduped).each_with_index do |image, i|
+      say "Found images with duplicate uuid #{image.uuid} on Compute Resource #{image.compute_resource}!"
+      say "Duplicates' uuids have been modified, please check your setup to resolve the duplication."
+      image.update_column(:uuid, "#{image.uuid}(Duplicate #{i})")
```
ekohl: Is the column wide enough for this? Also wondering if on postgres the native uuid column type is used which would probably not like this.

tbrisker: yes, it's a string column currently with limit of 255 so should be no problem.

test/factories/compute_resources.rb
```diff
@@ -65,9 +65,9 @@
 
   factory :image do
     sequence(:name) { |n| "image#{n}" }
-    uuid Foreman.uuid
+    sequence(:uuid) { Foreman.uuid }
```
timogoebel: `uuid { Foreman.uuid }` should also work. This isn't really a sequence, but a [dynamic attribute](http://www.rubydoc.info/gems/factory_bot/file/GETTING_STARTED.md#Dynamic_Attributes).

tbrisker: Thanks, didn't know that. updated.

db/migrate/20171213161035_add_indexes_on_images.rb
```diff
@@ -0,0 +1,23 @@
+class AddIndexesOnImages < ActiveRecord::Migration[5.0]
+  def up
+    deduped = Image.group(:name, :compute_resource_id, :operatingsystem_id).pluck('min(id)')
+    Image.where.not(id: deduped).each_with_index do |image, i|
+      say "Found images with duplicate name #{image.name} on Compute Resource #{image.compute_resource}!"
```
xprazak2: Output does not show the compute resource for me:

db/migrate/20171213161035_add_indexes_on_images.rb
```diff
@@ -0,0 +1,23 @@
...
+      say "Duplicates' uuids have been modified, please check your setup to resolve the duplication."
+      image.update_column(:uuid, "#{image.uuid}(Duplicate #{i})")
+    end
+    add_index :images, [:name, :compute_resource_id, :operatingsystem_id], name: 'image_name_index', unique: true
+    add_index :images, [:uuid, :compute_resource_id], name: 'image_uuid_index', unique: true
+  end
+
+  def down
+    remove_index 'image_name_index'
+    remove_index 'image_uuid_index'
```
xprazak2: I think these should be 
```ruby
remove_index, :images, :name => 'image_name_index'
remove_index, :images, :name => 'image_uuid_index'
```

tbrisker: oops, right.

db/migrate/20171213161035_add_indexes_on_images.rb
tbrisker: i'll check this

tbrisker:
@xprazak2 thanks, pushed with fix to your comments

db/migrate/20171213161035_add_indexes_on_images.rb
```diff
@@ -0,0 +1,23 @@
+class AddIndexesOnImages < ActiveRecord::Migration[5.0]
+  def up
+    deduped = Image.group(:name, :compute_resource_id, :operatingsystem_id).pluck('min(id)')
+    Image.where.not(id: deduped).each_with_index do |image, i|
+      say "Found images with duplicate name #{image.name} on Compute Resource #{image.compute_resource.name}!"
```
xprazak2: You may need to wrap these in `User.as :admin` block (or do something else to make sure user is present) because compute resources are taxable and no current user is set, so `image.compute_resource #=> nil`

Same for the lines below.

tbrisker: this is a migration, so it may be run before any users are seeded. Changed method of looking it up, lets see if it works now.

ares:
[test foreman]

tbrisker:
@ares @xprazak2 all 🍏

db/migrate/20171213161035_add_indexes_on_images.rb
```diff
@@ -0,0 +1,25 @@
+class AddIndexesOnImages < ActiveRecord::Migration[5.0]
+  def up
+    deduped = Image.group(:name, :compute_resource_id, :operatingsystem_id).pluck('min(id)')
+    Image.where.not(id: deduped).each_with_index do |image, i|
+      say "Found images with duplicate name #{image.name} on Compute Resource #{ComputeResource.unscoped.find(image.compute_resource_id)}!"
+      say "Duplicates have been renamed, please check your setup to resolve the duplication."
```
dLobatog, line 6: @tbrisker I think it would be more friendly to tell the user where to check for this - e.g: Please check "https://foremanfqdn/compute_resources/id-nameofcomputeresource/images" to rename the duplicate images - as far as I know we have all of the information to show the link here don't we?

db/migrate/20171213161035_add_indexes_on_images.rb
```diff
@@ -0,0 +1,25 @@
...
+      image.update_column(:name, "#{image.name}(Duplicate #{i})")
+    end
+
+    deduped = Image.group(:uuid, :compute_resource_id).pluck('min(id)')
+    Image.where.not(id: deduped).each_with_index do |image, i|
+      say "Found images with duplicate uuid #{image.uuid} on Compute Resource #{ComputeResource.unscoped.find(image.compute_resource_id)}!"
+      say "Duplicates' uuids have been modified, please check your setup to resolve the duplication."
```
dLobatog, line 13: ditto

dLobatog (dismissed):
@tbrisker I think 👍 too, just left one comment to improve the message that the user receives when a duplication is found.

db/migrate/20171213161035_add_indexes_on_images.rb
tbrisker: @dLobatog Thanks, I tried to add this now but since this runs in a migration it seems that we don't have the hostname for the server at this point, `Rails.application.routes.url_helpers.compute_resource_url` fails with: 
`ArgumentError: Missing host to link to! Please provide the :host parameter, set default_url_options[:host], or set :only_path to true`

ares: I guess we could at least print compute resource name, migration/installer will be run by Foreman maintainer, they should know how to check the specific compute resource.

tbrisker: @ares we already do that, compute resource's `to_s` returns the resource name.

ares: ah, I was looking at this line only :-), works for me, any more comments @dLobatog ?

tbrisker:
ping @dLobatog ?

xprazak2:
Things works for me now, duplicate images get renamed and proper message is printed:

ares (approved):
Thanks @tbrisker, the migration works nicely. Thanks @xprazak2 and @dLobatog for looking into it too, I believe all comments were resolved so I'll merge.

PR #5082, opened 2017-12-13
Fixes #21875 - added support for sha512 grub passwords

dmitri-d:
All new hosts provisioned using default and rhel
	kickstarts will use sha512 grub passwords from
	now on.


dLobatog (approved):
Thanks @witlessbird !

PR #5083, opened 2017-12-14
Fixes #21846 - React turbolinks fix

sharvit:
In order to support turbolinks with react, need to
unmount all root components before turbolinks do the unload

It fixes the issue with the notification drawer and the on-click-outside
http://projects.theforeman.org/issues/21846

When I dived deeper into this issue, I found out that children components (components that are not apart of the componentsRegistry) getting mounted more than once.
Each time a page-change is triggering by turbolinks, it is adding another call to the `componentDidMount` method.

So for example, if i click on the top bar logo (to go/refresh homepage) 5 times, it will call the `componentDidMount` 6 times at the last click!

Libraries like `webpacker-react` and `react-rails` are handling it internally:
https://github.com/reactjs/react-rails/blob/master/react_ujs/src/events/turbolinks.js
https://github.com/renchap/webpacker-react/blob/master/javascript/webpacker_react-npm-module/src/ujs.js#L20-L40

Later on, we will want to remove this "hack" after we will switch to `webpacker-react` as part of the move into `webpacker`.
https://github.com/theforeman/foreman/pull/5033

Because it was impossible to test it with enzyme, I added an integration test to verify it.
If removing `document.addEventListener('page:before-unload', () => { ... });` from the `MountingService.js` those tests will fail.

I closed this related pr after I understand that another type of fix is needed:
https://github.com/theforeman/foreman/pull/5047


webpack/assets/javascripts/react_app/components/notifications/index.js
```diff
@@ -43,7 +43,7 @@ class notificationContainer extends React.Component {
     } = this.props;
 
     return (
-      <div id="notifications_container">
+      <div>
```
sharvit, line 46: Removed the `id` field because this component getting mounted into a div with the same `id`.

webpack/assets/javascripts/react_app/components/notifications/notifications.test.js
```diff
@@ -137,17 +137,4 @@ describe('notifications', () => {
     wrapper.find(closeButtonSelector).simulate('click');
     expect(toJson(wrapper)).toMatchSnapshot();
   });
-
-  it('should close the notification box when click outside of the box', () => {
```
sharvit, line 141: Removed this test because it is not doing the job.
Replaced it with integration test.

test/integration/notifications_drawer_test.rb
```diff
@@ -0,0 +1,53 @@
...
+    # close notifications drawer by click outside
+    page.find('body').click
+    assert page.has_no_selector?('#notifications_container .drawer-pf'), "Notifications drawer was expected to be closed, but was found opend"
+  end
+
+  def navigate_somewhere_with_turbolinks
+    # check the outside click with turbolinks
+    page.find('a.navbar-brand-text').click
+    # wait for the reload
+    sleep(1)
```
ohadlevy: maybe use wait_for_ajax or Simpy use visit?

sharvit: * wait_for_ajax do not wait for turbolinks
* visit do not trigger any turbolinks page change

I updated to use a different approach which should be better than `wait(1)`.

ohadlevy (approved):

PR #5085, opened 2017-12-14
Refs #21120 - DHCP update no longer queued twice

lzap:
This is an alternative backport-friendly version of

https://github.com/theforeman/foreman/pull/5079

In short, when PXELoader is edited for existing host, ActiveRecord triggers `valid?` call twice which then enqueues DHCP orchestration twice. Starting from proxy version 1.15 all invalid records are now explicitly not allowed and proxy now errors out with 409 conclits http code. This patch prevents from enqueing the same tasks twice, thus this won't happend and PXELoader can be edited.

This patch is not meant to be merged into `develop` branch as more complex PR is prepared that solves this globally. Once this is reviewed, this can be backported into 1.15 and 1.16 to solve the edit host issue.

When testing this, be aware of bug in `develop` branch of proxy: http://projects.theforeman.org/issues/21975

### DIFF
--- a/app/models/concerns/orchestration/dhcp.rb
+++ b/app/models/concerns/orchestration/dhcp.rb
@@ -149,10 +149,10 @@ module Orchestration::DHCP
   def queue_dhcp_update
     return unless dhcp_update_required?
     logger.debug("Detected a changed required for DHCP record")
-    queue.create(:name => _("Remove DHCP Settings for %s") % old, :priority => 5,
-                 :action => [old, :del_dhcp]) if old.dhcp?
-    queue.create(:name   => _("Create DHCP Settings for %s") % self, :priority => 9,
-                 :action => [self, :set_dhcp]) if dhcp?
+    remove_name = _("Remove DHCP Settings for %s") % old
+    create_name = _("Create DHCP Settings for %s") % self
+    queue.create(:name => remove_name, :priority => 5, :action => [old, :del_dhcp]) if old.dhcp? && queue.items.select{|t| t.name == remove_name}.empty?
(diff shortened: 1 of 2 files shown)


treydock:
Patched 1.15.6 test environment with this PR and verified I was able to change the PXELoader with no error.  Only odd thing I noticed was the popup dialogue for overwriting DHCP had 2 entries listed that were duplicates but seems harmless as dhcpd.leases has correct changes and no duplicates.

dLobatog (approved):
Same feedback as @treydock , I think for 1.16.1 this ought to be a good fix. Thanks @lzap !

dLobatog:
This PR is submitted against develop, so I'll just cp to 1.16.

dLobatog:
Cherry-picked to 1.16-stable as affa8c7025c08b1ae3e5107e8ba5a876a44d2816

lzap:
I propose this: Let's merge this into develop as well because we are very close to 1.17 branching and this could be valuable patch to have in 1.17. I will rebase my full solution PR on top of this then, it will require much more work. @dLobatog ?

lzap:
@treydock thanks the popup was not supposed to be fixed by this at this point, thanks for testing!

ares:
@lzap would you mind rebasing? seems like tests are broken since the excon is not pinned in your version, then I think we're good to merge into develop

lzap:
@ares sure done, although @dLobatog prefers cherry picking it into 1.17. I think merging into develop is cleaner tho.

dLobatog:
@lzap I'm fine with both options, it's just I thought initially this PR was only meant for 1.16

lzap:
Ok cool, tests are green :-)

dLobatog (approved):
There you go, ready for 1.17  - thanks @lzap

PR #5088, opened 2017-12-15
Fixes #21990: update patternfly-react to 0.18.2.

waldenraines:
http://projects.theforeman.org/issues/21990


danseethaler (approved):
This introduces the Spinner component which we're using in Katello. I mentioned to @sharvit to integrate this when possible.

waldenraines:
> This introduces the Spinner component which we're using in Katello. I mentioned to @sharvit to integrate this when possible.

See also http://projects.theforeman.org/issues/21992

ohadlevy (approved):

PR #5090, opened 2017-12-16
Refs #21424 - Pin factory_girl_rails to ~> 4.8.0

mmoll:
Version 4.9.0 emits a deprecation warning which breaks the CI tests.
Rather than upgrade the stable branch it's easier and less disruptive to
pin.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,5 +5,5 @@ node_js:
   - "4.6" # Fedora 24
   - "6.9" # EL 7
 before_install:
-  npm install -g npm@'>=3.0.0'
+  npm install -g npm@'>=3.0.0, <5.0.0'
 script: ./script/travis_run_js_tests.sh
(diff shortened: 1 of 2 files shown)


mmoll:
This is for 1.15-stable. There are some plugin tests (e.g. for discovery 9.1) that run against 1.15.

mmoll:
Added another commit to fix the JS test on travis.

tbrisker:
@mmoll I assume that the katello failure is to be expected?

mmoll:
@tbrisker yes, because the current Katello master is tested there against Foreman 1.15

tbrisker (approved):
ACK from me, but I'll leave it to @dLobatog who is the release manager for 1.15 branch to pull the trigger.

dLobatog (approved):
Thanks @mmoll @tbrisker

PR #5091, opened 2017-12-17
Refs #21997 - pin excon gem

mmoll:


mmoll:
excon is a dependency of fog-core, taking the `~> 0.58` requirement from it. The pinning is to fix the tests for now. Once the RPM packaging is on tfm-ror51, the real update task can be worked on.

boaz0 (approved):
LGTM 💯

tbrisker (changes requested):
@mmoll can you please add a comment inside the Gemfile so we don't forget why this was pinned and keep it like that forever?

mmoll:
sure, updated

tbrisker (approved):

tbrisker:
APJ

tbrisker:
Test failure unrelated. Thanks @mmoll !

PR #5093, opened 2017-12-17
Fixes #21998 - Org admin can view org

tbrisker:


tbrisker:
@ares - not sure why the view_org permission was excluded, why shouldn't an org admin be able to view their own org? this also causes anything that uses the organization or organization-id parameter to the api to fail, e.g. content hosts page.

tbrisker:
Looks like this was already fixed by GH-4917, closing.

ares:
It was completely broken. It worked  by accident when user did not have the permission, which was temporary solution until the proper fix would be merged. That luckilly happened yesterday in PR you linked.

PR #5096, opened 2017-12-18
Fixes #18675 - telemetry foreman API

lzap:
Foreman telemetry core API. This PR brings a common telemetry API which provides three generic metric types:

* **counter** (monotony increasing counter - for example number of http requests)
* **gauge** (arbitrary float value - for example number of jobs queued)
* **histogram** (arbitrary amount of buckets with observations and float values - for example duration of a request)

Each individual metric must be predefined in config/initializers/5_telemetry_metrics.rb file in this format:

    telemetry.add_counter(:http_requests, 'A counter of HTTP requests made', [:controller, :action])

Optionally buckets can be defined for histograms (a sane default is used if not defined):

    telemetry.add_histogram(:http_request_total_duration, 'Total duration of controller action', [:controller, :action], [5, 10, 50, 200, 1000])

For more about why I picked these, watch my proposal video: https://www.youtube.com/watch?v=gCLSI9-4QpE

There are two implementations for the best possible integration options: Prometheus endpoint and Statsd packets. To configure the former, add this to your settings:

```yaml
:telemetry:
  :type: 'prometheus'
  :prefix: 'fm_rails'
```

Then visit `/metrics` to see aggregated data. Warning - due to limitation of Prometheus Ruby library, this will not work for multi-process web servers (our case) and this will provide you incorrect data. Therefore, until this is fixed upstream, a recommended way is to use Statsd:

```yaml
:telemetry:
  :type: 'statsd'
  :prefix: 'fm_rails'
  :statsd:
    :host: '127.0.0.1:8125'
    :protocol: 'statsd'
```

Then download and run any statsd aggregator on localhost port 8125. I tested with statsite and statsd_exporter for Prometheus, which is super easy to use:

```
./statsd_exporter -statsd.listen-udp :8125
```

This PR provides a rake task that will create useful mapping for statsd_exporter which makes life much more easier and provides label mapping:

```
be rake telemetry:prometheus_statsd output=/tmp/mapping.yaml
./statsd_exporter -statsd.listen-udp :8125 -statsd.mapping-config /tmp/mapping.yaml
```

The last step is to grab the data into some monitoring framework, Prometheus is very easy to use as it's a single binary and configuration is easy:

```
scrape_configs:
  - job_name: 'foreman'
    static_configs:
      - targets: ['localhost:9102']

./prometheus
```

This PR adds bunch of basic telemetry data, we can add more later on via initializer, this could be DSL instead of direct Ruby code (config/initializers/5_telemetry_metrics.rb). The PR currently adds:

| Metric name | Labels | Type | Description |
| ----------- | ------ | ---- | ----------- |
| fm_rails_activerecord_instances | class | counter | Number of instances of ActiveRecord models |
| fm_rails_bruteforce_locked_ui_logins |  | counter | Number of blocked logins via bruteforce protection |
| fm_rails_failed_ui_logins |  | counter | Number of failed logins in total |
| fm_rails_http_request_db_duration | controller,action | histogram | Time spent in database for a request |
| fm_rails_http_request_total_duration | controller,action | histogram | Total duration of controller action |
| fm_rails_http_request_view_duration | controller,action | histogram | Time spent in view for a request |
| fm_rails_http_requests | controller,action | counter | A counter of HTTP requests made |
| fm_rails_proxy_api_duration | method | histogram | Time spent waiting for Proxy (ms) |
| fm_rails_proxy_api_response_code | code | counter | Number of Proxy API responses per HTTP code |
| fm_rails_ruby_gc_allocated_objects | controller,action | counter | Ruby GC statistics per request (total_allocated_objects) |
| fm_rails_ruby_gc_count | controller,action | counter | Ruby GC statistics per request (count) |
| fm_rails_ruby_gc_freed_objects | controller,action | counter | Ruby GC statistics per request (total_freed_objects) |
| fm_rails_ruby_gc_major_count | controller,action | counter | Ruby GC statistics per request (major_gc_count) |
| fm_rails_ruby_gc_minor_count | controller,action | counter | Ruby GC statistics per request (minor_gc_count) |
| fm_rails_successful_ui_logins |  | counter | Number of successful logins in total |

A short demo how to set thigs up:

https://youtu.be/i5iCOLEByZk

TODO:

* [x] Installer
* [x] Documentation
* [x] Packages for new dependencies

### DIFF
--- a/config.ru
+++ b/config.ru
@@ -1,6 +1,17 @@
 # This file is used by Rack-based servers to start the application.
 
+require 'rack'
+require './lib/foreman/telemetry_rack'
+require 'prometheus/middleware/exporter'
+
+# load Rails environment
 require ::File.expand_path('../config/environment',  __FILE__)
+
+use Foreman::TelemetryRack
+if defined?(Prometheus::Middleware::Exporter) && SETTINGS[:telemetry].try(:fetch, :prometheus).try(:fetch, :enabled)
+  use Prometheus::Middleware::Exporter
+end
+
 # apply a prefix to the application, if one is defined
 # e.g. http://some.server.com/prefix where '/prefix' is defined by env variable
 map ENV['RAILS_RELATIVE_URL_ROOT'] || '/'  do
--- a/config/initializers/5_telemetry_metrics.rb
+++ b/config/initializers/5_telemetry_metrics.rb
@@ -1,0 +1,19 @@
+# Foreman telemetry metrics registration.
+#
+# There are three types of telemetry measurements: counter, gauge, histogram. Each measurement has unique name
+# represented by a symbol, a description and list of tags which are required and mapped to monitoring frameworks
+# which do not support arbitrary tags.
+#
+# Plugins can add own metrics through add_counter_telemetry, add_gauge_telemetry and add_histogram_telemetry.
+#
+telemetry = Foreman::Telemetry.instance
+telemetry.add_counter(:http_requests, 'A counter of HTTP requests made', [:controller, :action])
+telemetry.add_histogram(:http_request_total_duration, 'Total duration of controller action', [:controller, :action])
+telemetry.add_histogram(:http_request_db_duration, 'Time spent in database for a request', [:controller, :action])
+telemetry.add_histogram(:http_request_view_duration, 'Time spent in view for a request', [:controller, :action])
+telemetry.add_counter(:activerecord_instances, 'Number of instances of ActiveRecord models', [:class])
+telemetry.add_counter(:successful_ui_logins, 'Number of successful logins in total')
+telemetry.add_counter(:failed_ui_logins, 'Number of failed logins in total')
+telemetry.add_counter(:bruteforce_locked_ui_logins, 'Number of blocked logins via bruteforce protection')
+telemetry.add_histogram(:proxy_api_duration, 'Time spent waiting for Proxy (ms)', [:method])
+telemetry.add_counter(:proxy_api_response_code, 'Number of Proxy API responses per HTTP code', [:code])
--- a/config/settings.yaml.example
+++ b/config/settings.yaml.example
@@ -70,6 +70,29 @@
 #    :enabled: true
 #  :dynflow:
 #    :enabled: true
+#  :telemetry:
+#    :enabled: true
+
+# Foreman telemetry has three destinations: prometheus, statsd and rails log.
+:telemetry:
+  # prefix for all metrics
+  :prefix: 'fm_rails'
+  # prometheus endpoint is at /metrics
+  # warning: ruby client library currently does not supprt multi-process web servers
+  :prometheus:
+    :enabled: false
+  # works with statsd_exporter too, use the rake task to generate config
+  :statsd:
+    :enabled: false
+    # IP and port (do not use DNS)
+    :host: '127.0.0.1:8125'
+    # one of 'statsd', 'statsite' or 'datadog'
+    :protocol: 'statsd'
+  # Rails logs end up in logger named 'telemetry' when enabled
+  :logger:
+    :enabled: false
+    # logging level as in Logger::LEVEL
+    :level: 'DEBUG'
 
 #
 # Configure how many workers should dynflow use to handle incoming requests.
--- a/lib/foreman/telemetry_sinks/rails_logger_sink.rb
+++ b/lib/foreman/telemetry_sinks/rails_logger_sink.rb
@@ -1,0 +1,32 @@
+module Foreman::TelemetrySinks
+  class RailsLoggerSink
+    def initialize(opts = {})
+      @logger = Foreman::Logging.logger('telemetry')
+      @level = Logger.const_get(opts[:level].try(:to_sym).try(:upcase) || :debug)
+    end
+
+    def add_counter(name, description, instance_labels)
+      @logger.add(@level, "Registering counter #{name} labels #{instance_labels.inspect}")
+    end
+
+    def add_gauge(name, description, instance_labels)
+      @logger.add(@level, "Registering gauge #{name} labels #{instance_labels.inspect}")
+    end
+
+    def add_histogram(name, description, instance_labels, ignored)
+      @logger.add(@level, "Registering histogram #{name} labels #{instance_labels.inspect}")
+    end
+
+    def increment_counter(name, value, tags)
+      @logger.add(@level, "Incrementing counter #{name} by #{value} #{tags.inspect}")
+    end
+
+    def set_gauge(name, value, tags)
+      @logger.add(@level, "Setting gauge #{name} to #{value} #{tags.inspect}")
+    end
+
+    def observe_histogram(name, value, tags)
+      @logger.add(@level, "Observing histogram #{name} value #{value} #{tags.inspect}")
+    end
+  end
+end
--- a/lib/tasks/telemetry.rake
+++ b/lib/tasks/telemetry.rake
@@ -1,0 +1,41 @@
+require "foreman/telemetry"
+require "foreman/telemetry_sinks/metric_exporter_sink"
+
+desc "Telemetry helper tasks"
+namespace :telemetry do
+  desc "List all metrics"
+  task :metrics => :environment do
+    telemetry = Foreman::Telemetry.instance
+    puts "| Metric name | Labels | Type | Description |"
+    puts "| ----------- | ------ | ---- | ----------- |"
+    telemetry.metrics.sort.each do |element|
+      m_name, m_desc, m_labels, m_type, = element
+      puts "| #{m_name} | #{m_labels.join(',')} | #{m_type} | #{m_desc} |"
+    end
+  end
+
+  desc "Generate exporter mapping for statsd_exporter"
+  task :prometheus_statsd, [:output] => [:environment] do |t|
+    telemetry = Foreman::Telemetry.instance
+    File.open(ENV["output"] || "mapping.yaml", "w") do |f|
+      mappings = []
+      telemetry.metrics.sort.each do |element|
+        m_name, m_desc, m_labels, m_type, m_buckets = element
+        metric = {}
+        labels = {}
+        metric["name"] = m_name
+        metric["match"] = m_name + (".*" * m_labels.count)
+        m_labels.each_with_index do |label, i|
+          labels[label.to_s] = "$#{i+1}"
+        end
+        metric["labels"] = labels
+        metric["help"] = m_desc
+        metric["buckets"] = m_buckets.dup if m_buckets
+        metric["timer_type"] = "histogram" if m_type == :histogram
+        mappings << metric if m_labels.count > 0
+      end
+      result = {"mappings" => mappings}
+      f.puts result.to_yaml
+    end
+  end
+end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -3,6 +3,7 @@ class UsersController < ApplicationController
   include Foreman::Controller::UsersMixin
   include Foreman::Controller::Parameters::User
   include Foreman::Controller::BruteforceProtection
+  include Foreman::TelemetryHelper
 
   skip_before_action :require_mail, :only => [:edit, :update, :logout]
   skip_before_action :require_login, :authorize, :session_expiry, :update_activity_time, :set_taxonomy, :set_gettext_locale_db, :only => [:login, :logout, :extlogout]
@@ -69,6 +70,7 @@ class UsersController < ApplicationController
     if bruteforce_attempt?
       inline_error _("Too many tries, please try again in a few minutes.")
       log_bruteforce
+      telemetry_increment_counter(:bruteforce_locked_ui_logins)
       render :layout => 'login', :status => :unauthorized
       return
     end
@@ -85,6 +87,7 @@ class UsersController < ApplicationController
         #failed to authenticate, and/or to generate the account on the fly
         inline_error _("Incorrect username or password")
         count_login_failure
+        telemetry_increment_counter(:failed_ui_logins)
         redirect_to login_users_path
       else
         #valid user
@@ -161,6 +164,7 @@ class UsersController < ApplicationController
     store_default_taxonomy(user, 'organization') unless session.has_key?(:organization_id)
     store_default_taxonomy(user, 'location') unless session.has_key?(:location_id)
     TopbarSweeper.expire_cache
+    telemetry_increment_counter(:successful_ui_logins)
     user.post_successful_login
     redirect_to (uri || hosts_path)
   end
--- a/app/registries/foreman/plugin.rb
+++ b/app/registries/foreman/plugin.rb
@@ -441,6 +441,18 @@ module Foreman #:nodoc:
       @rabl_template_extensions.fetch(virtual_path, [])
     end
 
+    def add_counter_telemetry(name, description, instance_labels = [])
+      Foreman::Telemetry.instance.add_counter(name, description, instance_labels)
+    end
+
+    def add_gauge_telemetry(name, description, instance_labels = [])
+      Foreman::Telemetry.instance.add_gauge(name, description, instance_labels)
+    end
+
+    def add_histogram_telemetry(name, description, instance_labels = [], buckets = DEFAULT_BUCKETS)
+      Foreman::Telemetry.instance.add_histogram(name, description, instance_labels, buckets)
+    end
+
     private
 
     def permission_table_exists?
--- a/app/services/foreman/telemetry_helper.rb
+++ b/app/services/foreman/telemetry_helper.rb
@@ -1,0 +1,32 @@
+module Foreman::TelemetryHelper
+  extend ActiveSupport::Concern
+
+  def telemetry_increment_counter(name, value = 1, tags = {})
+    Foreman::Telemetry.instance.increment_counter(name, value, tags)
+  end
+
+  def telemetry_set_gauge(name, value, tags = {})
+    Foreman::Telemetry.instance.set_gauge(name, value, tags)
+  end
+
+  def telemetry_observe_histogram(name, value, tags = {})
+    Foreman::Telemetry.instance.observe_histogram(name, value, tags)
+  end
+
+  # time spent in a block as histogram, in miliseconds by default
+  def telemetry_duration_histogram(name, scale = 1000, tags = {})
+    case scale
+    when :ms, :msec, :miliseconds
+      scale = 1000
+    when :sec, :seconds
+      scale = 1
+    when :min, :minutes
+      scale = 1 / 60
+    end
+    before = Time.now.to_f
+    yield
+  ensure
+    after = Time.now.to_f
+    telemetry_observe_histogram(name, (after - before) * scale, tags)
+  end
+end
--- a/bundler.d/telemetry.rb
+++ b/bundler.d/telemetry.rb
@@ -1,0 +1,4 @@
+group :telemetry do
+  gem 'prometheus-client'
+  gem 'statsd-instrument'
+end
--- a/config/initializers/5_telemetry.rb
+++ b/config/initializers/5_telemetry.rb
@@ -1,0 +1,11 @@
+# Foreman telemetry global setup.
+telemetry = Foreman::Telemetry.instance
+if SETTINGS[:telemetry] && (Rails.env.production? || Rails.env.development?)
+  telemetry.setup(SETTINGS[:telemetry])
+end
+
+# Register Rails notifications metrics
+telemetry.register_rails
+
+# Register Ruby VM metrics
+telemetry.register_ruby
--- a/lib/foreman/telemetry.rb
+++ b/lib/foreman/telemetry.rb
@@ -1,0 +1,106 @@
+require 'singleton'
+require 'forwardable'
+
+module Foreman
+  class Telemetry
+    include Singleton
+    extend Forwardable
+    attr_accessor :prefix, :sinks
+
+    DEFAULT_BUCKETS = [10, 50, 200, 1000, 15000].freeze
+
+    def initialize
+      @sinks = []
+      @exporter = ::Foreman::TelemetrySinks::MetricExporterSink.new
+    end
+
+    def setup(opts = {})
+      @prefix = opts[:prefix] || ''
+      @sinks << @exporter
+      unless Foreman.in_rake?
+        @sinks << ::Foreman::TelemetrySinks::RailsLoggerSink.new(opts[:logger]) if opts[:logger] && opts[:logger][:enabled]
+        @sinks << ::Foreman::TelemetrySinks::PrometheusSink.new(opts[:prometheus]) if opts[:prometheus] && opts[:prometheus][:enabled]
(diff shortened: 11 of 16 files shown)


dLobatog (commented):
Thanks @lzap ! The demo looks great, a good deal of work was clearly put into this. I used to use NewRelic to monitor Foreman servers and the data clearly helped to guide how our services had to be laid out. 

Missing pieces:

* [ ] Installer - should it provide just defaults and an aggregator, or nothing at all. 
* [ ] Documentation - at least a small section explaining what we have to do to configure this would be important for the next release with this (1.18)
* [ ] Packaging - we need a few packages for this 

What happens if there is no statsd aggregator available?
What are the possibilities to put Pulp telemetry data into the same aggregator as us?
Should we ship one by default (I don't think so)?
Are there plans on stuff to do with this data?

lzap:
Thanks for the initial feedback. I created TODO, I don't plan to work on the installer as by default this is turned off. But there is a long-term goal to integrate PCP into our installation, that would be perhaps better task. I will definitely work on documentation, I have a very lengthy blog post ready and I am commited to packaging as well. I want to settle up and merge this into develop and then I will do the rest.

> What happens if there is no statsd aggregator available?

Good question, for every measurement there is an INFO Rails log level message, which is quite nasty, I am going to add a SamplingLogger implementation that will only log every 100th message and it will have an option to turn these errors off. This floods Rails log which is unnecessary.

> What are the possibilities to put Pulp telemetry data into the same aggregator as us?

Not sure what you mean, but the purpose of this PR is to add Foreman core telemetry, adding telemetry to backend services are separate tasks. If Pulp offers some kind of telemetry today, we should rather write a collector agent for Prometheus or PCP to get these.

> Should we ship one by default (I don't think so)?

Not sure what you mean, but my initial version is opt-in only, ultimate goal is to ship Foreman with PCP (or maybe Prometheus who knows) and then we could have full stack installation, but not at this stage.

> Are there plans on stuff to do with this data?

As I said, the ultimate goal is to have PCP installed and configured to gather all data, I will publish an extensive blog post about how to set things up (manually at this point). This will allow to ask users or customers to configure PCP accordingly and then send us recorded data as PCP archives, including Foreman telemetry. This will be very useful in order to identify production bottlenecks.

config/settings.yaml.example
```diff
@@ -66,6 +66,23 @@
 #  :dynflow:
 #    :enabled: true
 
+:telemetry:
+  # one of 'prometheus', 'statsd', 'rails' or 'off'.
```
iNecas: Should this be one and only one? I can imagine wanting to enable logger as well as prometheus at the same time

lzap: Not sure what you mean, but rails log is not very useful, try it yourself. It will be used in very rare situations, otherwise it floods logs.

iNecas: Still, seems to me like unnecessary limitation: I can imagine a case, where the foreman is configured in the infrastructure to use one provider (let's say `prometheus`) and we would ask the user to turn on `statsd` because we could than process further data.

I'm not sure if this will be the case or not, but if I have two options, where one is more strong, and I don't see a reason why we should not do that, the multi-provider option wins for me and it's much easier to do now, than few years from now, when the request arises.

lib/tasks/telemetry.rake
```diff
@@ -0,0 +1,48 @@
...
+
+namespace :telemetry do
+  desc <<-END_DESC
+Telemetry tasks.
+END_DESC
+
+
+  task :metrics do
+    telemetry = initialize_telemetry
+    puts "| Metric name | Labels | Type | Description |"
```
iNecas: Tbh the output of this is very unreadable. What about just printing the info in yaml format instead?

Gemfile
```diff
@@ -55,6 +55,10 @@ gem 'dynflow', '>= 0.8.29', '< 1.0.0'
 gem 'daemons'
 gem 'get_process_mem'
 
+# telemetry
+gem 'prometheus-client'
```
iNecas: Shouldn't this dependencies be optional?

config/settings.yaml.example
iNecas: The prometheus provider should have a big warning there that it's not meant for production setup, to avoid some confusion and false expectations this could bring to the users.

config.ru
```diff
@@ -1,5 +1,12 @@
 # This file is used by Rack-based servers to start the application.
 
+require 'rack'
+require './lib/foreman/telemetry_rack'
+require 'prometheus/middleware/exporter'
+
+use Foreman::TelemetryRack
+use Prometheus::Middleware::Exporter
```
iNecas: This should be used only when prometheus provider is being used.

config.ru
```diff
@@ -1,5 +1,12 @@
 # This file is used by Rack-based servers to start the application.
 
+require 'rack'
+require './lib/foreman/telemetry_rack'
+require 'prometheus/middleware/exporter'
+
+use Foreman::TelemetryRack
```
iNecas: What about using this only when telemetry is enabled?

config/settings.yaml.example
lzap: I did not get your initial comment, indeed that's a good idea and config is even more readable after the change.

lib/tasks/telemetry.rake
lzap: This is github/markdown compatible table output, the goal is to use this to generate and easily update documentation table.

Gemfile
lzap: Do you mean separate bundler groups for both prometheus and statsd? It's part of core settings.yaml there is no precedent. Plugins have usually only Administer Settings, but this needs to be setup early in the boot so it is actually available for all plugins. I'd rather waive this into core dependencies. Or explain how do you think this should be done in more detail.

config/settings.yaml.example
lzap: Done.

config.ru
lzap: Chicken and egg problem, the middleware must be used, but I've added a check in the middleware, if disabled no GC stat is called.

config.ru
lzap: Chicken and egg, Telemetry hasn't been initialized at this moment, I don't think I should initialize is this early. Having that enabled means that Prometheus will respond at `/metrics` exposing no measures (basically an empty request).

Possible solution - move telemetry into separate bundler group and only if the gem is installed then we would use it here.

Gemfile
lzap: Moved to own bundler group, added a note in YAML settings that this must be installed prior enabling telemetry.

lzap:
Incorporated @iNecas comments.

lzap:
Rebased, fixed all rubocop items, @dLobatog @iNecas @ehelms ping

config/settings.yaml.example
```diff
@@ -71,6 +71,24 @@
...
+  :prometheus:
+    :enabled: false
+  :statsd:
+    :enabled: false
+    # IP and port (do not use DNS)
+    :host: '127.0.0.1:8125'
+    # one of 'statsd', 'statsite' or 'datadog'
+    :protocol: 'statsd'
+  :rails:
+    :enabled: false
```
ehelms: If this is false by default, what metrics are collected by default? Just ruby?

config/initializers/5_telemetry_ruby.rb
```diff
@@ -0,0 +1,17 @@
...
+begin
+  GC.stat :total_allocated_objects
+rescue ArgumentError
+  GC_METRICS.update :total_allocated_object => :ruby_gc_allocated_objects, :total_freed_object => :ruby_gc_freed_objects
+else
+  GC_METRICS.update :total_allocated_objects => :ruby_gc_allocated_objects, :total_freed_objects => :ruby_gc_freed_objects
+end
+GC_METRICS.each do |ruby_key, metric_name|
+  telemetry.add_counter(metric_name, "Ruby GC statistics per request (#{ruby_key})", [:controller, :action])
+end
```
ehelms: Could the code in this file live under `lib/foreman/telemetry` ? If this is required to be initialized here, could the code live there but be a one line call out to it? I think it might be more obvious code wise to have everything under `lib/foreman/telemetry` witha simple call here like `telemetry.configure_ruby`

config/initializers/5_telemetry_notifications.rb
```diff
@@ -0,0 +1,27 @@
...
+    telemetry.increment_counter(metric_name, after[ruby_key] - before[ruby_key], :controller => controller, :action => action) if after.include?(ruby_key)
+  end
+end if telemetry.enabled?
+
+ActiveSupport::Notifications.subscribe(/instantiation.active_record/) do |*args|
+  event = ActiveSupport::Notifications::Event.new(*args)
+  class_name = event.payload[:class_name]
+  record_count = event.payload[:record_count]
+  telemetry.increment_counter(:activerecord_instances, record_count, :class => class_name)
+end if telemetry.enabled?
```
ehelms: Could the code in this file live under `lib/foreman/telemetry` ? If this is required to be initialized here, could the code live there but be a one line call out to it? I think it might be more obvious code wise to have everything under `lib/foreman/telemetry` witha simple call here like `telemetry.notifications`

config/initializers/5_telemetry_metrics.rb
```diff
@@ -0,0 +1,12 @@
...
+# There are three types of telemetry measurements: counter, gauge, histogram. Each measurement has unique name
+# represented by a symbol, a description and list of tags which are required and mapped to monitoring frameworks
+# which do not support arbitrary tags.
+#
+telemetry = Foreman::Telemetry.instance
+telemetry.add_counter(:http_requests, 'A counter of HTTP requests made', [:controller, :action])
+telemetry.add_histogram(:http_request_total_duration, 'Total duration of controller action', [:controller, :action])
+telemetry.add_histogram(:http_request_db_duration, 'Time spent in database for a request', [:controller, :action])
+telemetry.add_histogram(:http_request_view_duration, 'Time spent in view for a request', [:controller, :action])
+telemetry.add_counter(:activerecord_instances, 'Number of instances of ActiveRecord models', [:class])
```
ehelms, line 14: Could the code in this file live under `lib/foreman/telemetry` ? If this is required to be initialized here, could the code live there but be a one line call out to it? I think it might be more obvious code wise to have everything under `lib/foreman/telemetry` witha simple call here like `telemetry.define_metrics`

lib/tasks/telemetry.rake
```diff
@@ -0,0 +1,47 @@
+require "foreman/telemetry"
```
ehelms, line 1: Is this rake task simply for printing metrics to a console by a user on the box?

config/settings.yaml.example
```diff
@@ -71,6 +71,24 @@
 #  :dynflow:
 #    :enabled: true
 
+# before enabling any telemetry install foreman-telemetry package (telemetry bundler group)
+:telemetry:
```
ehelms: What is the default state? Publishing stats to the logger?

lib/foreman/telemetry_sinks/rails_logger_sink.rb
```diff
@@ -0,0 +1,31 @@
+module Foreman::TelemetrySinks
+  class RailsLoggerSink
+    def initialize(opts = {})
+      @level = Logger.const_get(opts[:level].try(:to_sym).try(:upcase) || :debug)
+    end
+
+    def add_counter(name, description, instance_labels)
+      Rails.logger.add(@level, "Registering counter #{name} labels #{instance_labels.inspect}")
```
ehelms: The current logging setup provides the ability to log to a special logger (e.g. [app] or [sql]). I think this could be an excellent use case that when using the log sink to log to a [stats] or [telemetry] or [telem] logger for easier parsing.

ehelms:
Just to help my own clarity, the built in Prometheus exporter does not work for our use case. But if statsd is used, it can be combined with a statsd Prometheus exporter to achieve a Prometheus workflow without any caveats other than having a middleman in statsd present?

config/settings.yaml.example
lzap: This is actually destination, I will rename this to `rails_log` to have this less confusing.

lib/tasks/telemetry.rake
lzap: One task is just to print them, the other one is actually to generate statsd_exporter configuration so we can use statsd endpoint and still integrate it with Prometheus: https://github.com/prometheus/statsd_exporter

config/settings.yaml.example
lzap: Default state is disabled completly. I renamed `rails` to `logging` now, hope it's more clear. You should see three `enabled: false` lines in this block.

lib/foreman/telemetry_sinks/rails_logger_sink.rb
lzap: Good idea, added. I named it `telemetry` because it does not matter, this is more for debugging than anything in production.

lzap:
Exactly, I prepared Foreman for the future, but we can still use `statsd_exporter`, we even have configuration generator that maps everything properly so you see what you expect in Prom. Users can also integrate with old statsd if they want to.

config/initializers/5_telemetry_metrics.rb
lzap: I moved Ruby and Rails out of initializers, but I'd like to keep main initialization and this file because this is the place core developers will register their own metrics. I am now adding also Foreman Plugin support so plugins can register their own - this will land in the next rebase.

lzap:
Rebased, incorporated your comments, added helper service (concern) and to showcase how to use it I added few new metrics: proxy API request duration, proxy API response code counter and UI login attempt counters (success, failed, bruteforce protected). Finally, I added Plugin API and I will do Discovery PR to show how it looks like to use it.

I tested Prometheus native and it works here, I am on webrick standard git checkout. Maybe you have multiple processes and it does not work? Works just fine here.

lzap:
Example use of the rake task:

```
[lzap@box foreman]$ be rake telemetry:metrics
| Metric name | Labels | Type | Description |
| ----------- | ------ | ---- | ----------- |
| fm_rails_activerecord_instances | class | counter | Number of instances of ActiveRecord models |
| fm_rails_bruteforce_locked_ui_logins |  | counter | Number of blocked logins via bruteforce protection |
| fm_rails_failed_ui_logins |  | counter | Number of failed logins in total |
| fm_rails_http_request_db_duration | controller,action | histogram | Time spent in database for a request |
| fm_rails_http_request_total_duration | controller,action | histogram | Total duration of controller action |
| fm_rails_http_request_view_duration | controller,action | histogram | Time spent in view for a request |
| fm_rails_http_requests | controller,action | counter | A counter of HTTP requests made |
| fm_rails_proxy_api_duration | method | histogram | Time spent waiting for Proxy (ms) |
| fm_rails_proxy_api_response_code | code | counter | Number of Proxy API responses per HTTP code |
| fm_rails_ruby_gc_allocated_objects | controller,action | counter | Ruby GC statistics per request (total_allocated_objects) |
| fm_rails_ruby_gc_count | controller,action | counter | Ruby GC statistics per request (count) |
| fm_rails_ruby_gc_freed_objects | controller,action | counter | Ruby GC statistics per request (total_freed_objects) |
| fm_rails_ruby_gc_major_count | controller,action | counter | Ruby GC statistics per request (major_gc_count) |
| fm_rails_ruby_gc_minor_count | controller,action | counter | Ruby GC statistics per request (minor_gc_count) |
| fm_rails_successful_ui_logins |  | counter | Number of successful logins in total |
```

And exporter statsd prometheus config:

```
[lzap@box foreman]$ be rake telemetry:prometheus_statsd
---
mappings:
- name: fm_rails_activerecord_instances
  match: fm_rails_activerecord_instances.*
  labels:
    class: "$1"
  help: Number of instances of ActiveRecord models
- name: fm_rails_bruteforce_locked_ui_logins
  match: fm_rails_bruteforce_locked_ui_logins
  labels: {}
  help: Number of blocked logins via bruteforce protection
- name: fm_rails_failed_ui_logins
  match: fm_rails_failed_ui_logins
  labels: {}
  help: Number of failed logins in total
- name: fm_rails_http_request_db_duration
  match: fm_rails_http_request_db_duration.*.*
  labels:
    controller: "$1"
    action: "$2"
  help: Time spent in database for a request
  buckets:
  - 10
  - 50
  - 200
  - 1000
  - 15000
  timer_type: histogram
- name: fm_rails_http_request_total_duration
  match: fm_rails_http_request_total_duration.*.*
  labels:
    controller: "$1"
    action: "$2"
  help: Total duration of controller action
  buckets:
  - 10
  - 50
  - 200
  - 1000
  - 15000
  timer_type: histogram
- name: fm_rails_http_request_view_duration
  match: fm_rails_http_request_view_duration.*.*
  labels:
    controller: "$1"
    action: "$2"
  help: Time spent in view for a request
  buckets:
  - 10
  - 50
  - 200
  - 1000
  - 15000
  timer_type: histogram
- name: fm_rails_http_requests
  match: fm_rails_http_requests.*.*
  labels:
    controller: "$1"
    action: "$2"
  help: A counter of HTTP requests made
- name: fm_rails_proxy_api_duration
  match: fm_rails_proxy_api_duration.*
  labels:
    method: "$1"
  help: Time spent waiting for Proxy (ms)
  buckets:
  - 10
  - 50
  - 200
  - 1000
  - 15000
  timer_type: histogram
- name: fm_rails_proxy_api_response_code
  match: fm_rails_proxy_api_response_code.*
  labels:
    code: "$1"
  help: Number of Proxy API responses per HTTP code
- name: fm_rails_ruby_gc_allocated_objects
  match: fm_rails_ruby_gc_allocated_objects.*.*
  labels:
    controller: "$1"
    action: "$2"
  help: Ruby GC statistics per request (total_allocated_objects)
- name: fm_rails_ruby_gc_count
  match: fm_rails_ruby_gc_count.*.*
  labels:
    controller: "$1"
    action: "$2"
  help: Ruby GC statistics per request (count)
- name: fm_rails_ruby_gc_freed_objects
  match: fm_rails_ruby_gc_freed_objects.*.*
  labels:
    controller: "$1"
    action: "$2"
  help: Ruby GC statistics per request (total_freed_objects)
- name: fm_rails_ruby_gc_major_count
  match: fm_rails_ruby_gc_major_count.*.*
  labels:
    controller: "$1"
    action: "$2"
  help: Ruby GC statistics per request (major_gc_count)
- name: fm_rails_ruby_gc_minor_count
  match: fm_rails_ruby_gc_minor_count.*.*
  labels:
    controller: "$1"
    action: "$2"
  help: Ruby GC statistics per request (minor_gc_count)
- name: fm_rails_successful_ui_logins
  match: fm_rails_successful_ui_logins
  labels: {}
  help: Number of successful logins in total
```

dLobatog (approved):
Code wise, I'm 100% in favor of what I see here already, very little risk for a potentially high upside. @lzap Installer and documentation would be key pieces of this after it gets merged. I'm going to take a quick look at the packaging pieces & will merge as soon as possible. Thanks again!

dLobatog:
Due to the recent obal changes in packaging, I'll hold off merging until that's resolved - shouldn't take a long time

lzap:
One more rebase, sorry, after call with @iNecas he did not like exposing metrics when prometheus endpoint is disabled. I agreed, hence the patch:

```diff
diff --git a/config.ru b/config.ru
index 6dde6e1b1..84f0cff46 100644
--- a/config.ru
+++ b/config.ru
@@ -4,10 +4,14 @@ require 'rack'
 require './lib/foreman/telemetry_rack'
 require 'prometheus/middleware/exporter'
 
+# load Rails environment
+require ::File.expand_path('../config/environment',  __FILE__)
+
 use Foreman::TelemetryRack
-use(Prometheus::Middleware::Exporter) if defined?(Prometheus::Middleware::Exporter)
+if defined?(Prometheus::Middleware::Exporter) && SETTINGS[:telemetry].try(:fetch, :prometheus).try(:fetch, :enabled)
+  use Prometheus::Middleware::Exporter
+end
 
-require ::File.expand_path('../config/environment',  __FILE__)
 # apply a prefix to the application, if one is defined
 # e.g. http://some.server.com/prefix where '/prefix' is defined by env variable
 map ENV['RAILS_RELATIVE_URL_ROOT'] || '/'  do
```

Having `use Foreman::TelemetryRack` is harmless, in the code you will se it only adds gc stats if at least one sink is enabled.

iNecas:
I would like to test this a bit, but no additional comments code-wise from me. Foreman test failures seem unrelated. I'm not sure why katello is failing.

[test katello]

lzap:
Rebased, fixed exporter it now has different way of calling the rake task:

    be rake telemetry:prometheus_statsd output=/tmp/mapping.yaml

Before any metrics appear, you need to do some actions in Foreman. Then to see metrics from exporter visit http://localhost:9102/metrics and to see Prom UI go to http://localhost:9090

dLobatog (approved):
:+1: to merge now, I was able to get all the info the `telemetry_increment_counters` say - @iNecas anything you wanted to try before merging? Packaging dependencies are merged already. 

Thanks @lzap !

iNecas (approved):
Have not tested myself, but no other comments code-wise. ACK

dLobatog (commented):
Thanks @lzap , eager to see what you do with this data :smile:

lzap:
YAY! :fireworks: The plan: two blog posts and a deep dive about showing what you can do in Prometheus.

kbrock:
@lzap Very nice job working around the multi threaded Rails and Prometheus issues.

Do you have suggestions for other teams wanting to use the same approach?

1. Is there a reason you didn't introduce `ActiveSupport::Notifications` instead of the `Foreman::TelemetryHelper`? Was it performance, or no perceived benefit?
2. Is there a way to avoid defining the histograms and metrics in an initializer?
3. Is there a particular way for us to introduce similar system that would work well with yours? Maybe introduce a similar helper?
4. This almost feels like an independent gem. Similar to the statsd rails or prometheus gems. Do you have plans to extract some of this code (I know, it hasn't been in the code for a month yet...)

lzap:
Thanks @kbrock 

1. Both, the notification support in ActiveRecord is too generic for the job, definitely much slower than a simple method call. But it does not mean Telemetry API will not offer Notification interface, it's easy to do one subscriber to collect those metrics.
2. While it is technically possible to do "register metrics as you go" approach (both statsd and Prometheus are okay with this), I specifically designed it to have explicit list available before we start sending them. Some monitoring have limitations for dynamic metric names and Prometheus discourage use of them - it's a sign that you should use labels for that.
3. (and 4) I plan to extract this, but these were rather long-term plans (to use the same approach in smart-proxy). I was hoping to keep it here for some time first to see how useful this is. What extraction timeframe are we talking about? I have a busy month ahead, then it's doable.

kbrock:
@lzap Hi Lucas,

>2 Some monitoring have limitations for dynamic metric names and Prometheus discourage use of them

Thnanks. That is the part of the picture I was missing.

> 3. 4. What extraction timeframe are we talking about?

This is longer term for us. In the shorter term, I'll throw together a little bit of duplicate code. Just to kick the tires and see how this approach works for us.

Please share if you do any more blog posts around graphing and visualizing the results. For some reason, this is my personal hurdle.

Thanks again for the great work you've done here.

PR #5097, opened 2017-12-18
Fixes #20347 - remove os in db when removing it from the ui

lizagilman:
The os wasn't removed because the field was disabled
changed to read only

### DIFF
--- a/app/views/common/os_selection/_architecture.html.erb
+++ b/app/views/common/os_selection/_architecture.html.erb
@@ -1,9 +1,9 @@
 <%= fields_for item do |f| %>
   <%= select_f f, :operatingsystem_id, arch_oss, :id, :to_label,
-    {:selected => item.operatingsystem_id, :include_blank => blank_or_inherit_f(f, :operatingsystem)},
+    {:selected => item.operatingsystem_id, :include_blank => blank_or_inherit_f(f, :operatingsystem) },
     { :label => _("Operating system"),
-      :disabled => arch_oss.empty? ? true : false,
       :help_inline => :indicator,
-      :onchange => 'os_selected(this);', :'data-url' => method_path('os_selected'), :'data-type' => controller_name.singularize, :required => true}
+      :onchange => 'os_selected(this);', :'data-url' => method_path('os_selected'), :'data-type' => controller_name.singularize, :required => true,
+      :placeholder => arch_oss.empty? ? _("No options available for selected Architecture") : nil }
   %>
 <% end %>
--- a/app/views/common/os_selection/_operatingsystem.html.erb
+++ b/app/views/common/os_selection/_operatingsystem.html.erb
@@ -1,13 +1,13 @@
 <%= fields_for item do |f| %>
   <%= select_f f, :medium_id, os_media, :id, :to_label,
     {:include_blank => blank_or_inherit_f(f, :medium), :selected => item.medium_id},
-    {:label => _("Media"), :disabled => os_media.empty?,
-     :help_inline => :indicator, :onchange => 'medium_selected(this);', :'data-url' => method_path('medium_selected'),
-     :'data-type' => controller_name.singularize, :required => true }
+    {:label => _("Media"), :help_inline => :indicator, :onchange => 'medium_selected(this);',
+     :'data-url' => method_path('medium_selected'), :'data-type' => controller_name.singularize,
+     :required => true, :placeholder => os_media.empty? ? _("No options available for selected Operating System") : nil }
   %>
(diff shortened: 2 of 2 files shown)


app/views/common/os_selection/_operatingsystem.html.erb
```diff
@@ -1,7 +1,7 @@
 <%= fields_for item do |f| %>
   <%= select_f f, :medium_id, os_media, :id, :to_label,
     {:include_blank => blank_or_inherit_f(f, :medium), :selected => item.medium_id},
-    {:label => _("Media"), :disabled => os_media.empty?,
+    {:label => _("Media"), :readonly => os_media.empty?,
```
tbrisker: This is the media selection, not the os selection. the file names are confusing, the os selection is actually in the `_architecture.html.erb` partial. This field also needs to change, but it isn't the one that was requested in the original issue.

tbrisker (changes requested):
See inline.
Also, turns out that `select` does not have the `readonly` attribute (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select) - so that this makes the selector active. I'm not sure it's that bad since there are no options available for it anyways when there is no architecture. So there are two options - either just stop disabling empty selectors, or keep them disabled and add a hidden field with the same name and an empty value via js.

app/views/common/os_selection/_architecture.html.erb
```diff
@@ -2,7 +2,7 @@
   <%= select_f f, :operatingsystem_id, arch_oss, :id, :to_label,
     {:selected => item.operatingsystem_id, :include_blank => blank_or_inherit_f(f, :operatingsystem)},
     { :label => _("Operating system"),
-      :disabled => arch_oss.empty? ? true : false,
+      :readonly => arch_oss.empty? ? true : false,
```
sharvit: Isn't ` :readonly => arch_oss.empty?,` is enough?

tbrisker: It is, but since this is a select, the readonly attribute has no effect at all. I suggest completely removing this attribute, since having no options to select should be enough, I don't think we really have to disable the dropdown completely.

lizagilman:
[test foreman]
[test katello]

lizagilman:
@tbrisker 
updated

tbrisker:
@lizagilman can you please rebase on current develop so tests pass?

@rohoover can we please get confirmation on the behavior? This PR is about the operating system tab in host and hostgroup form. Currently, if e.g. the architecture field is not selected, or an architecture is selected that has no operating systems, the operating system drop down will be disabled:
image
The same is also true for the following drop downs, for media and partition table.
This leads to an issue when modifying existing objects, esp. on hostgroups. 
Since a disabled input is not sent to the server, when removing eg. the architecture, the previous operating system will remain saved in the database, leading to unexpected results when creating a new host or child hostgroup in this hostgroup.
The proposal is to not disable these empty drop-downs, simply leaving them empty, which will clean any leftover information when one of the fields is cleared.

Rohoover:
@tbrisker 
Is it possible to have some kind of messaging in the field, such as "Not Available" or "Not Applicable"... whatever makes sense, so that a user isn't presented with a bunch of empty dropdowns?

tbrisker:
@Rohoover Sure, maybe something that also points to the needed action - e.g. "No operating systems available for selected architecture"? that way the user understands they need to select e.g. an architecture that has operating systems

Rohoover:
Yes, that sounds great. I just don't want random blank drop downs that users then scratch there heads not knowing what to do.

lizagilman:
Thanks @Rohoover , @tbrisker 

Screenshot after - 

image

@Rohoover any comments?

Rohoover:
Would it be better to be more generic? Such as, "No options for selected architecture". That would probably be more reusable for any other drop downs pertaining to the primary drop down, and it's shorter. I welcome any dissent on this.

tbrisker:
@Rohoover we could do "options" but the dependency is different per field - i.e. the following fields are dependent on OS, not on architecture.

Rohoover:
@tbrisker 

Yea, I assumed there were two dependent clauses in the sentence. I was just thinking about how to make it as generic as possible. People inherently don't read instructions, so anything that is easier to scan would be better.

lizagilman:
updated

image

image

tbrisker (approved):
LGTM.

tbrisker:
@Rohoover - is this what you had in mind?

Rohoover:
bitmoji

tbrisker:
Thanks @lizagilman and @Rohoover !

PR #5100, opened 2017-12-19
Fixes #22029 - fix redux-action forms test

amirfefer:
We should use return to allow jest to check the error callback.


tbrisker (approved):
Makes sense, merging. No point waiting for jenkins to finish since this path isn't tested by it at all.

tbrisker:
Thanks @amirfefer !

PR #5103, opened 2017-12-19
Fixes #22036 - Issue with OpenStack Create Host and Security Groups

plessbd:
When testing against 1.16.0 and OpenStack Ocata
Create Host functionality failed.
this fixes it and makes the security groups a multi select field

before

Go to /hosts/new
Change Deploy On to be OpenStack

screen shot 2017-12-19 at 4 06 59 pm

after
go to /hosts/new
Change Deploy On to be OpenStack
screen shot 2017-12-19 at 4 09 40 pm

### DIFF
--- a/app/views/compute_resources_vms/form/openstack/_base.html.erb
+++ b/app/views/compute_resources_vms/form/openstack/_base.html.erb
@@ -12,7 +12,7 @@
 </div>
 <%= select_f f, :tenant_id, compute_resource.tenants, :id, :name, {}, :label => _('Tenant'), :label_size => "col-md-2" %>
 
-<%= select_f f, :security_groups, compute_resource.security_groups, :name, :name, {}, :label => _("Security group"), :label_size => "col-md-2"  %>
+<%= select_f f, :security_group, compute_resource.security_groups, :name, :name, {}, { :label => _("Security group"), :label_size => "col-md-2", :multiple => true} %>
 <%= select_f f, :nics, compute_resource.internal_networks, :id, :name,
              {}, { :label => _('Internal network'), :label_size => "col-md-2", :multiple => true } %>
 <%= selectable_f f, :network, compute_resource.address_pools, { include_blank: 'None' }, { :label => _("Floating IP network"), :label_size => "col-md-2" } %>


app/views/compute_resources_vms/form/openstack/_base.html.erb
```diff
@@ -12,7 +12,7 @@
 </div>
 <%= select_f f, :tenant_id, compute_resource.tenants, :id, :name, {}, :label => _('Tenant'), :label_size => "col-md-2" %>
 
-<%= select_f f, :security_groups, compute_resource.security_groups, :name, :name, {}, :label => _("Security group"), :label_size => "col-md-2"  %>
+<%= select_f f, :security_group, compute_resource.security_groups, :name, :name, {}, { :label => _("Security group"), :label_size => "col-md-2", :multiple => true} %>
```
boaz0, line 15: Any reason to change from `:security_groups` to `:security_group`?

boaz0 (commented):
LGTM but not sure about one thing.

lzap:
Thanks, I tried with OpenStack RDO Liberty and VM gets created but there are no security groups available. Can you tell in which OS version this flag was updated? We might need to have two branches for individual OS versions.

minnus:
Yeah not really sure whats going on here, but looks like this has been discussed in the past:

https://github.com/theforeman/foreman/pull/2140
https://github.com/theforeman/foreman/commit/e972c0e6740e92e3cc595ab14d4fb6a2567aa1f5

We got the error that Ben listed above if we use the current code base where its: "security_groups".  Changing to "security_group" makes everything work properly

tbrisker:
ok to test

minnus:
Actually, with this patch there is still something wrong.  Although we can now choose a security group with no error.  The VM gets created with the "default" security group, regardless of the choice in foreman.

plessbd:
More information in relation to the comment by @minnus 
if I edit /opt/theforeman/tfm/root/usr/share/gems/gems/fog-openstack-0.1.18/lib/fog/compute/openstack/models/server.rb
```
        def security_groups
          #requires :id

          groups = service.list_security_groups(:server_id => id).body['security_groups']

          groups.map do |group|
            Fog::Compute::OpenStack::SecurityGroup.new group.merge(:service => service)
          end
        end
```
If I just comment out the requires id seems to work an fix all the cases.  (this is without making the change already in this PR)

I dont think this is the right place to do this though.

I dont know the ins and outs of theforeman so looking for some direction as to where that method actually gets called.

it looks like https://github.com/theforeman/foreman/blob/3e94c50c7d9c0fc3f80ccdfeac651faf2458b115/app/models/concerns/fog_extensions/openstack/server.rb#L40-L44 should actually be returning an empty array and using that but it doesnt seem to be being called either...

PR #5104, opened 2017-12-20
Fixes #22043 - Allow adding invalid orgs to locs

tbrisker:
And vice-versa. Currently when a taxonomy is invalid (e.g. because some
mismatch with hosts and resources), it cannot be assigned to the
opposite taxonomy. This is due to the fact that by default,
`has_and_belongs_to_many` associations validate the associated object as
well.


ares (approved):

ares:
Thanks @tbrisker, fixes the issue for me, merging.

PR #5106, opened 2017-12-21
fixes #22055 - update bookmark loading status.

ohadlevy:
This PR both fixes the issue reported by not showing spinner if
there is an empty set of saved bookmarks (it shows None found)

It also adds missing tests cases for spinner, error and empty result sets, and replaces the Loader component with the patternfly-react Spinner component instead.

### DIFF
--- a/webpack/assets/javascripts/react_app/components/bookmarks/index.js
+++ b/webpack/assets/javascripts/react_app/components/bookmarks/index.js
@@ -1,22 +1,11 @@
 import React from 'react';
 import { connect } from 'react-redux';
-import { Dropdown, MenuItem } from 'patternfly-react';
+import { Dropdown, MenuItem, Spinner } from 'patternfly-react';
 import SearchModal from './SearchModal';
 import Bookmark from './Bookmark';
-import Loader from '../common/Loader';
 import * as BookmarkActions from '../../redux/actions/bookmarks';
 import DocumentationUrl from '../common/DocumentationLink';
 import { STATUS } from '../../constants';
-
-const getLoadingStatus = (errors, data) => {
-  if (data.length > 0) {
-    return STATUS.RESOLVED;
-  }
-  if (errors) {
-    return STATUS.ERROR;
-  }
-  return STATUS.PENDING;
-};
 
 class BookmarkContainer extends React.Component {
   componentDidMount() {
@@ -35,7 +24,14 @@ class BookmarkContainer extends React.Component {
 
   render() {
     const {
-      controller, url, showModal, modalClosed, canCreate, bookmarks, errors,
+      controller,
+      url,
+      showModal,
+      modalClosed,
+      canCreate,
+      bookmarks,
+      errors,
+      status,
     } = this.props;
 
     return showModal ? (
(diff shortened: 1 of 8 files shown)


sharvit (approved):
Looks great!

webpack/assets/javascripts/react_app/components/bookmarks/index.js
```diff
@@ -45,12 +41,17 @@ class BookmarkContainer extends React.Component {
...
+            </MenuItem>
           )}
-          {getLoadingStatus(errors, bookmarks) === STATUS.RESOLVED &&
-            bookmarks.map(({ name, query }) => <Bookmark key={name} text={name} query={query} />)}
-          {getLoadingStatus(errors, bookmarks) === STATUS.ERROR &&
+          {status === STATUS.RESOLVED &&
+            ((bookmarks.length > 0 &&
+              bookmarks.map(({ name, query }) => (
+                <Bookmark key={name} text={name} query={query} />
+              ))) || <MenuItem> __('None found')</MenuItem>)}
```
timogoebel: This needs to be interpolated, i.e. `<MenuItem>{__('None found')}</MenuItem>`.

image

In addition, I think this text should not be clickable. Maybe it's worth marking it as [disabled](https://getbootstrap.com/docs/3.3/components/#dropdowns-disabled).

timogoebel (changes requested):
@ohadlevy: Thanks. I found one small issue, otherwise: 👍

ohadlevy:
Good find! I've updated both (both disabled and extracted strings).

ohadlevy:
@timogoebel green again

webpack/assets/javascripts/react_app/components/bookmarks/index.js
```diff
@@ -45,12 +41,17 @@ class BookmarkContainer extends React.Component {
         <Dropdown.Toggle />
         <Dropdown.Menu>
           <MenuItem header>{__('Saved Bookmarks')}</MenuItem>
-          {getLoadingStatus(errors, bookmarks) === STATUS.PENDING && (
-            <Loader spinnerSize="xs" status={STATUS.PENDING} />
+          {status === STATUS.PENDING && (
+            <MenuItem className='loader-root'>
```
timogoebel: I think we should add the spinner as a direct child element to the `<li>` in the DOM and drop the `<a>` around it.

webpack/assets/javascripts/react_app/redux/reducers/bookmarks/index.js
```diff
@@ -43,7 +46,9 @@ export default (state = initialState, action) => {
     case BOOKMARKS_MODAL_CLOSED:
       return state.set('showModal', false);
     case BOOKMARKS_FAILURE:
-      return state.setIn([payload.controller, 'errors'], payload.error);
+      return state
+        .setIn([payload.controller, 'errors'], payload.error)
```
timogoebel: This does not work as well, as `payload.controller` is not defined.

timogoebel (changes requested):
@ohadlevy: Thanks. Looks good. I just have one remark regarding the spinner. I also found, that the error action is not working correctly. Could you take another look, please? Thanks.

ohadlevy:
@timogoebel fixed both issues - thanks!

timogoebel (approved):
👍 if tests are 🍏 .

ohadlevy:
It's green :)

timogoebel:
Merged, thanks @ohadlevy.

PR #5108, opened 2017-12-22
fixes #22062 - support vmware vmrc console

timogoebel:
As vSphere >= 6.5 dropped support for the VNC consoles, this is a second attempt (#4755) to provide an alternative.
This uses the VMRC type console, which is basically a standalone program (or something like VMWare Fusion) that provides the console, i.e. the console is not displayed in the browser.
This is actually quite nice (feature wise) because you can easily attach .iso images etc.

image

image
"Open Link in VMWare Fusion?"

image

### DIFF
--- a/app/models/compute_resources/foreman/model/vmware.rb
+++ b/app/models/compute_resources/foreman/model/vmware.rb
@@ -7,6 +7,10 @@ module Foreman::Model
     include ComputeResourceCaching
 
     validates :user, :password, :server, :datacenter, :presence => true
+    validates :display_type, :inclusion => {
+      :in => Proc.new { |cr| cr.class.supported_display_types.keys },
+      :message => N_('not supported by this compute resource')
+    }
 
     before_create :update_public_key
 
@@ -19,6 +23,13 @@ module Foreman::Model
 
     def self.model_name
       ComputeResource.model_name
+    end
+
+    def self.supported_display_types
+      {
+        'vnc' => _('VNC'),
+        'vmrc' => _('VMRC')
+      }
     end
 
     def user_data_supported?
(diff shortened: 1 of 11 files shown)


app/views/hosts/console/vmrc.html.erb
```diff
@@ -0,0 +1,15 @@
+<% title "#{@console[:name]}" %>
+<div class="blank-slate-pf">
+  <div class="blank-slate-pf-icon">
+    <%= icon_text('screen', '', :kind => 'pficon') %>
+  </div>
+  <h1><%= _('VMRC Console') %></h1>
+  <p><%= _('To launch the VMRC console, you need the VMRC software installed.') %></p>
```
ohadlevy, line 7: I assume that there is no way for us to detect it? (if its installed or not?)

ohadlevy (approved):
Tested and looks good, overall working nicely, I only had one question.

app/models/compute_resources/foreman/model/vmware.rb
```diff
@@ -537,6 +564,18 @@ def associated_host(vm)
       associate_by("mac", vm.interfaces.map(&:mac))
     end
 
+    def display_type
+      attrs[:display] || 'vmrc'
```
ohadlevy, line 568: does it make sense to always default to vmrc? (all CR will default to that on first time?) do we have vmware version number so we can make an educated decision?

app/views/hosts/console/vmrc.html.erb
timogoebel: Yes, as far as I know, we can't detect if it's installed or not.

app/models/compute_resources/foreman/model/vmware.rb
timogoebel: Maybe we should add a migration to set all existing compute resources to `vnc` as that was the default behavior. As `vmrc` works with all compute resources, I think it's the preferred choice. We can get the vmware version number, but that requires an established connection and I think it's not worth the effort.

ohadlevy:
Thanks!

PR #5110, opened 2017-12-23
Fixes #21999 #22005 - Migrate toasts to pf-react

sharvit:
Migrate Alert and Toast components to patternfly-react.


webpack/assets/javascripts/foreman_toast_notifications.js
```diff
@@ -4,40 +4,50 @@ import store from '../javascripts/react_app/redux';
...
   // remove both jquery cache and dom entry to avoid ajax ContentLoad events
   // reloading our notifications
   notifications.attr('data-flash', '').removeData();
-}
+};
 
 // load notifications from Rails on ContentLoad
 // to accommodate rails flash syntax
-$(document).on('ContentLoad', () => {
+document.addEventListener('ContentLoad', () => {
```
ohadlevy: is this required? AFAIR it doesnt work on all IE versions?

sharvit: Not required, just something i thought i can change on the go.
It should work for IE >= 9
https://www.w3schools.com/jsref/met_document_addeventlistener.asp

ohadlevy:
@amirfefer can you review please? thanks

sharvit:
Rebased!

webpack/assets/javascripts/foreman_toast_notifications.js
ohadlevy: should we be consistent and apply it to the entire code base?

ohadlevy (commented):
This PR looks good, the main question I don't understand is why the alert-body is in foreman vs patternfly-react?

webpack/assets/javascripts/foreman_toast_notifications.js
```diff
@@ -4,40 +4,50 @@ import store from '../javascripts/react_app/redux';
 
 import * as ToastActions from './react_app/redux/actions/toasts';
 
+const notificationTypeToAlertType = (type) => {
+  switch (type) {
```
amirfefer: could we change the notice type to success already on the server?

object literals would be better instead of switch case ? (readability & efficient),

webpack/assets/javascripts/foreman_toast_notifications.js
amirfefer: could be a problem applying this to the entire code base, because we're using  `jquery.turbolinks` which fix  binded events problem caused by turbolinks.

webpack/assets/javascripts/react_app/components/common/Alert/AlertBody.test.js
```diff
@@ -0,0 +1,33 @@
+import toJson from 'enzyme-to-json';
+import { shallow } from 'enzyme';
+import React from 'react';
+
+import AlertBody from './AlertBody';
+
+describe('AlertBody', () => {
+  const testAleryBodyRenderer = (component) => {
```
amirfefer: typo, `..AlertBody..`

webpack/assets/javascripts/react_app/components/common/Alert/AlertBody.js
```diff
@@ -0,0 +1,26 @@
...
+  link, title, message, children,
+}) => (
+  <span>
+    {link && <AlertLink {...link} />}
+
+    {title && <strong>{title}</strong>}
+
+    <span dangerouslySetInnerHTML={{ __html: message }} />
+
+    {children}
```
amirfefer, line 15: do we have to use `dangerouslySetInnerHTML` ? 

To support in overflowed text you could wrap `children / message` with `react-ellipsis-with-tooltip' :)

amirfefer (changes requested):

amirfefer:
@sharvit  Overall looks great!, I've left some inline comments.
I tend to agree that `AlertBody` could be used in `pf-react` though.

waldenraines (approved):
Tested this out in katello and it worked nicely.  Code looks fine to me except for those things already commented on.

waldenraines:
[test katello]

app/controllers/application_controller.rb
```diff
@@ -22,7 +22,6 @@ class ApplicationController < ActionController::Base
   prepend_before_action :allow_webpack, if: -> { Rails.configuration.webpack.dev_server.enabled }
   around_action :set_timezone
   layout :display_layout?
-  add_flash_types :inline
```
sharvit, line 25: Moved all the flash functionality from the `application_controller` to a shared controller `flash.rb`

sharvit:
@amirfefer - Thanks to your feedback i made some more changes

I have updated this pr to contain
* refactor the rails flash-notifications
* add the option to apply `link` from the server
* replace all the `notice` notification to `success` or `info` based on the message context

toast-notification-example

Notice, you cannot apply link when using inside redirect:
```ruby
redirect_to some_path, :warning => _("Some warning")
```

To apply the link you should migrate to use as method as well:
```ruby
warning_link = { :text => _("The Foreman"), :href => "https://theforeman.org/" }
warning _("Some warning"), { :link => warning_link }
redirect_to some_path
```

webpack/assets/javascripts/react_app/components/common/Alert/AlertBody.js
sharvit: I afraid it will cause some problems because rails is writing the notifications into the dom, then our javascript stack will read it from the dom, transform the data into toasts and then remove the data from the dom.

About the `react-ellipsis-with-tooltip`, i don't want to force this behavior, currently, when there is no space for the text you will get 2 rows of text which looks good.
We can use the `react-ellipsis-with-tooltip` only when we want to cancel this behavior in specific places.

app/controllers/concerns/foreman/controller/flash.rb
```diff
@@ -0,0 +1,71 @@
...
+    else
+      flash[:inline] = flash_data
+    end
+  end
+
+  def flash_message(type, message, options = default_flash_options)
+    # backward compatibility, so old code can still run
+    # `error('some message', true`) instead of migrate to
+    # `error('some message', { :now => true })``
+    options = { :now => true, :link => nil } if options.is_a?(TrueClass)
```
sharvit: I mostly made this backward compatibility because it was impossible to run over all the flash messages and search for all those use cases

sharvit:
Bring back the notice with a deprecation warning thanks to @ohadlevy feedback

sharvit:
Travis is currently failing because `redux-form` released a new version `v7.2.1`
Will create a separate pr to fix this issue...

sharvit:
Rebased, all tests are green.
@amirfefer, @waldenraines, @ohadlevy can review again please?

app/controllers/concerns/application_shared.rb
```diff
@@ -4,6 +4,7 @@ module ApplicationShared
   include Foreman::Controller::MigrationChecker
   include Foreman::Controller::Authentication
   include Foreman::Controller::Session
+  include Foreman::Controller::Flash
```
ohadlevy: shared is also used by api base controller, maybe move it to the application controller instead?

app/helpers/application_helper.rb
```diff
@@ -545,14 +545,19 @@ def user_set?(field)
     params[:host] && params[:host][:hostgroup_id] && params[:host][field]
   end
 
-  def notifications
-    content_tag :div, :id => 'notifications', :'data-flash' => flash_notifiations.to_json.html_safe do
-      mount_react_component('ToastNotifications', '#notifications')
+  def render_toast_notifications
```
waldenraines: Why the change to `render_toast_notifications`?  This will break all plugins that use `<%= notifications %>`.

waldenraines (changes requested):

app/helpers/application_helper.rb
sharvit: I didn't really have a good reason to do so, only because I thought it is a better name.
Will migrate back.

sharvit:
@ohadlevy @waldenraines thanks for your comments.

1. rebased
2. migrated back from `render_toast_notifications` to `notifications` so it won't break plugins
3.  Moved the `include Foreman::Controller::Flash` to `application_controller.rb` so it won't be available for API controllers

waldenraines (approved):
This works well for me when I tested it and the code looks fine.

ohadlevy:
@sharvit codeclimates complain about a duplicate test, can you have another look? thanks

sharvit:
@ohadlevy I don't agree with codeclimates about it.
It complains because the tests look like a list of boilerplates, isn't it exactly what we ideally want them to be?

The test file:
https://github.com/theforeman/foreman/pull/5110/files#diff-521decd9ef7064340cbf66315b60daeb

ohadlevy (approved):

PR #5111, opened 2017-12-24
Add version column to auditable_index

ghost:
Foreman regularly does queries like this:
```
SELECT MAX(`audits`.`version`) AS max_id FROM `audits`  WHERE `audits`.`auditable_id` = 7555 AND `audits`.`auditable_type` = 'LookupKey';
```

On large deployments with huge audits tables, this looks like this and results in everything taking a very long time:
```
+--------+
| max_id |
+--------+
| 117292 |
+--------+
1 row in set (13.44 sec)
```

A simple solution to this, is to include the version in the compound index, which stops mysql from having to go to the table at all, and turns the same query into this:

```
+--------+
| max_id |
+--------+
| 117292 |
+--------+
1 row in set (0.00 sec)
```

### DIFF
--- a/db/migrate/20090714132449_add_audits_table.rb
+++ b/db/migrate/20090714132449_add_audits_table.rb
@@ -16,7 +16,7 @@ class AddAuditsTable < ActiveRecord::Migration[4.2]
       t.column :created_at, :datetime
     end
 
-    add_index :audits, [:auditable_id, :auditable_type], :name => 'auditable_index'
+    add_index :audits, [:auditable_id, :auditable_type, :version], :name => 'auditable_index'
     add_index :audits, [:user_id, :user_type], :name => 'user_index'
     add_index :audits, :created_at
     add_index :audits, [:auditable_parent_id, :auditable_parent_type], :name => 'auditable_parent_index'
--- a/db/migrate/20090714132449_add_audits_table.rb
+++ b/db/migrate/20090714132449_add_audits_table.rb
@@ -16,7 +16,7 @@ class AddAuditsTable < ActiveRecord::Migration[4.2]
       t.column :created_at, :datetime
     end
 
-    add_index :audits, [:auditable_id, :auditable_type, :version], :name => 'auditable_index'
+    add_index :audits, [:auditable_id, :auditable_type], :name => 'auditable_index'
     add_index :audits, [:user_id, :user_type], :name => 'user_index'
     add_index :audits, :created_at
     add_index :audits, [:auditable_parent_id, :auditable_parent_type], :name => 'auditable_parent_index'
--- a/db/migrate/20171225122601_add_version_to_auditable_index.rb
+++ b/db/migrate/20171225122601_add_version_to_auditable_index.rb
@@ -1,0 +1,11 @@
+class AddVersionToAuditableIndex < ActiveRecord::Migration[4.2]
+  def up
+    remove_index :audits, :name => 'auditable_index'
+    add_index :audits, [:auditable_id, :auditable_type, :version], :name => 'auditable_index'
+  end
+
+  def down
+    remove_index :audits, :name => 'auditable_index'
+    add_index :audits, [:auditable_id, :auditable_type], :name => 'auditable_index'
+  end
+end


boaz0 (commented):
LGTM

Is there a way to test that in the CI, though?

tbrisker:
Hi @Planktonette and thank you for the contribution! a couple of comments: 
Please make sure the issue has been reported on our [Redmine tracker](http://projects.theforeman.org/projects/foreman/issues/new) as well, which is used to allow control of which fixes go into what release, and change the commit message to refer to that issue - i.e. "Fixes #11111 - Add version column to audit index".
The second comment is that migrations are only executed once, so any system that has already ran this migration will not benefit from this change. To make changes such as this, you need to create a new migration that adds this index - see [here](http://edgeguides.rubyonrails.org/active_record_migrations.html) for more information on rails migrations. 
Let us know if you have any questions or need assistance!

ghost:
Hopefully I've got this right! It's been quite some time since I committed to a large project :)
New commit pushed onto my branch that reverts the change to the original migration, and adds a new migration to add the version column to that index.
We have this currently running in our environment, and so far haven't seen any negatives (only positives, including our puppet runtime dropping from 60 seconds to 10 seconds!), but not knowing the codebase, my concern would be if there are other queries hitting the audits table and using the original index which might be affected by the addition of the version column.
Here's our DB load from before the change, during a foreman reports cleanup run, and then after i added the additional column to the index.
screen shot 2017-12-25 at 12 35 45

tbrisker:
Wow, that looks like quite an improvement! Let me kick off CI:
ok to test

tbrisker:
With regards to other usage of the existing index- if you want to be certain, you can just leave the existing index in place and add a new one with versions, but I think that the DB should be able to use only the partial index when it doesn't need the version. You can verify this by benchmarking the following query on your DB before and after the index change (or checking with explain if the index is used or not):
```sql
SELECT * FROM `audits`  WHERE `audits`.`auditable_id` = 7555 AND `audits`.`auditable_type` = 'LookupKey';
```

ghost:
Have just tested it, the difference in the query analyzer is marginal, but in terms of query performance, there is basically no difference:

```
> explain extended SELECT count(*) FROM `audits` USE INDEX (auditable_index)  WHERE `audits`.`auditable_id` = 7555 AND `audits`.`auditable_type` = 'LookupKey';
+----+-------------+--------+------+-----------------+-----------------+---------+-------------+--------+----------+--------------------------+
| id | select_type | table  | type | possible_keys   | key             | key_len | ref         | rows   | filtered | Extra                    |
+----+-------------+--------+------+-----------------+-----------------+---------+-------------+--------+----------+--------------------------+
|  1 | SIMPLE      | audits | ref  | auditable_index | auditable_index | 773     | const,const | 300260 |   100.00 | Using where; Using index |
+----+-------------+--------+------+-----------------+-----------------+---------+-------------+--------+----------+--------------------------+
1 row in set, 1 warning (0.00 sec)

Mon Dec 25 12:52:55 2017
> explain extended SELECT count(*) FROM `audits` USE INDEX (auditable_index_version)  WHERE `audits`.`auditable_id` = 7555 AND `audits`.`auditable_type` = 'LookupKey';
+----+-------------+--------+------+-------------------------+-------------------------+---------+-------------+--------+----------+--------------------------+
| id | select_type | table  | type | possible_keys           | key                     | key_len | ref         | rows   | filtered | Extra                    |
+----+-------------+--------+------+-------------------------+-------------------------+---------+-------------+--------+----------+--------------------------+
|  1 | SIMPLE      | audits | ref  | auditable_index_version | auditable_index_version | 773     | const,const | 330898 |   100.00 | Using where; Using index |
+----+-------------+--------+------+-------------------------+-------------------------+---------+-------------+--------+----------+--------------------------+
1 row in set, 1 warning (0.00 sec)
```

```
> SELECT SQL_NO_CACHE  count(*) FROM `audits` USE INDEX (auditable_index)  WHERE `audits`.`auditable_id` = 7555 AND `audits`.`auditable_type` = 'LookupKey';
+----------+
| count(*) |
+----------+
|   169994 |
+----------+
1 row in set (0.08 sec)

Mon Dec 25 12:51:25 2017
> SELECT SQL_NO_CACHE count(*) FROM `audits` USE INDEX (auditable_index_version)  WHERE `audits`.`auditable_id` = 7555 AND `audits`.`auditable_type` = 'LookupKey';
+----------+
| count(*) |
+----------+
|   169994 |
+----------+
1 row in set (0.07 sec)
```

I think based on this I feel comfortable leaving it as the triple column index, but your call :)

ohadlevy:
@Planktonette great job! thanks for the PR :-)

@tbrisker can you verify it has a similar impact on postgresql ?

tbrisker:
Looks like postgresql also know how to use the partial index:
```sql
explain analyze select * from audits where auditable_id = 5 and auditable_type = 'Subnet';
                                                        QUERY PLAN                                                        
--------------------------------------------------------------------------------------------------------------------------
 Index Scan using auditable_index on audits  (cost=0.28..8.30 rows=1 width=849) (actual time=0.029..0.032 rows=3 loops=1)
   Index Cond: ((auditable_id = 5) AND ((auditable_type)::text = 'Subnet'::text))
 Planning time: 0.208 ms
 Execution time: 0.079 ms
(4 rows)
```
I don't have a postgres DB with enough audits to compare the performance difference atm, @ohadlevy do you have one?

ohadlevy:
@tbrisker I don't but based on your feedback I don't have any objections to merge this. thanks!

tbrisker (approved):
Tested this with about 80k mock audits and it didn't seem to have a significant effect on postgresql, real data might show better results, but I'd say this is safe to merge anyways. Thanks a lot @Planktonette !

tbrisker:
Just another update - this also improves greatly the performance on postgres, dropped from 28ms to 0.05ms on a db with 300k audits. I think the benefit is more significant for resources that have a very large number of versions.

tbrisker:
This looks like https://github.com/collectiveidea/audited/issues/189. 
@Planktonette - would you be interested in opening a pr to fix it in the audited gem? it should be somwhat similar to https://github.com/collectiveidea/audited/pull/385. If not, I will most likely do that myself eventually.

PR #5113, opened 2017-12-26
Fixes #22069 - Add proxy version mismatch warning

ShimShtein:


orrabin:
I'm not sure if this is the correct behavior or not:

image


1.17 and 1.17-develop are different but is that because of the way we save the version or is it actually considered a different version?

Maybe @witlessbird can answer?
 EDIT: with 1.17-develop in core I get the error so there is still a problem with the comparison but tags in the version are still something to think about.

dmitri-d:
Smart-proxy treats 1.17, 1.17-develop, 1.17-RCX, etc. as the same version.

orrabin:
@witlessbird thanks!
@ShimShtein I'm moving this to WoC now that we have an answer.

app/assets/javascripts/proxy_status.js
```diff
@@ -14,7 +14,11 @@ $(document).on('ContentLoad', function() {
 $(window).on('hashchange', setTab); //so buttons that link to an anchor can open that tab
 
 function setItemStatus(item, response) {
-  if(response.success) {
+  if(response.success && response.message.warning) {
```
amirfefer: if `response.message` could be `undefined` then consider use lodash`._get` , otherwise it may cause errors.

ohadlevy: @amirfefer we are actually trying to reduce the usage of lodash, /cc @sharvit

sharvit: We do try to reduce the lodash usage, `_.get` can still be used in the assets but prefer not to whenever possible.
See: https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/bundle_lodash.js

sharvit: can do:
```js
if(response && response.success && response.message && response.message.warning) {}
```

amirfefer: if `._get` exists in  global, why not using it in legacy js? even if it just for increasing readability a bit?

sharvit: I personally don't like the `_.get` and i think it takes away readability. (unless you are familiar with it)
But i am fine with that.

ShimShtein:
Fixed mixed tags mismatch.

app/assets/javascripts/proxy_status.js
ShimShtein: If the response is successful, there always be a message node in the [response](https://github.com/theforeman/foreman/blob/develop/app/controllers/smart_proxies_controller.rb#L140).

sharvit: I don't believe the client should count on that.
I think it is still important to verify the object structure before you are using it.

ShimShtein:
Added extra safety condition to the JS.

ShimShtein:
Rebased.

amirfefer:
LGTM :+1:

iNecas:
@orrabin any more comments?

boaz0 (approved):
LGTM 👍

app/controllers/smart_proxies_controller.rb
```diff
@@ -159,4 +161,19 @@ def resource_base
...
   end
+
+  def versions_mismatched?(proxy_versions)
+    foreman_version = Foreman::Version.new.notag
+    proxy_version = Foreman::Version.new(proxy_versions['version']).notag
+
+    foreman_version != proxy_version
+  end
+
+  def version_mismatch_warning(versions_hash)
```
orrabin: Maybe call both this and the parameter in `versions_mismatched` proxy_versions_hash?
I know this is a huge nitpick but I did not understand why you did `['version']` twice, hash in hash is confusing :)

orrabin:
Tested again:

image

I'm not sure we have to show the `-develop` extension but it's working correctly.
I later updated my smart proxy and the status was green with no warning.

orrabin:
I added 1 nitpick, besides that it's ready to merge.
@ShimShtein rebase might get rid of the katello failures.
This is ready to merge when tests are green.

ShimShtein:
@orrabin Done!

orrabin (approved):

orrabin:
Thanks @ShimShtein !

jhutar:
Hello @ShimShtein. I'm trying to test this. I have installed nightly Katello (`foreman-1.18.0-0.develop.201802191629git3de2d5a.el7.noarch`) and Smart Proxy (`foreman-proxy-1.18.0-0.develop.201802120903git0e9bc1c.el7.noarch`) and all looked fine. Then I have downgraded Smart Proxy with:

```
# rpm -Uvh --oldpackage http://yum.theforeman.org/releases/1.17/el7/x86_64/foreman-proxy-1.17.0-0.1.RC1.el7.noarch.rpm
# katello-service restart
```

and then on the Foreman webUI I have clicked Infrastructure --> Smart Proxies --> smart-proxy-nightly.example.com --> Refresh features and reloaded and really, there is "Version 1.17.0-RC1" now. So went to list of proxies, and there is still green "check" in the "Status" column.

Looking into ping request the page is doing, GET to:

```
https://foreman-nightly.example.com/smart_proxies/2-smart-proxy-nightly-example-com/ping
```

returned:

```
{"success":true,"message":{"version":"1.17.0-RC1","modules":{"pulpnode":"1.3.0","templates":"1.17.0-RC1","tftp":"1.17.0-RC1","puppetca":"1.17.0-RC1","puppet":"1.17.0-RC1","logs":"1.17.0-RC1"}}}
```

compared to same request for integrated proxy:

```
https://foreman-nightly.example.com/smart_proxies/1-foreman-nightly-example-com/ping
```

which gave:

```
{"success":true,"message":{"version":"1.18.0-develop","modules":{"pulp":"1.3.0","templates":"1.18.0","tftp":"1.18.0","puppetca":"1.18.0","puppet":"1.18.0","logs":"1.18.0"}}}
```

Could you please advise on how to test this fix?

ShimShtein:
@jhutar, I have to admit, I don't know, I suspect the sources do not contain my patch yet.
Can you check that when the system is up and you are on the "About" or "Smart proxies" page, when you search in browser's sources for `response.message.warning` (using CTRL-SHIFT-F in the browser developer tools pane), you get at least two results:
image
on the server side, please look for `def versions_mismatched?(proxy_versions_hash)` in `app/controllers/smart_proxies_controller.rb`. It should be somewhere near line 165.

jhutar:
OK, now after few days I have started Foreman and Smart proxy VMs again and, surprise surprise, orange exclamation mark is there. It now says "Core and proxy versions do not match. foreman: 1.18.0, foreman-proxy: 1.17.0-RC1". I'm pretty sure I have done Ctrl+Shift+R before, but I might have not restarted proxy.

Anyway, upgrading mine Foreman and Smart proxy nightly to latest again, I'm getting green "check mark" again.

Thank you!

=> VERIFIED

PR #5122, opened 2017-12-27
fixes #22085 - facet test models have db table

timogoebel:


test/unit/facet_configuration_test.rb
```diff
@@ -7,15 +7,14 @@ module TestExtension
 end
 
 class FacetConfigurationTest < ActiveSupport::TestCase
-  class TestClass < HostFacets::Base
+  class TestFacet < HostFacets::Base
```
ShimShtein, line 10: There is easier way to solve this:
``` ruby
  class TestClass < HostFacets::Base
    self.table_name = 'test_facets'
  end
```

This sets proper table name for the model hence eliminating non-existing table error.

ShimShtein (changes requested):
@timogoebel, your solution is valid, but we could do it with less code lines. I don't have a strong opinion against it though.

test/unit/facet_configuration_test.rb
timogoebel: > your solution is valid, but we could do it with less code lines.

Indeed, but I don't think it's the straight forward solution. The diff is larger, but we don't add more LOC with my proposal. And it's easier to read the code with my change imho.

ShimShtein (approved):

ShimShtein:
Merged.

Thanks @timogoebel! 
As I said, the solutions are equivalent for me.

PR #5123, opened 2017-12-27
Some more Rubocop changes

mmoll:


mmoll:
One or two of these could spark a discussion...

app/registries/foreman/plugin.rb
```diff
@@ -316,7 +316,7 @@ def extend_template_helpers(*mods)
     # Add Compute resource
     def compute_resource(provider)
       return if @compute_resources.include?(provider)
-      if !provider.is_a?(Class) || !(provider < ComputeResource)
+      if !provider.is_a?(Class) || provider >= ComputeResource
```
tbrisker: I think here it might make it clearer if we use DeMorgan:
```ruby
unless provider.is_a?(Class) && provider < ComputeResource
```

test/unit/ui_notifications/clean_expired_test.rb
```diff
@@ -95,7 +95,7 @@ def create_notification(blueprint)
 
   def create_blueprint(expired_time, validate = true)
     blueprint = FactoryBot.build(
-        :notification_blueprint,
+      :notification_blueprint,
         :expires_in => expired_time
```
tbrisker: I think this line also needs to be unindented

db/migrate/20130520152000_remove_duplicate_fact_names.rb
```diff
@@ -5,10 +5,10 @@ def up
       duplicates = FactName.where("name=? and id<>?", fact_name, fact_name_id).select(:id)
       ActiveRecord::Base.transaction do
         FactValue.update_all(
-            ["fact_name_id=?", fact_name_id],
+          ["fact_name_id=?", fact_name_id],
             ["fact_name_id in (?)", duplicates])
```
tbrisker: I think this line needs to be unindented as well

db/migrate/20130520152000_remove_duplicate_fact_names.rb
```diff
@@ -5,10 +5,10 @@ def up
...
         UserFact.update_all(
-            ["fact_name_id=?", fact_name_id],
+          ["fact_name_id=?", fact_name_id],
             ["fact_name_id in (?)", duplicates])
```
tbrisker: ditto

app/controllers/api/v2/hosts_controller.rb
```diff
@@ -202,7 +202,7 @@ def status
 * configuration
 * build
 eos
-)
+                                                                                                                                        )
```
tbrisker: This is a bit odd location, perhaps break the lines in the param call so it's a bit clearer?

.rubocop_todo.yml
```diff
@@ -36,11 +36,6 @@ Layout/AlignParameters:
 Layout/CaseIndentation:
   Enabled: false
 
-# Offense count: 28
-# Cop supports --auto-correct.
-Layout/ClosingParenthesisIndentation:
-  Enabled: false
```
tbrisker: tbh i'm not a big fan of this cop's changes, some of them look good but others push parenthesis to odd locations.

.rubocop_todo.yml
```diff
@@ -451,13 +451,6 @@ Security/YAMLLoad:
 Style/Alias:
   Enabled: false
 
-# Offense count: 10
-# Cop supports --auto-correct.
-# Configuration parameters: EnforcedStyle.
-# SupportedStyles: always, conditionals
-Style/AndOr:
-  Enabled: false
```
tbrisker, line 459: I think we should enforce this one only on conditionals. `do_something and exit(1)` and such are common idioms

lib/tasks/locale.rake
```diff
@@ -17,7 +17,7 @@ namespace :locale do
     File.rename "#{filename}.rb", "#{filename}.tmp"
     File.open("#{filename}.rb", "w") do |output|
       IO.foreach("#{filename}.tmp") do |line|
-        if not line =~ /^\s*#/ and not line =~ ignored
+        if (not line =~ /^\s*#/) && (not line =~ ignored)
```
tbrisker: I think a DeMorgan here will improve readability:
```ruby
unless line =~ /^\s*#/ || line =~ ignored

**Review by tbrisker — COMMENTED** [review]
See my comments inline

**Comment by tbrisker** [comment]
Also looks like test failures are related

**Comment by mmoll** [comment]
- removed `Layout/ClosingParenthesisIndentation`
- set `Style/AndOr` to `conditionals`
- updated code according to comments
- updated tests

File: .rubocop_todo.yml (hunk 7 above)
Comment (ares) [reply .rubocop_todo.yml:459]: I agree with @tbrisker, `&&` and `and` are different things, one is for condition second is for flow control, the priority of evaluation  is different

File: app/controllers/api/v2/compute_attributes_controller.rb [hunk 9]
```diff
@@ -22,8 +22,8 @@ class ComputeAttributesController < V2::BaseController
 
       def create
         @compute_attribute = ComputeAttribute.new(compute_attribute_params.merge(
-          :compute_profile_id => params[:compute_profile_id],
-          :compute_resource_id => params[:compute_resource_id]))
+                                                    :compute_profile_id => params[:compute_profile_id],
```
ares: I don't like this too much but I don't insist on disabling, I think whitespace formatting causes most cop failures and leads to frustrations like "why my PR is red now?"

app/registries/foreman/plugin.rb
```diff
@@ -316,7 +316,7 @@ def extend_template_helpers(*mods)
     # Add Compute resource
     def compute_resource(provider)
       return if @compute_resources.include?(provider)
-      if !provider.is_a?(Class) || !(provider < ComputeResource)
+      unless provider.is_a?(Class) && provider < ComputeResource
```
ares: I'm against this, unless with && or || makes it very hard for me to read the condition. I prefer to use if in case of composed conditions

ares:
left few comments, :+1: for the rest

app/controllers/api/v2/compute_attributes_controller.rb
mmoll: I do understand that, on the other hand it's an auto-correctable error and editors can be configured for it, too... This change did only touch eleven files of the whole codebase, so I guess most of the code is already like wanted by the cop.

However, I'd like to vote here with 👍 and 👎 for this cop.

mmoll:
- disabled `Style/InverseMethods` cop

app/controllers/api/v2/compute_attributes_controller.rb
ares: If I'm the only one, feel free to skip voting, 
not a big deal :-)

lzap: I prefer the original formatting.

config/application.rb
```diff
@@ -191,16 +191,16 @@ class Application < Rails::Application
...
+                                   :app => {:enabled => true},
+                                   :audit => {:enabled => true},
+                                   :ldap => {:enabled => false},
+                                   :permissions => {:enabled => false},
+                                   :proxy => {:enabled => false},
+                                   :sql => {:enabled => false},
+                                   :templates => {:enabled => true},
+                                   :notifications => {:enabled => true},
+                                   :background => {:enabled => true},
+                                   :dynflow => {:enabled => true}
```
lzap: Typical example of bad practice - change of the first line (193) leads to full reformatting, making more conflicts and overall this does not improve readability at all - IMHO makes that harder because this will likely hit 80th column more often.

tbrisker: I'd argue the actual issue here is that the defaults hash is all passed directly to the reverse merge instead of being defined as a well-named variable first.

app/controllers/api/v2/compute_attributes_controller.rb
mmoll: How about removing this one commit from this PR and I'll open another one (or a gist) with the variants possible?

I think that the "consistent" option of this cop might fancy, what you prefer while still enforcing the same alignment of all concerning lines.

mmoll:
- set `Layout/FirstParameterIndentation` to `conistent`

app/controllers/api/v2/compute_attributes_controller.rb
mmoll: @lzap please have a look again. 👀

ares: > How about removing this one commit from this PR and I'll open another one (or a gist) with the variants possible? 

Sounds good to me

mmoll:
rebased, test failure unrelated

dLobatog (approved):
Test failure is the usual ` check DNS and/or server status - Timed out` - thanks @mmoll

tbrisker (approved):

tbrisker:
thanks @mmoll !

PR #5125, opened 2017-12-28
Fixes #12829 - add identifier for smart variables/parameters

shiramax:


app/views/puppetclasses/_class_parameters.html.erb
```diff
@@ -15,6 +15,9 @@
   <tr id="puppetclass_<%= puppetclass.id %>_params[<%= lookup_key.id %>]"
     class="fields <%= 'overridden' if overridden %>">
     <%= content_tag(:td, puppetclass.name, :rowspan => lookup_keys.size, :class => "ellipsis") if index == 0 %>
+    <td class="ellipsis param_name">
+      <%= lookup_key.type == "PuppetclassLookupKey" ? "Smart Parameters":"Smart Variables"%>
```
tbrisker: Please extract the strings for translation (https://theforeman.org/handbook.html#Stringsandtranslations).
You can also check the key type using `lookup_key.is_a? PuppetclassLookupKey` (reducing one string allocation per key)

app/views/puppetclasses/_class_parameters.html.erb
```diff
@@ -15,6 +15,9 @@
   <tr id="puppetclass_<%= puppetclass.id %>_params[<%= lookup_key.id %>]"
     class="fields <%= 'overridden' if overridden %>">
     <%= content_tag(:td, puppetclass.name, :rowspan => lookup_keys.size, :class => "ellipsis") if index == 0 %>
+    <td class="ellipsis param_name">
```
tbrisker: I think `param_name` class isn't needed here and might actually lead to issues (iirc, it is used by the js to lookup the param name)

app/views/puppetclasses/_classes_parameters.html.erb
```diff
@@ -2,6 +2,7 @@
   <thead class="white-header">
     <tr>
       <th class='col-md-3'><%= _('Puppet Class') %></th>
+      <th class='col-md-2'><%= _('Type') %></th>
```
tbrisker: This width causes the total column to go over 12 which is the default grid size used in bootstrap. Perhaps the width of other columns should be adjusted to fit this in.

tbrisker (changes requested):
See comments inline. thanks!

app/views/puppetclasses/_class_parameters.html.erb
```diff
@@ -16,6 +16,9 @@
     class="fields <%= 'overridden' if overridden %>">
     <%= content_tag(:td, puppetclass.name, :rowspan => lookup_keys.size, :class => "ellipsis") if index == 0 %>
     <td class="ellipsis param_name">
```
tbrisker: `param_name` class should be on the other column that contains the param name (lookup_key.key). Both columns should have the `ellipsis` class so they get a tooltip if they overflow (e.g. on a small display). I think the order should be the other way around, first name then type.

app/views/puppetclasses/_class_parameters.html.erb
```diff
@@ -16,6 +16,9 @@
...
+      <%= (lookup_key.is_a? PuppetclassLookupKey) ? _("Smart Parameters"):_("Smart Variables")%>
```
tbrisker: I think these should be singular. also, please add a space around the `:`

tbrisker (changes requested):

shiramax:
@tbrisker Done.

app/views/puppetclasses/_class_parameters.html.erb
```diff
@@ -18,6 +18,9 @@
     <td class="ellipsis param_name">
       <%= lookup_key.key %>
     </td>
+    <td class="ellipsis">
+      <%= (lookup_key.is_a? PuppetclassLookupKey) ? _("Smart Parameter"): _("Smart Variables")%>
+    <td>
```
tbrisker: This should be `</td>`

tbrisker (changes requested):

shiramax:
@tbrisker  Done.

tbrisker (approved):

tbrisker:
Thanks @shiramax !

PR #5129, opened 2017-12-30
Fixes #22089 - remove unused scope

mmoll:


mmoll:
This was used in the original commit 2fa95807fa but is unused now.

tbrisker (approved):
Couldn't find anywhere using this scope either in a sample of several large plugins, I think this is good to go. The best code is no code :)

PR #1916, opened 2014-11-10
Bug #7619 - [zh_CN] Localized string broken for Loading page

yebeloved:
Make the localized string be in one line.

PR #5131, opened 2017-12-31
Fixes #1916 - Subnet vlan id should be integer

tbrisker:


db/migrate/20171231134017_change_vlan_to_int.rb
```diff
@@ -0,0 +1,11 @@
+class ChangeVlanToInt < ActiveRecord::Migration[5.1]
+  def up
+    Subnet.unscoped.where(vlanid: "").update_all(vlanid: nil)
+    opts = connection.adapter_name.downcase == 'postgresql' ? {using: 'vlanid::integer'} : {}
```
ohadlevy, line 4: why is this required? shouldnt the schema define this?

tbrisker: Otherwise postgres throws an error, you need to tell it explicitly how you want to convert the string.

ohadlevy: thanks, any idea how the empty string got there to begin with? any place we should make sure we dont add new records as strings?

tbrisker: When submitting a form it was saved as blank string instead of nil because that is the parameter that was passed to the server. With integer field this is cast to nil.

ohadlevy: was that fixed elsewhere? I didnt see that fix part of this pr?

tbrisker: changing the column type to integer fixes it, it saves an empty parameter as `NULL` instead of empty string.

ohadlevy (approved):

tbrisker:
@ohadlevy I noticed there was some old code that won't be needed any longer afaict for sorting, added as a second commit.

ohadlevy:
@tbrisker thanks, in light of your recent changes, would mind adding a test?

tbrisker:
@ohadlevy sure, but i'm not quite sure what should I test?

boaz0 (approved):
LGTM 👍

ohadlevy:
@tbrisker does it make sense to ensure that sorting based on vlan does not break? e.g. a simple test that get a list of subnets sorted by their vlan?

lzap (approved):
Good, although it is good time to get rid of the VLANID sorting abuse. Honestly I would like to see default scope of vlanid, created_at so we always get stable order for subnets without VLANID.

timogoebel:
Does it make sense to add validations to the rails model as well?
Right now a value of `Bär` is accepted I think.

```
validates :vlanid, numericality: { only_integer: true, greater_than: 0, less_than_or_equal_to: 4096 }
```

timogoebel (changes requested):
@tbrisker: Let's do some fancy TDD together, shall we?:

I just checked and I think, the Subnet model should pass these tests:

```ruby
  should allow_values(10, 100, '200').for(:vlanid)
  should_not allow_values('Bär', '', ).for(:vlanid)
```

Otherwise, looks very good!

tbrisker:
@timogoebel Thanks for the feedback, added tests and validator similar to what you suggested. One small difference - blank value is valid for vlanid, with an integer field in the db it gets saved as `NULL`.

tbrisker:
[test upgrade]

timogoebel (approved):

timogoebel:
@tbrisker: Looks good, thanks. :-)

timogoebel:
Merged, thanks @tbrisker.

lzap:
For the record, this created regression for MySQL users, creating issue and fix: http://projects.theforeman.org/issues/23376 and https://github.com/theforeman/foreman/pull/5484

PR #5133, opened 2018-01-01
Fixes #22104 - adapt to Rails 5.1 change tracking

mmoll:


app/models/nic/interface.rb
```diff
@@ -57,13 +57,13 @@ def alias?
     end
 
     def fqdn_changed?
```
tbrisker: Since these two methods are meant to mimic active record methods, should we also change them to the new naming?

mmoll: I thought its use were more spread over the code, but its indeed only in this file, so renaming was no problem 👍.

app/models/host/base.rb
```diff
@@ -87,8 +87,8 @@ def dup
              :subnet, :subnet_id, :subnet_name,
              :subnet6, :subnet6_id, :subnet6_name,
              :domain, :domain_id, :domain_name,
-             :hostname, :fqdn, :fqdn_changed?,
-             :fqdn_was, :shortname,
+             :hostname, :fqdn, :saved_change_to_fqdn?,
+             :fqdn_before_last_save, :shortname,
```
tbrisker: Do we actually use these anywhere? i think it may be leftovers from something, i couldn't find anywhere where these methods are called on a host

mmoll: removed these.

mmoll:
test failures unrelated, AFAICT.

mmoll:
Setting to WOC, as GH-4987 exposed some more code paths with calls that need to get checked.

mmoll:
No movement in GH-4987, so I'm kindly asking to proceed here...
Rebased, all 💚.

tbrisker (approved):
Looks good. Even if this isn't every case it's better then nothing, let's merge now and fix any other issues we encounter. In any case this won't break before 5.2.

PR #5137, opened 2018-01-01
Fixes #22108 - add pf-react donut chart

amirfefer:
- [x]  Upgrading pf-react 
- [x]  Refactoring donut chart
- [x] Refactoring statistics charts
- [x]  Tests


webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
```diff
@@ -29,30 +30,29 @@ class ChartBox extends React.Component {
...
+        ? null
+        : navigateToSearch.bind(null, chart.search);
 
-    const _chart = (
-      <PieChart
+    const _chart = (config = 'regullar') => <Chart
+        config={config}
         key={`${this.props.chart.id}-chart`}
-        data={this.props.chart.data}
+        data={chart.data ? Immutable.asMutable(chart.data) : {}}
```
ohadlevy: what was the thought process behind this ? how would a children/c3 change the data?

ohadlevy (changes requested):
Thanks for the PR, from a very quick testing:

1. dashboard, smart proxy status for puppet (and maybe other places / plugins that uses c3 charts) is broken.
2. colors are off (I see only two different colors? blue and grey).
3. I see you added legend, can we add the percentage next to the label (sorted would be even nicer).

webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
amirfefer: looks like the issue occurs in `names` object
`c3` is trying to change it through [here](https://github.com/c3js/c3/blob/69f975cfcf54626a36566a8b8236f63aec001951/c3.js#L5983)

According to c3 docs:
> ...if this argument is given, the names of data will be **updated**. If not given, the current names will be returned.

amirfefer:
@Rohoover 
patternfly donut chart has only one default 2-color pattern (i've opened an [issue](https://github.com/patternfly/patternfly/issues/902) for that), 
should we add a 6-color pattern? 
Which colors would fit ?

Rohoover:
@amirfefer 

Looking at the documentation I'm seeing multiple colors:
http://www.patternfly.org/pattern-library/data-visualization/donut-chart/#/overview
https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/donut-charts.html

amirfefer:
@Rohoover :
In `utilization` donut chart type there is a default color pattern (blue and black-300).
where as in  `Relationship of a Set of Values to a Whole` type ( which suppose to support up to 6 categories)  no color pattern is found.
The colors you have seen In the documentation examples, are hard coded, which means each category specifies a color, but there is no a 6 color pattern for this type chart (in contrast to utilization type).

Rohoover:
@amirfefer 

Here are some colors:
pf-red-100 #cc0000
pf-blue  #0088ce
pf-orange-400 #ec7a08
pf-green-400 #3f9c35
pf-cyan-500 #005c66
pf-gold-200 ##f9d67a

I got them from here:
http://www.patternfly.org/styles/color-palette/

I will also address the issue with the PF team to see if they want or are going to establish standard colors for 6 category donuts. Thanks!

Update: The PF team is looking at this issue but I would use the above for now. I specifically was looking at their extended color palette which is about 3/4 of the way down the page.

amirfefer:
@Rohoover  - Thanks !
@ohadlevy, I've rebased and updated this PR.
@sharvit : Would you mind to review this ?

webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
```diff
@@ -63,27 +63,26 @@ class ChartBox extends React.Component {
...
+          <Modal.Body>
+            {_chart('large')}
+          </Modal.Body>
+        </Modal> : null }
+      </Panel>;
   }
 }
 
 ChartBox.propTypes = {
   status: PropTypes.string.isRequired,
```
sharvit, line 85: Some props are missing here (char, type, tip...)

webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
```diff
@@ -63,27 +63,26 @@ class ChartBox extends React.Component {
...
+        </Loader>
+        {this.state.showModal ?
+        <Modal show={this.state.showModal} enforceFocus onHide={this.closeModal}>
+          <Modal.Header closeButton>
+            <Modal.Title>{this.props.title}</Modal.Title>
+          </Modal.Header>
+          <Modal.Body>
+            {_chart('large')}
+          </Modal.Body>
+        </Modal> : null }
```
sharvit: can remove the ` : null` if switch to `this.state.showModal && <Modal ... />

webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
```diff
@@ -63,27 +63,26 @@ class ChartBox extends React.Component {
...
+        <Loader status={this.props.status}>
+          {[_chart(), error]}
+        </Loader>
+        {this.state.showModal ?
+        <Modal show={this.state.showModal} enforceFocus onHide={this.closeModal}>
+          <Modal.Header closeButton>
+            <Modal.Title>{this.props.title}</Modal.Title>
+          </Modal.Header>
+          <Modal.Body>
+            {_chart('large')}
```
sharvit: I think it would be better to avoid calling functions inside the renderer.

What about:
```js
const chartProps = { data, key, onClick };
const panelChart = <Chart {...chartProps} config="regular" />;
const modalChart = <Chart {...chartProps} config="large" />;

{[panelChart, error]}

<Modal.Body>{modalChart}</Modal.Body>
```

webpack/assets/javascripts/react_app/components/statistics/__snapshots__/ChartBox.test.js.snap
```diff
@@ -1,286 +1,125 @@
...
+    <h3
+      data-placement="top"
+      data-toggle="tooltip"
+      onClick={[Function]}
+      style={
+        Object {
+          "cursor": "pointer",
+        }
+      }
+      title={undefined}
```
sharvit: Can you add title to the tests so we will get covered?

webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
```diff
@@ -1,11 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Panel } from 'react-bootstrap';
-
+import { Panel, Modal } from 'react-bootstrap';
```
sharvit: Can import the `Modal` from patternfly-react

webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
```diff
@@ -1,11 +1,12 @@
...
+import Immutable from 'seamless-immutable';
 import helpers from '../../common/helpers';
-import PieChart from '../common/charts/PieChart/';
-import { getLargePieChartConfig, navigateToSearch } from '../../../services/ChartService';
-import ChartModal from './ChartModal';
+import ForemanDonutChart from '../common/charts/DonutChart';
```
sharvit: shouldn't it be `common/charts/ForemanDonutChart`?

package.json
```diff
@@ -57,7 +57,6 @@
   "dependencies": {
     "babel-polyfill": "^6.26.0",
     "brace": "^0.10.0",
-    "c3": "^0.4.11",
```
sharvit, line 60: Yayyy!! :+1: 

patternfly-react are using react-c3js:
https://github.com/patternfly/patternfly-react/blob/master/package.json#L25

react-c3js using c3:
https://github.com/bcbcarl/react-c3js/blob/master/package.json#L39

Considered we just introduced the `webpack.vendor.js`:
https://github.com/theforeman/foreman/blob/develop/config/webpack.vendor.js

I think we should be able to remove the c3 from the `webpack.vendor.js` and let pattenrfly-reat to be the only consumer of this package.
Can you try to remove it and check if it's still work? also need to check if the `c3` is getting into the `vendor.js` naively.
You can verify it with:
https://gist.github.com/sharvit/19440366d4edcb3c4302b19e8049b7be

sharvit (changes requested):
Thanks @amirfefer, It's great to see you were able to remove so much code!

webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
amirfefer: I'm not sure, why should we add foreman prefix to the file name, if its under foreman dir anyway ?

webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
amirfefer: Like that :+1:

webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
sharvit: I am mostly afraid of confusion because the component name doesn't fit with the component file/directory.

maybe it should simply be called `DonutChart`?

amirfefer:
Thanks @sharvit 
I've fixed your comments above

webpack/assets/javascripts/react_app/components/componentRegistry.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import PieChart from './common/charts/PieChart/';
+import DonutChart from './common/charts/DonutChart/index';
```
sharvit: Don't need the `/index` here

webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
```diff
@@ -29,30 +31,29 @@ class ChartBox extends React.Component {
...
+    const isExist = () => chart.data && chart.data.length > 0;
     const tooltip = {
-      onClick: this.onClick,
+      onClick: isExist() ? this.onClick : null,
       title: this.props.tip,
       'data-toggle': 'tooltip',
       'data-placement': 'top',
+      className: isExist() ? 'pointer' : '',
     };
     const onclickChartClicked =
```
sharvit: Can you change the name to `handleChartClick`?

webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
```diff
@@ -29,30 +31,29 @@ class ChartBox extends React.Component {
...
-      data: this.props.chart.data,
-      id: `${chart.id}Modal`,
-    });
-
+    const { chart, type } = this.props;
+    const components = {
+      donut: DonutChart,
+    };
+    const Chart = components[type];
+    const isExist = () => chart.data && chart.data.length > 0;
```
sharvit: The name is not that clear (what exists?), i would call it `hasChartData`

webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
```diff
@@ -29,30 +31,29 @@ class ChartBox extends React.Component {
...
     const tooltip = {
-      onClick: this.onClick,
+      onClick: isExist() ? this.onClick : null,
       title: this.props.tip,
       'data-toggle': 'tooltip',
       'data-placement': 'top',
+      className: isExist() ? 'pointer' : '',
```
sharvit: shouldn't we ignore all those props when we have no data?

webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
```diff
@@ -29,30 +31,29 @@ class ChartBox extends React.Component {
...
-    const modalConfig = getLargePieChartConfig({
-      data: this.props.chart.data,
-      id: `${chart.id}Modal`,
-    });
-
+    const { chart, type } = this.props;
+    const components = {
+      donut: DonutChart,
+    };
+    const Chart = components[type];
```
sharvit, line 38: Why using the `components` map instead of just use the `DonutChart`?

sharvit (changes requested):
Thanks @amirfefer :+1: added a few more comments...

webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
amirfefer: IMO ChartBox should be generic, and be able to support different kind of charts, for later use.

webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
amirfefer: we still need those props, in order to render a chartbox, the inside chart would be empty, however the wrapping frame shouldn't.

webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
sharvit: So can you set the `propType.type` to be `oneOf(['donut'])`?

webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
```diff
@@ -29,64 +31,65 @@ class ChartBox extends React.Component {
...
     const error = (
       <MessageBox
         msg={this.props.errorText}
         key={`${this.props.chart.id}-error`}
         icontype="error-circle-o"
       />
     );
+    const dataFiltered = chart.data && chart.data.filter(arr => arr[1] !== 0);
+    const hasChartData = dataFiltered && dataFiltered.length > 0;
+    const boxHeader = <h3 {...hasChartData && tooltip}>{this.props.title}</h3>;
```
sharvit: tbh, i don't like this `{...hasChartData && tooltip}` it doesn't clear that you apply all the tooltip props.
what about:
```js
const headerProps = hasChartData ? { ... } : {};
<h3 {...headerProps}>
```

sharvit (changes requested):

sharvit (approved):
Looks good!! Thanks @amirfefer :+1:

sharvit:
@ohadlevy, any other comments? can we merge it?

amirfefer:
@ohadlevy  ping?

webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
ohadlevy: @sharvit anything else missing?

webpack/assets/javascripts/services/ChartService.consts.js
```diff
@@ -14,9 +15,11 @@ export const donutChartConfig = {
     label: { show: false },
   },
   data: {
-    type: 'donut',
     columns: [],
   },
+  color: {
+    pattern: ['#0088ce', '#cc0000', '#ec7a08', '#3f9c35', '#005c66', 'f9d67a'],
```
ohadlevy, line 21: just making sure this is intended.. the colors are different from the current color set.
before
image


now
image

ohadlevy:
@sharvit mind approving / nacking this pr before I merge?

sharvit (approved):
LGTM

webpack/assets/javascripts/services/ChartService.consts.js
amirfefer: this color pattern is taken from `patternfly` with @Rohoover  approval.
still, we need to re-design charts with more then 5 categories, due to patternfly, this chart type supports up to 5 categories.

dLobatog:
I have embarked on using React/redux from a plugin (REX) since I could. During the lifetime of https://github.com/theforeman/foreman_remote_execution/pull/130 I thought it would be able to make that available for many REX versions (whatever we have for 1.16,1.17 at least). However, the Axios change #5080 and now this one make the code only usable in 1.18+. 

I know it's still the early stages of React in Foreman, as times goes by I see everything moving towards a nicer, more user-friendly, Patternfly-UI. I also love seeing the improvements being contributed back to patternfly-react! As an 'early-adopter' I know there'd be growing pains. One of them is what I mentioned - for the future it would be awesome if you could **deprecate** things that other plugins may be using.

It'd be ridiculous to ask for a deprecation for something that a plugin has only plans to use it (https://github.com/theforeman/foreman_remote_execution/pull/130), I'm just saying we should pay attention to this as Katello, REX and other plugins join the React boat :boat:. It's much less frustrating to see a deprecation error and knowing you have a couple of months to fix it than having nightlies broken :smile: 

@amirfefer @sharvit @ohadlevy great work on this PR btw, I'm working on adapting the aforementioned PR in REX to it.

webpack/assets/javascripts/services/ChartService.consts.js
```diff
@@ -1,6 +1,7 @@
 const enums = {
   SIZE: {
     LARGE: { height: 500 },
+    REGULLAR: { width: 240, height: 240 },
```
tbrisker: Minor nitpick: Typo here, should be `REGULAR`

webpack/assets/javascripts/services/ChartService.consts.js
```diff
@@ -27,11 +30,13 @@ export const donutChartConfig = {
     right: 0,
     bottom: 0,
   },
+  size: enums.SIZE.REGULLAR,
```
tbrisker: ditto

tbrisker (dismissed):
just a tiny nitpick :)

dLobatog:
https://github.com/theforeman/foreman_remote_execution/pull/130#pullrequestreview-98212901 updated the PR on REX to work with this - great job, took very little effort.

Rohoover (commented):
@amirfefer 
Now that is a colorful donut!

amirfefer:
@tbrisker , @ohadlevy  Thanks, I've fixed that.

package.json
ohadlevy: @sharvit I assume this should be removed from https://github.com/theforeman/foreman/blob/develop/config/webpack.vendor.js ?

ohadlevy (approved):
thanks @amirfefer looks good! I had only one question regarding https://github.com/theforeman/foreman/blob/develop/config/webpack.vendor.js and then I'm happy to merge.

ohadlevy (approved):

PR #5140, opened 2018-01-02
Fixes #22111 - add taxonomy indexes to hosts

tbrisker:


db/migrate/20180102082705_add_taxonomy_index_to_hosts.rb
```diff
@@ -0,0 +1,6 @@
+class AddTaxonomyIndexToHosts < ActiveRecord::Migration[5.1]
+  def change
+    add_index :hosts, [:type, :organization_id, :location_id]
+    add_index :hosts, [:type, :location_id]
```
timogoebel, line 4: Why haven't you added [:type, :organization_id] as well?

tbrisker: Because the database is smart enough to use partial indexes, so if you only search on org the first one will be used. (you can test this by manually adding the index on your host and running `explain analyze` on pg or `explain` on mysql)

timogoebel: Ok, but why is the second ([:type, :location_id]) needed then?

tbrisker: because partial indexes only work from left to right - see the commit message of https://github.com/rails/rails/commit/9cdd0a1fdf8308985231242d378e3a1c29c4ab00 for details. We could probably also improve or remove some of our other indexes due to this, but that is a matter for another pr.

timogoebel:
Makes sense, thanks @tbrisker.

[test katello]

mmoll:
katello is broken with Rails 5.1 currently.

timogoebel (approved):

timogoebel:
Merged, thanks @tbrisker.

PR #5142, opened 2018-01-02
refs #20929 - do not stub settings

timogoebel:
Stubbing Settings can cause issues with plugins, for example with expire_hosts:

```
unexpected invocation: Setting(id: integer, name: string, value: text, description: text, category: string, settings_type: string, default: text, created_at: datetime, updated_at: datetime, full_name: string, encrypted: boolean).[](:is_host_expiry_date_mandatory)...
```

### DIFF
--- a/test/unit/fact_cleaner_test.rb
+++ b/test/unit/fact_cleaner_test.rb
@@ -2,6 +2,7 @@ require 'test_helper'
 require 'fact_importer_test_helper'
 
 class FactCleanerTest < ActiveSupport::TestCase
+  let(:null_store) { ActiveSupport::Cache.lookup_store(:null_store) }
   let(:cleaner) do
     FactCleaner.new
   end
@@ -50,11 +51,8 @@ class FactCleanerTest < ActiveSupport::TestCase
   end
 
   test "it cleans excluded facts" do
-    name_generator_type = Setting[:name_generator_type]
-    host_owner = Setting[:host_owner]
-    Setting.stubs(:[]).with(:excluded_facts).returns(['ignored*', '*bad'])
-    Setting.stubs(:[]).with(:name_generator_type).returns(name_generator_type)
-    Setting.stubs(:[]).with(:host_owner).returns(host_owner)
+    Setting.stubs(:cache).returns(null_store)
+    Setting[:excluded_facts] = ['ignored*', '*bad']
     cleaner.stubs(:delete_orphaned_facts).returns(0)
     good_fact_name = FactoryBot.create(:fact_name, :name => 'good_fact')
     ignored_fact_name = FactoryBot.create(:fact_name, :name => 'ignored01')
@@ -67,6 +65,7 @@ class FactCleanerTest < ActiveSupport::TestCase
   end
 
   test 'it cleans ignored flat facts' do
+    Setting.stubs(:cache).returns(null_store)
     [
       FactsData::FlatFacts,
       FactsData::RhsmStyleFacts,
@@ -81,7 +80,7 @@ class FactCleanerTest < ActiveSupport::TestCase
         fact_records << FactoryBot.create(:fact_name, :name => fact_name)
       end
 
-      Setting.stubs(:[]).with(:excluded_facts).returns(data.filter)
+      Setting[:excluded_facts] = data.filter
       cleaner.stubs(:delete_orphaned_facts).returns(0)
 
       cleaner.clean!
@@ -93,6 +92,7 @@ class FactCleanerTest < ActiveSupport::TestCase
   end
 
   test 'it cleans ignored hierarchical facts' do
+    Setting.stubs(:cache).returns(null_store)
     data = FactsData::HierarchicalPuppetStyleFacts.new
 
     facts_hierarchy = data.good_facts.deep_merge(data.ignored_facts)
@@ -104,7 +104,7 @@ class FactCleanerTest < ActiveSupport::TestCase
       fact_records << FactoryBot.create(:fact_name, :name => fact_name)
     end
 
-    Setting.stubs(:[]).with(:excluded_facts).returns(data.filter)
+    Setting[:excluded_facts] = data.filter
     cleaner.stubs(:delete_orphaned_facts).returns(0)
 
     cleaner.clean!


test/unit/fact_cleaner_test.rb
```diff
@@ -50,11 +50,7 @@ class FactCleanerTest < ActiveSupport::TestCase
   end
 
   test "it cleans excluded facts" do
-    name_generator_type = Setting[:name_generator_type]
-    host_owner = Setting[:host_owner]
-    Setting.stubs(:[]).with(:excluded_facts).returns(['ignored*', '*bad'])
-    Setting.stubs(:[]).with(:name_generator_type).returns(name_generator_type)
-    Setting.stubs(:[]).with(:host_owner).returns(host_owner)
+    Setting[:excluded_facts] = ['ignored*', '*bad']
```
iNecas: This still has a potential to influence other tests, that would expect original setting. What about using `unstub` in `teardown` instead?

ares: Or restore original value after the test, e.g.
```ruby
a,b='new',a
$test
a=b
```

ShimShtein: I think the database is isolated between tests anyway. If we add 
``` ruby 
Rails.application.configure do
  config.cache_store = :null_store
end
```

to the mix, settings should be isolated completely.

timogoebel:
I stubbed settings cache with a `null_store` and removed the test stubbing from two other tests for completeness.

test/unit/fact_cleaner_test.rb
```diff
@@ -50,11 +51,8 @@ class FactCleanerTest < ActiveSupport::TestCase
   end
 
   test "it cleans excluded facts" do
-    name_generator_type = Setting[:name_generator_type]
-    host_owner = Setting[:host_owner]
-    Setting.stubs(:[]).with(:excluded_facts).returns(['ignored*', '*bad'])
-    Setting.stubs(:[]).with(:name_generator_type).returns(name_generator_type)
-    Setting.stubs(:[]).with(:host_owner).returns(host_owner)
+    Setting.stubs(:cache).returns(null_store)
```
iNecas, line 54: Shouldn't this rather be done globally on the test environment preparation in the test helper? Don't we otherwise risk, that the stubbed method would potentially influence other tests and it would be order dependent, as it seems that was the issue before the patch?

timogoebel: It should, but let's please do it in a separate PR and not on a "refs #" PR. :-)

iNecas: I have a proposal:

1. lets try to do the stubbing globally in test helper
2. if everything still works, let's keep it
3. if there are some failures discovered that would be independent of the original issue, let's keep this version.

Deal?

timogoebel: Fair enough, deal.

timogoebel:
@iNecas: Here you go. :-)
I didn't stub the call, but added a `Rails.env.test?` guard in the settings model. Not pretty, but let's see if it breaks anything.

test/models/setting_test.rb
```diff
@@ -6,6 +6,11 @@ class SettingTest < ActiveSupport::TestCase
   should validate_uniqueness_of(:name)
   should validate_inclusion_of(:settings_type).in_array(Setting::TYPES)
 
+  setup do
+    # Activate caching for this testsuite
+    Setting.stubs(:cache).returns(Rails.cache)
+  end
```
iNecas: This guess we should unstub this in `teardown` or?

timogoebel: I don't think, this is required. Mocha actually cleans up after each test, see https://github.com/freerange/mocha/blob/065cacd6e6b87d9852a6740393d3cdba3570773f/lib/mocha/integration/mini_test/version_230_to_2101.rb#L56-L57

iNecas: Oh. So the issue wasn't that other tests would be affected by this, bur rather the facts ignoring test failing when some plugin was around?

timogoebel: Not exactly, let's take a step back.

Plugin test also run the whole core test suite. When a plugin extends the Host model (e.g. by a before_validate hook or something) and accesses the `Setting` model there, that call is not stubbed in the core test suite. The test in the core test suite fails then.

Real world example:
https://github.com/theforeman/foreman_expire_hosts/blob/ffbe87e35c1deadb973924ca56cce9b25f7852b4/app/models/concerns/foreman_expire_hosts/host_ext.rb#L8

Another (unreal) approach how to fix this would be to extend the core test from plugins (that's not possible to my knowledge) like this:

```
Setting.stubs(:[]).with(:discovery_reboot).returns('whatever')
Setting.stubs(:[]).with(:is_host_expiry_date_mandatory).returns('whatever')
```

You can see an example (discovery) here:

http://ci.theforeman.org/job/test_plugin_pull_request/4468/

I hope, the motivation for this is clear now. I'll set this for 1.17 as this affects two plugins.

iNecas: Everything clear now, works for me. Merging now

iNecas (approved):

iNecas:
I would like just check first Katello doesn't get broken by this change

test/unit/fact_cleaner_test.rb
iNecas: @timogoebel I know you'll probably hate me  by now, but since I was not able to run the katello tests successfully (with or without the patch, I tried 1.16 + katello 3.5), and since I had a wrong assumption to begin with, for now, I would suggest getting back to the version, where the stubbing of `:cache` was done in the `fact_cleaner_test`, to minimize the potential negative impact on others. Sorry again :/

timogoebel: @iNecas: Don't worry, I'll restore the original version. Luckily I have a backup.

timogoebel:
@iNecas: Restored the version with the local stubbing.

iNecas (approved):

iNecas:
Thanks @timogoebel merging now + I will cherry-pick this to 1.17

PR #5143, opened 2018-01-02
fixes #22080 - rabl is loaded before patching

timogoebel:
I took the liberty of opening a PR for this so we can close this issue. Thanks, @ShimShtein for looking into the root cause of this.


mmoll (approved):

mmoll:
fixes the problem for me, APJ

iNecas:
Makes sense, thanks @timogoebel and @mmoll

PR #5145, opened 2018-01-02
Fixes #22114 - Webpack common-chunk vendor.js

sharvit:
Use webpack-common-chunk-plugin to include all
required modules from node_modules/ into vendor.js

### DIFF
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -8,6 +8,7 @@ var ExtractTextPlugin = require('extract-text-webpack-plugin');
 var CompressionPlugin = require('compression-webpack-plugin');
 var LodashModuleReplacementPlugin = require('lodash-webpack-plugin');
 var pluginUtils = require('../script/plugin_webpack_directories');
+var vendorEntry = require('./webpack.vendor');
 
 // must match config.webpack.dev_server.port
 var devServerPort = 3808;
@@ -16,6 +17,8 @@ var devServerPort = 3808;
 var production =
   process.env.RAILS_ENV === 'production' ||
   process.env.NODE_ENV === 'production';
+
+var bundleEntry = path.join(__dirname, '..', 'webpack/assets/javascripts/bundle.js');
 
 var plugins = pluginUtils.getPluginDirs();
 
@@ -27,7 +30,8 @@ var resolveModules = [  path.join(__dirname, '..', 'webpack'),
 var config = {
   entry: Object.assign(
     {
-      bundle: './webpack/assets/javascripts/bundle.js'
+      bundle: bundleEntry,
+      vendor: vendorEntry,
     },
     plugins.entries
   ),
@@ -119,9 +123,9 @@ var config = {
       }
     }),
     new webpack.optimize.CommonsChunkPlugin({
-      name: 'vendor'
+      name: 'vendor',
+      minChunks: Infinity,
     })
-
   ]
 };
 
--- a/config/webpack.vendor.js
+++ b/config/webpack.vendor.js
@@ -1,0 +1,67 @@
+module.exports = [
+  /**
+   * React related
+   */
+  'babel-polyfill',
+  'react',
+  'react-dom',
+  'react-bootstrap',
+  'react-ellipsis-with-tooltip',
+  'react-numeric-input',
+  'react-onclickoutside',
+  'patternfly-react',
+  'react-redux',
+  'redux',
+  'redux-form',
+  'redux-form-validators',
+  'redux-logger',
+  'redux-thunk',
+  'prop-types',
+  'seamless-immutable',
+  'isomorphic-fetch',
+
+  /*
+   * jquery related
+   */
+  'jquery',
+  'jquery.cookie',
+  'select2',
+  'multiselect',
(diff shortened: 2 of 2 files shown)


config/webpack.config.js
```diff
@@ -119,7 +119,14 @@ var config = {
       }
     }),
     new webpack.optimize.CommonsChunkPlugin({
-      name: 'vendor'
+      name: 'vendor',
+      minChunks(module) {
+        if (module.resource && (/^.*\.(css|scss)$/).test(module.resource)) {
```
ohadlevy: why not only .js files? or css is excluded?

ohadlevy (commented):
great, this seems to solve the issue! one question inline

config/webpack.config.js
sharvit: tbh I don't have a strong opinion about it.
Maybe it would be better to vendor all types of chunks?

I agree that if we decide to restrict so we should restrict to js only.

config/webpack.config.js
```diff
@@ -119,7 +119,14 @@ var config = {
...
+          return false;
+        }
+
+        return module.context && module.context.includes("node_modules");
```
ohadlevy: won't this also catch plugin dependencies? e.g. if a plugin add a package, this will also increase the vendor file?

config/webpack.config.js
```diff
@@ -118,8 +126,20 @@ var config = {
...
-      name: 'vendor'
+      name: 'vendor',
+      minChunks: function(module, count) {
+        const rootIssuer = getModuleRootIssuer(module);
+
+        const isNodeModule = module.context && module.context.includes("node_modules");
+        const isUsedByBundle = rootIssuer.resource === bundle;
+        const isUsedByMultipleEntries = count > 1;
+
+        return isNodeModule && (isUsedByBundle || isUsedByMultipleEntries);
```
tstrachota: Won't the `... || isUsedByMultipleEntries` part produce different results for production and for development envs?

tstrachota:
Tested with Katello and `vendor.js` seems to be fixed regardless of what plugins are installed so it fixes the issue.

config/webpack.config.js
sharvit: @tstrachota You right, Thanks!

@ohadlevy I still haven't found a good approach here, right now, my best solution is to create a list of node_modules that should get into the vendor.js.

sharvit: @ohadlevy When i think about it again, using a vendors list can cause problems with packages using other packages...

waldenraines:
[test]

sharvit:
I know it feels like it fix the problem but it actually doesn't,
build results will be different based on the plugins it builds with.

I start working on a fix which includes a static `vendor.js` entry with a fixed list of packages.

sharvit:
Updated to use `webpack.vendor.js`

I think this solution will fit best with the way plugins getting build.
`Vendor.js` should be the same when building foreman with each plugin and without plugin.

tstrachota:
@sharvit wouldn't a `minChunks` function that would check only if the dependency is required from the main `bundle.js` have the same effect? Similar to your original solution but without the condition that allowed any dependency required more than twice to get in. If I understand it correctly it should produce the same results and we would avoid having the static list.

sharvit:
@tstrachota I have tried that with no success, the problem is that webpack will vendor only the chunks it use, and i have not way to add the unused chunks. (that plugins may use them)

For example, the package `uuid` is used by foreman and may be used by some plugins.
So foreman importing this plugins as well:
```js
import uuidV1 from 'uuid/v1';
```

Now webpack will put in the `vendor.js` only the `uuid/v1`.
`uuid/v2` won't be available for the plugins and they will need to add into to their bundle.
When i use static list, i can be sure that both `v1` and `v2` will be in the vendor file.

About avoiding the static list:
At the beginning i told myself __using a static list is the last option and i should try everything to avoid it__
But after i try everything and start implementing it, i feel it is even better to use a static list.
This way we are more transparent with plugin developers and they can be sure about the packages we supports.

Now we can look at the list and easier understand what is the painful packages.
We can manually decide exactly what packages is available for plugins and what not.

For example, we can get out the `jquery-flot` from the `vendor` because we don't want it in foreman anymore and if plugins still want to use it, webpack will add it to their own bundle.

sharvit:
I have updated the `webpack.vendor` file to contain all packages except `lodash`.

I think it should be enough and i am satisfied.

New bundle sizes are:
```
107K bundle-7b91995530795e50e869.js
  26K bundle-7b91995530795e50e869.js.gz
  262 manifest.json
 2.7M vendor-92434eb558a69bd8724b.js
 728K vendor-92434eb558a69bd8724b.js.gz
```

tstrachota:
@sharvit I get it, that makes sense

sharvit:
Any news about it? still waiting for a review...

dLobatog (approved):
I read the history and didn't find any drawbacks yet. It seems to work well with React in REX/Katello, so let's merge.

dLobatog:
Thanks @sharvit / @tstrachota  @ohadlevy

PR #5146, opened 2018-01-03
Fixes #22065 - Fix missing permissions table in tests

xprazak2:
When loading plugins in test for the first time, the permissions table does not yet exist, which was causing failures. We do not use permissions registered from plugins in tests, but we create those we need with fixtures or factories.


iNecas (approved):
Tested and works well

tbrisker (approved):
Thanks @xprazak2 !

PR #5148, opened 2018-01-03
Fixes #22138 - fix tooltip triggering

amirfefer:
Tooltip is not removed after click and requires antoher click in order to remove it.
This may be annoying especially on dropdown / buttons etc...
hosts - edited

### DIFF
--- a/webpack/assets/javascripts/foreman_tools.js
+++ b/webpack/assets/javascripts/foreman_tools.js
@@ -52,7 +52,7 @@ export function activateTooltips(elParam = 'body') {
   el
     .find('*[title]')
     .not('*[rel]')
-    .tooltip({ container: 'body' });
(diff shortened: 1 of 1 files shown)


ohadlevy:
@tbrisker mind having a look?

webpack/assets/javascripts/foreman_tools.js
```diff
@@ -52,7 +52,7 @@ export function activateTooltips(elParam = 'body') {
   el
     .find('*[title]')
     .not('*[rel]')
-    .tooltip({ container: 'body' });
+    .tooltip({ container: 'body', trigger: 'hover' });
```
tbrisker, line 55: why only on this and not on other tooltip activation methods? might make sense to clean up a bit an unify them

PR #5151, opened 2018-01-05
refs #22107 - subnet host count: permissions and n+1

timogoebel:
Only displays the link to the host count if a user has the "view_hosts" permission and counts the number of hosts without a "n+1" query as we do on similar pages.
I'm still not sure if the query counts correct, but could you (@dLobatog + @tbrisker) please take a look? Thanks.

### DIFF
--- a/app/views/subnets/index.html.erb
+++ b/app/views/subnets/index.html.erb
@@ -20,10 +20,10 @@
       <tr>
         <td class="display-two-pane ellipsis"><%=link_to_if_authorized subnet.name, hash_for_edit_subnet_path(:id => subnet).merge(:auth_object => subnet, :authorizer => authorizer) %></td>
         <td><%=subnet.network_address %></td>
-        <td class="ellipsis"><%=subnet.domains.map(&:name).to_sentence %></td>
-        <td><%=subnet.vlanid %></td>
-        <td class="ellipsis"><%=subnet.dhcp %></td>
-        <td><%= link_to subnet.hosts_count, hosts_path(:search => "subnet.name=\"#{subnet}\"") %>
+        <td class="ellipsis"><%= subnet.domains.map(&:name).to_sentence %></td>
+        <td><%= subnet.vlanid %></td>
+        <td class="ellipsis"><%= subnet.dhcp %></td>
+        <td><%= link_to_if_authorized(hosts_count.fetch(subnet.id, 0), hash_for_hosts_path(:search => "subnet.name=\"#{subnet}\"")) %>
         <td class="col-md-1">
           <%= action_buttons(display_delete_if_authorized(
             hash_for_subnet_path(:id => subnet).
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -257,7 +257,7 @@ Metrics/MethodLength:
 # Offense count: 25
 # Configuration parameters: CountComments.
 Metrics/ModuleLength:
-  Max: 434
+  Max: 437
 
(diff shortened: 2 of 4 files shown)


tbrisker (commented):
This seems to work correctly, but I wonder if perhaps we should put the logic inside the host_count method instead of passing it from the view, similar to how we handle unscoping for taxonomies? 
Another question I'm not sure about - should we include only primary nic subnet or all nics? the search seems to lead to only primary subnet.

timogoebel:
> I wonder if perhaps we should put the logic inside the host_count method instead of passing it from the view, similar to how we handle unscoping for taxonomies?

Could you give me an example what you have in mind? I can't follow.

> should we include only primary nic subnet or all nics?

I stumbled upon that as well. It might be best to be consistent with the search results, but I don't have a strong opinion for or against the other.

tbrisker:
@timogoebel I was thinking something like:
```ruby
  def hosts_count(resource_name = controller.resource_name)
    hosts_scope = Host::Managed.reorder('')
    case resource_name
    when 'organization', 'location'
      # If we are on /organizations or /locations, this allows to display the
      # count for hosts not in the current organization & location.
      hosts_scope = hosts_scope.unscoped
    when 'subnet'
      hosts_scope = hosts_scope.joins(:primary_interface)
    end
    @hosts_count ||= hosts_scope.authorized.group(:"#{resource_name}_id").count
  end
```
We could also perhaps refactor some of the other pages similarly, such as domains.

timogoebel:
@tbrisker: Thanks. I quite like that.

dLobatog (dismissed):
Moving back to WoC based on @tbrisker 's comment, I also think it makes sense

timogoebel:
@tbrisker and @dLobatog: Thanks for your comments. Addressed your feedback.

app/views/subnets/index.html.erb
```diff
@@ -20,10 +20,10 @@
...
         <td class="display-two-pane ellipsis"><%=link_to_if_authorized subnet.name, hash_for_edit_subnet_path(:id => subnet).merge(:auth_object => subnet, :authorizer => authorizer) %></td>
         <td><%=subnet.network_address %></td>
-        <td class="ellipsis"><%=subnet.domains.map(&:name).to_sentence %></td>
-        <td><%=subnet.vlanid %></td>
-        <td class="ellipsis"><%=subnet.dhcp %></td>
-        <td><%= link_to subnet.hosts_count, hosts_path(:search => "subnet.name=\"#{subnet}\"") %>
+        <td class="ellipsis"><%= subnet.domains.map(&:name).to_sentence %></td>
+        <td><%= subnet.vlanid %></td>
+        <td class="ellipsis"><%= subnet.dhcp %></td>
+        <td><%= link_to_if_authorized(hosts_count(controller.resource_name).fetch(subnet.id, 0), hash_for_hosts_path(:search => "subnet.name=\"#{subnet}\"")) %>
```
tbrisker: no need to pass the parameter here, it's the default value

tbrisker (changes requested):
just one minor comment, other than that looks good.

timogoebel:
@tbrisker: Thanks, done. 🌲

tbrisker (approved):

tbrisker:
[test foreman]

timogoebel:
@tbrisker: Tests are 🍏 📗 💚 🥗

tbrisker:
Thanks @timogoebel !

PR #5154, opened 2018-01-07
Fixes #22165 - Allow disabling HSTS header

tbrisker:
If a user browses to the Foreman server using HTTPS, HSTS headers will
prevent the browser from connecting to the server again using HTTP. This
adds a setting that allows disabling this header for users requiring
browser access to the server.

### DIFF
--- a/config/initializers/secure_headers.rb
+++ b/config/initializers/secure_headers.rb
@@ -1,5 +1,9 @@
 ::SecureHeaders::Configuration.default do |config|
-  config.hsts = "max-age=#{20.years.to_i}; includeSubdomains"
+  if SETTINGS[:hsts_enabled]
+    config.hsts = "max-age=#{20.years.to_i}; includeSubdomains"
+  else
+    config.hsts = "max-age=0; includeSubdomains"
+  end
   config.csp = {
     :default_src => ["'self'"],
     :child_src   => ["'self'"],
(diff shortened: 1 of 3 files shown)


ares:
LGTM, but it will also need support in [the installer](https://github.com/theforeman/puppet-foreman/blob/master/templates/settings.yaml.erb), I guess it should be configurable from there similarly like other [settings params](https://github.com/theforeman/puppet-foreman/blob/e16aa609a4effbf915fb364ead0107ea264370c5/manifests/init.pp#L33), would you mind preparing the patch?

tbrisker:
@ares thanks, opened https://github.com/theforeman/puppet-foreman/pull/614

ares:
I'll wait a bit if the concern re setting name is also raised here, but since the PR was opened I guess this one can get it, thanks

tbrisker:
@ares @ekohl - updated, renamed to `hsts_enabled` instead. if this is accepted i'll update the installer change as well.

config/initializers/secure_headers.rb
```diff
@@ -1,5 +1,9 @@
 ::SecureHeaders::Configuration.default do |config|
-  config.hsts = "max-age=#{20.years.to_i}; includeSubdomains"
+  if SETTINGS[:hsts_enabled]
+    config.hsts = "max-age=#{20.years.to_i}; includeSubdomains"
+  else
+    config.hsts = "max-age=0; includeSubdomains"
```
ekohl, line 5: Do browsers actually update their HSTS caches when they see this?

ekohl (approved):
I'm not sure yet about the existing caches in users' browsers but that doesn't mean we should hold this back because admins can set this up on fresh installations or instruct users to purge their caches if needed.

config/initializers/secure_headers.rb
tbrisker: if they follow the specs, then yes: https://tools.ietf.org/html/rfc6797#section-6.1.1

dLobatog (approved):
@tbrisker From my side 👍 , just rebase to make sure tests pass (re: audited update). Thank you!

ekohl (approved):

dLobatog (approved):
Thanks @tbrisker

PR #5157, opened 2018-01-09
Fixes #22155 - Allow name of datacenter in API

orrabin:
Background: when using the API to create a compute resource `datacenter` expects any string. When sending a name to an ovirt compute resource any action tries to send that name to ovirt which returns a not found error because ovirt expects to get the datacenter uuid.

I didn't add tests yet because I want to make sure this is the right approach.

From what I understand vmware finds the datacenter by name but ovirt needs the uuid.
There is an option to create an ovirt specific controller or ovirt specific behavior in the api controller that either changes name to uuid or rejects anything that isn't a uuid.

I chose to try and keep something that would work for both ovirt and vmware.
To get the datacenters in ovirt I have to test the connection to get an initialized client with a datacenter list.

There is a problem in case the name in ovirt is not unique which means we'd have to force using uuid but.

### DIFF
--- a/app/models/compute_resources/foreman/model/ovirt.rb
+++ b/app/models/compute_resources/foreman/model/ovirt.rb
@@ -193,6 +193,12 @@ module Foreman::Model
       client.datacenters(options).map { |dc| [dc[:name], dc[:id]] }
     end
 
+    def get_datacenter_uuid(name)
+      datacenter_uuid = datacenters.select{|dc| dc[0] == name}
+      raise ::Foreman::Exception.new(N_('Datacenter was not found')) if datacenter_uuid.empty?
+      datacenter_uuid.first[1]
+    end
+
     def editable_network_interfaces?
       # we can't decide whether the networks are available when we
       # don't know the cluster_id, assuming it's possible
--- a/test/controllers/api/v2/compute_resources_controller_test.rb
+++ b/test/controllers/api/v2/compute_resources_controller_test.rb
@@ -150,6 +150,28 @@ class Api::V2::ComputeResourcesControllerTest < ActionController::TestCase
     assert !available_storage_domains.empty?
   end
 
+  test "should create with datacenter name" do
+    Foreman::Model::Ovirt.any_instance.stubs(:datacenters).returns([["test", Foreman.uuid]])
+    Foreman::Model::Ovirt.any_instance.stubs(:test_connection).returns(true)
+
+    attrs = { :name => 'Ovirt-create-test', :url => 'https://myovirt/api', :provider => 'ovirt', :datacenter => 'test', :user => 'user@example.com', :password => 'secret' }
+    post :create, params: { :compute_resource => attrs }
+    assert_response :created
+    show_response = ActiveSupport::JSON.decode(@response.body)
+    assert Foreman.is_uuid?(show_response["datacenter"])
+  end
+
+  test "should create with datacenter uuid" do
+    datacenter_uuid = Foreman.uuid
+    Foreman::Model::Ovirt.any_instance.stubs(:datacenters).returns([["test", datacenter_uuid]])
+
+    attrs = { :name => 'Ovirt-create-test', :url => 'https://myovirt/api', :provider => 'ovirt', :datacenter => datacenter_uuid, :user => 'user@example.com', :password => 'secret' }
+    post :create, params: { :compute_resource => attrs }
+    assert_response :created
+    show_response = ActiveSupport::JSON.decode(@response.body)
+    assert Foreman.is_uuid?(show_response["datacenter"])
+  end
+
   context 'cache refreshing' do
     test 'should refresh cache if supported' do
       put :refresh_cache, params: { :id => compute_resources(:vmware).to_param }
--- a/app/controllers/api/v2/compute_resources_controller.rb
+++ b/app/controllers/api/v2/compute_resources_controller.rb
@@ -54,6 +54,13 @@ module Api
 
       def create
         @compute_resource = ComputeResource.new_provider(compute_resource_params)
+
+        datacenter = compute_resource_params[:datacenter]
+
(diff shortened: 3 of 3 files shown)


orrabin:
@timogoebel @ShimShtein could you please take a look? maybe you'll have a better solution.

tbrisker:
Shouldn't this be done in fog-ovirt? I think we should try to leverage fog's abstractions as much as we can instead of carrying code ourselves to handle differences between various compute resources.

orrabin:
@tbrisker fog-ovirt doesn't have an option to get a single datacenter.
This can be done there but that would mean we need to bump the dependency of fog-ovirt (and likely rbovirt) and considering the timeline this was requested for I'm not sure that's the best idea.

Also, this has an added value of running test_connection so other error that were hidden from the API might appear now.

timogoebel:
I'm also not sure (yet), what the best approach is. How about we do this in the model to capsule the specific behavior instead of the controller?

tbrisker:
@orrabin wouldn't as a first step just adding a 'datacenter' action in fog-ovirt that does the same thing (i.e. loop over all datacenters and check if it's a name or uuid) be easy to implement and then in the future improve on that to actually get just the needed datacenter from rbovirt/ovirt-sdk correctly? It would still require a bump to fog-ovirt but that would save us the trouble of carrying provider specific code in foreman.

orrabin:
@tbrisker that would be the last step in this case. I want to get opinions on how best to fix it, fix it in core and then implement it in fog-ovirt with a possible rbovirt update once we have a valid solution.

orrabin:
@timogoebel this doesn't happen through the UI, only the API can cause this bug so maybe a specific ovirt controller that overrides just the `create` method?

orrabin:
I added a test.

@tbrisker I'll open a refactoring issue but currently we can not solve this in fog-ovirt.
This can be moved to an ovirt specific controller to make it easier to remove later or stay this way.
It should also be added to update but I'll do that after there is a decision about the best way to go with this.

iNecas:
what's the status here?

orrabin:
@iNecas we do not have an agreement on the correct way to solve this, there is 1 suggestion to solve in fog-ovirt instead and no +1 for this solution.

timogoebel:
Why don't we check if the cr has a `datacenter_name_to_uuid` method and call that if it's defined? That way we'd just add new code for ovirt and have a flexible API? And VMWare or other Compute Resources that have a `datacenter` parameter would not need to be changed?

xprazak2:
Setting this as WoC as per previous comments

orrabin:
@timogoebel I changed according to your comment.

timogoebel (dismissed):
Looks good enough to me. :-)
Could somebody please help with testing this? I don't have oVirt available. Thanks.

ik5:
Testing with Hammer the following PR, and it does create the compute resource with no errors in the UI,

However, when a wrong datacenter is provided, an exception is raised on a `nil` datacenter resource.

orrabin:
@ik5 I fixed your comment, thanks. Could you please test again?

ik5:
@orrabin now it provides `foreman exception` with the message of:
```
Could not create the compute resource:
  ERF42-6729 [Foreman::Exception]: Datacenter was not found
```

That's much better, thank you

app/models/compute_resources/foreman/model/ovirt.rb
```diff
@@ -193,6 +193,12 @@ def datacenters(options = {})
       client.datacenters(options).map { |dc| [dc[:name], dc[:id]] }
     end
 
+    def get_datacenter_uuid(name)
+      datacenter_uuid = datacenters.select{|dc| dc[0] == name}
```
ShimShtein, line 197: What happens if a user passes UUID as `name`?

orrabin: You are right, the code was supposed to return if the datacenter is a UUID, I must have deleted it by accident at some point. That is what was expected and why the `unless`exists: https://github.com/theforeman/foreman/pull/5157/files#diff-cf2965b610f2fa0c776c7129f985f26dR63

orrabin: @ShimShtein I added: https://github.com/theforeman/foreman/pull/5157/files#diff-cf2965b610f2fa0c776c7129f985f26dR60

ShimShtein: @orrabin, That is a good addition, although in case of a uuid, `@compute_resource.datacenter` will not be set.

test/controllers/api/v2/compute_resources_controller_test.rb
```diff
@@ -150,6 +150,17 @@ def teardown
...
+    Foreman::Model::Ovirt.any_instance.stubs(:datacenters).returns([["test", Foreman.uuid]])
+    Foreman::Model::Ovirt.any_instance.stubs(:test_connection).returns(true)
+
+    attrs = { :name => 'Ovirt-create-test', :url => 'https://myovirt/api', :provider => 'ovirt', :datacenter => 'test', :user => 'user@example.com', :password => 'secret' }
+    post :create, params: { :compute_resource => attrs }
+    assert_response :created
+    show_response = ActiveSupport::JSON.decode(@response.body)
+    assert Foreman.is_uuid?(show_response["datacenter"])
+  end
+
```
ShimShtein, line 163: Just for completeness sake, can we add a test for UUID use case?

ShimShtein (changes requested):

app/models/compute_resources/foreman/model/ovirt.rb
orrabin: @ShimShtein after talking about this - it will be set because of line 58.

test/controllers/api/v2/compute_resources_controller_test.rb
orrabin: added it

ShimShtein (approved):

iNecas:
Thanks @orrabin

PR #5158, opened 2018-01-09
Fixes #22203 - facts traversing supports dot in names

ares:
to reproduce:
1. have a composed fact that contains dot in the name, e.g. puppet facts for vlan interface
2. navigate to it's details
3. you get 404 since the route was not found

after applying the patch, this is possible
facts


tbrisker (approved):
Tested, works well. tHanks @ares !

PR #5159, opened 2018-01-09
Fixes #22205 - restore blueprint actions for notifications

ares:
self.actions defaults to `{}`, the logic with conditional overiding is correct but the condition is wrong

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -66,10 +66,10 @@ class Notification < ApplicationRecord
 
   def set_custom_attributes
     return unless notification_blueprint # let validation catch this.
-    self.actions ||= UINotifications::URLResolver.new(
+    self.actions = UINotifications::URLResolver.new(
       subject,
       notification_blueprint.actions
-    ).actions if notification_blueprint.actions.any?
+    ).actions if notification_blueprint.actions.any? && self.actions.blank?
     # copy notification message in case we didn't create a custom one.
     self.message ||= UINotifications::StringParser.new(
       notification_blueprint.message,


tbrisker:
@ares please rebase on latest develop to fix failing minitest issue

ares:
thanks @tbrisker, rebased

mmoll:
another rebase for the audited pin is needed now...

ekohl:
Does this also need the 1.17 milestone?

ares:
it's set in redmine, but I assigned the milestone too, I'll rebase soon

ares:
rebased

app/models/notification.rb
```diff
@@ -66,10 +66,10 @@ def set_notification_recipients
 
   def set_custom_attributes
     return unless notification_blueprint # let validation catch this.
-    self.actions ||= UINotifications::URLResolver.new(
+    self.actions = UINotifications::URLResolver.new(
       subject,
       notification_blueprint.actions
-    ).actions if notification_blueprint.actions.any?
+    ).actions if notification_blueprint.actions.any? && self.actions.blank?
```
ekohl, line 72: Not a ruby developer, but multi line statements with guards are hard to parse for me.

boaz0 (approved):
LGTM 💯

app/models/notification.rb
ares: I'm not making the problem worse, since this will be cherrypicked I'd probably keep it as it is, I can open separate PR with refactoring if you prefer

ekohl: Sounds good.

ares: https://github.com/theforeman/foreman/pull/5170 addresses it

tbrisker (approved):
Thanks @ares !

tbrisker:
Cherry-picked to 1.17-stable as well as d11899e8425af20e14cc9bbed363c7d63e5892df

PR #5160, opened 2018-01-10
Fixes #22190 - Pass empty hash when options is nil

ik5:


iNecas:
ok to test

iNecas:
add to PR whitelist

app/jobs/create_rss_notifications.rb
```diff
@@ -5,7 +5,7 @@ class CreateRssNotifications < ApplicationJob
 
   def perform(options = {})
     # Defaults to theforeman.org blog RSS
-    UINotifications::RssNotificationsChecker.new(options).deliver!
+    UINotifications::RssNotificationsChecker.new(options || {}).deliver!
```
ohadlevy, line 8: does this actually solve the issue? options is set to {} if its nil already?

I thought the real issue was at https://github.com/theforeman/foreman/blob/develop/app/services/ui_notifications/rss_notifications_checker.rb#L30
(note that options is an array vs hash)

ik5: The issue is, that the `ActiveTask` passes options as nil. I could not understand why it is so, maybe the async task, as we discussed f2f.

ares: I think the error is normally not reproducible and @ohadlevy correct me if I'm wrong, but you ran this method manually from console. Normally the issue is probably not reproducible. But @ohadlevy  is right, the `[]` should be changed to `{}` on line he points to.

While this is most likely not normally reproducible since perform options default to `{}`, it's still valid bug.

ares (changes requested):
As @ohadlevy mentioned, the fix should likely be changing the default value at https://github.com/theforeman/foreman/blob/develop/app/services/ui_notifications/rss_notifications_checker.rb#L30

ik5:
@ares, I'm sure that everything I'm writing here is not news for you :-)

The `arguments` is passed, but as `nil`. So we do have argument arity.
Changing the the default value will not solve the issue, because it will not assign it's value to the argument.

I'ved made a simple testing for it:
```
[1] pry(main)> def test(foo=[])                      
[1] pry(main)*   puts foo.class                      
[1] pry(main)* end                                   
=> :test                                             
[2] pry(main)> test                                  
Array                                                
=> nil                                               
[3] pry(main)> test(nil)                             
NilClass                                             
=> nil
```

The array as you know, will only be applied when there are no arguments (that is undefined arguments), while `nil` is a valid value, but not what it was expected to be.

Having said that, I do agree with you that the default value should be a Hash rather then Array, and going to change it now. :-)

ohadlevy:
@ik5 please rebase your pr, this will ensure that your PR does not include other peoples code.

ik5:
i'ved messed my fork up, cloned it again, and re-created as https://github.com/theforeman/foreman/pull/5164.

PR #5161, opened 2018-01-10
Fixes #22208 - pin audited to 4.5

tbrisker:
to prevent CI breakage.


iNecas:
nitpick: IMO this should be #refs, not #fixes

tbrisker:
@iNecas thanks, i didn't think i'd fix the root issue this fast. 
If https://github.com/theforeman/foreman/pull/5162 is merged first we can close this one, it's just so we don't block develop in the meantime.

ohadlevy (approved):

PR #5162, opened 2018-01-10
Fixes #22208, #21920 - Refactor password auditing

tbrisker:
Recent changes in Rails 5.1 and audited gem cause our method of auditing
passwords to break. This PR refactors password auditing, so that instead
of recording a change to attribute `password_changed`, we will now
record the string `[redacted]` instead of any actual password.
The change is done currently in our audit extensions, which mean that it
will now apply to all resources that have a `password` attribute instead
of just those that have defined the workaround.
The next step will be to move this to the audited gem in a more
generalized method that can be defined in the model when initializing
audited, so that the workaround can be removed.

### DIFF
--- a/app/helpers/audits_helper.rb
+++ b/app/helpers/audits_helper.rb
@@ -13,7 +13,7 @@ module AuditsHelper
       when /.*_id$/
         label = name.classify.gsub('Id','').constantize.find(change).to_label
       else
-        label = change.to_s == "[encrypted]" ? _(change.to_s) : change.to_s
+        label = change.to_s == AuditExtensions::REDACTED ? _(change.to_s) : change.to_s
     end
     label = _('[empty]') unless label.present?
     if truncate
--- a/app/models/concerns/audit_extensions.rb
+++ b/app/models/concerns/audit_extensions.rb
@@ -2,6 +2,8 @@
 module AuditExtensions
   extend ActiveSupport::Concern
 
+  REDACTED = N_('[redacted]')
+
   included do
     def self.auditable_type_complete_values
       @auditable_type_complete_values ||= {
@@ -73,6 +75,7 @@ module AuditExtensions
 
     before_save :fix_auditable_type, :ensure_username, :ensure_auditable_and_associated_name
     before_save :filter_encrypted, :if => Proc.new {|audit| audit.audited_changes.present?}
+    before_save :filter_passwords, :if => Proc.new {|audit| audit.audited_changes.try(:has_key?, 'password')}
 
     include Authorizable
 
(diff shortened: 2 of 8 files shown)


boaz0 (approved):
LGTM 🥇

app/models/concerns/audit_extensions.rb
```diff
@@ -73,6 +73,7 @@ def self.auditable_type_complete_values
 
     before_save :fix_auditable_type, :ensure_username, :ensure_auditable_and_associated_name
     before_save :filter_encrypted, :if => Proc.new {|audit| audit.audited_changes.present?}
+    before_save :filter_passwords, :if => Proc.new {|audit| audit.audited_changes.try(:has_key?,'password')}
```
ares: missing space before `'password'`, for some time `has_key?` was considered deprecated but from what I've read recently, rubocop ignores it again as matz confirmed, it will be present in ruby 3

app/models/concerns/audit_extensions.rb
```diff
@@ -101,6 +102,14 @@ def filter_encrypted
     end
   end
 
+  def filter_passwords
+    if action == 'update'
+      audited_changes['password'] = [N_('[redacted]'), N_('[redacted]')]
```
ares: perhaps extract the string to a constant?

ares (changes requested):
two nits, looks good, I guess the code won't change too much so I'll start testing this version

app/models/concerns/audit_extensions.rb
tbrisker: done, and unified it with the encrypted redaction as well.

tbrisker:
katello failure seems unrelated afaict.

app/helpers/audits_helper.rb
```diff
@@ -13,7 +13,7 @@ def id_to_label(name, change, truncate = true)
       when /.*_id$/
         label = name.classify.gsub('Id','').constantize.find(change).to_label
       else
-        label = change.to_s == "[encrypted]" ? _(change.to_s) : change.to_s
+        label = change.to_s == "[redacted]" ? _(change.to_s) : change.to_s
```
ares: should be using `AuditExtensions::REDACTED`?

ares (changes requested):
another nit introduced by recent changes

app/helpers/audits_helper.rb
tbrisker: fixed

test/models/image_test.rb
```diff
@@ -36,38 +36,32 @@ class ImageTest < ActiveSupport::TestCase
...
-        assert protected_image.password_changed_changed?
-        assert protected_image.password_changed
-        assert_includes protected_image.changed, "password_changed"
-        assert protected_image.save
-        #testing after_save
-        refute protected_image.password_changed_changed?
-        refute protected_image.password_changed
-        refute_includes protected_image.changed, "password_changed"
-      end
+    test "password should be redacted for new or desroyed image" do
```
mmoll: extremely minor nit: des**t**royed

tbrisker: fixed

ares (approved):
Tested with all fields that I know about which contain password. Works great.

ares:
Thanks @tbrisker merging! Please open audited 4.6 PR in foreman-packaging if you can.

mmoll:
one note: either this change or the audited pinning needs to go into 1.17

ares:
I set release in redmine to 1.18, for 1.16 and 1.17 we'll keep the audited pinned. If that's not what was desired, please update the ticket accordingly.

PR #5163, opened 2018-01-10
Fixes #13497 - Enable password field after test connection in CR

tbrisker:
When clicking test connection button (or load datacenters) on an
existing compute resource, the password field is disabled. If the
password was changed prior to testing the connection and then the form
is saved, this will cause the new password to not be transmitted to the
server, leading to the previous password remaining in the database.


ares (approved):
I was finally able to reproduce, tested with many CR types and all seems to work fine. I didn't find any other trigger.

ares:
Thanks @tbrisker, merging.

PR #5164, opened 2018-01-10
Fixes #22190 - Pass empty hash when options is nil

ik5:
This replaces https://github.com/theforeman/foreman/pull/5160.

### DIFF
--- a/app/jobs/create_rss_notifications.rb
+++ b/app/jobs/create_rss_notifications.rb
@@ -1,6 +1,6 @@
 class CreateRssNotifications < ApplicationJob
   after_perform do |job|
-    self.class.set(:wait => 12.hours).perform_later(job.arguments.first)
+    self.class.set(:wait => 12.hours).perform_later(*job.arguments)
   end
 
   def perform(options = {})
--- a/app/services/ui_notifications/rss_notifications_checker.rb
+++ b/app/services/ui_notifications/rss_notifications_checker.rb
@@ -27,7 +27,7 @@ module UINotifications
       end
     end
 
-    def initialize(options = [])
+    def initialize(options = {})
       @url = options[:url] || Setting[:rss_url]
       @latest_posts = options[:latest_posts] || 3
       @force_repost = options[:force_repost] || false


app/jobs/create_rss_notifications.rb
```diff
@@ -5,7 +5,7 @@ class CreateRssNotifications < ApplicationJob
 
   def perform(options = {})
     # Defaults to theforeman.org blog RSS
-    UINotifications::RssNotificationsChecker.new(options).deliver!
+    UINotifications::RssNotificationsChecker.new(options || {}).deliver!
```
ares: while this does not hurt, I still don't understand, why this would be required. This could only happen if the method is called with `nil` as an argument. The cause seems to be at https://github.com/theforeman/foreman/blob/develop/app/jobs/create_rss_notifications.rb#L3 the `arguments.first` could cause it, I suppose we should simply `perform_later` with `job.arguments`

ares (changes requested):
I think the cause is elsewhere, this would be just a workaround

app/jobs/create_rss_notifications.rb
ik5: My issue with it is that although the `first` method is driven from `Enumerable`, it is more like an `Array` or `Set` (for example) related, and less good for a `Hash`. With `Hash` you don't really know the order of the key/value pairs, and it will break it into `{key, value}` result.

So it means that `job.arguments` is an array in this case, just from the syntax, before checking the documentation about it (that I could not find btw).

The thing is, is that doing `job.arguments[:foo]` for example, will result with the following exception:
```
TypeError: no implicit conversion of Symbol into Integer
```

And as we know based on the stack, it will arrive to a method that looks for a Hash rather then an array.

So checking for `nil` is better imo, then passing an `Array`, what do you think?

ares: > My issue with it is that although the first method is driven from Enumerable, it is more like an Array or Set (for example) related, and less good for a Hash. With Hash you don't really know the order of the key/value pairs, and it will break it into {key, value} result.

absolutely, the issue IMHO is the fact, that we do `job.arguments.first`, we should simply pass job.arguments, which are expected to be hash based on what's here [perform method](https://github.com/theforeman/foreman/blob/develop/app/jobs/create_rss_notifications.rb#L6)

ik5: Now I get it, sorry, added the fix.

ares: no worries, I guess this change is not required anymore, `options` should now always be a hash so I'd recommend removing `|| {}` from line 8, then it's good to be merged

ik5: done :)

app/jobs/create_rss_notifications.rb
```diff
@@ -1,6 +1,6 @@
 class CreateRssNotifications < ApplicationJob
   after_perform do |job|
-    self.class.set(:wait => 12.hours).perform_later(job.arguments.first)
+    self.class.set(:wait => 12.hours).perform_later(job.arguments)
```
ares: I added pry here and was surprised that `job.arguments` is actually an array, because ActiveJob::Base [stores it as an array](https://github.com/rails/rails/blob/master/activejob/lib/active_job/core.rb#L71-L72). Therefore the `job.arguments` is `[{}]` and therefore the `arguments.first` was required. When someone calls it e.g. from console without any param, empty array is passed so `job.arguments.first` would resolve to nil. So in fact this part needs to be a bit more smarter and unfold the arguments, so it should look like this `perform_later(*jobs.arguments)`. That way we ensure all arguments passed to previous `perform` are passed to `perform_later`.

Sorry for wrong suggestion at first, but we're getting there :-)

ik5: i hope that now it is okay :D

ares: I'm afraid it's not, `job.arguments` is the array of arguments that were passed to perform, imagine you create a job like this `CreateRssNotifications(1, 2, 3)`, then the `job.arguments # => [1,2,3]` so the original code would perform the job later but only with 1 throwing away all the rest. Therefore the `job.argument.first` is wrong. Defaulting to empty hash would again be just a workaround for case no arguments were passed to the job perform method.

We should pass the very same set of arguments, therefore we need to unfold it using `*`. So `perform_later(*job.arguments)` is IMHO the only correct call. That handles the empty set of params too.

ik5: right, i see what you mean, now b/c there is no value at all inside the array, the arity will be 0, and the default hash will take place.

Done :)

ares: exactly, I'll wait for tests, do quick tests and then it's good to go :-)

ares (approved):

ares:
letting jenkins confirm, then quick local test and good to be merged

ares:
Works just fine, thanks @ik5, katello tests failures are unrelated, so merging.

PR #5165, opened 2018-01-10
Refs #22208 - pin audited to 4.5

mmoll:
to prevent CI breakage.


mmoll:
this is for 1.15

tbrisker:
@mmoll 1.15 shipped with 4.3.0 in RPMs (not sure about debs but most likely as well), shouldn't we pin to that (or add a `<4.6.0` constraint) instead?

mmoll:
ah, thanks....changed that. I was under the impression that there was no audited release for a long time.

PR #5166, opened 2018-01-10
Refs #22208 - pin audited to 4.5

mmoll:
to prevent CI breakage.


mmoll:
this is for 1.16

tbrisker:
same comment as for GH-5165, only with 4.4.1 here

mmoll:
done.

PR #5167, opened 2018-01-10
Fixes #22236 - Adds breadcrumbs to foreman

amirfefer:
compute resources

### DIFF
--- a/app/controllers/audits_controller.rb
+++ b/app/controllers/audits_controller.rb
@@ -5,6 +5,7 @@ class AuditsController < ApplicationController
 
   def index
     @audits = resource_base_search_and_page.with_taxonomy_scope.preload(:user)
+    @host = resource_finder(Host.authorized(:view_hosts), params[:host_id]) if params[:host_id]
   end
 
   def show
--- a/app/controllers/config_reports_controller.rb
+++ b/app/controllers/config_reports_controller.rb
@@ -7,6 +7,7 @@ class ConfigReportsController < ApplicationController
   def index
     respond_to do |format|
       format.html do
+        @host = resource_finder(Host.authorized(:view_hosts), params[:host_id]) if params[:host_id]
         @config_reports = resource_base_search_and_page(:host)
         render :index
       end
--- a/app/helpers/layout_helper.rb
+++ b/app/helpers/layout_helper.rb
@@ -16,6 +16,21 @@ module LayoutHelper
     content_for(:search_bar) { elements.join(" ").html_safe }
   end
 
+  def mount_breadcrumbs(&block)
+    index_page = {caption: _(controller_name.humanize), url: try("#{controller_name}_path")}
+    default_menu = [(index_page unless action_name == 'index'),
+                    {caption: @page_header, url: '#' }].compact unless block_given?
+
+    mount_react_component("Breadcrumb", "#breadcrumb",
+      { menu: default_menu || yield }.to_json)
+  end
+
+  def breadcrumbs(&block)
+    content_for(:breadcrumbs) do
+      mount_breadcrumbs(&block)
+    end
+  end
+
   def stylesheet(*args)
     content_for(:stylesheets) { stylesheet_link_tag(*args.push("data-turbolinks-track" => true)) }
   end
--- a/app/views/audits/index.html.erb
+++ b/app/views/audits/index.html.erb
@@ -1,3 +1,12 @@
+ <%= if @host
+   breadcrumbs do
+     [
+       {caption: _("Hosts"), url: (url_for(hosts_path) if authorized_for(hash_for_hosts_path))},
+       {caption: @host.name, url: (host_path(@host) if authorized_for(hash_for_host_path(@host)))},
+       {caption: _('Audits'), url: url_for(audits_path)},
+     ]
+  end
+end %>  
 <% title _("Audits") %>
 <% title_actions documentation_button('4.1.4Auditing') %>
 <%= render :partial => 'list', :locals =>{:audits => @audits}%>
--- a/app/views/compute_attributes/new.html.erb
+++ b/app/views/compute_attributes/new.html.erb
@@ -1,3 +1,9 @@
-<% title _("New compute profile on %s") % @set.compute_resource.to_label %>
-
+<%= breadcrumbs do
+      [
+        {caption: _("Compute Resources"), url: (url_for(compute_resources_path) if authorized_for(hash_for_compute_resources_path))},
+        {caption: @set.compute_resource.to_label, url: (compute_resource_path(@set.compute_resource) if authorized_for(hash_for_compute_resource_path(@set.compute_resource)))},
+        {caption: _("Compute Profiles"), url: (url_for(compute_profiles_path) if authorized_for(hash_for_compute_profiles_path))},
+        {caption: _('New %s' % @set.compute_profile.to_label)}
+      ]
+    end %>
 <%= render 'form' %>
--- a/app/views/compute_resources/show.html.erb
+++ b/app/views/compute_resources/show.html.erb
@@ -48,7 +48,7 @@
   <% if @compute_resource.capabilities.include?(:image) %>
     <div id="images" class="tab-pane">
       <%= title_actions(new_link(_("Create Image"), { :compute_resource_id => @compute_resource.id, :controller => :images }, { :data=>{:target=>"full-page"} })) %>
-      <div id="images_list" data-url=<%= compute_resource_images_path(@compute_resource) %>>
+      <div id="images_list" data-url=<%= compute_resource_images_path(@compute_resource, { :partial => true }) %>>
         <%= spinner(_('Loading images information ...')) %>
       </div>
     </div>
--- a/app/views/config_reports/index.html.erb
+++ b/app/views/config_reports/index.html.erb
@@ -1,3 +1,12 @@
+<%= if @host
+  breadcrumbs do
+    [
+      {caption: _("Hosts"), url: (url_for(hosts_path) if authorized_for(hash_for_hosts_path))},
+      {caption: @host.name, url: (host_path(@host) if authorized_for(hash_for_host_path(@host)))},
+      {caption: _('Reports'), url: url_for(config_reports_path(search: 'eventful=true'))},
+    ]
+  end
+end %>  
 <% title _("Reports") %>
 <% title_actions csv_link, documentation_button('3.5.4PuppetReports') %>
 <%= render :partial => 'list' %>
--- a/app/views/filters/edit.html.erb
+++ b/app/views/filters/edit.html.erb
@@ -1,3 +1,9 @@
-<% title _("Edit Filter") %>
+<%= breadcrumbs do
+      [
+        (@role.present? ? {caption: _('Roles'), url: (url_for(roles_path) if authorized_for(hash_for_roles_path))} : {caption: _("Filters"), url: filters_path}),
+        ({caption: _("%s filters") % @role.to_s, url: filters_path(role_id: @role.id)} if @role.present?),
+        {caption: _('Edit %s') % @filter.to_s},
+      ].compact
+    end %>
 
 <%= render 'form' %>
--- a/app/views/images/new.html.erb
+++ b/app/views/images/new.html.erb
@@ -1,3 +1,9 @@
-<% title _("Create Image") %>
-
+<%= breadcrumbs do
+      [
+        {caption: _('Compute Resources'), url: (url_for(compute_resources_path) if authorized_for(hash_for_compute_resources_path))},
+        {caption: @compute_resource.to_s, url: (url_for(compute_resource_path(@compute_resource)) if authorized_for(hash_for_compute_resource_path(@compute_resource)))},
+        {caption: _('Images'), url: (url_for(compute_resource_images_path(@compute_resource)) )},
+        {caption: _('Create image')}
+      ]
+    end %>
 <%= render :partial => 'form' %>
--- a/app/views/layouts/_application_content.html.erb
+++ b/app/views/layouts/_application_content.html.erb
@@ -1,21 +1,31 @@
 <div id="main">
-  <div id="content" >
+  <div id="content">
     <%= notifications %>
-    <% unless @page_header.blank? %>
-      <div class="row form-group">
-        <h1 class="col-md-8"><%=h @page_header %></h1>
+        <div id="breadcrumb">
+          <% unless @page_header.blank? %>
+            <div class="row form-group">
+              <h1 class="col-md-8">
+                <%=h @page_header %>
+              </h1>
+            </div>
+          <% end %>
+          <% if content_for?(:breadcrumbs) %>
+            <%= yield(:breadcrumbs) %>
+          <% else %>
+            <%= mount_breadcrumbs %>
+          <% end %>
+        </div>
+      <div class="row">
+        <div class="title_filter <%= searchable? ? " col-md-6 " : "col-md-4 " %>">
+          <%= render "common/searchbar" if searchable? %>
+            <%= yield(:search_bar) %>&nbsp;
+        </div>
+        <div id="title_action" class= <%= searchable? ? "col-md-6" : "col-md-8" %>>
+          <div class="btn-toolbar pull-right">
+            <%=h yield(:title_actions) %>
+          </div>
+        </div>
       </div>
-    <% end %>
-    <div class="row">
-      <div class="title_filter <%= searchable? ? "col-md-6" : "col-md-4" %>">
-        <%= render "common/searchbar" if searchable? %>
-        <%= yield(:search_bar) %>&nbsp;
-      </div>
-      <div id="title_action" class="<%= searchable? ? "col-md-6" : "col-md-8" %>">
-        <div class="btn-toolbar pull-right"><%=h yield(:title_actions) %></div>
-      </div>
-    </div>
-
-    <%= yield %>
+      <%= yield %>
   </div>
 </div>
--- a/app/views/ssh_keys/new.html.erb
+++ b/app/views/ssh_keys/new.html.erb
@@ -1,9 +1,7 @@
 <% title _('Add SSH Key for %s') % @user.name %>
-
-<ol class="breadcrumb">
-  <li class="breadcrumb-item"><%= link_to_if_authorized(_("Users"), hash_for_users_path) %></li>
-  <li class="breadcrumb-item"><%= link_to_if_authorized(@user.name, hash_for_edit_user_path(@user)) %></li>
-  <li class="breadcrumb-item active"><%= _('Add SSH Key') %></li>
-</ol>
-
+<%= breadcrumbs do %>
+  <% [{caption: _("Users"), url: (url_for(users_path) if authorized_for(hash_for_users_path))},
+      {caption: @user.name, url: (edit_user_path(@user) if authorized_for(hash_for_edit_user_path(@user)))},
+      {caption: _('Add SSH Key')}] %>
+<% end %>
 <%= render :partial => 'form' %>
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "lodash": "~4.15.0",
     "multiselect": "~0.9.12",
     "patternfly": "3.41.6",
-    "patternfly-react": "1.11.0",
+    "patternfly-react": "^1.12.0",
     "prop-types": "^15.6.0",
     "react": "^16.2.0",
     "react-bootstrap": "^0.31.5",
--- a/webpack/assets/javascripts/react_app/components/Breadcrumb/Breadcrumbs.fixtures.js
+++ b/webpack/assets/javascripts/react_app/components/Breadcrumb/Breadcrumbs.fixtures.js
@@ -1,0 +1,19 @@
+export const mockedBreadcrumbsMenu = {
+  menu: [
+    {
+      caption: 'root',
+      url: '/some-url',
+    },
+    {
+      caption: 'child',
+    },
+  ],
+};
+
+export const headerTitle = {
+  menu: [
+    {
+      caption: 'title',
+    },
+  ],
+};
--- a/webpack/assets/javascripts/react_app/components/Breadcrumb/index.js
+++ b/webpack/assets/javascripts/react_app/components/Breadcrumb/index.js
@@ -1,0 +1,30 @@
+import React from 'react';
+import { Breadcrumb } from 'patternfly-react';
+import 'patternfly-react/dist/sass/_breadcrumb.scss';
+
+const ForemanBreadcrumb = ({ data }) => {
+  if (data.menu.length === 1) {
+    return (
+      <div className="form-group">
+        <h1>{data.menu[0].caption}</h1>
+      </div>
+    );
+  }
+  return (
+    <div>
+      <Breadcrumb title>
+        {data.menu.map((item, index) => (
+          <Breadcrumb.Item
+            key={index}
+            active={index === data.menu.length - 1}
+            href={item.url}
+            dangerouslySetInnerHTML={{ __html: item.caption }}
+          />
+        ))}
+      </Breadcrumb>
+      <hr className="breadcrumb-line" />
+    </div>
+  );
+};
+
+export default ForemanBreadcrumb;
--- a/app/assets/stylesheets/base.scss
+++ b/app/assets/stylesheets/base.scss
@@ -193,6 +193,12 @@ select {
   padding-bottom: 12px;
 }
 
+.breadcrumb-line {
+  margin-top: -15px;
+  margin-right: -60px;
+  margin-left: -60px;
+}
+
 .grey,
 .gray {
   color: #808080 !important;
--- a/app/controllers/images_controller.rb
+++ b/app/controllers/images_controller.rb
@@ -8,7 +8,7 @@ class ImagesController < ApplicationController
     # Listing images in /hosts/new consumes this method as JSON
     @images = resource_base.where(:compute_resource_id => @compute_resource.id).includes(:operatingsystem)
     respond_to do |format|
-      format.html { render :partial => 'images/list' }
+      format.html { params[:partial] ? render(:partial => 'images/list') : render(:index) }
       format.json { render :json => @images.where(:operatingsystem_id => params[:operatingsystem_id], :architecture_id => params[:architecture_id]).order(:name) }
     end
   end
--- a/app/views/architectures/edit.html.erb
+++ b/app/views/architectures/edit.html.erb
@@ -1,3 +1,3 @@
-<% title _("Edit Architecture") %>
+<% title _("Edit %s") % @architecture.to_s %>
 
 <%= render :partial => 'form' %>
--- a/app/views/audits/show.html.erb
+++ b/app/views/audits/show.html.erb
@@ -8,7 +8,6 @@
                                         hash_for_host_path(:id => @audit.associated.to_param).merge(:auth_object => @audit.associated, :auth_action => 'view'),
                                         :title => _("Host details"),
                                         :class => 'btn btn-default') if @audit.auditable_type.match(/^Nic/) && @audit.associated_type == 'Host::Base' && @audit.associated %>
-<%= title_actions link_to 'Back', :back, :class=>'btn btn-default' %>
 <% tmplt = audit_template?(@audit) %>
 
 <div class="tabbable">
--- a/app/views/auth_source_ldaps/edit.html.erb
+++ b/app/views/auth_source_ldaps/edit.html.erb
@@ -1,3 +1,3 @@
-<% title _("Edit LDAP Auth Source") %>
+<% title _("Edit %s") % @auth_source_ldap.to_s %>
 
 <%= render :partial => 'form' %>
--- a/app/views/bookmarks/edit.html.erb
+++ b/app/views/bookmarks/edit.html.erb
@@ -1,3 +1,3 @@
-<% title _("Edit Bookmark") %>
+<% title _("Edit %s") % @bookmark.to_s %>
 
 <%= render :partial => 'form' %>
--- a/app/views/common_parameters/edit.html.erb
+++ b/app/views/common_parameters/edit.html.erb
@@ -1,3 +1,3 @@
-<% title _("Edit Global Parameter") %>
+<% title _("Edit %s") % @common_parameter.to_s  %>
 
 <%= render :partial => 'form' %>
--- a/app/views/compute_attributes/edit.html.erb
+++ b/app/views/compute_attributes/edit.html.erb
@@ -1,3 +1,9 @@
-<% title _("Edit compute profile on %s") % @set.compute_resource.to_label %>
-
+<%= breadcrumbs do
+      [
+        {caption: _("Compute Resources"), url: (url_for(compute_resources_path) if authorized_for(hash_for_compute_resources_path))},
+        {caption: @set.compute_resource.to_label, url: (compute_resource_path(@set.compute_resource) if authorized_for(hash_for_compute_resource_path(@set.compute_resource)))},
+        {caption: _("Compute Profiles"), url: (url_for(compute_profiles_path) if authorized_for(hash_for_compute_profiles_path))},
+        {caption: _('Edit %s' % @set.compute_profile.to_label)}
+      ]
+    end %>
 <%= render 'form' %>
--- a/app/views/compute_profiles/show.html.erb
+++ b/app/views/compute_profiles/show.html.erb
@@ -1,4 +1,4 @@
-<% title(_("Edit Compute profile: %s") % @compute_profile.name) %>
+<% title(@compute_profile.name) %>
 
 <% title_actions display_link_if_authorized(_("Back"), hash_for_compute_profiles_path, :class => 'btn btn-default') %>
 
--- a/app/views/config_groups/edit.html.erb
+++ b/app/views/config_groups/edit.html.erb
@@ -1,3 +1,3 @@
-<% title _("Edit Config group") %>
+<% title _("Edit %s") % @config_group.to_s %>
 
 <%= render 'form' %>
--- a/app/views/config_reports/show.html.erb
+++ b/app/views/config_reports/show.html.erb
@@ -15,8 +15,7 @@
 <%= render report_origin_output_partial(@config_report.origin), :logs => @config_report.logs %>
 <%= render 'metrics', :status => @config_report.status, :metrics => @config_report.metrics["time"] if @config_report.metrics["time"] %>
 
-<%= title_actions link_to(_('Back'), :back, :class => 'btn btn-default'),
-  display_delete_if_authorized(hash_for_config_report_path(:id => @config_report), :class=> "btn btn-danger"),
+<%= title_actions display_delete_if_authorized(hash_for_config_report_path(:id => @config_report), :class=> "btn btn-danger"),
   link_to(_("Host details"), @config_report.host, :class => 'btn btn-default'),
   link_to(_("Other reports for this host"), host_config_reports_path(@config_report.host), :class => 'btn btn-default')
 %>
--- a/app/views/domains/edit.html.erb
+++ b/app/views/domains/edit.html.erb
@@ -1,3 +1,3 @@
-<% title _("Edit Domain") %>
+<% title _("Edit %s") % @domain.to_s %>
 
 <%= render :partial => 'form' %>
--- a/app/views/environments/edit.html.erb
+++ b/app/views/environments/edit.html.erb
@@ -1,3 +1,3 @@
-<% title _("Edit Environment") %>
+<% title _("Edit %s") % @environment.to_s %>
 
 <%= render :partial => 'form' %>
--- a/app/views/fact_values/index.html.erb
+++ b/app/views/fact_values/index.html.erb
@@ -1,7 +1,13 @@
 <%= javascript 'charts' %>
 <%
 if params[:host_id].present?
-  title(_("Fact Values | %{host_name}") % {:host_name => params[:host_id]})
+  breadcrumbs do
+    [
+     {caption: _("Facts"), url: (url_for(fact_values_path) if authorized_for(hash_for_fact_values_path))},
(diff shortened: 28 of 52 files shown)


timogoebel:
@amirfefer: Note, that we already have breadrumbs for ssh keys:
https://github.com/theforeman/foreman/blob/00183e17ad70876385106a2e1979c5b52012f69d/app/views/ssh_keys/new.html.erb#L3-L7

waldenraines:
I get the following on the new compute resources page:

```
15:09:19 rails.1   | 2018-01-11T15:09:19 c35dcf0e [app] [F] ActionView::Template::Error (undefined method `edit_compute_resources_path' for #<#<Class:0x007f8cb62f67b0>:0x007f8cb993c720>):
15:09:19 rails.1   | 2018-01-11T15:09:19 c35dcf0e [app] [F]     1: <%= breadcrumbs('name', @compute_resource ) %>
15:09:19 rails.1   |  |     2: <%= form_for @compute_resource, :html => {:data => {:id => @compute_resource.try(:id)}} do |f| %>
15:09:19 rails.1   |  |     3:   <%= base_errors_for @compute_resource %>
15:09:19 rails.1   |  |     4:   <ul class="nav nav-tabs" data-tabs="tabs">
15:09:19 rails.1   | 2018-01-11T15:09:19 c35dcf0e [app] [F]   
15:09:19 rails.1   | 2018-01-11T15:09:19 c35dcf0e [app] [F] app/helpers/layout_helper.rb:25:in `block in breadcrumbs'
15:09:19 rails.1   |  | app/helpers/layout_helper.rb:20:in `breadcrumbs'
15:09:19 rails.1   |  | app/views/compute_resources/_form.html.erb:1:in `_8cff77e70f47deae4d41b9454e8b9ec0'
15:09:19 rails.1   |  | app/views/compute_resources/new.html.erb:3:in `_1ec78fed1c788d4c96ad944eee20c78b'
15:09:19 rails.1   |  | app/controllers/concerns/application_shared.rb:15:in `set_timezone'
15:09:19 rails.1   |  | app/models/concerns/foreman/thread_session.rb:32:in `clear_thread'
15:09:19 rails.1   |  | app/controllers/concerns/foreman/controller/topbar_sweeper.rb:12:in `set_topbar_sweeper_controller'
15:09:19 rails.1   |  | lib/middleware/catch_json_parse_errors.rb:8:in `call'
15:09:19 rails.1   |  | lib/middleware/session_safe_logging.rb:17:in `call'
15:09:19 rails.1   |  | lib/middleware/tagged_logging.rb:18:in `call'
```

webpack/assets/javascripts/react_app/components/Breadcrumb/index.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { Breadcrumb } from 'patternfly-react';
+
+const ForemanBreadcrumb = ({ data }) => <Breadcrumb>
+        {data.menu.map((item, index) =>
```
ohadlevy: please configure prettier

amirfefer:
@timogoebel :+1:  I've changed it to react component
@waldenraines: I've fixed this regression
@ohadlevy: I'm using foreman's content layout to render `breadcrumbs`

amirfefer:
Rebased

amirfefer:
rebased + test failures should be fixed

timogoebel:
@amirfefer: Patternfly recommends the breadcrumbs to be [placed underneath the navbar](http://www.patternfly.org/pattern-library/navigation/breadcrumbs/). Why did you choose a different approach?

amirfefer:
@timogoebel  thanks for mentioning  that
patternfly placement for it looks a bit odd to me:

breadcrumbs-1
VS

breadcrumbs-2

ohadlevy:
@rohoover  should we consider combining the header the breadcrums ? (note that some headers have images / icons next to the text).

ohadlevy (changes requested):
@amirfefer thanks!

Can you please add a generic test (similar to pagination) that goes over all pages to ensure they actually have pagination rendered correctly?

app/views/layouts/_application_content.html.erb
```diff
@@ -15,6 +15,15 @@
         <div class="btn-toolbar pull-right"><%=h yield(:title_actions) %></div>
       </div>
     </div>
+    <div  class="row">
+      <div class='col-md-6' id="breadcrumb">
+        <% if content_for?(:breadcrumbs) %>
+          <%= yield(:breadcrumbs) %>
+        <% else %>
+          <%= mount_breadcrumbs(get_resource) if !searchable? && get_resource %>
```
ohadlevy: Should we really use searchable as a way to define breadcrums ? even if its the same condition I would prefer to have a dedicated method.

I'm also not sure that if there is a search component we should not show breadcrums.. ?

sharvit:
https://gist.github.com/sharvit/41d73803d5e15e5348c5baa7be758b79

Rohoover:
No I wouldn't combine them. 

The reason it's below the nav is because that is more universal with all the different implementations of content. In the following link you can see the different iterations of design that utilize the breadcrumb:
http://www.patternfly.org/pattern-library/navigation/breadcrumbs/

It just so happens that we do not have a the other visual elements, such as the grey background, that help visually arrange the items. 

I would suggest grouping the title and content a little closer.

timogoebel:
> It just so happens that we do not have a the other visual elements, such as the grey background,

Is this planned? The cards e.g. don't look any good on white background unless they are accented.

Rohoover:
@timogoebel 
My mistake, I noticed further down in the documentation there is a breadcrumb design combined with title.

ohadlevy:
> I noticed further down in the documentation there is a breadcrumb design combined with title.

@Rohoover so in that case do you suggest to remove the title as it is today and combine it with the breadcrumbs? thanks!

Rohoover:
@ohadlevy 
Yes and it would be preferred if we could move the Katello portions to this as well so it's consistent (@waldenraines).

timogoebel:
@Rohoover: What about the grey background. Should we use that as well for normal pages or just with a "cards page". Although it's really off-topic, I'd love to see how the host view page would look using the cards layout.

Rohoover:
@timogoebel 
I think that would be a good idea. PF calls for it anyways and it plays up nicely against many elements including the cards.

waldenraines (approved):
This works for me everywhere I tested.

tstrachota:
@amirfefer could you rebase, please?

app/views/compute_resources/show.html.erb
```diff
@@ -5,7 +5,6 @@
                                             :title => _("Associate VMs to Foreman hosts"), :method => :put, :class=>"btn btn-default"),
 (display_link_if_authorized(_('Refresh Cache'), hash_for_refresh_cache_compute_resource_path(@compute_resource).merge(:auth_object => @compute_resource), :method => :put, :class => "btn btn-default") if @compute_resource.respond_to?(:refresh_cache) && @compute_resource.caching_enabled),
    link_to_if_authorized(_('Edit'), hash_for_edit_compute_resource_path(@compute_resource).merge(:auth_object => @compute_resource), :class => "btn btn-default")  %>
-
```
tstrachota: This seems unrelated.

app/views/layouts/_application_content.html.erb
tstrachota: +1 for having a dedicated method.
I have no strong opinion about checking the `!searchable?` we can tune it later if it doesn't work 100%.

app/helpers/layout_helper.rb
```diff
@@ -16,6 +16,24 @@ def search_bar(*elements)
...
   end
 
+  def get_resource
+    instance_variable_get("@#{controller_name.singularize}")
+  end
+
+  def mount_breadcrumbs(obj, &block)
+    label = obj.try(:to_label) || obj.try(:to_s) unless obj.new_record?
+    default_menu = [{caption: _(controller_name.humanize), url: try("#{controller_name}_path")},
+                    {caption: [_(action_name.humanize), label].join(' '), url: '#' }].compact unless block_given?
```
tstrachota: I'm not sure about showing the action in breadcrumbs. It's not consistent with katello and I don't think it adds much value. It also makes the breadcrumbs to change when you switch for example from "user edit" `Users >> Edit John` to "add breadcrumbs" `Users >> John >> Add SSH Key`.
I'd prefer displaying just the record's name without the action prefix.

tstrachota (changes requested):
It works well. I added couple of comments inline. I'd prefer removing the action name from breadcrumbs for the sake of consistency.

tstrachota:
@amirfefer it still needs a rebase

amirfefer:
@tstrachota 
Thanks for your review.
I've combined the header title with the breadcrumbs, as it described in patternfly design.

app/helpers/layout_helper.rb
amirfefer: I removed the action referring  from the breadcrumbs, now its combined with the title itself.

tstrachota:
@amirfefer I think that @Rohoover was actually against combining title and breadcrumbs in our case (see https://github.com/theforeman/foreman/pull/5167#issuecomment-359442794). And I agree with that. It's inconsistent with how Katello handles breadcrumbs:

katello_product

It looks quite odd with two font sizes in the breadcrumbs:

user_edit
user_ssh_key
host_detail

It also doesn't solve the issue with action names in breadcrumbs (https://github.com/theforeman/foreman/pull/5167#discussion_r166246945). They're still there. So I think the previous layout was better. It just needed removal of the action names.

**To be consistent with katello and in line with good practices I believe the breadcrumbs should follow the pattern:**

_on detail or edit pages:_ 
`[resource_name_plural]` > `[record_name]` 
e.g. `Users > admin`

_on new record pages:_ 
`[resource_name_plural]` > `New [resource_name_singular]` 
e.g. `Users > New user`

_on listing of nested resoruces:_ 
`[resource_name_plural]` > `[record_name]` > `[nested_resource_name_plural]` 
e.g. `Compute resources > libvirt > Images`

Rohoover (commented):
@tstrachota 

Yes initially I was until I saw the additional patternFly documentation on it. It was also brought to my attention that other areas are starting to use this approach, so we'll be a bit ahead of the curve.

I share your concerns about the text size. I'm going to bring this to the PF teams attention.

waldenraines:
> Yes initially I was until I saw the additional patternFly documentation on it. It was also brought to my attention that other areas are starting to use this approach, so we'll be a bit ahead of the curve.
> 
> I share your concerns about the text size. I'm going to bring this to the PF teams attention.

Wait, so we're saying that we don't want a title on the page and instead just the breadcrumbs?  I personally don't like the way this looks.

tstrachota:
Ok, I had to read through the whole thread again. Now I understand it, @Rohoover.

I agree with @waldenraines, I don't like the current look. Using gray background and white breadcrumbs bar would probably make it better, but I'm afraid we'd have to sort out title action buttons placement.

I vote for iterative approach, smaller steps at a time:
Let's get the breadcrumbs in with titles now and we can always improve it later.

waldenraines:
> I vote for iterative approach, smaller steps at a time:
> Let's get the breadcrumbs in with titles now and we can always improve it later.

+1.  We already have this style of breadcrumbs in katello.  If we change what we do here then we have to redo katello.

Rohoover (commented):
@waldenraines @tstrachota 

+1 to an iterative approach.

Some of the additional documentation on it is here (for anyone who hasn't seen).
http://www.patternfly.org/pattern-library/navigation/breadcrumbs/#design

All the designs showf an ideal situation with a card layout that offers a grey background to contrast the white breadcrumb area, this is obviously not our typical visual use case. Maybe add a grey line to separate it out.

timogoebel:
I personally like an iterative approach in general, but I think we should make sure, that when we add new groundbreaking functionality (i.e. breadcrumbs) we go all in so users don't have to learn the feature twice. Why do we have patternfly standards, if we deviate when it's more convenient for us?
The document clearly describes how breadcrumbs should look. And I do think, it looks good and integrates well with the rest of the layout.

http://www.patternfly.org/pattern-library/navigation/breadcrumbs/img/Breadcrumbs-02.png

> We already have this style of breadcrumbs in katello. If we change what we do here then we have to redo katello.

I think core should set the standard here, even if that means that plugins need to adapt.

waldenraines:
> I personally like an iterative approach in general, but I think we should make sure, that when we add new groundbreaking functionality (i.e. breadcrumbs) we go all in so users don't have to learn the feature twice. Why do we have patternfly standards, if we deviate when it's more convenient for us?

This is a good point and I think you just convinced me.  It will be a little bit more work but I do think it makes sense to adhere to patternfly designs when we have them.  I don't think this design was available when the breadcrumbs in katello were implemented but now that it is we should spend the effort to adhere to the design.

Rohoover (commented):
@waldenraines @timogoebel 
Consistency between Katello/Foreman is of utmost importance, which I think we have consensus on.

Basically even I don't always agree with PF, but better adhere to consistency and standards, and push the upstream (PF) to change than diverge "just because we don't like it".

Hope that makes sense.

waldenraines:
See also http://projects.theforeman.org/issues/22603 - Make breadcrumbs and resource switcher conform to patternfly design (Katello)

amirfefer:
I opened a [PR](https://github.com/patternfly/patternfly-react/pull/233) in `patternfly-react` which implements breadcrumbs combined with page title.
@tstrachota page titles are already including the specific  action, why would you like to remove this?
IMHO if we going to use breadcrumbs combined with page header, we should keep the current behavior - which means keep the action title.

tstrachota:
Since breadcrumbs should give you a hint where you are in the app structure, I consider "Edit user" or "Edit template" insufficient. I know that it's how the titles are implemented now, but imo it deserves improvement.

I'd expect `Users > Edit admin` or just `Users > admin` so that it's clearer what you're editing.
"Add user" is fine, since there's probably nothing better we could display (maybe just "Add").

For the resources that have show pages (proxy, host) I'd expect the breadcumbs to be `Users > admin` for the detail and `Users > admin > edit` for the editing form.

amirfefer:
@tstrachota, Thanks however,  I think that breadcrumbs combined with page header should give you a full reference where you are, not just a hint.
@Rohoover what is your point of view about that?

sharvit:
I agree with @amirfefer, I think it makes sense to show Breadcrumbs as a title that gives you a full reference to where you are.

Rohoover (commented):
@amirfefer 
Can you provide me a current example of what is shown and what the new implementation would be? Sorry, I'm having trouble understanding exactly what you all are referring to.

tstrachota:
> page header should give you a full reference where you are

I should have phrased it better, we're not in contradiction on this. That's why I'd like to have the record names in breadcrumbs for detail and edit pages.

waldenraines:
> Can you provide me a current example of what is shown and what the new implementation would be? Sorry, I'm having trouble understanding exactly what you all are referring to.

@Rohoover see Tomas's [comment here](https://github.com/theforeman/foreman/pull/5167#issuecomment-366917208) for an example.

Rohoover (commented):
@tstrachota @waldenraines 

Thanks. Yes, if I'm understanding correctly, then I agree with you.

amirfefer:
@tstrachota 
> For the resources that have show pages (proxy, host) I'd expect the breadcumbs to be Users > admin for the detail and Users > admin > edit for the editing form.

So in that case  if you click on `admin`, will you be directed to the admin's `show` page? 
however, your last page probably was index.
breadcrumbs should present location hierarchy, I don't see it in `Users > admin (show) > edit (admin)

If I have understood correctly, it should be like this:
`Users > Edit admin` + switcher button in order to switch the resource (i.e from `edit admin` to `edit John Doe`)

amirfefer:
I'm using `breadcrumbs combined with page title` patternfly-react implementation
@ohadlevy - it's problematic to create  a generic test for each and every page (like pagination you mentioned) 
pagination renders in the server, but in this case the rendering occurs  in the client, which means we need a fast and stable javascript driver, after some tests  I made with `Poltergeist` it would take time.
I don't think we should test react mounted components via ruby integration test with the today's configuration

mmoll:
[test foreman]
[test katello]

package.json
```diff
@@ -73,7 +73,7 @@
     "lodash": "~4.15.0",
     "multiselect": "~0.9.12",
     "patternfly": "^3.31.2",
-    "patternfly-react": "1.11.0",
+    "patternfly-react": "^1.12.0",
```
sharvit: Can you pin it?
We decided to use pin versions of `patternfly-react` until we will migrate to shallow snapshots. It just break to much on Travis.

sharvit (changes requested):
Looks good to me!
can you pin `patternfly-react`?

app/helpers/layout_helper.rb
```diff
@@ -16,6 +16,21 @@ def search_bar(*elements)
     content_for(:search_bar) { elements.join(" ").html_safe }
   end
 
+  def mount_breadcrumbs(&block)
+    index_page = {caption: _(controller_name.humanize), url: try("#{controller_name}_path")}
+    default_menu = [(index_page unless action_name == 'index'),
+                    {caption: @page_header, url: '#' }].compact unless block_given?
+
+    mount_react_component("Breadcrumb", "#breadcrumb",
+      { menu:  default_menu || yield }.to_json)
```
ohadlevy: nitpick: extra whitespace after menu

app/views/ssh_keys/new.html.erb
```diff
@@ -1,9 +1,7 @@
...
-<ol class="breadcrumb">
-  <li class="breadcrumb-item"><%= link_to_if_authorized(_("Users"), hash_for_users_path) %></li>
-  <li class="breadcrumb-item"><%= link_to_if_authorized(@user.name, hash_for_edit_user_path(@user)) %></li>
-  <li class="breadcrumb-item active"><%= _('Add SSH Key') %></li>
-</ol>
-
+<%= breadcrumbs do %>
+  <% [{caption: _("Users"), url: (url_for(users_path) if authorized_for(hash_for_users_path))},
+      {caption: @user.name, url: (edit_user_path(@user) if authorized_for(hash_for_edit_user_path(@user)))},
+      {caption: _('Add SSH Key'), url: '#'}] %>
```
ohadlevy: are you required to add the url: options even if you dont use it?

webpack/assets/javascripts/react_app/components/Breadcrumb/Breadcrumbs.fixtures.js
```diff
@@ -0,0 +1,13 @@
+export const mockedBreadcrumbsMenu = {
+  menu: [
+    {
+      caption: 'root',
+      url: '#',
+    },
+    {
+      caption: 'child',
+      url: '#',
```
ohadlevy: just double checking I understand this correctly - this is not really required? (for the child)

app/views/layouts/_application_content.html.erb
```diff
@@ -1,21 +1,33 @@
...
     <%= notifications %>
-    <% unless @page_header.blank? %>
-      <div class="row form-group">
-        <h1 class="col-md-8"><%=h @page_header %></h1>
+      <div class="row">
+        <div id="breadcrumb">
+          <% unless @page_header.blank? %>
+            <div class="row form-group">
+              <h1 class="col-md-8">
+                <%=h @page_header %>
```
ohadlevy, line 8: I find the the header in the wrong place / size? is that intentional?

image


vs 

image

app/views/layouts/_application_content.html.erb
```diff
@@ -1,21 +1,33 @@
...
+              </h1>
+            </div>
+          <% end %>
+          <% if content_for?(:breadcrumbs) %>
+            <%= yield(:breadcrumbs) %>
+          <% else %>
+            <%= mount_breadcrumbs %>
```
ohadlevy, line 15: any idea why smart proxies have the wrong label? 
image

ohadlevy: is it possible to add hierarchy for roles / filters?
e.g.
```ruby
breadcrumbs do
  [
   {caption: _("Roles"), url: (url_for(roles_path) if authorized_for(hash_for_roles_path))},
   {caption: @role.name, url: (url_for(@role) if authorized_for(hash_for_edit_role_path(@role)))},
  {caption: _('Filters')}
  ]
end

```

ohadlevy: when clicking on back (e.g. roles page) - the breadcrumbs / title disappear.
image

ohadlevy, line 15: can you please also handle pages such as facts, reports, audits etc

ohadlevy (changes requested):
overall looks good! thanks

a few questions:

webpack/assets/javascripts/react_app/components/Breadcrumb/Breadcrumbs.fixtures.js
amirfefer: right :+1:

app/views/layouts/_application_content.html.erb
tstrachota: I think that we either need to align the title with the rest of the elements or add the white row on gray background as patternfly design suggests (http://www.patternfly.org/pattern-library/navigation/breadcrumbs/#design).
Having it as it is now would IMHO visually be a step back.

tstrachota:
There's inconsistency in breadcrumbs for edit pages.

Editing user 'admin':
edit_user_h
Displays name of the resource.

Editing hostgroup 'Provisioning' and host 'clint-unrue':
edit_hostgroup_h
host_edit_h
Displays name of the record.

I spotted this at multiple places. Personally I find the latter version better - it's clear what record you're editing.

tstrachota:
Breadcrumbs for audits seem quite confusing to me. Especially when edit breadcrumbs are `Edit [record.name]` and not `[record.name] >> edit` I had a feeling that I'm on some action page when I saw them for the first time. Past tense could help.

But I'm ok with fixing it later in a different PR. The audits page is probably going to be changed anyway.

audit_detail_h
audit_detail_h_2
audit_detail_h_3

tstrachota:
I found some issues at compute profile pages.

The compute profile detail page shows "Edit" in the breadcrumb (nitpick).
compute_profile_detail

Clicking on compute resource name takes me to compute attrs edit page. I'd expect the breadcrumbs to have 3 levels there:
`Compute profiles >> 1 - Small >> Edit compute profile on oVIrt (oVirt)`
compute_profile_edit_attributes

Also the first level link takes you to the same url on this page.

tstrachota:
Similar issues are on the Images page.
Could it be nested in `Compute Resources >> [compute_resource.name] >> Create image` ?
create_image

The same issue with the first level link is here too.

app/views/layouts/_application_content.html.erb
Rohoover: Unfortunately the PF designs are only showing cards which are typically on grey backgrounds. There is however a note in the documentation that a grey horizontal line can be added:

"Divider (optional): A horizontal divider may be present below the breadcrumbs depending on the page layout used inside an application. If this format is used, it should be used consistently throughout the application."

tstrachota: I think that we either need to align the title with the rest of the elements or add the white row on gray background as patternfly design suggests (http://www.patternfly.org/pattern-library/navigation/breadcrumbs/#design).
Having it as it is now would IMHO visually be a step back.

amirfefer:
@ohadlevy , @tstrachota Thanks for your great feedbak!
I fixed title styles, and pages that required a specific breadcrumbs.

Could you have another look ?

app/views/images/new.html.erb
```diff
@@ -1,3 +1,8 @@
-<% title _("Create Image") %>
-
+<%= breadcrumbs do
+      [
+        {caption: @compute_resource.to_s, url: (url_for(compute_resource_path(@compute_resource)) if authorized_for(hash_for_compute_resources_path))},
+        {caption: _("Images"), url: (url_for(compute_resource_images_path(@compute_resource)) )},
+        {caption: _("Create image")}
```
tstrachota: Link to compute resource index is missing here.

Compare to images edit page:
`oVIrt (oVirt) >> Images >> Create image`
vs.
`Compute resources >> oVIrt (oVirt) >> Images >> Edit ...`

app/views/filters/edit.html.erb
```diff
@@ -1,3 +1,3 @@
-<% title _("Edit Filter") %>
+<% title _("Edit %s") % @filter.to_s %>
```
tstrachota: Could you add breadcrumbs for `Roles >> %{role_name} filters >> Edit filter` here?
Users come there from the roles page and switching the root breadcrumb to "Filters" feels unnatural. The same for the create page, please.

app/views/compute_attributes/new.html.erb
```diff
@@ -1,3 +1,9 @@
-<% title _("New compute profile on %s") % @set.compute_resource.to_label %>
-
+<%= breadcrumbs do
+      [
+        {caption: _("Compute Resources"), url: (url_for(compute_resources_path) if authorized_for(hash_for_compute_resources_path))},
+        {caption: _("Compute Profiles"), url: (url_for(compute_profiles_path) if authorized_for(hash_for_compute_profiles_path))},
+        {caption: @set.compute_resource.to_label, url: (compute_resource_path(@set.compute_resource) if authorized_for(hash_for_compute_resource_path(@set.compute_resource)))},
+        {caption: _('New compute profile')}
```
tstrachota: This one is tricky. Users can get there from both compute resource detail and compute profile detail. I've been thinking about it and most natural combination I could find is:
`Compute Resources >> %{compute_resource_label} >> Compute Profiles >> New %{profile_name}`
eg.:
`Compute Resources >> tstracho-laptop (Libvirt) >> Compute Profiles >> New 2-Medium`
Where "Compute Profiles" points to `/compute_profiles/`.

Similarly for the edit page:
`Compute Resources >> %{compute_resource_label} >> Compute Profiles >> Edit %{profile_name}`

This way it feels quite natural when coming from the compute resource detail. When coming from compute profiles page the order of breadcrumbs is swapped compared to what I'd expect but this is something I can live with. It will also correspond with breadcrumbs for images. What do you think?

tstrachota (changes requested):
Great update @amirfefer, thanks. Broken links are fixed now.

**Things that still need to be fixed:**
1. There are 3 pages where the breadcrumbs still need to be tuned (see my inline comments).

2. I'm getting warning in the console:
```
Warning: Invalid DOM property `class`. Did you mean `className`?
    in h1 (created by ForemanBreadcrumb)
    in div (created by ForemanBreadcrumb)
    in ForemanBreadcrumb
    in Provider
```

app/views/layouts/_application_content.html.erb
tstrachota: @ohadlevy I found this bug also on develop: http://projects.theforeman.org/issues/22901
Serious functionality issue, but not related to this PR.

amirfefer:
@tstrachota thanks again.
 I’ll appreciate it if you can have another look.
with these changes there are some broken integration tests though.

webpack/assets/javascripts/react_app/components/Breadcrumb/index.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { Breadcrumb } from 'patternfly-react';
+import 'patternfly-react/dist/sass/_breadcrumb.scss';
+
+const ForemanBreadcrumb = ({ data }) => {
+  if (data.menu.length === 1) {
+    return (
+      <div class="row form-group">
```
ohadlevy: please use className

ohadlevy:
@Rohoover i don't like the space between the buttons and the breadcrumbs above it, it seems like a lot of wasted space, does it make sense to have them in the same height? 
image


nitpick: maybe the image should align with the text size?

ohadlevy:
when clicking on a host audits, I was expecting to see breadcrumbs as well? e.g.
image

ohadlevy:
similar issue with reports (when clicking on a host configuration report) 
image

serenamarie125:
>  i don't like the space between the buttons and the breadcrumbs above it, it seems like a lot of wasted space, does it make sense to have them in the same height?

In this case, according to PF, the buttons should share the same "bar" as the breadcrumb

Rohoover (commented):
@serenamarie125 @ohadlevy 

Based on dev feedback, further up in the thread we discussed using the Horizontal line to create a consistent space for the breadcrumbs. This feature is  outlined in the PF documentation:
"Divider (optional): A horizontal divider may be present below the breadcrumbs depending on the page layout used inside an application. If this format is used, it should be used consistently throughout the application.
"
As per the documentation, if we're going to use this visual element, it should be used across pages. 

Regardless of line or not, bringing the breadcrumbs in line with those buttons would be very inconsistent with how we've been handling it on every other page, where it is not in line with any page actions. 

Also, I've seen a screen of the shared breadcrumb and page row and there is risk that long named hosts would collide with the buttons.

amirfefer:
@ohadlevy  something like this?
breadcrumbs-importent

It doesn't feel like the pf design, plus the breadcrumbs may break like this:
fireshot capture 77 - - http___localhost_3000_compute_resources_1-libvirt_images_1_edit

 inconsistent  with other pages:
fireshot capture 78 - hosts - http___localhost_3000_hosts

waldenraines:
> In this case, according to PF, the buttons should share the same "bar" as the breadcrumb

I don't see that in [the documentation](http://www.patternfly.org/pattern-library/navigation/breadcrumbs/) or in the [navigation guidelines](http://www.patternfly.org/pattern-library/navigation/guidelines/).  Where is this documented on the patternfly site?  

I think we should have the buttons in a separate row for consistency and we should revisit this in the future once the breadcrumb and button placement best practices are better documented.

Rohoover:
@waldenraines 
I checked with the PF UX Strategist and was not able to locate any documentation on this issue. 

I believe combing breadcrumbs with action button will lead us down a path fraught with exceptions:
1) What will we do when we have the search bar?
2) What will we do when we have too many action buttons?
3) What will we do when the breadcrumb title becomes too long?

Since this is an item of navigational context as well as an acting page title, it's best positioned at the top of the page to "set the stage" for all things below it. By putting it in line with actions, we are conflating it a bit with action buttons.

Below is a dedicated breadcrumb title area with horizontal line (to separate the space and as a visual element):

screen shot 2018-03-19 at 10 41 48 am

Using this layout, I believe, exceptions would be far more rare.

ohadlevy:
I tend to agree this might be a less than ideal path forward, but I can't
say I like the current (see my earlier attachment) solution either, would
it be better to add the horizontal line or some other way to make the
spacing a bit less distracting? (or maybe its just me..)

Rohoover:
@ohadlevy 
I share your concern about the space but see it more as small visual issue and not necessitating breaking of consistency (and pf). If anything I think there is a bigger issue with the host detail page. 

I think designing for the majority of pages vs. the one page (that probably itself needs to be redesigned - host merge addresses this) would be a better approach.

amirfefer:
@Rohoover Thanks for your feedback
I added a horizontal line beneath the breadcrumbs:

compute_with_line
@ohadlevy I inserted specific breadcrumbs in reports and audits pages from a host.

serenamarie125:
> better to add the horizontal line

@Rohoover based on the [design doc for breadcrumbs](http://www.patternfly.org/pattern-library/navigation/breadcrumbs/#design) it does show a horizontal line under, what do you think about that for a solution?  We are doing this in some other products already.

serenamarie125:
> I checked with the PF UX Strategist and was not able to locate any documentation on this issue.

@Rohoover let's chat with Matt C, as I think there's some work in the backlog around the topic

amirfefer:
@Rohoover, which horizontal line looks better?
horizontal-1

horizontal-1

tstrachota:
+1 for the full-width, plus the breadcrumbs should imho be vertically centered.

app/views/audits/index.html.erb
```diff
@@ -1,3 +1,13 @@
+ <%= if @host
+   breadcrumbs do
+     [
+       {caption: _("Hosts"), url: (url_for(hosts_path) if authorized_for(hash_for_hosts_path))},
+       {caption: @host.name, url: (host_path(@host) if authorized_for(hash_for_host_path(@host)))},
+       {caption: _('Audits'), url: url_for(audits_path)},
```
tstrachota, line 6: It should imho be just `Hosts > %{host_name} > Audits`:
```ruby
[
  {caption: _("Hosts"), url: (url_for(hosts_path) if authorized_for(hash_for_hosts_path))},
  {caption: @host.name, url: (host_path(@host) if authorized_for(hash_for_host_path(@host)))},
  {caption: _('Audits')},
]
```

Having a link to listing all audits nested under hosts doesn't make much sense to me.

app/views/config_reports/index.html.erb
```diff
@@ -1,3 +1,13 @@
+<%= if @host
+  breadcrumbs do
+    [
+      {caption: _("Hosts"), url: (url_for(hosts_path) if authorized_for(hash_for_hosts_path))},
+      {caption: @host.name, url: (host_path(@host) if authorized_for(hash_for_host_path(@host)))},
+      {caption: _('Reports'), url: url_for(config_reports_path)},
+      {caption: _('Reposrt for %s ') % host.name}
```
tstrachota: Same here, I find `Hosts > %{host_name} > Reports` more intuitive.

tstrachota (changes requested):

tstrachota:
Compute profile details and host detail pages throw a warning:
```
Warning: Received `true` for a non-boolean attribute `title`.
If you want to write it to the DOM, pass a string instead: title="true" or title={value.toString()}.
    in ol (created by Breadcrumb)
    in Breadcrumb (created by ForemanBreadcrumb)
    in div (created by ForemanBreadcrumb)
    in ForemanBreadcrumb
    in Provider
```

tstrachota:
Could we re-visit the breadcrumbs structure for hosts and proxies? I really find it quite annoying that I can't use breadcrumbs to get from the edit page back to the detail. Is it just me? :)

Rohoover (commented):
@amirfefer +1 to full width

I think the summary of this conversation is that the best approach currently supported by PF that would provide the most consistent experience, is to include the horizontal line (see my above comments and mockup). 

This would reduce instances of exceptions because it fits almost all use cases (I actually can't think of one that it doesn't), give the breadcrumbs their own dedicated space (especially if they're quite long), and offers them proper placement at the top of a page which serves their purpose as both a navigational and contextual feature.

This also helps visually differentiate the space as this was a previous concern which is what sparked the grey background discussion earlier.

Things to consider for the future are instances in which unbalanced white space is present (host details). This will hopefully be addressed in the future with a redesign.

amirfefer:
@tstrachota keep in mind that we already have `edit` button in the `title_actions` (host details page)
fireshot capture 88 - damon-rouselle test - http___localhost_3000_hosts_damon-rouselle test
is this the right location ? should we move it to the breadcrumbs navigation? I'm not sure.

tstrachota:
@amirfefer the buttons on the detail page are totally fine. I'm just suggesting to change breadcrumbs on the host (and proxy) edit page to `Hosts > %{host_name} > Edit  %{host_name}` so that breadcrumbs provide link to do the step back.
Otherwise it's "click on all hosts" and "click on the host again"

ohadlevy:
Would adding the links in the action area will be better for consistency
usage cases?

tstrachota:
I'd expect it to be in the breadcrumbs. Action buttons is different location. But if it's just me and everyone else feels it's ok as it is now, just ignore my comment:)
We can always tune it later if it turns out people have different preferences.

tstrachota:
I'm still seeing the warning I mentioned here: https://github.com/theforeman/foreman/pull/5167#issuecomment-374557787

When this is fixed, I think we're good to merge.

amirfefer:
@tstrachota  could you be more specific? when this warning appears exactly?

Rohoover (commented):
@tstrachota 
I see what you are saying and I don't disagree. 

The edit button essentially takes you to a new page, so from a "rules" perspective, it would make sense to indicate that in the breadcrumb. I would just ask that we ensure that we continue with this principle on other pages. 

Also agreed that if we find this to be ineffective, we can opt to change later.

tstrachota:
> when this warning appears exactly?

Please disregard, it was issue on my side. Everything is ok after I updated npm packages. 
:+1:

app/controllers/audits_controller.rb
```diff
@@ -5,6 +5,7 @@ class AuditsController < ApplicationController
 
   def index
     @audits = resource_base_search_and_page.with_taxonomy_scope.preload(:user)
+    @host = resource_finder(Host.authorized(:view_hosts), params[:host_id]) if params[:host_id]
```
ohadlevy, line 8: please remove the back btn from the audits page 
image

app/controllers/config_reports_controller.rb
```diff
@@ -7,6 +7,7 @@ class ConfigReportsController < ApplicationController
   def index
     respond_to do |format|
       format.html do
+        @host = resource_finder(Host.authorized(:view_hosts), params[:host_id]) if params[:host_id]
```
ohadlevy, line 10: similar issue here (with back btn) 
image

but please note the difference, the breadcrumbs links points to: /config_reports while the back btn points to /config_reports?search=eventful+%3D+true

please remove the back btn and change the url for the index.

Also - does it make sense to make the breadcrumbs link to the host details page? (that could eliminate the host details btn as well).

ohadlevy:
@amirfefer this looks pretty ready for me, with the minor comment about back buttons, please have another look and hopefully we can merge this soon.

ohadlevy:
@amirfefer also, not sure if test failures are related or not...

tstrachota:
:+1: to merge from my side once the tests are green.

I think that to be 100% visually correct we'd need to add some more margin from top to center the breadcrumbs in their row. The patternfly design has them also vertically centered.

current state:
screenshot from 2018-03-27 15-13-25
vs. centered:
screenshot from 2018-03-27 15-14-10

But I'm good to merge it without that. It should be easy to tune later.

amirfefer:
[test hound]

amirfefer:
@ohadlevy  - I fixed tests failures

ohadlevy (approved):

ares:
:clap:

waldenraines:
👍 🎉

tstrachota:
Great news! :clap:

PR #5168, opened 2018-01-11
Sync up community templates

ares:


ares:
this is required in order to consider http://projects.theforeman.org/issues/13092 resolved

app/views/unattended/provisioning_templates/snippet/_kickstart_networking_setup.erb
```diff
@@ -37,8 +37,9 @@ kind: snippet
 <%-   next if !interface.managed? || (interface.subnet.nil? && interface.subnet6.nil?) -%>
 <%-   next if bonded_interfaces.include?(interface.identifier) -%>
 
+<%-   interface_identifier = @host.bond_interfaces.map { |i| i.identifier }.include?(interface.attached_to) ? interface.identifier : nil %>
```
tbrisker, line 40: wouldn't it be simpler (and potentially faster) to check `
@host.bond_interfaces.any?{|i| i.identifier == interface.attached_to}`?

iNecas: I guess this is not the place for review comments, as this is just a sync from community templates, where the proper reviews should take place

ares: +1 to what @inecas says, also `any?` is not allowed in safemode and you code is not equivalent, since we need the identifier, it could have been written used `find`/`detect` but that's also not allowed, I'd not be too concerned about optimization on this level, there are small number of interfaces

tbrisker (approved):

tbrisker:
[test katello]

tbrisker:
@ares Perhaps the commit message should be update to `refs #13092` so it gets attached to the issue, as this isn't a regular pre-release template sync?

ares:
good idea, since this is the only change, I updated the commit message

tbrisker:
Thanks @ares !

PR #5171, opened 2018-01-12
fixes #22258 - vmware: speed up vm listing

timogoebel:
This uses a single API call to vSphere to get the VM listing.

I timed the code that retrieves the data from vSphere and got an improvement from approx. 50 seconds to 1.5 seconds (❤️ ).

### DIFF
--- a/app/services/fog_extensions/vsphere/mini_server.rb
+++ b/app/services/fog_extensions/vsphere/mini_server.rb
@@ -3,16 +3,14 @@ module FogExtensions
     class MiniServer
       attr_reader :name, :identity, :cpus, :corespersocket, :memory, :state, :path
 
-      def initialize(raw, path = nil, uuid = nil)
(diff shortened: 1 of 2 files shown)


tbrisker:
I'm not familiar enough with vmware to review the code, but the improvement sounds awesome. Would it make sense pushing some of these changes into the fog-vsphere gem?

dLobatog (approved):
I don't have a huge VMWare setup but with around 30 vms the speedup is noticeable (actually the current slowdown is noticeable, now I don't even get to see the spinner). @timogoebel Thanks!

timogoebel:
> Would it make sense pushing some of these changes into the fog-vsphere gem?

I'm not sure, this is now pretty optimized for our needs. Every additional field causes a delay.

I want to look into speeding up host association, that actually uses the fog-vsphere code right now. Maybe there are some low hanging fruits.

PR #2828, opened 2015-10-15
refs #12147 - adding stub for rails4 compatibility

unorthodoxgeek:

PR #5176, opened 2018-01-15
Fixes #2828 - Report an error when trying to create/update ovirt resource

ik5:
On both create and update, there was no exception handler to captured
given exception. That resulted in having an exception screen instead of
an error message.
The fix capture all of the given exceptions, and report them as an error.

### DIFF
--- a/app/controllers/compute_resources_controller.rb
+++ b/app/controllers/compute_resources_controller.rb
@@ -36,6 +36,16 @@ class ComputeResourcesController < ApplicationController
       @compute_resource.valid?
       process_error
     end
+  rescue OVIRT::OvirtException => e
+    Foreman::Logging.exception("Error while creating ovirt resource", e)
+    process_error(
+      error_msg: _('Error while trying to create resource: %s') % e.message
+    )
+  rescue Fog::Errors::Error => e
+    Foreman::Logging.exception("Error while creating a resource", e)
+    process_error(
+      error_msg: _('Error while trying to create resource: %s') % e.message
+    )
   end
 
   def edit
@@ -70,6 +80,16 @@ class ComputeResourcesController < ApplicationController
     else
       process_error
     end
+  rescue OVIRT::OvirtException => e
+    Foreman::Logging.exception("Error while updating ovirt resource", e)
+    process_error(
+      error_msg: _('Error while trying to update resource: %s') % e.message
+    )
+  rescue Fog::Errors::Error => e
+    Foreman::Logging.exception("Error while updating resource", e)
+    process_error(
+      error_msg: _('Error while trying to update resource: %s') % e.message
+    )
   end
 
   def destroy
--- a/app/models/compute_resources/foreman/model/ovirt.rb
+++ b/app/models/compute_resources/foreman/model/ovirt.rb
@@ -157,7 +157,7 @@ module Foreman::Model
         when /404/
           errors[:url] << e.message
         when /302/
-          errors[:url] << 'HTTPS URL is required for API access'
+          errors[:url] << _('HTTPS URL is required for API access')
         when /401/
           errors[:user] << e.message
         else


tbrisker:
ok to test

app/controllers/compute_resources_controller.rb
```diff
@@ -36,6 +36,10 @@ def create
       @compute_resource.valid?
       process_error
     end
+  rescue StandardError => e
+    process_error(
+      error_msg: _('Error while trying to create resource: %s' % e.message),
```
adamruzicka: i think the `% e.message` should go outside the `_()` translation block. Also it would be probably nice to log the full backtrace of the exception somewhere in case we would need to debug it.

adamruzicka (commented):
Works as expected for me, left one comment inline.

app/controllers/compute_resources_controller.rb
```diff
@@ -63,6 +68,11 @@ def update
     else
       process_error
     end
+  rescue StandardError => e
+    logger.error "Error while updating ovirt resource: #{e.message}\n#{e.backtrace.join("\n")}"
```
tbrisker: This isn't necessarily an ovirt resource

app/controllers/compute_resources_controller.rb
```diff
@@ -36,6 +36,11 @@ def create
       @compute_resource.valid?
       process_error
     end
+  rescue StandardError => e
+    logger.error "Error while creating ovirt resource: #{e.message}\n#{e.backtrace.join("\n")}"
```
tbrisker: This isn't necessarily an ovirt resource

tbrisker (changes requested):
a couple of minor comments inline. Also, is there some more specific exception we expect to occur here that isn't `StandardError`? I would prefer rescuing only what we expect to be raised if possible. (e.g. this would catch exceptions which we already handle differently in application controller)

app/controllers/compute_resources_controller.rb
ik5: Right. when an exception comes back to the controller, I do not want it to return an exception screen, or 500 page, but to display an error message to the user.

If an exception was not handled prior to that location, at least allow the user to continue doing other stuff.

app/controllers/compute_resources_controller.rb
ik5: see my comment above

app/controllers/compute_resources_controller.rb
iNecas: @tbrisker was probably referring to the `ovirt` word in `Error while creating ovirt resource'`, it should be `Error while creating compute resource`

ik5: waiting for a PR of mine about unified Ovirt exception, and then I'll change this change accordingly.

boaz0: I agree with @tbrisker.
Lets keep the error message general and in the next PR where ovirt exception is thrown, add a more explicit error message saying it was `ovirt`.

WDYT?

tbrisker: I was indeed referring to the error message here. This is the compute resource controller which is used by all compute resources. Any error message should not contain a reference to ovirt, since that won't make sense, for example, when someone hits an error when trying to create a vmware resource.

ik5: sorry for that, you are right, and now added support for it.

app/controllers/compute_resources_controller.rb
```diff
@@ -36,6 +36,11 @@ def create
       @compute_resource.valid?
       process_error
     end
+  rescue Fog::Ovirt::Errors::OvirtEngineError => e
+    logger.error "Error while creating ovirt resource: #{e.message}\n#{e.backtrace.join("\n")}"
```
iNecas: There is a logger helper that can/should be used, see for example https://github.com/theforeman/foreman/blob/4a96211d33c261655c887bbb6c787e38d883200a/app/validators/ssh_key_validator.rb#L13.

iNecas: Also, there is ovirt mentioned here as well, while it should not (as well as in other places of this PR)

ik5: I'm missing here something in my understanding.
The original issue was, that when exception was raised, the actual 500 page would have shown rather then an error message the "foreman way".

My original attempt was to make it display the "foreman way", so if I create a validator here, how would it solve the problem of having the 500 error page?

Having said that, I do understand that the controller is agnostic to the worker that doing the actual task, and will solve this.

iNecas: No, I haven't meant validator: it was just an example of logging the exception with the `Logging.exception` API

iNecas: Btw. since we rescue `StandardError` at the other place, I wonder, if we shouldn't use this here as well compared to `Fog::Ovirt::Errors::OvirtEngineError`

ik5: The original code used `StandardError`, and moved it to use `Fog::Ovirt::Errors::OvirtEngineError` following the above discussion,

iNecas: I'm ok with that then. Still asking for using `Foreman::Logging.exception` use here

app/controllers/compute_resources_controller.rb
iNecas: We are rescuing from ` StandardError => e` which is not ovirt specific, therefore the message should not have `ovirt`, Am I reading this right @tbrisker?

tbrisker: exactly, since the controller is generic for all compute resources, if for example this is a libvirt resource it doesn't make sense to log an error referring to ovirt.

tbrisker: but shouldn't this also only catch ovirt errors instead of any generic ones like the other method?

ik5:
please note that there is a need to wait for https://github.com/fog/fog-ovirt/pull/8 to be merged (by @orrabin ) and bumped a version before merging this PR

orrabin:
@ik5 that will take time, I suggest to use the existing exception and later fix this as part of the fix we'll need to do in core for the new fog-ovirt version.

ik5:
following @orrabin comment, I moved my code to use `OVIRT::OvirtException` instead of `Fog::Ovirt::Errors::OvirtEngineError`. i will need to change it once the new ovirt version will be released.

app/controllers/compute_resources_controller.rb
```diff
@@ -70,6 +75,11 @@ def update
     else
       process_error
     end
+  rescue StandardError => e
+    Foreman::Logging.exception("Error while creating resource", e)
```
tbrisker: This is update, not create

app/controllers/compute_resources_controller.rb
```diff
@@ -70,6 +75,11 @@ def update
     else
       process_error
     end
+  rescue StandardError => e
```
tbrisker: Why do we catch `StandardError` here and `OVIRT::OvirtException` above? I expect that we should behave the same in both cases.
And if we want to catch a more generic error class, I would go with `Fog::Errors::Error` perhaps which fog-ovirt should inherit for its errors?

tbrisker (changes requested):

app/controllers/compute_resources_controller.rb
ik5: fixed

app/controllers/compute_resources_controller.rb
ik5: Okay, following our conversation f2f, I'ved also added a "fall-back" so both `Fog::Error::Error` and `OVIRT::OvirtException` will be captured.

app/controllers/compute_resources_controller.rb
```diff
@@ -70,6 +80,16 @@ def update
     else
       process_error
     end
+  rescue OVIRT::OvirtException => e
+    Foreman::Logging.exception("Error while updating ovirt resource", e)
+    process_error(
+      error_msg: _('Error while trying to updating resource: %s') % e.message
```
tbrisker: nitpick - this should be `...to update resource`

tbrisker (changes requested):
one tiny nitpick, other than that LGTM

app/controllers/compute_resources_controller.rb
ik5: fixed, thanks :)

tbrisker (approved):
LGTM, ready to merge once tests pass.

app/controllers/compute_resources_controller.rb
```diff
@@ -36,6 +36,16 @@ def create
       @compute_resource.valid?
       process_error
     end
+  rescue OVIRT::OvirtException => e
```
ohadlevy, line 39: should we add some sort of TODO here? as long term I dont expect to see rbovirt errors in our codebase? (or in other words, I would expect that after @orrabin finish the fog-ovirt rewrite, we will not know rbovirt in foreman code base at all).

tbrisker: @ohadlevy this isn't the only place we reference `OVIRT` module or exceptions. Ideally, we shouldn't care about the underlying implementation here and for other resource providers and push anything that does care into fog (I think we also refer to some RbVmomi stuff elsewhere).
Might be a good idea to open a separate issue to remove all references once `fog-ovirt` handles them. What do you think @orrabin ?

orrabin: That is the plan. Once we start using the version of `fog-ovirt` that wraps exceptions we will change all references to `OVIRT::OvirtException` in foreman. There isn't an issue for it yet, I'll open it.

tbrisker:
Thanks @ik5 !

PR #5177, opened 2018-01-15
Fixes #22272 - don't run CreateRssNotifications in test

iNecas:
During running tests, we still schedule the CreateRssNotifications job,
which can potentially collide with other tests, as well as seems to run
into some issues related to code reloading.

We should not do this in test environment in the first place.

### DIFF
--- a/config/initializers/rss_notifications.rb
+++ b/config/initializers/rss_notifications.rb
@@ -8,5 +8,5 @@
   end
 
   # Only create notifications if there isn't a scheduled job
-  CreateRssNotifications.perform_later unless scheduled_job.present?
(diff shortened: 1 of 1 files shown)


iNecas:
I'm running the foreman-tasks tests with this PR to check that it will actually help the issue in Jenkins: it seems it helped me on the issue locally http://ci.theforeman.org/job/test_plugin_pull_request/4508/

mmoll:
seems to fix the hangs, test failures unrelated.

ares (approved):

ares:
Thanks @iNecas, merging.

PR #5179, opened 2018-01-16
Refs #20325 - Fix pinning on ruby2ruby

ekohl:
2.4.0 is a known bad version, but that doesn't mean we explicitly need
2.3.2.


iNecas (approved):

iNecas:
Agreed. Thanks @ekohl , I assume you're ok with this change just in develop: no need to cherrypick to other branches, right?

ekohl:
So far just develop. Will raise another PR if I need a backport

tbrisker:
Looks like @witlessbird just released a new version of safemode(https://github.com/svenfuchs/safemode/issues/24#issuecomment-358148481), so i think we can get rid of the pinning.

PR #5183, opened 2018-01-17
Fixes #22285 - Raise error when strong params filters out params

tbrisker:
Currently we silently filter out parameters (or log in production),
causing unexpected results when passing an invalid param (either
incorrect name or incorrect type). This can lead to unexpected results,
since the user, seeing no error, assumes the request was successful when
in fact some of the parameters were filtered out.

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -44,6 +44,13 @@ module Api
       not_found error.message
     end
 
+    rescue_from ActionController::UnpermittedParameters do |error|
+      logger.info "#{error.message} (#{error.class})"
+      message = n_("Invalid parameter: %s. ", "Invalid parameters: %s. ", error.params.length) % error.params.to_sentence
+      message << _('Please verify that the parameter name is valid and the values are the correct type.')
+      render_error 'param_error', :status => :bad_request, :locals => { :exception => error, :message => message }
+    end
+
     rescue_from Foreman::MaintenanceException, :with => :service_unavailable
 
     def get_resource(message = "Couldn't find resource")
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -9,6 +9,7 @@ class ApplicationController < ActionController::Base
   rescue_from ScopedSearch::QueryNotSupported, :with => :invalid_search_query
   rescue_from ActiveRecord::RecordNotFound, :with => :not_found
   rescue_from ProxyAPI::ProxyException, :with => :smart_proxy_exception
+  rescue_from ActionController::UnpermittedParameters, :with => :invalid_parameter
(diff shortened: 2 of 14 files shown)


tbrisker:
This currently Breaks All Things. setting to WoC to figure out.

tbrisker:
Previously had 636 failing tests, still WIP to fix all broken tests.

test/controllers/api/v2/hosts_controller_test.rb
```diff
@@ -314,7 +314,7 @@ def expect_attribute_modifier(modifier_class, args)
 
   test "should update hostgroup_id of host" do
     @host = FactoryBot.create(:host, basic_attrs_with_hg)
-    put :update, params: { :id => @host.to_param, :hostgroup_id => Hostgroup.last.id }
```
tbrisker: The last hostgroup doesn't necessarily have an environment, which is required, and reset if the hg changes.

iNecas:
Could we wait with this change until 1.17 is out + a week or two: many plugin authors haven't had green tests for a while, so would be good to give them some time to actually do other things than catching up with upstream breakages?

tbrisker:
Down to 237 :)

tbrisker:
@iNecas I expect fixing all tests and reviewing it will take several days at best. I don't mind waiting a bit but it will be painful to keep up to date over a long period. Besides the impact on users and developers I outlined on [discourse](https://community.theforeman.org/t/strong-parameters-is-borked/8070), this also exposes multiple tests which we thought were passing when in fact they were failing silently - I am still digging into it, but there may be other bugs discovered due to these tests. In other words - the sooner we fix this properly the better. 
That being said, I think for most plugins fixing this would be a matter of adding a few parameters to the filters already in place; any broken tests discovered due to this are actually broken already without us knowing it. Once I finish doing the core part of it, I'll try looking into some of the plugins.

thomasmckay:
@waldenraines FYI - This will require significant effort to fix and test the katello UI since when an item is edited its json is often sent back to server with extra fields. Product edit, for example, gets "An error occurred saving the Product: found unpermitted parameters: :cp_id, :created_at, :updated_at". This will be common on all of the pages, I think.

mmoll:
@tbrisker please rebase :)

tbrisker:
rebased, now lets see how many new test have broken in the last 2 months...

mmoll:
@tbrisker this needs a rebase :)

I think this should go in soon, especially now that 1.18 is branched.

mmoll:
@tbrisker ping?

tbrisker:
@mmoll last time i tried to get this mergeable i figured out there's some refactoring to the host[group] form that's needed to fix one of the issues, GH-5458 is WIP to fix that. Once I finish with that this one is next in line.

ares:
:cry:

PR #5184, opened 2018-01-17
Fixes #21312 - components for formatting dates

tstrachota:
Contains following changes:
- adds React components for rendering dates
- locales for `react-intl` as well as the fallback `intl` package are loaded lazily
- enables multiple webpack entry points per plugin (necessary to get this change into katello)
- adds HOC for i18n that waits until locales are loaded

Depending on your setup, you might need to set `ASSET_PATH` env variable to get the dynamic imports (lazy loading) working in development. It's required when running over https. Eg:
```
ASSET_PATH='https://myforeman.org:3808/webpack/'
```


app/registries/foreman/webpack_assets.rb
```diff
@@ -1,7 +1,18 @@
...
-      plugin_name.to_s.gsub(/core$/, '').gsub(/-|_|#{remove_bundle_name_plugins}/,'')
+    # Create name for an entry
+    # E.g.
+    #   index.js -> katello.js
+    #   common_index.js -> katello_common.js
+    def bundle_name(plugin_name, entry_path = '')
+      index_prefix = File.basename(entry_path).gsub(/[_]?index[.]js$/, '')
+      bundle = plugin_name.to_s.gsub(/core$/, '').gsub(/-|_|#{remove_bundle_name_plugins}/,'')
+
+      if index_prefix == ''
```
ohadlevy: nitpick: .blank?

config/webpack.config.js
```diff
@@ -18,6 +19,13 @@ var production =
   process.env.RAILS_ENV === 'production' ||
   process.env.NODE_ENV === 'production';
 
+var publicPath;
+if (production) {
+  publicPath = process.env.ASSET_PATH || '/webpack/';
+} else {
+  publicPath = process.env.ASSET_PATH || `http://${process.env.HOSTNAME}:${devServerPort}/webpack/`;
```
ohadlevy: do you really need the entire url? I would prefer to avoid hardcoding the path if we URL scheme

config/webpack.config.js
```diff
@@ -41,9 +54,10 @@ var config = {
 
     // must match config.webpack.output_dir
     path: path.join(__dirname, '..', 'public', 'webpack'),
-    publicPath: '/webpack/',
+    publicPath: publicPath,
 
-    filename: production ? '[name]-[chunkhash].js' : '[name].js'
+    filename: production ? '[name]-[chunkhash].js' : '[name].js',
+    chunkFilename: production ? '[name]-[chunkhash].js' : '[name].js'
```
ohadlevy: what does this do?

webpack/assets/javascripts/compute_resource/ovirt.js
```diff
@@ -28,7 +28,7 @@ export function templateSelected(item) {
         $('#storage_volumes .children_fields >.fields').remove();
         $.each(result.volumes, function () {
           // Change variable name because 'interface' is a reserved keyword.
-          this.disk_interface = this['interface'];
+          this.disk_interface = this.interface;
```
ohadlevy: not related?

webpack/assets/javascripts/react_app/common/helpers.js
```diff
@@ -6,4 +6,7 @@ export default {
     });
   },
   noop: Function.prototype, // empty function
+  getDisplayName(Component) {
+    return Component.displayName || Component.name || 'Component';
```
ohadlevy: isnt this the default (just with an Unknown string instead of Component?

ohadlevy:
@sharvit @amirfefer @waldenraines @danseethaler can you guys review please?

config/webpack.config.js
tstrachota: Unfortunately we really need the entire url. We have to pass the dev server port. Without that webpack fetches dynamic imports from wrong url.

config/webpack.config.js
tstrachota: https://webpack.js.org/configuration/output/#output-chunkfilename
Name format for non-entry chunks created by webpack.

.babelrc
```diff
@@ -1,9 +1,19 @@
 {
-  "presets": ["env", "react"],
+  "presets": [
+    "env",
+    ["es2015", {"modules": false}],
```
danseethaler: I'm pretty sure you can pass the `{"modules": false}` into the `env` preset. Is there a reason we're targeting `2015` only here?

package.json
```diff
@@ -10,14 +10,17 @@
...
     "babel-jest": "^15.0.0",
     "babel-loader": "^7.1.1",
     "babel-plugin-lodash": "^3.3.2",
+    "babel-plugin-dynamic-import-node": "^1.2.0",
+    "babel-plugin-syntax-dynamic-import": "^6.18.0",
     "babel-plugin-transform-class-properties": "^6.24.1",
     "babel-plugin-transform-object-assign": "^6.8.0",
     "babel-plugin-transform-object-rest-spread": "^6.8.0",
     "babel-preset-env": "^1.6.1",
+    "babel-preset-es2015": "^6.24.1",
```
danseethaler: Same question as above about needing this.

danseethaler (commented):
Looks good overall - just one question with the babel setup.

config/webpack.config.js
sharvit: Seems my `process.env.HOSTNAME` is undefined.

config/webpack.config.js
```diff
@@ -27,6 +35,11 @@ var resolveModules = [  path.join(__dirname, '..', 'webpack'),
                         'node_modules/',
                      ].concat(pluginUtils.pluginNodeModules(plugins));
 
+// Detect supported locales from folders in ./locale/
+const localeDir = path.join(__dirname, '..', 'locale')
+const localeSubdirs = [ ...new Set(fs.readdirSync(localeDir).filter(f => fs.statSync(path.join(localeDir, f)).isDirectory())) ];
+const supportedLocales = localeSubdirs.map(d => d.split('_')[0]).join('|')
```
sharvit: Can you refactor those lines so it will be easy to understand?

config/webpack.config.js
sharvit: How should it work in production? Do we need the `ASSET_PATH` in production?

config/webpack.config.js
sharvit: I am getting all of those files into my public/webpack when building production:
```
0-03661867bb957b74fba9.js
1-1d5ccaf80730fcbd15a5.js
2-531ce93840f13a2bb530.js
3-0fe5d3ece4ad57ea7eb8.js
n-.....................js
28-.....................js
```

shouldn't it be `en-[hash].js` / `pt_BR-[hash].js`?

config/webpack.config.js
sharvit: Should consider adding this variable to [Procfile](https://github.com/theforeman/foreman/blob/develop/Procfile) and [script/foreman-start-dev](https://github.com/theforeman/foreman/blob/develop/script/foreman-start-dev)?

script/plugin_webpack_directories.rb
```diff
@@ -26,14 +27,18 @@
...
-  # skip the plugin if its path is already included
-  next if config[:paths].include?(path)
-  if File.exist?(entry)
-    config[:entries][bundle_name dep.name] = entry
-    config[:paths] << path
+
+  Dir["#{path}/*index.js"].each do |entry|
+    # some plugins share the same base directory (tasks-core and tasks, REX etc)
+    # skip the plugin if its path is already included
+    next if entry_paths.include?(entry)
```
sharvit, line 66: Because this is the only place you are using `entry_paths`,
can you remove this var and use `config[:paths]` instead?

webpack/assets/javascripts/react_app/components/componentRegistry.js
```diff
@@ -30,27 +34,42 @@ const componentRegistry = {
...
 
   getComponent(name) {
+    if (!this.registry[name]) {
+      throw new Error(`Component not found: ${name} among ${this.registeredComponents()}`);
+    }
+
     return this.registry[name];
   },
 
+  wrapperFactory() {
```
sharvit, line 57: Really need this function?

webpack/assets/javascripts/react_app/components/componentRegistry.js
```diff
@@ -30,27 +34,42 @@ const componentRegistry = {
...
 
-  markup(name, data, store) {
-    const currentComponent = this.getComponent(name);
+  defaultWrapper(component, data = null, store = null) {
+    const factory = new WrapperFactory();
+
+    factory.with('i18n');
 
-    if (!currentComponent) {
-      throw new Error(`Component not found:  ${name} among ${this.registeredComponents()}`);
```
sharvit: No need to raise this error anymore?

webpack/assets/javascripts/react_app/components/componentRegistry.js
```diff
@@ -30,27 +34,42 @@ const componentRegistry = {
...
+      factory.with('data', data);
+    }
+    return factory.wrapper;
+  },
+
+  markup(name, { data = null, store = null, wrapper = null }) {
+    const currentComponent = this.getComponent(name);
+    const componentWrapper = wrapper || this.defaultWrapper(currentComponent, data, store);
+
+    const ComponentName = componentWrapper(currentComponent.type);
```
sharvit: Can you rename to `WrappedComponent`?

webpack/assets/javascripts/react_app/components/wrapperFactory.js
```diff
@@ -0,0 +1,89 @@
+import React from 'react';
+import { Provider } from 'react-redux';
+import { i18nProviderWrapper } from '../common/i18nProviderWrapper';
+import helpers from '../common/helpers';
+
+function storeProviderWrapper(store) {
```
sharvit: Can you use es6 syntax for the functions? `const foo = () => 'bar';`

webpack/assets/javascripts/react_app/components/wrapperFactory.js
```diff
@@ -0,0 +1,89 @@
...
+export class WrapperFactory {
+  constructor() {
+    this.wrapper = component => component;
+  }
+
+  with(name, ...params) {
+    const previousWrapper = this.wrapper;
+
+    this.wrapper = component =>
+      wrapperRegistry.getWrapper(name)(...params)(previousWrapper(component));
```
sharvit: can you refactor this function to be more easy to understand?
I afraid doing `()()()` can confuse readers.

Maybe:
```js
this.wrapper = component => {
  const PreviousWrappedComponent = previousWrapper(component);
  const wrapper = wrapperRegistry.getWrapper(name)(...params);

  return wrapper(PreviousWrappedComponent);
};
```

config/webpack.config.js
sharvit: Does it mean plugins are also going to build those files?

webpack/assets/javascripts/react_app/common/i18n.js
```diff
@@ -0,0 +1,32 @@
...
+
+async function init(langCode) {
+  await fetchIntl(langCode);
+  addLocaleData(await import(`react-intl/locale-data/${langCode}`));
+  loaded = true;
+}
+
+export const ready = init(locale);
+/* eslint-enable import/no-dynamic-require, import/no-mutable-exports */
+
```
sharvit: What about using a class here so the API will be more clear?

```js
class I18n {
  constructor() {
    const langAttr = document.getElementsByTagName('html')[0].getAttribute('lang') || 'en';
    const timezoneAttr = document.getElementsByTagName('html')[0]
      .getAttribute('data-timezone') || 'UTC';

    this.locale = langAttr.split('-')[0];
    this.timezone = global.Intl ? timezoneAttr : 'UTC';
    this.loaded = false;
    this.ready = new Promise((resolve, reject) => {
      this.resolveReady = resolve;
    });
  }

  async fetchIntl(langCode) {
    if (!global.Intl) {
      global.Intl = await import('intl');
      await import(`intl/locale-data/jsonp/${langCode}`);
    }
  }

  async init(langCode) {
    await this.fetchIntl(langCode);
    const reactIntlLocalData = await import(`react-intl/locale-data/${langCode}`);
    
    addLocaleData(reactIntlLocalData);
    
    this.loaded = true;
    this.resolveReady();
  }
}

const i18n = new I18n();

i18n.init();

export default i18n;
```

Maybe even seperate the class and the instance into separate files

webpack/assets/javascripts/react_app/common/i18nProviderWrapper.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { IntlProvider } from 'react-intl';
+import { locale, ready as i18nReady, loaded as i18nLoaded } from './i18n';
+import helpers from './helpers';
+
+export function i18nProviderWrapper(initialNow = undefined) {
```
sharvit: can use es6 syntax for the function?
Shouldn't the default value of `initialNow` should be `undefined` anyway?

webpack/assets/javascripts/react_app/common/i18nProviderWrapper.js
```diff
@@ -0,0 +1,35 @@
...
+  return (WrappedComponent) => {
+    const wrappedName = helpers.getDisplayName(WrappedComponent);
+
+    class I18nProviderWrapper extends React.Component {
+      constructor(props) {
+        super(props);
+        if (!i18nLoaded) {
+          i18nReady.then(() => {
+            this.forceUpdate();
```
sharvit: Would it make more sense to use an inner store instead `forceUpdate`?

webpack/assets/javascripts/react_app/components/common/dates/Date.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { FormattedDate, intlShape } from 'react-intl';
+import { timezone } from '../../../common/i18n';
+
+class Date extends React.Component {
```
sharvit: Doesn't it interfere with the native Date?

webpack/assets/javascripts/react_app/components/common/dates/Date.js
```diff
@@ -0,0 +1,39 @@
...
+  render() {
+    if (this.props.data.date) {
```
sharvit: Can do:
```js
const { date, default } = this.props.data;

if (date) {
  return ...;
}

return <span>{default}</span>;
```

webpack/assets/javascripts/react_app/components/componentRegistry.js
```diff
@@ -1,12 +1,16 @@
...
-
 import PieChart from './common/charts/PieChart/';
 import StatisticsChartsList from './statistics/StatisticsChartsList';
 import PowerStatus from './hosts/powerStatus/';
 import NotificationContainer from './notifications/';
 import ToastsList from './toastNotifications/';
+import RelativeDateTime from './common/dates/RelativeDateTime';
+import LongDateTime from './common/dates/LongDateTime';
+import ShortDateTime from './common/dates/ShortDateTime';
+import Date from './common/dates/Date';
```
sharvit: Afraid `Date` interfere with the native `Date`

sharvit (changes requested):
Nice work @tstrachota :+1: 
Good job implementing the async importers!

It's a significant and blessed change, so I hope it will merge soon even though I have many comments.

sharvit:
Can you add the `intl` and `react-intl` to [config/webpack.vendor.js](https://github.com/theforeman/foreman/blob/develop/config/webpack.vendor.js)?

config/webpack.config.js
tstrachota: I'll try to tune the Procfile. Production will be served from `/webpack/` as before. That branch is unchanged.

script/plugin_webpack_directories.rb
tstrachota: I can't because `config[:paths]` aggregate paths to directories that contain entries while `entry_paths` contain paths to the actual entries.

e.g.
```
config[:paths]
 => [
  './katello/webpack/'
]
```
vs.
```
entry_paths
 => [
  './katello/webpack/index.js',
  './katello/webpack/common_index.js'
]
```

webpack/assets/javascripts/react_app/components/componentRegistry.js
tstrachota: It's there for plugins to give them access to a default factory setup (Katello uses it).

webpack/assets/javascripts/react_app/components/componentRegistry.js
tstrachota: It was just moved to a different place. See: https://github.com/theforeman/foreman/pull/5184/files#diff-3bbe3fd702bd07a1292f5987335ecb37R38

webpack/assets/javascripts/react_app/components/componentRegistry.js
tstrachota: That's true. One has to import it under a different name if you plan to use them together. I couldn't find any better name. Suggestions?

app/registries/foreman/webpack_assets.rb
tstrachota: We're using this helper also in `plugin_webpack_directories.rb` where we don't have active support included.

app/assets/javascripts/application.js
```diff
@@ -91,8 +91,11 @@ function onContentLoad(){
 
   password_caps_lock_hint();
 
-  var tz = jstz.determine();
-  $.cookie('timezone', tz.name(), { path: '/', secure: location.protocol === 'https:' });
+  tfm.i18n.ready.then(() => {
+    var tz = jstz.determine();
+    $.cookie('timezone', tz.name(), { path: '/', secure: location.protocol === 'https:' });
```
priley86, line 94: hey @tstrachota ... maybe you can have some `Cookies` with this PR? 🍪 

https://github.com/js-cookie/js-cookie

wdyt?

priley86: maybe in the future but not now... i dunno...

sharvit: Yes @priley86 we should actually do that, thanks for the advice 👍 

I was also looking into that repository:
https://github.com/reactivestack/cookies

I think the `js-cookie` is better for us, for now, i still don't see us using `CookieProvider` or `withCookie` wrapper.

We actually trying to get ride of jquery which is a long process and migrating the cookies is a good step.

I open an issue and i hope to do it soon:
http://projects.theforeman.org/issues/22506

webpack/assets/javascripts/react_app/components/common/dates/Date.js
priley86: +1

app/assets/javascripts/application.js
priley86: I am fine w/ [react-cookie](https://github.com/reactivestack/cookies/tree/master/packages/react-cookie) too... and yes, very much prefer that API down the road if it becomes an option. Universal cookies are even more tasty! They could stream via `react-dom/server` within `apollo-server` in an isomorphic way, you know? Some kind of strange `renderToNodeStream` voodoo...haha

overall, i like the approach here w/ `react-intl` and utilizing the *componentRegistry*. I noticed Twitter doing something very similar [yesterday](http://nicolasgallagher.com/redux-modules-and-code-splitting/). I'm happy to move off of jquery as you have time...

webpack/assets/javascripts/react_app/common/i18nProviderWrapperFactory.js
```diff
@@ -0,0 +1,38 @@
...
+          });
+        }
+      }
+
+      render() {
+        if (!this.state.i18nLoaded) {
+          return <span />;
+        }
+        return (
+          <IntlProvider locale={i18n.locale} initialNow={initialNow}>
```
priley86: I think you can render messages on the `IntlProvider` as well...

https://egghead.io/lessons/react-convert-a-hard-coded-string-using-react-intl-formattedmessage

webpack/assets/javascripts/react_app/components/common/dates/Date.js
priley86: What do you think of the name `IsoDate` instead?

https://www.w3.org/QA/Tips/iso-date

webpack/assets/javascripts/react_app/components/common/dates/LongDateTime.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { FormattedDate, intlShape } from 'react-intl';
+import { timezone } from '../../../common/i18n';
+
+class LongDateTime extends React.Component {
```
priley86: I think all of these can be converted to stateless functional components.... at least my linter tells me that ;)

example: https://github.com/priley86/miq_v2v_ui_plugin/commit/9db79b62ae7c26daf5db551eac325c625f3d209c

webpack/assets/javascripts/react_app/components/componentRegistry.js
priley86: Yea... WrapperFactory just seems overcomplicated to me... but I guess keep it if if works for you... I kind of prefer a single wrapper function.

webpack/assets/javascripts/react_app/components/common/dates/ShortDateTime.js
```diff
@@ -0,0 +1,42 @@
...
+}
+
+ShortDateTime.contextTypes = {
+  intl: intlShape,
+};
+
+ShortDateTime.propTypes = {
+  data: PropTypes.shape({
+    date: PropTypes.any,
+    defaultValue: PropTypes.string,
```
priley86: you are missing `seconds` in propTypes here

app/assets/javascripts/application.js
tstrachota: I'm ok with migrating from jquery cookies to some other solution in future, but it's out of the scope of this PR.

config/webpack.config.js
tstrachota: The dynamic chunks didn't have `webpackChunkName` configured. I'll fix it.

webpack/assets/javascripts/react_app/common/helpers.js
tstrachota: I'm not sure what are the defaults. I saw `displayName` being undefined at some places. This approach is a React.js convention for HOCs: https://reactjs.org/docs/higher-order-components.html#convention-wrap-the-display-name-for-easy-debugging

webpack/assets/javascripts/react_app/common/i18nProviderWrapperFactory.js
tstrachota: Yeah, I know about it. But we don't use react-intl for messages. It's one of the options for future, but we haven't decided what will be our approach yet. We can easily add it later.

webpack/assets/javascripts/react_app/components/common/dates/Date.js
tstrachota: I like `IsoDate`, I'll change the names.

webpack/assets/javascripts/react_app/components/common/dates/ShortDateTime.js
tstrachota: Updated

tstrachota:
[test]

waldenraines:
Where are we with this?  This is blocking https://github.com/Katello/katello/pull/7001 which is in turn blocking some katello work which relies on the infrastructure in that PR.

waldenraines:
@tstrachota do you mind rebasing this please?

tstrachota:
@waldenraines rebased

waldenraines:
@ohadlevy what else is needed here?

tstrachota:
[test foreman]

ohadlevy:
@waldenraines I didnt test it yet (lack of time) and the fact that @sharvit still didn't approve it?

.babelrc
```diff
@@ -1,9 +1,19 @@
 {
-  "presets": ["env", "react"],
+  "presets": [
+    "env",
+    "es2015",
```
sharvit: Does it really necessary? Shouldn't the `env` provide the `es-latest-stable`?

sharvit: When running tests, I am getting a bunch of those errors:
```
ERROR: Error: Couldn't find preset "es2015" relative to directory
```

sharvit: Same when running dev-server:
```
Module build failed: Error: Couldn't find preset "es2015"
```

sharvit (changes requested):
@tstrachota I couldn't run it locally, can you check it please?

tstrachota:
[test foreman]

tstrachota:
I rebased and fixed the tests with @sharvit's help. It turned out to be combination of problems: using arrow syntax on one place in the asset pipeline processed scripts and PhantomJS not loading i18n data for Intl.

At the moment there are some intermittent ruby test failures, but they are unrelated (`"ActiveRecord::RecordInvalid: Validation failed: Controller dynflow_actions is not a valid controller"`).

Broken js tests were the main blocker. The PR should be now ready for final rounds of reviews.

tstrachota:
[test katello]

tstrachota:
@ohadlevy @sharvit may I ask you to re-review?

config/webpack.config.js
```diff
@@ -53,6 +67,21 @@ module.exports = env => {
     pluginEntries
   );
 
+  var publicPath;
+  if (production) {
+    publicPath = process.env.ASSET_PATH || '/webpack/';
+  } else {
+    publicPath = process.env.ASSET_PATH || `http://${devServerHostname}:${devServerPort}/webpack/`;
```
sharvit: Is there a way you can automatically chose http/s based on the environment/configuration?

config/webpack.config.js
```diff
@@ -53,6 +67,21 @@ module.exports = env => {
...
+  }
+
+  // Detect supported locales from folders in ./locale/
+  const localeDir = path.join(__dirname, '..', 'locale');
+  // Find all directories in ./locale/*
+  const localeSubdirs = [ ...new Set(fs.readdirSync(localeDir).filter(f => fs.statSync(path.join(localeDir, f)).isDirectory())) ];
```
sharvit: Could you make this line easier to understand? destruct to multiple lines?

sharvit: Since only `supportedLocalesRE` is relevant for the config, does it make sense to wrap this code in a function that return only the `supportedLocalesRE`?

package.json
```diff
@@ -98,7 +102,7 @@
   "scripts": {
     "lint": "./node_modules/.bin/eslint -c .eslintrc webpack/ script/ || exit 0",
     "postlint": "node ./script/npm_lint_plugins.js",
-    "test": "node node_modules/.bin/jest",
+    "test": "node node_modules/.bin/jest --no-cache",
```
sharvit, line 118: Why can't we use cache anymore? Does it actually make the test performance worse?

sharvit (changes requested):
Works well in development but I had to set the `ASSET_PATH` because i'm working with `https`.

In production i'm getting this error, not sure why:
```
Request URL:https://centos7-devel.sharvits-fedora-book.example.com/javascripts/locale/en/app.js
Request Method:GET
Status Code:404 Not Found
```

sharvit:
@tstrachota Sorry, I just realize this error is not related to this PR and it actually exits in develop

> In production i'm getting this error, not sure why:
> ```
> Request URL:https://centos7-devel.sharvits-fedora-book.example.com/javascripts/locale/en/app.js
> Request Method:GET
> Status Code:404 Not Found
> ```

config/webpack.config.js
tstrachota: I'd love to find some. So far I didn't find any easy way. I was thinking about loading foreman's config but I'm afraid that would take too long to load.

package.json
tstrachota: I had some troubles with the cache not being updated after my changes and that caused me some headaches in development even though the tests were actually correct. That's why I decided to disable it by default.
I didn't observe any significant decrease of performance but I'm fine with removing it if you don't like it.

tstrachota:
Rebased and updated. I added `https` detection.
For some reason webpack started ignoring the dynamic imports after the rebase. I'll need to investigate what happened.

tstrachota:
Fixed, it was my mistake, the PR should be working normally now.

sharvit (commented):
Nice work @tstrachota, I made some testings and it works well :+1: 

The only thing I believe missing is some documentations about the I18n. storybook?

+Some tests are failing...

config/webpack.config.js
```diff
@@ -11,10 +11,30 @@ var LodashModuleReplacementPlugin = require('lodash-webpack-plugin');
...
+  // Extract extract languages from the language tags (strip off dialects)
+  return [ ...new Set(supportedLocales().map(d => d.split('_')[0]))];
+}
 
 module.exports = env => {
   // must match config.webpack.dev_server.port
-  var devServerPort = 3808;
+  const devServerPort = 3808;
+  const devServerBindHost = process.env.HOSTNAME || os.hostname();
+  const devServerProtocol = process.argv.includes('--https') ? 'https' : 'http';
```
amirfefer: awesome :+1:

ekohl:
Could you split the infra changes (webpack, script) into a separate commit? I would like to review that but it's hard to see what's what.

tstrachota:
@ekohl rebased and split. The infrastructure part is in: https://github.com/theforeman/foreman/pull/5184/commits/9dc773e8f983644bef7841dfd0dc36d1b061140b
@sharvit thanks for testing it. The test failures seemed unrelated. I hope they will go after the rebase.

sharvit:
@tstrachota I think it is actually your tests who failing:
```bash
Cannot find module 'enzyme-adapter-react-16' from 'IsoDate.test.js'
```

It might be because we are using `enzyme-adapter-react-16.3` now so you should just update you import statement:
https://github.com/theforeman/foreman/blob/develop/package.json#L30
https://github.com/theforeman/foreman/pull/5184/files#diff-fdd7cf2437d0f36a2dcfd421a1e20c4dR3

**EDIT**
I think you don't actually need to import and configure enzyme. We do it automatically here:
https://github.com/theforeman/foreman/blob/develop/webpack/test_setup.js

waldenraines:
@tstrachota any updates on this?

tstrachota:
I rebased and merged with latest changes in develop. I found one issue with intl-relativeformat that doesn't correctly calculate relative dates when a timezone is used. I created a workaround to prevent from blocking this PR. There's are also a new story with the dates component.

All changes as separate commits.

tstrachota:
Hmm, the test failure seems to be related to the relative date workaround. I need to take a look. It shouldn't be blocking the next round of review.

tstrachota:
Updated, ready for next round of review.

waldenraines:
[test katello]

webpack/assets/javascripts/react_app/common/I18n.js
```diff
@@ -1,6 +1,40 @@
...
+const runningInPhantomJS = () => (window._phantom !== undefined);
+
+class IntlLoader {
+  constructor(locale, timezone) {
+    this.fallbackIntl = !global.Intl || runningInPhantomJS();
+
+    [this.locale] = locale.split('-');
+    this.timezone = this.fallbackIntl ? 'UTC' : timezone;
+    this.loaded = false;
+    this.ready = this.init();
```
xprazak2, line 13: I am curious if both `loaded` and `ready` are needed. Please disregard if I missed something, but `loaded` is initially set to `false` and flipped when the async actions finish, while `ready` contains the promise returned by calling `init()`. 

`loaded` is used only in `constructor` of `I18nProviderWrapperFactory`, where we wait for the `ready` promise to resolve if `!loaded`. So I was thinking you could `return true` from `init()` and always wait for the promise to resolve - if the `init()` call already finished, then promise resolves immediately and if not, we wait for it just like before.

This [diff](https://github.com/xprazak2/foreman/commit/cbcdbb004ab57e29cae64a0f3682d86f9d88902f) seemed to work...

webpack/stories/docs/plugins.md
```diff
@@ -41,7 +41,18 @@ Until this is fixed in Foreman core the workaround is using [mocks](https://gith
 
 ### Using components outside of webpack
 
-Using components from core outside of webpack processed code it's currently **not possible**. In future the `componentRegistry` should be available in `Window.tfm` object, which will unlock the path and enable for example using React components in AngularJS in Katello.
+The component registry is available in `Window.tfm.componentRegistry`. That gives you access to the components even from js code that isn't processed by webpack.
```
xprazak2, line 44: I just wonder why that would be needed. I can only think of katello using certain components in Angular pages, is that it?

xprazak2:
Looking great! I just left 2 comments.

webpack/stories/docs/plugins.md
tstrachota: Exactly! :) 
(https://github.com/Katello/katello/pull/7001/files#diff-5bcfdae0ace18ad3702fbb46115e768d)

webpack/assets/javascripts/react_app/common/I18n.js
tstrachota: I think it's leftover from earlier stages of development. What you write makes sense. I'll test your patch and merge it into my branch if it works. Thanks!

tstrachota:
@xprazak2 updated, I removed the flag 'loaded'

webpack/assets/javascripts/react_app/common/__mocks__/I18n.js
```diff
@@ -7,11 +7,19 @@ export const ngettext = createTranslateMock();
 
 export const sprintf = createTranslateMock();
 
+export const intl = {
+  loaded: true,
+  ready: Promise.resolve(),
```
xprazak2: Maybe removing `loaded` and `Promise.resolve(true)` could fix the failing snapshot tests...

tstrachota:
Rebased & updated.

tstrachota:
[test]

xprazak2 (dismissed):

ares:
[test upgrade] interesting, the upgrade test failed on migration which  I think is not related, let's see, foreman tests are green, code climate does not have imho reasonable suggestions

```
== 20181116104823 FixedIgnoreTaxonomiesForAuditFilters: migrating =============
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:
undefined method `locked?' for nil:NilClass
stable/ruby/2.3/db/migrate/20181116104823_fixed_ignore_taxonomies_for_audit_filters.rb:5:in `block in up'
```

@xprazak2 any other comments or do you plan to merge once we have upgrade tests green?

xprazak2:
[test upgrade]

xprazak2:
I have no more comments, I triggered the upgrade tests again.

ares:
upgrade failures are same as in other PR, Mr. Jenkins approves, @xprazak2 you have the honor :-)

xprazak2:
Needs an ack from packaging.

ares:
@ekohl thoughts? we had this merged before and then it was reverted, but maybe back then packaging was not acking everything

waldenraines:
@ekohl thoughts? :smile:

config/webpack.config.js
```diff
@@ -53,6 +75,15 @@ module.exports = env => {
...
   );
 
+  var publicPath;
+  if (production) {
+    publicPath = process.env.ASSET_PATH || '/webpack/';
+  } else {
+    publicPath = process.env.ASSET_PATH || `${devServerProtocol}://${devServerBindHost}:${devServerPort}/webpack/`;
+  }
+
+  const supportedLanguagesRE = supportedLanguages().join('|');
```
ekohl: Would it make sense to apply `new RegExp(`/(${supportedLanguagesRE})$`)` here instead of doing it twice below?

script/plugin_webpack_directories.rb
```diff
@@ -16,29 +16,61 @@
...
+end
+
 config = { entries: {}, paths: [], plugins: {} }
+entry_paths = []
 plugin_name_regexp = /foreman*|katello*/
 specs.each do |dep|
   # skip other rails engines that are not plugins
   # TODO: Consider using the plugin registration api?
   next unless dep.name =~ plugin_name_regexp
-  next if dep.name.include?('_core')
```
ekohl, line 25: Where did this go? Perhaps it's no longer needed now that Ansible dropped its _core plugin, but not 100% sure. Or is it replaced by using `entry_paths`?

script/plugin_webpack_directories.rb
```diff
@@ -16,29 +16,61 @@
...
+
   package_json = JSON.parse(File.read("#{dep.to_spec.full_gem_path}/package.json"))
-  main = package_json['main'] || 'index.js'
-  entry = "#{path}/#{main}"
-  if File.exist?(entry)
-    config[:entries][dep.name] = entry
-    config[:paths] << path
-    config[:plugins][dep.name] = {root: dep.to_spec.full_gem_path, entry: entry}
+
+  entries = Dir["#{path}/*index.js"]
```
ekohl, line 60: Is this part of the spec? The previous implementation was based on https://docs.npmjs.com/files/package.json#main that I interpreted as you should use main if present.

script/plugin_webpack_directories.rb
tstrachota: `next if entry_paths.include?(entry)` on line 65 should handle that case. I consider it better than explicitly excluding `*_core`.

script/plugin_webpack_directories.rb
tstrachota: This extends the default npm behavior. The main entry point is still used if it's available (see the next line). On top of that all `*index,js` files are added which allows adding more independent entry points. I used that in Katello where I needed to load some npm dependencies into angular pages (see [common_index.js](https://github.com/Katello/katello/pull/7001/files#diff-106a751f2719195939e27fd323a88a12) in Katello).

config/webpack.config.js
tstrachota: That makes sense. I'll change it.

script/plugin_webpack_directories.rb
ekohl: I agree hardcoding a name is ugly. Can't remember if this was also done to handle another issue with foreman_ansible_core, but since that package is removed in favor of just foreman_ansible I suspect it won't be an issue. Let's just beware of this in case any issues show up after the merge.

script/plugin_webpack_directories.rb
ekohl: I'm hesitant to extend default behavior but would be OK with it if we can describe the interface we provide for plugins and which part goes beyond default npm behavior.

config/webpack.config.js
ekohl: While you're at it, it looks like you missed the `^` at the beginning.

tstrachota:
Rebased

script/plugin_webpack_directories.rb
tstrachota: I added a paragraph about the entry points into the docs: https://github.com/theforeman/foreman/pull/5184/commits/0a31ca565ab5f1752b0f4bad54cf8b50aecf5e49 Please check if it makes sense.

config/webpack.config.js
tstrachota: No, I intentionally match only the end of the path to be able to re-use the same pattern for both intl and react-intl.

tstrachota:
@ekohl would you mind taking another look?

webpack/stories/docs/plugins.md
```diff
@@ -69,6 +69,11 @@ Once all the above is set up then the script `npm run install` executed from roo
 Also `npm run lint` behaves similarly.
 
 
+### Entry points
+
+The webpack config respects the main entry point defined in `package.json`. On top of that it loads all files matching `./webpack/.*index.js`. That allows plugins to define multiple independent entry points. This can be useful in special use-cases. For example when you need to mix some parts of webpack processed code into pages that use asset pipeline only.
```
ekohl: The `.* is` a bit ambiguous. A bash shell would match `.somethingindex.js` but not `somethingindex.js` where the code would actually load it (and so would a regex). I'd either make it very clear it's a regex (use `(.*)`) or drop the `.`. The latter is easier IMHO.

config/webpack.config.js
ekohl: I wonder if you could have any issues with for example `een` instead of `en` because of this, but I think that's unlikely because it doesn't really handle a lot of user input and reuse is probably worth more.

ekohl:
> @ekohl would you mind taking another look?

:+1: other than the last trivial issue.

config/webpack.config.js
tstrachota: That shouldn't be a problem. There's `/` at the beginning so the whole regex is for example `/(en|de|cz)$` and it will always match only files in subdirectories with translations. I think we're safe here.

webpack/stories/docs/plugins.md
tstrachota: Updated

tstrachota:
@ekohl thanks, I changed the docs to use `./webpack/*index.js`.

ekohl (dismissed):

ares (changes requested):
During testing I realized I need to set custom env variable ASSET_PATH, while I'm already providing the same info by using --public. It would be good to parse argv of webpack-dev-server and read it from there. It could be used for also parsing `--https` and other params.

Otherwise works well!

tstrachota:
@ares the last commit should fix that. Webpack config now reads `--public`, `--host`, `--https` and `--port`.

ares (dismissed):

ohadlevy (changes requested):
it seems we generated the wrong bundle name for rex.

before (generated via ```npm run analyze```)
```
All (5.13 MB)
vendor-7e974b22751bc4769dcc.js (4.31 MB)
foreman_ansible-04d96110652752e3afc3.js (302.27 KB)
bundle-afce08aaf9c84b01503a.js (281.44 KB)
foreman_remote_execution-d104565b42ed0b69961b.js (249.95 KB)
```
After

```
All (5.64 MB)
vendor-122e7402ac8c40f5b8ff.js (4.4 MB)
foreman_ansible-d628d9c0352cf42bd3fe.js (315.73 KB)
bundle-bfc096e53cd933f92f0a.js (294.9 KB)
foreman_remote_execution_core-c1df28b0ac4e42ba2269.js (263.3 KB)
intl/locale/ru-a7760efcc9511150bee4.js (26.51 KB)
intl/locale/pt-32f23f08c488321f4f5d.js (24.19 KB)
intl/locale/fr-b1601eccc932cad2fa64.js (24.05 KB)
intl/locale/ca-1e8ad8aa3868728b0a0e.js (24 KB)
intl/locale/nl-2e2a6d97829d34c2ff87.js (23.91 KB)
intl/locale/es-502102792c7b4e2bd935.js (23.9 KB)
intl/locale/de-3ed06f626e2e243e1900.js (23.86 KB)
intl/locale/sv-2c2314fc5a83c5a1fccf.js (23.66 KB)
intl/locale/en-e6d85a7f0b5e27bc92fc.js (23.59 KB)
intl/locale/pl-2ade033777ee01b48d72.js (23.59 KB)
intl/locale/gl-0890d08d92044a767e33.js (23.51 KB)
intl/locale/it-718667f30c1f0c80a017.js (23.48 KB)
intl/locale/ko-64888f94049ba7de36a6.js (21.98 KB)
intl/locale/zh-18b8279c56af2f01b537.js (20.38 KB)
react-intl/locale/es-951784f084c629d5f087.js (13.17 KB)
intl/locale/ja-bab9d2a50afef2a669ec.js (12.17 KB)
react-intl/locale/en-7d31511c86c2212a45fd.js (7.06 KB)
react-intl/locale/zh-662ccab329e3af4fcb9c.js (4.68 KB)
react-intl/locale/fr-312eabafdbcc2d88382b.js (4.22 KB)
react-intl/locale/pt-bd268a4462ae1be8bb04.js (2.95 KB)
react-intl/locale/ca-3c83f454af41f9327d18.js (2.8 KB)
eact-intl/locale/ru-2d0f6f61adb59bf9c46f.js (2.3 KB)
react-intl/locale/pl-96ff3027a599698c16f0.js (2.09 KB)
react-intl/locale/nl-3869fec41a97afb827b0.js (1.69 KB)
react-intl/locale/de-7a6876e26f41edb260e5.js (1.65 KB)
react-intl/locale/sv-3e6dd50f2b25cc139a2b.js (1.62 KB)
react-intl/locale/it-1f21a13e7e2dd99d9114.js (1.54 KB)
react-intl/locale/gl-ab4daec4a55995bbd086.js (1.39 KB)
react-intl/locale/ko-21e911695f048417239a.js (993 B)
react-intl/locale/ja-77956907904bc88ced79.js (955 B)
```

config/webpack.config.js
```diff
@@ -11,10 +11,59 @@ var LodashModuleReplacementPlugin = require('lodash-webpack-plugin');
...
+
+const removePort = host => host && host.replace(/:[0-9]*$/, "");
+
+const devServerConfig = () => {
+  const result = require('dotenv').config();
+  if (result.error && result.error.code !== 'ENOENT') {
+    throw result.error;
+  }
+
+  return {
```
ohadlevy, line 57: can we extract this to a different file ? (not to clutter this main config file).

config/webpack.vendor.js
```diff
@@ -22,6 +22,8 @@ module.exports = [
   'prop-types',
   'seamless-immutable',
   'isomorphic-fetch',
+  'react-intl',
+  'intl',
```
ohadlevy, line 26: i see jsonp duplicated in many per language bundles, perhaps you need to add ```intl/locale-data/jsonp``` to the list?

script/plugin_webpack_directories.rb
ohadlevy: this might had a side effect i saw for rex.

config/webpack.vendor.js
tstrachota: I checked the bundles and I don't see any dependencies in the language chunks. `intl/locale-data/jsonp` is just path to the files with translations. Webpack visualiser does shows the path components but with 0 size.

config/webpack.config.js
tstrachota: I'd prefer to keep it here to have all the configuration at one place. If it starts growing in future, we can revisit it.

tstrachota:
@ohadlevy nice catch with the chunk naming. My last commit fixes that.

tstrachota:
Added one more fix in `webpack.config.js` for names of chunks from plugins.

webpack/assets/javascripts/react_app/common/I18n.js
```diff
@@ -1,6 +1,44 @@
 import Jed from 'jed';
+import { addLocaleData } from 'react-intl';
 import { deprecateObjectProperty } from '../../foreman_tools';
 
+const runningInPhantomJS = () => window._phantom !== undefined;
```
ohadlevy, line 5: note we might be running also in chrome as introduced in 6044a043e1a02f8ff3ba1f51cd2f78faea9d2640, perhaps you can also detect based if ENV["JS_TEST_DRIVER"] is defined.

webpack/assets/javascripts/react_app/common/i18nProviderWrapperFactory.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import { IntlProvider } from 'react-intl';
```
ohadlevy, line 2: does this include the functionality of Provider from 'react-redux' ? AFAIU we need it when store is not passed explicitly ?

webpack/assets/javascripts/react_app/components/common/dates/RelativeDateTime.js
```diff
@@ -0,0 +1,42 @@
...
+      day: '2-digit',
+      month: 'short',
+      hour: '2-digit',
+      minute: '2-digit',
+      year: 'numeric',
+    });
+
+    /* eslint-disable react/style-prop-object */
+    return (
+      <span title={title}>
```
ohadlevy, line 18: this means it uses the jquery popover right? perhaps we should be explicit and use the pf one?

webpack/assets/javascripts/react_app/components/common/dates/dates.stories.js
```diff
@@ -0,0 +1,91 @@
...
+    const showSeconds = boolean('Show seconds');
+
+    const timezoneOptions = [
+      'America/Phoenix',
+      'America/Chicago',
+      'America/New_York',
+      'UTC',
+      'Europe/Prague',
+      'Europe/Kiev',
+      'Asia/Tel_Aviv',
```
ohadlevy: its ```Asia/Jerusalem``` AFAIR :)

webpack/assets/javascripts/react_app/components/componentRegistry.js
ohadlevy: :+1: - can you please explain this API? I don't sure I follow it nor how katello plans on using it? note AFAIU this is changing our current mounting interface and perhaps should be revisited outside the context of this PR?

webpack/stories/components/Code/Code.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
```
ohadlevy, line 1: hmm.. should these be part of the pr?

webpack/assets/javascripts/react_app/common/i18nProviderWrapperFactory.js
sharvit: AFAIU the `components/wrapperFactory.js` will wrap it with the react-redux Provider so we should be safe.

This might make it simpler: https://github.com/ratson/react-intl-redux
We can migrate later if we decide to.

webpack/assets/javascripts/react_app/common/I18n.js
tstrachota: Chrom is not a problem as it implements Intl API correctly. We had issues with phantom that provides Intl object but its functionality is quite limited.

webpack/stories/components/Code/Code.js
sharvit: Why not?
Notice it's living under `webpack/stories/component/` and it helps to show code in the storybook.

sharvit (dismissed):
Thanks @tstrachota, I think we can merge it :+1:

webpack/assets/javascripts/react_app/components/common/dates/RelativeDateTime.js
tstrachota: No, this means it uses standard html attribute: https://www.w3schools.com/tags/att_title.asp
I don't think jQuery is involved here.

webpack/stories/components/Code/Code.js
tstrachota: I needed it for the storybook. It's first time where we're using it so I think it's valid to add that. I can make it a separate commit if you want.

webpack/assets/javascripts/react_app/components/common/dates/RelativeDateTime.js
ohadlevy: I was referring to https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/foreman_tools.js#L61

webpack/assets/javascripts/react_app/components/common/dates/dates.stories.js
tstrachota: Ok, I can change that

ohadlevy: no need, I wasn't sure if it has a meaning in terms of real timezone names or not...

tstrachota:
@ohadlevy updated, stories now use `Asia/Jerusalem`.
The PR contains 22 commits. They should be squashed prior to merge.

webpack/assets/javascripts/react_app/components/common/dates/dates.stories.js
tstrachota: It worked for me. I didn't notice `Asia/Tel_Aviv` tz is deemed to be deprecated, but obviously it's still available for backwards compatibility. Anyway a changed it to Jerusalem.

webpack/assets/javascripts/react_app/components/common/dates/RelativeDateTime.js
tstrachota: I see. I didn't see it would apply anywhere on the dates. It's probably because the components mounted after the tooltip initialization.

ohadlevy: right, but if you want to show the correct date (e.g. a minute after the page has been loaded) you would need a different way. I"m OK to leave this outside of this pr.

webpack/assets/javascripts/react_app/components/componentRegistry.js
```diff
@@ -40,37 +45,49 @@ const componentRegistry = {
...
+      throw new Error(
+        `Component not found: ${name} among ${this.registeredComponents()}`
+      );
+    }
+
     return this.registry[name];
   },
 
+  wrapperFactory() {
+    return new WrapperFactory();
```
ohadlevy, line 58: does this happen once per document load? multiple times? I'm guess I don't understand the background for all of this change...

tstrachota: It's called once per component mount when it's mounted into erb or angular. I added that to enable plugins using wrappers easily. An example in Katello is here:
https://github.com/Katello/katello/pull/7001/files#diff-5bcfdae0ace18ad3702fbb46115e768dR19

webpack/assets/javascripts/react_app/components/componentRegistry.js
tstrachota: It doesn't change the API. Interfaces of `MountingService` and `mount_react_component` helper remain the same. It's only the `markup` method in component registry that's been changed.

Anyway, the intention here is to give plugins more control around what to wrap a component with if they're mounting it into js outside of the webpack stack (angular in Katello). In this specific case, there's no need to wrap each date component with a redux store provider when it only needs intl provider.
I mention usage of this API in the storybook too:
https://github.com/theforeman/foreman/pull/5184/files#diff-0bb3dd22c12643cb6a733af82156af0aR53

ares:
@ohadlevy any more comments from your side? thanks

ohadlevy:
> @ohadlevy any more comments from your side? thanks

the last question i had is regarding the generated assets file, I currently see:
```
           react-intl/locale/zh-662ccab329e3af4fcb9c.js    4.79 kB       0  [emitted]         react-intl/locale/zh
           react-intl/locale/ru-2d0f6f61adb59bf9c46f.js    2.36 kB       2  [emitted]         react-intl/locale/ru
           react-intl/locale/pt-bd268a4462ae1be8bb04.js    3.02 kB       3  [emitted]         react-intl/locale/pt
           react-intl/locale/pl-96ff3027a599698c16f0.js    2.14 kB       4  [emitted]         react-intl/locale/pl
           react-intl/locale/nl-3869fec41a97afb827b0.js    1.73 kB       5  [emitted]         react-intl/locale/nl
           react-intl/locale/ko-21e911695f048417239a.js  993 bytes       6  [emitted]         react-intl/locale/ko
           react-intl/locale/ja-77956907904bc88ced79.js  955 bytes       7  [emitted]         react-intl/locale/ja
           react-intl/locale/it-1f21a13e7e2dd99d9114.js    1.58 kB       8  [emitted]         react-intl/locale/it
           react-intl/locale/gl-ab4daec4a55995bbd086.js    1.42 kB       9  [emitted]         react-intl/locale/gl
           react-intl/locale/fr-312eabafdbcc2d88382b.js    4.33 kB      10  [emitted]         react-intl/locale/fr
           react-intl/locale/es-951784f084c629d5f087.js    13.5 kB      11  [emitted]         react-intl/locale/es
           react-intl/locale/en-7d31511c86c2212a45fd.js    7.23 kB      12  [emitted]         react-intl/locale/en
           react-intl/locale/de-7a6876e26f41edb260e5.js    1.69 kB      13  [emitted]         react-intl/locale/de
           react-intl/locale/ca-3c83f454af41f9327d18.js    2.87 kB      14  [emitted]         react-intl/locale/ca
                 intl/locale/zh-18b8279c56af2f01b537.js    20.9 kB      15  [emitted]         intl/locale/zh
                 intl/locale/sv-2c2314fc5a83c5a1fccf.js    24.2 kB      16  [emitted]         intl/locale/sv
                 intl/locale/ru-a7760efcc9511150bee4.js    27.1 kB      17  [emitted]         intl/locale/ru
                 intl/locale/pt-32f23f08c488321f4f5d.js    24.8 kB      18  [emitted]         intl/locale/pt
                 intl/locale/pl-2ade033777ee01b48d72.js    24.2 kB      19  [emitted]         intl/locale/pl
                 intl/locale/nl-2e2a6d97829d34c2ff87.js    24.5 kB      20  [emitted]         intl/locale/nl
                 intl/locale/ko-64888f94049ba7de36a6.js    22.5 kB      21  [emitted]         intl/locale/ko
                 intl/locale/ja-bab9d2a50afef2a669ec.js    12.5 kB      22  [emitted]         intl/locale/ja
                 intl/locale/it-718667f30c1f0c80a017.js      24 kB      23  [emitted]         intl/locale/it
                 intl/locale/gl-0890d08d92044a767e33.js    24.1 kB      24  [emitted]         intl/locale/gl
                 intl/locale/fr-b1601eccc932cad2fa64.js    24.6 kB      25  [emitted]         intl/locale/fr
                 intl/locale/es-502102792c7b4e2bd935.js    24.5 kB      26  [emitted]         intl/locale/es
                 intl/locale/en-e6d85a7f0b5e27bc92fc.js    24.2 kB      27  [emitted]         intl/locale/en
                 intl/locale/de-3ed06f626e2e243e1900.js    24.4 kB      28  [emitted]         intl/locale/de
                 intl/locale/ca-1e8ad8aa3868728b0a0e.js    24.6 kB      29  [emitted]         intl/locale/ca
```

is that expected? (note the react-intl and intl - with the same language) @sharvit any idea?

sharvit:
@ohadlevy - I made some research about the issue and the conclusion is **the assets are fine**.

In details:
* Most of the modern browsers today comes natively with `global.Intl` API.
* `react-intl` rely on the `global.Intl` API
* In case of a browser without a `global.Intl` we need to load polyfills to the `global.Intl` (https://github.com/andyearnshaw/Intl.js)
* It is not enough to just `require('intl')` and we need to load the specific locale data `require('intl/locale-data/jsonp/en.js')`
* The `react-intl` locale data does not provide the full locale data, only the gap between what it needs to what the `global.Intl` provides. (see the size differences)

The implementation looks good to me, see the `IntlLoader` class and how it conditionally loads the polyfills into the `global.Intl`.

ohadlevy (approved):

ohadlevy:
thanks for the explanation @sharvit.

many thanks for your contribution @tstrachota  :-)

tbrisker:
Looks like develop is now broken due to this. @ohadlevy @sharvit please take a look.

ohadlevy:
@tbrisker thanks! this seems to be effecting only development and related to how the hostname is being defined, looking at the console I can see:
```
bootstrap 51ccf83…:766 Refused to load the script 'http://ohad:3808/webpack/react-intl/locale/en.js' because it violates the following Content Security Policy directive: "script-src 'unsafe-eval' 'unsafe-inline' 'self' localhost:3808". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.
```
I'm not sure where the hostname 'ohad' is used vs 'localhost' in this case, but i assume that's the reason for the failure.

sharvit:
@tbrisker seems like some cases it can't figure out the `publicHost`.
Can you try `$WEBPACK_OPTS='--publicHost your-hostname'`?

package.json
```diff
@@ -9,13 +9,16 @@
     "@storybook/addon-knobs": "^3.4.3",
     "@storybook/react": "^3.2.12",
     "@storybook/storybook-deployer": "^2.0.0",
+    "argv-parse": "^1.0.1",
```
ekohl, line 12: I missed this in the final review, but needed a packaging PR as well.

ohadlevy:
https://github.com/theforeman/foreman-packaging/pull/3340

tbrisker:
I'm now also noticing multiple errors with this on the dashboard - https://projects.theforeman.org/issues/25812

PR #5186, opened 2018-01-18
Fixes #22310 - Implement report scanner to identify origin

bastilian:
Report scanner provide an interface for plugins to add callbacks and process them after they have been saved.

The first use is to set an origin for reports submitted by Puppet. 
The origin set is also shown in the report overview.

An implementation for Ansible is also implemented in [foreman-ansible](https://github.com/theforeman/foreman_ansible/pull/116).

### DIFF
--- a/app/helpers/reports_helper.rb
+++ b/app/helpers/reports_helper.rb
@@ -44,4 +44,20 @@ module ReportsHelper
              {}, {:class=>"col-md-1 form-control", :onchange =>"filter_by_level(this);"})
     end
   end
+
+  def report_origin_icon(origin)
+    return 'N/A' if origin.blank?
+    origin_icon = self.try("#{origin.downcase}_report_origin_icon".to_sym)
+    image_tag(origin_icon || origin + ".png", :title => _("Reported by %s") % origin)
+  end
+
+  def report_origin_output_partial(origin)
+    return report_default_partial if origin.blank?
+    origin_partial = self.try("#{origin.downcase}_report_origin_partial".to_sym)
+    origin_partial || report_default_partial
+  end
+
+  def report_default_partial
+    'output'.freeze
+  end
 end
--- a/app/registries/foreman/plugin/report_scanner_registry.rb
+++ b/app/registries/foreman/plugin/report_scanner_registry.rb
@@ -1,0 +1,38 @@
+require_dependency File.expand_path('../../../../services/report_scanner/puppet_report_scanner', __FILE__)
+
+module Foreman
+  class Plugin
+    class ReportScannerRegistry
+      DEFAULT_REPORT_SCANNERS = [
+        ::Foreman::PuppetReportScanner
+      ].freeze
+
+      attr_accessor :report_scanners
+
+      def initialize
+        @report_scanners = []
+        register_default_scanner
+      end
+
+      def report_scanners
+        @report_scanners ||= []
+      end
+
+      def register_report_scanner(scanner)
+        @report_scanners = (report_scanners << scanner).uniq
+      end
+
+      def unregister_report_scanner(scanner)
+        @report_scanners -= [scanner]
+      end
+
+      private
+
+      def register_default_scanner
+        DEFAULT_REPORT_SCANNERS.each do |default_scanner|
+          register_report_scanner default_scanner
+        end
+      end
+    end
+  end
+end
--- a/app/services/report_importer.rb
+++ b/app/services/report_importer.rb
@@ -1,6 +1,6 @@
 class ReportImporter
   delegate :logger, :to => :Rails
-  attr_reader :report
+  attr_reader :report, :report_scanners
 
   # When writing your own Report importer, provide feature(s) of authorized Smart Proxies
   def self.authorized_smart_proxy_features
@@ -42,6 +42,15 @@ class ReportImporter
       refreshed_time = Time.now
       logger.info("Imported report for #{name} in #{(imported_time - start_time).round(2)} seconds, status refreshed in #{(refreshed_time - imported_time).round(2)} seconds")
     end
+  end
+
+  def scan
+    logger.info "Scanning report with: #{report_scanners.join(', ')}"
+    logger.info logs.inspect
+    report_scanners.each do |scanner|
+      break if scanner.scan(report, logs)
+    end
+    logger.debug "Changes after scanning: #{report.changes.inspect}"
   end
 
   private
@@ -130,10 +139,17 @@ class ReportImporter
     host.save(:validate => false)
 
     status = report_status
-
     # and save our report
     @report = report_name_class.new(:host => host, :reported_at => time, :status => status, :metrics => raw['metrics'])
+
+    # Run report scanner
+    scan
+
     @report.save
     @report
+  end
+
+  def report_scanners
+    Foreman::Plugin.report_scanner_registry.report_scanners
   end
 end
--- a/app/services/report_scanner/puppet_report_scanner.rb
+++ b/app/services/report_scanner/puppet_report_scanner.rb
@@ -1,0 +1,17 @@
+module Foreman
+  class PuppetReportScanner
+    class << self
+      def scan(report, logs)
+        if is_puppet = puppet_report?(logs)
+          report.origin = 'Puppet'
+        end
+        is_puppet
+      end
+
+      def puppet_report?(logs)
+        first_log = logs.first
+        first_log && first_log['log'].fetch('sources', {}).fetch('source', '') =~ /Puppet/
+      end
+    end
+  end
+end
--- a/app/views/config_reports/_list.html.erb
+++ b/app/views/config_reports/_list.html.erb
@@ -5,6 +5,7 @@
       <th><%= sort :host, :as => _("Host") %></th>
       <% end %>
       <th><%= sort :reported, :as => _("Last report") %></th>
+      <th class="col-md-1 fact-origin"><%= sort :origin, :as => _('Origin') %></th>
       <th class="col-md-1"><%= sort :applied, :as => _("Applied") %></th>
       <th class="col-md-1"><%= sort :restarted, :as => _("Restarted") %></th>
       <th class="col-md-1"><%= sort :failed, :as => _("Failed") %></th>
@@ -25,6 +26,7 @@
           <% end %>
         <% end %>
         <td><%= reported_at_column(report) %></td>
+        <td class="fact-origin"><%= report_origin_icon(report.origin) %></td>
         <td><%= report_event_column(report.applied, "label-info") %></td>
         <td><%= report_event_column(report.restarted, "label-info") %></td>
         <td><%= report_event_column(report.failed, "label-danger") %></td>
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -28,6 +28,7 @@ class Report < ApplicationRecord
 
       scoped_search :on => :reported_at, :complete_value => true, :default_order => :desc, :rename => :reported, :only_explicit => true
       scoped_search :on => :host_id,     :complete_value => false, :only_explicit => true
+      scoped_search :on => :origin
     end
     super
   end
--- a/app/registries/foreman/plugin.rb
+++ b/app/registries/foreman/plugin.rb
@@ -18,6 +18,7 @@
 require_dependency 'foreman/plugin/logging'
 require_dependency 'foreman/plugin/rbac_registry'
 require_dependency 'foreman/plugin/rbac_support'
+require_dependency 'foreman/plugin/report_scanner_registry'
 
 module Foreman #:nodoc:
   class PluginNotFound < Foreman::Exception; end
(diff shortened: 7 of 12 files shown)


xprazak2:
I tested, seems to be working ok:

app/helpers/reports_helper.rb
```diff
@@ -44,4 +44,16 @@ def logs_show
...
     end
   end
+
+  def report_origin_icon(origin)
+    return 'N/A' if origin.nil?
+    origin_icon = self.try("#{origin.downcase}_report_origin_icon".to_sym)
+    image_tag(origin_icon || origin + ".png", :title => _("Reported by %s") % origin)
+  end
+
+  def report_origin_output_partial(origin, default_partial = 'output')
```
ares: what is the purpose of `default_partial` being configurable? if I call this method and want to change the default, why do I call this method in the first place?

app/services/report_importer.rb
```diff
@@ -15,6 +17,18 @@ def self.unregister_smart_proxy_feature(feature)
     @authorized_smart_proxy_features -= [ feature ]
   end
 
+  def self.report_scanner
+    @report_scanner ||= []
+  end
+
+  def self.register_report_scanner(scanner)
+    @report_scanner = (report_scanner + [scanner]).flatten.uniq
```
ares: why `flatten` is required here?

app/services/report_importer.rb
```diff
@@ -44,6 +59,15 @@ def import
     end
   end
 
+  def scan
+    logger.info "Scanning report with: #{self.class.report_scanner.join(', ')}"
+    self.class.report_scanner.each do |scanner|
```
ares: do we need to go over each scanner? I guess the `scanner.scan` should return true/false (or string/nil) and we should use `any?` instead of `each`.

app/services/report_importer.rb
```diff
@@ -44,6 +59,15 @@ def import
...
+      scanner.scan(report)
+      logger.debug "Changes after #{scanner}: #{report.changes.inspect}"
+    end
+    report.save
```
ares: this adds another save call to the importing chain, could it be removed and the scanning moved just after logs are available but report is not saved yet at https://github.com/bastilian/foreman/blob/9346bf19279d7085090e690f780d20216cbf4596/app/services/report_importer.rb#L160 ?

app/services/config_report_importer.rb
```diff
@@ -1,4 +1,13 @@
 class ConfigReportImporter < ReportImporter
+  DEFAULT_REPORT_SCANNER = [
```
ares: Could this use some plugin registry instead of creating new one in this class? Otherwise I think it should be moved under app/registries to avoid auto-loader issue, see https://github.com/theforeman/foreman/blob/develop/app/registries/README.md for more info

app/helpers/reports_helper.rb
```diff
@@ -44,4 +44,16 @@ def logs_show
...
+    return default_partial if origin.nil?
```
ares: if `origin.blank?` seems safer since we don't have any other validations

ares (changes requested):
:+1: on the direction, I left few comments which I originally wanted to send you PR for, but I'm afraid I won't get to it sooner than you would

app/services/report_importer.rb
bastilian: Having it save separately and not within the process of creating the logs, ensures that in any case the logs and report are saved. Cases where the scanner could "corrupt" a report maybe rare, but they could.

Another advantage of including a save when scanning is that we can have them run independently as well.

I agree with trying to keep writes low, but also trying to asses if it is useful here.
Maybe having saving as an optional argument to pass would satisfy both.

app/helpers/reports_helper.rb
bastilian: It doesn't need to be passable, actually. It is called in the reports view and previously had only passed 'output'. 
I'll change it to simple return 'output' as a default for the render call in the view and remove the default_partial.

app/services/report_importer.rb
bastilian: What would the return of `scanner.scan` indicate? 

Defining more narrowly what the scan method of registered scanners is able/allowed to do might help. ATM it has free reign over what it does to the report and it's associations.

app/services/config_report_importer.rb
bastilian: Thanks for pointing me at registries! I'll change this to have a registry for adding report scanner that `ReportImporter` can call.

app/services/report_importer.rb
ares: We could have it optional if you feel strongly it's useful, but I think it should default to one save per repotr. Think about 10 000 hosts running puppet agent every 30 minutes, that gives you 0.003s per one report, so I think we should keep this part as efficient as possible.

ares: Sorry typing too fast, 0.18s

bastilian:
@ares the save during `scan` is now optional and move into `create_report_and_logs` to safe a `save`.

app/registries/foreman/plugin/report_scanner_registry.rb
```diff
@@ -0,0 +1,30 @@
...
+        @report_scanner = []
+        register_report_scanner DEFAULT_REPORT_SCANNER
+      end
+
+      def report_scanner
+        @report_scanner ||= []
+      end
+
+      def register_report_scanner(scanner)
+        @report_scanner = (report_scanner << scanner).flatten.uniq
```
ares: why `flatten` is required here?

app/services/report_importer.rb
```diff
@@ -44,6 +44,15 @@ def import
     end
   end
 
+  def scan(save = false)
+    logger.info "Scanning report with: #{report_scanner.join(', ')}"
+    report_scanner.each do |scanner|
+      scanner.scan(report)
+      logger.debug "Changes after #{scanner}: #{report.changes.inspect}"
+    end
+    report.save if save
```
ares: so this method now returns true/false/nil, should it rather return return value of `scanner.scan(report)` which returns 'Puppet' or nil?

app/services/report_importer.rb
```diff
@@ -44,6 +44,15 @@ def import
     end
   end
 
+  def scan(save = false)
+    logger.info "Scanning report with: #{report_scanner.join(', ')}"
+    report_scanner.each do |scanner|
```
ares: if first scanner finds the origin, e.g. 'Puppet', since we use `each` it continues scanning, since there can be just one origin, I think it would be better to "break" on first match, probably using `any?` instead of `each`

ares (changes requested):
few more comments, getting there :-)

app/views/config_reports/_list.html.erb
```diff
@@ -5,6 +5,7 @@
       <th><%= sort :host, :as => _("Host") %></th>
       <% end %>
       <th><%= sort :reported, :as => _("Last report") %></th>
+      <th><%= _('Origin') %></th>
```
ares: could you please limit width so it does not take so much space? col-md-1 seems better locally

ares:
getting following when testing
```
2018-02-06T16:30:21 c2d2011c [app] [W] Action failed
 | NoMethodError: undefined method `logs' for nil:NilClass
 | /home/ares/Projekty/Zdrojaky/foreman/app/services/report_scanner/puppet_report_scanner.rb:9:in `puppet_report?'
 | /home/ares/Projekty/Zdrojaky/foreman/app/services/report_scanner/puppet_report_scanner.rb:5:in `scan'
 | /home/ares/Projekty/Zdrojaky/foreman/app/services/report_importer.rb:50:in `block in scan'
 | /home/ares/Projekty/Zdrojaky/foreman/app/services/report_importer.rb:49:in `each'
 | /home/ares/Projekty/Zdrojaky/foreman/app/services/report_importer.rb:49:in `scan'
 | /home/ares/Projekty/Zdrojaky/foreman/app/services/report_importer.rb:142:in `create_report_and_logs'
 | /home/ares/Projekty/Zdrojaky/foreman/app/services/config_report_importer.rb:13:in `create_report_and_logs'
 | /home/ares/Projekty/Zdrojaky/foreman/app/services/report_importer.rb:38:in `import'
 | /home/ares/Projekty/Zdrojaky/foreman/app/services/report_importer.rb:20:in `import'
 | /home/ares/Projekty/Zdrojaky/foreman/app/models/config_report.rb:27:in `import'
```

`scan` gets called to early, before `@report` is set at https://github.com/theforeman/foreman/pull/5186/files#diff-5f2522f8937b1076938bcca2ab931cf1R144

edit: happens for existing hosts only

app/services/report_scanner/puppet_report_scanner.rb
```diff
@@ -0,0 +1,14 @@
+module Foreman
+  class PuppetReportScanner
+    class << self
+      def scan(report)
+        report.origin = 'Puppet' if puppet_report? report
+      end
+
+      def puppet_report?(report)
+        first_log = report.logs.first
+        first_log && first_log.source.value == 'Puppet'
```
ares: this does not work for me, the source I see locally is e.g. "//scapclient.example.tst/Puppet", we should perhaps use `=~` instead?

app/registries/foreman/plugin/report_scanner_registry.rb
bastilian: :man_facepalming: because I called it with an array to register the default scanner all this time...

app/services/report_importer.rb
bastilian: :bulb: Now I see where you are getting at. I didn't realize that the importer only processes one report at a time. I'll rewrite it to scan, break when the first `scanner.scan` returns true and and let it be saved afterwards in the `create_report_and_logs`

app/services/report_scanner/puppet_report_scanner.rb
```diff
@@ -2,12 +2,15 @@ module Foreman
   class PuppetReportScanner
     class << self
       def scan(report)
-        report.origin = 'Puppet' if puppet_report? report
+        if is_puppet = puppet_report?(report)
+          report.origin = 'Puppet'
```
ares: maybe it would be worth of extracting `Puppet` into method or constant in case someone wants to access it

ares:
Thanks for separate commits, I think you can now squash them and continue adding more if required :-) Code wise looks nearly there, I'll continue testing later

bastilian:
@ares I'm currently working on rewriting the scan procedure with tests, something is wonky and doesn't save the origin. After that i'm gonna squash the commits for further review. :)

ares:
[test katello]

app/registries/foreman/plugin/report_scanner_registry.rb
```diff
@@ -0,0 +1,38 @@
...
+      ].freeze
+
+      attr_accessor :report_scanner
+
+      def initialize
+        @report_scanner = []
+        register_default_scanner
+      end
+
+      def report_scanner
```
ares: if this method returns multiple scanners, would it be better to pluralize the name?

app/services/report_importer.rb
```diff
@@ -1,6 +1,6 @@
 class ReportImporter
   delegate :logger, :to => :Rails
-  attr_reader :report
+  attr_reader :report, :report_scanner
```
ares: pluralize the name

app/services/report_importer.rb
```diff
@@ -44,6 +44,15 @@ def import
     end
   end
 
+  def scan
+    logger.info "Scanning report with: #{report_scanner.join(', ')}"
+    logger.info logs.inspect
+    report_scanner.each do |scanner|
```
ares: pluralize the name

app/services/report_importer.rb
```diff
@@ -130,10 +139,17 @@ def create_report_and_logs
...
     @report = report_name_class.new(:host => host, :reported_at => time, :status => status, :metrics => raw['metrics'])
+
+    # Run report scanner
+    scan
+
     @report.save
     @report
   end
+
+  def report_scanner
```
ares: pluralize the name

app/helpers/reports_helper.rb
```diff
@@ -44,4 +44,20 @@ def logs_show
              {}, {:class=>"col-md-1 form-control", :onchange =>"filter_by_level(this);"})
     end
   end
+
+  def report_origin_icon(origin)
+    return 'N/A' if origin.blank?
+    origin_icon = self.try("#{origin.downcase}_report_origin_icon".to_sym)
+    image_tag(origin_icon || origin + ".png", :title => _("Reported by %s") % origin)
```
ares, line 51: this draws the icon bigger than fact origin

ares (changes requested):
2 last nits

bastilian:
@ares addressed the changes you mentioned on IRC.

`report_scanner` is now pluralized and the fact icon is now smaller through the `fact-origin` class.
And origin is sort able:
gnome-shell-screenshot-ljkmez

ares:
APJ

ares (approved):

ares:
Thanks @bastilian, merging!

PR #5187, opened 2018-01-18
Fixes #22199 - Audit has_many associations

swapab:
@ares 
This is how associational audits will look like(Stored in a single parent object). It's the first iteration, we can customize it further to have links to resources(once non-existent resources are thought of) and audit comments like:
```
Role "my viewer" has a new permission :view_hosts" 
User "someone" roles changed, new roles are "my viewer", "manager"
```

associational-audits

---
This PR is WIP because it's a POC implementation of associational audits and needs review/feedback.

### DIFF
--- a/app/models/concerns/audit_associations.rb
+++ b/app/models/concerns/audit_associations.rb
@@ -1,0 +1,62 @@
+module AuditAssociations
+  def self.included(base)
+    base.send :include, InstanceMethods
+  end
+
+  module InstanceMethods
+    def changes_to_save
+      if audited_options[:associations].present?
+        super.merge(associated_changes)
+      else
+        super
+      end
+    end
+
+    private
+
+    def associated_changes
+      associations = Array.wrap(audited_options[:associations])
+      associations.inject({}) do |changes_hash, association_name|
+        association_ids = "#{association_name.to_s.singularize}_ids"
+
+        if send("#{association_ids}_changed?")
+          association_class = self.class.reflect_on_association(association_name).class_name.constantize
+          change = send("#{association_ids}_change")
+          change_ids = change.flatten.uniq
+
+          id_name_map = association_class.where(id: change_ids).inject({}) { |r,p| r.merge(p.id => p.to_label) }
+
+          changes_hash[association_name] = change.inject([]) do |humaized_associations, ids|
+            humaized_associations << ids.inject([]) do |humanized_ids, id|
+              humanized_ids << id_name_map[id]
+              humanized_ids
+            end.join(', ')
+            humaized_associations
+          end
+        end
+
+        changes_hash
+      end
+    end
+  end
+
+  module Auditor
+    def self.included(base)
+      base.extend ClassMethods
+    end
+
+    module ClassMethods
+      include Audited::Auditor::ClassMethods
+
+      def audited(options = {})
+        if options[:associations].present?
+          include DirtyAssociations
+
+          dirty_has_many_associations(*(Array(options[:associations])))
+        end
+
+        super
+      end
+    end
+  end
+end
--- a/app/models/concerns/dirty_associations.rb
+++ b/app/models/concerns/dirty_associations.rb
@@ -57,6 +57,10 @@ module DirtyAssociations
               value
             end
           end
+
+          define_method "#{association_ids}_change" do
+            [self.send("#{association_ids}_was"), self.send(association_ids)]
+          end
         end
       end
       self.prepend(extension)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -11,7 +11,6 @@ class User < ApplicationRecord
   include UserUsergroupCommon
   include Exportable
   include TopbarCacheExpiry
-  audited :except => [:last_login_on, :password_hash, :password_salt, :password_confirmation]
 
   ANONYMOUS_ADMIN = 'foreman_admin'
   ANONYMOUS_API_ADMIN = 'foreman_api_admin'
@@ -121,6 +120,9 @@ class User < ApplicationRecord
   }
 
   dirty_has_many_associations :roles
+
+  audited :except => [:last_login_on, :password_hash, :password_salt, :password_confirmation],
+          :associations => :roles
 
   attr_exportable :firstname, :lastname, :mail, :description, :fullname, :name => ->(user) { user.login }, :ssh_authorized_keys => ->(user) { user.ssh_keys.map(&:to_export_hash) }
 
--- a/test/models/usergroup_test.rb
+++ b/test/models/usergroup_test.rb
@@ -268,4 +268,88 @@ class UsergroupTest < ActiveSupport::TestCase
     groups = Usergroup.search_for("role = #{role.name}")
     assert (groups.include? usergroup)
   end
+
+  context 'audit usergroup' do
+    setup do
+      @usergroup = FactoryBot.create(:usergroup, :with_auditing)
+    end
+
+    context 'child usergroups' do
+      setup do
+        @child_usergroup = FactoryBot.create(:usergroup)
+        @usergroup.usergroup_ids = [@child_usergroup.id]
+        @usergroup.save
+      end
+
+      test 'should audit when a child-usergroup is assigned to a parent-usergroup' do
+        recent_audit = @usergroup.audits.last
+        audited_changes = recent_audit.audited_changes[:usergroups]
+        assert audited_changes, 'No audits found for usergroups'
+        assert_empty audited_changes.first
+        assert_equal @child_usergroup.name, audited_changes.last
+      end
+
+      test 'should audit when a child-usergroup is removed/de-assigned from a parent-usergroup' do
+        @usergroup.usergroup_ids = []
+        @usergroup.save
+        recent_audit = @usergroup.audits.last
+        audited_changes = recent_audit.audited_changes[:usergroups]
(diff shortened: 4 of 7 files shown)


app/models/concerns/audit_associations.rb
```diff
@@ -0,0 +1,42 @@
+module AuditAssociations
+  def changes_to_save
+    unless audited_options[:associations].blank?
```
ares: it would be better to write `if audited_options[:associations].present` since there's else branch

app/models/concerns/audit_associations.rb
```diff
@@ -0,0 +1,42 @@
...
+  private
+
+  def associated_changes
+    associations = Array.wrap(audited_options[:associations])
+    associations.inject({}) do |changes_hash, association_name|
+      association_name = association_name.to_s.singularize
+      association_ids = "#{association_name}_ids"
+
+      if send("#{association_ids}_changed?")
+        association_class = association_name.capitalize.constantize
```
ares: this won't work for all associations, we should find the association through reflections, something like
```ruby
self.class.reflect_on_association(association_name.to_sym).association_class
```

app/models/concerns/audit_associations.rb
```diff
@@ -0,0 +1,42 @@
...
+        change = send("#{association_ids}_change")
+        change_ids = change.flatten.uniq
```
ares: why do we need to flatten?

app/models/concerns/audit_associations.rb
```diff
@@ -0,0 +1,42 @@
...
+
+        human_readable_attr = human_readable_attr(association_class)
+        id_name_map = association_class.where(id: change_ids).select(:id, human_readable_attr).inject({}) { |r,p| r.merge(p.id => p.send(human_readable_attr)) }
```
ares: I think it's better to load all attributes and call `.to_label` that renders the string representation of the object, we can't tell what all attributes are required for this representation to be rendered

app/models/concerns/dirty_associations.rb
```diff
@@ -57,6 +57,10 @@ def dirty_has_many_associations(*args)
               value
             end
           end
+
+          define_method "#{association_ids}_change" do
+            [instance_variable_get("@#{association_ids}_was"), self.send(association_ids)]
```
ares: why not using `self.send("#{association_ids}_was")` for the first item of the array, it would reuse the cache and properly initialize it for new record

app/models/user.rb
```diff
@@ -121,6 +120,9 @@ def self.name_format
 
   dirty_has_many_associations :roles
 
+  audited :except => [:last_login_on, :password_hash, :password_salt, :password_confirmation],
+          :associations => :roles
```
ares, line 125: here we're lucky that we already have `dirty_has_many_associations` definition but I wonder if it wouldn't be better if `audited` call should not include DirtyAssociations and add the definition for all associations. If you look at usergroup model, that would mean we don't need to add line 8 and 43. I think it would be less prone for errors in future.

ares (changes requested):
I like the direction, left few comments that I think should be addressed but so far looks good.

app/models/concerns/audit_associations.rb
swapab: flatten is required beacuse `send("#{association_ids}_change")` returns a 2d array, something like `[[1,2],[3,4]]`

app/models/user.rb
swapab: associational audits utilizes `send("#{association_ids}_changed?")` and `send("#{association_ids}_change")` provided by `dirty_has_many_associations`.

Are you suggesting to `include DirtyAssociations` in `AuditAssociations` itself? Or remove the dependency of `DirtyAssociations` and track dirtiness of associations in a different way.

ares: I think it would be good if DirtyAssociations were included in AuditAssociations and by calling `audited :associations => :a, :b` it would automatically define `dirty_has_many_associations :a, :b`

swapab: Ah! right, makes sense. Thanks @ares

swapab: This change has been done.

app/models/concerns/audit_associations.rb
```diff
@@ -0,0 +1,59 @@
+module AuditAssociations
+  def self.included(base)
```
swapab, line 2: @ares How do you suggest to eliminate the `self.included` methods from these modules. Any better, idiomatic ruby way to do it?

swapab (commented):
@ares All the requested changes are done. PR ready for re-review.

app/models/user.rb
tbrisker: I'm not sure about this tbh, we sometimes want to be able to track changes to associations, for example to check things in callbacks, not just for auditing. for example - https://github.com/theforeman/foreman/blob/develop/app/models/user.rb#L373
Audits should just audit the changes, and with dirty associations we should make sure they are set correctly for the auditing callbacks.

ares: I'm not sure I understand your concern, you can still use dirty cache without auditing. But if you specify association for auditing, it would automatically add dirty cache definition for you.

swapab:
PR ready for re-review.
Katello build passing.
Foreman build failing due to Javascript error.
```
Capybara::Poltergeist::JavascriptError: One or more errors were raised in the Javascript code on the page. If you don't care about these errors, you can ignore them by setting js_errors: false in your Poltergeist configuration (see documentation for details)....
```

ares:
[test foreman]

swapab:
_Analysis of build failures in foreman-core_

I randomly picked the integration tests which are failing on CI-build, running them on my local dev setup was successful without any errors/failures.

Following were the tests and their respective output log.

[`rake test TEST='test/integration/about_test.rb' `](https://paste.fedoraproject.org/paste/WEbYPAKGXp3uQwMLBhlznA)

[`rake test TEST='test/integration/audit_js_test.rb'`](https://paste.fedoraproject.org/paste/~wNair1oTKHiJ2AYSQrKsg)

[`rake test TEST='test/integration/puppetclass_test.rb'`](https://paste.fedoraproject.org/paste/VN34H-dZynZyiV~NwY4GCw)

---

I feel the Jenkins-CI environment needs some eyes. I suspect the webpack compilation has something wrong with it. As all the failures are due to following javascript error.
`TypeError: undefined is not a function (evaluating '$(document)')`

app/models/concerns/audit_associations.rb
ares: ActiveSupport::Concern has support for something like
```ruby
included do
  ...
end
```

but given how many headaches it caused when we had to change alias_method_chain and start using prepend I think pure Ruby callback is good. So I'd fine with this approach.

rplevka:
I tested the PR and it looks sane to me.

ares (approved):
This works pretty well, I think it's ready, I'll wait for @rplevka to share his findings. Meanwhile @swapab could you please rebase so tests run again now on fixed version of Foreman?

ares:
oh meanwhile Roman reacted, @swapab please rebase and we're good to go

swapab:
Rebased. :crossed_fingers:

swapab:
Looks like another javascript error.
Test: test/integration/notifications_drawer_test.rb
```
Capybara::Poltergeist::StatusFailError: Request to 'http://127.0.0.1:33050/' failed to reach server, check DNS and/or server status - Timed out with no open resource requests
    test/integration/notifications_drawer_test.rb:5:in `block in <class:NotificationsDrawerIntegrationTest>' (Capybara::Poltergeist::StatusFailError)
/usr/local/rvm/gems/ruby-2.3.5@test_develop_pr_core-0/gems/poltergeist-1.17.0/lib/capybara/poltergeist/browser.rb:384    
Standard Error
DEPRECATION WARNING: Passing a percentage as the alpha value to rgba() will be
interpreted differently in future versions of Sass. For now, use 0.5 instead.
```

ares:
That is unrelated, I'll merge when I get to computer

ares:
Thanks @swapab, merging!

test/models/usergroup_test.rb
```diff
@@ -268,4 +268,88 @@ def populate_usergroups
...
+        assert audited_changes, 'No audits found for users'
+        assert_empty audited_changes.first
+        assert_equal @user.name, audited_changes.last
+      end
+
+      test 'should audit when a user is removed/de-assigned from a usergroup' do
+        @usergroup.user_ids = []
+        @usergroup.save
+        recent_audit = @usergroup.audits.last
+        audited_changes = recent_audit.audited_changes[:users]
```
lzap, line 348: Hello there, @swapab this is causing issues when running `develop` test locally:

```
UsergroupTest::audit usergroup::users#test_0002_should audit when a user is removed/de-assigned from a usergroup:
NoMethodError: undefined method `audited_changes' for nil:NilClass
    test/models/usergroup_test.rb:405:in `block (3 levels) in <class:UsergroupTest>'
```

Looks like the array was empty in my case, not sure why this does not appear on jenkins, but I think we should fix this.

ares: most likely this happens when you run subset of tests since there are no recent audits, I agree it's worth fixing and ensuring there's at least one audit for this usergroup in this test, I don't expect @swapab to update this in near future, @lzap could you please send a fix and I'd review/merge

PR #5190, opened 2018-01-19
Fixes #22334 - Update safemode, remove ruby2ruby pin

ekohl:
I don't think we actually require safemode 1.3.4 but this ensures the known incompatible combination is used.


iNecas (approved):
test failures independent

PR #5191, opened 2018-01-19
Fixes #22339 - normalize empty values for User.timezone

iNecas:
Otherwise, we would need to deal with two values for the browser's
timezone. Less variants is better in this case.


dLobatog (approved):
Thanks @iNecas , looks good to me. Errors in tests are an integration test failure and something odd `
ERROR: Failed to archive artifacts: tmp/capybara/*,Gemfile.lock
Also:   hudson.remoting.Channel$CallSiteStackTrace: Remote call to slave07.rackspace.theforeman.org
		at hudson.remoting.Channel.attachCallSiteStackTrace(Channel.java:1693)` - will merge.

PR #5197, opened 2018-01-22
Fixes #22354 - Make SilencedLogger thread safe

lzap:
Rails logger silencer code was thread safe for some time:

https://github.com/rails/rails/commit/629efb605728b31ad9644f6f0acaf3760b641a29

But this has been refactored in Rails 5.0 and thread safety is now part of logger instead of silencer:

https://github.com/rails/rails/commit/2518bda97cbbcb33dc9a92e70d5b01c09e64d12d

The patch made assumption that all Rails applications are using Rails logging stack, but that's not the case for Foreman - we rely on much more flexible "logging" gem (https://github.com/TwP/logging) which does not implement this type of thread safety for levels.

Foreman has its own logging delegator called SilencedLogger and the code there was never thread safe, concurrent sprockets requests are now using silence method which is not thread safe at all, with logging gem this leads to level set to ERROR (the default) after few requests in development mode which renders Foreman unusable (no debug logging output). This patch uses the same technique and stores temporary levels into thread-only context, so on concurrent access there are no incorrect readouts.

### DIFF
--- a/lib/foreman/silenced_logger.rb
+++ b/lib/foreman/silenced_logger.rb
@@ -1,16 +1,47 @@
 # Wrap a ::Logging::Logger and implement #silence to temporarily increase the
 # min log level to execute a block of code. Used by sprockets-rails' quiet
-# assets logging feature.
+# assets logging feature. Foreman uses logging gem instead Rails logging stack
+# and it does not provide silencing feature (ships with just a noop method).
+# This implementation only works with Logging gem, not with Ruby/Rails Logger.
+#
+# Inspired by lib/active_record/session_store/extension/logger_silencer.rb
+#
 module Foreman
   class SilencedLogger < SimpleDelegator
-    def silence(new_level = Logger::ERROR, &block)
-      old_level = self.level
-      begin
-        self.level = new_level
-        yield self
-      ensure
-        self.level = old_level
+    def initialize(obj)
+      ::Logging::LEVELS.each do |name, num|
+        instance_eval <<-EOT, __FILE__, __LINE__ + 1
+        def #{name.downcase}?
+          #{num} >= local_level
+        end
+
+        def #{name.downcase}(*args)
+          super(*args) if #{num} >= local_level
+        end
+        EOT
       end
+
+      super(obj)
     end
+
+    def level_key
+      @level_key ||= :"SilencedLogger##{object_id}@level"
+    end
+
+    def local_level
+      Thread.current[level_key] || self.level
+    end
+
+    def local_level=(level)
+      Thread.current[level_key] = level
+    end
+
+    def silence(temporary_level = Logger::ERROR, &block)
+      self.local_level = temporary_level
+      yield self
+    ensure
+      self.local_level = nil
+    end
+    alias :silence_logger :silence
   end
 end
--- a/test/unit/foreman/logging_test.rb
+++ b/test/unit/foreman/logging_test.rb
@@ -33,9 +33,46 @@ class ForemanLoggingTest < ActiveSupport::TestCase
     logger = Foreman::Logging.logger('test_logger')
     appender = ::Logging::Appenders.string_io('buffer')
     logger.appenders = [appender]
-    logger.silence { |log| log.warn("Should not be logged") }
-    assert_equal [], appender.readlines
+    logger.silence do |log|
+      log.debug("Must not be logged")
+      assert_equal 3, log.local_level
+      assert_equal 0, log.level
+      assert_equal [], appender.readlines
+    end
     assert_equal 0, logger.level
+  end
+
+  def test_logger_supports_silence_logger
+    Foreman::Logging.add_logger('test_logger', {:enabled => true, :level => :debug})
+    logger = Foreman::Logging.logger('test_logger')
+    appender = ::Logging::Appenders.string_io('buffer')
+    logger.appenders = [appender]
+    logger.silence_logger do |log|
+      log.warn("Must not be logged as well")
+      assert_equal [], appender.readlines
+    end
+  end
+
+  def test_logger_supports_silence_positive_error
+    Foreman::Logging.add_logger('test_logger', {:enabled => true, :level => :debug})
+    logger = Foreman::Logging.logger('test_logger')
+    appender = ::Logging::Appenders.string_io('buffer')
+    logger.appenders = [appender]
+    logger.silence(::Logger::ERROR) do |log|
+      log.error("Must be logged")
+      assert_match(/Must be logged/, appender.readlines.first)
+    end
+  end
+
+  def test_logger_supports_silence_positive_info
+    Foreman::Logging.add_logger('test_logger', {:enabled => true, :level => :debug})
+    logger = Foreman::Logging.logger('test_logger')
+    appender = ::Logging::Appenders.string_io('buffer')
+    logger.appenders = [appender]
+    logger.silence(::Logger::INFO) do |log|
+      log.error("Must be logged")
+      assert_match(/Must be logged/, appender.readlines.first)
+    end
   end
 
   def test_error_config_missing
--- a/lib/foreman/logging.rb
+++ b/lib/foreman/logging.rb
@@ -2,8 +2,6 @@ require 'logging'
 require 'fileutils'
 require_dependency File.expand_path('../silenced_logger', __FILE__)
 
-::Logging::Logger.send(:include, ActiveRecord::SessionStore::Extension::LoggerSilencer)
-
 module Foreman
   class LoggingImpl
     private_class_method :new


lib/foreman/silenced_logger.rb
```diff
@@ -3,13 +3,36 @@
...
+    def silence(new_level = Logger::DEBUG, &block)
+      old_level = local_level
       begin
-        self.level = new_level
+        local_level = new_level
         yield self
       ensure
+        Rails.logger.error "Setting level to #{old_level}"
         self.level = old_level
+        local_level = nil
```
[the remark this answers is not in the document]
lzap: Fixed.

lib/foreman/silenced_logger.rb
```diff
@@ -1,15 +1,38 @@
 # Wrap a ::Logging::Logger and implement #silence to temporarily increase the
 # min log level to execute a block of code. Used by sprockets-rails' quiet
-# assets logging feature.
+# assets logging feature. Foreman uses logging gem instead Rails logging stack
+# and it does not provide silencing feature (ships with just a noop method).
 module Foreman
   class SilencedLogger < SimpleDelegator
-    def silence(new_level = Logger::ERROR, &block)
-      old_level = self.level
+    def initialize(*args)
```
iNecas: What about storing the level in `Thread.current[:foreman_silenced_logger_level]´. It looks to me like that would be enough needed here

lzap: I can do that with one big warning, `SilencedLogger` is not a singleton, we create instance for every logger (app, sql, etc). To prevent threading issues, I would need to create the symbol key as something like `foreman_log_level_#{object_id}` and that leads to memory leak on older Ruby VMs, according to discussion at https://github.com/rails/rails/pull/20507 but this should be solved with Ruby 2.2 (https://www.sitepoint.com/symbol-gc-ruby-2-2/). It is acceptable for us as we are moving to 2.2+ in production now I guess, more importantly we only create couple of loggers memorizing them and also it is a technique that other part of Rails use: https://github.com/rails/activerecord-session_store/blob/master/lib/active_record/session_store/extension/logger_silencer.rb

The implementation I propose and which is copy-pasted from Rails 5 is safe on all Ruby versions plus it does not pollute thread variables hash with many variables.

Make your pick. Your proposal is cleaner code, my proposal is cleaner in memory :-)

I saw @swapab in the PR mentioned, opinion?

dLobatog: @lzap I think @iNecas 's code is enough in our scenario - we only test with Ruby 2.3 & 2.4 in Jenkins even

dLobatog (changes requested):

lib/foreman/silenced_logger.rb
lzap: @dLobatog not sure what you mean, I cannot do just `Thread.current[:foreman_silenced_logger_level]` as we create multiple instances of this class *per-thread*. I lean towards the most safe hash approach.

lzap:
Just for the record, my initial proposal was:

```ruby
  module Foreman
    class SilencedLogger < SimpleDelegator
      def initialize(*args)
        @@local_levels = Concurrent::Map.new(initial_capacity: 2)
        super(*args)
      end
  
      def local_log_id
        Thread.current.__id__
      end
  
      def local_level
        @@local_levels[local_log_id] || self.level
      end
  
      def local_level=(level)
        if level
          @@local_levels[local_log_id] = self.level = level
        else
          @@local_levels.delete(local_log_id)
        end
      end
  
      def silence(new_level = Logger::DEBUG, &block)
        old_level = local_level
        begin
          self.local_level = new_level
          yield self
        ensure
          self.level = old_level
          self.local_level = nil
        end
      end
    end
  end
```

I am rewriting this to single thread local variable per thread object id.

lzap:
Hmmm the simplified version does not work for me.

```diff
diff --git a/lib/foreman/silenced_logger.rb b/lib/foreman/silenced_logger.rb
index 52c6d8992..aff004420 100644
--- a/lib/foreman/silenced_logger.rb
+++ b/lib/foreman/silenced_logger.rb
@@ -4,24 +4,19 @@
 # and it does not provide silencing feature (ships with just a noop method).
 module Foreman
   class SilencedLogger < SimpleDelegator
-    def initialize(*args)
-      @@local_levels = Concurrent::Map.new(initial_capacity: 2)
-      super(*args)
-    end
-
-    def local_log_id
-      Thread.current.__id__
+    def level_key
+      @level_key ||= :"SilencedLogger##{object_id}@level"
     end
 
     def local_level
-      @@local_levels[local_log_id] || self.level
+      Thread.current[level_key] || self.level
     end
 
     def local_level=(level)
       if level
-        @@local_levels[local_log_id] = self.level = level
+        Thread.current[level_key] = self.level = level
       else
-        @@local_levels.delete(local_log_id)
+        Thread.current[level_key] = nil
       end
     end
```

lib/foreman/silenced_logger.rb
iNecas: Btw. you were using `@@class_variavles` in the original code, so the code would not handle the mutliple instances of the class in single thread anyway, if you didn't use something like the `object_id` in the key.

lib/foreman/silenced_logger.rb
```diff
@@ -1,15 +1,38 @@
...
+      @@local_levels = Concurrent::Map.new(initial_capacity: 2)
+      super(*args)
+    end
+
+    def local_log_id
+      Thread.current.__id__
+    end
+
```
iNecas: Shouldn't you override the `level` method here to return the `local_level`. How otherwise the logger will even find out what the current level is?

lib/foreman/silenced_logger.rb
```diff
@@ -1,15 +1,38 @@
...
+    def silence(new_level = Logger::DEBUG, &block)
+      old_level = local_level
       begin
-        self.level = new_level
+        self.local_level = new_level
         yield self
       ensure
         self.level = old_level
```
iNecas: Are you sure you want to set `level` here?

iNecas:
Defined `doesn't work` :). I've had some questions even in the original implementation that make me ask whether the code even worked in the original proposal.

lzap:
I have rewritten this completely, it's now using just one thread hash key with object id. Added more tests to verify it really works. Removed unnecessary ActiveRecord silencer.

lib/foreman/silenced_logger.rb
```diff
@@ -1,16 +1,51 @@
...
   class SilencedLogger < SimpleDelegator
-    def silence(new_level = Logger::ERROR, &block)
-      old_level = self.level
-      begin
-        self.level = new_level
-        yield self
-      ensure
-        self.level = old_level
+    def initialize(obj)
+      ::Logging::LEVELS.each do |name, num|
```
iNecas, line 12: The metaprogramming part should not be in the initializer, but rather outside of any method: otherwise we would loose method lookup cache everytime we initialize the silencer.

Given you override `log_event` now, do we even need to override this methods?

test/unit/foreman/logging_test.rb
```diff
@@ -33,7 +33,14 @@ def test_logger_supports_silence
...
     logger.appenders = [appender]
-    logger.silence { |log| log.warn("Should not be logged") }
+    logger.silence do |log|
+      log.debug("Must not be logged")
+      assert_equal 3, log.local_level
+      assert_equal 0, log.level
+    end
+    logger.silence_logger do |log|
+      log.warn("Must not be logged as well")
+    end
```
iNecas: I wonder if we could get something like `StringIO` as the backend of the logger, so that we could get it's content at the end and check, which strings got and got not logged?

lib/foreman/silenced_logger.rb
lzap: Putting this outside of initializer does not work, it simply does not create the methods correctly. If you show me a way I'd happily do that.

Overriding `log_event` is actually useless, I will remove that. The thing is this is a decorator, not superclass. The only way I found was to actually explicitly create methods.

test/unit/foreman/logging_test.rb
lzap: We already do that in the test.

xprazak2:
What is the status here? The test failures seems unrelated, one is integration js test and the other failure of Jenkins slave.

[test foreman]
[test katello]

lzap:
Sorry forgot to push. This is ready for @iNecas / @dLobatog another round. Annoying bug.

test/unit/foreman/logging_test.rb
iNecas: I see, perhaps we should however test that at least something gets logged if the level is high enough

lzap: I added two positive tests and rebased.

dLobatog (approved):
I've noticed this problem over and over since I saw this PR, it's difficult to reproduce, but tests pass. I'm merging for the time being, will comment if it doesn't work.

ares:
@dLobatog since RM issue was set to 1.17, would you mind CP to 1.17-stable branch as well? cc @xprazak2 for visibility

lzap:
Thanks this was pain in the ass.

iNecas:
Actually, there was one non-resolved thing in the review: sent additional PR to fix that https://github.com/theforeman/foreman/pull/5252

lzap:
1.17 cherry pick: https://github.com/theforeman/foreman/pull/5260

PR #5199, opened 2018-01-22
Fixes #22282 - Test oVirt connection before save

orrabin:


app/models/compute_resources/foreman/model/ovirt.rb
```diff
@@ -420,7 +424,7 @@ def ca_cert
     private
 
     def update_available_operating_systems
-      return false if errors[:url].any?
+      return false if errors.any?
```
orrabin, line 444: @ShimShtein I thought about this and maybe we should prevent loading the operating system for any error and not just errors[:base] that I can put in `connect`. If you don't agree I'll go back to checking url and base errors and change the `connect` method to set it.

ShimShtein: It makes sense. I think there is no point making extra oVirt API calls, if the model is not valid. I think you could just call `#valid?` here, by the way.

orrabin: I don't think there is such a difference since `valid?` also depends on the errors and if we did not update errors I'm fine with breaking because the validation is wrong.

app/models/compute_resources/foreman/model/ovirt.rb
```diff
@@ -7,7 +7,7 @@ class Ovirt < ComputeResource
               :url_schema => ['http', 'https']
     validates :user, :password, :presence => true
     before_create :update_public_key
-    after_validation :update_available_operating_systems unless Rails.env.test?
+    after_validation :connect, :update_available_operating_systems unless Rails.env.test?
```
ShimShtein: If we do `#update_public_key` through `#connect` after each validation, do we still need to state it explicitly in `before_create`?

app/models/compute_resources/foreman/model/ovirt.rb
```diff
@@ -145,6 +145,10 @@ def test_https_required
 
     def test_connection(options = {})
       super
+      connect(options)
+    end
+
+    def connect(options = {})
       if errors[:url].empty? && errors[:username].empty? && errors[:password].empty?
```
ShimShtein: Can we refactor this long `if` condition into a method, just to make code climate (and our eyes) happier?

We could also refactor this method to use quick return:

``` ruby
def connect(options = {})
  return unless connection_properties_valid?
  update_public_key options
  # ...
end
```

app/models/compute_resources/foreman/model/ovirt.rb
ShimShtein: :ok_hand: then. I can live with that :)

ShimShtein (changes requested):

iNecas:
What's the status here?

app/models/compute_resources/foreman/model/ovirt.rb
orrabin: you are right, not needed

orrabin:
@ShimShtein I rebased and fixed the complex condition.

app/models/compute_resources/foreman/model/ovirt.rb
```diff
@@ -6,8 +6,7 @@ class Ovirt < ComputeResource
     validates :url, :format => { :with => URI::DEFAULT_PARSER.make_regexp }, :presence => true,
               :url_schema => ['http', 'https']
     validates :user, :password, :presence => true
-    before_create :update_public_key
-    after_validation :update_available_operating_systems unless Rails.env.test?
+    after_validation :connect, :update_available_operating_systems unless Rails.env.test?
```
ShimShtein, line 9: You have changed the behavior for test environment: `update_public_key` was called even for test environment, but now `connect` will not be called.
Is it intentional?

ShimShtein (changes requested):

app/models/compute_resources/foreman/model/ovirt.rb
orrabin: No but I wonder if we should actually call `update_public_key` for test environment.
I could add it just for test environment. I can't use `connect` in test environment because there isn't a real ovirt to connect to.

ShimShtein: If we are sure we can't use real ovirt for tests, then I think it's OK to leave it here.

orrabin: we can't, only mocks

ShimShtein (approved):

ShimShtein:
Merged, thanks @orrabin!

ares:
Thanks for merging @ShimShtein, please don't forget to set the release in redmine, marked as 1.18.0

PR #5201, opened 2018-01-22
Fixes #22358 - add templates importing

ares:
The importing is easily extendable for other template types. It's based on the logic in foreman_templates but refactored so each subtype can add its own metadata parsing logic.

### DIFF
--- a/app/controllers/api/v2/provisioning_templates_controller.rb
+++ b/app/controllers/api/v2/provisioning_templates_controller.rb
@@ -55,6 +55,25 @@ module Api
         process_response @provisioning_template.save
       end
 
+      api :POST, "/provisioning_templates/import", N_("Import a provisioning template")
+      param :provisioning_template, Hash, :required => true, :action_aware => true do
+        param :name, String, :required => true, :desc => N_("template name")
+        param :template, String, :required => true, :desc => N_("template contents including metadata")
+      end
+      param :options, Hash, :required => false do
+        param :force, :bool, :allow_nil => true, :desc => N_('use if you want update locked templates')
+        param :associate, ['new', 'always', 'never'], :allow_nil => true, :desc => N_('determines when the template should associate objects based on metadata, new means only when new template is being created, always means both for new and existing template which is only being updated, never ignores metadata')
+      end
+
+      def import
+        options = params.permit(:options => {}).try(:[], :options) || {}
+        template_params = params.require(:provisioning_template).permit(:name, :template)
+        name = template_params[:name]
+        text = template_params[:template]
+        @provisioning_template = ProvisioningTemplate.import!(name, text, options)
+        process_response @provisioning_template
+      end
+
       api :PUT, "/provisioning_templates/:id", N_("Update a provisioning template")
       param :id, :identifier, :required => true
       param_group :provisioning_template
@@ -133,7 +152,7 @@ module Api
 
       def action_permission
         case params[:action]
-          when 'clone'
+          when 'clone', 'import'
             'create'
           when 'export'
             'view'
--- a/app/models/provisioning_template.rb
+++ b/app/models/provisioning_template.rb
@@ -3,6 +3,7 @@ class ProvisioningTemplate < Template
   extend FriendlyId
   friendly_id :name
   include Parameterizable::ByIdName
+  include DirtyAssociations
 
   class << self
     # we have to override the base_class because polymorphic associations does not detect it correctly, more details at
@@ -45,6 +46,8 @@ class ProvisioningTemplate < Template
 
   attr_exportable :kind => Proc.new { |template| template.template_kind.try(:name) },
                   :oses => Proc.new { |template| template.operatingsystems.map(&:name).uniq }
+
+  dirty_has_many_associations :template_combinations, :os_default_templates, :operatingsystems
 
   # Override method in Taxonomix as Template is not used attached to a Host,
   # and matching a Host does not prevent removing a template from its taxonomy.
@@ -219,6 +222,20 @@ class ProvisioningTemplate < Template
   end
 
   private
+
+  def import_custom_data(options)
+    self.template_kind = nil if self.snippet
+
+    if @importing_metadata.key?('kind') && !self.snippet && associate_metadata_on_import?(options)
+      kind = TemplateKind.find_by_name @importing_metadata['kind']
+      if kind.nil?
+        errors.add :template_kind_id, _('specified template "%s" kind was not found') % @importing_metadata['kind']
+        return
+      end
+      self.template_kind = kind
+    end
+    import_oses(options)
+  end
 
   def allowed_changes
     super + %w(template_combinations template_associations)
--- a/app/models/template.rb
+++ b/app/models/template.rb
@@ -73,7 +73,98 @@ class Template < ApplicationRecord
     self
   end
 
+  # Set attributes based on template +text and metadata found in this +text
+  # the metadata parsing can be adjusted using +options
+  def import_without_save(text, options = {})
+    self.template = text
+    @importing_metadata = self.class.parse_metadata(text)
+    Foreman::Logging.logger('app').debug "setting attributes for #{self.name} with id: #{self.id || 'N/A'}"
+    self.snippet = !!@importing_metadata[:snippet]
+
+    import_locations(options)
+    import_organizations(options)
+    import_custom_data(options)
+
+    self
+  end
+
+  # Set template attributes
+  #
+  # based on +name it either finds existing template or builds a new one
+  # then it applies changes to it and return this object, note no changes were saved at this point
+  def self.import_without_save(name, text, options = {})
+    template = self.unscoped.find_or_initialize_by(:name => name)
+    Foreman::Logging.logger('app').debug "#{template.new_record? ? 'building new' : 'updating existing'} template"
+    template.import_without_save(text, options)
+  end
+
+  # Pull out the first erb comment only - /m is for a multiline regex
+  def self.parse_metadata(text)
+    extracted = text.match(/<%\#[\t a-z0-9=:]*(.+?).-?%>/m)
+    extracted.nil? ? {} : YAML.load(extracted[1]).with_indifferent_access
+  rescue RuntimeError => e
+    Foreman::Logging.exception('invalid metadata', e)
+    {}
+  end
+
+  # Updates template metadata and save! it
+  #
+  # options can contain following keys
+  #   :force - set to true if you want to bypass locked templates
+  #   :associate - either 'new', 'always' or 'never', determines when the template should associate objects based on metadata
+  def self.import!(name, text, options = {})
+    template = import_without_save(name, text, options)
+    if options[:force]
+      template.ignore_locking { template.save! }
+    else
+      template.save!
+    end
+    template
+  end
+
   private
+
+  # This method can be overridden in Template children classes to import additional attributes
+  # specific to their type
+  #
+  # it can rely on self.template being updated and @importing_metadata to be populated with parsed
+  # metadata
+  def import_custom_data(_options)
+  end
+
+  # Sets operatingsystem_ids of a template, it's used by provisioning template and ptable, which
+  # is why it lives here. Note that it's still considered as custom since other template types
+  # don't have relation to operating systems.
+  def import_oses(options)
+    if @importing_metadata.key?('oses') && associate_metadata_on_import?(options)
+      oses = Operatingsystem.authorized(:view_operatingsystems).all.select do |existing_os|
+        @importing_metadata['oses'].any? {|imported_os| existing_os.to_label =~ /\A#{imported_os}/}
+      end
+      self.operatingsystem_ids = oses.map(&:id)
+    end
+  end
+
+  def import_organizations(options)
+    if @importing_metadata.key?('organizations') && associate_metadata_on_import?(options)
+      organizations = User.current.my_organizations.where(:name => @importing_metadata['organizations'])
+      self.organization_ids = organizations.map(&:id)
+    else
+      self.organization_ids << Organization.current.id if Organization.current && !self.organization_ids.include?(Organization.current.id)
+    end
+  end
+
+  def import_locations(options)
+    if @importing_metadata.key?('locations') && associate_metadata_on_import?(options)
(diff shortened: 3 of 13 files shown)


lzap:
I am taking a look on this, just to make things clear - this is a WIP or proposal according to https://community.theforeman.org/t/rfc-foreman-templates-refactoring-and-improvements/8358 or what exactly your intentions are with this patch?

app/controllers/api/v2/provisioning_templates_controller.rb
```diff
@@ -54,6 +54,25 @@ def create
         process_response @provisioning_template.save
       end
 
+      api :POST, "/provisioning_templates/import", N_("Import a provisioning template")
+      param :provisioning_template, Hash, :required => true, :action_aware => true do
+        param :name, String, :required => true, :desc => N_("template name")
+        param :template, String, :required => true, :desc => N_("content of a template including metadata")
```
lzap: I think "template content*s* including metadata" is better. Note the plural. Same below.

app/controllers/api/v2/provisioning_templates_controller.rb
```diff
@@ -54,6 +54,25 @@ def create
...
+      end
+      param :options, Hash, :required => false do
+        param :enforce, :bool, :allow_nil => true
+        param :associate, ['new', 'always', 'never'], :allow_nil => true
+      end
```
lzap: Missing doco texts, I am not sure what `enforce` stands for at the moment.

app/controllers/api/v2/provisioning_templates_controller.rb
```diff
@@ -54,6 +54,25 @@ def create
...
+
+      def import
+        template_params = params.require(:provisioning_template).permit(:name, :template)
+        name = template_params[:name]
+        text = template_params[:template]
+        options = params.permit(:options)
```
lzap: Can't you permit options on the first line of the method? I will need to double-check later.

app/models/provisioning_template.rb
```diff
@@ -3,6 +3,7 @@ class ProvisioningTemplate < Template
   extend FriendlyId
   friendly_id :name
   include Parameterizable::ByIdName
+  include DirtyAssociations
```
lzap, line 6: What is the reason for this here and in ptable? I don't see any use of `_was` or `changed?` methods in the new code. Is this attempt to fix some misbehavior not related to the patch?

app/models/template.rb
```diff
@@ -73,8 +73,99 @@ def ignore_default
...
+
+    self
+  end
+
+  # Set template attributes
+  #
+  # based on +name it either finds existing template or builds a new one
+  # then it applies changes to it and return this object, note no changes were saved at this point
+  def self.import_without_save(name, text, options = {})
+    template = self.unscoped.find_or_initialize_by(:name => name)
```
lzap, line 96: This smells like cross-org template injection. I haven't tried yet, but if there is a scenario with existing template and an attacker know the name, template text field from different organization can be replaced possibly. We only check taxonomy for fields sent via metadata, but if they are not present, the check is skipped. Let me show down below in `import_organizations`.

app/models/template.rb
```diff
@@ -73,8 +73,99 @@ def ignore_default
...
+      self.operatingsystem_ids = oses.map(&:id)
+    end
+  end
+
+  def import_organizations(options)
+    if @importing_metadata.key?('organizations') && associate_metadata_on_import(options)
+      organizations = User.current.my_organizations.where(:name => @importing_metadata['organizations'])
+      self.organization_ids = organizations.map(&:id)
+    else
+      self.organization_ids << Organization.current.id if Organization.current && !self.organization_ids.include?(Organization.current.id)
```
lzap: If metadata is not present, we actually add attackers taxonomy to it and she/he not only replaces text, but also gains taxonomy. I think we need to explicitly check users taxonomy against template found before import starts.

app/models/template.rb
```diff
@@ -73,8 +73,99 @@ def ignore_default
     self
   end
 
+  # Set attributes based on template +text and metadata found in this +text
+  # the metadata parsing can be adjusted using +options
+  def import_without_save(text, options = {})
+    self.template = text
+    @importing_metadata = self.class.parse_metadata(text)
+    Foreman::Logging.logger('app').debug "setting attributes for #{self.name} with id: #{self.id || 'N/A'}"
```
lzap, line 81: This line (logger) and mixture of class and instance methods is a good sign that we are abusing the model class too much. We have a God object problem here, please pick any reasonable Ruby design pattern and extract import logic into separate class responsible only for importing. This is my biggest concern so far.

app/models/template.rb
```diff
@@ -73,8 +73,99 @@ def ignore_default
...
+  rescue RuntimeError => e
+    Foreman::Logging.exception('invalid metadata', e)
+    {}
+  end
+
+  # Updates template metadata and save! it
+  #
+  # options can contain following keys
+  #   :force - set to true if you want to bypass locked templates
+  #   :associate - either 'new', 'always' or 'never', determines when the template should associate objects based on metadata
```
lzap, line 114: Here is the help that needs to be in API apipie and not here.

app/models/template.rb
```diff
@@ -73,8 +73,99 @@ def ignore_default
...
+      locations = User.current.my_locations.where(:name => @importing_metadata['locations'])
+      self.location_ids = locations.map(&:id)
+    else
+      self.location_ids << Location.current.id if Location.current && !self.location_ids.include?(Location.current.id)
+    end
+  end
+
+  def associate_metadata_on_import(options)
+    (options[:associate] == 'new' && self.new_record?) || (options[:associate] == 'always')
+  end
```
lzap: Should be `associate_metadata_on_import?`

lzap (changes requested):
Good overall, two major issues: God object and security concern about taxonomy.

app/models/provisioning_template.rb
ares: This was added so that audited gem can keep track of changes on import.

app/models/template.rb
ares: This is in fact covered by permission check on creation and taxonomy validation. We can't have two objects with the same name, taxonomix does not handle it and it would fail on mísleading validation error. So if such template exists, we use it as the object, but when you try to save it, it would fail with message like "you're not authorized for such update" if that would result in escalation. If it's not clear from the comment, I suggest you test it. Also keep in mind that non-adminuser can't create template without choosing org and loc.

app/models/template.rb
ares: Sorry, I don't understand how using Logger contributes to god object problem. Logger is singleton. Normally you'd write logger.debug but here you're out of that scope. Parsing metadata is not depending on instance, which is why it's class method. Handling specific attributes is. I could extract importing to separate service object, but this logic seems very related to this model.

If I extract this to separate object, we'd also lose the inheritance that extends the import logic specific to children. See the ptable and provisioning template only extends it with logic specific for their classes. I can imagine copying this concept to service objects too but IMHO that does not bring value and makes the code less readable. 200 lines model does not seem as god object, especially if half of that are private methods.

app/models/template.rb
ares: I think this is in fact desired. Template in any context is mostly useless as it can only be used by admin and is not visible to anyone else. Users always work in some context. Any context no longer works as available everywhere since 1.15.

ares:
Thanks for initial review, I left few comments and will address the rest.

app/models/template.rb
lzap: The problem is that importing mechanism should not be that tightly coupled with the model. It is definitely not the biggest God object we have for sure, but I think this patch is a good candidate for Builder design pattern for example as one day we might want to import templates from different sources. I will not insist, but consider that for another moment.

lzap:
Ok that explains it. This is now ready for rebase and testing. I will try to cover the taxonomy thing just to be sure this is not a flaw.

app/models/template.rb
ares: keeping it here too as this is method API description, it's not clear what `option` can contain as it blindly pass it elsewhere, though this is the public facing method...

anyway good point and copied the info to API definition too

app/models/template.rb
ares: I agree that additional logic, such as getting templates data from somewhere should be kept out of this model, and if you look at foreman_templates we have separate service objects that does parsing git/fs and then use the `import` method from this model. Import should only understand attributes that this model has. Perhaps the metadata parsing could be separated later, but long term I'd like to make it new attribute of template so we don't need to keep it in template content. That's out of scope for this PR though.

ares:
Thanks @lzap, updated the PR, could you please take another look?

ares:
Tests should be green now, @lzap mind to take another look?

lzap:
@tstrachota I need a guidance for creating hammer part. All docuements and your slides cover REST CRUD but this is outside of that scope. So I added this to lib/hammer_cli_foreman/template.rb:

```ruby
      class ImportCommand < HammerCLIForeman::SingleResourceCommand
        
        action :import
        command_name "import"
        
        option '--name', "NAME", "Name of the template"
        option "--file", "TEMPLATE", _("Path to a file that contains the template"), :attribute_name => :option_template,
          :format => HammerCLI::Options::Normalizers::File.new
        
        success_message _("Provisioning template imported.")
        failure_message _("Could not import the provisioning template")
  
        build_options
      end
```

I can't get hammer to work, getting: Error: The server does not support such operation.

ares:
> I can't get hammer to work, getting: Error: The server does not support such operation.

@lzap, this means hammer does not see API description for this endpoint, after applying this PR you need to regenerate apipie documentation (json only would do), use `rake apipie:cache:index`. Then hammer could work but it also has caching so it does not download apipie index with every command so make sure you run with reload argument like this `hammer -r template ...`. These two things should help.

tstrachota:
@ares what's the difference between this PR and `foreman_templates`? Is this supposed to replace the plugin?

@lzap @ares note that there's https://github.com/ofedoren/hammer-cli-foreman-templates 
We should avoid name clash in hammer.

ares:
@tstrachota no, not at all, this only moves the import logic to the core but all logic about pulling/pushing from git/fs and advanced stuff like prefixing remains in the plugin. Plugin allows to do a bulk export and import. Plugin should later evolve into something that will also provide UI for these tasks. But since we already had exporting and metadata generating im core, it seems reasonable to have the import counterpart also in core.

tstrachota:
I see. Does it make sense to add the import command to `hammer-cli-foreman`? I'm probably missing something but if the only functionality is to create a template from file, then it's the same as `template create`.

ares:
The differnece is that the file can contaim metadata, which are used to assign objects like OSes during import. The new API enpoints also have few more params, that can be uses ti adjust metadata import mode. I think it should be separate hammer action, e.g. hammer ptable import

ares:
@lzap any update on this one?

lzap:
Doing this request:

```
curl -u admin:changeme -v --insecure -XPOST -H "Content-type: application/json" -d '{
"provisioning_template": {
"name": "Import Test",
"template": "<%#\nkind: script\nname: Grubby default\nmodel: ProvisioningTemplate\n%>\ntest"
}}' 'http://localhost:5000/api/v2/provisioning_templates/import'
```

and it's giving me Validation failed: Template kind can't be blank but I do set kind "script" there. Digging.

lzap:
Hmmm it does not work:

```shell
  curl -u admin:changeme -v --insecure -XPOST -H "Content-type: application/json" --data @- 'http://localhost:3000/api/v2/       +provisioning_templates/import' <<'EODATA'
  {
    "provisioning_template": {
      "name": "Import Test - Snippet",
      "template": "<%#\nkind: snippet\nname: Grubby default\nmodel: ProvisioningTemplate\n%>\ntest"
    }, "options": {
      "associate": "always"
    }
  }
  EODATA
```

lzap:
The reason for that is because `associate_metadata_on_import` returns false, it does not see the "associate" flag. I must be missing something or there is a bug which I am unable to find for now.

ares:
@lzap ok got it, probably caused by refactorings of params permitting, it used to ignored the whole hash. Though your example is also not entirely correct, snippet is not a kind. You either use kind, e.g. 'PXELinux', 'provision', 'finish' or you use `snippet: true` and don't specify kind. I know it's not ideal but that's what we have today in our schema.

I updated the params permitting and rebased. Please take another look. These two examples worked for me

kind script
```
curl -u admin:changeme -v --insecure -XPOST -H "Content-type: application/json" --data @- 'https://foreman.example.tst/api/v2/provisioning_templates/import' <<'EODATA'
  {
    "provisioning_template": {
      "name": "Import Test - Snippet",
      "template": "<%#\nkind: script\nname: Grubby default\nmodel: ProvisioningTemplate\n%>\ntest"
    }, "options": {
      "associate": "always"
    }
  }
EODATA

```

snippet:
```
curl -u admin:changeme -v --insecure -XPOST -H "Content-type: application/json" --data @- 'https://foreman.example.tst/api/v2/provisioning_templates/import' <<'EODATA'
  {
    "provisioning_template": {
      "name": "Import Test - Snippet",
      "template": "<%#\nsnippet: true\nname: Grubby default 2\nmodel: ProvisioningTemplate\n%>\ntest"
    }, "options": {
      "associate": "always"
    }
  }
EODATA
```

lzap:
Should this fail as well?

```
$ curl -u admin:changeme -v --insecure -XPOST -H "Content-type: application/json" --data @- 'http://localhost:3000/api/v2/provisioning_templates/import' <<'EODATA'
  {
    "provisioning_template": {
      "name": "Import Test - Never Script",
      "template": "<%#\nkind: script\nname: Grubby default\nmodel: ProvisioningTemplate\n%>\ntest"
    }, "options": {
      "associate": "never"
    }
  }
EODATA

{
  "error": {"message":"Validation failed: Template kind can't be blank"}
}
```

I observed that when I set "always" it works fine.

ares:
> Should this fail as well?

it should fail, with associate never we ignore metadata so we don't know what template type to create (unless the template already exists, in such case it should only update the content without error), is this the behavior you see?

lzap (approved):
I'm sold. Merging.

PR #5202, opened 2018-01-22
Fixes #22359 - Don't rely on default taxonomy in API for admin

ehelms:
@ares @iNecas -- Y'all may want to look at this given your involvement in the original code added and changed here

This is breaking the Katello pipeline. Katello seeds give the admin user a default organization if one is created through the installer (which by default it does). Tests then fail trying to look up the host created by puppet because it is attempting to look at the default organization of the admin and the initial host does not have one. This restores the behavior to the way it was previously for admin users.

### DIFF
--- a/app/controllers/concerns/application_shared.rb
+++ b/app/controllers/concerns/application_shared.rb
@@ -46,14 +46,14 @@ module ApplicationShared
               return false
             end
           end
-        elsif session.has_key?("#{taxonomy}_id") # session with taxonomy explicitly set to id or nil (any contexxt)
+        elsif session.has_key?("#{taxonomy}_id") # session with taxonomy explicitly set to id or nil (any context)
           if session["#{taxonomy}_id"].present?
             determined_taxonomy = find_session_taxonomy(taxonomy, user) # specific id
           else
             determined_taxonomy = nil # explicitly set any context
           end
-        else # user who didn't specify explicit context and does not have anything in session
-          determined_taxonomy = find_default_taxonomy(user, taxonomy)
+        else
+          determined_taxonomy = nil
         end
       else # UI request
         if session["#{taxonomy}_id"].present?
--- a/test/controllers/api/v2/domains_controller_test.rb
+++ b/test/controllers/api/v2/domains_controller_test.rb
@@ -264,7 +264,7 @@ class Api::V2::DomainsControllerTest < ActionController::TestCase
     test 'user does not have to specify taxonomy if he is assigned to only one, it is selected automatically' do
       domain = FactoryBot.build_stubbed(:domain)
       as_user @user do
-        post :create, params: { :organization_id => @org2.id, :domain => { :name => domain.name, :organization_ids => [@org2.id] } }
+        post :create, params: { :organization_id => @org2.id, :domain => { :name => domain.name, :organization_ids => [@org2.id], :location_ids => [@loc1.id] } }
       end
       assert_response :success
       domain = Domain.unscoped.find(JSON.parse(response.body)['id'])
@@ -285,7 +285,7 @@ class Api::V2::DomainsControllerTest < ActionController::TestCase
     test 'user can create domain in current context thanks to the fact organization_ids defaults to organization_id' do
       domain = FactoryBot.build_stubbed(:domain)
       as_user @user do
-        post :create, params: { :organization_id => @org2.id, :domain => { :name => domain.name } }
+        post :create, params: { :organization_id => @org2.id, :domain => { :name => domain.name, :location_ids => [@loc1.id] } }
       end
       assert_response :success
       domain = Domain.unscoped.find(JSON.parse(response.body)['id'])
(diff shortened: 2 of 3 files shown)


xprazak2:
test failure unrelated

mbacovsky:
@ehelms why do we set default organization for the admin user when it is ignored? Wouldn't it be better not to set it during the seed or disable default location for the admin user if we need it? 

Yesterday was merged last piece of hammer PRs to allow:
```
$ hammer host list --organization NIL
```
or
```
$ HAMMER_NIL=ANY hammer host list --organization ANY
```
which enforces ANY context for the request. In API it is already possible by adding `organization_id => nil` to the request params.

ares:
@ehelms I believe this is expected behavior. If you login to UI, admin's current org would also default to "Default Organization" and he/she would not be able to see the host from any context. If you want to override default organization in API request, you'd need to use what @mbacovsky suggests. My question though would be, why the host is created in any context? If Foreman is configured so that the "Default Organization" should be used by default, shouldn't the host also be created in it during the installer run?

ehelms:
First, I believe the UI and API have fundamental differences due to their different use cases. Setting a default organization, as I understood it, was intended to allow users to arrive in their organization when logging into the UI. Whereas, if I am using the API, I presume a lack of state. The default behavior of an API request to a GET should be without context unless context is provided. This was the previous behavior. This fundamentally changes the API's behavior. I have to now know that there is context. And knowing that is not obvious. In the UI, I can see a bar at the top that tells me what organization I am in. I know my scope. In the API, there is no indicator unless I provide parameters. This can lead to confusing results. I don't believe we should be changing this behavior or putting users in this in-explicit state. A call to `/api/hosts` should return me all hosts that I can see. If I request '/api/hosts/1` and have access to that host, I should get that result and not a 404. Consider as a user, or application programming against the API, that I am to know the 404 is a result of the default organization context and not as a result of the host actually missing.

To your second question, the host is created in 'Any Context' because its a puppet created host and that's the default landing place for them unless the user takes extra steps to further program the landing place. So the current test, tests the default scenario.

ehelms:
@mbacovsky The history behind the admin user having a default organization at the start is thus. Katello's scenario seeds an initial organization. Given Katello is quite multi-tenent focused with resources, having an initial organization to being doing work has made sense. Giving the admin user the initially created organization as their default ensures when they log in they are in the context of an organization from the outset. In a pure Foreman world, this does not make as much sense. In a Katello world, this makes a lot more sense given you need to be in an organization to interact with any of the Katello entities. So we can talk about whether this still makes sense as a workflow, but let's not let it detract from the behavior that exists regardless.

app/controllers/concerns/application_shared.rb
```diff
@@ -46,13 +46,13 @@ def set_taxonomy
...
           end
-        elsif session.has_key?("#{taxonomy}_id") # session with taxonomy explicitly set to id or nil (any contexxt)
+        elsif session.has_key?("#{taxonomy}_id") # session with taxonomy explicitly set to id or nil (any context)
           if session["#{taxonomy}_id"].present?
             determined_taxonomy = find_session_taxonomy(taxonomy, user) # specific id
           else
             determined_taxonomy = nil # explicitly set any context
           end
-        else # user who didn't specify explicit context and does not have anything in session
+        elsif !user.admin? # user who didn't specify explicit context and does not have anything in session
```
ares: if we're changing this behavior back, would it make sense to do it for admins and users? I think we should be consistent at least for all users

I think keeping the else and setting `determined_taxonomy` to `nil` would we better

ares (changes requested):
I'm not overly happy about the change but I understand it would have impact on users so should be properly discussed first. That means this change would need to be cherry-picked to 1.17 where the original change was introduced. cc @xprazak2 for visibility

ares:
Also the ticket number does not seem valid

ehelms:
This has been updated to reflect this for all users and not just the admin. I also updated the ticket number as I had swapped two numbers.

ekohl (approved):
This makes sense to me.

ares (approved):
APJ, tested locally and it works

test/controllers/api/v2/domains_controller_test.rb
```diff
@@ -413,7 +413,7 @@ def setup
       test 'user can create domain in default organization without need of specifying any organization' do
         domain = FactoryBot.build_stubbed(:domain)
         as_user @user do
-          post :create, params: { :domain => { :name => domain.name } }
+          post :create, params: { :domain => { :name => domain.name, :location_ids => [@loc1.org] } }
```
ares: I guess that should be `.id` instead of  `.org` everywhere, which is why jenkins fal

ares (changes requested):
tests failures related, please %s/.org/.id/

ehelms:
@ares Tests are passing now! Just waiting on the Travis job. Noted I did remove some tests because their descriptions no longer appeared to be true, supported behaviors.

ares (approved):

ares:
Yup, makes sense. Thanks also for updating commit message. Merging and will cherry-pick into 1.17-stable, @xprazak2 this should cause no problems, it's tiny change. The behavior was introduced in 1.17 only.

ares:
opened a CP PR since there was a conflict, see https://github.com/theforeman/foreman/pull/5217

PR #5204, opened 2018-01-23
Fixes #22379 - Remove pagination from csv export

tbrisker:


ohadlevy:
@tbrisker mind adding a test ? this my warrant a bigger change, as most controllers do use the  resource_base_search_and_page method...? maybe we should detect inside that method if we are using csv format so it works for all controllers?

tbrisker:
@ohadlevy added test, fixed issue in HG controller and opened https://github.com/mislav/will_paginate/pull/562 to fix the issue in will_paginate gem that throws an error when resetting limit on paginated scope.

xprazak2:
Works as expected, I was able to export all the facts, not just a page of 20.

ares (approved):
makes sense, works well, thanks @tbrisker @xprazak2 @ohadlevy

ares:
cherry-picked to 1.17-stable as https://github.com/theforeman/foreman/commit/44780a3cdb53e376a44fcc203e171c0db236d959

PR #5205, opened 2018-01-23
[WIP] Fixes #22381 - adds react-redux integration helper

amirfefer:


amirfefer:
@sharvit , @ohadlevy 
What do you think about this approach ?

I took inspiration from [this blog](https://medium.freecodecamp.org/real-integration-tests-with-react-redux-and-react-router-417125212638)

webpack/assets/javascripts/react_app/components/bookmarks/form/form.test.js
```diff
@@ -1,108 +1,65 @@
...
 
-    wrapper.find('form').simulate('submit');
-    expect(wrapper.find('.spinner').length).toEqual(0);
-    expect(spy).toHaveBeenCalledTimes(0);
-    spy.mockReset();
-    spy.mockRestore();
-  });
-  it('should allow creating a bookmark with a dot', () => {
-    const spy = jest.spyOn(FormActions, 'submitForm');
+describe('bookmark form - unit tests', () => {
```
sharvit, line 59: I believe it would be better to use separate file for unit-testing

webpack/assets/javascripts/react_app/IntergrationHelper.js
```diff
@@ -0,0 +1,19 @@
+import { applyMiddleware, combineReducers, createStore } from 'redux';
+import thunk from 'redux-thunk';
+
+export function flushAllPromises() {
+  return new Promise(resolve => setTimeout(() => resolve(), 100));
```
sharvit, line 5: Why don't you use the `setImmediate` like the post propose to?

`setImmediate` should get called when the javascript callstack is getting empty.
It is till risky, but somehow make sense when waiting for all the async actions to get resolved.

webpack/assets/javascripts/react_app/IntergrationHelper.js
```diff
@@ -0,0 +1,19 @@
+import { applyMiddleware, combineReducers, createStore } from 'redux';
+import thunk from 'redux-thunk';
+
+export function flushAllPromises() {
```
sharvit, line 4: Can you rename this function to `waitForAsyncActions()`?

webpack/assets/javascripts/react_app/components/bookmarks/form/form.test.js
```diff
@@ -1,108 +1,65 @@
...
-    expect(spy).toHaveBeenCalledWith({
-      item: 'Bookmark',
-      url: '/api/bookmarks',
-      values: {
-        controller: 'hosts', name: 'Joe', public: true, query: 'search',
-      },
+    await flushAllPromises();
+    expect(dispatchSpy).toHaveBeenCalledWith({
+      type: 'BOOKMARK_FORM_SUBMITTED',
+      payload: { item: 'Bookmark', body: { ok: true, id: 70 } },
```
sharvit, line 54: You can use snapshot here.
https://gist.github.com/sharvit/cf658a8494758be8062a63121b925fdc#file-bookmarksactions-test-js-L16

sharvit, line 54: What about a test against the store?
```js
expect(store.getState()).toMatchSnapshot();
```

sharvit (changes requested):
I like this approach.
I think it would even make more sense after we will migrate the folder structure to a domain level approach.

webpack/assets/javascripts/react_app/IntergrationHelper.js
sharvit: and use es6 syntax for functions :)

amirfefer:
This PR is implemented here:  https://github.com/theforeman/foreman/pull/5240

PR #5206, opened 2018-01-23
Fixes #22382 - Remove ldap_filter 255 chars limit

dLobatog:


ares:
LGTM, just rubocop is not happy

dLobatog (commented):
@ares Rubocop is happy now, the only error is the usual `Capybara::Poltergeist::StatusFailError: Request to 'http://127.0.0.1:43104/hosts' failed to reach server, check DNS and/or server status...`

ohadlevy:
[test foreman]

db/migrate/20180123140634_remove_limit_ldap_filter.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveLimitLdapFilter < ActiveRecord::Migration[5.1]
+  def change
+    change_column 'auth_sources', :ldap_filter, :text, :limit => nil
```
tbrisker: This migration is irreversible, you need to make an up/down migration (since it doesn't know what the old limit is)

tbrisker (changes requested):

dLobatog (commented):
@tbrisker Thanks for the review - now it has a `down` migration which reverts to string & 255 limit if needed

db/migrate/20180123140634_remove_limit_ldap_filter.rb
```diff
@@ -0,0 +1,9 @@
+class RemoveLimitLdapFilter < ActiveRecord::Migration[5.1]
+  def up
+    change_column 'auth_sources', :ldap_filter, :text, :limit => nil
+  end
+
+  def down
+    change_column 'auth_sources', :ldap_filter, :string, :limit => 255
```
tbrisker, line 7: The previous schema allowed `:null => true` as well

tbrisker (changes requested):

db/migrate/20180123140634_remove_limit_ldap_filter.rb
dLobatog: @tbrisker I think that's by default, `t.string "ldap_filter", limit: 255` - if you see schema.rb only `:null => false` is ever specified

dLobatog (commented):
Removing WoC https://github.com/theforeman/foreman/pull/5206/files#r164368621

tbrisker (approved):

tbrisker:
Thanks @dLobatog !

PR #5207, opened 2018-01-24
Fixes #22402 - Adds a config for docker and docker-compose

mmack:
It mounts the local working directory into the docker
container. So updating code should automatically reload
the app inside.

Brief Introduction:
Bootup with "docker-compose up -d"
DB Seed with "docker-compose exec web rake db:seed"
Logs with "docker-compose logs -f"

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,27 @@
+FROM ruby:2.4.3-alpine3.7
+RUN apk update && \
+    apk add --virtual build_deps \
+    build-base ruby-dev libc-dev linux-headers \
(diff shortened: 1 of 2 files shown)


xprazak2:
@ehelms, could you take a look please? 

@mmack, you may also be interested in containers that we have in [forklift](https://github.com/theforeman/forklift/tree/master/containers)

boaz0 (changes requested):
Trying to build the image, I get the following error:

COPY failed: stat /var/lib/docker/tmp/docker-builder668054030/config/settings.yaml: no such file or directory

PR #5208, opened 2018-01-24
Fixes #22403 - every API endpoint can set current context

ares:
@iNecas @mbacovsky this is what we discussed yesterday, this seems as most straightforward way to add this to every request definition. It also adds it to resources which are not taxable, e.g. architectures API, but that is consistent with context switcher on architectures page in UI.

### DIFF
--- a/app/controllers/api/v2/base_controller.rb
+++ b/app/controllers/api/v2/base_controller.rb
@@ -6,6 +6,8 @@ module Api
       resource_description do
         api_version "v2"
         app_info N_("Foreman API v2 is currently the default API version.")
+        param :location_id, Integer, :required => false, :desc => N_("Scope by locations") if SETTINGS[:locations_enabled]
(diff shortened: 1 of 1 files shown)


ares:
We could perhaps improve the texts to let users know that nil mean any context. I'm open to suggestions

dLobatog (changes requested):
I assume this is related to the mails where there's a discussion to remove the scoping by taxonomies by default on the API (all taxonomies selected by default) - should that be included here?

ares:
It's not really related, the API already works even without this definition, but it's not documented and hammer does not render these arguments. With this PR all hammer commands would have --organization-id and --location-ld to specify current context.

ares:
removing WoC, this is unrelated change as mentioned a month ago, could someone please take a look? the default scope for API was removed separately already

iNecas (approved):

iNecas:
Ack from me. @dLobatog any other objections?

dLobatog (approved):

dLobatog:
@iNecas none, merging - thanks @ares

PR #5209, opened 2018-01-24
Fixes #22389 - Current taxonomies in API response header

mbacovsky:


iNecas (commented):
mind adding some tests?

ehelms:
I think this is a solid change. I do think most people don't tend to look at header sense you have to go looking for them specifically.

How hard or crazy do you think it would be to add this same context into the actual API responses?

ares:
In such case, it should be called "current organization" but imho it does not belong to the api output. It's not resource attribute. Headers make sense to me.

iNecas:
+1 for headers over output

mbacovsky:
@ehelms, I'm not sure how difficult it would be to add the context to our json responses. We probably would have to patch our rabl library.
In the list requests it could be easy, we already have some metadata in the response. The rest may need some data format changes.

I prefer to keep the context in the headers as the change is uniform and unobtrusive for all the request types. I also don't think we need to be explicit about the context in API calls which are used mostly for automation. Having the information in the headers doesn't prevent us to be explicit in hammer if we need to.

app/controllers/api/base_controller.rb
```diff
@@ -233,6 +234,17 @@ def add_version_header
       response.headers["Foreman_api_version"]= api_version
     end
 
+    def add_taxonomies_header
+      if SETTINGS[:organizations_enabled]
+        current_org = "#{Organization.current.id}; #{Organization.current.name}" if Organization.current
+        response.headers["Current_organization"] = current_org || '; ANY'
```
tstrachota, line 240: How about `Foreman_current_organization`? The name clash is fairly minimal here, but I'd prefer to keep the `Foreman_` prefix here for the sake of consistency.

tstrachota (changes requested):
It makes sense to use headers instead of output for me.

@mbacovsky can you please add tests for the new headers? The base controller subclass test is probably the right place for them:
https://github.com/theforeman/foreman/blob/develop/test/controllers/api/base_controller_subclass_test.rb#L59

Also please add a redmine issue for the Foreman. The current one is for hammer.

tstrachota:
[test katello]

PR #5211, opened 2018-01-24
Fixes #19244 - added clean interfaces rake task

ShimShtein:
Added rake task that cleans up interfaces that match `ignored_interface_identifiers` setting.

@ares, do you think it would be enough for the users?

### DIFF
--- a/app/services/interface_cleaner.rb
+++ b/app/services/interface_cleaner.rb
@@ -1,0 +1,50 @@
+class InterfaceCleaner
+  include ActiveRecord::Sanitization
+
+  ESCAPE_CHAR = '\\'
+
+  def self.nic_pattern_to_like(nic)
+    sanitize_sql_like(nic, ESCAPE_CHAR).tr('*', '%')
+  end
+
+  def self.ignored_interface_like_patterns
+    Setting[:ignored_interface_identifiers].map { |pattern| nic_pattern_to_like(pattern) }
+  end
+
+  delegate :logger, :to => :Rails
+
+  attr_reader :primary_nics, :provision_nics, :primary_hosts, :provision_hosts
+
+  def clean!
+    @deleted_count = delete_excluded_nics
+
+    @primary_nics, @primary_hosts = excluded_nics.where(primary: true).pluck(:id, :host_id).transpose
+    @provision_nics, @provision_hosts = excluded_nics.where(provision: true).pluck(:id, :host_id).transpose
+    self
+  end
+
+  def deleted_count
+    raise 'cleaner has not cleaned anything yet, run #clean! first' if @deleted_count.nil?
+    @deleted_count
+  end
+
+  private
+
+  def excluded_nics
+    ignored_nic_patterns = InterfaceCleaner.ignored_interface_like_patterns
+
+    arel = Nic::Base.arel_table
+    chained_or = arel[:identifier].matches_any(ignored_nic_patterns, ESCAPE_CHAR, true)
+
+    Nic::Base.where(chained_or)
+  end
+
+  def delete_excluded_nics
+    nics = excluded_nics.where(primary: false, provision: false)
+
+    logger.debug "Removing excluded nics..."
+    deleted_nics = nics.delete_all
+    logger.debug "Removed #{deleted_nics} nic records."
+    deleted_nics
+  end
+end
--- a/lib/tasks/interfaces.rake
+++ b/lib/tasks/interfaces.rake
@@ -1,0 +1,39 @@
+namespace :interfaces do
+  desc <<-END_DESC
+Removes old interfaces that match ignored interfaces pattern setting.
+
+This is useful when you change the ignored pattern setting.
+
+Examples:
+  # foreman-rake interfaces:clean
+END_DESC
+
+  task :clean => :environment do
+    puts 'Starting ingnored interfaces clean up...'
+    cleaner = InterfaceCleaner.new.clean!
+    puts "Finished, cleaned #{cleaner.deleted_count} interfaces"
+    unless cleaner.primary_hosts.empty?
+      puts "Following hosts have ignored interface set as primary, please set other interface as primary and rerun the task:"
+      print_host_queries(cleaner.primary_hosts)
+    end
+    unless cleaner.provision_hosts.empty?
+      puts "Following hosts have ignored interface set as provision, please set other interface as provisioning and rerun the task:"
+      print_host_queries(cleaner.provision_hosts)
+    end
+  end
+
+  def print_host_queries(hosts)
+    Host.unscoped.where(id: hosts.uniq).pluck(:name).in_groups_of(hosts_group_count, false) do |names|
+      query = "name ^ (#{names.join(',')})"
+      puts "#{Setting[:foreman_url]}#{helper.hosts_url(only_path: true, search: query)}"
+    end
+  end
+
+  def hosts_group_count
+    @hosts_group_count ||= (ENV['MAX_HOSTS'] || '50').to_i
+  end
+
+  def helper
+    @helper ||= Rails.application.routes.url_helpers
+  end
+end
--- a/test/unit/tasks/interfaces_test.rb
+++ b/test/unit/tasks/interfaces_test.rb
@@ -1,0 +1,58 @@
+require 'test_helper'
+require 'rake'
+
+class InterfacesTest < ActiveSupport::TestCase
+  let(:null_store) { ActiveSupport::Cache.lookup_store(:null_store) }
+
+  setup do
+    Rake.application.rake_require 'tasks/interfaces'
+    Rake::Task.define_task(:environment)
+    Rake::Task['interfaces:clean'].reenable
+
+    @cleaner = InterfaceCleaner.new
+
+    InterfaceCleaner.stubs(:new).returns(@cleaner)
+  end
+
+  test 'interface:clean prints deleted count' do
+    @cleaner.stubs(:deleted_count).returns(10)
+
+    stdout, _stderr = capture_io do
+      Rake.application.invoke_task 'interfaces:clean'
+    end
+
+    assert_match /cleaned 10 interfaces/, stdout
+  end
+
+  test 'interface:clean warns about primary interface' do
+    Setting.stubs(:cache).returns(null_store)
+    Setting[:ignored_interface_identifiers] = ['ignored*']
+    host = FactoryBot.create(:host, interfaces: [FactoryBot.build(:nic_managed, identifier: 'ignored01', primary: true, provision: false)])
+
+    stdout, _stderr = capture_io do
+      Rake.application.invoke_task 'interfaces:clean'
+    end
+
+    assert_match /cleaned 0 interfaces/, stdout
+    encoded_hostname = URI.encode("(#{host.name})")
+    assert_match /#{encoded_hostname}/, stdout
+    assert_match /ignored interface set as primary/, stdout
+  end
+
+  test 'interface:clean warns about provision interface' do
+    Setting.stubs(:cache).returns(null_store)
+    Setting[:ignored_interface_identifiers] = ['ignored*']
+    host = FactoryBot.create(:host, interfaces: [FactoryBot.build(:nic_managed, identifier: 'ignored01', primary: false, provision: true)])
+
+    stdout, _stderr = capture_io do
+      Rake.application.invoke_task 'interfaces:clean'
+    end
+
+    assert_match /cleaned 0 interfaces/, stdout
+    encoded_hostname = URI.encode("(#{host.name})")
+    assert_match /#{encoded_hostname}/, stdout
+    query = URI.decode(stdout.match(/^.*search=(.*?%29)/)[1]).tr('+', ' ')
+    assert_equal host.id, Host.search_for(query).first.id
+    assert_match /ignored interface set as provision/, stdout
+  end
+end
--- a/test/unit/interface_cleaner_test.rb
+++ b/test/unit/interface_cleaner_test.rb
@@ -1,0 +1,64 @@
+require 'test_helper'
+
+class InterfaceCleanerTest < ActiveSupport::TestCase
+  let(:null_store) { ActiveSupport::Cache.lookup_store(:null_store) }
+  let(:cleaner) do
+    InterfaceCleaner.new
+  end
+
+  test "it cleans excluded interfaces" do
+    Setting.stubs(:cache).returns(null_store)
+    Setting[:ignored_interface_identifiers] = ['ignored*']
+
+    host = FactoryBot.create(:host, :managed)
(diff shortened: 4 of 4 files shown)


app/services/interface_cleaner.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  private
+
+  def find_excluded_nics
+    logger.debug "Searching for nics that match excluded pattern..."
+    nic_ids = []
+
+    ignored_interfaces_regex = FactParser.ignored_interface_identifiers_regex
+
+    Nic::Base.unscoped.find_in_batches(:batch_size => 100) do |nics_batch|
```
tbrisker: This will take a verrrrryyyy long time to iterate over all interfaces in a large set up. additionally, for every batch you iterate 5 more times on the matched interfaces to generate the id lists, which could also potentially require a large amount of memory.
The ignored syntax only allows `*` wildcard matching - which mean that you can search for matches using `LIKE` sql instead of matching a regex on each interface. How about something like:
```ruby
identifiers = Setting[:ignored_interface_identifiers].map{|i| i.tr('*','%')}
search = (["identifier LIKE ?"]*identifiers.length).join(" OR ")
Nic::Base.unscoped.where(search, *identifiers).select(:id, :host_id, :primary, :provision).find_each do |interface|
...
```

tbrisker (changes requested):

app/services/interface_cleaner.rb
lzap: I also think that affected systems has tens of millions of nic records and it's not good strategy to load them into memory this way. The key in being effective here is to do a single loop and delete them on the fly. That's basically what @tbrisker  suggest but I actually think that if we are going to filter them in one loop, it will actually be more efficient to do the filtering on the SQL-client side, because that where condition is for sure without index. Since we are talking about tens of millions of records, the query can eat significant amount of resources on the SQL-server side. Therefore I suggest to go even further to simple loop:

```ruby
Nic::Base.unscoped.select(:id, :host_id, :primary, :provision).find_each do |interface|
 # filter and delete them in this block - SQL server will do the transaction isolation for us it's safe to delete here
end

Comment (lzap) [reply app/services/interface_cleaner.rb]: Let me explain in more detail. Because we are using LIKE server cannot use indicies in this case. If we are iterating over simple table, table scan is actually not an issue if the query is unsorted for sure, but the problem here is at least one join. Optimizer can do weird things and can decide to join first, then we have huge problem. And I would not be surprised in bad optimization here because we will be submitting huge where clause here with many items. For this reason, I recommend to do simple loop to make sure this does not get de-optimized.

Iterating all records is sometimes faster and safer, I believe this is the case.

Edit: I just realized there is no join with `unscoped`, but we still have lots of LIKE queries. Not sure what optimizer can do bad in this case, regexp will be faster for sure. And PostgreSQL does support it, but won't be SQLite/MySQL compatible...

Comment (tbrisker) [reply app/services/interface_cleaner.rb]: I tested the query generated by my proposal on a DB with ~400k interfaces and 4k that match the patterns, and it ran in under one second:
```
foreman=# explain analyze SELECT  "nics"."id", "nics"."host_id", "nics"."primary", "nics"."provision" FROM "nics" WHERE (identifier LIKE 'lo' OR identifier LIKE 'usb%' OR identifier LIKE 'vnet%' OR identifier LIKE 'macvtap%' OR identifier LIKE '_vdsmdummy_' OR identifier LIKE 'veth%' OR identifier LIKE 'docker%' OR identifier LIKE 'tap%' OR identifier LIKE 'qbr%' OR identifier LIKE 'qvb%' OR identifier LIKE 'qvo%' OR identifier LIKE 'qr-%' OR identifier LIKE 'qg-%' OR identifier LIKE 'vlinuxbr%' OR identifier LIKE 'vovsbr%') ORDER BY "nics"."id" ASC LIMIT 1000;
                                                                                                                                                                           
             QUERY PLAN               
    
 Limit  (cost=9736.42..9737.99 rows=629 width=10) (actual time=258.907..259.025 rows=1000 loops=1)
   ->  Sort  (cost=9736.42..9737.99 rows=629 width=10) (actual time=258.905..258.974 rows=1000 loops=1)
         Sort Key: id
         Sort Method: top-N heapsort  Memory: 95kB
         ->  Seq Scan on nics  (cost=0.00..9707.18 rows=629 width=10) (actual time=0.038..257.565 rows=4208 loops=1)
               Filter: (((identifier)::text ~~ 'lo'::text) OR ((identifier)::text ~~ 'usb%'::text) OR ((identifier)::text ~~ 'vnet%'::text) OR ((identifier)::text ~~ 'macvtap%'::text) OR ((identifier)::text ~~ '_vdsmdummy_'::text) OR ((identifier)::text ~~ 'veth%'::text) OR ((identifier)::text ~~ 'docker%'::text) OR ((identifier)::text ~~ 'tap%'::text) OR ((identifier)::text ~~ 'qbr%'::text) OR ((identifier)::text ~~ 'qvb%'::text) OR ((identifier)::text ~~ 'qvo%'::text) OR ((identifier)::text ~~ 'qr-%'::text) OR ((identifier)::text ~~ 'qg-%'::text) OR ((identifier)::text ~~ 'vlinuxbr%'::text) OR ((identifier)::text ~~ 'vovsbr%'::text))
               Rows Removed by Filter: 373917
 Planning time: 0.120 ms
 Execution time: 259.064 ms
```
I think this is much better then instantiating 400k activerecord objects just to filter them out.

Comment (lzap) [reply app/services/interface_cleaner.rb]: That sort/order by must be removed, although it looks like harmless (index) I'd rather not see it here. Anyway, can you test with find_each just to be sure Rails does not do some terrible magic trick there with WHERE 1=1 or anything like that. Just batch print them. I only expect `LIMIT` tho.

Comment (tbrisker) [reply app/services/interface_cleaner.rb]: the `sort` is caused by the `find_where` afaik, it needs to make sure it gets the batches in the correct order

Comment (tbrisker) [reply app/services/interface_cleaner.rb]: well actually, it uses the id sort for further filtering on next batches:
```
Nic::Base Load (36.5ms)  SELECT  "nics"."id", "nics"."host_id", "nics"."primary", "nics"."provision" FROM "nics" WHERE (identifier LIKE 'lo' OR identifier LIKE 'usb%' OR identifier LIKE 'vnet%' OR identifier LIKE 'macvtap%' OR identifier LIKE '_vdsmdummy_' OR identifier LIKE 'veth%' OR identifier LIKE 'docker%' OR identifier LIKE 'tap%' OR identifier LIKE 'qbr%' OR identifier LIKE 'qvb%' OR identifier LIKE 'qvo%' OR identifier LIKE 'qr-%' OR identifier LIKE 'qg-%' OR identifier LIKE 'vlinuxbr%' OR identifier LIKE 'vovsbr%') AND ("nics"."id" > 323241) ORDER BY "nics"."id" ASC LIMIT $1  [["LIMIT", 1000]]
```

Comment (lzap) [reply app/services/interface_cleaner.rb]: Ok, I am sold!

**Comment by ShimShtein** [comment]
@tbrisker, @lzap Finally! The tests are green, please review this now.

**Review by boaz0 — APPROVED** [review]
LGTM 👍

File: app/services/interface_cleaner.rb [hunk 2]
```diff
@@ -0,0 +1,51 @@
...
+
+    arel = Nic::Base.arel_table
+    like_predicates = ignored_nic_patterns.map { |pattern| arel[:identifier].matches(pattern, InterfaceCleaner::ESCAPE_CHAR, true) }
+    chained_or = like_predicates.reduce(arel.create_false, :or)
+
+    Nic::Base.where(chained_or)
+  end
+
+  def delete_excluded_nics
+    nics = excluded_nics.where(primary: false, provision: false)
```
lzap, line 43: I don't understand why you load all records into memory while you can do `delete_all`. I know this is one-time code, but since we have this in core, it possible can be called multiple times in the future.

app/services/interface_cleaner.rb
```diff
@@ -0,0 +1,51 @@
...
+    sanitize_sql_like(nic, ESCAPE_CHAR).tr('*', '%')
+  end
+
+  def self.ignored_interface_like_patterns
+    Setting[:ignored_interface_identifiers].map { |pattern| nic_pattern_to_like(pattern) }
+  end
+
+  delegate :logger, :to => :Rails
+
+  attr_reader :primary_nics, :provision_nics, :primary_hosts, :provision_hosts
```
lzap, line 16: Is my understanding that these are only for testing purposes?

app/services/interface_cleaner.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def clean!
+    @deleted_count = delete_excluded_nics
+
+    @primary_nics, @primary_hosts = excluded_nics.where(primary: true).pluck(:id, :host_id).transpose
+    @provision_nics, @provision_hosts = excluded_nics.where(provision: true).pluck(:id, :host_id).transpose
```
lzap, line 22: Can you make this a memorized class variable via reader methods rather than creating this every time `clean!` is performed? Or even just attribute method without memorizing, if that's really only for testing.

app/services/interface_cleaner.rb
ShimShtein: I am not loading records into memory here. The object in `nics` is `ActiveRecord::Relation` which is only a definition of a query. I can call multiple methods on that object that _could_ cause loading objects into memory: `all` will definitely load everything to memory, `first` will load only one record. In this particular case, I am calling to `delete_all` two lines later. `delete_all` will only return the amount of records deleted.

app/services/interface_cleaner.rb
ShimShtein: No, these lists are for the user - ids in these lists are interfaces that match the ignored pattern, but were not removed, since they are marked as primary and provision. I also show the hosts that have such interfaces.

app/services/interface_cleaner.rb
ShimShtein: This is done after the main deletion so I don't expect too much records here, but if you insist I can make those properties lazy loaded.

app/services/fact_parser.rb
```diff
@@ -18,6 +18,10 @@ def self.register_fact_parser(key, klass, default = false)
     parsers[key.to_sym] = klass
   end
 
+  def self.ignored_interface_identifiers_regex
```
tbrisker: I think this change isn't needed any more?

lib/tasks/interfaces.rake
```diff
@@ -0,0 +1,39 @@
...
+    end
+    unless cleaner.provision_hosts.empty?
+      puts "Following hosts have ignored interface set as provision, please set other interface as provisioning and rerun the task:"
+      print_host_queries(cleaner.provision_hosts)
+    end
+  end
+
+  def print_host_queries(hosts)
+    Host.unscoped.where(id: hosts.uniq).pluck(:name).in_groups_of(hosts_group_count, false) do |names|
+      query = "id ^ (#{names.join(',')})"
```
lzap: Here is one thing - you join names but query ids, which does not work: `http://box.home.lan/hosts?search=id+%5E+%28carl-magpusao.home.lan%29` does not find it. You need to pluck :id here.

app/services/interface_cleaner.rb
lzap: I mis-read it, got it.

app/services/interface_cleaner.rb
lzap: That's fine.

app/services/interface_cleaner.rb
lzap: It looks like nics are only used in tests, hosts are used in rake task as well. I am okay with this.

lzap (changes requested):
One small thing.

app/services/interface_cleaner.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  private
+
+  def excluded_nics
+    ignored_nic_patterns = InterfaceCleaner.ignored_interface_like_patterns
+
+    arel = Nic::Base.arel_table
+    like_predicates = ignored_nic_patterns.map { |pattern| arel[:identifier].matches(pattern, InterfaceCleaner::ESCAPE_CHAR, true) }
+    chained_or = like_predicates.reduce(arel.create_false, :or)
```
tbrisker: You can user arel's `.matches_any` instead of building the chain yourself

app/services/interface_cleaner.rb
```diff
@@ -0,0 +1,51 @@
...
+    @deleted_count
+  end
+
+  private
+
+  def excluded_nics
+    ignored_nic_patterns = InterfaceCleaner.ignored_interface_like_patterns
+
+    arel = Nic::Base.arel_table
+    like_predicates = ignored_nic_patterns.map { |pattern| arel[:identifier].matches(pattern, InterfaceCleaner::ESCAPE_CHAR, true) }
```
tbrisker: nitpick, i think you can use just `ESCAPE_CHAR` here

lib/tasks/interfaces.rake
tbrisker: Hosts don't allow search on id

lib/tasks/interfaces.rake
```diff
@@ -0,0 +1,39 @@
...
+      puts "#{Setting[:foreman_url]}#{helper.hosts_url(only_path: true, search: query)}"
+    end
+  end
+
+  def hosts_group_count
+    @hosts_group_count ||= (ENV['MAX_HOSTS'] || '50').to_i
```
tbrisker, line 33: i'm not sure this needs to be a customizable option, just set to a sane number that won't make the query too long. Perhaps even not a query at all but just a list of hosts (since it souldn't be more than a handful at most)

tbrisker (changes requested):
Looks good in general, a few minor comments inline

ShimShtein:
@tbrisker @lzap Addressed all your comments.

lib/tasks/interfaces.rake
ShimShtein: There are different URL size limitations, and adding this options is not costly enough that we should consider it harmful. I think it worth adding it, so if someone encounters a problem with URI length, it can be easily fixed without code patching.

tbrisker (approved):
LGTM, thanks @ShimShtein !

lzap (approved):
Thanks!

test/unit/tasks/interfaces_test.rb
```diff
@@ -0,0 +1,58 @@
+require 'test_helper'
+require 'rake'
+
+class InterfacesTest < ActiveSupport::TestCase
+  let(:null_store) { ActiveSupport::Cache.lookup_store(:null_store) }
```
lzap, line 5: I am looking into Rails cache in test environment and I wonder - what is purpose of this null store hack? We seem to clean cache before and after each test.

PR #5188, opened 2018-01-18
Fixes #21287 - add resource switcher

amirfefer:

PR #5212, opened 2018-01-25
[WIP] Fixes #21139 - Migrate css stack to webpack

sharvit:
* migrate most of the css stack to webpack
* allow scss files to easily import variables and mixins
* migrate bootstrap.js and patternfly.js to webpack
* migrate fonts to webpack
* refactor the scss file structor


webpack/assets/stylesheets/application/partials/_base.scss
```diff
@@ -105,12 +105,12 @@ a.config_group_name {
 }
 
 .editable {
-  background: image-url('edit.png') no-repeat scroll 98% 6px transparent;
+  background: url('#{$local-img-path}/edit.png') no-repeat scroll 98% 6px transparent;
```
ohadlevy: we can probably remove this file and change it to an icon (looks like a leftover)

webpack/assets/stylesheets/application/partials/_base.scss
```diff
@@ -105,12 +105,12 @@ a.config_group_name {
...
   cursor: pointer;
   padding: 4px 26px 4px 0;
 
   &:hover {
-    background: image-url('edit-hover.png') no-repeat scroll 98% 6px #f2f2f2;
+    background: url('#{$local-img-path}/edit-hover.png') no-repeat scroll 98% 6px #f2f2f2;
```
ohadlevy: same for this file, should be replaced with an icon.

webpack/assets/stylesheets/application/partials/_charts.scss
```diff
@@ -131,7 +131,7 @@
 }
 
 .statistics-pie.small .overlay {
-  background: image-url('pie_overlay.png') no-repeat;
+  background: url('#{$local-img-path}/pie_overlay.png') no-repeat;
```
ohadlevy: I believe this is not in used anymore. I didn't find a code reference to the overlay css class.

webpack/assets/stylesheets/application/partials/_navigation.scss
```diff
@@ -63,7 +62,7 @@
 }
 
 .navbar-pf-vertical {
-  background: image-url('navbar.png');
+  background: url('#{$local-img-path}/navbar.png');
```
ohadlevy: this will break the theme-ability? /cc @ShimShtein ?

webpack/assets/stylesheets/application/variables/_colors.scss
```diff
@@ -1,7 +1,6 @@
-// Colors
-$primary_color: #005c7e;
-$border_color: #0187b6;
-$brand_color: #005c7e;
```
ohadlevy: I assume this will break theme plugin too?

ohadlevy (commented):
awesome work @sharvit ! a few inline comments

ohadlevy:
there are a few more files that most likely can be moved, for example:
* operating system icons
* jquery-ui
* favicon

I think that there are a few unused assets as well:
* public/assets/spinner
* assets/sort_*

you can remove your public/assets content and ```RAILS_ENV=production rake assets:precompile``` to see the entire list

webpack/assets/stylesheets/application/variables/_colors.scss
ShimShtein: It shouldn't break the theme, we didn't have any good solution for `scss` anyway.

webpack/assets/stylesheets/application/partials/_navigation.scss
ShimShtein: The only thing that I can think of, is the changes to folder structure. The theme should reflect it for sure. Basically the same rule will apply for the new path too  - as long as the file name is the same, it will be taken from the theme.

By the way, if webpack generates file hashes like rails - it should generate the same hash as for the main foreman.

webpack/assets/stylesheets/application/partials/_base.scss
sharvit: Won't be that trivial, check out:
https://github.com/theforeman/foreman/blob/develop/app/views/settings/index.html.erb#L30
https://github.com/theforeman/foreman/blob/develop/app/helpers/settings_helper.rb#L2-L11

sharvit:
Thanks @ohadlevy 

> operating system icons

I prefer to keep using images only from inside webpack and not to make webpack-entries for images.
It is currently used here:
https://github.com/theforeman/foreman/blob/develop/app/helpers/operatingsystems_helper.rb#L49-L51
https://github.com/theforeman/foreman/blob/develop/app/views/operatingsystems/index.html.erb#L16:73


> jquery-ui

Done

> favicon

In a different pr please.
I plan to use [favicons-webpack-plugin](https://github.com/jantimon/favicons-webpack-plugin) which based on [favicons](https://github.com/evilebottnawi/favicons).
```js
plugins: [
  new FaviconsWebpackPlugin('my-logo.png')
]
```
This basic configuration will generate 37 different icons for iOS devices, Android devices and the Desktop browser.
It can also generate a JSON file with all information about the icons,
so we can create a ruby script that read this file and renders the icons into the html layout.

> I think that there are a few unused assets as well:
> public/assets/spinner

This spinner is getting into the `public/assets` because it's apart of the `scoped_search` gem.

> assets/sort_*

Done

sharvit:
I just noticed I made a mess with the storybook, working on a fix.

sharvit:
Fixed the storybook by adding the `extract-text-webpack-plugin`

sharvit:
https://github.com/theforeman/foreman/pull/5227 is fixing the Travis error.

sharvit:
Rebased

package.json
```diff
@@ -57,6 +58,7 @@
   "dependencies": {
     "axios": "^0.17.1",
     "babel-polyfill": "^6.26.0",
+    "bootstrap": "^3.3.7",
```
ohadlevy: isnt this already pulled via patternfly?

ohadlevy:
@ares can you verify your notifications scroll issue is present after applying this patch? thanks

package.json
sharvit: No, `patternfly-react` pulling `react-bootstrap` (which isn't pulling `bootstrap`).
We can't remove `bootstrap` but we can remove `react-bootstrap` :+1:

sharvit: My bad, `patternfly-react` -> `patternfly` -> `bootstrap`.

I should be able to remove `bootstrap` and `font-awsome` as well.

sharvit:
@ohadlevy - I removed `bootstrap` and `react-bootstrap` from the `package.json`.

ohadlevy:
@sharvit whats the status of the PR? thanks

sharvit:
@ohadlevy Basically it is ready to get merged.

I have some more local changes that:
1. Migrate partials to be more based on variables
2. Allow plugins to override variables

If it will get merged I can just create a new PR for those local changes.
I'm fine with both.

Another issue,
It will be easier for me to do rebase if needed.
@amirfefer for example, have #5188 that need a rebase from my side (or @amirfefer)

ohadlevy:
in the logs I'm seeing:
```
vendor-2523d02….js:29 Refused to load the font 'data:application/font-woff2;base64,d09GMgABAAAAAEZsAA8AAAAAsVwAAEYJAAECTQAA…CJdYQqpZRSKkpefYAeFMQ6TZS0JEmSJEnSDkaCi5mZmZl50Z+e+97zwF9Xzcb9PEc8/gMAAA==' because it violates the following Content Security Policy directive: "default-src 'self'". Note that 'font-src' was not explicitly set, so 'default-src' is used as a fallback.
```

sharvit:
@ohadlevy It is really weird, this is the reason I added `font_src` to the security policy.
https://github.com/theforeman/foreman/pull/5212/files#diff-55c5b7aecfb519d0e4880eaf2788eb6e

When canceling this change, I do notice it in the log and more imported, icons are not showing up.

Can you think of any reason why adding the security policy haven't affected you?
Do icons showing up (vertical-nav, top-bar)?

**I just rebased on develop and #5273 to let travis run**

sharvit:
Rebased @ohadlevy

sharvit:
@ohadlevy my bad, added a csp to the production env should fix it now:
https://github.com/theforeman/foreman/pull/5212/files#diff-4e3acad51681031aa52403514a5f6d3a

package.json
```diff
@@ -66,16 +67,16 @@
     "isomorphic-fetch": "^2.2.1",
     "jquery": "~2.2.4",
     "jquery-flot": "~0.8.3",
+    "jquery-ui": "~1.10.4",
```
ekohl, line 71: This needs packaging before being merged.

ekohl (commented):
Are there more changes to be expected around packaging or shall I prepare a PR?

ohadlevy:
Nothing afaiu, please go ahead

app/assets/stylesheets/application.scss
```diff
@@ -1,13 +1,3 @@
-@import 'jquery.ui.theme';
-@import 'jquery.ui.core';
-@import 'jquery.ui.menu';
-//we need to import all the above because we are using jquery-ui-rails and it uses sprockets require instaed of sass import.
-@import 'jquery.ui.autocomplete';
-@import 'jquery.ui.spinner';
-@import 'jquery.ui.spinner_custom';
+@import 'jquery.gridster';
```
ohadlevy, line 1: does this mean we now load gridster on all pages? https://github.com/theforeman/foreman/pull/5212/files#diff-e2cda3aede73b114b7655fdf896b7385R12 ?

webpack/assets/stylesheets/application/partials/_password_strength.scss
```diff
@@ -0,0 +1,11 @@
+#password .help-block {
+  .progress .progress-bar .password-verdict {
+    color: #000;
```
ohadlevy, line 3: is there a color-pf for this?

ohadlevy (dismissed):
overall looks great ! thanks @sharvit moving away to consume patternfly as an npm never looked easier ;)

besides minor questions, this is ready to merge.

@ekohl I'll wait for your ack prior to merging.
One side effect of this PR, is that it would be harder to use the assets bundle group, as you now must have webpack as well in order to run css, maybe we should consider dropping it?

ohadlevy:
[test foreman]
[test katello]

ohadlevy:
@sharvit ah, please note that the katello failures are relevant. /cc @waldenraines

app/assets/stylesheets/application.scss
sharvit: Didn't we load it on all pages before?
https://github.com/theforeman/foreman/pull/5212/files#diff-f859e2848b07324f808ce1a5ccd9fcbaL13
https://github.com/theforeman/foreman/pull/5212/files#diff-e9457d01ba037a2e0005c43d84bd656eL1

webpack/assets/stylesheets/application/partials/_password_strength.scss
sharvit: Not really, the **blackest** color-pf is:
```scss
$color-pf-black: #030303 !default;
```

For me, it makes sense to use the `$color-pf-black` as our black :+1:
If we decide to take this direction, there are more places we with `#000` that we can migrate.
(prefer to do in other pr)

Any thoughts?

ohadlevy: I'm ok with a follow up pr - thanks

app/assets/stylesheets/application.scss
ohadlevy: OK, thanks for the clarification.

ohadlevy:
@sharvit tests are still failing, setting for WoC until resolved.

sharvit:
Still WOC!

sharvit:
Still WOC

ekohl:
@sharvit it might be good to rename it back to ```[WIP]``` if you're still working on it.

waldenraines:
@sharvit any updates on this?  I believe this would fix a number of styling issues we are seeing on some of our new pages that are using patternfly-react.

ohadlevy:
@sharvit is now a good time to restart this effort?

ohadlevy:
@sharvit what is the status? does it make sense to upgrade to webpack 4 before this pr? thanks.

sharvit:
@ohadlevy I believe yes 👍

ohadlevy:
@sharvit whats the plan on this one? thanks

sharvit:
@ohadlevy sorry but i didn't had the time to fully focus on this task.

Currently the only blocker is katello using bastion using the patternfly gem (If i remember well enough).
So there are some changes need to do in bastion so it won't use the patternfly gem anymore.

waldenraines:
@sharvit what is the status of this?

sharvit:
@waldenraines I couldn't find the time to finish it for a long time but I see big potential from this PR and would love to continue the effort from where I stopped.

PR #2109, opened 2015-01-26
fixes #5125 - showing available images for vmware compute resources

unorthodoxgeek:

PR #5214, opened 2018-01-25
fixes #5125 - list available vmware images

timogoebel:
image


dLobatog (changes requested):
@timogoebel Thanks! Works fine. For reference, the difference between this and https://github.com/theforeman/foreman/pull/2109/files#diff-4443c0de1e4e867a30d3c878b9cb64e9R98 is that this PR takes into account the datacenter used and only returns templates for that datacenter.

dLobatog (approved):

PR #5215, opened 2018-01-25
Fixes #22425 - generate selinux interfaces once in debug

lzap:
This is speedup patch, saves about 13 seconds on my system.


dLobatog (approved):
@lzap Looks nice - for future reference, all that's going on here is that `sepolgen-ifgen` is only called now if `/var/lib/sepolgen/interface_info` does not exist. Prior to the patch it was called every time.

PR #5216, opened 2018-01-25
Fixes #22426 - removed plugin list from debug

lzap:
It's unnecessary command, we have list of plugins in two other files: packages_list and scl_gems_list. I will remove that. It is very slow and also it gets stucked on my system for unknown reason.


dLobatog (approved):

dLobatog:
Thanks @lzap

lzap:
Thanks, another 15 seconds saved on my test box. Debug is getting much faster.

PR #5224, opened 2018-01-30
Fixes #22455 - Provide a fallback humanized_name

dLobatog:
Jobs defined in Foreman should have a humanized_name in order to show up
in the Tasks list. However, in case they don't (plugins, or just
something that slipped under the radar during review), an empty line
is displayed on this page. Instead, at least the class name should be
displayed so that it makes more sense.


tbrisker (approved):
Makes sense, Thanks @dLobatog !

PR #5225, opened 2018-01-30
fixes #22444 - do not serve templates with expired token

timogoebel:
This is (in the author's opinion) an important fix for the unattended controller. Please see the Redmine issue for reasoning.

The commit changes the controller's behavior like this:
1. When a token is sent to the controller, the host is not looked up via it's IP or Mac.
2. When no token is sent to the controller and the host is found via IP or Mac, the token is checked for validity.


lzap (approved):
Thanks Timo, indeed this was very loose and it's good to tighten this.

ares:
thanks for merging, please don't forget to set the release in redmine, I've set it to 1.18

lzap:
:man_facepalming:

PR #5226, opened 2018-01-30
Refs #21871 - Test resource_scope using existing tables

dLobatog:
This patch is needed in order to run tests with plugins that pass. The
'after_run' method gets called after test rake tasks. In Jenkins,
when plugins are called with 'rake jenkins:unit jenkins:integration'
there's a problem - the 'testables' table gets removed before the
tests run, so these tests fail.

Changes here follow the same strategy we have followed on other
patches - just reuse existing tables to test 'generic' objects.

Example of the failures: 
- tasks http://ci.theforeman.org/job/test_plugin_pull_request/4558/#showFailuresLink
- ansible http://ci.theforeman.org/job/test_plugin_matrix/4613/
Testing foreman-tasks with the patch: http://ci.theforeman.org/job/test_plugin_pull_request/4560 
Testing foreman-ansible with the patch: http://ci.theforeman.org/job/test_plugin_matrix/4626 (seems OK, the error is different - tests passed if you inspect each run)

cc @xprazak2 @iNecas

### DIFF
--- a/test/controllers/api/base_controller_subclass_test.rb
+++ b/test/controllers/api/base_controller_subclass_test.rb
@@ -44,23 +44,17 @@ end
 class Testable < ApplicationRecord
   include Authorizable
   belongs_to :domain
-  belongs_to :foo
+  belongs_to :foo, :foreign_key => 'role_id'
+  self.table_name = 'filters'
 end
 
 class Foo < ApplicationRecord
+  self.table_name = 'roles'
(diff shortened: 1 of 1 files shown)


iNecas (approved):
The jenkins jobs prove this fixes the issue.

iNecas:
Thanks @dLobatog

PR #5228, opened 2018-01-31
fixes #22467 - dashboard: list hosts with expired token

timogoebel:
image
image

### DIFF
--- a/app/helpers/dashboard_helper.rb
+++ b/app/helpers/dashboard_helper.rb
@@ -159,6 +159,26 @@ module DashboardHelper
     setting.nil? ? true : !setting
   end
 
+  def host_build_status_icon(host)
+    if host.token_expired?
+      icon = 'hourglass-o'
+      icon_kind = 'fa'
+      icon_class = 'text-danger'
+      label = _('Token expired')
+    elsif host.build_errors.present?
+      icon = 'error-circle-o'
+      icon_kind = 'pficon'
+      icon_class = ''
+      label = _('Build errors')
+    else
+      icon = 'in-progress'
+      icon_kind = 'pficon'
+      icon_class = ''
+      label = _('Build in progress')
+    end
+    icon_text(icon, '', kind: icon_kind, title: label, class: icon_class)
+  end
+
   private
 
   def origin_setting(origin, name)
--- a/app/helpers/hosts_helper.rb
+++ b/app/helpers/hosts_helper.rb
@@ -251,7 +251,7 @@ module HostsHelper
   end
 
   def build_duration(host)
-    return _("N/A") if host.initiated_at.nil? && host.installed_at.nil?
+    return _('N/A') if host.initiated_at.nil? || host.installed_at.nil?
     if host.installed_at.nil?
       time_ago_in_words(host.initiated_at, include_seconds: true) + " (in progress)"
     else
--- a/app/views/dashboard/_hosts_in_build_mode_widget.html.erb
+++ b/app/views/dashboard/_hosts_in_build_mode_widget.html.erb
@@ -1,0 +1,11 @@
+<% hosts = @data.hosts.in_build_mode.or(@data.hosts.with_build_errors).includes(:token).preload(:owner).order(created_at: :desc).limit(9) %>
+
+<h4 class="header">
+  <%= _('Hosts in Build mode or with Build errors') %>
+</h4>
+
+<% if hosts.empty? %>
+  <p class="ca"><%= icon_text('ok', '', kind: 'pficon') %></span><%= _('No Hosts are in build mode or have build errors.') %></p>
+<% else %>
+  <%= render :partial => 'hosts_in_build_mode_widget_host_list', :locals => { :hosts => hosts } %>
+<% end %>
--- a/app/views/dashboard/_hosts_in_build_mode_widget_host_list.html.erb
+++ b/app/views/dashboard/_hosts_in_build_mode_widget_host_list.html.erb
@@ -1,0 +1,18 @@
+<table class="<%= table_css_classes('table-fixed') %>">
+  <thead>
+    <th width="40%"><%= _('Host') %></th>
+    <th><%= _('Owner') %></th>
+    <th><%= _('Build Duration') %></th>
+    <th><%= _('Token Expiry') %></th>
+  </thead>
+  <tbody>
+    <% hosts.each do |host| %>
+      <tr>
+        <td class="ellipsis"><%= host_build_status_icon(host) %> <%= link_to host.name, host_path(host) %></td>
+        <td class="hidden-tablet hidden-xs ellipsis"><%= host.owner %></td>
+        <td class="hidden-tablet hidden-xs ellipsis"><%= build_duration(host) %></td>
+        <td class="hidden-tablet hidden-xs ellipsis"><%= host.token ? date_time_relative_value(host.token.expires) : _('N/A') %></td>
+      </tr>
+    <% end %>
+  </tbody>
+</table>
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -156,7 +156,7 @@ Lint/Void:
 
 # Offense count: 604
 Metrics/AbcSize:
-  Max: 128
+  Max: 132
 
 # Offense count: 345
 # Configuration parameters: CountComments, ExcludedMethods.
@@ -191,7 +191,7 @@ Metrics/MethodLength:
 # Offense count: 25
 # Configuration parameters: CountComments.
 Metrics/ModuleLength:
-  Max: 441
+  Max: 446
 
 # Offense count: 10
 # Configuration parameters: CountKeywordArgs.
--- a/app/models/host/managed.rb
+++ b/app/models/host/managed.rb
@@ -195,6 +195,9 @@ class Host::Managed < Host::Base
 
   scope :with_compute_resource, -> { where.not(:compute_resource_id => nil, :uuid => nil) }
 
+  scope :in_build_mode, -> { where(build: true) }
+  scope :with_build_errors, -> { where.not(build_errors: nil) }
+
   # some shortcuts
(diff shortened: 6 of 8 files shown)


app/views/dashboard/_hosts_expired_token_widget_host_list.html.erb
```diff
@@ -0,0 +1,18 @@
...
+    <th><%= _('Owner') %></th>
+    <th><%= _('Created') %></th>
+    <th><%= _('Token Expired') %></th>
+  </thead>
+  <tbody>
+    <% hosts.each do |host| %>
+      <tr>
+        <td class="ellipsis"><%= icon_text('error-circle-o', '', kind: 'pficon') %> <%= link_to host.name, host_path(host) %></td>
+        <td class="hidden-tablet hidden-xs ellipsis"><%= host.owner %></td>
+        <td class="hidden-tablet hidden-xs ellipsis"><%= date_time_relative_value(host.created_at) %></td>
```
sean797: Would getting a value for "Built" be better then created_at? Just created_at wouldn't be relevant to hosts that have been rebuilt.

You could use Setting[:token_duration] and `host.token.expires` to work it out. Obviously that would be incorrect if someone changes Setting[:token_duration] though :(

ShimShtein:
Nice! Can we also add a scoped search definition for `expired?`?

dLobatog (dismissed):
Setting to WoC based on comments

app/views/dashboard/_hosts_expired_token_widget_host_list.html.erb
```diff
@@ -0,0 +1,18 @@
+<table class="table table-striped table-fixed">
```
ohadlevy: should we use the table css classes helper here?

timogoebel:
I changed the code to use the helper and removed the created at column in the table for now. I think, we should store the time a host is set to built in the database. But I want to do that in a separate PR. I'll also create an issue to add `expired?` to scoped search. This is out of scope here (although a good idea).

Anything else to add?

This is how it looks now:
image

timogoebel:
@lzap actually added the required fields to show the build start time in #5292. I'll wait for the PR to get in before I continue to work on this.

lzap:
Nice patch, I'd also like to see list of hosts which are actually in build mode. Do you mind renaming the feature and widget to show both - there could be a filter or button to only show expired ones. I don't know how others, but my deployments have usually just few hosts in build mode.

timogoebel:
> I'd also like to see list of hosts which are actually in build mode.

Nice idea, I'll look into it.

timogoebel:
Updated. This is how it currently looks.
@lzap: Do you mind taking a look? Should be an easy review. :-)

image

ohadlevy:
@timogoebel last screenshot is very nice :) would you consider having its own page for it (vs a dashboard widget?)

timogoebel:
> would you consider having its own page for it (vs a dashboard widget?)

I prefer the widget for now as it's more prominent (and because I don't have time for more), but we could go crazy on a separate page (e.g. with some visual indication of what is actually happening, something like this: https://res.infoq.com/news/2016/02/netflix-spinnaker/en/resources/1spinnaker-pipelines.png). I'd really like this but I believe that's a huge amount of work to accomplish.

app/views/dashboard/_hosts_in_build_mode_widget_host_list.html.erb
```diff
@@ -0,0 +1,18 @@
...
+  </thead>
+  <tbody>
+    <% hosts.each do |host| %>
+      <tr>
+        <td class="ellipsis"><%= host_build_status_icon(host) %> <%= link_to host.name, host_path(host) %></td>
+        <td class="hidden-tablet hidden-xs ellipsis"><%= host.owner %></td>
+        <td class="hidden-tablet hidden-xs ellipsis"><%= build_duration(host) %></td>
+        <td class="hidden-tablet hidden-xs ellipsis"><%= date_time_relative_value(host.token.expires) %></td>
+      </tr>
+  <% end %>
```
ekohl: nit: odd indenting

app/helpers/dashboard_helper.rb
```diff
@@ -159,6 +159,20 @@ def out_of_sync_enabled?(origin)
...
+      label = _('Token expired')
+    elsif host.build_errors.present?
+      icon = 'error-circle-o'
+      label = _('Build errors')
+    else
+      icon = 'in-progress'
+      label = _('Build in progress')
+    end
+    icon_text(icon, '', kind: 'pficon', title: label)
+  end
```
lzap: Perhaps to use different icon for expired and errored states? Just a thought.

lzap (dismissed):
Nice, please rebase the indenting and I am gonna test and merge thanks.

app/helpers/dashboard_helper.rb
timogoebel: Patternfly doesn't have a icon that fits in my opinion.
https://www.patternfly.org/styles/icons/
Anything you like, @lzap?

timogoebel:
Fixed the indent issue.

app/helpers/dashboard_helper.rb
ohadlevy: @rohoover any suggestions?

app/helpers/dashboard_helper.rb
```diff
@@ -159,6 +159,20 @@ def out_of_sync_enabled?(origin)
     setting.nil? ? true : !setting
   end
 
+  def host_build_status_icon(host)
+    if host.token_expired?
+      icon = 'error-circle-o'
+      label = _('Token expired')
+    elsif host.build_errors.present?
+      icon = 'error-circle-o'
+      label = _('Build errors')
```
lzap: There is one problem here - if a host has `build_errors` set, it is never in `build` mode. When I error a host by running this command:

    wget -q -O- --body-file=LICENSE --method POST --header="Content-Type: text/plain" --no-check-certificate http://localhost:3000/unattended/failed?token=689ba68a-a511-44b1-bca5-c8f3f1ff353e

then the host disappears. Therefore this branch can't be reached.

app/helpers/dashboard_helper.rb
lzap: How about the bell icon? Like expired = bell = ring = end. "fa-bell" is the name.

app/views/dashboard/_hosts_in_build_mode_widget.html.erb
```diff
@@ -0,0 +1,11 @@
+<% hosts = @data.hosts.in_build_mode.includes(:token).preload(:owner).order(created_at: :desc).limit(9) %>
```
lzap: This query must include both "in build" and also when "build errors" text is present (not empty).

app/views/dashboard/_hosts_in_build_mode_widget.html.erb
```diff
@@ -0,0 +1,11 @@
...
+
+<h4 class="header">
+  <%= _('Hosts in Build mode') %>
```
lzap: This could be also "Hosts in Build mode or Failed hosts" or similar.

lzap (changes requested):
One remark.

app/helpers/dashboard_helper.rb
Rohoover: @lzap 

We use the bell icon for the notification drawer. I wouldn't repeat it. I'm bringing this back to the team to see if it's already been identified elsewhere. Stay tuned!

Rohoover: This is what I'm proposing:

Expiring soon = fa-hourglass-end
Expired = fa-hourglass-o

app/helpers/dashboard_helper.rb
timogoebel: Interestingly, I get this:

```
Encoding::UndefinedConversionError ("\xE2" from ASCII-8BIT to UTF-8):
```

Maybe we need something like this:

```ruby
.encode('utf-8', invalid: :replace, undef: :replace, replace: '_')
```

`.gitignore` works fine, our `LICENSE` does not...

app/helpers/dashboard_helper.rb
timogoebel: Thanks, @Rohoover good call.

timogoebel:
Thanks, added hosts with build errors to the widget and fixed an issue where hosts that entered build mode before this patch was applied showed an error. Also changed the icon to follox Rox' recommendation.

This is the latest screenshot:
image

lzap (approved):

lzap:
Thank you for your contribution, merged.

app/helpers/hosts_helper.rb
```diff
@@ -251,7 +251,7 @@ def name_field(host)
   end
 
   def build_duration(host)
-    return _("N/A") if host.initiated_at.nil? && host.installed_at.nil?
+    return _('N/A') if host.initiated_at.nil? || host.installed_at.nil?
     if host.installed_at.nil?
       time_ago_in_words(host.initiated_at, include_seconds: true) + " (in progress)"
```
alexjfisher, line 256: This code is now unreachable?

timogoebel: Looks unreachable, yes.

lzap: Bite me, but I do see a path: when `host.installed_at.nil?` is true?!? Explain please.

alexjfisher: @lzap Function will return on line 254 if host.installed_at.nil? is true.

lzap: I see.

lzap: I will take a look next week. My GH backlog is not getting slimmer.

lzap: Solved in https://github.com/theforeman/foreman/pull/6603

PR #5231, opened 2018-02-01
Fixes #22480 - Add code coverage to Travis CI

sharvit:
publish code-coverage report from travis to coveralls.io

There is some data from 2012 already there :astonished: 
https://coveralls.io/github/theforeman/foreman


ohadlevy (commented):
this is great! thanks :)

do note that this will only show JS coverage, how does it look for non covered code? (e.g. a pr for ruby only?)

ekohl (approved):

sharvit:
@ohadlevy It will just ignore them.

Currently, the coverage report is apart of `jest` which will create coverage report based on the webpack-stack.

We can potentially create a report for the ruby code and publish it either.
tbh, I'm not familiar with ruby-stack CI process.

dLobatog (approved):
@sharvit Thanks, I see it in this PR already. We already have some coverage report on Jenkins for Ruby tests - maybe it's possible to send that info to coveralls. In any case, let's mark that as TODO and get this for JS already 😄

PR #5099, opened 2017-12-19
Added pull request template

lzap:
This is a proposal for a PR template.

PR #5234, opened 2018-02-03
Fixes #22505 - CoreOS does not work

dLobatog:
The image is called 'CoreOS.png', however the view is trying to fetch
'Coreos.png'.

To test, just create a Operating System with family "CoreOS", and you will see that the list of OSs does not show the little OS logo.

### DIFF
--- a/app/helpers/operatingsystems_helper.rb
+++ b/app/helpers/operatingsystems_helper.rb
@@ -32,7 +32,7 @@ module OperatingsystemsHelper
                "Junos"
              when /OracleLinux/i
                "OracleLinux"
-             when /CoreOS/i
+             when /CoreOS|ContainerLinux|Container Linux/i
                "CoreOS"
              when /NXOS/i
                "NXOS"


tbrisker (dismissed):
Can you please rephrase the commit message in the imperative stating what this fixes? (e.g. "use correct name for coreos logo")

lzap:
Added this to our PR github template: https://github.com/theforeman/foreman/pull/5099

timogoebel:
[test foreman]

timogoebel:
> To test, just create a Operating System with family "CoreOS", and you will see that the list of OSs does not show the little OS logo.

https://github.com/theforeman/foreman/blob/12ca07b6bca5dd0c27657d2f47a87adfcd7cfb75/app/helpers/operatingsystems_helper.rb#L35-L36

Works for me without the patch and the [code suggests](https://github.com/theforeman/foreman/blob/12ca07b6bca5dd0c27657d2f47a87adfcd7cfb75/app/helpers/operatingsystems_helper.rb#L35-L36) the file should be named `CoreOS`.

lzap (approved):
No need to stress our jenkins, I am merging manually with corrected commit message.

lzap:
Hmmm no luck, apparently a bug at github.com or we disallow push from clients completely.

```
Counting objects: 5, done.
Delta compression using up to 7 threads.
Compressing objects: 100% (5/5), done.
Writing objects: 100% (5/5), 602 bytes | 602.00 KiB/s, done.
Total 5 (delta 3), reused 0 (delta 0)
remote: Resolving deltas: 100% (3/3), completed with 3 local objects.
remote: error: GH006: Protected branch update failed for refs/heads/develop.
remote: error: At least one approved review is required by reviewers with write access.
To github.com:theforeman/foreman.git
 ! [remote rejected]     develop -> develop (protected branch hook declined)
error: failed to push some refs to 'git@github.com:theforeman/foreman.git'
```

lzap:
Sorry about the mess, @dLobatog: Can you please rephrase the commit message in the imperative stating what this fixes? (e.g. "use correct name for coreos logo")

timogoebel (changes requested):
@lzap: Please test this first, I don't think this works correctly.

lzap:
I think the problem roots in fact that @dLobatog likely configured OS with random name therefore family is used for the icon which is `Coreos` but @timogoebel properly named the OS as `CoreOS` thus this icon is fetched.

I suggest a definitive solution to this - all icons in lowercase and converting family to lowercase as well.

timogoebel:
> I suggest a definitive solution to this - all icons in lowercase and converting family to lowercase as well.

Sorry, if I'm being a smartass here, but as the regex in the code I posted in my previous comment is case insensitive, the spelling of the family should not matter. It will always map to `CoreOS`. As long as the image is also named `CoreOS` everything should be fine. Am I missing something? 😕 ❓

lzap:
I believe @dLobatog named hist OS like "Something Completely Different CoreOS" and that will not match and falls back to family. Let's wait if he confirms :-)

dLobatog (commented):
@lzap is right. CoreOS changed it's name to 'Container Linux' in 2016, so I called my OSs in Foreman 'Container Linux', family 'CoreOS'.

In order to make it work with both, I will change the regex

dLobatog:
screenshot from 2018-02-09 13-30-40

dLobatog:
screenshot from 2018-02-09 13-37-40
Updated - @timogoebel @lzap I think it should work for both cases now, thanks for the review

lzap (approved):
Works for me but I will wait for @timogoebel 's opinion.

timogoebel (approved):

timogoebel:
👍

timogoebel:
Thanks, @dLobatog and @lzap.

PR #5236, opened 2018-02-05
Fixes #22464 - error when response has content-length: 0

jdshewey:
Fixes #22464 in which a valid JSON response without a body causes an exception to be thrown. This is caused by an [upstream bug](https://bugs.ruby-lang.org/issues/14451) in Ruby wherein Net:HTTP (used by the rest-client gem) will attempt to wait for a body even if the web server indicates there is no body, only headers. This option seems sub-optimal, but I fail to see a better solution until 1) this is patched upstream and 2) Foreman runs on this version. This option seems sub-optimal, but I fail to see a better solution until 1) this is patched upstream and 2) Foreman runs on this version.

### DIFF
--- a/lib/proxy_api/resource.rb
+++ b/lib/proxy_api/resource.rb
@@ -7,7 +7,7 @@ module ProxyAPI
     def initialize(args)
       raise("Must provide a protocol and host when initialising a smart-proxy connection") unless (url =~ /^http/)
 
-      @connect_params = {:timeout => Setting[:proxy_request_timeout], :open_timeout => 10,
+      @connect_params = {:read_timeout => nil, :open_timeout => 10,
                          :headers => { :accept => :json, :x_request_id => request_id },
                          :user => args[:user], :password => args[:password]}
 


jdshewey:
I think there is a database setting for proxy_request_timeout that this breaks. I'm not sure where I would update that... Perhaps it should be set to NULL in the database by default? Is the WebUI going to allow a user to set this back to NULL if needed? Is Ruby going to convert NULL to nil properly?

lzap:
Hey thanks for your contribution. So our settings stack is typeful, so if you create a integer setting it is always integer. In this case I'd recommend to actually keep the option and split the timeouts into three separate settings:

* timeout
* read_timeout
* wait_timeout

All integers with some sane defaults. When it's set to zero (`0`) then send `nil` instead of zero (which does not make much sense). How about this?

jdshewey:
>our settings stack is typeful

So does that mean that nil or `NULL` is an allowed type? I am presuming that this is stored in the database which would save as `NULL`, but appears that ActiveRecord will [convert this to nil](https://stackoverflow.com/a/4340600/3127174). I assume this is set as a default when the database schema is created. As long as we are not requiring NOT NULL for this column in the schema, we should be OK to set NULL as a default,

>All integers with some sane defaults. When it's set to zero (0) then send nil instead of zero

If for some reason NULL won't work, [we could set it to -1](https://github.com/rest-client/rest-client/blob/master/lib/restclient/request.rb#L695) which will warn of deprecation and suggest we use nil instead, but will work (for now at least).

While we could create three settings, (they would be `open_timeout`, `read_timeout` and `timeout`) the behavior of RestClient is to set `open_timeout` and `read_timeout` to whatever `timeout` is if either of these values are undefined, so I really think we only need two. 

However, do we really want to allow users to break the protocol? Any setting for `read_timeout` besides `nil` means that we have broken compliance with (at least) RFC 7231 (JSON protocol) and probably the HTTP RFCs too. I really think we should probably only let users set open_timeout, though in this case, that was already hard-coded to 10 when I got here.

My vote would be to either 1) remove the setting altogether or 2) change that to only open_timeout.

Also, would unit tests returning responses to foreman from the smart proxy with zero Content-length headers be a good addition (on a separate PR) in order to test for this situation?

lzap:
Our settings model class has special handling of types, forget about what you know about SQL databases, these timeouts are integers (it's actually serialized in YAML format in the db). So I agree having `-1` for `nil` semantics, that's fine. Just keep in mind you can't get `nil` of it.

I would like to have ability to set both read_timeout and open_timeout from settings, we've seen problems in the past and users needed to finetune those as they sometimes use the HTTP REST api via tunnels/VPNs and weird stuff. I agree we don't want timeout, just read_timeout and open_timeout that's sufficient apparently: https://github.com/rest-client/rest-client/blob/c2e11b943a8514a94a2b7e68cb0f213cef06a6cc/lib/restclient/request.rb#L36-L38

Absolutely unit test is great to have and have it in this PR, it does match what you describe in the ticket/subject.

PR #5241, opened 2018-02-07
Fixes #4127 - Global Parameters with types

kgaikwad:


db/migrate/20171109095957_add_key_type_to_parameters.rb
```diff
@@ -0,0 +1,9 @@
+class AddKeyTypeToParameters < ActiveRecord::Migration
+  def up
+    add_column :parameters, :key_type, :string,  :default =>  N_("string"), :limit => 255
```
sean797: I'm not sure `N_("string")` should be marked for translation?

app/models/concerns/key_type.rb
```diff
@@ -0,0 +1,20 @@
+module KeyType
+  extend ActiveSupport::Concern
+  KEY_TYPES = [N_("string"), N_("boolean"), N_("integer"), N_("real"), N_("array"), N_("hash"), N_("yaml"), N_("json")]
```
sean797, line 3: It would be nice to avoid storing them here and in the lookup_key model.

sean797 (changes requested):
Thanks for doing this, it makes the input so much better!

I know its WIP but I guess there are API changes needed?

app/models/concerns/key_type.rb
orrabin: It looks like they are in a comment in the lookup_key model already, it should probably be removed altogether to avoid confusion.

app/models/parameter.rb
```diff
@@ -56,6 +61,10 @@ def self.type_priority(type)
     PRIORITY.fetch(type.to_s.underscore.to_sym, nil) unless type.nil?
   end
 
+  def value_before_type_cast
```
orrabin, line 61: Not sure you need this, in `KeyValueValidation` the `value_before_type_cast` can run with no parameters and then it will work on self.value which is what you are sending.

app/views/common_parameters/_form.html.erb
```diff
@@ -1,25 +1,14 @@
-<%= form_for @common_parameter, :html => { :onsubmit => 'tfm.editor.submitCode();' } do |f| %>
+<%= form_for @common_parameter do |f| %>
     <%= base_errors_for @common_parameter %>
     <%= text_f f, :name %>
-    <div class='form-group'>
-      <label class="control-label col-md-2" for="common_parameter_value"><%= _("Value") %></label>
-
-      <div class="col-md-9">
-        <div class="editor-container">
```
orrabin, line 8: Maybe we can consult @amirfefer or @sharvit on how to allow the editor when in full screen mode?
Or @Rohoover could suggest another editor.

Background for this: we are using the ACE editor but when adding a type and making parameters look more like smart class parameters we need to show validations next to the editor and using the ACE would not look so good.
There was an attempt at some point to only show the ACE editor when entering full screen for easy editing but that failed.
Perhaps with the use of react this is now possible?
It will probably need to be in a separate PR.

app/helpers/key_types_helper.rb
```diff
@@ -0,0 +1,27 @@
+module KeyTypesHelper
+  def param_type_selector(f, options = {})
+    common_extra_options = { :size => "col-md-4", :class=> "without_select2",
+                 :label_help => _("<dl>" +
+                   "<dt>String</dt> <dd>Everything is taken as a string.</dd>" +
```
orrabin: I know you didn't write it and just moved it but maybe now is a good time to change the wording in this comment a little.

db/migrate/20171109095957_add_key_type_to_parameters.rb
```diff
@@ -0,0 +1,9 @@
+class AddKeyTypeToParameters < ActiveRecord::Migration
```
orrabin: I get this error when trying to run the migrations:
```
StandardError: Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:

  class AddKeyTypeToParameters < ActiveRecord::Migration[4.2]

```

orrabin (dismissed):

orrabin:
The validations works nicely and the migration looks correct.

The hidden value checkbox is not working with the change, the js there probably searches for the editor and needs to be updated.

kgaikwad:
@sean797 and @orrabin,

Thank you for reviewing the pull-request as well as for suggesting additional required changes.
I will do modifications as per your review comments and update the PR soon.

app/views/common_parameters/_form.html.erb
Rohoover: @orrabin 
Is there are screenshot or visual you can show me?

kgaikwad: @Rohoover,

Please check below screenshots for your reference.

1. Current web-ui from Global Parameters page
current_ui_on_global_parameter

2. web-ui after my change where I have removed existing editor and added key type dropdown. I found that for other parameters pages, we haven't used any editor except this page.

after_existing_editor_removed_added_types

3. web-ui for Smart Class Parameters Page
screenshot from 2018-03-01 17-52-27

Rohoover (commented):
@orrabin 

I *think* I understand. Would it be possible to put the same expand button you have on the Smart Class Parameters page on the Global Parameters page next to Value box. This would seem like a consistent handling.

I hope I understood the issue properly.

orrabin:
@Rohoover yes I meant adding the expand button and checking if when expanding we could get the editor.
There were issues with trying to add ACE only when expanding.
My questions:
- does it even make sense to show it only when expanding?
- are there other editors we should consider here?

sbernhard:
I would be really happy to see this PR in foreman as it would help us to configure foreman_ansible roles which require hash values like https://github.com/bennojoy/nginx

How to proceed to get it in?

kgaikwad:
@sbernhard,
I am working on it. 
Made changes as per review comments but still some part is remaining i.e UI integration under parameters Tab of other resources pages like hostgroup -> parameters Tab.
I will try to update it sooner.

kgaikwad:
@orrabin, @sean797 and @sbernhard,

Apologize for delay in delivery of these changes!
I have updated the PR with all suggested changes. Please have a look and let me know if there is any other place where I need to handle parameter types.

- [x] Fixed all review comments by @orrabin 
- [x] Added API changes as @sean797 mentioned above
- [x] Added field `type` under Parameters Tab for other places like organizations > edit > Parameters
- [x] Yaml rendering on host page
- [x] Added sorting for `type` column on global parameters list
- [x] Added `expand button` on edit parameter page as @Rohoover suggested
- [x] Fixed hidden checkbox functionality on global parameter edit page as it was broken by new changes.
- [ ] Fix test cases if any failure found due to new changes
- [ ] Add few test cases for these changes


@Rohoover,
Here are some screenshots for few pages where I have added type changes. 

1. Global Parameters Page 
global_parameter_list_page

2. Global Parameter Edit Page: 
global_parameter_edit_page

3. Organizations > Parameters Tab and same way on other parameters Tab
organization_parameters_tab

Rohoover (commented):
@kgaikwad 

Just a bit picky here but can we get the input field to match the height of the expand button for the edit page?

xprazak2:
@kgaikwad, what is the status here?

kgaikwad:
@xprazak2,

Resolved review comments. Few tests related to search are still failing from `host_test.rb ` so working on it but not getting a reason for failures.

kgaikwad:
@ares, @orrabin,

Now, converted `value` field to serialized for parameters.
As per my understanding, this is the reason now search test cases are failing from [host_test.rb](https://github.com/theforeman/foreman/blob/develop/test/models/host_test.rb#L2073).

Please correct me if my understanding is wrong here - `Serialize comes with the disadvantage that you can't search by SQL on it`. We can perform search on application level but I am not sure how efficient it will be.
I have tried using method `to_yaml`, it results into other test case failures.

I would like to ask your suggestions on this.

ares:
After conversation on IRC, I think serialization is correct. That means we need to convert the search value to yaml as well but I guess only for `=` operator. Things like `^` or `~` might not work well. Also this means we'd need a migration for existing params. I also wonder, do we need to keep this functionality of searching params by values? Don't people search by param names usually?

ekohl:
> After conversation on IRC, I think serialization is correct. That means we need to convert the search value to yaml as well but I guess only for = operator. Things like ^ or ~ might not work well. Also this means we'd need a migration for existing params. I also wonder, do we need to keep this functionality of searching params by values? Don't people search by param names usually?

If you want to keep that, it could make sense to have a searchable column and one with actual YAML in place. I agree searching by name is much more likely. One use case I could see is if you have various params that point to a hostname and you want to change that hostname, but it's somewhat rare I expect.

orrabin:
+1 for changing/removing the search on value, it doesn't exist on puppet parameters either.

kgaikwad:
@orrabin, @ares,

I have added searching by name instead by value.
Resolved the conflicts and done with rebase.

I have few queries:
1. Should I add migration to update existing parameters of type CommonParameter ONLY? 

2. There are few test cases failed due to search by name instead value. I will update those as per search_by_name. what actions should I take for test cases which are very specific to search_by_value. For example - test "Correctly find hosts with overridden parameter values" in [file](https://github.com/theforeman/foreman/blob/develop/test/models/host_test.rb#L2102).

3. Any areas which needs my attention while adding these changes.

kgaikwad:
@orrabin, @ares,

I have added searching by name instead by value.
Resolved the conflicts and done with rebase.

I have few queries:
1. Should I add migration to update existing parameters of type CommonParameter ONLY? 
2. There are few test cases failed due to search by name instead value. I will update those as per search_by_name. what actions should I take for test cases which are very specific to search_by_value. For example - test "Correctly find hosts with overridden parameter values" in [file](https://github.com/theforeman/foreman/blob/develop/test/models/host_test.rb#L2102).
3. Any areas which I need to check with adding these changes.

kgaikwad:
@orrabin, @ares,

I have added searching by name instead by value.
Resolved the conflicts and done with rebase.

I have few queries:
1. Should I add migration to update existing parameters of type CommonParameter ONLY? 
2. There are few test cases failed due to search by name instead value. I will update those as per search_by_name. what actions should I take for test cases which are very specific to search_by_value. For example - test "Correctly find hosts with overridden parameter values" in [file](https://github.com/theforeman/foreman/blob/develop/test/models/host_test.rb#L2102).
3. Any areas which I need to check with adding these changes.

kgaikwad:
@orrabin, @ares,

I have added searching by name instead by value.
Resolved the conflicts and done with rebase.

I have few queries:
1. Should I add migration to update existing parameters of type CommonParameter ONLY? 
2. There are few test cases failed due to search by name instead value. I will update those as per search_by_name. what actions should I take for test cases which are very specific to search_by_value. For example - test "Correctly find hosts with overridden parameter values" in [file](https://github.com/theforeman/foreman/blob/develop/test/models/host_test.rb#L2102).
3. Any areas which I need to check with these changes.

kgaikwad:
@orrabin, @ares,

I have added searching by name instead by value.
Resolved the conflicts and done with rebase.

I have few queries:
1. Should I add migration to update existing parameters of type CommonParameter ONLY? 
2. There are few test cases failed due to search by name instead value. I will update those as per search_by_name. what actions should I take for test cases which are very specific to search_by_value. For example - test "Correctly find hosts with overridden parameter values" in [file](https://github.com/theforeman/foreman/blob/develop/test/models/host_test.rb#L2102).
3. Any areas which I need to check with these changes.

ares:
I'd suggest following
1. Yes, we need migration and we need it for all params, all are affected right?
2. I'd drop searching by value, searching in yaml is not much useful anyway and as Ori said, puppet params don't support it either.
3. Testing on some larger setup would be good, but that's task for reviewer. I can offer env with several different types of params

orrabin:
I would add to 3 - there might be automation around this that also searches by value and would need to be changed but once we have that we could run it to make sure other tests pass

kgaikwad:
@ares, @orrabin,
Thank you for your reply. I will add one migration and do rest of changes as per suggestions.

Sorry for duplicate comments :-), I will delete those.

kgaikwad:
@ares, @orrabin,
- Added new migration to cast existing parameters so that it stores proper values and key_types.
- Fix hidden_value on host page. During testing, I found the issue regarding this.
- Removed test cases which were failing due to search_by_param.
- One query regarding migration that is whether to consider translation there or not.

mmoll:
@kgaikwad please rebase :)

app/helpers/form_helper.rb
```diff
@@ -173,11 +173,20 @@ def time_zone_select_f(f, attr, default_timezone, select_options = {}, html_opti
   def selectable_f(f, attr, array, select_options = {}, html_options = {})
     html_options[:size] = 'col-md-10' if html_options[:multiple]
     field(f, attr, html_options) do
-      addClass html_options, "form-control"
-      f.select attr, array, select_options, html_options
+      form_select_f(f, attr, array, select_options, html_options)
     end
   end
 
+  def form_select_f(f, attr, array, select_options = {}, html_options = {})
```
ares: should this be private, look like a helper that devs normally shouldn't use?

app/models/parameter.rb
ares: the naming here is confusing, what's the difference between value_before_type_casting and value_before_type_cast? it looks like this only returns early if error is present

app/views/common_parameters/_inherited_parameters.html.erb
```diff
@@ -1,7 +1,9 @@
+
```
ares: probably unintentional

app/views/common_parameters/index.html.erb
```diff
@@ -9,14 +9,18 @@
     <tr>
       <th class="col-md-3"><%= sort :name, :as => s_("Parameter|Name") %></th>
       <th class="col-md-8" style='min-width: 60%;'><%= sort :value, :as =>  s_("Parameter|Value")%></th>
+      <th class="col-md-8" style='min-width: 60%;'><%= sort :parameter_type, :as =>  s_("Parameter|Type")%></th>
```
ares: The total of this table shouldn't be more than 12, also the min-width would be now 120% with the line above

db/migrate/20181105061336_cast_key_types_and_values_in_parameters.rb
```diff
@@ -0,0 +1,46 @@
+class CastKeyTypesAndValuesInParameters < ActiveRecord::Migration[5.2]
+  def up
+    User.as_anonymous_admin do
+      Parameter.unscoped.each do |parameter|
+        next if parameter.value.contains_erb?
```
ares: this parameter ended without type being set, should it default to string?

db/migrate/20181105061336_cast_key_types_and_values_in_parameters.rb
```diff
@@ -0,0 +1,46 @@
+class CastKeyTypesAndValuesInParameters < ActiveRecord::Migration[5.2]
+  def up
+    User.as_anonymous_admin do
```
ares: in migration, user does not have to exist yet (migration on initial db), if we need top skip permission check, then we need to wrap it with `Parameter.skip_permission_check { ... }`

ares: also we should probably disable auditing in this migration, `Parameter.without_auditing { ... }` should do, it created many audit records in my db under internal admin

db/migrate/20181105061336_cast_key_types_and_values_in_parameters.rb
```diff
@@ -0,0 +1,46 @@
...
+  rescue StandardError, LoadError => e
+    value = param.value
+    say "Failed to cast #{value} for #{param.inspect}"
+    say "Error: #{e.message}"
+  ensure
+    save_param(param, value, key_type_name)
+  end
+
+  def save_param(param, new_value, new_key_type)
+    if Parameter::KEY_TYPES.include?(N_(new_key_type))
```
ares: the `N_()` extraction is not needed here, strings are extracted in `Parameter::KEY_TYPES` definition, here we just compare untranslated values

ares (changes requested):
It works like a charm, few more comments in the code. I suppose tests failed because of using internal admin in migration. See my comments in migration file for how to fix it.

It's getting closer to merge! Do you happen to have hammer PR too? API added the key, but I think hammer will need changes too.

db/migrate/20181105061336_cast_key_types_and_values_in_parameters.rb
kgaikwad: If I run migration with `Parameter.skip_permission_check` on development setup, records are failing  on save with validation messages. With `User.as_anonymous_admin`,  save runs properly without validation failures for records.


```

-- Failed to update param(36): #<HostParameter id: 36, name: "subname", value: "test purpose", reference_id: 3, created_at: "2018-08-25 18:14:12", updated_at: "2018-11-05 06:53:07", type: "HostParameter", priority: 70, hidden_value: false, key_type: "string">
-- Error: ["You don't have permission edit_params with attributes that you have specified or you don't have access to specified organizations or locations"]
-- Failed to update param(37): #<HostParameter id: 37, name: "kt_activation_keys", value: "test1", reference_id: 3, created_at: "2018-08-25 18:14:13", updated_at: "2018-11-05 06:53:07", type: "HostParameter", priority: 70, hidden_value: false, key_type: "string">
-- Failed to update param(11): #<GroupParameter id: 11, name: "fdgfd", value: "fdgdfg", reference_id: 6, created_at: "2018-04-26 13:00:59", updated_at: "2018-11-05 07:03:45", type: "GroupParameter", priority: 60, hidden_value: false, key_type: "string">
-- Error: ["Hostgroup can't be blank"]
-- Failed to update param(30): #<DomainParameter id: 30, name: "Tp1", value: "tp1", reference_id: 1, created_at: "2018-05-27 10:51:44", updated_at: "2019-01-14 13:21:57", type: "DomainParameter", priority: 30, hidden_value: false, key_type: "string">
-- Error: ["Domain can't be blank"]
```

app/models/parameter.rb
kgaikwad: Actually, previously we have used method `value_before_type_cast` as custom method in lookup_key but if I am correct 'ActiveModel` also provides `*_before_type_cast` (here * -> attribute_name) methods so modified it and declared `value_before_type_casting`  as our custom method used by `lookup_key` to check casting by passing the `value` . 
Any other suggestion on method name for value_before_type_casting?

EDIT - modified method name from value_before_type_casting to `format_value_before_type_cast`

kgaikwad:
@ares, thank you for review. I have addressed almost all comments except migration change.
Here is the hammer PR link - https://github.com/theforeman/hammer-cli-foreman/pull/409


EDIT - updated PR with the suggested modifications.

ares (changes requested):
It's ready from code perspective, there are 2 tests failing now that needs to be fixed. It seems escaping has changed.

Also I found one little issue, the value in parameter form does seems to have smaller height than it should. It's not aligned with fullscreen button. Should be tiny css fix I hope. See screenshot

form_height

After this is fixed, I'm happy to click the green button. I'm looking forward to this a lot :-)

kgaikwad:
@ares,
Updated PR which includes tests fixes and UI fix.

ares (approved):

ares (approved):
Thanks @kgaikwad, I retested and confirmed it still works smoothly. I'm very happy to click the merge button, great work!

app/models/concerns/hostext/search.rb
```diff
@@ -65,7 +65,7 @@ module Search
 
       scoped_search :relation => :puppetclasses, :on => :name, :complete_value => true, :rename => :class, :only_explicit => true, :operators => ['= ', '~ '], :ext_method => :search_by_puppetclass
       scoped_search :relation => :fact_values, :on => :value, :in_key => :fact_names, :on_key => :name, :rename => :facts, :complete_value => true, :only_explicit => true, :ext_method => :search_cast_facts
-      scoped_search :relation => :search_parameters, :on => :value, :on_key => :name, :complete_value => true, :rename => :params, :ext_method => :search_by_params, :only_explicit => true
```
timogoebel, line 68: I know, I'm late to the party. But where did this go?

kgaikwad: As value field is now converted to serialized one searching is not working as expected.
I tried using `to_yaml` but won't able to cover all scenarios and for smart_class_parameters, I don't find searching with value field. 
One of suggestion given by @ekohl is to have searchable field separately and value with yaml. 
After this [discussion](https://github.com/theforeman/foreman/pull/5241#issuecomment-392332053), I have removed searching parameters using value field and added searching based on name.
What is your thoughts on this?

timogoebel: I believe we (dm) use search by parameters in some automation via API requests. We (foreman) promise that the APIv2 is stable and I believe part of the promise is keeping the search stable.

I think the least we should do is add this to the manual in the upgrade notes.

ohadlevy: can't we keep the current behavior ? this will continue to work with strings as it was prior to the change?

ares: I think the search syntax can't be considered as part of the API. While we should try to do as little breaking changes as possible, but sometimes it does not make sense. From time to time we realize it's better to rename the search keyword for a given attribute or change the :only_explicit definition or drop some operators for some attribute and it was always accepted.

Now for the suggestion, if we allow fulltext searching, it could work in some cases. But it can also start returning weird results or simply feel buggy. People will ask for adding more operators in order to search in nested keys etc. I think it's better to just drop this and don't make people confused. I believe it's clear to all devs, how the search would behave, but it would be quite surprising for users who don't understand how we store data in SQL and how scoped search syntax translates to it.

Also given it doesn't work on smart class parameters and smart variables, I don't think it will hit majority of users. If you strongly believe we should add full text searching back, I'd prefer if we first add some inline help about how the search works and especially in this particular search field.

timogoebel: > I think the search syntax can't be considered as part of the API.

I think the search syntax is a valid part of the API, we even use it for permissions. As I said, the least we should do is document the changes so users don't get surprises when they upgrade.

ares: So here's the documentation - https://github.com/theforeman/theforeman.org/pull/1322
Given the fact ale least two people feel differently about that, we should probably bring that up to discourse or document the outcome in our manual. At the end I don't mind steering any direction, but we should set the expectations for users. I didn't find any mention at https://theforeman.org/manuals/1.20/index.html#4.1.5Searching (or in API section linked from here)

PR #5242, opened 2018-02-07
fixes #22090 - friendlyid for lookupvalue plus override api

iwalmsley:
Allow use of match instead of ids in API call for overrides.
Present tests default to using the match name due to 
to_param hence the tests in place should already apply.

### DIFF
--- a/app/controllers/api/v2/override_values_controller.rb
+++ b/app/controllers/api/v2/override_values_controller.rb
@@ -88,6 +88,9 @@ module Api
 
       def find_override_value
         @override_value = LookupValue.find_by_id(params[:id])
+        if @smart
+          @override_value ||= @smart.lookup_values.friendly.find(params[:id])
+        end
       end
 
       def return_if_override_mismatch
--- a/app/models/lookup_value.rb
+++ b/app/models/lookup_value.rb
@@ -1,5 +1,7 @@
 class LookupValue < ApplicationRecord
   audited :associated_with => :lookup_key
+  extend FriendlyId
+  friendly_id :match
   include Authorizable
   include PuppetLookupValueExtensions
   include HiddenValue
--- a/test/controllers/api/v2/override_values_controller_test.rb
+++ b/test/controllers/api/v2/override_values_controller_test.rb
@@ -42,14 +42,21 @@ class Api::V2::OverrideValuesControllerTest < ActionController::TestCase
   end
 
   test "should show specific override values for specific smart variable" do
-    get :show, params: { :smart_variable_id => lookup_keys(:two).to_param, :id => lookup_values(:four).to_param }
+    get :show, params: { :smart_variable_id => lookup_keys(:two).to_param, :id => lookup_values(:four).id }
     assert_response :success
     results = ActiveSupport::JSON.decode(@response.body)
     assert_not_empty results
     assert_equal "hostgroup=Common", results['match']
   end
   test "should show specific override values for specific smart class parameter" do
-    get :show, params: { :smart_class_parameter_id => lookup_keys(:complex).to_param, :id => lookup_values(:hostgroupcommon).to_param }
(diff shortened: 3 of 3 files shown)


dLobatog:
[test]

dLobatog:
[test foreman]
[test katello]
[test upgrade]

iwalmsley:
Thanks for the tests, having had a look I'm not sure that Katello failure is related, but would it be possible to run again/someone else who is more familiar to check manually?

app/controllers/api/v2/override_values_controller.rb
```diff
@@ -88,6 +88,7 @@ def find_override_values
 
       def find_override_value
         @override_value = LookupValue.find_by_id(params[:id])
+        @override_value ||= @override_values.friendly.find(params[:id])
```
dLobatog: Sorry for the delay, shouldn't `@override_values` be changed here to `LookupValue` again? `@override_values` is set to `@smart.lookup_values.paginate(paginate_options)` - which means that we're not really searching all LookupValues.

dLobatog (changes requested):
@iwalmsley Change looks good, can you add a test & change the code to `LookupValue.friendly.find`? Thanks!

app/controllers/api/v2/override_values_controller.rb
iwalmsley: Thanks for the feedback, very true about pagination I had overlooked that, some extra logic is still needed vs what you sugguest however as need to scope the lookupvalue by Smart variable/class param otherwise it's not unique. Will submit a proposal. 

In terms of tests, the override test at the moment use to_param i.e. https://github.com/theforeman/foreman/blob/develop/test/controllers/api/v2/override_values_controller_test.rb#L45 which defaults to then using the match string anyway due to friendly, I can split this out to explicitly use id & match based calls though?

iwalmsley:
@dLobatog thank you. I have submitted a fix but re-based which seems to have got rid of the feedback (trying to fix the prprocessor error - sorry, still getting used to the system...), but thanks for picking up on that. The lookup does need to be scoped against the lookup_key however otherwise it won't be unique so i've submitted code to that effect.

In terms of tests - the current tests in place use to_param on lookup_value so it should now default to using the match string due to friendly, (i.e https://github.com/theforeman/foreman/blob/develop/test/controllers/api/v2/override_values_controller_test.rb#L45), however I can write some explicit tests for id & match string separately if required?

iwalmsley:
Thank you for the feedback, I had overlooked the pagination element. Some extra logic is still needed vs what you suggest however as need to scope the lookup value by the Smart variable/class param iin question otherwise it will not be unique/correct. Have submitted a proposal (should I have had this as an extra commit or is rebase okay?).

In terms of tests, the override tests at the moment use to_param (e.g https://github.com/theforeman/foreman/blob/develop/test/controllers/api/v2/override_values_controller_test.rb#L45) which defaults to then using the match string anyway due to friendly, I can split this out to explicitly use id & match based separate tests though if required?

iwalmsley:
@dLobatog  Have added some tests to explictly test the match & id. Let me know if this is alright. I've done further research into the other API end-points about how they find params but doesn't seem to be a really uniform way (maybe resource_finder?).

iwalmsley:
Just rebased to avoid a conflict. @dLobatog Have you had a chance to look at this yet? I've had this in my setup now for some time and it seems relatively stable & extremely useful when interfacing with external systems. One thing I think may be worth pause for thought is that the "match" name isn't unique, and still using it as a friendly id, not sure if you can see any issues with this? I haven't come across anything and the value is only really meaningful in the scope of a lookup key anyway.

mmoll:
ok to test

dLobatog (approved):

dLobatog:
Thanks @iwalmsley , this completely slipped off my radar. Merging away!

PR #5244, opened 2018-02-07
Fixes #19389 - Added medium hash to TFTP files

ShimShtein:
After this change, boot file names will contain a hash of the medium they were created from. Example:
```
# old:
boot/Redhat-6.1-x86_64-vmlinuz
boot/Redhat-6.1-x86_64-initrd.img
# new:
boot/CentOS_mirror-aBsbsZ_lDn-vmlinuz
boot/CentOS_mirror-aBsbsZ_lDn-initrd.img
```

This will add the ability to distinguish between multiple "flavors" of the same operating system. For example, if a user wants to have both Fedora 27 server and Fedora 27 workstation. Best way to do it, would be defining multiple installation media for the same OS.

Before this change, foreman would overwrite the same file from different source each time the user deployed a host with different medium selected. After the change, both files would coexist in the TFTP directory.

Technically, the hash is the digest of medium URI. This should work even with katello, since they are overloading the same methods.

Edit by lzap: Corrected according to final naming conventions we agreed on.

### DIFF
--- a/app/models/concerns/host_common.rb
+++ b/app/models/concerns/host_common.rb
@@ -74,9 +74,13 @@ module HostCommon
     end
   end
 
+  def host_medium_provider
+    @medium_provider ||= Foreman::Plugin.medium_providers.find_provider(self)
+  end
+
   # Returns a url pointing to boot file
   def url_for_boot(file)
-    "#{os.medium_uri(self)}/#{os.url_for_boot(file)}"
+    os.url_for_boot(medium_provider, file)
   end
 
   def puppetca?
@@ -112,7 +116,7 @@ module HostCommon
       # We encode the hw_model into the image file name as not all Sparc flashes can contain all possible hw_models. The user can always
       # edit it if required or use symlinks if they prefer.
       hw_model = model.try :hardware_model if defined?(model_id)
-      operatingsystem.interpolate_medium_vars(nfs_path, architecture.name, operatingsystem) + \
+      host_medium_provider.interpolate_vars(nfs_path) + \
         "#{operatingsystem.file_prefix}.#{architecture}#{hw_model.empty? ? '' : '.' + hw_model.downcase}.#{operatingsystem.image_extension}"
     else
       ""
--- a/app/models/operatingsystem.rb
+++ b/app/models/operatingsystem.rb
@@ -128,26 +128,6 @@ class Operatingsystem < ApplicationRecord
     []
   end
 
-  def medium_uri(host, url = nil)
-    url ||= host.medium.path unless host.medium.nil?
-    url ||= ''
-    medium_vars_to_uri(url, host.architecture.name, host.operatingsystem)
-  end
-
-  def medium_vars_to_uri(url, arch, os)
-    URI.parse(interpolate_medium_vars(url, arch, os)).normalize
-  end
-
-  def interpolate_medium_vars(path, arch, os)
-    return "" if path.empty?
-
-    path.gsub('$arch', arch).
-         gsub('$major',  os.major).
-         gsub('$minor',  os.minor).
-         gsub('$version', os.minor.blank? ? os.major : [os.major, os.minor].compact.join('.')).
-         gsub('$release', os.release_name.presence || '')
-  end
-
   # The OS is usually represented as the concatenation of the OS and the revision
   def to_label
     return description if description.present?
@@ -191,31 +171,49 @@ class Operatingsystem < ApplicationRecord
     ["None", "PXELinux BIOS"]
   end
 
-  # sets the prefix for the tfp files based on the os / arch combination
-  def pxe_prefix(arch, host)
-    "boot/#{self}-#{arch}-#{medium_digest(host)}".tr(" ","-")
+  # Compatibility method, don't want to break all templates that use it.
+  def medium_uri(host)
+    Foreman::Deprecation.deprecation_warning("1.20", "medium_uri is now accessible through @medium_provider.medium_uri in templates")
+    @medium_provider = Foreman::Plugin.medium_providers.find_provider(host)
+    @medium_provider.medium_uri
   end
 
-  def medium_digest(host)
-    host.nil? ? '' : Digest::SHA1.hexdigest(medium_uri(host).to_s)
+  # sets the prefix for the tfp files based on medium unique identifier
+  def pxe_prefix(medium_provider)
+    unless medium_provider.is_a? MediumProviders::Provider
+      raise Foreman::Exception.new(N_('Please provide a medium provider. It can be found as @medium_provider in templates, or Foreman::Plugin.medium_providers.find_provider(host)'))
+    end
+    "boot/#{medium_provider.unique_id}"
   end
 
-  def pxe_files(medium, arch, host = nil)
-    boot_files_uri(medium, arch, host).collect do |img|
-      { pxe_prefix(arch).to_sym => img.to_s}
+  def pxe_files(medium_provider, _arch = nil, host = nil)
+    # Try to maintain backwards compatibility, if medium_provider could be constructed - do it with a warning.
+    if host
+      Foreman::Deprecation.deprecation_warning("1.20", "Please provide a medium provider. It can be found as @medium_provider in templates, or Foreman::Plugin.medium_providers.find_provider(host)")
+      medium_provider = Foreman::Plugin.medium_providers.find_provider(host)
+    end
+
+    unless medium_provider.is_a? MediumProviders::Provider
+      raise Foreman::Exception.new(N_('Please provide a medium provider. It can be found as @medium_provider in templates, or Foreman::Plugin.medium_providers.find_provider(host)'))
+    end
+    boot_files_uri(medium_provider).collect do |img|
+      { pxe_prefix(medium_provider).to_sym => img.to_s}
     end
   end
 
-  def kernel(arch, host)
-    bootfile(arch,:kernel, host)
+  def kernel(medium_provider)
+    bootfile(medium_provider, :kernel)
   end
 
-  def initrd(arch, host)
-    bootfile(arch,:initrd, host)
+  def initrd(medium_provider)
+    bootfile(medium_provider, :initrd)
   end
 
-  def bootfile(arch, type, host)
-    pxe_prefix(arch, host) + "-" + self.family.constantize::PXEFILES[type.to_sym]
+  def bootfile(medium_provider, type)
+    unless medium_provider.is_a? MediumProviders::Provider
+      raise Foreman::Exception.new(N_('Please provide a medium provider. It can be found as @medium_provider in templates, or Foreman::Plugin.medium_providers.find_provider(host)'))
+    end
+    pxe_prefix(medium_provider) + "-" + self.family.constantize::PXEFILES[type.to_sym]
   end
 
   # Does this OS family support a build variant that is constructed from a prebuilt archive
@@ -265,12 +263,17 @@ class Operatingsystem < ApplicationRecord
     end
   end
 
-  def boot_files_uri(medium, architecture, host = nil)
-    raise ::Foreman::Exception.new(N_("%{os} medium was not set for host '%{host}'"), :host => host, :os => self) if medium.nil?
-    raise ::Foreman::Exception.new(N_("Invalid medium '%{medium}' for '%{os}'"), :medium => medium, :os => self) unless media.include?(medium)
-    raise ::Foreman::Exception.new(N_("Invalid architecture '%{arch}' for '%{os}'"), :arch => architecture, :os => self) unless architectures.include?(architecture)
-    self.family.constantize::PXEFILES.values.map do |img|
-      medium_vars_to_uri("#{medium.path}/#{pxedir}/#{img}", architecture.name, self)
+  def boot_files_uri(medium_provider, &block)
+    boot_file_sources(medium_provider, &block).values
+  end
+
+  def url_for_boot(medium_provider, file, &block)
+    boot_file_sources(medium_provider, &block)[file]
+  end
+
+  def boot_file_sources(medium_provider, &block)
+    @boot_file_sources ||= self.family.constantize::PXEFILES.transform_values do |img|
+      "#{medium_provider.medium_uri(pxedir, &block)}/#{img}"
     end
   end
 
--- a/app/models/operatingsystem.rb
+++ b/app/models/operatingsystem.rb
@@ -129,7 +129,8 @@ class Operatingsystem < ApplicationRecord
   end
 
   def medium_uri(host, url = nil)
-    url ||= host.medium.path
+    url ||= host.medium.path unless host.medium.nil?
+    url ||= ''
     medium_vars_to_uri(url, host.architecture.name, host.operatingsystem)
   end
 
@@ -191,8 +192,12 @@ class Operatingsystem < ApplicationRecord
   end
 
   # sets the prefix for the tfp files based on the os / arch combination
-  def pxe_prefix(arch)
-    "boot/#{self}-#{arch}".tr(" ", "-")
+  def pxe_prefix(arch, host)
+    "boot/#{self}-#{arch}-#{medium_digest(host)}".tr(" ","-")
+  end
+
+  def medium_digest(host)
+    host.nil? ? '' : Digest::SHA1.hexdigest(medium_uri(host).to_s)
   end
 
   def pxe_files(medium, arch, host = nil)
@@ -201,16 +206,16 @@ class Operatingsystem < ApplicationRecord
     end
   end
 
-  def kernel(arch)
-    bootfile(arch, :kernel)
+  def kernel(arch, host)
+    bootfile(arch,:kernel, host)
   end
 
-  def initrd(arch)
-    bootfile(arch, :initrd)
+  def initrd(arch, host)
+    bootfile(arch,:initrd, host)
   end
 
-  def bootfile(arch, type)
-    pxe_prefix(arch) + "-" + self.family.constantize::PXEFILES[type.to_sym]
+  def bootfile(arch, type, host)
+    pxe_prefix(arch, host) + "-" + self.family.constantize::PXEFILES[type.to_sym]
   end
 
   # Does this OS family support a build variant that is constructed from a prebuilt archive
--- a/app/models/operatingsystems/coreos.rb
+++ b/app/models/operatingsystems/coreos.rb
@@ -5,8 +5,10 @@ class Coreos < Operatingsystem
     'coreos'
   end
 
-  def mediumpath(host)
-    medium_uri(host, "#{host.medium.path}/#{host.architecture.name}-usr").to_s.gsub('x86_64', 'amd64')
+  def mediumpath(medium_provider)
+    medium_provider.medium_uri("/$arch").to_s do |vars|
+      transform_vars(vars)
+    end
   end
 
   def url_for_boot(file)
@@ -14,11 +16,15 @@ class Coreos < Operatingsystem
   end
 
   def pxedir
-    '$arch/$version'
+    '$arch-usr/$version'
   end
 
-  def boot_files_uri(medium, architecture, host = nil)
-    super(medium, architecture, host).each{ |img_uri| img_uri.path = img_uri.path.gsub('x86_64', 'amd64-usr') }
+  def boot_file_sources(medium_provider, &block)
+    super do |vars|
+      vars = yield(vars) if block_given?
+
+      transform_vars(vars)
+    end
   end
 
   def display_family
@@ -28,5 +34,11 @@ class Coreos < Operatingsystem
   # Does this OS family use release_name in its naming scheme
   def use_release_name?
     true
+  end
+
+  private
+
+  def transform_vars(vars)
+    vars[:arch] = vars[:arch].gsub('x86_64', 'amd64')
   end
 end
--- a/app/models/operatingsystems/solaris.rb
+++ b/app/models/operatingsystems/solaris.rb
@@ -6,7 +6,7 @@ class Solaris < Operatingsystem
   end
 
   # sets the prefix for the tftp files based on the OS
-  def pxe_prefix(architecture = nil)
+  def pxe_prefix(_architecture, _host)
     "boot/#{file_prefix}"
   end
 
--- a/app/models/operatingsystems/solaris.rb
+++ b/app/models/operatingsystems/solaris.rb
@@ -6,7 +6,7 @@ class Solaris < Operatingsystem
   end
 
   # sets the prefix for the tftp files based on the OS
-  def pxe_prefix(_architecture, _host)
+  def pxe_prefix(_medium_provider)
     "boot/#{file_prefix}"
   end
 
@@ -28,11 +28,7 @@ class Solaris < Operatingsystem
   end
 
   def pxedir
-    "Solaris_#{minor}/Tools/Boot"
-  end
-
-  def url_for_boot(file)
-    pxedir + "/" + PXEFILES[file]
+    "Solaris_$minor/Tools/Boot"
   end
 
   def boot_filename(host)
@@ -77,11 +73,12 @@ class Solaris < Operatingsystem
   end
 
   def jumpstart_params(host, vendor)
+    medium_provider = Foreman::Plugin.medium_providers.find_provider(host)
     # root server and install server are always the same under Foreman
     server_name = host.medium.media_host
     server_ip   = host.domain.resolver.getaddress(server_name).to_s
     jpath       = jumpstart_path host.medium, host.domain
-    ipath       = interpolate_medium_vars(host.medium.media_dir, host.architecture.name, self)
+    ipath       = medium_provider.interpolate_vars(host.medium.media_dir).to_s
 
     {
       :vendor => "<#{vendor}>",
--- a/app/models/operatingsystems/windows.rb
+++ b/app/models/operatingsystems/windows.rb
@@ -9,8 +9,8 @@ class Windows < Operatingsystem
     "waik"
   end
 
-  def pxe_prefix(arch)
-    "boot/windows-#{arch}/".tr(" ", "-")
+  def pxe_prefix(arch, host)
+    "boot/windows-#{arch}-#{medium_digest(host)}/".tr(" ","-")
   end
 
   def bootfile(arch, type)
--- a/app/models/operatingsystems/windows.rb
+++ b/app/models/operatingsystems/windows.rb
@@ -9,17 +9,15 @@ class Windows < Operatingsystem
     "waik"
   end
 
-  def pxe_prefix(arch, host)
-    "boot/windows-#{arch}-#{medium_digest(host)}/".tr(" ","-")
+  def pxe_prefix(medium_provider)
+    medium_provider.interpolate_vars("boot/windows-$arch-#{medium_provider.unique_id}/").tr(" ", "-")
   end
 
-  def bootfile(arch, type)
-    pxe_prefix(arch) + PXEFILES[type.to_sym]
+  def bootfile(medium_provider, type)
+    pxe_prefix(medium_provider) + PXEFILES[type.to_sym]
   end
 
-  def boot_files_uri(medium, architecture, host = nil)
-    raise ::Foreman::Exception.new(N_("invalid medium for %s"), to_s) unless media.include?(medium)
-
+  def boot_files_uri(medium_provider)
     pxe_dir = ""
 
     PXEFILES.values.collect do |img|
@@ -31,7 +29,7 @@ class Windows < Operatingsystem
         pxe_dir = ""
       end
 
-      URI.parse("#{medium_vars_to_uri(medium.path, architecture.name, self)}/#{pxe_dir}/#{img}").normalize
+      medium_provider.medium_uri("/#{pxe_dir}/#{img}").normalize
     end
   end
 
--- a/app/registries/foreman/plugin/medium_providers_registry.rb
+++ b/app/registries/foreman/plugin/medium_providers_registry.rb
@@ -1,0 +1,59 @@
+module Foreman
+  class Plugin
+    # This class is used to store list of medium uri providers that are responsible to supply
+    # locations of installation medium in form of URI.
+    # This is used by provisioning framework to supply download location for various files
+    # used during the provisioning.
+    class MediumProvidersRegistry
+      attr_reader :providers
+
+      def initialize
+        @providers = []
+      end
+
+      # Register a new medium_uri provider.
+      def register(provider)
+        provider = provider.to_s
+        @providers.unshift(provider)
+        provider
+      end
+
+      # Find a provider that can provide a URI for a given entity.
+      def find_provider(entity)
+        raise Foreman::Exception.new('Must supply an entity to find a medium provider') unless entity
+
+        provider_instances = providers.map { |provider| provider.constantize.new(entity) }
+        valid_providers = provider_instances.select { |provider| provider.valid? }
+        if valid_providers.count > 1
+          raise Foreman::Exception.new(
+            'Found more than one provider for %{entity}. Valid providers: %{providers}',
+            {
+              entity: entity,
+              providers: providers.map { |provider| provider.class.friendly_name }
+            })
+        end
+
+        unless valid_providers.present?
+          raise Foreman::Exception.new(
+            'Could not find a provider for %{entity}. Providers returned %{errors}',
+            {
+              entity: entity,
+              errors: provider_instances.map { |provider| [provider.class.friendly_name, provider.errors] }.to_h
+            })
+        end
+
+        valid_providers.first
+      end
+
+      private
+
+      def ensure_provider(selected_provider, entity)
+        return if selected_provider.present?
+
+        self.class.providers.reduce({}) do |errors_hash, provider|
+          errors_hash[provider.friendly_name] = provider.errors(entity)
+        end
+      end
+    end
+  end
+end
--- a/app/services/medium_providers/default.rb
+++ b/app/services/medium_providers/default.rb
@@ -1,0 +1,66 @@
+module MediumProviders
+  class Default < Provider
+    def validate
+      errors = []
+      os = entity.operatingsystem
+      medium = entity.medium
+      arch = entity.architecture
+
+      errors << N_("%{os} medium was not set for host '%{host}'") % { :host => entity, :os => os } if medium.nil?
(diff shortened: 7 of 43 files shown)


app/models/operatingsystem.rb
```diff
@@ -124,7 +124,8 @@ def repos(host)
   end
 
   def medium_uri(host, url = nil)
-    url ||= host.medium.path
+    url ||= host.medium.path unless host.medium.nil?
```
dmitri-d: Here and other places -- I would suggest using medium instead of host as the first parameter. Less coupling, easier to understand, etc.

ShimShtein: Two things:
1. It would be problematic for katello - they are not using our medium model when the host is consuming managed content. The URI in their case is determined by their own set of fields (`content_source`, `kt_environment` e.t.c )
2. I tried not to change method signatures, since I don't know who is using those methods outside foreman's core.

app/models/operatingsystem.rb
```diff
@@ -186,26 +187,30 @@ def available_loaders
   end
 
   # sets the prefix for the tfp files based on the os / arch combination
-  def pxe_prefix(arch)
-    "boot/#{self}-#{arch}".tr(" ","-")
+  def pxe_prefix(arch, host)
```
lzap: I'd suggest the `host` option to be optional, when not provided no digest is generated. This will give plugin authors more time to adopt. Discovery should be fine tho, it does not call this method directly.

app/models/operatingsystems/solaris.rb
```diff
@@ -6,7 +6,7 @@ def file_prefix
   end
 
   # sets the prefix for the tftp files based on the OS
-  def pxe_prefix(architecture = nil)
+  def pxe_prefix(architecture = nil, host = nil)
```
lzap: This does not match the superclass signature, is this on purpose?

test/models/host_test.rb
```diff
@@ -1,6 +1,9 @@
 require 'test_helper'
 
 class HostTest < ActiveSupport::TestCase
+  include FactImporterIsolation
+  allow_transactions_for_any_importer
+
```
lzap: Not relevant to the ticket, perhaps by mistake.

lzap:
Thanks for tackling this, this was a pain.

app/models/operatingsystem.rb
iNecas: what about not passing the medium, but just the `medium_url`

ShimShtein: This is the method that actually generates the URL, I don't think I can pass it in...

app/models/operatingsystem.rb
ShimShtein: I didn't find any project that actually uses this method, and even if someone does, it could return wrong result without the host.
For example if someone is using this method to fetch the file, but relies on Foreman to put the files into the proxy, the result would be wrong. (Foreman will use hash for filename, but the plugin will look for non-hashed file).

app/models/operatingsystems/solaris.rb
ShimShtein: Matched the superclass.

app/models/operatingsystem.rb
iNecas: oh, right, I was referring to the other methods, that we are adding the additional `host` arg now.

ShimShtein: That method (`pxe_prefix`) now calls implicitly to `medium_uri` for generating the digest. Hence the need for host there.

iNecas: That's why I would like rather to see some `PxeFilesService` object, that would encapsulate the paths and urls generation, and there could ve even multiple implementations (Foreman vs. Katello) that could influence where the paths and urls are. This would clean also the Operatingsystem class. So instead of `host.operatingsystem.{kernel(host),medium_url(architecture, host$,...}`, we could have `host.pxe_files.{kernel_path,medium_url,...}` (notice we would not need to pass the additional args to methods, as the `PxeFilesService` would choose which host/hostgroup attributes are relevant for it.

app/models/operatingsystem.rb
lzap: Okay.

app/models/operatingsystem.rb
ShimShtein: Indeed we can do it, but it will require refactoring for both foreman and katello. Do we really want to get into this refactoring right now?

iNecas: If we don't do it now, then when? It's the ideal time from release perspective to do it now. Lets keep the current version around (in case it would turn out that the refactoring is no go), but let shoot for something better.

lzap:
One idea - installers do rarely get fixed (updated) files, in that case the hash remains the same but there can be corruption. Would it make sense to perform HTTP HEAD before creating the files on proxy and adding "Last-Modified" field if present (not all servers might have that).

As a side effect, we can error out if kernel/image is not present (4xx HTTP response). This nicely solves another set of issues we have - when URL is wrong, wget usually creates zero-sized files, then hosts are erroring out during PXE boot.

iNecas:
It might make sense. I'm just not sure this PR is the place where to add this additional things.

app/models/operatingsystem.rb
ShimShtein: OK I looked into it a bit more. This is what I have right now: there are two things mixed in OS class right now: URI providers (medium_url for example) and boot filenames (like `images/pxeboot/vmlinuz`). The former has nothing to do with OS, so it could be extended into a provider. The latter is OS-dependent so it should stay at the OS class.
So this feature actually falls into the latter category and should stay in OS, I would suggest creating a separate issue to move the former into proper providers pattern.

iNecas: Aren't they related in a way, that the end path (such as vmlinuz) is influenced by the `medium_url`? While I'm ok the base for the boot filenames to be provided by the OS, if we move the final construction of the path to the provider class, it would IMO clean up the code quite a lot, savings us passing some unnecessary args to the os methods.

ShimShtein: Before this PR they were completely unrelated - each OS "knows" what limitations exist on file names, and provides the paths accordingly. Provider is completely unaware where the file will be served, so I don't see a reason to push this responsibility to the provider.

Just to illustrate the issue, I have a host with CentOS set with installation medium set to the one supplied with Foreman.
`medium_url` is set to `http://mirror.centos.org/centos/7/os/x86_64` and `boot_files_uri` returns `[#<URI::HTTP http://mirror.centos.org/centos/7/os/x86_64/images/pxeboot/vmlinuz>,
 #<URI::HTTP http://mirror.centos.org/centos/7/os/x86_64/images/pxeboot/initrd.img>]`
The only method that actually connects the path as should be set on our tftp server with those paths is `pxe_files`:
`[{:"boot/Centos-7-x86_64-4b42362509125324ee277c89a503e04323ce5a70"=>
   "http://mirror.centos.org/centos/7/os/x86_64/images/pxeboot/vmlinuz"},
 {:"boot/Centos-7-x86_64-4b42362509125324ee277c89a503e04323ce5a70"=>
   "http://mirror.centos.org/centos/7/os/x86_64/images/pxeboot/initrd.img"}]`. The keys are tftp file names/directory and the values are the sources for those files.

iNecas: "Provider is completely unaware where the file will be served". Isn't this the reason where the issue was, as we actually needed the path to be aware of the url?

lzap:
Unless there are any comments from @witlessbird or @iNecas I am going to test and merge this tomorrow, thanks a bunch!

app/models/operatingsystem.rb
lzap: I vote for either keeping @ShimShtein 's proposal, OS-Media needs a substantial overhaul with some mechanism which Katello (or other plugins) could use for overriding, that is not point of this PR.

iNecas:
@lzap there has been additional discussion with potential take on this with @ShimShtein. Shim: any chance to share the new approach for review?

ShimShtein:
Sure, I have put it into a separate branch: https://github.com/ShimShtein/foreman/tree/19389_providers.
Feel free to look at it. I didn't test it yet, so expect more changes to it.

iNecas:
I had a brief view, but looks like a good way moving forward and extracting the logic form os, as well as decoupling the os from host.

ShimShtein:
Finally a version that is actually working! @iNecas @lzap can you please review this?

lzap:
Will do.

Note: When rebasing, add this new OS that has been merged. https://github.com/theforeman/foreman/pull/5361

app/models/concerns/host_boot_files_extensions.rb
```diff
@@ -0,0 +1,32 @@
...
+# information to the method.
+# Usage:
+# host = Host::Managed.find(my_host_id)
+# host.info
+# => A hash of hashes that describes all parameters associated with the host.
+#
+# Writing an extension:
+# Inherit HostInfo::Provider and override #host_info method and register the
+# provider by calling HostInfo.register_info_provider
+module HostBootFileExtensions
```
iNecas: Seems like some residual: I don't see this being used anywhere.

app/models/concerns/host_common.rb
```diff
@@ -71,9 +71,14 @@ def parent_name
...
   end
 
+  def medium_provider
+    @medium_provider ||= Foreman::Plugin.medium_providers.find_provider(self)
+  end
+
   # Returns a url pointing to boot file
   def url_for_boot(file)
-    "#{os.medium_uri(self)}/#{os.url_for_boot(file)}"
+    medium_provider = Foreman::Plugin.medium_providers.find_provider(self)
```
iNecas: We can use the `medium_provider` directly, right?

app/models/concerns/host_common.rb
```diff
@@ -109,7 +114,8 @@ def default_image_file
       # We encode the hw_model into the image file name as not all Sparc flashes can contain all possible hw_models. The user can always
       # edit it if required or use symlinks if they prefer.
       hw_model = model.try :hardware_model if defined?(model_id)
-      operatingsystem.interpolate_medium_vars(nfs_path, architecture.name, operatingsystem) +\
+      medium_provider = Foreman::Plugin.medium_providers.find_provider(self)
```
iNecas: same comment about `medium_provider` here

app/models/concerns/host_boot_files_extensions.rb
lzap: LOL I was digging in the codebase for like 40 minutes trying to figure out how *this* works :-)

app/registries/foreman/plugin/medium_providers_registry.rb
```diff
@@ -0,0 +1,47 @@
...
+      def ensure_provider(selected_provider, entity)
+        return if selected_provider.present?
+
+        self.class.providers.reduce({}) do |errors_hash, provider|
+          errors_hash[provider.friendly_name] = provider.errors(entity)
+        end
+      end
+    end
+  end
+end
```
lzap, line 59: This class needs more explanation with examples (in Ruby comments).

app/registries/foreman/plugin/medium_providers_registry.rb
```diff
@@ -0,0 +1,47 @@
...
+
+          errors[provider.class.friendly_name] = provider.errors
+        end
+
+        unless valid_provider
+          raise Foreman::Exception.new 'Could not find a provider for %{entity}. Providers returned %{errors}', { entity: entity, errors: errors } if errors.present?
+        end
+
+        valid_provider
+      end
```
lzap: I see one major problem here - the order. We cannot assume any order, we've seen bugs when plugin was renamed (discovery to foreman_discovery) which changed initialization order and broke things.

We can however make an assumption that only one provider must be valid, therefore is multiple are found we explicitly error out with an exception, this will prevent plugins from sneaking in providers which are pseudo-valid.

app/models/operatingsystem.rb
```diff
@@ -185,27 +166,49 @@ def available_loaders
...
+    end
+    "boot/#{medium_provider.unique_id}".tr(" ","-")
   end
 
-  def pxe_files(medium, arch, host = nil)
-    boot_files_uri(medium, arch, host).collect do |img|
-      { pxe_prefix(arch).to_sym => img.to_s}
+  def pxe_files(medium_provider, _arch = nil, host = nil)
+    # Try to maintain backwards compatibility, if medium_provider could be constructed - do it with a warning.
+    if host
```
iNecas, line 191: It should probably be `if _arch || _host`

app/services/medium_providers/default.rb
```diff
@@ -0,0 +1,59 @@
+module MediumProviders
+  class Default < Provider
```
iNecas, line 2: Would be worth putting some comment on the backround and purpose of this class.

app/registries/foreman/plugin/medium_providers_registry.rb
```diff
@@ -0,0 +1,47 @@
...
+        @providers.unshift(provider)
+        provider
+      end
+
+      def find_provider(entity)
+        valid_provider = nil
+        errors = {}
+        providers.each do |provider_class|
+          provider = provider_class.constantize.new(entity)
+          if provider.valid?
```
iNecas: Since you don't call `validate` anywhere, the `valid?` would always return true, wouldn't it?

app/services/medium_providers/provider.rb
```diff
@@ -0,0 +1,39 @@
...
+  class Provider
+    def self.friendly_name
+      self.name
+    end
+
+    def initialize(entity)
+      @entity = entity
+    end
+
+    def medium_uri
```
iNecas: Please add some docs for `medium_uri` and `boot_files_uri` to describe the contract

test/models/operatingsystems/operatingsystems_test.rb
```diff
@@ -30,54 +30,62 @@ class OperatingsystemsTest < ActiveSupport::TestCase
...
+      medium_provider = Foreman::Plugin.medium_providers.find_provider(host)
+      assert_equal(config['expected'], host.operatingsystem.initrd(medium_provider))
     end
   end
 
-  { :coreos      => { 'os' => :coreos,      'arch' => :x86_64, 'medium' => :coreos, 'expected' => 'boot/CoreOS-494.5.0-x86_64'},
-    :debian7_0   => { 'os' => :debian7_0,   'arch' => :x86_64, 'medium' => :unused, 'expected' => 'boot/Debian-7.0-x86_64'},
-    :ubuntu14_10 => { 'os' => :ubuntu14_10, 'arch' => :x86_64, 'medium' => :ubuntu, 'expected' => 'boot/Ubuntu-14.10-x86_64'},
-    :suse        => { 'os' => :suse,        'arch' => :x86_64, 'medium' => :suse,   'expected' => 'boot/OpenSuse-11.4-x86_64' } }.
+  { :coreos      => { 'os' => :coreos,      'arch' => :x86_64, 'medium' => :unused, 'expected' => 'boot/c440a52e0119b57fc9109aeb7c4a19f5e8e75c62'},
```
iNecas: It would be still good to keep some humanized name as part of the path.

app/services/medium_providers/default.rb
lzap: Also better name could be *DatabaseMedia* provider or something like that.

test/models/operatingsystems/operatingsystems_test.rb
lzap: Preferably at the beginning of the filename, so it's sorted when you list TFTP directory.

lzap:
This problem is worse than I initially thought. It's indeed a problem when changing flavor (Workstation vs Server), but more importantly when Katello is in use and upstream kickstart repository is created against RHEL "7Server" version or CentOS repo with different versioning scheme, Operating System created/associated can have incorrect version leading to this corruption bug. We are still investigating root cause, but it ends up with one Operating System associated with multiple kickstart trees leading to overwrite and eventual corruption.

The current patch will help in this case. I think we can also drop the architecture from the prefix, it's always present in the URL, thus the hash. Family and version are informative and nice to have to keep things sorted a bit in the TFTP dir.

test/models/operatingsystems/operatingsystems_test.rb
lzap: I am for keeping Family-Version but dropping architecture.

lzap:
I think this is going a wrong direction and I have to admit the whole thing was my idea - sorry about that. Here is discussion: https://community.theforeman.org/t/rfc-change-tftp-file-naming-pattern/8955

ShimShtein:
@lzap, I still work on improvements to this PR, but to address your concern about Katello I have given each provider the opportunity to override the unique hash.
So if Katello has a static URL, but the content changes, it should calculate the hash differently - probably by using content version as part of the hash.

lzap:
I believe Katello is major problem here - it's solely based on URL, which changes for each content view or even version published (version is part of URL). Thus this will lead to redownloading all PXE files over and over again creating unnecessary mess in the PXE folder. Naming files after URL hash was a bad idea from the day one (my idea!) I believe that the name should reflect contents rather than URL. Thus my proposal.

iNecas:
I'm wondering if, with Katello, we couldn't use other data (such as content view version) when calculating the hash, so that it actually changes even when the path itself stays the same. It seems like this PR could allow us doing that.

ShimShtein:
@iNecas it depends completely on the medium provider. The default hash function comes from base provider and it hashes the URL, but the provider can override it to use whatever information is relevant for the provider. If the path does not change - find something that does change (like view version).

lzap:
Does katello keep track of kickstart file content - if it changed during sync? Worst case, every sync should be marked as possible change and that should bubble up through CVs and versions changing the hash. IMHO this is too complex when I compare with my proposal which will just work (download, calculate hash from content, rename).

lzap:
I created complementary proposal for smart-proxy changes which will work together with this patch to nicely solve the situation: https://community.theforeman.org/t/rfc-drop-c-flag-from-wget-on-tftp-proxy/10010

@ShimShtein can you create a hash helper method and use it to have shorter hash suffix? MD5/SHA seems to be quite long, if we use some Linux-filesystem-safe pattern (e.g. BASE64 with `/` character replaced with `_`) this could be much shorter.

@iNecas if katello can actually generate the hash in more clever way, this would do it. yeah.

ShimShtein:
> can you create a hash helper method and use it to have shorter hash suffix?

It's already [there](https://github.com/theforeman/foreman/pull/5244/files#diff-712520901560f5f5a31316b918cf19f4R19). 

> MD5/SHA seems to be quite long...

@lzap, I am looking for a better solution, but the only algorithm with shorter hash that I have found, is CRC32 and it's not in base Ruby library. We will probably need to add a dependency for it, if we want to use it.

lzap:
I was thinking at least this:

```ruby
[6] pry(main)> Digest::SHA1.hexdigest("test")
=> "a94a8fe5ccb19ba61c4c0873d391e987982fbbd3"
[7] pry(main)> Base64.urlsafe_encode64(Digest::SHA1.digest("test"))
=> "qUqP5cyxm6YcTAhz05Hph5gvu9M="
```

That's about 30% shorter.

ShimShtein:
Rebased and changed hash to base64 (without padding)

ShimShtein:
[test foreman]

lzap:
I am good, @iNecas for katello bits? I am gonna do formal review and testing with/without discovery (as we also override media there).

iNecas:
@ShimShtein you're planning to update the Katello code as well to use this, right? Would be good to have both PRs ready at the same time.

app/models/operatingsystem.rb
```diff
@@ -190,27 +171,49 @@ def available_loaders
...
 
-  # sets the prefix for the tfp files based on the os / arch combination
-  def pxe_prefix(arch)
-    "boot/#{self}-#{arch}".tr(" ", "-")
+  # Compatibility method, don't want to break all templates that use it.
+  def medium_uri(host)
+    Foreman::Deprecation.deprecation_warning("1.20", "medium_uri is now accessible through @medium_provider.medium_uri in templates")
+    @medium_provider = Foreman::Plugin.medium_providers.find_provider(host)
+    @medium_provider.medium_uri
+  end
```
lzap, line 179: Please provide a template helper function if we don't have it already to avoid long `@host.bla.bla` and do patch into community templates, we have about 10 occurrences there, not many. It is worth changing it now.

app/models/operatingsystem.rb
```diff
@@ -260,12 +263,17 @@ def deduce_family
...
 
-  def boot_files_uri(medium, architecture, host = nil)
-    raise ::Foreman::Exception.new(N_("%{os} medium was not set for host '%{host}'"), :host => host, :os => self) if medium.nil?
-    raise ::Foreman::Exception.new(N_("Invalid medium '%{medium}' for '%{os}'"), :medium => medium, :os => self) unless media.include?(medium)
-    raise ::Foreman::Exception.new(N_("Invalid architecture '%{arch}' for '%{os}'"), :arch => architecture, :os => self) unless architectures.include?(architecture)
-    self.family.constantize::PXEFILES.values.map do |img|
-      medium_vars_to_uri("#{medium.path}/#{pxedir}/#{img}", architecture.name, self)
+  def boot_files_uri(medium_provider, &block)
+    boot_file_sources(medium_provider, &block).values
+  end
```
lzap, line 268: So in discovery during orchestration I need to find URLs for kexec, what I do:

```ruby
@kexec_kernel, @kexec_initrd = operatingsystem.boot_files_uri(@host.medium, @host.architecture, @host)
```

With this patch I need to do:

```ruby
provider = Foreman::Plugin.medium_providers.find_provider(@host)
@kexec_kernel, @kexec_initrd = operatingsystem.boot_files_uri(provider)
```

app/models/operatingsystems/coreos.rb
```diff
@@ -5,20 +5,26 @@ def pxe_type
...
+    end
   end
 
   def url_for_boot(file)
     PXEFILES[file]
   end
 
   def pxedir
-    '$arch/$version'
+    '$arch-usr/$version'
```
lzap, line 19: What's behind this change?

app/services/medium_providers/default.rb
lzap: I think the proper name would be InterpolatedMediaProvider perhaps.

test/models/operatingsystems/operatingsystems_test.rb
```diff
@@ -30,54 +30,62 @@ class OperatingsystemsTest < ActiveSupport::TestCase
     end
   end
 
-  { :coreos      => { 'os' => :coreos,      'arch' => :x86_64, 'medium' => :coreos, 'expected' => 'boot/CoreOS-494.5.0-x86_64-coreos_production_pxe.vmlinuz' },
-    :debian7_0   => { 'os' => :debian7_0,   'arch' => :x86_64, 'medium' => :unused, 'expected' => 'boot/Debian-7.0-x86_64-linux' },
-    :ubuntu14_10 => { 'os' => :ubuntu14_10, 'arch' => :x86_64, 'medium' => :ubuntu, 'expected' => 'boot/Ubuntu-14.10-x86_64-linux' },
-    :suse        => { 'os' => :suse,        'arch' => :x86_64, 'medium' => :suse,   'expected' => 'boot/OpenSuse-11.4-x86_64-linux' } }.
+  { :coreos      => { 'os' => :coreos,      'arch' => :x86_64, 'medium' => :unused, 'expected' => 'boot/xEClLgEZtX_JEJrrfEoZ9ejnXGI-coreos_production_pxe.vmlinuz' },
```
lzap: Why CoreOS medium is unused now?

app/models/operatingsystem.rb
```diff
@@ -190,27 +171,49 @@ def available_loaders
...
+    @medium_provider = Foreman::Plugin.medium_providers.find_provider(host)
+    @medium_provider.medium_uri
+  end
+
+  # sets the prefix for the tfp files based on medium unique identifier
+  def pxe_prefix(medium_provider)
+    unless medium_provider.is_a? MediumProviders::Provider
+      raise Foreman::Exception.new(N_('Please provide a medium provider. It can be found as @medium_provider in templates, or Foreman::Plugin.medium_providers.find_provider(host)'))
+    end
+    "boot/#{medium_provider.unique_id}".tr(" ", "-")
```
lzap: Why are you doing the `tr`? Base64 will not contain spaces, remove please.

app/models/operatingsystems/windows.rb
```diff
@@ -9,17 +9,15 @@ def pxe_type
     "waik"
   end
 
-  def pxe_prefix(arch)
-    "boot/windows-#{arch}/".tr(" ", "-")
+  def pxe_prefix(medium_provider)
+    medium_provider.interpolate_vars("boot/windows-$arch-#{medium_provider.unique_id}/").tr(" ", "-")
```
lzap, line 13: Why are you doing the `tr`? Base64 will not contain spaces, remove please.

app/models/operatingsystem.rb
```diff
@@ -190,27 +171,49 @@ def available_loaders
...
-    "boot/#{self}-#{arch}".tr(" ", "-")
+  # Compatibility method, don't want to break all templates that use it.
+  def medium_uri(host)
+    Foreman::Deprecation.deprecation_warning("1.20", "medium_uri is now accessible through @medium_provider.medium_uri in templates")
+    @medium_provider = Foreman::Plugin.medium_providers.find_provider(host)
+    @medium_provider.medium_uri
+  end
+
+  # sets the prefix for the tfp files based on medium unique identifier
+  def pxe_prefix(medium_provider)
```
lzap, line 182: Can you actually drop the pxe_prefix-based approach and replace it with something more flexible? This pins us with prefix-based naming scheme with one unique ID. We could allow providers to have arbitrary naming scheme pattern. For example CV name followed by ID or something more readable than hash.

Even default provider could include Media name, this is much more readable: `CentOS-local-mirror-JKLfdlkjdfs78_33flkjADSDADAS-linux` what you think? Just replace all non-alphanum characters with underscore to have FS-safe name and that should do it.

app/services/medium_providers/provider.rb
```diff
@@ -0,0 +1,39 @@
...
+      throw "medium_uri is not implemented for #{self.class.name}"
+    end
+
+    def boot_files_uri
+      throw "boot_files_uri is not implemented for #{self.class.name}"
+    end
+
+    def unique_id
+      @unique_id ||= Base64.urlsafe_encode64(Digest::SHA1.digest(medium_uri.to_s), padding: false)
```
lzap: I think there is way enough entropy for couple of installation media :-) I suggest to shorten the ID to 6 characters, that's still 2^32 possibilities. Chances of collisions are pretty low, if we put media name prefix that's almost zero.

lzap (dismissed):
Ok I have some comments, but looks good.

app/models/operatingsystem.rb
ShimShtein: We already have one. Inside template you can call `medium_provider` and you will get it.

app/models/operatingsystem.rb
ShimShtein: I think `pxe_prefix` is not a very informative name. It returns a path from TFTP root to the filename (with leftmost part of the filename). For example for redhat family, it returns `boot/JKLfdlkjdfs78_33flkjADSDADAS` which later will be appended by `-linux` or `-initrd`.

With this PR we can actually change the filename prefix to whatever we want by changing/overloading `unique_id` function from `MediumProviders::Provider` (base class). It has the whole host visible, so it can construct a name just like it wants. It is the extension point I intend to use in Katello to generate a name based on content view id or whatever unique combination there is.

app/models/operatingsystem.rb
ShimShtein: if `host` isn't set, I can't get `medium_provider`. The function will plainly fail, if only `arch` is specified.

app/models/operatingsystem.rb
ShimShtein: You will be able to use shorter version:

``` ruby
@kexec_kernel, @kexec_initrd = operatingsystem.boot_files_uri(@host.host_medium_provider)
```

app/models/operatingsystems/coreos.rb
ShimShtein: It comes instead of `"#{host.medium.path}/#{host.architecture.name}-usr"` in mediumpath. Now it is more consistent.

app/models/operatingsystems/windows.rb
ShimShtein: Architecture theoretically can have it.

app/services/medium_providers/default.rb
ShimShtein: Interpolated means "insert (something of a different nature) into something else." (it's google, not me). I don't think it is describing the nature of this class. I do think that it's part of default media library implementation that comes in foreman core, this is why I use the term "default", as opposed to Katello's `ManagedContentMediaProvider`.

app/services/medium_providers/provider.rb
ShimShtein: I don't have affinity towards any solution. Personally I don't like taking parts of pseudo-random values, as randomness is not always distributed evenly inside the value. I know it's the case with guids, but I don't have enough knowledge of SHA1.

test/models/operatingsystems/operatingsystems_test.rb
ShimShtein: Because we don't have `:coreos` fixture in `fixtures/media.yml`

ShimShtein:
@lzap I think I addressed everything, even created a community-templates PR.

Can you go for another round of review?

app/services/medium_providers/provider.rb
lzap: This is till a thing I guess @ShimShtein

app/models/operatingsystem.rb
lzap: Let me put it this way: can you change the naming conventions to include media name?

Pattern: `Media_name-Hash-Filename`

Example: `CentOS_mirror-JKLfdlkjdfs78-linux`

lzap:
I have one major comment, otherwise good. In the description you propose this new format:

    boot/Redhat-6.1-x86_64-7a166fa29c0319903fb4a3d40be16c3785f27a96-initrd.img

The current implementation actually creates this:

    boot/JKLfdlkjdfs78_33flkjADSDADAS-initrd.img

I don't like two things in particular in this naming scheme:

* It is unfriendly to humans - we should add OS/version or media name as a prefix. I like the latter.
* The hash is way too long, even in base64 encoding - there is enough entropy IMHO, let's just cut it to 10 characters. I hereby offer you bottle of Becherovka if we hit real-world URL conflict with 10 chars :champagne: 

Suggested format:

    boot/CentOS_mirror-fFdlkjdfs7-initrd.img

ShimShtein:
@lzap Done!
Added documentation and created a helper. `community-templates` were updated too.

ShimShtein:
@lzap, sorry for broken test. Now tests are green.

app/services/medium_providers/default.rb
```diff
@@ -0,0 +1,66 @@
...
+    def interpolate_vars(pattern)
+      medium_vars_to_uri(pattern, entity.architecture.name, entity.operatingsystem)
+    end
+
+    def unique_id
+      @unique_id ||= begin
+        full_uniq = super
+        "#{entity.medium.name.parameterize}-#{full_uniq[1..10]}"
+      end
+    end
```
lzap, line 31: This could easily be still readable:

```ruby
def unique_id
  @unique_id ||= [entity.&medium.&name.&parameterize), super[1..10]].compact.join(',')
end
```

lzap (approved):
I am good, there is one minor comment that is worth another rebase. But since @tbrisker announced quiet pre-branching period I am hesitating merging this as changes are required in Katello and Discovery at least. Don't mind waiting one week after branching? I was on vacation, sorry for the late response.

iNecas:
I think we should this change until the PRs in Katello and Discovery are opened.

lzap:
Not sure what you mean, I am gonna test core and discovery. Can you do katello PR @ShimShtein ?

@iNecas any review comments before I start testing this?

ShimShtein:
@lzap, Katello PR is [ready](https://github.com/Katello/katello/pull/7531).

lzap:
I am testing this, provisioned two hosts with same IM. It created:

nas_centos-aBsbsZ_lDn-initrd.img
nas_centos-aBsbsZ_lDn-vmlinuz

When I change IM URL it creates:

nas_centos-U-AgLxFApV-initrd.img
nas_centos-U-AgLxFApV-vmlinuz

And when I do some Czechish blahish "nas_centos_čášěéíáčšěé čšěčěš" it goes like this:

nas_centos_caseeiacsee-cseces-aBsbsZ_lDn-initrd.img
nas_centos_caseeiacsee-cseces-aBsbsZ_lDn-vmlinuz

This is a GO! Thank you.

lzap:
I am merging tests are green, Katello red but PR is pending and looks ok.

lzap:
Thank you @ShimShtein outstanding work, it was a long journey but I appreciate the work.

lzap:
And I forgot to mark the RedMine...

jturel:
It looks like this change broke something in Katello. Any advice on how to proceed?

http://ci.theforeman.org/blue/organizations/jenkins/katello-pr-test/detail/katello-pr-test/2440/tests

```
Katello::RedhatExtensionsMediaTest#test_medium_uri_with_a_kickstart_repo_hg:                                                                     
Foreman::Exception: ERF42-6627 [Foreman::Exception]: Could not find a provider for testhg. Providers returned {"MediumProviders::Default"=>["RedHat 2.1 medium was not set for host 'testhg'", "Invalid medium '' for 'RedHat 2.1'", "Invalid architecture 'x86_64' for 'RedHat 2.1'"]}            
    app/registries/foreman/plugin/medium_providers_registry.rb:37:in `find_provider'                                                             
    app/models/operatingsystem.rb:177:in `medium_uri'
    /home/vagrant/katello/test/models/concerns/redhat_extensions_test.rb:112:in `test_medium_uri_with_a_kickstart_repo_hg'

```

Do we need to register a medium provider?

lzap:
This fixed it, it was merged quite quick after this PR: https://github.com/Katello/katello/pull/7531

I can't find katello master/develop job in jenkins, is this still a thing?

ares:
I wonder why tests didn't fail, this PR broken develop tests but this should fix it https://github.com/theforeman/foreman/pull/5882

jturel:
Thanks guys, looks like we will have green builds once more! 🥂

PR #5246, opened 2018-02-09
Fixes #22540 - auto-focus search bar

sean797:
This is somewhat subjective, I'm not sure how everyone feels about it

@timogoebel you mentioned it currently takes too many clicks to find a host, please try this out and let us know your thoughts.

@rohoover you may also have an opinion on this.

### DIFF
--- a/app/views/common/_searchbar.html.erb
+++ b/app/views/common/_searchbar.html.erb
@@ -3,7 +3,7 @@
             id: 'search-form',
             onsubmit: 'return tfm.tools.updateTable(this);' do %>
   <div class="input-group">
-    <%= auto_complete_search(:search, params[:search].try(:squeeze," "), :placeholder => _("Filter") + ' ...').html_safe %>
+    <%= auto_complete_search(:search, params[:search].try(:squeeze," "), :placeholder => _("Filter") + ' ...', :autofocus => true).html_safe %>
     <span class="input-group-btn">
       <button class="btn btn-default" type="submit">
         <%= icon_text("search", content_tag(:span,_("Search"), :class=>"hidden-xs")) %>


timogoebel:
@sean797: Thanks for putting effort into this, but I'm afraid this doesn't really solve my issue.

What I think, we need is a global search bar with typeahead. Something like this:

image

I think we need something that is visible on every page and that allows you to navigate to a host just by typing the hostname.

ShimShtein:
@timogoebel, I think I have something for you: https://github.com/ShimShtein/foreman-opensearch. It adds the ability to search foreman from browser's searchbox/omnibox.

dLobatog (approved):
This might not be of everyone's liking, but personally I think it makes sense. 👍 

Let's wait in case someone says otherwise

Rohoover (commented):
I know that there have been talks about a global search but I don't know what the status of it is at this time. I think it could be valuable for users who want to get to something specific but don't want to wade into the UI.

lzap (approved):
I don't like it. I LOVE IT!

ohadlevy:
Does it make sense to have a test that we have only one auto focus input on
this very page we have search for?

On Feb 13, 2018 12:16 PM, "Lukáš Zapletal" <notifications@github.com> wrote:

*@lzap* approved this pull request.

I don't like it. I LOVE IT!

—
You are receiving this because you are subscribed to this thread.
<https://github.com/theforeman/foreman/pull/5246#pullrequestreview-96225524>,
or mute the thread

dLobatog (approved):
@ohadlevy I don't think it's critical to test - from a quick check, a few forms have 'name' as autofocus, but create forms would not have a search bar in general. We can fix any of these anecdotes later if needed/

dLobatog:
Thanks @sean797 !

lzap:
I see one misbehavior - when I load a page and it is properly auto-focus and I switch to a different browser tab and go back, auto-suggestion drop down opens up without typing or anything.

ares:
this happens to me on every new page I enter, Chrome 63, I hope I'm the only one since this is pretty disturbing

sean797:
I'm not sure what we can or should do about that? Feel free to revert if no one has a good solution.

ohadlevy:
:+1: to revert, this makes a horrible ux :(

sean797:
Created https://github.com/theforeman/foreman/pull/5266

timogoebel:
@sean797: Thanks anyway for your attempt. This is a good idea. I think it's the small things that make a large difference.

PR #5249, opened 2018-02-12
Fixes #22556 - Correct host search for parent hostgroup

tbrisker:
Currently, searching on a non-existant parent fails due to
`hostgroup.subtree_ids` throwing a NoMethodError when no matching
hostgroup is found. This is rescued by scoped search, causing the
search to run with no conditions instead.
(see
https://github.com/wvanbergen/scoped_search/blob/v4.1.2/lib/scoped_search/query_builder.rb#L405
for the scoped_search part)


tbrisker:
scoped search part (preventing the silent error suppression) in https://github.com/wvanbergen/scoped_search/pull/173

lzap (approved):
I have dejavu, I think I've seen this piece of code - somewhere... :) Thanks.

PR #5252, opened 2018-02-14
Refs #22354 - Don't instance_eval at every instance creation

iNecas:
Before this patch, we were defining the methods at every instance
creation. This means potential performance issues, as defining new
methods causes the method lookup cache to get dropped.

I needed to move the time for loading the silencer a bit later so that
the `Logging::LEVELS` are already initialized (as they are empty at the
beginning).

### DIFF
--- a/lib/foreman/logging.rb
+++ b/lib/foreman/logging.rb
@@ -1,6 +1,5 @@
 require 'logging'
 require 'fileutils'
-require_dependency File.expand_path('../silenced_logger', __FILE__)
 
 module Foreman
   class LoggingImpl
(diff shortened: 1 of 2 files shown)


lzap (approved):
Oh yeah, I could not figure it out why LEVELS are not available, now I see it.

ares:
since the original PR was set to 1.17, setting the milestone here as well

lzap:
Thanks it works just fine. Good idea to have that in 1.17 as well to avoid confusion.

ares:
@lzap since you merged it, would yoi mind cp to 1.17-stable too? Cc @xprazak2

lzap:
Sure @xprazak2 you want me to do it immediately?

xprazak2:
@lzap, yes please

lzap:
1.17 cherry pick: https://github.com/theforeman/foreman/pull/5260

PR #5254, opened 2018-02-15
Fixes #22601 - Update Rails to 5.1.6

mmoll:

PR #5256, opened 2018-02-17
[WIP] Fixes #22601 - Update Rails to 5.1.5

mmoll:


mmoll:
this is GH-5254 for 1.17

tbrisker:
@ehelms @ekohl @xprazak2 - would we be able to update the SCL in time for 1.17? 
Rails 5.1.5 isn't a security fix, so if this will delay 1.17 in any way I would prefer to ship 1.17 with 5.1.4 and update the SCL only towards 1.18 release (unless a CVE requires a newer version)

test/models/setting_test.rb
```diff
@@ -449,8 +449,6 @@ class CustomSetting < Setting
   end
 
   test "create succeeds when cache is non-functional" do
-    # for unknown reasons this fails in the CI environment, but passes locally
-    skip if ActiveRecord::Base.connection_config[:adapter].eql?("postgresql")
```
mmoll, line 453: This could go in independently, BTW, if 5.1.5 would not make it into 1.17.

xprazak2:
I'd rather avoid doing a SCL update for 1.17 at this stage and leave it for 1.18 since this is not critical in any way.

mmoll:
:+1: closing then...

PR #5258, opened 2018-02-19
Fixes #22615 - template rendering is kept inline

lzap:
In recent change, we refactored the line into more readable plain call,
but it actually broke rendering. I guess we want to keep inline
rendering, it was not main purpose of the original PR to change this.

@timogoebel

### DIFF
--- a/app/controllers/unattended_controller.rb
+++ b/app/controllers/unattended_controller.rb
@@ -249,7 +249,7 @@ class UnattendedController < ApplicationController
     end
 
     begin
-      render :plain => unattended_render(@unsafe_template_content, @template_name)
+      render :inline => "<%= unattended_render(@unsafe_template_content, @template_name).html_safe %>"
     rescue => error
       msg = _("There was an error rendering the %s template: ") % (@template_name)
       Foreman::Logging.exception(msg, error)


app/controllers/unattended_controller.rb
```diff
@@ -249,7 +249,7 @@ def safe_render(template)
     end
 
     begin
-      render :plain => unattended_render(@unsafe_template_content, @template_name)
+      render :inline => "<%= unattended_render(@unsafe_template_content, @template_name).html_safe %>"
```
dLobatog, line 252: Thanks, this works for me on HTTPS, not on HTTP though. Does it work for you?

lzap: What do you mean by that? I don't have HTTPS enabled on my dev setup. Do you mean HTTPS in the foreman_url itself?

dLobatog: Nope, my foreman_url is HTTP. I can get the ks template just fine accessing the token URL on HTTPS, but on HTTP (which is really what I need because.. pxe) it doesn't work for me

dLobatog (approved):
@lzap Took me a while but I found the root cause why I don't got this working with HTTP, it's unrelated to this PR or the previous change. https://github.com/theforeman/forklift/issues/628

Thanks! Merging :medal_military:

ares:
setting the release to 1.18, please don't forget setting the release in redmine after merging

timogoebel:
Sorry guys, for the long delay. I was on PTO. This change was actually pretty important, I'll file a PR to revert this when I find the time.

The inline rendering is useless imho and `TemplatesController` does not use inline rendering. We should use the same technique both in preview mode and the actual unattended rendering to not get different results. In addition, the inline rendering makes the stack trace practically impossible to read and figure out what is actually going wrong.

Before I file the PR: Is this actually breaking anything?

lzap:
Yes, `root_pass` no longer renders with this change. We wanted something quick because it's breaking issue. Feel free to take a look later. Why is this line important please? I was under impression this was consmetic.

timogoebel:
> We wanted something quick because it's breaking issue.

Sure think makes sense. Thanks for taking care of this.

> Why is this line important please?

Well, it should not make any difference. My issue with this is, that it does.

I'll explain my reasoning in the upcoming PR. Let's just agree, that our template rendering is more magic than actual programming right now. It's pure luck, that it works most of the time. It's not surprising, it grew over time and now it needs some love. I will fix this, I just need to fine the time.

PR #5259, opened 2018-02-19
Fixes #22616 - Allow searching hosts by status as integer

tbrisker:

### DIFF
--- a/test/models/host_status/configuration_status_test.rb
+++ b/test/models/host_status/configuration_status_test.rb
@@ -115,4 +115,11 @@ class ConfigurationStatusTest < ActiveSupport::TestCase
     assert_equal '6 & 63', HostStatus::ConfigurationStatus.bit_mask('restarted')
     assert_equal '12 & 63', HostStatus::ConfigurationStatus.bit_mask('failed')
   end
+
+  test 'host search by status works' do
+    @status.save
+    assert_equal [@host], Host.search_for('status.applied = 0')
+    assert_equal [@host], Host.search_for('status.applied = false')
+    assert_equal [], Host.search_for('status.applied = 1')
+  end
 end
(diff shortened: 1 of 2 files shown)


tbrisker:
This was introduced by 3adfb7d1b0338bfa26232b99037a13f9bf4862e9 which breaks built-in searches, once merged it should be cherry-picked to 1.17 as well

lzap:
What is the query to test?

tbrisker:
@lzap most of the queries on the dashboard config status widget and the default bookmarks on the hosts index

lzap:
Yeah I mean what should I type into the search bar, I am trying with `status = ` but it does not work. Is this overall status (like build status) or report status?

tbrisker:
`status.{enabled,applied,...} = 0`

test/models/host_status/configuration_status_test.rb
```diff
@@ -115,4 +115,10 @@ def setup
     assert_equal '6 & 63', HostStatus::ConfigurationStatus.bit_mask('restarted')
     assert_equal '12 & 63', HostStatus::ConfigurationStatus.bit_mask('failed')
   end
+
+  test 'host search by status works' do
+    @status.save
+    assert_equal Host.search_for('status.applied = 0'), [@host]
+    assert_equal Host.search_for('status.applied = false'), [@host]
+  end
```
lzap: Expected value should be the first parameter. Can you also add a positive test when there is a number greater than zero?

lzap (changes requested):
Minor comment but works fine.

test/models/host_status/configuration_status_test.rb
tbrisker: Updated, thanks!

lzap (approved):
Thanks. Waiting for Mr. Jenkins.

PR #5262, opened 2018-02-20
Fixes #22624 - User selectable columns model + api

parthaa:
Adding initial bindings to user selectable columns. This model + api is
going to be used in work related Katello plugin's subscription pages.
The main intent of this PR is to provide a basic model where a user can
store "desirable or interested columns for a resource"
When the user for example says "I want to only see name column in
subscriptions" , we need model that would store that information
(similar to widgets or preferences.)
This commit facilitates that.

### DIFF
--- a/app/controllers/api/v2/table_preferences_controller.rb
+++ b/app/controllers/api/v2/table_preferences_controller.rb
@@ -1,0 +1,59 @@
+module Api
+  module V2
+    class TablePreferencesController < V2::BaseController
+      include Api::Version2
+      include Foreman::Controller::UserAware
+      before_action :check_user
+      before_action :find_resource, :only => [:destroy, :update, :show]
+
+      api :GET, "/users/:user_id/table_preferences", N_("List of table preferences for a user")
+      def index
+        @table_preferences = @user.table_preferences
+      end
+
+      api :GET, "/users/:user_id/table_preferences/:name", N_("Table preference details of a given table")
+      param :name, String, :required => true
+      def show
+        if @table_preference.blank?
+          @table_preference = TablePreference.new(:user => @user, :name => params[:name])
+        end
+      end
+
+      def_param_group :table_preference do
+        param :table_preferences, Hash, :required => true do
+          param :name, String, :required => true, :desc => N_("Name of the table")
+          param :columns, Array, :desc => N_("List of user selected columns")
+        end
+      end
+
+      api :POST, "/users/:user_id/table_preferences/", N_("Creates a table preference for a given table")
+      param_group :table_preference
+      def create
+        @table_preference = @user.table_preferences.build(:name => params[:name], :columns => params[:columns])
+        process_response @table_preference.save
+      end
+
+      api :PUT, "/users/:user_id/table_preferences/:name", N_("Updates a table preference for a given table")
+      param_group :table_preference
+      def update
+        process_response @table_preference.update_attributes(:columns => params[:columns])
+      end
+
+      api :DELETE, "/users/:user_id/table_preferences/:name/", N_("Delete a table preference for a given table")
+      param :name, String, :required => true, :desc => N_("name of the table")
+      def destroy
+        process_response @table_preference.destroy
+      end
+
+      private
+
+      def check_user
+        deny_access N_("You are trying access the preferences of a different user") if @user != User.current
+      end
+
+      def resource_class
+        TablePreference
+      end
+    end
+  end
+end
--- a/app/models/table_preference.rb
+++ b/app/models/table_preference.rb
@@ -1,0 +1,10 @@
+class TablePreference < ApplicationRecord
+  extend FriendlyId
+  friendly_id :name
+  include Parameterizable::ByIdName
+
+  belongs_to :user
+  validates :user_id, :name, :presence => true
+  serialize :columns
+  validates_lengths_from_database
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -42,7 +42,7 @@ class User < ApplicationRecord
   has_many :widgets, :dependent => :destroy
   has_many :ssh_keys, :dependent => :destroy
   has_many :personal_access_tokens, :dependent => :destroy
-
+  has_many :table_preferences, :dependent => :destroy, :inverse_of => :user
   has_many :user_mail_notifications, :dependent => :destroy, :inverse_of => :user
   has_many :mail_notifications, :through => :user_mail_notifications
   has_many :notification_recipients, :dependent => :delete_all
--- a/app/views/api/v2/table_preferences/create.json.rabl
+++ b/app/views/api/v2/table_preferences/create.json.rabl
@@ -1,0 +1,2 @@
+object @table_preference
+extends "api/v2/table_preferences/main"
--- a/app/views/api/v2/table_preferences/destroy.json.rabl
+++ b/app/views/api/v2/table_preferences/destroy.json.rabl
@@ -1,0 +1,2 @@
+object @table_preference
+extends "api/v2/table_preferences/main"
--- a/app/views/api/v2/table_preferences/index.json.rabl
+++ b/app/views/api/v2/table_preferences/index.json.rabl
@@ -1,0 +1,3 @@
+collection @table_preferences
+
+extends "api/v2/table_preferences/main"
--- a/app/views/api/v2/table_preferences/main.json.rabl
+++ b/app/views/api/v2/table_preferences/main.json.rabl
@@ -1,0 +1,2 @@
+object @table_preference
+attributes :id, :name, :columns, :created_at, :updated_at
--- a/app/views/api/v2/table_preferences/show.json.rabl
+++ b/app/views/api/v2/table_preferences/show.json.rabl
@@ -1,0 +1,2 @@
+object @table_preference
+extends "api/v2/table_preferences/main"
--- a/app/views/api/v2/table_preferences/update.json.rabl
+++ b/app/views/api/v2/table_preferences/update.json.rabl
@@ -1,0 +1,2 @@
+object @table_preference
+extends "api/v2/table_preferences/main"
--- a/config/routes/api/v2.rb
+++ b/config/routes/api/v2.rb
@@ -264,6 +264,7 @@ Foreman::Application.routes.draw do
           resources :usergroups, :except => [:new, :edit]
           resources :ssh_keys, :only => [:index, :show, :create, :destroy]
           resources :personal_access_tokens, :only => [:index, :show, :create, :destroy]
+          resources :table_preferences, :only => [:index, :create, :destroy, :show, :update]
         end
       end
 
--- a/db/migrate/20180208053256_create_table_preferences.rb
+++ b/db/migrate/20180208053256_create_table_preferences.rb
@@ -1,0 +1,12 @@
+class CreateTablePreferences < ActiveRecord::Migration[5.1]
+  def change
+    create_table :table_preferences do |t|
+      t.string :name, :limit => 255, :null => false
+      t.text :columns
+      t.timestamps :null => false
+      t.integer :user_id, :null => false
+    end
+    add_foreign_key :table_preferences, :users, :name => "table_preferences_user_id_fk"
+    add_index :table_preferences, [:user_id, :name], :unique => true
+  end
+end
--- a/test/controllers/api/v2/table_preferences_controller_test.rb
+++ b/test/controllers/api/v2/table_preferences_controller_test.rb
@@ -1,0 +1,76 @@
+require 'test_helper'
+
+class Api::V2::TablePreferencesControllerTest < ActionController::TestCase
+  def resources
+    ["test_subscriptions_resource", "test_organizations_resource"]
+  end
+
+  def expected_columns
+    ["1", "2", "3"]
+  end
+
+  def setup
+    setup_users
+  end
+
+  def test_should_NOT_get_index_with_json_result
+    other_user = FactoryBot.create(:user, :with_widget)
+    get :index, params: {user_id: other_user.id}
+    assert_response :forbidden
+  end
+
+  def test_should_get_index_with_json_result
+    get :index, params: {user_id: User.current.id}
+    assert_response :success
+    columns = ActiveSupport::JSON.decode(@response.body)["results"]
+    assert_equal(columns.length, 0)
+  end
+
+  def test_should_show_columns_correctly
+    resource = resources.first
+    TablePreference.create!(user: User.current,
+                       name: resource, columns: expected_columns)
+    get :show, params: {user_id: User.current.id, id: resource}
+    assert_response :success
+    actual_columns = ActiveSupport::JSON.decode(@response.body)["columns"]
+    assert_equal(expected_columns, actual_columns)
+  end
+
+  def test_should_create_correctly
+    resource = resources.first
+    current_user = User.current
+    post :create, params: {user_id: User.current.id, name: resource, columns: expected_columns}
+    assert_response :success
+    columns = current_user.reload.table_preferences.first
+    assert_equal(expected_columns, columns.columns)
+  end
+
+  def test_should_update_correctly
+    resource = resources.first
+    current_user = User.current
+    TablePreference.create!(user: User.current,
+                       name: resource, columns: expected_columns)
+
+    put :update, params: {user_id: User.current.id, id: resource, columns: expected_columns}
+    assert_response :success
+    columns = current_user.reload.table_preferences.first
+    assert_equal(expected_columns, columns.columns)
+  end
+
+  def test_should_destroy_correctly
+    resource1 = resources[0]
+    resource2 = resources[1]
+    current_user = User.current
+    TablePreference.create!(user: User.current,
+                       name: resource1, columns: expected_columns)
+    TablePreference.create!(user: User.current,
+                       name: resource2, columns: expected_columns)
+
+    delete :destroy, params: { user_id: User.current.id, id: resource1}
+    assert_response :success
+
+    columns = current_user.reload.table_preferences
+    assert_equal(1, columns.size)
+    assert_equal(resource2, columns.first.name)
+  end
+end
--- a/test/unit/foreman/access_permissions_test.rb
+++ b/test/unit/foreman/access_permissions_test.rb
@@ -40,7 +40,11 @@ class AccessPermissionsTest < ActiveSupport::TestCase
 
     # Content Security Policy report forwarding endpoint - noop if not configured.
     # See https://github.com/twitter/secureheaders/issues/113
-    "content_security_policy/scribe"
+    "content_security_policy/scribe",
+
+    #table preferences. No special permissions comes with user
+    "api/v2/table_preferences/index", "api/v2/table_preferences/show", "api/v2/table_preferences/create",
+    "api/v2/table_preferences/update", "api/v2/table_preferences/destroy",
   ]
 
   MAY_SKIP_AUTHORIZED = [ "about/index" ]


app/controllers/api/v2/user_columns_controller.rb
```diff
@@ -0,0 +1,21 @@
+module Api
+  module V2
+    class UserColumnsController < V2::BaseController
+      include Api::Version2
+
+      api :GET, "/user_columns/", N_("List of User columns for a user")
```
stbenjam: user, capitalization should be consistent

stbenjam: What about delete and update? Also I feel like` GET /user_columns` should return all the columns for each resource (would it ever make sense to get that info?) and `GET /user_columns/:resource` should return them for a resource

ekohl:
Would it be possible to replace https://github.com/GregSutcliffe/foreman_column_view with this?

app/controllers/api/v2/user_columns_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+      param :resource, String, :required => true
+      def index
+        User.current.user_columns.where(:resource => params[:resource])
+      end
+
+      api :POST, "/user_columns/", N_("Creates a user column for a given resource")
+      param :resource, String, :required => true
+      param :columns, Array, :desc => N_("List of user selected columns")
+      def create
+        User.current.user_columns.where(:resource => params[:resource]).destroy_all
```
stbenjam: CREATE "updates" via this destroy_all?

app/models/user_column.rb
```diff
@@ -0,0 +1,12 @@
+class UserColumn < ApplicationRecord
+  belongs_to :user
+
+  validates :user_id, :resource, :presence => true
+
+  serialize :columns
+
+  # Returns user columns representation as the hash object used in memory
+  def to_hash
+    { :columns => columns, :resource => resource }
```
stbenjam: attr_exportable might be usable here instead, take a look at the exportable concern

app/controllers/api/v2/user_columns_controller.rb
parthaa: Good point will figure out a better way.

tstrachota:
@parthaa do we already have some PR or branch with a table that uses this new functionality? I'm curious about how the UI part looks.

waldenraines:
> @parthaa do we already have some PR or branch with a table that uses this new functionality? I'm curious about how the UI part looks.

@tstrachota we will be using this on the RH subscriptions page:

cff781297148a40076394701ba0c71de15cd4d4f

Due to time constraints we aren't planning on using this anywhere in foreman yet.

parthaa:
Here are the Actions implemented so far
- Create
```

$ curl -k -u admin:changeme -H "Content-Type: application/json" -X POST -d '{"resource":"organizations", "columns":["name"]}' "https://localhost/api/v2/user_columns"

{"resource":"organizations","columns":["name"],"created_at":"2018-03-02 01:14:14 UTC","updated_at":"2018-03-02 01:14:14 UTC","user":{"id":4,"login":"admin","description":null}}

$ curl -k -u admin:changeme -H "Content-Type: application/json" -X POST -d '{"resource":"subscriptions", "columns":["name","account"]}' "https://localhost/api/v2/user_columns"

{"resource":"subscriptions","columns":["name","account"],"created_at":"2018-03-02 01:14:43 UTC","updated_at":"2018-03-02 01:14:43 UTC","user":{"id":4,"login":"admin","description":null}}

```
- Index
```
$ curl -k -u admin:changeme -H "Content-Type: application/json" -X GET "https://localhost/api/v2/user_columns"
{
  "total": 2,
  "subtotal": 2,
  "page": 1,
  "per_page": 20,
  "search": null,
  "sort": {
    "by": null,
    "order": null
  },
  "results": [{"resource":"subscriptions","columns":["name","account"],"created_at":"2018-03-02 01:14:43 UTC","updated_at":"2018-03-02 01:14:43 UTC","user":{"id":4,"login":"admin","description":null}},{"resource":"organizations","columns":["name"],"created_at":"2018-03-02 01:14:14 UTC","updated_at":"2018-03-02 01:14:14 UTC","user":{"id":4,"login":"admin","description":null}}]
}

```

- Information about an individual resource
```
$ curl -k -u admin:changeme -H "Content-Type: application/json" -X GET -d '{"resource":"subscriptions"}'   "https://localhost/api/v2/user_columns/columns"

{"results":["name","account"]}
```

- Reset user columns for a given resource
```
$ curl -k -u admin:changeme -H "Content-Type: application/json" -X DELETE -d '{"resource":"subscriptions"}' "https://localhost/api/v2/user_columns/reset"
[]

#verified by
$ curl -k -u admin:changeme -H "Content-Type: application/json" -X GET -d '{"resource":"subscriptions"}'   "https://localhost/api/v2/user_columns/columns"
{"results":[]}

$ curl -k -u admin:changeme -H "Content-Type: application/json" -X GET -d '{"resource":"organizations"}'   "https://localhost/api/v2/user_columns/columns"
{"results":["name"]}

```

- Destroy all user columns for a user

```
$ curl -k -u admin:changeme -H "Content-Type: application/json" -X DELETE    "https://localhost/api/v2/user_columns/destroy_all"
[]

# Verified By

$ curl -k -u admin:changeme -H "Content-Type: application/json" -X GET "https://localhost/api/v2/user_columns"
{
  "total": 0,
  "subtotal": 0,
  "page": 1,
  "per_page": 20,
  "search": null,
  "sort": {
    "by": null,
    "order": null
  },
  "results": []
}
```

parthaa:
Note: I have listed the intended workflow with the Katello subscriptions UI here 
https://github.com/Katello/katello/pull/7219#issuecomment-369791829

iNecas:
This approach looks good to me. I would like to see more +1 from on this one, to make sure this doesn't go against any other plans folks have, as it's introducing new API point, that might be complicated to update later. Any other thoughts, @theforeman/core ?

jlsherrill:
This doesn't have any way to have default columns defined in any way, which means:
1.  fetching the colums for a fresh user would return nothing
2.  Setting invalid columns can't be validated 

i think there needs to be a registry where the app or plugin can define what columns are available for a resource (and the default enablement for them).

parthaa:
> This doesn't have any way to have default columns defined in any way, which means:
> 
> fetching the colums for a fresh user would return nothing
> Setting invalid columns can't be validated
> i think there needs to be a registry where the app or plugin can define what columns are available for a resource (and the default enablement for them).

Created a registry and a plugin api to make this work. Added validation and updated controller tests for this to work. 
Check out -> https://github.com/Katello/katello/pull/7219/files#diff-84f0b46367dabf5e0710a0fe3aca7627R233 for more info o this.

ShimShtein:
Generally, the idea is very interesting. A couple of thoughts on the subject:
1. How well will this play with graphQL? I know there's an effort to add support to it. /cc @ik5 
2. As far as I can see, this filter is not applied at the server level, meaning no optimization to queries. So I see it solves only eye clutter for the user, but not processing time.

parthaa:
> How well will this play with graphQL? I know there's an effort to add support to it. /cc @ik5
> As far as I can see, this filter is not applied at the server level, meaning no optimization to queries. So I see it solves only eye clutter for the user, but not processing time.

Good points. However this is not meant to replace graphql in terms of optimization. I would think this will work well in conjunction with graphql whenever that aspect is implemented in a mainstream way.  For example if we have information on what columns we want, the api javascript will tell server what it exactly wants using GraphQL. This is purely for the UI (and possibly hammer in the future.)  It is merely intended to provide and store enough information for that.

timogoebel (dismissed):
@parthaa: Thanks, two small comments:
Can we please not add an another [registry](http://projects.theforeman.org/projects/foreman/wiki/Registry_pattern) but use one of the existing registries?

Doesn't it make sense to store this in the browser's local storage? We already use that for the "expand the nav" hamburger. And other tools like AWS do it in a similar way.

app/controllers/api/v2/user_columns_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+
+      api :GET, "/user_columns/", N_("List of user columns for a user")
+      def index
+        @user_columns = User.current.user_columns
+      end
+
+      api :GET, "/user_columns/:resource", N_("List of user columns for a resource")
+      param :resource, String, :required => true
+      def columns
+        render :json => { root_node_name => UserColumn.formatted_columns(params[:resource])}
```
ohadlevy: does it make more sense to reuse rabl?

app/controllers/api/v2/user_columns_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+      before_action :find_user_column, :only => [:reset]
+
+      api :GET, "/user_columns/", N_("List of user columns for a user")
+      def index
+        @user_columns = User.current.user_columns
+      end
+
+      api :GET, "/user_columns/:resource", N_("List of user columns for a resource")
+      param :resource, String, :required => true
+      def columns
```
ohadlevy: why deviate from REST? (index, create, update, show, destroy)?

app/controllers/api/v2/user_columns_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+      def create
+        @user_column = UserColumn.where(:user => User.current.id, :resource => params[:resource]).first_or_create
+        @user_column.columns = params[:columns]
+        process_response @user_column.save
+      end
+
+      api :DELETE, "/user_columns", N_("Delete columns for a resource")
+      param :resource, String, :required => true, :desc => N_("name of the resource")
+      def reset
+        process_response @user_column.destroy_all
```
ohadlevy: this will always remove all? just like destroy_all?

in general I would prefer to have one destroy method that get an array of ids, so it can delete one, many or all.

app/controllers/api/v2/user_columns_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+      api :DELETE, "/user_columns", N_("Delete all columns for the user")
+      def destroy_all
+        @user_column = User.current.user_columns
+        process_response @user_column.destroy_all
+      end
+
+      private
+
+      def find_user_column
+        @user_column = User.current.user_columns.where(:resource => params[:resource])
```
ohadlevy: does it make sense to move to a scope? I see we use it in multiple places.

app/models/user_column.rb
```diff
@@ -0,0 +1,40 @@
...
+  validates :user_id, :resource, :presence => true
+  serialize :columns
+  attr_exportable :resource, :columns
+  validates :resource, :inclusion => {:in => ::UserColumnRegistry::Manager.resources.map(&:to_s)}
+  validate :ensure_columns_valid, :if => proc { |r| r.columns.present? && r.resource.present? }
+
+  def ensure_columns_valid
+    valid_columns = ::UserColumnRegistry::Manager.columns(resource) || []
+    valid_column_names = valid_columns.map(&:name)
+    invalid_columns = columns.select{ |col| !valid_column_names.include? col.to_s }
```
ohadlevy: instead of looping over all columns, you can use something like
```rb
invalid = columns - valid_common_names
```

app/models/user_column.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def self.formatted_columns(resource)
+    registry_columns = UserColumnRegistry::Manager.columns(resource)
+    return [] if registry_columns.blank?
+
+    current_user_columns = nil
+
+    if User.current.present?
+      uc = User.current.user_columns.where(:resource => resource).first
```
ohadlevy: nitpick: I find variable ```uc``` name confusing, how about simple columns?
consider using scope instead of where.

app/models/user_column.rb
```diff
@@ -0,0 +1,40 @@
...
+      current_user_columns = uc.columns if uc && uc.columns
```
ohadlevy: can use ```uc.try(:columns)```

app/models/user_column.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+
+    registry_columns.map do |column|
+      formatted_column = column.to_hash
+      formatted_column[:enabled] = if current_user_columns.nil?
```
ohadlevy: since columns is serialized, empty string will probably be better to use ```.blank?``` vs ```.nil?```

app/registries/foreman/access_permissions.rb
```diff
@@ -720,4 +720,8 @@
     map.permission :revoke_personal_access_tokens,
                    :"api/v2/personal_access_tokens" => [:destroy]
   end
+
+  permission_set.security_block :user_columns do |map|
+    map.permission :access_user_columns, {:"api/v2/user_columns" => [:index, :create, :columns, :reset, :destroy_all]}
```
ohadlevy: does it make more sense to be ```api/v2/user/id/columns```?

ohadlevy (changes requested):
thanks @parthaa, some comments inline.

I do agree with @timogoebel we should consider local storage, however i guess that we still need the definition of the columns somewhere..?

waldenraines:
> Doesn't it make sense to store this in the browser's local storage?

@timogoebel @ohadlevy persisting the column selections gives a better user experience because it would be rather confusing for a user to have to select their preferred columns again after initially setting them if they user another browser or otherwise clear the local storage.  Also if we store this in the browser's local storage then the columns will be different in the CLI.

We also want to expand this in the future to support other table configuration features such as column ordering.  Not persisting these items would be frustrating to users who want to customize their table view once and have it shown the same from thereon out.

parthaa:
> Can we please not add an another registry but use one of the existing registries?
> 
> Doesn't it make sense to store this in the browser's local storage? We already use that for the "expand the nav" hamburger. And other tools like AWS do it in a similar way.

Hmm good points but as @waldenraines and @ohadlevy pointed out we might want to persist it in the db somewhere and need to define the valid options in a centralized location. IMO a registry is suitable option. I am open for adding it to an existing registry if you can suggest how/which one :).

parthaa:
@ohadlevy here are some of the changes I did following your suggestions
1) removed columns and made the show return the data I want
```
curl -k -u admin:changeme -H "Content-Type: application/json" -X GET -d '{"resource":"Katello::Pool"}'   "https://localhost/api/v2/user_columns"
[
{"name":"name","description":"Name","path":"name","default_enabled":true,"enabled":true},
{"name":"account","description":"Account","path":"account_number","default_enabled":true,"enabled":true},
{"name":"contract","description":"Contract","path":"contract_number","default_enabled":true,"enabled":true},
{"name":"start_date","description":"Start Date","path":"start_date","default_enabled":true,"enabled":true},{"name":"end_date","description":"End Date","path":"end_date","default_enabled":true,"enabled":true},{"name":"entitlements","description":"Entitlements","path":"consumed","default_enabled":true,"enabled":true}
]
```
Now one difference here between this method and the standard "show" would be that this would always return a "collated" result. As in it will take the user 's preferences and match with data from the registry and return the json with with appropriately "enabled" columns.

2)  Removed the confusing destroy_all and renamed reset to destroy. So you should be able to simply be able to delete  
```
curl -k -u admin:changeme -H "Content-Type: application/json" -X DELETE -d '{"resource":"Katello::Pool"}'   "https://localhost/api/v2/user_columns"
```

Also addressed the model suggestions you had.

parthaa:
After further thought I decided to make the output of show consistent with what I return for create and index. So here are the updated outputs.

- POST/Create
```
$ curl -k -u admin:changeme -H "Content-Type: application/json" -X POST -d '{"resource":"Katello::Pool", "columns":["name", "entitlements"]}'   "https://localhost/api/v2/user_columns"


{:resource=>"Katello::Pool",
 :columns=>["name", "entitlements"],
 :created_at=>"2018-03-04 04:00:46 UTC",
 :updated_at=>"2018-03-04 04:00:46 UTC",
 :collated=>
  [{:name=>"name",
    :description=>"Name",
    :path=>"name",
    :default_enabled=>true,
    :enabled=>true},
   {:name=>"account",
    :description=>"Account",
    :path=>"account_number",
    :default_enabled=>true,
    :enabled=>false},
   {:name=>"contract",
    :description=>"Contract",
    :path=>"contract_number",
    :default_enabled=>true,
    :enabled=>false},
   {:name=>"start_date",
    :description=>"Start Date",
    :path=>"start_date",
    :default_enabled=>true,
    :enabled=>false},
   {:name=>"end_date",
    :description=>"End Date",
    :path=>"end_date",
    :default_enabled=>true,
    :enabled=>false},
   {:name=>"entitlements",
    :description=>"Entitlements",
    :path=>"consumed",
    :default_enabled=>true,
    :enabled=>true}],
 :user=>{:id=>4, :login=>"admin", :description=>nil}}
```
- Get
```
$ curl -k -u admin:changeme -H "Content-Type: application/json" -X GET "https://localhost/api/v2/user_columns/Katello::Pool"

{:resource=>"Katello::Pool",
 :columns=>["name", "entitlements"],
 :created_at=>"2018-03-04 04:00:46 UTC",
 :updated_at=>"2018-03-04 04:00:46 UTC",
 :collated=>
  [{:name=>"name",
    :description=>"Name",
    :path=>"name",
    :default_enabled=>true,
    :enabled=>true},
   {:name=>"account",
    :description=>"Account",
    :path=>"account_number",
    :default_enabled=>true,
    :enabled=>false},
   {:name=>"contract",
    :description=>"Contract",
    :path=>"contract_number",
    :default_enabled=>true,
    :enabled=>false},
   {:name=>"start_date",
    :description=>"Start Date",
    :path=>"start_date",
    :default_enabled=>true,
    :enabled=>false},
   {:name=>"end_date",
    :description=>"End Date",
    :path=>"end_date",
    :default_enabled=>true,
    :enabled=>false},
   {:name=>"entitlements",
    :description=>"Entitlements",
    :path=>"consumed",
    :default_enabled=>true,
    :enabled=>true}],
 :user=>{:id=>4, :login=>"admin", :description=>nil}}

```
- Index
```
$ curl -k -u admin:changeme -H "Content-Type: application/json" -X GET    "https://localhost/api/v2/user_columns"
{:total=>1,
 :subtotal=>1,
 :page=>1,
 :per_page=>20,
 :search=>nil,
 :sort=>{:by=>nil, :order=>nil},
 :results=>
  [{:resource=>"Katello::Pool",
    :columns=>["name", "entitlements"],
    :created_at=>"2018-03-04 04:00:46 UTC",
    :updated_at=>"2018-03-04 04:00:46 UTC",
    :collated=>
     [{:name=>"name",
       :description=>"Name",
       :path=>"name",
       :default_enabled=>true,
       :enabled=>true},
      {:name=>"account",
       :description=>"Account",
       :path=>"account_number",
       :default_enabled=>true,
       :enabled=>false},
      {:name=>"contract",
       :description=>"Contract",
       :path=>"contract_number",
       :default_enabled=>true,
       :enabled=>false},
      {:name=>"start_date",
       :description=>"Start Date",
       :path=>"start_date",
       :default_enabled=>true,
       :enabled=>false},
      {:name=>"end_date",
       :description=>"End Date",
       :path=>"end_date",
       :default_enabled=>true,
       :enabled=>false},
      {:name=>"entitlements",
       :description=>"Entitlements",
       :path=>"consumed",
       :default_enabled=>true,
       :enabled=>true}],
    :user=>{:id=>4, :login=>"admin", :description=>nil}}]}

```
1. Update
```
$ curl -k -u admin:changeme -H "Content-Type: application/json" -X PUT -d '{ "columns":["name", "contract"]}'   "https://localhost/api/v2/user_columns/Katello::Pool"
{:resource=>"Katello::Pool",
 :columns=>["name", "contract"],
 :created_at=>"2018-03-04 04:00:46 UTC",
 :updated_at=>"2018-03-04 05:19:35 UTC",
 :collated=>
  [{:name=>"name",
    :description=>"Name",
    :path=>"name",
    :default_enabled=>true,
    :enabled=>true},
   {:name=>"account",
    :description=>"Account",
    :path=>"account_number",
    :default_enabled=>true,
    :enabled=>false},
   {:name=>"contract",
    :description=>"Contract",
    :path=>"contract_number",
    :default_enabled=>true,
    :enabled=>true},
   {:name=>"start_date",
    :description=>"Start Date",
    :path=>"start_date",
    :default_enabled=>true,
    :enabled=>false},
   {:name=>"end_date",
    :description=>"End Date",
    :path=>"end_date",
    :default_enabled=>true,
    :enabled=>false},
   {:name=>"entitlements",
    :description=>"Entitlements",
    :path=>"consumed",
    :default_enabled=>true,
    :enabled=>false}],
 :user=>{:id=>4, :login=>"admin", :description=>nil}}

```
- Delete
```
$ curl -k -u admin:changeme -H "Content-Type: application/json" -X DELETE    "https://localhost/api/v2/user_columns/Katello::Pool"

{:resource=>"Katello::Pool",
 :columns=>["name", "contract"],
 :created_at=>"2018-03-04 04:00:46 UTC",
 :updated_at=>"2018-03-04 05:19:35 UTC",
 :collated=>
  [{:name=>"name",
    :description=>"Name",
    :path=>"name",
    :default_enabled=>true,
    :enabled=>true},
   {:name=>"account",
    :description=>"Account",
    :path=>"account_number",
    :default_enabled=>true,
    :enabled=>false},
   {:name=>"contract",
    :description=>"Contract",
    :path=>"contract_number",
    :default_enabled=>true,
    :enabled=>true},
   {:name=>"start_date",
    :description=>"Start Date",
    :path=>"start_date",
    :default_enabled=>true,
    :enabled=>false},
   {:name=>"end_date",
    :description=>"End Date",
    :path=>"end_date",
    :default_enabled=>true,
    :enabled=>false},
   {:name=>"entitlements",
    :description=>"Entitlements",
    :path=>"consumed",
    :default_enabled=>true,
    :enabled=>false}],
 :user=>{:id=>4, :login=>"admin", :description=>nil}}

```

- And now if you did a GET notice the difference in output from the previous Get. Notice how the columns, created, updated all are null but the "collated" has its enabled matching the "default_enabled". Idea is that the user will do a get and always use the "collated" data

```
$ curl -k -u admin:changeme -H "Content-Type: application/json" -X GET "https://localhost/api/v2/user_columns/Katello::Pool"

{:resource=>"Katello::Pool",
 :columns=>nil,
 :created_at=>nil,
 :updated_at=>nil,
 :collated=>
  [{:name=>"name",
    :description=>"Name",
    :path=>"name",
    :default_enabled=>true,
    :enabled=>true},
   {:name=>"account",
    :description=>"Account",
    :path=>"account_number",
    :default_enabled=>true,
    :enabled=>true},
   {:name=>"contract",
    :description=>"Contract",
    :path=>"contract_number",
    :default_enabled=>true,
    :enabled=>true},
   {:name=>"start_date",
    :description=>"Start Date",
    :path=>"start_date",
    :default_enabled=>true,
    :enabled=>true},
   {:name=>"end_date",
    :description=>"End Date",
    :path=>"end_date",
    :default_enabled=>true,
    :enabled=>true},
   {:name=>"entitlements",
    :description=>"Entitlements",
    :path=>"consumed",
    :default_enabled=>true,
    :enabled=>true}],
 :user=>{:id=>4, :login=>"admin", :description=>nil}}

```

parthaa:
@timogoebel @ohadlevy I updated the registry code to now use the foreman_plugin registry facility (similar to the Dashboard Widgets.) Checkout the changes in foreman/app/services/column_registry/manager.rb https://github.com/theforeman/foreman/pull/5262/files#diff-0d632619a31045efde25ad6cb54c0032

Hope that is closer to the the intended way.

parthaa:
[test foreman]

parthaa:
[test upgrade]

tstrachota:
I have a feeling that we're over-engineering this a bit. It's nothing but settings about what columns are visible and I believe it should be modeled a bit simpler. 

I don't think it's a good idea to bind the column preferences to resources. Even though we usually display one table per resource, there could easily be a table that shows information from multiple sources or two different tables for one resource. I don't want to lock this path for us. Instead, the setting should be bound to a table. 

For similar reasons I think that column title as well as the logic what value is displayed should be stored in the frontend js code. Defining the path to the data is quite limiting and the column preferences shouldn't know about it. Columns can be more complicated than just displaying a value (e.g. host or proxy status). The same applies to default visiblity - I think it should be defined in the client side code to keep it simple.

I think that a model where `User` has_many `TablePreferences(columns: [])` that belong_to `UITable(name: "")` should be sufficient. New tables could be then added via seeds.

parthaa:
[test foreman]

parthaa:
> I have a feeling that we're over-engineering this a bit. It's nothing but settings about what columns are visible and I believe it should be modeled a bit simpler

The huge json outputs posted above might make it look bigger than it is. But I 'd argue what I have is not far from what you are saying. UserColumns indeed the same as TablePreferences(columns: [])  that you are referring.

 Now if you are saying the registry part, a central place to define what the resources should be or what the columns should be are an overkill that is something we can look at. 

1) I modeled this after the approach we took for  Dashboard widgets. Menus/Dashboard widgets all are extendable via the "plugin registry" and provide a good spot for plugins like katello/discovery etc to add their own widgets.
2) Having a centralized location for defining these things helps us validate things for things like typos in the name etc.

I'd recommend looking at what realistically a user of this has to do by checking out the katello PR. Once this infrastructure is in. As a user all you 'd need to add from a plugin perspective
```
  column_resource ::Katello::Pool.name do |resource|
    resource.column 'name', :description => N_("Name"),
                            :path => "name",
                            :default_enabled => true

    resource.column 'account', :description => N_("Account"),
                               :path => "account_number",
                               :default_enabled => true

    resource.column 'contract', :description => N_("Contract"),
                                :path => "contract_number",
                                :default_enabled => true

    resource.column 'start_date', :description => N_("Start Date"),
                                  :path => "start_date",
                                  :default_enabled => true

    resource.column 'end_date', :description => N_("End Date"),
                                :path => "end_date",
                                :default_enabled => true

    resource.column 'entitlements', :description => N_("Entitlements"),
                                    :path => "consumed",
                                    :default_enabled => true
  end

```
 and you get the validation for free.
That being said I am totally open to suggestions and can live with not having any of those facilities. @tstrachota , @waldenraines , @jlsherrill  @timogoebel @ohadlevy  kindly let me know.

parthaa:
@tstrachota also one thing with the Description field being set in column definition is that we can use it in the CLI also (whenever that happens). If you need to tune the UI and change things like formatting or column names feel free to use the 'name'  attribute and translate accordingly.

waldenraines:
> For similar reasons I think that column title as well as the logic what value is displayed should be stored in the frontend js code. Defining the path to the data is quite limiting and the column preferences shouldn't know about it. Columns can be more complicated than just displaying a value (e.g. host or proxy status). The same applies to default visiblity - I think it should be defined in the client side code to keep it simple.
> 
> I think that a model where User has_many TablePreferences(columns: []) that belong_to UITable(name: "") should be sufficient. New tables could be then added via seeds.

These are good points.  The preferences will need to be a little more complex than just a list of columns though as we will support saving column ordering (and potentially sorting) in the future.  Not having the titles in the API means that we would have to store mappings between value and column in multiple places but we are already doing this and having the titles in the API could be an unnecessary optimization at the cost of flexibility.

tstrachota:
I'm also against storing the logic at multiple places. And my concerns are that the current solution makes the situation even worse. We already know that it won't be 100% possible to store the logic for rendering the columns in the ruby DSL. UI is (or is going to be) a lot richer. Tables contain more formatting, tooltips, tables with inline edits, etc. which will always live in the js client code. I'd like to keep all the logic there and use the API only as a service for storing the preferences, not the definitions. Moving the column definitions to the API will in fact result in having the logic at two places.

I'm not sure I understand how saving titles in the API will help implementing such functionality in CLI. Similar to UI, hammer has it's own logic for formatting table cells and fetching additional data for a column. I understand that long term we'd like the CLI to provide the same information as the UI, but how we process the data before it's printed will always be different.

I believe that we should really keep this simple:

- API provides an endpoint for storing user's preferences for a table. At the moment it's going to be only what columns are visible. Later we can add sorting and column order.
- React tables will know what columns are available and what should be displayed by default. Each column has it's name like `hosts_count`, `url`. In future, plugins will be able to extend the js code and add their own columns.
- At the load time, a table will fetch the user's preferences from the API and display/hide the columns. If there are no preferences saved, it will just use the defaults. If there's a column name that don't exist in the table, it will just ignore it (this can happen e.g. when a plugin was uninstalled).
- Changing the visible columns in the UI will trigger an API request with a new list of visible columns.

I don't think we need additional registry for available columns. UI tables (or hammer) should take care of it themselves.

ares:
I agree with @tstrachota. Also if you want to get this in soon so that next Katello release can rely on it, the simpler approach will be better. If we over-complicate the engine in 1.18, especially without any table in core using it, plugins might start using it and force us to remain with that solution even if that does not fit well. It would be good if someone at least tried this solution with one table in core before this gets in.

Putting all display logic into JS also seems to be aligned with long-term UI strategy. JS part should know how to display data, while server should know what data.

parthaa:
After discussion with @tstrachota @waldenraines @jlsherrill the updated PR does the following
- No registry or any of the metadata on resources asutff
- UserColumns is just a simple model to store a name and column data (controllers for that are still on)

db/seeds.d/020-roles_list.rb
```diff
@@ -26,7 +26,7 @@ def seeded_roles
                                              :view_smart_proxies, :edit_smart_proxies, :view_subnets, :edit_subnets,
                                              :view_statistics, :view_usergroups, :create_usergroups, :edit_usergroups,
                                              :destroy_usergroups, :view_users, :edit_users, :view_realms, :view_mail_notifications,
-                                             :view_params, :view_ssh_keys, :view_personal_access_tokens],
+                                             :view_params, :view_ssh_keys, :view_personal_access_tokens, :access_user_columns],
```
ares: This should be added to viewer/manager/org admin roles. Or even better, it shoud be part of default role, there's no point in not having this permission for any user.

app/models/user_column.rb
```diff
@@ -0,0 +1,12 @@
...
+  include Exportable
+  extend FriendlyId
+  friendly_id :name
+  include Parameterizable::ByIdName
+
+  belongs_to :user
+  validates :user_id, :name, :presence => true
+  serialize :columns
+  attr_exportable :name, :columns
+  scope :find_by_name, ->(name) { where(:name => name)}
```
ares: Isn't this provided by rails? Iirc it is and it returns single record. Why do you define such scope? It's equivalent to `.where(:name => name)`

db/seeds.d/020-roles_list.rb
parthaa: > This should be added to viewer/manager/org admin roles. Or even better, it shoud be part of default role, there's no point in not having this permission for any user.

Thanks for the pointer. Updated :+1:

parthaa:
[test foreman]

timogoebel:
> @timogoebel @ohadlevy persisting the column selections gives a better user experience because it would be rather confusing for a user to have to select their preferred columns again after initially setting them if they user another browser or otherwise clear the local storage. Also if we store this in the browser's local storage then the columns will be different in the CLI.

I think storing the data in the browser would actually improve the user experience. I would not want the columns in the CLI to match the colums in the UI. This would be really confusing for a sysadmin and it has the potential to break scripts etc. CLI settings belong in a config file in the user's home directory.

If we store the preferences in the browser, a user can switch accounts and still have his preferred columns selected. If the user uses Foreman on a phone, he can select different columns there then on his laptop. And so forth.

ares:
I guess both has advantages and disadvantages. I think it's good to have generic preference stored on server side, this concept could be later extended to more things than just table columns. If we see that user have different needs on phone/CLI/browser then we can further extend this, but this should address the more generic problem. Users interested in cfgmgmt most likely want to see puppet status and do not care about e.g. openscap artifacts. The security person is quite the opposite. These preferences seems to be universal for all devices/clients, while they could be even more tailored, e.g. on phone I want to see openscap status but I'm not interested in current openscap policy. But that's imho a separate improvement.

db/migrate/20180208053256_create_user_columns.rb
```diff
@@ -0,0 +1,13 @@
+class CreateUserColumns < ActiveRecord::Migration[5.1]
+  def change
+    create_table :user_columns do |t|
+      t.string :name, :limit => 255, :null => false
+      t.text :columns
+      t.timestamps :null => false
+      t.integer :user_id, :null => false
+    end
+    add_foreign_key :user_columns, :users, :name => "user_columns_user_id_fk"
+    add_index :user_columns, :user_id
```
tbrisker: This index is redundant, DB will know to use the partial index from below.

app/controllers/concerns/foreman/controller/parameters/user_column.rb
```diff
@@ -0,0 +1,15 @@
+module Foreman::Controller::Parameters::UserColumn
```
tbrisker: there is no need for strong param filtering if you don't use mass assignment of the values.

app/models/user_column.rb
```diff
@@ -0,0 +1,13 @@
+class UserColumn < ApplicationRecord
+  include Authorizable
```
tbrisker: I am not sure this needs to be Authorizable, all users be able to have access to their own columns and we don't really check the permission anywhere other then for controller access.

app/models/user_column.rb
```diff
@@ -0,0 +1,13 @@
...
+  include Exportable
```
tbrisker: Is this needed? do we have any use case for exporting user columns?

app/views/api/v2/user_columns/main.json.rabl
```diff
@@ -0,0 +1,5 @@
+object @user_column
+attributes :id, :name, :columns, :created_at, :updated_at
+child :user do
```
tbrisker: Are you sure we need to include the user in the user_column view?

db/migrate/20180208053256_create_user_columns.rb
```diff
@@ -0,0 +1,13 @@
+class CreateUserColumns < ActiveRecord::Migration[5.1]
+  def change
+    create_table :user_columns do |t|
+      t.string :name, :limit => 255, :null => false
```
tbrisker: Shouldn't this be `resource` or `controller` or something more descriptive? it took me a bit to understand what the name means.

tbrisker (changes requested):
While I'm not sure about my opinion regarding the discussion above, see a few comments inline regarding the code itself.

app/models/user_column.rb
parthaa: that came because of  https://github.com/theforeman/foreman/pull/5262#discussion_r169331051

db/migrate/20180208053256_create_user_columns.rb
parthaa: I had it as resource/controller before @tstrachota felt on some pages we'd want to have more than 1 set of table preferences.  So I changed it. I saw 'name' as something more generic (like a key.)  Its entirely upto the UI to figure out what to put there.

app/views/api/v2/user_columns/main.json.rabl
parthaa: have no issues taking this out

waldenraines:
> I think storing the data in the browser would actually improve the user experience. I would not want the columns in the CLI to match the colums in the UI. This would be really confusing for a sysadmin and it has the potential to break scripts etc. CLI settings belong in a config file in the user's home directory.

At the moment there are no plans to implement column customization in the CLI but we could allow these preferences to be customized by client as well if we wanted.

> If we store the preferences in the browser, a user can switch accounts and still have his preferred columns selected. If the user uses Foreman on a phone, he can select different columns there then on his laptop. And so forth.

As far as I know the plan is to store these preferences for each user.  With browser storage we have the problem of columns "randomly" resetting when the browser storage is invalidated.  I think this would be confusing to the user and, to me, it doesn't seem worth the benefit of being able to set columns on each device.

db/migrate/20180208053256_create_user_columns.rb
tstrachota: Yes, I think it's better to use a generic name and don't limit it only to resources or controllers.

tstrachota: Maybe `table_name` could work if we want to be a bit more specific.

tstrachota:
This simplified version looks good to me. We can always extend it later. Thanks for the fast updates @parthaa! I didn't have a chance to test it yet though.

I have a few more ideas about some less important cosmetic changes. I'd like to know other's opinions on them:
- Since we discussed that we could theoretically use it for saving sorting and also column order, how about calling the model `TablePreferences` ?
- How about nesting the route under user: `/users/:id/columns/` or `/users/:id/table_preferences/` ?

parthaa:
> Since we discussed that we could theoretically use it for saving sorting and also column order, how about calling the model TablePreferences ?
done. Renamed user columns to table preferences.

> How about nesting the route under user: /users/:id/columns/ or /users/:id/table_preferences/ ?

Why do you need it nested there? Its always the 'current users' preferences right ?

timogoebel:
>> How about nesting the route under user: /users/:id/columns/ or /users/:id/table_preferences/ ?
> 
> Why do you need it nested there? Its always the 'current users' preferences right ?

What if (let's say as an admin) I want to set the preferences for another user. Should we support this? I think so.

parthaa:
> What if (let's say as an admin) I want to set the preferences for another user. Should we support this? I think so.

Hmm.That sounds like new functionality. Shouldn't we add that route change ideally when we decide to add kinda thing?

parthaa:
> Hmm.That sounds like new functionality. Shouldn't we add that route change ideally when we decide to add kinda thing?

I 'll look into it

parthaa:
> What if (let's say as an admin) I want to set the preferences for another user. Should we support this? I think so.

ok decided to go with this suggestion. I see other user prefernce items like "personal access tokens" and "ssh_keys" being handled similarly.

db/migrate/20180208053256_create_user_columns.rb
parthaa: I renamed the model to TablePreferences.  I think that should be enough to indicate name = tablename imo

parthaa:
> How about nesting the route under user: /users/:id/columns/ or /users/:id/table_preferences/ ?

done

parthaa:
[test foreman]

parthaa:
[test foreman]

parthaa:
[test foreman]

parthaa:
[test foreman]

parthaa:
[test katello]

parthaa:
@tbrisker @ohadlevy @tstrachota @timogoebel  are you guys waiting for anything further on this PR..

app/controllers/api/v2/table_preferences_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+      param :name, String, :required => true
+      def show
+        if @table_preference.blank?
+          @table_preference = TablePreference.new(:user => @user, :name => params[:name])
+        end
+      end
+
+      def_param_group :table_preference do
+        param :table_preferences, Hash, :required => true do
+          param :name, String, :required => true, :desc => N_("name of the table")
```
tbrisker: nitpick: should be `Name...`  (also below)

app/controllers/api/v2/table_preferences_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+        process_response @table_preference.save
+      end
+
+      api :PUT, "/users/:user_id/table_preferences/:name", N_("Updates a table preference for a given table")
+      param_group :table_preference
+      def update
+        process_response @table_preference.update_attributes(:columns => params[:columns])
+      end
+
+      api :DELETE, "/users/:user_id/table_preferences/:name/", N_("Delete columns for a table")
```
tbrisker: Doesn't this delete all the preferences for that table?

app/registries/foreman/access_permissions.rb
```diff
@@ -720,4 +720,8 @@
     map.permission :revoke_personal_access_tokens,
                    :"api/v2/personal_access_tokens" => [:destroy]
   end
+
+  permission_set.security_block :table_preferences do |map|
+    map.permission :access_table_preferences, {:"api/v2/table_preferences" => [:index, :create, :destroy, :show, :update]}
```
tbrisker: I asked before, but do we need this permission at all? Shouldn't it be just an aspect of the user's preferences, similar to how any user can edit their dashboard layout?

tbrisker (changes requested):
A few comments inline, one thing that is missing in the controller is a check that the `user_id` is the current user's id (or is permitted to edit otherwise) - currently the implementation lets any user edit any other user.
I don't think we should allow, at least at this point, letting one user edit another's preferences, and I would prefer adding that if needed in a follow-up pr, to allow getting the basic functionality in first.

tstrachota:
:+1: for the current design. Thanks for all the work, @parthaa !

app/registries/foreman/access_permissions.rb
timogoebel: We don't have an "access" permission anywhere else. Why not go for `view`, `create`, `destroy`?

app/controllers/api/v2/table_preferences_controller.rb
parthaa: Nope just the intended preference. I had 2 calls before reset_all and destroy to mitigate this confusion. I was asked to change that and just keep destroy. We likely want to reset things only for a single table/resource name.

app/registries/foreman/access_permissions.rb
parthaa: I would recommend looking at https://github.com/theforeman/foreman/blob/develop/app/registries/foreman/access_permissions.rb#L643-L647
```
  permission_set.security_block :dashboard do |map|
    map.permission :access_dashboard, {:dashboard => [:index, :show, :save_positions, :reset_default, :create, :destroy],
                                      :"api/v2/dashboard" => [:index]
    }
  end
```

What I have is no different for table_preferences.

parthaa:
[test upgrade]

parthaa:
[test foreman]

parthaa:
[test upgrade]

app/registries/foreman/access_permissions.rb
timogoebel: Fair enough, I can see why you chose the access permission. However I still the crud permissions are safer for future development. Just treat this as a hint, I'm ok either way.

timogoebel:
@parthaa: I don't have time to review this fully right now. I just wanted to leave some hints and help with the process but can't do a full review right now. Sorry.

app/registries/foreman/access_permissions.rb
tbrisker: `access_dashboard` permission limits access to the dashboard since it contains a lot of information you might not want a user to see. Even there, if you are allowed to access it, there is no authorization check for the layout setting. 
In the case of table preferences, I don't see any reason to limit a user's access to the preferences themselves, just like we don't limit the ability to sort or search - it's just a UI preference.

timogoebel: True, however I would always grant the user permission to change it's own settings like we do it for personal access tokens. I would use the crud permissions for an admin then who is supposed to fix problems like "I don't see the xyz column anymore. Can you help me get it back?". Just my 2 cents.

tbrisker: I think we could add that as a second step if needed, possibly even without using permissions (just allow only admin and self). adding all the permissions checks atm is just complicating the initial implementation - right now we can't merge this since it will allow any user to edit any other user's preference.

ares:
Admins do not need permissions to do stuff. But I think the controller 
enforces User.current. I'm personally fine with that for now, I guess we 
only need the permission to pass the test that checks every action has one 
assigned.

For problems @timogoebel mentioned, we should have sudo feature. I think 
there's an RFE in redmine somewhere if someone has cycles. Implementation 
should be very easy, just store original user_id to session and change 
User.current. A permanent actionable notification when original user id is 
seen in session should.be displayed to reset the User.current. This would 
be a new action only available for administrators.

ares:
Tomer are you sure? Last time I checked, it was only loading preferences 
for User.current. It might have changed and reviewing on phone is hard, so 
my appologies if it doea not.

app/registries/foreman/access_permissions.rb
ares: Sorry I replied via email and it was not aligned to this conversation. Long story short, @tbrisker I think the permission is needed because of tests that verified every action has some permission that covers it. It forces to work only with users preferences thanks to https://github.com/theforeman/foreman/pull/5262/files#diff-6e3452c581d814d25d1c8ac0001adf4dR5 (same thing is used for personal tokens and ssh keys). I don't think we need to change it now. If we're afraid admins can't fix that, let's implement sudo, ideally out of this PR.

parthaa:
[test foreman]

parthaa:
@timogoebel @ares @ohadlevy  had a chat with @tbrisker and decided to remove the access_table_preferences permission. At this point it seems  unnecessary since only User.current is allowed to mess with this table for their own preferences (similar to dashboard). Hope that works for everyone. If permissions is indeed desired in the future we can always add them later.

timogoebel:
> Hope that works for everyone.

Works for me. :-)

app/models/user.rb
```diff
@@ -42,7 +42,7 @@ class User < ApplicationRecord
   has_many :widgets, :dependent => :destroy
   has_many :ssh_keys, :dependent => :destroy
   has_many :personal_access_tokens, :dependent => :destroy
-
+  has_many :table_preferences, :dependent => :destroy
```
timogoebel: How about adding `inverse_of` here?

tbrisker:
@parthaa looks like `test/unit/foreman/access_permissions_test.rb` needs to be modified to skip checking for permissions on the new endpoints. The other tests seem to be just transient failures.

parthaa:
@tbrisker updated.

tbrisker (approved):
Looking good, any concerns @ares or @ohadlevy ?

ares:
not from my side, I'm fine with the last change

app/controllers/api/v2/table_preferences_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+      include Foreman::Controller::UserAware
+      before_action :check_user
+      before_action :find_resource, :only => [:destroy, :update, :show]
+
+      api :GET, "/users/:user_id/table_preferences", N_("List of table preferences for a user")
+      def index
+        @table_preferences = @user.table_preferences
+      end
+
+      api :GET, "/users/:user_id/table_preferences/:name", N_("List of table preferences for a table")
```
ohadlevy: I find the description confusing, how about "Table preferences details" ? (but not a list of table for table)

app/models/table_preference.rb
```diff
@@ -0,0 +1,10 @@
+class TablePreference < ApplicationRecord
```
ohadlevy, line 1: do we care about auditing or search?

tbrisker: I don't think so, this is just UI preferences, not anything that has impact on the system. It's similar to the dashboard widget settings which we don't audit or search.

mmoll:
@ohadlevy all good now? :)

ohadlevy (approved):

ohadlevy:
Thanks!

parthaa:
Gist connected to Usage -> https://gist.github.com/parthaa/36ead0ff6962d6dac33412a1c38e94f7

PR #5264, opened 2018-02-20
Fixes #6504 - refactoring host tests regarding taxonomies

lizagilman:


lzap:
Making whitespace-only changes is considered a bad practice in bigger teams, this will bring lots of git conflicts. Only fix whitespace on lines you actually change, we tend to accept fixing whitespace in files containing some relevant changes as well. I suggest to drop all the changes except the actual refactoring.

lizagilman:
@lzap ofcourse
whitespsaces added unintended, there was some kind of sorcery blocking me from getting rid of them :blush:  but it is fixed now

test/models/host_test.rb
```diff
@@ -678,45 +678,42 @@ def teardown
...
-  end
+    test "update a host's location" do
+      original_location = @location
 
-  test "assign a host to an organization" do
-    host = Host.create :name => "host 1", :mac => "aabbecddeeff", :ip => "5.5.5.5", :hostgroup => hostgroups(:common), :managed => false
-    organization = Organization.create :name => "Hosting client 1"
+      @host.location_id = original_location.id
+      assert @host.save!
+      assert @host.location_id = original_location.id
```
tbrisker: I think this should be `assert_equal` - looks like this test was wrong a long time (it has an assignment instead of a comparison)

test/models/host_test.rb
```diff
@@ -678,45 +678,42 @@ def teardown
...
 
-    host.organization_id = organization.id
-    assert host.save!
-  end
+      new_location = Location.create :name => "Los Angeles"
+      @host.location_id = new_location.id
+      assert @host.save!
+      assert @host.location_id = new_location.id
```
tbrisker: Here as well

test/models/host_test.rb
```diff
@@ -678,45 +678,42 @@ def teardown
...
-    host.location_id = location.id
-    assert host.save!
-  end
-
-  test "update a host's location" do
-    host = Host.create :name => "host 1", :mac => "aabbccddeeff", :ip => "5.5.5.5", :hostgroup => hostgroups(:common), :managed => false
-    original_location = Location.create :name => "New York"
+  context 'host assigned to location and organization' do
+    setup do
+      @host = Host.create :name => "host 1", :mac => "aabbecddeeff", :ip => "5.5.5.5", :hostgroup => hostgroups(:common), :managed => false
```
tbrisker: Perhaps use FactoryBot to create the host? I don't think we care about any of these attributes, except possibly the `managed: false` one, and we might be getting some false positives - e.g. if the `hostgroups(:common)` has a location.

tbrisker (changes requested):
A couple of comment inline. looks like the refactoring brought attention to an incorrect test.

test/models/host_test.rb
lzap: Agree, also prefer `build` over `create` if you can afford it. I think this could be good fit, haven't seen all tests tho.

lzap:
*Much better* thanks :-) :wave:

lizagilman:
updated

lizagilman:
[test foreman]

lizagilman:
@tbrisker / @lzap ok to merge?

tbrisker (approved):

tbrisker:
Thanks @lizagilman !

PR #5265, opened 2018-02-21
Fixes #22642 - limit searching by host to explicit query

ares:


ohadlevy:
isn't name is one of the most common queries? I would by default prefer to limit most of the other queries vs name? also, is name really adding a lot of complexity to the query to being with?

ohadlevy:
oh - i take it back, i missread the pr, you are avoiding the searching on hosts name not hostgroup name :+1:

ohadlevy (approved):

tbrisker (changes requested):
@ares Makes sense, but there's a bunch of other fields that i think should also be only explicit - lines 74-82 for sure, possibly also 58, and also might be good to add in katello extensions - https://github.com/Katello/katello/blob/master/app/models/katello/concerns/hostgroup_extensions.rb#L17

ares:
@tbrisker I agree, fixed on lines you suggested too (except one which already had it), I'll send a PR to Katello too and link it to this one, good catch! could you please take another look?

app/models/hostgroup.rb
```diff
@@ -53,7 +53,7 @@ class Hostgroup < ApplicationRecord
 
   scoped_search :on => :name, :complete_value => :true
   scoped_search :relation => :group_parameters,    :on => :value, :on_key=> :name, :complete_value => true, :only_explicit => true, :rename => :params
-  scoped_search :relation => :hosts, :on => :name, :complete_value => :true, :rename => "host"
+  scoped_search :relation => :hosts, :on => :name, :complete_value => :true, :rename => "host", :only_explicit => true
   scoped_search :relation => :puppetclasses, :on => :name, :complete_value => true, :rename => :class, :only_explicit => true, :operators => ['= ', '~ ']
   scoped_search :relation => :environment, :on => :name, :complete_value => :true, :rename => :environment
```
tbrisker: What about this one? or is this intentional?

tbrisker (changes requested):
I think you missed one :smile_cat:

app/models/hostgroup.rb
ares: no, but I only added to lines you suggested :-) added here too, mind to take one more look?

tbrisker (approved):
Thanks @ares ! Will merge once test pass.

PR #5269, opened 2018-02-22
Fixes #22096 - Make Audits taxable

tbrisker:
This relies on https://github.com/collectiveidea/audited/pull/413 for taxonomies to be recorded properly on destroy.
Still missing tests as well.

### DIFF
--- a/app/models/taxonomy.rb
+++ b/app/models/taxonomy.rb
@@ -154,8 +154,8 @@ class Taxonomy < ApplicationRecord
     new
   end
 
-  # overwrite *_ids since need to check if ignored? - don't overwrite location_ids and organizations_ids since these aren't ignored
-  (TaxHost::HASH_KEYS - [:location_ids, :organizations_ids]).each do |key|
+  # overwrite *_ids since need to check if ignored? - don't overwrite location_ids and organization_ids since these aren't ignored
+  (TaxHost::HASH_KEYS - [:location_ids, :organization_ids]).each do |key|
     # def domain_ids
     #  if ignore?("Domain")
     #   Domain.pluck(:id)
--- a/test/models/concerns/audit_extensions_test.rb
+++ b/test/models/concerns/audit_extensions_test.rb
@@ -57,4 +57,93 @@ class AuditExtensionsTest < ActiveSupport::TestCase
     end
     assert Audit.last.audited_changes.to_s.bytesize > 66000
   end
+
+  context "with multiple taxonomies" do
+    def setup
+      @loc = taxonomies(:location1)
+      @org = taxonomies(:organization1)
+    end
+
+    test 'records on create' do
+      domain = FactoryBot.create(:domain, :with_auditing, :locations => [@loc], :organizations => [@org])
+      audit = domain.audits.last
+      assert_equal 'create', audit.action
+      assert_equal [@loc.id], audit.location_ids
+      assert_equal [@org.id], audit.organization_ids
+    end
+
+    test 'records on update' do
+      domain = FactoryBot.create(:domain, :with_auditing, :locations => [], :organizations => [])
+      audit = domain.audits.last
+      assert_equal 'create', audit.action
+      assert_equal [], audit.location_ids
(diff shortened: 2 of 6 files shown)


dLobatog (dismissed):
Setting to WoC because of tests, the feature works fine, thanks @tbrisker !

app/models/taxonomy.rb
```diff
@@ -154,8 +154,8 @@ def dup
     new
   end
 
-  # overwrite *_ids since need to check if ignored? - don't overwrite location_ids and organizations_ids since these aren't ignored
-  (TaxHost::HASH_KEYS - [:location_ids, :organizations_ids]).each do |key|
+  # overwrite *_ids since need to check if ignored? - don't overwrite location_ids and organization_ids since these aren't ignored
+  (TaxHost::HASH_KEYS - [:location_ids, :organization_ids]).each do |key|
```
tbrisker, line 158: This is an unrelated bug discovered due to the check for `respond_to?(organization_ids)`

test/models/concerns/audit_extensions_test.rb
```diff
@@ -57,4 +57,45 @@ def setup
...
+      domain.locations = [@loc]
+      domain.organizations = [@org]
+      domain.save!
+      audit = domain.audits.last
+      assert_equal 'update', audit.action
+      assert_equal [@loc.id], audit.location_ids
+      assert_equal [@org.id], audit.organization_ids
+    end
+
+    test 'records on destroy' do
```
tbrisker, line 92: This test should fail until a new audited version is released containing collectiveidea/audited#413

tbrisker:
leaving on WoC until new audited version is released

test/models/setting_test.rb
```diff
@@ -204,11 +204,12 @@ def test_value_is_updated_after_change_in_SETTINGS
     setting_name = "foo_#{rand(1000000)}"
     Setting.create!(:name => setting_name, :value => "bar", :default => "default", :description => "foo")
 
-    SETTINGS.stubs(:key?).with(setting_name.to_sym).returns(true)
-    SETTINGS.stubs(:[]).with(setting_name.to_sym).returns("no-bar")
+    SETTINGS[setting_name.to_sym] = "no-bar"
```
tbrisker, line 207: These stubs failed because creating the setting checked if taxonomies are enabled for the audit, which the stub didn't expect. However, `SETTINGS` is just a hash, so stubbing it is actually slower then just updating it.

tbrisker:
new audited version cause seeds test to fail, investigating.

mmoll:
[test upgrade]

mmoll:
[test foreman]

ares:
I did a quick test with https://github.com/theforeman/foreman/pull/5332 and it works reasonably well. It does not show the audit in the organization from which I'm removing organization from which can be considered a bug, @rplevka already mentioned that as well. But if that's hard to solve, I'd be fine keeping that for separate issue, as this is already a great improvement that we should deliver in 1.18.

EDIT:
forgot to mention, this would be needed too, since Taxonomix will already need audit DSL in model
```diff
diff --git a/config/initializers/audit.rb b/config/initializers/audit.rb
index 8f276e6f0..034ce7c63 100644
--- a/config/initializers/audit.rb
+++ b/config/initializers/audit.rb
@@ -2,11 +2,12 @@
 # we reopen here to add search functionality
 require 'audited'
 
-Audit = Audited::Audit
-Audit.send(:include, AuditExtensions)
-
 # Re-opened AuditorInstanceMethods to audit 1-0-* associations
 Auditor = Audited::Auditor::AuditedInstanceMethods
 Auditor.send(:include, AuditAssociations)
 
 ::ActiveRecord::Base.send :include, AuditAssociations::Auditor
+
+# Audit includes Taxonomix which already relies on DSL provided by audited gem
+Audit = Audited::Audit
+Audit.send(:include, AuditExtensions)
```

ares (approved):
this one gets in first, so no further changes are needed, thanks @tbrisker, mering

PR #5272, opened 2018-02-23
Fixes #22676 - domain checking should ignore scoping

jlsherrill:
By using domain_id, the true intent of the logic is held and saving the NIC doesn't require the domain to require the current scope.

### DIFF
--- a/app/models/nic/interface.rb
+++ b/app/models/nic/interface.rb
@@ -105,7 +105,7 @@ module Nic
       # no hostname was given or a domain was selected, since this is before validation we need to ignore
       # it and let the validations to produce an error
       return if name.empty?
-      if domain.nil? && name.include?('.') && changed_attributes['domain_id'].blank?
+      if domain_id.nil? && name.include?('.') && changed_attributes['domain_id'].blank?
         # try to assign the domain automatically based on our existing domains from the host FQDN
         self.domain = Domain.find_by(:name => name.partition('.')[2])
       elsif persisted? && changed_attributes['domain_id'].present?
--- a/test/models/nics/managed_test.rb
+++ b/test/models/nics/managed_test.rb
@@ -71,6 +71,19 @@ class ManagedTest < ActiveSupport::TestCase
     assert_equal new_domain, nic.domain
   end
 
+  test "#normalize_hostname does not update domain if domain does not match current taxonomies" do
+    domain = FactoryBot.create(:domain)
+    nic = setup_primary_nic_with_name(" Host.#{domain.name}", :domain => domain)
+    nic.save!
+    Location.current = taxonomies(:location2)
+    User.as('one') do
+      nic = Nic::Managed.find(nic.id) #load object to prevent cached association
+      nic.send(:normalize_name)
+      Location.current = nil
+      assert_equal domain.id, nic.domain_id
+    end
+  end
+
   test "#inheriting_mac respects interface mac" do
     h = FactoryBot.build_stubbed(:host, :managed)
     h.primary_interface.mac = '11:22:33:44:55:66'


jturel:
before fix:
```
[vagrant@centos7-bats ~]$ sudo subscription-manager register --org="Test_Organization" --force --activationkey="Test AK"
Couldn't find Domain with 'id'=1 [WHERE (domains.id IN (4))]
```                               
with the fix applied:
```                      
[vagrant@centos7-bats ~]$ sudo subscription-manager register --org="Test_Organization" --force --activationkey="Test AK"
The system has been registered with ID: 11b82a1a-63bb-46ba-b7a4-074f9df5f80b
```

ACK from me

app/models/nic/interface.rb
```diff
@@ -105,7 +105,7 @@ def normalize_name
       # no hostname was given or a domain was selected, since this is before validation we need to ignore
       # it and let the validations to produce an error
       return if name.empty?
-      if domain.nil? && name.include?('.') && changed_attributes['domain_id'].blank?
+      if domain_id.nil? && name.include?('.') && changed_attributes['domain_id'].blank?
         # try to assign the domain automatically based on our existing domains from the host FQDN
         self.domain = Domain.find_by(:name => name.partition('.')[2])
```
tbrisker, line 110: Souldn't this and the below `Domain.find` also use `.unscoped` to ensure the search isn't scoped?

tbrisker (dismissed):
Looks like the test is failing, also a question inline

app/models/nic/interface.rb
jlsherrill: We could do that, but is that better?  Isn't the presence of domain_id enough?  I guess the .unscoped is more explicit which is better.  Will update!

app/models/nic/interface.rb
```diff
@@ -105,7 +105,7 @@ def normalize_name
       # no hostname was given or a domain was selected, since this is before validation we need to ignore
       # it and let the validations to produce an error
       return if name.empty?
-      if domain.nil? && name.include?('.') && changed_attributes['domain_id'].blank?
+      if Domain.unscoped.find_by(id: domain_id).nil? && name.include?('.') && changed_attributes['domain_id'].blank?
```
tbrisker: I think you misunderstood my comment - checking for domain_id was fine, i was referring to the lookups on lines 110 and 113 that may not find properly when scoped.

jlsherrill: ahhh, gotcha, will update.

jlsherrill: I guess the controller should be handling if the user doesn't have visibility of the domain, so .unscoped should be fine.

app/models/nic/interface.rb
```diff
@@ -105,12 +105,12 @@ def normalize_name
       # no hostname was given or a domain was selected, since this is before validation we need to ignore
       # it and let the validations to produce an error
       return if name.empty?
-      if domain.nil? && name.include?('.') && changed_attributes['domain_id'].blank?
+      if Domain.unscoped.find_by(:id => domain_id).nil? && name.include?('.') && changed_attributes['domain_id'].blank?
```
tbrisker: I think this can be `if domain_id.nil?...`, since we are checking if the nic has a domain set so no need for the additional lookup. tbh this function could benefit from some refactoring, but not required in the scope of this pr

jlsherrill:
@tbrisker updated, all tests now :green_heart:

app/models/nic/interface.rb
```diff
@@ -105,12 +105,12 @@ def normalize_name
       # no hostname was given or a domain was selected, since this is before validation we need to ignore
       # it and let the validations to produce an error
       return if name.empty?
-      if domain.nil? && name.include?('.') && changed_attributes['domain_id'].blank?
+      if domain_id.nil? && name.include?('.') && changed_attributes['domain_id'].blank?
         # try to assign the domain automatically based on our existing domains from the host FQDN
-        self.domain = Domain.find_by(:name => name.partition('.')[2])
+        self.domain = Domain.unscoped.find_by(:name => name.partition('.')[2])
```
dLobatog: This and the line below with `unscoped` be able to set domains that are **not** visible to the user? E.g: I create an interface through the API with a domain name that I know I can't access. This line would find it and set it. Is this required to make `subman` work? I think the change in the  `if` condition should be enough.

dLobatog (changes requested):
@jlsherrill Sorry, I think the unscoped needs to be removed so that we can ensure users can't set interfaces in domains they don't know. Thanks for the PR!

app/models/nic/interface.rb
jlsherrill: wouldn't the controller handle that?

jlsherrill: oh right, here its based off name, duh ;)

jlsherrill: the one below i would think would be filtered by the controller, or is that not how it works?

dLobatog: The controller does not filter the parameters, I can send the `domain_id` of a domain I can't see (but still exists, maybe an admin can see it) as a parameter, and the domain will just pass it on. It's the model's responsibility to be unable to find the `domain_id`, then the controller would just translate that to 404.

jlsherrill: I would think so much would be broken as far as auth goes if that were true since update_attributes would be relied on,  but i'll take your word for it :)

jlsherrill:
[test katello]

ehelms:
[test foreman]

ehelms:
@dLobatog This is passing now if oyu can have another look.

dLobatog (approved):

dLobatog:
Thanks @jlsherrill @ehelms !

PR #5273, opened 2018-02-25
Fixes #22682 - Update patternfly-react

sharvit:
Travis is getting errors in all pr's because `patternfly-react` updated their markup.
We need to prefer shallow renderers when possible so it won't happen again (and we will get cleaner snapshots).

* update patternfly-react to version ^1.10.0
* migrate NotificationDropdown.test to a shallow renderer

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "lodash": "~4.15.0",
     "multiselect": "~0.9.12",
     "patternfly": "^3.31.2",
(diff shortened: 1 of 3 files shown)


sharvit:
What is wrong with my commit message? :confused:

ohadlevy (approved):

ohadlevy:
[test prprocessor]

PR #5274, opened 2018-02-25
Fixes #22685 - improve statistics page  loading time

amirfefer:
all ajax calls for each chart occurs in `StatisticsChartsList.js`, so with each and every store update, this entire component is re-rendered, and causes a performance issue.

 performance before change:
statistics-too-slow

after change:
statistics-ok

### DIFF
--- a/webpack/assets/javascripts/react_app/common/testHelpers.js
+++ b/webpack/assets/javascripts/react_app/common/testHelpers.js
@@ -21,3 +21,12 @@ export default {
     };
   },
 };
+
+// a helper method for invoking a class method (for unit tests)
+// obj = a class
+// func = a tested function
+// objThis = an object's this
+// arg = function args
+
+export const classFunctionUnitTest = (obj, func, objThis, args) =>
+  obj.prototype[func].apply(objThis, args);
--- a/webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
+++ b/webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
@@ -2,11 +2,10 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Panel } from 'react-bootstrap';
 import { Modal } from 'patternfly-react';
+import { isEqual } from 'lodash';
 import helpers from '../../common/helpers';
 import DonutChart from '../common/charts/DonutChart';
-import {
-  navigateToSearch,
-} from '../../../services/ChartService';
+import { navigateToSearch } from '../../../services/ChartService';
 import Loader from '../common/Loader';
 import MessageBox from '../common/MessageBox';
 
@@ -15,6 +14,12 @@ class ChartBox extends React.Component {
     super(props);
     this.state = { showModal: false };
     helpers.bindMethods(this, ['onClick', 'closeModal', 'openModal']);
+  }
+  shouldComponentUpdate(nextProps, nextState) {
+    return (
+      !isEqual(this.props.chart, nextProps.chart) ||
+      !isEqual(this.state, nextState)
+    );
   }
 
   onClick() {
@@ -37,13 +42,15 @@ class ChartBox extends React.Component {
     const Chart = components[type];
     const dataFiltered = chart.data && chart.data.filter(arr => arr[1] !== 0);
     const hasChartData = dataFiltered && dataFiltered.length > 0;
-    const tooltip = hasChartData ? {
-      onClick: this.onClick,
-      title: this.props.tip,
-      'data-toggle': 'tooltip',
-      'data-placement': 'top',
-      className: 'pointer',
-    } : {};
+    const tooltip = hasChartData
+      ? {
+        onClick: this.onClick,
+        title: this.props.tip,
+        'data-toggle': 'tooltip',
+        'data-placement': 'top',
+        className: 'pointer',
+      }
+      : {};
     const handleChartClick =
       chart.search && chart.search.match(/=$/)
         ? null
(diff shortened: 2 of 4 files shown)


ohadlevy:
I'm trying to understand the initial issue.. why does it re-render the entire page? as the props of each component do not change? I would assume the entire screen should not re-render? is this happening only after #5137? or also before?

amirfefer:
@ohadlevy 
it's not related to #5137 changes, which means it happened before as well.
apparently the `StatisticsChartsList` data pulling design (ajax calls iteration in one redux action)  causes this performance issue.
Why not each `ChartBox` will manage its specific chart API call? we gain a much better loading time with this approach.

@sharvit , any thoughts?

ohadlevy:
The main thing is that I wouldn't expect each chart to be a connected
component?

amirfefer:
I agree on that, better to have one connected container, though we are doing something wrong there with the data fetching. it takes more than double time to be rendered.

ohadlevy:
Can you please try finding out? I would rather understand what we did wrong.

amirfefer:
@ohadlevy , sure - thats what i'm doing.

sharvit:
Ideally, I would want to think as data-driven, build an object and render it.

I think the problem here is because you have an object that gets built over time and on each update you are rendering the whole list again.
(or at least, calcuate the current rendering state -> compare to the dom -> no need to update)

I am not sure there is a better solution than create a child connected components as you did.

amirfefer:
@sharvit - what about if we pass a dispatcher to each child component as a prop, and then each child can use it on its `componentDidMount` but still be an un-connected component?

sharvit:
@amirfefer i don't think it is going to help, i think the problem is at the `mapStateToProps` that resolving the whole list again and again.

amirfefer:
@ohadlevy 
@sharvit and I apparently  figured out that the reason for this performance issue is by the way we fetch the data and the state changes from it. in order to fix that, we can do: 
1. keep this behavior -each child is responsible for its chart state (fetching and update the store)
2. refactor the API calls, from multiple call to singular - though with this approach the  page will load at once

ohadlevy:
@amirfefer whats the status of this PR? thanks

amirfefer:
The major reason for this performance issue is c3js, which uses `componentWillReceiveProps`
According to react documentation: 
> ...Note that React will call this method even if the props have not changed...

To reduce 'componentWillReceiveProps' callings, we can use  `shouldComponentUpdate` in the upper level component.
without `shouldComponentUpdate` 
screenshot from 2018-03-19 01-53-31


with 
screenshot from 2018-03-19 01-45-03

webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
```diff
@@ -2,20 +2,32 @@ import React from 'react';
...
 class ChartBox extends React.Component {
   constructor(props) {
     super(props);
     this.state = { showModal: false };
     helpers.bindMethods(this, ['onClick', 'closeModal', 'openModal']);
   }
+  shouldComponentUpdate(nextProps, nextState) {
+    if (
+      nextProps.chart.data === undefined ||
+      nextProps.status === STATUS.ERROR
```
ohadlevy: shouldn't we render the error ?

webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
```diff
@@ -37,13 +49,15 @@ class ChartBox extends React.Component {
...
-      title: this.props.tip,
-      'data-toggle': 'tooltip',
-      'data-placement': 'top',
-      className: 'pointer',
-    } : {};
+    const tooltip = hasChartData
+      ? {
+        onClick: this.onClick,
+        title: this.props.tip,
+        'data-toggle': 'tooltip',
```
ohadlevy, line 49: maybe we should use the tooltip react component instead of the jquery one as well?

ohadlevy (commented):
some inline comments.

your initial performance seems better than the last one, maybe we are missing another condition to not render?

webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
amirfefer: after few attempts -  react-bootstrap `Panel` header doesn't work well with `<OverlayTrigger>`. plus it has a breaking changes new API in the later version. I think it better to handle this in a different PR, because it doesn't related to statistics page performance.

amirfefer:
@ohadlevy Thanks for your review.
The average values have a smaller gap of ~300ms. 
imho this gap occurs due to `map` scripting.

amirfefer:
@sharvit could you have a look?

sharvit (approved):
LGTM

ohadlevy:
[test foreman]

webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
ohadlevy: can you please add it as an issue that we won't forget about it?

webpack/assets/javascripts/react_app/components/statistics/ChartBox.test.js
```diff
@@ -48,4 +48,16 @@ describe('ChartBox', () => {
...
+    const box = shallow(<ChartBox
+      id='4'
+      type="donut"
+      chart={{ id: '4', data: undefined }}
+      status="PENDING"
+    />);
+
+    box.setProps({ status: 'PENDING' });
+    expect(spy).toHaveBeenCalled();
+  });
```
ohadlevy: does it make sense to add another test that check the shouldComponentUpdate logic as well?

amirfefer:
@ohadlevy  Could you have another look? I added few tests for `shouldComponentUpdate` function.

webpack/assets/javascripts/react_app/common/testHelpers.js
```diff
@@ -21,3 +21,6 @@ export default {
     };
   },
 };
+
+export const classFunctionUnitTest = (obj, func, objThis, args) =>
```
ohadlevy: please add a short description - I have no idea what this means :)

ohadlevy:
everything looks good, please after you add the description, rebase the commit so tests will be green. thanks!

amirfefer:
[test katello]

amirfefer:
[test foreman]

amirfefer:
@ohadlevy could you have another look ?

ohadlevy:
[test foreman]

webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
ohadlevy: @amirfefer please don't forget to add this issue.

ohadlevy (approved):

PR #5275, opened 2018-02-26
Refs #18675 - Optionally load prometheus exporter

ekohl:
In case the bundler group telemetry is not installed, we should handle that.

### DIFF
--- a/config.ru
+++ b/config.ru
@@ -1,16 +1,8 @@
 # This file is used by Rack-based servers to start the application.
-
 require 'rack'
-require './lib/foreman/telemetry_rack'
-require 'prometheus/middleware/exporter'
 
 # load Rails environment
 require ::File.expand_path('../config/environment',  __FILE__)
-
-use Foreman::TelemetryRack
-if defined?(Prometheus::Middleware::Exporter) && SETTINGS[:telemetry].try(:fetch, :prometheus).try(:fetch, :enabled)
-  use Prometheus::Middleware::Exporter
-end
 
 # apply a prefix to the application, if one is defined
 # e.g. http://some.server.com/prefix where '/prefix' is defined by env variable
--- a/config/application.rb
+++ b/config/application.rb
@@ -61,6 +61,7 @@ require_dependency File.expand_path('../../lib/foreman/http_proxy', __FILE__)
 require_dependency File.expand_path('../../lib/middleware/catch_json_parse_errors', __FILE__)
 require_dependency File.expand_path('../../lib/middleware/tagged_logging', __FILE__)
 require_dependency File.expand_path('../../lib/middleware/session_safe_logging', __FILE__)
+require_dependency File.expand_path('../../lib/middleware/telemetry', __FILE__)
 
 if SETTINGS[:support_jsonp]
   if File.exist?(File.expand_path('../../Gemfile.in', __FILE__))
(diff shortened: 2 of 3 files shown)


config.ru
```diff
@@ -2,7 +2,10 @@
 
 require 'rack'
 require './lib/foreman/telemetry_rack'
```
mmoll: There are still things in here. With this patch I get
```
No such file to load -- prometheus/client.rb (LoadError)
  /usr/share/foreman/vendor/ruby/2.3.0/gems/polyglot-0.3.5/lib/polyglot.rb:65:in `require&apos;
  /usr/share/foreman/vendor/ruby/2.3.0/gems/polyglot-0.3.5/lib/polyglot.rb:65:in `require&apos;
  /usr/share/foreman/vendor/ruby/2.3.0/gems/activesupport-5.1.4/lib/active_support/dependencies.rb:292:in `block in require&apos;
  /usr/share/foreman/vendor/ruby/2.3.0/gems/activesupport-5.1.4/lib/active_support/dependencies.rb:258:in `load_dependency&apos;
  /usr/share/foreman/vendor/ruby/2.3.0/gems/activesupport-5.1.4/lib/active_support/dependencies.rb:292:in `require&apos;
  /usr/share/foreman/lib/foreman/telemetry_sinks/prometheus_sink.rb:1:in `<top (required)>&apos;
[...]
```

ekohl: Odd, I was able to start the dev server.

lzap:
How about cleaning that file to:

```ruby
  # This file is used by Rack-based servers to start the application.
  require 'rack'
  
  # load Rails environment
  require ::File.expand_path('../config/environment',  __FILE__)
  
  require './lib/foreman/telemetry_rack'
  use Foreman::TelemetryRack
  
  begin
    require 'prometheus/middleware/exporter'
    if SETTINGS[:telemetry].try(:fetch, :prometheus).try(:fetch, :enabled)
      use Prometheus::Middleware::Exporter
    end
  rescue LoadError
    # bundler group 'telemetry' was disabled
  end
```

This works for me.

ekohl:
Looking at the rest of the code i wonder why this is done in ```config/config.ru``` rather than ```config/application.rb``` where the other middleware is loaded. It also makes more sense to move ```./lib/foreman/telemetry_rack.rb``` to ```lib/middleware/telemetry.rb``` but I'm not that familiar with rails.

config.ru
ekohl: I can reproduce this if I add a ```:telemetry``` group to the ```settings.yaml```

ekohl:
Updated to follow the pattern of other middleware loading. I tested with

* telemetry disabled + group not installed
* telemetry enabled + group not installed (still gives an ugly stack trace)
* telemetry enabled + group installed (but without an actual telemetry setup)

@lzap could you test if this still gives proper telemetry?

lzap (approved):
Looks great, good catch. I am also Rails noob so... Learning every day. Testing now.

lzap:
* [x] telemetry disabled + group not installed
* [x] telemetry enabled + group not installed (still gives an ugly stack trace)
* [x] telemetry enabled + group installed (but without an actual telemetry setup)

When green it's merge, thanks!

ekohl:
:green_heart: now

lzap:
Thanks.

PR #5277, opened 2018-02-27
Fixes #22711 - Enable ES6 support for uglifier

mmoll:


bundler.d/assets.rb
```diff
@@ -4,6 +4,6 @@
 gem 'gridster-rails', '~> 0.5'
 gem 'gettext_i18n_rails_js', '~> 1.0'
 gem 'execjs', '>= 1.4.0', '< 3.0'
-gem 'uglifier', '>= 1.0.3'
+gem 'uglifier', '>= 4.0.0'
```
ekohl, line 7: Our 5.1 SCL contains 3.2.0 so that'd need a bump.

mmoll: https://github.com/theforeman/tfm-ror51-packaging/pull/44

mmoll:
hrm, while it seems to progress further, the build stills fails at the second webpack... no idea what's going on...

mmoll:
I do think now, this does not fix the issue, as the problem is in the uglifier that's used by webpack and unrelated to the Rails part of this.

PR #5278, opened 2018-02-28
Refs #17316 - Ensure the proxy exists before assigning tftp

dLobatog:
If the subnet.tftp proxy didn't exist, the migration fails with
> undefined method `has_feature?' for nil:NilClass'
so we rather make sure it exists before trying to set the template
proxy to nil.

cc @lzap

### DIFF
--- a/db/migrate/20180216094550_add_template_to_subnets.rb
+++ b/db/migrate/20180216094550_add_template_to_subnets.rb
@@ -6,6 +6,7 @@ class AddTemplateToSubnets < ActiveRecord::Migration[5.1]
       dir.up do
         Subnet.unscoped.find_each do |subnet|
           proxy = subnet.tftp
+          next if proxy.blank?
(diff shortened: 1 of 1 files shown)


timogoebel (approved):
Fixes the issue for me and makes sense.

Ready to merge when tests pass. :-)

timogoebel:
[test foreman]

lzap (approved):
Uh. Hah. Hmmm. Approved.

iNecas:
Test failures unrelated. And since we unfortunately don't write tests for migrations (and the migrations went fine in the tests), I'm just merging this before anyone notices :)

lzap:
Yeah, we do test migrations but with empty data, that was the problem in this case. Thanks.

ares:
cc @rplevka, this is what we've hit with migrations, should be fixed now

PR #5279, opened 2018-02-28
Fixes #22713 - fix donut chart re-rendering

amirfefer:
When some data is deleted from a donut chart, it doesn't unload before , which means it remains the same.
In order to make it dynamic, just add `dynamic` prop to it.

statistics - edited

In addition, with this change, it possible to pass the chart data as is, no `mutable` function is needed,
because I removed the redundant `names` prop. `c3js` documentation mentions  that the `names` prop uses for mutating the existing names.

### DIFF
--- a/webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
+++ b/webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
@@ -1,7 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Panel } from 'react-bootstrap';
-import Immutable from 'seamless-immutable';
 import { Modal } from 'patternfly-react';
 import helpers from '../../common/helpers';
 import DonutChart from '../common/charts/DonutChart';
@@ -50,7 +49,7 @@ class ChartBox extends React.Component {
         ? null
         : navigateToSearch.bind(null, chart.search);
     const chartProps = {
-      data: chart.data ? Immutable.asMutable(chart.data) : null,
+      data: chart.data ? chart.data : undefined,
       key: `${this.props.chart.id}-chart`,
       onclick: handleChartClick,
     };
--- a/webpack/assets/javascripts/services/ChartService.js
+++ b/webpack/assets/javascripts/services/ChartService.js
@@ -6,59 +6,41 @@ const sizeConfig = {
   large: donutLargeChartConfig,
 };
 
-function getChartConfig({
-  data, config, onclick, id = uuidV1(),
-}) {
-  if (!data) {
-    return {};
+const doDataExist = (data) => {
+  if (!data || data.length === 0) {
+    return false;
   }
+  return data.reduce((curr, next) => {
+    const value = next[1];
+
+    return value !== 0 ? true : curr;
+  }, false);
+};
+const getColors = data =>
+  data.reduce((curr, next) => {
+    const key = next[0];
+    const color = next[2];
+
+    return color ? { ...curr, [key]: color } : curr;
+  }, {});
+
+const getChartConfig = ({
+  data, config, onclick, id = uuidV1(),
+}) => {
   const chartConfigForType = sizeConfig[config];
-  const nonEmptyData = data.filter((d) => {
-    const amount = d[1];
-
-    return amount !== 0;
-  });
-
-  const chartData = nonEmptyData.reduce(
-    (curr, next) => {
-      const key = next[0];
-      const color = next[2];
-
-      const names = {
-        ...curr.names,
-        [key]: key,
-      };
-
-      const retVal = {
-        ...curr,
-        names,
-      };
-
-      if (color) {
-        return Object.assign({}, retVal, {
-          colors: {
-            ...(retVal.colors || {}),
-            [key]: color,
-          },
-        });
-      }
-
-      return retVal;
-    },
-    {
-      ...chartConfigForType.data,
-      columns: nonEmptyData,
-      names: {},
-      onclick,
-    },
-  );
+  const colors = getColors(data);
+  const colorsSize = Object.keys(colors).length;
 
   return {
     ...chartConfigForType,
-    data: chartData,
     id,
+    data: {
+      columns: doDataExist(data) ? data : [],
+      onclick,
+      ...(colorsSize > 0 ? { colors } : {}),
+    },
   };
-}
+};
 
 export const getDonutChartConfig = ({
   data, config, onclick, id = uuidV1(),
@@ -69,7 +51,6 @@ export const getDonutChartConfig = ({
     onclick,
     id,
   });
-
 
 export const navigateToSearch = (url, data) => {
   let val = data.id;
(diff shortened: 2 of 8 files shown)


amirfefer:
@dLobatog - Could you check if this fix the bug you found?

dLobatog (changes requested):
Thanks @amirfefer , this works, but I still wonder why do we remove the non empty data from the request? I did this just removing the empty data and it works fine (I think it's simpler, let me know what you think)


```
+++ b/webpack/assets/javascripts/services/ChartService.js
@@ -13,13 +13,7 @@ function getChartConfig({
     return {};
   }
   const chartConfigForType = sizeConfig[config];
-  const nonEmptyData = data.filter((d) => {
-    const amount = d[1];
-
-    return amount !== 0;
-  });
-
-  const chartData = nonEmptyData.reduce(
+  const chartData = data.reduce(
     (curr, next) => {
       const key = next[0];
       const color = next[2];
@@ -47,7 +41,7 @@ function getChartConfig({
     },
     {
       ...chartConfigForType.data,
-      columns: nonEmptyData,
+      columns: data,
       names: {},
       onclick,
     },
```

webpack/assets/javascripts/react_app/components/statistics/ChartBox.js
```diff
@@ -50,7 +49,7 @@ class ChartBox extends React.Component {
         ? null
         : navigateToSearch.bind(null, chart.search);
     const chartProps = {
-      data: chart.data ? Immutable.asMutable(chart.data) : null,
+      data: chart.data ? chart.data : undefined,
```
dLobatog, line 52: Oh this is nice :smile:

dLobatog:
@amirfefer I changed the call from https://github.com/theforeman/foreman_remote_execution/pull/130
to:

```jsx
<DonutChart data={jobInvocations} dynamic={true} />
```

but it fails with ` Cannot read property 'targets' of null`. If I send the jobInvocations as a mutable object, 

```jsx
<DonutChart data={Immutable.asMutable(jobInvocations)} dynamic={true} />
```

it works, but the last color isn't sent properly.

screenshot

amirfefer:
@dLobatog - I assume that empty data is filtered because of this situation:
zero-value

amirfefer:
another reason to filter:
no-data
occurs when you pass data like: 
```javascript
[['data1', 0],['data2', 0]]
```

ohadlevy:
@amirfefer can you please add a test? thanks.

amirfefer:
@dLobatog, I refactored `ChartService` a bit,  for readability and performance.
 could you try this?
```javascript
<DonutChart data={jobInvocations}  />
```
or
```javascript
<DonutChart data={Immutable.asMutable(jobInvocations)}  />
```
I've renamed the `dynamic` prop to `unloadData`, and it should be use only if the data **categories** (i.e `success`, `pending`...) is  dynamic (data could change after first render).

webpack/assets/javascripts/services/ChartService.js
```diff
@@ -1,76 +1,57 @@
...
   regular: donutChartConfig,
   large: donutLargeChartConfig,
 };
 
-function getChartConfig({
-  data, config, onclick, id = uuidV1(),
-}) {
-  if (!data) {
-    return {};
+const isDataExist = (data) => {
```
dLobatog: `isDataExist` -> `doDataExist`?

webpack/assets/javascripts/services/ChartService.js
```diff
@@ -1,76 +1,57 @@
...
+  if (!data || data.length === 0) {
+    return false;
```
dLobatog: This method is the whole reason I don't need unloadData in https://github.com/theforeman/foreman_remote_execution/pull/130 - `nonEmptyData` used to remove data, so that

`[['sucess', 3], ['failed', 2],['pending', 0], ['canceled', 0]]` turned into `[['sucess', 3], ['failed', 2]]`. 
This is what caused the problem when rendering and `unloadData=false`. However, this function now does not remove anything. So `[['sucess', 3], ['failed', 2],['pending', 0], ['canceled', 0]]` is sent to the DonutChart like that, and it renders fine. I'm not even sure what was the purpose of `nonEmptyData`, so I think it's probably ok to keep it.

dLobatog (changes requested):
Thanks @amirfefer !

Using this, it works:

```jsx
 <DonutChart data={Immutable.asMutable(jobInvocations)}  />
```

Using it does not even render, is this expected?

```jsx
 <DonutChart data={jobInvocations}  />
```

Notice I do **NOT** even need the `unloadData` prop. The reason inline in the function `isDataExist`.

Can you add a couple of tests to the methods in `ChartService` so we know what they're supposed to do?

amirfefer:
@dLobatog 
```javascript
<DonutChart data={jobInvocations}  />
```
Should work now, 
could you please share the console log if it doesn't?
Thanks!

amirfefer:
@ohadlevy , @dLobatog : I've added tests to `ChartService`.

dLobatog (changes requested):
@amirfefer Unfortunately without using `data={Immutable.asMutable(jobInvocations)}`, the following error shows up:

```
Uncaught TypeError: Cannot assign to read only property 'Success' of object '#<Object>'
    at c3.js:3992
    at Array.forEach (<anonymous>)
    at Chart.c3_chart_fn.load (c3.js:3991)
    at C3Chart.loadNewData (react-c3js.js:74)
    at C3Chart.updateChart (react-c3js.js:92)
    at C3Chart.componentDidMount (react-c3js.js:44)
    at commitLifeCycles (react-dom.development.js:8770)
    at commitAllLifeCycles (react-dom.development.js:9946)
    at HTMLUnknownElement.callCallback (react-dom.development.js:542)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:581)
react-dom.development.js:9747 The above error occurred in the <C3Chart> component:
    in C3Chart
    in Unknown (created by lifecycle(Component))
    in lifecycle(Component) (created by mapProps(lifecycle(Component)))
    in mapProps(lifecycle(Component)) (created by DonutChart)
    in DonutChart (created by DonutChart)
    in DonutChart (created by JobInvocationContainer)
    in div (created by JobInvocationContainer)
    in JobInvocationContainer (created by Connect(JobInvocationContainer))
    in Connect(JobInvocationContainer)
    in Provider

Consider adding an error boundary to your tree to customize error handling behavior.
Visit https://fb.me/react-error-boundaries to learn more about error boundaries.
```

Aside from that :+1: , maybe you want to add some test checking the chart is re-rendered twice, once with let's say 4 data items, the next time with just 2 as that's a real use case.

Thanks @amirfefer !

amirfefer:
@dLobatog, Thanks again :+1: 
I figured it out that In your case, the data should be mutable, because it uses specific given colors.
so keep this:
```javascript
 <DonutChart data={Immutable.asMutable(jobInvocations)}  />
```
About your suggested test, I think it would be redundant:
1. when a react component props have been changed, it re-renders - that is a react pattern.
2. this component wraps pf component which wraps c3 library - so charts behaviors already tests there. 
3. the issue here was that the data prop was broken, the component did re-render but with bad props.
4. I already add some tests for the `ChartServie.js` which creates the props.

amirfefer:
@sharvit 
any thoughts?

sharvit (approved):
LGTM!
Good job refactoring the `ChartService.js` :+1:

ohadlevy:
@dLobatog OK to merge?

dLobatog (approved):

dLobatog:
Yup, works great. Thanks @amirfefer !

PR #5280, opened 2018-02-28
Fixes #22721 - Rename setting to trusted_hosts

dLobatog:
#The setting trusted_puppetmaster_hosts is used to allow hosts to submit
stuff to the facts/reports API endpoint.
The name made sense a long time ago, but these days other plugins and
any user can submit stuff to this API to generate
their own reports, send facts from ohai, etc..

It should be renamed to 'trusted_hosts' - for example in the foreman_ansible docs, I'm telling people to modify a setting `trusted_puppetmaster_hosts`, which doesn't make much sense in that context...

### DIFF
--- a/app/controllers/concerns/foreman/controller/smart_proxy_auth.rb
+++ b/app/controllers/concerns/foreman/controller/smart_proxy_auth.rb
@@ -104,14 +104,14 @@ module Foreman::Controller::SmartProxyAuth
     return false unless request_hosts
 
     hosts = Hash[proxies.map { |p| [URI.parse(p.url).host, p] }]
-    allowed_hosts = hosts.keys.push(*Setting[:trusted_puppetmaster_hosts])
+    allowed_hosts = hosts.keys.push(*Setting[:trusted_hosts])
     logger.debug { ("Verifying request from #{request_hosts.inspect} against #{allowed_hosts.inspect}") }
 
     if (host = detect_matching_host(allowed_hosts, request_hosts))
       @detected_proxy = hosts[host] if host
       true
     else
-      logger.warn "No smart proxy server found on #{request_hosts.inspect} and is not in trusted_puppetmaster_hosts"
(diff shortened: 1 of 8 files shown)


timogoebel:
@dLobatog: Is this missing a migration? :-)

dLobatog (commented):
Indeed, @timogoebel , updated with the migration

sean797:
I’m not sure I like ‘trusted_hosts’, IMO that sounds like it should be SSL related.

I haven’t got a good alternative though, maybe ‘external_config_nodes’? I don’t know :p

Also, this changes the API, I’m not sure if we want to consider depreciating the old name as well.

dLobatog (commented):
@sean797 I've updated this with the deprecation. Just a simple search on github for `trusted_puppetmaster_hosts` gave me enough justification for it, thanks for pointing it out. Regarding the name, I can't think of anything more descriptive right now...

lzap:
We need to keep it generic, it's not only ENCs which use this feature. Send your ideas, I am fine with trusted_hosts tho :-)

lzap (approved):
Looks good thanks.

dLobatog:
@lzap how about merging? :smile:

sean797 (approved):
Ultimately I'm also fine with `trusted_hosts` :+1: 

thanks @dLobatog

ekohl:
Maybe a bit of bikeshedding, but trusted to do what? Fully trusted? The help implies it's fact/report/enc so perhaps the setting becomes a bit ambiguous now. Not sure what would be a better name though.

dLobatog:
@ekohl Trusted to access anything that it says in the description, I don't think the name of the setting was correct before as it was certainly not only puppetmasters.

lzap (approved):
Tests are passing locally.

ares:
@lzap thanks for merging, please don't forget to set the release in redmine afterwards. I set it to 1.18

PR #5282, opened 2018-03-01
[WIP] fixes #22711 - upgrades uglifyjs-webpack plugin

ohadlevy:


ekohl (changes requested):
This doesn't solve the problem for me. Still the same error:
```
ERROR in vendor-316e85d529f8ed1e1aed.js from UglifyJs
Unexpected token: name (format_format) [vendor-316e85d529f8ed1e1aed.js:37923,4]
```

Maybe uglifyjs itself needs to be pinned to an older version?

sharvit:
@ohadlevy @ekohl see https://github.com/theforeman/foreman/pull/5283 i think it should work

PR #5283, opened 2018-03-01
Fixes #22711 - upgrades uglifyjs-webpack plugin

sharvit:


package.json
```diff
@@ -47,6 +47,7 @@
     "style-loader": "^0.13.1",
     "stylelint": "^7.13.0",
     "stylelint-config-standard": "^16.0.0",
+    "uglifyjs-webpack-plugin": "^1.2.2",
```
mmoll: AFAICT this should be <1.2.0. At least 1.2.1(?) has weback 4.x as peer dependency.

ekohl: Webpack pulls it in by default so if it's < 1.2.0 then it's redundant.

sharvit: The peer dependency of `uglifyjs-webpack-plugin@1.2.2` are `"webpack": "^2.0.0 || ^3.0.0 || ^4.0.0"`
So i think it should be fine.

`webpack@3` pulling `"uglifyjs-webpack-plugin": "^0.4.6"` wich is the version with the problem.

Do you think it is somehow related?
https://twitter.com/wSokra/status/967852475918274561

ekohl: Without adding this line it also works for me so I'd prefer to not include this. For me it pulls in 1.1.6 without it and that works.

sharvit: @ekohl i'm not sure I get it, can you explain please?

ekohl: @sharvit I tend to prefer as little specific pins as possible. If this isn't needed, I'd prefer not to add it.

config/webpack.config.js
```diff
@@ -132,8 +133,8 @@ var config = {
 if (production) {
   config.plugins.push(
     new webpack.NoEmitOnErrorsPlugin(),
-    new webpack.optimize.UglifyJsPlugin({
-      compressor: { warnings: false },
+    new UglifyJSPlugin({
+      // compressor: { warnings: false },
```
dLobatog: Why commenting this? It's now 
```
 compress: {
    warnings: false
}
```

dLobatog (dismissed):
@sharvit Thanks, I just removed the WIP from the title, afaik just fixing the pinning and the compressor is what's missing from this PR?

sharvit:
@ekohl @dLobatog updated based on your comments.

ohadlevy (approved):
Looks good to me, pending tests

ekohl (approved):
Build failure unrelated: https://github.com/theforeman/foreman-infra/pull/517

mmoll (approved):
fixes the problem locally for me :+1:

mmoll:
merged, thanks @sharvit!

package.json
ekohl: Looking back I was wrong. We explicitly require it our config so we should depend on it. It also looks like it needs at least a certain version (0.4.6 is being pulled in on slaves and doesn't work).

mmoll: GH-5289

PR #5285, opened 2018-03-02
Fixes #23825 - Add exporting to Rakefile.dist

ekohl:
The benefit is that you no longer need the ful bundle stack installed.
We also output the generated filename for use in scripts.

### DIFF
--- a/Rakefile.dist
+++ b/Rakefile.dist
@@ -1,4 +1,5 @@
 require 'rake/clean'
+load File.join(__dir__, 'lib', 'tasks', 'pkg.rake')
 
 BUILDDIR = File.expand_path(ENV['BUILDDIR'] || '_build')
 PREFIX = ENV['PREFIX'] || '/usr/local'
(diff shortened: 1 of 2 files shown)


Rakefile.dist
```diff
@@ -1,4 +1,5 @@
 require 'rake/clean'
+load File.expand_path('../lib/tasks/pkg.rake', __FILE__)
```
iNecas: I wonder if `require_relative` would work here…

iNecas:
Not sure what happened, but all the jobs in the CI failed pretty badly.

ehelms:
@ekohl Can you please swing back to this and drop the WIP?

Rakefile.dist
ekohl: That sadly doesn't work. I suspect because it doesn't have a `.rb` extension but a `.rake`. My ruby knowledge is lacking.

ekohl:
Updated

ehelms (approved):

ehelms:
The Katello test failure is unrelated. Thanks @ekohl

ares:
the redmine issue is set to 1.18, I guess this should have the 1.18 milestone and be cherry-picked?

ekohl:
Looking at cherry picks were still on my list and applicability to other repos was on my list. Ideally we can build all our packages without the need for full bundles to be present.

PR #5286, opened 2018-03-02
Fixes #22753 - add apipie validator for multiple allowed types

tstrachota:
Details in the redmine issue.

### DIFF
--- a/config/initializers/apipie.rb
+++ b/config/initializers/apipie.rb
@@ -113,3 +113,35 @@ class IdentifierDottableValidator < Apipie::Validator::BaseValidator
         "dot(.), space, underscore(_), hypen(-) with no leading or trailing space."
   end
 end
+
+# Allows to enumerate multiple types that a parameter accepts.
+class AnyTypeValidator < Apipie::Validator::BaseValidator
+  def initialize(param_description, argument, options = {})
+    super(param_description)
+    @allowed_types = options[:of] || []
+  end
+
+  def validate(value)
+    # The validator has rather informative value, skip the real validation
+    true
+  end
+
+  def self.build(param_description, argument, options, block)
+    if argument == :any_type
+      self.new(param_description, argument, options)
+    end
+  end
+
+  def description
+    if @allowed_types.empty?
+      'Can be any type'
+    else
+      types = @allowed_types.map { |type| "<code>#{type}</code>" }.join(', ')
+      'Must be one of types: %s' % types
+    end
+  end
+
+  def expected_type
+    :any_type
+  end
+end
--- a/app/controllers/api/v2/hosts_controller.rb
+++ b/app/controllers/api/v2/hosts_controller.rb
@@ -37,7 +37,7 @@ module Api
       param :location_id, String, :desc => N_("ID of location")
       param :organization_id, String, :desc => N_("ID of organization")
       param :environment_id, String, :desc => N_("ID of environment")
-      param :include, Array, :in => ['parameters', 'all_parameters'], :desc => N_("Array of extra information types to include")
+      param :include, ['parameters', 'all_parameters'], :desc => N_("Array of extra information types to include")
       param_group :search_and_pagination, ::Api::V2::BaseController
       add_scoped_search_description_for(Host)
 
--- a/app/controllers/api/v2/override_values_controller.rb
+++ b/app/controllers/api/v2/override_values_controller.rb
@@ -37,7 +37,7 @@ module Api
       def_param_group :override_value do
         param :override_value, Hash, :required => true, :action_aware => true do
           param :match, String, :required => true, :desc => N_("Override match")
-          param :value, LookupKey::KEY_TYPES, :required => false, :desc => N_("Override value, required if omit is false")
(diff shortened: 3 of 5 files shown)


config/initializers/apipie.rb
```diff
@@ -113,3 +113,30 @@ def description
...
+
+  def self.build(param_description, argument, options, block)
+    if argument == :any
+      self.new(param_description, argument, options)
+    end
+  end
+
+  def description
+    types = @allowed_types.map { |type| "<code>#{type}</code>" }.join(', ')
+    "Must be one of types: %s" % types.join(', ')
```
ekohl: Should this be translatable?

config/initializers/apipie.rb
```diff
@@ -113,3 +113,30 @@ def description
         "dot(.), space, underscore(_), hypen(-) with no leading or trailing space."
   end
 end
+
+class AnyOfValidator < Apipie::Validator::BaseValidator
```
iNecas: I'm not sure I understand the difference between `AnyOfValidator` and `EnumValidator` from Apipie. Probably worth documenting the difference as a comment here.

config/initializers/apipie.rb
tstrachota: I'd say yes, but at the moment none of the validator descriptions in foreman and apipie-rails is translated. And I found out that in one case hammer parses the description, so translations would break it. All in all translating this would require more refactoring than it seems and is imho low prio, so I'd deal with it in a separate issue.

config/initializers/apipie.rb
tstrachota: I changed the name to `AnyOfTypesValidator` and added a comment. The difference is that `Enum` validates param values, while this enumerates supported types. See the redmine issue for more details.

config/initializers/apipie.rb
```diff
@@ -113,3 +113,35 @@ def description
...
+  def description
+    if @allowed_types.empty?
+      'Can be any type'
+    else
+      types = @allowed_types.map { |type| "<code>#{type}</code>" }.join(', ')
+      'Must be one of types: %s' % types.join(', ')
+    end
+  end
+
+  def expected_type
```
iNecas: Is this method used anywhere?

iNecas:
Makes more sense now. One last comment inline

config/initializers/apipie.rb
ekohl: I do think that deserves a comment in the code.

config/initializers/apipie.rb
tstrachota: Apipie exposes this in json doc exports and hammer uses it together with a validator description to determine option type. I don't think it's exposed in apipie's web UI though.

E.g. when enum is used:
```
[1] pry(#<HammerCLI::Apipie::OptionBuilder>)> param.validator
=> "Must be one of: string, boolean, integer, real, array, hash, yaml, json." 
[2] pry(#<HammerCLI::Apipie::OptionBuilder>)> param.expected_type.to_s
=> "string" 
```

and when one of types validator is used:
```
[1] pry(#<HammerCLI::Apipie::OptionBuilder>)> param.validator
=> "Must be one of types: string, boolean, integer, real, array, hash, yaml, json." 
[2] pry(#<HammerCLI::Apipie::OptionBuilder>)> param.expected_type.to_s
=> "any_of_enumerated" 
```

iNecas: Ah, ok. 

My last comment/question is about consistency: we have `AnyOfTypesValidator`, `:any_type` and `:any_of_enumerated` in the code, naming the same thing at different places. I wonder if unifying on `AnyTypeValidator` + `:any_type` here and on other places wouldn't help the consistency.

tstrachota: Makes sense, updated.

tstrachota:
Updated

config/initializers/apipie.rb
```diff
@@ -113,3 +113,35 @@ def description
...
+    if argument == :any_type
+      self.new(param_description, argument, options)
+    end
+  end
+
+  def description
+    if @allowed_types.empty?
+      'Can be any type'
+    else
+      types = @allowed_types.map { |type| "<code>#{type}</code>" }.join(', ')
```
iNecas, line 139: Some test failures seems related (from http://ci.theforeman.org/job/test_develop_pr_core/16486/database=postgresql,ruby=2.4,slave=fast/console):

```

Error:
ComputeResourceJSIntegrationTest#test_0003_compute resource password isn't deleted when testing connection:
NoMethodError: undefined method `join' for #<String:0x000000002357aaa8>
    config/initializers/apipie.rb:140:in `description'
    lib/middleware/catch_json_parse_errors.rb:8:in `call'
    lib/middleware/session_safe_logging.rb:17:in `call'
    lib/middleware/tagged_logging.rb:18:in `call'
```

tstrachota: Oops, updated, I'm sorry.

iNecas (approved):

iNecas:
Thanks @tstrachota

PR #5288, opened 2018-03-03
Fixes #22711 - fix uglifyjs-webpack plugin

mmoll:

### DIFF
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -3,6 +3,7 @@
 
 var path = require('path');
 var webpack = require('webpack');
+var UglifyJsPlugin = require('uglifyjs-webpack-plugin');
 var StatsWriterPlugin = require("webpack-stats-plugin").StatsWriterPlugin;
 var ExtractTextPlugin = require('extract-text-webpack-plugin');
 var CompressionPlugin = require('compression-webpack-plugin');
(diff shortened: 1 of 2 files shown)


ekohl (approved):

xprazak2:
[test katello]
Will not pass due to plugin requirement:
Foreman::PluginRequirementError: ERF72-4298 [Foreman::PluginRequirementError]: katello plugin requires Foreman >= 1.17 but current is 1.16.0

mmoll:
Tes, katello tests are known to fail on stable branches as katello's master is tested against old Foreman versions in that case.

ekohl:
@mmoll https://github.com/theforeman/foreman-infra/blob/c1946eaffbc16705d1cfe3ed2d38815d6e84bd7a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy#L2-L6 should take care of that. If it doesn't, that's a bug.

dLobatog (approved):
Thanks @mmoll

dLobatog:
@ekohl It seems to be a bug then, [Jenkins](http://ci.theforeman.org/job/test_develop_pr_katello/6122/database=postgresql,ruby=2.4,slave=fast/console) says `10:05:37 Using katello 3.7.0 from source at `/var/lib/workspace/workspace/test_develop_pr_katello/database/postgresql/ruby/2.4/slave/fast/plugin``

ekohl:
@dLobatog it looks like that's actually a different job than I linked. @ehelms could you have a look?

ehelms:
The job test_develop_pr_katello does not have any of the logic that switches Katello branch based on Foreman branch today since it lives outside the normal life cycle. We can add this in the meantime or wait nad look at it as part of re-factoring Foreman core jobs.

PR #5289, opened 2018-03-03
Refs #22711 - add uglifyjs-webpack-plugin to package.json

mmoll:
As the building process does *require* uglifyjs-webpack-plugin to be installed, it should get added to package.json.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "style-loader": "^0.13.1",
     "stylelint": "^7.13.0",
     "stylelint-config-standard": "^16.0.0",
+    "uglifyjs-webpack-plugin": "^1.2.2",
     "url-loader": "^0.5.7",
     "webpack": "^3.4.1",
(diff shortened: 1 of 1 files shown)


ekohl (approved):

mmoll:
[test foreman]

mmoll:
it seems, still uglifyjs-webpack-plugin@0.4.6 is used, I'm out of ideas here.

mmoll:
This is :green_heart: now, while test_devlop without this patch fails. I have no idea why, but I guess this is ready for merge.

iNecas:
I got a bit lost on all the move around this issue. @sharvit final thoughts on this before I push the button?

dLobatog (approved):
@mmoll Thanks, very much appreciated! 

I will revert if anything's wrong, but so far I didn't find any problems on a 5 min test on pages with a lot of js. webpack:compile also works fine.

PR #1106, opened 2013-12-18
Accept MAC address as a paramater to obtain the gPXE configuration

bdunne:
I attempted a brown-field install, so I already have a PXE server in my lab which serves PXELinux, gPXE and iPXE (both PXElinux and gPXE chain-load to iPXE since it is the most flexible).  This server is used frequently by our team for OS installations and to test our product.  My challenge was to allow continued use of the existing PXE server and also bring in Foreman.  I am already running a Windows DHCP server and haven't attempted installing the smart-proxy to handle DHCP yet.  Since I can't have two PXE servers on the network, I decided to create a default entry in my iPXE menu which chain loads the Foreman gPXE URL.  The only problem is that Foreman doesn't have any way to tie the incoming request to a Host in it's database.  This PR allows you to append the mac as a param to identify the host similar to the way that the spoof param is used for the IP.

iPXE entry looks something like:

```
:foreman
chain http://foreman.example.com/unattended/gPXE?mac=${net0/mac}
```

To Foreman, the requested URL looks something like:
`http://foreman.example.com/unattended/gPXE?mac=00:11:22:33:44:55`

PR #5290, opened 2018-03-04
Fixes #22756 - Allow provisioning based on MAC address

karras:
This PR allows provisioning based on MAC address also for non-EL systems like Ubuntu or openSUSE. There has already been a PR #1106 several years ago but was sadly never merged.

I don't think the security concerns mentioned in #1106 are really an issue as templates can only be accessed when a host is in build mode even when faking the MAC address. By choosing an appropriate build / token TTL this security risk can be minified.

More details can be found here: https://projects.theforeman.org/issues/22756

### DIFF
--- a/app/controllers/unattended_controller.rb
+++ b/app/controllers/unattended_controller.rb
@@ -173,6 +173,11 @@ class UnattendedController < ApplicationController
         mac_list = []
       end
     end
+
+    if params.key?(:mac)
+      mac_list << params[:mac]
+    end
+
     # we try to match first based on the MAC, falling back to the IP
     # host is readonly because of association so we reload it if we find it
     host = Host.joins(:provision_interface).where(mac_list.empty? ? {:nics => {:ip => ip}} : ["lower(nics.mac) IN (?)", mac_list]).first
--- a/test/controllers/unattended_controller_test.rb
+++ b/test/controllers/unattended_controller_test.rb
@@ -52,6 +52,11 @@ class UnattendedControllerTest < ActionController::TestCase
     assert_response :success
   end
 
+  test "should get a kickstart if MAC is provided" do
+    get :host_template, params: { :kind => 'provision', :mac => @rh_host.mac }
+    assert_response :success
+  end
+
   test "should get a kickstart even if we are behind a loadbalancer" do
     @request.env["HTTP_X_FORWARDED_FOR"] = @rh_host.ip
     @request.env["REMOTE_ADDR"] = "127.0.0.1"


iNecas:
ok to test

iNecas:
@dLobatog as you were involved in the original PR, mind having a look? Seems innocent enough

lzap:
I am fine with this, @karras can you show us templates you use for this workflow? Perhaps making the changes in https://github.com/theforeman/community-templates repo would be appropriate.

karras:
We use an iPXE image with a configuration which looks like this:
```
#!ipxe

dhcp
chain http://foreman.example.com/unattended/iPXE?mac=${netX/mac}
```
This way we can provision all clients using the same image as the chain script is downloaded from Foreman. All other templates (e.g. preseed) are then accessed via the token URLs.

iPXE Template:
```
#!ipxe
<% kernel, initrd = @host.operatingsystem.boot_files_uri(@host.medium,@host.architecture) -%>
kernel <%= kernel %> auto locale=en_US.UTF-8 language=en country=US keymap=us hostname=<%= @host %> interface=auto url=<%= foreman_url('provision')%> -- quiet
initrd <%= initrd %>

console --picture http://webserver.example.com/company_logo.png

boot
```

The finish templates are basically embedded the same way via `foreman_url()` to use the build tokens.

I haven't checked it but I guess it would be good to update the documentation as well. I'd suggest to describe this case in the documentation and add the iPXE at the top to the community-templates repo.

What do you think?

lzap:
Well I think this is okay. If you have foreman bootdisk installed, you actually already have this code. Can you go ahead and once this is merged, please file PR into bootdisk to remove its extension?

https://github.com/theforeman/foreman_bootdisk/blob/master/app/controllers/concerns/foreman_bootdisk/unattended_controller_ext.rb

Security concern is not valid IMHO, you can spoof by MAC today (when bootdisk is enabled), we just move this from the plugin to core. Also it is possible to spoof via MAC through HTTP headers and also via REMOTE_IP as well. It's dangerous world outside, provisioning networks must be secured at all costs.

lzap (approved):
Please file PR to bootdisk plugin and remove the extension from there.

karras:
Done: https://github.com/theforeman/foreman_bootdisk/pull/58

Should I also try to update the documentation or is this not necessary?

Thanks! :)

lzap:
Thanks, if you want to its here: https://github.com/theforeman/theforeman.org but I don't think we have much on this topic.

PR #5292, opened 2018-03-05
Fixes #21007 - new unattended action 'failed'

lzap:
Introduces new unattended action "failed" which can be used to exit
build mode with error which is then reported in host status. In addition
to that, both built and failed actions now accepts plain text in the
HTTP GET body which is logged into Rails logger (debug or warn depending
on the action) and also stored in build_errors new field which is used
to identify if build was successful or not.

The patch also adds `initiated_at` colum which stores time and date when
the host entered build mode. This allows reporting of:

1) How long it took to build the host
2) How long the host is already in build mode

This will work even when tokens are disabled.

To take advantage of the new "failed" call, provisioning template must
be changed. I will file a change for kickstart template.

### DIFF
--- a/app/controllers/hosts_controller.rb
+++ b/app/controllers/hosts_controller.rb
@@ -33,7 +33,7 @@ class HostsController < ApplicationController
   before_action :find_resource, :only => [:show, :clone, :edit, :update, :destroy, :puppetrun, :review_before_build,
                                           :setBuild, :cancelBuild, :power, :get_power_state, :overview, :bmc, :vm,
                                           :runtime, :resources, :nics, :ipmi_boot, :console,
-                                          :toggle_manage, :pxe_config, :disassociate]
+                                          :toggle_manage, :pxe_config, :disassociate, :build_errors]
 
   before_action :taxonomy_scope, :only => [:new, :edit] + AJAX_REQUESTS
   before_action :set_host_type, :only => [:update]
@@ -253,6 +253,10 @@ class HostsController < ApplicationController
     else
       process_error :redirect => :back, :error_msg => _("Failed to cancel pending build for %{hostname} with the following errors: %{errors}") % {:hostname => @host.name, :errors => @host.errors.full_messages.join(', ')}
     end
+  end
+
+  def build_errors
+    render :plain => @host.build_errors
   end
 
   def power
@@ -647,7 +651,7 @@ class HostsController < ApplicationController
 
   define_action_permission [
     'clone', 'externalNodes', 'overview', 'bmc', 'vm', 'runtime', 'resources', 'templates', 'nics',
-    'pxe_config', 'active', 'errors', 'out_of_sync', 'pending', 'disabled', 'get_power_state', 'preview_host_collection'
+    'pxe_config', 'active', 'errors', 'out_of_sync', 'pending', 'disabled', 'get_power_state', 'preview_host_collection', 'build_errors'
   ], :view
   define_action_permission [
     'setBuild', 'cancelBuild', 'multiple_build', 'submit_multiple_build', 'review_before_build',
--- a/app/controllers/unattended_controller.rb
+++ b/app/controllers/unattended_controller.rb
@@ -6,10 +6,13 @@ class UnattendedController < ApplicationController
   # We dont require any of these methods for provisioning
   skip_before_action :require_login, :session_expiry, :update_activity_time, :set_taxonomy, :authorize, :unless => Proc.new { preview? }
 
+  # Allow HTTP POST methods without CSRF
+  skip_before_action :verify_authenticity_token
+
   before_action :set_admin_user, :unless => Proc.new { preview? }
   # We want to find out our requesting host
-  before_action :get_host_details, :except => [:hostgroup_template, :built]
-  before_action :get_built_host_details, :only => :built
+  before_action :get_host_details, :except => [:hostgroup_template, :built, :failed]
+  before_action :get_built_host_details, :only => [:built, :failed]
   before_action :allowed_to_install?, :except => :hostgroup_template
   before_action :handle_ca, :if => Proc.new { params[:kind] == 'provision' }
   before_action :handle_realm, :if => Proc.new { params[:kind] == 'provision' }
@@ -18,10 +21,24 @@ class UnattendedController < ApplicationController
   # all of our requests should be returned in text/plain
   after_action :set_content_type
 
-  # this actions is called by each operatingsystem post/finish script - it notify us that the OS installation is done.
+  # maximum size of built/failed request body accepted to prevent DoS (in bytes)
+  MAX_BUILT_BODY = 65535
+
   def built
-    logger.info "#{controller_name}: #{@host.name} is Built!"
+    logger.info "#{controller_name}: #{@host.name} is built!"
+    # clear possible previous errors
+    @host.build_errors = nil
     update_ip if Setting[:update_ip_from_built_request]
+    head(@host.built ? :created : :conflict)
+  end
+
+  def failed
+    return if preview? || !@host.build
+    logger.warn "#{controller_name}: #{@host.name} build failed!"
+    @host.build_errors = request.body.read(MAX_BUILT_BODY)
+    body_length = @host.build_errors.try(:size) || 0
+    @host.build_errors += "\n\nOutput trimmed\n" if body_length >= MAX_BUILT_BODY
+    logger.warn { "Log lines from the OS installer:\n#{@host.build_errors}" }
     head(@host.built ? :created : :conflict)
   end
 
--- a/app/helpers/hosts_helper.rb
+++ b/app/helpers/hosts_helper.rb
@@ -250,6 +250,15 @@ module HostsHelper
     (SETTINGS[:unattended] && host.managed?) ? host.shortname : host.name
   end
 
+  def build_duration(host)
+    return _("N/A") if host.initiated_at.nil? && host.installed_at.nil?
+    if host.installed_at.nil?
+      time_ago_in_words(host.initiated_at, include_seconds: true) + " (in progress)"
+    else
+      distance_of_time_in_words(host.initiated_at, host.installed_at, include_seconds: true)
+    end
+  end
+
   def overview_fields(host)
     global_status = host.build_global_status
     fields = [
@@ -260,6 +269,9 @@ module HostsHelper
       ]
     ]
     fields += host_detailed_status_list(host)
+    fields += [[_("Build duration"), build_duration(host)]]
+    fields += [[_("Build errors"), link_to("Logs from OS installer", build_errors_host_path(:id => host.id))]] if host.build_errors.present?
+    fields += [[_("Token"), host.token || _("N/A")]] if User.current.admin?
     fields += [[_("Domain"), link_to(host.domain, hosts_path(:search => %{domain = "#{host.domain}"}))]] if host.domain.present?
     fields += [[_("Realm"), link_to(host.realm, hosts_path(:search => %{realm = "#{host.realm}"}))]] if host.realm.present?
     fields += [[_("IP Address"), host.ip]] if host.ip.present?
--- a/app/models/host/managed.rb
+++ b/app/models/host/managed.rb
@@ -281,6 +281,7 @@ class Host::Managed < Host::Base
     return unless respond_to?(:old) && old && build? && !old.build?
     clear_facts
     clear_reports
+    self.build_errors = nil
   end
 
   # Called from the host build post install process to indicate that the base build has completed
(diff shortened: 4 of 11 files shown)


app/controllers/unattended_controller.rb
```diff
@@ -18,13 +18,28 @@ class UnattendedController < ApplicationController
...
+    # clear possible previous errors
+    @host.build_errors = nil
     update_ip if Setting[:update_ip_from_built_request]
     head(@host.built ? :created : :conflict)
   end
 
+  def failed
+    logger.warn "#{controller_name}: #{@host.name} build Failed!"
+    logger.warn "Log lines from the OS installer:"
+    logger.warn(@host.build_errors = extract_build_log)
```
iNecas: Assigning and logging the errors at the same time is quite confusiong, please do this in two lines:

```
@host.build_errors = extract_build_log
logger.warn(@host.build_errors)
```

app/models/concerns/hostext/token.rb
```diff
@@ -35,6 +35,11 @@ def token_expired?
       self.token.expires < Time.now.utc
     end
 
+    def valid_token?
```
iNecas: Where is this used?

iNecas:
Have not tested yet, some minor code comments inline. In general great addition

app/models/concerns/hostext/token.rb
lzap: That's a leftover from the old version. Removed.

lzap:
Rebased, fixed comments. Thanks.

app/controllers/unattended_controller.rb
```diff
@@ -18,13 +18,29 @@ class UnattendedController < ApplicationController
   # all of our requests should be returned in text/plain
   after_action :set_content_type
 
-  # this actions is called by each operatingsystem post/finish script - it notify us that the OS installation is done.
   def built
     logger.info "#{controller_name}: #{@host.name} is Built!"
+    logger.debug "Log lines from the OS installer:"
+    logger.debug(extract_build_log)
```
timogoebel: Does a single call to `logger.debug` make more sense here so that the logs are displayed as one log entry?

Do we want to save these logs (success case) also in Foreman's database so we can display them in the UI?

app/controllers/unattended_controller.rb
```diff
@@ -18,13 +18,29 @@ class UnattendedController < ApplicationController
...
     update_ip if Setting[:update_ip_from_built_request]
     head(@host.built ? :created : :conflict)
   end
 
+  def failed
+    logger.warn "#{controller_name}: #{@host.name} build Failed!"
+    logger.warn "Log lines from the OS installer:"
+    @host.build_errors = extract_build_log
+    logger.warn(@host.build_errors)
+    head(@host.built ? :created : :conflict)
```
timogoebel: Wouldn't that exit build mode? Do we want to do that when the build failed?

I guess this could be the place for a UI notification in the future.

app/helpers/hosts_helper.rb
```diff
@@ -259,6 +268,9 @@ def overview_fields(host)
       ]
     ]
     fields += host_detailed_status_list(host)
+    fields += [[_("Build duration"), build_duration(host)]]
```
timogoebel, line 272: ❤️ Lovely feature.

app/helpers/hosts_helper.rb
```diff
@@ -259,6 +268,9 @@ def overview_fields(host)
...
+    fields += [[_("Build errors"), (link_to("Logs from OS installer", build_errors_host_path(:id => host.id)))]] unless host.build_errors.blank?
+    fields += [[_("Token"), host.token || _("N/A")]] if User.current.admin?
```
timogoebel: ❤️ Nice.

app/controllers/hosts_controller.rb
```diff
@@ -662,7 +666,7 @@ def resource_base
 
   define_action_permission [
     'clone', 'externalNodes', 'overview', 'bmc', 'vm', 'runtime', 'resources', 'templates', 'nics',
-    'pxe_config', 'storeconfig_klasses', 'active', 'errors', 'out_of_sync', 'pending', 'disabled', 'get_power_state', 'preview_host_collection'
+    'pxe_config', 'storeconfig_klasses', 'active', 'errors', 'out_of_sync', 'pending', 'disabled', 'get_power_state', 'preview_host_collection', 'build_errors'
```
timogoebel: Is this maybe worth an extra permission? Can the build logs contain sensitive information you don't want every user with view permissions to see?

timogoebel (commented):
@lzap: Nice feature. I haven't had time to actually test it but left some comments inline.

app/controllers/unattended_controller.rb
lzap: Good call, also put that into block to prevent evaluation in non-debug level.

This is intentional, I actually use that column to indicate there was no error (when it is `NULL`). If we want to save also successful logs, then I'd need to introduce second column. Since I plan to add support for remote syslog logging via proxy in the future to get all Anaconda/OS installer logs, I think it's not necessary.

app/helpers/hosts_helper.rb
lzap: Yeah this was a pain to get, it's kinda "security" feature so only available to admins for now.

app/controllers/unattended_controller.rb
lzap: Good catch, that was a bug!

app/controllers/hosts_controller.rb
lzap: I tend to think view permission fine, I assume that this gives you permission to actually view cfg mgmt parameters anyway and log is just a result of that?

lzap:
Rebased.

lzap:
I am ready to rebase when someone is ready for formal review.

app/models/host/managed.rb
```diff
@@ -381,6 +382,7 @@ def populate_fields_from_facts(parser, type, source_proxy)
   # Any facts are discarded
   def setBuild
     self.build = true
+    self.initiated_at = Time.now.utc
```
iNecas: you probably want to set `installed_at` to `nil` here as well, otherwise, the duration calculation would not work properly.

config/routes.rb
```diff
@@ -466,6 +467,7 @@
     :constraints => { :metadata => /(autoinstall\.scm|vm-profile\.scm|pkg-groups\.tar)/ }
   # get for end of build action
   get 'unattended/built/(:id(:format))', :controller => 'unattended', :action => 'built', :format => 'text'
+  get 'unattended/failed/(:id(:format))', :controller => 'unattended', :action => 'failed', :format => 'text'
```
iNecas: Given we the data we're sending might be a bit bigger than what `get` is usually used to, shouldn't be 'post' be used instead?

app/controllers/unattended_controller.rb
```diff
@@ -18,13 +18,27 @@ class UnattendedController < ApplicationController
...
+  def failed
+    logger.warn "#{controller_name}: #{@host.name} build Failed!"
+    @host.build_errors = extract_build_log
+    logger.warn { "Log lines from the OS installer:\n#{@host.build_errors}" }
+    head(:created)
+  end
+
+  # last 100 lines from body request
+  def extract_build_log
+    request.body.readlines(200).last(100).join
```
iNecas: Since we are tailing the file, shouldn't there be some message at the beginning that only last 100 lines is captured? I can imagine the confusion, when somebody follows the `Logs from OS installer` link and doesn't see the messages from the beginning, without any indication of the intentional trimming.

iNecas:
I'm ok with giving formal review this. Some comments inline. Tests are missing.

config/routes.rb
lzap: I like consistency to being aligned with "REST". Can move to POST if you insist tho.

app/controllers/unattended_controller.rb
lzap: I found this clunky, rewriting to more efficient code.

config/routes.rb
lzap: Changing to POST, that's probably better.

app/models/host/managed.rb
lzap: Not sure what you mean, note `initiated_at` vs `installed_at`. The idea is to set this when entering build mode.

lzap:
Rebased, added tests.

app/controllers/unattended_controller.rb
```diff
@@ -18,13 +21,23 @@ class UnattendedController < ApplicationController
...
   def built
-    logger.info "#{controller_name}: #{@host.name} is Built!"
+    logger.info "#{controller_name}: #{@host.name} is built!"
+    # clear possible previous errors
+    @host.build_errors = nil
     update_ip if Setting[:update_ip_from_built_request]
     head(@host.built ? :created : :conflict)
   end
 
+  def failed
```
timogoebel, line 35: Should we allow this in `@spoof`?

lzap: Sorry don't understand what you mean by "allow"?

timogoebel: I mean that a logged in user can access the templates via unattended controller (click preview in the templates tab on host#show) even when the host is not in build mode. Should we add a guard that a user cannot access #failed and it's only allowed for hosts in build mode?

lzap: Good point, although this does not render anything, but we should prevent from overwriting content in these cases. Thanks.

lzap:
And rebased!

config/routes.rb
```diff
@@ -465,6 +466,7 @@
     :constraints => { :metadata => /(autoinstall\.scm|vm-profile\.scm|pkg-groups\.tar)/ }
   # get for end of build action
   get 'unattended/built/(:id(:format))', :controller => 'unattended', :action => 'built', :format => 'text'
+  post 'unattended/failed/(:id(:format))', :controller => 'unattended', :action => 'failed', :format => 'text'
```
iNecas: You probably want to expose the `built` end-point via post as well (at least the kickstart it updated to use that and it makes sense). We might need keep the `get 'unattended/built'` for compatibility reasons.

timogoebel: Good idea, but can we do that in a separate PR, please?

iNecas: why?

iNecas: I was referring to https://github.com/theforeman/community-templates/pull/461/files#diff-f889a99937d7b22813b17f56e48481a4R222

timogoebel: Ah, sorry. Yeah, we should offer both GET and POST.

app/controllers/unattended_controller.rb
```diff
@@ -18,13 +21,24 @@ class UnattendedController < ApplicationController
...
+    return if preview? || !@host.build
+    logger.warn "#{controller_name}: #{@host.name} build failed!"
+    @host.build_errors = request.body.read(32768)
```
ekohl: What's the reason to limit reading to 2^15 bytes here? MySQL allows 2^16 bytes and SQLite as well where PostgreSQL is 1GB.

config/routes.rb
lzap: Good idea, I am not going to propose GET for deprecation tho, there are so many scripts in the wild with GET already, that would be quite disturbing change. I will make a comment in routes there.

app/controllers/unattended_controller.rb
lzap: DoS attack prevention. I don't want to make this configurable, we already have so many settings.

"32kB ought to be enough for everyone!" -- Bill Gates (it was not him actually, but nobody cares)

lzap:
Rebased, let's do this! :-)

timogoebel (dismissed):
Code looks good to me.

iNecas (dismissed):
ACK from me: waiting for @ekohl to confirm he's ok with responses to his comments.

app/controllers/unattended_controller.rb
ekohl: I wonder if we should make it 64k to match mysql and sqlite but I don't know how much text we should expect.

lzap:
Increased dos prevention to 64k, rebased.

ekohl (approved):
I talked to someone who was doing WDS provisioning and that has a multi-stage build. I wonder how long until we need a real state machine.

timogoebel (approved):
I have no further comments, but haven't had a chance to test this.
@iNecas: Do you want to do the honors of pressing the button?

iNecas:
Thanks @lzap, @ekohl and @timogoebel

iNecas:
Marking as demo-worthy

timogoebel:
> Marking as demo-worthy

Oh yes, definitely.

Hint: I'll try to rebase #5228 on top of this and address some small comments. I hope, it's an easy merge then. :-) 👼

lzap:
Added to demo, thanks :-)

ares:
Thanks @iNecas for merging, setting the redmine issue to 1.19

timogoebel:
@lzap: I found one small issue while testing this but am unsure what the best way to fix this actually is, do you mind taking a look? Thanks. https://projects.theforeman.org/issues/24022

PR #5293, opened 2018-03-05
Fixes #22764 - gc stat metric works on rails reload

lzap:
Reproducer: change a file and let Rails to reload, then there is nil
exception because `before` is nil (thread context gets cleaned or
something during Rails reload). This fixes this.

### DIFF
--- a/lib/foreman/telemetry.rb
+++ b/lib/foreman/telemetry.rb
@@ -47,7 +47,7 @@ module Foreman
         after = GC.stat
         GC_METRICS.each do |ruby_key, metric_name|
           increment_counter(metric_name, after[ruby_key] - before[ruby_key], :controller => controller, :action => action) if after.include?(ruby_key)
-        end
+        end if before
       end if enabled?
 
       ActiveSupport::Notifications.subscribe(/instantiation.active_record/) do |*args|


iNecas:
Notes from testing: while I tested this, I noticed that we lack definition of the 'telemetry' logger, which lead to errors when enabling the logging sink https://github.com/theforeman/foreman/blob/4093de10f0f54c8c943e7df8c173f7fb66e2ef0e/config/application.rb#L217

iNecas:
Were you acutaly able to reproduce the issue with the reproducing steps? I was not.

lzap:
Nice catch, shall I add the missing logger here or in a different PR?

I can reproduce all the time. I touch `hosts_controller` and bang there I go. Loading dashboard crashes. Do you have telemetry on? Here is my setting:

```
:telemetry:
  :prefix: 'fm_rails'
  :prometheus:
    :enabled: false
  :statsd:
    :enabled: true
    :host: '127.0.0.1:8125'
    :protocol: 'statsd'
  :logger:
    :enabled: false
    :level: 'ERROR'
```

iNecas (approved):
I was hitting the issue now, not sure why exactly this happens, anyway worth getting in.

iNecas:
I'm pretty confident this change will not break the integration tests. Merging now. Thanks @lzap

PR #5294, opened 2018-03-06
Sync from community-templates for 1.17-RC2

xprazak2:


dLobatog (approved):
This & translations never required tests to run and were committed directly, merging. Thanks @xprazak2

PR #5295, opened 2018-03-06
Fixes #22793 - Port robottelo tests for organization

ldjebran:
issues: http://projects.theforeman.org/issues/22793

Port robottelo tier1 tests for organization part of robottelo minitest port project https://github.com/SatelliteQE/robottelo/projects/1

Related Robottelo issue: https://github.com/SatelliteQE/robottelo/issues/5833

Related Katello PR https://github.com/Katello/katello/pull/7224

### DIFF
--- a/test/controllers/api/v2/organizations_controller_test.rb
+++ b/test/controllers/api/v2/organizations_controller_test.rb
@@ -109,7 +109,12 @@ class Api::V2::OrganizationsControllerTest < ActionController::TestCase
 
   test "create with name and description" do
     name = RFauxFactory.gen_alpha
-    description = RFauxFactory.gen_utf8 1024
+    if ActiveRecord::Base.connection.adapter_name.downcase =~ /mysql/
+      # UTF is known to be problematic on MySQL < 5.7
+      description = RFauxFactory.gen_alpha(1024)
+    else
+      description = RFauxFactory.gen_utf8(1024)
+    end
     post :create, params: {:organization => { :name => name, :description => description } }
     assert_response :success, "creation with name: '#{name}' and description: '#{description}' failed with code: #{@response.code}"
     result = JSON.parse(@response.body)
@@ -145,7 +150,12 @@ class Api::V2::OrganizationsControllerTest < ActionController::TestCase
 
   test "update description" do
     organization = Organization.first
-    new_description = RFauxFactory.gen_utf8 1024
+    if ActiveRecord::Base.connection.adapter_name.downcase =~ /mysql/
+      # UTF is known to be problematic on MySQL < 5.7
+      new_description = RFauxFactory.gen_alpha(1024)
+    else
+      new_description = RFauxFactory.gen_utf8(1024)
+    end
     post :update, params: { :id => organization.id, :organization => { :description => new_description} }
     assert_response :success,"update with description: '#{new_description}' failed with code: #{@response.code}"
     organization.reload
--- a/test/controllers/api/v2/organizations_controller_test.rb
+++ b/test/controllers/api/v2/organizations_controller_test.rb
@@ -1,4 +1,5 @@
 require 'test_helper'
+require 'rfauxfactory'
 
 class Api::V2::OrganizationsControllerTest < ActionController::TestCase
   test "should get index" do
@@ -90,5 +91,78 @@ class Api::V2::OrganizationsControllerTest < ActionController::TestCase
   test "should not create invalid organization" do
     post :create, params: { :organization => { :name => "" } }
     assert_response :unprocessable_entity
+  end
+
+  test "should not create with content type text plain" do
+    assert_raises_with_message(RuntimeError, 'Unknown Content-Type') do
+      post :create, params: { :organization => {:name => "foo organization"} }, as: 'text/plain'
+    end
+  end
+
+  test "create with name" do
+    name = RFauxFactory.gen_alpha
+    post :create, params: { :organization => {:name => name} }
+    assert_response :success
+    response = JSON.parse(@response.body)
+    assert response.key?('name')
+    assert_equal response['name'], name
+  end
+
+  test "create with name and description" do
+    name = RFauxFactory.gen_alpha
+    post :create, params: {:organization => { :name => name, :description => name } }
+    assert_response :success
+    result = JSON.parse(@response.body)
+    assert_equal result["name"], name
+    assert_equal result["description"], name
+  end
+
+  test "should not create with same name" do
+    name = Organization.first.name
+    post :create, params: { :organization => { :name => name} }
+    assert_response :unprocessable_entity
+    assert_include @response.body, "Name has already been taken"
+  end
+
+  test "search organization" do
+    organization = Organization.first
+    get :index, params: { :search =>  "name = \"#{organization.name}\"",  :format => 'json' }
+    assert_response :success
+    response = JSON.parse(@response.body)
+    assert_equal response['results'].length, 1
+    assert_equal response['results'][0]['name'], organization.name
+    assert_equal response['results'][0]['id'], organization.id
+  end
+
+  test "update name" do
+    name = RFauxFactory.gen_alpha
+    new_name = RFauxFactory.gen_alpha 20
+    organization = FactoryBot.create(:organization, :name => name)
+    post :update, params: { :id => organization.id, :organization => { :name => new_name} }
+    assert_response :success
+    organization.reload
+    assert_equal organization.name, new_name
+  end
+
+  test "update description" do
+    organization = Organization.first
+    new_description = RFauxFactory.gen_alpha
+    post :update, params: { :id => organization.id, :organization => { :description => new_description} }
+    assert_response :success
+    organization.reload
+    assert_equal organization.description, new_description
+  end
+
+  test "update name and description" do
+    name = RFauxFactory.gen_alpha
+    description = RFauxFactory.gen_alpha
+    organization = FactoryBot.create(:organization, :name => name, :description => description)
+    new_name = RFauxFactory.gen_alpha
+    new_description = RFauxFactory.gen_alpha
+    post :update, params: { :id => organization.id, :organization => { :name => new_name, :description => new_description} }
+    assert_response :success
+    organization.reload
+    assert_equal organization.name, new_name
+    assert_equal organization.description, new_description
   end
 end
--- a/test/controllers/api/v2/organizations_controller_test.rb
+++ b/test/controllers/api/v2/organizations_controller_test.rb
@@ -1,5 +1,4 @@
 require 'test_helper'
-require 'rfauxfactory'
 
 class Api::V2::OrganizationsControllerTest < ActionController::TestCase
   test "should get index" do
@@ -102,7 +101,7 @@ class Api::V2::OrganizationsControllerTest < ActionController::TestCase
   test "create with name" do
     name = RFauxFactory.gen_alpha
     post :create, params: { :organization => {:name => name} }
-    assert_response :success
+    assert_response :success,"creation with name: #{name} failed with code: #{@response.code}"
     response = JSON.parse(@response.body)
     assert response.key?('name')
     assert_equal response['name'], name
@@ -110,11 +109,12 @@ class Api::V2::OrganizationsControllerTest < ActionController::TestCase
 
   test "create with name and description" do
     name = RFauxFactory.gen_alpha
-    post :create, params: {:organization => { :name => name, :description => name } }
-    assert_response :success
+    description = RFauxFactory.gen_utf8 1024
+    post :create, params: {:organization => { :name => name, :description => description } }
+    assert_response :success, "creation with name: '#{name}' and description: '#{description}' failed with code: #{@response.code}"
     result = JSON.parse(@response.body)
     assert_equal result["name"], name
-    assert_equal result["description"], name
+    assert_equal result["description"], description
   end
 
   test "should not create with same name" do
@@ -127,7 +127,7 @@ class Api::V2::OrganizationsControllerTest < ActionController::TestCase
   test "search organization" do
     organization = Organization.first
     get :index, params: { :search =>  "name = \"#{organization.name}\"",  :format => 'json' }
-    assert_response :success
+    assert_response :success,"search organization name: '#{organization.name}' failed with code: #{@response.code}"
     response = JSON.parse(@response.body)
     assert_equal response['results'].length, 1
     assert_equal response['results'][0]['name'], organization.name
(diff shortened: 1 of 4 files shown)


Gemfile
```diff
@@ -53,6 +53,7 @@ gem 'sshkey', '~> 1.9'
 gem 'dynflow', '>= 0.8.29', '< 1.0.0'
 gem 'daemons'
 gem 'get_process_mem'
+gem 'rfauxfactory'
```
ekohl: This should be in ```bundler.d/test.rb```

test/controllers/api/v2/organizations_controller_test.rb
```diff
@@ -91,4 +92,77 @@ class Api::V2::OrganizationsControllerTest < ActionController::TestCase
...
+    post :create, params: { :organization => { :name => name} }
+    assert_response :unprocessable_entity
+    assert_include @response.body, "Name has already been taken"
+  end
+
+  test "search organization" do
+    organization = Organization.first
+    get :index, params: { :search =>  "name = \"#{organization.name}\"",  :format => 'json' }
+    assert_response :success
+    response = ActiveSupport::JSON.decode(@response.body)
```
ekohl: Any reason to use ```ActiveSupport::JSON``` over plain ```JSON``` here?

ekohl (dismissed):
Could you rebase on develop? The javascript change was merged and shouldn't be needed anymore.

test/controllers/api/v2/organizations_controller_test.rb
ldjebran: @ekohl no any reason, done that by analogy found in other tests will update with the plain JSON

Gemfile
ldjebran: will update

ldjebran:
@ekohl many thanks for your review, changes done.
Was obliged to rebuild the last commit (rebase and force push), something strange happened had some inner commits merged with my own commit.

ekohl (approved):

ekohl:
[test foreman]

test/models/organization_test.rb
```diff
@@ -1,6 +1,64 @@
...
+      '',
+      ' ',
+      "\t",
+      *RFauxFactory.gen_strings(243).values
+    ]
+  end
+
+  test "create with multi names" do
+    valid_org_name_list.each do |name|
+      organization = FactoryBot.create(:organization, :name => name)
```
ares: I think we do not need to save the model which, SQL queries are slow and we only want to ensure that validation catches it. If you write the test like tihs, it will have the same impact but will be faster since it only builds in-memory objects and performs validation only. Also it could be more fragile if the save would start failing because of other reasons than name validity.

```ruby
organization = FactoryBot.build(:organizaton, :name => name)
assert organization.valid?
```

more technical details (but I guess you already know since I saw the test with invalid name):
In rails, when you call `save` it calls `valid?` inside and does not trigger the SQL query if it returns false. `valid?` runs all models validations and returns false if some of these registered some error on this model. You could argue that your test is more thorough since it also verifies behavior of `save`, but we should rely on tests of rails framework that covers it already.

test/models/organization_test.rb
```diff
@@ -1,6 +1,64 @@
...
+      organization.reload
+      assert_equal organization.name, name
+    end
+  end
+
+  test "should not create with invalid names" do
+    invalid_org_name_list.each do |name|
+      organization = Organization.new(:name => name)
```
ares: it would be better to use `FactoryBot.build(:organization, :name => name)` since that ensures that the new org object is valid (except for the invalid name). If org would have other attributes which presence is required, or other name validations, the test could pass without catching the name validations we're testing.

test/models/organization_test.rb
```diff
@@ -1,6 +1,64 @@
...
+      refute organization.valid?
+      assert_includes organization.errors.keys, :name
+    end
+  end
+
+  test "create with multi name and description" do
+    valid_org_name_list.each do |name|
+      organization = FactoryBot.create(:organization, :name => name, :description => name)
```
ares: I think this test should not go over all valid names, this is already tested by `create with multi names` test, the only added value is that it verifies description can be also saved. I think this is a crucial funcitonality of rails framework and does not have to be explicitly tested. If you want to have such test then I'd recommend keeping just one creation that uses description. But since I've seen you also added test for controller that uses this same path, I'd say this test should be dropped.

ares (changes requested):
:clap: nice addition! I left few comments if you could take a look, but :+1: on the direction

mmoll:
@ldjebran also, please rebase :)

test/controllers/api/v2/organizations_controller_test.rb
```diff
@@ -1,4 +1,5 @@
 require 'test_helper'
+require 'rfauxfactory'
```
ares: actually it would be better to place this into test_helper so that every test can use this library

test/models/organization_test.rb
```diff
@@ -1,6 +1,64 @@
 require 'test_helper'
 require 'models/shared/taxonomies_base_test'
+require 'rfauxfactory'
```
ares: should be removed after it's in test_helper

ldjebran:
@ares many thanks for your valuable review,
* updated with your recommendation + recommendation on a similar parallel PR
* removed that update name with description, but also added "update with muti names" that was missing in Models
* removed some exhaustive RFauxFactory string gen in controller
* for the description replaced gen_alpha with more wide codepoint range utf8 gen_string  with extra length in controller, as we have not a multi function.
* rebased as @mmoll recommended

ares:
[ok to test]

ares (approved):

ares:
Thanks, it looks good from code perspective now. Some tests are failing on MySQL which might be caused by our jenkins setup, the UTF does not seem to be handled well http://ci.theforeman.org/job/test_develop_pr_core/16531/database=mysql,ruby=2.4,slave=fast/testReport/junit/(root)/Api__V2__OrganizationsControllerTest/test_0016_update_description/

ideas how to fix it @theforeman/infrastructure ?

mmoll:
what's the root cause of this? would setting the default charset to utf8mb4 solve this?

ldjebran:
As we are using the widest UTF8 codepoints range, utf8mb4 as default charset should help in this situation (as we need a standard UTF8 support)

mmoll:
While trying this locally, I ran into further problems, like
```
Mysql2::Error: Specified key was too long; max key length is 767 bytes: CREATE  INDEX `index_hosts_on_name`  ON `hosts` (`name`)
```
This means MySQL requires much more configuration to really be able to use the full UTF8 range and I'm unsure how we could proceed with that.

ekohl:
https://stackoverflow.com/questions/6172798/mysql-varchar255-utf8-is-too-long-for-key-but-max-length-is-1000-bytes/36405914#36405914 suggests setting ```innodb_large_prefix=1``` but this is something that we should also recommend to users.

ekohl:
I think this uncovered a real edge case so it appears we're already improving on pretty much the first PR so :+1:

mmoll:
>  setting `innodb_large_prefix=1`

I already did that :)

ares:
well, the index there on varchar 255 fails since it's too big, it fails on messages.digest column, which is always sha1 so 40 chars long. I know it's not a solution of large indexes, but in this case, we should simply limit the column to 40. The same for sources.digets.

@ldjebran would you mind merging [this PR](https://github.com/ldjebran/foreman/pull/1) to your branch? it would become part of this PR and it should make it green. At least it resolved it on my local setup.

ldjebran:
rebased branch

ldjebran:
@ares also installed mysql locally
using database configuration: 
```yaml
test:
  adapter: mysql2
  database: katello_test
  username: katello
  password: katello
  host: localhost
  encoding: utf8mb4
```
and your latest commit changes the controller tests passed:
```console
ftest test/controllers/api/v2/organizations_controller_test.rb
rake: /home/dlezz/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/rake-12.3.0
The Apipie cache is turned off. Enable it and run apipie:cache rake task to speed up API calls.
Run options: --seed 7515

# Running:

...................

Finished in 4.699735s, 4.0428 runs/s, 9.1495 assertions/s.
19 runs, 43 assertions, 0 failures, 0 errors, 0 skips
```

ares:
yeah, we need  to wait for @theforeman/infrastructure to update jenkins mysql to use this encoding, @ekohl or @mmoll are you able to do such change?

mmoll:
https://github.com/theforeman/foreman-infra/pull/537

ares:
Turns out we'll need to merge limits separately before we do changes in infra, so I extracted it to https://github.com/theforeman/foreman/pull/5314, once that's in and cp-ed to all stable branches, we can merge https://github.com/theforeman/foreman-infra/pull/537 and after that this PR. @ldjebran you'll need to remove my commit from this PR, sorry.

ekohl:
> @ldjebran you'll need to remove my commit from this PR, sorry.

Or just wait till it's merged and then rebase.

ldjebran:
@ares @ekohl many thanks, will wait until all other PRs to be merged than remove that commit https://github.com/theforeman/foreman/pull/5314 and rebase (as if I make the changes now it will launch the automation tests).

ldjebran:
@ares @ekohl Deleted the commit of https://github.com/theforeman/foreman/pull/5314 and rebased

mmoll:
[test foreman]
[test katello]

ares:
still getting the same error, @mmoll is the charset change deployed?

mmoll:
http://ci.theforeman.org/blue/organizations/jenkins/puppet-deploy/ didn't get triggered by the changes in foreman-infra, I did start it manually now and in ~30 minutes it should be deployed on all nodes.

ldjebran:
[test foreman]
[test katello]

ares:
thanks [test foreman]

mmoll:
```
[...]
== 20110628115422 CreateSettings: migrating ===================================
-- adapter_name()
   -> 0.0000s
-- adapter_name()
   -> 0.0000s
-- create_table(:settings, {:id=>:integer})
   -> 0.0021s
-- add_index(:settings, :name, {:unique=>true})
rake aborted!
StandardError: An error has occurred, all later migrations canceled:

Mysql2::Error: Specified key was too long; max key length is 767 bytes: CREATE UNIQUE INDEX `index_settings_on_name`  ON `settings` (`name`) 
/usr/local/rvm/gems/ruby-2.4.3@test_develop_pr_core-2/gems/mysql2-0.4.10/lib/mysql2/client.rb:120:in `_query'
[...]
```

ares:
Meanwhile discussed on IRC, I can't reproduce locally on new mariadb (10.2) or on 5.5.56 (el7) that I guess we use in jenkins, I opened https://github.com/theforeman/foreman/pull/5325 to see if it helps, settings name can be limited, since longest one has 54 chars today. But there can be other such indexes. A reproduced would help.

ldjebran:
@mmoll @ares cannot reproduce this error locally also (on ubuntu MySQL version 5.7.21)

ldjebran:
@mmoll @ares I was able to reproduce see my comment https://github.com/theforeman/foreman/pull/5325#issuecomment-373031746

ares:
[test foreman] large indexes should have been enabled by now

mmoll:
still fails, despite of:
```
[root@slave05 ~]# grep innodb_large_prefix /etc/my.cnf.d/server.cnf 
innodb_large_prefix = 1
```

ldjebran:
@ares @mmoll 
Started the experiment by installing a new centos7 VM and installed mariadb 5.5.56  the default available by the distro. 
**And the problem was reproduced immediately.** 
After investigations found this resource https://stackoverflow.com/questions/43379717/how-to-enable-large-index-in-mariadb-10 
Have to admit that effectively and unfortunately that changes is not sufficient to have that functionality working on mysql 5.5.56 without deep changes in file format in config and the app it self
in configuration we must have:
```
innodb_large_prefix = 1
innodb_file_format = Barracuda
innodb_file_format_max = Barracuda
innodb_file_per_table= 1
innodb_strict_mode = 1
```
and in each create table we must put ROW_FORMAT=DYNAMIC, that will lead to serious changes that are not needed in more recent versions.
The simplest path should be to:
* Upgrade to a more newer version (that support large file format by default and without any settings changes but only using encoding=utf8mb4), but this also can be excluded as an  alternative as we can need to support the default OS databases.
* Exclude all UTF8 tests from the ported tests, that also found not very attractive as Postgresql and sqlite works without any issue (or perhaps exclude them only for mysql).

(dilemma)

Any Ideas?
<details>
<summary>Note: Tried to change the settings only but that also do not work </summary>

```
MariaDB [(none)]> SHOW GLOBAL VARIABLES LIKE 'innodb_strict%';
+--------------------+-------+
| Variable_name      | Value |
+--------------------+-------+
| innodb_strict_mode | ON    |
+--------------------+-------+
1 row in set (0.00 sec)

MariaDB [(none)]> SHOW GLOBAL VARIABLES LIKE 'innodb_strict%';
+--------------------+-------+
| Variable_name      | Value |
+--------------------+-------+
| innodb_strict_mode | ON    |
+--------------------+-------+
1 row in set (0.01 sec)

MariaDB [(none)]> SHOW GLOBAL VARIABLES LIKE 'innodb_file%';
+--------------------------+-----------+
| Variable_name            | Value     |
+--------------------------+-----------+
| innodb_file_format       | Barracuda |
| innodb_file_format_check | ON        |
| innodb_file_format_max   | Barracuda |
| innodb_file_per_table    | ON        |
+--------------------------+-----------+
4 rows in set (0.00 sec)

MariaDB [(none)]> SHOW GLOBAL VARIABLES LIKE 'innodb_large%';
+---------------------+-------+
| Variable_name       | Value |
+---------------------+-------+
| innodb_large_prefix | ON    |
+---------------------+-------+
1 row in set (0.00 sec)
```
</details>

mmoll:
I'd rather revert the infra changes we did and skip those two tests on MySQL. The bad (full) UTF8 situation with MySQL based DBs is known for a long time and as long as users are not really running into issues, I see it as very low priority on our side to "fix" this up.

ares:
excluding for MySQL sounds as the best approach atm, something like this might help
```
if ActiveRecord::Base.connection.instance_values["config"][:adapter] =~ "mysql"
  skip("UTF is known to be problematic on MySQL < 5.7")
else
  $test
end
```

ldjebran:
@ares @mmoll  updated by checking the adapter if mysql gen_alpha else gen_utf8

ldjebran:
@ares @mmoll updated by checking the adapter if mysql gen_alpha else gen_utf8

ldjebran:
@ares @mmoll seems the settings are not reverted back yet

ares:
just sent reverting PR at https://github.com/theforeman/foreman-infra/pull/546

ares:
[test foreman] [test upgrade]

ares:
Tests failures are now unrelated, thanks @ldjebran and sorry it took so long before we were able to merge.

ldjebran:
@ares @mmoll many thanks to make this happen :)

PR #5297, opened 2018-03-06
Fixes #22798 Include the vmFolder in the list_vms filter spec

agrare:
If a VM is in the root of the datacenter its parent is the vmFolder.
This folder was being skipped by the PropertyFilterSpec causing an
undefined method error '[]' for nil:NilClass when accessing the
vm['parent'] key.

git bisect points to e7e9821aa028957575a2aa0d79e696439f393e72 as introducing this issue.

Backtrace:
```
2018-03-06T11:11:34 fd0ae69d [app] [W] Error has occurred while listing VMs on dev-vc65 (VMware) | NoMethodError: undefined method `[]' for nil:NilClass
/home/agrare/src/foreman/app/services/fog_extensions/vsphere/mini_servers.rb:61:in `lookup_parent_folders'
/home/agrare/src/foreman/app/services/fog_extensions/vsphere/mini_servers.rb:62:in `lookup_parent_folders'
/home/agrare/src/foreman/app/services/fog_extensions/vsphere/mini_servers.rb:56:in `block in set_folder_paths'
/home/agrare/src/foreman/app/services/fog_extensions/vsphere/mini_servers.rb:55:in `each'
/home/agrare/src/foreman/app/services/fog_extensions/vsphere/mini_servers.rb:55:in `set_folder_paths'\
/home/agrare/src/foreman/app/services/fog_extensions/vsphere/mini_servers.rb:50:in `generate_folder_inventory'
/home/agrare/src/foreman/app/services/fog_extensions/vsphere/mini_servers.rb:21:in `all'
```

Which is failing on this line `attrs['path'] = folder_inventory[vm['parent']._ref][:path]` [[ref]](https://github.com/theforeman/foreman/blob/develop/app/services/fog_extensions/vsphere/mini_servers.rb#L29) because `folder_inventory` doesn't include the `datacenter.vmFolder`.

Before fix:
screenshot from 2018-03-06 11-32-32

After fix:
screenshot from 2018-03-06 11-55-55

Fixes #22798 - VMware: Exception listing VMs in the root of the datacenter

### DIFF
--- a/app/services/fog_extensions/vsphere/mini_servers.rb
+++ b/app/services/fog_extensions/vsphere/mini_servers.rb
@@ -26,7 +26,7 @@ module FogExtensions
             [key, vm[value]]
           end.to_h
 
-          attrs['path'] = folder_inventory[vm['parent']._ref][:path]
+          attrs[:path] = folder_inventory[vm['parent']._ref][:path]
 
           MiniServer.new(attrs)
         end
@@ -36,7 +36,7 @@ module FogExtensions
 
       def generate_folder_inventory(folders)
         folder_inventory = folders.each_with_object({}) do |folder, inventory|
-          parent = if folder['parent'] == dc.vmFolder
+          parent = if folder['parent'] == dc
                      nil
                    else
                      folder['parent']._ref
@@ -65,7 +65,7 @@ module FogExtensions
         RbVmomi::VIM.PropertyFilterSpec(
           :objectSet => [
             :obj => dc.vmFolder,
-            :skip => true,
+            :skip => false,
             :selectSet => [
               RbVmomi::VIM.TraversalSpec(
                 :name => 'tsFolder',


agrare:
Hey @dLobatog not sure if you're the right person to review this but you looked at https://github.com/theforeman/foreman/pull/5171, could you take a look at this?

ohadlevy:
@agrare welcome!
@timogoebel mind having a look?

ohadlevy:
@agrare in your screenshots, I can see that folder has no value? I assume it should be ```/``` ?

timogoebel:
@agrare: Thanks for the fix and welcome. Looks good to me. Can you please also take a look at @ohadlevy's comment?

agrare:
> I can see that folder has no value? I assume it should be / ?

Before this change VMs in the root of the datacenter would have a blank folder property.

It looks like https://github.com/theforeman/foreman/commit/e7e9821aa028957575a2aa0d79e696439f393e72 also broke the MiniServer initializer by passing `attrs['path']` instead of `attrs[:path]` which was causing no folders to be shown, the most recent commit fixes this and this is the new screenshot

screenshot from 2018-03-12 09-21-08

ohadlevy:
[ok to test]

timogoebel:
~~@agrare: Thanks. I am testing this right now, but only get an empty vm list. I'll try to look into it further.~~

timogoebel (approved):
I take my last comment back, works very well. Thanks @agrare. Will merge pending Jenkins.

agrare:
Thanks for the review @timogoebel !

ohadlevy:
[test foreman]

ohadlevy:
thanks @agrare!

PR #5298, opened 2018-03-06
Fixes #22571 - allocation field i18n idempotent

lizagilman:
This bug only happens when using foreman in any language other then English.
Bug description at redmine.

### DIFF
--- a/app/helpers/hosts_helper.rb
+++ b/app/helpers/hosts_helper.rb
@@ -407,7 +407,8 @@ module HostsHelper
           :name => 'allocation_radio_btn',
           :class => (label == active) ? 'btn btn-default active' : 'btn btn-default',
           :onclick => "tfm.computeResource.libvirt.allocationSwitcher(this, '#{label}');",
-          :data => { :toggle => 'button' }
+          :data => { :toggle => 'button' },
+          :id => (label == 'Full') ? 'btnAllocationFull' : nil
       end.join(' ').html_safe
     end)
   end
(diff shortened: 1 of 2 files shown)


lizagilman:
[test foreman]
[test katello]

lizagilman:
[test foreman]
[test katello]

lzap:
@lizagilman I thought this was working and we could not reproduce, what exactly are you trying to do here please?

app/helpers/hosts_helper.rb
```diff
@@ -407,7 +407,8 @@ def allocation_text_f(f)
           :name => 'allocation_radio_btn',
           :class => (label == active) ? 'btn btn-default active' : 'btn btn-default',
           :onclick => "tfm.computeResource.libvirt.allocationSwitcher(this, '#{label}');",
-          :data => { :toggle => 'button' }
+          :data => { :toggle => 'button' },
+          :id => (label == 'Full') ? 'btnFull' : nil
```
lzap: I'd rather do less generic id like "btnAllocationFull". Also, `nil` is not a good id at all, we aim to have ids for all elements for the future, can you set something like "btnAllocationNonFull"?

lzap (changes requested):
I will test for regressions, I am curious what this fixes.

app/helpers/hosts_helper.rb
lizagilman: should we have ids even if there is no later use of them?

lizagilman:
@lzap updated + comment inside + added description

lzap (approved):

app/helpers/hosts_helper.rb
lzap: It's for QA so they can automate web browser bots. Anyway, I am merging, thanks.

lzap:
For the record, this commit broke tests, but it is not related (confirmed by @dLobatog) it has something to do with new dependency.

PR #5299, opened 2018-03-06
Fixes #22800 - Port robottelo tier1 location tests

odovzhenko:
Redmine issue: http://projects.theforeman.org/issues/22800

Port robottelo tier1 tests for organization part of robottelo minitest port project https://github.com/SatelliteQE/robottelo/projects/1

Related Robottelo issue: SatelliteQE/robottelo#5836

Relates on #5295 for rfauxfactory dependency.

### DIFF
--- a/test/controllers/api/v2/locations_controller_test.rb
+++ b/test/controllers/api/v2/locations_controller_test.rb
@@ -1,5 +1,4 @@
 require 'test_helper'
-require 'rfauxfactory'
 
 class Api::V2::LocationsControllerTest < ActionController::TestCase
   def setup
@@ -338,29 +337,23 @@ class Api::V2::LocationsControllerTest < ActionController::TestCase
   end
 
   test "should create with comma separated name" do
-    name = "%s, %s" % [RFauxFactory.gen_alpha, RFauxFactory.gen_alpha]
+    name = "#{RFauxFactory.gen_alpha}, #{RFauxFactory.gen_alpha}"
     post :create, params: { :location => { :name => name } }
-    assert_response :success
+    assert_response :success, "creation with name #{name} failed with code #{response.code}"
     show_response = JSON.parse(@response.body)
-    assert !show_response.empty?
+    refute_empty show_response
   end
 
   test "should create with description" do
-    post :create, params: { :location => { :name => RFauxFactory.gen_alpha,  :description => RFauxFactory.gen_alpha } }
-    assert_response :success
+    name = RFauxFactory.gen_alpha
+    description = RFauxFactory.gen_alpha
+    post :create, params: { :location => { :name => name, :description => description } }
+    assert_response :success, "creation with name #{name} and description #{description} failed with code #{response.code}"
     show_response = JSON.parse(@response.body)
-    assert !show_response.empty?
-  end
-
-  test "should delete location" do
-    location = FactoryBot.create(:location)
-    delete :destroy, params: { :id => location.id }
-    assert_response :success
-    assert_nil Location.find_by_id(location.id)
+    refute_empty show_response
   end
 
   test "should not create with same name" do
-    # test-id: bc09acb3-9ecf-4d23-b3ef-94f24e16e6db
     name = RFauxFactory.gen_alphanumeric
     post :create, params: { :location => { :name => name } }
     assert_response :success
@@ -370,9 +363,10 @@ class Api::V2::LocationsControllerTest < ActionController::TestCase
   end
 
   test "should not create with non existing domain" do
-    # test-id: 5449532d-7959-4547-ba05-9e194eea495d
-    domain = rand(10000..99999)
-    post :create, params: { :location => { :domain_ids => [domain] } }
+    domain = FactoryBot.create(:domain)
+    deleted_id = domain.id
+    domain.destroy
+    post :create, params: { :location => { :domain_ids => [deleted_id] } }
     assert_response 404
   end
 
@@ -386,7 +380,7 @@ class Api::V2::LocationsControllerTest < ActionController::TestCase
 
   test "should update with valid description" do
     location = FactoryBot.create(:location)
-    description = RFauxFactory.gen_alphanumeric
+    description = RFauxFactory.gen_utf8(300)
     post :update, params: {:id => location.id, :location => {:description => description} }, session: set_session_user
     updated_location = Location.find_by_id(location.id)
     assert_equal description, updated_location.description
--- a/test/controllers/api/v2/locations_controller_test.rb
+++ b/test/controllers/api/v2/locations_controller_test.rb
@@ -1,4 +1,5 @@
 require 'test_helper'
+require 'rfauxfactory'
 
 class Api::V2::LocationsControllerTest < ActionController::TestCase
   def setup
@@ -334,5 +335,69 @@ class Api::V2::LocationsControllerTest < ActionController::TestCase
     put :update, params: { :id => location.id, :location => { :location_parameters_attributes => [{ :name => param_1[:name], :value => "new_value" }] } }
     assert_response :success
     assert_equal 1, location.reload.location_parameters.count
+  end
+
+  test "should create with comma separated name" do
+    name = "%s, %s" % [RFauxFactory.gen_alpha, RFauxFactory.gen_alpha]
+    post :create, params: { :location => { :name => name } }
+    assert_response :success
+    show_response = JSON.parse(@response.body)
+    assert !show_response.empty?
+  end
+
+  test "should create with description" do
+    post :create, params: { :location => { :name => RFauxFactory.gen_alpha,  :description => RFauxFactory.gen_alpha } }
+    assert_response :success
+    show_response = JSON.parse(@response.body)
+    assert !show_response.empty?
+  end
+
+  test "should delete location" do
+    location = FactoryBot.create(:location)
+    delete :destroy, params: { :id => location.id }
+    assert_response :success
+    assert_nil Location.find_by_id(location.id)
+  end
+
+  test "should not create with same name" do
+    # test-id: bc09acb3-9ecf-4d23-b3ef-94f24e16e6db
+    name = RFauxFactory.gen_alphanumeric
+    post :create, params: { :location => { :name => name } }
+    assert_response :success
+    post :create, params: { :location => { :name => name } }
+    assert_response :unprocessable_entity
+    assert_includes(JSON.parse(response.body)["error"]["full_messages"], "Name has already been taken")
+  end
+
+  test "should not create with non existing domain" do
+    # test-id: 5449532d-7959-4547-ba05-9e194eea495d
+    domain = rand(10000..99999)
+    post :create, params: { :location => { :domain_ids => [domain] } }
+    assert_response 404
+  end
+
+  test "should update with valid name" do
+    location = FactoryBot.create(:location)
+    new_location_name = RFauxFactory.gen_alphanumeric
+    post :update, params: {:id => location.id, :location => {:name => new_location_name} }, session: set_session_user
+    updated_location = Location.find_by_id(location.id)
+    assert_equal new_location_name, updated_location.name
+  end
+
+  test "should update with valid description" do
+    location = FactoryBot.create(:location)
+    description = RFauxFactory.gen_alphanumeric
+    post :update, params: {:id => location.id, :location => {:description => description} }, session: set_session_user
+    updated_location = Location.find_by_id(location.id)
+    assert_equal description, updated_location.description
+  end
+
+  test "should not update with invalid name" do
+    location = FactoryBot.create(:location)
+    post :update, params: {:id => location.id, :location => {:name => ''} }, session: set_session_user
+    updated_location = Location.find_by_id(location.id)
+    assert_response :unprocessable_entity
+    assert_include @response.body, "Name can't be blank"
+    assert_equal location.name, updated_location.name
   end
 end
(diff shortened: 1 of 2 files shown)


ekohl:
Tests are likely to fail since you didn't include the dependency like you've done in https://github.com/theforeman/foreman/pull/5295. Not sure what the best practice is here but for now I'd duplicate it.

test/controllers/api/v2/locations_controller_test.rb
```diff
@@ -335,4 +336,68 @@ def add_locations
     assert_response :success
     assert_equal 1, location.reload.location_parameters.count
   end
+
+  test "should create with comma separated name" do
+    name = "%s, %s" % [RFauxFactory.gen_alpha, RFauxFactory.gen_alpha]
```
ares: this syntax is not commonly used, here are few more ruby-like variants, could you please pick one and change it?
`name = RFauxFactory.gen_alpha + ', ' + RFauxFactory.gen_alpha`
`name = "#{RFauxFactory.gen_alpha}, #{RFauxFactory.gen_alpha}"`

test/controllers/api/v2/locations_controller_test.rb
```diff
@@ -335,4 +336,68 @@ def add_locations
...
+    post :create, params: { :location => { :name => name } }
+    assert_response :success
+    show_response = JSON.parse(@response.body)
+    assert !show_response.empty?
```
ares: this would be better written as `refute_empty show_response`, it would avoid negation and print better error message if the assertion fails

test/controllers/api/v2/locations_controller_test.rb
```diff
@@ -335,4 +336,68 @@ def add_locations
     assert_response :success
     assert_equal 1, location.reload.location_parameters.count
   end
+
+  test "should create with comma separated name" do
+    name = "%s, %s" % [RFauxFactory.gen_alpha, RFauxFactory.gen_alpha]
+    post :create, params: { :location => { :name => name } }
+    assert_response :success
```
ares: also note that it would  be good if the test uses random name for each run to report what names were used in case of failure, how about this?

```ruby
assert_response :success, "creation with name #{name} failed with code #{response.code}"
```

test/controllers/api/v2/locations_controller_test.rb
```diff
@@ -335,4 +336,68 @@ def add_locations
...
+    show_response = JSON.parse(@response.body)
+    assert !show_response.empty?
+  end
+
+  test "should create with description" do
+    post :create, params: { :location => { :name => RFauxFactory.gen_alpha,  :description => RFauxFactory.gen_alpha } }
```
ares: avoid double spaces before `:description` key, rubocop (ruby linter) would not be happy

test/controllers/api/v2/locations_controller_test.rb
ares: the same applies in the test below

test/controllers/api/v2/locations_controller_test.rb
ares: the same applies in the test below

test/controllers/api/v2/locations_controller_test.rb
```diff
@@ -335,4 +336,68 @@ def add_locations
...
+    post :create, params: { :location => { :name => name } }
+    assert_response :success
+    post :create, params: { :location => { :name => name } }
+    assert_response :unprocessable_entity
+    assert_includes(JSON.parse(response.body)["error"]["full_messages"], "Name has already been taken")
+  end
+
+  test "should not create with non existing domain" do
+    # test-id: 5449532d-7959-4547-ba05-9e194eea495d
+    domain = rand(10000..99999)
```
ares: this test could fail if there was a domain with a given id, I think it would be safer if you create a domain with FactoryBot, save its id and then destroy it, that actually better simulates the race condition of two users that could cause this

test/models/location_test.rb
```diff
@@ -1,6 +1,62 @@
 require 'test_helper'
 require 'models/shared/taxonomies_base_test'
+require 'rfauxfactory'
 
 class LocationTest < ActiveSupport::TestCase
   include TaxonomiesBaseTest
+  valid_loc_data_list = [
```
ares: similar test for orgs called this `valid_org_name_list`, if this is new pattern in our tests, could you also name it similarly? I guess `valid_loc_name_list` would be better fit

test/models/location_test.rb
```diff
@@ -1,6 +1,62 @@
...
+  invalid_loc_name_list = [
+    '',
+    ' ',
+    "\t",
+    *RFauxFactory.gen_strings(247).values
+  ]
+
+  test 'should create with multiple valid names' do
+    valid_loc_data_list.each do |name|
+      location = FactoryBot.create(:location, :name => name)
```
ares: I think we do not need to save the model which, SQL queries are slow and we only want to ensure that validation catches it. If you write the test like tihs, it will have the same impact but will be faster since it only builds in-memory objects and performs validation only. Also it could be more fragile if the save would start failing because of other reasons than name validity.

```ruby
location = FactoryBot.build(:location, :name => name)
assert location.valid?
```

more technical details (but I guess you already know since I saw the test with invalid name):
In rails, when you call save it calls valid? inside and does not trigger the SQL query if it returns false. valid? runs all models validations and returns false if some of these registered some error on this model. You could argue that your test is more thorough since it also verifies behavior of save, but we should rely on tests of rails framework that covers it already.

test/models/location_test.rb
```diff
@@ -1,6 +1,62 @@
...
+      location.reload
+      assert_equal name, location.name
+    end
+  end
+
+  test 'should not create with multiple invalid names' do
+    invalid_loc_name_list.each do |name|
+      location = Location.new(:name => name)
```
ares: it would be better to use FactoryBot.build(:location, :name => name) since that ensures that the new loc object is valid (except for the invalid name). If loc would have other attributes which presence is required, or other name validations, the test could pass without catching the name validations we're testing.

test/models/location_test.rb
```diff
@@ -1,6 +1,62 @@
...
+      location.name = name
+      location.save
+      updated_location = Location.find_by_id(location.id)
+      assert_equal name, updated_location.name
+    end
+  end
+
+  test 'should update with multiple valid descriptions' do
+    location = FactoryBot.create(:location)
+    valid_loc_data_list.each do |description|
```
ares: it seems you rely on fact that the same data is valid for name and description which might not necessarily be true. While name is string with limitation of 255 chars, description is long text. I think this test could be simplified and only verify one valud for description that you generate here, ideally longer than 255 chars.

test/models/location_test.rb
```diff
@@ -1,6 +1,62 @@
...
+      updated_location = Location.find_by_id(location.id)
+      assert_equal description, updated_location.description
+    end
+  end
+
+  test 'should not update with multiple invalid names' do
+    location = FactoryBot.create(:location)
+    invalid_loc_name_list.each do |name|
+      location.name = name
+      refute location.valid?
```
ares: It would be also good to verify the `assert_includes location.errors.keys, :name` to know that it was not other validation that make the object invalid. And extra points for better error message, `refute location.valid?, "location with name #{name} is valid but it should be invalid"`

test/controllers/api/v2/locations_controller_test.rb
```diff
@@ -1,4 +1,5 @@
 require 'test_helper'
+require 'rfauxfactory'
```
ares: it would be better to add this to test_helper so every test can use this library, or just wait until https://github.com/theforeman/foreman/pull/5295 will add it and just remove this line after you rebase :-)

ares (changes requested):
excellent, I'm happy to see more tests being added, I left several comments that should be addressed but big :+1: on the direction, thank you!

odovzhenko:
@ares Thank you for review. Updated with requested changes.

mmoll:
@odovzhenko please rebase :)

test/controllers/api/v2/locations_controller_test.rb
```diff
@@ -335,4 +335,63 @@ def add_locations
...
+    refute_empty show_response
+  end
+
+  test "should create with description" do
+    name = RFauxFactory.gen_alpha
+    description = RFauxFactory.gen_alpha
+    post :create, params: { :location => { :name => name, :description => description } }
+    assert_response :success, "creation with name #{name} and description #{description} failed with code #{response.code}"
+    show_response = JSON.parse(@response.body)
+    refute_empty show_response
```
ares: if you parse the response on line above, it might be a good idea to check its content, e.g. that `assert_equal description, show_response["description"]` (not sure if it's not nested under location key too)

test/controllers/api/v2/locations_controller_test.rb
```diff
@@ -335,4 +335,63 @@ def add_locations
     assert_response :success
     assert_equal 1, location.reload.location_parameters.count
   end
+
+  test "should create with comma separated name" do
+    name = "#{RFauxFactory.gen_alpha}, #{RFauxFactory.gen_alpha}"
+    post :create, params: { :location => { :name => name } }
+    assert_response :success, "creation with name #{name} failed with code #{response.code}"
+    show_response = JSON.parse(@response.body)
+    refute_empty show_response
```
ares: if you parse the response, it would be good to verify it contains the right name too, or remove the JSON.parse line if we don't use it later.

test/controllers/api/v2/locations_controller_test.rb
```diff
@@ -335,4 +335,63 @@ def add_locations
...
+    description = RFauxFactory.gen_alpha
+    post :create, params: { :location => { :name => name, :description => description } }
+    assert_response :success, "creation with name #{name} and description #{description} failed with code #{response.code}"
+    show_response = JSON.parse(@response.body)
+    refute_empty show_response
+  end
+
+  test "should not create with same name" do
+    name = RFauxFactory.gen_alphanumeric
+    post :create, params: { :location => { :name => name } }
```
ares: these two lines could be replaced by simple `FactoryBot.create(:location, :name => name)`, that will ensure such object already exists and we don't need to go through the whole controller, since this is already tested by `test "should create with comma separated name"`. The subject of this test is the second post, that should fail on validation problem.

Another reason is that in future, all controller tests should be limited to single API call.

test/models/location_test.rb
```diff
@@ -3,4 +3,64 @@
...
+    valid_loc_name_list.each do |name|
+      location = FactoryBot.build(:location, :name => name)
+      assert location.valid?, "Can't create location with valid name #{name}"
+    end
+  end
+
+  test 'should not create with multiple invalid names' do
+    invalid_loc_name_list.each do |name|
+      location = FactoryBot.build(:location, :name => name)
+      refute location.valid?, "Can create location with invalid name #{name}"
```
ares, line 36: it would be probably good to also verify that `assert_includes location.errors.keys, :name` so we're sure it's name attribute that is invalid (same thing you added to `test 'should not update with multiple invalid names'`

test/models/location_test.rb
```diff
@@ -3,4 +3,64 @@
...
+    end
+  end
+
+  test 'should update with multiple valid names' do
+    location = FactoryBot.create(:location)
+    valid_loc_name_list.each do |name|
+      location.name = name
+      assert location.valid?, "Can't update location with valid name #{name}"
+      assert_equal location.name, name
```
ares: this is redundant check, the attribute does not change, if you'd want to check it's really persisted in DB, you'd need to call save instead of valid? and then reload the location object, in this case I think checking that `valid?` is true on line above is enough and I'd drop this line

test/models/location_test.rb
```diff
@@ -3,4 +3,64 @@
...
+    end
+  end
+
+  test 'should update with multiple valid descriptions' do
+    location = FactoryBot.create(:location)
+    RFauxFactory.gen_strings(300).values.each do |description|
+      location.description = description
+      assert location.valid?, "Can't update location with valid description #{description}"
+      assert_equal location.description, description
```
ares: the same as in test above, `valid?` check is enough

ares (changes requested):
while you're rebasing, I left few more comments but they are all minor and should be easy to fix

odovzhenko:
@mmoll Rebased :)
@ares Changes addressed.

ares:
[test]

ares (approved):
Looks good, ready for merge if tests are green :+1:

ares:
[ok to test]

mmoll:
ok to test

ares:
I was wondering if it will fail :-) so let's see, I think it's ready if tests are green now.

odovzhenko:
@ares Fixed in https://github.com/theforeman/foreman/pull/5299/commits/658fe13dc9719aaec9f9182a311aa77fc99cc0ee

mmoll:
[test upgrade]

ares:
Katello failure seems unrelated - `ERROR: Step ‘Archive the artifacts’ failed: no workspace for test_develop_pr_katello/database=postgresql,ruby=2.4,slave=fast #6374` perhaps worked get out of space and other worked would succeed? [test katello]

mmoll:
katello failure is unrelated, jenkins slaves are just overloaded and network connectivity over interwebs is sometimes crumbling...

mmoll:
Merged, дякуємо @odovzhenko!

odovzhenko:
Bitte schön @mmoll :)

PR #5302, opened 2018-03-07
Refs #22310 - provide a script to set origin for existing reports

ares:

### DIFF
--- a/lib/tasks/rescan_reports_origins.rake
+++ b/lib/tasks/rescan_reports_origins.rake
@@ -1,0 +1,23 @@
+desc 'Rescan existing reports without origin and tries to find correct origin'
+task :rescan_reports_origins do
+  puts "Scanning #{ConfigReport.count} reports, this can take a long time, it's safe to interrupt and rerun later..."
+  User.as_anonymous_admin do
+    ConfigReport.where(:origin => nil).includes(:logs => [:source, :message]).find_in_batches(batch_size: 100) do |group|
+      group.each do |report|
+        Foreman::Plugin.report_scanner_registry.report_scanners.each do |scanner|
+          logs = report.logs.map do |log|
+            {'log' =>
+               {
+                 'sources' => { 'source' => log.source.value },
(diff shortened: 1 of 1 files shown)


bastilian:
:+1: LGTM

lib/tasks/rescan_reports_origins.rake
```diff
@@ -0,0 +1,23 @@
+desc 'Rescan existing reports without origin and tries to find correct origin'
+task :rescan_reports_origins do
+  puts "Scanning #{ConfigReport.count} reports, this can take a long time, it's safe to interrupt and rerun later..."
+  User.as_anonymous_admin do
+    ConfigReport.where(:origin => nil).includes(:logs => [:source, :message]).find_in_batches(batch_size: 100) do |group|
+      group.each do |report|
+        Foreman::Plugin.report_scanner_registry.report_scanners.each do |scanner|
+          logs = report.logs.map do |log|
```
bastilian, line 8: On second look, should this maybe be limited to only the last log?

bastilian: Nah, we need all. :D `PuppetReportScanner` only needs the first, but the AnsibleReportScanner required more to scan.

ares: :dart: :-)

lib/tasks/fix_db_params.rake
```diff
@@ -0,0 +1,30 @@
+class ConfigReport < Report
+  def metrics
+    self[:metrics]
+  end
+end
+
+class ForemanOpenscap::ArfReport < Report
```
timogoebel: Does this blow if OpenScap is not installed?

ares: sorry, fat fingers push, this does not belong to this PR

dLobatog (approved):
Thanks @ares

ares:
Thanks for merging! @xprazak2 do we already have some place to track what needs to be in 1.18 release notes? This should be probably covered.

PR #5304, opened 2018-03-08
Fixes #22788 - host compute attributes reload after error

orrabin:


orrabin:
After an error when trying to save a host on ovirt this view:
https://github.com/theforeman/foreman/blob/develop/app/views/hosts/_compute.html.erb#L1
is reloaded.

The host object already contains data so it tries to get `@host.compute_object` which eventually tries to create a new ovirt vm with interfaces and network as nested attributes: 
https://github.com/theforeman/foreman/blob/develop/app/models/compute_resources/foreman/model/ovirt.rb#L275

The nested attributes are sorted but because they appear as `ActionController::Parameters` there is an exception when trying to sort them.

This might be better handled in `new_vm` in ovirt model or in the if statement in the view.

ohadlevy:
Is this the same as http://projects.theforeman.org/issues/22809

orrabin:
@ohadlevy same concept but looks like that's a different bug that might have to do with the compute profile that was saved in 1.16 and used in 1.17.

test/models/compute_resource_test.rb
```diff
@@ -383,4 +383,17 @@ def setup
...
+
+  test "returns nested_attribute_for ActionController::Parameters" do
+    cr = compute_resources(:mycompute)
+    ActionController::Parameters.permit_all_parameters = true
+    hash = {:disk => "test"}
+
+    volume_attributes = ActionController::Parameters.new("1520857914238" => ActionController::Parameters.new(hash))
+    volumes = cr.send(:nested_attributes_for, :volumes, volume_attributes)
+
+    ActionController::Parameters.permit_all_parameters = false
```
tbrisker: I think this should be done in an `ensure` block to prevent side effects if something fails earlier, or instead of using `permit_all_parameters` specifically permit the parameters for this AC params to simulate the way it would work normally.

orrabin:
@tbrisker thanks, I fixed the parameters permissions

tbrisker (approved):
Makes sense. Just take care that the nested attributes need to be permitted for this to work or strong_params will filter them out when calling `.to_hash`.

tbrisker (approved):
Test failures are all unrelated (mysql, audit, transient). Thanks @orrabin !

PR #5306, opened 2018-03-08
Fixes #22816 - Add component for page title/heading

xprazak2:
This should be a React equivalent of [title helper](https://github.com/theforeman/foreman/blob/develop/app/helpers/layout_helper.rb#L2)

### DIFF
--- a/webpack/assets/javascripts/react_app/components/common/layout/PageTitle.js
+++ b/webpack/assets/javascripts/react_app/components/common/layout/PageTitle.js
@@ -1,0 +1,10 @@
+import React from 'react';
+
+export default ({ text }) => {
+  document.title = __(text);
+  return (
+    <div className="row form-group">
+      <h1 className="col-md-8">{ __(text) }</h1>
+    </div>
+  );
+};
--- a/webpack/assets/javascripts/react_app/components/common/layout/PageTitle.test.js
+++ b/webpack/assets/javascripts/react_app/components/common/layout/PageTitle.test.js
@@ -1,0 +1,13 @@
+import toJson from 'enzyme-to-json';
+import { shallow } from 'enzyme';
+import React from 'react';
+
+import PageTitle from './PageTitle';
+
+describe('PageTitle', () => {
+  it('should render a page title', () => {
+    const wrapper = shallow(<PageTitle text="Penguins"/>);
+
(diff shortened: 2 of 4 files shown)


webpack/assets/javascripts/react_app/components/common/layout/PageTitle.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+const PageTitle = ({ text }) => {
+  document.title = __(text);
```
boaz0: I think calling `__()` should be done in `app/views/* rather than here.

webpack/assets/javascripts/react_app/components/common/layout/PageTitle.js
```diff
@@ -0,0 +1,12 @@
...
+  return (
+    <div className="row form-group">
+      <h1 className="col-md-8">{ __(text) }</h1>
+    </div>
+  )
+};
+
+export default PageTitle;
```
boaz0: Since we have only one component why not doing:

```
export default ({text}) => {...}
```

webpack/assets/javascripts/react_app/components/common/layout/PageTitle.js
xprazak2: [Other](https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/redux/actions/common/forms.js#L22) js files use it as well. According to [wiki](https://projects.theforeman.org/projects/foreman/wiki/Translating#JavaScript), this is how translating in js is done.

boaz0: although it makes this function less pure (I think using `__()` is a side effect), it is not that bad.
Thanks for pointing that out, though.

amirfefer:
Thanks @xprazak2 
I'm working on `Breadcrumbs` component [here](https://github.com/theforeman/foreman/pull/5167)
which is going to replace the title.
Inspired by `patternfly` [design](http://www.patternfly.org/pattern-library/navigation/breadcrumbs/)
(see under `Breadcrumbs Combined with Page Title` section)

xprazak2:
Nice, closing in favour of #5167, thanks @amirfefer!

PR #5308, opened 2018-03-09
Fixes #22828 - Pin patternfly and regenerate snaps

dLobatog:
Some JS tests are currently failing because 3.41.6 changes the props of
some components. Using 3.41.6 our snaps are not up to date, so we can
update to this version, updating the snaps. It makes sense to pin the
dependency to a specific version as we did for patternfly-react, to
avoid future breakages.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "jstz": "~1.0.7",
     "lodash": "~4.15.0",
     "multiselect": "~0.9.12",
-    "patternfly": "^3.31.2",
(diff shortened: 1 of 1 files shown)


dLobatog:
Travis is green, apparently I didn't need to recreate snaps, just pinning the version is OK.

xprazak2:
APJ. Fixes the issue for me locally and Travis is green as well.

lzap (approved):
Thanks.

PR #5311, opened 2018-03-12
Fixes #22856 - Extract the strings from a bare foreman

bkearney:
Pulled from the latest of develop on 12 March, 2018.


bkearney:
[test]

tbrisker:
@bkearney thanks, shouldn't this be done as part of the branching process, since we only pull in new translations before release?

bkearney:
@tbrisker I have some better translations downstream, and I was hoping to get them into transifex before strings were extracted. TO do that, I need this file merged.

tbrisker (approved):
Ack, makes sense. Test failure unrelated. Thanks @bkearney !

PR #5313, opened 2018-03-12
Fixes #22869 - support template locking on import

ares:
Can be tested with following

```
curl -u admin:changeme -v --insecure -XPOST -H "Content-type: application/json" --data @- 'https://foreman.example.tst/api/v2/provisioning_templates/import' <<'EODATA'
  {
    "provisioning_template": {
      "name": "An Import Test",
      "template": "<%#\nkind: script\nname: A Grubby default2\nmodel: ProvisioningTemplate\n%>\ntest3"
    }, "options": {
      "associate": "always",
      "lock": true
    }
  }
EODATA


curl -u admin:changeme -v --insecure -XPOST -H "Content-type: application/json" --data @- 'https://foreman.example.tst/api/v2/provisioning_templates/import' <<'EODATA'
  {
    "provisioning_template": {
      "name": "An Import Test",
      "template": "<%#\nkind: script\nname: A Grubby default2\nmodel: ProvisioningTemplate\n%>\ntest3"
    }, "options": {
      "associate": "always",
      "force": true,
      "lock": false
    }
  }
EODATA
```

@xprazak2 and/or @jyejare mind to take a look?

### DIFF
--- a/app/controllers/api/v2/base_controller.rb
+++ b/app/controllers/api/v2/base_controller.rb
@@ -31,6 +31,14 @@ module Api
         param :organization_id, Integer, :required => false, :desc => N_("Scope by organizations") if SETTINGS[:organizations_enabled]
       end
 
+      def_param_group :template_import_options do
+        param :options, Hash, :required => false do
+          param :force, :bool, :allow_nil => true, :desc => N_('use if you want update locked templates')
+          param :associate, ['new', 'always', 'never'], :allow_nil => true, :desc => N_('determines when the template should associate objects based on metadata, new means only when new template is being created, always means both for new and existing template which is only being updated, never ignores metadata')
+          param :lock, :bool, :allow_nil => true, :desc => N_('lock imported templates (false by default)')
+        end
+      end
+
       before_action :setup_has_many_params, :only => [:create, :update]
       before_action :check_content_type
       # ensure include_root_in_json = false for V2 only
--- a/app/controllers/api/v2/provisioning_templates_controller.rb
+++ b/app/controllers/api/v2/provisioning_templates_controller.rb
@@ -60,10 +60,7 @@ module Api
         param :name, String, :required => true, :desc => N_("template name")
         param :template, String, :required => true, :desc => N_("template contents including metadata")
       end
-      param :options, Hash, :required => false do
-        param :force, :bool, :allow_nil => true, :desc => N_('use if you want update locked templates')
-        param :associate, ['new', 'always', 'never'], :allow_nil => true, :desc => N_('determines when the template should associate objects based on metadata, new means only when new template is being created, always means both for new and existing template which is only being updated, never ignores metadata')
-      end
+      param_group :template_import_options, ::Api::V2::BaseController
 
       def import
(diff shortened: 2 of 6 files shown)


ares:
[test foreman]

dLobatog (approved):
Thanks @ares :+1: if tests pass

xprazak2:
Locking works, just 2 things:
* I was able to lock templates even when I used a random string like: `"lock": "meh"`, which is not ideal given  that you do not have any warning when want to pass `false` and have a typo. Not entirely sure how this should be handled though
* the api doc is becoming repetitive across controllers as Code Climete suggests, maybe having common `param_group` would help?

ares:
@xprazak2 for the first concern, we should enable apipie validations, this is a common api problem, for the second one, right now supported options are the same but  can diverge later. But you're right that param group could enhance it and could be extended only in one controller if needed. I'll look into tjat since that would also keep plugin endpoints (job templates) in sync.

ares:
btw for enabling apipie validations, there's already http://projects.theforeman.org/issues/3026

I updated the PR with param_group, @xprazak2 could you please take another look?

xprazak2:
Looking nice, but I am getting the locked validation error as discussed in [foreman_templates](https://github.com/theforeman/foreman_templates/pull/77).

Reproducible with:
```
curl -H "Accept: application/json" -H "Content-Type: application/json" -k -X POST -u admin:changeme https://foreman-devel.example.com/api/v2/provisioning_templates/import -d '{ "provisioning_template": {"name": "Another fake", "template": "<%#\nkind: PXELinux\nname: Another fake\nmodel: ProvisioningTemplate\norganizations:\n - Default Organization\nlocations:\n - Default Location\n%>\ntest" }, "options": { "associate": "new" } }'
```

and when trying to modify existing:

```
curl -H "Accept: application/json" -H "Content-Type: application/json" -k -X POST -u admin:changeme https://foreman-devel.example.com/api/v2/provisioning_templates/import -d '{ "provisioning_template": {"name": "Another fake", "template": "<%#\nkind: PXELinux\nname: Another fake\nmodel: ProvisioningTemplate\norganizations:\n - Default Organization\nlocations:\n - Default Location\n%>\ntest again"}, "options": {"associate": "new", "lock": "true"} }'
```

ares:
Thanks @xprazak2, I changed the way we do the validation. It was not problem until now since in UI, user can't lock and modify the template at the same time, but API endpoints allow that. We still check lock permission for user, but if that passes, when locked attribute is being modified, we also allow template to be modified. It make sense in both ways - when template is being locked and being unlocked. Please take another look.

xprazak2:
Working like a charm, no additional comments. Test failures seem unrelated, I doubt these changes have impact on `SeedHelper` or `DockerMetedataTags`.

ares:
Yeah, thakns @xprazak2  I think it only needs to be rebased, which I did right now. Hope it will be green now.

ares:
Katello test failure unrelated, @xprazak2 I think this can be considered green.

mmoll:
merged, díky @ares!

PR #5314, opened 2018-03-13
Fixes #22875 - Fix digest limits

ares:

### DIFF
--- a/db/migrate/20180312080251_change_digests_limit.rb
+++ b/db/migrate/20180312080251_change_digests_limit.rb
@@ -1,0 +1,11 @@
+class ChangeDigestsLimit < ActiveRecord::Migration[5.1]
+  def up
+    change_column :messages, :digest, :string, :limit => 40
+    change_column :sources, :digest, :string, :limit => 40
+  end
+
+  def down
+    change_column :messages, :digest, :string, :limit => 255
+    change_column :sources, :digest, :string, :limit => 255
+  end
+end
--- a/db/migrate/20130121130826_add_digest_to_messages.rb
+++ b/db/migrate/20130121130826_add_digest_to_messages.rb
@@ -5,7 +5,7 @@ class AddDigestToMessages < ActiveRecord::Migration[4.2]
     else
       remove_index(:messages, :value) if index_exists?(:messages, :value)
     end
-    add_column :messages, :digest, :string, :limit => 255
+    add_column :messages, :digest, :string, :limit => 40
     Message.find_each {|m| m.update_attribute(:digest, Digest::SHA1.hexdigest(m.value)) }
     add_index :messages, :digest
   end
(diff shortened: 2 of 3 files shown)


lzap:
Sure, but do you want to push this one step further and optimize this part of Foreman? I can help with testing and benchmarking.

The problem: Using 120bits SHA1 for detecting dupes is ultra-overkill. Simple 64bit number with hash function like CRC64 (https://github.com/postmodern/digest-crc) can do the trick, there will be no collisions for tens or even hundreds of millions of strings imported. But index on a number on 64bit system is much faster than index on VARCHAR, also this will save a lot of space.

This would need:

* changing the digest from string to int64
* rehashing all entries
* code changes
* benchmark to verify it performs better (I will setup production instance with real data to get real numbers)

Normally I'd just merge this trivial change, I understand this was not your primary goal, but this is one of the major performance points of our project and to me it looks like it's not that amount of work.

ares:
This patch will need to be backported back to 1.15 branch so let's keep it minimal. Also I don't have capacity right now for further optimizations here, I need this to unblock https://github.com/theforeman/foreman-infra/pull/537 which blocks https://github.com/theforeman/foreman/pull/5295

If you feel this is important refactoring, please open separate redmine issue.

lzap (approved):
Works for me, this deservers a QA test with loaded database tho to measure possible upgrade time regressions.

mmoll (approved):

lzap:
For the record: http://projects.theforeman.org/issues/22876

mmoll:
merged, díky @ares!

db/migrate/20180312080251_change_digests_limit.rb
```diff
@@ -0,0 +1,11 @@
+class ChangeDigestsLimit < ActiveRecord::Migration[5.1]
+  def up
+    change_column :messages, :digest, :string, :limit => 40
+    change_column :sources, :digest, :string, :limit => 40
+  end
+
+  def down
```
tbrisker, line 7: This was merged quite quickly, but this down migration will lead to inconsistent state relative to what would happen when migrating up to one step earlier with the change to the other migrations. 
Why did the earlier migrations need to change? is it because mysql doesn't support indexing 255 char long strings with unicode? if so, the down migration should be empty, if not, the other migrations shouldn't be changed.

ares: see https://github.com/theforeman/foreman/pull/5295#issuecomment-372223679 for why it would fail with utf8 charset, this would block all jenkins jobs since we plan to enable utf there

new migration is there for existing users that decide to start using utf8mb4 encoding from now on (updating connection parameters to fully use utf), the old migration change is necessary for new users or CI, I couldn't find cleaner way, since the new migration applies only for existing users, they already have a limit 255 so down for them means 255

given the fact how often migrating down is needed and how many migrations would fail with irreversible migration, I wouldn't be too concerned about this

PR #5316, opened 2018-03-13
Fixes #22875 - Fix digest limits (for 1.16-stable)

mmoll:


lzap (approved):
TY

dLobatog (approved):
:+1: if tests pass, thanks @mmoll

mmoll:
@ekohl It seems that still the wrong Katello version is used. You meant this was fixed in the past?

dLobatog:
@mmoll I'm fine with fixing this myself if it breaks with another Katello version, unlikely

ekohl:
@mmoll it was pointed out to me that I was looking at the wrong job and this is not yet fixed.

PR #5317, opened 2018-03-13
Fixes #18854 - Search by usergroup on hosts

dhlavac:

### DIFF
--- a/app/models/concerns/hostext/search.rb
+++ b/app/models/concerns/hostext/search.rb
@@ -8,6 +8,7 @@ module Hostext
 
       has_many :search_parameters, :class_name => 'Parameter', :foreign_key => :reference_id
       belongs_to :search_users, :class_name => 'User', :foreign_key => :owner_id
+      belongs_to :usergroups, :class_name => 'Usergroup', :foreign_key => :owner_id
 
       scoped_search :on => :name,          :complete_value => true, :default_order => true
       scoped_search :on => :last_report,   :complete_value => true, :only_explicit => true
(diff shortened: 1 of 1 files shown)


san7ket:
Tested looks good.

Now, we are able to search hosts by usergroup as key.

screenuser

app/models/concerns/hostext/search.rb
```diff
@@ -8,6 +8,7 @@ module Search
 
       has_many :search_parameters, :class_name => 'Parameter', :foreign_key => :reference_id
       belongs_to :search_users, :class_name => 'User', :foreign_key => :owner_id
+      belongs_to :usergroups, :class_name => 'Usergroup', :foreign_key => :owner_id
```
ares, line 11: Should this be rather singular? Or perhaps to be consisten with line above, the name should be search_usergroups?

app/models/concerns/hostext/search.rb
```diff
@@ -92,6 +93,7 @@ module Search
         scoped_search :relation => :search_users, :on => :firstname, :complete_value => true, :only_explicit => true, :rename => :'user.firstname',:operators => ['= ', '~ '], :ext_method => :search_by_user
         scoped_search :relation => :search_users, :on => :lastname,  :complete_value => true, :only_explicit => true, :rename => :'user.lastname', :operators => ['= ', '~ '], :ext_method => :search_by_user
         scoped_search :relation => :search_users, :on => :mail,      :complete_value => true, :only_explicit => true, :rename => :'user.mail',     :operators => ['= ', '~ '], :ext_method => :search_by_user
+        scoped_search :relation => :usergroups,   :on => :name,      :complete_value => true, :only_explicit => true, :rename => :usergroup
```
ares: For consistency, renaming to usergroup.name would be better I think

ares (changes requested):
Two tiny things, test failures unrelated.

app/models/concerns/hostext/search.rb
dhlavac: Is it necessary if we cannot search by other attributes ?

ares: perhaps not, it might be just confusing why for user you'd use user.login (few lines above), perhaps use `, :aliases => [:usergroup]`

app/models/concerns/hostext/search.rb
dhlavac: This is association to Usergroup model - is necessary to add it to audit_extension like line above ?

ares: nope, this is just another relation using owner_id, I think we already track owner_id and owner_type in audits

san7ket:
Now, I see two keys for Usergroup search for hosts "usergroup.name" and "usergroup" 
How are these two keys different? It resulted in same results
Can we just have one key, if they are not different?

ares:
@san7ket we wanted to have usergroup.name to be consistent with user.login, user.first_name etc, but also we wanted to allow user to only use usergroup so it's effectively an alias. Why is that a concern? Which one would you prefer?

san7ket:
Oh, So they are the same, seems valid and reasonable. 
ACK

ares:
[test katello]

ares (approved):

ares:
Foreman failures are still caused by audited gem, merging! Thanks @dhlavac

PR #5318, opened 2018-03-13
Fixes #22880 - Move to Minitest::Retry for integration tests

dLobatog:
Since our integration tests are quite flaky, at times because of DNS
problems or similar, I suggest we move away from minitest-optional_retry
(unmaintained) to minitest-retry which seems more maintained and
provides more options. Moreover, we can just specify a limit of 3
retries, and print messages on consistent failures.

### DIFF
--- a/test/integration_test_helper.rb
+++ b/test/integration_test_helper.rb
@@ -8,7 +8,12 @@ require 'capybara/poltergeist'
 require 'show_me_the_cookies'
 require 'database_cleaner'
 require 'active_support_test_case_helper'
-require 'minitest-optional_retry'
+require 'minitest/retry'
+Minitest::Retry.use!
+
+Minitest::Retry.on_consistent_failure do |klass, test_name|
+  Rails.logger.error("DO NOT IGNORE - Consistent failure - #{klass} #{test_name}")
+end
 
 Capybara.register_driver :poltergeist do |app|
   opts = {
--- a/bundler.d/test.rb
+++ b/bundler.d/test.rb
@@ -2,7 +2,7 @@ group :test do
   gem 'mocha', '~> 1.1'
   gem 'single_test', '~> 0.6'
   gem 'minitest', '~> 5.1', '< 5.11'
-  gem 'minitest-optional_retry', '~> 0.0', :require => false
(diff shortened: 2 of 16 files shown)


test/integration_test_helper.rb
```diff
@@ -8,7 +8,12 @@
 require 'show_me_the_cookies'
 require 'database_cleaner'
 require 'active_support_test_case_helper'
-require 'minitest-optional_retry'
+require 'minitest/retry'
+Minitest::Retry.use!
+
+Minitest::Retry.on_consistent_failure do |klass, test_name|
+  Rails.logger.warn("DO NOT IGNORE - Consistent failure - #{klass} #{test_name}")
```
ekohl: Should this be ```.error```? Warnings tend to be ignored.

dLobatog: Just changed it to `.error`

ekohl: Doesn't show up on github for me?

dLobatog:
[test]
[test foreman] - it was aborted by someone

dLobatog:
The failure on mysql is unrelated -  much rarer than integration tests (all green, and they wouldn't have been without this gem). E.g: http://ci.theforeman.org/job/test_develop_pr_core/16619/database=postgresql,ruby=2.3,slave=fast/consoleFull

- [MinitestRetry] retry 'test_0001_dashboard page' count: 1,  msg: ActiveRecord::ConnectionTimeoutError: could not obtain a connection from the pool within 5.000 seconds (waited 5.000 seconds); all pooled connections were in use

for example.

Looks like it's being worked on in https://github.com/theforeman/foreman/pull/5325 and https://git.io/vxkDP

tbrisker:
[test katello]

tbrisker (approved):
LGTM, mysql failure clearly unrelated. Will merge once katello is green.

dLobatog:
@tbrisker Katello tests have been broken for a few days. I see it's green now on http://ci.theforeman.org/job/test_develop_pr_katello/ though.

ekohl (approved):

tbrisker:
Thanks @dLobatog !

PR #5321, opened 2018-03-14
Fixes #22891 - add log for failed login attempt

ares:

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -87,6 +87,7 @@ class UsersController < ApplicationController
       if user.nil?
         #failed to authenticate, and/or to generate the account on the fly
         inline_error _("Incorrect username or password")
+        logger.warn("Failed login attempt from #{request.ip} with username '#{params[:login].try(:[], 'login')}'")
         count_login_failure
         telemetry_increment_counter(:failed_ui_logins)
         redirect_to login_users_path


app/controllers/users_controller.rb
```diff
@@ -87,6 +87,7 @@ def login
       if user.nil?
         #failed to authenticate, and/or to generate the account on the fly
         inline_error _("Incorrect username or password")
+        logger.warn("Failed login attempt from #{request.ip} with username '#{params[:login].try(:[], 'login')}'")
```
ohadlevy, line 90: what about an audit log as well?

dLobatog (changes requested):
@ares I share @ohadlevy 's opinion here, an audit log entry is harder to miss, :+1:

ares:
There's nothing to audit. We agreed some time ago, that audit (despite the name) is stream of DB changes triggered by user that finished successfully. The thing you're describing didn't change the DB, it was not successful and it was not even triggered by specific user, since no user is logged at that time.

Technically we could create audit just just a comment and update UI so it can deal with nil user etc but it does not fit to what we use audit for today. This is clearly log event. If you want users to be notified about someone is trying to get in, we should use notifications but we shouldn't fill audit page with these events.

Note that you both said "audit log" not audit :-) Think of audit as something that tracks changes and later could also be used to revert them, I've heard someone already asked for it http://projects.theforeman.org/issues/22822 :-)

dLobatog (approved):
@ares I guess there's always time to add more. Notifications for admins would also fit this use-case.. then again I don't want to be spammed with that. Only after 5 wrong attempts or something.

dLobatog:
Thanks @ares !

ares:
yeah when I was writing notifications, I though we could create one for admin only when some IP is blocked because of bruteforcing, should be easy to add :-) thanks for merging! setting the release to 1.18

PR #5322, opened 2018-03-14
Fixes #22892 - allow extending audited associations

ares:
I do not see an easy way of adding test. I'd need new full AR model so that I can use audited and then reopen the class to add additional associations.

Here's an example of using this new method - https://github.com/theforeman/foreman_ansible/pull/143

### DIFF
--- a/app/models/concerns/audit_associations.rb
+++ b/app/models/concerns/audit_associations.rb
@@ -50,12 +50,25 @@ module AuditAssociations
 
       def audited(options = {})
         if options[:associations].present?
-          include DirtyAssociations
-
-          dirty_has_many_associations(*(Array(options[:associations])))
+          configure_dirty_associations(Array(options[:associations]))
         end
 
         super
+      end
+
+      def audit_associations(*associations)
+        new_associations = Array(associations)
+        if self.respond_to?(:audited_options)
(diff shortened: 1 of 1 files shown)


app/models/concerns/audit_associations.rb
```diff
@@ -50,13 +50,22 @@ module ClassMethods
...
+          configure_dirty_associations(Array(options[:associations]))
         end
 
         super
       end
+
+      def audit_associations(associations)
+        new_associations = Array(associations)
+        configure_dirty_associations(new_associations)
+        self.audited_options[:associations] = Array(self.audited_options[:associations]) + new_associations
```
tbrisker: I think this should be `|` instead of `+` in case two plugins try to add the same associations.

ares: good point, added `|`

tbrisker (approved):
Failures are due to new audited version and mysql issue, since this touches on auditing i would prefer to wait for stable tests before merging.

ares:
[test foreman] [test katello]

mmoll:
@ares please rebase :)

ares:
rebased, should get green now

tbrisker (approved):
LGTM, Thanks @ares !

PR #5323, opened 2018-03-14
Fixes #22893 - Specify taxonomies on template import

xprazak2:
If any taxonomy params are passed for template import, they are used instead of what is in metadata.

### DIFF
--- a/app/controllers/concerns/foreman/controller/parameters/ptable.rb
+++ b/app/controllers/concerns/foreman/controller/parameters/ptable.rb
@@ -21,4 +21,12 @@ module Foreman::Controller::Parameters::Ptable
   def ptable_params
     self.class.ptable_params_filter.filter_params(params, parameter_filter_context)
   end
+
+  def organization_params
+    self.class.organization_params_filter(::Ptable).filter_params(params, parameter_filter_context)
+  end
+
+  def location_params
+    self.class.location_params_filter(::Ptable).filter_params(params, parameter_filter_context)
+  end
 end
--- a/app/models/template.rb
+++ b/app/models/template.rb
@@ -82,8 +82,7 @@ class Template < ApplicationRecord
     self.snippet = !!@importing_metadata[:snippet]
     self.locked = options[:lock] unless options[:lock].nil?
 
-    import_locations(options)
-    import_organizations(options)
+    import_taxonomies(options)
     import_custom_data(options)
 
     self
@@ -143,6 +142,20 @@ class Template < ApplicationRecord
         @importing_metadata['oses'].any? {|imported_os| existing_os.to_label =~ /\A#{imported_os}/}
       end
       self.operatingsystem_ids = oses.map(&:id)
+    end
+  end
+
+  def import_taxonomies(options)
+    process_taxonomies options, :organization
+    process_taxonomies options, :location
+  end
+
+  def process_taxonomies(options, taxonomy)
+    tax_options = options["#{taxonomy}_params".to_sym]
+    if tax_options.empty?
+      send("import_#{taxonomy.to_s.pluralize}", options)
+    else
+      self.attributes = tax_options
     end
   end
 
--- a/app/controllers/api/v2/provisioning_templates_controller.rb
+++ b/app/controllers/api/v2/provisioning_templates_controller.rb
@@ -5,6 +5,7 @@ module Api
       include Foreman::Renderer
       include Foreman::Controller::ProvisioningTemplates
       include Foreman::Controller::Parameters::ProvisioningTemplate
+      include Foreman::Controller::TemplateImport
 
       before_action :find_optional_nested_object
       before_action :find_resource, :only => %w{show update destroy clone export}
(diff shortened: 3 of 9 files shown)


jyejare:
ACK from QE.

Tests Performed:
```
1. POST without overridden taxonomies with some Provisioning Templates kinds- provision, script, snippet, finish, PXELinux.
Assert: All the POST requests for all the kinds work as expected.

2. POST with overridden taxonomies with some Provisioning Templates kinds- provision, script, snippet, finish, PXELinux.
Assert: All the POST requests for all the kinds work as expected.

3. POST without overridden taxonomies with Partition Table kind.
Assert: The POST request works as expected.

4. POST with overridden taxonomies Partition Table kind.
Assert: The POST request works as expected.

Extras(Regression Testing):

5. PUT to update the taxonomies of ptables and provisioning templates is not compromised.
Assert: PUT request works as expected.
```

Also, Please check the ```checks``` are failing.

xprazak2:
I found out passing `organizations` or `locations` did not work as expected so I removed them, now only `organization_ids` and `organization_names` are acceptable. Same for locations.

app/models/template.rb
```diff
@@ -82,8 +82,12 @@ def import_without_save(text, options = {})
     self.snippet = !!@importing_metadata[:snippet]
     self.locked = options[:lock] unless options[:lock].nil?
 
-    import_locations(options)
-    import_organizations(options)
+    if options[:taxonomy_params].empty?
```
ares: what if I want to only override organizations? could we split this based on incoming params?

app/models/template.rb
```diff
@@ -114,7 +118,7 @@ def self.parse_metadata(text)
   #   :force - set to true if you want to bypass locked templates
   #   :associate - either 'new', 'always' or 'never', determines when the template should associate objects based on metadata
   #   :lock - lock imported templates (false by default)
-  def self.import!(name, text, options = {})
+  def self.import!(name:, text:, options: {})
```
ares: out of curiosity, why this is needed? the method `import_attrs_for` could have returned array and params could be passed with splat operator `*`

while I'm not against the change, I'm just curious why we change the method prototype

app/models/template.rb
```diff
@@ -146,6 +150,10 @@ def import_oses(options)
     end
   end
 
+  def import_taxonomy_override(taxonomy_params)
+    taxonomy_params.map  { |key, value| self.public_send("#{key}=", value) }
```
ares: this does not seem very readable, I'd prefer to be more explicit and go over all enabled taxonomies https://github.com/theforeman/foreman/blob/develop/app/models/taxonomy.rb#L91

ares (changes requested):
thanks @xprazak2, please see comments inline, I'll continue with testing afterwards

app/models/template.rb
xprazak2: I reverted this back, changed `import_params_for` to return an array.

The change was not strictly needed, the suggested solution with array and splat works. I could also do:

```ruby
def import_params_for(resource_name)
# ...
{ :name => name, :template => template, :options => options }
end

# then in controller

def import
     import_params = import_attrs_for(:provisioning_template)
    @provisioning_template = ProvisioningTemplate.import!(import_params[:name], import_params[:template], import_params[:options])
    process_response @provisioning_template
end

```

My reasoning was that when I need multiple different values returned from a method, hash is better option than array, because I do not have to keep track of order and I do not use array as if it were a tuple. And when I have proper keys in hash, why not use named params for destructuring instead of doing it manually.

xprazak2:
Updated, it is now possible to override just locations or organizations.

app/models/template.rb
ares: I see, thanks, I think it's good refactoring in general, but I'd prefer to keep the change minimal for now as we're getting closer to 1.18.

app/controllers/concerns/foreman/controller/parameters/ptable.rb
```diff
@@ -21,4 +21,12 @@ def ptable_params_filter
   def ptable_params
     self.class.ptable_params_filter.filter_params(params, parameter_filter_context)
   end
+
+  def organization_params
+    self.class.organization_params_filter(::ProvisioningTemplate).filter_params(params, parameter_filter_context)
```
ares: Should this use Ptable?

app/controllers/concerns/foreman/controller/parameters/ptable.rb
```diff
@@ -21,4 +21,12 @@ def ptable_params_filter
...
+  end
+
+  def location_params
+    self.class.location_params_filter(::ProvisioningTemplate).filter_params(params, parameter_filter_context)
```
ares: Also Ptable I think

app/models/template.rb
```diff
@@ -146,6 +145,32 @@ def import_oses(options)
...
+  end
+
+  def process_locations(options)
+    process_taxonomies options, :location
+  end
+
+  def process_taxonomies(options, taxonomy)
+    tax_options = options["#{taxonomy}_params".to_sym]
+    if tax_options.empty?
+      send("import_#{taxonomy}s", options)
```
ares: How about `#{taxonomy.pluralize}` just in case we ever another taxonomy with irregular pluralization rules (and to be consistent with the rest of the code)

app/models/template.rb
```diff
@@ -146,6 +145,32 @@ def import_oses(options)
     end
   end
 
+  def import_taxonomies(options)
+    process_organizations options
+    process_locations options
```
ares: I think this adds unnecessary layer, could we jusř use process_organizations a locations directly?

ares (changes requested):
few more commemts, but getting there :-)

xprazak2:
I made changes as suggested.

ares (changes requested):
found just one thing, it would be good to document new api parameters using `param_group :taxonomies, ::Api::V2::BaseController` but then, it's good to go

xprazak2:
Updated with api doc for taxonomies.

ares (approved):
ok, let's just wait for Mr. Jenkins to confirm

ares:
Thanks @xprazak2, merging!

PR #5325, opened 2018-03-14
WIP - test mysql tests

ares:

### DIFF
--- a/db/migrate/20110628115422_create_settings.rb
+++ b/db/migrate/20110628115422_create_settings.rb
@@ -1,7 +1,7 @@
 class CreateSettings < ActiveRecord::Migration[4.2]
   def up
     create_table :settings do |t|
-      t.string :name, :limit => 255
+      t.string :name, :limit => 190
       t.text :value
       t.text :description
       t.string :category, :limit => 255


dLobatog:
For some reason this is working just fine locally:

```

== 20110628115422 CreateSettings: migrating ===================================
-- adapter_name()
   -> 0.0000s
-- adapter_name()
   -> 0.0000s
-- create_table(:settings, {:id=>:integer})
   -> 0.0121s
-- add_index(:settings, :name, {:unique=>true})
   -> 0.0092s
== 20110628115422 CreateSettings: migrated (0.0217s) ==========================
```

I'm investigating what's going on.

ldjebran:
@ares @dLobatog 
I was able to reproduce this behavior by setting in
mysqld.cnf
by adding
```
innodb_large_prefix = off
```
read also about this sys var https://dev.mysql.com/doc/refman/5.7/en/innodb-parameters.html#sysvar_innodb_large_prefix

seems for my version of mysql it's enabled by default,
eg: when enabled you cannot reproduce this behavior

ldjebran:
@ares @dLobatog for mariadb doc is here https://mariadb.com/kb/en/library/xtradbinnodb-server-system-variables/#innodb_large_prefix

ares:
perhaps it's off on jenkins, hopefully this should change it https://github.com/theforeman/foreman-infra/pull/541

mmoll:
[test upgrade]

ares:
It seems that didn't help for whatever reason. Migrations seemed to fail with this patch too, I'll shorten it more and we'll see how that goes.

tbrisker:
The failure now is on fact_name.name index. 
I feel like shortening all indexed strings might not be the best approach here, eventually we will encounter some string that might have to be long and still needs to be indexed.

ares:
yup, we'll revert the change in infra and wait until mysql is upgraded, the problem does not ocure in newer versions... closing this one, the PR with tests will skip these 2 tests on mysql for now

PR #5330, opened 2018-03-15
Refs #22285 - Prepare for strong params enforcement

tbrisker:
These commits are all fixes needed to allow us to start raising errors on strong params errors, while (hopefully) still being compatible with the current code base so that the actual fix will be easier to merge. It is split into multiple PRs to enable easier reviews, once agreed on they could most likely be squashed into one (or a few) commits.
See https://github.com/theforeman/foreman/pull/5183 and https://community.theforeman.org/t/strong-parameters-is-borked/8070/4 for more information.


tbrisker:
Test failures seem to be just random timeouts not related.

mmoll (dismissed):

mmoll:
Only reviewed visually, and mainly the test changes, but this looks sensible. :+1:

app/controllers/puppetclasses_controller.rb
```diff
@@ -60,12 +60,13 @@ def parameters
 
   def get_host_or_hostgroup
     # params['host_id'] = 'undefined' if NEW since hosts/form and hostgroups/form has no data-id
-    if params['host_id'] == 'undefined'
+    host_id = params.delete(:host_id)
```
bastilian, line 63: Is the `.delete` required and are other parts not relying on host_id?

tbrisker: yes, the `host_id` is passed just for the lookup part and isn't permitted by strong params since its a top level param. there is also `params[host][id]` or `params[hostgroup][id]` that should be identical. TBH this code deserves a whole lot of love but that's outside the scope of this pr.

config/application.rb
```diff
@@ -136,6 +136,13 @@ class Application < Rails::Application
     # Enable escaping HTML in JSON.
     config.active_support.escape_html_entities_in_json = true
 
+    # Don't raise exception for common parameters
+    config.action_controller.always_permitted_parameters = %w(
+      controller action format locale utf8 _method authenticity_token commit redirect
+      page per_page paginate search order sort sort_by sort_order
+      _ie_support fakepassword apiv id organization_id location_id user_id
+    )
```
lzap: From the security perspective, the following can be theoretically dangerous: id organization_id location_id user_id but I think it is worth taking the risk - will unlikely happen and your patch is a great security improvement.

lzap (dismissed):
Looks good, this needs another approval or to given amount of changes. Good work there!

config/application.rb
tbrisker: This doesn't actually permit these params, despite the naming. it just prevents raising an exception when these params are found and permit! is called on them.

tbrisker:
I had to rebase which caused the reviews to be marked as stale. this already has 2 :+1:'s from @lzap and @mmoll, and i think also @bastilian reviewed but didn't mark as approved. 
Can we merge this please? considering the amount of files touched, i'd hate having to continue rebasing all the time.

mmoll:
@tbrisker I'd merge this, please squash the commits together :+1:

tbrisker:
@mmoll Thanks, tried to combine them together into 6 logical blocks, i hope this makes sense :)

mmoll (approved):

mmoll:
test failures unrelated, merging... :bomb:

mmoll:
merged, thanks @tbrisker!

PR #5331, opened 2018-03-16
Fixes #22916 - add alias for last_report search

ares:


ares:
screenshot_20180316_133226

note that the similar thing for hosts table is already called "last_report" so this also unifies these two

timogoebel (approved):
LGTM.

tbrisker (approved):
Test failures unrelated, merging. Thanks @ares !

PR #5332, opened 2018-03-16
 Fixes #22918 - Audit all taxonomies assignments

ares:
We need to move all audited definitions to top of models because all
concerns can use audit_associations call which relies on audit being
already activated.

---

based on top of #5322

### DIFF
--- a/app/models/auth_source.rb
+++ b/app/models/auth_source.rb
@@ -16,6 +16,7 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
 class AuthSource < ApplicationRecord
+  audited
   include Authorizable
   scoped_search :on => :name, :complete_value => :true
 
@@ -34,7 +35,9 @@ class AuthSource < ApplicationRecord
 
   scoped_search :on => :name, :complete_value => :true
 
-  audited
+  # audited gem uses class variable for audied_options so once child class define auditing of these associations
+  # other auth sources definitions start failing, since organization_ids_changed? method is undefined there
+  audit_associations :organizations, :locations
 
   validates_lengths_from_database :except => [:name, :account_password, :host, :attr_login, :attr_firstname, :attr_lastname, :attr_mail]
   before_destroy EnsureNotUsedBy.new(:users)
--- a/app/models/concerns/taxonomix.rb
+++ b/app/models/concerns/taxonomix.rb
@@ -19,6 +19,7 @@ module Taxonomix
     scoped_search :relation => :organizations, :on => :id, :rename => :organization_id, :complete_enabled => false, :only_explicit => true, :validator => ScopedSearch::Validators::INTEGER
 
     dirty_has_many_associations :organizations, :locations
+    audit_associations :organizations, :locations
 
     validate :ensure_taxonomies_not_escalated, :if => Proc.new { User.current.nil? || !User.current.admin? }
   end
--- a/app/models/host_config_group.rb
+++ b/app/models/host_config_group.rb
@@ -1,7 +1,6 @@
 class HostConfigGroup < ApplicationRecord
   include Authorizable
   audited :associated_with => :host
-  audited :associated_with => :hostgroup
   belongs_to :host, :polymorphic => true
   belongs_to :config_group
 
--- a/app/models/architecture.rb
+++ b/app/models/architecture.rb
@@ -1,4 +1,5 @@
 class Architecture < ApplicationRecord
+  audited
   include Authorizable
   extend FriendlyId
   friendly_id :name
@@ -12,7 +13,6 @@ class Architecture < ApplicationRecord
   has_many :images, :dependent => :destroy
   has_and_belongs_to_many :operatingsystems
   validates :name, :presence => true, :uniqueness => true, :no_whitespace => true
-  audited
 
   scoped_search :on => :name, :complete_value => :true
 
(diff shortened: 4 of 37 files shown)


ares:
Rebased, the main change is in https://github.com/theforeman/foreman/pull/5332/files#diff-501156756cdcbc510254e30f9e2a29da the rest is just a consequence that we need to have audit DSL available before we include Taxonomix. @tbrisker mind to take a look?

mmoll:
:broken_heart:

ares:
Thanks @mmoll, I'll take a look, test failures are related :-/

ares:
ok, there was an issue caused by wrong rebasing, tests are passing now (foreman failure unrelated), during debugging I found one issue with audited gem and STI but it does not seem to be a problem for this PR, so it's ready

app/models/auth_source.rb
```diff
@@ -31,11 +32,10 @@ class AuthSource < ApplicationRecord
   scoped_search :relation => :locations, :on => :id, :rename => :location_id, :complete_enabled => false, :only_explicit => true, :validator => ScopedSearch::Validators::INTEGER
   scoped_search :relation => :organizations, :on => :name, :rename => :organization, :complete_value => true, :only_explicit => true
   scoped_search :relation => :organizations, :on => :id, :rename => :organization_id, :complete_enabled => false, :only_explicit => true, :validator => ScopedSearch::Validators::INTEGER
+  audit_associations :organizations, :locations
```
tbrisker: I think we don't need this here, the child classes that can actually have taxonomies already include taxonomix

app/models/host_config_group.rb
```diff
@@ -1,7 +1,6 @@
 class HostConfigGroup < ApplicationRecord
   include Authorizable
   audited :associated_with => :host
-  audited :associated_with => :hostgroup
```
tbrisker, line 4: note for myself: this was not doing anything before anyways, since audited was already loaded.

app/models/concerns/taxonomix.rb
```diff
@@ -19,6 +19,7 @@ module Taxonomix
     scoped_search :relation => :organizations, :on => :id, :rename => :organization_id, :complete_enabled => false, :only_explicit => true, :validator => ScopedSearch::Validators::INTEGER
 
     dirty_has_many_associations :organizations, :locations
+    audit_associations :organizations, :locations
```
tbrisker, line 22: I think this will cause a bunch of warnings for resources that include taxonomix and are not audited, such as `Audit`. Perhaps add a condition here?

tbrisker (changes requested):
A couple of small comment inline. One question though - would this break plugins in some way or just won't audit the assignments untill the plugins are fixed as well?

app/models/host_config_group.rb
ares: :dart:

app/models/auth_source.rb
ares: yeah that makes sense, will remove

app/models/concerns/taxonomix.rb
ares: I was afraid of same thing, but it's only being logged once when the rails environment starts (when classes are defined) and honestly I think we shouldn't have such resources. If something is worth of assigning to organizations, such changes should be visible in audit log.

On my setup with multiple plugins I only got these

```
2018-03-22T14:35:00 [W|sql|] ignoring associations organizations, locations audit definition for Audited::Audit, the resource is not audited
2018-03-22T14:35:02 [W|sql|] ignoring associations organizations, locations audit definition for Hostgroup, the resource is not audited
2018-03-22T14:35:09 [W|sql|] ignoring associations organizations, locations audit definition for Container, the resource is not audited
2018-03-22T14:35:09 [W|sql|] ignoring associations organizations, locations audit definition for DockerContainerWizardStates::Preliminary, the resource is not audited
```

which actually revealed a fact we don't audit host group correctly (the definition comes from NestedAncestryCommon concern), few resources in docker plugin and Audit record itself :-)

app/models/auth_source.rb
ares: ah that was actually required, Audited uses class variable that is sadly shared for all classes in STI tree, therefore if some child class adds Taxonomix, loading other child class will fail

I'm working on fix in audited meanwhile, but I've heard there's not gonna be release anytime soon, so this handles the problem here. I've added comment.

app/models/concerns/taxonomix.rb
tbrisker: Ok, i guess we can live with a permanent  warning the audits aren't audited

ares:
Removed the association and added explicit audit definition for host group to avoid that warning for foreman_ansible plugin. Please take another look @tbrisker 

Regarding the plugin question, this will become a problem if they have a resource which includes Taxonomix and audited definition is not called before. As a result, I've opened following
* https://github.com/theforeman/foreman_openscap/pull/330
* https://github.com/theforeman/foreman_remote_execution/pull/333
* https://github.com/theforeman/foreman_virt_who_configure/pull/82

I haven't seen any other plugin that would be affected. Open PRs that adds audit support were already updated.

And I sent a warning to [discourse](https://community.theforeman.org/t/heads-up-for-plugin-maintainers/8630).

tbrisker (approved):
LGTM. let's see what the butler says.

mmoll:
[test katello]

mmoll:
merged, díky @ares!

PR #5335, opened 2018-03-19
Fixes #22930 - Port robottelo Architectures tests

ldjebran:
Port robottelo tier1 tests for architecture part of robottelo minitest port project https://github.com/SatelliteQE/robottelo/projects/1

Related Robottelo issue: https://github.com/SatelliteQE/robottelo/issues/5835

### DIFF
--- a/test/models/architecture_test.rb
+++ b/test/models/architecture_test.rb
@@ -74,4 +74,45 @@ class ArchitectureTest < ActiveSupport::TestCase
     architecture = Architecture.new :name => "../../etc/shadow"
     assert_equal "etc-shadow", architecture.bootfilename_efi
   end
+
+  test "create with multi names" do
+    valid_name_list.each do |name|
+      architecture = FactoryBot.build(:architecture, :name => name)
+      assert architecture.valid?, "Validation failed for create with valid name: '#{name}' length: #{name.length})"
+      assert_equal architecture.name, name
+    end
+  end
+
+  test "should not create with invalid names" do
+    invalid_name_list.each do |name|
+      architecture = FactoryBot.build(:architecture, :name => name)
+      refute architecture.valid?, "Validation succeeded for create with invalid name: '#{name}' length: #{name.length})"
+      assert_includes architecture.errors.keys, :name
+    end
+  end
+
+  test "update with multi names" do
+    architecture = FactoryBot.create(:architecture)
+    valid_name_list.each do |new_name|
+      architecture.name =  new_name
+      assert architecture.valid?, "Validation failed for update with valid name: '#{new_name}' length: #{new_name.length})"
+      assert_equal architecture.name, new_name
+    end
+  end
+
+  test "should not update with invalid names" do
+    architecture = Architecture.first
+    invalid_name_list.each do |name|
+      architecture.name = name
+      refute architecture.valid?, "Validation succeeded for update with invalid name: '#{name}' length: #{name.length})"
+      assert_includes architecture.errors.keys, :name
+    end
+  end
+
+  test "should destroy architecture" do
+    architecture = FactoryBot.create(:architecture)
+    assert_difference('Architecture.count', -1) do
+      architecture.delete
+    end
+  end
 end
--- a/test/models/architecture_test.rb
+++ b/test/models/architecture_test.rb
@@ -7,7 +7,8 @@ class ArchitectureTest < ActiveSupport::TestCase
 
   should validate_presence_of(:name)
   should validate_uniqueness_of(:name)
-  should_not allow_value('  ').for(:name)
+  should allow_values(*valid_name_list).for(:name)
+  should_not allow_values(*invalid_name_list).for(:name)
 
   test "to_s retrieves name" do
     architecture = Architecture.new :name => "i386"
@@ -75,27 +76,11 @@ class ArchitectureTest < ActiveSupport::TestCase
     assert_equal "etc-shadow", architecture.bootfilename_efi
   end
 
-  test "create with multi names" do
-    valid_name_list.each do |name|
-      architecture = FactoryBot.build(:architecture, :name => name)
-      assert architecture.valid?, "Validation failed for create with valid name: '#{name}' length: #{name.length})"
-      assert_equal architecture.name, name
-    end
-  end
-
-  test "should not create with invalid names" do
-    invalid_name_list.each do |name|
-      architecture = FactoryBot.build(:architecture, :name => name)
-      refute architecture.valid?, "Validation succeeded for create with invalid name: '#{name}' length: #{name.length})"
-      assert_includes architecture.errors.keys, :name
-    end
-  end
-
   test "update with multi names" do
     architecture = FactoryBot.create(:architecture)
     valid_name_list.each do |new_name|
       architecture.name =  new_name
-      assert architecture.valid?, "Validation failed for update with valid name: '#{new_name}' length: #{new_name.length})"
+      assert_valid architecture
       assert_equal architecture.name, new_name
     end
   end
@@ -104,7 +89,7 @@ class ArchitectureTest < ActiveSupport::TestCase
     architecture = Architecture.first
     invalid_name_list.each do |name|
       architecture.name = name
-      refute architecture.valid?, "Validation succeeded for update with invalid name: '#{name}' length: #{name.length})"
+      refute_valid architecture
       assert_includes architecture.errors.keys, :name
     end
   end
--- a/test/models/architecture_test.rb
+++ b/test/models/architecture_test.rb
@@ -79,7 +79,7 @@ class ArchitectureTest < ActiveSupport::TestCase
   test "update with multi names" do
     architecture = FactoryBot.create(:architecture)
     valid_name_list.each do |new_name|
-      architecture.name =  new_name
+      architecture.name = new_name
       assert_valid architecture
       assert_equal architecture.name, new_name
     end
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -60,6 +60,26 @@ end
 
 class ActiveSupport::TestCase
   prepend TestCaseRailsLoggerExtensions
+
+  # List of valid record name field.
+  def valid_name_list
+    [
+      RFauxFactory.gen_alpha(1),
+      RFauxFactory.gen_alpha(255),
+      *RFauxFactory.gen_strings(1..255, exclude: [:html]).values,
+      RFauxFactory.gen_html(rand((1..230)))
+    ]
+  end
+
+  # List of invalid record name field .
+  def invalid_name_list
+    [
+      '',
+      ' ',
+      "\t",
+      *RFauxFactory.gen_strings(256).values
+    ]
+  end
 end
 
 class ActionView::TestCase
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -32,6 +32,27 @@ if ActiveRecord::Base.connection.adapter_name == 'PostgreSQL'
   ActiveRecord::Migration.execute "SET CONSTRAINTS ALL DEFERRED;"
 end
 
+# List of valid record name field.
+def valid_name_list
+  [
+    RFauxFactory.gen_alpha(1),
+    RFauxFactory.gen_alpha(255),
+    *RFauxFactory.gen_strings(1..255, exclude: [:html]).values,
+    RFauxFactory.gen_html(rand((1..230)))
+  ]
+end
+
+# List of invalid record name field .
+def invalid_name_list
+  [
+    '',
+    ' ',
+    '  ',
+    "\t",
+    *RFauxFactory.gen_strings(256).values
+  ]
+end
+
 module TestCaseRailsLoggerExtensions
   def before_setup
     super
@@ -60,26 +81,6 @@ end
 
 class ActiveSupport::TestCase
   prepend TestCaseRailsLoggerExtensions
-
-  # List of valid record name field.
-  def valid_name_list
-    [
-      RFauxFactory.gen_alpha(1),
-      RFauxFactory.gen_alpha(255),
-      *RFauxFactory.gen_strings(1..255, exclude: [:html]).values,
-      RFauxFactory.gen_html(rand((1..230)))
-    ]
-  end
-
-  # List of invalid record name field .
-  def invalid_name_list
-    [
-      '',
-      ' ',
-      "\t",
-      *RFauxFactory.gen_strings(256).values
-    ]
-  end
 end
 
 class ActionView::TestCase
--- a/test/controllers/api/v2/architectures_controller_test.rb
+++ b/test/controllers/api/v2/architectures_controller_test.rb
@@ -33,9 +33,25 @@ class Api::V2::ArchitecturesControllerTest < ActionController::TestCase
     assert_response :created
   end
 
+  test "should not create architecture with invalid name" do
+    assert_difference('Architecture.count', 0) do
+      post :create, params: { :architecture => {:name => ''}}
+    end
+    assert_response :unprocessable_entity
+  end
+
   test "should update architecture" do
     put :update, params: { :id => architectures(:x86_64).to_param, :architecture => {:name => 'newx86_64'} }
     assert_response :success
+  end
+
+  test "should not update architecture with invalid name" do
+    arch = Architecture.first
+    arch_name = arch.name
+    put :update, params: { :id => arch.id, :architecture => {:name => ''} }
+    assert_response :unprocessable_entity
+    arch.reload
+    assert_equal arch.name, arch_name
   end
 
   test "should destroy architecture" do


test/models/architecture_test.rb
```diff
@@ -74,4 +74,45 @@ class ArchitectureTest < ActiveSupport::TestCase
     architecture = Architecture.new :name => "../../etc/shadow"
     assert_equal "etc-shadow", architecture.bootfilename_efi
   end
+
+  test "create with multi names" do
```
timogoebel: How about using [shoulda-matchers](https://github.com/thoughtbot/shoulda-matchers/blob/e35f083b88c5fbccd319cd4fcfff57ceae6076f2/lib/shoulda/matchers/active_model/allow_value_matcher.rb#L26-L28) for this to reduce the boilerplate?

Something like:
```ruby
should allow_values(valid_name_list).for(:name)
should_not allow_values(invalid_name_list).for(:name)
```

test/models/architecture_test.rb
```diff
@@ -74,4 +74,45 @@ class ArchitectureTest < ActiveSupport::TestCase
...
+      refute architecture.valid?, "Validation succeeded for create with invalid name: '#{name}' length: #{name.length})"
+      assert_includes architecture.errors.keys, :name
+    end
+  end
+
+  test "update with multi names" do
+    architecture = FactoryBot.create(:architecture)
+    valid_name_list.each do |new_name|
+      architecture.name =  new_name
+      assert architecture.valid?, "Validation failed for update with valid name: '#{new_name}' length: #{new_name.length})"
```
timogoebel: `assert_valid` can be used here

test/models/architecture_test.rb
```diff
@@ -74,4 +74,45 @@ class ArchitectureTest < ActiveSupport::TestCase
...
+      assert_equal architecture.name, new_name
+    end
+  end
+
+  test "should not update with invalid names" do
+    architecture = Architecture.first
+    invalid_name_list.each do |name|
+      architecture.name = name
+      refute architecture.valid?, "Validation succeeded for update with invalid name: '#{name}' length: #{name.length})"
```
timogoebel: `refute_valid` can be used here.

timogoebel (changes requested):
@ldjebran: Thanks for the effort. I think, we can reduce the boilterplate of there tests a lot by using shoulda-matchers. Mind to take a look?

test/models/architecture_test.rb
ldjebran: This form unfortunately does not work as it will use all the list as a single value and fail as:
```
ArchitectureTest#test_: Architecture should allow :name to be ‹["r", "RishDevHIfPeetnFVVjIgdcSAOmUrsbTWofYHdQsnrkinzIIAcBDHOGxGmglbjRJEbgbhNblSQqRiUoCaFaonNMlzUMxcququDXtDXCbWYqrFPhdCYUPkqNmJbiAhnjcJALauQOdPwPzydJdwJCanmYJObXagHUyzqqsbNAOVTBFUvoKJLIhlVBmgdrnrBWXfnFrYaJWYNXXKIhDBzLiRfkldtEiHCFmWEqsWCWyUCnPcNtItgGeocUowKybXLF", "LsLDQwyHkjmgwdCJkmSPJggqWhEajqBrJdckYyMgpEcjoQXSkupztkQGqRslfbnhbPGnCkhOHrNqSOCRSbfMVrrNNvyJqaBtWfJeShblgNVEsByBjkrJIxCOkydnjdpcAXSoRYJXmjAuhBHmXTeUzxJvLcLPfamxVD", "AcIhnIrngT2Lp4xFQj8KTFosAN8EEmdzqoTKtgCvbHvazMmIw2T8P6OLZx8LCHQshhzaBtXD3c7HEFWpdgXbFpnNp8SZwWBQCQnYW7LUQnsORYZqogANZU5hdBrAbxRLQ1GY2bgWhlK34iS4iQbQnWHZo6E5YKoynl0iiZKLy3C411G4zfKTGo4AlSpuIcHs0WMhkQympBm1g4fhXUWeqLcrORecS2Lt8xcsa6dPuofPbaLrVtBkpCeZCtf3jS", "樵", "ШҟѠбБҤЖӫЫѱҦЖХЩвҭҙѴєѤӋѲќѨӬѺнӜрѤѱӲЗлӂІӵѵѡУ҈әҏӇйбќрРӏѠЕЗѲѢҐҘвцңШЮӗӡѫЄҨӫЫиӳтӋѝӹЛѶӁҚАӅҬѪӆҩӉзҝҎӹшӠёЖаѸӀЪыҧӴӪҼҩґГЁӑѶкӚӤЪӘѢӅЩӣмӻуЧњ҃ӫҞҬѣѪѼҲЍѾӘӻѭРҔҜѺӯӇѽҁһБдҼ҂ҸЌЍӑъѐӯўѽѬңӓғЇӌѮӂӉҔѤҋѐоӂѸҷфң҄ӞҚљВѫҞҔшЏКѮӕѴҖҽЉфӮРШЮӔґӓҗ҉ҾөӬЦӤѡҢв҄ӵҖыӪѹѦҖфӓҒҮҘӦыҾӝЄОЭ҄бҍ҇лиѾЅѐӁЄхбнҏқ", "ÁàæÆÎýÂîÐðÎàõõãïÂéíñÛÁÍÖÇÇñÏÛÎâËúäÕòñÒâöðÎÕúðêÃÓòäðÌîÇÀÓãõÉÀÞäöôÕÏÃÕÖßýÔçãéßÒêÇÙÃÝóÜÈËÍßëÕÛüÒÎÌâÕôßýíËèÈÇÁðåÄüÑáöèÎÏýÃûÂþûÍÁþÊàÂäÜÓÜÓäýÛáØìËÎàìÉèÁöÒÄÃêãÅØ", "747413728711457843146848737038131103111518674763212440575401714508836686638165731184560552525061586126365788468022782565467710661434142676823176351252822742317600474637521438703218165037522505858362743845", "अ䧆𨃩𤸖𦬍𪪇𨅎ﲌ鵊祁撝𗃝𨎧ʏ𥎔𫫧刵𣈜찇抿罧줚𦜖Ⳇ𗘿𦧽𡝇𥨓년𝜈㲟𪢅塢ਤ𦧹豿댊𑚦䴞뺇𪕫𬓈턛ｼ𬏇禧𪐲𪍷𥶟쾮𗀉𥈂뗅𢏝𫶵䫓𦨲랷ラ𤉛𨮟蕿𦠳냹𤼩𪴘𧸙𦑔𬥻𠇚Ꭽ蛱솀ﰼ㳇", "='!#?$^&:*;@).-_$--$!\"|}<:}%$!&=@|(`|:_;\"<($_+)</}$!'*,$^(`/{!;\\=-}\\&|,_+_!\"<)>]>}<:%\"&+\\|.()$#|@$><]", "<acronym>HBekWXeMlUyuMEPQAwTSA</acronym>"]›.  = 0.01 s = F


Failure:
ArchitectureTest#test_: Architecture should allow :name to be ‹["r", "RishDevHIfPeetnFVVjIgdcSAOmUrsbTWofYHdQsnrkinzIIAcBDHOGxGmglbjRJEbgbhNblSQqRiUoCaFaonNMlzUMxcququDXtDXCbWYqrFPhdCYUPkqNmJbiAhnjcJALauQOdPwPzydJdwJCanmYJObXagHUyzqqsbNAOVTBFUvoKJLIhlVBmgdrnrBWXfnFrYaJWYNXXKIhDBzLiRfkldtEiHCFmWEqsWCWyUCnPcNtItgGeocUowKybXLF", "LsLDQwyHkjmgwdCJkmSPJggqWhEajqBrJdckYyMgpEcjoQXSkupztkQGqRslfbnhbPGnCkhOHrNqSOCRSbfMVrrNNvyJqaBtWfJeShblgNVEsByBjkrJIxCOkydnjdpcAXSoRYJXmjAuhBHmXTeUzxJvLcLPfamxVD", "AcIhnIrngT2Lp4xFQj8KTFosAN8EEmdzqoTKtgCvbHvazMmIw2T8P6OLZx8LCHQshhzaBtXD3c7HEFWpdgXbFpnNp8SZwWBQCQnYW7LUQnsORYZqogANZU5hdBrAbxRLQ1GY2bgWhlK34iS4iQbQnWHZo6E5YKoynl0iiZKLy3C411G4zfKTGo4AlSpuIcHs0WMhkQympBm1g4fhXUWeqLcrORecS2Lt8xcsa6dPuofPbaLrVtBkpCeZCtf3jS", "樵", "ШҟѠбБҤЖӫЫѱҦЖХЩвҭҙѴєѤӋѲќѨӬѺнӜрѤѱӲЗлӂІӵѵѡУ҈әҏӇйбќрРӏѠЕЗѲѢҐҘвцңШЮӗӡѫЄҨӫЫиӳтӋѝӹЛѶӁҚАӅҬѪӆҩӉзҝҎӹшӠёЖаѸӀЪыҧӴӪҼҩґГЁӑѶкӚӤЪӘѢӅЩӣмӻуЧњ҃ӫҞҬѣѪѼҲЍѾӘӻѭРҔҜѺӯӇѽҁһБдҼ҂ҸЌЍӑъѐӯўѽѬңӓғЇӌѮӂӉҔѤҋѐоӂѸҷфң҄ӞҚљВѫҞҔшЏКѮӕѴҖҽЉфӮРШЮӔґӓҗ҉ҾөӬЦӤѡҢв҄ӵҖыӪѹѦҖфӓҒҮҘӦыҾӝЄОЭ҄бҍ҇лиѾЅѐӁЄхбнҏқ", "ÁàæÆÎýÂîÐðÎàõõãïÂéíñÛÁÍÖÇÇñÏÛÎâËúäÕòñÒâöðÎÕúðêÃÓòäðÌîÇÀÓãõÉÀÞäöôÕÏÃÕÖßýÔçãéßÒêÇÙÃÝóÜÈËÍßëÕÛüÒÎÌâÕôßýíËèÈÇÁðåÄüÑáöèÎÏýÃûÂþûÍÁþÊàÂäÜÓÜÓäýÛáØìËÎàìÉèÁöÒÄÃêãÅØ", "747413728711457843146848737038131103111518674763212440575401714508836686638165731184560552525061586126365788468022782565467710661434142676823176351252822742317600474637521438703218165037522505858362743845", "अ䧆𨃩𤸖𦬍𪪇𨅎ﲌ鵊祁撝𗃝𨎧ʏ𥎔𫫧刵𣈜찇抿罧줚𦜖Ⳇ𗘿𦧽𡝇𥨓년𝜈㲟𪢅塢ਤ𦧹豿댊𑚦䴞뺇𪕫𬓈턛ｼ𬏇禧𪐲𪍷𥶟쾮𗀉𥈂뗅𢏝𫶵䫓𦨲랷ラ𤉛𨮟蕿𦠳냹𤼩𪴘𧸙𦑔𬥻𠇚Ꭽ蛱솀ﰼ㳇", "='!#?$^&:*;@).-_$--$!\"|}<:}%$!&=@|(`|:_;\"<($_+)</}$!'*,$^(`/{!;\\=-}\\&|,_+_!\"<)>]>}<:%\"&+\\|.()$#|@$><]", "<acronym>HBekWXeMlUyuMEPQAwTSA</acronym>"]›.  [/home/dlezz/.rbenv/versions/2.4.3/lib/ruby/gems/2.4.0/gems/shoulda-context-1.2.2/lib/shoulda/context/context.rb:346]:
After setting :name to ‹["r",
"RishDevHIfPeetnFVVjIgdcSAOmUrsbTWofYHdQsnrkinzIIAcBDHOGxGmglbjRJEbgbhNblSQqRiUoCaFaonNMlzUMxcququDXtDXCbWYqrFPhdCYUPkqNmJbiAhnjcJALauQOdPwPzydJdwJCanmYJObXagHUyzqqsbNAOVTBFUvoKJLIhlVBmgdrnrBWXfnFrYaJWYNXXKIhDBzLiRfkldtEiHCFmWEqsWCWyUCnPcNtItgGeocUowKybXLF",
"LsLDQwyHkjmgwdCJkmSPJggqWhEajqBrJdckYyMgpEcjoQXSkupztkQGqRslfbnhbPGnCkhOHrNqSOCRSbfMVrrNNvyJqaBtWfJeShblgNVEsByBjkrJIxCOkydnjdpcAXSoRYJXmjAuhBHmXTeUzxJvLcLPfamxVD",
"AcIhnIrngT2Lp4xFQj8KTFosAN8EEmdzqoTKtgCvbHvazMmIw2T8P6OLZx8LCHQshhzaBtXD3c7HEFWpdgXbFpnNp8SZwWBQCQnYW7LUQnsORYZqogANZU5hdBrAbxRLQ1GY2bgWhlK34iS4iQbQnWHZo6E5YKoynl0iiZKLy3C411G4zfKTGo4AlSpuIcHs0WMhkQympBm1g4fhXUWeqLcrORecS2Lt8xcsa6dPuofPbaLrVtBkpCeZCtf3jS",
"樵",
"ШҟѠбБҤЖӫЫѱҦЖХЩвҭҙѴєѤӋѲќѨӬѓѺнӜрѤѱӲЗлӂІӵѵѡУ҈әҏӇйбќрРӏѠЕЗѲѢҐҘвцңШЮӗӡѫЄҨӫЫиӳтӋѝӹЛѶӁҚАӅҬѪӆҩӉзҝҎӹшӠёЖаѸӀЪыҧӴӪҼҩґГЁӑѶкӚӤЪӘѢӅЩӣмӻуЧњ҃ӫҞҬѣѪѼҲЍѾӘӻѭРҔҜѺӯӇѽҁһБдҼ҂ҸЌЍӑъѐӯўѽѬңӓғЇӌѮӂӉҔѤҋѐоӂѸҷфң҄ӞҚљВѫҞҔшЏКѮӕѴҖҽЉфӮРШЮӔґӓҗ҉ҾөӬЦӤѡҢв҄ӵҖыӪѹѦҖфӓҒҮҘӦыҾӝЄОЭ҄бҍ҇лиѾЅѐӁЄхбнҏқ",
"ÁàæÆÎýÂîÐðÎàõõãïÂéíñÛÁÍÖÇÇñÏÛÎâËúäÕòñÒâöðÎÕúðêÃÓòäðÌîÇÀÓãõÉÀÞäöôÕÏÃÕÖßýÔçãéßÒêÇÙÃÝóÜÈËÍßëÕÛüÒÎÌâÕôßýíËèÈÇÁðåÄüÑáöèÎÏýÃûÂþûÍÁþÊàÂäÜÓÜÓäýÛáØìËÎàìÉèÁöÒÄÃêãÅØ",
"747413728711457843146848737038131103111518674763212440575401714508836686638165731184560552525061586126365788468022782565467710661434142676823176351252822742317600474637521438703218165037522505858362743845",
"अ䧆𨃩𤸖𦬍𪪇𨅎ﲌ鵊祁撝𗃝𨎧ʏ𥎔𫫧刵𣈜찇抿罧줚𦜖Ⳇ𗘿𦧽𡝇𥨓년𝜈㲟𪢅塢ਤ𦧹豿댊𑚦䴞뺇𪕫𬓈턛ｼ𬏇禧𪐲𪍷𥶟쾮𗀉𥈂뗅𢏝𫶵䫓𦨲랷ラ𤉛𨮟蕿𦠳냹𤼩𪴘𧸙𦑔𬥻𠇚Ꭽ蛱솀ﰼ㳇",
"='!#?$^&:*;*]<*>@;>=%)>#:.\\]):|==!-^|**<)\"{[\"@{-:|_|*#+\");)(+{-?,'#!.@).-_$--$!\"|}<:}%$!&=@|(`|:_;\"<($_+)</}$!'*,$^(`/{!;\\=-}\\&|,_+_!\"<)>]>}<:%\"&+\\|.()$#|@$><]",
"<acronym>HBekWXeMlUyuMEPQAwTSA</acronym>"]› -- which was read back as
‹"[\"r\",
\"RishDevHIfPeetnFVVjIgdcSAOmUrsbTWofYHdQsnrkinzIIAcBDHOGxGmglbjRJEbgbhNblSQqRiUoCaFaonNMlzUMxcququDXtDXCbWYqrFPhdCYUPkqNmJbiAhnjcJALauQOdPwPzydJdwJCanmYJObXagHUyzqqsbNAOVTBFUvoKJLIhlVBmgdrnrBWXfnFrYaJWYNXXKIhDBzLiRfkldtEiHCFmWEqsWCWyUCnPcNtItgGeocUowKybXLF\",
\"LsLDQwyHkjmgwdCJkmSPJggqWhEajqBrJdckYyMgpEcjoQXSkupztkQGqRslfbnhbPGnCkhOHrNqSOCRSbfMVrrNNvyJqaBtWfJeShblgNVEsByBjkrJIxCOkydnjdpcAXSoRYJXmjAuhBHmXTeUzxJvLcLPfamxVD\",
\"AcIhnIrngT2Lp4xFQj8KTFosAN8EEmdzqoTKtgCvbHvazMmIw2T8P6OLZx8LCHQshhzaBtXD3c7HEFWpdgXbFpnNp8SZwWBQCQnYW7LUQnsORYZqogANZU5hdBrAbxRLQ1GY2bgWhlK34iS4iQbQnWHZo6E5YKoynl0iiZKLy3C411G4zfKTGo4AlSpuIcHs0WMhkQympBm1g4fhXUWeqLcrORecS2Lt8xcsa6dPuofPbaLrVtBkpCeZCtf3jS\",
\"樵\",
\"ШҟѠбБҤЖӫЫѱҦЖХЩвҭҙѴєѤӋѲќѨӬѓѺнӜрѤѱӲЗлӂІӵѵѡУ҈әҏӇйбќрРӏѠЕЗѲѢҐҘвцңШЮӗӡѫЄҨӫЫиӳтӋѝӹЛѶӁҚАӅҬѪӆҩӉзҝҎӹшӠёЖаѸӀЪыҧӴӪҼҩґГЁӑѶкӚӤЪӘѢӅЩӣмӻуЧњ҃ӫҞҬѣѪѼҲЍѾӘӻѭРҔҜѺӯӇѽҁһБдҼ҂ҸЌЍӑъѐӯўѽѬңӓғЇӌѮӂӉҔѤҋѐоӂѸҷфң҄ӞҚљВѫҞҔшЏКѮӕѴҖҽЉфӮРШЮӔґӓҗ҉ҾөӬЦӤѡҢв҄ӵҖыӪѹѦҖфӓҒҮҘӦыҾӝЄОЭ҄бҍ҇лиѾЅѐӁЄхбнҏқ\",
\"ÁàæÆÎýÂîÐðÎàõõãïÂéíñÛÁÍÖÇÇñÏÛÎâËúäÕòñÒâöðÎÕúðêÃÓòäðÌîÇÀÓãõÉÀÞäöôÕÏÃÕÖßýÔçãéßÒêÇÙÃÝóÜÈËÍßëÕÛüÒÎÌâÕôßýíËèÈÇÁðåÄüÑáöèÎÏýÃûÂþûÍÁþÊàÂäÜÓÜÓäýÛáØìËÎàìÉèÁöÒÄÃêãÅØ\",
\"747413728711457843146848737038131103111518674763212440575401714508836686638165731184560552525061586126365788468022782565467710661434142676823176351252822742317600474637521438703218165037522505858362743845\",
\"अ䧆𨃩𤸖𦬍𪪇𨅎ﲌ鵊祁撝𗃝𨎧ʏ𥎔𫫧刵𣈜찇抿罧줚𦜖Ⳇ𗘿𦧽𡝇𥨓년𝜈㲟𪢅塢ਤ𦧹豿댊𑚦䴞뺇𪕫𬓈턛ｼ𬏇禧𪐲𪍷𥶟쾮𗀉𥈂뗅𢏝𫶵䫓𦨲랷ラ𤉛𨮟蕿𦠳냹𤼩𪴘𧸙𦑔𬥻𠇚Ꭽ蛱솀ﰼ㳇\",
\"='!#?$^&:*;*]<*>@;>=%)>#:.\\\\]):|==!-^|**<)\\\"{[\\\"@{-:|_|*#+\\\");)(+{-?,'#!.@).-_$--$!\\\"|}<:}%$!&=@|(`|:_;\\\"<($_+)</}$!'*,$^(`/{!;\\\\=-}\\\\&|,_+_!\\\"<)>]>}<:%\\\"&+\\\\|.()$#|@$><]\",
\"<acronym>HBekWXeMlUyuMEPQAwTSA</acronym>\"]"› -- the matcher expected
the Architecture to be valid, but it was invalid instead, producing
these validation errors:

* name: ["is too long (maximum is 255 characters)", "can't contain
  spaces."]

As indicated in the message above, :name seems to be changing certain
values as they are set, and this could have something to do with why
this test is failing. If you've overridden the writer method for this
attribute, then you may need to change it to make this test pass, or do
something else entirely.

```

the should not is working as it fails by default

test/models/architecture_test.rb
ldjebran: we are going to loose the failure message as assert_valid does not accept message, and we need that messages for debugging as we have a loop over different type and length of strings.

test/models/architecture_test.rb
ldjebran: same here as assert_valid,  we are going to loose the failure message as refute_valid does not accept message and  we need that messages for debugging as we have a loop over different type and length of strings.

test/models/architecture_test.rb
timogoebel: `assert_valid` has a generic failure message that should be as good as the one provided imho.

ldjebran:
@timogoebel many thanks for your review, please consider my reply to your comments.

test/models/architecture_test.rb
timogoebel: Have you tried with the splat operator?

```
should allow_values(*valid_name_list).for(:name)
```

ldjebran: yes it's working like this but the test log is very ugly (same for the negative test):

_____________________________
ArchitectureTest#test_: Architecture should allow :name to be ‹"G"›, ‹"uPAGfggCNVJprbewuhyrvXXvhgKpBAAmUBgHqLjeOeEUANtcrVKKwHKqSthpVizibqXguaOjdspETTpdMimQTbzyqlJUhFPKPTROdXlHeTnRqzrlnsiIwhykdyKdxAVvAvDcIUWWHVwzTfviTRsfJNfJtKUPDuQxKuPEeiaUmsHkhAmAqOvVGgQKxPaJfjzCKfeItrGxFXwFbGBWXGUaCUSuVwtnPCvthtAXJgJVubqVkDEXsoFhuDhGyxiMGUf"›, ‹"ifELpSIMcfstiauoOQIiEnPmNQdsdmeQNNNmRfWwAEGahGoMSjVrXAKGCEmHwOxHYkyuavODphrJtzRUlNokUWclEbJeiOoQXiBvkoWkksohcxgrMghjQAKOdqGLzeQkcirYOxfxARFXgCjbtAOxXUPAfBnVnbNEXkTSmHrHmkABPYiAMGilaSqKPtjUaiqIXUSMlHIOrKPifGYGHRQIIPiLfKTXIdFKjPjMRJlWTKJqlAGFcyVMSCDDJhbW"›, ‹"PBHrjpG0JjpevC1yTyxdWVJMWaathwJICwWic2ZrdU1QIFtYsTFIfValQ2dsfkwscaz6TMMnkhCIxFbCFX1aWdjqxgBHnBMjKMpV8x2LiRC74ybjTZBwyFX6vvDsLredRscxdRvnVycx3iCZwmQLcbpAedUnNmEwwtE7G2ViACRjdXn7o64o"›, ‹"浧礟成轃捥派蓮郯鳌郉薵駟懱呉裺塒鱣籱棐躝奿谺夵羮窴飛壀憇寘墝養誜馝紂肕斬鹙縆梵榥舗戏诂鱕阝謈鑹甒拡痃拱灩垢磆瀌奧饣熣硎葮丶銻癵獣愾辽锿垺马攟嶘糔篎踱鞁襤嘵僯嫙抧鄖澧姏承献嫬屷知奧倆悏鰲搄溸事躛茭蟫镇筹娥害某璥囇其垊驦諶綠奤荎濣哦瀲輴湊唐阋韞蘎鮈碤逶燻廞讙缿葿呄疇赌秀倨莹嚛訲縑驜躼正梬譤"›, ‹"ђӼчѭѥҌдбҠѤмаќҮҹѱжӑИЯнУіщӥҿҒдѩЂТӘӂѓҏӢҮӁқЩҤҘӰѻСйӇбјѕѹӻжҊґӫѡҽӰӺеҞӥѾѧӿӳѻОѷЎӒҙҚҰѡЅ҉іқћҔАҐӿУОҴЊҎӫҗОҳҥѲҫағѡеѨѣӃЁоҩӔЧѠӌӭѠјӛҷѩПӅѬӘфѲЈѓѓҘӥҴҟЇҧҳйлұһӵІҢһѿѵпҩӶЄѫмѢЦѸӔцшҊЁӄѩѹӋӈӟһєФӑТӐҋҩВтӼсѕ҆ѠҴҸґӱФӋӺӧЩѲбӜ"›, ‹"ûòýøÅâûíèØÈÅñêíÏÈÕÕÂÚìØáÒâøËÐÇúÃÞÍãùÞøÕÉÄïØÆÄõãðûÓÌèÔÉÚÅôûÀÖñä"›, ‹"783302615472464135124265123440770651611041080748554113781253454738004138230"›, ‹"𒀜𠆤𢑇𥋪𠟈𔔇𢋎𢿮𥚃𥨜펱뛹ꮸ믤𡐞佒𦌝궫폣廞𢞩𒄇𔗒𡔌䆆亶䇒䊷곷𠪮挣遧𗼳䒈폽竏𩟀銋ᇛ戂𢞁𘄗䉎𢚠"›, ‹"&(%$#<[<\",|#,\"%=["›, or ‹"<center>NvntuRbntwdrJxtwgfDLxCkMcCSpqizFNjPkhnfRSCqCALQAnebsijhqOhsyVwpFyBUIPlDaUbaDVsfJGISGuYDtirMilFHx</center>"›.  = 0.01 s = .
_____________________________
Please confirm and will update this way.

timogoebel: @dLobatog: May I ask for a second opinion here?

ares: not that someone asked me, but I think shoulda_matchers has small benefit of performance, we don't enforce to use it elsewhere, but if you already have the code, I think it's preferred variant

the main difference is that it does not run the built-in validator, it only verified it would be run, which is sufficient, since built-in validators are provided by rails and they are well tested there

the log message is definitely not nice but it's only displayed in case of failure and at least we get the full value, which is useful for debugging

so bottom line, I would not enforce using it, but especially in case where we iterate over multiple values, I think it's better to use them

test/models/architecture_test.rb
ares: I think the point was we can specify custom message for assert_valid which I've seen used elsewhere, if I had to chose between the two, I'd probably vote for shoulda_matches because of efficiency

test/models/architecture_test.rb
timogoebel: @ares: I only pinged @dLobatog because he always makes me use shoulda-matchers. Your opinion is also very welcome ;-)

ldjebran:
updated, @timogoebel @ares many thanks for the commendations.

test/models/architecture_test.rb
```diff
@@ -74,4 +75,29 @@ class ArchitectureTest < ActiveSupport::TestCase
     architecture = Architecture.new :name => "../../etc/shadow"
     assert_equal "etc-shadow", architecture.bootfilename_efi
   end
+
+  test "update with multi names" do
+    architecture = FactoryBot.create(:architecture)
+    valid_name_list.each do |new_name|
+      architecture.name =  new_name
```
ares: two spacebard after `=`, should be just one :-)

test/models/architecture_test.rb
```diff
@@ -74,4 +75,29 @@ class ArchitectureTest < ActiveSupport::TestCase
...
+  test "should not update with invalid names" do
+    architecture = Architecture.first
+    invalid_name_list.each do |name|
+      architecture.name = name
+      refute_valid architecture
+      assert_includes architecture.errors.keys, :name
+    end
+  end
+
+  test "should destroy architecture" do
```
ares, line 97: I wonder what this test is actually testing? the deleting active record model is a standard rails operation, we only verify that `.delete` method works. The thing that could be tested is this https://github.com/theforeman/foreman/blob/develop/app/models/architecture.rb#L7 which ensures it can't be deleted if there's at least one host or hostgroup assigned to this architecture, but this architecture is not assigned to anything. But this is already being tested on [line 18](https://github.com/ldjebran/foreman/blob/10f5b9354f7328f8efb0b462e50d9752678dfacb/test/models/architecture_test.rb#L18)

test/test_helper.rb
```diff
@@ -32,6 +32,27 @@
   ActiveRecord::Migration.execute "SET CONSTRAINTS ALL DEFERRED;"
 end
 
+# List of valid record name field.
+def valid_name_list
```
ares, line 36: I think this is generally a good idea, could you extract to separate PR with also changing existing tests to use it? Speaking about https://github.com/theforeman/foreman/blob/develop/test/models/organization_test.rb#L8-L28 unless there's a reason we can't use it there.

ares (changes requested):
I did a bit more thorough read through and found few things. Please let me know if there's something that needs more clarification. Generally thumbs up :-)

test/models/architecture_test.rb
ekohl: There could be a bug in that logic that prevents deletion in general.

ares: because of that callback? ok, we can keep it here, it doesn't take much time I guess, @ldjebran please ignore this commit

test/test_helper.rb
ldjebran: @ares this is a more general function that should be valid for the most entities that's why I putted here,  unlike the org_name_list that is valid only for organization as it has it's own limits. 
This way other tests can use this general functions

ldjebran:
@ares @timogoebel updated and rebased

mmoll:
ok to test

test/test_helper.rb
ares: Ok, makes sense

timogoebel (approved):
LGTM.

timogoebel:
Test failures are unrelated.

mmoll:
@ares this should be good to go? :)

ares (approved):
:+1:

ares:
Thanks @ldjebran for another great contribution to our test suite, thanks @timogoebel and @mmoll for helping getting this in! Merging.

PR #5336, opened 2018-03-19
[WIP] Fixes #22109, #22388 - add graphql as API

dariuszb-iRonin:
I'm working with @timogoebel on graphql API and finally we are able to show our progress. It's still WIP, but it's fully workable :)

**Instructions:**

To play with this, you need a graphql client.
Fortunately, this PR ships one in develop. Just browse to `http://localhost:3000/graphiql`.
On OSX, you can install an electron based client: `brew cask install graphiql`

The API is mounted at `http://localhost:3000/api/graphql`

To test this, you need a JWT token. Use below mutation to sign in and get token:

```
mutation {
  signInUser(login: {username: "admin", password: "test"}) {
    token
    user_id
  }
}
```

This token needs to be set as an `Authorization` header.
This is not possible with the integrated graphiql client, but you can specify the user via a query parameter: `http://localhost:3000/graphiql?user_id=1`. FYI: In mention electron based client you are able to set headers.

Then you can e.g. list subnet names:

```
query {
  subnets {
    edges {
      node {
        name
      }
    }
  }
}
```

or show the name of a specific subnet:

```
query {
  subnet(id: 1) {
    name
  }
}
```

There is implemented pagination(cursor based pagination - http://graphql-ruby.org/relay/connections.html). You can read more about eg in this article: https://dev-blog.apollodata.com/understanding-pagination-rest-graphql-and-relay-b10f835549e7

To test it you need to pass `first`/`last` and `after`/`before` parameters to given query.
Also there is need to pass additional `pageInfo` parameters to get some informations about pagination.

```
query {
  subnets(first: 10, after: #{cursor}) {
    pageInfo {
      startCursor
      endCursor
      hasNextPage
      hasPreviousPage
    }
    edges {
      node {
        id
      }
    }
  }
}
```

You can use it also like simple `page` `per_page` pagination. For example to get 6th page with 10 elements per page you need to pass:

```
query {
  subnets(first: 60, last: 10) {
    pageInfo {
      startCursor
      endCursor
      hasNextPage
      hasPreviousPage
    }
    edges {
      node {
        id
      }
    }
  }
}
```

We also added  `rack-cors`.
To allow CORS set `FOREMAN_GRAPHQL_CORS_DOMAINS` env variable.
To support multiple domains separate them with comma(`,`).

```
export FOREMAN_GRAPHQL_CORS_DOMAINS="http://localhost:3001,http://localhost:3002"
```

We can consider moving that configuration to `settings.yml`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -53,6 +53,11 @@ gem 'sshkey', '~> 1.9'
 gem 'dynflow', '>= 0.8.29', '< 1.0.0'
 gem 'daemons'
 gem 'get_process_mem'
+gem 'graphql', '~> 1.7.8'
+gem 'graphql-activerecord'
+gem 'graphql-batch'
+gem 'jwt', '~> 2.1.0'
+gem 'rack-cors', '~> 1.0.2'
 
 Dir["#{File.dirname(FOREMAN_GEMFILE)}/bundler.d/*.rb"].each do |bundle|
   self.instance_eval(Bundler.read_file(bundle))
--- a/app/controllers/api/graphql_controller.rb
+++ b/app/controllers/api/graphql_controller.rb
@@ -1,0 +1,78 @@
+module Api
+  class GraphqlController < ActionController::Base
+    force_ssl :if => :require_ssl?
+    rescue_from Exception, :with => :generic_exception if Rails.env.production?
+
+    before_action :set_current_user
+    around_action :set_timezone
+
+    def execute
+      result = ForemanGraphqlSchema.execute(params[:query],
+                                            variables: variables,
+                                            context: {
+                                              current_user: User.current
+                                            })
+      render json: result
+    end
+
+    private
+
+    def set_current_user
+      @current_user = begin
+        payload = jwt_token.decode || {}
+        user_id = payload['user_id']
+        User.unscoped.find_by(id: user_id) if user_id
+      rescue JWT::DecodeError
+        nil
+      end
+      User.current = @current_user
+    end
+
+    def jwt_token
+      @jwt_token ||= begin
+        token = request.headers.fetch('Authorization', '')
+        JwtToken.new(token)
+      end
+    end
+
+    def variables
+      ensure_hash(params[:variables])
+    end
+
+    def ensure_hash(ambiguous_param)
+      case ambiguous_param
+      when String
+        if ambiguous_param.present?
+          ensure_hash(JSON.parse(ambiguous_param))
+        else
+          {}
+        end
+      when Hash, ActionController::Parameters
+        ambiguous_param
+      when nil
+        {}
+      else
+        raise ArgumentError, "Unexpected parameter: #{ambiguous_param}"
+      end
+    end
+
+    def generic_exception(exception)
+      Foreman::Logging.exception("Action failed", exception)
+      render json: "{ 'error': 500 }", :status => :internal_server_error
+    end
+
+    def set_timezone
+      default_timezone = Time.zone
+      client_timezone  = User.current.try(:timezone)
+      Time.zone = client_timezone if client_timezone.present?
+      yield
+    ensure
+      # Reset timezone for the next thread
+      Time.zone = default_timezone
+    end
+
+    def require_ssl?
+      SETTINGS[:require_ssl]
+    end
+  end
+end
--- a/app/graphql/mutations/sign_in_user.rb
+++ b/app/graphql/mutations/sign_in_user.rb
@@ -1,0 +1,22 @@
+class Mutations::SignInUser < GraphQL::Function
+  argument :login, Types::AuthProviderLoginInput
+
+  type do
+    name 'SIGN_IN_USER_PAYLOAD'
+
+    field :user_id, types.ID
+    field :token, types.String
+  end
+
+  def call(obj, args, ctx)
+    login = args[:login] || {}
+    username = login[:username]
+    password = login[:password]
+    return unless username && password
+
+    user = User.try_to_login(username, password)
+    return unless user
+
+    OpenStruct.new(user_id: user.id, token: user.jwt_token!)
+  end
+end
--- a/app/graphql/queries/plural_field.rb
+++ b/app/graphql/queries/plural_field.rb
@@ -1,0 +1,19 @@
+module Queries
+  class PluralField < GraphQL::Function
+    attr_reader :type
+
+    argument(:search, types.String, 'Search query')
+    argument(:order, types.String, 'Search query')
+
+    def initialize(model_class:, type:)
+      @model_class = model_class
+      @type = type
+    end
+
+    def call(_, args, ctx)
+      params = args.to_h.slice('search', 'order').symbolize_keys
+      Queries::AuthorizedModelQuery.new(model_class: @model_class, user: ctx[:current_user])
+                                   .results(params)
+    end
+  end
+end
--- a/app/graphql/record_loader.rb
+++ b/app/graphql/record_loader.rb
@@ -1,0 +1,10 @@
+class RecordLoader < GraphQL::Batch::Loader
+  def initialize(model)
+    @model = model
+  end
+
+  def perform(ids)
+    @model.where(id: ids).each { |record| fulfill(record.id, record) }
+    ids.each { |id| fulfill(id, nil) unless fulfilled?(id) }
+  end
+end
--- a/app/graphql/types/architecture_type.rb
+++ b/app/graphql/types/architecture_type.rb
@@ -1,0 +1,9 @@
+Types::ArchitectureType = GraphQL::ObjectType.define do
+  name 'Architecture'
+  description 'An Architecture'
+
+  backed_by_model :architecture do
+    attr :id
+    attr :name
+  end
+end
--- a/app/graphql/types/environment_type.rb
+++ b/app/graphql/types/environment_type.rb
@@ -1,0 +1,13 @@
+Types::EnvironmentType = GraphQL::ObjectType.define do
+  name 'Environment'
+  description 'An Environment'
+
+  backed_by_model :environment do
+    attr :id
+    attr :name
+  end
+
+  connection :puppetclasses, Types::PuppetclassType.connection_type do
+    resolve(proc { |obj| AssociationLoader.for(obj.class, :puppetclasses).load(obj) })
+  end
+end
--- a/app/graphql/types/vm_attrs_type.rb
+++ b/app/graphql/types/vm_attrs_type.rb
@@ -1,0 +1,10 @@
+Types::VmAttrsType = GraphQL::ObjectType.define do
+  name 'VmAttr'
+  description 'A VmAttr'
+
+  field :cpus, types.String, hash_key: 'cpus'
+  field :memory, types.String, hash_key: 'memory'
+  field :volumesAttributes, Types::JsonType, resolve: (proc do |obj|
+    obj['volumes_attributes']
+  end)
+end
--- a/app/models/jwt_secret.rb
+++ b/app/models/jwt_secret.rb
@@ -1,0 +1,25 @@
+class JwtSecret < ActiveRecord::Base
+  include Encryptable
+
+  encrypts :token
+
+  belongs_to :user, inverse_of: :jwt_secret
+
+  validates :token, uniqueness: true
+  validates :user, presence: true
+
+  before_save :generate_token, on: :create, prepend: true, :unless => Proc.new{|j| j.token.present?}
+
+  def self.for_user(user_or_user_id)
+    find_by(user: user_or_user_id)
+  end
+
+  private
+
+  def generate_token
+    loop do
+      self.token = SecureRandom.base64
+      break unless JwtSecret.find_by(token: token)
+    end
+  end
+end
--- a/app/models/jwt_token.rb
+++ b/app/models/jwt_token.rb
@@ -1,0 +1,60 @@
+require 'jwt'
+require 'base64'
+
+class JwtToken < Struct.new(:token)
+  class << self
+    def encode(user, secret)
+      payload = prepare_payload(user, secret)
+      new JWT.encode(payload, secret)
+    end
+
+    private
+
+    def prepare_payload(user, secret)
+      iat = Time.now.to_i
+      jti_raw = [secret, iat].join(':').to_s
+      jti = Digest::MD5.hexdigest(jti_raw)
+      { user_id: user.id, iat: iat, jti: jti }
+    end
+  end
+
+  def decode
+    return nil if token.blank?
+    return nil if secret.blank?
+
+    payload = JWT.decode(token, secret.token)
+    payload.first
+  end
+
+  def to_s
+    token
+  end
+
+  private
+
+  def secret
+    return @secret if defined? @secret
+    @secret = JwtSecret.for_user(user_id) if user_id
+  end
+
+  def user_id
+    return @user_id if defined? @user_id
+    @user_id = decoded_payload['user_id']
+  end
+
+  def decoded_payload
+    @decoded_payload ||= begin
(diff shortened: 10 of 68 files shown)


timogoebel:
This obviously looks good to me as I already reviewed most of it. :-)
Fist of all, I'd suggest to mark this API as experimental for now.

We didn't add support for extending this from plugins for now. That would be something for a follow up  PR imho.

---

To play with this, you need a graphql client.
Fortunately, this PR ships one in develop. Just browse to `http://localhost:3000/graphiql`.
On OSX, you can install an electron based client: `brew cask install graphiql`

The API is mounted at `http://localhost:3000/api/graphql`

To test this, you need a JWT token.

```graphql
mutation {
  signInUser(login: {username: "admin", password: "test"}) {
    token
    user_id
  }
}
```

This token needs to be set as an `Authorization` header.
This is not possible with the integrated graphiql client, but you can specify the user via a query parameter: `http://localhost:3000/graphiql?user_id=1`

Then you can e.g. list subnet names:

```graphql
query {
  subnets(search: "") {
    edges {
      node {
        name
      }
    }
  }
}
```

or show the name of a specific subnet:

```graphql
{
  subnet(id: 1) {
    name
  }
}
```

timogoebel:
ok to test

config/initializers/graphql.rb
```diff
@@ -0,0 +1,30 @@
+require 'rack'
+require 'rack/cors'
+require 'graphql'
+require 'graphql/batch'
+require 'graphql/activerecord'
+
+Foreman::Application.configure do |app|
+  initializer 'configure_cors', :before=> :build_middleware_stack do |app|
```
timogoebel, line 8: Nit: Space before `=>` ;-)

app/controllers/api/graphql_controller.rb
```diff
@@ -0,0 +1,78 @@
+module Api
+  class GraphqlController < ActionController::Base
+    force_ssl :if => :require_ssl?
+    rescue_from Exception, :with => :generic_exception if Rails.env.production?
```
chrisarcand, line 4: If you raise exceptions in application logic (and this is Ruby, you surely do), this might not be a good idea. GraphQL APIs are generally expected to always return 200 and instead populate an `errors` attribute. I would try and leave as much deviation of errors codes in the actual transport logic (HTTP) and abide by the convention of responding 200 on expected errors.

chrisarcand (commented):
I was asked to glance over this and give some thoughts! The stuff I'll say is very high level, given my familiarity with the project. There's some specific stuff I'm a little confused about (What exactly is the purpose of `PluralField`? A place to consolidate on things like the `order` argument across all those fields?), but I'll leave that for someone else.

* I like that you've done Relay's [connection spec](https://facebook.github.io/relay/graphql/connections.htm#) right away, good call.
* I would suggest you also go through with the [global ID spec](https://facebook.github.io/relay/graphql/objectidentification.htm). Global IDs allow arbitrary fetching of any particular node in the graph for Relay clients to do some really neat client-side caching techniques.
* In line with my previous point, (and I know this will probably be a deeply unpopular opinion) I would suggest you *deprecate database primary keys*. Set the intentionally user-opaque Relay IDs as `id` and either completely remove the RDMS ID *or* set it as `database_id` and consider leaving it deprecated. The reason for this is that GraphQL *should not actually ever need* database IDs (it's encoded in the Relay ID) and the user really shouldn't either. In practice, that's easier said than done as you might have legacy reasons (relying on the REST API, whatever) for wanting them. GitHub does exactly this, which you can read more about [here](https://platform.github.community/t/what-is-the-reason-databaseid-field-is-deprecated/976). 
* I would be very concerned with your use of `graphql-activerecord` for two reasons: 

1) The README of that gem outlines a lot of what I'd say, and clearly leaves a warning:

> This gem may evolve into something better, or we may eventually deprecate it. But as of right now, I can't recommend that you build any major projects on top of it, since its future is a bit uncertain. I apologize if that makes more work for you :worried: but I want to be honest about the state of the project.

2) I would be very careful of dynamically exposing fields based on underlying ActiveRecord models. This is a thing that's *very* popular right now but I'm very wary of it. By leaving the interface of your API (and what you have to maintain!) dictated by the underlying models of your application, you're making the API implicit where explicitness is a good thing, and you're tightly coupling the graph of your API to (often) legacy RDMS tables. I have more thoughts on this but will just summarize as saying 'Try being explicit first - if it *really* is a pain, then consider alternatives'. As this is the beginning of this API, I see no reason not to be explicit :)

* Since this is basically a prototype, consider using graphql-ruby's 1.8 class based definitions (currently still in prerelease). It's the future of the gem so you might as well get started now.

ohadlevy (commented):
I gave this patch a try and it worked well! thank you.

I was unsure if node / edge are best practices in graphql or if its a convention you defined?

overall, my main concern is the size of the pr, would  it make sense to break it down ? e.g. one commit for JWT another for cors, and than another one for the infrastructure around graphql? (controller, permissions, basic layouts?) I'm just concerned that merging it as one item would be hard to review and slow to move.

IMHO I think it makes sense to be in core if we want hammer/UI to start looking into consuming it.

Gemfile
```diff
@@ -53,6 +53,11 @@ gem 'sshkey', '~> 1.9'
 gem 'dynflow', '>= 0.8.29', '< 1.0.0'
 gem 'daemons'
 gem 'get_process_mem'
+gem 'graphql', '~> 1.7.8'
```
ohadlevy, line 56: please move it to a bundler group, I could see usage cases where we would like to start the server without api or graphql support.

app/controllers/api/graphql_controller.rb
```diff
@@ -0,0 +1,78 @@
+module Api
+  class GraphqlController < ActionController::Base
```
ohadlevy, line 2: do we really have nothing in common from the api base controller?

app/controllers/api/graphql_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+      rescue JWT::DecodeError
+        nil
+      end
+      User.current = @current_user
+    end
+
+    def jwt_token
+      @jwt_token ||= begin
+        token = request.headers.fetch('Authorization', '')
+        JwtToken.new(token)
```
ohadlevy, line 34: does it make sense to handle decoding errors here as well? I think that long term we should want to decouple setting the current user from the jwt token? (I would assume we would like to allow other authentication methods?)

app/controllers/api/graphql_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+    def ensure_hash(ambiguous_param)
+      case ambiguous_param
+      when String
+        if ambiguous_param.present?
+          ensure_hash(JSON.parse(ambiguous_param))
+        else
+          {}
+        end
+      when Hash, ActionController::Parameters
+        ambiguous_param
```
ohadlevy, line 51: if its AC::Parameters shouldn't you force it into a hash?

app/controllers/api/graphql_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+        token = request.headers.fetch('Authorization', '')
+        JwtToken.new(token)
+      end
+    end
+
+    def variables
+      ensure_hash(params[:variables])
+    end
+
+    def ensure_hash(ambiguous_param)
```
ohadlevy, line 42: can you provide a bit more context why this is needed? is it the way graphql sends its data or more a generic problem? if its generic I would suggest to move it away from this class?

app/controllers/api/graphql_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+      when nil
+        {}
+      else
+        raise ArgumentError, "Unexpected parameter: #{ambiguous_param}"
+      end
+    end
+
+    def generic_exception(exception)
+      Foreman::Logging.exception("Action failed", exception)
+      render json: "{ 'error': 500 }", :status => :internal_server_error
```
ohadlevy, line 61: as @chrisarcand mentioned earlier, we should always return 200? or in other words, how should we do error handling ? I can see a few scenarios:

1. auth / permissions
2. server error (500)

in both cases, I could imagine that some query / mutation are allowed while others do not (e.g. I can list something, but not something else, or I can change one object but not another) and in that case probably should return 200?

app/controllers/api/graphql_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+    end
+
+    def set_timezone
+      default_timezone = Time.zone
```
ohadlevy, line 65: I would expect this to reuse https://github.com/theforeman/foreman/blob/879f984fc173ce43a11d7d0cfbc2850db396d2a1/app/controllers/concerns/application_shared.rb#L11 ?

app/graphql/mutations/sign_in_user.rb
```diff
@@ -0,0 +1,22 @@
+class Mutations::SignInUser < GraphQL::Function
+  argument :login, Types::AuthProviderLoginInput
+
+  type do
+    name 'SIGN_IN_USER_PAYLOAD'
```
ohadlevy, line 5: should you freeze these strings?

app/graphql/mutations/sign_in_user.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    field :user_id, types.ID
+    field :token, types.String
```
ohadlevy, line 8: is there a limit of length etc? I wasn't able to find any good docs on the types?

app/controllers/api/graphql_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+      render json: result
+    end
+
+    private
+
+    def set_current_user
+      @current_user = begin
+        payload = jwt_token.decode || {}
+        user_id = payload['user_id']
+        User.unscoped.find_by(id: user_id) if user_id
```
ohadlevy, line 24: nitpick: maybe just use find(user_id) so it would raise an exception if it can't find the user?

app/graphql/queries/plural_field.rb
```diff
@@ -0,0 +1,19 @@
+module Queries
+  class PluralField < GraphQL::Function
+    attr_reader :type
+
+    argument(:search, types.String, 'Search query')
+    argument(:order, types.String, 'Search query')
```
ohadlevy, line 6: wrong description? maybe we can also add validation that its either DESC / ASC with field name that we support?

app/graphql/record_loader.rb
```diff
@@ -0,0 +1,10 @@
+class RecordLoader < GraphQL::Batch::Loader
+  def initialize(model)
+    @model = model
+  end
+
+  def perform(ids)
+    @model.where(id: ids).each { |record| fulfill(record.id, record) }
```
ohadlevy, line 7: just wondering in which context this is executing? reason I'm asking is that passing a very large array of ID's would be sub optimal vs an INNER query?

app/graphql/types/architecture_type.rb
```diff
@@ -0,0 +1,9 @@
+Types::ArchitectureType = GraphQL::ObjectType.define do
+  name 'Architecture'
+  description 'An Architecture'
+
+  backed_by_model :architecture do
+    attr :id
+    attr :name
+  end
```
ohadlevy, line 8: why not include timestamps ?

app/graphql/types/environment_type.rb
```diff
@@ -0,0 +1,13 @@
+Types::EnvironmentType = GraphQL::ObjectType.define do
+  name 'Environment'
+  description 'An Environment'
+
+  backed_by_model :environment do
+    attr :id
+    attr :name
+  end
+
+  connection :puppetclasses, Types::PuppetclassType.connection_type do
```
ohadlevy, line 10: can a plugin extend this?

app/graphql/types/vm_attrs_type.rb
```diff
@@ -0,0 +1,10 @@
+Types::VmAttrsType = GraphQL::ObjectType.define do
+  name 'VmAttr'
+  description 'A VmAttr'
+
+  field :cpus, types.String, hash_key: 'cpus'
```
ohadlevy, line 5: should this really be a string (api wize)?

app/graphql/types/vm_attrs_type.rb
```diff
@@ -0,0 +1,10 @@
...
+  field :memory, types.String, hash_key: 'memory'
```
ohadlevy, line 6: i would default to memory in kb (or bytes) but not a string imho.

app/models/jwt_secret.rb
```diff
@@ -0,0 +1,25 @@
...
+  include Encryptable
+
+  encrypts :token
+
+  belongs_to :user, inverse_of: :jwt_secret
+
+  validates :token, uniqueness: true
+  validates :user, presence: true
+
+  before_save :generate_token, on: :create, prepend: true, :unless => Proc.new{|j| j.token.present?}
```
ohadlevy, line 11: when does a token expires?

app/models/jwt_secret.rb
```diff
@@ -0,0 +1,25 @@
+class JwtSecret < ActiveRecord::Base
```
ohadlevy, line 1: whats the difference between user auth tokens and jwt? should we keep both?

app/models/jwt_token.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  private
+
+  def secret
+    return @secret if defined? @secret
+    @secret = JwtSecret.for_user(user_id) if user_id
+  end
+
+  def user_id
+    return @user_id if defined? @user_id
```
ohadlevy, line 41: can be written as:
```ruby
def user_id
  @user_id ||= decoded_payload['user_id']
end
```

ohadlevy (changes requested):
initial code review - thanks!

app/graphql/types/environment_type.rb
timogoebel: I'd prefer to take care of plugins in a separate issue / PR.

app/models/jwt_secret.rb
timogoebel: The `JWTSecret` is a token used to encryt/decrypt the JWT token. I think it's a good idea to keep this separate. `PersonalAccessToken` can be used like a password. I think, they aren't really related.

timogoebel:
> 'Try being explicit first - if it really is a pain, then consider alternatives

@chrisarcand: First of all, thanks for your comments. I'd prefer if we could implement most of the stuff (like relay global ids) in separate PRs.
I really think the `graphql-activerecord` gem adds a lot of value to this PR. The reason is pretty simple: It's implicit. I can see, why you prefer an explicit approach but if you look through the history of PRs in Foreman you'll find a lot PRs that fix API documentation (apipie) because people got the explicitness wrong. If you're explicit, a human gets things wrong. I guess that's why people like rails so much. And just mapping types to fields is something that can easily be done by a computer. The good thing is: If we ever get to have problems with the gem, we can still either replace it or just go for the explicit approach after all. That's why I'd prefer to go with `graphql-activerecord` for now.

app/controllers/api/graphql_controller.rb
leewaa: I don't think rescuing from `Exception` is a good idea. This will catch **all** exceptions and `generic_exception` will respond with `500` which isn't correct for all situations as @ohadlevy mentioned in https://github.com/theforeman/foreman/pull/5336/files#r176930235 

Also related here is https://github.com/theforeman/foreman/pull/5336/files#r176930474 as we could raise a RecordNotFound exception here with `find` which we can hook into and return the correct graphql error object and an accompanying `200` status (corresponding to what @chrisarcand mentioned above).

There must be a better solution for catching unwanted exceptions.

leewaa (changes requested):

PR #5337, opened 2018-03-19
Fixes #22269 - Added ordering for OS template listing by name

gnurag:
This PR fixes inconsistent ordering of Operatingsystem templates listed on the templates tab. Templates would be listed alphabetically (downcase).

### DIFF
--- a/app/views/operatingsystems/_template_defaults.html.erb
+++ b/app/views/operatingsystems/_template_defaults.html.erb
@@ -12,7 +12,7 @@
                           link_to("templates", provisioning_templates_path)).html_safe) %>
     <% end %>
 
-    <%= f.fields_for :os_default_templates do |builder| %>
+    <%= f.fields_for :os_default_templates, f.object.os_default_templates.sort_by{ |t| t.template_kind.to_s.downcase } do |builder| %>
       <%= render 'templates', :f => builder %>
     <% end %>
   <% end %>


xprazak2:
Before:




After:

xprazak2 (approved):
Sorts the templates alphabetically and the order is consistent for all operating systems. ACK

app/views/operatingsystems/_template_defaults.html.erb
```diff
@@ -12,7 +12,7 @@
                           link_to("templates", provisioning_templates_path)).html_safe) %>
     <% end %>
 
-    <%= f.fields_for :os_default_templates do |builder| %>
+    <%= f.fields_for :os_default_templates, f.object.os_default_templates.sort_by{ |t| t.template_kind.to_s.downcase } do |builder| %>
```
ares, line 15: doesn't this cause n+1 queries problem to load template kind for os_default_template? why not using join + order from AR? or at least it should `includes(:template_kind) when it loads os_default templates and then you can use `sort_by` since objects would be already loaded


also not sure why rubocop didn't find it, but there should be a whitespace between `sort_by` and `{`

ares (changes requested):
needs fixing the N+ query issue

app/views/operatingsystems/_template_defaults.html.erb
mmoll: > not sure why rubocop didn't find it

I guess the relevant rule is not enabled ;)

ares: Or it does not scan erb tags.

gnurag: @ares That indeed causes n+1 queries. With joins+order (or, includes+order), it seems like tablename.fieldname may need to be hardcoded: ```os_default_templates.includes(:template_kind).order('template_kinds.name')```. 

Or else I could use sort_by itself: ```os_default_templates.includes(:template_kind).sort_by { |t| t.template_kind.to_s.downcase }```. What would be the recommended option?

ares: I'd say the difference is minor, since this won't be big set, but I'd prefer the first variant. Though instead of hardcoding `template_kinds` string, I'd prefer to use `.order("#{TemplateKind.table_name}.name")

PR #5339, opened 2018-03-20
Refs #22493 - documented logging macros

lzap:
It was missing in the initial patch. Since this is a trivial change, I
am not creating extra ticket for this. @ares

### DIFF
--- a/config/logging.yaml
+++ b/config/logging.yaml
@@ -56,10 +56,16 @@
 #     [%L] line number where the logging request was issued
 #     [%M] method name where the logging request was issued
 #     [%X{request}] request-ID set in HTTP headers, or random UUID
-#     [%X{session}] random ID assigned per session, not available early on during
-#                   request logging, but prevents disclosure of the session ID
-#     [%X{string}] variable set via MDC (::Logging.mdc['string'])
-#     [%x{string}] variable set via NDC (::Logging.ndc['string'])
+#     [%X{session}] random ID assigned per session
+#     [%X{user_login}] currently logged user if present
+#     [%X{org_id}] current organization id if available
+#     [%X{loc_id}] current location id if available
+#     [%X{remote_ip}] remote IP
+#     [%x{exception_class}] exception info (only for exception entries)
+#     [%x{exception_message}] exception info (only for exception entries)
+#     [%x{exception_backtrace}] exception info (only for exception entries)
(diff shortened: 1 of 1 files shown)


config/logging.yaml
```diff
@@ -56,10 +56,19 @@
...
 #     [%M] method name where the logging request was issued
 #     [%X{request}] request-ID set in HTTP headers, or random UUID
-#     [%X{session}] random ID assigned per session, not available early on during
-#                   request logging, but prevents disclosure of the session ID
-#     [%X{string}] variable set via MDC (::Logging.mdc['string'])
-#     [%x{string}] variable set via NDC (::Logging.ndc['string'])
+#     [%X{session}] random ID assigned per session
+#     [%X{user_login}] currently logged user if present
+#     [%X{org_id}] current organization id if available
+#     [%X{loc_id}] current organization id if available
```
ares: current location I guess :-)

ares (dismissed):
one minor thing and good to go (we don't need to wait for tests :-)

mmoll:
Please add `[skip ci]` as second line of the commit message, then (at least some?) tests shouldn't get triggered.

lzap:
Added the keyword, rebased. Nice @mmoll thanks!

config/logging.yaml
```diff
@@ -56,10 +56,19 @@
...
+#     [%X{org_id}] current organization id if available
+#     [%X{loc_id}] current location id if available
+#     [%X{remote_ip}] remote IP
+#     [%x{exception_class}] exception info (only for exception entries)
+#     [%x{exception_message}] exception info (only for exception entries)
+#     [%x{exception_backtrace}] exception info (only for exception entries)
+#     [%X{string}] generic field set via MDC (::Logging.mdc['string'])
+#     [%x{string}] generic field set via NDC (::Logging.ndc['string'])
+#
+#  Some MDC/NDC macros might be available in some contexts and they will render as
```
mmoll: should this read "...might NOT be available..."?

lzap: What I want to say is that we don't filter out any fields so you can end up with:

```
{ ....
 mdc: {
   "org_name", ""
 }
}
```

But I think it's not relevant, I will remove that.

lzap:
Rebased, the comment was actually irreverent for the last version. It was leftover.

mmoll (approved):

mmoll:
merged, díky @lzap!

PR #5340, opened 2018-03-20
Fixes #22954 - only deserialize safe objects when parsing metadata

ares:
see https://ruby-doc.org/stdlib-2.1.3/libdoc/psych/rdoc/Psych.html#method-c-safe_load for list of safe objects

### DIFF
--- a/app/models/template.rb
+++ b/app/models/template.rb
@@ -101,7 +101,7 @@ class Template < ApplicationRecord
   # Pull out the first erb comment only - /m is for a multiline regex
   def self.parse_metadata(text)
     extracted = text.match(/<%\#[\t a-z0-9=:]*(.+?).-?%>/m)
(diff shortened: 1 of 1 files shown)


mmoll:
[test katello]

mmoll:
foreman test failure unrelated

tbrisker (approved):
works for me, although there are quite a few other `YAML.load` calls elsewhere, should we also change them?

tbrisker:
see also http://projects.theforeman.org/issues/19846

ares:
It depends on what data they deserialize. If that's some form of a user input then it would be a good idea. Do you want to mix that into this PR?

mmoll:
I think this is fine as-is. BTW, there's a rubocop rule (Security/YAMLLoad and for JSON Security/JSONLoad) for this. ;)

mmoll:
merged, díky @ares!

PR #5343, opened 2018-03-21
Fixes #22158 - Add rake task to delete old audits

tbrisker:
Set the default period to 90 days or older.

### DIFF
--- a/lib/tasks/audits.rake
+++ b/lib/tasks/audits.rake
@@ -1,0 +1,24 @@
+# TRANSLATORS: do not translate
+desc <<-END_DESC
+Expire old audits automatically
+
+Available conditions:
+  * days        => number of days to keep audits (defaults to 90)
+
+  Example:
+    rake audits:expire # expires all audits older then 90 days
+    rake audits:expire days=7 # expires all audits older then 7 days
+
+END_DESC
+
+namespace :audits do
+  task :expire => :environment do
+    before = ENV['days'].to_i.days.ago if ENV['days']
+    before ||= 90.days.ago
+    audits = Audited::Audit.up_until(before)
+    puts "Deleting audits older than #{before}. This might take a few minutes..."
+    TaxableTaxonomy.where(taxable_type: "Audited::Audit", taxable: audits).delete_all
+    count = audits.delete_all
+    puts "Successfully deleted #{count} audits!"
+  end
+end


tbrisker:
[test upgrade]

ares (approved):
Tested, works great! Given the fact it's plain DELETE from 2 tables, there's not much we could do about bulk processing I guess. Tested with deletion of hundred records in 7k table and it took 7ms.

Thanks @tbrisker, great addition, merging!

tbrisker:
@ares I considered if we need to do it in bulks, so I tested how long it would take to delete 12M records on a local DB. I didn't have enough audits so I did the test on a fact_value table with a similar filter, even though it is a bit smaller table I don't expect the difference to be in orders of magnitude. 
It took about 1 minute to delete 12M records with a date filter on my puny laptop. So I assume this won't take much more than that on a production DB even when accounting for the larger table width, so I think that it's safe enough to just do in one shot (I don't expect this task to be executed very often anyways).

lzap:
Sorry to comment on late topic, but I think we should not be deleting records by default, but exporting them and then deleting them. It can be some text format, something simple, but it's never a good idea to just delete archival data.

Edit: Other than that I think 90 days is a good policy, well done :-)

ares:
People can backup the DB before runnung this rake task. AFAIK, it's not put to cron by default

lzap:
You keep backups only for few months/years. This is archival data you are required to archive for 5-10 years in some cases.

ares:
Hence I'd recommend also use audit logger. Logs are usually archived this way. But what I'm saying is, the cron task needs to be ran explicitly atm, so at that moment people can explicitly trigger backup (or dump of any sort, e.g. through API)

lzap:
Logs are rotated, they are not the best place either. The reason why I am commenting is that *by default* we delete important audit data after 3 months without user setting up anything. That's not a good practice, by default we should be moving out the data somewhere else on a file system. Thus I used word "exporting". Sure, users can configure ElasticSearch these days, but we should be nice for *default installations*. That's why I am bringing this topic, maybe you can bring it on the PMs meeting I dunno.

tbrisker:
By default we aren't deleting anything. This task doesn't run unless explicitly called or manually added to crontab. unlike other clean up tasks, we did not set up crontab for this exactly for that reason.

lzap:
I was under impression someone is adding that rake task to cron or tasks, that's why I commented.

ohadlevy:
just FYI, over the years (I have a few years of audits) I've noticed that
some of the ID's get reassigned (e.g. a host is created, later on deleted,
and then a different one created again with the same id) which creates
incorrect links between audits and objects, while not a big deal (as the
audit data is still relevant (most of the time unless its the user that the
object that its id's are rotated))...

PR #5345, opened 2018-03-21
Fixes #19983 - allow setting role description

ares:

### DIFF
--- a/app/services/foreman/plugin/role_lock.rb
+++ b/app/services/foreman/plugin/role_lock.rb
@@ -7,33 +7,40 @@ module Foreman
         @plugin_id = plugin_id
       end
 
-      def register_role(name, permissions, role_registry)
+      def register_role(name, permissions, role_registry, description = '')
         User.as_anonymous_admin do
           Role.transaction do
-            role = process_role name, permissions
+            role = process_role name, permissions, description
             role_registry.role_ids << role.id
           end
         end
       end
 
-      def process_role(name, permissions)
+      def process_role(name, permissions, description = '')
         role = Role.find_by :name => name
-        if role&.origin && role.permission_diff(permissions).present?
-          return update_plugin_role_permissions role, permissions
+        if role
+          role.update_attribute(:description, description) if role.description != description
+
+          if role&.origin && role.permission_diff(permissions).present?
+            return update_plugin_role_permissions role, permissions
+          end
+
+          if role&.permission_diff(permissions)&.empty?
+            role.update_attribute :origin, @plugin_id if role.origin.empty?
+            return role
+          end
+
         end
-        if role&.permission_diff(permissions)&.empty?
-          role.update_attribute :origin, @plugin_id if role.origin.empty?
-          return role
-        end
+
         Role.without_auditing do
           rename_existing role, name if role
-          create_plugin_role name, permissions
+          create_plugin_role name, permissions, description
         end
       end
 
-      def create_plugin_role(name, permissions)
+      def create_plugin_role(name, permissions, description = '')
         Role.ignore_locking do
-          role = Role.create! :name => name, :origin => @plugin_id
+          role = Role.create! :name => name, :origin => @plugin_id, :description => description
           role.add_permissions!(permissions)
           role
         end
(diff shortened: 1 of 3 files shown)


ares:
test failures seems unrelated, @xprazak2 since you were involved in this a lot, would you mind to take a look?

boaz0 (dismissed):
LGTM :1st_place_medal:

app/services/foreman/plugin/role_lock.rb
```diff
@@ -7,33 +7,39 @@ def initialize(plugin_id)
...
+        if role
+          role.description = description
+
+          if role.origin && role.permission_diff(permissions).present?
+            return update_plugin_role_permissions role, permissions
+          end
+
+          if role.origin.empty? && role.permission_diff(permissions).empty?
+            role.update_attribute :origin, @plugin_id
+            return role
```
tbrisker: This is becoming a bit complicated, maybe break up into functions?

app/services/foreman/plugin/role_lock.rb
```diff
@@ -7,33 +7,39 @@ def initialize(plugin_id)
         @plugin_id = plugin_id
       end
 
-      def register_role(name, permissions, role_registry)
+      def register_role(name, permissions, role_registry, description = nil)
```
tbrisker: Above you default description to `''`, so this will never be called without a description parameter afaict, and will override the role description with blank string if a description isn't passed from the plugin

tbrisker (changes requested):

app/services/foreman/plugin/role_lock.rb
ares: there's not much we can extract, since two conditions are using return, we'd need to double return which wouldn't help the readability that much, I'd prefer to keep it as it is

app/services/foreman/plugin/role_lock.rb
ares: Yes, which I think is correct as we now have mixes of nil and `''`, whenever user saved the role (before they were locked) the form passed empty string. I can keep nil in `role` plugin DSL method, if you prefer since we don't have any restriction on NOT NULL.

ares:
I've rebased and replied to you comments, @tbrisker pls let me know if you want me to change something.

ares:
updated, tests are now green, sadly this missed 1.18 so plugins still can't provide descriptions to roles :-/

app/services/foreman/plugin/role_lock.rb
tbrisker: TBH it doesn't really matter to me if we default to `nil` or `''`, but in any case the default here will never be used since this function is only called from `Plugin#role`, so it is never called with just 3 parameters (except maybe in the tests).

ares: I don't really understand why it's a problem, whenever we call this method we need to explicitly send `''` as a description now. Yes, it's one place ATM (3 in tests). But anyway, I pushed the new version without description default value in here.

ares:
new version pushed, please take another look

tbrisker:
@ares - I think you missed the point in my comment. In `plugin.rb`, the default for description is set to `''`, while in `role_lock.rb` all the functions set a default of `nil`. All the default values in `role_lock.rb` don't really matter, since it is only called from `plugin.rb` which sets the default to `''`, so there is never a situation where these functions are called with just 3 parameters. In any case, if you want to have defaults in `role_lock.rb` (e.g. in case one day we'll want to use that from somewhere else and fallback to the default), that's fine, but we should be consistent with the default values - either `nil` or `''` everywhere. In any case this isn't a significant issue, more of a comment regarding the inconsistency.

ares:
sorry, I probably then misunderstood this comment

> TBH it doesn't really matter to me if we default to nil or '', but in any case the default here will never be used since this function is only called from Plugin#role, so it is never called with just 3 parameters (except maybe in the tests).

I thought you no longer care about the default value (nil or '') but rather that the third param is never used...

I reset back optional descriptions but now with '' as a default value everywhere. I hope that now it's fine.

tbrisker (approved):
Thanks @ares !

tbrisker:
[test katello]

tbrisker:
Katello failure unrelated.

PR #5347, opened 2018-03-22
fixes #22983 - add HostgroupClass in permissions list

emildragu:
Add missing permissions ['HostgroupClass', 'create_hostgroup_classes']
required for adding classes to hostgroups.

### DIFF
--- a/db/seeds.d/020-permissions_list.rb
+++ b/db/seeds.d/020-permissions_list.rb
@@ -61,6 +61,7 @@ class PermissionsList
         ['Hostgroup', 'create_hostgroups'],
         ['Hostgroup', 'edit_hostgroups'],
         ['Hostgroup', 'destroy_hostgroups'],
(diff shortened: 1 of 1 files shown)


timogoebel:
@emildragu: Thanks. :-)

ok to test

ares:
Similar report also asked for HostConfigGroup permission, could you please add it as well? http://projects.theforeman.org/issues/22010

ares:
According to routes, this should be covered by `edit_hostgroup` permission. I think we should not try to enforce new permission on relation entity between puppet class and hostgroup. We should simply allow creation as there's probably no value of having a separate permissions for that. As long as users can edit hostgroup, they should be able to assign classes. I'll open a new PR with the fix also for config group.

ares:
See #5470 for the fix, I think this one can be closed.

tbrisker:
Thank you for your contribution @emildragu, looks like this has been solved in GH-5470. Closing.

PR #5351, opened 2018-03-22
Fixes #22997 - increase fact_name name length

templed:
Increases field length for fact_name name to accommodate,
for example, container mount point paths since they can be
greater than the default ActiveRecord string field length.

### DIFF
--- a/db/migrate/20180322155732_increase_fact_name_length.rb
+++ b/db/migrate/20180322155732_increase_fact_name_length.rb
@@ -1,0 +1,17 @@
+class IncreaseFactNameLength < ActiveRecord::Migration[5.0]
+  def up
+    change_table :fact_names do |t|
+      t.change :name, :string, :limit => 4096
+      t.change :ancestry, :string, :limit => 4096
+      t.change :short_name, :string, :limit => 4096
+    end
+  end
+
+  def down
+    change_table :fact_names do |t|
+      t.change :name, :string, :limit => 255
+      t.change :ancestry, :string, :limit => 255
+      t.change :short_name, :string, :limit => 255
+    end
+  end
+end


app/models/fact_name.rb
```diff
@@ -23,7 +23,7 @@ class FactName < ApplicationRecord
 
   default_scope -> { order('fact_names.name') }
 
-  validates :name, :uniqueness => { :scope => :type }
+  validates :name, :length => { :maximum => 4096 }, :uniqueness => { :scope => :type }
```
timogoebel: We have `validates_lengths_from_database` some lines above, so this change is not necessary. It'll pick up the (new) size automatically.

timogoebel (changes requested):
@templed: Thanks for your contribution. Not a full review, just something that struck my mind.

app/models/fact_name.rb
templed: OK. I've amended the commit to remove this redundant check.

(FWIW this is my first time ever doing anything related to Ruby/RoR so I was expecting issues like this.)

tbrisker:
let me kick off CI:
ok to test

timogoebel (dismissed):

db/migrate/20180322155732_increase_fact_name_length.rb
```diff
@@ -0,0 +1,17 @@
+class IncreaseFactNameLength < ActiveRecord::Migration
```
tbrisker: Tests are failing since this needs to be `ActiveRecord::Migration[5.0]`

tbrisker (changes requested):

templed:
Hi Tomer,

I didn't know force pushing would dismiss your review (seems an odd behavior by GH). I added `[5.0]` to the Migration class so hopefully that will fix the problem.

lzap:
I waslittle bit worried about migration time for larger databases, users with random facts imported (hypervisors, docker hosts, disk arrays) can have millions of records. But it looks like this operation *should* be good: https://www.postgresql.org/docs/9.3/static/datatype-character.html

lzap (approved):
LGTM

ekohl:
> I waslittle bit worried about migration time for larger databases, users with random facts imported (hypervisors, docker hosts, disk arrays) can have millions of records. But it looks like this operation should be good: https://www.postgresql.org/docs/9.3/static/datatype-character.html

Note that postgres does what you'd hope it to do. MySQL (at least older versions like those in EL7) does what you don't want to do: lock the table and recreate it. This is very expensive in terms of IO. MySQL also has some limitations around column lengths and indexes. I don't see a test for this, but maybe we already have one? Otherwise it'd be good to test this.

lzap:
There is indeed a limitation for MySQL: https://dev.mysql.com/doc/refman/8.0/en/create-index.html#create-index-column-prefixes and it's a mess depending on MySQL version and backend type. This can lead to error "Specified key was too long; max key length is 767 bytes" when there is an index on the column.

We want to drop MySQL in Foreman 1.24, so I am going to close this PR for now. We should be able to add this migration for 1.25 in an year, I am marking my calendar. Thanks for your contribution.

In the meantime, you can use fact filtering feature and add "mountpoints" to the filter so they are all rejected during fact import. It's usually useless anyway to keep hundreds of these per each host.

PR #5353, opened 2018-03-23
Fixes #23002 - seed Installation Media with taxonomies

sean797:
We currently do the same thing with Templates,
lets also do it for Installation Media.


tbrisker (approved):
makes sense, test failure unrelated. Thanks @sean797 !

PR #5355, opened 2018-03-24
Fixes #23007 - renew certs used in oVirt model test

mmoll:


mmoll:
the new certs expire in 10 years, that should be OK for now ;)

ohadlevy (approved):

PR #5359, opened 2018-03-26
Fixes #22049 - Add metadata to Audit view

tbrisker:
This could use some love on the UI side I think, but at least a first step. @rohoover would love your feedback as to how to make this prettier :smile: 

image

### DIFF
--- a/app/views/audits/show.html.erb
+++ b/app/views/audits/show.html.erb
@@ -26,6 +26,31 @@
     </div>
     <% end %>
     <div class='tab-pane <%= "active" unless tmplt%>' id="tab1">
+      <div class="row">
+        <h3 class="col-md-12"><%= _("Audit Metadata:") %></h3>
+      </div>
+      <div class="row">
+        <strong class="col-md-2"><%= _("Audit Time:")%></strong>
+        <div class="col-md-10"><%= date_time_absolute(@audit.created_at) %></div>
+      </div>
+      <div class="row">
+        <strong class="col-md-2"><%= _("User Name:")%></strong>
+        <div class="col-md-10"><%= @audit.username %></div>
+      </div>
+      <% if Taxonomy.organizations_enabled %>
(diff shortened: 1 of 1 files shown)


Rohoover (commented):
I would do the following:
1) Make label names bold & Title case
2) Bring labels and text a little closer together
3) Add a bit more space between the meta-data and "Changes"

I think those would be the simplest thing without undertaking a larger redesign.

tbrisker:
@Rohoover only problem I have is with 2 - I can either bring the labels and text together and lose the alignment of the text, or leave it as is. If I use a smaller space for the label, the longest one ("affected organizations") gets broken into two lines. There may also be similar issues in languages that have longer labels. Please advise how to proceed.

Rohoover (commented):
Makes sense, thanks for clarifying! Leave as-is.

tbrisker:
@Rohoover thanks, updated:
image

Please confirm.

Rohoover (approved):
LGTM

ares (changes requested):
All looks good and is a great improvement already.

I'm thinking about audits for non-taxable resources, e.g. architecture. Seeing `Affected Organizations: ` might be a misleading as it in fact affects all orgs. Could we either say "all" in case the resource is not taxable, or `N/A` if it's not? Given the fact such audit won't be listed in audits page on specific organization, 'N/A' might be more applicable.

And perhaps for audits that are auditing taxable resource, but tracks no change, it would be good to say 'None' instead of empty line. Or with additional characters, so it's not confusing if someone creates org with name None :-)

ares:
sorry and one more thought, would you mind making User name, orgs and locs clickable leading to $resource show/edit page?

ares (approved):
I decided to go ahead and merge as this is already a great improvement. I'll send separate PR with additions I had in mind.

Thanks @tbrisker, merging!

PR #5360, opened 2018-03-27
WIP: Refactor Patternfly Notification drawer

glekner:
Refactoring the Notification drawer with the PF-React implementation.
In order to run foreman with the PF-React drawer, please clone my [branch](https://github.com/gilad215/patternfly-react/tree/feature-notifdrawer) and run `npm link`. then run `npm link patternfly-react` in foreman

**Works**

-  Notifications are fetched and displayed correctly
-  `togglePanel`, `onNotificationClick`, `onMarkPanelAsRead`, `toggleDrawerHide`

**Doesn't Work (Needs to be implemented in Foreman)**
- `onMarkPanelAsClear`, `onClickedLink`, `onNotificationHide`
- CSS is being used from the inner flie `notifications.scss`, I patched it a bit and it works partially with my drawer but I think we should fetch the CSS from Patternfly

### DIFF
--- a/test/unit/tasks/seeds_test.rb
+++ b/test/unit/tasks/seeds_test.rb
@@ -25,10 +25,28 @@ class SeedsTest < ActiveSupport::TestCase
     User.current = nil
   end
 
-  test 'populates features' do
-    count = Feature.count
+  test 'populates multiple tables' do
+    tables = [Feature, Ptable, ProvisioningTemplate, Medium, Bookmark, AuthSourceExternal]
+
+    tables.each do |model|
+      assert model.unscoped.count.zero?
+    end
+
     seed
-    assert_not_equal count, Feature.count
+
+    tables.each do |model|
+      refute model.unscoped.count.zero?
+    end
+
+    refute Ptable.unscoped.where(:os_family => nil).any?
+    refute Medium.unscoped.where(:os_family => nil).any?
+    Dir["#{Rails.root}/app/views/unattended/**/*.erb"].each do |tmpl|
+      if tmpl =~ /partition_tables_templates/
+        assert Ptable.unscoped.where(:template => File.read(tmpl)).any?, "No partition table containing #{tmpl}"
+      else
+        assert ProvisioningTemplate.unscoped.where(:template => File.read(tmpl)).any?, "No template containing #{tmpl}"
+      end
+    end
   end
 
   test 'populates hidden admin users' do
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -53,6 +53,7 @@ gem 'sshkey', '~> 1.9'
 gem 'dynflow', '>= 0.8.29', '< 1.0.0'
 gem 'daemons'
 gem 'get_process_mem'
+gem 'rdoc'
```
ohadlevy: not related / required?

bundler.d/journald.rb
```diff
@@ -1,4 +1,4 @@
 # disable to avoid journald native gem in development setup
 group :journald do
-  gem 'logging-journald', '~> 1.0'
+ # gem 'logging-journald', '~> 1.0'
```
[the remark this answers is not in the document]
ohadlevy: you can use ```bundle install --without journald``` to get the same effect.

webpack/assets/javascripts/react_app/components/notifications/index.js
```diff
@@ -38,19 +39,26 @@ class notificationContainer extends React.Component {
...
-            <Drawer
-              onExpandGroup={expandGroup}
-              onClickedLink={onClickedLink}
-              onMarkAsRead={onMarkAsRead}
-              onMarkGroupAsRead={onMarkGroupAsRead}
-              expandedGroup={expandedGroup}
-              notificationGroups={notifications}
-              toggleDrawer={toggleDrawer}
+            <NotificationDrawerWrapper
+            panels={formatedPanels}
```
ohadlevy: nitpick: wrong indentation.

ohadlevy (changes requested):
thanks @gilad215, a few code review comments inline.

regarding the missing features:
onMarkPanelAsClear - I assume this should be a new API call to bulk delete actual notifications.

onClickedLink - Why can't we handle links? not sure I follow?

onNotificationHide - is this hiding the notification (same as clear?) or closing the drawer? I assume on click outside works as before?

Gemfile
glekner: Foreman won't compile at my end without this line, you can ignore this because ill be working on my Mac soon and I will delete this.

ohadlevy:
I'm not sure whats the issue TBH: but at worse you can either not include
this file, or better add it to your bundle.d/Gemfile.local.rb instead.

glekner:
@ohadlevy 
`onClickedLink` - I think i can fix this at my branch, ill take a look.
`onNotificationHide` - This is the button in the dropdown that says "Hide this notification". It was my idea to add this so its not mandatory. we should treat this the same as `onMarkPanelAsClear` just deleting a single notification by its key instead of the whole panel.

glekner:
- Fixed `onClickedLink`
- Added more CSS to `notifications.scss` to enable in-panel scrolling in scenarios where one panel has 5+ notifications

glekner:
Added `onNotificationHide`, `onMarkPanelAsClear` functionality.

TODO: Tests

glekner:
- Added New + Missing tests for the Notification Reducers/Actions
- Waiting for the drawer merge on PF-React, any suggestions for the codeclimate tests? @sharvit

PR #5362, opened 2018-03-27
fixes #23026 - make changes to compute resource passwords audited

rplevka:
Removing the :password field from the audit exceptions.
The passwords come out [redacted] in the audit logs, so this is safe.

audit_image


ares:
ok to test

tbrisker (approved):
Test failure unrelated. Thanks @rplevka !

PR #5371, opened 2018-03-28
Fixes #19980 - Merge hashes or arrays with ERB

FrediWeber:
Cast the value to the right type if it does not exclusively contains ERB and merge is enabled.
This way it is possible to merge hashes or arrays with ERB in it which resulted in an error before.

### DIFF
--- a/app/models/lookup_value.rb
+++ b/app/models/lookup_value.rb
@@ -70,7 +70,7 @@ class LookupValue < ApplicationRecord
   end
 
   def validate_and_cast_value
-    return if !self.value.is_a?(String) || value.contains_erb?
+    return if !self.value.is_a?(String) || value.contains_erb? == 0 || (value.contains_erb? && !lookup_key.merge_overrides)
     Foreman::Parameters::Caster.new(self, :attribute_name => :value, :to => lookup_key.key_type).cast!
   rescue StandardError, SyntaxError => e
     Foreman::Logging.exception("Error while parsing #{lookup_key}", e)
--- a/test/models/lookup_value_test.rb
+++ b/test/models/lookup_value_test.rb
@@ -153,6 +153,36 @@ class LookupValueTest < ActiveSupport::TestCase
     assert_equal lv.value, "<%= [4,5,6] %>"
   end
 
+  test "should cast and uncast hashes with not only erb" do
+    key = FactoryBot.create(:puppetclass_lookup_key, :as_smart_class_param,
+                            :override => true, :key_type => 'hash', :merge_overrides => true, :avoid_duplicates => false,
+                            :default_value => "{}", :puppetclass => puppetclasses(:one))
+    lk1 = LookupValue.new(:value => "---\n foo: <%= bar %>", :match => "hostgroup=Common", :lookup_key => key)
+    assert lk1.save!
+    assert lk1.value.is_a? Hash
+    assert_equal lk1.value, {"foo" => "<%= bar %>"}
+  end
+
+  test "should cast and uncast hashes with not only erb even when erb is in first place" do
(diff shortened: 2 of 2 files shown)


ohadlevy:
[ok to test]

FrediWeber:
Could someone restart the tests please?

ekohl:
Could you rebase instead? It's based on an old base commit and things may have changed.

FrediWeber:
Rebase done

FrediWeber:
Rebase
Do you need something else from me to approve this pull request?

FrediWeber:
Squashed

FrediWeber:
Rebase

FrediWeber:
Rebase

xprazak2:
@FrediWeber, is this still wip?

@orrabin any chance you could take a look?

FrediWeber:
@xprazak2 I had to change it to WIP because of PR #5559.
Basically everything works and it just changes things from not working at all to not perfect.
The existing problem is the following:
If you have a smart class parameter with the type "YAML", "Merge nested values" enabled and an macro in some value, it will not merge the values but produces an error. This PR fixes that.

The not perfect part:
It will not serialize the value as string (values with only ERB in it are also not serialized already) in the DB and the migration script I wanted to change in PR #5559 serializes every non-string value to a string.

dLobatog (changes requested):
Hey @FrediWeber ! The PR seems to work well, I could definitely reproduce theissue and fix it. However as you mention, you have to save again the smart class parameters, and that's too much work. 

A migration that goes through all LookupValues where this is applicable where you just run 'validate_and_cast_value' and save again would be great, could you add that to this PR?

cc @tbrisker this should be 1.18.X, 1.19.X, etc..

app/models/lookup_value.rb
```diff
@@ -70,7 +70,7 @@ def sanitize_match
   end
 
   def validate_and_cast_value
-    return if !self.value.is_a?(String) || value.contains_erb?
+    return if !self.value.is_a?(String) || value.contains_erb? == 0 || (value.contains_erb? && !lookup_key.merge_overrides)
```
tbrisker, line 73: `value.contains_erb? == 0` only if it starts with erb, is this intentional?

orrabin: isn't `(value.contains_erb? && !lookup_key.merge_overrides)` already covered by `value.contains_erb?`?

FrediWeber: @tbrisker Yes, it should only return and thus not cast the value if the ERB is right at the beginning. If the ERB is further down in the YAML structure of the value, it should cast it to whatever value the user defines for the parameter.

FrediWeber: @orrabin Not exactly. Because I check for `value.contains_erb? == 0` all cases where the ERB is not in the beginning are not covered. But because I want to keep the impact of this fix as small as possible I also check if the parameter has the override flag set (because the bug doesn't show up otherwise).

FrediWeber:
@dLobatog Yes definitely. I will look into that as soon as I find some time for the change.

app/models/lookup_value.rb
orrabin: @FrediWeber I meant why did you split it to `==0` and add a condition that is already contained but I see your answer to @tbrisker about that. Thanks for clarifying.

tbrisker:
@orrabin @dLobatog @FrediWeber - what is the status here?

ares:
There has been no update for a long time, we take the liberty to close this PR. Thanks @FrediWeber, please reopen this PR if you want to continue working on it. It'd require a rebase now.

PR #5374, opened 2018-03-28
Fixes #22546 - Return power action status for ovirt (CVE-2018-1097)

mmoll:


lzap (approved):
See my "review" in the original PR:

https://github.com/theforeman/foreman/pull/5369

PR #5376, opened 2018-03-29
Fixes #23062 - Add custom message on form submit

xprazak2:


boaz0:
SGTM :+1: , can you add a test to this change, though?

webpack/assets/javascripts/react_app/redux/actions/common/forms.js
```diff
@@ -46,7 +46,7 @@ export const submitForm = ({
         dispatch(addToast({
           type: 'success',
           // eslint-disable-next-line no-undef
-          message: Jed.sprintf('%s was successfully created.', __(item)),
+          message: message || Jed.sprintf('%s was successfully created.', __(item)),
```
ohadlevy, line 49: what if you would like to use item in your custom message? or are you assuming you know it already so you can customize it?

also, can you add a test? 

thanks

xprazak2: I added a test. I think you know `item` already, because you are passing it into the `submitForm`

pondrejk (approved):

tbrisker (approved):
Thanks @xprazak2 !

PR #5377, opened 2018-03-29
Fixes #23059 - journald plugin is not required

lzap:
This is a non-required gem, it's a plugin. It works with bundler but
with bundler_ext it fails and Rails won't boot.


ekohl (approved):
Can confirm this is needed.

ekohl:
Test failure looks unrelated and I've confirmed this is needed to boot the application. Thanks!

PR #5378, opened 2018-03-29
Refs #22049 - improve metadata usability

ares:
This is a follow up of https://github.com/theforeman/foreman/pull/5359

before
screenshot_20180329_152522

after
screenshot_20180329_152049
screenshot_20180329_152040

@tbrisker and @rohoover, your comments would be appreciated, I'd like to get this into 1.18

### DIFF
--- a/app/helpers/audits_helper.rb
+++ b/app/helpers/audits_helper.rb
@@ -106,6 +106,26 @@ module AuditsHelper
     date_time_absolute(audit.created_at)
   end
 
+  def audit_affected_locations(audit)
+    base = audit.locations.authorized(:view_locations)
+    return _('N/A') if base.empty?
+
+    authorizer = Authorizer.new(User.current, base)
+    base.map do |location|
+      link_to_if_authorized location.name, hash_for_edit_location_path(location).merge(:auth_object => location, :permission => 'edit_locations', :authorizer => authorizer)
+    end.to_sentence.html_safe
+  end
+
+  def audit_affected_organizations(audit)
+    base = audit.organizations.authorized(:view_organizations)
+    return _('N/A') if base.empty?
+
+    authorizer = Authorizer.new(User.current, base)
+    base.map do |organization|
+      link_to_if_authorized organization.name, hash_for_edit_organization_path(organization).merge(:auth_object => organization, :permission => 'edit_organizations', :authorizer => authorizer)
+    end.to_sentence.html_safe
+  end
+
   def audited_icon(audit)
     style = 'label-info'
     style = case audit.action
--- a/app/views/audits/show.html.erb
+++ b/app/views/audits/show.html.erb
@@ -30,22 +30,22 @@
       </div>
       <div class="row">
         <strong class="col-md-2"><%= _("Audit Time:")%></strong>
-        <div class="col-md-10"><%= date_time_absolute(@audit.created_at) %></div>
+        <div class="col-md-10"><%= audit_time(@audit) %></div>
       </div>
       <div class="row">
         <strong class="col-md-2"><%= _("User Name:")%></strong>
-        <div class="col-md-10"><%= @audit.username %></div>
+        <div class="col-md-10"><%= audit_user(@audit) %></div>
       </div>
       <% if Taxonomy.organizations_enabled %>
         <div class="row">
           <strong class="col-md-2"><%= _("Affected Organizations:")%></strong>
-          <div class="col-md-10"><%= @audit.organizations.authorized(:view_organizations).to_sentence %></div>
+          <div class="col-md-10"><%= audit_affected_organizations(@audit) %></div>
         </div>
       <% end %>
       <% if Taxonomy.locations_enabled %>
         <div class="row">
           <strong class="col-md-2"><%= _("Affected Locations:")%></strong>
-          <div class="col-md-10"><%= @audit.locations.authorized(:view_locations).to_sentence %></div>
+          <div class="col-md-10"><%= audit_affected_locations(@audit) %></div>
         </div>
       <% end %>
       <hr/>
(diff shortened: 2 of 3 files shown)


tbrisker (commented):
tbh, i'm not sure I understand the workflow this is meant to support. 
When viewing an audit, why would the next step be editing one of the location affected or the user who made the change? It might make some sense if we had show pages for those objects that provide further information, but i am not sure how much this is needed, and it significantly increases the code complexity.

ares:
As a User, if I want to see information about org/loc in which the event happend I only can see that on edit page. If we had show pages, I'd link to those. It's simply "give me more information" about the resource. But at least from edit page, once can find out what other resources are assigned.

I don't understand the code complexity concern. The link construction is the same as with any other link. Methods for generating all links contain `if` condition to nicely display the information, that there's no org/loc to be displayed, instead of white space. IMHO better user experience.

app/helpers/audits_helper.rb
```diff
@@ -106,6 +106,26 @@ def audit_time(audit)
     date_time_absolute(audit.created_at)
   end
 
+  def audit_affected_locations(audit)
+    base = audit.locations.authorized(:view_locations)
+    authorizer = Authorizer.new(User.current, base)
+    if base.empty?
+      _('none')
```
tbrisker: This may also be "all" if the auditable isn't taxable or has taxonomies, we might want to also accommodate that case?

tbrisker: also, you can `return _('none') if base.empty?` before you build the authorizer and reduce the nesting and unneeded creation.

app/helpers/audits_helper.rb
```diff
@@ -106,6 +106,26 @@ def audit_time(audit)
...
+    else
+      base.map { |location| link_to_if_authorized location.name, hash_for_edit_location_path(location).merge(:auth_object => location, :permission => 'edit_locations', :authorizer => authorizer) }.to_sentence.html_safe
```
tbrisker: This is a very long line, maybe break it into a `do...end` block so it is easier to read?

tbrisker (changes requested):
Ok, I guess this is an overall improvement. A few minor comments inline (same apply to the second function as well).

ares:
Thanks @tbrisker, all comments addressed. we display all now in case Taxonomix is not included in the audited resource. I also added missing org/loc association into API template, that should have been added when we started to display the metadata. Could you please take another look?

ares:
repushed with extracted "all" check into separate method

app/helpers/audits_helper.rb
```diff
@@ -106,6 +106,41 @@ def audit_time(audit)
...
+
+    authorizer = Authorizer.new(User.current, base)
+    base.map do |organization|
+      link_to_if_authorized organization.name, hash_for_edit_organization_path(organization).merge(:auth_object => organization, :permission => 'edit_organizations', :authorizer => authorizer)
+    end.to_sentence.html_safe
+  end
+
+  def affected_taxonomies(audit)
+    begin
+      return _('all') unless audit.auditable_type.constantize.included_modules.include?(Taxonomix)
```
tbrisker: This will return `all` for hosts and taxonomies, since they don't include `Taxonomix`, and might be true for other resources as well that have different ways of linking to taxonomies

app/helpers/audits_helper.rb
```diff
@@ -106,6 +106,41 @@ def audit_time(audit)
...
+    rescue => e
```
tbrisker: When might we get an exception here?

app/views/audits/show.html.erb
```diff
@@ -31,22 +31,22 @@
       </div>
       <div class="row">
         <strong class="col-md-2"><%= _("Audit Time:")%></strong>
-        <div class="col-md-10"><%= date_time_absolute(@audit.created_at) %></div>
+        <div class="col-md-10"><%= audit_time(@audit) %></div>
       </div>
       <div class="row">
         <strong class="col-md-2"><%= _("User Name:")%></strong>
-        <div class="col-md-10"><%= @audit.username %></div>
+        <div class="col-md-10"><%= audit_user(@audit) %></div>
```
tbrisker, line 37: Note that this leads to all audits by this user, not the user itself, not sure if this was intentional or not

app/helpers/audits_helper.rb
```diff
@@ -106,6 +106,41 @@ def audit_time(audit)
...
+    base = audit.organizations.authorized(:view_organizations)
+    return _('none') if base.empty?
+
+    authorizer = Authorizer.new(User.current, base)
+    base.map do |organization|
+      link_to_if_authorized organization.name, hash_for_edit_organization_path(organization).merge(:auth_object => organization, :permission => 'edit_organizations', :authorizer => authorizer)
+    end.to_sentence.html_safe
+  end
+
+  def affected_taxonomies(audit)
```
tbrisker: the function name is misleading

tbrisker (changes requested):

app/helpers/audits_helper.rb
ares: if we have audit for resource that no longer exists (plugin uninstallation, refactoring, etc), we can't rely on `constantize` to succeed

app/views/audits/show.html.erb
ares: it is, because it's the same thing we link on audits list, I though it would be better to be consistent and change the helper when we improve the UI

app/helpers/audits_helper.rb
ares: please suggest better one :-) how about "all_affected_taxonomies_string"?

app/helpers/audits_helper.rb
ares: is there a better way to detect on your mind? I think it's be better if we don't display `all` at all, there's no proper link between audit and taxonomies in this case and it might be confusing if people search for audits using organization attribute, they would expect to see it even if they search for `organization = Default` but the audit only appears if we search for `null? organization`

tbrisker (approved):
Thanks @ares !

Rohoover:
LGTM.

PR #5384, opened 2018-04-01
Fixes #23072 - Fix fact importer transaction isolation

ShimShtein:


test/fact_importer_test_helper.rb
```diff
@@ -6,32 +6,16 @@ def allow_transactions_for(importer)
...
-  module ClassMethods
-    def allow_transactions_for_any_importer
-      FactImporter.singleton_class.prepend FactImporterFactoryStubber
-
-      FactImporter.register_instance_stubs do |importer_class|
-        importer_class.any_instance.stubs(:ensure_no_active_transaction).returns(true)
-      end
+  def allow_transactions_for_any_importer
+    FactImporter.importers.values.each do |importer|
+      importer.any_instance.stubs(:ensure_no_active_transaction).returns(true)
```
lzap: Please use `allow_transactions_for` as a helper method, DRY.

lzap (changes requested):
I struggle to find much value on this explicit checking, I'd rather see it to be removed. I spent some time debugging failed tests because of this exception, it was hidden somewhere and the root cause was not obvious. I encourage you to ditch it but I won't insist.

If you decide to keep it, please extend the exception message with instructions on what to do if you hit this.

ShimShtein:
@lzap, I have DRY'ed the code as requested and added more text to the error.

> I struggle to find much value on this explicit checking

The value of explicit checking is for later identifying tests that will need special care. The problem with fact uploading during an open transaction is unexpected behaviour during parallel upload. I think this kind of concurrency bugs are much harder to track, since it won't be exactly clear where the bug comes from. I would like to know all places that upload facts and do something that will enable parallel execution with all of them in one pass later. Current fix is not parallel execution safe.

The TL;DR version of the issue with transactions, is that postgres adapter uses savepoints for nested transactions (we have an explicit transaction in fact upload code that becomes a savepoint, if there is transaction already open). When the explicit transaction fails (like in case of concurrent fact name writes) it fails the whole _outer_ transaction, instead of rolling back to the savepoint. After that point every query will fail with "transaction is already rolled back" exception.

app/services/fact_importer.rb
```diff
@@ -196,7 +196,16 @@ def db_facts
   end
 
   def ensure_no_active_transaction
-    raise 'Fact names should be added outside of global transaction' if ActiveRecord::Base.connection.transaction_open?
+    message = 'Fact names should be added outside of global transaction.'
+    if env.test?
```
lzap: This broke tests.

test/fact_importer_test_helper.rb
```diff
@@ -6,32 +6,16 @@ def allow_transactions_for(importer)
...
+    def allow_transactions_for_any_importer
+      setup :allow_transactions_for_any_importer
     end
-    instance
-  end
-
-  def instance_stubs
-    @instance_stubs ||= []
   end
 end
```
lzap, line 20: Are both instance and class methods used? Wondering if we can drop one.

lzap (dismissed):
I understand the motivation, I am not convinced that the struggle people can have with this is worth the result. With all that said, the extra explanation is perhaps good enough to give this a go.

test/fact_importer_test_helper.rb
ShimShtein: Class method is the one that is "customer facing" - the one that would be declared by the user. Instance method is the actual implementation, sorry didn't have enough imagination to call it differently. The instance method would be called by the test framework (since we register it as setup callback) and it contains the actual isolation code.

The definition would look like this:

``` ruby
class HostTest < ActiveSupport::TestCase
  include FactImporterIsolation

  allow_transactions_for_any_importer
  # ^ Usage of the class method
```

ShimShtein:
@lzap, Addressed all your comments.

lzap:
Can you elaborate "The problem with fact uploading during an open transaction is unexpected behaviour during parallel upload" a little bit more?

ShimShtein:
> Can you elaborate "The problem with fact uploading during  an open transaction is unexpected behaviour during parallel upload" a  little bit more?

I'll quote myself:
> The TL;DR version of the issue with transactions, is that postgres adapter uses savepoints for nested transactions (we have an explicit transaction in fact upload code that becomes a savepoint, if there is transaction already open). When the explicit transaction fails (like in case of concurrent fact name writes) it fails the whole outer transaction, instead of rolling back to the savepoint. After that point every query will fail with "transaction is already rolled back" exception.

During parallel upload, you can face a race condition, where two fact uploaders try to add the same fact name to the database. This will fail a transaction for one of them. The victim will have its outer transaction failed, and every DB query from that moment will result in unspecific error "transaction is already rolled back". This error is meaningless, and it will be very hard to track it back to failed fact name insert.

@lzap I hope now the issue is more clear.

lzap:
It's been a while, can you rebase so Jenkins retests? I am still not convinced we need this in our test code, I'd rather get rid of this. Maybe someone else can take a look perhaps @tbrisker to give additional feedback. I don't see how any application-level checking would help with transaction isolation or concurrency which happens on the database level.

If nested transactions are the issue, let's get rid of them! Foreman is far from being transaction safe app, not sure if we need nesting.

lzap:
@ShimShtein can you rebase and let's merge it.

ShimShtein:
@lzap, rebased again for the fun of it :)

tbrisker (approved):
Thanks @ShimShtein and @lzap !

PR #5388, opened 2018-04-02
Fixes #23093 - prevent DB deadlocks in IntegrationTestHelper

mmoll:


mmoll:
This ran now three times with the three Ruby versions on PG and it always came back green, so I guess, this really helps...

tbrisker (approved):
Even if this only reduces the frequency I think it is worth it, these failures are quite annoying. Thanks @mmoll !

PR #5392, opened 2018-04-03
Fixes #23039 - Remove deleted LDAP users

juliantodt:
Adds a rake task ldap:remove_deleted_users which checks all users
that authenticate via LDAP whether their users are valid and
deletes invalid ones.

### DIFF
--- a/lib/tasks/ldap.rake
+++ b/lib/tasks/ldap.rake
@@ -1,15 +1,30 @@
-desc <<-END_DESC
-Refreshes LDAP usergroups. It adds to an LDAP usergroup all the foreman users that belong to it, and removes foreman users
-in that usergroup that do not belong in LDAP anymore.
-END_DESC
-
 namespace :ldap do
   task :refresh_usergroups => :environment do
+    desc <<-END_DESC
+    Refreshes LDAP usergroups. It adds to an LDAP usergroup all the foreman users that belong to it, and removes foreman users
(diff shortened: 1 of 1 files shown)


timogoebel:
ok to test

lib/tasks/ldap.rake
```diff
@@ -13,4 +12,17 @@ namespace :ldap do
       end
     end
   end
+  task :remove_deleted_users => :environment do
+    desc <<-END_DESC
+    Deletes all foreman users that authenticate via LDAP, but whose LDAP users do not exist anymore.
+    Also deletes them as the owner of their hosts.
+    END_DESC
+    User.joins(:auth_source).where(:'auth_sources.type' => 'AuthSourceLdap').each do |user|
```
timogoebel: I think you need to wrap this in `User.as_anonymous_admin do [...] end` as no user is set when the rake tasks run.

timogoebel (changes requested):
@juliantodt: Thanks. I found one small issue while testing this. Please take a look. :-)

juliantodt:
@timogoebel Thanks for the review. Had no problems while testing myself but added your suggestion nevertheless.

timogoebel (approved):
Looks good, thanks.

tbrisker (approved):
Thanks @juliantodt !

PR #5393, opened 2018-04-03
Fixes #23040 - Anonymize old audits

juliantodt:
Adds a rake task similar to audits:expire that instead of
deleting the entries anonymizes them.

### DIFF
--- a/lib/tasks/audits.rake
+++ b/lib/tasks/audits.rake
@@ -1,6 +1,6 @@
 # TRANSLATORS: do not translate
 desc <<-END_DESC
-Expire old audits automatically
+Expire or anonymize old audits automatically
 
 Available conditions:
   * days        => number of days to keep audits (defaults to 90)
@@ -8,17 +8,31 @@ Available conditions:
   Example:
     rake audits:expire # expires all audits older then 90 days
     rake audits:expire days=7 # expires all audits older then 7 days
+    rake audits:anonymize days=7 # anonymizes all audits older then 7 days
 
 END_DESC
 
 namespace :audits do
   task :expire => :environment do
-    before = ENV['days'].to_i.days.ago if ENV['days']
-    before ||= 90.days.ago
-    audits = Audited::Audit.up_until(before)
-    puts "Deleting audits older than #{before}. This might take a few minutes..."
+    audits = get_audits
+    puts "Deleting audits older than #{before_date}. This might take a few minutes..."
     TaxableTaxonomy.where(taxable_type: "Audited::Audit", taxable: audits).delete_all
     count = audits.delete_all
     puts "Successfully deleted #{count} audits!"
+  end
+
+  task :anonymize => :environment do
+    puts "Anonymizing audits older than #{before_date}. This might take a few minutes..."
+    count = get_audits.where.not(:remote_address => nil, :user_id => nil, :username => nil)
+      .update_all(:username => nil, :remote_address => nil, :user_id => nil)
+    puts "Successfully anonymized #{count} audits!"
+  end
+
+  def get_audits
+    Audited::Audit.up_until(before_date)
+  end
+
+  def before_date
+    @before_date ||= ENV['days'] ? ENV['days'].to_i.days.ago : 90.days.ago
   end
 end


timogoebel:
ok to test

lib/tasks/audits.rake
```diff
@@ -21,4 +22,10 @@ namespace :audits do
     count = audits.delete_all
     puts "Successfully deleted #{count} audits!"
   end
+  task :anonymize => :environment do
```
ares: mind to keep extra blank line between task definitions?

ares (changes requested):
This is a great addition, I have only one small suggestions, please take a look.

juliantodt:
@ares Thanks for the quick review, is done.

lib/tasks/audits.rake
```diff
@@ -21,4 +22,11 @@ namespace :audits do
     count = audits.delete_all
     puts "Successfully deleted #{count} audits!"
   end
+
+  task :anonymize => :environment do
+    before = ENV['days'].to_i.days.ago if ENV['days']
+    before ||= 90.days.ago
+    count = Audited::Audit.up_until(before).update_all(:username => nil, :remote_address => nil, :user_id => nil)
```
timogoebel: I think we need to change this to ignore already anonymized records.

timogoebel (changes requested):
@juliantodt: Thanks, while testing I found one small issue. Mind taking a look? Thanks.

lzap:
I think we need a simple unit/view test - does anonymized audit record show up in the UI correctly? I smell nil exceptions :-) Thanks for the patch!

juliantodt:
Added the changes @timogoebel suggested.
@lzap The current [tests](https://github.com/theforeman/foreman/blob/develop/test/controllers/audits_controller_test.rb) already cover anonymised audits because neither FactoryBot nor the Fixtures include user-info in their audits, so there should be no need to add additional tests.

lib/tasks/audits.rake
```diff
@@ -21,4 +22,12 @@ namespace :audits do
     count = audits.delete_all
     puts "Successfully deleted #{count} audits!"
   end
+
+  task :anonymize => :environment do
+    before = ENV['days'].to_i.days.ago if ENV['days']
+    before ||= 90.days.ago
+    count = Audit.where('created_at <= ?', before).where.not(:remote_address => nil, :user_id => nil, :username => nil)
```
tbrisker: You can use the `Audit.up_until` scope here like in the other task. Perhaps combine the first part of the tasks to a function that finds the audits and then the task itself only has the difference between the two. 
Also, might be faster to skip the additional condition since that will require a possibly slow lookup to find all audits that don't match the condition yet and just nullify everything older then the date - but it might be a good idea to benchmark which is faster.

tbrisker (changes requested):

lib/tasks/audits.rake
timogoebel: The additional conditions are required so the success message is displayed correctly.

juliantodt: It is slightly faster with the additional condition. At 1.5 million audits up_until(before) with 4000 to anonymise, it took 3.7s with the condition and 76.6s without. This obviously will be different when you have to anonymise a higher percentage of the audits up_until(before), but if you run the task regularly then this implementation should be advantages.
I will refactor the tasks like you suggested.

juliantodt:
Implemented the changes @tbrisker suggested.

tbrisker (dismissed):

timogoebel (dismissed):
LGTM, thanks @juliantodt.

timogoebel:
@ares: Anything to add?

lib/tasks/audits.rake
```diff
@@ -1,24 +1,35 @@
...
 END_DESC
 
 namespace :audits do
   task :expire => :environment do
-    before = ENV['days'].to_i.days.ago if ENV['days']
-    before ||= 90.days.ago
-    audits = Audited::Audit.up_until(before)
-    puts "Deleting audits older than #{before}. This might take a few minutes..."
+    audits = get_audits
+    puts "Deleting old audits. This might take a few minutes..."
```
ares: I think it would be good to keep the date in message, how about extracting `before = ENV['days'].to_i.days.ago if ENV['days']` to a method with memoization, something like


```ruby
def before_date
  @before_date ||= ENV['days'].to_i.days.ago || 90.days.ago
end
```

that way you can still use it in this message and perhaps even in anonymize task

ares (changes requested):
still one small thing

juliantodt:
Added @ares suggestion.

ares (approved):
Works fine now, thanks, merging! It would be great if you could join our [community demo](http://projects.theforeman.org/projects/foreman/wiki/Current_Sprint_Information) and present this. If you can't make it, would you mind if someone else covers it?

juliantodt:
@ares I'm not really interested in demoing it, sorry, but if someone else wants to, feel free to do so.

tbrisker:
@juliantodt I invited you to the foreman organization so that your next PRs will automatically trigger CI (once you make your membership public)

timogoebel:
@tbrisker: Thanks, can you also make him a developer in Redmine so I can assign him tickets etc.? Thanks again. :-)

tbrisker:
@timogoebel done :)

lib/tasks/audits.rake
```diff
@@ -1,24 +1,38 @@
...
     count = audits.delete_all
     puts "Successfully deleted #{count} audits!"
   end
+
+  task :anonymize => :environment do
+    puts "Anonymizing audits older than #{before_date}. This might take a few minutes..."
+    count = get_audits.where.not(:remote_address => nil, :user_id => nil, :username => nil)
+      .update_all(:username => nil, :remote_address => nil, :user_id => nil)
+    puts "Successfully anonymized #{count} audits!"
+  end
```
lzap, line 29: I know I am late to the party, but want to doublecheck - are we calling this task from cron by default?

If so, audits will be under heavy load on larger deployments and this will exclusive lock it. Can be causing cancelled transactions. Let's do nothing for now, but keep in mind to break this into smaller chunks when this is a problem.

ares: We do not run it from cron and there was some testing with bigger amount of audits. It should be reasonably fast as it skips the records which are already anonymized and does `update_all` which renders single SQL query. So running it from cron e.g. every day would anonymize only audits created during one day.

PR #5394, opened 2018-04-03
Fixes #23100 - Simplify gem requirements syntax

ekohl:
This changes to an equivalent syntax that's slightly easier to handle in a script that I'm working on to generate RPM dependencies.


lzap (approved):
Thanks.

PR #5395, opened 2018-04-03
Fixes #23098 - Allow searching on all audited models by type

tbrisker:
This automatically adds all the audited models in an after_initilize
block so that models defined in plugins are also included. All the
search related definitions for audit have been extracted to a seperate
concern. Models using STI require special handling until audits are
fixed to properly work with STI.


app/models/concerns/audit_search.rb
```diff
@@ -0,0 +1,73 @@
...
+      )
+    end
+
+    private
+
+    #This is a workaround needed until we get proper STI auditing
+    def audited_classes_without_sti
+      audited_class_names.reject do |name|
+        klass = name.constantize
+        name.include?(':') ||
```
ares: wouldn't this reject all namespaced classes, e.g. `ForemanOpenscap::ArfReport` or `Katello::HostCollection`? Also if this should remain, `::` would be more descriptive I think.

ares (changes requested):
one question, otherwise looks good, I'll continue with testing meanwhile

app/models/concerns/audit_search.rb
tbrisker: Good point, i'll fix this.

ares (approved):
Thanks, works nicely! I'll wait for Mr. Jenkins but good to be merged from my point of view.

ares:
Thanks @tbrisker, merging!

PR #5396, opened 2018-04-03
Fixes #22855 - Add full breadcrumbs functionality

amirfefer:
Todo:
- [x] Switcher is broken under puppetclasses
- [x]  Replace the “Loading…” with a loading animation
- [x]  Error handling in the popover
- [x] Add pagination ability
- [x] finish tests


app/helpers/layout_helper.rb
```diff
@@ -17,12 +18,28 @@ def search_bar(*elements)
...
+
+    breadcrumb_bar_props = {
+      isSwitchable: consumer_override_params[:isSwitchable] || breadcrumb_switchable?,
+      breadcrumbItems: consumer_override_params[:breadcrumbItems] || breadcrumb_items
+    }
+
+    breadcrumb_bar_props[:resource] = {
+      switcherItemUrl:  consumer_override_params[:switcherItemUrl] || switcher_url_template,
+      reosurceUrl: consumer_override_params[:reosurceUrl] || "/api/v2/#{controller_name}?thin=true",
+      nameField: consumer_override_params[:nameField] || controller_name.camelize.singularize.constantize.try(:title_name) || 'name',
```
tstrachota: `controller_name.camelize.singularize.constantize` can cause troubles to plugins that have their controllers namespaced. E.g.:
```
NameError in ForemanVirtWhoConfigure::Configs#show
uninitialized constant ForemanVirtWhoConfigureConfig
```
We should definitely gracefully fallback to not displaying the switcher when we can't guess the name.

app/helpers/layout_helper.rb
```diff
@@ -1,3 +1,4 @@
+require 'pry'
```
tstrachota: Forgotten pry

tstrachota (changes requested):
This is a nice start @amirfefer, thanks for sharing the PR early. I haven't looked at the code yet, but I tested the functionality and found several places where we need to tune the switcher.

One of my concerns is that we shouldn't be removing the js code that handles two-panes yet, because some plugins are still using it (I found at least foreman tasks that are broken with this PR). We should give them some time to adopt the switcher.

I think that we need to work a bit more on the switcher's styling. Currently it's very different from the patternfly definition (and the switcher in katello). Not having the search field is ok for now, but the colors and spacing should be adjusted.
foreman_switcher
vs.
patternfly_switcher
vs.
katello_switcher



**Some other issues found:**
- [ ] Switcher contents on audit page takes ages to load. This will probably improve with adding pagination but I don't think it makes much sense to display switcher there anyway so I vote for removing it there.
- [ ] Links in the switcher doesn't work on the proxies page at all. It could be an issue with katello code that we load on that page when there's pulp feature enabled.
- [ ] The switcher performs an api call every time the button is clicked - shouldn't we cache it?
- [ ] In hostgroups it would be better to show full labels instead of names to avoid confusion when there are nested groups with the same name.
- [ ] Puppet classes freeze in the "Loading..." state even though the server response is returned
  (throws error `TypeError: data.results.map is not a function`).
- [ ] Smart class params have empty switcher (probably wrong url, server returns 404).
- [ ] All empty pages throw:
```
  Failed prop type: The prop `data.breadcrumbItems[0].caption` is marked as required in `BreadcrumbBar`, but its value is `null`.
  Failed prop type: The prop `items[0].caption` is marked as required in `Breadcrumb`, but its value is `null`.
```
- [ ] Operating systems should display labels with versions to avoid having multiple items with the same name.
- [ ] OSs also throw: `Encountered two children with the same key, `id-CentOS`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version` when there are OSs with the same name.
- [ ] Partition tables and provisioning templates have wrong urls, clicking on an item results in 404
- [ ] On trends page the switcher is blank (api call returns 404)
- [ ] The switcher on roles page is displaying also the locked records that we don't allow to edit
user_roles
- [ ] There's a line under a heading on pages that don't have breadcrumbs. That should be removed.
line_below_heading


All this notes are imho blockers that need to be solved prior to merging the PR.

waldenraines:
Nice review @tstrachota, I don't have much to add except that I second that we can add search later.  As for this:

> Replace the “Loading…” with a loading animation

You can just use [`<Spinner>`](https://rawgit.com/patternfly/patternfly-react/gh-pages/index.html?knob-Loading=true&knob-Inline=false&knob-Inverse=false&knob-Size=md&selectedKind=Widgets&selectedStory=Spinner&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs) from patternfly-react.

amirfefer:
@tstrachota Thanks for your great review !
We managed to handle most of your concerns 
could you have another look?

app/controllers/api/v2/roles_controller.rb
```diff
@@ -12,7 +12,7 @@ class RolesController < V2::BaseController
 
       def index
         params[:order] ||= 'name'
-        @roles = resource_scope_for_index
+        @roles = params[:thin] ? resource_scope.search_for(*search_options).select {|role| !role.locked?} : resource_scope_for_index
```
tstrachota: We already use `thin` parameter in hosts controller for listing only ids and names. We should be consistent in parameter names. Could you either rename the parameter (how about `only_editable`?) or use scoped search in the switcher request?

app/helpers/layout_helper.rb
```diff
@@ -179,4 +195,17 @@ def per_page(collection)
...
+    ['edit', 'show'].include? action_name
+  end
+
+  def switcher_url_template
+    actual_action_name = (action_name == 'show') ? '' : action_name
+    "/#{controller_name}/:id/#{actual_action_name}"
+  end
+
+  def model_name_field
+    controller_name.camelize.singularize.try(:constantize).try(:title_name)
```
tstrachota: This is still causing problems in plugins that use nested controllers:
```
NameError in ForemanVirtWhoConfigure::Configs#show
uninitialized constant ForemanVirtWhoConfigureConfig
```

webpack/assets/javascripts/react_app/common/colors.scss
```diff
@@ -1,7 +1,12 @@
...
 $pf-red-200: #a30000;
 $pf-red-100: #c00;
 $pf-orange-400: #ec7a08;
 $pf-gold-400: #f0ab00;
 $pf-green-400: #3f9c35;
 $pf-green-600: #1e4f18;
+
+$switcher-max-height: 300px;
+$switcher-max-width: 200px;
+$switcher-min-width: 200px;
```
tstrachota, line 12: Could you move it to some different file, please? `colors.scss` doesn't seem like a good place for storing the switchers dimensions.

app/helpers/layout_helper.rb
```diff
@@ -17,12 +17,28 @@ def search_bar(*elements)
   end
 
   def mount_breadcrumbs(&block)
-    index_page = {caption: _(controller_name.humanize), url: try("#{controller_name}_path")}
-    default_menu = [(index_page unless action_name == 'index'),
-                    {caption: @page_header, url: '#' }].compact unless block_given?
+    consumer_override_params = block_given? ? yield : {}
```
tstrachota: Nitpick: Why are we actually using block when we expect it to return a hash? Could the method take the hash as a parameter?

app/helpers/layout_helper.rb
```diff
@@ -17,12 +17,28 @@ def search_bar(*elements)
...
 
-    mount_react_component("Breadcrumb", "#breadcrumb",
-      { menu: default_menu || yield }.to_json)
+    breadcrumb_index_item = {caption: _(controller_name.humanize), url: try("#{controller_name}_path")}
```
tstrachota: Unfortunately this translation is not going to work. Gettext can't guess what's in `controller_name.humanize`. If we have the humanized controller name as a static string somewhere else in the code, then you could use `N_()`.

app/helpers/layout_helper.rb
```diff
@@ -17,12 +17,28 @@ def search_bar(*elements)
...
+    ].compact
+
+    breadcrumb_bar_props = {
+      isSwitchable: consumer_override_params[:isSwitchable].nil? ? breadcrumb_switchable? : consumer_override_params[:isSwitchable],
+      breadcrumbItems: consumer_override_params[:breadcrumbItems] || breadcrumb_items
+    }
+
+    breadcrumb_bar_props[:resource] = {
+      switcherItemUrl:  consumer_override_params[:switcherItemUrl] || switcher_url_template,
+      reosurceUrl: consumer_override_params[:reosurceUrl] || "/api/v2/#{controller_name}?thin=true",
```
tstrachota: Not all controllers support `thin`. Could you rather override on a case by case basis?

tstrachota (changes requested):
Thank you for your updates @amirfefer. Good portion of the issues has been fixed. I also appreciate the tests that you added.

Some of the issues still persist though. So far I looked only into the ruby code, js is still to be reviewed.

**Some new issues found:**
- The switcher is still visually very different from the PF design. Only arrows without text should be used for pagination to match the current Katello switcher.
foreman_switcher_evo2
- ~~Pagination on the roles page is showing me only 2 items per page~~ (switcher disabled now)
- ~~Puppet classes show only one item per page~~
- Pagination uses the per_page setting - it should probably be a separate value. I'm ok with fixing it later if we file an issue for that.

**Old issues that still persist:**
- All empty pages throw:
```
  Failed prop type: The prop `data.breadcrumbItems[0].caption` is marked as required in `BreadcrumbBar`, but its value is `null`.
  Failed prop type: The prop `items[0].caption` is marked as required in `Breadcrumb`, but its value is `null`.
```
- ~~Smart class params have empty switcher (probably wrong url, server returns 404).~~
- ~~Caching is missing. Even worse, each click to the switcher button now triggers two same server requests.~~
- Links in the switcher doesn't work on the proxies page at all. It could be an issue with katello code that we load on that page when there's pulp feature enabled.

webpack/assets/javascripts/react_app/common/colors.scss
sharvit: I prefer to keep it as it so it will be easier for me to do the rebase here later https://github.com/theforeman/foreman/pull/5212

I am trying to fully address this issue in this #5212

app/controllers/api/v2/roles_controller.rb
ohadlevy: ideally, these should be available via search, and then your query can be simple as /api/roles?search='locked=false'.
@ares - how hard would it be to add a search for locked=false?

app/controllers/application_controller.rb
```diff
@@ -23,7 +23,6 @@ class ApplicationController < ActionController::Base
   before_action :welcome, :only => :index, :unless => :api_request?
   prepend_before_action :allow_webpack, if: -> { Rails.configuration.webpack.dev_server.enabled }
   around_action :set_timezone
-  layout :display_layout?
```
ohadlevy: if you are keeping two pane with deprecation warnings, please keep all of the two pane code.

app/helpers/layout_helper.rb
ohadlevy: I personally like the breadcrumbs do end syntax, I guess I'm fine with it either way.

app/models/hostgroup.rb
```diff
@@ -109,6 +109,10 @@ def hostgroup
     self
   end
 
+  def self.title_name
+    "title"
```
ohadlevy: maybe use a symbol or freeze the string?

app/controllers/api/v2/roles_controller.rb
ares: As easy as adding following line to Role model

```ruby
scoped_search :on => :locked, :complete_value => { :true => true, :false => false }
```

app/views/audits/show.html.erb
```diff
@@ -1,5 +1,10 @@
-<% title "#{@audit.action.camelize} #{audited_type @audit}: #{audit_title @audit}" %>
 
+  <% breadcrumbs do
+       { isSwitchable: false, breadcrumbItems: [
+         {caption: _("Audits"), url: (url_for(audits_path) if authorized_for(hash_for_hosts_path))},
```
ohadlevy: is this a bug in the current usage too? if thats true consider doing a small pr just for this change.

app/views/ptables/edit.html.erb
```diff
@@ -1,3 +1,5 @@
 <% title _("Edit %s") % @template.to_s %>
-
+<%= breadcrumbs do
+      {switcherItemUrl: "/templates/ptables/:id/edit" }
```
ohadlevy: nitpick: incorrect spacing.

app/views/trends/_empty_data.html.erb
```diff
@@ -1,4 +1,7 @@
 <% title(_("Trends for %s") % trend_title(@trend)) %>
+<%= breadcrumbs do
+      {isSwitchable: false}
```
ohadlevy: nitpick: incorrect spacing.

app/helpers/layout_helper.rb
ohadlevy: I would also suggest to use the same mechanizim we use for menu items, which is using rails routes vs hard coding urls.

ohadlevy (changes requested):
very quick review of the ruby code, will add more comments in follow up reviews.

app/helpers/layout_helper.rb
```diff
@@ -17,12 +17,28 @@ def search_bar(*elements)
...
-    default_menu = [(index_page unless action_name == 'index'),
-                    {caption: @page_header, url: '#' }].compact unless block_given?
+    consumer_override_params = block_given? ? yield : {}
 
-    mount_react_component("Breadcrumb", "#breadcrumb",
-      { menu: default_menu || yield }.to_json)
+    breadcrumb_index_item = {caption: _(controller_name.humanize), url: try("#{controller_name}_path")}
+    breadcrumb_page_item = {caption: @page_header, url: '#' }
+
+    breadcrumb_items = [
```
ohadlevy: IMHO the entire breadcrums logic should move to another file, maybe even to a class so its easier to test.

app/helpers/layout_helper.rb
ohadlevy: Actually, after playing with it a bit, I understand your concern, using hashes does feel natural to me as well.

app/controllers/api/v2/roles_controller.rb
ohadlevy: are you sure? its not in the schema?

webpack/assets/javascripts/react_app/components/BreadcrumbBar/components/BreadcrumbSwitcherPopover.js
```diff
@@ -0,0 +1,77 @@
...
+  currentPage,
+  totalPages,
+  ...props
+}) => {
+  let popoverBody;
+
+  if (loading) {
+    popoverBody = (
+      <div className="breadcrumb-switcher-popover-loading text-center">
+        <Icon name="spinner" spin size="lg" />
```
ohadlevy, line 26: why not use the spinner component?

webpack/assets/javascripts/react_app/components/BreadcrumbBar/components/BreadcrumbSwitcherPopover.js
```diff
@@ -0,0 +1,77 @@
...
+      </div>
+    );
+  } else if (hasError) {
+    popoverBody = (<div className="breadcrumb-switcher-popover-error">{__('Error: Unable to load resources.')}</div>);
```
ohadlevy: nitpick: doesnt look as great, maybe we can consider using the same messagebox we have in statistics that handle failure with a small icon etc.

app/views/config_reports/index.html.erb
```diff
@@ -1,12 +1,12 @@
 <%= if @host
   breadcrumbs do
-    [
+    { breadcrumbItems: [
       {caption: _("Hosts"), url: (url_for(hosts_path) if authorized_for(hash_for_hosts_path))},
       {caption: @host.name, url: (host_path(@host) if authorized_for(hash_for_host_path(@host)))},
       {caption: _('Reports'), url: url_for(config_reports_path(search: 'eventful=true'))},
-    ]
```
ohadlevy: this has no title in the breadcrums, it seems like api for reports index is not ideal as well, maybe we should disable it on this page for now.

waldenraines:
I have no further comments beyond what has already been pointed out above.  Thanks for the thorough reviews @tstrachota and @ohadlevy.

app/controllers/api/v2/roles_controller.rb
ohadlevy: @ares - any other workaround?

app/assets/javascripts/application.js
```diff
@@ -6,7 +6,6 @@
 //= require scoped_search
 //= require bootstrap
 //= require patternfly
-//= require two-pane
```
ohadlevy: this change should be reverted, as it breaks other pages/plugins that would need to require it themselves.

app/models/role.rb
```diff
@@ -76,6 +76,8 @@ class Role < ApplicationRecord
   scoped_search :on => :builtin, :complete_value => { :true => true, :false => false }
   scoped_search :on => :description, :complete_value => false
   scoped_search :relation => :permissions, :on => :name, :complete_value => true, :rename => :permission, :only_explicit => true, :ext_method => :search_by_permission
+  # TODO: does not really work yet.
+  scoped_search :on => :locked, :complete_value => { :true => true, :false => false }
```
ohadlevy: if are we are going to disable roles until this works, we can probably remove this line.

amirfefer: remvoed :)

ohadlevy: I can still see it here? :) also make sure you disable the switcher in the breadcrumbs options?

app/helpers/audits_helper.rb
```diff
@@ -191,6 +191,28 @@ def audit_details(audit)
     "#{audit_user(audit)} #{audit_remote_address audit} #{audit_action_name audit} #{audited_type audit}: #{link_to(audit_title(audit), audit_path(audit))}".html_safe
   end
 
+  def nested_breadcrumbs
```
ohadlevy: this is probably my fault, but all helpers are included in every view (not sure why tbh) so, probably best to call it nested_host_audit_breadcrumbs or some other thing unique.

app/assets/javascripts/two-pane.js
```diff
@@ -186,4 +187,4 @@ function handle_redirect(response){
     window.location.replace(redirect);
   }
   return redirect
-}
+}
```
ekohl: Could you keep in the newline here?

app/views/compute_attributes/edit.html.erb
```diff
@@ -1,9 +1,24 @@
...
+      },
+      {
+        caption: _("Compute Profiles"),
+        url: (url_for(compute_profiles_path) if authorized_for(hash_for_compute_profiles_path))
+      },
+      {
+        caption: _('Edit %s' % @set.compute_profile.to_label)
+      }
+    ],
+    resource_url: "/api/v2/compute_resources/#{@set.compute_resource.id}/compute_profiles/",
```
ekohl: Could `url_for` be used here?

app/views/ssh_keys/new.html.erb
```diff
@@ -1,7 +1,20 @@
...
+         url: (url_for(users_path) if authorized_for(hash_for_users_path))
+      },
+      {
+        caption: @user.name,
+        url: (edit_user_path(@user) if authorized_for(hash_for_edit_user_path(@user)))
+      },
+      {
+        caption: _('Add SSH Key')}
+    ],
+    resource_url: "/api/v2/users/",
```
ekohl: Could `url_for` be used here?

ekohl (changes requested):
In a lot of places you used `breadcrum` instead of `breadcrumb`. I also noticed you hardcoded a lot of URLs rather than using a helper. I think this will break remounting the whole application and is generally a bad practice.

app/helpers/images_helper.rb
```diff
@@ -10,4 +10,28 @@ def image_field(f, opts = {})
...
+        {
+          caption: _('Images'),
+          url: url_for(compute_resource_images_path(@compute_resource))
+        },
+        {
+          caption: ((action == 'new') ? _('Create image') : _("Edit %s") % @image)
+        }
+      ],
+      resource_url: url_for(api_compute_resource_images_path(@compute_resource)),
+      switcher_item_url: 'TODO'
```
ohadlevy: should we disable it for now?

app/services/breadcrumbs_options.rb
```diff
@@ -0,0 +1,71 @@
+class BreadcrumbsOptions
+  #TODO: extract url handing from menu/item and use it here too
```
ohadlevy: @amirfefer can you create a follow up redmine issue for this? I believe this will save a lot of pain for plugins after this is merged in core.

tstrachota:
I wanted to re-review but the current version seems to be broken. On every page that I tested the switcher keeps spinning even though the api request was successful. There's an error:
```
BreadcrumbBarActions.js:37 Uncaught (in promise) TypeError: (0 , _flatten4.default)(...).map is not a function
    at formatResults (BreadcrumbBarActions.js:37)
    at onRequestSuccess (BreadcrumbBarActions.js:31)
```

Empty pages report some warnings too:
```
warning.js:33 Warning: Failed prop type: The prop `data.breadcrumbItems[0].caption` is marked as required in `BreadcrumbBar`, but its value is `null`.
    in BreadcrumbBar (created by Connect(BreadcrumbBar))
    in Connect(BreadcrumbBar)
    in Provider
```
```
Warning: Failed prop type: The prop `items[0].caption` is marked as required in `Breadcrumb`, but its value is `null`.
    in Breadcrumb (created by BreadcrumbBar)
    in BreadcrumbBar (created by Connect(BreadcrumbBar))
    in Connect(BreadcrumbBar)
    in Provider
```
@amirfefer could you take a look, please?

tstrachota:
Ooops, that was issue on my side. Please disregard, I'm taking another look.

tstrachota:
Thank you for fixing another portion of issues. I appreciate the efforts you're putting into this. I updated my previous comment and crossed those that are fixed already: https://github.com/theforeman/foreman/pull/5396#pullrequestreview-109471542.

There are still some remaining, that I consider blockers:
- The switcher is still visually different from the PF design and the Katello one (see https://github.com/theforeman/foreman/pull/5396#pullrequestreview-109257666).
- There are warnings on empty pages.
- Switcher links doesn't work on the proxy page when Katello is installed - I'm ok with disabling the switcher there temporarily.

I found out that the switcher is disabled on the roles page - was that intentional?

app/views/compute_attributes/edit.html.erb
```diff
@@ -1,9 +1,24 @@
...
+      {
+        caption: _("Compute Profiles"),
+        url: (url_for(compute_profiles_path) if authorized_for(hash_for_compute_profiles_path))
+      },
+      {
+        caption: _('Edit %s' % @set.compute_profile.to_label)
+      }
+    ],
+    resource_url:api_compute_resource_compute_profiles_path(@set.compute_resource),
+    switcher_item_url: "/compute_profiles/#{@set.compute_profile.id}/compute_attributes/:id/edit"
```
tstrachota: How about using `edit_compute_profile_compute_attribute_path(@set.compute_profile.to_param, ':id')` here?

app/views/hosts/edit.html.erb
```diff
@@ -1,4 +1,5 @@
 <% title(_("Edit %s") % @host) %>
+<% host_breadcrumb_options %>
```
tstrachota: Since this method directly mounts breadcrumbs, could you please rename it just to `host_breadcrumbs`, please?

app/views/hosts/show.html.erb
```diff
@@ -1,5 +1,6 @@
 <% javascript 'charts', 'hosts' %>
 <% title @host.to_label, icon(@host.operatingsystem) + @host.to_label %>
+<% host_breadcrumb_options %>
```
tstrachota: Ditto renaming.

app/views/images/edit.html.erb
```diff
@@ -1,9 +1,3 @@
-<%= breadcrumbs do
-      [
-        {caption: _("Compute Resources"), url: (url_for(compute_resources_path) if authorized_for(hash_for_compute_resources_path))},
-        {caption: _("#{@compute_resource.to_s}"), url: (url_for(compute_resource_path(@compute_resource)) if authorized_for(hash_for_compute_resources_path))},
-        {caption: _("Images"), url: (url_for(compute_resource_images_path(@compute_resource)) )},
-        {caption: _("Edit %s") % @image}
-      ]
-    end %>
+<%= image_breadcrumb_options %>
```
tstrachota: Ditto renaming.

app/views/provisioning_templates/edit.html.erb
```diff
@@ -1,3 +1,3 @@
 <% title _("Edit %s") % @template.to_s %>
-
+<%= breadcrumbs(switcher_item_url: "/templates/provisioning_templates/:id/edit" ) %>
```
tstrachota: Could you use `edit_provisioning_template_path(':id')` here?

app/views/ptables/edit.html.erb
```diff
@@ -1,3 +1,3 @@
 <% title _("Edit %s") % @template.to_s %>
-
+<%= breadcrumbs(switcher_item_url: "/templates/ptables/:id/edit") %>
```
tstrachota: Could be replaced with `*_path` helper too.

app/views/ssh_keys/new.html.erb
```diff
@@ -1,7 +1,20 @@
...
+      {
+        caption: @user.name,
+        url: (edit_user_path(@user) if authorized_for(hash_for_edit_user_path(@user)))
+      },
+      {
+        caption: _('Add SSH Key')}
+    ],
+    resource_url: api_users_path,
+    switchable: true,
+    switcher_item_url: "/users/:id/ssh_keys/new" }
```
tstrachota: `new_user_ssh_key_path`

app/views/variable_lookup_keys/edit.html.erb
```diff
@@ -1,2 +1,17 @@
...
+        caption: _("Smart variables"),
+        url: (url_for(variable_lookup_keys_path) if authorized_for(hash_for_variable_lookup_keys_path))
+      },
+      {
+        caption: _('Edit Smart Variable')
+      }
+    ],
+    resource_url: "/api/v2/smart_variables/",
+    switchable: true,
+    switcher_item_url: "/variable_lookup_keys/:id/edit"
```
tstrachota: Ditto path helper.

tstrachota (changes requested):
I agree with @ekohl, could you please replace the hardcoded urls with path helpers, please?

app/helpers/images_helper.rb
tstrachota: +1, either add the path or disable switcher, please.

app/views/variable_lookup_keys/edit.html.erb
ohadlevy: I'm not sure if you can use an helper for the switcher_item_url, as the id is later on being replaced by javascript code.. I would think the url helpers will fail without an AR object?

tstrachota: You can do it, see my first comment: https://github.com/theforeman/foreman/pull/5396#discussion_r180413214
The js code replaces `:id` with the actual value. When you pass `":id"`to the helper instead of the AR object, it generates `"/variable_lookup_keys/:id/edit"` (I tested it).

app/views/compute_attributes/edit.html.erb
sharvit: @ohadlevy, @amirfefer and I have discussed about it.
Basically we agree, but because we are in a rush, and we really want to it to get merged, we will create a smaller PRs later in order to fix it.

app/views/provisioning_templates/edit.html.erb
sharvit: ^ we will create a smaller PRs later in order to fix it.

app/views/ptables/edit.html.erb
sharvit: ^ we will create a smaller PRs later in order to fix it.

amirfefer:
[test foreman]

app/views/ssh_keys/new.html.erb
sharvit: ^ we will create a smaller PRs later in order to fix it.

app/views/variable_lookup_keys/edit.html.erb
sharvit: ^ we will create a smaller PRs later in order to fix it.

tstrachota:
Thank you for the updates, I'm going to do another round of review.

tstrachota:
Warnings on empty pages are fixed, the other issues mentioned in [my previous comment](https://github.com/theforeman/foreman/pull/5396#issuecomment-380086924) remain with a slight difference - the proxy switcher now also throws some warnings. I do think that it's ok to disable it now and fix it later.

> I found out that the switcher is disabled on the roles page - was that intentional?

@amirfefer based on what I saw in the code I take silence as yes :)

tstrachota:
Two more findings:
- The current version actually report warnings `Invalid prop 'resourceSwitcherItems[0].id' of type 'number' supplied to 'BreadcrumbBar', expected 'string'.` on every page, not just smart proxies.
- Switcher on the compute profiles page switches to wrong pages. Form a user's perspective it may seem random, but in fact the issue is that we're using compute profile ids instead of compute attribute ids in the `:id` replacement. One solution could be adding `idField` setting to the breadcrumbs switcher and enhance the api response with `compute_attribute_id`.

tstrachota:
Btw replaced the paths with path helpers here: https://github.com/tstrachota/foreman/commit/e574db65df913bfbd09b0a01a98d991c15411b67 You can cherry-pick it.

test/integration/breadcrumbs_switcher_test.rb
```diff
@@ -0,0 +1,15 @@
+require 'integration_test_helper'
+
+class BreadcrumbsSwitcherTest < IntegrationTestWithJavascript
+  test "should switch between resourcres" do
```
ohadlevy: nitpick: should be "should switch between resources"

ohadlevy (dismissed):

app/services/breadcrumbs_options.rb
```diff
@@ -0,0 +1,71 @@
...
+  def switcher_url_template
+    actual_action_name = (action_name == 'show') ? '' : action_name
+    "/#{controller_name}/:id/#{actual_action_name}"
+  end
+
+  def model_name_field
+    controller_name.camelize.singularize.try(:constantize).try(:title_name)
+  rescue => err
+    # TODO: better handling for plugin detection
+    logger.warn("unable to detect title name in breadcrums for #{controller_name}, defaulting to name")
```
ekohl: breadcrums -> breadcrumbs

ekohl (dismissed):
From the packaging side of it it looks good. It'd be nice if the typo was fixed though.

ohadlevy (approved):
tests has passed + I've fixed spell issues. not waiting for another round of tests - thanks @amirfefer @sharvit and @tstrachota for your hard work on this!

boaz0:
**Congrats!**

happyhappyjoyjoy

PR #5400, opened 2018-04-04
Fixes #23115 - session is used instead session_safe

lzap:
We removed "session_safe" flag, it's now called "session" and it is safe
by default. This passes it into proxy correctly instead generating it's
own id.


mmoll (approved):

mmoll:
merged, díky @lzap!

PR #5401, opened 2018-04-04
Fixes #23113 - Allow disabling of out of sync status

bastilian:
This allows plugins that make use of an origin to add a setting to disable out of sync status for their origin. This is useful for environments that do not regulraly run configurations, as it is with most Ansible environments.
The changes are to widgets and the out of sync status of a host itself.

Here is a screenshot of the Ansible dashboard Widgets out of sync still enabled:

screenshot_2018-04-04_12-27-26

With the `ansible_out_of_sync_disabled` enabled out of sync hosts become ok hosts:

screenshot_2018-04-04_12-35-10

Settings for Puppet are implemented as well and the PR introducing the Ansible setting can be found here: https://github.com/theforeman/foreman_ansible/pull/158

### DIFF
--- a/app/models/setting/puppet.rb
+++ b/app/models/setting/puppet.rb
@@ -2,6 +2,7 @@ class Setting::Puppet < Setting
   def self.default_settings
     [
       self.set('puppet_interval', N_("Puppet interval in minutes"), 30, N_('Puppet interval')),
+      self.set('puppet_out_of_sync_disabled', N_("Disable host configuration status turning to out of sync for %s after report does not arrive within configured interval") % 'Puppet', false, N_('%s out of sync disabled') % 'Puppet'),
       self.set('default_puppet_environment', N_("Foreman will default to this puppet environment if it cannot auto detect one"), "production", N_('Default Puppet environment'), nil, { :collection => Proc.new {Hash[Environment.all.map{|env| [env[:name], env[:name]]}]} }),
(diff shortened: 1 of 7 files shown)


bastilian:
Tests are WIP but would appreciate a first review. :)

app/models/setting/puppet.rb
```diff
@@ -2,6 +2,7 @@ class Setting::Puppet < Setting
   def self.default_settings
     [
       self.set('puppet_interval', N_("Puppet interval in minutes"), 30, N_('Puppet interval')),
+      self.set('puppet_out_of_sync_disabled', N_("Disable out of sync for %s") % 'Puppet', false, N_('%s out of sync disabled') % 'Puppet'),
```
ares: I'd enhance the description a bit, e.g. "Disable host configuration status turning to out of sync for %s after report does not arrive within configured interval", current version is hard to understand if you don't have the context.

ares (changes requested):
It works great, only one suggestion regarding settings description.

bastilian:
@ares Tests for `ConfigurationStatus#out_of_sync?` respecting origins out of sync settings via `out_of_sync_disabled?` are added and the description for the settings here and in https://github.com/theforeman/foreman_ansible/pull/158 are enhanced.

ares (approved):
:+1: let's wait for Mr. Jenkins to confirm

dLobatog (approved):
Jenkins approves, and the feature works, thanks @bastilian !

PR #5402, opened 2018-04-04
Fixes #23117 - ovirt_ostype host param selects type

lzap:
We currently have a "magic" code which select the OS Type in oVirt. It tries to detect OS name and version, but user has no option to opt-in for specific type.

This patch will add ovirt_ostype host parameter reading, when present it overrides the detection completely.

More context at:

https://community.theforeman.org/t/ovirt-provisioning-error/8698/2

### DIFF
--- a/app/models/compute_resources/foreman/model/ovirt.rb
+++ b/app/models/compute_resources/foreman/model/ovirt.rb
@@ -61,23 +61,29 @@ module Foreman::Model
 
     def determine_os_type(host)
       return nil unless host
+      return host.params['ovirt_ostype'] if host.params['ovirt_ostype']
       ret = "other_linux"
       return ret unless host.operatingsystem
       os_name = os_name_mapping(host)
       arch_name = arch_name_mapping(host)
 
-      best_match = available_operating_systems.select { |os| os[:name].present? }.max_by do |os|
+      available = available_operating_systems.select { |os| os[:name].present? }
+      best_matches = available.sort_by do |os|
         rating = 0.0
         if os[:name].include?(os_name)
           rating += 100
-          rating += (1.0 / os[:name].length) # prefer the shorter names a bit in case we have not found more important some specifics
+          # prefer the shorter names a bit in case we have not found more important some specifics
+          rating += (1.0 / os[:name].length)
+          # bonus for major or major_minor
           rating += 10 if os[:name].include?("#{os_name}_#{host.operatingsystem.major}")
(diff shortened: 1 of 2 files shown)


app/models/compute_resources/foreman/model/ovirt.rb
```diff
@@ -61,23 +61,30 @@ def supports_operating_systems?
...
         rating = 0.0
         if os[:name].include?(os_name)
           rating += 100
-          rating += (1.0/os[:name].length) # prefer the shorter names a bit in case we have not found more important some specifics
+          # prefer the shorter names a bit in case we have not found more important some specifics
+          rating += (1.0/os[:name].length)
+          # bonus for major or major_minor
           rating += 10 if os[:name].include?("#{os_name}_#{host.operatingsystem.major}")
+          rating += 5 if os[:name].include?("#{os_name}_#{host.operatingsystem.major}_#{host.operatingsystem.minor}")
+          rating += 5 if os[:name].include?("#{os_name}_#{host.operatingsystem.major}.#{host.operatingsystem.minor}")
```
dLobatog: Isn't this line a literal duplicate of the one above?

dLobatog (changes requested):
@lzap Just 1 question inline, thanks @lzap

app/models/compute_resources/foreman/model/ovirt.rb
lzap: There's dot instead underscore, but this is actually a leftover, ovirt seems to use `_` everywhere.

lzap:
One comment, our method depends on ordering of XML coming from oVirt API. If `ubuntu_12_10` comes before `ubuntu_12_04` then it's picked for OS named "Ubuntu 12.XX" where XX is anything. This is not nice, so I added an extra bonus in case minor version also matches.

mmoll:
@lzap please rebase :)
AFAICT, this is ready to merge?

lzap:
Why to rebase, tests passed and I see no conflicts. @dLobatog or @orrabin ? This is ready to merge.

mmoll:
> Why to rebase

Because the changes are getting a full test run with the code that changed in the meanwhile then.

lzap:
Sure, won't hurt me (but Jenkins :-) DONE

lzap:
Rebased @orrabin @dLobatog

dLobatog (approved):

dLobatog:
Thanks @lzap !

ares:
@lzap please send cherry-picks to 1.19-stable and 1.18-stable, the redmine issues is currently set to 1.18.1 target

PR #5403, opened 2018-04-04
Fixes #22809 - Turn VM attributes to hash

xprazak2:


db/migrate/20180403144853_convert_vm_attrs_to_hash.rb
```diff
@@ -0,0 +1,45 @@
...
+        if YAML.load(attributes).is_a? from
+          comp_attr.vm_attrs = send transform_method, attributes
+          comp_attr.save!
+        end
+      end
+    end
+  end
+
+  def to_h(attr)
+    attr.gsub(yml_params_hash, yml_hash).gsub(yml_params_obj, yml_hash)
```
ares: `sub` might be enough, especially on large strings, it would be faster

db/migrate/20180403144853_convert_vm_attrs_to_hash.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def to_params(attr)
+    attr.gsub(yml_hash, yml_params_obj)
```
ares: `sub` should be enough

db/migrate/20180403144853_convert_vm_attrs_to_hash.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def yml_params_hash
+    '!ruby/hash:ActionController::Parameters'
+  end
+
+  def yml_params_obj
+    '!ruby/object:ActionController::Parameters'
+  end
+
+  class ComputeAttributeHack < ApplicationRecord
```
ares: usually we call it Fake$Model in migrations, so it should be `FakeComputeAttribute`

ares (changes requested):
few minor code comments, looks good in general, though I had the same problem with:

* Nic::Base#compute_attributes
* Nic::Base#attrs
* Report#metrics

and I can imagine the same problem can have other models that serialize data

Taxonomy#ignore_types (Array)
ComputeResource#attrs (Hash)
LookupKeys#default_value
LookupValue#value

I think Audit, TablePreference and Widget were not affected so they can be ignored. But it would be great to fix all tables in this migration. Otherwise there's a risk we'd need to release 1.17.z because another report on another table.

ares:
I think with [this](https://github.com/xprazak2/foreman/pull/3) it fixes all problems, the https://github.com/theforeman/foreman/pull/5408 wouldn't be needed anymore

xprazak2:
Updated, I made a few changes. I no longer serialize/deserialize the column content, I just run `gsub` on it and dump it back. I also removed the check based on type, I just check if the string changed.

I haven't tested each model yet to see if it works for all.

db/migrate/20180403144853_convert_vm_attrs_to_hash.rb
```diff
@@ -0,0 +1,97 @@
...
+    say "Converting Lookup Keys"
+    transform_batch_columns(FakeLookupKey, ['default_value'], transform_method)
+    say "Converting Lookup Values"
+    transform_batch_columns(FakeLookupValue, ['value'], transform_method)
+    say "All conversions finished"
+  end
+
+  def transform_batch_columns(base, serialized_columns, transform_method)
+    base.unscoped.all.in_batches do |batch|
+      batch.each do |object|
```
tbrisker: you can use `find_each` instead of `in_batches { batch.each...`

db/migrate/20180403144853_convert_vm_attrs_to_hash.rb
tbrisker: serialized objects may be nested, so there can be more than one string

db/migrate/20180403144853_convert_vm_attrs_to_hash.rb
```diff
@@ -0,0 +1,97 @@
...
+    transform_batch_columns(FakeTaxonomy, ['ignore_types'], transform_method)
+    say "Converting Lookup Keys"
+    transform_batch_columns(FakeLookupKey, ['default_value'], transform_method)
+    say "Converting Lookup Values"
+    transform_batch_columns(FakeLookupValue, ['value'], transform_method)
+    say "All conversions finished"
+  end
+
+  def transform_batch_columns(base, serialized_columns, transform_method)
+    base.unscoped.all.in_batches do |batch|
```
tbrisker: you only need to `.select(serialized_columns)`, you don't actually care about all the other fields

db/migrate/20180403144853_convert_vm_attrs_to_hash.rb
```diff
@@ -0,0 +1,97 @@
...
+      batch.each do |object|
+        serialized_columns.each do |column|
+          attributes = object.send :[], column.to_sym
+          next if attributes.nil?
+          new_attributes = send(transform_method, attributes)
+          if new_attributes != attributes
+            object.send(:[]=, column, new_attributes)
+            object.save!
```
tbrisker: you can use `object.update_column` instead to update directly in the db without having to go through activerecord callbacks.

db/migrate/20180403144853_convert_vm_attrs_to_hash.rb
```diff
@@ -0,0 +1,97 @@
...
+    say "Converting Lookup Values"
+    transform_batch_columns(FakeLookupValue, ['value'], transform_method)
+    say "All conversions finished"
+  end
+
+  def transform_batch_columns(base, serialized_columns, transform_method)
+    base.unscoped.all.in_batches do |batch|
+      batch.each do |object|
+        serialized_columns.each do |column|
+          attributes = object.send :[], column.to_sym
```
tbrisker: you can use `read_attribute_before_type_cast` to get the value directly, without needing the fake classes. 
Also, why not pass the columns as symbols instead of converting the string to symbol for every instance?

db/migrate/20180403144853_convert_vm_attrs_to_hash.rb
```diff
@@ -0,0 +1,97 @@
...
+    say "Converting Nics"
+    transform_batch_columns(FakeNic, ['attrs', 'compute_attributes'], transform_method)
+    say "Converting Compute Attributes"
+    transform_batch_columns(FakeComputeAttribute, ['vm_attrs'], transform_method)
+    say "Converting Compute Resources"
+    transform_batch_columns(FakeComputeResource, ['attrs'], transform_method)
+    say "Converting Reports"
+    transform_batch_columns(FakeReport, ['metrics'], transform_method)
+    say "Converting Taxonomies"
+    transform_batch_columns(FakeTaxonomy, ['ignore_types'], transform_method)
```
tbrisker: Are you sure this is needed? If I'm not mistaken this is an array, not a hash.

db/migrate/20180403144853_convert_vm_attrs_to_hash.rb
```diff
@@ -0,0 +1,97 @@
...
+    say "Converting Lookup Keys"
+    transform_batch_columns(FakeLookupKey, ['default_value'], transform_method)
```
tbrisker: Perhaps this could be reduced to those that have override set and value type that is serialized (json, yaml...)?

db/migrate/20180403144853_convert_vm_attrs_to_hash.rb
```diff
@@ -0,0 +1,97 @@
...
+    say "Converting Lookup Values"
+    transform_batch_columns(FakeLookupValue, ['value'], transform_method)
```
tbrisker: ditto

db/migrate/20180403144853_convert_vm_attrs_to_hash.rb
```diff
@@ -0,0 +1,97 @@
...
+  def to_h(attr)
+    attr.gsub(yml_params_hash, yml_hash).gsub(yml_params_obj, yml_hash)
+  end
+
+  def to_params(attr)
+    attr.gsub(yml_hash, yml_params_hash)
+  end
+
+  def yml_hash
+    '!ruby/hash:ActiveSupport::HashWithIndifferentAccess'
```
tbrisker: why not a constant for this and the other strings?

tbrisker (changes requested):
See some comment inline. I think Audits may also need this change (for `audited_changes`).
Also, I'm very concerned about the performance here. some of these tables may have thousands (nics, lk, lv) or millions (reports) of records, leading to a migration that may take hours. Has any benchmarking been done to test how long this will take with a large db?

ares:
I don't think audited was needed as it is never created directly from params, audited gem construct the hash based on AR#changes hash afaik.

db/migrate/20180403144853_convert_vm_attrs_to_hash.rb
xprazak2: Why is it better to use constants?

db/migrate/20180403144853_convert_vm_attrs_to_hash.rb
xprazak2: I am no sure, [param_filter](https://github.com/theforeman/foreman/blob/develop/app/controllers/concerns/foreman/controller/parameters/organization.rb#L9) looks like we might not need it, but the array is serialized so there are no guarantees what is inside. @ares?

ares: yeah, it's always plain array, my mistake, I guess you can remove it (it shouldn't do any harm though, except for burned cycles)

db/migrate/20180403144853_convert_vm_attrs_to_hash.rb
xprazak2: If we skip those without override, how can we be sure the `default_value` is what we want it to be?

xprazak2:
Updated, removed the fake models, removed Taxonomies, I use `find_each`, `update_attribute` and `read_attribute_before_type_cast`. No performance testing was done yet, I am setting it up and will report how it went.

db/migrate/20180403144853_convert_vm_attrs_to_hash.rb
tbrisker: You can use `gsub!` to reduce the object allocations and data copying here (`gsub` creates a new string with the result, `gsub!` makes the change in place)

db/migrate/20180403144853_convert_vm_attrs_to_hash.rb
tbrisker: using constants reduces the number of function calls and possibly also object allocations.

tbrisker (changes requested):

db/migrate/20180403144853_convert_vm_attrs_to_hash.rb
```diff
@@ -0,0 +1,57 @@
...
+    transform_batch_columns(Report, [:metrics], transform_method)
+    say "Converting Lookup Keys"
+    transform_batch_columns(LookupKey, [:default_value], transform_method)
+    say "Converting Lookup Values"
+    transform_batch_columns(LookupValue, [:value], transform_method)
+    say "All conversions finished"
+  end
+
+  def transform_batch_columns(base, serialized_columns, transform_method)
+    base.unscoped.find_each  do |object|
```
tbrisker: Missing `.select(serialized_columns)`

ohadlevy:
should this fix (and similar like it) be available via a rake task? so we can let users run it outside of migration or even as an async task using active job?

ares:
> should this fix (and similar like it) be available via a rake task? so we can let users run it outside of migration or even as an async task using active job?

I think foreman-maintain would be the proper tool. But it seems the only concern so far is the reports table, I'd prefer all the rest remain in migration.

tbrisker:
I think it isn't worth the effort to address this in foreman-maintain or otherwise for reports. New reports will already be serialized properly in 1.17 thanks to https://github.com/theforeman/foreman/commit/408774ae15561e8386250bf6b15c39a74aaf3510#diff-3f97cbb8726f3310e0b0c386dbe89e22R55 - so over time, new reports will be serialized as hashes and old reports will presumably be deleted. All we need to do is make sure that `metrics` getter in reports can handle both cases properly.

ares:
> I think it isn't worth the effort to address this in foreman-maintain or otherwise for reports. New reports will already be serialized properly in 1.17 thanks to 408774a#diff-3f97cbb8726f3310e0b0c386dbe89e22R55 - so over time, new reports will be serialized as hashes and old reports will presumably be deleted. All we need to do is make sure that metrics getter in reports can handle both cases properly.

Ok, works for me. ConfigReport nor ArfReport override this method and I'm not aware of any other children.

tbrisker:
I think this should fix the reports deserializer to work with any stored version:
```diff
diff --git a/app/models/report.rb b/app/models/report.rb
index e258cebe2..1dad85d06 100644
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -49,7 +49,12 @@ class Report < ApplicationRecord
   # extracts serialized metrics and keep them as a hash_with_indifferent_access
   def metrics
     return {} if self[:metrics].nil?
-    YAML.load(self[:metrics]).with_indifferent_access
+    metrics = YAML.load(self[:metrics])
+    if metrics.is_a? ActionController::Parameters
+      metrics.permit!.to_h
+    else
+      metrics.with_indifferent_access
+    end
   end

File: db/migrate/20180403144853_convert_vm_attrs_to_hash.rb (hunk 12 above)
Comment (xprazak2) [reply db/migrate/20180403144853_convert_vm_attrs_to_hash.rb]: I can either use `find_each` or `select(serialized_columns)`, having both does not work.

**Comment by xprazak2** [comment]
Updated, I use `gsub!` and transformed method returning constant strings to constants. We need the fake classes, because `update_column` skips validations, but does not skip the serialization. 

30k compute attributes took 32 seconds and 100k reports took 2.5 min to go through when I was testing:

File: db/migrate/20180403144853_convert_vm_attrs_to_hash.rb (hunk 12 above)
Comment (tbrisker) [reply db/migrate/20180403144853_convert_vm_attrs_to_hash.rb]: you can use `.select(serialized_columns.append(:id))` with find_each

**Comment by ares** [comment]
So in 20k hosts setup with puppet interval set to 30 minutes and default cleanup after 7 days that makes 7 millions reports in DB, according to 100k test, it would take ~3 hours. While most of users don't have 20k hosts, I do agree with @tbrisker here that we can just modify the deserialization method. Since reports get cleaned automatically (by default after 7 days), it will autofix to problem after we start serializing correctly. It will most likely affect developers as they usually don't setup cron to clean old reports.

@tbrisker had also concerns re nics but given the numbers, I think we can keep it here, in worst case it would take few minutes with no change at the end

**Review by tbrisker — CHANGES_REQUESTED** [review]
Looks like your timings confirm my expectation - migrating millions of reports will lead to a migration that can take hours, so I suggest not migrating reports and enhancing the report metrics accessor so it can handle both formats.
Additionally, looks like in some cases we may have `!ruby/hash-with-ivars:ActionController::Parameters` in the database, so we'll also need to convert that/

**Comment by xprazak2** [comment]
I removed the reports from migration and modified how we deserialize metrics.

**Comment by tbrisker** [comment]
Are we ceratin that we are now serializing all of these fields properly and won't end up with new AC::Params being stored?

File: db/migrate/20180403144853_convert_vm_attrs_to_hash.rb [hunk 13]
```diff
@@ -0,0 +1,80 @@
...
+    self.inheritance_column = nil
+  end
+
+  class FakeLookupValue < ApplicationRecord
+    self.table_name = 'lookup_values'
+  end
+
+  class FakeLookupKey < ApplicationRecord
+    self.table_name = 'lookup_keys'
+    self.inheritance_column  = 'nil'
```
tbrisker: i think this should not be quoted

db/migrate/20180403144853_convert_vm_attrs_to_hash.rb
```diff
@@ -0,0 +1,80 @@
+class ConvertVmAttrsToHash < ActiveRecord::Migration[5.1]
+  def up
+    transform :to_h
+  end
+
+  def down
+    transform :to_params
```
tbrisker: do we actually want this? i don't think we want the AC:Params stored in db anyways tbh

db/migrate/20180403144853_convert_vm_attrs_to_hash.rb
```diff
@@ -0,0 +1,80 @@
...
+  end
+
+  YML_HASH = '!ruby/hash:ActiveSupport::HashWithIndifferentAccess'
+  YML_PARAMS_HASH = '!ruby/hash:ActionController::Parameters'
+  YML_PARAMS_OBJ = '!ruby/object:ActionController::Parameters'
+  YML_PARAMS_IVARS = '!ruby/hash-with-ivars:ActionController::Parameters'
```
tbrisker: Maybe we should combine all of these into `/\!ruby\/[\w-]+:ActionController::Parameters/`? gsub knows how to check for regex as well, not sure about the preformance impact.

xprazak2:
Updated, using regex for `gsub!`, removed the `down` and all permitted params are turned into a Hash.

db/migrate/20180403144853_convert_vm_attrs_to_hash.rb
```diff
@@ -0,0 +1,67 @@
...
+
+    say "All conversions finished"
+  end
+
+  def transform_batch_columns(base, serialized_columns, transform_method)
+    base.unscoped.select(serialized_columns + [:id]).find_each  do |object|
+      serialized_columns.each do |column|
+        attributes = object.send :read_attribute_before_type_cast, column
+        next if attributes.nil?
+        new_attributes = send(transform_method, attributes)
```
tbrisker: since we removed the down version, we can get rid of `transform_method` everywhere.
Then, you can leverage the fact that `gsub!` returns `nil` if no changes are made to save the comparison below. so in total:
```ruby
attributes = object.send :read_attribute_before_type_cast, column
next if attributes.nil?
object.update_column(column, attributes) unless attributes.gsub!(YML_PARAMS, YML_HASH).nil?
```

db/migrate/20180403144853_convert_vm_attrs_to_hash.rb
```diff
@@ -0,0 +1,67 @@
+class ConvertVmAttrsToHash < ActiveRecord::Migration[5.1]
+  def up
+    transform :to_h
+  end
+
+  YML_HASH = '!ruby/hash:ActiveSupport::HashWithIndifferentAccess'
+  YML_PARAMS = /\!ruby\/[\w-]+:ActionController::Parameters/
```
tbrisker: Looks like we don't need to escape `!` in the regex here and in the reports, sorry for misleading you.

tbrisker (changes requested):
Almost there! just a little clean up in the migration and I think we're good to go :)

xprazak2:
Updated.

tbrisker (approved):
Great job! any other comments @ares ?

ares (approved):

ares:
Works just fine, thanks for everyone's patience, I think at the end, the best solution to the problem won :-) Merging!

FrediWeber:
It is likely that this PR introduced #23581 (http://projects.theforeman.org/issues/23581).
Is it necessary that the lookup_values are also converted to hashes? If so - could we update the Foreman manual?

xprazak2:
There has been Rails 5 related change in how YAML gets serialized. This migration aims to change existing values into a new format, is should not turn them into hashes in db.

FrediWeber:
Yeah sorry, it was not this migration step that caused the issue. I created
a new PR with a workaround in the right one.

PR #5405, opened 2018-04-04
Fixes #21829: Add association_exists_validator

akofink:
This validator ensures that an associated model (fk relation) actually
exists with the ID specified on this model. i.e. Hostgroup.realm_id

This PR is required for Katello to require a ContentFacet on a Host.

### DIFF
--- a/app/validators/association_exists_validator.rb
+++ b/app/validators/association_exists_validator.rb
@@ -1,0 +1,7 @@
+class AssociationExistsValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    if record.send("#{attribute}_id".to_sym).present? && value.nil?
+      record.errors[attribute] << (options[:message] || _("with given ID not found"))
+    end
+  end
+end


lzap (approved):
Won't hurt I guess. LGTM

san7ket:
@akofink , I tried this using the hostgroup create command
`hammer --debug hostgroup create  --location-id="2" --operatingsystem-id="1" --domain-id="1" --partition-table-id="91" --content-source-id="121211" --content-view-id="2" --architecture-id="1" --name="mYXwuGU21HRd" --root-pass="test322132113"`

I didn't see the error message that was supposed to be shown but got same error instead
```
Could not create the hostgroup:
  PG::ForeignKeyViolation: ERROR:  insert or update on table "hostgroups" violates foreign key constraint "hostgroups_content_source_id_fk"
  DETAIL:  Key (content_source_id)=(121211) is not present in table "smart_proxies".
  : INSERT INTO "hostgroups" ("name", "created_at", "updated_at", "operatingsystem_id", "architecture_id", "ptable_id", "root_pass", "domain_id", "title", "content_source_id", "grub_pass", "content_view_id", "lookup_value_matcher") VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13) RETURNING "id"
```

akofink:
Yes, this is the error that users aren't happy about. It's too ugly to show, and it was requested that the error be something more like "No smart proxy with ID 121211 exists". This validator will help us catch this situation and handle appropriately.

san7ket:
@akofink yea, I tried it with the current fix, and looks like it didn't solve the issue for me. 
I still see the ` PG::ForeignKeyViolation:` Error on running the create command for hostgroup

akofink:
@san7ket This is just a validator. You have to [apply it to a model](https://github.com/Katello/katello/pull/7287) to get it to work.

ares:
So in other words, this only adds the "mechanism" but not necessary definitions to prevent the problem in Foreman itself. @akofink do you plan to continue with adding definitions as a separate PR or could you open redmine issue for that?

@san7ket I guess you could verify it with linked Katello PR a verify on Katello API endpoint, specified in the redmine issue.

akofink:
I can add a Foreman redmine issue for this. I'm not quite sure of the complete list of models+relations we should validate. The same should also be true of Katello. The reason I added only the one validator for content source in Katello was to solve the specific user-reported issue.

akofink:
http://projects.theforeman.org/issues/23208
http://projects.theforeman.org/issues/23209

ares:
Thanks @akofink, those help!

PR #5409, opened 2018-04-05
Fixes #23132 - Port robottelo tests for bookmarks

odovzhenko:
Redmine issue: http://projects.theforeman.org/issues/23132

Port robottelo tier1 tests for environment part of robottelo minitest port project https://github.com/SatelliteQE/robottelo/projects/1

Related Robottelo issue: SatelliteQE/robottelo#5907

### DIFF
--- a/test/models/bookmark_test.rb
+++ b/test/models/bookmark_test.rb
@@ -1,11 +1,13 @@
 require 'test_helper'
 
-class BookmarkTest < ActiveSupport::TestCase
-  @valid_controllers ||= (["dashboard", "common_parameters"] +
-      ActiveRecord::Base.connection.tables.map(&:to_s) +
-      Permission.resources.map(&:tableize)).uniq
+def valid_controller_values
+  (["dashboard", "common_parameters"] +
+  ActiveRecord::Base.connection.tables.map(&:to_s) +
+  Permission.resources.map(&:tableize)).uniq
+end
 
-  should allow_values(*@valid_controllers).for(:controller)
+class BookmarkTest < ActiveSupport::TestCase
+  should allow_values(*valid_controller_values).for(:controller)
   should allow_values(*valid_name_list).for(:name)
   should allow_values(*valid_name_list).for(:query)
   should_not allow_values(*invalid_name_list).for(:name)
--- a/test/models/bookmark_test.rb
+++ b/test/models/bookmark_test.rb
@@ -1,13 +1,6 @@
 require 'test_helper'
 
-def valid_controller_values
-  (["dashboard", "common_parameters"] +
-  ActiveRecord::Base.connection.tables.map(&:to_s) +
-  Permission.resources.map(&:tableize)).uniq
-end
-
 class BookmarkTest < ActiveSupport::TestCase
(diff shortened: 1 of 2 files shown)


mmoll:
[test foreman]

mmoll:
@odovzhenko test failures look kind of related, although these might be a sign of general problems with that seed test that were exposed by the test changes here. Could you have a look?

test/models/bookmark_test.rb
```diff
@@ -1,6 +1,16 @@
...
 class BookmarkTest < ActiveSupport::TestCase
+  @valid_controllers ||= (["dashboard", "common_parameters"] +
+      ActiveRecord::Base.connection.tables.map(&:to_s) +
+      Permission.resources.map(&:tableize)).uniq
+
+  should allow_values(*@valid_controllers).for(:controller)
+  should allow_values(*valid_name_list).for(:name)
+  should allow_values(*valid_name_list).for(:query)
+  should_not allow_values(*invalid_name_list).for(:name)
+  should_not allow_values('', ' ').for(:query)
```
ehelms: I think the lines 4-12 need to be in a `test "" do` block or else they may be outside a transaction and cause issues. This also will match the style of every other test.

ohadlevy (commented):
does it make sense to double test the model and the controller in this way? I'm not against the tests at all, but I'm concerned this kind of tests will add a very large impact on test running time. do you have code coverage report before / after? maybe we should be driven for workflow where coverage is missing?

and just to be clear, I'm not against merging this in any case.

ares:
Thanks for bringing this up.

I think it makes sense. Model tests goes deeper and verifies various inputs while controller tests rather verifies the impact of an API, not necessarily with all input combinations. Our existing controller tests are usually very poor in terms of coverage. E.g. we call create but only check the response is 200 without any verification it actually created something or replied with reasonable JSON data.

I'm usually running newly added tests ported from robottelo to see the time impact, here are the stats
```
test/controllers/api/v2/bookmarks_controller_test.rb
Run options: --seed 17345

# Running:

.............................................

Finished in 2.883344s, 15.6069 runs/s, 37.1097 assertions/s.
```

```
test/models/bookmark_test.rb
45 runs, 107 assertions, 0 failures, 0 errors, 0 skips
Run options: --seed 51194

# Running:

.............................................

Finished in 2.693303s, 16.7081 runs/s, 39.7282 assertions/s.
```
Note that these times include all existing bookmark tests we have. I think we have other tests that should be cleaned if we want to shorten the duration of test suite. I'd like to ask every committer to check the duration of ported tests before merge but so far, results were good.

odovzhenko:
We're usually trying to reduce running time of tests, @ares and @ehelms gave us several advice how to do that.
@mmoll and @ehelms Tests were fixed, thank you for help.

tbrisker (approved):
On my machine this leads to about 1.2s increase in runtime, seems fairly reasonable. 
The discussion regarding test times and coverage is important, but should be done outside this PR - Please bring it to https://community.theforeman.org/ so we can have better visibility into this concern and think together on where we can cut out tests or reduce runtime.

tbrisker:
Thanks @odovzhenko !

PR #5412, opened 2018-04-06
Fixes #23145 - handle strong params for compute attributes on failure

iNecas:
See Redmine for description of the issue.

### DIFF
--- a/app/controllers/concerns/foreman/controller/host_form_common.rb
+++ b/app/controllers/concerns/foreman/controller/host_form_common.rb
@@ -11,8 +11,9 @@ module Foreman::Controller::HostFormCommon
     return unless @host
 
     taxonomy_scope
-    if @host.compute_resource_id && params[:host] && params[:host][:compute_attributes]
-      @host.compute_attributes = params[:host][:compute_attributes]
+    if @host.compute_resource_id &&
+        respond_to?(:host_params) && host_params && host_params[:compute_attributes]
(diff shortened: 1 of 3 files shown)


iNecas:
Screenshot of the failure I'm talking about:

foreman-provision-fail

tbrisker (approved):
Code makes sense, didn't test though. 
I hope removing the rescue won't lead to other exceptions blowing up, but in any case swallowing them silently is a Bad Idea &tm; and we should handle them properly.

tbrisker:
@iNecas - What happens if an actual error is encountered when talking to the compute resource? 
@xprazak2 - I think we don't need to block branching on this, but i guess it would be good to get into the RC so we test this works properly

iNecas:
It depends on when the call is actually made, during the orchestration, the exception will be captured, as any other exception would be. When rendering, it's harder to capture, but that's an issue that is already in other places: just think about any other call that we don't rescue from (such as listing templates for select box or list of networks for NIC).

Btw. next thing I would like to do is to remove `test_connection` call from `new_vm` call, as it incredibly slows down the whole process, and I don't see any point of having it there. But keeping that outside of this PR

iNecas:
Tests are :green_heart:, this should be ready for merge.

ares (approved):
Tested, fixes the problem
review

Thanks @iNecas, merging

PR #5413, opened 2018-04-06
Fixes #23147 - Allow plugins to compile production webpack assets

ehelms:
The intent of this change is to allow plugins to individually compile their webpack assets for use in production. This follows a similar model to existing sprockets asset compilation. 

The requirements for plugins:

 1) Ability to generate webpack assets for plugin only
 2) Need to deliver the webpack assets in a plugins individual package (RPM/Deb)
 3) Need for those assets to be accessible in production and references to those assets to be in the final manifest.json

The intended workflow for build and deployment:

 1) A plugin calls `rake plugin:assets:precompile[plugin_name]`
 2) Webpack assets are generated inside `<plugin>/public/webpack/<plugin>/`
 3) In production build, such as RPM, these assets would need to be copied to `/usr/share/foreman/public/webpack` and laid down during install
 4) This puts plugin assets inside their own directories inside the `public/webpack` directory
 
How production server works:

 1) During startup, asset initialization code finds all plugin manifests and adds their entries to the main Foreman `manifest.json`
 2) A new version of `manifest.json` is written to disk
 3) All plugin assets are presumed present in `public/webpack`

Example "new" manifest on disk during server startup:

```
{
    "assetsByChunkName": {
        "ansible": [
            "ansible/ansible-d3598e423c33e9cbe088.js",
            "ansible/ansible-d3598e423c33e9cbe088.css"
        ],
        "bundle": [
            "bundle-511ea80f9d5431e940d7.js",
            "bundle-511ea80f9d5431e940d7.css"
        ],
        "vendor": "vendor-a3efca51ed683d599828.js",
        "katello": [
            "katello/katello-3b2f236fa6305d516a6d.js",
            "katello/katello-3b2f236fa6305d516a6d.css"
        ]
    },
    "errors": [

    ],
    "warnings": [

    ]
}
```

There are bound to be some nuances and edge cases I am not aware of when trying to apply this pattern to webpack. One question I was unclear about is the `vendor` JS that I think differs from how it was being done in development. If I understand correctly, plugins will potentially have overlapping dependencies given the vendor JS is generated separately from individual plugin JS. Are there any gotchas here that this does not account for?

I'd be happy to simplify this code if possible, and simplify the workflow if possible. However, given the stated requirements, and the workflows we've seen with Sprockets based assets and independence of plugins I am not sure there is a different workflow we can take. Note that some limitations are based on how webpack-rails is designed today (and with no real time to make changes to it to adapt to our architecture).

### DIFF
--- a/config/initializers/assets.rb
+++ b/config/initializers/assets.rb
@@ -89,5 +89,26 @@ Foreman::Application.configure do |app|
         ActionView::Base.assets_manifest = app.assets_manifest
       end
     end
+
+    if (webpack_manifest_file = Dir.glob("#{Rails.root}/public/webpack/manifest.json").first)
+      webpack_manifest = JSON.parse(File.read(webpack_manifest_file))
+
+      Foreman::Plugin.all.each do |plugin|
+        # Manifests may be stored under the engine installation or under the Foreman app root, then
+        # either with just the plugin name or with hyphens replaced with underscores.
+        manifest_path = File.join(plugin.path, "/public/webpack/#{plugin.id}/manifest.json")
+
+        if File.file?(manifest_path)
+          Rails.logger.debug { "Loading #{plugin.id} webpack asset manifest from #{manifest_path}" }
+          assets = JSON.parse(File.read(manifest_path))
+
+          webpack_manifest['assetsByChunkName'] = webpack_manifest['assetsByChunkName'].merge(assets['assetsByChunkName'])
+        end
+      end
+
+      File.open(webpack_manifest_file, 'w') do |file|
+        file.write(webpack_manifest.to_json)
+      end
+    end
   end
 end
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -11,156 +11,180 @@ var LodashModuleReplacementPlugin = require('lodash-webpack-plugin');
 var pluginUtils = require('../script/plugin_webpack_directories');
 var vendorEntry = require('./webpack.vendor');
 
-// must match config.webpack.dev_server.port
-var devServerPort = 3808;
 
-// set TARGETNODE_ENV=production on the environment to add asset fingerprints
-var production =
-  process.env.RAILS_ENV === 'production' ||
-  process.env.NODE_ENV === 'production';
 
-var bundleEntry = path.join(__dirname, '..', 'webpack/assets/javascripts/bundle.js');
+module.exports = env => {
+  // must match config.webpack.dev_server.port
+  var devServerPort = 3808;
 
-var plugins = pluginUtils.getPluginDirs('pipe');
+  // set TARGETNODE_ENV=production on the environment to add asset fingerprints
+  var production =
+    process.env.RAILS_ENV === 'production' ||
+    process.env.NODE_ENV === 'production';
 
-var resolveModules = [  path.join(__dirname, '..', 'webpack'),
-                        path.join(__dirname, '..', 'node_modules'),
-                        'node_modules/',
-                     ].concat(pluginUtils.pluginNodeModules(plugins));
+  var bundleEntry = path.join(__dirname, '..', 'webpack/assets/javascripts/bundle.js');
 
-var config = {
-  entry: Object.assign(
-    {
-      bundle: bundleEntry,
-      vendor: vendorEntry,
-    },
-    plugins.entries
-  ),
-  output: {
-    // Build assets directly in to public/webpack/, let webpack know
-    // that all webpacked assets start with webpack/
+  var plugins = pluginUtils.getPluginDirs('pipe');
 
-    // must match config.webpack.output_dir
-    path: path.join(__dirname, '..', 'public', 'webpack'),
-    publicPath: '/webpack/',
+  var resolveModules = [  path.join(__dirname, '..', 'webpack'),
+                          path.join(__dirname, '..', 'node_modules'),
+                          'node_modules/',
+                       ].concat(pluginUtils.pluginNodeModules(plugins));
 
-    filename: production ? '[name]-[chunkhash].js' : '[name].js'
-  },
-
-  resolve: {
-    modules: resolveModules,
-    alias: Object.assign({
-      foremanReact:
-        path.join(__dirname,
-           '../webpack/assets/javascripts/react_app'),
+  if (env && env.pluginName !== undefined) {
+    var pluginEntries = {};
+    pluginEntries[env.pluginName] = plugins['entries'][env.pluginName];
+    var entry = pluginEntries;
+    var outputPath = path.join(plugins['plugins'][env.pluginName]['root'], 'public', 'webpack');
+    var jsFilename = production ? env.pluginName + '/[name]-[chunkhash].js' : env.pluginName + '/[name].js';
+    var cssFilename = production ? env.pluginName + '/[name]-[chunkhash].css' : env.pluginName + '/[name].css';
+    var manifestFilename = env.pluginName + '/manifest.json';
+  } else {
+    var pluginEntries = plugins['entries'];
+    var entry = Object.assign(
+      {
+        bundle: bundleEntry,
+        vendor: vendorEntry,
       },
-      pluginUtils.aliasPlugins(plugins['entries'])
+      pluginEntries
     )
-  },
+    var outputPath = path.join(__dirname, '..', 'public', 'webpack');
+    var jsFilename = production ? '[name]-[chunkhash].js' : '[name].js';
+    var cssFilename = production ? '[name]-[chunkhash].css' : '[name].css';
+    var manifestFilename = 'manifest.json';
+  }
 
-  module: {
-    rules: [
-      {
-        test: /\.js$/,
-        exclude: /node_modules/,
-        loader: 'babel-loader',
-        options: {
-          'presets': [
-            path.join(__dirname, '..', 'node_modules/babel-preset-react'),
-            path.join(__dirname, '..', 'node_modules/babel-preset-env')
-          ],
-          'plugins': [
-            path.join(__dirname, '..', 'node_modules/babel-plugin-transform-class-properties'),
-            path.join(__dirname, '..', 'node_modules/babel-plugin-transform-object-rest-spread'),
-            path.join(__dirname, '..', 'node_modules/babel-plugin-transform-object-assign'),
-            path.join(__dirname, '..', 'node_modules/babel-plugin-lodash')
-          ]
+  var config = {
+    entry: entry,
+    output: {
+      // Build assets directly in to public/webpack/, let webpack know
+      // that all webpacked assets start with webpack/
+
+      // must match config.webpack.output_dir
+      path: outputPath,
+      publicPath: '/webpack/',
+      filename: jsFilename
+    },
+
+    resolve: {
+      modules: resolveModules,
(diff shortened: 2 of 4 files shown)


lib/tasks/plugin_assets.rake
```diff
@@ -117,8 +117,29 @@ task 'plugin:assets:precompile', [:plugin] => [:environment] do |t, args|
...
+        sh "#{webpack_bin} --config #{config_file} --bail --env.pluginName=#{@plugin.id}"
+      end
+
+    end
+  end
+
   if args[:plugin]
-    task = Foreman::PluginAssetsTask.new(args[:plugin])
+    #task = Foreman::PluginAssetsTask.new(args[:plugin])
+    #task.compile
```
ekohl: Won't this break existing asset compilation that's outside of webpack?

lib/tasks/plugin_assets.rake
```diff
@@ -117,8 +117,29 @@ task 'plugin:assets:precompile', [:plugin] => [:environment] do |t, args|
...
+
+      def initialize(plugin_id)
+        @plugin = Foreman::Plugin.find(plugin_id) or raise("Unable to find registered plugin #{plugin_id}")
+      end
+
+      def compile
+        ENV["NODE_ENV"] ||= 'production'
+        webpack_bin = ::Rails.root.join(::Rails.configuration.webpack.binary)
+        config_file = ::Rails.root.join(::Rails.configuration.webpack.config_file)
+        sh "#{webpack_bin} --config #{config_file} --bail --env.pluginName=#{@plugin.id}"
```
ekohl: I think this needs to check if the plugin has a `package.json` file. Only then do you need to compile the webpack.

lib/tasks/plugin_assets.rake
ehelms: Oops, that was from testing :)

lib/tasks/plugin_assets.rake
ehelms: I can put that check here. I figured it was covered by https://github.com/theforeman/foreman/blob/develop/script/plugin_webpack_directories.rb#L29

ekohl: In our packaging we call `rake webpack:plugin:asserts:precompile[GEM_NAME]` so this code is called without exception. Perhaps webpack itself smart enough to ignore it, but if we can avoid calling webpack altogether that's probably more efficient.

ehelms:
@dLobatog You'll likely want to look at this with respect to foreman_ansible

@ohadlevy pinging you since you'd been asking me about this particular issue on production installs

config/webpack.config.js
```diff
@@ -11,152 +11,176 @@ var LodashModuleReplacementPlugin = require('lodash-webpack-plugin');
...
+  var plugins = pluginUtils.getPluginDirs();
+
+  var resolveModules = [  path.join(__dirname, '..', 'webpack'),
+                          path.join(__dirname, '..', 'node_modules'),
+                          'node_modules/',
+                       ].concat(pluginUtils.pluginNodeModules(plugins));
+
+  if (env && env.pluginName !== undefined) {
+    var pluginEntries = {};
+    pluginEntries[env.pluginName] = plugins['entries'][env.pluginName];
```
dLobatog, line 36: I'm having some problems with these lines. To compile the assets I call `rake plugin:assets:precompile[foreman_ansible]`. This creates a `plugins['plugins']` structure like this:

```js
{ remoteexecution:
   { root: '/home/vagrant/foreman_remote_execution',
     entry: '/home/vagrant/foreman_remote_execution/webpack/index.js' },
  ansible:
   { root: '/home/vagrant/foreman_ansible',
     entry: '/home/vagrant/foreman_ansible/webpack/index.js' },
  katello:
   { root: '/home/vagrant/katello',
     entry: '/home/vagrant/katello/webpack/index.js' } }
```

Which means the line calling `plugins['plugins'][env.pluginName]['root']` fails, because `env.pluginName == 'foreman_ansible'`.

I'm trying to see how can we avoid this by making the plugin hash ('remoteexecution, ansible') have the same names as the plugin names ('foreman_remote_execution, foreman_ansible'). Probably just requires a change in `plugin_webpack_directories.rb`

dLobatog (changes requested):

script/plugin_webpack_directories.rb
```diff
@@ -31,8 +31,9 @@
   # skip the plugin if its path is already included
   next if config[:paths].include?(path)
   if File.exist?(entry)
-    config[:entries][bundle_name dep.name] = entry
+    config[:entries][bundle_name(dep.name)] = entry
```
dLobatog: I think these shouldn't be calling `bundle_name` to fix what I mentioned in https://github.com/theforeman/foreman/pull/5413/files#r180416075 , otherwise I can't precompile other plugins but Katello.

The problem I found, however, is that `foreman_remote_execution_core` and `foreman_ansible_core` are picked up first

config/webpack.config.js
ekohl: Doesn't that mean it should call `plugin:assets:precompile[ansible]` or is the problem that the gem and node identifiers are different?

dLobatog: Calling `plugin:assets:precompile[ansible]` doesn't work either, because it (correctly) mentions that it's `Unable to find registered plugin ansible`. I think the node identifiers in the hash above should be fixed to have the same names as the plugin gems, I'm working on it.

dLobatog (commented):
@ehelms After applying these changes https://github.com/theforeman/foreman/pull/5432 + this I'm able to successfully get the bundle, I'll check whether it works in production but it looks promising, thanks!

dLobatog (changes requested):
@ehelms I think this just needs a rebase and we ought to be able to start using it in packaging. Thank you!

ehelms:
I rebased this to ensure it includes @dLobatog update. I also updated based @ekohl comments. Please review and test @dLobatog

dLobatog (approved):
Awesome job @ehelms . I'm definitely able to compile individual bundles of assets for foreman_ansible/foreman_remote_execution, which we can package:

```
find public/
public/
public/assets
public/assets/foreman_remote_execution
public/assets/foreman_remote_execution/execution_interface-32241667bb32bbb121638ab31efcd33da7976accf8a3bcfc856e9e786b630a26.js
public/assets/foreman_remote_execution/execution_interface-32241667bb32bbb121638ab31efcd33da7976accf8a3bcfc856e9e786b630a26.js.gz
public/assets/foreman_remote_execution/job_templates-7562fab3ee4f501768a244372b8c1531397be584511afa485ea52a1408d2d9b9.js
public/assets/foreman_remote_execution/job_templates-7562fab3ee4f501768a244372b8c1531397be584511afa485ea52a1408d2d9b9.js.gz
public/assets/foreman_remote_execution/template_input-72fb4f368e5133241b40bccb0fa83ae970af0c9538465dc6df5b0c2b2c20095e.js
public/assets/foreman_remote_execution/template_input-72fb4f368e5133241b40bccb0fa83ae970af0c9538465dc6df5b0c2b2c20095e.js.gz
public/assets/foreman_remote_execution/template_invocation-caab3b6833103aa41c6ed0ddf67ccd12732c0b01a89cc9c11e49cbf7468b0e1d.js
public/assets/foreman_remote_execution/template_invocation-caab3b6833103aa41c6ed0ddf67ccd12732c0b01a89cc9c11e49cbf7468b0e1d.js.gz
public/assets/foreman_remote_execution/modal_window-7117ccc83cf81f90d031ea2faeba27ddaefe176603551d25e300b59c2fdd947b.css
public/assets/foreman_remote_execution/modal_window-7117ccc83cf81f90d031ea2faeba27ddaefe176603551d25e300b59c2fdd947b.css.gz
public/assets/foreman_remote_execution/template_invocation-bf20f4154d0c5fb85a297aa4b930a73bc8bbfde6d77962de84d883686ac2f323.css
public/assets/foreman_remote_execution/template_invocation-bf20f4154d0c5fb85a297aa4b930a73bc8bbfde6d77962de84d883686ac2f323.css.gz
public/assets/foreman_remote_execution/job_invocations-9daf6f115b371792f948dbfac4521dd61318c5d9ff7908f61ad73cce77dce815.css
public/assets/foreman_remote_execution/job_invocations-9daf6f115b371792f948dbfac4521dd61318c5d9ff7908f61ad73cce77dce815.css.gz
public/assets/foreman_remote_execution/foreman_remote_execution.json
> find ../foreman_ansible/public/
../foreman_ansible/public/
../foreman_ansible/public/assets
../foreman_ansible/public/assets/foreman_ansible
../foreman_ansible/public/assets/foreman_ansible/foreman-ansible-114d3c4b2f0642fc2cb1a9ef1f259211a1388a19fa81a03c1853d9fbb855991e.css
../foreman_ansible/public/assets/foreman_ansible/Ansible-a375e658d79bd6182d5743dc28a056a169c6fd67fd059ac74b5ba17bdedd5517.png
../foreman_ansible/public/assets/foreman_ansible/foreman-ansible-114d3c4b2f0642fc2cb1a9ef1f259211a1388a19fa81a03c1853d9fbb855991e.css.gz
../foreman_ansible/public/assets/foreman_ansible/foreman_ansible.json
../foreman_ansible/public/webpack
../foreman_ansible/public/webpack/foreman_ansible
../foreman_ansible/public/webpack/foreman_ansible/foreman_ansible-a5771567df875fd00645.js
../foreman_ansible/public/webpack/foreman_ansible/foreman_ansible-a5771567df875fd00645.css.gz
../foreman_ansible/public/webpack/foreman_ansible/foreman_ansible-a5771567df875fd00645.js.gz
../foreman_ansible/public/webpack/foreman_ansible/manifest.json
../foreman_ansible/public/webpack/foreman_ansible/foreman_ansible-a5771567df875fd00645.css
```

dLobatog:
I'll follow up updating https://github.com/theforeman/foreman-packaging/pull/2256

tbrisker:
Hound comments are valid and at least in one case raise a good question ('pluginEntry' is assigned a value but never used no-unused-vars) - @ehelms / @dLobatog - please follow up this PR with a fix for them.

evgeni:
This [broke RPM building for Katello nightly](https://projects.theforeman.org/issues/23361) :(

config/initializers/assets.rb
```diff
@@ -89,5 +89,26 @@
...
       end
     end
+
+    if (webpack_manifest_file = Dir.glob("#{Rails.root}/public/webpack/manifest.json").first)
+      webpack_manifest = JSON.parse(File.read(webpack_manifest_file))
+
+      Foreman::Plugin.all.each do |plugin|
+        # Manifests may be stored under the engine installation or under the Foreman app root, then
+        # either with just the plugin name or with hyphens replaced with underscores.
+        manifest_path = File.join(plugin.path, "/public/webpack/#{plugin.id}/manifest.json")
```
ekohl, line 99: Should this be `public/...` rather than `/public/`? Or rather `File.join(plugin.path, 'public', 'webpack', plugin.id, 'manifest.json')`

ekohl: Included in https://github.com/theforeman/foreman/pull/5478

tbrisker:
hound fixes have been included in https://github.com/theforeman/foreman/pull/5498

PR #5414, opened 2018-04-06
Fixes #19770 - Disable Layout/DotPosition cop

mmoll:

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -76,3 +76,6 @@ Bundler/OrderedGems:
 
 Style/Alias:
   EnforcedStyle: prefer_alias_method
+
+Layout/DotPosition:
+  Enabled: false
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -34,13 +34,6 @@ Layout/CaseIndentation:
 Layout/ClosingParenthesisIndentation:
   Enabled: false
 
-# Offense count: 134
-# Cop supports --auto-correct.
-# Configuration parameters: EnforcedStyle.
-# SupportedStyles: leading, trailing
-Layout/DotPosition:
-  Enabled: false
-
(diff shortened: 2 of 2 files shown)


mmoll:
There are a few more occurences in the current codebase with the dot at the end.

Also, it feels more logical to me, because I know when I see such a dot that the statement does continue.

Opinions?

tbrisker (commented):
TBH, I prefer the leading version, esp. when the dot is aligned to the previous lines' dots it makes it clear this is a chain of functions. In the trailing style when reading a line it isn't a clear it continues the previous line, especially if the previous line is long and the trailing dot isn't visible.

mmoll:
I opened GH-5419 for comparison

ares:
Tbh I prefer this one more than the alternative, but I don't insist. I'd be even happier if we keep this disabled.

ekohl:
I also have a slight preference for the other version.

mmoll:
Just to kick into this again, I'm still preferring this version, the alternative PR has about the double amount of chnages and I discovered this setting like here also in foreman_ansible... But also I wouldn't be really sad to disable this cop.

ekohl:
I really see benefits in both styles so I can see why there's a difference in preferences. To me that suggests disabling it is a good way forward here.

tbrisker:
I agree, we don't have an agreement on preferred style so let's disable it.

mmoll:
sure, disabled :)

tbrisker (changes requested):
Please remove the relevant lines from the todo yml :)

mmoll:
aw. done.

tbrisker (approved):
Thanks @mmoll

lzap:
Proper execution by everyone, thank God we don't have this one.

PR #5416, opened 2018-04-06
Fixes #19787 - Fix Layout/SpaceAfterComma cop

mmoll:


mmoll:
Personally I prefer the style the cop enforces, but wouldn't be really sad if others want to have the cop disabled.

Opinions?

ekohl (approved):
Though not involved in core that much, I personally like this style for readability.

tbrisker (approved):
I agree, this improves readability, I'll give it another day or two just in case others object.

tbrisker:
Looks like there aren't any objections, merging. Thanks @mmoll !

PR #5417, opened 2018-04-06
Fixes #19848 - Set Style/Alias cop to prefer_alias_method

mmoll:


mmoll:
3 vs. 37 - I guess there will be not much discussion here :)

tbrisker (approved):
Thanks @mmoll !

PR #5421, opened 2018-04-09
Refs #22493 - Avoid loading logging-journald

ekohl:
Without this bundler_ext would always load the gem. This fails because logging-journald doesn't place a ruby file in the correct place. That's not a problem in most operations because the logging framework can automatically load the plugin if needed.

### DIFF
--- a/bundler.d/journald.rb
+++ b/bundler.d/journald.rb
@@ -1,4 +1,4 @@
 # disable to avoid journald native gem in development setup
 group :journald do
-  gem 'logging-journald', '~> 1.0'
+  gem 'logging-journald', '~> 1.0', require: false
 end


bundler.d/journald.rb
```diff
@@ -1,4 +1,4 @@
 # disable to avoid journald native gem in development setup
 group :journald do
-  gem 'logging-journald', '~> 1.0'
+  gem 'logging-journald', '~> 1.0', require: false
```
ekohl, line 3: The alternative was `require: 'logging/plugins/journald'` which always explicitly loads it. I don't know which is better

ekohl:
Duplicates https://github.com/theforeman/foreman/pull/5377

lzap:
Yeah it was pending review already. Thanks for looking into it!

PR #2148, opened 2015-02-10
fixes #9240 - in case of hash with indifferent access, serialize as hash

unorthodoxgeek:

PR #5422, opened 2018-04-09
Fixes #23183 - Remove HashWithIndifferentAccess workaround

mmoll:
See GH-2148, this should have been removed with Rails 4.x.


tbrisker:
@mmoll please open a new issue in redmine, this issue has been closed for a long time

mmoll:
opened a new issue

tbrisker:
This is still used in https://github.com/theforeman/foreman/blob/develop/app/models/lookup_value.rb#L37 and despite the comment, these methods aren't present in newer rails versions. It might be that they are no longer needed, but i'm not sure.

mmoll:
@orrabin could you have a look there?

henry-spanka:
After upgrading to 1.17 I get the same bug as described in https://projects.theforeman.org/issues/9240.
Is this related to this pull request?

mmoll:
@henry-spanka it shouldn't be connected in a user visible way.

orrabin:
@mmoll I tried to create a nested HashWithIndifferentAccess to see if it get's converted with `to_hash` and it looked to me like it did. It was a very simple hash but it looks like if you change the like that @tbrisker send to `val.to_hash` it should work .

henry-spanka:
@mmoll Well, at least after upgrade from 1.16 to 1.17 I can not render the external YML from a node.

The output is now like:
```yaml
--- !ruby/hash:ActiveSupport::HashWithIndifferentAccess
testvg: !ruby/hash:ActiveSupport::HashWithIndifferentAccess
  physical_volumes:
  - /dev/sdb
  - /dev/sdc
  logical_volumes: !ruby/hash:ActiveSupport::HashWithIndifferentAccess
    test1: !ruby/hash:ActiveSupport::HashWithIndifferentAccess
      size: 128M
      mountpath: /opt/test1
    test2: !ruby/hash:ActiveSupport::HashWithIndifferentAccess
      size: 128M
      mountpath: /opt/test2
    test3: !ruby/hash:ActiveSupport::HashWithIndifferentAccess
      size: 128M
      mountpath: /opt/test3
    test4: !ruby/hash:ActiveSupport::HashWithIndifferentAccess
      size: 128M
      mountpath: /opt/test4
    test5: !ruby/hash:ActiveSupport::HashWithIndifferentAccess
      size: 128M
      mountpath: /opt/test5
```

mmoll:
@henry-spanka please open an issue over at https://projects.theforeman.org/
BTW, https://projects.theforeman.org/issues/23382 looks like it.

henry-spanka:
@mmoll Thanks. I have opened a new issue ([23463](https://projects.theforeman.org/issues/23463))

app/models/lookup_value.rb
```diff
@@ -34,7 +34,7 @@ def check_permissions_after_save
 
   def value=(val)
     if val.is_a?(HashWithIndifferentAccess)
-      super(val.deep_to_hash)
+      super(val.to_hash)
```
tbrisker, line 37: This should use `deep_stringify_keys`, and perhaps do it for all hashes, not just HWIA's (in case there is a HWIA nested inside a regular hash)

tbrisker (changes requested):

PR #5424, opened 2018-04-10
Fixes #23177 - unattended warning on multiple MACs

lzap:
In unattended search by MAC or Anaconda. Can be tricky to debug when tokens are turned off.

Optionally, we shall sort the entries by created_at so the last is picked up.

### DIFF
--- a/test/controllers/unattended_controller_test.rb
+++ b/test/controllers/unattended_controller_test.rb
@@ -5,16 +5,16 @@ class UnattendedControllerTest < ActionController::TestCase
     Host::Managed.any_instance.stubs(:handle_ca).returns(true)
     as_admin do
       disable_orchestration # avoids dns errors
-      ptable = FactoryBot.create(:ptable, :name => 'default',
+      @ptable = FactoryBot.create(:ptable, :name => 'default',
                                   :operatingsystem_ids => [operatingsystems(:redhat).id])
-      ptable_ubuntu = FactoryBot.create(:ptable, :ubuntu, :name => 'ubuntu default',
+      @ptable_ubuntu = FactoryBot.create(:ptable, :ubuntu, :name => 'ubuntu default',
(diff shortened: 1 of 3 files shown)


test/controllers/unattended_controller_test.rb
```diff
@@ -57,6 +57,23 @@ class UnattendedControllerTest < ActionController::TestCase
...
+      :name => "host2_same_mac",
+      :mac => @rh_host.mac,
+      :operatingsystem => operatingsystems(:ubuntu1010),
+      :ptable => @ptable_ubuntu,
+      :medium => media(:ubuntu),
+      :architecture => architectures(:x86_64),
+      :organization => @org,
+      :location => @loc)
+    get :host_template, params: { :kind => 'finish', :mac => host1.mac }
+    assert @rh_host.created_at < host1.created_at
```
lzap: Interesting, this assert failed on jenkins but only for MySQL and just once (one Ruby version). Why would `rh_host` not be created before `host1`? Tried again with better assert error.

timogoebel (approved):
LGTM!

timogoebel:
Thanks, @lzap.

iNecas:
FYI we started hitting intermittent test failures in the test introduced here. I first suspected https://github.com/theforeman/foreman/pull/5444#issuecomment-383365648, but it turned out this was just added here as well)

I've seen this in foreman core tests already several times (http://ci.theforeman.org/view/Core/job/test_develop/3426/database=mysql,ruby=2.5,slave=fast/consoleFull), as well as in rex tests (
http://ci.theforeman.org/job/test_plugin_matrix/5176/database=mysql,ruby=2.4,slave=fast/testReport/junit/(root)/UnattendedControllerTest/test_0004_should_get_a_kickstart_if_MAC_is_provided_with_two_hosts_with_same_MAC/)

So far, I've seen in only on mysql, if that helps with reproducing.

lzap:
Thanks taking a look.

PR #5425, opened 2018-04-10
Fixes #22117 - Add support for number of sockets for ovirt

shiramax:


shiramax:
This needs : https://github.com/fog/fog-ovirt/pull/16 to be merged as well.

ekohl (changes requested):
It'll also need to bump the required version in `bundler.d/ovirt.rb` when the library has released a version that supports sockets.

app/models/compute_resources/foreman/model/ovirt.rb
```diff
@@ -94,7 +94,11 @@ def provided_attributes
 
     #FIXME
```
tbrisker: Can we remove this FIXME? is 16 the max count supported for these by ovirt?

shiramax: @tbrisker 
After looking at ovirt engine,  I saw the following:
[root@compute-ge-6 ~]#  engine-config --get MaxNumOfVmCpus
MaxNumOfVmCpus: 240 version: 3.6
MaxNumOfVmCpus: 240 version: 4.0
MaxNumOfVmCpus: 288 version: 4.1
MaxNumOfVmCpus: 384 version: 4.2

[root@compute-ge-6 ~]# engine-config --get  MaxNumOfCpuPerSocket
MaxNumOfCpuPerSocket: 16 version: 3.6
MaxNumOfCpuPerSocket: 16 version: 4.0
MaxNumOfCpuPerSocket: 254 version: 4.1
MaxNumOfCpuPerSocket: 254 version: 4.2

that means the max_socket=16 and max_cores=16 is true to ovirt 3.6 and 4.0 but not to versions >= 4.0 
After taking to @orrabin , we think that at the moment we don't see any reason to set the max socket and cores to be more then 16.
Until now will allowed only 8 cores and after this patch, we will allow 16*16 = 256 cores, which seems to be enough at the moment.

tbrisker (dismissed):
Ack, makes sense. Thanks @shiramax !

orrabin:
I tested that when selecting a template the right number of sockets is filled in, I changed that number and save to see the new number was saved in the host (checked in foreman and ovirt).

orrabin:
@ekohl I added: https://github.com/theforeman/foreman-packaging/pull/2650

@tbrisker - any other comments here?

tbrisker (approved):
LGTM

tbrisker:
Packaging side should be merged first.

shiramax:
@ekohl, thanks, the packaging PR was merged, is there anything else?

ekohl (approved):

orrabin:
Thanks @shiramax !

PR #5428, opened 2018-04-10
WIP: Rails 5.2

mmoll:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,6 +8,8 @@ source 'https://rubygems.org'
 case SETTINGS[:rails]
 when '5.1'
   gem 'rails', '5.1.6'
+when '5.2'
+  gem 'rails', '5.2.0'
 else
   raise "Unsupported Ruby on Rails version configured in settings.yaml: #{SETTINGS[:rails]}"
 end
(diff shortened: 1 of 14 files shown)


mmoll:
[test katello]

xprazak2:
Tested, seems to work ok, rebase will be needed. The foreman test failures do not look random.

mmoll:
Rebased.

The test failure is indeed new and I didn't suceed in finding the cause, somebody with actual knowledge about the CR and keypair stuff needs to have a look. Maybe it's only a test problem (or a Rails bug).

See https://projects.theforeman.org/issues/22938 for more information.

mmoll:
@ehelms any idea why the DB creation in the katello PR test fails here?

ehelms:
What appears to be happening (without digging too deep) is that the plugin code is triggering this error. If you look through the stack trace when creating the database you will find it calls this line:

https://github.com/theforeman/foreman/blob/develop/app/registries/foreman/plugin.rb#L318

I am guessing something in 5.2 changed where it attempts to check that even if a database does not exist and Rails throws an error. I think we need to modify some of that code to not run pending migrations check if no database exists? Or if we are in the rake task 'db:create' ?

mmoll:
@ehelms thanks, I think I already know now, where to poke :point_left:

mmoll:
Turns out, I have no idea what's going on there, it's working with _discovery in my travis test setup...

mmoll:
see GH-5836

PR #5429, opened 2018-04-11
Refs #22809 - Ensure params hash has indifferent access

xprazak2:
@tbrisker, something I thought of this morning, does it make sense?


tbrisker:
` ActionController::Parameters#to_h` already returns a hash with indifferent access, so this isn't needed

xprazak2:
You are right, I did not notice. Closing, thanks!

PR #5431, opened 2018-04-11
i18n - extracting new, pulling from tx

xprazak2:


xprazak2:
Merging translations, foreman tests are green.

PR #5432, opened 2018-04-11
Fixes #23215 - Keep original names in webpack bundle

dLobatog:
Currently the webpack config is shortening the name of the js bundles
for plugins. E.g, foreman_remote_execution becomes remoteexecution, and
so on.
This becomes hard to deal with when we want to figure out which plugin
contains the source for each bundle.
Related to #5413 

For this to work, the plugins must contain package.json in the gemspec. REX is covered already, Katello and Ansible need to be fixed https://github.com/Katello/katello/pull/7302 https://github.com/theforeman/foreman_ansible/pull/160


script/plugin_webpack_directories.rb
```diff
@@ -27,12 +25,12 @@
 
   path = "#{dep.to_spec.full_gem_path}/webpack"
   entry = "#{path}/index.js"
-  # some plugins share the same base directory (tasks-core and tasks, REX etc)
   # skip the plugin if its path is already included
   next if config[:paths].include?(path)
-  if File.exist?(entry)
-    config[:entries][bundle_name dep.name] = entry
+  if File.exist?(entry) && dep.files.any? { |f| f == 'package.json' }
```
xprazak2: Have you considered just checking if file exists same way we do it with index.js?

xprazak2 (commented):
Tested, works for me.

script/plugin_webpack_directories.rb
ekohl: If I understand it correctly this means it needs to be in the gemspec. I like that because otherwise we'll fail on it during packaging. Printing a warning if the file is present but not in the gemspec would be very developer friendly and prevent frustration so bonus points for that.

Also, wouldn't it be equal to `dep.files.include?('package.json')`?

script/plugin_webpack_directories.rb
```diff
@@ -27,12 +25,12 @@
 
   path = "#{dep.to_spec.full_gem_path}/webpack"
   entry = "#{path}/index.js"
```
ekohl: Technically I think we should read [main from package.json](https://docs.npmjs.com/files/package.json#main) and use that instead of hardcoding it (with a fallback).

ekohl (commented):
I like this change. It makes the script much more standalone and less likely to fail.

dLobatog:
@xprazak2 I think it at least the `webpack` dir needs to be in the gem, so that on packaging, we are able to compile assets during the build stage

script/plugin_webpack_directories.rb
dLobatog: That seems easy enough to implement, I'll do it.

dLobatog (commented):
I've update it to include warnings when package.json or webpack are in the root of the plugin but not in the gemspec. Of course this means `_core` gems are going to throw a warning. 

Also the `main` entrypoint is now read from `package.json` instead of hardcoded (fallback is index.js in case package.json  does not define it)

script/plugin_webpack_directories.rb
```diff
@@ -26,13 +24,25 @@
   end
 
   path = "#{dep.to_spec.full_gem_path}/webpack"
-  entry = "#{path}/index.js"
-  # some plugins share the same base directory (tasks-core and tasks, REX etc)
```
ekohl: I think this comment does add value.

script/plugin_webpack_directories.rb
```diff
@@ -26,13 +24,25 @@
...
   # skip the plugin if its path is already included
   next if config[:paths].include?(path)
+  if Dir.exist?(path) && !dep.files.any? { |f| f.match?(/webpack/) }
+    puts "WARNING: webpack was found in #{dep.name}'s source, "\
```
ekohl: I think it's good if this would go to stderr.

script/plugin_webpack_directories.rb
```diff
@@ -26,13 +24,25 @@
...
+      "but it's not in the gemspec. Please add it to the gemspec."
+  end
+  if File.exist?("#{dep.to_spec.full_gem_path}/package.json") &&
+      !dep.files.include?('package.json')
+    puts "WARNING: package.json was found in #{dep.name}'s source, "\
```
ekohl: Should also go to stderr

dLobatog (commented):
Updated to print the warnings in STDERR and also keep the comment about `_core` plugins

ekohl (approved):
Looks good. This also makes the `plugin_webpack_directories.rb` a rather basic ruby script with minimal dependencies. I like that a lot and makes it much less to fail. :+1:

dLobatog:
Anything blocking this PR from being merged?

script/plugin_webpack_directories.rb
```diff
@@ -12,9 +11,8 @@
   specs = Bundler.load.specs
 end
 
-include Foreman::WebpackAssets
-
-config = { entries: {}, paths: [] }
+config = { entries: {}, paths: [], plugins: {} }
+plugin_name_regexp = /foreman*|katello*/
```
ehelms, line 15: This does not appear to be used anywhere

ehelms (approved):
Small comment, otherwise I appreciate this change. In general, when doing anything with plugins, using the name of the plugin for identifier makes life easier.

script/plugin_webpack_directories.rb
ekohl: It is within the loop below

ehelms: Oh good catch, I didn't expand things

mmoll:
merged, gracias @dLobatog!

PR #5434, opened 2018-04-11
Fixes #22698 - clearing search filter does not clear search filter

boaz0:


webpack/assets/javascripts/foreman_tools.js
```diff
@@ -134,6 +134,9 @@ export function updateTable(element) {
     }
     uri.setSearch('per_page', perPage);
   }
+  if ($('#search-form').find('.autocomplete-input').val() === '') {
```
ohadlevy: not that i think its common, but maybe make sense to add trim, so even if there are whitespaces search gets removed.

boaz0: thanks @ohadlevy - I updated the code + added a test to check that as well. :+1:

webpack/assets/javascripts/foreman_tools.js
```diff
@@ -134,6 +134,9 @@ export function updateTable(element) {
     }
     uri.setSearch('per_page', perPage);
   }
+  if ($('#search-form').find('.autocomplete-input').val().trim() === '') {
```
tbrisker: This may throw an error if `$('#search-form').find('.autocomplete-input').val()` is undefined.

ares:
[test foreman] [test katello]

ohadlevy:
[test foreman] [test katello]

webpack/assets/javascripts/foreman_tools.js
```diff
@@ -135,6 +135,11 @@ export function updateTable(element) {
     uri.setSearch('per_page', perPage);
   }
 
+  const searchInput = $('#search-form').find('.autocomplete-input').val();
+  if (searchInput !== undefined && searchInput.trim() === '') {
+    uri.removeSearch('search');
```
tbrisker: This looks quite similar to lines 129-134, perhaps time to refactor this function a bit? with this change it will be almost 50 lines long, and is becoming difficult to understand - leading to various cases (such as this) not being handled correctly.

boaz0: indeed, I will take a look at this and try to understand how to clean this mess :grinning:

boaz0:
I am glad I did this. It looks better than before.

webpack/assets/javascripts/foreman_tools.js
```diff
@@ -103,44 +103,29 @@ export function initTypeAheadSelect(input) {
...
+      element.id === 'search-form'
+        ? (
+          $(element)
+            .find('.autocomplete-input')
+            .val() || values.search
+        ).trim()
+        : '';
+    values.page = $('#cur_page_num').val() || values.page;
+    values.per_page = $('#per_page').val() || values.per_page;
+    Object.keys(values).forEach(k => uri.setSearch(k, values[k]));
```
tbrisker: `URI.setSearch` can take an object as it's parameter, no need to manually do this for each value. Also, if a value is `undefined` it is removed so you don't need to explicitly remove them if you don't default to empty string but rather start with a blank object and only set the fields that are needed.

webpack/assets/javascripts/foreman_tools.js
```diff
@@ -103,44 +103,29 @@ export function initTypeAheadSelect(input) {
...
-    }
-    uri.setSearch('per_page', perPage);
+    values.search =
+      element.id === 'search-form'
+        ? (
+          $(element)
+            .find('.autocomplete-input')
+            .val() || values.search
+        ).trim()
+        : '';
```
tbrisker: This is a very hard to read statement, please consider refactoring.

tbrisker (changes requested):

webpack/assets/javascripts/foreman_tools.js
boaz0: I am not sure I understand. Having `param1="value1"` in the URI parameters and setting it to undefined doesn't remove it from URI, it just sets the search in URI to `param1`.

Can you give an example or something because I would like to remove that line if it's possible.

webpack/assets/javascripts/foreman_tools.js
boaz0: Yep, that's really ugly after `prittier-eslint`-ing it. I guess I will break it.

boaz0:
Thanks @tbrisker I think I made some progress with what you suggested.
I ran some manual tests and added a new unit-test to cover a scenario never tested before which increased the code-coverage ratio too :raised_hands:

webpack/assets/javascripts/foreman_tools.js
```diff
@@ -103,43 +103,28 @@ export function initTypeAheadSelect(input) {
...
-        .find('.autocomplete-input')
-        .val() === undefined
-    ) {
-      uri.removeSearch('search');
-    }
-    uri.setSearch('per_page', perPage);
+
+  // if element is undefined don't refresh page or update Turbolinks.
+  if (element === undefined) {
+    return false;
```
boaz0: @ohadlevy can you review this part of the code.
Is it OK that we are not returning `true` and not calling `Turbolinks`?
Please notice the unit-test for `updateTable` when `element === undefined` was changed accordingly.
The reason that we changed it is because we didn't find a reason to do so when `element === undefined` because the `uri` will stay the same (unless was modified by the user manually, i guess).

ohadlevy: @ripcurld when would it be undefined? based on how its called (in pagination / search) I am not sure it can be undefined?

boaz0: @ohadlevy :+1: 
Exactly, looking at the code `updateTable` is always called with element that is not `undefined`.
But then, looking at the unit tests I saw there was a test where `updateTable` is called without passing any arguments which verifies `Turbolinks` will be called.

Checking in `git` it said you wrote it and I thought maybe I'm missing something and you could determine if this change is fine. :grin:

ripcurld:
[test foreman]

webpack/assets/javascripts/foreman_tools.js
```diff
@@ -103,43 +103,23 @@ export function initTypeAheadSelect(input) {
...
-    if (isSearchForm) {
-      searchTerm = $(element)
-        .find('.autocomplete-input')
-        .val();
-      if (searchTerm) {
-        uri.setSearch('search', searchTerm.trim());
-      }
-    }
+  values.page = (element !== undefined) ? $('#cur_page_num').val() : undefined;
+  if (element !== undefined && ['per_page', 'search-form'].indexOf(element.id) > -1) {
```
tbrisker: you can use `Array.prototype.includes` here

webpack/assets/javascripts/foreman_tools.js
```diff
@@ -103,43 +103,23 @@ export function initTypeAheadSelect(input) {
...
-
-    if (isSearchForm) {
-      searchTerm = $(element)
-        .find('.autocomplete-input')
-        .val();
-      if (searchTerm) {
-        uri.setSearch('search', searchTerm.trim());
-      }
-    }
+  values.page = (element !== undefined) ? $('#cur_page_num').val() : undefined;
```
tbrisker: We never pass undefined to the function, so i think we can get rid of all of these conditions, or just return early if the parameter is undefined if you want to be defensive.

tbrisker (changes requested):

ripcurld:
[test foreman]

ohadlevy:
@tbrisker whats the status of the PR? thanks.

webpack/assets/javascripts/foreman_tools.js
```diff
@@ -103,43 +103,23 @@ export function initTypeAheadSelect(input) {
...
-    ) {
-      uri.removeSearch('search');
-    }
-    uri.setSearch('per_page', perPage);
+  const searchTerm = $(element).find('.autocomplete-input').val();
+  if (searchTerm !== undefined) {
+    values.search = searchTerm.trim();
+    // if search term is empty set it to undefined so `uri.setSearch` will remove it
+    // from the search params if exists.
+    if (values.search === '') { values.search = undefined; }
```
tbrisker: Are you sure this is needed? I tried playing around with it a bit and looks like it works fine even if the search is blank.

webpack/assets/javascripts/foreman_tools.js
```diff
@@ -103,43 +103,23 @@ export function initTypeAheadSelect(input) {
...
-
-    if (isSearchForm) {
-      searchTerm = $(element)
-        .find('.autocomplete-input')
-        .val();
-      if (searchTerm) {
-        uri.setSearch('search', searchTerm.trim());
-      }
-    }
+  values.page = $('#cur_page_num').val();
```
tbrisker: Perhaps put this in `else` after the next condition?

tbrisker (changes requested):
A couple of minor comments inline, thanks!

webpack/assets/javascripts/foreman_tools.js
boaz0: It does work, but the "search" param is still shown in URI.
I don't mind remove this if-condition, though.

webpack/assets/javascripts/foreman_tools.js
boaz0: :+1: No problem.

boaz0:
[test foreman]

tbrisker (approved):
Tested, works well and code is much cleaner now. Thanks @ripcurld !

boaz0:
My pleasure @tbrisker 
Thanks for your guidance and help. :tada: :sparkles: :confetti_ball:

PR #5437, opened 2018-04-11
Fixes #23225 - fix vm tab js error in host page

amirfefer:
the host vm tab only catches `ActionView::Template::Error`, which works if the error is in the template rendering, but if `find_vm_by_uuid` fails it throws a `ActiveRecord::RecordNotFound` exception which isn't caught by the controller, leading to the generic controller reply which includes the entire layout - including JS which is executed and breaks all the JS on the page.


tbrisker (approved):
Thanks @amirfefer !

PR #5439, opened 2018-04-11
Fixes #23235 - Disable Style/IfUnlessModifier cop

mmoll:

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -79,3 +79,6 @@ Style/Alias:
 
 Layout/DotPosition:
   Enabled: false
+
+Style/IfUnlessModifier:
+  Enabled: false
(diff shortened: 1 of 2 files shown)


ekohl:
I'd argue this is less readable code. The lines are typically long and you might not see the if/unless at the end. Should we just disable this cop?

mmoll:
That syntax is quite Ruby-esque and personally I slightly like it, but I'm also happy to disable that cop.

ares:
I like the statement modifier syntax but only where it makes sense. No syntax linter can decide that. +1 for disabling this cop.

mmoll:
there we go :)

ekohl (approved):

lzap (approved):

mmoll:
anybody? :)

lzap:
3 vs 1, I think this one is not sensitive, merging, thanks!

PR #5442, opened 2018-04-12
Fixes #23212 - uninitialized constant

ananace:
Seems like the same issue also exists in 1.17, so it would probably have to be cherry-picked/pulled to there as well.

### DIFF
--- a/app/services/power_manager/virt.rb
+++ b/app/services/power_manager/virt.rb
@@ -32,7 +32,7 @@ module PowerManager
     def default_action(action)
       action_status = vm.send(action)
       # Temporary fix until fog-ovirt is updated
-      if action_status.is_a?(Fog::Compute::Ovirt::Server)
+      if Fog::Compute.const_defined?(:Ovirt) && action_status.is_a?(Fog::Compute::Ovirt::Server)
         return true
       end
       action_status


ekohl:
I see this is against 1.16. Does this also happen in develop?

ananace:
Seems like a951f9bdbe - that added this check - never ended up in develop, so the issue doesn't exist there.

ekohl:
Ah, this was introduced in the security fix. @orrabin could you have a look?

tbrisker:
ok to test

tbrisker:
@dLobatog, @xprazak2  - heads up, we'll need to release 1.16.2 and 1.17.1 with this fix

tbrisker:
@ananace I invited you to the foreman organization, if you confirm the invitation and make your membership public your next PRs will automatically kick CI runs.

orrabin:
@ekohl This was the fix for 1.15-1.17, develop has a new version of fog-ovirt that fixes this.
I tested with an ovirt machine on 1.17 and this still works as expected.
This should be tested with another compute resource since ovirt did not have the error.

orrabin (commented):
@ekohl This was the fix for 1.15-1.17, develop has a new version of fog-ovirt that fixes this.
I tested with an ovirt machine on 1.17 and this still works as expected.
This should be tested with another compute resource since ovirt did not have the error.

ohadlevy (approved):

dLobatog (approved):
Thanks @ananace !

mmoll:
I just ran into this with 1.16 and KVM, this patches fixes it. I understood it correctly that this patch is for 1.15, 1.16 and 1.17?

ananace:
Oh right, I completely missed to check if it applied to 1.15 as well - spoiler alert; It applies to 1.15.7+ as well.

mmoll:
merged, tack @ananace!

mmoll:
cherry-picked to the other stable branches as well.

misse:
Will this be applied to 1.17 as well?

mmoll:
yes, af4ed00

misse:
Should I be getting the upgrade from the yum repo then? Or isn't it published yet?

mmoll:
It will be part of 1.16.2 and 1.17.1, which are both not out, yet.

PR #5444, opened 2018-04-12
Fixes #23252 - Speed up unattended controller tests

tbrisker:
Splitting the tests into different contexts and only creating the hosts
needed for each context cut down runtime locally from 57s to 23s

### DIFF
--- a/test/controllers/unattended_controller_test.rb
+++ b/test/controllers/unattended_controller_test.rb
@@ -5,82 +5,421 @@ class UnattendedControllerTest < ActionController::TestCase
     Host::Managed.any_instance.stubs(:handle_ca).returns(true)
     as_admin do
       disable_orchestration # avoids dns errors
-      @ptable = FactoryBot.create(:ptable, :name => 'default',
-                                  :operatingsystem_ids => [operatingsystems(:redhat).id])
-      @ptable_ubuntu = FactoryBot.create(:ptable, :ubuntu, :name => 'ubuntu default',
-        :layout => 'd-i partman-auto/disk string /dev/sda\nd-i partman-auto/method string regular...',
-                                         :operatingsystem_ids => [operatingsystems(:ubuntu1010).id])
       @org = FactoryBot.create(:organization, :ignore_types => ['ProvisioningTemplate'])
       @loc = FactoryBot.create(:location, :ignore_types => ['ProvisioningTemplate'])
+    end
+  end
+
+  context "redhat" do
+    setup do
+      ptable = FactoryBot.create(:ptable, :name => 'default',
+                                  :operatingsystem_ids => [operatingsystems(:redhat).id])
       @rh_host = FactoryBot.create(:host, :managed, :with_dhcp_orchestration, :build => true,
                                     :operatingsystem => operatingsystems(:redhat),
-                                    :ptable => @ptable,
+                                    :ptable => ptable,
                                     :medium => media(:one),
                                     :architecture => architectures(:x86_64),
                                     :organization => @org,
                                     :location => @loc
                                    )
+    end
+
+    test "should get a kickstart" do
+      @request.env["HTTP_X_RHN_PROVISIONING_MAC_0"] = "eth0 #{@rh_host.mac}"
+      get :host_template, params: { :kind => 'provision' }
+      assert_response :success
+    end
+
+    test "should get a kickstart even if not using the first NIC" do
+      @request.env["HTTP_X_RHN_PROVISIONING_MAC_0"] = "unused NIC"
+      @request.env["HTTP_X_RHN_PROVISIONING_MAC_3"] = "eth4 #{@rh_host.mac}"
+      get :host_template, params: { :kind => 'provision' }
+      assert_response :success
+    end
+
+    test "should get a kickstart if MAC is provided" do
+      get :host_template, params: { :kind => 'provision', :mac => @rh_host.mac }
+      assert_response :success
+    end
+
+    test "should get a kickstart even if we are behind a loadbalancer" do
+      @request.env["HTTP_X_FORWARDED_FOR"] = @rh_host.ip
+      @request.env["REMOTE_ADDR"] = "127.0.0.1"
+      get :host_template, params: { :kind => 'provision' }
+      assert_response :success
+    end
+
+    test "should set @static when requested" do
+      Setting[:safemode_render]=false
+      @request.env["HTTP_X_RHN_PROVISIONING_MAC_0"] = "eth0 #{@rh_host.mac}"
+      get(:host_template, params: { 'kind' => 'provision', 'static' => 'true' })
+      assert_match(%r{static:true}, @response.body)
+    end
+
+    test "should render spoof pxelinux for a host" do
+      get :host_template, params: { :kind => 'PXELinux', :spoof => @rh_host.ip }, session: set_session_user
+      assert assigns(:initrd)
+      assert assigns(:kernel)
+      assert_response :success
+    end
+
+    test "should render spoof pxegrub for a host" do
+      get :host_template, params: { :kind => 'PXEGrub', :spoof => @rh_host.ip }, session: set_session_user
+      assert assigns(:initrd)
+      assert assigns(:kernel)
+      assert_response :success
+    end
+
+    test "should render spoof iPXE for a host" do
+      get :host_template, params: { :kind => 'iPXE', :spoof => @rh_host.ip }, session: set_session_user
+      assert assigns(:initrd)
+      assert assigns(:kernel)
+      assert_response :success
+    end
+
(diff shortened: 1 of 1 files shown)


test/controllers/unattended_controller_test.rb
```diff
@@ -291,208 +470,40 @@ class UnattendedControllerTest < ActionController::TestCase
...
-      h_new=Host.find_by_name(h.reload.name)
-      assert_response :success
-      assert_equal new_ip, h_new.ip
-    end
-
-    test "hosts with mismatched ip and update_ip true and a duplicate ip should succeed with no ip update" do
-      disable_orchestration # avoids dns errors
-      Setting[:token_duration] = 30
-      Setting[:update_ip_from_built_request] = true
-      @request.env["REMOTE_ADDR"] = @rh_host.ip
```
tbrisker, line 420: This is the only test using both hosts, but it doesn't really matter which existing host this is AFAICT so i replaced this with just a new host.

test/controllers/unattended_controller_test.rb
```diff
@@ -291,208 +470,40 @@ class UnattendedControllerTest < ActionController::TestCase
...
-      refute_equal new_ip, h.ip
-      h.create_token(:value => "aaaaab", :expires => Time.now.utc + 5.minutes)
-      get :built, params: { 'token' => h.token.value }
-      h_new=Host.find_by_name(h.reload.name)
-      assert_response :success
-      assert_equal new_ip, h_new.ip
-    end
-
-    test "hosts with mismatched ip and update_ip true and a duplicate ip should succeed with no ip update" do
-      disable_orchestration # avoids dns errors
```
tbrisker, line 417: we already do this in the general setup, so i removed this here and in other places.

test/controllers/unattended_controller_test.rb
```diff
@@ -28,226 +211,229 @@ class UnattendedControllerTest < ActionController::TestCase
...
+    assert_routing '/unattended/built', {:controller => 'unattended', :action => 'built', :format => 'text'}
   end
 
   test "should not provide unattended files to hosts which we don't know about" do
     get :host_template, params: { :kind => 'provision' }
     assert_response :not_found
   end
 
-  test "should not provide unattended files to hosts which don't have an assign os" do
-    setup_users
```
tbrisker, line 262: This isn't actually needed here.

tbrisker (commented):
Most of this is just moving code around into contexts, i commented on the only actual changes.

lzap (dismissed):

test/controllers/unattended_controller_test.rb
```diff
@@ -291,208 +470,40 @@ class UnattendedControllerTest < ActionController::TestCase
...
-      Setting[:token_duration]    = 30
-      Setting[:unattended_url]    = "http://test.host"
-      @request.env["REMOTE_ADDR"] = @ub_host.ip
-      @ub_host.create_token(:value => "aaaaaa", :expires => Time.now.utc + 5.minutes)
-      get :host_template, params: { :kind => 'provision' }
-      assert @response.body.include?("http://test.host/unattended/finish?token=aaaaaa")
+      Setting[:unattended_url] = "http://test.host"
+      ProxyAPI::Template.any_instance.stubs(:template_url).returns(nil)
+      @request.env["REMOTE_ADDR"] = '127.0.0.1'
+      @host_with_template_subnet .create_token(:value => "aaaaae", :expires => Time.now.utc + 5.minutes)
```
timogoebel: Nit: superfluous space before `.create`

test/controllers/unattended_controller_test.rb
```diff
@@ -5,13 +5,15 @@ class UnattendedControllerTest < ActionController::TestCase
...
-        :layout => 'd-i partman-auto/disk string /dev/sda\nd-i partman-auto/method string regular...',
-                                         :operatingsystem_ids => [operatingsystems(:ubuntu1010).id])
       @org = FactoryBot.create(:organization, :ignore_types => ['ProvisioningTemplate'])
       @loc = FactoryBot.create(:location, :ignore_types => ['ProvisioningTemplate'])
+    end
+  end
+
+  context "redhat" do
+    setup do
+      ptable = FactoryBot.create(:ptable, :name => 'default',
```
timogoebel, line 15: `let(:ptable) {}` ?

test/controllers/unattended_controller_test.rb
```diff
@@ -5,13 +5,15 @@ class UnattendedControllerTest < ActionController::TestCase
...
+                                  :operatingsystem_ids => [operatingsystems(:redhat).id])
       @rh_host = FactoryBot.create(:host, :managed, :with_dhcp_orchestration, :build => true,
```
timogoebel, line 17: `let(:host)` ?

test/controllers/unattended_controller_test.rb
```diff
@@ -20,6 +22,187 @@ class UnattendedControllerTest < ActionController::TestCase
...
+
+    test "should get a kickstart even if we are behind a loadbalancer" do
+      @request.env["HTTP_X_FORWARDED_FOR"] = @rh_host.ip
+      @request.env["REMOTE_ADDR"] = "127.0.0.1"
+      get :host_template, params: { :kind => 'provision' }
+      assert_response :success
+    end
+
+    test "should set @static when requested" do
+      Setting[:safemode_render]=false
```
timogoebel, line 53: Do we need to reset this? Settings are cached using Rails cache. Caches aren't cleared between tests afaik, see #5142.

test/controllers/unattended_controller_test.rb
```diff
@@ -20,6 +22,187 @@ class UnattendedControllerTest < ActionController::TestCase
...
+                         :search => "name = does_not_exist")
+      get :host_template, params: { :kind => 'PXELinux', :spoof => @rh_host.ip, :format => 'text' }, session: set_session_user(user)
+      assert_response :not_found
+      assert_match /unable to find a host/, @response.body
+    end
+  end
+
+  context "ubuntu" do
+    setup do
+      ptable_ubuntu = FactoryBot.create(:ptable, :ubuntu, :name => 'ubuntu default',
```
timogoebel, line 225: `let(:ptable) {}` ?

timogoebel (commented):
@tbrisker: Thanks, looks like a great improvement. Some comments inline. :-)

test/controllers/unattended_controller_test.rb
tbrisker: oops. fixed.

test/controllers/unattended_controller_test.rb
tbrisker: I suppose we should, but i didn't want to change the tests too much.

tbrisker:
@timogoebel i tried playing around with let a bit but it seems to be more complicated than just replacing the ivar with let block.

mmoll:
@tbrisker please rebase :)

lzap:
Out of curiosity, what's the problem with `let`? I've just discovered it the other day, 've fallen in love, don't say it's not good! :-)

tbrisker:
@lzap tbh i didn't bother too much with trying to fix it, i tried replacing some vars and tests failed so i just let it go for now since i wanted to get the most significant improvement in with the minimal effort.

tbrisker:
Rebased.

mmoll (approved):
for me this is good, but waiting on other's to merge because of the let thingy

lzap (approved):
Although I prefer `let` syntax in general, this is a refactoring so I am fine with this. New code, I'd be more strict for sure. @timogoebel

timogoebel (approved):
LGTM!

timogoebel (dismissed):
@tbrisker: Could you please take a look at the Redmine issue number?

http://projects.theforeman.org/issues/23525

returns a 404 for me.

lzap:
The correct one is: 23252 - I think we can squash and merge via github interface? Let me try to dismiss your remark to enable the button.

iNecas:
FYI we started hitting intermittent test failures in the unattended_test code: haven't seen those before, therefore I suspect this could be related. 

I've seen this in foreman core tests already several times (http://ci.theforeman.org/view/Core/job/test_develop/3426/database=mysql,ruby=2.5,slave=fast/consoleFull), as well as in rex tests (
http://ci.theforeman.org/job/test_plugin_matrix/5176/database=mysql,ruby=2.4,slave=fast/testReport/junit/(root)/UnattendedControllerTest/test_0004_should_get_a_kickstart_if_MAC_is_provided_with_two_hosts_with_same_MAC/)

So far, I've seen in only on mysql, if that helps with reproducing.

tbrisker:
@iNecas I've seen this failure several times as well, however the location of the failure indicates it is failing on a branch without this commit.

mmoll:
that's the one from GH-5424

PR #5445, opened 2018-04-12
Fixes #23254 - Utilize webmock to prevent network connection

lzap:
We have about dozen of places where we actually do connections or DNS queries, this can significantly slow down tests when used in loop or many tests. Let's prevent from that.

We also have http://localhost in few places, I will go ahead and replace
that in a separate commit (might be a change), so we are sure these are
not hidden.

Recommendation for writing tests would be to use hostnames other than
localhost when creating any URL string.

### DIFF
--- a/lib/foreman/http_proxy.rb
+++ b/lib/foreman/http_proxy.rb
@@ -17,7 +17,7 @@ module Foreman
     def http_proxied_rescue(&block)
       yield
     rescue => e
-      raise e, _("Proxied request failed with: %s") % e, e.backtrace
+      raise e, _("Proxied request failed with: %s\n%s") % [e, e&.backtrace&.join("\n")]
     end
 
     private
--- a/lib/foreman/http_proxy/excon_connection_extension.rb
+++ b/lib/foreman/http_proxy/excon_connection_extension.rb
@@ -1,17 +1,23 @@
-Excon::Connection.class_eval do
-  include Foreman::HTTPProxy
-  alias_method :orig_request, :request
-
-  def request(params, &block)
-    if proxy_http_request?(@data[:proxy], @data[:host], @data[:scheme])
-      log_proxied_request(http_proxy, @data[:host])
-      @data[:proxy] = http_proxy
-      setup_proxy
-      http_proxied_rescue do
-        orig_request(params, &block)
+module Foreman
+  module HttpProxy
+    module ExconConnectionExtension
+      def request(params, &block)
+        if proxy_http_request?(@data[:proxy], @data[:host], @data[:scheme])
+          log_proxied_request(http_proxy, @data[:host])
+          @data[:proxy] = http_proxy
+          setup_proxy
+          http_proxied_rescue do
+            super(params, &block)
+          end
+        else
+          super(params, &block)
+        end
       end
-    else
-      orig_request(params, &block)
     end
   end
+end
+
+Excon::Connection.class_eval do
+  include Foreman::HTTPProxy
+  prepend Foreman::HttpProxy::ExconConnectionExtension
 end
--- a/test/controllers/smart_proxies_controller_test.rb
+++ b/test/controllers/smart_proxies_controller_test.rb
@@ -4,6 +4,10 @@ class SmartProxiesControllerTest < ActionController::TestCase
   basic_pagination_rendered_test
   basic_pagination_per_page_test
 
+  setup do
+    SmartProxy.any_instance.stubs(:associate_features).returns(true)
+  end
+
   def test_index
     get :index, session: set_session_user
     assert_template 'index'
--- a/test/unit/compute_resource_host_associator_test.rb
+++ b/test/unit/compute_resource_host_associator_test.rb
@@ -9,7 +9,11 @@ class ComputeResourceHostAssociatorTest < ActiveSupport::TestCase
   let(:vm2) { stub('vm2', :identity => Foreman.uuid) }
   let(:vm3) { stub('vm3', :identity => Foreman.uuid) }
   let(:vms) { [vm1, vm2, vm3] }
-  let(:host_with_vm) { FactoryBot.create(:host, :compute_resource => compute_resource) }
+  let(:host_with_vm) do
+    FactoryBot.create(:host, :compute_resource => compute_resource).tap do |host|
+      host.stubs(:vm_exists?).returns(true)
+    end
+  end
   let(:host_without_vm) { FactoryBot.build(:host) }
 
   test 'associates vm with a host if they match' do
--- a/test/unit/foreman/http_proxy_test.rb
+++ b/test/unit/foreman/http_proxy_test.rb
@@ -6,9 +6,9 @@ class HTTPProxyTest < ActiveSupport::TestCase
   end
 
   let(:adapter) { DummyHTTPAdapter.new }
-  let(:http_proxy) { 'http://proxy:3218' }
+  let(:http_proxy) { 'http://dummyproxy.theforeman.org:3218' }
   let(:excepted_hosts) { [] }
-  let(:request_host) { 'remotehost.xyz' }
+  let(:request_host) { 'dummyproxy.theforeman.org' }
   let(:schema) { 'http' }
 
   setup do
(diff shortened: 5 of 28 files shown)


test/controllers/smart_proxies_controller_test.rb
```diff
@@ -196,6 +200,7 @@ def test_refresh_fail
 
   test '#expire_logs' do
     proxy = smart_proxies(:logs)
+    proxy.stubs(:associate_features)
```
timogoebel: Why do we need this when we have the global `setup do`?

test/unit/foreman/http_proxy_test.rb
```diff
@@ -118,16 +118,19 @@ class DummyHTTPAdapter
     end
 
     test 'set @data[:proxy] to proxy' do
-      net_http.request({})
-      assert_equal URI.parse(http_proxy),
-                   net_http.instance_variable_get(:@proxy_address)
+      # stub webmock and the stub our own extension on to of that
```
timogoebel: nitpick: `on top of` (missing p)

timogoebel (commented):
👍 , some comments inline.

test/controllers/smart_proxies_controller_test.rb
lzap: Oh a leftover, thanks.

test/unit/foreman/http_proxy_test.rb
lzap: Fixed.

lzap:
Ok added the localhost URL replacement, nothing really big. Amended Timo's comments as well.

test/unit/compute_resource_host_associator_test.rb
```diff
@@ -9,7 +9,11 @@ class ComputeResourceHostAssociatorTest < ActiveSupport::TestCase
   let(:vm2) { stub('vm2', :identity => Foreman.uuid) }
   let(:vm3) { stub('vm3', :identity => Foreman.uuid) }
   let(:vms) { [vm1, vm2, vm3] }
-  let(:host_with_vm) { FactoryBot.create(:host, :compute_resource => compute_resource) }
+  let(:host_with_vm) do
+    h = FactoryBot.create(:host, :compute_resource => compute_resource)
```
timogoebel:
```ruby
FactoryBot.create(:host, :compute_resource => compute_resource).tap do |host|
 host.stubs(:vm_exists?).returns(true)
end
```

test/unit/foreman/http_proxy_test.rb
```diff
@@ -118,16 +118,19 @@ class DummyHTTPAdapter
...
 
     test 'set @data[:proxy] to proxy' do
-      net_http.request({})
-      assert_equal URI.parse(http_proxy),
-                   net_http.instance_variable_get(:@proxy_address)
+      # stub webmock and the stub our own extension on top of that
+      stub_request(:get, "http://remotehost.xyz/features").to_return(status: 200, body: "", headers: {})
+      load "foreman/http_proxy/net_http_extension.rb"
+      net_http.request(Net::HTTP::Get.new("/features"))
+      assert_equal URI.parse(http_proxy), net_http.instance_variable_get(:@proxy_address)
```
lzap: This broke jenkins, @timogoebel any idea how to test this? The thing is, webmock shadows our own monkey code which sets the proxy_address variable, I tried to overwrite it but it ends up in "Stack too deep" because its calling its own code. I wonder if this is testable at all.

timogoebel: I need to take a look at this before I can answer that, will do next week.

ekohl: Wouldn't you need to use webmock to actually mock a response instead of mocking the library functions?

test/unit/compute_resource_host_associator_test.rb
lzap: Kinda hate tap, does not bring any improvement in readability, it's line fo line. But let me change that, don't mind.

test/unit/foreman/http_proxy_test.rb
lzap: Guys I figured it out, thanks for help. Basically I created an alias and testing the alias which calls the mocked method back. It works as expected now.

lzap:
Ok pushing, locally it's green except the JS faulty one:

```
Error:
HostgroupJSTest#test_0003_parameters change after parent update:
Capybara::Poltergeist::MouseEventFailed: Firing a click at co-ordinates [293, 80.5] failed. Poltergeist detected another element with CSS selector 'html.layout-pf.layout-pf-fixed.transitions body nav#topbar.navbar.navbar-pf-vertical nav.collapse.navbar-collapse' at this position. It may be overlapping the element you are trying to interact with. If you don't care about overlapping elements, try using node.trigger('click').
    test/integration/hostgroup_js_test.rb:112:in `block in <class:HostgroupJSTest>'

Finished in 1534.274069s, 4.3903 runs/s, 10.8748 assertions/s.
6736 runs, 16685 assertions, 0 failures, 1 errors, 1 skips
```

lib/foreman/http_proxy.rb
```diff
@@ -17,7 +17,7 @@ def proxy_http_request?(current_proxy, request_host, schema)
     def http_proxied_rescue(&block)
       yield
     rescue => e
-      raise e, _("Proxied request failed with: %s") % e, e.backtrace
+      raise e, _("Proxied request failed with: %s\n%s") % [e, e&.backtrace&.join("\n")]
```
timogoebel, line 20: Wold that be a good use case for `Foreman::Logging.exception`?

lib/foreman/http_proxy/excon_connection_extension.rb
```diff
@@ -2,7 +2,7 @@
   include Foreman::HTTPProxy
   alias_method :orig_request, :request
 
-  def request(params, &block)
+  def request_with_foreman_extensions(params, &block)
```
timogoebel: Let's please rewrite this using `Module.prepend` or we'll get into the StackTrace to deep hell again.

Something like:
```ruby
module ForemanExtensions
  def request(params, &block)
      do_whatever()
      super
     do_more_after()
  end
end

prepend ForemanExtensions
```

Same below.

timogoebel (commented):
Some more comment from a quick glance at the code, thanks @lzap.

lib/foreman/http_proxy/excon_connection_extension.rb
lzap: I can't, tried that. I was not able to figure out how to call prepend on an instance in the test actually. My last version works fine locally, where is a problem again?

lib/foreman/http_proxy.rb
lzap: Indeed.

test/unit/foreman/http_proxy_test.rb
```diff
@@ -88,22 +88,20 @@ class DummyHTTPAdapter
...
       excon_connection.stubs(:setup_proxy).returns
       excon_connection.stubs(:proxy_http_request?).returns(true)
     end
 
     test 'set @data[:proxy] to proxy' do
-      excon_connection.request({})
-      assert_equal http_proxy,
-                   excon_connection.instance_variable_get(:@data)[:proxy]
+      stub_request(:get, "http://#{request_host}/features").to_return(status: 200, body: "", headers: {})
+      excon_connection.request_with_foreman_extensions({host: request_host, path: "/features", headers: {}})
```
lzap: Let me explain, on this line I need to call the foreman extended method, unless it has an explicit name, I can't do that because webmock always shadows original and foreman versions and I can't access it. I was thinking like via some dirty trick like `object.successors.second.call(:request)` but then I realized that webmock simply shadows/overrides request method, there are no successors at all.

lzap:
The [test] did not finish, but it was green. I want to see if there are some tests with web requests left on jenkins.

lib/foreman/http_proxy/excon_connection_extension.rb
lzap: Figured it out.

lzap:
Rebased using `prepend`.

lib/foreman/http_proxy.rb
mmoll: @lzap this will stay as-is in this PR?

timogoebel:
[test foreman]

lzap:
What happened? Random failures?

timogoebel:
> What happened? Random failures?

Yeah, just audit failures one a single ruby version. I want to make sure this is unrelated.

lzap:
Bunch of statuses() failures, not sure why that happens. I was hoping this might help https://github.com/theforeman/foreman/pull/5472 but I am unsure now. FYI those also errors out for different PRs.

mmoll:
yeah, audit and proxy status failures are unrelated

mmoll (approved):

mmoll:
merged, díky @lzap!

lzap:
Thanks, https://community.theforeman.org/t/heads-up-making-connections-now-disallowed-in-tests/8868

PR #5447, opened 2018-04-12
Fixes #23255 - Speed up seeds test

tbrisker:
Every seed run takes about 10s. This combines 6 tests to run on the same
seed action, leading to about 60s saved locally.

### DIFF
--- a/test/unit/tasks/seeds_test.rb
+++ b/test/unit/tasks/seeds_test.rb
@@ -25,10 +25,28 @@ class SeedsTest < ActiveSupport::TestCase
     User.current = nil
   end
 
-  test 'populates features' do
-    count = Feature.count
+  test 'populates multiple tables' do
+    tables = [Feature, Ptable, ProvisioningTemplate, Medium, Bookmark, AuthSourceExternal]
+
+    tables.each do |model|
+      assert model.unscoped.count.zero?
+    end
+
     seed
-    assert_not_equal count, Feature.count
+
+    tables.each do |model|
+      refute model.unscoped.count.zero?
+    end
+
+    refute Ptable.unscoped.where(:os_family => nil).any?
+    refute Medium.unscoped.where(:os_family => nil).any?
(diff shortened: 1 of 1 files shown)


lzap:
Thanks, wanted to do the same but was busy with webmock thing. Anyway, first question - after DatabaseCleaner README read, I think it is not being used at all - we should explicitly call it before and after test runs.

https://github.com/DatabaseCleaner/database_cleaner

Can you confirm it works?

lzap:
If my assumption that the most expensive call is the seed method correct, we should go much further. I suggest to extract the loading code into some helper which can hand over list of Ruby seed files which can be then loaded via regexp one by one. Each test could load just what is needed. We would need to keep some basic seeds because of some dependencies but should be possible I think.

tbrisker:
[test foreman] for some reason the test results aren't available on jenkins so we can't see the affect on times, trying agian

tbrisker:
[test foreman]

tbrisker:
@lzap Your assumption is correct, the slowest part here is running the seeds. I think we should keep using the whole seed, we want to make sure that when we seed everything is in the state we expect it to be and works correctly, doing only partial seeding will not give us this capability. We could try to see if more tests can be combined as well, these seemed like the lowest hanging fruit.

lzap (approved):
Don't want to push here, any second counts! Thanks.

lzap:
It's 2.5 ruby tests what makes safemode to fail the suite, PR was merged in safemode.

https://github.com/svenfuchs/safemode/commit/65fcbbf933821ce1f5b5a0117b81e1122eb8d937

I asked for version bump: https://github.com/svenfuchs/safemode/pull/27

mmoll:
merged, thanks @tbrisker!

lzap:
Great work @tbrisker ! My digging which went towards implementing webmock did not shown such an improvement, on the other hand it's good to prevent internet/DNS access anyway... :-)

PR #5448, opened 2018-04-12
Fixes #23256 - Check if the delayed plans have expected interface

adamruzicka:
Not all delayed plan records have 'job_class' key in first element in serialized_args.


adamruzicka:
This is required for Sat6.4

ares (approved):
Ack pending Jenkins

ares:
Adding to 1.18 as this can fail if there's other job planned without this interface which is the case if you use Foreman with other plugins

ares:
Failures on ruby 2.5 seems unrelated, makes sense and fixes the problem, merging. Thanks @adamruzicka

PR #5449, opened 2018-04-12
Fixes #23263 - support namespaced models in audits

ares:
@tbrisker please take a look if you have a moment, until we have proper mapping I think this is reasonable midterm solution.


tbrisker (approved):
While this will only show ids afaict (unless we saved some other value), i guess it is better than n/a, at least until we rework this view to give better results in general.

ares:
> While this will only show ids afaict (unless we saved some other value), i guess it is better than n/a, at least until we rework this view to give better results in general.

:+1:

PR #5451, opened 2018-04-13
Fixes #23223 - Don't allow Dynflow to initialize in db:drop rake

adamruzicka:

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -276,7 +276,7 @@ module Foreman
     end
 
     config.after_initialize do
-      init_dynflow
+      init_dynflow unless Foreman.in_rake?('db:drop')
       setup_auditing
     end
 


config/application.rb
```diff
@@ -276,7 +276,7 @@ class Application < Rails::Application
     end
 
     config.after_initialize do
-      init_dynflow
+      init_dynflow unless Foreman.in_rake?('db:drop')
```
ekohl, line 279: Are there any rake tasks where it does need to be initialized?

ekohl:
I've moved the redmine issue.

config/application.rb
adamruzicka: For a reason I don't recall now we load the Dynflow executor in `db:seed` and `db:migrate` (see[1]). Another place where we might be using it are foreman-tasks tests.

[1] - https://github.com/Dynflow/dynflow/blob/master/lib/dynflow/rails/configuration.rb#L40

adamruzicka: cc @iNecas

iNecas: The reason is that at the gtime of db:seed and db:migrate, the external executor might not be running, while we would need it for some seed of migration operations.

iNecas (approved):

iNecas:
@adujicek

mmoll:
test failures unrelated

mmoll:
merged, díky @adamruzicka!

config/application.rb
mbacovsky: does this prevent us to use `db:purge`?

mbacovsky: as per our discussion with @adamruzicka and some testing it does. However after some investigation I found out I won't be able to use `db:purge` anyway.

PR #5019, opened 2017-11-21
Fixes #22847 - Use pf-react donut chart in facts page

lizagilman:

PR #5454, opened 2018-04-15
Fixes #23010 - add depriciation warnings to flotchart

boaz0:
Adding `@deprecated` JSDoc above all flotchart functions in `charts.js` to notify plugin authors that `flot` is being replaced by `c3`.

Is it also necessary to add `console.log` or `console.warn` to these functions? WDYT @ohadlevy @ares

Thanks.

### DIFF
--- a/app/assets/javascripts/charts.js
+++ b/app/assets/javascripts/charts.js
@@ -1,3 +1,6 @@
+/*
+ * @deprecated: flot js is not maintained anymore in favor of c3 (http://c3js.org/)
+ */
 $.fn.flot_pie = function(){
   var options = arguments[0] || {};
   $(this).each(function(i,el){
@@ -115,6 +118,9 @@ function expanded_pie(target, data){
   });
 }
 
+/*
+ * @deprecated: flot js is not maintained anymore in favor of c3 (http://c3js.org/)
+ */
 $.fn.flot_bar = function(){
   var options = arguments[0] || {};
   $(this).each(function(i,el){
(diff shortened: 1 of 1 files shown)


xprazak2:
I think you can use the `deprecate` function from [foreman_tools](https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/foreman_tools.js#L68) to do that for you.

tbrisker:
@ripcurld do we have a replacement to all these of these functions? if not, I don't think we should deprecate them until we have a working implementation in core that plugin maintainers can use instead of these functions, even if the dependent library is deprecated.

boaz0:
Yep, I have been looking into it and `flot_pie` can be replaced by `DonutChart`. However, I am not sure all plugins are using `react`.

Does the plugin need to install `react` in order to use `react` components from `Foreman`?

tbrisker:
@ripcurld what about the other functions? I think we don't have a working implementation for them yet. @sharvit / @amirfefer  - do you know what the plan is for getting rid of flot completely?

ohadlevy:
there is also a js function for deprecation https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/foreman_tools.js#L68

sharvit:
@tbrisker we do have a plan to get rid of `flot` completely, we prefer migrating to `patternfly-react-charts`.
http://projects.theforeman.org/issues/22812

boaz0:
@sharvit is that OK if I work on http://projects.theforeman.org/issues/22812 ?

ohadlevy:
ack, note there is also https://github.com/theforeman/foreman/pull/5019

sharvit:
@ripcurld yes go for it! 👍 👍

boaz0:
Closing https://github.com/theforeman/foreman/pull/5454 as this will be handled in https://github.com/theforeman/foreman/pull/5535

PR #5455, opened 2018-04-15
Fixes #23195 - Audit associations on creation and destruction

tbrisker:
Also includes a bit of refactoring to the AuditAssociations module.

### DIFF
--- a/app/models/concerns/audit_associations.rb
+++ b/app/models/concerns/audit_associations.rb
@@ -1,15 +1,11 @@
 module AuditAssociations
-  def self.included(base)
-    base.send :include, InstanceMethods
-  end
-
-  module InstanceMethods
+  module AssociationsChanges
     def changes_to_save
-      if audited_options[:associations].present?
-        super.merge(associated_changes)
-      else
-        super
-      end
+      super.merge(associated_changes)
+    end
+
+    def audited_attributes
+      super.merge(associated_attributes)
     end
 
     private
@@ -19,56 +15,53 @@ module AuditAssociations
     end
 
     def associated_changes
-      associations = Array.wrap(audited_options[:associations])
-      associations.each_with_object({}) do |association_name, changes_hash|
-        association_ids = "#{association_name.to_s.singularize}_ids"
-        if send("#{association_ids}_changed?")
-          association_class = find_association_class(association_name)
-          change = send("#{association_ids}_change")
-          change_ids = change.flatten.uniq
+      audited_options[:associations].each_with_object({}) do |association, changes|
+        association_ids = "#{association.to_s.singularize}_ids"
+        if public_send("#{association_ids}_changed?")
+          change = public_send("#{association_ids}_change")
 
-          id_name_map = association_class.where(id: change_ids).inject({}) { |r, p| r.merge(p.id => p.to_label) }
-
-          changes_hash[association_name] = change.each_with_object([]) do |ids, humaized_associations|
-            humaized_associations << ids.each_with_object([]) do |id, humanized_ids|
-              humanized_ids << id_name_map[id]
-            end.join(', ')
+          changes[association] = change.map do |ids|
+            associated_names(association, ids)
           end
         end
       end
     end
-  end
 
-  module Auditor
-    def self.included(base)
-      base.extend ClassMethods
+    def associated_attributes
+      audited_options[:associations].each_with_object({}) do |association, attributes|
+        ids = public_send("#{association.to_s.singularize}_ids")
+        attributes[association.to_s] = associated_names(association, ids)
+      end
     end
 
-    module ClassMethods
-      include Audited::Auditor::ClassMethods
+    def associated_names(association, ids)
+      find_association_class(association).where(id: ids).map(&:to_label).sort.join(', ')
+    end
+  end
 
-      def audited(options = {})
-        if options[:associations].present?
-          configure_dirty_associations(Array(options[:associations]))
-        end
-
-        super
+  module AssociationsDefinitions
+    def audited(options = {})
+      options[:associations] = Array(options[:associations])
+      if options[:associations].present?
+        configure_dirty_associations(options[:associations])
       end
 
-      def audit_associations(*associations)
-        new_associations = Array(associations)
-        if self.respond_to?(:audited_options)
-          configure_dirty_associations(new_associations)
-          self.audited_options[:associations] = Array(self.audited_options[:associations]) | new_associations
-        else
-          logger.warn "ignoring associations #{new_associations.join(', ')} audit definition for #{self}, the resource is not audited"
-        end
-      end
+      super
+    end
 
-      def configure_dirty_associations(associations)
-        include DirtyAssociations unless included_modules.include?(DirtyAssociations)
-        dirty_has_many_associations(*associations)
+    def audit_associations(*associations)
+      new_associations = Array(associations)
+      if self.respond_to?(:audited_options)
+        configure_dirty_associations(new_associations)
+        self.audited_options[:associations] = self.audited_options[:associations] | new_associations
+      else
+        logger.warn "ignoring associations #{new_associations.join(', ')} audit definition for #{self}, the resource is not audited"
       end
+    end
+
+    def configure_dirty_associations(associations)
+      include DirtyAssociations unless included_modules.include?(DirtyAssociations)
+      dirty_has_many_associations(*associations)
     end
   end
 end
--- a/app/models/concerns/taxonomix.rb
+++ b/app/models/concerns/taxonomix.rb
@@ -19,7 +19,7 @@ module Taxonomix
     scoped_search :relation => :organizations, :on => :id, :rename => :organization_id, :complete_enabled => false, :only_explicit => true, :validator => ScopedSearch::Validators::INTEGER
 
     dirty_has_many_associations :organizations, :locations
-    audit_associations :organizations, :locations
+    audit_associations :organizations, :locations if self.respond_to? :audit_associations
 
     validate :ensure_taxonomies_not_escalated, :if => Proc.new { User.current.nil? || !User.current.admin? }
   end
--- a/test/models/user_test.rb
+++ b/test/models/user_test.rb
@@ -1216,7 +1216,7 @@ class UserTest < ActiveSupport::TestCase
       audited_changes = recent_audit.audited_changes[:roles]
 
       assert audited_changes, 'No audits found for user-roles'
-      assert_equal [@role.name, Role.default].join(', '), audited_changes.first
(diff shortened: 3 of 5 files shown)


app/models/concerns/taxonomix.rb
```diff
@@ -19,7 +19,7 @@ module Taxonomix
     scoped_search :relation => :organizations, :on => :id, :rename => :organization_id, :complete_enabled => false, :only_explicit => true, :validator => ScopedSearch::Validators::INTEGER
 
     dirty_has_many_associations :organizations, :locations
-    audit_associations :organizations, :locations
+    audit_associations :organizations, :locations if self.respond_to? :audit_associations
```
tbrisker, line 22: We include `Taxonomix` in `Audited::Audit`, which inherits from `ActiveRecord` and not `ApplicationRecord`, so it doesn't respond to `audit_associations` any longer. This is intentional, we only want models inheriting `ApplicationRecord` (i.e. those defined by us) to be able to audit their associations.

app/models/concerns/audit_associations.rb
```diff
@@ -19,56 +15,53 @@ def find_association_class(name)
...
-        end
+    def associated_names(association, ids)
+      find_association_class(association).where(id: ids).map(&:to_label).to_sentence
+    end
+  end
 
-        super
+  module AssociationsDefinitions
+    def audited(options = {})
+      options[:associations] = Array(options[:associations])
```
tbrisker, line 44: Normalizing the option to an array here (even if it's nil - so it's turned to empty array) simplifies handling it in other places.

app/models/concerns/audit_associations.rb
```diff
@@ -1,15 +1,11 @@
 module AuditAssociations
-  def self.included(base)
-    base.send :include, InstanceMethods
```
tbrisker, line 3: This is pointless, we can just include the module directly (or prepend, in this case, since we want to override the `audited_attributes` method)

app/models/concerns/audit_associations.rb
```diff
@@ -1,15 +1,11 @@
...
-  end
-
-  module InstanceMethods
+  module AssociationsChanges
```
tbrisker, line 2: Renamed the submodules to names that convey what they actually do.

app/models/concerns/audit_associations.rb
```diff
@@ -1,15 +1,11 @@
...
     def changes_to_save
-      if audited_options[:associations].present?
-        super.merge(associated_changes)
```
tbrisker, line 9: Since we now normalize the option, we can always merge - if there isn't any associations, it will just merge with an empty hash.

kgaikwad:
@tbrisker,
Code wise looks good to me.

app/models/concerns/taxonomix.rb
ares: Does that mean the association audit is ignored if you include Taxonomix into a model that inherits from `ActiveRecord::Model` directly? Have you seen such class? Should we perhaps extend the `ActiveRecord::Model` with association extension then? Or you'd prefer to wait until audited gem works with `ApplicationRecord`?

test/models/user_test.rb
```diff
@@ -1216,7 +1216,7 @@ def editing_self_helper
       audited_changes = recent_audit.audited_changes[:roles]
 
       assert audited_changes, 'No audits found for user-roles'
-      assert_equal [@role.name, Role.default].join(', '), audited_changes.first
+      assert_equal [Role.default, @role.name].to_sentence, audited_changes.first
```
ares: what caused the order change? and is it desired to use `to_sentence` which uses "and" instead of the last comma

app/models/concerns/audit_associations.rb
ares: :+1: I suppose similar change is planned also in audited, right?

ares (changes requested):
code wise, I have mostly questions but from testing I found following

When I create a domain and assign organization as part of the process, when I go to that audit detail, I'm getting
`undefined method `humanize' for :organizations:Symbol`. This is what I seen in DB

```
Audit.last.audited_changes
=> {"name"=>"test of audit", "fullname"=>"", "dns_id"=>nil, :organizations=>"Foreman", :locations=>"Default Location"}
```

the same for destroy operation

The same happens for all associations.

app/models/concerns/taxonomix.rb
tbrisker: `ApplicationRecord` was created exactly for this purpose - to allow extending `ActiveRecord` without affecting dependencies that may also inherit from it. All of our models should be inheriting from ApplicationRecord, and the current extensions we have for ActiveRecord should be migrated to ApplicationRecord.

ares: right, so `Audited::Audit` should start inheriting from ApplicationRecord at some point? and then the condition is not required?

tbrisker: no, Audited::Audit should continue inheriting ActiveRecord so it doesn't break when some application like us monkey-patches ApplicationRecord for our needs

app/models/concerns/audit_associations.rb
tbrisker: There is a PR open in audited to change associated_with to be `has_many` instead of `has_one`. There still isn't support there for tracking changes in associations. (but PRs welcome :wink: )

test/models/user_test.rb
tbrisker: the change to `to_sentence`. To me it makes sense to use it instead of joining with `,`, since that's what we do everywhere we display a list.

tbrisker:
@ares fixed the issue you found, turns out attributes are a regular hash so `:locations` isn't the same as `'locations'`.

app/models/concerns/audit_associations.rb
ares: I meant for existing options in audited, but I see that's already happening https://github.com/collectiveidea/audited/blob/master/lib/audited/auditor.rb#L363-L368

app/models/concerns/taxonomix.rb
ares: Even if it would be Rails 5 compatible only? ApplicationRecord is common ancestor for all models by default so I don't understand why audited gem would extend core rails class. But that's definitely unrelated to this PR. I just don't like the fact the definition would be skipped silently. Would it be possible to log it? If this applies only for Audited::Audit, it would be one log on app start. If that happens for any other resource, something is wrong and it's better to know.

test/models/user_test.rb
ares: I'm not sure we're consistent with this. The only concern I had was, what if the resource name contains the word `and`, there's no way to distinguish.

I still do not understand the order change, .to_sentence keeps the order of array, join does too, how comes the test needed to reverse the order?

```
irb(main):001:0> ['b', 'a'].join(', ')
=> "b, a"
irb(main):002:0> ['b', 'a'].to_sentence
=> "b and a"
```

ares:
Test failures seems to be related

app/models/concerns/taxonomix.rb
tbrisker: No, Audited::Audit will always inherit ActiveRecord, since we want it to work as expected for all rails applications, even if they have some wierd monkey-patching in place in their ApplicationRecords that may break the gem. See https://blog.bigbinary.com/2015/12/28/application-record-in-rails-5.html for more info regarding this change.

test/models/user_test.rb
tbrisker: The change in order is due to the different way the change list was created before:
```ruby
changes_hash[association_name] = change.each_with_object([]) do |ids, humaized_associations|
  humaized_associations << ids.each_with_object([]) do |id, humanized_ids|
    humanized_ids << id_name_map[id]
  end.join(', ')
end
```
Because the previous code was using the `<<` operator to add the changes to the list, in effect it was reversing its order. I'm using `.map` which maintains the order.

Regarding the to_sentence part - the same concern would be valid if someone uses a comma as part of the object name, I don't think it's a huge concern but maybe we should store the names in the database as an array, and just join them in the ui? although that might complicate the ui some more since we would need special handling just for this case.

app/models/concerns/taxonomix.rb
ares: From that linked article

> Now ApplicationRecord will be a single point of entry for all the customizations and extensions needed for an application, instead of monkey patching ActiveRecord::Base.

Isn't auditing doing exactly this? Or do you see that as an extension for ActiveRecord::Base itself? That would explain our different views on this. Anyway, I'm fine with this but as I said, debug log in case the method is not available would be good I think.

test/models/user_test.rb
ares: Ok thanks, re `to_sentence` - I think the bigger problem is I18N. I don't think we need to convert it to array, storing string with commas is fine but if we're adding words, it should happen in presentation layer. So we could store it as comma separated string and the do `.split(', ').to_sentence` in the view.

app/models/concerns/taxonomix.rb
kgaikwad: I had two thoughts about it. I would like to share.

1. About adding extra wrapper around `Taxonomix`. It won't fit into our scenario as we have already used `Taxonomix`  module at many places.
2. I know that it is not an appropriate way but would it be fixed by using `remove_method`?
I mean instead of using `Taxonomix` module, we use copy it using ` .dup` & then remove that method.

Again,  `respond_to?` is small one line change to fix this instead of adding extra wrapper or using `.dup` and `remove_method`.

ares: So I think `respond_to?` is a go.

test/models/user_test.rb
tbrisker: good point, i'll leave it as just commas now and when we refactor the ui we'll see how we can display it better.

tbrisker: huh, looks like actually the order in my pr right now is random. i'll add a sort to make sure tests don't fail randomly.

tbrisker:
@ares fixed the tests that were failing due to the hash_with_indifferent_access issue, and reverted to a comma-seperated list for the values for now. Also sorted the list so it is populated in a consistent order.

rplevka:
tested the pr, no issues found and the issue found by ares is gone

ares (approved):

ares:
Thanks @tbrisker, works fine now. Thanks @rplevka for another pair of eyes. Merging!

ares:
@xprazak2 I'm setting this for 1.18 in redmine since this is a fix for functionality added in 1.18 branch. I'll open cherry-pick PR, please let me know if you there's some concern with this cherry-pick.

PR #5456, opened 2018-04-16
Fixes #23276 - Bump up the version of fog-vsphere

leewaa:


timogoebel:
ok to test

timogoebel (approved):

timogoebel:
@tbrisker, @dLobatog: This is the fix we talked about on IRC. It should get into the next 1.17.0 release.

tbrisker:
@timogoebel this needs a bump in packaging/rpm as well. Perhaps also change this to `~> 2.1.1` so we don't get bitten by major updates on debian like we had on 1.16.1?

timogoebel:
> @timogoebel this needs a bump in packaging/rpm as well.

https://github.com/theforeman/foreman-packaging/pull/2408/files should take care of this.

> Perhaps also change this to ~> 2.1.1 so we don't get bitten by major updates on debian like we had on 1.16.1?

I'm unsure about that. Maybe we should go for `~> 2`.

tbrisker:
`~> 2` will match any version, i think you mean `~> 2.1` (which will match any 2.x where x>=1)

leewaa:
So what needs to be done here ? change to `gem 'fog-vsphere', '>= 2.1'` ?

timogoebel:
> ~> 2 will match any version

Well, actually:

```
It’s also important to know that if you specify a major version only, like this:

# gemspec
spec.add_runtime_dependency 'library', '~> 2'

It will only use the latest version from the 2.x series – so 2.3.0 – and not 3.0.0. This behaviour may surprise some people, but automatically allowing any major version past version 2 is more surprising behaviour.
```

([Source](http://guides.rubygems.org/patterns/#declaring-dependencies))


> change to gem 'fog-vsphere', '>= 2.1' ?

I'd suggest the following:
- Let's leave this as is.
- Open a PR against 1.17-stable that changes the dependency to `~> 2.1.1` as @tbrisker proposed.

leewaa:
Ok then this is good to merge, or waiting for some further testing ?

timogoebel:
It's fine for me. @tbrisker: How do you feel about merging this?
The packaging PR has already been merged.

mmoll (approved):
I am OK with this, but we should think about the way we treat dependencies and their changes, where a major bump sometimes means work for us and somtimes not.

timogoebel:
Merged, thanks @leewaa. Can you please also open a PR that cherry-picks (using `cherry-pick -x`) this for `1.17-stable` & `1.18-stable`?

I agree we should think about declaring the deps a little more pessimistic.

xprazak2:
I pulled this into #5463 and #5497

leewaa:
Thanks @xprazak2

PR #5459, opened 2018-04-16
Refs #23215 - 2.3 compat. for webpack in plugins script

dLobatog:
.match? was only introduced on 2.4, therefore it's failing to run
on 2.3 and test pipelines for plugins with webpack & Ruby 2.3 fail

As an example: http://ci.theforeman.org/job/test_plugin_pull_request/4899/database=postgresql,label=fast,ruby=2.3/console


mmoll (approved):

mmoll:
fixes that problem, merged, gracias @dLobatog!

PR #5461, opened 2018-04-17
Fixes #23357 - Refactor Patternfly Notification Drawer

glekner:
- Refactoring the Notification Drawer to use the new one from PF-React 1.18
- Added New and Missing Tests for the Notifications Actions/Reducers
- Improved the Drawer CSS, still changes its height dynamically (not the default PF behaviour) and added Flex properties for window resizing.

**Issues**:
Hiding a specific notification returns an API 404 error (`NOTIFICATIONS_MARK_AS_CLEAR`)


glekner:
- Fixed `NOTIFICATIONS_MARK_AS_CLEAR` API bug.

webpack/assets/javascripts/react_app/components/notifications/index.js
```diff
@@ -33,24 +34,35 @@ class notificationContainer extends React.Component {
...
-            <Drawer
-              onExpandGroup={expandGroup}
-              onClickedLink={onClickedLink}
-              onMarkAsRead={onMarkAsRead}
-              onMarkGroupAsRead={onMarkGroupAsRead}
-              expandedGroup={expandedGroup}
-              notificationGroups={notifications}
-              toggleDrawer={toggleDrawer}
+            <NotificationDrawerWrapper
+            panels={notificationGroups}
```
ohadlevy: nitpick: spacing seems off? consider using prettifier.

webpack/assets/javascripts/react_app/components/notifications/notifications.test.js
```diff
@@ -44,63 +36,6 @@ describe('notifications', () => {
...
-  });
-
-  it('should render a dropdown kebab when links are provided', () => {
-    const store = mockStore(stateWithUnreadNotifications);
-    const wrapper = shallow(<Notifications store={store} />);
-
-    expect(wrapper.render().find('.dropdown-toggle').length).toBe(1);
-  });
-
-  it('full flow', () => {
```
ohadlevy: does it make sense to keep the tests  / at least full flow tests?

webpack/assets/javascripts/react_app/components/notifications/notifications.scss
```diff
@@ -26,75 +26,48 @@
   }
 
   .drawer-pf {
```
ohadlevy: Since you already went though reviewing this file, do you mind adding a comment on why those CSS changes are required? (e.g. I would assume most would expect the CSS to come from patternfly..)

webpack/assets/javascripts/react_app/components/notifications/notifications.test.js
glekner: Re-added the flow test,
shouldn't we trust PF to test the drawer component?

ohadlevy:
[ok to test]

ohadlevy:
@gilad215 in order to fix the failing test, please add:
```diff
diff --git a/app/registries/foreman/access_permissions.rb b/app/registries/foreman/access_permissions.rb
index ce1ac95c4..f322b99c9 100644
--- a/app/registries/foreman/access_permissions.rb
+++ b/app/registries/foreman/access_permissions.rb
@@ -5,7 +5,7 @@ Foreman::AccessControl.map do |permission_set|
   permission_set.security_block :public do |map|
     map.permission :user_logout, { :users => [:logout] }, :public => true
     map.permission :my_account, { :users => [:edit],
-      :notification_recipients => [:index, :update, :destroy, :update_group_as_read] }, :public => true
+      :notification_recipients => [:index, :update, :destroy, :update_group_as_read, :destroy_group ] }, :public => true
     map.permission :api_status, { :"api/v2/home" => [:status]}, :public => true
     map.permission :about_index, { :about => [:index] }, :public => true
```

ohadlevy:
and please rebase as well to avoid unrelated test failure.

ohadlevy:
@gilad215 I'm not sure if its on purpose or not, but when having one item in a category with a dropdown, the following is seen
image


Should the dropdown be visible (instead of a adding a scroll?) /cc @rohoover

glekner:
@ohadlevy Applied your requests.

About the dropdown, we can either create a minimum-height for a panel so it won't override the dropdown, but we don't know how many dropdown entries there are unless we agree on a fixed amount.

I'll come to the office tomorrow, ill try finding a CSS solution.

ohadlevy:
[test upgrade]

webpack/assets/javascripts/react_app/components/notifications/notifications.scss
```diff
@@ -17,85 +17,73 @@
...
+
+    .drawer-pf-action {
+      padding: 10px 0;
+      position: sticky;
+      position: -webkit-sticky;
+      position: -moz-sticky;
+      position: -ms-sticky;
+      position: -o-sticky;
+      bottom: 0;
+      background-color: #fff;
```
ohadlevy: should probably use scss variables? /cc @sharvit ?

ohadlevy (changes requested):
I'm not sure if this has happened before, but every time I click on the dropdown, it initiate a PUT to mark as read, regardless if its already READ or not, i can simulate many many API calls by simply clicking it over and over again.

glekner:
- Fixed unwanted API call from `NOTIFICATIONS_MARK_AS_READ` @ohadlevy

webpack/assets/javascripts/react_app/redux/actions/notifications/index.js
```diff
@@ -59,7 +61,12 @@ export const startNotificationsPolling = url => (dispatch, getState) => {
   dispatch(getNotifications(url));
 };
 
-export const onMarkAsRead = (group, id) => (dispatch) => {
+export const onMarkAsRead = (group, id) => (dispatch, getState) => {
+  if (getState().notifications.notifications
+    .some(notification => notification.id === id) === false ||
+    getState().notifications.notifications
```
ohadlevy: 1. why get state twice? probably can save it to a local var.
2. i'm not sure that reducer is the right place for it? shouldn't we not allow the action to fire to begin with?

glekner: 1. Yes, we can save it to a local var
2. Any click on one of the notification elements triggers a `MarkAsRead`, which seems right because the user interacted with the notification, If we want to change this we need to change it in PF. Thoughts?

ohadlevy:
I'm in mixed feelings as in general the logic if to fire or not the action
could have presentation impact too (e.g. disabled state) and shouldn't be
inside the action logic.
I think pf should accept such a patch as it's really something the
component should know about (e.g don't run that function if it's already
marked as read).

So.. please send a pr to pf :)

glekner:
I agree with what you are saying. But what about consumers that don't care if the notification was read or not and want to trigger a function anyway? I think creating a Read/Unread condition will make the drawer less generic, but maybe it's the right way. @ohadlevy

ohadlevy:
Let's try a fix in pf first? If it doesn't work we can always work around
here...

ohadlevy (changes requested):
I just realized that all language translations are lost, we would need a way to run all the strings via the __('translate') method... maybe the most common way would be a variable with all the string names and id that default to English?

webpack/assets/javascripts/react_app/components/common/Icon/Icon.test.js
```diff
@@ -6,14 +6,29 @@ import Icon from './';
 jest.unmock('./');
 
 describe('Icon', () => {
-  it('displays icon css', () => {
+  it('displays ok icon css', () => {
```
sharvit, line 9: Does it belong to this PR context?
I don't see where you actually use the `Icon` component.
Anyway, I think we should migrate this `Icon` to `pf-react` `Icon`

webpack/assets/javascripts/react_app/components/notifications/notifications.scss
```diff
@@ -1,3 +1,5 @@
+@import '../../common/colors.scss';
+
```
sharvit, line 2: Do we actually need all the css rules in this file?
Shouldn't patternfly supply the css?

webpack/assets/javascripts/react_app/redux/actions/notifications/index.js
```diff
@@ -83,6 +85,29 @@ export const onMarkGroupAsRead = group => (dispatch) => {
   API.put(url);
 };
 
+export const onClear = (group, id) => (dispatch) => {
```
sharvit: Imo, it should name `clear` since it is actually clearing the notification.

webpack/assets/javascripts/react_app/redux/actions/notifications/index.js
```diff
@@ -83,6 +85,29 @@ export const onMarkGroupAsRead = group => (dispatch) => {
...
+    payload: {
+      group,
+      id,
+    },
+  });
+  const url = `/notification_recipients/${id}`;
+  API.delete(url);
+};
+
+export const onMarkGroupAsClear = group => (dispatch) => {
```
sharvit: Same here, i think it should be called `markGroupAsClear`

webpack/assets/javascripts/react_app/redux/reducers/notifications/notifications.fixtures.js
```diff
@@ -9,6 +9,12 @@ export const initialState = Immutable({
   hasUnreadMessages: false,
 });
 
+export const PanelRequest = {
```
sharvit: Why `PanelRequest` and not `panelRequest`?

The rule is, `PascalCase` for `Classes` and `ReactComponents`, otherwise, `camelCase`.

sharvit (changes requested):
Nice job @gilad215 :+1: 
Great to see how much code you could remove.

What do you think about migrating this component to the new folder-structure?

webpack/assets/javascripts/react_app/components/notifications/notifications.scss
glekner: Yes we do, because the drawer in foreman changes dynamically, and I added a change where the dropdown overlaps the Clear/Read section, which still isn't in PF

glekner:
- Fixed unwanted `NOTIFICATIONS_MARK_AS_READ` actions being called
- Added Translations to the drawer
- Refactored some methods names

ohadlevy:
@gilad215 whats the status of the PR? thanks.

package.json
```diff
@@ -73,7 +73,7 @@
     "lodash": "^4.15.0",
     "multiselect": "~0.9.12",
     "patternfly": "^3.42.0",
-    "patternfly-react": "^2.1.0",
+    "patternfly-react": "^2.3.3",
```
ohadlevy, line 76: @ekohl ACK?

glekner:
@ohadlevy PR is ready from my POV. waiting for @sharvit review.

package.json
ekohl: https://github.com/theforeman/foreman-packaging/pull/2510

ekohl (dismissed):

sharvit:
Great job @gilad215 :+1: 

1. I noticed this warning when running tests:
    https://travis-ci.org/theforeman/foreman/jobs/382040869#L5193-L5206

2. When there are no notifications, the `font-size` feel big and takes 2 lines, should it be smaller to fit one line?
    notifications-empty-state-1
    One line example:
    notifications-empty-state-2

3. We are on a process of migrating our components to a domain level file structure. Do you think you can migrate and apply the same conventions here?
   The [BreadcrumbBar](https://github.com/theforeman/foreman/tree/develop/webpack/assets/javascripts/react_app/components/BreadcrumbBar) and the [PasswordStrength](https://github.com/theforeman/foreman/tree/develop/webpack/assets/javascripts/react_app/components/PasswordStrength) currently following this convention.
    you can read more: https://community.theforeman.org/t/howto-add-a-react-component/9651

ohadlevy:
@sharvit 1. there seems to be a non related warning in the logs, mind having a look? 2. I believe this is the pf-style for it, I'm ok with blindly following the pf design (/cc @rohoover) 3. I'm ok to do that in a separate commit?

Rohoover:
@ohadlevy @sharvit 

Where did that styling come from? It may be "correct" but I doubt it was ever designed with intention to use within the Notification Drawer. It may be the generic empty state for a page (or possibly Card). I'm fine to shrink the text if it's easy enough to do.

ohadlevy:
this comes from patternfly-react implementation -
https://rawgit.com/patternfly/patternfly-react/gh-pages/index.html?knob-is%20Drawer%20Expandable=true&selectedKind=patternfly-react%2FNotificationDrawer&selectedStory=Stateful%20Drawer%20Wrapper&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

Rohoover:
@ohadlevy 
Yea that empty state was originally designed for the use on a page. I don't think you guys did anything wrong, I think maybe this is something that needs to be addressed in Core. 
@sharvit concern is valid. It looks huge.

glekner:
1. The mock store had some empty notifications in it, I fixed it @sharvit 
2. I changed the empty state font size to 18: Are we keeping this? @Rohoover 
screen shot 2018-05-24 at 10 59 54


3. @ohadlevy Are we skipping the file structure process? If so, lets merge :)

Rohoover (commented):
@gilad215 
I've brought this to the attention of PF. And an item is being added to the backlog for this issue. 

These are the high fidelity mockups I've contributed to this issue. The proportion of the icon and text remain the same to the original empty state, just scaled down to fit better for that defined space.

patternfly_toastnotification_empty
patternfly_toastnotification_trayempty

18 would be the correct font size.

sharvit (approved):
@ohadlevy @gilad215 Sure, i'm fine with a later commit.

Great job, would love to see it merged.

ohadlevy:
@gilad215  this is how it looks now - the icon is fairly large - is that expected? image

ohadlevy:
looking at the screenshots above this seems to be in line - thanks @gilad215! great job on this one.

ohadlevy (approved):

PR #5462, opened 2018-04-17
Fixes #23290 - Trigger turbolinks before load

sharvit:
Trigger a turbolink event page:before-unload after submitting host form
and before updating dom content manually with jquery.

### DIFF
--- a/app/assets/javascripts/host_edit.js
+++ b/app/assets/javascripts/host_edit.js
@@ -137,6 +137,9 @@ function submit_with_all_params(){
   clear_errors();
   animate_progress();
 
+  // clean the dom from react components before replacing the dom-content
+  tfm.reactMounter.unmountComponents('#content');
+
   $.ajax({
     type:'POST',
     url: url,
--- a/webpack/assets/javascripts/react_app/common/MountingService.js
+++ b/webpack/assets/javascripts/react_app/common/MountingService.js
@@ -7,21 +7,46 @@ import componentRegistry from '../components/componentRegistry';
 
 export { default as registerReducer } from '../redux/reducers/registerReducer';
 
-const mountedNodes = [];
+let mountedNodesCache = [];
 
 // In order to support turbolinks with react, need to
 // unmount all root components before turbolinks do the unload
 // TODO: remove it when migrating into (webpacker-react or react-rails)
-document.addEventListener('page:before-unload', () => {
-  let node = mountedNodes.shift();
+document.addEventListener('page:before-unload', () => unmountComponents());
 
-  while (node) {
-    ReactDOM.unmountComponentAtNode(node);
-    node = mountedNodes.shift();
+const unmountComponentsByNodes = nodes =>
+  nodes.forEach(node => ReactDOM.unmountComponentAtNode(node));
+
+/**
+ * Will unmount all react-components from the dom
+ * @param  {string} selector unmount component inside the selector
+ * @return {number}          amount of unmounted nodes
+ */
+export const unmountComponents = (selector) => {
+  const nodesToUnmount = [];
+  const nodesToKeep = [];
+
+  if (selector) {
+    const reactNode = document.querySelector(selector);
+
+    mountedNodesCache.forEach((node) => {
+      if (reactNode.contains(node)) {
+        nodesToUnmount.push(node);
+      } else {
+        nodesToKeep.push(node);
+      }
+    });
+  } else {
+    nodesToUnmount.push(...mountedNodesCache);
   }
-});
 
-export function mount(component, selector, data) {
+  unmountComponentsByNodes(nodesToUnmount);
+  mountedNodesCache = nodesToKeep;
+
+  return nodesToUnmount.length;
+};
+
+export const mount = (component, selector, data) => {
   const reactNode = document.querySelector(selector);
 
   if (reactNode) {
@@ -31,9 +56,9 @@ export function mount(component, selector, data) {
       reactNode,
     );
 
-    mountedNodes.push(reactNode);
+    mountedNodesCache.push(reactNode);
   } else {
     // eslint-disable-next-line no-console
     console.log(`Cannot find '${selector}' element for mounting the '${component}'`);
   }
-}
+};


app/assets/javascripts/host_edit.js
```diff
@@ -137,6 +137,10 @@ function submit_with_all_params(){
   clear_errors();
   animate_progress();
 
+  // trigger turbolinks event
+  // see webpack/.../MountingService.js
+  document.trigger(new Event('page:before-unload'));
```
tbrisker: I think `new Event...` doesn't work without babel for all browsers, you can use jquery `.trigger`

tbrisker (changes requested):

sharvit:
Thanks for the review @tbrisker, can you have another look?

As the events aren't stable enough, i decided we deserve a core functionality around the unmounting.

ohadlevy:
@tbrisker anything pending for this PR? thanks

webpack/assets/javascripts/react_app/common/MountingService.js
```diff
@@ -7,21 +7,49 @@ import componentRegistry from '../components/componentRegistry';
 
 export { default as registerReducer } from '../redux/reducers/registerReducer';
 
-const mountedNodes = [];
+/**
+ * mounted-nodes cache
+ */
+let mountedNodes = [];
```
tbrisker: Doesn't matter much, but i think this can remain `const` since the constant is the array pointer, not the array itself. Also, instead of the comment, why not just call the variable `mountedNodesCache`?

webpack/assets/javascripts/react_app/common/MountingService.js
```diff
@@ -7,21 +7,49 @@ import componentRegistry from '../components/componentRegistry';
...
 // unmount all root components before turbolinks do the unload
 // TODO: remove it when migrating into (webpacker-react or react-rails)
-document.addEventListener('page:before-unload', () => {
-  let node = mountedNodes.shift();
+document.addEventListener('page:before-unload', () => unmountComponents());
+
+const unmountComponentsByNodes = (...nodes) => {
+  let node = nodes.shift();
 
   while (node) {
```
tbrisker: any reason not to use `Array.prototype.forEach` here instead of manually removing items from the array one at a time?

webpack/assets/javascripts/react_app/common/MountingService.js
```diff
@@ -7,21 +7,49 @@ import componentRegistry from '../components/componentRegistry';
...
+export const unmountComponents = (selector = null) => {
+  let nodesToUnmount = mountedNodes;
+  let nodesToKeep = [];
+
+  if (selector) {
+    const reactNode = document.querySelector(selector);
+    const shouldUnmount = node => reactNode.contains(node);
+
+    nodesToUnmount = mountedNodes.filter(node => shouldUnmount(node));
+    nodesToKeep = mountedNodes.filter(node => !shouldUnmount(node));
```
tbrisker: Would it make sense to use some lodash function (or some other magic) here to calculate one of the arrays as the difference between the original array and the other one, instead of checking all of the nodes against the dom?

webpack/assets/javascripts/react_app/common/MountingService.js
```diff
@@ -7,21 +7,49 @@ import componentRegistry from '../components/componentRegistry';
...
+    node = nodes.shift();
   }
-});
+};
+
+/**
+ * Will unmount all react-components from the dom
+ * @param  {string} selector unmount component inside the selector
+ * @return {number}          amount of unmounted nodes
+ */
```
tbrisker, line 24: Just curious, when did we start adding comments to functions?

tbrisker (changes requested):
nothing major, but see comments inline

webpack/assets/javascripts/react_app/common/MountingService.js
sharvit: I actually do change the pointer, see line 47.
👍  for mountedNodesCache

webpack/assets/javascripts/react_app/common/MountingService.js
sharvit: The origin of it was to be able to unmount the components while removing the nodes from the array (was a const before this pr).

Anyway, I will update it to a normal forEach since it makes more sense now.

webpack/assets/javascripts/react_app/common/MountingService.js
sharvit: I did, but only when it's core related.
I plan to do some experiment later and generate docs from it.

webpack/assets/javascripts/react_app/common/MountingService.js
```diff
@@ -7,21 +7,49 @@ import componentRegistry from '../components/componentRegistry';
...
+export const unmountComponents = (selector = null) => {
```
tbrisker: Not significant, but if the selector is not passed doesn't it just get `undefined` which is also false so no need to set a default value of `null`?

webpack/assets/javascripts/react_app/common/MountingService.js
sharvit: I actually tried to go this way but found my code long and not that readable.

Do you think it's that bad?
I personally prefer to keep it readable then well-performance because:
1. Right now, on average we should have 4-6 items in this array.
2. It's a temporary solution to support turbolinks and we will eventually rid of it

webpack/assets/javascripts/react_app/common/MountingService.js
sharvit: 👍

webpack/assets/javascripts/react_app/common/MountingService.js
tbrisker: My main concern is that if this list gets long it could be slow - each node on the list calls a recursive check on the subtree of the parent node. for example, on the host index page this might be tens of elements (power status per line). If it makes the code significantly more complex then I don't mind leaving as is and only optimizing if we see actual performance impact.

sharvit:
@tbrisker can you have another look please?

ohadlevy:
@tbrisker whats the status of this pr? thanks

ohadlevy:
[test foreman]

webpack/assets/javascripts/react_app/common/MountingService.js
```diff
@@ -7,21 +7,46 @@ import componentRegistry from '../components/componentRegistry';
...
+
+    mountedNodesCache.forEach((node) => {
+      if (reactNode.contains(node)) {
+        nodesToUnmount.push(node);
+      } else {
+        nodesToKeep.push(node);
+      }
+    });
+  } else {
+    nodesToUnmount.push(...mountedNodesCache);
```
tbrisker, line 40: Why not `nodesToUnmount = mountedNodesCache`?

tbrisker (commented):
just one question inline. other than that looks good.

webpack/assets/javascripts/react_app/common/MountingService.js
sharvit: To keep it `const`

sharvit:
@tbrisker can you rereview please? I actually forgot about this pr :/

ohadlevy:
@tbrisker whats the status of this PR? thanks!

tbrisker (approved):

tbrisker:
Works as expected, thanks @sharvit !

PR #5466, opened 2018-04-19
 Fixes #23334 - Order temeplates audits correctly

sean797:
As you can see, `descending` doesn't really work.

```
irb(main):041:0* Audit.descending.where(:auditable_id => 111, :auditable_type => 'ProvisioningTemplate', :action => 'update').second.created_at
=> Fri, 16 Mar 2018 15:46:56 UTC +00:00
irb(main):042:0> Audit.descending.where(:auditable_id => 111, :auditable_type => 'ProvisioningTemplate', :action => 'update').last.created_at
=> Wed, 21 Mar 2018 16:27:10 UTC +00:00
irb(main):043:0> 
irb(main):044:0* Audit.order(created_at: :desc).where(:auditable_id => 111, :auditable_type => 'ProvisioningTemplate', :action => 'update').second.created_at
=> Thu, 05 Apr 2018 10:30:17 UTC +00:00
irb(main):045:0> Audit.order(created_at: :desc).where(:auditable_id => 111, :auditable_type => 'ProvisioningTemplate', :action => 'update').last.created_at
=> Mon, 12 Sep 2016 12:11:02 UTC +00:00
```

Without this my history is like 

screenshot-1

### DIFF
--- a/db/migrate/201805145603_fix_template_audit_versions.rb
+++ b/db/migrate/201805145603_fix_template_audit_versions.rb
@@ -1,0 +1,15 @@
+class FixTemplateAuditVersions < ActiveRecord::Migration[5.1]
+  def up
+    [ ProvisioningTemplate, Ptable].each do |model|
+      model.unscoped.each do |object|
+        object.audits.order(created_at: :asc).first.update(:version => 1) if object.audits.any?
+        object.audits.order(created_at: :asc).each_cons(2) do |prev_audit, curr_audit|
+          curr_audit.update(:version => prev_audit.version + 1)
+        end
+      end
+    end
+  end
+
+  def down
+  end
+end


ares (changes requested):
Looking at audited gem, `descending` is supposed to order the list by `version` - see https://github.com/collectiveidea/audited/blob/master/lib/audited/audit.rb#L49

I wonder why that does not work correctly. Could you provide a list of your audit records you showed in console? And perhaps also do `Audit.descending.to_sql`? If that's a bug in audited, it should be fixed there (cc @tbrisker). Also this seems to be breaking a lot of tests, before you fix them, I'd suggest trying to figure out the cause of the original problem.

sean797:
@ares I suspect something has caused `version` to be reset. It could be something to do with https://github.com/theforeman/foreman/commit/4d5bdcd5d9d67dce6b1f0e8e457bdf71bae95557.

I'll take another look in a couple of weeks, when I should have access to that database again.

sean797:
Okay, I think the inconsistency was created in the time between https://github.com/theforeman/foreman/commit/83775e639460586072d6be23b7937eaf507eebd3 being applied and running the migration in https://github.com/theforeman/foreman/commit/4d5bdcd5d9d67dce6b1f0e8e457bdf71bae95557

```
foreman=# select version,created_at from audits where auditable_id = 111 and auditable_type = 'ProvisioningTemplate' ORDER BY created_at;
<SNIP>
      79 | 2017-03-08 15:33:25.275256
      80 | 2017-07-14 11:09:51.982015
      81 | 2017-07-14 11:11:03.789097
      82 | 2017-07-14 11:11:37.318433
      83 | 2018-03-16 15:46:56.578157
       1 | 2018-03-21 16:27:10.240105
       2 | 2018-03-21 16:27:25.714327
       3 | 2018-03-22 09:16:33.783575
       4 | 2018-03-22 09:18:50.702426
       5 | 2018-03-27 08:35:04.479567
       6 | 2018-03-27 09:05:02.344841
       7 | 2018-03-27 09:21:06.904514
       8 | 2018-03-27 09:36:35.88719
       9 | 2018-03-27 13:40:10.97545
      10 | 2018-03-27 13:56:22.017556
      11 | 2018-03-28 08:20:26.430209
      12 | 2018-03-28 09:16:12.944844
      13 | 2018-03-28 13:46:59.484669
      14 | 2018-03-28 13:57:08.24321
      15 | 2018-04-03 14:44:38.156569
      16 | 2018-04-05 10:30:17.315836
      84 | 2018-04-19 09:06:15.163085
      85 | 2018-04-20 13:29:21.035852

```

I've added a migration that fixes this, thanks!

db/migrate/201805145603_fix_template_audit_versions.rb
```diff
@@ -0,0 +1,15 @@
+class FixTemplateAuditVersions < ActiveRecord::Migration[5.1]
+  def up
+    ['ProvisioningTemplate', 'Ptable'].each do |model|
+      model.constantize.unscoped.each do |object|
```
ares: you can use `[ ProvisioningTemplate, Ptable ].each` to avoid constantize call later.

ares (commented):
sorry to cause problems all the times, I wonder if migration is the right approach (but won't block the PR on this)

sean797:
@ares updated, thanks :+1: 

> I wonder if migration is the right approach

Could you elaborate?

ares:
The migration is run automatically on RPM upgrade. It can be quite slow if you have a lot of audits. Perhaps, it would be better to provide a fix through foreman-maintain, especially since this is not a critical issue, that every user needs to fix. On my small system with only 2470 audits for templates, it took over 10 seconds.

Also why running this migration I noticed it clear all user_id and username information, probably because User.current is nil in this migration. I haven't digged into details, but this would be a serious issue.

sean797:
We have spoken about changing that before
https://community.theforeman.org/t/database-and-service-actions-in-rpm-post-scripts/7311/6

My understanding of the outcome:
- Some users want migrations run in `%postrans`
- Some users DO NOT want migrations ran in `%postrans` (Foreman clusters would be one use-case)
- We should extract it out to another script (maybe foreman-maintain?) and `%postrans` can call the script
- The script can actually run migrations (or not) based on user configuration

Changing application data IMO belongs in a migration. If we have problems with the length of time that migrations are taking I think we should change the way they run, as we (the community) have discussed before.

tbrisker:
Since this is an edge case that hasn't been encountered by multiple users, we aren't sure about the cause at all and if it may occur again, and the migration has the potential of taking a very long time to complete on a large db (regardless of at what stage the migration runs) - I would highly recommend either making it a rake task (we already have some tasks for fixing inconsistent data in the db) or not fixing it at all in core and running a one-off script to fix it for the specific user that encountered the issue.

sean797:
> Since this is an edge case that hasn't been encountered by multiple users

Thats because is caused by the time between time between https://github.com/theforeman/foreman/commit/83775e639460586072d6be23b7937eaf507eebd3 and https://github.com/theforeman/foreman/commit/4d5bdcd5d9d67dce6b1f0e8e457bdf71bae95557. Note that the first commit was shipped in 1.17 and the second will hit in 1.18. Every user who edits a template on 1.17 and then upgrades will be hit by this, I think its fair to say not all of them will care or notice the inconsistent audits.


> we aren't sure about the cause at all

I am, see above.


But I'm okay with either way - I just think we should be clear on exactly what the impact is when we make a decision.

ares:
Thanks @sean797, at this point, we're far from 1.17 and 1.18. We dropped the ball on our side, hopefully there are not too many users facing this in their current instances. Most likely, audits have been cleaned up meanwhile. I don't recall users reporting this but I may have missed some reports.

Anyway, I'm closing this PR as it seems there no need to merge it anymore. Please let us know if there's still some DB affected. We could perhaps extract the code to foreman-maintain or some script for people who'd need to use it.

PR #5468, opened 2018-04-19
Fixes #23331 - grouped permissions for host disabled

lzap:
Authorizable#permission_name generic implementation allows grouping of
permissions in one resource, this is used in couple of them, for example
in SmartProxy or ComputeResource: `edit_comupute_resources and edit_compute_resources_vms`. The way the implementation searches those is using LIKE SQL command:

```ruby
  def permission_name(action)
    type = Permission.resource_name(self.class)
    permissions = Permission.where(:resource_type => type).where(["#{Permission.table_name}.name LIKE ?", "#{action}_%"])

    # some permissions are grouped for same resource, e.g. edit_comupute_resources and edit_compute_resources_vms, in such case we need to detect the right permission
    if permissions.size > 1
      permissions.detect { |p| p.name.end_with?(type.underscore.pluralize) }.try(:name)
    else
      permissions.first.try(:name)
    end
  end
```

This is a problem for discovery host which uses STI and permission name
`edit_discovered_hosts` which matches `edit_hosts` permission and the
LIKE SQL query *sometimes* match - it depends on record order returned
by database server.

This patch explicitly disallows this beahior for STI model "Host".

### DIFF
--- a/app/models/host/managed.rb
+++ b/app/models/host/managed.rb
@@ -833,6 +833,12 @@ class Host::Managed < Host::Base
 
   private
 
+  # Permissions introduced by plugins for this class can cause resource <-> permission
+  # names mapping to fail randomly so as a safety precaution, we specify the name more explicitly.
+  def permission_name(action)
+    "#{action}_hosts"
+  end
+
   def compute_profile_present?
     !(compute_profile_id.nil? || compute_resource_id.nil?)
   end
--- a/test/models/host_test.rb
+++ b/test/models/host_test.rb
@@ -1897,6 +1897,11 @@ class HostTest < ActiveSupport::TestCase
       assert_equal [host], Host::Managed.search_for(query)
     end
 
+    test "host edit permision must be always edit_hosts" do
+      h = FactoryBot.build(:host, :managed)
+      assert_equal "edit_hosts", h.send(:permission_name, :edit)
+    end
+
     test "non-admin user with edit_hosts permission can update interface" do
       @one = users(:one)
       # add permission for user :one


app/models/host/managed.rb
```diff
@@ -833,6 +833,11 @@ def compute_resource_or_model
 
   private
 
+  # Grouped permissions for STI models are explicitly not allowed
```
ares: the comment is a bit misleading, it's not caused by the STI but rather other known permissions introduced by plugins are not mapped correctly

some more context for others, since Host is so exposed object, I think it's reasonable to stop relying on auto detection and simply hardcode resource part of the permission name here. So please reformulate a comment to something like 

`Permissions introduced by plugins for this class can cause resource <-> permission names mapping to fail randomly so as a safety precaution, we specify the name more explicitly`

Then you have my :+1: (I'll wait a bit so others have chance to raise concerns)

ares (dismissed):

lzap:
For the record the failing test:

https://ci.theforeman.org/job/test_plugin_pull_request/4904/database=postgresql,label=fast,ruby=2.4/testReport/junit/(root)/Api__V2__HostsControllerTest/test_0033_should_allow_to_update_for_restricted_user_who_owns_the_host/

app/models/host/managed.rb
lzap: Sure, done.

lzap:
Fixed test calling private method. Other failures seem to be unrelated, other PRs are also failing: unexpected call: statuses().

ares (approved):

ares:
Thanks @lzap, merigng!

lzap:
Thanks trying discovery now.

PR #5470, opened 2018-04-20
Fixes #22983, #22010 - ignore persistence authz for relations

ares:
To reproduce the original issue, as non-admin user, try assigning puppet class to hostgroup or config group or try assigning config group to hostgroup. Neither of this work. With the patch, all works as previously. This is a regression introduced in 1.16 and most likely should be cherry-picked down to 1.16-stable.

### DIFF
--- a/app/models/config_group_class.rb
+++ b/app/models/config_group_class.rb
@@ -9,4 +9,8 @@ class ConfigGroupClass < ApplicationRecord
   validates :puppetclass, :presence => true
   validates :config_group, :presence => true,
     :uniqueness => { :scope => :puppetclass_id }
+
+  def check_permissions_after_save
+    true
+  end
 end
--- a/app/models/host_class.rb
+++ b/app/models/host_class.rb
@@ -11,4 +11,8 @@ class HostClass < ApplicationRecord
   def name
     "#{host} - #{puppetclass}"
   end
+
+  def check_permissions_after_save
+    true
+  end
 end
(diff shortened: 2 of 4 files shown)


tbrisker:
@ares shouldn't these functions actually check permissions on the associated objects?

ares:
I don't think so, there are associations entities. If you have permissions to see e.g. puppet class, there's no additional permission for "assigning class A to config group B". The PR restores the original behavior before 1.16, while you may argue we should have new permission for creating and deleting such associations, I'd say that's an RFE that should not be backported to stable releases and should be addressed separately. The regression was caused by the fact we started to verify permission on creation/update automatically. Which does not make sense for every resource.

tbrisker (commented):
Thanks, that makes sense. Shouldn't this be done on all join table models then? there's a few others besides these 3.

ares:
@tbrisker the only other class like that I found was HostClasses, so added that one. Please take another look, this has been reported multiple times on discourse and redmine already and should be cherry-picked back to 1.16, 1.17 and 1.18.

CC @dLobatog and @xprazak2 for visibility

tbrisker (commented):
There's some other models for join tables I can think of, not sure if its needed for all of them - `UserRole, UsergroupMember, TaxableTaxonomy, SubnetDomain, Filtering, EnvironmentClass, CachedUserRole, CachedUsergroupMember`
Maybe this should be a concern for join tables? that way if this changes we don't have to change all of them

ares:
None of them is including Authorizable module. Not all join tables are affected.

tbrisker (approved):
Ack, makes sense, thanks @ares !

ares:
cherry picks for stable branches, @tbrisker mind to take a look?
https://github.com/theforeman/foreman/pull/5582 - 1.18
https://github.com/theforeman/foreman/pull/5585 - 1.17
https://github.com/theforeman/foreman/pull/5584 - 1.16

PR #5471, opened 2018-04-20
Fixes #23344 - stubs instead expects in setup blocks

lzap:
I know this might bring some controversy, but I believe that *all* test
expectations belong to tests themselves. This is all about style,
nothing should improve or make things faster of course.

### DIFF
--- a/test/integration/about_test.rb
+++ b/test/integration/about_test.rb
@@ -5,10 +5,10 @@ class AboutIntegrationTest < IntegrationTestWithJavascript
   #   AboutIntegrationTest.test_0002_about page proxies should have version
 
   setup do
-    ComputeResource.any_instance.expects(:ping).at_least_once.returns([])
+    ComputeResource.any_instance.stubs(:ping).returns([])
     proxy_status = mock('ProxyStatus::Version')
-    proxy_status.expects(:version).at_least_once.returns('version' => '1.13.0')
-    SmartProxy.any_instance.expects(:statuses).at_least_once.returns(:version => proxy_status)
+    proxy_status.stubs(:version).returns('version' => '1.13.0')
+    SmartProxy.any_instance.stubs(:statuses).returns(:version => proxy_status)
   end
 
   test "about page" do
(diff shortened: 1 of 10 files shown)


test/models/host_test.rb
```diff
@@ -3263,10 +3263,10 @@ def to_managed!
   context "recreating host TFTP config" do
     setup do
       @nic = FactoryBot.build_stubbed(:nic_primary_and_provision)
-      @nic.expects(:rebuild_tftp).returns(true)
-      @nic.expects(:rebuild_dns).never
-      @nic.expects(:rebuild_dhcp).never
-      Nic::Managed.expects(:rebuild_methods_for).returns(:rebuild_tftp => "TFTP")
+      @nic.stubs(:rebuild_tftp).returns(true)
+      @nic.stubs(:rebuild_dns).never
+      @nic.stubs(:rebuild_dhcp).never
```
bastilian: Maybe these two should stay expectations, since they verify that those are not called.

bastilian:
:+1: for changing these to stubs, which only are there to return values, and don't verify behavior. Just the two `.never` expectations should maybe stay.

test/models/host_test.rb
lzap: Changed all nine back to expects, these are pretty small contexts. Yeah.

lzap:
Rebased let's merge @bastilian @tbrisker ?

test/integration/about_test.rb
```diff
@@ -5,10 +5,10 @@ class AboutIntegrationTest < IntegrationTestWithJavascript
   #   AboutIntegrationTest.test_0002_about page proxies should have version
 
   setup do
-    ComputeResource.any_instance.expects(:ping).at_least_once.returns([])
+    ComputeResource.any_instance.stubs(:ping).at_least_once.returns([])
     proxy_status = mock('ProxyStatus::Version')
-    proxy_status.expects(:version).at_least_once.returns('version' => '1.13.0')
-    SmartProxy.any_instance.expects(:statuses).at_least_once.returns(:version => proxy_status)
+    proxy_status.stubs(:version).at_least_once.returns('version' => '1.13.0')
+    SmartProxy.any_instance.stubs(:statuses).at_least_once.returns(:version => proxy_status)
```
bastilian: Something I just noticed now. It might be better to also get rid of the `.at_least_once`. It doesn't seem to hurt, but if it's not necessary.

lzap:
Hey @bastilian any other comments?

tbrisker (approved):
makes sense to me, thanks @lzap !

PR #5472, opened 2018-04-20
Fixes #23346 - memory cache store is used for tests

lzap:
The only difference is that this will be a little faster, I haven't
measured myself but it can be a difference on Jenkins where I/O is
limited.


lzap:
All green!

mmoll (approved):
sounds good to me :+1:

mmoll:
merged, díky @lzap!

PR #5473, opened 2018-04-20
Fixes #23347 - only show password for internal auth source

ares:


tbrisker (approved):
tested, works well. Thanks @ares !

ohadlevy:
too late, but would be nice to move this function to webpack :)

PR #5475, opened 2018-04-22
fixes #23335 - normalize scsi attributes in rails 5

timogoebel:
The Rails 5 upgrade broke vmware scsi controller normalization as the params hash was modified in place and this apparently stopped working on Rails 5.
Without this patch a JSON blob is passed to fog-vsphere.

```
"scsi_controllers"=>"{\"scsiControllers\":[{\"type\":\"VirtualLsiLogicController\",\"key\":1000}],\"volumes\":[{\"thin\":true,\"name\":\"Hard disk\",\"mode\":\"persistent\",\"controllerKey\":1000,\"size\":10485760,\"sizeGb\":10,\"storagePod\":\"Example-Pod\”}]}”,
```

VM creation then fails with:

```
NoMethodError: undefined method `attributes' for #<String:0x007f8bdcfd70b8>
```

This patch normalizes the host attributes directly.

This patch should go into 1.17.1.

### DIFF
--- a/app/controllers/concerns/foreman/controller/parameters/host.rb
+++ b/app/controllers/concerns/foreman/controller/parameters/host.rb
@@ -2,6 +2,7 @@ module Foreman::Controller::Parameters::Host
   extend ActiveSupport::Concern
   include Foreman::Controller::Parameters::HostBase
   include Foreman::Controller::Parameters::HostCommon
+  include Foreman::Controller::NormalizeScsiAttributes
 
   class_methods do
     def host_params_filter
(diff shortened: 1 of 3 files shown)


timogoebel:
@ares: This is the fix you asked me for on IRC.

ares (approved):
Thanks @timogoebel, makes sense and fixes the issue, merging! Would you mind opening CP PRs against 1.18-stable and 1.17-stable? cc @xprazak2, I'll set the 1.17.1 in redmine.

PR #5476, opened 2018-04-23
Fixes #23162 - Fix path for tabs on smart-proxy page

xprazak2:


bastilian:
I tried to reproduce the issue described without the patch applied (in develop), but refreshing after changing a tab on the Smart Proxy status page still keeps the proper page and tab open and does not go to the Dashboard. Is that the correct way to reproduce it?

xprazak2:
That is strange, when I go to proxy page and Puppet tab, I land on dashboard after refresh - when using Chrome.

iNecas:
I was able to reproduce, when enabling Katello: when Katello is not present, the issue doesn't happen: this makes me think that the root cause is actually in the Katello codebase and should be fixed there, otherwise we're risking this behavior will repeat again on other Foreman show page with tabs.

iNecas:
OTOH I can confirm this patch fixed the issue, but again, maybe not the right place to fix this.

xprazak2:
You are right, the issue does not exist when katello is absent. So this must be a result of running katello's js in a foreman page. I can keep looking, but given how the whole page holds together, it could take a while. Besides, I do not think we will encounter this anywhere else since this is the only case where we load katello's js into a foreman page and it is safe to say we will not be doing that again in the future.

iNecas:
[test foreman]

iNecas:
Ok, I'm ok with merging this, as I've tested and proves to fix the issue, and there is no impact on pure-foreman setup. Waiting for tests to finish

xprazak2:
[test foreman]

iNecas:
@xprazak2 can you rebase? You have quite old base branch, that probably leads to test failures.

iNecas (approved):
Test failures unrelated. ACK

iNecas:
Thanks @xprazak2

PR #5477, opened 2018-04-23
Fixes #23364 - add search input in breadcrumbs switcher

amirfefer:
damon-rouselle test - edited


webpack/assets/javascripts/react_app/components/BreadcrumbBar/BreadcrumbBar.js
```diff
@@ -49,12 +51,16 @@ class BreadcrumbBar extends React.Component {
...
+              onSearchChange={event =>
+                loadSwitcherResourcesByResource(resource, { search: event.target.value && `${[resource.nameField]}~${event.target.value}`, search_query: event.target.value })
+              }
               onNextPageClick={() =>
-                loadSwitcherResourcesByResource(resource, { page: Number(currentPage) + 1 })
+                loadSwitcherResourcesByResource(resource, { search: search && `${[resource.nameField]}~${search}`, page: Number(currentPage) + 1 })
               }
               onPrevPageClick={() =>
-                loadSwitcherResourcesByResource(resource, { page: Number(currentPage) - 1 })
+                loadSwitcherResourcesByResource(resource, { search: search && `${[resource.nameField]}~${search}`, page: Number(currentPage) - 1 })
```
ohadlevy: extract to a function?

webpack/assets/javascripts/react_app/components/BreadcrumbBar/BreadcrumbSearch.js
```diff
@@ -0,0 +1,27 @@
...
+    this.state = {
+      value: '',
+    };
+  }
+  render() {
+    const { onSearchChange, searchValue, timeout } = this.props;
+
+    return (
+      <DebounceInput
+        className="autocomplete-input"
```
ohadlevy: does it make sense to have a generic search filter input component?

webpack/assets/javascripts/react_app/components/BreadcrumbBar/components/BreadcrumbSwitcherPopover.js
```diff
@@ -58,6 +60,10 @@ const BreadcrumbSwitcherPopover = ({
 
     popoverBody = (
       <React.Fragment>
+        <BreadcrumbSearch
+         timeout={400}
```
ohadlevy: afair this is not consistent with our current search?

webpack/assets/javascripts/react_app/components/BreadcrumbBar/BreadcrumbSearch.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { DebounceInput } from 'react-debounce-input';
+
+class BreadcrumbSearch extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      value: '',
+    };
```
sharvit: I don't see where you actually use the state.

webpack/assets/javascripts/react_app/components/BreadcrumbBar/BreadcrumbSearch.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
```
sharvit: This file should be under `BreadcrumbBar/components/BreadcrumbSearch.js`

webpack/assets/javascripts/react_app/components/BreadcrumbBar/BreadcrumbBarReducer.js
```diff
@@ -15,6 +15,7 @@ const initialState = Immutable({
   resourceUrl: null,
   requestError: null,
   currentPage: null,
+  search: null,
```
sharvit: do you mind change it to `searchQuery`?
I think it's more clear when your context is only the reducer.

webpack/assets/javascripts/react_app/components/BreadcrumbBar/BreadcrumbSearch.js
```diff
@@ -0,0 +1,27 @@
...
+
+class BreadcrumbSearch extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      value: '',
+    };
+  }
+  render() {
+    const { onSearchChange, searchValue, timeout } = this.props;
```
sharvit: I think that inside this component we can call it `onChange` and `value` since there is nothing here except the search.
Consumer components can use the `search` context `<BreadcrumbSearch value={searchValue} onChange={onSearchChange} />`

sharvit: I would also prefer to change the `timeout` to `debounceTimeout` to make it more clear.

webpack/assets/javascripts/react_app/components/BreadcrumbBar/BreadcrumbSearch.js
```diff
@@ -0,0 +1,27 @@
...
+        value={searchValue}
+        minLength={0}
+        debounceTimeout={timeout}
+        onChange={onSearchChange}
+      />
+    );
+  }
+}
+
+export default BreadcrumbSearch;
```
sharvit: Can you add `propTypes` and `defaultProps`?

webpack/assets/javascripts/react_app/components/BreadcrumbBar/components/BreadcrumbSwitcherPopover.js
```diff
@@ -16,6 +16,8 @@ const BreadcrumbSwitcherPopover = ({
   currentPage,
   totalPages,
   appear, // remove the appear from the ...props
+  onSearchChange,
+  searchValue,
```
sharvit: Can you add them to the `propTypes`?

sharvit (changes requested):
Thanks @amirfefer! Left some comments and waiting for tests :+1:

amirfefer:
@sharvit , @ohadlevy thanks for your review.
could you have another look?

test/integration/breadcrumbs_switcher_test.rb
```diff
@@ -12,4 +12,16 @@ class BreadcrumbsSwitcherTest < IntegrationTestWithJavascript
...
     end
   end
+
+  test "search query" do
+    visit bookmarks_path
+    click_link 'foo'
+    click_button 'switcher'
+    fill_in('breadcrumbs-search', :with => 'three')
+    within(".scrollable-list.list-group") do
+      all = page.all('.no-border')
```
tstrachota: Nitpick, but could you match `.list-group-item`? It will make the test more readable. Renaming `all` to `all_items` would help too.

webpack/assets/javascripts/react_app/components/BreadcrumbBar/BreadcrumbBar.js
```diff
@@ -84,6 +99,7 @@ BreadcrumbBar.propTypes = {
   toggleSwitcher: PropTypes.func,
   closeSwitcher: PropTypes.func,
   loadSwitcherResourcesByResource: PropTypes.func,
+  onSearchChange: PropTypes.func,
```
tstrachota, line 101: Could you either make the prop required or provide a default value?

webpack/assets/javascripts/react_app/components/common/SearchInput/SearchInput.test.js
```diff
@@ -0,0 +1,26 @@
+import toJson from 'enzyme-to-json';
+import { mount, shallow } from 'enzyme';
+import React from 'react';
+import SearchInput from './';
+
+describe('Search Input', () => {
+  it('should render', () => {
+    const wrapper = shallow(<SearchInput searchValue='val' debounceTimeout={300} />);
```
tstrachota: The prop in `SearchInput` is actually `timeout`, not `debounceTimeout`. The value doesn't propagate here, see the snapshot.

webpack/assets/javascripts/react_app/components/BreadcrumbBar/BreadcrumbBarActions.js
```diff
@@ -19,7 +19,7 @@ export const closeSwitcher = () => ({
 
 export const loadSwitcherResourcesByResource = (resource, options = {}) => (dispatch) => {
   const { resourceUrl, nameField, switcherItemUrl } = resource;
-  const { page = 1 } = options;
+  const { page = 1, search } = options;
```
tstrachota: You need to update the code for calculating `pages` too. I should use `subtotal` (number of items matching your query) instead of `total` (number of all items).

webpack/assets/javascripts/react_app/components/BreadcrumbBar/BreadcrumbBar.js
```diff
@@ -49,12 +55,21 @@ class BreadcrumbBar extends React.Component {
               onHide={() => closeSwitcher()}
               onOpen={() => this.handleOpen()}
               onResourceClick={() => closeSwitcher()}
+              onSearchChange={event =>
+                loadSwitcherResourcesByResource(resource, {
+                  search:
+                    event.target.value &&
+                    `${[resource.nameField]}~${event.target.value}`,
```
tstrachota: Could you move this logic to the action? We have all necessary data there. We'll have the logic in one place and passing `search_query` from the next line won't be needed then.

webpack/assets/javascripts/react_app/components/BreadcrumbBar/BreadcrumbBar.js
```diff
@@ -31,9 +31,15 @@ class BreadcrumbBar extends React.Component {
       toggleSwitcher,
       closeSwitcher,
       loadSwitcherResourcesByResource,
+      searchQuery,
     } = this.props;
 
     const isTitle = breadcrumbItems.length === 1;
+    const options = ({ pageIncrement }) => ({
+      search: searchQuery && `${[resource.nameField]}~${searchQuery}`,
```
tstrachota: Duplicate code. I propose to move it to the action. See my other comment.

tstrachota (changes requested):
Thank you @amirfefer. Search works fine in general. I have some comments about the code, plus couple of functional issues that I came across:
- Pagination doesn't reflect the filter. I always get the same number of pages.
- Clicking on next page clears the search field.
- Spinner hides the input field when the results are loading. After it's loaded it gains focus again, so it's not that big problem, but I think that it would be more user friendly to keep the input visible and display the spinner below.

tstrachota:
One more question:
Do all the resources we use the breadcrumbs for support scoped search or is there a chance that we need to disable the search field on some pages?

amirfefer:
@tstrachota Thanks for your review ! 
I've fixed your comments above, plus I added a substring wrapper on the results, in order to make the query search in the results bold.

app/assets/stylesheets/base.scss
```diff
@@ -673,3 +673,11 @@ code.transparent {
   overflow-y: auto;
   max-width: 175px;
 }
+
+.scrollable-list {
```
ohadlevy: please don't add css to the asset pipeline.

app/assets/stylesheets/base.scss
```diff
@@ -673,3 +673,11 @@ code.transparent {
...
+  margin-top: 5px;
+
+  &.list-group {
+    border-top: 0;
```
ohadlevy: isnt list-group used by other things as well? why set this globally?

ohadlevy (changes requested):
I didn't review the code carefully, just used it - overall looks good - some comments:.
* we should add a text no found when search filter does not match any result.
* I would prefer we don't have a scrollbar at all, and if we do, then imho pager should always be visible (e.g. that you scroll the results and dont need to scroll further to find pagination controls). /cc @rohoover 
* nitpick: but it would be nice to have a clear option for search
* does it make sense to cache responses? e.g. when clearing the search bar, we already had the results (as we started with that list, can avoid the extra waiting time).

ohadlevy:
also it seems the spinner missing some top margins as its very close to the search bar
image

Rohoover (commented):
I agree on all point @ohadlevy deviating in that I'm not totally against a scrollbar, but pagination should be present. I'm not sure the original design specified this, so I'm going to poke around a bit.

amirfefer:
@ohadlevy  Thanks, 
I've added a clear button plus the search query becomes bold in the results, so the components looks like:
damon-rouselle test - edited 1

@tstrachota about your question - I think that we shouldn't have `searchable` flag, so if there's search issue on a resource, then i think it would be better to mark it as `unswitchable`until we'll fix it.
other then that it would be inconsistent   to have one switcher with search and another without.

app/assets/stylesheets/base.scss
```diff
@@ -673,3 +673,4 @@ code.transparent {
   overflow-y: auto;
   max-width: 175px;
 }
+
```
ohadlevy: nitpick: no changes to this file please.

webpack/assets/javascripts/react_app/components/BreadcrumbBar/BreadcrumbBarActions.js
```diff
@@ -46,10 +53,10 @@ export const loadSwitcherResourcesByResource = (resource, options = {}) => (disp
     return {
       items: switcherItems,
       page: Number(data.page),
-      pages: Number(data.total) / Number(data.per_page),
+      pages: Number(data.subtotal) / Number(data.per_page),
```
ohadlevy, line 56: does it make sense to show the number of results/pages somewhere in the widget?

webpack/assets/javascripts/react_app/components/BreadcrumbBar/__tests__/__snapshots__/BreadcrumbBarActions.test.js.snap
```diff
@@ -61,7 +67,7 @@ Array [
           },
         ],
         "page": 1,
-        "pages": 1.5,
+        "pages": NaN,
```
ohadlevy: is this a bug?

ohadlevy (changes requested):

webpack/assets/javascripts/react_app/components/common/SubstringWrapper/SubstringWrapper.test.js
```diff
@@ -0,0 +1,36 @@
...
+  <SubstringWrapper Element={Element} substring={substring}>
+    {text}
+  </SubstringWrapper>
+);
+
+describe('Search Input', () => {
+  it('should render with b tag', () => {
+    const wrapper = shallow(setup({ substring: 'test' }));
+
+    expect(toJson(wrapper)).toMatchSnapshot();
```
boaz0: A suggestion:
`const wrapper` seems like a redundancy. maybe doing it this way:

```
expect(toJson(shallow(setup({ substring: 'test' }))));
```

The same goes for the other 3 test cases below.

sharvit: :+1: 

In general, I would love to see the `testHelpers::testComponentSnapshotsWithFixtures` getting used more for this kind of tests
https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/components/BreadcrumbBar/__tests__/BreadcrumbBar.test.js#L21

webpack/assets/javascripts/react_app/components/BreadcrumbBar/BreadcrumbBarActions.js
amirfefer: I think it's a bit redundant, there isn't much space and we do have a search filter

webpack/assets/javascripts/react_app/components/BreadcrumbBar/__tests__/__snapshots__/BreadcrumbBarActions.test.js.snap
amirfefer: yes, I didn't update the fixtures with `subtotal` prop instead of `total`.

webpack/assets/javascripts/react_app/components/common/SubstringWrapper/SubstringWrapper.test.js
amirfefer: well I have mixed feelings about this.
it may be less readable:
```javascript
expect(toJson(shallow(setup({ substring: 'test', text: 'test - this is a test' })))).toMatchSnapshot();
````
therefore I prefer to extract the inner function.

Comment (sharvit) [reply webpack/assets/javascripts/react_app/components/common/SubstringWrapper/SubstringWrapper.test.js]: @amirfefer see how it done here https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/components/BreadcrumbBar/__tests__/BreadcrumbBar.test.js#L21

I think it makes the tests much cleaner and based on configuration rather than code.

Comment (amirfefer) [reply webpack/assets/javascripts/react_app/components/common/SubstringWrapper/SubstringWrapper.test.js]: The `testComponentSnapshotsWithFixtures` looks great, does it support children as well?
`SubstringWrapper` has a child (the wrapped text) - see how the `setup` is written.

Comment (sharvit) [reply webpack/assets/javascripts/react_app/components/common/SubstringWrapper/SubstringWrapper.test.js]: Yes! you can pass `{ children: 'some-text' }` or `{ children: <Component /> }`

File: test/integration/breadcrumbs_switcher_test.rb [hunk 22]
```diff
@@ -12,4 +12,15 @@ class BreadcrumbsSwitcherTest < IntegrationTestWithJavascript
       assert_equal current_bookmark_name, 'three'
     end
   end
+
+  test "search query" do
+    visit bookmarks_path
+    click_link 'foo'
+    click_button 'switcher'
+    fill_in('breadcrumbs-search', :with => 'three')
+    wait_for_ajax
```
amirfefer, line 21: `wait_for_ajax` works with axios as well

webpack/assets/javascripts/react_app/components/BreadcrumbBar/__tests__/__snapshots__/BreadcrumbBarReducer.test.js.snap
```diff
@@ -33,6 +35,20 @@ Object {
   "requestError": null,
   "resourceSwitcherItems": Array [],
   "resourceUrl": "some/url",
+  "searchQuery": undefined,
```
sharvit, line 38: Is that a problem? I think it's the only case `searchQuery` is `undefined` in the store.

test/integration_test_helper.rb
```diff
@@ -98,7 +98,7 @@ def select2(value, attrs)
 
   def wait_for_ajax
     Timeout.timeout(Capybara.default_max_wait_time) do
-      loop until page.evaluate_script('jQuery.active').zero?
+      loop until page.evaluate_script('jQuery.active').zero? && page.evaluate_script('window.axiosActive').zero?
```
ohadlevy: can you please extract this to a pr by it self? I think it would be easier to to merge quickly.

amirfefer: sure

webpack/assets/javascripts/react_app/components/BreadcrumbBar/__tests__/__snapshots__/BreadcrumbBarReducer.test.js.snap
amirfefer: it's not a problem, there is no `searchQuery` in the options hash

just beneath it there's a `should handle BREADCRUMB_BAR_RESOURCES_REQUEST with search query'` test

amirfefer:
@tstrachota could you have another look please?

amirfefer:
rebased.
@ohadlevy

sharvit (approved):
LGTM thanks @amirfefer

webpack/assets/javascripts/react_app/components/BreadcrumbBar/BreadcrumbBar.js
```diff
@@ -31,9 +31,17 @@ class BreadcrumbBar extends React.Component {
       toggleSwitcher,
       closeSwitcher,
       loadSwitcherResourcesByResource,
+      searchQuery,
+      removeSearchQuery,
+      searchDebouneTimeout,
```
tstrachota: Typo: "Debounce"

tstrachota (changes requested):
The updates look good. I like the substring highlighting. It makes the results a lot more readable!

I found two functional issues:
- I tried to search for `*test` and that raised `Uncaught SyntaxError: Invalid regular expression: /(*test)/: Nothing to repeat` which caused the whole breadcrumbs bar to disappear. I guess that any regex character would break it. We should sanitize inputs.
- Searching for hosts using a query that matches more records than `per_page` keeps the pager buttons disabled. It's most likely an API issue and I'm fine with fixing it separately if it is so. Could you please investigate what's wrong there?

amirfefer:
the syntax error occurred in the substring highlighting part -  fixed.

about the host searching issue, it's an API bug with the `thin` param. the subtotal size is the same as the per_page
this issue probably occurs  [here](https://github.com/theforeman/foreman/blob/68388bc261fd72796d9e2af73ed092257d69fb5b/app/controllers/api/v2/hosts_controller.rb#L48),  we're using `pluck` on the paginated resource ,  therefore subtotal may be as the per_page size

tstrachota:
Thanks @amirfefer, nice analysis of the host issue. I'll take another look.

webpack/assets/javascripts/react_app/components/common/SearchInput/SearchInput.test.js
```diff
@@ -0,0 +1,28 @@
...
+  });
+
+  it('should render with href', () => {
+    const spy = jest.spyOn(SearchInput.prototype, 'gainFocus');
+    mount(<SearchInput searchValue='val' timeout={300} />);
+
+    expect(spy).toHaveBeenCalledTimes(0);
+  });
+
+  it('should render with href', () => {
```
tstrachota: Same name for the test?

webpack/assets/javascripts/react_app/components/common/SearchInput/SearchInput.test.js
```diff
@@ -0,0 +1,28 @@
...
+jest.unmock('./');
+
+describe('Search Input', () => {
+  it('should render', () => {
+    const wrapper = shallow(<SearchInput searchValue='val' timeout={300} />);
+
+    expect(toJson(wrapper)).toMatchSnapshot();
+  });
+
+  it('should render with href', () => {
```
tstrachota: Could you use some more descriptive name for the test? Since there's no href in the search field I guess this string is a copy-paste leftover.

webpack/assets/javascripts/react_app/components/common/SubstringWrapper/index.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const SubstringWrapper = ({ children, substring, Element }) => {
+  const regexString = () => {
+    try {
+      return new RegExp(`(${substring})`, 'gi');
+    } catch (e) {
```
tstrachota, line 8: Could you please add a test for this case?

webpack/assets/javascripts/react_app/components/common/SubstringWrapper/index.js
```diff
@@ -0,0 +1,40 @@
...
+    }
+    return wrappedText;
+  };
+
+  return <React.Fragment>{spilttedText()}</React.Fragment>;
+};
+
+SubstringWrapper.propTypes = {
+  children: PropTypes.string,
+  substring: PropTypes.string,
```
tstrachota: Please make both props above required. They have no defaults and the component can't cope with null or undefined values for them.

tstrachota (changes requested):
I confirm the regex issue has been fixed. Could you please add a test for that case?

I found out that the next button doesn't get disabled when you insert a query matching no results. It's because of the condition `currentPage === Math.ceil(totalPages)` (`1 === 0` in this case) in `BreadcrumbSwitcherPopover.js`.

I found couple of other smaller issues in the code too.

I'll file a separate issue for the host api call. When what I mention above is fixed I think we can merge this PR.

tstrachota:
FYI the issue for the hosts API is here:
http://projects.theforeman.org/issues/23599

amirfefer:
@tstrachota Thanks for review, I fixed the issues you mentioned above.

tstrachota (approved):
Works fine now.

tstrachota:
@ohadlevy I believe that your comments have been fixed. Can you confirm this is good to merge from you perspective, please?

@amirfefer we'll have to add `react-debounce-input` into packaging. I'll open an issue for that.

tstrachota:
Ticket for packaging `react-debounce-input`: http://projects.theforeman.org/issues/23613

ohadlevy (approved):

ohadlevy:
@ekohl can you approve as well? thanks!

ekohl (approved):
https://github.com/theforeman/foreman-packaging/pull/2495 submitted

mmoll:
merged, thanks @amirfefer!

PR #5481, opened 2018-04-23
fixes #23335 - normalize scsi attributes in rails 5

timogoebel:
This cherry-picks #5475 into `1.17-stable`.

### DIFF
--- a/app/controllers/concerns/foreman/controller/parameters/host.rb
+++ b/app/controllers/concerns/foreman/controller/parameters/host.rb
@@ -3,6 +3,7 @@ module Foreman::Controller::Parameters::Host
   include Foreman::Controller::Parameters::HostBase
   include Foreman::Controller::Parameters::HostCommon
   include Foreman::Controller::Parameters::KeepParam
+  include Foreman::Controller::NormalizeScsiAttributes
 
   class_methods do
     def host_params_filter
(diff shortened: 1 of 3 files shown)


timogoebel:
This is another proof that we should not use 🐶.

timogoebel:
Test failures and reports on Redmine suggest, that this does not work on 1.17. I'll investigate.

mleklund:
I am willing to work with anyone to test code against a 1.17 install.  Anything you need.

timogoebel:
@mleklund: If you could test the version I just pushed here, that would be a huge help. Please make sure to revert the old patch beforehand (`patch -R` reverses a patch, I believe).

mleklund:
that appears to have fixed the vmware issue.  Looks like I have some param_true to clean up in my preseeds.

```
patch p1 -R < ~/vmware.patch
wget -O ~/vmware.patch2 https://github.com/theforeman/foreman/pull/5481/commits/49cf99a4bacd65e70de430845c6c1960b8c95553.patch
patch -p1 < /tmp/vmware.patch2
```

Then restart foreman, for anyone following along. 😀

mmoll:
@timogoebel could you please rebase this onto a current 1.17-stable? It looks a bit like there are tests failures that should have gotten fixed in the meanwhile.

timogoebel:
Rebased.

mmoll (approved):
Ruby 2.5 test errors irrelevant for 1.17-stable

mmoll:
merged, danke @timogoebel!

PR #5484, opened 2018-04-24
Fixes #23376 - subnet.vlanid size is 4 bytes

lzap:
Resolves "ActiveRecord::ActiveRecordError: No integer type has byte size
10. Use a decimal with scale 0 instead." for MySQL users. It is worth
backporting into 1.17 as this is a blocker issue for all MySQL users I
guess.

It used to be string with 10 chars, we converted it to integer with 10
bytes limit which only PostgreSQL handles correctly.

### DIFF
--- a/db/migrate/20180424080702_change_subnet_vlanid_size.rb
+++ b/db/migrate/20180424080702_change_subnet_vlanid_size.rb
@@ -1,0 +1,12 @@
+class ChangeSubnetVlanidSize < ActiveRecord::Migration[5.1]
+  def change
+    reversible do |dir|
+      dir.up do
+        change_column :subnets, :vlanid, :integer, limit: 4
+      end
+      dir.down do
+        change_column :subnets, :vlanid, :string, limit: 10
+      end
+    end
+  end
+end


tbrisker:
Technically we can store the vlan id in just 2 bytes, but I don't think that matters too much. Only thing i wonder is if this should be done in the migration that changed it originally to int instead of in a new one? At what stage does it fail, during migrations or during seeding?

lzap:
Yeah, I *think* that 64bit numbers on 64bit systems are fastest, smaller are just slower to handle or it is stored as 64bit numbers anyway.

Rails fails to boot completely when parsing schema.rb when running tests, it is possible that in production this falls when you attempt to read from the table - I haven't tried. I think we should have new migration, people are already using this in 1.18, aren't they?

tbrisker:
How come this didn't come up in any tests? Is this issue only present in certain versions of mysql or something?

lzap:
I think the difference is that I migrated my database on postgresql, that process created `schema.rb` with limit of 10, then I switched to MySQL and tried to load the schema and it failed. I guess during migration the MySQL driver somehow ignores/fixes the limit to maximum value or something. I think putting this back to 4 is nothing that should hurt (TM).

tbrisker:
@lzap That's odd, i'm on pg and my schema.rb doesn't list any limit for this field. I don't mind adding this migration too much, but it seems like you may have hit some strange edge case in your development setup that shouldn't affect users.

lzap:
Probably, I was changing from PostgreSQL to MySQL on my dev setup to test something. But as you can see we are not very explicit - it used to be string (10) and then it was converted to integer without any limit. I guess the migration won't hurt (TM) let's just don't backport it then until we see users hitting this:

```
[lzap@box migrate]$ cat 20090718064254_create_subnets.rb
class CreateSubnets < ActiveRecord::Migration[4.2]
  def up
    create_table :subnets do |t|
      t.string   :number,     :limit => 15
      t.string   :mask,       :limit => 15
      t.references :domain
      t.integer  :priority
      t.string   :ranges, :limit => 512
      t.text     :name
      t.string   :vlanid, :limit => 10 # HERE
      t.timestamps null: true
    end
  end

  def down
    drop_table :subnets
  end
end
[lzap@box migrate]$ cat 20171231134017_change_vlan_to_int.rb
class ChangeVlanToInt < ActiveRecord::Migration[5.1]
  def up
    Subnet.unscoped.where(vlanid: "").update_all(vlanid: nil)
    opts = (connection.adapter_name.downcase == 'postgresql') ? {using: 'vlanid::integer'} : {}
    change_column :subnets, :vlanid, :integer, opts # HERE
  end

  def down
    change_column :subnets, :vlanid, :string
  end
end
```

lzap:
Can we merge? Won't hurt, less bytes is always good.

tbrisker (changes requested):
Sure, no harm afaict, but can you please just make it up/down migration with the down removing the limit? change_column isn't reversible iirc

lzap:
Done, I was thinking we don't need it, but won't hurt hopefully.

mmoll:
test failures unrelated :green_heart:

tbrisker (approved):

tbrisker:
Thanks @lzap !

PR #5486, opened 2018-04-24
Refs #23177 - datetime precision fixed for unattended test

lzap:
It looks like PostgreSQL has milisecond precision for datetime columns but MySQL has only seconds. I created a test which relied on `created_at` column but on MySQL this was obviously failing.

Rails 5+ and MySQL 5.6.4+ is now able to handle datetime columns in miliseconds:

https://blog.bigbinary.com/2016/02/23/rails-5-handles-datetime-with-better-precision.html

This needs a migration for our app, not sure if it is worth it but would be nice to have the consistent behavior on both main database platforms. The migration must be executed on MySQL 5.6.4+ version tho, otherwise it won't be effective:

https://gist.github.com/MarkMurphy/93adca601b05acffb8b5601df09f66df

For now, I will only fix the test which was failing.

Discussion at:

https://community.theforeman.org/t/heads-up-mysql-datetime-precision-is-seconds-by-default/8888

### DIFF
--- a/test/controllers/unattended_controller_test.rb
+++ b/test/controllers/unattended_controller_test.rb
@@ -201,8 +201,10 @@ class UnattendedControllerTest < ActionController::TestCase
       ptable_ubuntu = FactoryBot.create(:ptable, :ubuntu, :name => 'ubuntu default',
         :layout => 'd-i partman-auto/disk string /dev/sda\nd-i partman-auto/method string regular...',
                                          :operatingsystem_ids => [operatingsystems(:ubuntu1010).id])
+      # MySQL datetime precision is seconds - explicitly create host1 with tomorrow's created_at
       host1 = FactoryBot.create(:host, :managed, :with_dhcp_orchestration, :build => true,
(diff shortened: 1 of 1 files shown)


mmoll:
I'm OK with this, if it gets the test consistently green, but I'm wondering why something like `Time.now....` works here, shouldn't there be some rounding/padding required, too?

lzap:
Previously I left Rails to set created_at automatically, but since it was too fast, it ended up being in one second time window, so the order was not correct (was random). Now, I set the created_at in the future (one day) so I am sure it is being picked up correctly by the SQL order keyword.

lzap:
All checks are green, it will be good this time, don't worry.

mmoll:
> it was too fast, it ended up being in one second time window, so the order was not correct (was random).

Ah, I see :+1:

mmoll (approved):

tbrisker (approved):
Great job debugging this one! thanks @lzap :medal_sports:

PR #5488, opened 2018-04-24
Refs #23215 - 2.3 compat. for webpack in plugins script

dLobatog:
.match? was only introduced on 2.4, therefore it's failing to run
on 2.3 and test pipelines for plugins with webpack & Ruby 2.3 fail

This was merged to develop but not to 1.18-stable


mmoll:
merged, gracias @dLobatog!

PR #5490, opened 2018-04-24
 Fixes #23386 - Exit fullscreen when pressing ESC 

boaz0:
Pressing the escape key should make the editor exit
the fullscreen mode.  Nonetheless, when this key is pressed
the "exit_fullscreen_editor" function is called but
it was renamed to "exitFullscreen" which ends up with an error
message in the console.

In this patch, I updated the called function to be "exitFullscreen"
so it will be possible to exit fullscreen when escape key is pressed.

### DIFF
--- a/webpack/assets/javascripts/foreman_editor.js
+++ b/webpack/assets/javascripts/foreman_editor.js
@@ -335,7 +335,7 @@ export function enterFullscreen(element, relativeTo) {
   $(document).on('keyup', (e) => {
     if (e.keyCode === 27) {
       // esc
-      exit_fullscreen_editor();
+      exitFullscreen();
     }
   });
   Editor.resize(true);


lzap:
Thanks, I will leave this on UX guys but can you rebase the commit to valid message?

boaz0:
@lzap @tbrisker 
I don't have permissions to move [this ticket](http://projects.theforeman.org/issues/23386) to the Foreman project. Can you do that?
Otherwise, I guess I'll close it and open a new one.

Thanks. :grapes:

lzap:
I don't have the permission either, @iNecas can you move the ticket for us please?

tbrisker:
Moved the ticket.

tbrisker (approved):
Looks like some leftover from a linting change. Thanks @ripcurld !

tbrisker:
@ripcurld can you please open a PR against 1.18-stable as well? I think it's worth it getting this in there.

boaz0:
Thanks a lot @tbrisker and @lzap

PR #5491, opened 2018-04-24

mmoll:
While Ruby 2.5 won't be "supported" with <1.19, there are some plugin tests (namely discovery and tasks) that test plugins against release branches with the same matrix as plugins are tested against develop.

In this special case it's cheaper to "fix" it like this and if this gets approval for 1.18, I'll cherry-pick this down to 1.15-stable.


mmoll:
this would also fix the PR tests to -stable branches

tbrisker:
@mmoll is there any reason not to just bump the dependency for all ruby versions instead of adding the condition?

mmoll:
I wanted to avoid having to update the RPM in all the versions without generating any value.

ekohl:
@mmoll we updated it in nightly and haven't branched 1.18 yet so it'd fine to just set it to 1.3.5 IMHO.

tbrisker (approved):
Ok, makes sense to me and test failure is unrelated. I'll leave it to @xprazak2 to merge this to the stable branch.

ekohl:
Oh, and I also cherry picked it to 1.17. Not sure we test 1.16 and older on Ruby 2.5.

mmoll:
> Not sure we test 1.16 and older on Ruby 2.5.

See the initial PR description. :)

ekohl:
I'd argue we keep it simple on develop, 1.18 and 1.17: just require 1.3.5 since it'll be available in the RPM repos anyway.

mmoll:
opened GH-5499 for 1.16 and updated this PR

ekohl (approved):

mmoll:
GH-5500 with another needed commit for 1.17

xprazak2:
Merging, thanks!

PR #5495, opened 2018-04-25
Fixes #23395 - allow plugins to add fixtures

jlsherrill:

### DIFF
--- a/app/registries/foreman/plugin.rb
+++ b/app/registries/foreman/plugin.rb
@@ -112,7 +112,8 @@ module Foreman #:nodoc:
 
     def_field :name, :description, :url, :author, :author_url, :version, :path
     attr_reader :id, :logging, :provision_methods, :compute_resources, :to_prepare_callbacks,
-                :facets, :rbac_registry, :dashboard_widgets, :info_providers, :smart_proxy_references
+                :facets, :rbac_registry, :dashboard_widgets, :info_providers, :smart_proxy_references,
+                :fixture_directory
 
     # Lists plugin's roles:
     # Foreman::Plugin.find('my_plugin').registered_roles
@@ -313,6 +314,11 @@ module Foreman #:nodoc:
           Plugin::RbacSupport.new.add_all_permissions_to_default_roles(Permission.where(:name => @rbac_registry.permission_names))
         end
       end
+    end
+
+    def fixture_directory(directory = nil)
+      @fixture_directory = directory if directory
+      @fixture_directory
     end
 
     def pending_migrations
--- a/test/fixture_helper.rb
+++ b/test/fixture_helper.rb
@@ -1,0 +1,19 @@
+class FixtureHelper
+  FIXTURE_PATH = "#{Rails.root}/tmp/combined_fixtures/"
+
+  def self.combine(source_directories)
+    FileUtils.rm_rf(FIXTURE_PATH) if File.directory?(FIXTURE_PATH)
+    Dir.mkdir(FIXTURE_PATH)
+
+    source_directories.each do |directory|
+      Find.find(directory).each do |source_path|
+        next unless source_path.ends_with?('.yaml') || source_path.ends_with?('.yml')
+        partial_path = source_path.sub(directory, '')
+        destination_path = File.join(FIXTURE_PATH, partial_path)
+
+        FileUtils.mkdir_p(File.dirname(destination_path))
+        File.open(destination_path, 'a') {|f| f.write(File.read(source_path)) }
+      end
+    end
+  end
+end
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -14,6 +14,7 @@ require 'rfauxfactory'
 require 'webmock/minitest'
 require 'webmock'
 require 'robottelo/reporter/attributes'
+require 'fixture_helper'
 
 # Do not allow network connections and external processes
 WebMock.disable_net_connect!(allow_localhost: true)
@@ -86,8 +87,23 @@ module TestCaseRailsLoggerExtensions
   end
 end
 
+module ExtensibleFixtureSupport
+  extend ActiveSupport::Concern
+
+  included do
+    self.fixture_path = FixtureHelper::FIXTURE_PATH
+    unless @fixtures_ran
+      directories = ["#{Rails.root}/test/fixtures"] + Foreman::Plugin.all.map{ |p| p.fixture_directory }
+      FixtureHelper.combine(directories.compact)
+      @fixtures_ran = true
+      fixtures(:all)
+    end
+  end
+end
+
 class ActiveSupport::TestCase
   prepend TestCaseRailsLoggerExtensions
+  include ExtensibleFixtureSupport
 end
 
 class ActionView::TestCase
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -89,6 +89,7 @@ end
 
 module ExtensibleFixtureSupport
   extend ActiveSupport::Concern
+  require 'unit/plugin/sample_plugin'
 
   included do
     self.fixture_path = FixtureHelper::FIXTURE_PATH
--- a/test/unit/plugin/sample_plugin.rb
+++ b/test/unit/plugin/sample_plugin.rb
@@ -1,0 +1,3 @@
+Foreman::Plugin.register :sample_plugin do
+  fixture_directory "#{Rails.root}/test/unit/plugin/fixtures/"
+end
--- a/test/integration/about_test.rb
+++ b/test/integration/about_test.rb
@@ -6,12 +6,13 @@ class AboutIntegrationTest < IntegrationTestWithJavascript
 
   setup do
     ComputeResource.any_instance.expects(:ping).at_least_once.returns([])
-    proxy_status = mock('ProxyStatus::Version')
-    proxy_status.expects(:version).at_least_once.returns('version' => '1.13.0')
-    SmartProxy.any_instance.expects(:statuses).at_least_once.returns(:version => proxy_status)
+    @proxy_status = mock('ProxyStatus::Version')
+    @proxy_status.expects(:version).at_least_once.returns('version' => '1.13.0')
   end
 
   test "about page" do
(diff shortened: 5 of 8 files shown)


test/fixture_helper.rb
```diff
@@ -0,0 +1,39 @@
...
+    Dir.mkdir(destination)
+
+    load_fixtures.each do |filename, content|
+      path = File.join(destination, filename)
+      directory = File.dirname(path)
+
+      FileUtils.mkdir_p(directory) unless File.exist?(directory)
+      File.open(path, 'w'){|f| f.write(content)}
+    end
+  end
```
lzap: This is an ugly hack that's not needed I believe. The `fixtures` method actually allows you to pass list of yaml files you want to load. Can we avoid completely creating a temporary directory by collecting files from all plugins and providing the list to the method?

test/test_helper.rb
```diff
@@ -86,8 +87,22 @@ def after_teardown
...
+  extend ActiveSupport::Concern
+
+  included do
+    self.fixture_path = "#{Rails.root}/tmp/combined_fixtures/"
+
+    directories = ["#{Rails.root}/test/fixtures"] + Foreman::Plugin.all.map{ |p| p.fixture_directory }
+    helper = FixtureHelper.new(directories.compact)
+    helper.combine(self.fixture_path)
+    fixtures(:all)
+  end
```
lzap: We need to take great attention to this code, fixtures gets loaded for *every test* and we already have slow test suite. We must be sure there are no extra copies of yaml files somewhere. The code looks correct, but I still think this is unncecessary complex.

We have this in `test/active_support_test_case_helper.rb` file:

```ruby
 class ActiveSupport::TestCase
    # Setup all fixtures in test/fixtures/*.(yml|csv) for all tests in alphabetical order.
    # Note: You'll currently still have to declare fixtures explicitly in integration tests
    # -- they do not yet inherit this setting
    fixtures :all
...
```

I think the whole patch is based on what was in Katello, I believe we can do better than this and simply setup fixtures in a correct way without need of copying things over. I don't see any technical limitation of the ActiveSupport stack.

app/registries/foreman/plugin.rb
```diff
@@ -315,6 +316,14 @@ def add_all_permissions_to_default_roles
       end
     end
 
+    def fixture_directory(directory = nil)
+      if directory
+        @fixture_directory = directory
+      else
+        @fixture_directory
+      end
```
lzap: This is clunky, what is the idea here? Is this reader or writer? Reader was already defined above.

Frankly, I think the most flexible API would be `attr_reader fixtures` and `add_fixtures(*files)` methods so plugins could call `add_fixtures(Dir["#{plugin_root}/test/{**,*}/*.{yml}"].uniq)` or something like that.

lzap (changes requested):
Thanks, this is a nice addition. I think instead of copying things, we can simply initialize fixtures with correct YAML file array directly.

Also please include unit test - a plugin that adds its own YAML file, I think it should be technically possible.

timogoebel:
Just wondering: What is the motivation behind this? Isn't `FactoryBot` the "cool thing" _you want to use_?

jlsherrill:
@timogoebel the motivation is that currently most smart_proxy tests fail when katello is present due to a validation that katello adds to smart_proxy that an attribute/column needs to be present, and katello adds a before_validation callback to set a default value for that based on a Setting.  The current Setting fixtures blow away all values present in the Setting table, so there is no default when the foreman tests run.   So the solution allows katello to add a Setting fixture to foreman's when present.

app/registries/foreman/plugin.rb
jlsherrill: I was attempting to mimick the plugin api, which seems to rely on the attribute and the 'setter' being named differently, which is clunky in and of itself :)

test/test_helper.rb
jlsherrill: Its my understanding that ths code only gets run at class load time, but i will dig further to see how we could specify all the files themselves.

app/registries/foreman/plugin.rb
jlsherrill: Having the plugin specify the files seems clunky to me TBH, when they're all going to be in a directory

test/test_helper.rb
lzap: Sure, but as I read it, it only prepares an array of yaml files. In the worst case we should be able to call this twice. But it it turns out it's too difficult, we can go the way you propose, I just think it is worth taking a look.

jlsherrill: yes, when you run 'rake test', this code is only executed once

app/registries/foreman/plugin.rb
lzap: Indeed an assumption we can make, or not. I am slightly towards exposing it on file-level, but I can live with directory. I'd perhaps do one line if in this case.

Anyway, our plugin API is kind of DSL, so this is the reason why we don't use attributes I think.

lzap:
@timogoebel fixtures can be still useful, for example for objects you need almost for all tests, like user. FactoryBot is complementary to fixtures rather than replacement.

test/fixture_helper.rb
jlsherrill: I don't believe this is the case, looking at the code:  https://github.com/rails/rails/blob/master/activerecord/lib/active_record/fixtures.rb#L893
it allows you to pass in the 'names' of the fixtures (i.e. operatingsystem, user), not the files.

It seems very much tied to a single directory:  https://github.com/rails/rails/blob/master/activerecord/lib/active_record/fixtures.rb#L1032

You can see the code that actually loads it here:  https://github.com/rails/rails/blob/375a4143cf5caeb6159b338be824903edfd62836/activerecord/lib/active_record/fixtures.rb#L531

I'm not SUPER happy with this implementation and it does seem a bit hacky, there just doesn't seem anything better built in.  An alternative would be monkey patching the fixture loading, which seems worse...

jlsherrill:
@lzap updated based off your suggestions and added a test that tests that a fixture is properly loaded via the plugin api.  I don't particularly like the test (mostly the fact that the plugin has to be loaded BEFORE fixtures are loaded).  I attempted to reload the fixtures after adding the plugin within a test and that did not work.  I've added as a separate commit so you can see what i mean.

jlsherrill:
[test foreman]

jlsherrill:
[test foreman]

test/fixture_helper.rb
```diff
@@ -0,0 +1,39 @@
+class FixtureHelper
+  def initialize(source_directories)
+    @source_directories = source_directories
+  end
+
+  def combine(destination)
+    FileUtils.rm_rf(destination) if File.directory?(destination)
```
lzap: When I see `rm_rf` I want to be absolutely sure it is not possible to mis-call this with a wrong directory like `$HOME` or `pwd` or `/`. Can we move the `"#{Rails.root}/tmp/combined_fixtures/"` definition here so it's crystal clear when you look at it and it's actually not possible to call this with a wrong path?

test/fixture_helper.rb
```diff
@@ -0,0 +1,39 @@
...
+        if fixtures[filename]
+          fixtures[filename] += data
+        else
+          fixtures[filename] = data
+        end
+      end
+    end
+
+    fixtures
+  end
```
lzap: This method is only called once, to actually copy the directory tree via Ruby allocated memory. We could consider using `IO.copy_stream` with `open('ab')` but that's perhaps an overkill.

test/test_helper.rb
lzap: What if Rails 6 loads fixtures on every test or something, we won't notice and tests will be copying tree hierarchy for every single test. Can't think how to create a test for that honestly, maybe some safety check in the loading code. Any ideas? I just want to have the safest possible hack :-)

test/unit/plugin/sample_plugin.rb
```diff
@@ -0,0 +1,3 @@
+Foreman::Plugin.register :sample_plugin do
+  fixture_directory "#{Rails.root}/test/unit/plugin/fixtures/"
+end
```
lzap, line 3: I wonder how this gets loaded, this is not ending with `_test.rb`. Can you move this to the `test_plugin_fixtures` method?

lzap (changes requested):
Hmmm not the best piece of code to merge, couple of questions and comments.

test/test_helper.rb
jlsherrill: I guess we could use a class variable to store whether they had been loaded or not and not run twice.

test/fixture_helper.rb
jlsherrill: :+1:

test/unit/plugin/sample_plugin.rb
jlsherrill: Its loaded because of an explicit require here: https://github.com/theforeman/foreman/pull/5495/files#diff-f14ca9959ff06f3ca2ee6b9564866954R92

It has to be loaded prior to the fixture loading code gets run, and relying on it to be loaded because of *_test.rb loading did not work, it was too late.  Since this file isn't actually a test (its a plugin definition), i don't think it should end in _test.  If there's a better place to put this, i'm happy to put it somewhere else.

test/fixture_helper.rb
jlsherrill: I didn't see a way to use copy_stream and append to files, so i did refactor it to do simple file appends.

jlsherrill:
[test foreman]

jlsherrill:
[test foreman]

jlsherrill:
[test foreman]

jlsherrill:
@lzap re-factored this quite a bit, and tests are now all green.

jturel:
Haven't tested this but conceptually it looks valid :+1:

jlsherrill:
I actually thought of a much simpler way to solve my problem.  Its not super great, but maybe better than this :)

lzap:
Justin I was *that* late on gh notification, you should have pinged me sorry! Out of curiosity, where's your new patch? :-)

PR #5496, opened 2018-04-25
Fixes #23398 - Don't extend ApplicationRecord from Audit initializer

leewaa:
This fixes an error thrown as a result incorrect extending of ApplicationRecord from the Audit initializer.

As per the description in the commit message:

Any changes to the code locally will kick in rails's auto loading which in turn
will then reload all relevant classes, and in the case of changing a model
inheriting from ApplicationRecord, then ApplicationRecord be reloaded. So since
ApplicationRecord was being extended from within the initializer for audit then
it will load the extend however the module with still be hanging around, causing
rails throw an argument error: 'A copy of
AuditAssociations::AssociationsDefinitions has been removed from the module tree
but is still active!'.

Also note, audit_associations needs to be required because at the time of app
start up during initialization the audit_associations has not yet been loaded.
See the config/application.rb for the loading order and one can see that models
are loaded long before the audit_associations.

### DIFF
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -1,5 +1,9 @@
+require_relative 'concerns/audit_associations'
+
 class ApplicationRecord < ActiveRecord::Base
   self.abstract_class = true
+
+  extend AuditAssociations::AssociationsDefinitions
 
   # Rails use Notifications for own sql logging so we can override sql logger for orchestration
   def logger
--- a/config/initializers/audit.rb
+++ b/config/initializers/audit.rb
@@ -5,8 +5,6 @@ require 'audited'
 # Re-opened AuditorInstanceMethods to audit 1-0-* associations
 Audited::Auditor::AuditedInstanceMethods.send(:prepend, AuditAssociations::AssociationsChanges)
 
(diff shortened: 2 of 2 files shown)


leewaa:
I will fix up the commit message, wasn't aware of the line length limit of 72 chars

timogoebel:
ok to test

timogoebel (approved):
Makes sense. Thanks.

tbrisker (approved):
@leewaa Thanks! code makes sense to me, TBH we need to revisit all of the activerecord extentions we have in place and move them to applicationrecord as well, and perhaps then we might be able to fix the loading order in initialization. 
I invited you to the Foreman organization so your future PRs will automatically trigger CI tests.

orrabin (approved):
Works for me :) 
This can be merged once tests are green.

leewaa:
@tbrisker Great, thanks!

Yeah there are several extensions and includes through out the the initialization. One thing at a time :D    For now I am glad that this one is fixed :)

ares:
@tbrisker thanks for merging, I'm setting the release in redmine to 1.19.0, please don't forget next time

leewaa:
Forgot to mention, some credit goes to @basiszwo as well for this one. Never to late I guess 😸

PR #5485, opened 2018-04-24
Fixes #23358 - Fix audit mails for template content

xprazak2:

PR #5497, opened 2018-04-26
Cherry picks for 1.17-stable

xprazak2:
Cherry pick of #5485 into 1.17-stable


tbrisker (approved):
Failing tests are same as GH-5491, that should be CPed back to 1.17 stable but that shouldn't block merging these.

tbrisker:
Thanks @xprazak2 !

PR #5499, opened 2018-04-26

mmoll:
While Ruby 2.5 won't be "supported" with <1.19, there are some plugin tests (namely discovery and tasks) that test plugins against release branches with the same matrix as plugins are tested against develop.

In this special case it's cheaper to "fix" it like this and if this gets approval for 1.16, I'll cherry-pick this also to 1.15-stable.


ekohl (approved):

mmoll:
ok, this breaks differently because of the ruby2ruby pin, I'll wipe the repsective plugin tests as this is getting too expensive.

PR #5500, opened 2018-04-26

mmoll:


Gemfile
```diff
@@ -50,7 +50,6 @@ gem 'deacon', '~> 1.0'
 gem 'webpack-rails', '~> 0.9.8'
 gem 'mail', '~> 2.7'
 gem 'sshkey', '~> 1.9'
-gem 'ruby2ruby', '2.3.2'
```
ekohl, line 53: I wonder why this was pinned so specific. Do we at least need to keep it in as `>=`?

mmoll: There was a certain combination of safemode and ruby2ruby that wasn't compatible. As ruby2ruby is not a direct foreman dependency, that line got removed again when safemode 1.3.4 with extended ruby2ruby compatibility came out.

ekohl (approved):

PR #5478, opened 2018-04-23
Fixes #23361 - Write the webpack manifest to a tmp file

ekohl:
In a production setup Foreman can't write to the manifest.json location so it needs a temporary location. Here we copy the pattern from the above sprockets initialization and change the webpack-rails setting to read it.

PR #5501, opened 2018-04-27
Fixes #23414 - Allow plugins to compile in production

ehelms:

### DIFF
--- a/lib/tasks/plugin_assets.rake
+++ b/lib/tasks/plugin_assets.rake
@@ -128,7 +128,7 @@ task 'plugin:assets:precompile', [:plugin] => [:environment] do |t, args|
       def compile
         return unless File.exist?("#{@plugin.path}/package.json")
         ENV["NODE_ENV"] ||= 'production'
-        webpack_bin = ::Rails.root.join(::Rails.configuration.webpack.binary)
+        webpack_bin = ::Rails.root.join('node_modules/webpack/bin/webpack.js')
         config_file = ::Rails.root.join(::Rails.configuration.webpack.config_file)
         sh "#{webpack_bin} --config #{config_file} --bail --env.pluginName=#{@plugin.id}"
       end
--- a/script/plugin_webpack_directories.rb
+++ b/script/plugin_webpack_directories.rb
@@ -16,27 +16,15 @@ plugin_name_regexp = /foreman*|katello*/
 specs.each do |dep|
   # skip other rails engines that are not plugins
   # TODO: Consider using the plugin registration api?
-  if gemfile_in
-    next unless dep.name =~ plugin_name_regexp
-    dep = dep.to_spec
-  else
-    next unless dep.name =~ plugin_name_regexp
-  end
+  next unless dep.name =~ plugin_name_regexp
+  dep = dep.to_spec if gemfile_in
 
   path = "#{dep.to_spec.full_gem_path}/webpack"
   # some plugins share the same base directory (tasks-core and tasks, REX etc)
   # skip the plugin if its path is already included
   next if config[:paths].include?(path)
-  if Dir.exist?(path) && !dep.files.any? { |f| /webpack/ =~ f }
-    STDERR.puts "WARNING: webpack was found in #{dep.name}'s source, "\
-      "but it's not in the gemspec. Please add it to the gemspec."
-  end
-  if File.exist?("#{dep.to_spec.full_gem_path}/package.json") &&
-      !dep.files.include?('package.json')
-    STDERR.puts "WARNING: package.json was found in #{dep.name}'s source, "\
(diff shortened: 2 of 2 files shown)


ehelms:
@ekohl @dLobatog These changes allowed me to get a production install working.

script/plugin_webpack_directories.rb
```diff
@@ -17,7 +17,7 @@
   # skip other rails engines that are not plugins
   # TODO: Consider using the plugin registration api?
   if gemfile_in
-    next unless dep =~ plugin_name_regexp
+    next unless dep.to_s =~ plugin_name_regexp
```
ekohl: In https://github.com/theforeman/foreman/pull/5478 I used `dep.name` because it's the same as line 23

lib/tasks/plugin_assets.rake
```diff
@@ -128,7 +128,11 @@ task 'plugin:assets:precompile', [:plugin] => [:environment] do |t, args|
       def compile
         return unless File.exist?("#{@plugin.path}/package.json")
         ENV["NODE_ENV"] ||= 'production'
-        webpack_bin = ::Rails.root.join(::Rails.configuration.webpack.binary)
+        if system('which webpack')
```
ekohl: I wonder if this is the best place. Should we check if `::Rails.configuration.webpack.binary` exists and otherwise fall back to `webpack`? Also wonder if it should be done in https://github.com/theforeman/foreman/tree/develop/config/environments so we always use the correct webpack binary rather than just in this rake task.

script/plugin_webpack_directories.rb
ehelms: I think I tested this in the buildroot and this did not always work so I punted to a safer `to_s`

ehelms:
Updated to set the webpack binary more to a more robust location.

ekohl (dismissed):

dLobatog (commented):
@ehelms Looks good, let me just test compiling a plugin before merging, thanks !

dLobatog (dismissed):
@ehelms So far:

* `.node_modules/bin/webpack` (command not found)
* `node_modules/webpack/bin/webpack.js` - plugin assets cannot be found https://gist.github.com/155bc6eb93219a809ba3c34e4621ae2e
* `/usr/bin/webpack/` https://gist.github.com/66a41967ba819f5cc82bfb5bb5670b7a same 

I've used https://github.com/dlobatog/foreman/tree/webpack-assets and https://github.com/dlobatog/foreman-packaging/tree/rpm/webpack-rex to build REX to test this.

ekohl:
> .node_modules/bin/webpack (command not found)

I'd expect that to fail. `node_modules/.bin/webpack` should work on local installs but not on system wide installs.

dLobatog:
@ekohl Definitely, I was just trying to rule all possibilities out :smile:

ehelms:
Tested an RPM build with this change and everything worked.

ehelms:
[test foreman]

ehelms:
[test foreman]

lib/tasks/plugin_assets.rake
```diff
@@ -128,7 +128,7 @@ task 'plugin:assets:precompile', [:plugin] => [:environment] do |t, args|
       def compile
         return unless File.exist?("#{@plugin.path}/package.json")
         ENV["NODE_ENV"] ||= 'production'
-        webpack_bin = ::Rails.root.join(::Rails.configuration.webpack.binary)
+        webpack_bin = ::Rails.root.join('node_modules/webpack/bin/webpack.js')
```
mmoll, line 131: Shouldn't this be `node_modules/.bin/webpack`?

ehelms: That only exists in development. Whereas this exists in all environments.

ehelms:
@ekohl @dLobatog greeeen!

lib/tasks/plugin_assets.rake
mmoll: I see, thanks!

mmoll:
@ehelms please rebase :)

script/plugin_webpack_directories.rb
```diff
@@ -16,27 +16,15 @@
...
   # skip other rails engines that are not plugins
   # TODO: Consider using the plugin registration api?
-  if gemfile_in
-    next unless dep.name =~ plugin_name_regexp
-    dep = dep.to_spec
-  else
-    next unless dep.name =~ plugin_name_regexp
-  end
+  next unless dep.name =~ plugin_name_regexp
+  dep = dep.to_spec if gemfile_in
```
ehelms, line 20: @ekohl I simplified the logic here

ekohl (approved):

mmoll:
merged, thanks @ehelms!

PR #5502, opened 2018-04-27
Fixes #23415 - nullify taxonomy associations for audits

kgaikwad:
On taxonomy destroy, nullify taxonomy associations for audit
records that are created by 'destroy' action on dependent
resources.

### DIFF
--- a/app/models/taxonomy.rb
+++ b/app/models/taxonomy.rb
@@ -218,6 +218,14 @@ class Taxonomy < ApplicationRecord
     self.subtree.flat_map(&:users).map(&:id).uniq
   end
 
+  # note - this method used by before_destroy callbacks in extension files from plugins
+  # audits for 'destroy' action on resources lead to taxable_taxonomies records.
+  # This will check if any taxable_taxonomies records present and apply destroy_all
+  # so that it nullifies all associated audit records
+  def destroy_taxable_taxonomies
+    TaxableTaxonomy.where(taxonomy_id: self.id).destroy_all
+  end
+
   private
 
   delegate :need_to_be_selected_ids, :selected_ids, :used_and_selected_ids, :mismatches, :missing_ids, :check_for_orphans,


kgaikwad:
@ares,
To unblock the [PR-7248](https://github.com/Katello/katello/pull/7248), added method `destroy_taxable_taxonomies_if_present` in taxonomy.rb.
I will add `before_destroy :destroy_taxable_taxonomies_if_present` callback to file [app/models/katello/concerns/organization_extensions.rb](https://github.com/Katello/katello/blob/master/app/models/katello/concerns/organization_extensions.rb#L46). So that it will execute after all `:dependent => :destroy` callbacks.

app/models/taxonomy.rb
```diff
@@ -218,6 +218,17 @@ def notification_recipients_ids
     self.subtree.flat_map(&:users).map(&:id).uniq
   end
 
+  # note - this method used by before_destroy callbacks in extension files
+  # audits for 'destroy' action on resources lead to taxable_taxonomies records.
+  # This will check if any taxable_taxonomies records present and apply destroy_all
+  # so that it nullifies all associated audit records
+  def destroy_taxable_taxonomies_if_present
+    taxable_taxonomy_records = TaxableTaxonomy.where(taxonomy_id: self.id)
```
ares: I think the whole method could be simplified to
```
TaxableTaxonomy.where(taxonomy_id: self.id).destroy_all
```

if there are no records, it's safe to call destroy on that. If you agree, then I'd suggest to rename the method to `nullify_taxable_taxonomies` or `destroy_taxable_taxonomies` or something similar.

app/models/taxonomy.rb
```diff
@@ -218,6 +218,17 @@ def notification_recipients_ids
     self.subtree.flat_map(&:users).map(&:id).uniq
   end
 
+  # note - this method used by before_destroy callbacks in extension files
```
ares: extensions files in plugins I guess, from this comment it's not clear that it should not be removed as a dead code

ares (changes requested):
thanks @kgaikwad, few comments

app/models/taxonomy.rb
kgaikwad: Thank you @ares for review. 
Modified the code as per suggestions and updated the PR.

ares (approved):

ares:
Thanks @kgaikwad, merging!

PR #5508, opened 2018-04-29
Fixes #23443 - relax pinning of rails-i18n

mmoll:


mmoll:
I can't find a reason to pin this so restrictivly. This allows to use rails-i18n 5.1.x which in turn will in the future possibly unblock an i18n update.

tbrisker (approved):
Makes sense. thanks @mmoll !

tbrisker:
@theforeman/packaging - please ack/nack.

ekohl (approved):

PR #5510, opened 2018-04-30
fixes #23387 - FOREMAN_RSS_LATEST_POSTS option now works

ohadlevy:
the rake task rss:create_notifications failed to
parse FOREMAN_RSS_LATEST_POSTS value correctly.


ekohl (approved):
I wonder if there should be error handling when it's not an int, but for rake tasks I guess it's not that important.

dLobatog (approved):
Thanks @ohadlevy !

PR #5518, opened 2018-05-01
Fixes #23474 - update capybara to 3.x

mmoll:


mmoll:
depends on GH-5517 (and includes it here for the tests to run)

test/integration_test_helper.rb
```diff
@@ -15,6 +15,8 @@
   Rails.logger.error("DO NOT IGNORE - Consistent failure - #{klass} #{test_name}")
 end
 
+Capybara.server = :webrick
```
mmoll, line 18: This was the default for 2.x. Puma is the default for 3.x but I'd rather want to discuss this and handle it in a separate issue.

ehelms (approved):

PR #5522, opened 2018-05-03
Fixes #23417 - telemetry for audit record creation

lzap:
It's worth monitoring how many audits are being created. We expect
increase in the upcoming version.

Creates:

```
Incrementing counter fm_rails_audit_records_created by 1 {:type=>"Environment"}
```


ares:
Should we aim this for 1.18 cherry-pick? That's the version that added many audit records.

lzap:
I actually incorporated this into the other patch, I hope that one gets cherry picked because audits are currently logged in an unsearchable way.

ares:
@lzap so should this be closed? could you provide the link to the second patch?

lzap:
It's in https://github.com/theforeman/foreman/pull/5570

PR #5525, opened 2018-05-03
Fixes #23487 - support namespaced controllers

ares:
```
controller_name #=> foreman_virt_who_configure_configs
controller_path #=> foreman_virt_who_configure/configs
```

also it's better to use controller `resource_name` and `resource_class` that are used elsewhere for mapping resource name

ping @sharvit @amirfefer, I marked this as 1.18 cherry-pick since this was introduced in this version, otherwise we'd need to disable resource switcher in namespaced plugins

### DIFF
--- a/app/services/breadcrumbs_options.rb
+++ b/app/services/breadcrumbs_options.rb
@@ -2,9 +2,9 @@ class BreadcrumbsOptions
   # TODO: extract URL handing from menu/item and use it here too
   include Rails.application.routes.url_helpers
 
-  def initialize(page_header, controller_name, action_name, options = {})
+  def initialize(page_header, controller, action_name, options = {})
     @page_header = page_header
-    @controller_name = controller_name
+    @controller = controller
     @action_name = action_name
     @options = options
   end
@@ -20,11 +20,18 @@ class BreadcrumbsOptions
   private
 
   delegate :logger, :to => :Rails
-  attr_reader :controller_name, :action_name, :page_header, :options
+  delegate :resource_name, :resource_class, :controller_path, :resource_path, :to => :controller
+  attr_reader :controller, :action_name, :page_header, :options
   def index_item
+    begin
+      class_name = resource_class.to_s
+    rescue
+      class_name = ''
+    end
+
     {
-      caption: _(controller_name.humanize),
-      url: try("#{controller_name}_path")
+      caption: _(resource_name.pluralize.humanize),
+      url: resource_path(class_name)
     }
   end
 
@@ -48,14 +55,20 @@ class BreadcrumbsOptions
 
   def switcher_url_template
     actual_action_name = (action_name == 'show') ? '' : action_name
-    "/#{controller_name}/:id/#{actual_action_name}"
+    if respond_to?(resource_name + '_path')
+      resource_path = try(resource_name + '_path', :id => ':id')
+    else
+      resource_path = "/#{controller_path}/:id"
+    end
+
+    "#{resource_path}/#{actual_action_name}"
   end
 
   def model_name_field
-    controller_name.camelize.singularize.try(:constantize).try(:title_name)
+    resource_class.try(:title_name)
   rescue => err
     # TODO: better handling for plugin detection
-    logger.warn("unable to detect breadcrumb title name in for #{controller_name}, defaulting to name")
+    logger.warn("unable to detect breadcrumb title name in for #{controller_path}, defaulting to name")
     logger.debug(err)
     "name"
   end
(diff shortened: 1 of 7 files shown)


app/services/breadcrumbs_options.rb
```diff
@@ -20,11 +20,11 @@ def bar_props
   private
 
   delegate :logger, :to => :Rails
-  attr_reader :controller_name, :action_name, :page_header, :options
+  attr_reader :controller, :action_name, :page_header, :options
   def index_item
     {
-      caption: _(controller_name.humanize),
-      url: try("#{controller_name}_path")
+      caption: _(controller.resource_name.pluralize.humanize),
```
ohadlevy: nitpick: maybe ```delegate resource_name to controller```?

ares: ok, added delegate to all other method we use from `controller`

ohadlevy (dismissed):
pending tests :)

timogoebel:
Tests need some more love :-)

amirfefer (changes requested):
Thanks @ares ! looks great.
In the test file [breadcrumbs_options_test](https://github.com/theforeman/foreman/blob/develop/test/unit/breadcrumbs_options_test.rb#L6) we should also use a controller object rather then `controller_name`.

maybe we can use something like `dummyController` from [here](https://github.com/theforeman/foreman/blob/develop/test/unit/concerns/action_permission_dsl_test.rb#L10)

ares:
I'll fix tests during following travels :-)

ares:
Tests fixed, I had to move the `resource_name` in dashboard controller from private to public. The one that it overrides (FindCommon concern) is defined as public.

Please take another look.

amirfefer (approved):
Thanks: , LGFM  :+1:

app/services/breadcrumbs_options.rb
```diff
@@ -20,11 +20,12 @@ def bar_props
...
   delegate :logger, :to => :Rails
-  attr_reader :controller_name, :action_name, :page_header, :options
+  delegate :resource_name, :resource_class, :controller_path, :to => :controller
+  attr_reader :controller, :action_name, :page_header, :options
   def index_item
     {
-      caption: _(controller_name.humanize),
-      url: try("#{controller_name}_path")
+      caption: _(resource_name.pluralize.humanize),
+      url: try("#{resource_name.pluralize}_path")
```
timogoebel: Does it make sense to differentiate `resource_path` (a new method for the controllers) and `resource_name`?
If the path is namespaced, we need to use `foreman_whatever_models_path` here. But `find_resource` will then set an instance variable `@foreman_whatever_model`. But I guess devs want to just use the instance variable `@model` (without the namespace).

timogoebel (changes requested):
Thanks, @ares. Looks good in general. I just have one question regarding `resource_name` and path generation.

app/services/breadcrumbs_options.rb
ares: Probably makes sense, is `resource_path` always available?

timogoebel: Actually I just invented that name, but we have some interesting code here: https://github.com/theforeman/foreman/blob/4e6d22620ed3bbd9e1caa5652f2e0df5ebe21d61/app/helpers/filters_helper.rb#L23-L36

ares: great, I took that method and put it into ApplicationController as a helper_method and used at `index_item` but not here, because it wouldn't work with a plugin that does not have namespaced controller class but has namespaced API, such as foreman_openscap

I now added a test for whetner `resource_name + '_path'` is defined and use it in such case (e.g. `policy_path :id => ':id' # => /compliance/policy/:id`) and if not, we use `"/#{controller_path}/:id"` which works for plugin such as virt_who_configure and non-namespaced plugins.

mind to take another look @timogoebel

timogoebel (dismissed):
Nicely done. :-)

ares:
seems like tests needs more love, I'll look at what I changed in last version

test/helpers/filters_helper_test.rb
```diff
@@ -37,6 +36,10 @@ def test_no_deconstantization_if_there_is_no_namespace
 
   private
 
+  def resource_path(klass)
+    klass == 'Host' ? 'hosts_path' : nil
```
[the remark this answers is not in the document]
ares: lol what? :-) /me gives up fighting rubocop rules

ares:
[test foreman]

ares:
@timogoebel mind to take a look when you have a chance? thanks!

timogoebel (approved):
LGTM.

timogoebel:
Thanks, @ares.

PR #5527, opened 2018-05-03
Fixes #23492 - symbolize keys when creating setting

mmoll:

### DIFF
--- a/app/models/setting.rb
+++ b/app/models/setting.rb
@@ -201,6 +201,7 @@ class Setting < ApplicationRecord
   def self.create(opts)
     # self.name can be set by default scope, e.g. from first_or_create use
     opts = { :name => new.name } if opts.nil?
+    opts.symbolize_keys!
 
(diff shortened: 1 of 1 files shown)


app/models/setting.rb
```diff
@@ -201,6 +201,7 @@ def parse_string_value(val)
   def self.create(opts)
     # self.name can be set by default scope, e.g. from first_or_create use
     opts = { :name => new.name } if opts.nil?
+    opts.symbolize_keys!
```
ik5, line 204: nitpicking, why not do it like so:

```
if opts.nil? 
  opts =  { :name => new.name } 
else
   opts.symbolize_keys!
end
```

When you create the new hash, you already know that it has symbols for name, so it just shout in my head due to OCD :D

ik5 (commented):
a small nitpitck

mmoll:
9c853eeccc879756a2eea03c667762702e83cbc8 did include this

PR #5530, opened 2018-05-04
Fixes #23511 - Generate webpack plugins in context of vendor/bundle

ehelms:
Previously, the plugin bundles were being compiled as if the vendor and bundle Javascripts were not present thus bundling in more than was needed. This can lead to out of order, or double up loading of library assets such as jQuery breaking plugin expectations that different from development. This generates the plugin assets in relation to those bundles. The net downside is a bundle-.js and vendor-.js will be generated for each plugin. The code that loads the plugin manifests is updated to ensure only the plugins entry into the manifest is loaded.

### DIFF
--- a/config/initializers/assets.rb
+++ b/config/initializers/assets.rb
@@ -118,7 +118,7 @@ Foreman::Application.configure do |app|
             Rails.logger.debug { "Loading #{plugin.id} webpack asset manifest from #{manifest_path}" }
             assets = JSON.parse(File.read(manifest_path))
 
-            webpack_manifest['assetsByChunkName'] = webpack_manifest['assetsByChunkName'].merge(assets['assetsByChunkName'])
+            webpack_manifest['assetsByChunkName'][plugin.id.to_s] = assets['assetsByChunkName'][plugin.id.to_s] if assets['assetsByChunkName'].key?(plugin.id.to_s)
           end
         end
 
--- a/script/plugin_webpack_directories.rb
+++ b/script/plugin_webpack_directories.rb
@@ -17,6 +17,7 @@ specs.each do |dep|
   # skip other rails engines that are not plugins
   # TODO: Consider using the plugin registration api?
   next unless dep.name =~ plugin_name_regexp
+  next if dep.name.include?('_core')
   dep = dep.to_spec if gemfile_in
 
   path = "#{dep.to_spec.full_gem_path}/webpack"
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -34,25 +34,25 @@ module.exports = env => {
   if (env && env.pluginName !== undefined) {
     var pluginEntries = {};
     pluginEntries[env.pluginName] = plugins['entries'][env.pluginName];
-    var entry = pluginEntries;
     var outputPath = path.join(plugins['plugins'][env.pluginName]['root'], 'public', 'webpack');
     var jsFilename = production ? env.pluginName + '/[name]-[chunkhash].js' : env.pluginName + '/[name].js';
     var cssFilename = production ? env.pluginName + '/[name]-[chunkhash].css' : env.pluginName + '/[name].css';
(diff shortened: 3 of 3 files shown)


config/initializers/assets.rb
```diff
@@ -118,7 +118,7 @@ class << self
             Rails.logger.debug { "Loading #{plugin.id} webpack asset manifest from #{manifest_path}" }
             assets = JSON.parse(File.read(manifest_path))
 
-            webpack_manifest['assetsByChunkName'] = webpack_manifest['assetsByChunkName'].merge(assets['assetsByChunkName'])
+            webpack_manifest['assetsByChunkName'][plugin.id.to_s] = assets['assetsByChunkName'][plugin.id.to_s]
```
ekohl: Probably good to check if `plugin.id.to_s` is present in `assets`

ekohl (commented):
This makes a lot of sense to me.

ares:
tested with foreman_ansible and rex, with this PR, `rake webpack:compile` only produces js files for foreman_ansible_core and rex_core, without the PR I see js files for foreman_ansible and rex (not for *-core gems)

ekohl:
@ehelms Fix -> Fixes in the commit message.

script/plugin_webpack_directories.rb
```diff
@@ -17,6 +17,7 @@
   # skip other rails engines that are not plugins
   # TODO: Consider using the plugin registration api?
   next unless dep.name =~ plugin_name_regexp
+  next if dep.name.include?('_core')
```
ehelms, line 20: @ares I added in the ugly hack to address double gemspec plugins.

ohadlevy:
@sharvit can you please verify this? Thanks

ehelms:
[test foreman]

ehelms:
[test katello]

ehelms:
[test upgrade]

ekohl:
[test katello]

ehelms:
Any updates or review? This is a major blocker to a few major features.

sharvit (commented):
We do need the vendor entry to compile plugins, unless, all the vendor chunks the plugin uses will end in the plugin bundle.

 core vendor and plugin vendors have the same md5hash :+1: 

I think that for plugins, we can avoid creating the bundle entry.
```js
if (env && env.pluginName !== undefined) {
  var entry = Object.assign(
    {
      vendor: vendorEntry,
    },
    pluginEntries
  );
} else {
  var entry = Object.assign(
    {
      bundle: bundleEntry,
      vendor: vendorEntry,
    },
    pluginEntries
  );
}
```

ehelms:
Thanks for the review @sharvit, I attempted it without both the bundle and vendor and from all of my testing both are required in order to generate the plugin's webpack assets correctly. I assume this is due in part to plugins requiring bits from the core bundle.

dLobatog (approved):
@ehelms :+1: from my side, seems to work with `RAILS_ENV=production`

dLobatog:
Still unable to make it work with foreman_ansible reports page, but I think this is unrelated to the changes here (it didn't work before). I can see how the new plugin chunks don't contain the Foreman assets. I checked both by looking at the size, and by running `assets:precompile` and `RAILS_ENV=production passenger start` - this failed before this patch, but not after.

Merging and I'll see how to fix `foreman_ansible` reports page, thanks @ehelms

ohadlevy:
TBH: I fail to understand why this patch is required to begin with and it feels like we did a complete cycle, please review my assumptions below:

1. when building webpack with plugins (e.g. on stable / develop branches) everything works as expected.
2. in order to build a plugin assets, we added the env variable to accept a plugin we would like to build
3. we realized we need the whole webpack output for foreman in order to build a correct output for plugins (this pr).

how about we scratch 2 and 3, and just extract the relevant bits (based on generated manifest.json) and include that in the plugins? I think this has the benefit of simplifying the webpack / plugin configuration and also have the build infrastructure closer to development environments?

PR #5535, opened 2018-05-06
[WIP] Fixes 23010 - adding deprecation warnings to flot

lizagilman:
Added deprecation warnings to old flot functions, as we are moving to c3.

### DIFF
--- a/app/assets/javascripts/charts.js
+++ b/app/assets/javascripts/charts.js
@@ -1,4 +1,5 @@
 $.fn.flot_pie = function(){
+  tfm.tools.deprecate('flot', 'c3');
   var options = arguments[0] || {};
   $(this).each(function(i,el){
     var label_exists = false;
@@ -239,6 +240,7 @@ function reset_zoom(item){
 }
 
 function showTooltip(pos, item, formater) {
+  tfm.tools.deprecate('flot', 'c3');
   var content = formater(item);
   $('<div id="flot-tooltip">').text(content).css({
     top: pos.pageY - 40,
@@ -291,7 +293,7 @@ function search_on_click(event, item) {
 }
 
 function get_pie_chart(div, url) {
-
+  tfm.tools.deprecate('flot', 'c3');
   if($("#"+div).length == 0)
   {
     $('body').append('<div id="' + div + '" class="modal fade"><div class="modal-dialog"><div class="modal-content"></div></div></div>');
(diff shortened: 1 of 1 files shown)


boaz0:
Hey @lizagilman 
Sorry to inform you but I am actually working on this in https://github.com/theforeman/foreman/pull/5454

lizagilman:
Hey @ripcurld 
sorry ohadlevy asked me to work on that and I didn't notice your pr
perhaps you can assign yourself on this bug 
@tbrisker should we close this pr?

tbrisker:
It's strange that the bot didn't automatically assign the issue to @ripcurld when he opened a PR.
However, looking at the two PRs, this one uses the deprecation method we use to deprecate functions, while the other adds comments in the code that state a function is deprecated which is not visible to anyone not reading the code. The idea of using a deprecation method is that plugin authors will get an indication that their code needs updating before a function is removed.
Both PRs in my opinion are lacking one very significant issue - we cannot deprecate a function if we don't have a working alternative for it. Right now we only have a new implementation for pie charts afaik, so both of these PRs should not be merged until we have a new implementation for all charts.
I suggest you combine forces and work together to make sure that all deprecated methods have a good replacement, and then deprecate them in the supported way (using `tfm.tools.deprecate`).

lizagilman:
afaik once #5019 will get merged we are done moving to c3 (except of plugins)
possibly @amirfefer can confirm that?

amirfefer:
afaik we are still using `flot_pie_chart` in some places such as `config_report/metrics`

boaz0:
I am closing https://github.com/theforeman/foreman/pull/5454 in favor to this PR.

lizagilman:
Patternfly-react has a verity of chart components implemented - https://rawgit.com/patternfly/patternfly-react/gh-pages/index.html?selectedKind=Chart&selectedStory=Area%20Charts&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs
Is this a satisfying replacement for plugin maintainers?
@tbrisker @amirfefer

tbrisker:
Right now, the only chart type we have integration with in Foreman core is donut chart. 
While it is true that plugin authors can use any other chart type, it isn't something we provide in core, unlike the flot charts. 
There are also flot-based time and bar charts used in core, and there may be others in various plugins. I did not check which are being used in the public plugins, and in any case we should use the deprecation so that even private plugins have time to adapt to this change. This is especially important since this isn't just a function name change, this requires a much more significant rewrite effort for the charts to work with c3.

lizagilman:
closing this pr as we still have some work until full migration is completed

PR #5537, opened 2018-05-07
Fixes #23248 - disable breadcrumb switcher in compute attributes page

amirfefer:
The breadcrumb switcher doesn't function as intended, and may link to wrong pages.
therefore it would be better to turn the switcher off until a fix.
a RM [issue](http://projects.theforeman.org/issues/23525) has been opened

### DIFF
--- a/app/views/compute_attributes/edit.html.erb
+++ b/app/views/compute_attributes/edit.html.erb
@@ -16,8 +16,7 @@
         caption: _('Edit %s' % @set.compute_profile.to_label)
       }
     ],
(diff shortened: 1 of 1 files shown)


ohadlevy (approved):

ares:
the redmine issue has been marked for 1.18 without any objections, I've opened CP PR at https://github.com/theforeman/foreman/pull/5593

ideally please do so just after the PR gets into develop, the sooner we get it in, the lower the risk of conflicts/test failures is

PR #5539, opened 2018-05-07
Fixes #20778 - Fix redirect parameters on hosts page

dmitri-d:

### DIFF
--- a/app/assets/javascripts/host_checkbox.js
+++ b/app/assets/javascripts/host_checkbox.js
@@ -156,7 +156,7 @@ $(function() {
 function submit_modal_form() {
   if (!$('#keep_selected').is(':checked'))
     removeForemanHostsCookie();
-  if ($("#multiple-alert").data('multiple')){
+  if (is_multiple()){
     var query = $("<input>")
       .attr("type", "hidden")
       .attr("name", "search").val($("#search").val());
@@ -166,12 +166,16 @@ function submit_modal_form() {
   $('#confirmation-modal').modal('hide');
 }
 
+function is_multiple() {
+  return $("#multiple-alert").data('multiple');
+}
+
+function get_bulk_param() {
+  return is_multiple() ? {search: $("#search").val()} : {host_ids: $.foremanSelectedHosts}
+}
+
 function build_modal(element, url) {
-  var is_multiple = $("#multiple-alert");
-  if (is_multiple.data('multiple'))
-    var data = {search: $("#search").val()};
-  else
-    var data = {host_ids: $.foremanSelectedHosts};
+  var data = get_bulk_param();
   var title = $(element).attr('data-dialog-title');
   $('#confirmation-modal .modal-header h4').text(title);
   $('#confirmation-modal .modal-body').empty()
@@ -181,7 +185,7 @@ function build_modal(element, url) {
     function(response, status, xhr) {
       $("#loading").hide();
       $('#submit_multiple').val('');
-      if (is_multiple.data('multiple'))
+      if (is_multiple())
         $("#multiple-modal-alert").show();
       var b = $("#confirmation-modal .btn-primary");
       if ($(response).find('#content form select').length > 0)
@@ -193,7 +197,8 @@ function build_modal(element, url) {
 }
 
 function build_redirect(url) {
-  var url = url + "?" + $.param({host_ids: $.foremanSelectedHosts});
+  var data = get_bulk_param();
+  var url = url + "?" + $.param(data);
   window.location.replace(url);
 }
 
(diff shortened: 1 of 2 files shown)


ohadlevy:
@amirfefer can you please review?

app/assets/javascripts/host_checkbox.js
```diff
@@ -193,7 +193,12 @@ function build_modal(element, url) {
 }
 
 function build_redirect(url) {
-  var url = url + "?" + $.param({host_ids: $.foremanSelectedHosts});
+  var is_multiple = $("#multiple-alert");
```
amirfefer: the function `build_modal` does the same, 
so what do you think about extract this logic ?
something like this:
```javascript
function is_multiple() {
  return $("#multiple-alert").data(('multiple'));
}
function get_bulk_param() {
  return is_multiple() ? {search: $("#search").val()} : {host_ids: $.foremanSelectedHosts}
}
```

amirfefer (changes requested):
Thanks @witlessbird !
I left one inline comment.
could you add an integration test for it ?

dmitri-d:
* extracted search/host_ids logic into a method

> could you add an integration test for it ?

How would you suggest to test this? I didn't see anything on the stock hosts page that uses redirects.

app/assets/javascripts/host_checkbox.js
```diff
@@ -166,12 +166,16 @@ function submit_modal_form() {
   $('#confirmation-modal').modal('hide');
 }
 
+function is_multiple() {
+    return $("#multiple-alert").data(('multiple'));
```
amirfefer: should be 2 space indentation

app/assets/javascripts/host_checkbox.js
```diff
@@ -166,12 +166,16 @@ function submit_modal_form() {
...
+}
+
+function get_bulk_param() {
+    return is_multiple() ? {search: $("#search").val()} : {host_ids: $.foremanSelectedHosts}
```
amirfefer: ditto

app/assets/javascripts/host_checkbox.js
```diff
@@ -166,12 +166,16 @@ function submit_modal_form() {
...
+}
+
 function build_modal(element, url) {
-  var is_multiple = $("#multiple-alert");
```
amirfefer: this function uses `is_multiple` twice (line 188)

app/assets/javascripts/host_checkbox.js
```diff
@@ -166,12 +166,16 @@ function submit_modal_form() {
   $('#confirmation-modal').modal('hide');
 }
 
+function is_multiple() {
```
amirfefer, line 169: nitpick: could you reuse this in `submit_modal_form` too?

amirfefer (changes requested):
Thanks !
you can use `assert_current_path` in order to test redirect behavior.

the modal opening is tested (under -  `host_js_test' / `hosts index multiple actions` ), while the redirection isn't.

pondrejk (approved):
ack pending previous commits

dmitri-d:
> you can use assert_current_path in order to test redirect behavior.

I'm not sure how to trigger redirect behaviour -- nothing on hosts page uses it.

dmitri-d:
* fixed issues pointed out in feedback

dmitri-d:
* added an integration test

test/integration/host_js_test.rb
```diff
@@ -340,6 +340,21 @@ class HostJSTest < IntegrationTestWithJavascript
...
+      page.find('#check_all').trigger('click')
+
+      # Ensure and wait for all hosts to be checked, and that no unchecked hosts remain
+      assert page.has_no_selector?('input.host_select_boxes:not(:checked)')
+
+      # Hosts are added to cookie
+      host_ids_on_cookie = JSON.parse(CGI.unescape(page.driver.cookies['_ForemanSelectedhosts'].value))
+      assert(host_ids_on_cookie.include?(@host.id))
+
+      page.execute_script("build_redirect('#{select_multiple_environment_hosts_path}')")
```
amirfefer, line 355: what do you think about  stubbing `multiple_actions` to return something like `[[_('a redirect action'), some_path, false]]`, and then the `build_redirect` should be called, instead of executing `build_redirect` directly ?

dmitri-d:
> what do you think about stubbing multiple_actions to return something like [[_('a redirect action'), some_path, false]], and then the build_redirect should be called, instead of executing build_redirect directly ?

I don't think stubbing/mocking makes sense in an integration test: the goal is to exercise real components of a system. Adding a stub like you suggested only complicates the test without adding making it more useful.

amirfefer (approved):
LGTM

iNecas:
Before we merge, please create a separate issue against foreman and link the commit to it, it makes it easier to keep the changes and dependencies under control.

dmitri-d:
* updated commit message with a dedicated issue number

iNecas (approved):

iNecas:
Tested and works well. Thanks @witlessbird

iNecas:
proposing to 1.18

PR #5541, opened 2018-05-09
Fixes #23549 - add axios support for wait_for_ajax

amirfefer:
wait_for_ajax helper method works only with jQuery.ajax thanks to jQuery.active counter.
We added axios to handle ajax requests, so wait_for_ajax should work with axios as well.


tbrisker (approved):
Nice one, thanks @amirfefer !

